PR #2092, opened 2015-03-25
Improve packaging, bundling and module boilerplate

### LINKED ISSUE #2026: Sidebar won't open on the Guardian
Issue on the Guardian deploying the chrome extension.  no sidebar appears.  
"uncaught error: duplicate define: jquery"

http://www.theguardian.com/environment/2015/mar/10/working-towards-practical-solutions

screen shot 2015-03-12 at 6 04 43 am

tilgovi:
Prevent negative interactions with module systems present in injected
pages and refactor browserify interactions with angular to increase
CommonJS accessibility / legibility and isolate angular module system
boilerplate.

- Avoid issues due to mis-detected module environments

  - In the injection script, load only Annotator and jQuery in the
    global scope and then both removed with `noConflict()`. Both are
    loaded with browserify-shim, which prevents jQuery from registering
    itself against require.js, if present (fix #2026).

  - Be explicit about registration of common cross-frame components.

    - Lift angular glue into app.coffee. Individual components are
      exported as CommonJS modules.

    - Lift Annotator glue into bootstrap.js (renamed hypothesis.js).

- Break the helpers bundle up and create services and directives.

  - formHelpers -> {formInput, formValidate, formRespond}
  - stringHelpers -> unicode
  - tagHelpers -> tags
  - timeHelpers -> time
  - uiHelpers -> {tabbable, tabReveal}

- Rename all services, dropping the -service suffix, to match their
  injected names. This is consistent with how directives are being
  treated, and controllers as well (controllers are registered and
  injected done with the "Controller" suffix).

- Move Annotator subclasses into the annotator subdirectory. This
  change avoids having `Annotator.Host` clash with the `host`
  service and makes it easy to glob dependencies in assets.yaml to
  minimally rebuild either the app or the inject.

- Split up directives.coffee, services.coffee, filters.coffee, and
  controllers.coffee.


tilgovi:
You don't want to look at this too closely. Your eyes may bleed from all the diff noise. You want to read the summary, then open app.coffee, and you want to be happy, and you want to hit the green button.

tilgovi:
static/directives.coffee -> static/directive/*.coffee
static/filters.coffee -> static/filter/*.js (yes, these are JavaScript, let the conversion begin)

tilgovi:
The most important part, the bug I was trying to fix, had to do with not having the `if angular? ... else Annotator` module registrations for the common cross-frame stuff.

Never one to do something half-way and wanting so badly to see the beautiful app.coffee the way it is now, I just moved All The Things.

tilgovi:
I will personally help rebase the PRs of anyone who says "+1" here in speedy fashion. :)

nickstenning:
Holy shit! You only went and did it. I'll help review this tomorrow.

csillag:
> You don't want to look at this too closely. Your eyes may bleed from all the diff noise.

I agree, up to this point.

> You want to read the summary, then open app.coffee, and you want to be happy, and you want to hit the green button.

Now that you mention, personally, I want to hit the gray button (with the label saying *Close and comment*), and ask you to resubmit this in, like, 5 different PRs, because it's nigh impossible to meaningfully review a work of this size. (Or at least it takes exponentially more time than reviewing it multiple steps.)

What does commonJS and Angular boilerplate code has to do with the Coffee -> JS conversion? Is there any reason to attempt to do it in one step?

tilgovi:
There are no functionality changes.

The only coffee -> js conversation was the filters. I was breaking them up into new files anyway and they are _so tiny_ that it was next to zero effort to convert them. Since this is breaking up one file into many anyway, the diff would be no less noisy if I had kept it coffee. I saved you another noisy review.

If you want to bike shed about project layout or name, that kind of commentary is welcome and I can make small changes before we merge this.

I made changes in a few places to constructor patterns (using the revealing module pattern) so that everything that isn't a `provider` can be a `service` (no `factory`) and most services are instantiated the same way. I can revert some of that _if you have a good reason_.

This doesn't have to all happen at once, but making it happen in pieces is more work for me and the benefit is unclear.

csillag:
Imho, considering the depth of our current testing (or the lack thereof), merging a giant PR like this days before inviting loads of users, is a pretty reliably recipe for embarrassing ourselves by introducing loads of new errors.

> making it happen in pieces is more work for me and the benefit is unclear.

The benefit of doing it in pieces would be that we could keep the software fully functional and disruption-free that way, whereas if we merge it in one piece, it's almost *guaranteed* to break a few things. Sure, we can repair those later, but I would prefer not breaking them, or at least not right now - which is perfectly possible, if we keep a few simple rules. (One of which is, no monster PRs.)

tilgovi:
> The benefit of doing it in pieces would be that we could keep the software fully functional and disruption-free that way, whereas if we merge it in one piece, it's almost guaranteed to break a few things. 

How about you check it out and test it? There are 469 unit tests that pass. I've tested most of application myself (replies, page search, search, tags, tag autocompletion, auth), etc.

Really, there are no behavior changes. I didn't make any serious refactorings of anything. This is just shuffling boilerplate around. It also fixed _two_ blocking bugs.

tilgovi:
I understand we're trying to get this thing out to people right now but:

1) This fixes two pretty crucial bugs that block that.
2) People looking at our product means people looking at our code and this makes it a whole lot more consistent.

What I would really like from you is your _help_ checking that this doesn't somehow introduce regressions. I hear you that this could be easier in small pieces, but really it's not that big a deal. None of it's designed to change any behavior or code other than boilerplate and formatting. All of those diffs would be noisy. I actually thought it was saving everyone time (review, rebase, etc) to do it all at once.

csillag:
> What I would really like from you is your _help_ checking that this doesn't somehow introduce regressions. 

I am *trying to help* to avoid those regressions by reminding you that a PR of this size almost certainly *does* introduce them, despite our best efforts. I *will* click around, but that doesn't mean that we will discover all the problems.

> None of it's designed to change any behavior or code other than boilerplate and formatting.

Up to this point, we have introduced ~98% of our regressions with code changes that were not designed to impact those behaviors in any way, but they still did - so this hardly means anything.

csillag:
> static/filters.coffee -> static/filter/*.js (yes, these are JavaScript, let the conversion begin)

Do we have the infrastructure (ie. [Babel](http://babeljs.io/)) in place that would allow us to use EcmaScript 6, if we wanted to? It doesn't seem to be...

csillag:
> Do we have the infrastructure (ie. [Babel](http://babeljs.io/)) in place that would allow us to use EcmaScript 6, if we wanted to? It doesn't seem to be...

I have opened https://github.com/hypothesis/h/issues/2093 to track that.

gergely-ujvari:
I really want this PR to be split in 3-4 different PRs.
We have:
- directory restructuring
- service (and service functions) renaming
- file splitting
- language converting
- exports structure changing
- random modifications in a one and huge PR.
The changes are untrackable.

tilgovi:
Splitting this is a waste of my time. I'm giving you all until I wake up
tomorrow to suggest changes and then I'm hitting the button myself.

gergely-ujvari:
... so PR reviewing is a waste of time?  Really?

csillag:
> I'm giving you all until I wake up tomorrow to suggest changes

I already suggested changes: don't do it this way.

> [...] and then I'm hitting the button myself.

Doing so would make our whole PR and code review system worthless.
(I thought we have all agreed that nobody is supposed to merge his own PRs.)

gergely-ujvari:
For clarification: I want this PR to be split because it is theoretically impossible to verify if any functionality has been altered/broken/etc.

Yes, I've seen your comments that it isn't, but that is just not true.
For example in the annotation directive you've changed the way we store tags into the localStore. Now with your changes (and I'm not a fan of that) if we save an edited annotation, then we would save the tags which weren't newly added, again. And again in every edit. (Originally I prevented that, but it was just modified without a single comment)

I know this because I've worked on that, and accidentally spotted the change, but if that happened, a dozen more "hidden" change could have been introduced, but the huge diff just hides them.

How should we find this out, or verify it? Our test coverage is growing, but it is still incomplete.

Let's introduce this set of huge (and good) changes incrementally.

About self-merging: absolutely no!
And given the size of the PR forcefully giving 1 day to make suggestions is just unfair and unrealistic. (Anyway I gave my suggestions)

tilgovi:
Please suggest changes to the substance of the request, if you have any.

Process is in place as a best practice. We are voluntarily bound to it. To suggest that it wastes my time to split this up is not to suggest that code review is worthless, nor that our process is deficient or inapplicable generally. Such reactions are hyperbolic and derailing.

I'm bending this process to save us time, in the form of more large reviews to the same lines of code. I make it available for feedback as a sanity check, but also just as an "FYI" out of courtesy.

In the end, I'm exercising my prerogative as a lead maintainer to direct the "editorial voice" in the code and my best judgment to determine that doing it all at once is the least disruptive path. I don't require your approval for this, but I solicit it for your awareness, education, and comfort.

I put in about 14hrs yesterday making purely stylistic changes to bring this level of consistency to the code. That's not the kind of weeding I really want to do more of right now.

Please think again whether you really want to hit me over the head with process or if you would rather have a beautiful code base and move on.

When I merge this, if no one else does, you have a choice in how you feel about it.

tilgovi:
Good catch, @gergely-ujvari. I'll add that back. That should have been separate.

Thank you for the review.

csillag:
I am saying that the process is there for a very good reason: to help us to avoid shooting ourselves in the foot. Or in the head. Sure, it's voluntary, but the bullet will still hurt all the same, when you chose to ignore the safety rules.

csillag:
> Good catch, @gergely-ujvari. I'll add that back. That should have been separate.
> 
> Thank you for the review.

That finding was *not* his review; it his example showing why a review would be necessary, *while impossible in the current form*.

tilgovi:
And by "separate" I mean, of course, plain wrong. I'll make a regression test for that as a separate commit and take that regression out of here.

I'm not aware of anywhere else in this PR that I've even attempted to change implementation like that.

tilgovi:
Fixed the tags slip-up. I went back through the whole set of changes and the only other thing I think I did was change `viewFilter`'s checkers to `fields`. That, thankfully, has tests.

tilgovi:
Well, a couple other renaming's but nothing that wasn't "in place".

tilgovi:
I was kinda awful on this thread. I'm sorry.

It's not my desire to be dismissive of anyone else's role in shaping this.

I do want to give you these changes, which really do a lot of nice things to the packaging and organization of everything on the front end.

Please take you're time, please glance through it. Large blocks of red and green are likely unchanged at all. Careful attention was and can be paid to places where I phased out constructors.

I'm not super concerned about the size, but understand yours. I apologize for it, but I didn't conceive this all in advance, it sort of just unravelled in a furious binge.

Please don't make me comb it into nice pieces, it's such tedious stuff to begin with.

gergely-ujvari:
The stream page in this is branch is throwing an error (I think a service name wasn't renamed here)
So go to localhost:5000/stream

and for me the result is in the console.log:
```
"Error: [$injector:unpr] Unknown provider: queryParserProvider <- queryParser
http://errors.angularjs.org/1.2.25/$injector/unpr?p0=queryParserProvider%20%3C-%20queryParser
minErr/<@http://0.0.0.0:5000/assets/scripts/vendor/angular.js:78:5
...
```

gergely-ujvari:
Standalone page also throws a similar exception to me (should be trivial to fix):
```
"Error: [$injector:unpr] Unknown provider: streamfilterProvider <- streamfilter
http://errors.angularjs.org/1.2.25/$injector/unpr?p0=streamfilterProvider%20%3C-<main class="content ng-scope" ng-view="">treamfilter
minErr/<@http://localhost:5000/assets/scripts/vendor/angular.js:78:5
```

nickstenning:
Rebased to take account of #2100, and two commits added to fix the issues identified by @gergely-ujvari.

h/static/scripts/annotation-controller.coffee
```diff
@@ -0,0 +1,38 @@
+angular = require('angular')
+
+
+module.exports = class AnnotationController
```
nickstenning: I don't think renaming this to AnnotationController is a great idea. We now have two distinct objects called AnnotationController. Let's keep this as AnnotationViewerController for now, even if these two things need unifying later.

h/static/scripts/annotator/guest.coffee
```diff
@@ -1,15 +1,7 @@
-$ = require('jquery')
 Annotator = require('annotator')
-Channel = require('jschannel')
+$ = Annotator.$
 
-require('diff-match-patch')
```
nickstenning: I'm not sure I see the logic of moving these into `hypothesis.js`. Doesn't it make most sense to keep these requirements as close to the code that actually uses them as possible?

h/static/scripts/app.coffee
```diff
@@ -93,50 +76,81 @@ setupStreamer = [
...
-require('./annotation-ui-sync')
-require('./bridge')
-require('./discovery')
+.controller('AppController', require('./app-controller'))
+.controller('AnnotationController', require('./annotation-controller'))
+.controller('AnnotationUIController', require('./annotation-ui-controller'))
+.controller('StreamController', require('./stream-controller'))
+.controller('WidgetController', require('./widget-controller'))
+
+.directive('annotation', require('./directive/annotation'))
```
nickstenning: Good work on getting all the directives consistent. I'll note that I've double checked we've got the same set of directives before and after :)

h/static/scripts/filter/persona.js
```diff
@@ -0,0 +1,14 @@
+module.exports = function () {
+  return function (user, part) {
+    if (typeof(part) === 'undefined') {
+      part = 'username';
+    }
+    var index = ['term', 'username', 'provider'].indexOf(part);
+    var groups = user ? user.match(/^acct:([^@]+)@(.+)/) : null;
```
nickstenning: I'm not entirely sure whether it matters, but you have changed the behaviour here. `foo = user?.bar()` in CoffeeScript translates into `var foo = typeof user !== "undefined" && user !== null ? user.bar() : undefined;`.

I'd feel more comfortable with that.

h/static/scripts/filter/persona.js
```diff
@@ -0,0 +1,14 @@
...
+    if (groups) {
+      return groups[index];
+    } else if (part != 'provider') {
```
nickstenning: `!==`

h/static/scripts/app.coffee
```diff
@@ -93,50 +76,81 @@ setupStreamer = [
...
+.service('threading', require('./threading'))
+.service('unicode', require('./unicode'))
+.service('viewFilter', require('./view-filter'))
+
+.value('xsrf', token: null)
+.value('AnnotationSync', require('./annotation-sync'))
+.value('AnnotationUISync', require('./annotation-ui-sync'))
+.value('Discovery', require('./discovery'))
+
+.run(setupCrossFrame)
```
nickstenning: Just double checking: this is no longer conditional on being in the test environment because the tests no longer load the tested services through this angular module.

h/static/scripts/auth.coffee
```diff
@@ -70,11 +70,9 @@ class Auth
...
-        @user = null
+      if auth.user is undefined and not _checkingToken
+        auth.user = null
 
     identity.watch {onlogin, onlogout, onready}
-
-
-angular.module('h')
-.service('auth', Auth)
+    auth
```
nickstenning: Could you make this return explicit? I can't stand CoffeeScript's implicit return. It's so easy to miss this line but the entire service will be screwed if it's missing.

h/static/scripts/directive/form-input.coffee
```diff
@@ -0,0 +1,33 @@
+module.exports = ->
+  link: (scope, elem, attr, [model, validator]) ->
+    return unless  model
```
nickstenning: This has also changed. The changes here look fine to me but I would really suggest you resist the urge to fix all the things at once in future. It's *so* much easier to review the pieces where the only thing that's changed is `module.exports` at the top and `angular.module` at the bottom.

nickstenning: Are you sure that nothing relies on the fact that this link function was previously a no-op if no validator was provided?

nickstenning: The more I think about this, the more I'd be much happier if you could back out these second-order changes.

h/static/scripts/host.coffee
```diff
@@ -1,138 +1,31 @@
...
+        break
+
+    channelListeners =
+      back: -> host.hideSidebar()
+      open: -> host.showSidebar()
+
+    for own channel, listener of channelListeners
+      bridge.on(channel, listener)
+
+    host
```
nickstenning: Ditto wrt explicit return.

h/static/scripts/tags.coffee
```diff
@@ -14,24 +14,24 @@ createTagHelpers = ['localstorage', (localstorage) ->
...
   # timestamp for every tags used.
-  storeTags: (tags) ->
-    savedTags = localstorage.getObject TAGS_MAP_KEY
+  store: (tags) ->
+    savedTags = localStorage.getObject TAGS_MAP_KEY
     savedTags ?= {}
 
     for tag in tags
-      if savedTags[tag.text]?
+      if savedTags[tag]?
```
nickstenning: It isn't immediately obvious to me that this is safe change to make. Are tags just objects? If so are there object identity issues associated with keying savedTags by objects rather than strings?

h/static/scripts/tags.coffee
```diff
@@ -14,24 +14,24 @@ createTagHelpers = ['localstorage', (localstorage) ->
 
   # Add newly added tags from an annotation to the stored ones and refresh
   # timestamp for every tags used.
-  storeTags: (tags) ->
-    savedTags = localstorage.getObject TAGS_MAP_KEY
+  store: (tags) ->
+    savedTags = localStorage.getObject TAGS_MAP_KEY
```
nickstenning: You've changed the format of `savedTags`, but you don't have any code that handles backwards compatibility with the previous format.

h/static/scripts/tags.coffee
nickstenning: Ahhhh, sorry. I'd misread. Tags are guaranteed to just be strings now. That makes more sense. My point about backwards compatibility still stands, however.

nickstenning:
So, I've been through this with a fine toothcomb and I have the following observations:

- Contrary to what GitHub says, very little has changed. The fact that the diff tools are pretty poor at identifying what moved where doesn't itself mean that this is a huge change. Checking the branch out and using `git diff -C` made reviewing a much more manageable task.

- That said, the bits that were changed that weren't directly related to improving module isolation did account for about 50% of the review time. It's probably wise to avoid the temptation when making changesets like this.

- I have comments, mostly on pieces that aren't about module encapsulation. In particular the changes to `form-input.coffee` and `tags.coffee` seem like they could be dropped, giving greater confidence in the safety of the changes and without undoing a lot of hard work.

- There's a behaviour change in the persona filter that needs careful review. For the sake of simplicity replacing with code that is in exact parity with CoffeeScript is probably simplest.

- And finally, a nitpick: Annotator is 4-space-indented JavaScript. I'd really like it if we did the same here.

nickstenning:
Lastly it would make me happier if you could account for 3 missing tests :)

master: `PhantomJS 1.9.8 (Mac OS X): Executed 470 of 470 SUCCESS (2.844 secs / 1.659 secs)`

this branch: `PhantomJS 1.9.8 (Mac OS X): Executed 467 of 467 SUCCESS (3.36 secs / 1.59 secs)`

tilgovi:
The tests are related to the validator changes in formValidate and
formInput. Tests were testing the interaction between the two rather than
each as a unit.

h/static/scripts/annotator/guest.coffee
tilgovi: The guest can function without. The plugins themselves expect these in the
global scope. The entry point module seemed like the only sensible place.

h/static/scripts/app.coffee
tilgovi: As far as I know. Seems to work!

h/static/scripts/tags.coffee
tilgovi: I'll double check this, because I didn't think I changed anything that
would be incompatible. On the other hand, I'm not sure we've ever even
shipped this in production yet.

tilgovi:
Thanks _so_ much for the detailed review, @nickstenning.

The tags stuff was a definite slip-up. I've restored all that. Everything else you asked I addressed.

Here's what happened with the missing tests:

- In annotation-test, we had a test that asserted '$render' was called on the first digest. That's no longer testable because we're not exporting the controller at all so we can't instantiate it within instantiating the directive and triggering a digest. So our test can't get a handle to the controller until after the first digest. "No problem", I thought, the next test asserts that '$render' is called on every digest so there's no reason to specifically assert it happens on the first.

- The other missing test was for the form helpers, "should apply an error class if the form receives errors after a submit action". This is the test that was confused about its purpose. The form-input directive doesn't deal with the submit action, it deals with maintaining a custom error class that's conditional upon the field being dirty and resetting the 'response' validation key whenever the model changes. The form-validate directive triggers validation and rendering of controls on submit. So this test was testing the interaction between the two. This test is trying to be too clever about treating "response" as a special kind of error, but really it's like any other validation key and the form-input directive already thoroughly tests that invalid fields trigger the right behavior. Furthermore, this test was pretty broken anyway, I think, because it was trying to test that the "response" key triggered the error class but it wasn't setting a valid input anyway, so the error class was there regardless.

tilgovi:
Rebased, squashed, ready to go.

h/static/scripts/annotator/guest.coffee
csillag: These dependencies belong to various Annotator plugins. I have browserified those libraries once in the past...  let me look those up.

nickstenning:
I've pushed one small change to this to remove an empty file, and I've added a summary of non-trivial movements to the end of the commit message.

Looking pretty good to me. I'm still not sure about the rename of `AnnotationViewerController` to `AnnotationController`, given that that name already has a meaning in the codebase, but I'm not prepared to block this on that.

Any specific comments from others? If not I'll merge this in a little while and deploy to stage.

tilgovi:
Fine to change that name.

nickstenning:
Done.

tilgovi:
Anything else?

tilgovi:
Lemme know if there's anything else.

tilgovi:
Shoot... Said that twice. I'm putting the phone down until I'm fully awake.

nickstenning:
:fireworks: This is a really great set of changes. Good work.

PR #2100, opened 2015-03-26
Fix broken util.split_user

nickstenning:
The stream view assumed that split_user returned a re.Match object
rather then a tuple of match groups.

This commit makes split_user return None if the passed username is not
splittable, and places the responsibility on the caller to check for
this case.


seanh:
LGTM :+1: , see if tests pass

PR #2101, opened 2015-03-26
Scroll to annotation editor when opening it

### LINKED ISSUE #2053: Sometimes you have to search for the new annotation cards if many are present

Result:

seanh:
This fixes #2053:

- Go to a page that has more than a screenful of
  annotations in the sidebar
- Scroll the sidebar to the very bottom
- Scroll the page to the very top, select some text and create a new annotation
  that will appear above the other annotations in the sidebar

=> The new annotation appears in the sidebar with its edit form open,
but it's off screen. From the user's POV nothing happens.

You can get the same bug by scrolling the sidebar to the top and
annotating the bottom of the page.

This commit fixes the issue by having AnnotationController emit an
"editorOpened" event whenever it opens its editor form, and having
AppController catch this event and scroll the sidebar to the annotation
that emitted the event.

### DIFF
--- a/h/static/scripts/app-controller.coffee
+++ b/h/static/scripts/app-controller.coffee
@@ -87,6 +87,9 @@ module.exports = class AppController
     $rootScope.$on 'beforeAnnotationCreated', ->
       $scope.clearSelection()
 
+    $rootScope.$on 'editorOpened', (annotation) ->
+      $rootScope.scrollTo(annotation)
+
     $scope.login = ->
       $scope.dialog.visible = true
       identity.request {oncancel}
--- a/h/static/scripts/directive/annotation.coffee
+++ b/h/static/scripts/directive/annotation.coffee
@@ -116,6 +116,7 @@ AnnotationController = [
       @action = if model.id? then 'edit' else 'create'
       @editing = true
       @preview = 'no'
+      $rootScope.$emit("editorOpened", @annotation)
 
     ###*
     # @ngdoc method


seanh:
Questions:

- Is an emit() and on() the right approach, here?
- Do we have any guidelines on what parts of the code should and should not emit and catch signals?
- Any guidelines on naming signals? (I'm sure "editorOpened" is not specific enough)

When you have an emit() and on() system like this it can be very convenient: lines of code all over the place emit things, other lines catch them and do things. It's almost like everything function in your code is a global! :) Slightly worried about creating spaghettic code by doing too much of this.

seanh:
Oh, and tests?

tilgovi:
I don't understand how this solves the problem described. It seems as though, as written, this should be causing the page to scroll to the annotation but the issue describes having the _sidebar_ scroll to it.

csillag:
@seanh, the name of your branch (`2052-scroll-to-sign-in-form-on-clicking-sign-in-link`) is *slightly* misleading, since it's describing something completely different.

But more importantly, I get this error (when testing with Chromium):

```
TypeError: undefined is not a function
    at app.min.js:1328
    at Scope.$emit (angular.js:12915)
    at edit (app.min.js:1847)
    at new AnnotationController (app.min.js:2063)
    at invoke (angular.js:3966)
    at Object.instantiate (angular.js:3977)
    at angular.js:7281
    at angular.js:6670
    at forEach (angular.js:332)
    at nodeLinkFn (angular.js:6657)
```

The cause seems to be the `$rootScope` has no `scrollTo()` method.

csillag:
@tilgovi:

> It seems as though, as written, this should be causing the page to scroll to the annotation

I guess you might be confusing `ViewControllers'` `scrollTo()` method (which scrolls the page), and `$rootScope`'s alleged `scrollTo()` method (called within the `AppController`), which doesn't exist for me, but is probably supposed to scroll the sidebar.

seanh:
Hmm, I'm sure this was working when I tested it yesterday :( I'll look again

nickstenning:
Rebased this for you.

seanh:
Closing, replaced with https://github.com/hypothesis/h/pull/2121

PR #2110, opened 2015-03-29
Remove ES test

nickstenning:
Only one test in our codebase actually depends upon a running ElasticSearch server, and it doesn't really need to (and/or is testing the wrong thing).

So, remove it!


nickstenning:
Just ignore the Buildkite build status here.

gergely-ujvari:
LGTM.

PR #2115, opened 2015-03-31
Remove Dockerfile entrypoint

nickstenning:
The Dockerfile entrypoint is a nice trick, but it actually makes running
one-off commands in containers rather confusing if you're not expecting
it.

Before:

    docker run hypothesis/h --skip-runit -- hypothesis token --sub acct:foo@bar.com conf/production.ini

After

    docker run hypothesis/h hypothesis token --sub acct:foo@bar.com conf/production.ini

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -57,6 +57,5 @@ RUN pip install -r requirements.txt
 ADD ./svc /etc/service
 
 # Startup and ports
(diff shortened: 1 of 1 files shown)


nickstenning:
Apologies for the very exciting title. Managed to open a pull request at the exact same moment I spilled water on my laptop.

PR #2116, opened 2015-03-31
Simplify auth-controller-test.coffee

seanh:
- Remove variables $scope, $timeout, auth, session and controller that were
initialized in beforeEach() functions and shared between tests.

- Instead add a factory function (_controller()) that creates and returns
these variables, have each test that needs them call it to get just the
ones it needs.

- Remove the beforeEach() that was $providing a bunch of variables to the
Angular injector, it's not needed anymore.

- We don't need $provide() at all anymore, and we use inject() only to inject
$rootScope and $controller (which I think is unavoidable). This helps to
decouple our tests from Angular and make them easier to follow, without
having to understand the Angular injector.

- Only set the session spy in tests that need it

This decouples the tests from each other by removing shared variables and
shared beforeEach() behavior, making it easier to modify tests or add new
tests without affecting existing tests.

If a test doesn't need a particular variable, fixture or piece of setup code -
it just doesn't call it (also saves CPU cycles).

It also makes the tests easier to read - each test can be more easily
understood by just looking at its own code, without having to understand how
a bunch of global variables are created by Angular providing and injecting.
Each test function declares all the variable names that it uses inside its
own function code.

### DIFF
--- a/h/static/scripts/account/test/auth-controller-test.coffee
+++ b/h/static/scripts/account/test/auth-controller-test.coffee
@@ -18,15 +18,36 @@ class MockSession
     $promise:
       finally: sandbox.stub()
 
-mockFlash = info: sandbox.spy()
-mockFormRespond = sandbox.spy()
-
-describe 'h:AuthController', ->
+# Return an AuthController instance and associated objects for tests to use.
+_controller = (mocksession) ->
+  controller = null
   $scope = null
   $timeout = null
-  auth = null
   session = null
-  $controller = null
+  mockFlash = info: sandbox.spy()
+  mockFormRespond = sandbox.spy()
+  inject(($controller, $rootScope) ->
+    session = mocksession or new MockSession()
+    $scope = $rootScope.$new()
+    $timeout = sandbox.spy()
+    params =
+      $scope: $scope
+      $timeout: $timeout
+      flash: mockFlash
+      session: session
+      formRespond: mockFormRespond
+    controller = $controller('AuthController', params)
+  )
+  return(
+    auth: controller
+    $scope: $scope
+    $timeout: $timeout
+    session: session
+    mockFlash: mockFlash
+    mockFormRespond: mockFormRespond
+  )
+
+describe 'h:AuthController', ->
 
   before ->
     angular.module('h', [])
@@ -35,26 +56,13 @@ describe 'h:AuthController', ->
   beforeEach module('h')
   beforeEach module('h.templates')
 
-  beforeEach module ($provide) ->
-    $provide.value '$timeout', sandbox.spy()
-    $provide.value 'flash', mockFlash
-    $provide.value 'session', new MockSession()
-    $provide.value 'formRespond', mockFormRespond
-    return
-
-  beforeEach inject (_$controller_, $rootScope, _$timeout_, _session_) ->
-    $scope = $rootScope.$new()
-    $timeout = _$timeout_
-    $controller = _$controller_
-    auth = $controller 'AuthController', {$scope}
-    session = _session_
-    sandbox.spy session, 'login'
-
   afterEach ->
     sandbox.restore()
 
   describe '#submit()', ->
     it 'should call session methods on submit', ->
+      {auth, session} = _controller()
+      sandbox.spy(session, 'login')
 
       auth.submit
         $name: 'login'
@@ -64,6 +72,8 @@ describe 'h:AuthController', ->
       assert.called session.login
 
     it 'should do nothing when the form is invalid', ->
+      {auth, session} = _controller()
+      sandbox.spy(session, 'login')
       auth.submit
         $name: 'login'
         $valid: false
@@ -72,6 +82,7 @@ describe 'h:AuthController', ->
       assert.notCalled session.login
 
     it 'should apply validation errors on submit', ->
+      {auth, mockFormRespond} = _controller()
       form =
         $name: 'register'
         $valid: true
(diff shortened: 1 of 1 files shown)


seanh:
What do people think of this? It might just be me hating on Angular because I don't understand it, but in auth-controller-test.coffee I didn't like: having a bunch of "global"/shared variables and beforeEach() fixtures that the tests were all depending on; using $provide and inject for things that could just be simple params passed to a function.

seanh:
I feel like we shouldn't really need to inject `$rootScope.$new()` in these tests either, we should just be able to pass a `{}` for the scope for testing, but we can't because `AuthController` access a lot of properties (and nested properties) and methods of `$scope`, and it doesn't look like it creates any of these properties itself.

Using `$scope` in this way seems like a pretty nasty anti-pattern to me. Into `AuthController` tightly coupled to whatever code is creating all these `$scope` properties like `.account.tab`, `.form`, etc?

Wonder if we should use the "controller as" pattern instead.

h/static/scripts/account/test/auth-controller-test.coffee
```diff
@@ -18,15 +18,36 @@ class MockSession
...
-  auth = null
   session = null
-  $controller = null
+  mockFlash = info: sandbox.spy()
+  mockFormRespond = sandbox.spy()
+  inject(($controller, $rootScope) ->
+    session = mocksession or new MockSession()
+    $scope = $rootScope.$new()
+    $timeout = sandbox.spy()
+    params =
```
tilgovi: These would typical be called `locals` in the angular docs.

tilgovi:
:+1: to using the `locals` argument of `$controller` rather than using `$provide` to send the mock injections.

I like the way this looks, but a lot of that is the destructuring assignment. I'd also be happy, for the purpose of tests, to just keep things as local variables of the `describe` scope as we've been doing. That might translate with less boilerplate to JS. No strong opinion, really, but I don't see any gain from doing it this way and the rest of our test suite does it the other way. See any of the many test files with a `createController` function.

seanh:
CoffeeScript destructuring assignment is really nice :)

Normally if I had unit tests and some of them needed test object A, some needed B, and some C, assuming it took more than one line of code to create each of these test objects I'd have test helper methods `_a()`, `_b()` and `_c()`. Then each test would only call the methods it needs to get the test objects it needs. In this case though all 5 test objects are actually needed to construct the controller object that every test needs, so we end up with just one test helper function doing it all and every test calls that helper. But then the destructuring still helps to make the tests more readable by having each test only "take" what it needs.

What I'm trying to do here is have tests that support "get in and get out fast" development where if I have to fix a failing test, or add a new test, I can jump in and do it with the minimal amount of code reading and understanding, and definitely without having to fix any _other_ test methods that are broken by the one I changed or added.

For example I don't want to have to start by saying "Ok, how does _this_ test module work?" and figuring out all the module's before and afters and shared variables. I just want to jump straight to the one test method I'm interested in, and not have to look at anything else.

h/static/scripts/account/test/auth-controller-test.coffee
seanh: Thanks. I really wanted to just pass an inline anonymous object literal to the function call below, but it looks like there isn't a way to do this on CoffeeScript (unless the object literal fits on one line)? Will rename to `locals`

tilgovi:
I don't have a super-strong opinion here.

h/static/scripts/account/test/auth-controller-test.coffee
```diff
@@ -18,15 +18,36 @@ class MockSession
...
+      $scope: $scope
+      $timeout: $timeout
+      flash: mockFlash
+      session: session
+      formRespond: mockFormRespond
+    controller = $controller('AuthController', params)
+  )
+  return(
```
nickstenning:
```coffeescript
return(
  foo: bar
)
```

seems a bit perverse to me

```coffeescript
return {
  foo: bar
}
```

is, well, JavaScript, more-or-less, and would be preferred.

h/static/scripts/account/test/auth-controller-test.coffee
```diff
@@ -18,15 +18,36 @@ class MockSession
...
   $scope = null
   $timeout = null
-  auth = null
   session = null
-  $controller = null
+  mockFlash = info: sandbox.spy()
+  mockFormRespond = sandbox.spy()
+  inject(($controller, $rootScope) ->
+    session = mocksession or new MockSession()
+    $scope = $rootScope.$new()
```
nickstenning: Is there a reason why you couldn't go the whole hog and make `$scope` and `$timeout` simply locals as well? Why are we continuing to close over these?

h/static/scripts/account/test/auth-controller-test.coffee
nickstenning: Yup, this should be `locals`.

h/static/scripts/account/test/auth-controller-test.coffee
```diff
@@ -18,15 +18,36 @@ class MockSession
...
+_controller = (mocksession) ->
+  controller = null
   $scope = null
   $timeout = null
-  auth = null
   session = null
-  $controller = null
+  mockFlash = info: sandbox.spy()
+  mockFormRespond = sandbox.spy()
+  inject(($controller, $rootScope) ->
```
nickstenning: Wouldn't it be simpler to use `inject` to capture `$controller` and `$rootScope` only, and then you don't need to close over `controller`, you just have:

```coffeescript
$controller = null
$rootScope = null

inject (_$controller_, _$rootScope_) -> 
  $controller = _$controller_
  $rootScope = _$rootScope_

locals = {
  $scope: $rootScope.$new()
  $timeout: sandbox.spy()
  flash: mockFlash
  session: mocksession or new MockSession()
  ...
}

auth = $controller('AuthController', locals)
locals.auth = auth

return locals
```

nickstenning:
Overall I like this, because it replaces Angular magic (`$provide`) with straightforward function calls. I have made one suggestion for further simplification above.

tilgovi:
@seanh before we completely forget about this, maybe check @nickstenning's comment and then let's merge this.

seanh:
Yeah I was planning to get around to this..

seanh:
I have not forgotten this...

nickstenning:
I'm going to go ahead and proactively close this, just for the sake of reducing cognitive overhead. If you get a chance to come back to it, great: feel free to reopen.

PR #2118, opened 2015-03-31
Use domain as citation title if no document

### LINKED ISSUE #1542: Source URL not visible on older annotations
Go here:
https://stage.hypothes.is/a/atbunuGSQd-EniOgqgJxfQ

There's no source URL / page title visible.

seanh:
An annotation's document.title is used as the citation when displaying
an annotation. If this is an older annotation that has no document, fall
back to using the domain as the citation title instead.

Fixes #1542


seanh:
Work in progress, I want to add a test, and maybe refactor this controller a bit..

seanh:
I think this is ready for review now

nickstenning:
Pretty uncontroversial.

PR #2119, opened 2015-04-01
CLI utility to send invitations

tilgovi:


h/accounts/models.py
```diff
@@ -123,6 +123,17 @@ def subscriptions(self, value):
         else:
             self.status = (self.status or 0) & ~0b100
 
+    @property
+    def invited(self):
+        return bool((self.status or 0) & 0b100)
```
nickstenning: God this is ugly. Also, even if we must continue using this godawful bitfield, this should clearly be `0b1000`, not `0b100`.

nickstenning: Also maybe add a `# FIXME` or some other marker to remove this once we've been through the claim cycle? There won't be any more use for this code after we've sent out all these emails.

h/claim/invite.py
```diff
@@ -0,0 +1,182 @@
...
+    '-k',
+    '--key',
+    metavar='KEY',
+    help='Mandrill API key (defaults to MANDRILL_APIKEY variable)',
+    default=os.environ.get('MANDRILL_APIKEY'),
+)
+
+
+def get_users(session, limit=None):
+    return session \
```
nickstenning: I think I prefer wrapping things like this in parens. Not too fussed.

h/claim/invite.py
```diff
@@ -0,0 +1,182 @@
...
+:mod:`h.claim.invite` is a utility to invite users to claim accounts
+and is exposed as the command-line utility hypothesis-invite.
+"""
+import argparse
+import logging
+import os
+import sys
+import time
+
+import mandrill
```
nickstenning: The build is complaining about this. You can either add the extra to `.travis.yml` or `# noqa` this. Probably the former.

h/claim/invite.py
```diff
@@ -0,0 +1,182 @@
...
+            models.User.password == u'',
+            models.User.status.op('&')(0b1000) == 0  # noqa
+        ) \
+        .limit(limit) \
+        .all()
+
+
+def get_merge_vars(request, users):
+    for user in users:
+        claim = generate_claim_url(request, str(user.id))
```
nickstenning: Does this work? I think `generate_claim_url` is [*intended* to take acct URIs](https://github.com/hypothesis/h/blob/d488f9c491c0334ee4f46210ba17127df169a9e6/h/claim/test/utils_test.py#L16)?

nickstenning:
Mostly minor nitpicks other than the use of `generate_claim_url`.

h/claim/invite.py
tilgovi: It works fine because the get_by_id is overloaded and falls back to the
base class implementation.

If you think that's too gross to stomach I'll change it.

In fact, I think I'll do so anyway, since we have the proper base url.

tilgovi:
All good.

nickstenning:
:ship:

PR #2120, opened 2015-04-01
Bump the number of annotations loaded per request to 200

nickstenning:
The overhead associated with a single search call is high enough that
loading only 20 annotations at once is too little. In a naive test this
reduces the time take to request about 300 annotations from 5 seconds to
about 800ms (which, IMO, is still far too slow.)


gergely-ujvari:
LGTM

PR #2121, opened 2015-04-01
Automatically scroll to new annotations

### LINKED ISSUE #2053: Sometimes you have to search for the new annotation cards if many are present

seanh:
When the user adds a new annotation scroll the sidebar to show it.

The way this has been implemented causes annotations to be scrolled to
when their edit buttons are clicked as well, not just when new
annotations are created, which causes things to jump around confusingly
when clicking edit buttons. This needs to be fixed.

This fixes #2053 - in cases when there are more than a screenful of
annotations in the sidebar, creating a new annotation causes a new
annotation to appear in the sidebar but this may happen off screen
leaving the user confused. (For example: scroll to the very bottom of
the sidebar then create a new annotation at the top of the page, or
vice-versa.) The fix is to scroll the sidebar to show the new
annotation.

The diff for annotation.html is a mess - we've only actually changed a
couple of things, but we had to change the indentation of the whole file
so git thinks we changed everything.

- We use #fragment links to scroll to the new annotations.
- This requires the HTML elements for the annotations to have unique id
  attributes.
- We can't use the IDs of the annotations from the server because new
  annotations that haven't been saved don't have IDs yet.
- So instead we use timestamps.
- We use a new scrolltome directive that scrolls the page to its element
  when instantiated.
- We attach this scrolltome directive to an element within the
  annotation that has ng-if="vm.editing", whenever this expression
  becomes true the directive will be instantiated and cause a scroll.
  The expression is true when new annotations are added.
- We wrap the HTML for an annotation in a new div to hold the id that
  we're going to link to. We need an element that contains all of the
  HTML for the annotation so that the whole annotation is always
  scrolled into view, whether we're scrolling up or down to get to it.
  We can't use the outermost <article> of the annotation in thread.html
  because that's not within the scope of the AnnotationController where
  the element_created_at property that we need is located. So we add a
  new element.

### DIFF
--- a/h/static/scripts/directive/annotation.coffee
+++ b/h/static/scripts/directive/annotation.coffee
@@ -45,6 +45,7 @@ AnnotationController = [
     @hasDiff = false
     @showDiff = undefined
     @timestamp = null
+    @element_created_at = new Date().getTime()
 
     model = $scope.annotationGet()
     highlight = model.$highlight
--- a/h/static/scripts/app.coffee
+++ b/h/static/scripts/app.coffee
@@ -112,6 +112,7 @@ module.exports = angular.module('h', [
 .directive('match', require('./directive/match'))
 .directive('tabbable', require('./directive/tabbable'))
 .directive('tabReveal', require('./directive/tab-reveal'))
+.directive('scrolltome', require('./directive/scrolltome'))
 
 .filter('converter', require('./filter/converter'))
 .filter('moment', require('./filter/moment'))
--- a/h/static/scripts/directive/scrolltome.coffee
+++ b/h/static/scripts/directive/scrolltome.coffee
@@ -1,0 +1,8 @@
+module.exports = ['$location', '$anchorScroll', ($location, $anchorScroll) ->
+  return (
+    restrict: 'A'
+    link: (scope, element, attrs) ->
+      $location.hash(attrs.scrolltome)
+      $anchorScroll()
+  )
+]
--- a/h/templates/client/annotation.html
+++ b/h/templates/client/annotation.html
@@ -1,161 +1,164 @@
-<header class="annotation-header">
-  <!-- Deletion notice -->
-  <span ng-if="!vm.editing && vm.annotation.deleted">Annotation deleted.</span>
+<div id="{{vm.element_created_at}}">
+  <header class="annotation-header">
+    <!-- Deletion notice -->
+    <span ng-if="!vm.editing && vm.annotation.deleted">Annotation deleted.</span>
 
-  <!-- User -->
-  <span ng-if="vm.annotation.user">
-    <a class="annotation-user"
-       target="_blank"
-       ng-href="{{vm.baseURI}}u/{{vm.annotation.user}}"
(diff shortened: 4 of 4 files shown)


seanh:
- [ ] Needs tests
- [ ] needs docs (using ngdoc apparently)
- [ ] need to scroll only when creating new annotations, not when editing existing ones

@tilgovi What do you think of this solution? It's a shame the diff is so awful. To avoid inserting the new wrapper div, I'd have to move the unique ID to the `<article>` in thread.html (so it would now belong to which controller? not sure) and then the ID would have to be passed in to annotation.html or AnnotationController somehow

seanh:
This actually leaves a small part of the annotation off the top or bottom of the screen, the border/padding/margin that belongs to the `<article>` that's outside of the div I inserted, so we really should move the ids onto the articles, just got to figure out how

h/static/scripts/directive/annotation.coffee
```diff
@@ -45,6 +45,7 @@ AnnotationController = [
     @hasDiff = false
     @showDiff = undefined
     @timestamp = null
+    @element_created_at = new Date().getTime()
```
gergely-ujvari: Let's use camelCase here, as the rest of client side code uses that.

seanh:
I wonder what I'm gonna need to do to tell the difference between just-created annotations and older ones that have just had their editors opened..

h/static/scripts/directive/annotation.coffee
tilgovi: maybe uuid instead.

Can we hope for sub-millisecond linking of annotation directives? Maybe. Maybe not.

gergely-ujvari:
Yes, uuid is way better. (And maybe we still use the annotation.$$tags property for syncing?)

h/static/scripts/directive/annotation.coffee
seanh: Heh. I just used Date because it seemed to be easier to get in JS than a UUID. Happy to replace it

tilgovi: We've got the UUID library already as a dependency.

```coffee
uuid = require('node-uuid')
id = uuid.v4()
```

tilgovi:
More simply, maybe you could put `id="thread-{{$index}}"` on the thread `li` in "h/templates/client/viewer.html".

I would start to worry about the separation of threads and annotations here, but I think we're okay because threads do deal in `id` and `reference` even if they don't care about the rest of the properties of the annotation. So, then you could have the `thread` directive check:

- do I have an `id` attribute on my element?
- does my message have no `id` key (not persisted aka should be the latest draft)?
- okay, scroll to me

seanh:
And this $index is available anywhere inside the ng-repeat (even sub-templates with different controllers?) That would be simpler..

tilgovi:
I think it would be available, but that feels dirty. If you're not in the template that defined the `ng-repeat` you probably shouldn't assume you're in one.

seanh:
Closing, replaced with https://github.com/hypothesis/h/pull/2123

PR #2123, opened 2015-04-01
Scroll to new annotations on creation

### LINKED ISSUE #2053: Sometimes you have to search for the new annotation cards if many are present

seanh:
If necessary scroll the sidebar to show a newly created annotation (when
the user clicks the new annotation button).

Add unique ids to the <li> elements that contain the annotations, then
on creating a new annotation set the URL fragment to #<id> and scroll to
show the element.

Fixes #2053.

### DIFF
--- a/h/static/scripts/directive/thread.coffee
+++ b/h/static/scripts/directive/thread.coffee
@@ -1,3 +1,5 @@
+uuid = require('node-uuid')
+
 ###*
 # @ngdoc type
 # @name thread.ThreadController
@@ -148,6 +150,8 @@ ThreadController = [
       else
         0
 
+    this.uuid = uuid.v4()
+
     this
 ]
 
@@ -174,8 +178,8 @@ isHiddenThread = (elem) ->
 # Directive that instantiates {@link thread.ThreadController ThreadController}.
 ###
 module.exports = [
-  '$parse', '$window', 'pulse', 'render',
-  ($parse,   $window,   pulse,   render) ->
+  '$parse', '$window', '$location', '$anchorScroll', 'pulse', 'render',
+  ($parse,   $window, $location, $anchorScroll, pulse,   render) ->
     linkFn = (scope, elem, attrs, [ctrl, counter, filter]) ->
 
       # We would ideally use require for this, but searching parents only for a
@@ -183,6 +187,7 @@ module.exports = [
       ctrl.parent = elem.parent().controller('thread')
       ctrl.counter = counter
       ctrl.filter = filter
+      ctrl.id = attrs.id
 
       # Track the number of messages in the thread
       if counter?
@@ -210,6 +215,10 @@ module.exports = [
         render ->
           ctrl.container = thread
           scope.$digest()
+          if (ctrl.id and not ctrl.container?.message?.id)
+            # Scroll the sidebar to show new annotations.
+            $location.hash(ctrl.id)
+            $anchorScroll()
 
     controller: ThreadController
     controllerAs: 'vm'
--- a/h/static/scripts/directive/thread.coffee
+++ b/h/static/scripts/directive/thread.coffee
@@ -150,7 +150,7 @@ ThreadController = [
       else
         0
 
-    this.uuid = uuid.v4()
+    this.id = uuid.v4()
 
     this
 ]
@@ -187,7 +187,6 @@ module.exports = [
       ctrl.parent = elem.parent().controller('thread')
       ctrl.counter = counter
       ctrl.filter = filter
-      ctrl.id = attrs.id
 
       # Track the number of messages in the thread
       if counter?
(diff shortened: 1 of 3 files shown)


seanh:
Needs tests.

I want to replace `if (ctrl.id and not ctrl.container?.message?.id)` with a nice documented (maybe even tested) property.

There's a weird "bounce" happening if you do this:

- Reload the page (so there are no annotations with their editors open)
- Scroll sidebar to very top
- Make a new annotation at the very bottom of the page
- Sidebar scrolls down to the bottom, does a weird bounce/flicker

Otherwise it seems to work fine

seanh:
@tilgovi This is a much better solution, less code and doesn't have any of the problems the previous one had

h/static/scripts/directive/thread.coffee
```diff
@@ -174,8 +178,8 @@ isHiddenThread = (elem) ->
 # Directive that instantiates {@link thread.ThreadController ThreadController}.
 ###
 module.exports = [
-  '$parse', '$window', 'pulse', 'render',
-  ($parse,   $window,   pulse,   render) ->
+  '$parse', '$window', '$location', '$anchorScroll', 'pulse', 'render',
+  ($parse,   $window, $location, $anchorScroll, pulse,   render) ->
```
csillag: Indentation: please add some spacing so that `$location` ends up directly under `'$location'`, etc.

h/static/scripts/directive/thread.coffee
```diff
@@ -210,6 +215,10 @@ module.exports = [
         render ->
           ctrl.container = thread
           scope.$digest()
+          if (ctrl.id and not ctrl.container?.message?.id)
+            # Scroll the sidebar to show new annotations.
+            $location.hash(ctrl.id)
+            $anchorScroll()
```
csillag: Funny to see another name collision arising; this time for anchor.
(True, given the existing terms used in HTML world, it always lingered around, but at least we didn't really use that term inside our code, up now now.)

That is not to say that this is wrong: we could rename the other `anchors`, as discussed earlier. (Maybe to `TargetLocation`, or something like that.)

csillag:
> I want to replace if (ctrl.id and not ctrl.container?.message?.id) with a nice documented (maybe even tested) property.

Yes, that would be nice. We have many states for annotations; it's hard to keep all those corner cases in mind. (Top-level annotations, replies, highlights, page comments, deleted annotations, whatever I am missing right now...)

Otherwise, LGTM.

tilgovi:
Only thing that feels strange to me is setting `.id` and `.uuid` on the controller to the same thing.

seanh:
Ok, so how can I simulate what happens when the user adds a new annotation (creating a new ThreadController instance and everything that depends on) so that I can test the `isNew()` method? Will look for something in the existing tests..

seanh:
The tests that I added seem a bit artificial. In our code we assume that a new annotation does not have a container.message.id but an old annotation does have one. The same assumption is built into the tests. Is there a way we can test this assumption? Actually create a ThreadController for a new annotation and one for an old annotation in the same way that the app would?

seanh:
@nickstenning suggested the bounce might be caused by existing annotations from the server being loaded on scroll, and I think he migh

tilgovi:
Does it only happen at the bottom? Could it also be layout changing as the
annotation directive links?

seanh:
I've only be able to reproduce it at the very bottom of the sidebar. Also if you load the page then scroll to the bottom of the sidebar (it'll bounce) then scroll back to the top, _then_ add an annotation to the bottom of the doc, it auto scrolls the sidebar to the bottom with no bounce

tilgovi:
I don't see the symptoms you describe in FF or Chromium.

tilgovi:
I'm inclined to merge this and deal with this bouncing if we can isolate it later.

tilgovi:
Regardless, the experience is vastly improved over not knowing where your annotation is :).

tilgovi:
Nice work, @seanh.

seanh:
Cheers

tilgovi:
Had to revert this. I restored the branch. The stream is flickering and changing hash like crazy.

seanh:
Ok, sounds like isNew() is wrong probably, I'll investigate

seanh:
Hmm, I can't reproduce the problem, stream seems fine on my dev instance.

tilgovi:
I'll check it out again.

seanh:
@tilgovi Bump. This one is a priority bug for this print, I haven't been able to reproduce the problem with the stream flickering yet. Do you think we'll be able to fix this pr and get it in, or do we need a different approach?

tilgovi:
I'll give it another look today.

csillag:
> Had to revert this. I restored the branch. The stream is flickering and changing hash like crazy.

I can't reproduce this problem on this branch, either.

PR #2125, opened 2015-04-02
[WIP] Push handling annotations to the template controllers …

### LINKED ISSUE #1916: Too many message not available on stream
bad_stream
 
gergely-ujvari:
AppController is no longer responsible for handling streamer updates,
nor cleaning up annotations in login/logout.

Instead the three template controllers are responsible.
Widget-Controller for the sidebar.
Annotation-Viewer-Controller for the standalone page.
And Streamer-Controller for the stream.

They listen to streamer.onmessage, the annotation events and,
the 'cleanupAnnotations' event emitted at login/logout.

Fix #1916 

CleanupAnnotation has changed, no longer keeping any annotations,
just dropping everything and reloading all.

### DIFF
--- a/h/static/scripts/annotation-viewer-controller.coffee
+++ b/h/static/scripts/annotation-viewer-controller.coffee
@@ -3,18 +3,23 @@ angular = require('angular')
 
 module.exports = class AnnotationViewerController
   this.$inject = [
-    '$location', '$routeParams', '$scope',
-    'streamer', 'store', 'streamFilter', 'annotationMapper'
+    '$location', '$rootScope', '$routeParams', '$scope',
+    'annotationMapper', 'drafts', 'streamer', 'store', 'streamFilter',
+    'threading'
   ]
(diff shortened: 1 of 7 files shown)


tilgovi:
Rebased for a trivial conflict I introduced on master.

tilgovi:
Simplify the syntax for the listeners in the stream controller a little bit.

tilgovi:
Sorry, I meant to say "simplified". I did it and push it.

tilgovi:
Not sure about this anymore. I'm going to try to make a diagram of some flows. We may just want to use the angular event and start broadcasting the lifecycle events.

tilgovi:
I think we're going to need to revisit the `annotationMapper` and the role of events in the angular application before we can continue with this.

Right now, I think this is splitting some things (good) but it's a lot of duplication (bad) and it doesn't address these other root problems.

PR #2127, opened 2015-04-02
Test form-input interaction with form-validate

tilgovi:
Mock and test that half of the interaction, and in doing so show that
it's also optional to use the validator with the input.


nickstenning:
Sure. :eggplant:

PR #2128, opened 2015-04-02
Show spinner in search bar during requests

### LINKED ISSUE #930: Loading... icon for streams
@shawninder says:

> First the page loads with two sentences: 
>
> >   Please begin your search by adding terms above
>
> and
>
> >   No results were found. Listening for new matches...
> 
>
> Almost immediately after, the "No results..." sentence disappears
> (so fast it's hard to read). I am now left with only the 
> "Please begin your search..." sentence.
>
>Nearly 4 seconds later (!), the "Please begin you search..."  sentence
> disappears and the search results appear.
> 
> I think the page should load with a "Loading..." sentence or icon.
> Then, when the search results come in (ideally less than 4 seconds later),
> it can be replaced with either the search results or the
> "No results..." sentence. Only if the page is requested without 
> any GET parameters should the "Please begin your search" sentence appear.

tilgovi:
Fix #930

### DIFF
--- a/h/static/scripts/directive/simple-search.coffee
+++ b/h/static/scripts/directive/simple-search.coffee
@@ -1,8 +1,5 @@
-module.exports = ['$parse', ($parse) ->
-  uuid = 0
+module.exports = ['$http', '$parse', ($http, $parse) ->
   link: (scope, elem, attr, ctrl) ->
-    scope.viewId = uuid++
-
     scope.reset = (event) ->
       event.preventDefault()
       scope.query = ''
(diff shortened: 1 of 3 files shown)


tilgovi:
loading-icon

dwhly:
Really nice.

nickstenning:
This looks great to me, but I would question whether this is the right place for the code. This spinner represents open HTTP requests across all activity, not just searches, so finding it in `simple-search.coffee` would I think be quite surprising to me.

Is there an easy way we could reorganise responsibility for this part of the UI? If not, happy to merge as-is.

tilgovi:
So I did work out a branch where this listens to route change{Start,End}
but to do that I had to make it so routes fetch their initial loading data
in a resolve function.

I'll bring that up in a separate PR because it needs some discussion, I
think.

For now, this was a way to get this working quickly without any real
spaghetti code.

Happy to wait on this, or not.

PR #2131, opened 2015-04-03
Let users change their email addresses

### LINKED ISSUE #174: When there are "0" annotations above or below, hide up/down tab. [f1]

seanh:
Fixes https://github.com/hypothesis/vision/issues/174

account.html:

- Add a new "Change Your Email Address" form

accounts/views.py:

- Add email to the user dict returned at /app?__formid__=profile (or
  session.profile() in Angular)

- When resetting the form after a successful submission, put the new (or
  unchanged) email address into the user dict returned in the response.
  AccountController needs this (see below).

account-controller.coffee:

- Add $scope.email (the user's current email address), to use for placeholder
  text in the email form field.

- After a successful form submission set $scope.email to the one from the
  response. This sets the placeholder text in the email field to the new value,
  instead of continuing to show the old email address as placeholder.

- Add changeEmail() method to receive email form submissions.

### DIFF
--- a/h/accounts/schemas.py
+++ b/h/accounts/schemas.py
@@ -189,6 +189,11 @@ class EditProfileSchema(CSRFSchema):
         default='',
         missing=colander.null
     )
+    emailAgain = colander.SchemaNode(
+        colander.String(),
+        default='',
+        missing=colander.null,
+    )
     password = colander.SchemaNode(
         colander.String(),
         title=_('Password'),
--- a/h/accounts/schemas.py
+++ b/h/accounts/schemas.py
@@ -186,6 +186,7 @@ class EditProfileSchema(CSRFSchema):
     )
     email = colander.SchemaNode(
         colander.String(),
+        validator=colander.All(colander.Email(), unique_email),
         default='',
         missing=colander.null
     )
--- a/h/accounts/test/views_test.py
+++ b/h/accounts/test/views_test.py
@@ -11,6 +11,7 @@ from horus.schemas import ProfileSchema
 from horus.forms import SubmitForm
 from horus.strings import UIStringsBase
 
+import h.accounts.views
 from h.accounts.views import RegisterController
 from h.accounts.views import ProfileController
 from h.models import _
@@ -35,25 +36,6 @@ def configure(config):
     config.registry.registerUtility(MagicMock(), IRegisterForm)
     config.registry.feature = MagicMock()
     config.registry.feature.return_value = None
-
-
-def _mock_db_session_query_method():
-    """Return a mock session.query() that passes unique_email validation.
-
-    Return a callable that returns an object with a filter() method that
-    returns an object with a first() method that returns None.
-
-    When attached to a dummy_db_session object as dummy_db_session.query()
-    this callable will cause Horus's unique_email validation check to pass.
-
-    """
-    result = MagicMock()
-    result.first.return_value = None
-    filter_ = MagicMock()
-    filter_.filter.return_value = result
-    def query(*args, **kwargs):
-        return filter_
-    return query
 
 
 def _get_fake_request(username, password, with_subscriptions=False, active=True):
@@ -105,85 +87,44 @@ def test_profile_invalid_password(config, user_model):
     assert any('pwd' in err for err in result['errors'])
 
 
-@pytest.mark.usefixtures('activation_model',
-                         'dummy_db_session')
-def test_edit_profile_with_non_matching_emails(
-        config, user_model, dummy_db_session):
-    """edit_profile() should return an error if the emails don't match."""
-    request = _get_fake_request('john', 'doe')
-    request.params["email"] = "example@example.com"
-    request.params["emailAgain"] = "different@example.com"
+@pytest.mark.usefixtures('activation_model', 'dummy_db_session')
+def test_edit_profile_with_validation_failure(config, user_model):
+    """If _validate_edit_profile_request() raises it should return an error.
+
+    If _validate_edit_profile_request() raises an exception then edit_profile()
+    should return a dict with an "errors" list containing a list of the
+    error(s) from the exception's .errors property.
+
+    """
     configure(config)
+    profile = ProfileController(DummyRequest())
+    errors = [
+        ("email", ["That email is invalid", "That email is taken"]),
+        ("emailAgain", "The emails must match."),
+        ("password", ["That password is wrong"])
+    ]
 
-    dummy_db_session.query = _mock_db_session_query_method()
+    with patch("h.accounts.views._validate_edit_profile_request") as validate:
+        validate.side_effect = (
+            h.accounts.views._InvalidEditProfileRequestError(errors=errors))
+        result = profile.edit_profile()
 
-    profile = ProfileController(request)
-    result = profile.edit_profile()
-
-    assert "errors" in result
-    assert result["errors"][0][1] == "The emails must match."
-
-
-@pytest.mark.usefixtures('activation_model',
-                         'dummy_db_session')
-def test_edit_profile_with_invalid_email(config, user_model, dummy_db_session):
-    """edit_profile() should return an error if the new email is invalid."""
-    request = _get_fake_request('john', 'doe')
-    request.params["email"] = "foo@bar"
-    request.params["emailAgain"] = "foo@bar"
-    configure(config)
-
-    dummy_db_session.query = _mock_db_session_query_method()
-
-    profile = ProfileController(request)
-    result = profile.edit_profile()
-
-    assert result["errors"][0][1][0] == "Invalid email address"
+    assert result["errors"] == errors
 
 
-@pytest.mark.usefixtures('activation_model',
-                         'dummy_db_session')
-def test_edit_profile_with_taken_email(config, user_model, dummy_db_session):
-    """edit_profile() should return an error if the email is already taken."""
-    request = _get_fake_request('john', 'doe')
-    request.params["email"] = "john@doe.com"
-    request.params["emailAgain"] = "john@doe.com"
-    configure(config)
-
-    # Mock dummy_db_session.query() so that Horus's unique_email validation
-    # check will fail.
-    first = MagicMock()
-    first.email = "john@doe.com"
-    results = MagicMock()
-    results.first.return_value = first
-    filter_ = MagicMock()
-    filter_.filter.return_value = results
-    def query(*args, **kwargs):
-        return filter_
-    dummy_db_session.query = query
-
-    profile = ProfileController(request)
-    result = profile.edit_profile()
-
-    assert "errors" in result
-    assert result["errors"][0][1][0] == (
-        "Sorry, an account with the email john@doe.com already exists. Try "
-        "logging in instead.")
-
-
-@pytest.mark.usefixtures('activation_model',
-                         'dummy_db_session')
-def test_edit_profile_successfully(config, user_model, dummy_db_session):
+@pytest.mark.usefixtures('activation_model', 'dummy_db_session')
+def test_edit_profile_successfully(config, user_model):
     """edit_profile() returns a dict with key "form" when successful."""
-    request = _get_fake_request('john', 'doe')
-    request.params["email"] = "example@example.com"
-    request.params["emailAgain"] = "example@example.com"
     configure(config)
+    profile = ProfileController(DummyRequest())
 
-    dummy_db_session.query = _mock_db_session_query_method()
-
-    profile = ProfileController(request)
-    result = profile.edit_profile()
+    with patch("h.accounts.views._validate_edit_profile_request") as validate:
+        validate.return_value = {
+            "username": "johndoe",
+            "pwd": "password",
+            "subscriptions": []
+        }
+        result = profile.edit_profile()
 
     assert "form" in result
     assert "errors" not in result
--- a/h/accounts/test/views_test.py
+++ b/h/accounts/test/views_test.py
@@ -76,7 +76,7 @@ class TestProfile(object):
 
 class TestEditProfile(object):
 
-    """Unit tests for the edit_profile() function."""
+    """Unit tests for ProfileController's edit_profile() method."""
 
     @pytest.mark.usefixtures('activation_model', 'dummy_db_session')
     def test_profile_invalid_password(self, config, user_model):
--- a/h/accounts/test/views_test.py
+++ b/h/accounts/test/views_test.py
@@ -70,80 +70,81 @@ def test_profile_returns_email(config, user_model, authn_policy):
     assert profile["model"]["email"] == "test_user@test_email.com"
 
 
-@pytest.mark.usefixtures('activation_model',
-                         'dummy_db_session')
-def test_profile_invalid_password(config, user_model):
-    """Make sure our edit_profile call validates the user password"""
-    request = _get_fake_request('john', 'doe')
-    configure(config)
+class TestEditProfile(object):
 
-    # With an invalid password, get_user returns None
-    user_model.get_user.return_value = None
+    """Unit tests for the edit_profile() function."""
 
-    profile = ProfileController(request)
-    result = profile.edit_profile()
+    @pytest.mark.usefixtures('activation_model', 'dummy_db_session')
+    def test_profile_invalid_password(self, config, user_model):
+        """Make sure our edit_profile call validates the user password."""
+        request = _get_fake_request('john', 'doe')
+        configure(config)
 
-    assert result['code'] == 401
-    assert any('pwd' in err for err in result['errors'])
+        # With an invalid password, get_user returns None
+        user_model.get_user.return_value = None
 
-
-@pytest.mark.usefixtures('activation_model', 'dummy_db_session')
-def test_edit_profile_with_validation_failure(config, user_model):
-    """If _validate_edit_profile_request() raises it should return an error.
-
-    If _validate_edit_profile_request() raises an exception then edit_profile()
-    should return a dict with an "errors" list containing a list of the
-    error(s) from the exception's .errors property.
-
-    """
-    configure(config)
-    profile = ProfileController(DummyRequest())
-    errors = [
-        ("email", ["That email is invalid", "That email is taken"]),
-        ("emailAgain", "The emails must match."),
-        ("password", ["That password is wrong"])
-    ]
-
-    with patch("h.accounts.views._validate_edit_profile_request") as validate:
-        validate.side_effect = (
-            h.accounts.views._InvalidEditProfileRequestError(errors=errors))
-        result = profile.edit_profile()
-
-    assert result["errors"] == errors
-
-
-@pytest.mark.usefixtures('activation_model', 'dummy_db_session')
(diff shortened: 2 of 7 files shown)


seanh:
A problem I had with implementing this is that the user's current email address didn't seem to be available to the frontend. The session service seemed to be the right place to get user info, with `session.profile()`. So I made a change in `views.py` to add the email to this. If you go to `/app?__formid__=profile` there'll now be an email in the JSON. If you just go to `/app` though you get the same JSON with no email :(

- [ ] Tests
- [ ] Refactor the controller a little (duplication)
- [ ] Redesign the form - it has three fields for the current password, two Update buttons. The three separate forms could be merged into one smaller form. (Maybe a separate pr).

tilgovi:
We currently require users to verify their email address before they can log in and make annotations. I feel it only makes sense to extend that to the new email address.

Need answers to questions like:

- If I try to change my email address, should I have to confirm my new email? (I think, "yes".)
- If I try to change my email address, should I be able to log in with the old one until I confirm?
  - If not, and I typo my new email address, do I have/need any recourse other than support@.
  - If not, is there enough info logged anywhere that support@ could verify the old email address and restore access?

etc

seanh:
Hmm, this adds a lot of work! But you're right we can't just let people enter any old email address. I guess what we should do is hold any email address changes (previous address setting remains) until an activation link sent to the new address has been clicked on?

So I think the answers to both your questions above are yes.

I'll look into how we currently do address verification and how we can implement this..

seanh:
Reminder to myself: we need to handle the case where the user changes their email address, then changes it again without clicking on the activation link. Should this send out another acitvation email? With the same link or a different one? What if we send another email, then they click on the link in the first email?

seanh:
So on looking into this and talking to @nickstenning about it in irc, I think verifying new emails would require significant additions or changes to horus and we don't want to put effort into that at the mo, so our suggestion is to add a second field (type your email twice so hopefully you don't get it wrong) and allow changing emails without activation links for now

seanh:
@judell This one is half done. I'm going to leave it here as it's not in our current sprint, but when to get finished in a later sprint maybe

tilgovi:
I'm not super comfortable with not verifying new email addresses unless we want to simply say that activation is never required and do away with the entire issue.

nickstenning:
I think adding this without email verification is fine. What's worse:

- users can't change their own email addresses at all
- users can change their own email addresses to an unverified address, but only after they've already signed up with a verified email address

Let's add a story to add verification to email address changes later. For now I think this materially improves things.

tilgovi:
Alright. I think this is fine, then, no?

seanh:
No! It doesn't make users type their email address twice to verify it yet. But it's on hold for now as not in this sprint

tilgovi:
@seanh you sure? @judell put it in this milestone.

seanh:
Hmm, well it wasn't in Trello anyway

judell:
It wasn't, sorry, Trello/GH sync glitch.

Do you want it in this sprint?

seanh:
I may as well continue on this now..

judell:
Absolutely, thanks!

seanh:
As it turned out I ended up doing other things ... Let's get this one done soon though, it doesn''t need much more

seanh:
This is ready for review, then I'll rebase when people are happy. I'm happy with how the frontend tests turned out in the end. One thing I know @tilgovi doesn't like about this is adding a new `email` variable to the top-level of the scope. But where should I put it instead? This variable should maybe be renamed as well - it's specifically used only for the placeholder text in the email input fields.

h/accounts/schemas.py
```diff
@@ -189,6 +189,11 @@ class EditProfileSchema(CSRFSchema):
         default='',
         missing=colander.null
     )
+    emailAgain = colander.SchemaNode(
```
nickstenning: Hmm. I could have sworn there was a deform widget type that would just render two fields and automatically set up validation that they were the same...

Yes, [there is](https://deform.readthedocs.org/en/latest/api.html#deform.widget.CheckedInputWidget).

We might not need to do any of the validation ourselves if we just set the `widget` kwarg on the email node to one of these.

seanh: > We might not need to do any of the validation ourselves if we just set the widget kwarg on the email node to one of these.

This results in a great big stack trace from deform.

I'll try to figure it out but to be honest I feel this is a ludicrously complicated way to compare two strings (involving two third-party libraries). How many forms do we actually have in h? Wondering whether having deform and colander in our code at all is justifiable.

tilgovi: If deform were rendering it....

But isn't the form hand written?

tilgovi: I'm leaning away from deform lately. Unless we really like deform2 for some
reason.

nickstenning: No that's fine, my mistake. I thought deform was rendering this.

seanh: I am glad not to have to change this to use deform :)

h/accounts/test/views_test.py
```diff
@@ -32,6 +32,8 @@ def configure(config):
     config.registry.registerUtility(SubmitForm, IProfileForm)
     config.registry.registerUtility(MagicMock(), IRegisterSchema)
     config.registry.registerUtility(MagicMock(), IRegisterForm)
+    config.registry.feature = MagicMock()
```
nickstenning: `feature` has moved onto the config object so this either needs changing or perhaps is no longer necessary.

seanh:
Rebased onto master

h/accounts/test/views_test.py
seanh: Has it moved? This line on master is still accessing `registry.feature`: https://github.com/hypothesis/h/blob/master/h/accounts/views.py#L308

seanh: Ok, <https://github.com/hypothesis/h/pull/2239>, I'll try to figure out what's going on here

seanh: Ok that line is actually crashing on master currently (but there's not test failing).

@tilgovi @nickstenning I'm not sure how Pyramid controllers are supposed to access `feature()` now? Back when we were doing the original atom stream feature, there was an `includeme()` in `api_client/__init__.py` that would add a request method to the request object for getting an api client. includeme() would access api_url from settings to do this. But the idea was that view code didn't need to access settings, it just used the api client that had been placed on the request object for it.

So what about features? I could have an includeme call feature() and add a notifications_feature boolean to the request but this seems silly as we just removed feature() from request and we're gonna end up polluting the request object with a lot of new feature on/off properties.

What's the intention here? (Can't find an example in the code)

seanh: > Ok that line is actually crashing on master currently (but there's not test failing).

On master: login and choose Account from the username dropdown menu in the top right:

```
File "/home/seanh/Projects/hypothesis/h/h/accounts/views.py", line 308, in profile
    if request.registry.feature('notification'):
AttributeError: 'Registry' object has no attribute 'feature'
```

nickstenning: I've reverted that changeset: ad569a0

h/templates/client/settings/account.html
```diff
@@ -1,5 +1,51 @@
...
   <form class="account-form form"
+        name="changeEmailForm"
+        ng-submit="changeEmail(changeEmailForm)"
+        novalidate form-validate>
+
+    <h2 class="form-heading"><span>Change Your Email Address</span></h2>
+
+    <div class="form-field">
+      <label class="form-label" for="field-email">Email Address:</label>
+      <input id="field-email" class="form-input" type="email" name="email" placeholder="{{email}}" required ng-model="changeEmail.email" />
```
nickstenning: I'm not keen on the placeholder. I find it super confusing to interact with:

placeholder-badness

Either set default values (i.e. the value attribute) or just skip this entirely.

nickstenning: See also: http://www.nngroup.com/articles/form-design-placeholders/

nickstenning:
How are we validating email addresses... at the moment this form accepts "foo@bar" as valid, which doesn't seem wise.

nickstenning:
If I try and use the form twice in quick succession is not changed, I see no errors in the page, and this in the console:

screen shot 2015-05-20 at 15 03 15

seanh:
> How are we validating email addresses... at the moment this form accepts "foo@bar" as valid, which doesn't seem wise.

Following the existing code there is no server-side email address validation (I can add this if we think it's worth while). Angular automatically validates them on the frontend (this is triggered by the `<input type=email`). It is working, "foobar" will be rejected for example, but you're right that it lets "foo@bar" through. What do you think? Add better validation server-side?

seanh:
Rebased again since master changed again.

Changed the form:

screenshot from 2015-05-20 15 48 20

nickstenning:
> It is working, "foobar" will be rejected for example, but you're right that it lets "foo@bar" through. What do you think? Add better validation server-side?

Ew! Yes. I'm happy to not block merging this PR but doing input validation on the client only is a BIG "no no" in my experience. Let's make fixing that a priority.

nickstenning:
Still seeing a couple of issues with this.

- [ ] strange bug where email doesn't get fully replaced in the template:

    

- [ ] can't fill out the form twice in quick succession [as mentioned above](https://github.com/hypothesis/h/pull/2131#issuecomment-103876990)

seanh:
> If I try and use the form twice in quick succession is not changed, I see no errors in the page, and this in the console:
> 
> screen shot 2015-05-20 at 15 03 15

This one's a pain. I can reproduce it but the error message tells me nothing and I have no idea what's causing it

seanh:
> - strange bug where email doesn't get fully replaced in the template:
> 
>     

I'm not able to reproduce this (maybe fixed by other changes)

seanh:
Forgot some of my own todos from way up in this pr:

- [ ] `/app?__formid__=profile` is crashing if you're not logged in

seanh:
I don't see what the duplication I was talking about is, so checking it as done

seanh:
@nickstenning I'm expecting the tests to pass on Travis, think all the issues are addressed

seanh:
Interesting link about form placeholders by the way

seanh:
Todo:

- [ ] Something about this bug (or at least investiage it)

   

- [ ] Validate email addresses server-side

seanh:
@nickstenning I think that fixes the bug (seems to have fixed it for me), if you can retest please

I'll get on with fixing the server-side validation as discussed in sprint planning

seanh:
@nickstenning OK, ready for review again

nickstenning:
I promise I'm not trying to drag this out unnecessarily, but it does seem to me like the testing strategy for this is a little bit too "functional" for my taste. You're testing, in places, that colander works, which I would consider outside the scope of a sensible controller view function unit test.

I'd recommend breaking up the tests substantially, so that you

a) test custom validators (e.g. `unique_email` is currently untested)
b) (possibly) test that a schema field has the right validators, although even this might be overkill
c) test that the view function is using the correct schema

rather than testing the entire form handler end-to-end.

seanh:
I'm not sure. I think when faced with this kind of code it might be best to treat it as a black box and test end to end inputs and outputs, rather than testing the details of how it's implemented.

`unique_email` is a horus thing (I suppose I could submit a pull request with unit tests to horus, if it doesn't have any).

> c) test that the view function is using the correct schema

I think the problem with this is that the view function doesn't simply use the right schema. It has to pass in `_emails_must_match_validator()`, call `validate()` on a form constructed with the schema, catch `ValidationError`, format an errors dict to return, etc. If it gets any of these steps wrong it won't work. So do I write unit tests for each of these, mocking out all of the many implicit dependencies hidden in a view function like `edit_profile()`'s body?

The problem with testing that the view function does steps 1 through 7 is that you're encoding into your test the assumption that doing those steps will produce the right result. It feels to me a bit like looking at the implementation and saying ok it does steps 1, 2 and 3 so I'll write tests that it does 1, 2 and 3. Rather than asking what is it supposed to achieve in terms of inputs and outputs and then testing that.

I agree this is starting to look more like an integration test than a unit test. But I think that's because methods like `edit_profile()` aren't really units like say a `unique_email()` could be. They're integration methods that tie together several units at once.

I agree that this feels a little brittle.

The reason might be that Horus-based code not being testable, rather than that we can improve it much by testing differently. Stuff like this `exists = get_session(req).query(User).filter(User.email.ilike(val)).count()` right in the middle of a schema function means that if you're testing using a mock database session, it needs to be one with mock query, filter and count methods that return the right mocked objects. So your test ends up depending on a bunch of implementation details of the code you're testing. Horus is full of this sort of thing.

Alternative would be to mock one level higher: mock out horus rather than mocking the db session. Actually horus is called by colander which is called by deform, so you end up mocking deform. Here's a working example of one of the tests rewritten this way:

```
@pytest.mark.usefixtures('activation_model',
                         'dummy_db_session')
def test_edit_profile_with_invalid_email(config, user_model):
    """edit_profile() should return an error if the new email is invalid."""
    request = _get_fake_request('john', 'doe')
    request.params["email"] = "foo@bar"
    request.params["emailAgain"] = "foo@bar"
    configure(config)

    profile = ProfileController(request)
    with patch("deform.Form.validate") as mock_validate:
        error = MagicMock()
        error.children = [[None, ["Invalid email address"]]]
        mock_validate.side_effect = deform.ValidationFailure(
            "field", "cstruct", error)
        result = profile.edit_profile()

    assert result["errors"][0][1][0] == "Invalid email address"
```

The problem with that is your test just ends up depending on implementation details of deform instead of horus. The test has to actually construct a `deform.ValidationFailure()` with the right args and also mimick the ridiculous error object structure.

Maybe this way is better, or maybe it's just as bad and we may as well couple tests to horus and get the benefit of them being more end-to-end.

seanh:
It might be a case of just finding the right place to do the mocking. Doing it on mock_db_session or on deform's validate() is complicated because those have complicated interfaces. But horus's unique_email() function has a simple interface and should be easy:

```
@pytest.mark.usefixtures('activation_model', 'dummy_db_session')
def test_edit_profile_with_non_matching_emails(config, user_model):
    """edit_profile() should return an error if the emails don't match."""
    request = _get_fake_request('john', 'doe')
    request.params["email"] = "example@example.com"
    request.params["emailAgain"] = "different@example.com"
    configure(config)

    profile = ProfileController(request)

    with patch("horus.schemas.unique_email") as unique_email:
        unique_email.return_value = None
        result = profile.edit_profile()

    assert "errors" in result
    assert result["errors"][0][1] == "The emails must match."
```

Unfortunately that patch doesn't work, colander still ends up calling the real unique_email(), have to find what other name this function has.

tilgovi:
unique_email is not horus it's in h/accounts/schema.py, no?

Or it could be.

seanh:
https://github.com/hypothesis/h/blob/master/h/accounts/schemas.py#L7

seanh:
Patching "h.accounts.schemas.unique_email" doesn't seem to work either :(

tilgovi:
I was misremembering, thinking about the username hook.

Just pull out what you need. I'm slowly try to work us away from horus.

And it's not patching because I was bad and imported it directly rather
than importing the module. This came up in our conversation about import
style.

seanh:
So `patch("horus.schemas.unique_email")` doesn't work because of the `from horus.schemas import unique_email` in `h/accounts/schemas.py`. But that means that `patch("h.accounts.schemas.unique_email")` should work, shouldn't it? (It doesn't)

seanh:
I think there's something weird going on (with

seanh:
Guessing this is to do with when the `RegisterSchema` code in `schemas.py` is run: this gets run _before_ the first line of the test method is run so there hasn't been a chance to patch anything yet

seanh:
Nope, I changed my code to do the patch before schemas.py gets imported - it still calls the real unique_email().' Sheesh.

tilgovi:
It's referencing it directly, not via any module.

If it were `horus.schemas.unique_email` I think it would work, no?

tilgovi:
But also, thinking about what Nick said, not sure whether/if the code that actually instantiates this schema needs to run for these tests.

seanh:
If I were to do this in a more unit test style (by which I mean mocking out all the dependencies and just testing how it interacts with them), I think the complete set of tests for `edit_profile()` would look something like this:

* Unit test `_emails_must_match_validator()` (mocking form and value, patching `colander.Invalid`)
* Mock `self.request` with a fake request
* Patch `h.accounts.schemas.EditProfileSchema`
* Test that an `EditProfileSchema` is initialized passing `_emails_must_match_validator()` to `__init__()`.
* Test that the `bind()` method of the mock `EditProfileSchema` is called with the mock request as arg
* Patch `deform.Form`
* Test that a mock `deform.Form` is initialized and passed the mock `EditProfileSchema`
* Test that if the mock form's `validate()` method raises `deform.ValidationFailure()` then `edit_profile()` returns an errors dict
* Make the mock `validate()` return a mock appstruct dict
* (I'm going to ignore the whole `if subscriptions:` block)
* Mock `self.User` 
* Test that `self.User.get_user(...)` is called with the right params
* Test that if `self.User.get_user(...)` returns `None` then `edit_profile()` returns an errors dict
* Patch `horus.views.ProfileController.edit_profile()`
* Test that if `self.User.get_user(...)` returns a mock user then `edit_profile()` sets `request.context` to the mock user
* Test that if `self.User.get_user(...)` returns a mock user then `edit_profile()` calls `horus.views.ProfileController.edit_profile()` (via `super`)

Normally I would totally agree with what @nickstenning is saying here. We have units A, B and C (let's say they are simple functions without many dependencies or side-effects) and we can unit test each of them by testing their inputs and outputs. Then we have unit D which is a higher-level function that integrates A, B and C and we just add another set of unit tests that D calls A, B and C when it should (and doesn't when it shouldn't) and passes the right params to them. Sometimes it may need to get a return value from A and pass that on to B. But hopefully it wouldn't get much more complicated than that, and it can all be done with mocks of A, B and C.

The problem with `edit_profile()` in my opinion is that the interface between it and request, EditProfileSchema, deform, horus etc is too complicated making this kind of testing produce a kind of ridiculous list of unit tests.

But the more integration-tests approach that I've implemented does involve some brittle-looking mocking so I'm happy to rewrite it this way if you guys prefer.

seanh:
I haven't added tests  for `EditProfileSchema`. In theory I think you could delete e.g. the email validation from the schema and it'd be broken but the tests would still pass. It's not just the `EditProfileSchema` class but also the `_emails_must_match_validator()` function that has to get passed to its `__init__()`. I could factor this into a `_get_edit_profile_schema()` helper function (which `_validate_edit_profile_request()` would call) and then write tests against the object that it returns.

seanh:
I'm expecting landscape to have some issues with this (it might be useful to figure a way to get landscape's output locally: diff of the prospector output on this branch vs master).

seanh:
Rebased

seanh:
@tilgovi I think this is finished again, since Nick's on holiday do you want to review it?

seanh:
Just rebased this on master again

nickstenning:
LGTM. Merging.

dwhly:
:tada:

PR #2087, opened 2015-03-25
Mobile Support

RawKStar77:
Adds the ability to create new annotations on touch screen devices running iOS or Android. You can play around with it on: https://touch.dokku.hypothes.is/

This closes:
https://github.com/hypothesis/h/issues/2130
https://github.com/hypothesis/h/issues/587
https://github.com/hypothesis/vision/issues/167

Issue #587, opened 2013-08-07
Cannot select text with iPad or iPhone (Both Chrome and Safari)


RawKStar77:
Apple grabs the text selection event to produce its own context menu. Is there a way around this? People are going to want to annotate using their iPads. 

Other than text selection, everything else seems to work. We can login, read annotations, and even reply. 

This is what happens when text is selected:
![photo](https://f.cloud.github.com/assets/521978/928440/dfa0fba2-ffb6-11e2-97bf-384dda0f830b.PNG)

Could someone with an Android tablet explore whether or not the application works?

RawKStar77:
Closed by https://github.com/hypothesis/h/pull/2137

robertknight:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;zvmbTi4a&#x2F;328-spike-suggest-improvements-to-mobile-touch-support"><img src="https:&#x2F;&#x2F;github.trello.services&#x2F;images&#x2F;trello-icon.png" width="12" height="12"> [SPIKE] Suggest improvements to mobile&#x2F;touch support</a>

robertknight:
A better fix is coming in https://github.com/hypothesis/h/pull/3347

JakeHartnell:
> A better fix is coming in #3347

:tada:

PR #2137, opened 2015-04-08
Mobile support

RawKStar77:
Android and iOS devices can now create annotations. This work derives from my previous closed PR https://github.com/hypothesis/h/pull/2087, but doesn't include the UX work (which is in a separate PR). It has been tested this on the most recent versions of iOS and Android, and you can try it for yourself at http://mobile-support.dokku.hypothes.is/

@tilgovi, I think you will be happier with the code here. The events are now bound together, and we don't do user agent sniffing (except in one case the that is *only* relevant to mobile Safari).

Closes:
#587
hypothesis/vision#167

### DIFF
--- a/h/static/scripts/annotator/guest.coffee
+++ b/h/static/scripts/annotator/guest.coffee
@@ -196,19 +196,9 @@ module.exports = class Guest extends Annotator
   _setupDynamicStyle: -> this
   _setupViewer: -> this
   _setupEditor: -> this
-  _setupDocumentEvents: ->
-    $(document).bind({
-      # omit the "mouseup" check
-      "mousedown": this.checkForStartSelection
-    })
-    this
+  _setupDocumentEvents: -> this
 
   destroy: ->
-    $(document).unbind({
-      "mouseup":   this.checkForEndSelection
-      "mousedown": this.checkForStartSelection
-    })
-
     $('#annotator-dynamic-style').remove()
 
     @adder.remove()
@@ -262,13 +252,6 @@ module.exports = class Guest extends Annotator
       params: (a.$$tag for a in annotations)
 
   onAnchorMousedown: ->
-
-  checkForStartSelection: (event) =>
-    # Override to prevent Annotator choking when this ties to access the
-    # viewer but preserve the manipulation of the attribute `mouseIsDown` which
-    # is needed for preventing the panel from closing while annotating.
-    unless event and this.isAnnotator(event.target)
-      @mouseIsDown = true
 
   # This is called to create a target from a raw selection,
   # using selectors created by the registered selector creators
--- a/h/static/scripts/annotator/plugin/toolbar.coffee
+++ b/h/static/scripts/annotator/plugin/toolbar.coffee
@@ -1,15 +1,18 @@
 $ = Annotator.$
 
 makeButton = (item) ->
-  anchor = $('<a></a>')
+  anchor = $('<button></button>')
   .attr('href', '')
   .attr('title', item.title)
+  .attr('name', item.name)
   .on(item.on)
   .addClass(item.class)
   button = $('<li></li>').append(anchor)
   return button[0]
 
 class Annotator.Plugin.Toolbar extends Annotator.Plugin
+  HIDE_CLASS = 'annotator-hide'
+
   events:
     'setVisibleHighlights': 'onSetVisibleHighlights'
 
@@ -24,7 +27,8 @@ class Annotator.Plugin.Toolbar extends Annotator.Plugin
 
     items = [
       "title": "Toggle Sidebar"
-      "class": "annotator-toolbar-toggle h-icon-chevron-left"
+      "class": "h-icon-chevron-left"
+      "name": "sidebar-toggle"
       "on":
         "click": (event) =>
           event.preventDefault()
@@ -37,6 +41,7 @@ class Annotator.Plugin.Toolbar extends Annotator.Plugin
     ,
       "title": "Hide Highlights"
       "class": "h-icon-visibility"
+      "name": "highlight-visibility"
       "on":
         "click": (event) =>
           event.preventDefault()
@@ -46,6 +51,7 @@ class Annotator.Plugin.Toolbar extends Annotator.Plugin
     ,
       "title": "New Note"
       "class": "h-icon-insert-comment"
+      "name": "insert-comment"
       "on":
         "click": (event) =>
           event.preventDefault()
@@ -65,10 +71,12 @@ class Annotator.Plugin.Toolbar extends Annotator.Plugin
 
   onSetVisibleHighlights: (state) ->
     if state
-      $(@buttons[1]).children().removeClass('h-icon-visibility-off')
-      $(@buttons[1]).children().addClass('h-icon-visibility')
-      $(@buttons[1]).children().prop('title', 'Hide Highlights');
+      $('[name=highlight-visibility]')
+      .removeClass('h-icon-visibility-off')
+      .addClass('h-icon-visibility')
+      .prop('title', 'Hide Highlights');
     else
-      $(@buttons[1]).children().removeClass('h-icon-visibility')
-      $(@buttons[1]).children().addClass('h-icon-visibility-off')
-      $(@buttons[1]).children().prop('title', 'Show Highlights');
+      $('[name=highlight-visibility]')
+      .removeClass('h-icon-visibility')
+      .addClass('h-icon-visibility-off')
+      .prop('title', 'Show Highlights');
--- a/h/static/scripts/annotator/host.coffee
+++ b/h/static/scripts/annotator/host.coffee
@@ -64,19 +64,19 @@ module.exports = class Host extends Guest
     @frame.removeClass 'annotator-collapsed'
 
     if @toolbar?
-      @toolbar.find('.annotator-toolbar-toggle')
+      @toolbar.find('[name=sidebar-toggle]')
       .removeClass('h-icon-chevron-left')
       .addClass('h-icon-chevron-right')
 
   hideFrame: ->
-      @frame.css 'margin-left': ''
-      @frame.removeClass 'annotator-no-transition'
-      @frame.addClass 'annotator-collapsed'
+    @frame.css 'margin-left': ''
+    @frame.removeClass 'annotator-no-transition'
+    @frame.addClass 'annotator-collapsed'
 
-      if @toolbar?
-        @toolbar.find('.annotator-toolbar-toggle')
-        .removeClass('h-icon-chevron-right')
-        .addClass('h-icon-chevron-left')
+    if @toolbar?
+      @toolbar.find('[name=sidebar-toggle]')
+      .removeClass('h-icon-chevron-right')
+      .addClass('h-icon-chevron-left')
 
 
   _addCrossFrameListeners: ->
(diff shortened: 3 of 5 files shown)


h/static/scripts/annotator/guest.coffee
```diff
@@ -198,15 +201,37 @@ module.exports = class Guest extends Annotator
   _setupEditor: -> this
   _setupDocumentEvents: ->
     $(document).bind({
-      # omit the "mouseup" check
+      "touchstart": this.checkForStartSelection
       "mousedown": this.checkForStartSelection
     })
+
+    # This event is needed for iOS Safari only.
```
tilgovi: Does it break anything to add this for other browsers?

RawKStar77: No.

RawKStar77: Do you want me to remove the userAgent sniffing here then? : )

h/static/scripts/annotator/guest.coffee
```diff
@@ -298,7 +323,8 @@ module.exports = class Guest extends Annotator
     if immediate
       # Create an annotation
       @onAdderClick event
-    else
+    # We don't show the adder on mobile since it is not correctly placed and competes with native popups.
+    else if not ('ontouchstart' in window)
```
tilgovi: This should be `true` on any device that has a touch screen. For example, if I open this in Windows 8 in Chrome on my laptop I won't get an adder.

I think you should make two separate functions that you bind independently to each type of event. This method would no longer show the adder. Instead, both event handlers would call this method to extract the selection and only one of them would then show the adder.

h/static/scripts/annotator/plugin/toolbar.coffee
```diff
@@ -15,13 +15,24 @@ class Annotator.Plugin.Toolbar extends Annotator.Plugin
...
     '.annotator-toolbar mouseenter': 'show'
     '.annotator-toolbar mouseleave': 'hide'
-    'setTool': 'onSetTool'
     'setVisibleHighlights': 'onSetVisibleHighlights'
 
   html: '<div class="annotator-toolbar annotator-hide"></div>'
 
   pluginInit: ->
     @annotator.toolbar = @toolbar = $(@html)
+    # <UX-note>: If we detect a touch screen events, we use a different UX.
```
tilgovi: Why not just show it on all browsers?

h/static/scripts/annotator/plugin/toolbar.coffee
```diff
@@ -55,7 +66,17 @@ class Annotator.Plugin.Toolbar extends Annotator.Plugin
         "click": (event) =>
           event.preventDefault()
           event.stopPropagation()
-          @annotator.onAdderClick target: dataset: action: "comment"
```
tilgovi: You should either not change this line or follow the same convention in the highlight button. It's oddly inconsistent the way you've done it.

h/static/scripts/annotator/plugin/toolbar.coffee
```diff
@@ -63,6 +84,9 @@ class Annotator.Plugin.Toolbar extends Annotator.Plugin
     @buttons.appendTo(list)
     @toolbar.append(list)
 
+    # Hide highlight button.
+    $(@buttons[3]).hide()
```
tilgovi: This risks getting broken when we change the order of buttons. It's unnecessarily fragile. I would instead change the `class` key on the button definition above, in line 73, to have `"h-icon-border-color #{HIDE_CLASS}"` and define `HIDE_CLASS` at the top of the module to be `annotator-hide`. Then toggle that class on/off in the `showHighlightButton` method.

I understand that the other methods below here are using indexes into the buttons list in the same manner I've just recommended that you not, but let's try to make problems better not worse. In the future we could change the rest of the places we do this to use attributes to more explicitly filter for the right button.

h/static/scripts/annotator/plugin/toolbar.coffee
```diff
@@ -15,13 +15,24 @@ class Annotator.Plugin.Toolbar extends Annotator.Plugin
...
 
   pluginInit: ->
     @annotator.toolbar = @toolbar = $(@html)
+    # <UX-note>: If we detect a touch screen events, we use a different UX.
+    # This is because Mobile browsers often present their own popup menus in
+    # response to a user selection.
+    if 'ontouchstart' in window
+      # When there is a selection on touch devices show/hide highlighter button
+      # Check whether there is a user selection.
+      document.addEventListener "selectstart", =>
```
tilgovi: You need to extract this into a function you can keep a reference to so that it can be unbound in a `detroy` method.

h/static/scripts/annotator/guest.coffee
tilgovi: Yes, please.

h/static/scripts/annotator/guest.coffee
```diff
@@ -1,15 +1,18 @@
...
   SHOW_HIGHLIGHTS_CLASS = 'annotator-highlights-always-on'
 
   # Events to be bound on Annotator#element.
   events:
+    "setVisibleHighlights": "setVisibleHighlights"
     ".annotator-adder button click":     "onAdderClick"
     ".annotator-adder button mousedown": "onAdderMousedown"
     ".annotator-adder button mouseup":   "onAdderMouseup"
-    "setVisibleHighlights": "setVisibleHighlights"
+    "document onselectionchange": "userSelectionChanged"
```
tilgovi: I don't _think_ this works because our typical injection target is the body and `document` is not a selector underneath the body, it's above it. You will have to do this in `_setupDocumentEvents` and then unbind it yourself in `destroy`.

h/static/scripts/annotator/guest.coffee
```diff
@@ -198,15 +201,37 @@ module.exports = class Guest extends Annotator
...
       "mousedown": this.checkForStartSelection
     })
+
+    # This event is needed for iOS Safari only. 
+    if /iPhone|iPad|iPod/i.test(navigator.userAgent)
+      $(window).bind 'selectionEnd', ->
+        # reset selection timeout
+        selectionEndTimeout = null
+        # get user selection
+        selectedText = getSelectionText()
```
tilgovi: This doesn't appear to do anything because `selectedText` isn't defined outside this block or used anywhere else?

h/static/scripts/annotator/guest.coffee
```diff
@@ -198,15 +201,37 @@ module.exports = class Guest extends Annotator
...
+        return
     this
 
+  # Needed to get text selection working on iOS. Only used for touch screen devices.
+  userSelectionChanged: ->
+    # Wait 500 ms after the last selection change event to capture 
+    if selectionEndTimeout
+      clearTimeout selectionEndTimeout
+    selectionEndTimeout = setTimeout((->
+      Util.getGlobal().trigger 'selectionEnd'
```
tilgovi: It seems strange to trigger a custom event here only to catch it in exactly one other known place that's right above here in the code. Did you perhaps copy this from the answer http://stackoverflow.com/a/22787135/945546 has? I would just inline the code. Of course, that's assuming the code does anything (see comment on #L214)

RawKStar77:
@tilgovi thanks for your help in reviewing this. I was able to remove a bunch of code! I just tested on iOS 8.1 and Android and everything is still working. : )

h/static/scripts/annotator/guest.coffee
```diff
@@ -298,6 +302,7 @@ module.exports = class Guest extends Annotator
     if immediate
       # Create an annotation
       @onAdderClick event
+    # We don't show the adder on mobile since it is not correctly placed and competes with native popups.
```
tilgovi: This comment can be removed now.

h/static/scripts/annotator/plugin/toolbar.coffee
```diff
@@ -36,7 +39,7 @@ class Annotator.Plugin.Toolbar extends Annotator.Plugin
             @annotator.triggerHideFrame()
     ,
       "title": "Hide Highlights"
-      "class": "h-icon-visibility"
+      "class": "annotator-highlight-visibility h-icon-visibility"
```
tilgovi: Consistency in class names would be appreciated. You've got `annotator-highlight-visibility` here but `annotator-toolbar-highlight-btn` below. I'd also prefer if you used a data attribute for these things since I don't really like making code semantics depend on CSS classes. As a designer, one should be able to change CSS classes to their liking without worrying it will break code.

tilgovi: That might mean adding another key to each of these button definitions that you use to set the data attribute in `makeButton`.

tilgovi:
Reviewing this more, do you think we should just enable/disable the button rather than show/hide it? We literally just merged your PR which made the other buttons visible all the time.

tilgovi:
I mean, there's a difference of course, because the other buttons are applicable always, but I wonder if it makes sense to just leave it disabled rather than hiding it. Otherwise, I'd want to re-introduce some animation here. The appearance/disappearance also couples this strangely to the bucket bar, a coupling I'm eliminating in my code editor as I type this.

RawKStar77:
> The appearance/disappearance also couples this strangely to the bucket bar, a coupling I'm eliminating in my code editor as I type this.

Another alternative is that we could not support highlighting on mobile *for the time being* until we design a better way (experimenting with a disabled logo or perhaps tap to highlight). Making annotations is more important I think, and there are other priorities at the moment than mobile. But I'm glad I can make annotations on my phone now and that users with iPads will be able to as well. It's a start, so for the purpose of the PR I can remove the highlight button and we solve that problem at a later time.

Thoughts?

tilgovi:
I'm into that.

RawKStar77:
@tilgovi anything else?

judell:
"Making annotations is more important I think"

+1 and thanks for this. Enabling iPads to annotate is a huge win.

h/static/scripts/annotator/plugin/toolbar.coffee
```diff
@@ -24,7 +27,8 @@ class Annotator.Plugin.Toolbar extends Annotator.Plugin
 
     items = [
       "title": "Toggle Sidebar"
-      "class": "annotator-toolbar-toggle h-icon-chevron-left"
+      "class": "h-icon-chevron-left"
+      "data_attribute": "data-toolbartoogle"
```
tilgovi: I would make this `camelCase`. Or just call it `attr` or whatever.

h/static/scripts/annotator/plugin/bucket-bar.coffee
```diff
@@ -4,7 +4,7 @@ $ = Annotator.$
 
 class Annotator.Plugin.BucketBar extends Annotator.Plugin
   # prototype constants
-  BUCKET_THRESHOLD_PAD: 30
+  BUCKET_THRESHOLD_PAD: 105
```
tilgovi: I just myself made this 106 on master. Could you just rebase on that and this PR doesn't need to concern itself.

tilgovi:
Just a couple tiny nitpicks. Then please rebase and squash and let's merge it.

tilgovi:
I'm on it.

tilgovi:
- removed some obsolete stuff around annotator event states that allow us to remove all the selection stuff events from guest
- switched to using the `name` attribute on the buttons and made them actual button elements
- corrected typo in data-toolbartoogle (hehe) and made it sidebar-toggle to reflect what it does better

Thanks, @RawKStar77

PR #2139, opened 2015-04-08
Analytics

tilgovi:
This pull request adds some basic Google Analytics support.

- Server-rendered pages track page views
- Claim invitation emails send campaign code
- Single-page application tracks page views using angulartics (luisfarzati/angulartics)

Angulartics can work with any number of providers so the front-end is actually enabled for various analytics providers with a little tweaking. We can start to add support for logging analytics events from various UI actions, too, if we desire. See documentation for angulartics.

### DIFF
--- a/h/assets.yaml
+++ b/h/assets.yaml
@@ -37,6 +37,17 @@ angular_toastr:
       contents: h:static/scripts/vendor/angular-toastr/angular-toastr.css
 
 
+# Analytics
+angulartics:
+  filters: uglifyjs
+  output: scripts/vendor/angulartics.min.js
+  contents: h:static/scripts/vendor/angulartics.js
+angulartics_ga:
+  filters: uglifyjs
+  output: scripts/vendor/angulartics-ga.min.js
+  contents: h:static/scripts/vendor/angulartics-ga.js
+
+
 # jQuery
 jquery:
   filters: uglifyjs
@@ -100,6 +111,8 @@ app:
     - angular_sanitize
     - angular_tags_input
     - angular_toastr
+    - angulartics
+    - angulartics_ga
     - autofill
     - bind
     - katex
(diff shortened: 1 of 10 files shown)


tilgovi:
Need to figure out why this breaks the top-level auth pages (/login, /register), etc.

judell:
"We can start to add support for logging analytics events from various UI actions, too, if we desire."

That's huge, Can/should all this be conditional on a feature flag?

Probably related:

"Don't specify any property tracking id in your config/environment and it won't be included."

 Is that, or can/should that, be a parameter to embed.js, so usable from bookmarklets etc?

tilgovi:
> That's huge, Can/should all this be conditional on a feature flag?

I thought about it, but decided it offered no benefit. See below.

>
> Probably related:
>
> "Don't specify any property tracking id in your config/environment and it
won't be included."
>
> Is that, or can/should that, be a parameter to embed.js, so usable from
bookmarklets etc?
>

Hmm. Seems like you're suggesting that a publisher might want analytics on
how their embed is being used.

That seems reasonable.

It's not what I'm trying to address here, though. I'm trying to see how our
application is being used wherever it's embedded, regardless by whom or how.

Let's maybe punt on that until we start getting questions about it.

tilgovi:
I said "see below" then forgot to say what I thought I was going to.

You need to specify a google analytics property ID in your configuration
for it to be enabled. Adding another environment variable to enable it
seemed pointless when it can be so easily inferred.

tilgovi:
This is now good to go. It's not capable of tracking behavior _within_ the extension yet, but I'd like to merge this and then send out our first invites.

dwhly:
+1

judell:
Fantastic! Thank you, Randall.

tilgovi:
Alright, then. This is going live today.

PR #2143, opened 2015-04-08
Introduce _splitTerm() for searchFilter service

### LINKED ISSUE #2142: Stream: We split search expressions with the : characters
Seeing this both the current main branch and on our prod.

1. Open the stream
2. Open the Network tab to check requests/responses
3. Clear all previous network traffic and set it to monitor every new.
4. Type `uri:http://localhost:5000` (or any else valid uri) into the search bar and search for that uri.
5. Check the network tab
6. Observe the one new search request sent to our search api and check the `Request URL` parameter.

Excepted result: the parameter is rightly set. I.e. `http://localhost:5000/api/search?limit=10&uri=http:%2F%2Flocalhost:5000%2F`

Actual result: It is set to `http://localhost:5000/api/search?limit=10&uri=http` that means we give back all annotations and only filter them client side.

gergely-ujvari:
Regression fix.

Moved out from _tokenize() as a separate function to split
a search term properly into [filter, term] because toObject()
must call this too.

Fix #2142

### DIFF
--- a/h/static/scripts/search-filter.coffee
+++ b/h/static/scripts/search-filter.coffee
@@ -2,6 +2,22 @@
 # It expects a search query string where the search term are separated by space character
 # and collects them into the given term arrays
 module.exports = class SearchFilter
+  # Splits a search term into filter and data
+  # i.e.
+  #   'user:johndoe' -> ['user', 'johndoe']
+  #   'example:text' -> [null, 'example:text']
+  _splitTerm: (term) ->
+    filter = term.slice 0, term.indexOf ":"
+    unless filter?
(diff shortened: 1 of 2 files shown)


tilgovi:
Looks great. I would suggest that some test case should have both valid and invalid keys and ensure that some are split but others aren't. I notice you've got "other" but it's never a thing with a colon except in the last case, where there are no other tokens.

gergely-ujvari:
@tilgovi: Added a test case for that.

tilgovi:
Nice. Thanks!

PR #2156, opened 2015-04-10
Fix IE problems

### LINKED ISSUE #2155: IE11 annotations are not loaded
With my recent testing with IE11 it seems that if you open the welcome page with it, then no annotations are loaded.

The cause is that the `crossframe.providers` array is empty, which causes zero search requests for the backend. It seems there is an issue with the discovery/crossframe registration.

This not affect the standalone nor the stream page.

gergely-ujvari:
This PR has to components:
- Reintroduces conditional polyfill, ES6Promise, because IE doesn't have native support for it.
- Upgrades angularjs from 1.2.25 to 1.2.28 (just bugfixing changes) because with 1.2.25 and the polyfill in place IE throws exceptions. According to the [changelog](https://github.com/angular/angular.js/blob/master/CHANGELOG.md) angularjs 1.2.27 introduces many IE related fixes and also solves our problem.

With that annotations are loaded into the sidebar with IE too.

Fix #2155 

All the tests pass, and as I see we do not have tests for polyfills so I haven't added one either.

### DIFF
--- a/h/assets.yaml
+++ b/h/assets.yaml
@@ -71,6 +71,10 @@ bind:
   filters: uglifyjs
   output: scripts/vendor/polyfills/bind.min.js
   contents: h:static/scripts/vendor/polyfills/bind.js
+promise:
+  filters: uglifyjs
+  output: scripts/vendor/polyfills/promise.min.js
+  contents: h:static/scripts/vendor/polyfills/promise.js
 wgxpath:
   filters: uglifyjs
   output: scripts/vendor/polyfills/wgxpath.install.min.js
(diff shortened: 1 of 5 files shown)


tilgovi:
I think I'll merge this as is, but we should reassess how we're using polyfills soon. I might rather us use `require` to get a handle to the promise directly rather than executing the global polyfill. Similarly, with `URL`. The wgxpath lib is bigger, so it's more important that it not be included via `require()` but there might be a good way to externalize it. We'll see!

tilgovi:
Thanks, @gergely-ujvari.

judell:
Outstanding, thanks!

PR #2157, opened 2015-04-10
Add a blocklist of sites we don't launch on

seanh:
Add a blocklist of sites we don't launch on
The blocklist is configured in production.ini like so:

    h.blocklist = {
        "seanh.cc": {},
        "finance.yahoo.com": {},
        "twitter.com": {},
        "*.google.com"
    }

The Chrome extension handles the blocklist in the same way that it
handles other launch errors (e.g. when the user hasn't given the
extension permission to annotate file:/// URLs): it adds a (!) error
badge to the "browser action" button, if the button is clicked again it
opens a "help" page saying that the site isn't supported yet.

It doesn't seem to be possible to pass data into the Chrome extension's
help pages (as far as I can tell) so the error is just "Hypothesis
doesn't work on this page yet" and not e.g. "Hypothesis doesn't work on
twitter.com".

The bookmarklet handles the blocklist in the same way that it handles
other launch errors (e.g. trying to annotate a PDF when PDF.js isn't
available): it uses window.alert() to show an error.

### DIFF
--- a/h/browser/chrome/lib/sidebar-injector.js
+++ b/h/browser/chrome/lib/sidebar-injector.js
@@ -27,6 +27,20 @@
       throw new TypeError('isAllowedFileSchemeAccess must be a function');
     }
 
+    /* Return a Promise whose value is the blocklist from the blocklist.json
+     * file that's packaged with the Chrome extension, as an object.
+     */
+    function loadBlocklist() {
+      return new Promise(function(resolve, reject) {
+        var request = new XMLHttpRequest();
+        request.onload = function() {
+          resolve(JSON.parse(this.responseText));
+        };
+        request.open('GET', '/blocklist.json');
+        request.send(null);
+      });
+    }
+
     /* Injects the Hypothesis sidebar into the tab provided.
      *
      * tab - A tab object representing the tab to insert the sidebar into.
@@ -35,11 +49,19 @@
      * otherwise it will be rejected with an error.
      */
     this.injectIntoTab = function (tab) {
-      if (isFileURL(tab.url)) {
-        return injectIntoLocalDocument(tab);
-      } else {
-        return injectIntoRemoteDocument(tab);
-      }
+      return loadBlocklist().then(function(blocklist) {
+        if (h.blocklist.isBlocked(tab.url, blocklist)) {
+          return Promise.reject(
+            new h.BlockedSiteError(
+              "Hypothesis doesn't work on this site yet."));
+        }
+
+        if (isFileURL(tab.url)) {
+          return injectIntoLocalDocument(tab);
+        } else {
+          return injectIntoRemoteDocument(tab);
+        }
+      });
     };
 
     /* Removes the Hypothesis sidebar from the tab provided.
--- a/h/browser/chrome/help/index.html
+++ b/h/browser/chrome/help/index.html
@@ -101,6 +101,10 @@
         <h1 class="help-item__heading">We’re sorry, Hypothesis doesn’t work on this page…</h1>
         <p class="center">This extension can only be used on pages served over HTTP/HTTPS and FTP.</p>
       </section>
+      <section id="blocked-site" class="help-item">
+        <img class="help-item__icon" src="sad-annotation.svg" />
+        <h1 class="help-item__heading">We’re sorry, Hypothesis doesn't work on this site yet.</h1>
+      </section>
     </article>
     <script src="./index.js"></script>
   </body>
--- a/h/browser/chrome/karma.config.js
+++ b/h/browser/chrome/karma.config.js
@@ -27,7 +27,8 @@ module.exports = function(config) {
       'lib/browser-action.js',
       'lib/help-page.js',
       'lib/hypothesis-chrome-extension.js',
-      'test/*-test.js'
+      'test/*-test.js',
+      '../../static/scripts/blocklist.js'
     ],
 
 
--- a/h/browser/chrome/lib/errors.js
+++ b/h/browser/chrome/lib/errors.js
@@ -23,9 +23,16 @@
   }
   RestrictedProtocolError.prototype = Object.create(ExtensionError);
 
+  function BlockedSiteError(message) {
+    Error.apply(this, arguments);
+    this.message = message;
+  }
+  BlockedSiteError.prototype = Object.create(ExtensionError);
+
   h.ExtensionError = ExtensionError;
   h.LocalFileError = LocalFileError;
   h.NoFileAccessError = NoFileAccessError;
   h.RestrictedProtocolError = RestrictedProtocolError;
+  h.BlockedSiteError = BlockedSiteError;
 
 })(window.h || (window.h = {}));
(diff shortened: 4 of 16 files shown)


seanh:
@judell wanted to make the blocklist editable by users without doing a deploy. That's not going to be included in this initial PR, but if we did do it I'd say store the blocklist in the db instead of the config file and add a simple API for viewing the list of blocked sites, blocking a site, and unblocking a site.

seanh:
What do people think about tests for this? I can certainly write tests for the blocklist config setting validation in Python. But tests for `embed.js`?

I think the two things remaining to be done here are tests and replacing `window.alert()`

seanh:
> Biggest problem is using window.alert() is ugly. But how to make this better? Since we're deliberately
not injecting our Angular app. Possible solution: inject a different Angular app instead! A really simple standalone one that just requires toastr and displays one message.

Err ... Or just call toastr direct from js without angular (facepalm smiley)

judell:
Wildcards?

We will almost certainly have to say *.google.com for example.

seanh:
Wildcards can be done fairly easy. I'm having more trouble getting a nice way of showing the notification working

seanh:
The toast notification in commit <https://github.com/hypothesis/h/commit/664d70887d640b07a51c4b05e9f1f4abc71144a2> is working fine with the bookmarklet, but doesn't work with the Chrome extension on any site (https and csp or not): Uncaught ReferenceError: toastr is not defined at embed.js line 49.

There's no other error (e.g from the browser refusing to inject the script). If I put a breakpoint right before the line where it crashes and inspect the HTML, the `<script>` for toastr _has_ been inserted in the `<head>` (though I guess that doesn't necessarily mean the script has been executed).

I tried vendoring the files as well. So now when I inspect the HTML in the debugger I can see a `<script>` for `chrome-extension://oldbkmekfdjiffgkconlamcngmkioffd/public/scripts/vendor/toastr.min.js` has been injected, and if I go to that URL I do see the js file. Same error though: toastr is undefined.

csillag:
> Uncaught ReferenceError: toastr is not defined at embed.js line 49.
> [...] 
>  the `<script>` for toastr _has_ been inserted in the `<head>`

I haven't really tried to debug this, just a question: you are surely aware that there are two different spaces here? Part of the code of the chrome extension runs in a different scope, separate from the HTML page. That could explain not finding the libraries you wanted to load. (But of course it's also possible that all the code you are talking about is already running inside the page; as I said, I haven't looked too closely.)

seanh:
@csillag Thanks, I think you're onto it. Not tested a solution yet but I think the problem is probably that when injected by the chrome extension (sidebar-inject.jr does it with chrome.tabs.executeScript()) embed.js runs as a _content script_ and apparently content scripts cannot "Use variables or functions defined by web pages or by other content scripts". <https://developer.chrome.com/extensions/content_scripts>

So that would explain it, and why the same code works fine when injected by the bookmarklet.

Basically I don't know much about Chrome extensions (never done one before), thinking this was a problem that should be solvable quickly I just kept going at it for half a day or more, and never bothered to step back and go away and learn about Chrome extensions then come back. Rookie mistake.

I'll come back to this when I've figured a workaround, then we can test whether this was in fact the problem

tilgovi:
The content script can't really "inject" code, it can add a script tag. That script tag is run in the same context as any others, not the content script context.

seanh:
@tilgovi Yep :) Thanks. I'm also rewriting this so that the Chrome extension will do what it currently does for other kinds of unsupported pages (e.g. when it doesn't have permission to annotate a `file:///` URL): shows an errored badge on the button, if you click it again opens a help page in a new tab. And the bookmarklet will do what _it_ currently does for other errors, which is use the browser's `alert()` function.

seanh:
Just force pushed this with a completely new version, and edited the first comment at the top of this page.

seanh:
untitled screencast - cropped

seanh:
What does this still need?

- [ ] Feature flag? Seems unnecessary to me, if you've no blocklist setting in your ini nothing will be blocked
- [ ] Tests
- [ ] Wildcard support (regexes is probably easiest)

seanh:
Think this can be reviewed now

Thoughts:

- Is the button badge then help page the right thing to do here? We may be able to do something nicer (Chrome notifications API maybe? Or a popup window on the button?) but maybe that belongs in a separate PR that changes this behaviour for the other errors we already handle as well

- No tests for the bookmarklet

- Had to add `h.parseUrl` to make the function testable. Also there may be a better place for this function.

seanh:
Oh, still needs wildcard support

h/browser/chrome/lib/sidebar-injector.js
```diff
@@ -1,6 +1,40 @@
...
   'use strict';
 
+  /* Parse the given URL and return an object with its different components.
+   *
+   * Any or all of the components returned may be undefined.
+   * For example for the URL "http://twitter.com" port, path query and anchor
+   * will be undefined.
+   *
+   */
+  function parseUrl(url) {
```
tilgovi: https://developer.mozilla.org/en-US/docs/Web/API/URL

seanh: Yeah, I did use URL originally but it didn't seem to work in PhantomJS

seanh:
The search for current URL in blocklist code is duplicated in two places currently, would be nice if we could move it into somewhere shared. One of the two places where it's duplicated is `embed.js` though

seanh:
Oops, got a `.only(` in there

seanh:
The biggest remaining problem I have with this pr is this function, in the Chrome extension:

```javascript
    /* Return true if the given url is blocked by the given blocklist. */
    function isBlocked(url, blocklist) {
      var regexSpecialChars = '^$.+?=|\/()[]{}'; //  '*' deliberately omitted.
      for (var pattern in blocklist) {
        if (blocklist.hasOwnProperty(pattern)) {
          // Escape regular expression special characters.
          for (var i = 0; i < regexSpecialChars.length; i++) {
            var c = regexSpecialChars.charAt(i);
            pattern = pattern.replace(c, '\\' + c);
          }

          // Turn * into .* to enable simple patterns like "*.google.com".
          pattern = pattern.replace('*', '.*');

          // Blocklist patterns should match from the start of the URL.
          // This means that "google.com" will _not_ match "mail.google.com",
          // for example. (But "*.google.com" will match it.)
          pattern = '^' + pattern;

          if (url.match(pattern)) {
            return true;
          }
        }
      }
      return false;
    }
```

That's fairly complicated and it's duplicated (without wildcard support yet) in `embed.js` for the bookmarklet. How can I share this function between the Chrome extension and the bookmarklet?

seanh:
@tilgovi @nickstenning The last commit introduces some test fails that I don't understand. Any clues?

The commit gets rid of some warnings printed in the test output. Each time it makes an XMLHttpRequest for blocklist.json that fails it prints out a warning. It's actually fine - the code can handle this and the tests pass. But messes up the test runner output. So I just made blocklist.json available to all the tests. But for some reason this breaks a few.

seanh:
Tests should be passing now, but still need to get the wildcard support in the bookmarklet..

seanh:
That's the wildcard support done and duplication removed, and rebased. Ready for review!

tilgovi:
Got a lint failure here.

tilgovi:
ftfy

Issue #1688, opened 2014-11-14
Prevent double embedding


BigBlueHat:
Pages like this [JEP post about Hypothes.is](http://quod.lib.umich.edu/j/jep/3336451.0017.201?view=text;rgn=main) include our `embed.js` by default. If you then turn on the Chrome extension or use the bookmarklet, you'll end up with two Hypothes.is (the Firefox extension currently has a different bug...which I'll post shortly).

Previous conversation around this issue can be found here:
https://github.com/hypothesis/h/commit/f80109b430e18ef9bf65a53d21be72e194fd6720#commitcomment-8561557

csillag:
Solved in https://github.com/hypothesis/h/pull/2166. Moved this do "Done".

csillag:
> Solved in https://github.com/hypothesis/h/pull/2166. Moved this do "Done".

Should we also close this?

tilgovi:
I'm pushing to stage and going to give a quick test against extension and via.

tilgovi:
The double embedding is indeed fixed.

judell:
Hallelujah, and thanks.

PR #2166, opened 2015-04-16
Re-add protection against double deployment

csillag:
This seems to do the trick for https://github.com/hypothesis/h/issues/1688.

### DIFF
--- a/h/static/scripts/hypothesis.js
+++ b/h/static/scripts/hypothesis.js
@@ -1,5 +1,5 @@
 if (window.annotator) {
-    // Not deploying Hypothes.is, in order to avoid double deployment
+    // Not deploying app, in order to avoid double deployment
     return;
 }
 
--- a/h/static/scripts/hypothesis.js
+++ b/h/static/scripts/hypothesis.js
@@ -1,3 +1,8 @@
+if (window.annotator) {
+    console.warn("Not deploying Hypothes.is, in order to avoid double deployment.");
+    return;
+}
+
 var Annotator = require('annotator');
 
 // Monkeypatch annotator!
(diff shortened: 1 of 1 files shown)


csillag:
Tested in the following environments:
 * Activating the bookmarklet on welcome page. (Used to be broken, this fixes it.)
 * Activating the bookmarklet several times on a normal page. (Used to be broken, this fixes it.)
 * Activating the extension on a page with H embedded. (Used to be broken, this fixes it.)

  * * *

There are various points in the deployment code-path where this could be injected; as far as I can tell, the current position is able to catch all deployment methods, but if you have a different preference, let me know...

dwhly:
Can we also test activating the extension and bookmarklet on a via page?

tilgovi:
@dwhly all the activations follow this same code path in the same context. This should work fine.

h/static/scripts/hypothesis.js
```diff
@@ -1,3 +1,8 @@
+if (window.annotator) {
+    console.warn("Not deploying Hypothes.is, in order to avoid double deployment.");
```
tilgovi: I think we should remove this log. It's not likely to be seen by or help a user.

tilgovi: We also just got rejected from the Mozilla add-on review for logging messages to the console.

tilgovi: "Your add-on prints debugging information to the Console, which is generally not allowed in production add-ons."

Let's default to quiet and only log things when we have a very strong reason to believe it will help someone.

csillag:
Removed log line.

h/static/scripts/hypothesis.js
```diff
@@ -1,3 +1,8 @@
+if (window.annotator) {
+    // Not deploying Hypothes.is, in order to avoid double deployment
```
tilgovi: s/Hypothes.is/Hypothesis/ if you must mention the product in the comment.

csillag: Replaced product name with the more generic "app".

tilgovi:
Cheers, @csillag.

RawKStar77:
:cake:

judell:
Excellent.

PR #2168, opened 2015-04-16
Get Promise explicitly through requires

tilgovi:
This eliminates a lot of polyfill cruft. Also, we had two different
promise polyfills in the vendor directory (O.o).


nickstenning:
Looks great, but the tests say you've missed the extension.

tilgovi:
I think that should do it.

Issue #2129, opened 2015-04-03
Annotations fail to anchor on chrome


dwhly:
On chrome, go here:
https://via.hypothes.is/h/http://arstechnica.com/gaming/2015/04/valve-software-co-founder-gabe-newell-becomes-a-video-game-character/

No annotations.  Even thought the stream shows some here:
https://hypothes.is/stream?q=uri:http:%2F%2Farstechnica.com%2Fgaming%2F2015%2F04%2Fvalve-software-co-founder-gabe-newell-becomes-a-video-game-character%2F

On safari, the annotations show.  Tried OSX / Safari 8 via Saucelabs.

RawKStar77:
A possibly important detail: these annotations were made with Safari.

nickstenning:
I don't see annotations at either link.

dwhly:
@RawKStar77 Can you and I try to repro this again?

judell:
See also: 

http://www.a2gov.org/services/Pages/OpenCityHall.aspx?pd_url=http%3a//www.peakdemocracy.com/portals/116/forum_home?admin_mode%3Doff#peak_democracy

Compare:

1. What the extension sees

2. https://hypothes.is/stream?q=uri:http:%2F%2Fwww.a2gov.org%2Fservices%2FPages%2FOpenCityHall.aspx%3Fpd_url%3Dhttp%253a%2F%2Fwww.peakdemocracy.com%2Fportals%2F116%2Fforum_home%3Fadmin_mode%253Doff%23peak_democracy

3. https://hypothes.is/api/search?uri=http%3A%2F%2Fwww.a2gov.org%2Fservices%2FPages%2FOpenCityHall.aspx%3Fpd_url%3Dhttp%253a%2F%2Fwww.peakdemocracy.com%2Fportals%2F116%2Fforum_home%3Fadmin_mode%253Doff%23peak_democracy

4. # of annotations displayed in bucket (1) vs seen in extension (3)

gergely-ujvari:
We should separate the two phenomena, as they're separate issues, namely the failed anchoring and the problems with our search api. (Stream uses our search api for loading the initial annotations so there should be no difference between those two).

gergely-ujvari:
I checked @judell's link to a2gov and found out the following:

[...]

PR #2169, opened 2015-04-16
Uri indexing - remove end hashmark

gergely-ujvari:
Adding a new pattern to capture for the uri analyzer: whole uri without the hashmark from the end

This'll fix a great deal of #2129 
Reindexing will be needed.


tilgovi:
What am I missing? Why isn't this problem already handled by the search and index settings we have, both of which pull out everything betwen // and the hash as a token.

tilgovi:
We already capture the URL without the fragment or the scheme in both the index and search patterns. So, I don't think this helps. My latest diagnosis is that the recent problems we've seen were a simple matter of not reindexing. See https://github.com/hypothesis/h/issues/2129#issuecomment-93869450

PR #2170, opened 2015-04-16
Center icons in toolbar buttons.

RawKStar77:
This is a really simple PR. The icons in the toolbar were slightly off center, this centers them.

Before:
selection_047

After:
selection_048

### DIFF
--- a/h/static/styles/inject.scss
+++ b/h/static/styles/inject.scss
@@ -162,7 +162,6 @@ $base-font-size: 14px;
   margin-top: 5px;
   overflow: hidden;
   position: relative;
-  text-align: center;
   height: 30px;
   width: 30px;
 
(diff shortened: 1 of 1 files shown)


dwhly:
ty!

tilgovi:
Hmmm.. I might merge it just because it's so simple but something is off. These are supposed to be text-align center already, so this shouldn't be necessary (the whole absolute positioning of these should be unnecessary).

RawKStar77:
> Hmmm.. I might merge it just because it's so simple but something is off. These are supposed to be text-align center already, so this shouldn't be necessary (the whole absolute positioning of these should be unnecessary).

You're probably right. We can rewrite it later. : )

tilgovi:
Or right now :).

The 1px error was coming from the default padding on the button. `padding: 0` fixed it.

There was also a 1px vertical error, coming from some combination of strange things having to do with positioning, line height, and whatever. Removing the explicit positioning and the line height seems to leave it in top shape.

PR #2172, opened 2015-04-17
[wip] Add jscs JavaScript linting to `make lint`

seanh:

PR #2173, opened 2015-04-17
[wip] Add jshint to `make lint`

seanh:


seanh:
Lots of jshint errors: <https://github.com/hypothesis/h/pull/2173> I think some of the globals and environment options I added in jshintrc are too broad as well, they should only apply to some of the files not all.

nickstenning:
I'm not going to advocate doing a massive cleanup on these just to get the tests passing. In my experience that's a great way to break stuff.

For this and #2172, I think we should look into a way of computing lint/health/etc, without necessarily failing the build.

nickstenning:
For example, we can configure [Landscape](https://landscape.io) to comment on pull requests that make things worse. I presume that we can do the same for JavaScript somehow.

nickstenning:
One option worth looking into: https://houndci.com/

nickstenning:
Closing in favour of #2246.

tilgovi:
Keep the branch around or delete?

seanh:
Deleted. You can restore pull request branches on GitHub, forever as far as I can see.

PR #2183, opened 2015-04-19
Remove browserify-pipe

tilgovi:
This is totally safe unless there is another input filter before
browserify, but let's just say that's not allowed because it
simplifies things nicely to not have this extra script.


tilgovi:
For @nickstenning

tilgovi:
I'm going with this. As an input filter, browserify should be acting on each file independently. The only issue would be if another filter was wrapped inside browserify.

I think that's got to be considered incorrect usage. If you need to filter things at that point, the browserify transform stack is the place to do it.

nickstenning:
Retroactive :+1:

Issue #2095, opened 2015-03-26
Normalize URL syntax and query behavior across /api and /stream


judell:
NOW:

https://hypothes.is/stream?q=tag:%27h-examples%27 -> 1 result
https://hypothes.is/api/search?tags=%27h-examples%27 -> many results

IDEAL:

https://hypothes.is/stream?q=tag:%27h-examples%27 -> 1 result
https://hypothes.is/api/search?q=tag:%27h-examples%27 -> 1 result

or (i..e. pick either style, just be consistent)

https://hypothes.is/stream?tags=%27h-examples%27 -> 1 result
https://hypothes.is/api/search?tags=%27h-examples%27 -> 1 result

Also it's unclear whether or in which context compound queries work, e.g.:

?user=dwhly&tags=w3c

?user=dwhly&tags=w3c,spec

Again, ideally same pattern as between /stream and /api

judell:
IRC conversation from 3/27/2015: https://gist.github.com/judell/bd5c2a6374dba27338e4

tilgovi:
Confirmed fixed on stage.

judell:
I'm not seeing a difference between stage and production for the above comparison:

https://hypothes.is/stream?q=tag:%27h-examples%27 -> 1 result
https://hypothes.is/api/search?tags=%27h-examples%27 -> many results

Formerly the production stream matched only the whole tag, the production api matched tag fragments.

Now the production stream and api match tag fragments. As do stage stream and api. So I'm missing what it is that we are confirming here.

judell:
#2153 as a reminder that we need user feedback about partial-match search.

tilgovi:
@judell I think you just caught this in the minute I was cutting stage back over to the standard index alias and doing the final reindexing and cut-over for prod. Prod is up now and prod and stage both show 14 annotations for those URLs for me.

Sorry about that.

judell:
No worries. As advertised, things that were inconsistent are now consistent. Thanks!

Issue #2181, opened 2015-04-18
Stream uri: search gives different results for urls terminated by /


dwhly:
Results differ for:
https://hypothes.is/stream?q=uri:hypothes.is

And

https://hypothes.is/stream?q=uri:hypothes.is/

tilgovi:
I've determined that the real issue here is ambiguity due to still not having clear user stories for the stream.

The same analysis cannot serve the needs finding all annotations on a site and all annotations on a page. The issue here only demonstrates the most obvious ambiguity wherein it's not clear whether a user means to find all the annotations on a home / root page or all annotations on a site / domain.

#2184 introduces new analyzers and field mappings that will enable the stream to have a "site" facet in addition (or in place of) "uri".

tilgovi:
Confirmed fixed on stage just now.

judell:
These match:

https://stage.hypothes.is/stream?q=uri:hypothes.is

vs

https://stage.hypothes.is/stream?q=uri:hypothes.is/

Just checking/confirming that, as per https://github.com/hypothesis/h/pull/2184#issuecomment-98879105, this:

https://stage.hypothes.is/stream?q=site:hypothes.is

is (while a good idea) not part of the current batch of changes.

tilgovi:
Correct.

dwhly:
:fireworks:

judell:
Still an issue. https://github.com/hypothesis/h/issues/2217

nickstenning:
For posterity, this and #2217 are different bugs. This is fixed.

Issue #2180, opened 2015-04-18
Stream uri: search gives different results for urls prefixed with http://


dwhly:
Results differ for:

https://hypothes.is/stream?q=uri:http:%2F%2Fwww.theatlantic.com%2Ftechnology%2Farchive%2F2014%2F08%2Fadvertising-is-the-internets-original-sin%2F376041%2F

and 

https://hypothes.is/stream?q=uri:www.theatlantic.com%2Ftechnology%2Farchive%2F2014%2F08%2Fadvertising-is-the-internets-original-sin%2F376041%2F

tilgovi:
Open PR for this at #2184

tilgovi:
Cannot confirm fix on stage.

tilgovi:
Possibly diagnosis: the "document" type doesn't use the new analysis so the URI alias expansion doesn't work correctly.

tilgovi:
Issue as described is fixed, so I'm opening another issue for the new problem.

tilgovi:
#2213

Issue #2141, opened 2015-04-08
Stream: uri escaping problems


gergely-ujvari:
Coming from #2129 

Problem: Stream does not give back search results for an uri which has annotations.

Hunch: The specified uri already has html escaped parts and double escape happens.

To reproduce:
uri: http://www.a2gov.org/services/Pages/OpenCityHall.aspx?pd_url=http%3a//www.peakdemocracy.com/portals/116/forum_home?admin_mode%3Doff#peak_democracy
(notice that it already has some html escaping)

1. Open the uri, enable the extension
2. Observe that the sidebar has three public annotations
3. Confirm the same with our search api: https://hypothes.is/api/search?uri=http://www.a2gov.org/services/Pages/OpenCityHall.aspx?pd_url=http%3a//www.peakdemocracy.com/portals/116/forum_home?admin_mode%3Doff#peak_democracy
4. Open the stream page: https://hypothes.is/stream
5. Search for `uri:http://www.a2gov.org/services/Pages/OpenCityHall.aspx?pd_url=http%3a//www.peakdemocracy.com/portals/116/forum_home?admin_mode%3Doff#peak_democracy`

Expected result: see the three annotations in the stream

Actual result: No result for the search.

tilgovi:
Confirm fix on stage.

tilgovi:
*fixed

judell:
Double-checked and confirmed.

PR #2184, opened 2015-04-20
Overhaul URI analysis

tilgovi:
Change the analyzers for URI fields so that they are consistently
indexing for three separate needs:

- Match anything queries
  This analysis will return matches to any piece of the URI by
  splitting it on various separators and their URL-encoded equivalents.
  When the user is typing free text without specifying a facet, this
  will match any part of a URI.

- Site / domain queries
  When the user wants to see all the annotations for a particular site,
  they will type only a domain. This analysis indexes only the domain
  part and strips a leading 'www'.

- Page queries
  This is the "standard" application query for getting annotations
  associated with a URI. It ignores the scheme, query string, fragment
  and strips any trailing slash. It is expected that this will return
  false positives for pages where the query string affects the content
  but that is probably better than the alternative, given the many
  non-content related uses of query strings (tracking variables, etc).

The new mappings are introduced under the "uri.parts" and "site" keys.

### DIFF
--- a/h/api/views.py
+++ b/h/api/views.py
@@ -278,7 +278,7 @@ def _add_any_field_params_into_query(search_params):
         'multi_match': {
             'query': any_terms,
             'type': 'cross_fields',
-            'fields': ['quote', 'tags', 'text', 'uri', 'user']
+            'fields': ['quote', 'tags', 'text', 'uri.parts', 'user']
         }
     }
 
--- a/h/models.py
+++ b/h/models.py
@@ -1,4 +1,5 @@
 # -*- coding: utf-8 -*-
+from __future__ import unicode_literals
 from annotator import annotation, document
 from pyramid.i18n import TranslationStringFactory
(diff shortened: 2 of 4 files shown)


tilgovi:
Needs tests

csillag:
@tilgovi wrote:

> Needs tests

Are the tests coming soon?
(The problems solved by this PR have been identified as a high-priority for the current interim sprint.)

tilgovi:
This is updated now with tests.

From my testing this fixes: #2095, #2141, #2180, #2181

dwhly:
Phenomenal.

judell:
Outstanding. I will try to exercise these things using the Selenium script I'm developing.

tilgovi:
I pulled out the "site" query so I can add it back in a new PR with an update to the client-side query parser and I'll add a requirement for myself that I include documentation on the available facets. So for now this doesn't change the public interfaces it should just make things work as they were likely expected to all along.

tilgovi:
I'm pulling the trigger and we're going to test this on stage. Stand by.

PR #2196, opened 2015-04-27
Keep JWZ and threading idTables in sync

### LINKED ISSUE #2182: Cannot press cancel on annotation replies in the stream
Create an annotation, go to the stream.  Click "reply".  Now click cancel.  Nothing happens.

Console message:

TypeError: Cannot read property 'children' of undefined
    at Threading.module.exports.Threading.annotationDeleted (app.min.js?36c6163a:3)
    at app.min.js?36c6163a:3
    at Scope.$emit (angular.min.js?5edec867:5)
    at revert (app.min.js?36c6163a:2)
    at angular.min.js?5edec867:4
    at callback (angular.min.js?5edec867:6)
    at Scope.$eval (angular.min.js?5edec867:5)
    at Scope.$apply (angular.min.js?5edec867:5)
    at HTMLButtonElement.<anonymous> (angular.min.js?5edec867:6)
    at HTMLButtonElement.jQuery.event.dispatch (jquery.min.js?74d54652:2)

gergely-ujvari:
When StreamController creates a new idTable, it also gives this new idTable to the threading service. 
Before this fix they were out of Sync.

Fix #2182

### DIFF
--- a/h/static/scripts/stream-controller.coffee
+++ b/h/static/scripts/stream-controller.coffee
@@ -17,7 +17,7 @@ module.exports = class StreamController
     $scope.search = {}
 
     # XXX: Reset the threading service
-    threading.createIdTable([])
+    threading.idTable = threading.createIdTable([])
     $scope.threadRoot = threading.root = mail.messageContainer()
 
     # Initialize the base filter
--- a/h/static/scripts/test/stream-controller-test.coffee
+++ b/h/static/scripts/test/stream-controller-test.coffee
@@ -63,7 +63,8 @@ describe 'StreamController', ->
     }
 
     fakeThreading = {
-      createIdTable: sandbox.spy()
+      createIdTable: sandbox.stub().returns('a')
+      idTable: null
     }
 
     $provide.value 'annotationMapper', fakeAnnotationMapper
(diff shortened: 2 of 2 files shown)


csillag:
LGTM.

csillag:
Travis failed with this message:

> Your test run exceeded 50 minutes.
> One possible solution is to split up your test run.

This might be caused by the temporary spike in load (see [here](http://www.traviscistatus.com/incidents/yk24wrkb4082)), so I don't think we need to worry about this, for now.

gergely-ujvari:
I've restarted the travis build.

csillag:
> I've restarted the travis build.

I think it will timeout again.

tilgovi:
I would like you to explain to me why this is correct.

Looking at the code for JWZ, `idTable` is a getter exposing a constructor-scoped variable that `createIdTable` assgins to (without the `var` keyword).

I don't understand how it could ever be out of sync, unless we assign to it elsewhere.

tilgovi:
In other words, if I'm understanding the code correctly, this PR is creating the same problem it's trying to solve.

tilgovi:
See here: https://github.com/hypothesis/h/pull/2199

gergely-ujvari:
@tilgovi: The problem lies around  [extending](https://github.com/hypothesis/h/blob/master/h/static/scripts/threading.coffee#L10) is not real inheritance. Checking #2199

gergely-ujvari:
closing in favor of #2199

Issue #1142, opened 2014-04-01
User does not get notified when saving an annotation fails on server-side


gergely-ujvari:
If any exception occur during the saving of an annotation, then in the client side:
- The error message not always pop-up (so the "Your annotation was not saved" is not always showed, and even when yes, only with a great delay)
- The javascript console has the 500 Internal error exception 
- The annotation card is not showed, but the annotation itself is counted in the tabs. This is inconsistent.

BigBlueHat:
Just hit this on http://quod.lib.umich.edu/j/jep/3336451.0017.201?view=text;rgn=main

As did our partner...

Please prioritize this one, k? :smiley:

dwhly:
When an annotation fails to save, can we notify [insert monitoring app]?

tilgovi:
@dwhly probably nothing to do there.

We already track all exceptions in Sentry on the server. If an annotation fails to save and it's not already logged there then the most likely cause is connectivity and the monitoring would likely fail, too, for the same reason.

dwhly:
Tx

tilgovi:
Closed by #2197

PR #2197, opened 2015-04-28
Add a client-side error when saving an annotation fails

seanh:
When creating a new annotation, on the front-end, if the server fails
when trying to save the annotation, then:

1. Show an error message to the user
2. Don't close the annotation editor

### DIFF
--- a/h/static/scripts/annotation-mapper.coffee
+++ b/h/static/scripts/annotation-mapper.coffee
@@ -25,5 +25,5 @@ module.exports = [
     deleteAnnotation: (annotation) ->
       annotation.$delete(id: annotation.id).then ->
         $rootScope.$emit('annotationDeleted', annotation)
-      annotation
+        annotation
 ]
--- a/h/static/scripts/directive/annotation.coffee
+++ b/h/static/scripts/directive/annotation.coffee
@@ -11,6 +11,14 @@ validate = (value) ->
   (value.target?.length and not worldReadable)
 
 
+# Return an error message based on a server response.
+errorMessage = (reason) ->
+  message = reason.status + " " + reason.statusText
+  if reason.data.reason
+    message = message + ": " + reason.data.reason
+  message
+
+
 ###*
 # @ngdoc type
 # @name annotation.AnnotationController
@@ -105,8 +113,10 @@ AnnotationController = [
     this.delete = ->
       $timeout ->  # Don't use confirm inside the digest cycle
         if confirm "Are you sure you want to delete this annotation?"
+          onRejected = (reason) =>
+            flash.error(errorMessage(reason), "Deleting annotation failed")
           $scope.$apply ->
-            annotationMapper.deleteAnnotation model
+            annotationMapper.deleteAnnotation(model).then(null, onRejected)
       , true
 
     ###*
@@ -123,6 +133,16 @@ AnnotationController = [
     ###*
     # @ngdoc method
     # @name annotation.AnnotationController#view
+    # @description Switches the view to a viewer, closing the editor controls
+    #              if they are open.
+    ###
+    this.view = ->
+      @editing = false
+      @action = 'view'
+
+    ###*
+    # @ngdoc method
+    # @name annotation.AnnotationController#revert
     # @description Reverts an edit in progress and returns to the viewer.
     ###
     this.revert = ->
@@ -131,8 +151,7 @@ AnnotationController = [
         $rootScope.$emit('annotationDeleted', model)
       else
         this.render()
-        @action = 'view'
-        @editing = false
+        this.view()
 
     # Calculates the visual diff flags from the targets
     #
@@ -146,6 +165,13 @@ AnnotationController = [
         t.diffHTML? and not t.diffCaseOnly
 
       {hasDiff, shouldShowDiff}
+
+    # Update the given annotation domain model object with the data from the
+    # given annotation view model object.
+    updateDomainModel = (domainModel, viewModel) ->
+        angular.extend(
+          domainModel, viewModel,
+          {tags: (tag.text for tag in viewModel.tags)})
 
     ###*
     # @ngdoc method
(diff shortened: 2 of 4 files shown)


seanh:
Needs:

- [ ] Show the actual error from the server (when it sends a json error)
- [ ] Also handle errors when editing and deleting annotations
- [ ] Tests
- [ ] Refactor to add a `close()` method, remove duplication

tilgovi:
This is looking great!

judell:
https://github.com/hypothesis/h/issues/1142

seanh:
I wanted to add a test for when you edit an annotation, change its body text, hit save, get an error back from the server, then hit cancel. This discards your changes and returns the annotation body text (and tags) to its previous value. I can't figure out how to test this though

seanh:
I think I'm happy with this now, other than the test that I mentioned above that I'd like to add.

seanh:
I don't understand this. When testing manually I can do this: Edit an annotation, change its body text, Save, get an error back from the server which I've hacked to fail on saving annotations, the annotation editor form stays open, click Cancel - annotation body text is restored to original value.

The Edit button is this.edit() in annotation.coffee, the save button is this.save() and the cancel button is this.restore(). The text in the textarea is this.annotation.text. If I put breakpoints at the start and end of the revert() method I can see that it's called when I click Cancel, this.annotation.text is the edited text at the start of the method and is set back to the original text by the end of the revert() method.

Yet if I try to simulate this in an automated test, this.annotation.text doesn't get restored. This test in annotation-test.coffee fails:

```
    it "restores the original text when editing is cancelled", ->
      annotation.$update.returns(Promise.reject({
          status: 500,
          statusText: "Server Error",
          data: {}
        })
      )

      original_text = controller.annotation.text
      controller.edit()
      controller.annotation.text = "changed by test code"
      controller.save().then(->
        assert controller.annotation.text == "changed by test code"
        controller.revert()
        assert controller.annotation.text == original_text
      )
```

The onRejected() function in this.save() in annotation.coffee _does_ get called when this test runs, a flash error message with the error from the server is displayed. revert() just doesn't seem to revert the text.

tilgovi:
I can't tell what `controller.annotation.text` starts out to be. I only see you manipulating the view model, so I don't know what `revert()` would `render()`. Which assertion fails and what's the failure?

seanh:
It's the final `assert` that fails. In the test, after clicking _Cancel_ / calling `revert()` `this.annotation.text` is not restored to its original value, it remains `"changed by test code"`. But outside of the test it does get restored.

I've made some progress. It centers around this bit in the test code: <https://github.com/hypothesis/h/blob/eaca2667a816561061634ca932eaacd6d4f8bff8/h/static/scripts/directive/test/annotation-test.coffee#L102>. For some reason that creates an `$scope.annotation` with no `.text` attribute. Maybe that makes sense, for the existing tests in that file? But for my test it breaks the test, and is different from what happens when the app is actually running.

If I add a `text: ` with an initial value to that object then my test passes, but weirdly two other tests then start to fail:

```
PhantomJS 1.9.8 (Linux) annotation when the annotation is a highlight persists upon login FAILED
        AssertError: expected stub to be called once but was called 0 times
            at fail (/home/seanh/Projects/hypothesis/h/h/static/scripts/vendor/sinon.js:5409)
            at failAssertion (/home/seanh/Projects/hypothesis/h/h/static/scripts/vendor/sinon.js:5367)
            at /home/seanh/Projects/hypothesis/h/h/static/scripts/vendor/sinon.js:5390
            at /tmp/ea8bd23305fb12389202500135fcd892.browserify:6587:0 <- null:0:0
PhantomJS 1.9.8 (Linux) annotation when the annotation is a highlight is private FAILED
        AssertionError: expected undefined to deeply equal { read: [ 'justme' ] }
            at /home/seanh/Projects/hypothesis/h/h/static/scripts/vendor/chai.js:925
            at assertEql (/home/seanh/Projects/hypothesis/h/h/static/scripts/vendor/chai.js:1434)
            at /home/seanh/Projects/hypothesis/h/h/static/scripts/vendor/chai.js:3627
            at /home/seanh/Projects/hypothesis/h/h/static/scripts/vendor/chai.js:2523
            at /tmp/ea8bd23305fb12389202500135fcd892.browserify:6594:0 <- /home/seanh/Projects/hypothesis/h/h/static/scripts/directive/test/annotation-test.coffee:136:47
```

At an initial glance I don't understand why they're failing:

```
    it 'persists upon login', ->
      delete annotation.id
      delete annotation.user
      createDirective()
      $scope.$digest()
      assert.notCalled annotation.$create
      annotation.user = 'acct:ted@wyldstallyns.com'
      $scope.$digest()
      assert.calledOnce annotation.$create
```

Apparently the presence of a `.text` attribute on `$scope.annotation` causes `annotation.$create()` to not get called?

And:

```
    it 'is private', ->
      delete annotation.id
      createDirective()
      $scope.$digest()
      assert.deepEqual annotation.permissions, {read: ['justme']}
```

Adding a `.text` attribute on `$scope.annotation` causes `annotation.permissions` to be `undefined`.

I'll keep looking, see if I can either fix these two tests or change my test to not effect them.

seanh:
@tilgovi Ok I finally got this to pass by moving the test into its own top-level `describe()` where it doesn't clash with other tests.

This also gave me a chance to figure out how to apply my Angular unit testing method to directives, which is a bit trickier than controllers, and I think you have to use `$provide`, have a look at the commit and tell me what you think: https://github.com/hypothesis/h/commit/85a7cd2363a3b7bd7c68bfdcc01b01320b448abd

I made `createAnnotationDirective()` do only the minimal stubbing necessary to create the directive without crashing, then any more stubbing needed for the actual test to run is done in the test itself. I wanted to reduce the amount of stub behavior inherited by all tests, since inheriting an unwanted bit of stub behavior is exactly what caused me to be stuck on getting this test to pass for a week. But this does make for quite a lot of boilerplate stubbing in the test method, so maybe some of it should move into `createAnnotationDirective()`.

seanh:
This pr can be reviewed now (and if people are happy with it I'll rebase)

tilgovi:
That looks great. Let's rebase and merge it.

Once we solidify some of these patterns you've developed, we might see a chance to pull them out into a helper module that all the tests can access, just as angular-mocks provides the global `module`, `inject`, etc, we could have some additional helpers around that if we need.

But it's looking great as is. My brain recoils a little bit at the volume of comment in the test code, but maybe that's good.

seanh:
Ready!

tilgovi:
@seanh I added a generic error message for service unavailable when there is no status line, such as happens when there's no connection.

LGTM. Feel free to merge if you are happy with that addition.

seanh:
Looks good to me, waiting to see if Travis passes

tilgovi:
Green. :)

tilgovi:
I'll do it.

PR #2198, opened 2015-04-28
[chrome] run embed code in page context

tilgovi:
Rather than running the embed code in the content script context and
duplicating the double embed prevention in the extension, run the
embed code in the page context since it now handles the double embed
prevention itself.

This is also simpler to reason about because the embed code runs in
a similar way for all injection methods (boomarklet, script embed, or
extension).


tilgovi:
@hypothesis/maintainers-h this one is blocking upgrading the extension right now and I'd like to be able to roll out a new version.

tilgovi:
Nick checked this over in person with me on Monday. Since then it's only had comments and tests added. I'm going to merge this one myself so it unblocks release.

PR #2208, opened 2015-05-03
Convert some forms.scss to ems

tilgovi:
There should be only negligible change to any dimensions, but this way our forms are resizable using our base font size variables on the body.


tilgovi:
Don't merge this because it actually will only make sense within the context of a different PR I'll submit soon.

PR #2209, opened 2015-05-03
Distinguish brand color from error color

tilgovi:
We don't want our brand to be associated with errors and warnings
by virtue of using the same colors for both.

This clarifies things a bit by choosing a suitable orange red for
the error base.

As an additional touch-up for the colors this adds a function to
do return color "weights" giving color values similar to those in
Google's Material Design color palettes.

### DIFF
--- a/h/static/styles/common.scss
+++ b/h/static/styles/common.scss
@@ -89,7 +89,7 @@ html {
 }
 
 .red {
-  color: $hypothered;
+  color: $brand-color;
 }
 
 .pull-left {
--- a/h/static/styles/forms.scss
+++ b/h/static/styles/forms.scss
@@ -102,7 +102,7 @@
 .form-error-list {
   position: relative;
   display: none;
-  background: $hypothered;
+  background: $error-color;
   margin-top: (10 / 13 * 1em);
   padding: (4 / 13 * 1em) (11 / 13 * 1em);
   float: left;
@@ -122,8 +122,8 @@
     width: 0;
     position: absolute;
     pointer-events: none;
-    border-color: rgba($hypothered, 0);
-    border-bottom-color: $hypothered;
+    border-color: rgba($error-color, 0);
+    border-bottom-color: $error-color;
     border-width: 4px;
     margin-left: -4px;
   }
@@ -132,7 +132,7 @@
 .form-error {
   font-size: 13px;
   line-height: 1.5;
-  color: $error-red;
+  color: $error-color;
 }
 
 .form-checkbox-item {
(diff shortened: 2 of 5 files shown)


tilgovi:
brand-error-color-distinction

dwhly:
If we're worried that our H. red dot is somehow reminiscent of errors and warnings then it's probably the red dot we should change and not the color of our errors. 

Lots of logos use a bit of red, not sure this is an issue.

tilgovi:
If it were only the red dot I wouldn't worry much.

But I want to use this color elsewhere. It's currently plastered across the top of our home page.

If you seriously want to change our logo, that's another discussion. I'm proposing a simple, practical and considered change to some colors and would appreciate feedback that actually addresses that proposal.

tilgovi:
It's more than that even. Our whole website is themed around this red. The buttons, the link colors, the dot.

tilgovi:
And to be clear and address what you did say, I am _not_ worried that our brand color is reminiscent of errors, generally, but it is a problem to me if _our_ errors are the same color.

Within our design we can make colors mean what we like, to some extent, but using the same color for our logo and our error messages is busted to me.

tilgovi:
Screenshot shows new, proposed error colors and (in the top right) the brand color on the "Show Code" tab.

csillag:
Makes sense to me.

judell:
Thanks for this. Let's separate 2 issues:

1 Ability to distinguish brand color from error color

2 Deciding what those colors should be

This PR achieves 1, and makes a choice about the error color that we can have our future designer revisit.

tilgovi:
In lieu of a future designer I acted in this capacity and chose a color based on:

- Using complementary color schemes to pick a compatible orange.
- Using a monochromatic color scheme to find an appropriate base orange from that.

Unless there are super strong objections to the choice, I'd like to merge it and let a future designer replace it if they wish to.

tilgovi:
/me is not understanding why there is any hesitation here.

While technically, from a _code only_ perspective, it's not ridiculous to say this can possibly achieve (1) without (2), just by renaming some variables, that's really not the point of this.

csillag:
> /me is not understanding why there is any hesitation here.

I don't think @judell has expressed any hesitation.

tilgovi:
More to the point, this is already done since I have chosen a color, not arbitrarily. Unless someone else wants to put in that work now, I think it's pretty silly not to move forward with this. If it were a choice between doing this work and leaving it for a hypothetical future person, there's an argument that time could be better spent elsewhere, but it's too late for that since I've already done it.

judell:
I don't feel at all ambivalent. Yes let's do this now. I'm fine with the color choice. If a future designer wants to change it, he or she will now have that option. But absolutely, do this now and move on.

PR #2210, opened 2015-05-03
Break out client configuration into a Bundle

tilgovi:
In this manner, the client configuration is made more easily
replaceable in downstream deployments through the asset configuration.

Additionally, code shared between the various application pages that
configures the identity and session providers for our account system
lives in one place holding those cross-cutting concerns.


gergely-ujvari:
LGTM

PR #2012, opened 2015-03-09
Add ability to generate via links from the sidebar.

RawKStar77:
This PR adds the ability to generate via links from the sidebar. It is available when clicking "Share this page" inside the username menu:
selection_017

It then shows a dialogue with the link selected and a short set of instructions:
selection_016

Issue #2193, opened 2015-04-26
[hack day] Search sharing (with VIA)


csillag:
It would be nice to be able to share a page (via VIA) with a specific set of annotations selected.

Specification is [here](https://docs.google.com/a/hypothes.is/document/d/1RhiT-EgNK_pINALgD0Lv9rpWTSplz-oz_O-Almg2nHQ/edit?usp=sharing).

Being implemented [here](https://github.com/hypothesis/h/tree/search-sharing) by @csillag, @ikreymer and @gergely-ujvari, as per the feature request by @emvincent, with input from @jeremydean and @tilgovi.

Parts needed: 
 - [ ]  Server-side API to save and retrieve search queries - @gergely-ujvari is working on this
 - [ ] VIA: ability to take a parameter from the URL, and pass it to embed.js - @ikreymer can tackle this
 - [ ] H embed code: ability to take a parameter from the URL
 - H client side:
    - [x] *Share this search* feature - @csillag
    - [x] *Share this selection* feature - @csillag
    - [ ] Ability to load and execute a selected search or selection - @csillag
    - [ ] Extra emphasis for selected annotations (See [PR](https://github.com/hypothesis/h/pull/2036)) - @csillag

judell:
A possible simplification: No server-side API, just pass annotation ids (trivially, a single one) through via.

![Wordnik example](http://jonudell.net/images/wordnik-balloon.html)

#2215 
#2012

tilgovi:
I think we quickly run into a situation where we have to pick a deliberately strange and obfuscated thing to tack on the end of the URL hoping that it won't be ambiguous with respect to an actual query parameter for the original page.

And then we have to ask ourselves why we shouldn't just use local storage or something.

ikreymer:
I think there are two options as far as via. I think thus far we've been leaning towards 1) (I haven't had a chance to add it yet).

1) `https://via.hypothes.is/<ID>/http://example.com/`
Via parses <ID> server side and passes in to javascript as part of the inject script.
This requires some server-side changes but not much.

2) `https://via.hypothes.is/http://example.com/#<ID>
This requires essentially no server side changes, but is a bit more tricky, as the embed script should essentially strip of the `h=<ID>` first thing, in case there is an existing anchor and the site will use the anchor in some way. Care must be taken to ensure `<ID>` is actually is part of hypothesis and not the site's own anchor. But its doable, since the inject script runs first thing (I think). I think, however, option 1) has been preferred thus far.

[...]

Issue #1562, opened 2014-10-09
Implement the new filter bar design


aron:
Code is partially implemented here: https://github.com/hypothesis/h/tree/filter-bar-styles

This moves the viewer sorting out of the stream and into the filter bar in preparation for the pending viewer work. The intention is for this menu to grow in future with more filtering options such as facets etc. But the first step is just sorting.

With regard to the above branch, I’ve put the example markup in the app.pt for the moment. The idea is that the bar is hidden by default. When you click the icon the filter bar appears, clicking the icon again hides it. When a filter is active the icon is red. For the moment we only have sort filters, so those can stay the same.

![filter-bar](https://cloud.githubusercontent.com/assets/47144/4580257/0135bf96-4fd4-11e4-9a23-08007c83e704.png)

robertknight:
A new design for the filter bar has been implemented in master. The dropdown is not the same as the design above, although the filter indicator in the toolbar is quite similar. I see there were a bunch of other issues being discussed here, they can be handled separately.

PR #2215, opened 2015-05-06
Add ability to share via links from the sidebar.

JakeHartnell:
Did a little bit more design work and thinking around sharing today, I'm submitting this PR as an alternative to #2012.

Here's what the new share dialog looks like:
Share dialog

The changes include:
- Use the canonical url for generating via links. 
- Generate via uri in the share-dialog directive. (Thanks @csillag and @gergely-ujvari!)
- Introduce toolbar share icon, the tooltip says "Share this page"
- Update help/welcome page to cover sharing
- Add simple sharing via Facebook, Twitter, Google plus, and email
- Add social media icons to icomoon

**Edit:** We used have a share button on each annotation that links to the standalone annotation page. In the name of simplifying, I had removed it for the time being (thinking we should add it back when we can directly link to an annotation *in context*), but put it back after some outcry. Changes were made in a separate PR: https://github.com/hypothesis/h/pull/2231


dwhly:
While I'm waiting for the dokku to build:

> Does anyone violently disagree with that?

Yes, I do.  I agree that direct linking is superior, and that the standalone page is generally what almost no one wants when wanting to share a link to the annotation, however I'm quite passionately opposed to removing the share link to individual annotations/replies completely in the interim.

1- With this change there's zero way to link to specific annotations / replies, etc. at all.  At least w/ the standalone page, one can point to a specific part of a conversation.
2- This represents the link to the web resource that is the canonical identifier for the annotation.  In a sense, our whole narrative around how annotations are web resources is diminished by this change.
3- We probably always want *some* way to link to this, even when a direct linking approach is offered.
4- I talk to this share link all the time in demos, it's an important part of what we're trying to do.

Look forward to the rest of the PR, looks like some interesting thinking.

h/static/scripts/app-controller.coffee
```diff
@@ -46,6 +46,13 @@ module.exports = class AppController
     oncancel = ->
       $scope.dialog.visible = false
 
+    $scope.dialog = visible: false
+    $scope.viaLinkVisible = false
+
+    # Check to see if we are on the stream page.
```
tilgovi: What's this change about?

h/static/scripts/config/via.coffee
```diff
@@ -0,0 +1,5 @@
+angular = require('angular')
+
+angular.module('h')
+.value('via', url: 'https://via.hypothes.is/')
+
```
tilgovi: whitespace

RawKStar77:
Thanks for the nice response @dwhly, always good to get other perspectives. I think you've convinced me at least that it doesn't belong in this PR . Commit https://github.com/hypothesis/h/commit/72f83f53b5060b8d605776c8ee6afec7258aaee1 re-adds the share button on annotations as things were before.

h/static/styles/simple-search.scss
```diff
@@ -3,6 +3,9 @@
 
 .simple-search {
   overflow: hidden;
+  position: absolute;
```
tilgovi: Please try to find a way to do this layout without this. If it's possible to let the layout flow that's always better than absolute positioning in my mind.

h/templates/app.html
```diff
@@ -16,6 +16,7 @@
           --><span class="provider" ng-show="auth.user">/{% raw %}{{ auth.user|persona:'provider' }}{% endraw %}</span><!--
           --><i class="h-icon-arrow-drop-down"></i></span>
         <ul class="dropdown-menu pull-right" role="menu">
+          <li ng-hide="isStream"><a href="" ng-click="viaLinkVisible = true">Share this page</a></li>
```
tilgovi: If we have a share button in the search bar, let's maybe not have it in the dropdown.

h/templates/app.html
```diff
@@ -28,6 +29,15 @@
          ng-click="login()"
          ng-switch-when="null">Sign in</a>
 
+      <!-- Share this view -->
+      <span class="share-via-link">
+        <i class="h-icon-share"
+           title="Share this view"
+           ng-hide="isStream"
```
tilgovi: Oh, I see. Fine for the moment.

tilgovi:
Generally, yes, I love the idea to share to these services directly from here.

I would reconsider copying my bad patterns, though ;).

I would ake the share button itself, the one in the top bar, part of the via link dialog directive and make that directive have a template that pops open and out of the top bar when it's open.

That saves a lot of nastiness:

- The dialog is even closer to the button you click to open it
- The markup can live in one place, rather than split over two (not counting the drop-down)
- The app controller doesn't have to know about the state of the dialog

In short, the dialog can be self contained.

Try to get that nailed and I'll help you remove then need for isStream.

jeremydean:
is it relevant for me to ask here: why reddit and google plus? 

does any really use the latter? and in terms of the former, does reddit really call for the kind of elegant share card i assume will populate on twitter? doesn't a naked link suffice there? note i'm not a reddit user. 

also why not facebook?

tilgovi:
:+1: all of @jeremydean's questions.

judell:
Late to this party, sorry. https://sharevia.dokku.hypothes.is/ down now?

tilgovi:
Probably. That's my fault. I ran some upgrades on dokku and ended up wiping
the docker partition when I ran into mega problems.

RawKStar77:
> Probably. That's my fault. I ran some upgrades on dokku and ended up wiping the docker partition when I ran into mega problems.

That's why I've had trouble pushing out a dokku for this... I keep getting:
```
To dokku@dokku.hypothes.is:sharevia
 ! [remote rejected] sharing-via-link-from-extension -> master (pre-receive hook declined)
error: failed to push some refs to 'dokku@dokku.hypothes.is:sharevia'
```

tilgovi:
Well if that's still the case after you recreate the app let me know.

The upgrades were partly because I was experiencing similar.

judell:
This got me pondering direct linking again. #2193 envisions a server-side API that persists annotations sets. Would the simplification proposed there -- just pass an annotation id straight through via into the app -- get us to MVP on this?

There's a world of difference between this:

http://jonudell.net/images/wordnik-balloon.jpg

and this:

https://hypothes.is/a/JdjfuZBCTJWXk5NYnhTJBg

Jake's instinct is correct, the latter is lame. Dan's right too, we currently rely on it.

There's already some momentum on #2193, it isn't in the current sprint whereas #2012 (and maybe #2215 now too), can/should we combine efforts?

judell:
I will answer my own question: Not now. One step at a time.

RawKStar77:
@jeremydean, I didn't include Facebook (*at first* -- it is now included) because they only describe adding a share button to the page with JavaScript. I didn't want facebook Javascript to load on all the pages the extension does, but I eventually found what I needed: a link we can pass parameters into (*like URL*).

csillag:
> There's already some momentum on #2193, it isn't in the current sprint whereas #2012 (and maybe #2215 now too), can/should we combine efforts?

Yes, those should definitely be combined.
But let's finish this first, and the build #2215 on top of this.

csillag:
What I wanted to say:

> let's finish this first, and the**n** build #2215 on top of this.

csillag:
> -Introduce toolbar share icon as well as a username menu item (both are hidden on the steam)
> toolbar icon

I think having the share icon next to the search bar is slightly confusing; some people might get the mistaken idea that the thing being shared has something to do with the search query or search results ... which *will* be supported with #2193 (hopefully), but not right now.

judell:
Note for future enhancement: A user setting to choose which share buttons to display by default. LinkedIn will matter in some circles, Reddit in others, etc. No need to be constantly facing a button you'll never use.

tilgovi:
Still wish this had worked out: https://github.com/xauth/xauth

RawKStar77:
> some people might get the mistaken idea that the thing being shared has something to do with the search query or search results ... which will be supported with #2193 (hopefully), but not right now.

That is exactly what I was going for. : ) When #2193 is done, we should change the text of the tool tip and dialog to "share this view" or something similar. I first had this idea as I was working on groups and also needed a quick to get the link to the group again, so putting the share button in the toolbar seemed like a natural way to grab a link to what every you are looking at: *a page*, *a search query*, or soon even *a group*. Still need to do user testing naturally.

RawKStar77:
Note: the screen shot is out of date, the tooltip when hovering over the share icon says: "Share this page." When #2193 is done we can change it.

judell:
What's up with https://sharevia.dokku.hypothes.is/?

judell:
@tilgovi https://github.com/xauth/xauth Yeah, the field is littered with corpses. The IIW (Internet Identity Workshop) folks refer to what xAuth wanted to solve as the [NASCAR problem](http://ecx.images-amazon.com/images/I/913P5ngGVOL._SL1500_.jpg)

tilgovi:
If this is nearing ready it needs a rebase.

RawKStar77:
Rebased. This just needs review, in which a dokku would be very helpful! (Still having problems there.)

I've made some last minute improvements after doing a bit of usability testing, mostly adding `readonly` to the input with the via link so that people don't accidentally change or delete the link they are trying to copy. : ) 

Cleaned up the CSS a bit too.

Lastly, I played around with the copy and order of the dialog, it now looks/reads like this:
Share dialog

tilgovi:
Looking good. Copy could still be simpler, to me.

"Share the link below to show anyone these annotations and invite them to
contribute their own."

I don't think you need more than that. The bit about loading Hypothesis is,
IMO, too much technical detail. Words like "loaded" totally unnecessary.
Words like 'click' are mouse-centric.

Lastly, a question about the links. Is "post" preferred to "share" for
Facebook but not G+ for some reason?

Maybe a verb on the email (I know, email can be a verb) but maybe "Send via
email" to balance out the line lengths a bit.

Also, it may still be the case that more people share articles by email
than any other way. Do a quick search, this is a not infrequently reported
on thing. The answer is out there. And my bad if you already did this. But
if true, maybe email should be at the top.

tilgovi:
I'm skeptical of the style changes but I can check it out if you like. What's the max-width for? Does the link need to float left? Why?

Should we really be changing all the dialog stuff like this, or should that be better encapsulated in the directive.

I'd like to get the dialog template not to have the top level wrapper that it has. That makes it too aware of its placement and as such less relocatable as a component. It also introduces noise into the diff, including the changes to distinguish it from the other dialog that should also not be bleeding all over the AppController.

I know a lot of this is that I did things poorly with the account controller and now you're copying me, so no worries, but let's always be improving.

If you need help making these changes, just ask, or pair with someone on it.

judell:
A quick survey: 

image

At this point the icons are, well, iconic. Text is minimized to the point that some of these sites don't even bother with tooltips.

Available choices, and the ordering of them, varies according to audience. Since we appeal to no specific audience I don't think there's a right answer, and would eventually want our users to be able to configure both choices and order.

tilgovi:
Ah, this is what I was looking for: http://www.sharethis.com/blog/2015/01/21/q4-2014-consumer-sharing-trends-report/#sthash.P2hBdzEp.dpbs

It seems that things have changed in the last few years and email has lost ground. No surprises.

Anyway, I'm also with @judell about possibly not needing any text for these.

Let's fiddle with the design and see how clean this can be.

RawKStar77:
Ok, dokku is finally up here: https://sharevia.dokku.hypothes.is

Thanks @tilgovi and @judell for the many suggestions, sorry I've been a bit slow to respond. I've made another iteration on the design based on your wonderful thoughts, which now looks like this:
share via link

judell:
+1.

tilgovi:
So nice.

RawKStar77:
Rebased.

RawKStar77:
> I'd like to get the dialog template not to have the top level wrapper that it has. That makes it too aware of its placement and as such less relocatable as a component. It also introduces noise into the diff, including the changes to distinguish it from the other dialog that should also not be bleeding all over the AppController.

Feels like a different issue I think (since it touches on both the account, sharing, and future dialogs), but one we should certainly work on. I do agree that we need a better way to do dialogs, and I'm happy to work on that later.

Anything else holding things up?

tilgovi:
Let me take a quick look at that later today.

BigBlueHat:
@RawKStar77 couple things. I agree with @csillag's point about the odd placement of the share button. There's nothing else "click-able" in that region except the close sidebar button--which is a very different intent.

Given your thoughts on "eventually this will share the current view" what do you think about moving the share icon down into the gray space floated right of the "Sorted on location" drop down.

Like this (or similar):
2015-05-23 11_41_31-hypothesis
(obviously with some tweaking :wink:)

That would move it in the same conceptual space as the filter, and the same click-focused region as the user drop down list and other "action" click-ables.

Also, it'd be best (for similar reasons) to right align the share buttons (so they match all the other buttons) and also make them match the same color/hover experience as the other buttons.

Let me know if you want to chat through these or anything. Cheers!

dwhly:
The "social" view that will be the primary context for sharing (when groups is implemented) is up in the toolbar above tho.

dwhly:
I suppose it could move down as well, but they should probably stay close together.

BigBlueHat:
That one makes sense where it is:
2015-05-23 12_03_00-hypothesis

Putting the group name (and I assume eventually a drop-down to change it?) has similar meaning to the username: "I am taking action as / for this Entity."

The share button (even in the groups demo), still feels off all by itself in an area one isn't looking to click--except to type in a search.

dwhly:
> The share button (even in the groups demo), still feels off all by itself in an area one isn't looking to click--except to type in a search.

Yeah, I think the reasoning is that the share button should be next to filters from which it will inherit context. Search being one of those. How best to do that is the question.

dwhly:
Future thought: it might be nice if the Share dialog could explicitly list which filter values it was capturing (at the point that these are implemented).

tilgovi:
So, what's actually being shared? Here, it's the page.

Have we considered share as a toolbar icon, of the variety that floats inside the page even when the sidebar is closed?

RawKStar77:
> Have we considered share as a toolbar icon, of the variety that floats inside the page even when the sidebar is closed?

Yes, that was the first thing I considered, and works well for sharing just a page, but I chose the current design over that option because I had a thought towards the future when we could share not only just a page, but the current view: filters, search queries, groups, etc. I also think by having the share icon in the sidebar makes it ultra clear that we are sharing *this layer* over the page and *not* the page itself.

tilgovi:
I wonder if that actually just muddles things. Would we necessarily not want two entry points to sharing, in different places and thus implying different context, if we introduce other features? For the purposes of this one, though, does the toolbar really not make sense?

If it isn't sharing the page with annotations, it wouldn't be there at all. Users can copy/paste the location bar or use whatever in-browser sharing feature they usually would.

Even if it were to surprise them that it shares the page with annotations, aren't we aiming to be so unobtrusive as to have that not be an unpleasant surprise?

Is there a scenario where having it in the toolbar would rightly anger a user?

tilgovi:
In any case, if we think there's potential confusion should we consider instead to make the whole thing more explicit and have language for it somewhere?

I'm not trying to hard sell any one design, but I'm asking questions. I don't see enough of the thinking that has happened yet to see anything that really justifies the search bar as the best place for this to me.

judell:
"Is there a scenario where having it in the toolbar would rightly anger a user?"

Now that it's possible to leave the (test) extension active I have been doing so. As a result, I notice when it obscures things and am annoyed by that. Growing the toolbar will obscure more things and annoy me more.

I am concerned about clutter, and while https://github.com/hypothesis/h/issues/2252 puts even more pressure on scarce space, I think it's necessary and appropriate because it's really about notification, not just RSS.

Here's a legend to help think about these things.

[<] = share control (our share icon)

[R] = notify control (R denotes RSS icon)

[O] = view control (O denotes the globe icon for All or Public)

[A] = app control (the app menu)

Each is clickable, each expands to offer affordances:

[<] = open a panel, offer a page-level share link (possibly scoped by [O]

[O] = open a dropdown list, offer to select a page-level scope

[R] = open a panel, offer page-level notification (by RSS/Atom, email, possibly other)

[A] = open a menu, offer app-level actions

The odd man out here is actually search. On a call just now we talked about suppressing the search box until the search icon is clicked. Straw man proposal: do that but then open a panel for search, within which we can also offer the currently-undocumented search affordances (user and tag facets).

Let's call that one [S]. So now we have: 

[S] = search control

[R] = notify control (R denotes RSS icon)

[<] = share control (our share icon)

[O] = view control (O denotes the globe icon for All or Public)

[A] = app control (the app menu)

We can debate the order of these things but what we have here is a top-level app menu where everything item is a subsidiary interaction.

RawKStar77:
Sharing is important, it deserves to have visual prominence in our app, and I think the best place for that is in the top bar. Jon hit on a key point, which is that we need to be forward thinking about what other UI elements we are going to be adding to the app soon (such as notifications and needing to grab a link with filters, queries, or groups applied).

I do not think we should put the share dialog toggle in the toolbar as @tilgovi has suggested, even though I confess this was also one of my first thoughts. However, I eventually decided against the toolbar as I don't  want to be continually adding things to it. I agree with @judell:

> As a result, I notice when it obscures things and am annoyed by that. Growing the toolbar will obscure more things and annoy me more.

Are we going to add notifications to the tool bar? What else? I'm not a big fan of adding things to the toolbar unless we absolutely have to.

I think putting things in the topbar is a better alternative, as there is more space, and again with an eye towards the future where we will be able to share search queries, groups, or filters. Here's an example from the groups prototype:
Groups prototype sharing

I can imagine something similar with search parameters, and I think having the share dialog toogle in the topbar (as proposed by this PR) will make it clearer what is happening when we have this functionality... but time and users will give us much more information.

Additionally, IMO having the share icon in the sidebar makes it ultra clear that we are sharing *this layer over the page* and not *the page itself*. For what it's worth, in informal user tests with a few of my friends no one has failed to share the page yet.

@BigBlueHat's proposal was quite tempting, but I think the top bar is a better place for the ultra important activities of sharing and notifications. It's important that it's fixed too, since it is still selectable when scrolling through annotations:
Example

As far as space in the top bar goes, I'm not to worried. I think eventually we should pick up on @aron's idea to implement a new filter bar design: https://github.com/hypothesis/h/issues/1562 as it will address the problem of not having filters accessible when scrolling through annotations an provide us with a bit more space.

As for @tilgovi's point of making the whole thing more explicit and having language for it 'somewhere.' I've added a new section to the help page. The icon is also fairly recognizable, and it has a tooltip... no different than it would be if it were in the toolbar. Originally, I also had a separate menu item in the dropdown user menu that read "Share this page," though I removed it after [someone questioned the need for the repetition](https://github.com/hypothesis/h/pull/2215#discussion_r29731887). I admit I was on the fence about this as well.

I shall also make the point that this is by no means the end-all 'final' design of how sharing (or even the topbar) will look in our application. We are going to add sharing back on the cards when we have direct linking, and we should be exploring other options too (such as what sharing might look like on the stream?). 

I feel like we can talk about the placement of this icon to death, when in reality is very easy to move and rather than wasting too much moving it to various places I would prefer continuing with the current design, get it in front of users, and iterate. Does anyone strongly object?

RawKStar77:
Dokku here: https://sharevia.dokku.hypothes.is/

judell:
"get it in front of users, and iterate"

+1

nickstenning:
This has been open forever, and it's now looking pretty good. I'd love to merge it, but it doesn't seem to have any tests. Could you add some tests for the share-dialog directive?

RawKStar77:
@nickstenning, I added a simple test to App-Controller-Test.coffee. There's probably more that can be added with regard to to this part of the share-dialog directive:
```
        scope.$watchCollection (-> crossframe.providers), ->
            if crossframe.providers?.length
                # XXX: Consider multiple providers in the future
                p = crossframe.providers[0]
                if p.entities?.length
                    e = p.entities[0]
                    scope.viaPageLink = 'https://via.hypothes.is/' + e
```
But I'm not quite sure how to test that. Any suggestions? @seanh, I know you've been working a lot with tests of late.

Should I make a fake crossframe.providers collection and then add something to it and see if viaPageLink is set?

nickstenning:
Ok @RawKStar77, yes, you need to add some tests for the directive. So create a `h/static/scripts/directive/test/share-dialog-test.coffee` file -- `status-button-test.coffee` looks like a good example to copy.

You can fake out the `crossframe` service. There's an example of this in `h/static/scripts/test/widget-controller-test.coffee` (search for `fakeCrossFrame` in that file) and copy the pattern in your tests.

Then you want some tests which set `crossframe.providers` to some values you might expect, and ensure that `$scope.viaPageLink` is updated appropriately. You'll probably have to manually call `$scope.digest()` each time you change `crossframe.providers`.

JakeHartnell:
Ok, after some help from @nickstenning in the test writing department, this is ready. Someone want to push the button?

h/static/scripts/directive/share-dialog.coffee
```diff
@@ -0,0 +1,28 @@
+###*
+# @ngdoc directive
+# @name Share Dialog
+# @restrict A
```
nickstenning: This line is out of date.

JakeHartnell: Good catch, fixed.

JakeHartnell:
Bump. @nickstenning, @tilgovi, or @seanh... is there anything else I need to do or can we hit the button?

h/static/scripts/directive/share-dialog.coffee
```diff
@@ -0,0 +1,29 @@
+###*
```
tilgovi: This file is inconsistent with the rest of our CoffeeScript. It has four space indents instead of two.

h/static/scripts/test/app-controller-test.coffee
```diff
@@ -117,7 +117,11 @@ describe 'AppController', ->
 
   it 'does not show login form for logged in users', ->
     createController()
-    assert.isFalse($scope.dialog.visible)
+    assert.isFalse($scope.accountDialog.visible)
+
+  it 'does not show shareDialog by default', ->
```
tilgovi: Try to write these in a natural language descriptive style, rather than referencing the implementation. 'does not show the share dialog at start' or something. I'm totally nitpicking, but `shareDialog` is an object, it can't be shown. The "share dialog" is a thing that can be shown.

h/static/styles/simple-search.scss
```diff
@@ -9,7 +9,7 @@
   @include clearfix;
   position: relative;
   padding: 0 1.5385em;
-  color: $gray;
+  color: $gray-darker;
```
tilgovi: The changes to the colors in this file could be a separate PR, maybe.

h/static/styles/simple-search.scss
```diff
@@ -23,14 +23,13 @@
...
 :not(:focus) ~ .simple-search-icon {
-  color: $gray-lighter;
+  color: $gray-light;
 }
 
 input.simple-search-input {
   float: left;
   outline: none;
   color: $text-color;
-  width: 100%;
```
tilgovi: Are we sure this causes no problems? I thought it was necessary or we'd end up with a short field.

h/static/styles/topbar.scss
```diff
@@ -41,7 +41,7 @@
 
     & > *:first-child, & > *:last-child {
       line-height: 28px;
-      margin: 0 .72em;
+      margin-right: .72em;
```
tilgovi: This doesn't seem right. It feels like, if we want to have no margin on the share button, that this should be an override done by the share button. The default for the topbar (on the stream and other pages) is definitely to have symmetric margins.

h/templates/app.html
```diff
@@ -28,6 +28,14 @@
          ng-click="login()"
          ng-switch-when="null">Sign in</a>
 
+      <!-- Share this page -->
+      <span class="share-dialog-toggle" ng-hide="isStream">
```
tilgovi: I'd give this its own variable. `isStream` is handled by the route controllers. It's a bare property on the scope, rather than a nested property of a scope object, and so even though it's not modified it still goes against the best practice of not shadowing scope variables.

I'd just be explicit about this. `showShareButton` or something.

In the future, we will eventually follow through on separating the stream and viewer widget templates, and maybe the share dialog will move into the latter. But for now, let's at least keep our language clear and not conflate things.

h/templates/client/share_dialog.html
```diff
@@ -0,0 +1,29 @@
+<div id="dialog" class="sheet content">
```
tilgovi: Indentation in this template is inconsistent with the rest of the project.

h/templates/help.html
```diff
@@ -84,8 +84,8 @@ <h3 class="feature-heading"><i class="feature-icon h-icon-reply"></i> Replies</h
         <h2 id="key-features" class="help-page-heading">Privacy</h2>
         <p class="help-page-lede">Annotations are either public and viewable by everyone
           or private and visible only to you.</p>
-        <div class="grid feature-content">
-          <section class="feature column-desktop-1-2">
```
tilgovi: Is this change intentional?

tilgovi:
@JakeHartnell gave some detailed feedback on some things. Most should be easy to change. I think the CSS for the share button smells a bit, but if you don't see an easy way to improve it I'm happy to take a look.

h/static/styles/simple-search.scss
JakeHartnell: It's to make them consistent with the new share icon, so I think it's related.

h/static/styles/topbar.scss
JakeHartnell: Yeah, but the reason, I changed this is because it makes the search icon look off on the stream. You can see this live on hypothes.is/stream:
selection_089

See how the search icon doesn't line up with sort.

tilgovi: Then you should adjust the CSS around media queries and the topbar to make this a symmetric change.

tilgovi: What I mean is that if you resize the view to respond to a smaller screen you'll see that it does line up. Make that always the case.

h/static/scripts/directive/share-dialog.coffee
JakeHartnell: Ah yes, sorry about that. Fixed.

JakeHartnell:
@tilgovi I made the changes you requested. 

> I think the CSS for the share button smells a bit, but if you don't see an easy way to improve it I'm happy to take a look.

The CSS for the *whole topbar* smells a bit IMO. As we are adding more things to the top bar, we're going to have to work on the CSS for it at some point, because with all the different floats going on, things are getting crazy. I think a rewrite is out of scope for this PR though, and I want to get this in front of users for feedback sooner rather than later. : )

tilgovi:
I didn't say we needed to write the whole top bar. It may be that we have to. If you don't know how to improve it without a rewrite, that's fine. That's why I offered to take a look, which I'll now do.

JakeHartnell:
Thanks @tilgovi.

JakeHartnell:
Bump. Can I get either explicit feedback about what I need to do to get someone to push the merge button, or can we merge and then re-factor later (which we will be doing anyway soon, there is much more work to do with regards to sharing)?

tilgovi:
Sure.

JakeHartnell:
Thanks @tilgovi!

robertknight:
I couldn't find the share facility anywhere until I read this bug report just now. I completely missed the positioning next to the search bar.

I was looking for it alongside the the rest of the actions relating to a specific annotation, I think for a couple of reasons:

* They are red so they function as a visible call-to-action
* I wanted to share not just the annotations on the whole page but a link that would direct someone to a specific annotation within the page.

dwhly:
> I was looking for it alongside the the rest of the actions relating to a specific annotation, I think for a couple of reasons.

We actually had a share link on the annotation card for a while, in the place of the link icon now-- which simply exposes a link to the standalone page for the annotation.

We plan to reintroduce it when direct linking (https://github.com/hypothesis/vision/issues/87) is finished.  (It's near the top of the backlog now).

PR #2216, opened 2015-05-06
Add nipsa flag column to user table

gergely-ujvari:
Trying the split tasks into small PRs, this is the first one.
This PR adds the new column to the account model and provides the migration script.

The new column is never used in this point, the next PRs will use it (behind a feature flag).


tilgovi:
Great.

PR #2220, opened 2015-05-06
Add adding a changelog entry to the pr docs

seanh:
- Add docs/CHANGES.rst, stub file that includes CHANGES.txt into Sphinx
- Add the changelog to the Sphinx toc
- Add a line to the _How to submit a pull request_ saying to add a line to the
  changelog

### DIFF
--- a/docs/CHANGES.rst
+++ b/docs/CHANGES.rst
@@ -1,0 +1,4 @@
+Changelog
+#########
+
+.. include:: ../CHANGES.txt
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,5 +1,5 @@
 0.4.0 (2015-05-05)
-=================
+==================
 
 Highlights
 ----------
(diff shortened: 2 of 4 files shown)


docs/CHANGES.rst
```diff
@@ -0,0 +1 @@
+../CHANGES.rst
```
tilgovi: Is it possible to have the header here and then link in the rest of the content from the CHANGES file?

I think I consciously used .txt instead of .rst because, along with LICENSE, I viewed the changelog as a file that should open in anyones default text edit when they unpack a tarball and have a minimum of formatting noise.

That may be a flawed thought, but it was a thought I had.

seanh: I can replace docs/CHANGES.rst with an actual file (not a symlink) like this:

```
Changelog
#########

.. include:: ../CHANGES.txt
```

As long as people remember when editing CHANGES.txt that it should be valid rst, I'm happy with that.

seanh: (I think putting the title actually in index.rst might mess up Sphinx's navigation structure)

seanh:
Force pushed, think @tilgovi will prefer this version

tilgovi:
Nice.

PR #2222, opened 2015-05-08
Remove via links from cards

gergely-ujvari:
Strips the via url from the annotation uri when rendering the annotation cards.

This PR will be simplified when #2215 will be merged as it'll use the via config defined here.
It handles annotations created with both the old and the new via url.

(An example annotation created with the old via url: https://api.hypothes.is/annotations/CosfKz2ZQpq2r9i3M5b1pg )

### DIFF
--- a/h/static/scripts/directive/annotation.coffee
+++ b/h/static/scripts/directive/annotation.coffee
@@ -50,7 +50,9 @@ AnnotationController = [
     highlight = model.$highlight
     original = null
     vm = this
-
+    via =
+      url: 'https://via.hypothes.is/'
+      oldUrls: ['https://via.hypothes.is/h/']
     ###*
     # @ngdoc method
     # @name annotation.AnnotationController#tagsAutoComplete.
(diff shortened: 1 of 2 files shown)


nickstenning:
To be honest, this doesn't seem to me like a problem that should be solved in the frontend, or at least not in a way that's specific to via. Doing it here creates new requirements for all API consumers, which doesn't seem right.

I'd suggest something more like:

- if a page doesn't have a `<link rel=canonical ...>`, then via should inject one.
- we use document canonical URIs as the preferred display URI
- we migrate existing annotations on via URLs to include the appropriate metadata

Yes, what I'm proposing is undoubtedly more work, but this is a generic problem associated with rewriting proxies, not just via. Manually fixing the display for each of these proxies in the frontend is a solution that can't really work in the long run.

tilgovi:
:+1:

And if we still have a mixture of Arrays and Strings for some fields can we just run a migration? It's silly to be haunted by something like that forever.

gergely-ujvari:
Okay, closing this in favor of the solution @nickstenning has suggested.

gergely-ujvari:
Okay, it seems we already had the pieces for this, but a bug in the document plugin prevented us from using the canonical uri.
We were already calling the document plugin `uri()` to get the URI for annotation.uri, but sadly there was an error in this function which prevented returning the canonical URI.
PR to fix it is here: https://github.com/openannotation/annotator/pull/529
call chain:
- https://github.com/hypothesis/h/blob/master/h/static/scripts/annotator/guest.coffee#L71
- https://github.com/hypothesis/h/blob/master/h/static/scripts/annotator/guest.coffee#L146
- https://github.com/hypothesis/h/blob/master/h/static/scripts/annotator/guest.coffee#L143
- https://github.com/openannotation/annotator/blob/v1.2.x/src/plugin/document.coffee#L13

So, after merging that PR we can update annotator and the newly created annotations will be automatically created with the canonical uri.

gergely-ujvari:
Just plus one short note. Via already injects the canonical uri into the web-page: https://github.com/hypothesis/via/blob/master/templates/banner.html#L2

gergely-ujvari:
Annotator version updated: https://github.com/hypothesis/h/commit/ee2c76c3b998da433eedad3cc7365e46d8b2b8a1

It this will go to prod, newly created annotations will be saved with the right uri.
All that remains is the migration script.

PR #2225, opened 2015-05-11
Hide the widget panel until ready for input

### LINKED ISSUE #2207: indicate when app is ready to respond
I just activated Hypothesis on [45 sites](https://hypothes.is/stream?q=user:judell&tag=hypothesis-in-the-press) in rapid succession. The chevron always appears quickly, and signals that the app is ready to respond, but the app doesn't actually become responsive for a while. The delay is unpredictable and varies widely. 

Until we are sufficiently fast and unobtrusive to remain always active, we should provide a clear indication that the app is ready to respond.

Straw man proposals:

- A spinner instead of the chevron/eye/note trio, until we are ready

- Grey the chevron/eye/note trio until we are ready

I care less about how, only that, we implement this. We should not always need it, but for now we do.

tilgovi:
Close #2207

### DIFF
--- a/h/static/scripts/annotator/host.coffee
+++ b/h/static/scripts/annotator/host.coffee
@@ -33,6 +33,7 @@ module.exports = class Host extends Guest
     .attr('src', src)
 
     @frame = $('<div></div>')
+    .css('display', 'none')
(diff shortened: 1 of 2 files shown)


nickstenning:
LGTM.

PR #2231, opened 2015-05-12
Change annotation card action from 'share' to 'link'

RawKStar77:
Updated text, tooltip, and icon for the link to the standalone page on our annotation. It now looks like this:
Annotation card with design update

This small PR was inspired by a conversation I had with @judell today and one with @dwhly earlier in the week about the share action that sits on our annotation cards currently.

*Rational for change:* with the work happening in #2215, I don't want to confuse the share iconography. Users expect that a share link on an annotation is a direct link that will take the person to the annotation *in context*. The link to the standalone page is useful because it conveys that this annotation is a true web resource, but having it under the share action is not very clear. This has come up in user testing before.

We should have a share button on annotations of course, but only when we have direct linking IMO.

Thoughts?


judell:
That feels right to me. In the olden days it would have been called a *permalink*. As Randall has pointed out, such a link may well outlive the page it was anchored to, but it remains a web resource that may be referenced in other contexts. 

That said, a separate Share action for the standalone annotation will be an echo of the broader app-level Share action which will (eventually) share a selected annotation (or maybe a search set) in context. Size, styling, and tooltips ("Share this view" vs "Share this annotation") will need to differentiate between the major and minor sense of Share.

Given that we will have "minor" share before "major" share, where will the share icon appear for now?

RawKStar77:
> Given that we will have "minor" share before "major" share, where will the share icon appear for now?

For now, here:
toolbar icon

And as soon as we can offer a direct to context link for a single annotation, then *also* on the cards.

judell:
OK, thanks. (Wishing for sharevia.dokku.hypothes.is so I can remind myself what's there.)

tilgovi:
Could I ask you to add the icon from from the content section of the icon set (http://google.github.io/material-design-icons/). The one you have is from the editor section. While they are visually identical, they don't have the same markup. Doing it as you've done breaks the association between the icon description and its function. It means that if, say, a refreshed v2 of the icon set were released, we'd have to check each icon to make sure it still makes sense where it's being used, otherwise a change to an icon might make sense for one context in which we use it but not another.

It's a minor thing, but I think it makes sense. Please, keep it clean. The names of the icons are not arbitrary. This is about linking to content, not adding a link to an editor.

RawKStar77:
Makes sense. Done.

tilgovi:
Nice. Thanks!

PR #2234, opened 2015-05-15
Stricter prospector linting

nickstenning:
Now that we don't fail the build on lint errors (but we do receive
feedback on pull requests) this commit ups the strictness of the
prospector run to "high".

I've special-cased some names which pylint complains about which are (in
my opinion) perfectly reasonable names.

### DIFF
--- a/.prospector.yaml
+++ b/.prospector.yaml
@@ -1,14 +1,30 @@
 output-format: grouped
 inherits:
   - full_pep8
-  - strictness_low
+  - strictness_high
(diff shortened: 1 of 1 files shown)


nickstenning:
:raised_hands:

seanh:
Can we do `strictness: high` rather than `inherits: - strictness_high`? Also for full pep8. Don't get why we use inherits

tilgovi:
Closed in favor of #2237

PR #2235, opened 2015-05-15
Introduce NIPSA

gergely-ujvari:
This PR implements a throw-away basic functionality for handling Nipsa flag for users.
Includes:
 - Add "NIPSA" feature flag
 - Add admin package (under accounts, behind the nipsa feature flag)
 - Add "admins" list in the config directory
 - Put "admin" flag to the claim data at authentication
 - Implement nipsa filtering
   - to search api
   - to realtime update
   - to stream load_more
- Minimal UI for setting NIPSA flag (see below the accounts menu entry)

Limitations & problems:
- Important: The nipsa flag is being read is read from db for every search, this can create a performance bottleneck. The list could be cached for a few minutes and/or the nipsa flag must receive an index immediately if this goes to prod.
-  The administration of admin users is very crude, simple text file at conf/admins (the exact names should be sorted out)
- No fancy UI for listing currently flagged users (but the information is available as a simple json dump)
- The total number of results returned by nipsa-filtering might not always be accurate. (We filter the data in the backend, at version 2 it should be done as part of the ES query. But with the current architecture it'd be far from trivial to put all relevant information (currently logged in user, feature flag, nipsa flagged- users) to the place (models.py) where the elasticsearch query is generated)

General notes:
- Our current feature flag system is too closely coupled with the pyramid registry which is cumbersome to say the least, this should be changed. (But not in the scope of this PR)
- Our python module separation is currently just an illusion. (We have to import many things i.e from api, account, etc.)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -25,6 +25,7 @@ clean:
 	find . -type d -name "__pycache__" -delete
 	rm -f h/static/scripts/vendor/*.min.js
 	rm -f h/static/scripts/account.*js
+	rm -f h/static/scripts/admin.*js
 	rm -f h/static/scripts/app.*js
 	rm -f h/static/scripts/config.*js
 	rm -f h/static/scripts/hypothesis.*js
--- a/conf/admins
+++ b/conf/admins
@@ -1,0 +1,6 @@
+dwhly
+judell
+tilgovi
+nickstenning
+csillag
+ujvari
--- a/conf/development.ini
+++ b/conf/development.ini
@@ -27,6 +27,7 @@ h.feature.api: True
 h.feature.claim: True
 h.feature.streamer: True
 h.feature.notification: False
+h.feature.nipsa: True
 
 horus.login_redirect: stream
 horus.logout_redirect: index
--- a/conf/production.ini
+++ b/conf/production.ini
@@ -23,6 +23,7 @@ h.feature.api: True
 h.feature.claim: False
 h.feature.streamer: True
 h.feature.notification: True
+h.feature.nipsa: False
 
 # User and group framework settings -- see horus documentation
 # Used by the local authentication provider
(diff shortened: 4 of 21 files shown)


csillag:
> Limitations & problems:

Oh, one more limitation: no new tests, for now.

tilgovi:
This is in the accounts package so it's not ignorant of groups and users, both of which already have tables.

So, no config file to list admins. Either the first user is an admin, or the command line can add admins, or something like this.

Don't worry about caching and please don't try to implement it in application code if you do. Anywhere we have SQL bottlenecks like this we should use dogpile.cache with Redis.

Please say more about module separation and maybe we can find a way to improve it for this PR.

gergely-ujvari:
A clarification: I've never intended to do caching in the app code. Only we have a comment about it. (Maybe the comment should be reworded)

About module separation:
Currently the feature flags are not reachable in the modules like api, for that to happen I needed to include it in the [create_api()](https://github.com/hypothesis/h/blob/nipsa2/h/app.py#L103) which is not the best. (This has been confirmed by @nickstenning )

Overall I think that tying the feature flags to pyramid config is not the best practice. The feature flags switch on/off pieces of code and not every piece of code has an easy access for the pyramid request/config objects and it is IMO not the best practice to "drag/carry" the request object into all our function calls.

My next concern was that our User db access is fragile from other modules. Maybe a gateway/interface should be needed, but currently I had to register a Session object for the IDBSession interface when I tried to access [user information from api](https://github.com/hypothesis/h/blob/nipsa2/h/api/views.py#L394)
I do hope that there is a more simple solution, but overall, from the api module I expect that I do not have to do any boilerplate to get info out of the account - User object.

So these are the things which are in my mind when talking about module separation issues.

PR #2237, opened 2015-05-15
Make our Prospector settings much stricter

seanh:


seanh:
There are probably quite a lot of specific messages that we want to disable as well

seanh:
Potential messages to ignore:

* pylint missing docstring
* pylint too few public methods
* pylint method could be a function (I actually like this one but it complains about things like test methods)
* Oh god I can't be bothered reading any more of this output!

nickstenning:
Oops. #2234. Not a big deal, and I probably should have commented on #2232 as you did.

nickstenning:
Yeah, that's a bit harsh.

nickstenning:
Let's crank it up a bit at a time, rather than just telling the whole world via a badge that we are terrible people. (P.S. We're not.)

seanh:
@nickstenning Can we go with this pull request then? What settings do you want to crank down?

nickstenning:
Well, I think you should pull in the good-names setting, certainly. Many of those are just noise.

seanh:
Waiting for landscape...

seanh:
Vulture was just crashing for me anyway

seanh:
A lot of this is still noise. I'm happy to go through it and disable messages (next week probably) and accept suggestions.

I think we should go for a middle-ground here. I have [a vim plugin](https://github.com/scrooloose/syntastic) that shows me prospector output in the sidebar as I code. The higher I have prospector cranked, the better the new code I write is going to be (though we should still disable messages we don't care about). Turning prospector up bit by bit may mean we're actually introducing more warnings as we write new code and just not seeing them cause we don't have it cranked high enough.

nickstenning:
Agree on some of your points, but I think it's important that the SNR is reasonably high, otherwise we'll just ignore landscape, and that would be sad.

seanh:
Alright. So let's see how high we can get it, by ignoring as much noise as possible? I assume landscape bot is gonna comment with another score at some point

seanh:
That was 86% with strictness high, 83% with veryhigh. I think it's worth keeping it up. Anyone see any more messages worth disabling, either globally or case-by-case? There probably are some: <https://landscape.io/github/hypothesis/h/770/messages>. There are probably some particular problem files worth going through and fixing too: <https://landscape.io/github/hypothesis/h/770/modules>

Our biggest problem by far is invalid names.

seanh:
If we get it up to I think 85% then we'll be Excellent instead of just Good

seanh:
I want to disable pylint's protected-access warning, but only in tests

tilgovi:
Would be happy to see it merge and I can look more once it's on master at what's worth fixing and disable things that profoundly annoy me. Not concerned about having an ~80% on our README.

tilgovi:
Thanks for this.

seanh:
Disabled protected-access in some test files (has to be done on a file-by-file basis afaict)

seanh:
Rebased. Let's just wait and see what Travis and Landscape make of this in case the rebase broke something, but after that I'm happy with this.

seanh:
I'm super happy to have this - I love having prospector on in my vim, and h's low prospector settings were kind of spoiling it. Just wish we had something like this for the rest of our codebase

seanh:
We've pushed up into the _Excellent!_ rating :)

tilgovi:
LGTM

tilgovi:
Actually, there are a few of these I'd like to change. I'll comment in the diff.

.prospector.yaml
```diff
@@ -1,14 +1,58 @@
...
+max-line-length: 79
 pep8:
-  options:
-    max-line-length: 79
+  full: true
 pylint:
   enable:
     - relative-import
+  disable:
+    - C0111  # Missing %s docstring
```
tilgovi: Undecided about this one. We could fix so many of these in less than an hour of adding some docstrings.

.prospector.yaml
```diff
@@ -1,14 +1,58 @@
...
+    - R0201  # Method could be a function
```
tilgovi: I like this one because I like the non-OO push it provides.

.prospector.yaml
```diff
@@ -1,14 +1,58 @@
...
+    - W0704  # Except doesn't do anything
```
tilgovi: Definitely want this one enabled. We should not have any empty except clauses. At the very least we should usually be logging them.

.prospector.yaml
```diff
@@ -1,14 +1,58 @@
...
+    # - k,v: dict iteration
+    # - db,fn: common abbreviations
+    # - fp: python idiom for file handles
+    #
+    # Some good "constant" names that pylint would otherwise reject:
+    #
+    # - log: common in "log = logging.getLogger(__name__)" pattern
+    # - parser: common in modules that use argparse
+    #
+    good-names: _,i,j,k,v,e,db,fn,fp,log,parser
```
tilgovi: Nice.

.prospector.yaml
```diff
@@ -1,14 +1,58 @@
...
+
+    # Allow longer function and method names.
```
tilgovi: Could we handle this instead by saying '(test_)|(<the default definition>)'?

tilgovi:
What's protected access about?

tilgovi:
Finally, go have a weekend :)

seanh:
It's accessing a function/method with an underscore, from another module/class. If you want to unit test that module/class you have to do it. (In Java the compiler wouldn't let you do this even in tests, you always had to test the public methods of classes only.)

.prospector.yaml
seanh: Good idea! That should work let's try it

.prospector.yaml
seanh: That was @nickstenning !

.prospector.yaml
seanh: Cool, let's put it back

.prospector.yaml
seanh: I like this one as well but it triggers very often in test methods. We could disable it in the test files specifically instead

.prospector.yaml
seanh: I'm happy for this to go back if you want

seanh:
Watching a movie now :) Ciao

tilgovi:
I sort of buy into the idea that testing protected methods and functions is not good.

If it's not exercisable from the public interface, should it be tested?

I think keeping that warning maybe encourages us to create better public interfaces and to test behavior rather than implementation.

tilgovi:
I'm going to get this on master. Do more tweaks if you like from there.

PR #2238, opened 2015-05-17
Replace gnsq with nsq-py

tilgovi:
nsq-py seems at least as well maintained, it is closer in API to
the official pynsq client, supports TCP connections to nsqd,
and supports the iteration protocol as a way to consume Reader
messages, which is simpler than the signal-based callbacks and
feels a bit more Pythonic.

Finally, it doesn't depend on gevent, and it's always nice to bake
that assumption into fewer parts of the system.

### DIFF
--- a/h/api/views.py
+++ b/h/api/views.py
@@ -194,7 +194,7 @@ def _publish_annotation_event(request, annotation, action):
         'annotation': annotation,
         'src_client_id': request.headers.get('X-Client-Id'),
     }
-    queue.publish('annotations', json.dumps(data))
+    queue.pub('annotations', json.dumps(data))
 
 
 def _api_error(request, reason, status_code):
--- a/h/notification/worker.py
+++ b/h/notification/worker.py
@@ -16,9 +16,7 @@ def run(request):
     the same channel. NSQ will distribute messages among available workers in a
     round-robin fashion.
     """
-    def handle_message(reader, message=None):
-        if message is None:
-            return
+    for message in request.get_queue_reader('annotations', 'notification'):
         data = json.loads(message.body)
         action = data['action']
         annotation = Annotation(**data['annotation'])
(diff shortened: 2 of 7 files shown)


nickstenning:
I'm not sure about this. You say "at least as well maintained", but compare the PyPI pages for [gnsq](https://pypi.python.org/pypi/gnsq) and [nsq-py](https://pypi.python.org/pypi/nsq-py). 

- gnsq has a README, a CHANGELOG, and was downloaded 1600 times in the last month. nsq-py has neither, and was downloaded 700 times. 
- The last released version of gnsq dates from February, the last released version of nsq-py dates from last July.
- gnsq [has docs](https://gnsq.readthedocs.org/en/latest/), nsq-py does not, afaict.

Also, I'm not sure what you mean by "supports TCP connections to nsqd" -- so does gnsq, but it doesn't really make a lot of sense to use a persistent connection for queue writes, so we don't. If we decide it does, then it's easy enough to change.

tilgovi:
I misread gnsq and thought it was only HTTP for sending. I've been noodling about the percolation replacement for the streaming filters and so I have in mind use where connected senders might be good.

tilgovi:
I do prefer the nsq-py API, which is almost identical to the official pynsq. I much prefer iteration and to worry about multiplexing listeners myself than to deal with this blinker thing and I think it's a little strange to have Reader/Nsqd rather than Reader/Writer and to change mpub and pub to multipublish and publish. I don't like the arguments http_port and tcp_port instead of nsqd_tcp_addresses, which supports multiple nsqd for publishing.

I understand why all the things are the way they are, but gnsq feels clunkier to me to use.

tilgovi:
I also have no idea why any client would use the HTTP interface instead of the TCP one, since the HTTP connection obviously requires a TCP connection anyway. It's not like it's UDP or something where one doesn't have to connect.

tilgovi:
None of that is to say we shouldn't use it.

PR #2239, opened 2015-05-18
Features cleanup

tilgovi:
This is a set of commits that cleans up the feature flag usage a bit and propagates feature flags to the frontend.

- Rather than use `request.feature` the function is available as a config directive: `config.feature('name')`. This is slightly more attractive and pyramidal to me than setting properties on the registry.

- Export the feature test function to jinja2 templates by adding it as a global for all the configured renderers during configuration time rather than during the `BeforeRender` event.

- Add jinja2 templating to the asset pipeline, including feature flags. Since browserify must be the inner-most filter, the variable substitution delimiters are surrounded with quotes so that simple templates with variable substitutions can pass through browserify before jinja2.

- Create a features configuration in the frontend as part of the config package.

- Make the `h` angular module depend on `h.config` and make the config package use the latter. By defining this package in the main app bundle but augmenting it in the config bundle we ensure a few things:
  - The application can load without a config bundle.
  - The application's providers are available to the config bundle for configuration.
  - The application will not start until the config bundle is loaded.
  - The feature service is available to all service constructors in the application.

### DIFF
--- a/conf/development.ini
+++ b/conf/development.ini
@@ -57,6 +57,7 @@ secret_key: notverysecretafterall
 sqlalchemy.url: sqlite:///.h.db
 
 jinja2.extensions: h.jinja_extensions:IncludeRawExtension
+jinja2.filters: json = json.dumps
 
 webassets.base_dir: h:static
 webassets.base_url: assets
--- a/conf/production.ini
+++ b/conf/production.ini
@@ -56,6 +56,7 @@ sqlalchemy.url: sqlite:///.h.db
 
 # http://docs.pylonsproject.org/projects/pyramid-jinja2/en/latest/index.html#jinja2-extensions
 jinja2.extensions: h.jinja_extensions:IncludeRawExtension
+jinja2.filters: json = json.dumps
 
 # Static asset configuration -- see webassets documentation
 # Until the next pyramid_webassets, don't change these.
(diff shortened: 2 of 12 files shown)


nickstenning:
I'm afraid I've had to revert this:

- it introduced a crashing bug on master (see https://github.com/hypothesis/h/pull/2131#discussion_r30692024)
- feature flags are no longer accessible by view functions (configurator object is not available from the request), thus this blocks #2131

The reason `feature` is on the registry is because the registry is available both at configuration time and during the request cycle. While obviously the parts of this PR that expose feature flags to the frontend are :+1:, I think the changes to the backend need a little more thought.

Also, review is good. Let's seek review proactively for changes like this.

tilgovi:
Okay....

I don't see where that pull request tries to access a feature flag from a
view, though.

I'll look at the crash you say.

tilgovi:
I'm not understanding which comment is about a crashing bug on master.

nickstenning:
[This is the use of `feature`](https://github.com/hypothesis/h/blob/bcb423a9a01309de621183b7d3dd2859464bfef5/h/accounts/views.py#L308) that is present both on master and (hence) #2131.

tilgovi:
Ahh, I thought I grepped for all these. I had a version of the change that had a request property for feature but I didn't see it used. May I just add that and restore this?

tilgovi:
And sorry for not understanding which comment, apparently the android client for GitHub doesn't know how to link to a comment.

nickstenning:
No worries -- I'm not averse to the change, I just wanted to unbreak master! By all means submit a PR.

PR #2241, opened 2015-05-19
Display KaTeX parse error upon faliure to parse

### LINKED ISSUE #1898: LateX errors fail silently and confuse user
Writing an erroneous formula in LateX resuts in the rendered content not updating at all. Clicking 'Preview' then shows the last succesful render, and clicking 'Save' does so too (until page reload, which shows an empty annotation).
Desirable improvements, from basic to nice:
- [ ] Show an error that rendering failed when pressing Preview or Save. (bare minimum fix)
- [ ] In the error message (or in the editor), link to an overview of supported latex.
- [ ] Tell in the error message what the error was (e.g. "unknown latex command \bf").
- [ ] Render the rest of the annotation normally, and in place of the failing formula show an error symbol, which shows the error message on hover.

Future improvements could be to support more commands (why do `\bf` or `\textbf` fail?), and having a wysiwyg editor for direct feedback (and as writing aid), but both are separate issues.

mbbaig:
Fix #1898 

unsupported latex:
latex-to-parse

errors displayed:
errors-displayed

### DIFF
--- a/h/static/scripts/directive/markdown.coffee
+++ b/h/static/scripts/directive/markdown.coffee
@@ -313,14 +313,19 @@ module.exports = ['$filter', '$sanitize', '$sce', '$timeout', ($filter, $sanitiz
         else
           endMath = index
         if startMath != null and endMath != null
-          math = katex.renderToString(textToCheck.substring(startMath, endMath))
-          textToCheck = (
-            textToCheck.substring(0, (startMath - 2)) + math +
-            textToCheck.substring(endMath + 2)
-          )
-          startMath = null
-          endMath = null
-          return renderInlineMath(textToCheck)
+          try
+            math = katex.renderToString(textToCheck.substring(startMath, endMath))
+            textToCheck = (
+              textToCheck.substring(0, (startMath - 2)) + math +
+              textToCheck.substring(endMath + 2)
+            )
+            startMath = null
+            endMath = null
+            return renderInlineMath(textToCheck)
+          catch
+            loadMathJax()
+            mathJaxFallback = true
+            $sanitize textToCheck.substring(startMath, endMath)
       return textToCheck
 
     # Re-render the markdown when the view needs updating.


tilgovi:
That looks really nice! How does it interact with the MathJax fallback?

mbbaig:
I couldn't find a way to get it to fallback to MathJax even without the error detection code. How would I do that?

tilgovi:
@RawKStar77 ??

mbbaig:
Okay. It actually uses the fallback. I think that's why it continues to render even after the error. So it renders what it can I suppose and displays errors when it encounters them.  I think :confused:

tilgovi:
Hmmm. I'm just curious why the whole block of math isn't rejecting.

tilgovi:
I'd have to study this code again. I'll take a look when I can.

mbbaig:
Yeah that is odd. For inline it does just display the error and not the text.

Okay I'll do the same.

RawKStar77:
Ok, so what's happening is it is catching the KaTeX errors, but the Mathjax fallback is working and the math is being rendered by Mathjax.

I think we should be catching Mathjax errors, since if KaTeX doesn't work we load Mathjax as a fallback. KaTeX should fail silently, but if Mathjax fails, we should *then* show an error message. You can read about how to listen and retrieve formatting errors for Mathjax here: http://docs.mathjax.org/en/latest/api/hub.html

pkra:
@RawKStar77 let us know if we can help with anything.

mbbaig:
Okay so I've been messing around with MathJax like @RawKStar77 suggested. I find that it doesn't need error detection code since it already displays errors quiet nicely.

correct-latex
This is the LaTex that I tested with since it's not supported by KaTex which would force a fallback.

correct-latex-preview
Here is what it looks like without any errors.

incorrect-latex-preview
And here is what it looks like with incorrect format.

As you can see it highlights the exact part that is wrong. The only think I found that could be potentially confusing is when \begin or \end are misspelled it doesn't highlight them.

So I'm not sure if there is a bug here. What do you guys think?

mbbaig:
@tilgovi anything on the error?

tilgovi:
I don't have a strong opinion here. What do you think should be done?

mbbaig:
In my opinion it displays the errors clearly enough. I couldn't replicate the problem from the original bug. So, unless someone else can, I say this is already solved.

mbbaig:
ok so the actual problem is in the inline rendering. I'm looking into it now.

mbbaig:
I added the mathjax fallback to the inline rendering method as well. Should now have the same functionality as the dedicated rendering.

RawKStar77:
This looks great to me! Thank you @mbbaig for looking into this issue! I think you might be right, and I think this is alas a lapse on my part since I wrote the code for this section with a substantial amount of help from @tilgovi.

mbbaig:
@RawKStar77 no problem :smiley: . Let me know if any of this code needs to be changed. I'll be looking into others bugs as well. Hopefully I learn more.

tilgovi:
Nice. Thanks!

PR #2243, opened 2015-05-19
Prospector tweaks

seanh:
@tilgovi


seanh:
I can't be bothered to look through all the new errors, landscape paginates them into 60 pages with no way to see them all on one page, but I suspect they're mostly missing docstrings

seanh:
Yep, tested locally and 516 of the new messages are docstrings.

nickstenning:
LGTM.

PR #2245, opened 2015-05-20
Upgrade to Angular 1.3

tilgovi:
Explanation for changes:

- toastr plugin has differing support between versions
- keys beginning with "$" are no longer stripped by angualar.toJson, so the store has a request interceptor now
- the directive controller function is called as a constructor; it should not return an object
- changes to `ngModel.$setViewValue`
  - parsers don't run if `setViewModel` does not change the value, so privacy directive must initially parse to undefined in order for `setLevel` to work
  - parsers can return objects and they are copied automatically (not sure the manually copying was ever necessary, anyway
- `ngModel` has a `$validators` key, so the match directive is much simpler
  - and one of the tests was bogus, since it was testing the internals of the isolated scope


nickstenning:
Okay, I'll try and check this out later. Given the magnitude of the change I think our test strategy probably needs to include some manual poking for this one.

h/static/scripts/directive/form-validate.coffee
```diff
@@ -2,15 +2,15 @@ module.exports = ->
   controller: ->
     controls = {}
 
-    addControl: (control) ->
+    this.addControl = (control) ->
```
nickstenning: Not sure I understand this change. From here it looks like controller is just a function, not a constructor, so does this even work?

nickstenning: Argh you already answered that question. Sorry.

nickstenning: But that said, if this is the case then `controls` should become an instance property rather than a closed-over variable.

nickstenning:
Sign up form and password change form validation both seem to be rather broken in this branch.

nickstenning:
I'm going to close this for now.

PR #2246, opened 2015-05-20
Add jshint and hound config

nickstenning:
Building on #2173 but implemented as a [Hound CI](https://houndci.com/) configuration.

Only JavaScript at the moment. We can turn on CoffeeScript later if we wish.

### DIFF
--- a/.hound.yml
+++ b/.hound.yml
@@ -1,0 +1,9 @@
+coffeescript:
+  enabled: false
+javascript:
+  config_file: .jshintrc
+  ignore_file: .jshintignore
+ruby:
+  enabled: false
+scss:
+  enabled: false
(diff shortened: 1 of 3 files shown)


seanh:
Running `jshint .` locally on this branch I'm getting 300+ errors. Most of them look reasonable but I think jshint can't handle embed.js (which is actually a jinja template), ignore that?

seanh:
This probably needs some more tweaking as well (I'm not familiar enough with our JavaScript yet to do it reasonably quickly) but I'd be happy to see this merged now with embed.js and then we work from there.

nickstenning:
Updated to ignore the `templates/` directory. For me `make clean && jshint .` now reports 84 errors.

nickstenning:
It's also worth noting that Hound only comments on diffs, so will only report on *new* violations.

seanh:
82 errors here and they all look like useful one to me

PR #2247, opened 2015-05-20
Features cleanup again

tilgovi:


tilgovi:
Maybe unnecessary to add it to the jinja globals if its available on the request, but for consistency with the asset pipeline templating which doesn't have the request maybe it's nice.

conf/development.ini
```diff
@@ -58,6 +58,7 @@ secret_key: notverysecretafterall
 sqlalchemy.url: sqlite:///.h.db
 
 jinja2.extensions: h.jinja_extensions:IncludeRawExtension
+jinja2.filters: json = json.dumps
```
nickstenning: It seems odd to me that this is app configuration. For example, you haven't currently added this to `test.ini`, so if the tests at any point in the future try and render `features.coffee` (or any other template which in the future uses the `json` filter), they'll explode.

Perhaps this (and, actually, the extensions configuration) should move into the code?

nickstenning:
I know we chatted about this briefly the other day, but on reflection I'm really not sure that we should be compiling feature flags into the built assets, which is what the jinja-fication and `features.coffee` implies.

Feature flags are not really application configuration, in the sense that they should be able to change during the application's runtime. I freely confess that there are places where this doesn't work at the moment, but I should be able to change feature flags in production without having to ship a new Chrome extension.

It would make more sense to me feature flags were exposed by the server, for example at `/app` along with other user state, and the client application retrieves those when it loads.

tilgovi:
Closing for now.

PR #2249, opened 2015-05-21
Fix display regression on help/first-run pages

nickstenning:
So, once upon a time in the mists of yesteryear, the help/first-run page looked like this:

Nice, innit?

Well unfortunately it now looks like this:

Note, in particular:

- misaligned logotype
- text [for ants](http://i.imgur.com/EqLMD2w.jpg)
- incorrect line-height (see bookmarklet button and the help text where the sidebar should go)

Turns out, this is due to the bundling of `topbar.scss` by 1f4ac1b. For reasons that aren't clear to me, `topbar.scss` interferes with `font-size` on `body`, and commits a couple of other heinous crimes before getting on with what I assume it was supposed to be for, namely styling the top bar.

Anyway, I've fixed this by:

- moving the scope-creepy CSS from `topbar.scss` into `app.scss`
- removing `app.scss` from the included code for the help page
- turning `help-page.scss` into a new top-level stylesheet, to be used by... wait for it... the help page

And here's what it looks like now:

And here's a diff between the "before" image (the first in this message) and the result of this PR:

You can see that the text has moved a few pixels down, but is otherwise unchanged. The movement of a few pixels is just a result of the styling of the header, which no longer relies on overriding styles from `app.scss`.

### DIFF
--- a/h/assets.yaml
+++ b/h/assets.yaml
@@ -216,3 +216,24 @@ inject_css:
         - h:static/styles/reset.scss
         - h:static/styles/variables.scss
         - h:static/styles/mixins/icons.scss
+
+
+# Help page
+help_page:
+  contents:
+    - help_page_css
+
+help_page_css:
+  output: styles/help-page.min.css
+  filters:
+    - cleancss
+    - cssrewrite
+  contents:
+    - h:static/styles/vendor/icomoon.css
+    - output: styles/help-page.css
+      filters: compass
+      contents:
+        - h:static/styles/help-page.scss
+      depends:
+        - h:static/styles/*.scss
+        - h:static/styles/mixins/*.scss
--- a/h/assets.yaml
+++ b/h/assets.yaml
@@ -155,26 +155,8 @@ app_css:
       contents:
         - h:static/styles/app.scss
       depends:
-        - h:static/styles/base.scss
-        - h:static/styles/annotations.scss
-        - h:static/styles/mixins/forms.scss
-        - h:static/styles/mixins/icons.scss
-        - h:static/styles/mixins/responsive.scss
-        - h:static/styles/mixins/grid.scss
-        - h:static/styles/common.scss
-        - h:static/styles/grid.scss
-        - h:static/styles/threads.scss
-        - h:static/styles/forms.scss
-        - h:static/styles/markdown-editor.scss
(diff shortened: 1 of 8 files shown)


nickstenning:
FWIW I'm not sure I'm a fan of `pyramid_layout`. It seems to add rather a lot of overhead to making relatively straightforward changes such as this one, and it doesn't really seem to be doing that much for us at the moment. Thoughts on stripping it out and just relying on some template inheritance and webassets?

judell:
Nice catch, thanks!

nickstenning:
Seeing as you were the last one to play with the CSS here, @tilgovi, can I request a review from you?

nickstenning:
Bump, @tilgovi.

tilgovi:
I think this is probably fine. I have one thing we need to check that I'll just flag since I haven't time right now and one thing that I think we should consider changing.

To check: that this doesn't break the layout or do anything nasty to the /login /register et al.

Te change: we should find a name for the layout and use it. It's a little strange to me that we have one template (help.html) that like three different pages (help, onboarding, dev/admin welcome) and that the layout is called "help". Even if we have a single layout and it's called "Page" for now, that's better than "Help" probably.

nickstenning:
I've double checked `/login` and `/register` and they are are identical before and after this change.

As for the name, I personally think "Help" is fine. All three pages this renders are help pages of one kind or another, and it's consistent with the name of the asset, which is consistent with the internal names of the classes within the CSS file. I'm assuming you don't want me to rename all of that.

tilgovi:
Not terribly important but I think we will at some point. Want to avoid
making all the things for every page.

nickstenning:
Let's do that when there's a page using that layout that isn't a help page. Yes?

tilgovi:
Sure.

h/assets.yaml
```diff
@@ -155,26 +155,8 @@ app_css:
       contents:
         - h:static/styles/app.scss
       depends:
```
tilgovi: We don't want `inject.scss` in this bundle.

We should consider starting to group our bundles on the filesystem, but right now the parallel construction we have suggests we should just move `inject.scss` into a `styles/annotator` folder.

tilgovi:
Going to merge and then do what I suggested.

PR #2253, opened 2015-05-26
Fix links in readme

seanh:


seanh:
Unfortunately the IRCCloud link gives an error a lot of the time, not because the link is wrong but looks like a problem on their end. If you keep refreshing the page it'll work eventually. Hopefully they'll fix this

PR #2254, opened 2015-05-26
Add pyroma dependency

seanh:
Need to run prospector


tilgovi:
You probably don't need to PR these things.

PR #2255, opened 2015-05-26
Add sphinx-httpdomain dependency

seanh:
Needed to build the docs

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -71,7 +71,8 @@ INSTALL_REQUIRES = [
     'repoze.sendmail<4.2',
 ]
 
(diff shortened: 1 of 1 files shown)


seanh:
How do people feel about adding a test that builds the docs, checks that it exits with status 0, and parses the Sphinx output to check there are no errors or warnings? That would have caught this missing dependency and will also catch mistakes people make when editing the docs.

tilgovi:
Yes to the test. The `-W` flag will make warnings into errors rather than having to parse the output.

tilgovi:
Alternatively, maybe readthedocs can email us on error. They're already building the docs on every push, so if that can be done no reason to have travis do it, too.

PR #2260, opened 2015-05-27
Add pyramid_redis_sessions dev dependency

seanh:
Needed to build the Chrome extension with production.ini

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -72,7 +72,7 @@ INSTALL_REQUIRES = [
 ]
 
 DEV_EXTRAS = ['pyramid_debugtoolbar>=2.1', 'prospector[with_pyroma]', 'pep257',
-              'sphinxcontrib-httpdomain']
+              'sphinxcontrib-httpdomain', 'pyramid_redis_sessions']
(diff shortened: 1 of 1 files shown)


tilgovi:
I'm not sure I think that makes it a dev dependency :-/.

It would certainly be better if the extension build process didn't require the session plugin to be loaded.

nickstenning:
+1 -- I don't think this is a dev dependency, we should rather do one of:

1. add `pip install -r requirements.txt` to the documentation for building a prod extension
2. remove the requirement to spin up the entire application in order to build the extension (more work, but ultimately preferable, I think)

seanh:
I hadn't even noticed we had a requirements.txt file. When do we use it? (Grepping for requirements.txt gives no results, and afaik we don't use it in the install process)

seanh:
> add pip install -r requirements.txt to the documentation for building a prod extension

I'll probably do this just because A) I want the docs for building the Chrome extension to work without error and B) I want to get past this quickly and work on nipsa instead

tilgovi:
Please update the docs, that'd be awesome.

requirements.txt is used in our Dockerfile

It's also used automatically by PaaS systems likes Heroku and Dokku.

seanh:
It's gone into the docs pr

Issue #1992, opened 2015-03-04
Highlight disappears after clicking adder icon


BigBlueHat:
At least not until you focus on that card in the sidebar.

In the intervening time it looks like something broke--as your select text disappears.

New annotations should probably have their highlights stay on the page--with or without hover.

Or better yet, "Show Annotations" (at least mine) should be on when I'm annotating...

csillag:
We used to have a class called `annotation-hl-temporary`, which had exectly this rule.
(It decorated the HL of the annotation that was currently being created.)

Later it was removed, but I don't remember whether or not that was a conscious decision or not, and if it was, what was the rationale.

Anyway, I agree with the proposal in the OP. We should bring it back.

tilgovi:
The decision was conscious. It was made my be. The styling was no different than a regular highlight, so I just let that be the rule.

BigBlueHat:
The problem isn't that they matched, but that the non-temporary one doesn't
show up soon enough--and we need something there as a place holder (or make
the real in-progress highlight stay around some other way) while the person
makes their changes.

Lots of ways to come at this (as ever). No preference (here) on how it's
fixed.

csillag:
So @tilgovi, would you be OK with using the default style for regular highlights for the annotation that is currently being created?

tilgovi:
Sounds reasonable. But there may be some trickiness around selection color and making sure that we both a) show the new highlight the same as existing ones and b) still make it possible to see your overlap selections. Might be a matter of making sure we deselect on adder click?

csillag:
> there may be some trickiness around selection color and making sure that we both
> a) show the new highlight the same as existing ones and
> b) still make it possible to see your overlap selections.
>
> Might be a matter of making sure we deselect on adder click?

I'm not sure I fully understand the concern here. We [deselect](https://github.com/hypothesis/h/blob/25efdd689f7d7008bf1f5f3e4ef1bbba2576ef01/h/static/scripts/guest.coffee#L402) on adder click.
Fwiw, I'll just add a separate CSS class for now, making it easier to tweak the style later.

RawKStar77:
More discussion relevant to this issue was had here: https://github.com/hypothesis/h/pull/2264

I  think a way to solve this issue is to modify how the set visibility conrtol works so that highlights that I make are always visible, highlights that others make are toggled by the control.

This would solve this issue. 

@judell commented on #2264:

[...]

PR #2264, opened 2015-06-01
setVisibleHighlights to true when making new annotation

RawKStar77:
Meant to resolve https://github.com/hypothesis/h/issues/1992: *Highlight disappears after clicking adder icon.*

This simply sets visible highlights to true when making a new annotation, like we do when making a highlight. While we could do something complicated like introducing more CSS classes to highlights, I think this is a good enough usability improvement for now.

### DIFF
--- a/h/static/scripts/annotator/guest.coffee
+++ b/h/static/scripts/annotator/guest.coffee
@@ -363,6 +363,7 @@ module.exports = class Guest extends Annotator
         this.setVisibleHighlights true
         this.setupAnnotation(this.createHighlight())
       when 'comment'
+        this.setVisibleHighlights true
         this.setupAnnotation(this.createAnnotation())
         this.triggerShowFrame()
     Annotator.Util.getGlobal().getSelection().removeAllRanges()
--- a/h/static/scripts/annotator/test/guest-test.coffee
+++ b/h/static/scripts/annotator/test/guest-test.coffee
@@ -288,6 +288,23 @@ describe 'Guest', ->
       guest.onAdderMouseup(event)
       assert.isTrue(event.isPropagationStopped())
 
+  describe 'onAdderClick', ->
+    it 'it sets visiblehighlights to true when making a highlight', () ->
+      guest = createGuest()
+      event = jQuery.Event('click')
+      guest.setVisibleHighlights = sandbox.stub()
+      guest.onAdderClick target: dataset: action: "highlight"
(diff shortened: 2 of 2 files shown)


nickstenning:
Needs a test :)

tilgovi:
If we're going to do this maybe we should just disable the set visible highlights control. Otherwise, as a user, if I want to not have highlights I have to turn them off after everything I do.

RawKStar77:
@nickstenning, added some tests.

@tilgovi, I think we should certainly think more about the set visible highlights control. Perhaps the set visible highlights control should hide/show all but the user's highlights? In other words, *highlights that I make are always visible, highlights that others make are toggled by the control.*

Rational, when *I* make annotations I want to see them, but often I don't want to see a bunch of other people's highlights on the page (especially if there are lots of people annotating a document). 

IMHO, that would make the set visible highlights control a bit more useful.

tilgovi:
Yes, and if we think the same goes for annotations as well we could just tie this to the container scope, so rather than turning highlights on/off for that use case one would flip themselves into the "my notebook" / "my annotations" / whathaveyou group view.

judell:
"when *I* make annotations I want to see them, but often I don't want to
see a bunch of other people's highlights"

For me it depends on who they are and why we are together in the document.

When I was co-editing a document with Maryann and Dan last week I often
toggled between no highlights, so I could read without disruption, and
highlights to co-edit.

If I'm annotating a public page I might want the same kind of experience,
or might prefer to see only mine.

In all of these scenarios we have a mostly-hidden mechanism that, if/when
exposed, would imply granular control of highlights.

At http://blog.jonudell.net/2015/01/22/a-federated-wikipedia/ there are
highlights from judell and BigBlueHat. When I discovered that search could
limit the sidebar by user, I expected that user:judell would not only hide
BigBlueHat's annotations in the sidebar, but also his highlights. Turns out
it doesn't.

Are we going to reveal the sidebar search facets (
https://github.com/hypothesis/h/pull/2215#issuecomment-106010871)?

tilgovi:
@judell whatever you can recommend here would be great, I just wanted to bring up the question here about whether it would be annoying to have the highlights flip on every time a user makes any type of annotation.

RawKStar77:
Closing this because this is quite the solution we are looking for, but I feel like we've had some good discussion here on what we actually need to do to resolve #1992 and improve the set visible highlights control.

PR #2265, opened 2015-06-02
Add docs for building the Chrome extension

seanh:


docs/hacking/chrome-extension.rst
```diff
@@ -0,0 +1,60 @@
...
+
+   Once it has generated an ID for your local build of the extension Chrome
+   will always use the same ID (even when you rebuild and reload the
+   extension), so you'll only have to do this once:
+
+   .. code-block:: bash
+
+      hypothesis-buildext conf/development.ini chrome
+          --base   'http://127.0.0.1:5000'
+          --assets 'chrome-extension://notarealid/public'
```
nickstenning: Hmm. This isn't how I usually build a dev extension. Just omit `--assets` and it will load assets from the application. That gets around the need to build everything twice.

docs/hacking/firefox-extension.rst
```diff
@@ -16,11 +17,10 @@ To build the browser extensions, use the ``hypothesis-buildext`` tool::
         chrome          build the Google Chrome extension
         firefox         build the Mozilla Firefox extension
 
-At this point, a working extension should exist in either ``./build/chrome``
-or ``./build/firefox``. If the development configuration was used, static
-assets are loaded from the server. Start the application and ensure that the
-assets are built by visiting the start page or by running the ``assets``
-command::
+At this point, a working extension should exist in either ``./build/firefox``.
```
nickstenning: `s/either//`

docs/hacking/chrome-extension.rst
seanh: Isn't that an unnecessary difference between dev and prod extensions though?

seanh: I'll fix this after lunch, I've got an idea

nickstenning: Well, it's a difference, and perhaps not an ideal one, but it's mighty convenient if you want to iterate on the application code without rebuilding your Chrome extension every three seconds.

seanh:
@nickstenning Ok, force pushed this. I've redesigned these docs. I've realised that there are 3 valid ways to build the extension:

1. Standard dev extension. Won't workon https sites (including the hypothes.is welcome page it redirects to!) or csp sites

2. The same but with https://127.0.0.1:5000. Still won't work on GitHub

3. How we build it for prod: with production.ini and packed assets. But with your local (https) base URL and extension ID. Will work on GitHub.

So I've documented those three and also a variety of errors that you can get depending on how you've built the extension and that have confused me and wasted my time.

There are various invalid ways to combine the `hypothesis-buildext` options, for example:

* Using conf/development.ini and --assets 'chrome-extension://<id>/public together builds a broken extension (documented)

* Running h on https and building the extension with http (or vice-versa) causes confusing errors (documented)

* You can build with packed assets and an http base URL. You'll get past the CSP on GitHub but run straight into mixed content. Are there any sites that are gonna block us with CSP but not be running on https?

Too many ways to combine the options of this command, several of which are just invalid, too many confusing errors you can get depending on how you've combined those options and how you're running h ... Makes me wonder about simplifying the command to support only the 3 valid configs above.

But for now, docs.

seanh:
Ok, really done rebasing now, ready for review

seanh:
PS The only reason why the first way of building the extension is needed is that the secone two are more inconvenient, and the only reason the second is needed is because the third is inconvenient. Otherwise there'd never be any reason to use http or not to pack the assets.

nickstenning:
This is awesome and you are wonderful. Thank you. :fireworks:

judell:
Echoing Nick's thanks. 

(When you say "work on GitHub" you mean any CSP-enforcing site?)

tilgovi:
So nice.

seanh:
> (When you say "work on GitHub" you mean any CSP-enforcing site?)

Yes (well, GitHub is the only one I tested on, I think it'll work on any site that our prod extension works on though)

PR #2271, opened 2015-06-03
Don't explode on empty bodies in AsyncFormViewMapper

nickstenning:
This is a quick fix for an error currently in staging which prevents logout from working.

Obviously this should have tests, but testing this code is totally horrendous. I've nearly finished some work to improve this substantially on the `dehorusify-profilecontroller` branch, and I'll be sure to test this code path there. For now, I think we should merge this and push it to staging.

### DIFF
--- a/h/accounts/views.py
+++ b/h/accounts/views.py
@@ -30,7 +30,10 @@ def ajax_form(request, result):
 
     if isinstance(result, httpexceptions.HTTPRedirection):
         request.response.headers.extend(result.headers)
-        result = result.json
+        if result.body != '':
+            result = result.json
+        else:
+            result = {}
         result["status"] = "okay"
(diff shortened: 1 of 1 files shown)


h/accounts/views.py
```diff
@@ -30,7 +30,10 @@ def ajax_form(request, result):
 
     if isinstance(result, httpexceptions.HTTPRedirection):
         request.response.headers.extend(result.headers)
-        result = result.json
+        if result.body != '':
```
seanh: I take it that accessing `result.json` raises an exception? Could it be better to do a try and catch the exception and set result to `{}` in the `except`.

nickstenning: I don't want to catch cases where we're actually returning invalid JSON from a view function. I just want to to catch the case where the body is empty.

PR #2272, opened 2015-06-03
Set a server-side default for the NIPSA column

nickstenning:
This should really be done for all migrations that involve adding a non-nullable column. This makes it possible to run the migration in staging without breaking production.

You will need to throw away your development database (or do something fiddly like downgrade to 4a97f680ecca first and then upgrade to 29d0200ba8a9).


seanh:
Merged

tilgovi:
Thanks, @nick

tilgovi:
Thanks, @nickstenning

PR #2274, opened 2015-06-03
Dehorusify ProfileController

nickstenning:
As a first step towards remove horus completely from our accounts system, here's the ProfileController reworked to not depend on horus. Along the way I've fixed the following bugs:

- logout not working on master (also fixed temporarily by #2270)
- update anyone's subscriptions simply by picking a subscription id

The tests are now an order of magnitude easier to write.

### DIFF
--- a/h/accounts/test/views_test.py
+++ b/h/accounts/test/views_test.py
@@ -208,13 +208,12 @@ def test_asyncformviewmapper_preserves_email_in_response():
 
 @pytest.mark.usefixtures('activation_model',
                          'dummy_db_session')
-def test_disable_invalid_password(config, form_validator, user_model):
+def test_disable_user_with_invalid_password(config, form_validator, user_model):
     """
     Make sure our disable_user call validates the user password
     """
-    request = _get_fake_request('john', 'doe')
+    request = DummyRequest(method='POST')
     form_validator.return_value = (None, {"username": "john", "pwd": "doe"})
-    configure(config)
 
     # With an invalid password, get_user returns None
     user_model.get_user.return_value = None
@@ -226,15 +225,13 @@ def test_disable_invalid_password(config, form_validator, user_model):
     assert any('pwd' in err for err in result['errors'])
 
 
-@pytest.mark.usefixtures('activation_model',
-                         'dummy_db_session')
-def test_user_disabled(config, form_validator, user_model):
+@pytest.mark.usefixtures('activation_model', 'dummy_db_session')
+def test_disable_user_sets_random_password(config, form_validator, user_model):
     """
     Check if the user is disabled
     """
-    request = _get_fake_request('john', 'doe')
+    request = DummyRequest(method='POST')
     form_validator.return_value = (None, {"username": "john", "pwd": "doe"})
-    configure(config)
 
     user = FakeUser(password='abc')
     user_model.get_user.return_value = user
--- a/h/accounts/test/views_test.py
+++ b/h/accounts/test/views_test.py
@@ -45,21 +45,6 @@ def configure(config):
     config.registry.feature.return_value = None
 
 
-def _get_fake_request(username, password):
-    fake_request = DummyRequest()
-
-    def get_fake_token():
-        return 'fake_token'
-
-    fake_request.method = 'POST'
-    fake_request.params['csrf_token'] = 'fake_token'
-    fake_request.session.get_csrf_token = get_fake_token
-    fake_request.POST['username'] = username
-    fake_request.POST['pwd'] = password
-
-    return fake_request
-
-
 # A fake version of colander.Invalid for use when testing validate_form
 FakeInvalid = namedtuple('FakeInvalid', 'children')
 
--- a/h/accounts/test/views_test.py
+++ b/h/accounts/test/views_test.py
@@ -22,6 +22,7 @@ from horus.strings import UIStringsBase
 
 from h.accounts import schemas
 from h.accounts import views
+from h.accounts.views import ajax_form
 from h.accounts.views import validate_form
 from h.accounts.views import RegisterController
 from h.accounts.views import ProfileController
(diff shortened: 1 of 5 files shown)


h/accounts/test/views_test.py
```diff
@@ -48,228 +46,275 @@ def configure(config):
...
+    assert result['reason'] == 'I asplode!'
 
-    if with_subscriptions:
-        subs = '{"active": activestate, "uri": "username", "id": 1}'
-        subs = subs.replace('activestate', str(active).lower()).replace('username', username)
-        fake_request.POST['subscriptions'] = subs
-    return fake_request
 
+def test_validate_form_passes_data_to_validate():
+    idata = {}
```
seanh: idata? Does this mean input_data?

h/accounts/views.py
```diff
@@ -64,6 +60,16 @@ def ajax_form(request, result):
     return result
 
 
+def validate_form(form, data):
+    """Validate POST payload data for a form."""
+    try:
```
seanh: Nice

h/accounts/views.py
```diff
@@ -234,126 +240,98 @@ class AsyncRegisterController(RegisterController):
...
-    except deform.ValidationFailure as err:
-        raise _InvalidEditProfileRequestError(errors=err.error.children)
-
-
 @view_auth_defaults
 @view_config(attr='edit_profile', route_name='edit_profile')
 @view_config(attr='disable_user', route_name='disable_user')
 @view_config(attr='profile', route_name='profile')
-class ProfileController(horus.views.ProfileController):
+class ProfileController(object):
```
seanh: Yes :) I love classes that inherit from just object

nickstenning:
> For what it's worth the value in wrapping the tests for a given function in a class is making it easier to find the tests for that function and know that you've found all of them

I get that, but don't you achieve the same thing by:

- putting them all in the same place
- giving them good names? (i.e. `test_edit_profile_does_foo`, `test_edit_profile_does_bar`)

seanh:
`edit_profile()` couild do with being broken up into multiple methods, but it was already as long before this pr

seanh:
This all looks good to me, code is much simpler and easier to understand.

I'm a little disturbed by how long `edit_profile()` is and how few tests we have for `edit_profile()` and `ProfileSchema`, that isn't this PR's fault, but it might be good to prioritise writing more tests for those.

seanh:
> > For what it's worth the value in wrapping the tests for a given function in a class is making it easier to find the tests for that function and know that you've found all of them
I get that, but don't you achieve the same thing by:
>
> putting them all in the same place
giving them good names? (i.e. test_edit_profile_does_foo, test_edit_profile_does_bar)

Sort of. But it's the difference between scrolling through a long tests file and noticing "Hey, the function names are starting with test_edit_profile_ now" versus seeing a class with a docstring that says "This is where all the tests for edit_profile() go", which I think is clearer. The indentation difference and ability to search for "class " makes it easier to jump around in and get an overview of the file.

I'm happy enough either way though, in past projects I've tried to insist on not having test classes so that people wouldn't be tempted to start using `self`.

seanh:
I think some of landscape's complaints about the tests are actually valid: unused variables, unused imports, == True, line too long. But not all the missing docstring and other docstring complaints, not the redefined name or the unused arguments. And I'm not sure if the valid complaints are really introduced by this pr.

seanh:
I'm happy with this, recommend @tilgovi has a quick look as I think he understands horus, colander and deform better than I do.

tilgovi:
I'd say if test files are getting so long that it becomes a pain to identify and find a subset of them that you need it might mean the tested module has grown similarly too large and needs to be broken up. I'm not familiar with the pain @seanh is suggestion, but let's watch for it to occur.

tilgovi:
I'm general a fan of fewer classes in most things, so I'm tempted to agree with @nickstenning here for now.

h/accounts/test/views_test.py
tilgovi: If we're looking for better names for these things I would suggest maybe the ones that colander and deform use:

cstruct and appstruct. So cstruct is short for "colander struct". http://docs.pylonsproject.org/projects/colander/en/latest/glossary.html#glossary

The idea is that your form sends a formencoded sequence that is a colander struct, and colander deserializes this into an appstruct, or application structure.

In our case, with AJAX and JSON, this isn't actually a cstruct, probably, since we're skipping any deserialization colander might do, afaik, since part of the purpose of colander is to take a sequence of formencoded values submitted by a dumb, unscripted form, and turn that into a hierarchical structure that represents the logical form rather than a sequence of input controls.

<shrug>

So, it might be `in_appstruct`. Honestly, I'm not picky about the tests here, though.

h/accounts/views.py
tilgovi: :+1:

h/accounts/views.py
```diff
@@ -234,126 +240,98 @@ class AsyncRegisterController(RegisterController):
...
+        self.schema = schemas.ProfileSchema().bind(request=self.request)
+        self.form = deform.Form(self.schema)
+
     def edit_profile(self):
-        try:
-            appstruct = _validate_edit_profile_request(self.request)
-        except _InvalidEditProfileRequestError as err:
-            return dict(errors=err.errors)
+        """Handle POST payload from profile update form."""
+        if self.request.method != 'POST':
```
tilgovi: This can also be done by adding the `request_method` predicate to the view decorator, but this is okay.

h/accounts/views.py
```diff
@@ -234,126 +240,98 @@ class AsyncRegisterController(RegisterController):
...
-        form = deform.Form(schema)
+            response['model']['email'] = user.email = email
 
-        try:
-            appstruct = form.validate(request.POST.items())
-        except deform.ValidationFailure as e:
-            return dict(errors=e.error.children)
+        password = appstruct.get('password')
+        if password:
+            user.password = password
```
tilgovi: I think we're forgetting to add the updated user to the db session so it commits at the end of the request.

h/accounts/views.py
```diff
@@ -234,126 +240,98 @@ class AsyncRegisterController(RegisterController):
...
         username = appstruct['username']
         pwd = appstruct['pwd']
 
         # Password check
-        user = self.User.get_user(request, username, pwd)
+        user = User.get_user(self.request, username, pwd)
         if user:
             # TODO: maybe have an explicit disabled flag in the status
-            user.password = self.User.generate_random_password()
-            self.db.add(user)
```
tilgovi: I think this needs to be here.

h/accounts/views.py
```diff
@@ -367,7 +345,6 @@ def unsubscribe(self):
         subscription = Subscriptions.get_by_id(request, subscription_id)
         if subscription:
             subscription.active = False
-            self.db.add(subscription)
```
tilgovi: This, too.

tilgovi:
Totally owe you :beers: for this one. Makes me so happy.

Only thing missing is actually adding the changes to the db :).

h/accounts/views.py
nickstenning: You don't need to add objects that came from the database in the first place. sqlalchemy does dirty checking. You only need to add new objects.

nickstenning: And [here's a reference for that assertion](http://docs.sqlalchemy.org/en/latest/orm/session_basics.html#adding-new-or-existing-items) :)

nickstenning:
I'll just take a quick look through some of the things Sean raised, so don't merge it juuust yet.

tilgovi:
Okay. I've also just checked this locally and was able to change my password, so I'm confused about SQLAlchemy sessions and the need for an explicit `add`. If you can clarify why it's not needed that would help me.

tilgovi:
http://docs.sqlalchemy.org/en/latest/orm/session_state_management.html#quickie-intro-to-object-states

Right. I was mistaken because I'd seen excessive use of `add` in other code, I think.

If the object came from the DB it's already persistent and its attributes are managed so when the transaction commits at the end of the request it's done.

nickstenning:
Oh, sorry, I did already link to some sqla docs in the line comment above.

tilgovi:
No worries.

tilgovi:
LGTM. Whenever @seanh is happy.

seanh:
LGTM, I'll leave it up to Nick to correct as many of my points as he wants, none of them were blockers

nickstenning:
Ready for merge when the lights go green.

Issue #2273, opened 2015-06-03
make dev broken (gunicorn ssl error)


seanh:
On master for me today:

```
seanh@mistakenot ~/P/h/h make dev
2015-06-03 12:18:51,646 [8573] [gunicorn.error:INFO] Starting gunicorn 19.3.0
2015-06-03 12:18:51,646 [8573] [gunicorn.error:INFO] Listening at: http://127.0.0.1:5000 (8573)
2015-06-03 12:18:51,647 [8573] [gunicorn.error:INFO] Using worker: h.server.Worker
2015-06-03 12:18:51,648 [8580] [gunicorn.error:INFO] Booting worker with pid: 8580
2015-06-03 12:18:51,655 [8580] [gunicorn.error:ERROR] Exception in worker process:
Traceback (most recent call last):
  File "/home/seanh/.virtualenvs/h/local/lib/python2.7/site-packages/gunicorn/arbiter.py", line 507, in spawn_worker
    worker.init_process()
  File "/home/seanh/.virtualenvs/h/local/lib/python2.7/site-packages/gunicorn/workers/ggevent.py", line 185, in init_process
    self.patch()
  File "/home/seanh/.virtualenvs/h/local/lib/python2.7/site-packages/gunicorn/workers/ggevent.py", line 65, in patch
    monkey.patch_all(subprocess=True)
  File "/home/seanh/.virtualenvs/h/local/lib/python2.7/site-packages/gevent/monkey.py", line 200, in patch_all
    patch_ssl()
  File "/home/seanh/.virtualenvs/h/local/lib/python2.7/site-packages/gevent/monkey.py", line 146, in patch_ssl
    patch_module('ssl')
  File "/home/seanh/.virtualenvs/h/local/lib/python2.7/site-packages/gevent/monkey.py", line 62, in patch_module
    gevent_module = getattr(__import__('gevent.' + name), name)
  File "/home/seanh/.virtualenvs/h/local/lib/python2.7/site-packages/gevent/ssl.py", line 5, in <module>
    from gevent import _sslgte279 as _source
  File "/home/seanh/.virtualenvs/h/local/lib/python2.7/site-packages/gevent/_sslgte279.py", line 46, in <module>
    orig_SSLContext = __ssl__.SSLContext
AttributeError: 'module' object has no attribute 'SSLContext'
Traceback (most recent call last):
[...]

PR #2275, opened 2015-06-03
Fix gunicorn ssl crash on Python < 2.7.9

seanh:
Possible fix for #2273. I'm not sure what <https://github.com/hypothesis/h/commit/c3c99186549e6ecdae54e6743aa90b9baae78cca> is actually about (@tilgovi - more links?) but it makes gunicorn crash for me and googling it sounds like it may be because SSLContext is a Python 2.7.9 thing and I'm on Python 2.7.6 (Ubuntu 14.04). I guess this may be a gunicorn or gevent bug. This commit gets rid of the crash at least

### DIFF
--- a/gunicorn.conf.py
+++ b/gunicorn.conf.py
@@ -1,5 +1,6 @@
 # -*- coding: utf-8 -*-
 import os
+import sys
 import urlparse
 
 # Smart detect heroku stack and assume a trusted proxy.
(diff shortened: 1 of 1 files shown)


nickstenning:
I'm not sure this makes sense. The whole point of that commit was to trick gunicorn into treating Python versions < 2.7.9 as Python >= 2.7.9 when the `_sslwrap` function is missing. This is because Debian/Ubuntu have (apparently) backported the changes that were made to 2.7.9 to earlier Python versions. If it's exploding it's because the fix is wrong, not because it's being applied to Python < 2.7.9.

seanh:
Closing, replaced with https://github.com/hypothesis/h/pull/2277

PR #2276, opened 2015-06-03
Don't run prospector over test packages

nickstenning:
My motivation for this change is to reduce noise in the Prospector messages. PR #2274 pretty clearly improves "health", in my (biased) opinion, but its Prospector score is dragged down because I added [a bunch of test functions without docstrings](https://github.com/hypothesis/h/pull/2274/files#diff-bb25916b4db2b824f06e8d65a00b2945R53).

I don't think those test functions would be improved by docstrings, and might actually be made worse -- more for the eye to scan can actually hinder understanding.

In addition, there are some patterns for `py.test` (function params shadowing global fixture functions) which PyLint complains about but for which there is no easy fix.

### DIFF
--- a/.prospector.yaml
+++ b/.prospector.yaml
@@ -1,6 +1,5 @@
 output-format: grouped
 strictness: veryhigh
-test-warnings: true
 doc-warnings: true
 max-line-length: 79
 pep8:


seanh:
I lean towards leaving prospector on for tests but disabling certain messages (and if we can disable them across all tests in one place that'd be good). But I don't mind merging this if I'm outvoted. @tilgovi ?

nickstenning:
I've had a look and I don't think you can easily disable named checks for a directory. Happy to be wrong though.

tilgovi:
Fine for this and going to merge it. @seanh if you are able to find a way to disable certain messages for the tests that'd be great, but until then I think this is better than the noise it causes.

PR #2279, opened 2015-06-03
Keep app css and inject css separate

tilgovi:
Structure is good. Bonus fewer false dependencies on asset rebuilds
for the app.


nickstenning:
Great stuff.

PR #2281, opened 2015-06-03
Dehorusify the AuthController

nickstenning:
Remove the AuthController's dependency on horus. Part of the ongoing process to remove our dependency on this library.

(I couldn't resist the temptation because this one looked so easy. Expect more like this tomorrow...)

### DIFF
--- a/h/accounts/views.py
+++ b/h/accounts/views.py
@@ -11,6 +11,7 @@ from horus.interfaces import IForgotPasswordForm
 from horus.interfaces import IForgotPasswordSchema
 from pyramid import httpexceptions
 from pyramid.view import view_config, view_defaults
+from pyramid.security import forget
 from pyramid.url import route_url
 from pyramid_mailer import get_mailer
 from pyramid_mailer.message import Message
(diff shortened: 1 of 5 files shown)


h/accounts/views.py
```diff
@@ -115,27 +116,34 @@ def wrapper(context, request):
...
-            user = self.form.validate(self.request.POST.items())['user']
-        except deform.ValidationFailure as e:
-            return {
-                'status': 'failure',
-                'errors': e.error.children,
-                'reason': e.error.msg,
-            }
+        err, appstruct = validate_form(self.form, self.request.POST.items())
+        if err is not None:
+            return err
```
seanh: I've seen this pattern a few times. Just a thought: would it be nice if controller methods could return an error to the browser by raising an exception, rather than by returning an error object? Because then they could just blindly call `appstruct = validate_form()` which could raise the exception for them if validation fails.

seanh:
Landscape is complaining about missing docstrings on `login()` and `logout()` but I think they are obvious enough

seanh:
These controllers are _so_ easy to understand now

tilgovi:
All the :beers:.

h/accounts/views.py
nickstenning: Agreed. In time :) -- trying to make the minimal changes that kill off horus atm. I think there's loads more we can do on this whole module once that's done.

PR #2282, opened 2015-06-04
Refactor search

seanh:
@nickstenning @tilgovi This is a refactoring that I need for nipsa, but I think it'd be good to review and merge this on its own first. Makes the search API code a lot easier to work with imo. I need to change the search dict that we ultimately pass to Elasticsearch (well, to Annotator's search_raw()) from:

    "query": {...}

to:

    "query": {
        "filtered": {
            "filter": {...},
            "query": {...}
    }

This refactoring will allow me to make that change easily. Before it was not so easy.


seanh:
If we ever want to change h's search API params (e.g. to make them more like Elasticsearch's search lite API, or like the h app's stream's params) this refactoring will make that really easy to do as well: you just need to modify one short function that takes a NestedMultiDict and returns a dict and has loads of tests.

h/api/test/views_test.py
```diff
@@ -65,7 +66,7 @@ def user(monkeypatch):
     return user
 
 
-def test_index():
+def test_index(replace_io):
```
nickstenning: I think if you're not using it in the function it would be cleaner to use the decorator:

```python
@pytest.mark.usefixtures('replace_io')
```

nickstenning: Oh and if you're doing this in combination with `@patch`, make sure `@patch` is on the "outside" (i.e. first as you read the file).

nickstenning: Actually, on reflection, I'm not sure that matters.

nickstenning: On further reflection, I'm sure that doesn't matter.

h/api/test/views_test.py
```diff
@@ -258,3 +224,362 @@ def _assert_event_published(action):
         'delete': ['alice'],
     },
 }
+
+
+class TestSearch(object):
```
nickstenning: We should be consistent about this. No classes in tests. I get what you're saying about grouping, but it has downsides and visual noise such as the useless `self` parameters and the pylint pragma.

nickstenning:
I think the refactoring is great, but I'm not sure about testing things that are apparently private API (i.e. start with an underscore).

I'd suggest three things:

- Move the search helper functions into their own module, `h.search`.
- Make both `search` and `query` (perhaps `build_query`?) public API for that module.
- The bulk of the tests should be testing `build_query`, not `search`. If you do that then you don't need to patch `Annotation.search_raw` all over the place: `build_query` can be pure.

h/api/test/views_test.py
seanh: Yes, sorry, these were started a while ago, happy to remove it

seanh:
About patching search_raw(): the reason is that the old (pre-refactoring) h code sometimes calls Annotator's search_raw(), and it sometimes calls Annotator's search() when does some fiddling with the arguments then passes them on to search_raw(). I'm now removing both the refactored h

seanh:
About patching search_raw(): the reason is that the old (pre-refactoring) h code sometimes calls Annotator's search_raw(), and it sometimes calls Annotator's search() when does some fiddling with the arguments then passes them on to search_raw(). I'm now removing both the refactored h code and Annotator's search() from the equation and I need to be sure I haven't changed behaviour. In all cases calling Annotator's search_raw() is ultimately what ends up happening so if I test there I can be sure.

But yes I agree patching it is bad!

Maybe since I have used these tests pre- and post-refactoring to (try to) ensure I didn't change anything, it's now okay to refactor further including changing the tests to not patch search_raw().

seanh:
> Make both search and query (perhaps build_query?) public API for that module.

But `query()` / `build_query()` is a helper function that's only used by the `search()` function in the same module, so why should it be public? This seems to be making a function that isn't publicly used public, just to avoid calling an `_` function in tests.

seanh:
I think moving stuff into a search module seems good and I'm happy to change the tests to mostly call _build_query(), but I'm not sure about making that function public

nickstenning:
1. It's totally okay to change the design of your software in order to make it easier to test.
2. Pure functions are easier to test than non-pure functions.
3. What harm could result from exposing `build_query` as public API for the search module?

seanh:
I'm fine with making a pure function and testing that. The difference is with making that function public, just to avoid calling a private function from tests. What's wrong with calliong

seanh:
I'm fine with making a pure function and testing that. The difference is with making that function public, just to avoid calling a private function from tests. The harm in making `_build_query()` public is that that signals that it's intended to be used by other modules (not just test modules), which it isn't. I want to keep the public API of the module mininal. What's wrong with calling a private function from tests?

I think this conflates two things:

1. Designing the public API of the search module by deciding which functions to make public and private
2. Breaking up the code into different functions to make it easy to test

If you say "don't call private functions from tests" then you either have to make a function public that doesn't need to be, or you have to test a private function indirectly via a public one. So I would just call the private function from the tests.

nickstenning:
I'm not so sure. If you're testing a function called `_build_query`, then in my view you're doing at least one of these undesirable things:

1. testing implementation rather than interface
2. exposing a public symbol with a private name

The question is ultimately one of layering. `build_query` is public API to *something* otherwise you wouldn't be testing it. Perhaps the "correct" thing to do would be to put it in its own "querybuilder" module? In practice I think this would be overkill, and am struggling to see what harm could be caused by making it public -- you're already testing it, so it's not like you can refactor it away.

seanh:
@nickstenning How's that? I could add more tests for `h.api.search.search()` but it doesn't really seem necessary, it's so simple.

There's an `annotations_index()` view function that doesn't seem to be called from anywhere in our code and doesn't have any tests. But it's part of our API and I guess the client-side might be calling it. I moved it into `search` as well.

seanh:
I ran into a problem with `from` imports in `views.py`. There's already a view-callable function defined in there called `search()` so you can't do `from h.api import search` 'cos you'll get a name collision. So just using the full name.

h/api/search.py
```diff
@@ -0,0 +1,115 @@
...
+
+    query = build_query(request_params)
+    results = Annotation.search_raw(query, user=user)
+    count = Annotation.search_raw(query, {'search_type': 'count'},
+                                  raw_result=True)
+    return {"rows": results, "total": count["hits"]["total"]}
+
+
+def index(user=None):
+    """Return the 20 most recent annotations, most-recent first.
```
nickstenning: This seems like a docstring that's bound to be wrong at some point. It should probably say "return annotations from a default search".

h/api/test/search_test.py
```diff
@@ -0,0 +1,324 @@
+"""Tests for h/api/search.py."""
+# pylint: disable=protected-access
```
nickstenning: You can presumably get rid of this now?

h/api/test/search_test.py
```diff
@@ -0,0 +1,324 @@
+"""Tests for h/api/search.py."""
```
nickstenning: Is this docstring really adding anything? We have a strong convention on test module naming. This just seems like noise to me.

h/api/test/search_test.py
```diff
@@ -0,0 +1,324 @@
...
+# pylint: disable=protected-access
+import mock
+import webob.multidict
+
+import h.api.search as search
```
nickstenning: `from h.api import search`

nickstenning:
Looks great other than a few minor niggles in line comments, plus:

- [ ] the tests are a bit inconsistent about whether they pass a dict or a MultiDict. I'd suggest pass a dict unless you're testing MultiDict behaviour.
- [ ] slight typo ("Mae") in commit message 346fb4b

seanh:
@nickstenning Ready again.

nickstenning:
LGTM. Will :ship: as soon as Travis agrees.

PR #2284, opened 2015-06-04
Dehorusify the ForgotPasswordController

nickstenning:
Remove the ForgotPasswordController's dependency on horus. Part of the ongoing process to remove our dependency on this library.

This will almost certainly need a rebase after merging #2281 (I've added the same helper functions in both branches). Suggest merging that first and then pinging me here.

### DIFF
--- a/h/accounts/test/views_test.py
+++ b/h/accounts/test/views_test.py
@@ -301,24 +301,6 @@ forgot_password_fixtures = pytest.mark.usefixtures('activation_model',
                                                    'routes_mapper',
                                                    'user_model')
 
-@forgot_password_fixtures
-def test_forgot_password_forbids_GET():
-    request = DummyRequest()
-
-    result = ForgotPasswordController(request).forgot_password()
-
-    assert isinstance(result, httpexceptions.HTTPMethodNotAllowed)
-
-
-@forgot_password_fixtures
-def test_forgot_password_redirects_when_logged_in(authn_policy):
-    request = DummyRequest(method='POST')
-    authn_policy.authenticated_userid.return_value = "acct:jane@doe.org"
-
-    result = ForgotPasswordController(request).forgot_password()
-
-    assert isinstance(result, httpexceptions.HTTPFound)
-
 
 @forgot_password_fixtures
 def test_forgot_password_returns_error_when_validation_fails(authn_policy,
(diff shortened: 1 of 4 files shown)


nickstenning:
Rebased and ready for review.

seanh:
May as well fix "No blank lines allowed after function docstring"

seanh:
Ok, stopping myself from rushing a review of this at 10pm while sick, I should get to it tomorrow

h/accounts/test/views_test.py
```diff
@@ -292,6 +293,264 @@ def test_logout_response_has_forget_headers(authn_policy):
...
+def test_forgot_password_fetches_user_by_form_email(authn_policy,
+                                                    form_validator,
+                                                    user_model):
+    request = DummyRequest(method='POST')
+    authn_policy.authenticated_userid.return_value = None
+    form_validator.return_value = (None, {"email": "giraffe@thezoo.org"})
+
+    ForgotPasswordController(request).forgot_password()
+
+    user_model.get_by_email.assert_called_with(request, "giraffe@thezoo.org")
```
seanh: I'm not sure about this. Just testing that `get_by_email()` was called doesn't mean that the code is right, maybe it throws away the result and uses something else. I would rather test what it does with the user it gets - test that it calls `mailer.send()` with the right message

h/accounts/test/views_test.py
```diff
@@ -292,6 +293,264 @@ def test_logout_response_has_forget_headers(authn_policy):
...
+                                                              authn_policy,
+                                                              form_validator):
+    request = DummyRequest(method='POST')
+    authn_policy.authenticated_userid.return_value = None
+    form_validator.return_value = (None, {"email": "giraffe@thezoo.org"})
+    activation_model.return_value.code = "abcde12345"
+
+    ForgotPasswordController(request).forgot_password()
+
+    reset_link.assert_called_with(request, "abcde12345")
```
seanh: Again, I'd say this test seems unnecessary given the test below it. The only reason that forgot_password() calls reset_password_link() is to pass the link on to reset_password_email(), and the test below tests that. It seems there's no way this test could be failing if the test below were not also failing.

h/accounts/test/views_test.py
nickstenning: That's exactly what happens in later tests. This is a collaboration test, and is checking that the user is fetched using  the email address provided in the form.

nickstenning: And I'd add that no single test can ensure that the code is right. That's why we don't write one test per function :). But if you made changes to `forgot_password` in such a way that `User.get_by_email` is never called, or is called with something other than the email address provided in the form, then I think that's a red flag and it's right that a test should start failing.

h/accounts/test/views_test.py
nickstenning: This is testing that `reset_password_link` is passed the activation code. That's not tested in the following test.

nickstenning:
> May as well fix "No blank lines allowed after function docstring"

Should be fixed by latest push.

PR #2290, opened 2015-06-05
Report form level errors

nickstenning:
When we call `form.validate()` and the validation fails, the raised exception is a `deform.ValidationFailure`. This object has an `error` attribute which represents the underlying validation error object for the entire form.

Previously, if the validation failure were a result of top-level validation errors (such as the fact that an unactivated user is trying to log in), this error would be lost, because we only reported errors for the forms fields, or "children".

This commit changes `h.accounts.views.validate_form` so that it converts the entire `colander.Invalid` object into a dictionary using its `asdict()` instance method. By doing this, we get two immediate benefits:

- top-level validation errors are reported in the '' (empty string) field
- we avoid the need to aggregate form field errors by hand in ajax_form

In addition, we need to deal with this case on the frontend, so this commit also changes the formRespond directive so that if no overall "reason" is provided for failure, then an empty-string member of the "errors" object can set the overall validation status (and error message) for the form.

(This PR also removes support for error flash messages from `ajax_form`. We don't use them anywhere in this package and hopefully won't ever have to.)

### DIFF
--- a/h/accounts/test/views_test.py
+++ b/h/accounts/test/views_test.py
@@ -123,30 +123,6 @@ def test_ajax_form_includes_flash_data(pop_flash):
     assert result['flash'] == {'success': ['Well done!']}
 
 
-def test_ajax_form_sets_status_code_400_on_flash_error(pop_flash):
-    request = DummyRequest()
-    pop_flash.return_value = {'error': ['I asplode!']}
-    _ = ajax_form(request, {'some': 'data'})
-
-    assert request.response.status_code == 400
-
-
-def test_ajax_form_sets_status_failure_on_flash_error(pop_flash):
-    request = DummyRequest()
-    pop_flash.return_value = {'error': ['I asplode!']}
-    result = ajax_form(request, {'some': 'data'})
-
-    assert result['status'] == 'failure'
-
-
-def test_ajax_form_sets_reason_on_flash_error(pop_flash):
-    request = DummyRequest()
-    pop_flash.return_value = {'error': ['I asplode!']}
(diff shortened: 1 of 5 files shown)


h/accounts/test/views_test.py
```diff
@@ -68,34 +73,36 @@ def test_ajax_form_handles_http_error_as_error():
 
 def test_ajax_form_sets_failure_status_on_errors():
     request = DummyRequest()
-    result = ajax_form(request, {'errors': 'data'})
+    result = ajax_form(request, {'errors': {}})
```
nickstenning: These changes are just to make it clear that errors is supposed to be a dict (and only a dict) now.

seanh:
Get some errors weirdness:

screenshot from 2015-06-05 17 08 06

nickstenning:
Oh that's exciting. Will take a look.

nickstenning:
Fixed!

seanh:
Login, go to edit profile form, fill in change email for with two valid and matching emails, and the wrong password, submit: spinner is spinning forever, and in the console:

screenshot from 2015-06-05 18 52 31

nickstenning:
You're quite right. I'm struggling to understand how on earth my changes could have resulted in that suddenly not working, but I'll take a look on Monday.

nickstenning:
It seems like `errors` inside the `formRespond` directive is an array, which isn't right at all...

nickstenning:
Aha, found it.

nickstenning:
Aaaaand fixed again, I hope.

PR #2291, opened 2015-06-05
Dehorusify the RegisterController

nickstenning:
Wahey! The last of the bunch. This commit removes the last major dependency on horus from our accounts controllers.

Notable changes:

- User activation is no longer optional (i.e. configurable)
- After initial registration (i.e. while waiting for an email) users are taken to the index. Previously they were taken to the stream, which doesn't make a lot of sense as they can't log in until they've activated their account.
- Users are now automatically logged in after activation, and are taken to the index page as before.
- Hitting the register endpoint as a logged-in user will redirect you to the stream page.

### DIFF
--- a/conf/development.ini
+++ b/conf/development.ini
@@ -29,10 +29,6 @@ h.feature.notification: False
 h.feature.queue: False
 h.feature.streamer: False
 
-horus.register_redirect: stream
-
-horus.require_activation: False
-
 mail.default_sender: "Annotation Daemon" <no-reply@localhost>
 
 pyramid.debug_all: True
--- a/h/accounts/models.py
+++ b/h/accounts/models.py
@@ -16,7 +16,6 @@ from horus.models import (
 )
 from horus.strings import UIStringsBase
 from pyramid_basemodel import Base, Session
-from pyramid.threadlocal import get_current_request
 import sqlalchemy as sa
 from sqlalchemy import or_
 from sqlalchemy.ext.declarative import declared_attr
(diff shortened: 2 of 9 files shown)


h/accounts/models.py
```diff
@@ -39,18 +38,7 @@
 
 
 class Activation(ActivationMixin, Base):
```
seanh: This is still inheriting from a horus class?

conf/development.ini
```diff
@@ -29,10 +29,6 @@ h.feature.notification: False
 h.feature.queue: False
 h.feature.streamer: False
 
-horus.register_redirect: stream
-
-horus.require_activation: False
```
seanh: Removing differences between dev and production, removing unnecessary configurations - I like this

seanh:
Just rebase this for you

seanh:
When you register you get redirected to the sign-in form and the _Your account is not active. Please check your e-mail._ error message is already showing even though you hadn't tried to login and actually since no username is entered in the form it's not clear which account it's talking about. Is this intentional? It is sort of a useful message to see at that moment so willing to let it stay.

h/accounts/models.py
nickstenning: Yep. Horus isn't gone yet.

conf/development.ini
nickstenning: :+1:

nickstenning:
> When you register you get redirected to the sign-in form and the Your account is not active. Please check your e-mail.

I don't see this, for what it's worth. Registering either through the sidebar or the `/register` endpoint sends me to the login page with a flash message of:

> Thank you for registering! Please check your e-mail now. You can continue by clicking the activation link we have sent you.

If you can consistently reproduce the behaviour you describe, I'd like to dig into that.

seanh:
Hmm.. Just did a quick check on master and didn't get the error message again. Never mind.

PR #2295, opened 2015-06-06
Deduplicate password validation

seanh:


seanh:
I'm still working on this. I want to remove the client-side password validation which further duplicates our simple "at least two characters" password validation logic - just blindly submit the form and let the server validate it.

Currently when changing your password the confirmation password is not sent to the server - validation that you typed the password twice is done client-side only. I might want to move that to the server (not sure).

When registering a new account it doesn't ask you to type your password twice - it probably should.

nickstenning:
This doesn't merge so I'm going to close this for now. If you want to reopen later feel free.

Issue #2257, opened 2015-05-26
QueryParsingException: No text specified for text query


seanh:
After setting `h.feature.queue: True` and `h.feature.streamer: True` to get around [another issue](https://github.com/hypothesis/h/issues/2256) I'm getting the traceback below on master.

In the `send_annotations()` method in `streamer.py` on line 470 `self.query.query["query"]["bool"]["should"][0]["match"]["uri"]["query"]` is an empty list, and elasticsearch doesn't like this. Appending an arbitrary string to the list gets rid of the traceback.

In tracing where this empty list comes from: it seems to come into streamer.py via greenlet.py and websocket.py as a message. Not sure who sends it.

```
Traceback (most recent call last):
  File "/home/seanh/Projects/hypothesis/h/h/streamer.py", line 526, in received_message
    self.send_annotations()
  File "/home/seanh/Projects/hypothesis/h/h/streamer.py", line 470, in send_annotations
    annotations = Annotation.search_raw(query=self.query.query, user=user)
  File "/home/seanh/.virtualenvs/h/local/lib/python2.7/site-packages/annotator/annotation.py", line 91, in search_raw
    raw_result=raw_result)
  File "/home/seanh/.virtualenvs/h/local/lib/python2.7/site-packages/annotator/elasticsearch.py", line 182, in search_raw
    **params)
  File "/home/seanh/.virtualenvs/h/local/lib/python2.7/site-packages/elasticsearch/client/utils.py", line 69, in _wrapped
    return func(*args, params=params, **kwargs)
  File "/home/seanh/.virtualenvs/h/local/lib/python2.7/site-packages/elasticsearch/client/__init__.py", line 504, in search
    params=params, body=body)
  File "/home/seanh/.virtualenvs/h/local/lib/python2.7/site-packages/elasticsearch/transport.py", line 307, in perform_request
    status, headers, data = connection.perform_request(method, url, params, body, ignore=ignore, timeout=timeout)
  File "/home/seanh/.virtualenvs/h/local/lib/python2.7/site-packages/elasticsearch/connection/http_urllib3.py", line 89, in perform_request
    self._raise_error(response.status, raw_data)
  File "/home/seanh/.virtualenvs/h/local/lib/python2.7/site-packages/elasticsearch/connection/base.py", line 105, in _raise_error
    raise HTTP_EXCEPTIONS.get(status_code, TransportError)(status_code, error_message, additional_info)
[...]

PR #2300, opened 2015-06-08
Nipsa

seanh:


seanh:
I'm less chuffed with how `views_test.py` turned out. `models_test.py` was really clean and simple, possibly because `models.py` hardly does anything anymore, but it had no patching and was always testing the inputs, outputs and side-effects not testing implementation.

`views_test.py`: Loads of detailed patching and mocking of lots of different things.

`Session` and `NipsaUser` need to be patched in every test (even tests that don't use the mocks) or the tests will be unnecessarily accessing the db.

The tests are pretty repetitive - largely the same tests repeated four times for the `index()`, `create()`, `delete()` and `read()` functions.

Leaves me wondering if it could be factored differently to make the tests better. But maybe they're ok.

I think some clarity can be added with a few more test function docstrings and comments.

seanh:
Reminder to self: these view callables need tests for missing and invalid params

tilgovi:
@seanh I'm not picky about imports in tests at all, if you were asking that.

seanh:
@nickstenning @tilgovi I think we got into a confusion about what import style we had actually agreed on. Or at least it certainly confused the hell out of me. I thought that in our email thread about it that we had agreed on `import package`, `import module` and `from package.subpackage import module` but then Nick dinged me for `import pyramid_basemodel` and said we had agreed to import symbols from within modules directly (`from pyramid_basemodel import Session` etc).

I think we need to know what we agreed on here so we don't keep butting heads about it / I need to know what we're doing so I don't get told off.

Can I suggest we do what we probably should have done in the first place and just follow [what the Google Python Style Guide says about it](https://google-styleguide.googlecode.com/svn/trunk/pyguide.html#Imports)? I just realised that it's in there, and by coincidence (or maybe not so coincidental) it says to do exactly what Randall was arguing for in the email thread:

* `import json` (for package/module names with no dots in them)
*  `from package.subpackage import module` the rest of the time
* Import packages and modules only, not symbols
* No relative imports
* One import per line

I think this is also a good compromise between what I originally wanted to do (`import` everywhere and no `from`, max context wherever an imported name is used) vs always importing symbols (max conciseness) - it gives some context and is pretty concise.

nickstenning:
> Can I suggest we do what we probably should have done in the first place and just follow what the Google Python Style Guide says about it?

That's fine with me. Could you start a styleguide repository in the `hypothesis` org and put what you want  to codify in there? I'd much rather it was explicitly documented somewhere?

seanh:
I actually found it here https://h.readthedocs.org/en/latest/hacking/code-style.html though if we want to have an org-wide style guide, maybe we should move that to a style guide repo?

tilgovi:
Happy to use hypothesis/vision for this sort of thing.

seanh:
> Reminder to self: these view callables need tests for missing and invalid params

Ooh, actually I don't think it's possible to pass an invalid param :)

seanh:
TODO:

- [ ] Rebase on latest master (significant changes)
- [ ] Tests for a few parts of the already-written code
- [ ] Add an annotations API db, instead of using the app's one. The work done in https://github.com/hypothesis/h/pull/2320 will be useful
- [ ] Add the concept of admin users to h including a way to create the first admin user, an api for managing admin users and a way for the accounts api to send the info that a user is an admin to the annotations api. This should probably be done in a separate pull request and merged first. See this email thread about it https://groups.google.com/a/list.hypothes.is/forum/#!topic/dev/sUZbfkXSw2o
- [ ] Make it so that only admin users can use the nipsa api
- [ ] Probably some import styles to fix
- [ ] Fix landscape's issues

seanh:
Another email thread with ideas for this feature: https://groups.google.com/a/list.hypothes.is/forum/#!topic/dev/PlCkND_1YdI

seanh:
Just rebased this onto master

seanh:
@tilgovi @nickstenning This is finally ready for review now. Brief summary of how this works:

- Admins can access an HTML page at `/nipsa` for nipsa'ing and unnipsa'ing users

- For ease of use this page accepts usernames like `"seanh"` instead of ids like `"acct:seanh@hypothes.is"`. What is actually stored in the nipsa table in the db though are the full ids. The id is constructed from the submitted username using request.domain. Note that the user whose id we're constructing is the user to be nipsa'd, _not_ the authorized admin user who is submitting the form. So I'm not sure whether using request.domain is correct here. The alternative would be to require users to type the full `"acct:seanh@hypothes.is"`.

- The NIPSA API provides logic functions that the HTML page imports and calls. There are no HTTP API endpoints for NIPSA / it isn't exported as part of our public HTTP API. Arguably we should export a JSON API for it - because that would make NIPSA a complete usable feature in the API, without the app. Currently you need the app's HTML page to use it. But we don't need an HTTP API now so I've left it out.

- When you nipsa or unnipsa a user a worker function is kicked off to add the nipsa flag to or remove the flag from all of the user's past annotations. This uses Es's scan and bulk update which is the most efficient way to do this (recommended by Es docs).

- When a nipsa'd user creates a new annotation the nipsa flag is added to the annotation at create time.

- The search API now does filtered Es queries. All nipsa'd annotations are filtered out, but if there's an authorized user then all of their annotations are filtered in so nipsa'd users can see their own annotations but no one else can see them.

- Filtering is how the Es docs recommend to do this sort of thing - it is very efficient as documents can be filtered out before starting the query (before ranking and scoring and everything). Also it means that Es features such as limits and offsets continue to work against the remaining annotations.

- So "public site areas" is defined as the search API, meaning nipsa'd annotations won't appear in the stream or sidebar for example.

- The app/API split is respected by storing the nipsa flags in their own database table belonging to the API, instead of in the app's user table. This means there's no need to communicate the nipsa'd state from app to API: the API simply checks its own model.

- I've got h/api/nipsa/models.py simply importing h.db in order to use db.Base, but this means that the API is importing code from the app. Note sure how you want to do this.

This implementation is pretty much [what I outlined before I started coding](https://groups.google.com/a/list.hypothes.is/forum/#!searchin/dev/nipsa/dev/xZyevd0wFPo/FJHXoUvbeJsJ) (a month and a half ago!). Despite the landscape having changed in the meantime, I still think this implementation has a lot to recommend it over the many alternatives that have been suggested:

1. It's probably the simplest and most straightforward way to implement nipsa

2. I think flagging and filtering out not-safe-for-work content is a sensible feature to have in our API, that many sites/web apps/desktop or mobile apps that might use our API might want

3. The code is clean. In particular, it's almost all in `h/api/nipsa/` except for a couple of one-liners elsewhere in the API, and an HTML page in the app. The existing code is hardly touched at all by NIPSA.

4. It doesn't introduce any new abstract concepts such as a "public" sphere etc which may turn out to be premature. I think since we haven't shipped nipsa and haven't even started to implement groups, we don't know how those features will turn out and evolve in response to users, so we don't know if any given abstraction over both of them is a good one yet. No abstraction is better than a bad abstraction.

5. The way it respects the app/API split is really simple and obvious.

6. Doesn't do anything like accessing Es directly from the app (a -1, even if not strictly forbidden) or adding "generic" functions to the API just to enable one feature in the app

seanh:
@landscape-bot doesn't seem to have been commenting on any of my PRs lately

h/api/nipsa/logic.py
```diff
@@ -0,0 +1,55 @@
...
+    if nipsa_user:
+        request.db.delete(nipsa_user)
+
+    _publish(request, json.dumps({"action": "unnipsa", "user_id": user_id}))
+
+
+def is_nipsad(user_id):
+    """Return True if the given user is on the NIPSA list, False if not."""
+    user_nipsa = models.NipsaUser.get_by_id(user_id)
+    if user_nipsa:
```
nickstenning: `return user_nipsa is not None` would be more idiomatic here.

h/api/nipsa/models.py
```diff
@@ -0,0 +1,30 @@
+import sqlalchemy
+from sqlalchemy.orm import exc
+
+from h import db
+
+
+class NipsaUser(db.Base):
```
nickstenning: So as I mentioned in the document that we discussed yesterday, the API should probably have its own declarative base class.

I'm happy to overlook this for now, but if you do fancy having a stab at this it should be reasonably straightforward. You'd probably just need to add something like the following to `h/api/db.py`:

```python
from sqlalchemy.ext.declarative import declarative_base

Base = declarative_base()

def use_session(session, base=Base):
    base.query = session.query_property()

def bind_engine(engine, base=Base, should_create=False, should_drop=False):
    base.metadata.bind = engine
    if should_drop:
        base.metadata.drop_all(engine)
    if should_create:
        base.metadata.create_all(engine)
```

And then in `h/db.py#includeme` at an appropriate point add

```python
api.db.use_session(Session)
```

And then at the end of `h/db.py#bind_engine` add

```python
api.db.bind_engine(engine, should_create=should_create, should_drop=should_drop)
```

And hopefully that will give you a distinct Base class you can use, configured with a query property connected to the shared session.

h/api/nipsa/models.py
```diff
@@ -0,0 +1,30 @@
...
+                                index=True)
+
+    def __init__(self, user_id):
+        self.user_id = user_id
+
+    @classmethod
+    def get_by_id(cls, user_id):
+        """Return the NipsaUser object for the given user_id, or None."""
+        try:
+            return db.Session.query(cls).filter(
```
nickstenning: There's a query property on the base class so `db.Session.query(cls)` can be written more simply as `cls.query`.

h/api/search.py
```diff
@@ -31,7 +35,20 @@ def _match_clause_for_uri(uristr):
     }
 
 
-def build_query(request_params):
+def _es_client():
+    """Return an elasticsearch.Elasticsearch client object."""
+    return elasticsearch.Elasticsearch([{"host": "localhost", "port": 9200}])
```
nickstenning: *Ahem*

h/migrations/versions/534754b1bf54_add_nipsa_table.py
```diff
@@ -0,0 +1,25 @@
...
+down_revision = '49764e8a7e69'
+
+from alembic import op
+import sqlalchemy as sa
+
+
+def upgrade():
+    op.create_table(
+        'nipsa',
+        sa.Column('user_id', sa.UnicodeText, primary_key=True, index=True)
```
nickstenning: We're consistent in the rest of the code base about this being called `userid`, without an underscore. It should be the same here (and in the corresponding model).

seanh:
Note to self: We may need to re-enable nipsaing of users who don't exist (currently prevented by the html view) so we can hide annotations of user accounts that have been deleted.

seanh:
Ah, I'm getting an Es traceback when loading the sidebar when not logged in...

seanh:
Still need to fix this bug when you're not logged in...

seanh:
> Ah, I'm getting an Es traceback when loading the sidebar when not logged in...

Think this was a red herring, it's just an Es error that's already on master https://github.com/hypothesis/h/issues/2257

seanh:
@landscape-bot

seanh:
@nickstenning Ok, that's everything done.

I ended up adding `request.es_client` which we could make use of in other places as well.

I also removed checking whether the user exists before nipsaing them, b/c I think the code is better without it (see commit message).

seanh:
Oh wait some tests are broken now..

seanh:
That should fix it..

seanh:
@nickstenning OK now it's ready

h/api/nipsa/logic.py
```diff
@@ -0,0 +1,51 @@
...
+def nipsa(request, userid):
+    """NIPSA a user.
+
+    Add the given user's ID to the list of NIPSA'd user IDs.
+    If the user is already NIPSA'd then nothing will happen (but a "nipsa"
+    message for the user will still be published to the queue).
+
+    """
+    nipsa_user = models.NipsaUser.get_by_id(userid)
+    if not nipsa_user:
```
tilgovi: I would prefer explicit `is None` here.

h/api/nipsa/logic.py
```diff
@@ -0,0 +1,51 @@
...
+        nipsa_user = models.NipsaUser(userid)
+        request.db.add(nipsa_user)
+
+    _publish(request, json.dumps({"action": "nipsa", "userid": userid}))
+
+
+def unnipsa(request, userid):
+    """Un-NIPSA a user.
+
+    If the user isn't NIPSA'd then nothing will happen (but an "unnipsa"
```
tilgovi: Why publish the event anyway?

h/api/nipsa/logic.py
```diff
@@ -0,0 +1,51 @@
...
+
+    """
+    nipsa_user = models.NipsaUser.get_by_id(userid)
+    if nipsa_user:
+        request.db.delete(nipsa_user)
+
+    _publish(request, json.dumps({"action": "unnipsa", "userid": userid}))
+
+
+def is_nipsad(userid):
```
tilgovi: Maybe `has_nipsa` since NIPSA is more of a flag than a verb.
I would `add_nipsa` and `remove_nipsa` instead of `nipsa` and `unnipsa`, but that's personal preference, maybe.

h/api/nipsa/search.py
```diff
@@ -0,0 +1,83 @@
+import copy
+
+
+def nipsa_filter(query, userid=None):
```
tilgovi: The name of this function is `nipsa_filter` and it doesn't really change the `query`, it just nests it inside the filtered query. What if instead this function just returned the filter and its caller could combine the query and the filter to make a filtered query?

h/api/nipsa/search.py
```diff
@@ -0,0 +1,83 @@
...
+        "filtered": {
+            "filter": {"bool": {"should": should_clauses}},
+            "query": query["query"]
+        }
+    }
+
+    return query
+
+
+def query_for_users_annotations(userid):
```
tilgovi: Similarly, this could just return `{"term", {"userid": userid}}` and then it could be used by `nipsa_filter` to construct and return a boolean should filter.

h/api/nipsa/logic.py
seanh: https://google-styleguide.googlecode.com/svn/trunk/pyguide.html?showone=True/False_evaluations#True/False_evaluations

The reason for using `if nipsa_user is None:` seems to be in case `nipsa_user` is some falsey value other than `None`, for example in case of a method that may return either `None` or an empty list,  but in this case I know that `NipsaUser` objects aren't falsey and don't expect `get_by_id()` to return any falsey values except `None` so I think _Use the "implicit" false if at all possible_ applies.

h/api/nipsa/worker.py
```diff
@@ -0,0 +1,69 @@
...
+    return {
+        "_op_type": "update",
+        "_index": annotator.es.index,
+        "_type": "annotation",
+        "_id": annotation["_id"],
+        "script": "ctx._source.remove(\"not_in_public_site_areas\")"
+    }
+
+
+def add_or_remove_nipsa(userid, action, es_client):
```
tilgovi: The fact that we use `add` and `remove` here makes me think we definitely should change `nipsa` and `unnipsa` in the logic module.

h/api/nipsa/logic.py
seanh: 1. It's simpler to just always publish, doesn't do any harm and there's no need to increase the complexity of the logic and add more tests

2. This way enables re-unnipsa'ing users. For example if I unnipsa a user but the worker process crashes and doesn't get all of their annotations, I just unnipsa them again to try again. This is now a bit of a mute point since the actual HTML UI for nipsa will make you re-nipsa them in order to unnipsa them again anyway! But originally I had a JSON API (and we may want one again one day) and you could re-nipsa and re-unnipsa directly.

h/api/nipsa/worker.py
```diff
@@ -0,0 +1,69 @@
...
+    """Add/remove the NIPSA flag to/from all of the user's annotations."""
+    assert action in ("nipsa", "unnipsa")
+
+    if action == "nipsa":
+        query = nipsa_search.not_nipsad_annotations(userid)
+    else:
```
tilgovi: Maybe we should be defensive and say explicitly `if action == "unnipsa"`?

h/api/nipsa/worker.py
```diff
@@ -0,0 +1,69 @@
...
+        query = nipsa_search.nipsad_annotations(userid)
+
+    annotations = search.scan(es_client=es_client, query=query, fields=[])
+
+    if action == "nipsa":
+        actions = [add_nipsa_action(a) for a in annotations]
```
tilgovi: Could set a local variable above to the value `add_nipsa_action` or `remove_nipsa_action` and save the duplicated conditional. Might read more cleraly.

h/api/search.py
```diff
@@ -31,7 +34,41 @@ def _match_clause_for_uri(uristr):
     }
 
 
-def build_query(request_params):
+def scan(es_client, query, fields):
```
tilgovi: This doesn't seem to add anything at all to `helpers.scan`. You could just export `helpers.scan` here if you wanted to. Or maybe you're deliberately trying to hide the extra functionality of `helpers.scan`? Any reason?

h/api/search.py
```diff
@@ -31,7 +34,41 @@ def _match_clause_for_uri(uristr):
...
+    :type query: dict
+
+    :param fields: The list of fields to return for each hit
+    :type fields: list
+
+    """
+    return helpers.scan(es_client, query=query, fields=fields)
+
+
+def bulk(es_client, actions):
```
tilgovi: Same with this. Is there a need for us to just re-export stripped down versions of elasticsearch functions?

h/api/nipsa/search.py
seanh: That sounds like the kind of thing we may need in the future when we have a function for nipsa filtering, a function for group filtering, and a function for combining them.

But for now all logic about Es filters (including how to nest them in a query with a "filtered" dict containing "filter" and "query" keys) is encapsulated in this one function, which is nice.  I believe there's no code anywhere else that knows anything about Es filters.

h/api/search.py
```diff
@@ -41,6 +78,11 @@ def build_query(request_params):
         h search API
     :type request_params: webob.multidict.NestedMultiDict
 
+    :param userid: the ID of the authorized user (optional, default: None),
+        if a userid is given then this user's annotations will never be
```
tilgovi: This bit seems like too much implementation detail. The argument is the authorized userid, whether/how that factors into a feature like NIPSA seems not related to this interface to me.

h/api/nipsa/search.py
seanh: But again I don't want `worker.py` to do any more work than it already does, or to know anything about the structure of Es dicts. Encapsulation. But I could add a `_user_filter()` helper function to return `{"term": {"user": userid}}` that both `query_for_users_annotations()` and `nipsa_filter()` could call, save a little bit of duplication at the cost of readability.

h/api/nipsa/logic.py
tilgovi: I think of the models as being `dict`-like and while I guess it's not going to be an empty dict because it has keys it still makes it clearer for me to think about explicit non-existence if I see `is None`.

tilgovi: But I won't start a fight over it.

h/api/nipsa/logic.py
tilgovi: Hmmm. If the worker crashes, the message won't be acknowledged, and it will still be in the queue, but I guess I see your point about simplicity of the logic.

h/api/nipsa/search.py
tilgovi: It would make the rest of the module simpler to read, in my opinion, and it would save you a couple tests like asserting that this doesn't modify the query (that it never needed in the first place).

h/api/search.py
seanh: I'm trying to encapsulate elasticsearch in this module. So this module calls elasticsearch, other modules call this module. I know we said in a call recently that encapsulating Es in one place was desirable but not if there was a good reason for some other code to access it directly. When I wrote this code though our rule was simpler: encapsulate it here.

So I guess the options would be:

* worker.py imports elasticsearch directly and calls scan and bulk itself, but I consider accessing it directly a -1 and not sure there's a compelling enough reason to do it

* `from elasticsearch import bulk` in `search.py` instead of a wrapper function, but then you're gonna get unused import warnings from prospector

h/api/views.py
```diff
@@ -222,6 +224,9 @@ def _create_annotation(fields, user):
     annotation['user'] = user.id
     annotation['consumer'] = user.consumer.key
 
+    if nipsa.is_nipsad(user.id):
+        annotation["not_in_public_site_areas"] = True
```
tilgovi: any reason not to just make this read `"nipsa"`? I think Elasticsearch may compress the keys but when we refer to this as `nipsa` absolutely everywhere else I don't see why we should spell it out here.

h/api/search.py
tilgovi: If we didn't have to pass the client in I would understand the encapsulation argument. We would actually be hiding the details of the implementation.

But when our arguments include elasticsearch client instances and elasticsearch-formatted query objects I think the encapsulation is bogus anyway and it's better not to pretend.

h/api/nipsa/logic.py
seanh: Well, I guess the point is that a) re-nipsa'ing people doesn't do any harm and b) maybe one day there'll be a reason why we want to re-nipsa someone so why write unnecessary logic that prevents re-nipsa'ing?

This is the same reason why I didn't add logic to prevent you from nipsaing a user id that doesn't exist in the users table - because I can imagine (possibly unlikely) future scenarios where we may _want_ to nipsa users who aren't in the users table, and adding logic to prevent it is just unnecessary

h/api/views.py
seanh: This was just to give users looking at the documents from our api a clue what it means (though I now recall that we discussed hiding it from the output..)

h/api/search.py
seanh: Hmm, well the user just has to pass `request.es_client`, doesn't have to actually know anything about how Es works. But passing that through everything is boilerplatey. This makes me think that maybe the functions in `search.py` should be combined into an `EsClient` class, then `request.es_client` could become an instance of this class, then the code that wants to use it just does `request.es_client.scan(query, fields)`. (This is how `request.api_client works too). What do you think?

h/api/views.py
tilgovi: Up to you. I find it unwieldy and think we could solve it with documentation.

h/api/search.py
seanh: So

    search.scan(es_client=es_client, query=query, fields=[])

becomes

    request.es_client.scan(query=query, fields=[])

or in fact just

    es_client.scan(query=query, fields=[])

since `add_or_remove_nipsa()` would just be passed the `EsClient` instance not the whole request.

tilgovi: I would really just skip the abstraction entirely and we'll figure out what makes sense later.

h/api/nipsa/search.py
seanh: I really think that spreading our knowledge and logic about what the structure of Es queries is and how to construct them in different modules throughout the app is a pretty big negative..

h/api/search.py
tilgovi: One does have to know how ES works in order to use these. One has to first know about scanning and bulk operations, and about the query object format.

h/api/nipsa/search.py
tilgovi: Nonsense. The only places that call this are `streamer` and `search`. The `streamer` code already knows _far_ _far_ too much about Elastisearch. That would only ever change, maybe, if we had it use some totally abstracted search call, which maybe `api` already exposes well enough. And `search` is already constructed the query object anyway in order to pass that in.

Basically, if the caller has to know about the structure of an ES query in order to pass it to this function I find the claim that having the caller instead receive the filter and have to construct an ES query from it is any more exposing details about ES totally bogus.

seanh: It's _filters_ that are encapsulated here, `api/search.py` knows about Es queries of course but it doesn't know anything about filters yet, including how to insert them into query dicts, all that knowledge is in this one function in `api/nipsa/search.py`.

Filtering was something I was only introducing into the code for nipsa so for the sake of not adding new "generic" functions into other parts of the code that are only used for the nipsa function, and to minimize nipsa's footprint outside of `h/api/nipsa/` (that footprint being just a couple of one-liners was one of my favourite things about this approach to nipsa), I encapsulated it here.

I imagine once we have nipsa and groups, two features that are probably both gonna use filters, then we'll need some generic inserting filters in queries code in `api/search.py`, though. I would wait until we have two users but I can do it now if you really want

tilgovi: Yes. I would really prefer that.

tilgovi: It's about making this function simple. There's no reason I can see why it needs to accept an argument that it has a contractual obligation not to modify, and therefore has to deep copy, when it never inspects that argument or modifies it anyway, just to prevent the caller from having to know about how to construct filtered queries even though it already knows its using elasticsearch and constructing elasticsearch queries.

The result is only going to be a proliferation of code that modifies elasticsearch query objects, like this code, rather than code that constructs the useful pieces of them to be assembled as needed by the caller.

h/api/nipsa/worker.py
seanh: That's what the assert above is for

tilgovi: Ah, okay. Totally didn't catch that.

seanh:
@tilgovi Ok that's all the changes done. I have to run now, but that was a lot of changes (including a lot of big renames and stuff) so I want to manually test that everything's working and look over it again tomorrow before we merge it

tilgovi:
I'm going to rebase it, check it out, and see if I can't send it straight in. Thanks for all the effort and patience.

tilgovi:
I rebased this and I also changed the worker so that removal uses the `index` operation rather than `update` because it as easier to figure out how/where to turn on dynamic scripting support / document that / figure out if that works for hosting ES instances / store a script / blah blah blah. I could have gone with `nipsa: false` instead, as an update, but I'm just not that concerned about performance here.

tilgovi:
Last remaining blocker is that the `should_send_event` function in `h.streamer` has no idea how to handle NIPSA and so annotations from flagged users will show up in the stream if they are created/updated while a user is watching the stream or if they are revealed by infinite scroll. That's pretty poor NIPSA :(.

Unfortunately, that code is kind of a mess and I don't know what to do with it. It's going to be tricky to figure out when to let NIPSA hide real-time updates if we keep doing things as we have been. Obviously on the stream, yes, because that's public, but what about when on the standalone annotation page for a NIPSA'd annotation? I think clearly we should get that update. But right now the difference between the two cases is buried, from the server's persective, in the crazy payload of pseudo-Elasticsearch the client passes it over the WebSocket.

:rage:

seanh:
What?? I was pretty sure annotations were being filtered out of the stream and infinite scroll but I'll double check. I didn't think about annotations that get created or updated while someone is looking at the page though

seanh:
Ok, I think the existing streamer.py solution does work for infinite scrolling:

1. Create user fred, make some public annotations. Verify that both fred and non-logged-in visitors can see fred's annotations in the stream and sidebar.
2. Create user seanh, make him an admin using the command line
3. Use seanh to nipsa fred (make sure you have the nipsa worker process running)
4. Verify that fred can still see his own annotations in the stream and sidebar, but neither seanh nor not-logged-in users can see them
5. Use seanh to create lots of annotations that will appear before fred's in the stream, enough to push fred's annotations off the first page load and into the infinite scroll
6. Verify that fred receives his annotation from infinite scroll, but seanh and not-logged-in users don't

seanh:
I discovered this issue:

1. Log in as Fred (who is nipsa'd) and you can see Fred's annotations in the sidebar
2. Log out, Fred's annotations are still visible
3. Log in as Bob, Fred's annotations are still visible
4. Reload the page and the nipsa'd annotations disappear

This doesn't happen with private annotations. What I did for nipsa was find the part of the code in streamer.py when it queries elasticsearch for the annotations, and insert the nipsa filter in there. So this seems to suggest that on logout/login it doesn't simply throw away all the annotations and request them again with the new authorization, but instead re-filters the already-fetched annotations somewhere.

I'm gonna vote for not bothering to fix this for nipsa.

seanh:
Ok, I figured out that I have to turn on the queue and streamer features in the db now, and then I can reproduce the problem that Randall mentioned: New (and possibly also updated) annotations will appear on the page via the websocket

seanh:
Ok that's the issue with new/updated annotations coming down the websocket fixed

seanh:
Ok, so live updates of annotations (nipsa'd or not) on individual annotation pages does seem to be working if the user looking at the page is the annotation creator himself (the only code I wrote to filter annotations out of the websocket only filters them out if they're nipsa'd _and_ not created by the authorized user).

But there's a bigger problem - individual annotation pages are broken if the annotation is nipsa'd, and the authorized user is not the annotation creator. This is because the page loads a template with no annotation and then Angular makes an XHR to the _search_ API searching for an annotation by ID, and of course nipsa'd annotations are filtered out.

I think the correct solution here is for the frontend to use the API endpoint for retrieving a single annotation, since with the current situation there's no (nice) way for the backend to detect these requests and treat them differently from other search requests.

tilgovi:
Sounds right. We definitely want to make a single annotation request.

tilgovi:
Ugh, but I bet the point of that was to get the replies somehow...

Which leads to the question: if I'm viewing a NIPSA annotation, should I
see NIPSA replies?

Or, if I'm viewing a non-NIPSA annotation, should I see NIPSA replies?
Probably not... but how do we tell the difference between these?

seanh:
Argh, I hadn't even considered replies.

There is a way the backend can detect these individual annotation requests and treat them differently, it can look at the Referer header for example. It's ugly, but...

seanh:
I think for replies, replies to nipsa'd annotations should be filtered out whenever the annotation is filtered out, but that's something I'll need to test for now and maybe fix

seanh:
Ok just wanna go through this systematically. The current implementation:

1. Nipsa'd user replies to their own nipsa'd annotation:

   * The user can see their own annotation and reply
   * Other users will see neither.

2. Nipsa'd user 1 replies to nipsa'd user 2's annotation (this requires user 2 to have been nipsa'd _after_ the reply was made):

   * User 2 sees their annotation but not the reply.
   * User 1 sees neither, it seems kind of bad that they can't see their own reply, but then how would we even display it?
   * Other users see neither

3. Nipsa'd user replies to non-nipsa'd annotation:

   * The nipsa'd user sees both annotation and reply
   * Others (including annotation author) see just the annotation, not the reply

4. Non-nipsa'd user replies to nipsa'd annotation (requires the user to be nipsa'd _after_ the reply) (this is the same as 2):

   * The nipsa'd user sees both annotation and reply
   * Other users (including the author of the reply) see neither the annotation nor the reply. Weird that they can't see their own reply but not sure what we could do about it

seanh:
I'd say the right behaviour for an individual annotation page is to always show the annotation and all its replies, regardless of whether the annotation author or reply authors are nipsa'd and regardless of who the authorized user is. So in other words we just need to completely disable the nipsa filtering on those pages, it's currently enabled..

h/accounts/__init__.py
```diff
@@ -16,9 +16,9 @@ class NoSuchUserError(Error):
     pass
 
 
-def make_admin(user_id):
+def make_admin(userid):
     """Make the given user an admin."""
-    user = models.User.get_by_username(user_id)
+    user = models.User.get_by_username(userid)
```
nickstenning: "acct:foo@bar.com" is a `userid`, "foo" is a username. `get_by_username` accepts a username. This param name should probably change.

tilgovi: Or use `get_by_id`.

seanh:
Ok that's the individual annotation pages fixed to use the annotation read API instead of search

seanh:
Ok, ready again. To summarise the last couple of days:

- [x] Responded to Nick's review comments
- [x] Responded to Randall's review comments
- [x] There's now a `request.es_client` with the Elasticsearch URL from the config file
- [x] The API has its own sqlalchemy base class
- [x] You can nipsa a userid that doesn't exist in the user table
- [x] "Live update" annotations are filtered for nipsa
- [x] Checked that "infinite scroll" annotations were already being filtered
- [x] [Checked behaviour of nipsa with replies](https://github.com/hypothesis/h/pull/2300#issuecomment-123349996)
- [x] Annotation permalink pages now use annotation read URL not search, don't get nipsa'd
- [x] `annotation-viewer-controller-test.coffee` has been rewritten, is working now

Maybe I'll realise in the morning something I missed or did wrong

seanh:
@nickstenning @tilgovi This is ready for review again, whichever one of you wants it. I'd just ask that you do a quick manual test of everything, just because there were quite a lot of last-second changes (admins can nipsa and unnipsa users; the worker process runs and old annotations appear and disappear from the stream; new annotations from nipsad users are hidden; nipsad users can see their own annotations; everyone can see a nipsad annotation if they go to its permalink page; nipsad annotations are hidden from infinite scroll and live update...)

h/accounts/views.py
```diff
@@ -605,15 +605,15 @@ def index():
     def create(self):
         """Make a given user an admin."""
         try:
-            user_id = self.request.params['add']
+            userid = self.request.params['add']
```
nickstenning: Same comment here about `userid` vs `username`.

h/api/nipsa/logic.py
nickstenning: I'm also not going to block merging this over an issue as small as this, but as previously expressed I'd much rather see explicit `is None` checks where that's what's being checked.

h/api/nipsa/search.py
```diff
@@ -0,0 +1,61 @@
...
+        The returned filtered query won't filter out this user's annotations,
+        even if the annotations have the NIPSA flag.
+    :type userid: unicode
+
+    """
+    # If any one of these "should" clauses is true then the annotation will
+    # get through the filter.
+    should_clauses = [{"not": {"term": {"nipsa": True}}}]
+
+    if userid:
```
nickstenning: ... such as here. IMO this should be `if userid is not None:`

h/api/test/views_test.py
```diff
@@ -99,6 +101,34 @@ def test_create(mock_create_annotation, user):
...
+@patch("h.api.views.get_user")
+def test_create_does_not_add_nipsa_flag(get_user, has_nipsa, _):
+    get_user.return_value = Mock(
+        id="test_id", consumer=Mock(key="test_key"))
+    has_nipsa.return_value = False
+    request = DummyRequest(json_body=_new_annotation)
+
+    annotation = views.create(request)
+
+    assert not annotation.get("nipsa")
```
nickstenning: Another example of where this should be an explicit `is None` check, in my view. This test will pass if the nipsa field is set to the empty string, but that probably won't play well with the elasticsearch filters elsewhere in the PR.

h/conftest.py
```diff
@@ -91,6 +92,8 @@ def db_session(request, settings):
     engine = db.make_engine(settings)
     db.bind_engine(engine, should_create=True, should_drop=True)
 
+    api_db.use_session(db.Session)
```
nickstenning: This isn't necessary. You're already calling `db.make_engine` three lines earlier, and that calls this, no?

h/api/nipsa/models.py
```diff
@@ -0,0 +1,30 @@
...
+
+from h.api import db
+
+
+class NipsaUser(db.Base):
+
+    """A NIPSA entry for a user (SQLAlchemy ORM class)."""
+
+    __tablename__ = 'nipsa'
+    userid = sqlalchemy.Column(sqlalchemy.UnicodeText, primary_key=True,
```
nickstenning: You probably want `Unicode`, not `UnicodeText` here.

h/api/nipsa/models.py
```diff
@@ -0,0 +1,30 @@
...
+                               index=True)
```
nickstenning: AIUI there's nothing gained by adding an index on a column that already has a PRIMARY KEY constraint.

h/script.py
```diff
@@ -61,13 +61,13 @@ def init_db(args):
 def admin(args):
     """Make a user an admin."""
     paster.bootstrap(args.config_uri, request=Request.blank(''))
-    accounts.make_admin(unicode(args.user_id, sys.getfilesystemencoding()))
+    accounts.make_admin(unicode(args.userid, sys.getfilesystemencoding()))
```
nickstenning: This isn't a userid.

h/api/nipsa/logic.py
nickstenning: The distinction here (in my head, at least) is between:

1. Checking for falsiness. This should always be `if foo:`, not `if foo == True:` or `if foo is True:`
2. Checking for the *sentinel value* `None`. This should always be `if foo is None:`/`if foo is not None:`

This isn't a check for falsiness, it's a check for a sentinel value.

seanh:
In my defense, all of the requested userid -> username changes are on parts of the code that are already using user_id instead of username on master, they only appear in this diff because I changed user_id to userid everywhere. Admittedly user_id is on master because of an earlier pull request by me, though. Anyway I'll make the changes.

h/conftest.py
seanh: No, `make_engine()` doesn't call `use_session()`, the `includeme()` does. I could refactor `h/db.py` and `h/api/db.py` to make this line unnecessary here though.

h/api/test/views_test.py
seanh: Hmm, I would have thought Es would play fine with that (the Es filter is a positive one: filter out anything that does have `nipsa: true`). But I guess what I really want here (if we want to test specifically what the actual behaviour is) is `"nipsa" not in annotation`

h/api/nipsa/search.py
seanh: I think what this comes down to is what we think it should do if the user code passes in an empty string for the userid: current behaviour is not to add any userid filter, the Es query will work as if there's no authorized user. If I changed it to `is None` it would add a `{"term": {"user": ""}}` filter which I'm guessing will cause a traceback from Es. Or we could explicitly check for the empty string and raise our own exception here. I think the current implementation is as defensible as any of the other options, isn't it?

h/conftest.py
nickstenning: Oh, shit, sorry. My misreading. No, it's fine.

h/api/nipsa/search.py
nickstenning: I don't really understand. If you pass `user=123`, or `user=sys.stdout`, this code is still going to do The Wrong Thing. As far as I'm concerned, this is again a sentinel-vs-falsy distinction: the API of this function says "the `userid` parameter must be a valid userid, or the sentinel None, to indicate that there is no authenticated user". So we we should be precise and check for that sentinel.

nickstenning:
Okay, I've checked this out and tested everything I can think of testing. It all seems to work!

h/api/nipsa/search.py
seanh: So you're saying we assume the value is either `None` or a valid userid (so not the empty string and not falsy). In that case `if userid:` and `if userid is not None:` are the same behaviour-wise and it comes down to style. Our style guide says to use the implicit false _if at all possible_ and here it is possible to get the right behaviour by using it ... So unless we want to introduce a new style rule along the lines of always use `is None` instead of the implicit false, or always use it in case of `None` as a sentinel value, or something.

tilgovi: I think it should be explicit here and I don't think that violates the
style guide, either.

For me, it's just not the place for this function to judge whether an empty
string is a valid username. I know that sounds silly, but we're really
trying to check "was I passed an argument", no? Check that.

nickstenning:
It's been a long road, but I think we're there.

Thank you for all your hard work on this Sean.

:cake:

PR #2303, opened 2015-06-09
Add API CORS support

### LINKED ISSUE #2186: Add Access-Control-Allow-Origin: * to JSON responses for anonymous search requests
I'd like to be able to search the API for annotations of the current page, in client-side Javascript.

Currently there's no `Access-Control-Allow-Origin: *` header on the JSON response, so it's not possible to call the API from a different domain (unlike via.hypothes.is, which is able to call the API as it's on the same domain).

As the search results are sometimes filtered according to the current user, and the `Access-Control-Allow-Origin: *` prevents the response to authenticated requests being read, the header should only be added to anonymous requests (meaning that only non-authenticated requests would be readable cross-domain).

I think it needs to be added somewhere around https://github.com/hypothesis/h/blob/master/h/api/views.py#L77

nickstenning:
Adds basic CORS support to the API. This is implemented as ~~a tween, but perhaps might be better as~~ plain ol' WSGI middleware.

Notable limitations:

- returned access control headers are not customisable on a per-view basis
- sending credentials in CORS requests is currently disabled -- this means that you cannot redeem a session cookie from the h application for an auth token

Fixes #2186.

### DIFF
--- a/h/api/middleware.py
+++ b/h/api/middleware.py
@@ -1,0 +1,69 @@
+from webob import Response
+from webob.dec import wsgify
+from webob.exc import HTTPBadRequest
+
+
+@wsgify.middleware
+def permit_cors(req,
+                app,
+                allow_credentials=False,
+                allow_headers=None,
+                allow_methods=None,
+                expose_headers=None,
+                max_age=86400):
+    """
+    WSGI middleware that enables CORS support across an application.
+
+    CORS stands for "Cross-Origin Resource Sharing," and is a protocol
+    implemented in browsers to allow safe dispatch of XMLHttpRequests across
+    origins.
+    """
+    # If the request is anything other than an OPTIONS request, we just
+    # pass it through and add "A-C-A-O: *" to the response headers.
+    if req.method != 'OPTIONS':
+        resp = req.get_response(app)
+        resp.headers['Access-Control-Allow-Origin'] = '*'
+        return resp
+
+    # Otherwise, we're dealing with a CORS preflight request, which,
+    # according to the spec:
+    #
+    #  http://www.w3.org/TR/cors/#resource-preflight-requests
+    #
+    # ...MUST have an Origin header.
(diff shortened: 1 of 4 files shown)


nickstenning:
On reflection, I think I probably do want this to be just WSGI middleware... so hold your horses.

nickstenning:
Reworked as plain WSGI middleware. Review away. @seanh?

nickstenning:
@tilgovi I'm not sure @seanh is going to get to this today, so if you fancy taking a look, that would be awesome.

tilgovi:
It does what it says on the tin and I think that means I'm okay to merge it for now, but I am not sure I agree with the decision to do it as middleware, although a tween is not right either.

Maybe best would be via a response callback, like this: http://stackoverflow.com/a/22862087/945546

Convenience would be to have `request.add_cors_headers(...)` or some such call, and ultra convenience would be to have a view predicate for it or something. Cornice does this pretty right, I think: http://cornice.readthedocs.org/en/latest/api.html

It's worth considering which parts of our application actually want CORS. I would be surprised it it were more than a day of work to make our annotations API a cornice application, too.

nickstenning:
I did start by implementing this as a view decorator, but it's not trivial to get it right, because implementing it as a decorator implies a one-to-one mapping between preflight requests and view functions. And there is no trivial one-to-one mapping between preflight requests and two view functions like:

```
PUT /annotations/:id
DELETE /annotations/:id
```

That is, because you only get one preflight:

```
OPTIONS /annotations/:id
```

And the correct response includes `Access-Control-Allow-Methods: DELETE, PUT, OPTIONS`, which at that point involves introspecting the views.

Another option is to make the individual views also predicated on the request

tilgovi:
Yeah, that's my conclusion, too. Cornice is able to do it because it's creating all the view registrations for you, after being provided with all the details about the service.

tilgovi:
Let's go with it right now.

PR #2310, opened 2015-06-12
Don't flash a message when profile is successfully edited

nickstenning:
We already have little "Saved" spinner/status indicators on the profile forms, so an additional flash message is superfluous.

### DIFF
--- a/h/accounts/views.py
+++ b/h/accounts/views.py
@@ -447,7 +447,6 @@ class ProfileController(object):
         if password:
             user.password = password
 
-        FlashMessage(self.request, _('Changes saved!'), kind='success')
         return response
 
     def disable_user(self):


judell:
The screenshot was made using the stage extension (4.1.166) and repeatedly clicking *Someone replies to one of my annotations*

image

I'm seeing the reverse of what this PR implies. The production client only spins, does not flash. The new client flashes (repeatedly).

tilgovi:
You're testing the notifications settings not the profile (password / email), @judell.

The profile has a save button that shows the spinner and success feedback but the notifications settings don't have a save button.

Open an issue if you want to change that and let's pick some other means of showing success or decide, consciously, that we don't need to.

judell:
OK, did not realize that we don't consider the notification settings to be part of the profile.

PR #2312, opened 2015-06-12
Remove use of horus models in h.accounts

nickstenning:
This PR:

- removes all trace of the horus model mixins from `h.accounts.models`, adding the needed attributes and classmethods into our existing User and Activation models.
- removes all trace of horus from `h.accounts.schemas`.
- removes the unused Group and UserGroup models.
- *does not* include a migration to:
  + remove the User.security_code field (which is unused and luckily nullable)
  + remove the group and user_group tables

The lack of a migration is because we're going to need to do this in multiple stages in order to not break production:

1. migration: make any columns that are unused nullable
2. deploy and run the migration
3. remove any python model code that references those columns
4. deploy
5. migration: remove any unused columns
6. deploy and run the migration

### DIFF
--- a/h/accounts/models.py
+++ b/h/accounts/models.py
@@ -1,51 +1,60 @@
 # -*- coding: utf-8 -*-
+from datetime import datetime
+from datetime import timedelta
 import re
 
-from hem.interfaces import IDBSession
+import cryptacular.bcrypt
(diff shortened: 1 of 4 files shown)


tilgovi:
\o/ \o/

PR #2313, opened 2015-06-12
Don't depend on horus for "flash" messages

nickstenning:
This commit simplifies handling of session flash messages to just use plain old

    request.session.flash(message, queue)


tilgovi:
Ace

PR #2314, opened 2015-06-12
Visually truncate long bodies and quotes

### LINKED ISSUE #487: Cards aren't visually truncated
We used to have this sorted out earlier.

1. Select a longer piece of text. (Several paragraphs.)
2. Make an annotation about it.
3. When the annotation is displayed, the whole quote is shown. 

With texts above a given length, we should only show a smaller part, with a small link or button to show the whole text.

- [ ] Quotes, when bodies are uncollapsed, should be truncated to approximately two lines (around 160 characters on avg) of text.  When bodies are collapsed, quotes should not be shown.
- [ ] Bodies, when annotations are collapsed, should be truncated to one line.  When bodies are uncollapsed should be shown in full.
- [ ] Truncation should result in ellipses "... " shown at the end of lines via CSS.  Clicking the ellipses should expose the remainder of the truncated text.

JakeHartnell:
- Introduce collapsing bodies and quotes which expand when you click more.
- Meant to resolve #487 
- Uses CSS only to truncate the quotes.

Dokku here: http://more-less.dokku.hypothes.is/

Notes:
While implementing this, I attempted to follow the approach laid out in #487 (mostly succeeding). However, I ran into a problem with excerpts in the annotation body: *sometimes* overflow hidden will cause text to be cut in half. This is because we support stylized text (and thus have varying lineheights, font sizes, etc). 

To address this in this first attempt, I implemented a small gradient at the bottom of long bodies where the "More..." text appears like so:
Annotation with long quote and long body

I'm open to other ideas and suggestions.

We still need to determine how long we want the annotation bodies to be. I think there is general agreement that we want them to be longer than the quotes. I'm curious as to how people feel about card length after playing around with the dokku.

### DIFF
--- a/h/static/scripts/directive/excerpt.coffee
+++ b/h/static/scripts/directive/excerpt.coffee
@@ -1,0 +1,28 @@
+###*
+# @ngdoc directive
+# @name excerpt
+# @restrict C
+# @description Checks to see if text is overflowing its container.
+# If so, it prepends/appends expanders/collapsers.
+###
+module.exports = ->
+  link: (scope, elem, attr, ctrl) ->
+    scope.$evalAsync ->
+      if elem[0].scrollHeight > elem[0].clientHeight
+        elem.prepend angular.element '<span class="more"> More...</span>'
+        if elem.hasClass('annotation-quote')
+          offset = 6
+        else if elem.hasClass('annotation-body')
+          offset = 9
+        else offset = 0
+        elem.find('.more').css({
+            top: elem[0].clientHeight + offset
+        })
+        elem.append angular.element '<span class="less"> Less ^</span>'
+        elem.find('.more').on 'click', ->
+          $(this).hide()
+          elem.addClass('show-full-excerpt')
+        elem.find('.less').on 'click', ->
+          elem.find('.more').show()
+          elem.removeClass('show-full-excerpt')
+  restrict: 'C'
--- a/h/static/scripts/directive/excerpt.coffee
+++ b/h/static/scripts/directive/excerpt.coffee
@@ -5,10 +5,22 @@
 # @description Checks to see if text is overflowing its container.
 # If so, it prepends/appends expanders/collapsers.
 ###
-module.exports = ->
+module.exports = [ '$timeout', ($timeout) ->
   link: (scope, elem, attr, ctrl) ->
-    scope.$watch (-> elem[0].scrollHeight), (scrollHeight) ->
-      scope.$evalAsync ->
+    # Check for excerpts when annotations are created.
+    scope.$on 'annotationCreated', (event) ->
+      scope.excerpt()
+
+    # Check for excerpts when annotations are updated.
+    scope.$on 'annotationUpdated', (event) ->
+      scope.excerpt()
+
+    # Check for excerpts on thread collapse toggle.
+    scope.$on 'threadToggleCollapse', (value) ->
+      scope.excerpt()
+
+    do scope.excerpt = ->
+      $timeout ->
         if elem.find('.more').length == 0
           if elem[0].scrollHeight > elem[0].clientHeight
             elem.prepend angular.element '<span class="more"> More...</span>'
@@ -28,3 +40,4 @@ module.exports = ->
               elem.find('.more').show()
               elem.removeClass('show-full-excerpt')
   restrict: 'C'
+]
--- a/h/static/scripts/directive/excerpt.coffee
+++ b/h/static/scripts/directive/excerpt.coffee
@@ -7,22 +7,24 @@
 ###
 module.exports = ->
   link: (scope, elem, attr, ctrl) ->
-    scope.$evalAsync ->
-      if elem[0].scrollHeight > elem[0].clientHeight
-        elem.prepend angular.element '<span class="more"> More...</span>'
-        if elem.hasClass('annotation-quote')
-          offset = 6
-        else if elem.hasClass('annotation-body')
-          offset = 9
-        else offset = 0
-        elem.find('.more').css({
-            top: elem[0].clientHeight + offset
-        })
-        elem.append angular.element '<span class="less"> Less ^</span>'
-        elem.find('.more').on 'click', ->
-          $(this).hide()
-          elem.addClass('show-full-excerpt')
-        elem.find('.less').on 'click', ->
-          elem.find('.more').show()
-          elem.removeClass('show-full-excerpt')
+    scope.$watch (-> elem[0].scrollHeight), (scrollHeight) ->
+      scope.$evalAsync ->
+        if elem.find('.more').length == 0
+          if elem[0].scrollHeight > elem[0].clientHeight
+            elem.prepend angular.element '<span class="more"> More...</span>'
+            if elem.hasClass('annotation-quote')
+              offset = 6
+            else if elem.hasClass('annotation-body')
+              offset = 9
+            else offset = 0
+            elem.find('.more').css({
+                top: elem[0].clientHeight + offset
+            })
+            elem.append angular.element '<span class="less"> Less ^</span>'
+            elem.find('.more').on 'click', ->
+              $(this).hide()
+              elem.addClass('show-full-excerpt')
+            elem.find('.less').on 'click', ->
+              elem.find('.more').show()
+              elem.removeClass('show-full-excerpt')
   restrict: 'C'
--- a/h/static/styles/annotations.scss
+++ b/h/static/styles/annotations.scss
@@ -66,12 +66,20 @@
 
 .annotation-quote {
   @include quote;
+  max-height: 2.9em;
+  overflow: hidden;
+
   del {
     background:#ffe6e6;
   }
   ins {
     background:#e6ffe6;
   }
+}
+
+.annotation-body {
+  max-height: 13.5em;
+  overflow: hidden;
 }
 
 .annotation-citation-domain {
--- a/h/templates/client/annotation.html
+++ b/h/templates/client/annotation.html
@@ -56,7 +56,7 @@
 <!-- Excerpts -->
 <section class="annotation-section"
          ng-repeat="target in vm.annotation.target track by $index">
-  <blockquote class="annotation-quote"
+  <blockquote class="annotation-quote excerpt"
               ng-hide="target.diffHTML && vm.showDiff"
               ng-bind-html="selector.exact"
               ng-repeat="selector in target.selector
@@ -77,7 +77,7 @@
 
 <!-- Body -->
 <section name="text"
-         class="annotation-body"
+         class="annotation-body excerpt"
          ng-model="vm.annotation.text"
          ng-readonly="!vm.editing"
          markdown>
--- a/h/static/scripts/app.coffee
+++ b/h/static/scripts/app.coffee
@@ -98,6 +98,7 @@ module.exports = angular.module('h', [
 
 .directive('annotation', require('./directive/annotation'))
 .directive('deepCount', require('./directive/deep-count'))
+.directive('excerpt', require('./directive/excerpt'))
 .directive('formInput', require('./directive/form-input'))
 .directive('formValidate', require('./directive/form-validate'))
 .directive('markdown', require('./directive/markdown'))
--- a/h/static/scripts/directive/annotation.coffee
+++ b/h/static/scripts/directive/annotation.coffee
@@ -393,12 +393,14 @@ module.exports = [
         # Propagate changes through the counters.
         scope.$watch (-> ctrl.editing), (editing, old) ->
           if editing
+            elem.addClass('editing')
             counter.count 'edit', 1
             # Disable the filter and freeze it to always match while editing.
             if thread? and threadFilter?
               threadFilter.active(false)
               threadFilter.freeze(true)
           else if old
+            elem.removeClass('editing')
             counter.count 'edit', -1
             threadFilter?.freeze(false)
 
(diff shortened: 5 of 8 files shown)


h/static/scripts/directive/markdown.coffee
```diff
@@ -328,6 +328,23 @@ module.exports = ['$filter', '$sanitize', '$sce', '$timeout', ($filter, $sanitiz
             $sanitize textToCheck.substring(startMath, endMath)
       return textToCheck
 
+    scope.checkLength = (text) ->
```
tilgovi: I don't understand what this section is for. Isn't the excerpt directive supposed to be responsible for all this?

h/static/styles/annotations.scss
```diff
@@ -162,3 +178,70 @@ privacy {
...
+  background: #fff;
+  font-style: italic;
+  font-family: Georgia,"Bitstream Charter","Times","Times New Roman",serif;
+
+  &:hover {
+    color: #BD1C2B;
+  }
+}
+
+.gradient {
```
tilgovi: Why a separate gradient class? It doesn't suggest to me, as a reader, what kind of gradient it is (is it horizontal? vertical? what color? etc etc) so I don't find it super usable.

If there isn't a time when we wish to have the excerpt more/less without a gradient then just use the gradient always, as I did in the original patch I posted.

h/static/styles/styled-text.scss
```diff
@@ -1,5 +1,26 @@
...
+.excerpt {
+  &.show-full-body {
+    .styled-text {
+      position: inherit;
+    }
+  }
+
+  .styled-text {
+    position: relative;
+    top: -40px;
```
tilgovi: What's this -40px doing and why does the position need to be set and reset?

h/templates/client/annotation.html
```diff
@@ -45,11 +45,13 @@
 <section class="annotation-section"
          ng-repeat="target in vm.annotation.target track by $index">
   <blockquote class="annotation-quote"
+              excerpt=""
```
tilgovi: When directives take no arguments and are tightly coupled to styling it might just make sense to restrict them to classes instead and to invoke them by applying the class to the element rather than an attribute.

h/static/styles/annotations.scss
```diff
@@ -71,6 +78,15 @@
   }
 }
 
+.annotation-body {
```
tilgovi: These styles should probably all be handled by the excerpt directive and it's corresponding classes otherwise it becomes entangled with the annotation.

Writing the `excerpt` as a directive makes the most sense if it can be isolated. If the annotation needs its own styles in order to interact with it things look less like modular components and more like spaghetti that spreads out across multiple files and modules that are not, in fact, really separate.

There are certainly occasions where it makes sense to have contextual special cases for style interactions, but if that's the case I recommend getting the excerpt directive to control semantically meaningful class names and then use combinations of selectors in the CSS to make the annotation behave accordingly.

tl;dr `show-full-body` is explicitly about the annotation body, and is much too narrow. It means something has to be aware of the interaction with the more/less excerpts _and_ understand the context in which its used (the annotation body) in order to toggle the appropriate classes. Much better would be to have `.excerpt-expanded .annotation-body` or even define things the inverse way by saying what `.excerpt .annotation-body` looks like and causing the `excerpt` class to be removed.

h/static/scripts/directive/markdown.coffee
```diff
@@ -349,6 +367,7 @@ module.exports = ['$filter', '$sanitize', '$sce', '$timeout', ($filter, $sanitiz
       scope.preview = false
       output.style.height = ""
       ctrl.$render()
+      $timeout -> scope.checkLength()
```
tilgovi: Wherever you are using `$timeout` to allow the entire DOM subtree to be evaluated by Angular you should investigate whether `$scope.$evalAsync` works better. Otherwise, `$timeout` will introduce unnecessary delays digest cycles. `evalAsync` will batch, whereas every `$timeout` is individually scheduled by the JS runtime through a call to `window.setTimeout`.

h/static/scripts/directive/excerpt.coffee
```diff
@@ -0,0 +1,14 @@
+module.exports = [ '$timeout', ($timeout) ->
+    link: (scope, elem, attr, ctrl) ->
+      $timeout -> if elem[0].scrollHeight > elem[0].clientHeight
+        elem.prepend angular.element '<span class="more-quote"> More...</span>'
+        elem.append angular.element '<span class="less-quote"> Less ^</span>'
+        elem.find('.more-quote').on 'click', ->
+          $(this).hide()
+          elem.addClass('show-full-quote')
+        elem.find('.less-quote').on 'click', ->
```
tilgovi: I would make these more semantically general and meaningful or the utility of `excerpt` is greatly diminished. It may be that `excerpt` is the wrong name if it's implying that we're showing a quotation when in fact all we're doing is truncating / collapsing / or by whatever other term affecting the visual presentation of any arbitrary thing (quotation, comment, image, whatever) in its original context (not excerpted).

JakeHartnell:
Thanks @tilgovi for the awesome comments! I'll get to work on the changes. :+1:

tilgovi:
@JakeHartnell you have a commit in there with the message "Squash" that I think you meant to squash instead.

h/static/scripts/directive/excerpt.coffee
```diff
@@ -0,0 +1,12 @@
+module.exports = ->
```
tilgovi: It be great if you could add some angular-style docstrings here.

```
###*
# @ngdoc directive
# @name excerpt
# @restrict C
# @description
# Blah blah blah
###
```

h/static/scripts/directive/excerpt.coffee
```diff
@@ -0,0 +1,12 @@
...
+  link: (scope, elem, attr, ctrl) ->
+    scope.$evalAsync -> if elem[0].scrollHeight > elem[0].clientHeight
```
tilgovi: This is strange indentation. I would prefer if you indent the rest of the block another level and put the `if` statement on its own line.

h/static/scripts/directive/excerpt.coffee
```diff
@@ -0,0 +1,12 @@
...
+      elem.prepend angular.element '<span class="more"> More...</span>'
+      elem.append angular.element '<span class="less"> Less ^</span>'
+      elem.find('.more').on 'click', ->
+        $(this).hide()
+        elem.addClass('show-full-excerpt')
```
tilgovi: Might be better to namespace all these in a typical style: `excerpt`, `excerpt-reveal` (or whatever word you like), `excerpt-show-more`, `excerpt-show-less`, or something like this.

h/static/styles/annotations.scss
```diff
@@ -63,12 +63,62 @@
...
+
+  .more {
+    position: relative;
+    top: 3.35em;
+    float: right;
+    margin-top: -2em;
+
+    &:before {
+      opacity: .9;
+      content: " ";
```
tilgovi: What's this for?

h/static/styles/annotations.scss
```diff
@@ -63,12 +63,62 @@
...
+      margin-left: -3em;
+      padding-right: 40px;
```
tilgovi: Should this be expressed in ems?

h/static/styles/annotations.scss
```diff
@@ -63,12 +63,62 @@
...
+      background: -webkit-gradient(linear, left top, right top, from(rgba(255, 255, 255, 0)), to(white), color-stop(50%, white));
```
tilgovi: http://compass-style.org/reference/compass/css3/images/#function-linear-gradient

h/static/styles/annotations.scss
```diff
@@ -63,12 +63,62 @@
 
 .annotation-quote {
   @include quote;
+  height: 100%;
```
tilgovi: Is this height necessary?

h/static/styles/annotations.scss
```diff
@@ -63,12 +63,62 @@
...
+  overflow: hidden;
+  position: relative;
+
+  &.show-full-excerpt {
+    max-height: 100%;
+  }
+
+  .more {
+    top: 12em;
+    padding-top: 1.5em;
```
tilgovi: It's strange to me to have this padding if it's positioned absolutely. If you want to use the padding to make the click target easier, make it symmetrical. If you're using it for positioning, just use positioning.

h/static/styles/annotations.scss
```diff
@@ -162,3 +212,25 @@ privacy {
     color: $text-color;
   }
 }
+
+.more, .less {
+  color: rgba(189, 28, 43, 0.5);
```
tilgovi: What's this color / where does it come from? Is there a theme color you can use that makes sense?

h/static/styles/annotations.scss
```diff
@@ -162,3 +212,25 @@ privacy {
...
+  background: #fff;
```
tilgovi: Does this need a background? Won't that do odd things against the gradient behind it?

h/static/styles/annotations.scss
```diff
@@ -162,3 +212,25 @@ privacy {
...
+  font-style: italic;
+  font-family: Georgia,"Bitstream Charter","Times","Times New Roman",serif;
```
tilgovi: This should probably be from one of our font variables rather than an explicit set of families unique to this component.

h/static/styles/annotations.scss
```diff
@@ -162,3 +212,25 @@ privacy {
...
+
+  &:hover {
+    color: #BD1C2B;
+  }
+}
+
+.editing {
+  .annotation-body {
+    .more {
```
tilgovi: You can combine these with the selector `.more, .less`.

h/static/styles/annotations.scss
```diff
@@ -63,12 +63,62 @@
...
+  }
+
+  .more {
+    top: 12em;
+    padding-top: 1.5em;
+    width: 100%;
+    height: 3em;
+    position: absolute;
+    z-index: 5;
+    background: -webkit-gradient(linear, left bottom, right bottom, from(rgba(255, 255, 255, 0)), to(white), color-stop(50%, white));
```
tilgovi: These seem the same as the `.more:before` block above. Do we need both?

tilgovi:
Do you think the "Less ^" is clear? If we're going to use an ellipsis on "More" let's use a unicode ellipsis. I wonder if we shouldn't just say "Show More" and "Show Less". Also, ellipsis on a link implies a subsequent interaction to me, like a confirmation step, rather than that this is going to reveal more.

JakeHartnell:
@tilgovi, this works well with `scope.$watch` on the element's scrollHeight. Is that ok for now or should I find a way of rewriting it to not use `scope.$watch`?

tilgovi:
You should rewrite it to watch the values that actually affect the height rather than watching the height itself.

Checking for layout-dependent characteristics is expensive and is typically one of the biggest sources of slowness in front end applications. Doing it in every digest cycle is sure to be pain.

JakeHartnell:
Ok, yeah, I was reading stuff that was recommending avoiding using $watch too much.

tilgovi:
It's true that many watchers slows an Angular application down, but that's not the problem with this. Very fast watch expressions slow the application down only small amounts. Very slow watch expressions exacerbate the issue. Asking the browser to evaluate the CSS and consult the layout engine to determine the height of something is a notoriously slow thing to do.

So, while having many watches is a problem, it doesn't really get at the heart of the issue I have with this. A watch expression that checks whether the body text has changed would likely be orders of magnitude faster than a watch expression that checks the height.

tilgovi:
If the watch expression that checks for the body text to have changed takes 100 microseconds and you have 100 annotations, that's adding 10ms to the digest cycle. If the watch expression takes 1ms that's adding 100ms to the digest cycle. That could be the difference between smooth and janky animation, a responsive or a sluggish application.

JakeHartnell:
Ok, @tilgovi, I took a stab at not using `$watch.` Step in the right direction? If so I'll clean up the CSS and write some tests.

tilgovi:
No. You're still not understanding. Use `$watch`. Just watch the properties that cause the height to change rather than watching the height directly.

tilgovi:
I'll help out with this this week.

JakeHartnell:
I've been trying to get everything to work by 'watching the properties that cause the height to change rather than watching the height directly.' This works great with editing, but I'm a little stuck with thread collapsing. I've tried watching whether the parent is collapsed, but that only works for the direct parent, leading to a situation where the excerpt doesn't always apply correctly...

This is why I injected scope into the thread controller to do `$scope.$broadcast('threadToggleCollapse', value)`. 

So I need a bit more direction on what I should be watching to get excerpts to work correctly with threads. I'll try another stab this week, but any thoughts would be helpful.

tilgovi:
I think we probably need to be doing something smarter about how/where the `.more` is inserted in the hierarchy or what its styling is so that it doesn't need to be explicitly positioned. That would make it unnecessary to do anything for the collapse toggling.

JakeHartnell:
> I think we probably need to be doing something smarter about how/where the .more is inserted in the hierarchy or what its styling is so that it doesn't need to be explicitly positioned. That would make it unnecessary to do anything for the collapse toggling.

Should we just do character or word counting like I had initially?

tilgovi:
No. That is not what I'm suggesting. I want to set `max-height` and be done with it.

JakeHartnell:
Closing, will reopen with changes.

PR #2320, opened 2015-06-15
Simplify database session handling

nickstenning:
Our current use of `pyramid_basemodel` and `hem` to set up and retrieve database sessions involves a fair amount of indirection. Additionally, `pyramid_basemodel` doesn't support having two distinct scoped sessions within the same application, which is necessary if we wish to keep the API's database access distinct from the app's.

Given that we are not using any of `pyramid_basemodel`'s more fancy features, this PR removes our use of it, and of the support library `hem`, in favour of a few dozen lines of code to handle our own database connections. This way, if we want a distinct database session for the API, we can simply duplicate this code into the API application.

Of note:

- we still use `pyramid_tm` to integrate session management with pyramid's request-response cycle
- views no longer need to `from hem import get_session; db = get_session(request)` everywhere: instead they just refer to `request.db`
- model classes no longer need `request` passed into every single class method, as they can simply use the scoped session as intended.
- model classes now all have a class property, `query`, which either they or others can use to query the database inside a request:

  ```python
  User.query.filter(User.name.like(u'%ilgov%')).first()  # => <User: tilgovi>
  ```
- the `should_create_all`/`should_drop_all` config variables have been moved under our config namespace.

It's worth noting explicitly that this PR doesn't actually change how database sessions are set up or managed. Access to the database session through the request object via `hem` suggested that the session was being passed explicitly and was not in fact a thread local, but this isn't true: under the hood `pyramid_basemodel` uses the same simple and battle-tested thread-local `scoped_session` as most sqlalchemy applications.


h/accounts/models.py
```diff
@@ -279,8 +258,5 @@ def _username_to_uid(username):
     return username.replace('.', '').lower()
 
 
-def includeme(config):
-    registry = config.registry
-
-    if not registry.queryUtility(IDBSession):
-        registry.registerUtility(Session, IDBSession)
+def includeme(_):
```
tilgovi: Should we just remove this?

tilgovi:
Nice.

Issue #1691, opened 2014-11-16
Visibility setting is not always clear to users


dwhly:
@aron and I discussed a possible banner that would alert early users to the implications of the visibility status they've selected.   Either that what they're about to post is public, or that what they're about to post is private.  Either can be potentially confusing, depending on what the default setting is.  The banner might be able to be dismissed, or might only show for the first annotation.

gergely-ujvari:
@tilgovi: Yes, IMO we want a separate issue for introducing a library for walkthroughs. Meanwhile let's decide on how to solve this issue too.

So let's make a poll.
- Do we want to implement @Treora's idea about changing the save to publish for public visibility?
- Do we want to move the visibility settings closer to the save/publish button?
- etc?

BigBlueHat:
This just came up on the support@ list. I dug a bit into where the user was looking, and here was the response:
> I was looking at the bottom of the window – because I was ready to
"Save" and then paused, thinking "I remember there was some mention in
the website description of being able to set an annotation as public
or private," so I was hunting around for something in the vicinity of
the "Save" button – probably on the idea that it would either "save as
public" or "save as private".

So. I did this mockup. :smile_cat: 

---
![permissions rework](https://cloud.githubusercontent.com/assets/43209/6856350/ed3297e0-d3d5-11e4-9e3d-cb091e6897e2.png)

---
The one on the far right is the proposed solution. It puts the visibility options near the `Save` button (where folks are looking). Puts the current `CC0` note next to the visibility choice (which should hide if you switch to "Only Me"--note: it currently doesn't). It's also about 0.7em shorter thanks to dropping the user name out of the header. :smile: It was lonely, unnecessary, taking up space, and already in the top-right of the sidebar.

@RawKStar77 would love your thoughts on this draft. It should be fairly easy to implement...for someone who knows our angularz. :wink:

tilgovi:
Sounds like a totally reasonable proposal.

RawKStar77:
I would put the Save and Cancel button below the visibility setting. In the flow of creating an annotation, Save is the last thing you click, so it feels weird to me to put the visibility setting below the save button.

BigBlueHat:
I'd actually wanted it up next to the `Save` button, but I couldn't quite make that work without more drastic UI movement.

Above it is OK by me, but I (personally) don't want to have to skip over it visually every time. It's a secondary thought to the `Save` action--again, for me personally. :smile: I think "I want to save this...oh...who else will see this."

[...]

Issue #2285, opened 2015-06-04
Visually distinguish private notes


judell:
I often make private notes without intending to. So do others. (From the support forum today: "When I share an annotation the link doesn't work for other people, only for me.)

An alternate background as shown below is not necessarily the solution, but we need *something* more than the Public / Only Me dropdown to alert people to the fact that Only Me may be stickily in effect when composing a note. It's way too easy to not notice the state of the Public / Only Me dropdown. (And perhaps something more than the lock icon to signal that a note is private.)

![distinguish-private-notes](https://cloud.githubusercontent.com/assets/46509/7991850/1a89a0f6-0ab0-11e5-853e-fae5a8637ada.jpg)

dwhly:
I'd say the most effective way to communicate this is at the point of grabbing the share link.  It should state clearly there that the link is to a private annotation.

judell:
That too. But I have repeatedly made annotations that were unintentionally private, and then been puzzled to not find them in the stream or sidebar when I'm not logged in.

dwhly:
I think @RawKStar77's new styling that will repeat the permission with a text explanation via a banner at the bottom of the annotation when being created will help there. I really like the feel of that.

RawKStar77:
This is what @dwhly is referring too:
![explicitely state annotation visibility setting](https://cloud.githubusercontent.com/assets/521978/7993698/c9dde27c-0abc-11e5-847d-4474304bda92.png)

judell:
That'll help! What branch is that on, and what is it's status?

JakeHartnell:
It's in the groups prototype. I can easily break it out into its own PR, just put it in the backlog.

tilgovi:
@JakeHartnell in general, break out as many things as you have the patience for. It makes reviewing the eventual PR for groups easier.

JakeHartnell:
:+1:

JakeHartnell:
Resolved with https://github.com/hypothesis/h/pull/2322

PR #2322, opened 2015-06-15
Make annotation visibility level more explicit

JakeHartnell:
- Meant to resolve https://github.com/hypothesis/h/issues/1691 and https://github.com/hypothesis/h/issues/2285
- Privacy directives calls setLevel on the parent annotation's scope to display a clarifying message about who will be able to see the annotation
on save.
- Small styling refinements have been added to make the annotation-license and the newly added clarifying text the same size.
- The word "Private" has been added to the lock icon.

Private annotations are called out with the lock-icon and text:
Private annotation

A clarifying note is added when editing/creating annotation, explicitly stating what the annotation visibility level is:
Making a new annotation

### DIFF
--- a/h/templates/client/annotation.html
+++ b/h/templates/client/annotation.html
@@ -8,8 +8,11 @@
        target="_blank"
        ng-href="{{vm.baseURI}}u/{{vm.annotation.user}}"
        >{{vm.annotation.user | persona}}</a>
-    <i class="h-icon-lock" ng-show="vm.isPrivate() && !vm.editing"></i>
     <i class="h-icon-border-color" ng-show="vm.isHighlight() && !vm.editing" title="This is a highlight. Click 'edit' to add a note or tag."></i>
+    <span ng-show="vm.isPrivate() && !vm.editing"
+          title="This annotation is viewable only to you.">
+      <i class="h-icon-lock"></i> Private
+    </span>
     <span class="annotation-citation"
           ng-if="!vm.embedded"
           ng-show="vm.document.title">
@@ -24,6 +27,7 @@
       <privacy ng-click="$event.stopPropagation()"
                ng-if="vm.annotation.permissions && vm.editing && action != 'delete'"
                ng-model="vm.annotation.permissions"
+               level="vm.privacyLevel"
                user="{{vm.annotation.user}}"
                class="dropdown privacy pull-right"
                name="privacy" />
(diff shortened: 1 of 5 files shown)


h/templates/client/annotation.html
```diff
@@ -97,6 +97,13 @@
 </div>
 <!-- / Tags -->
 
+<div class="annotation-section small" ng-if="vm.editing">
+  <p ng-show="level.text == 'Only Me'">
```
tilgovi: You can just use `vm.isPrivate()` here and then you don't need `setLevel` and you don't need to reach "up" out of the isolated privacy directive scope (which is naughty thing to do).

JakeHartnell: Hah! Duh. Will fix asap.

JakeHartnell: I forgot I wasn't dealing with groups anymore.

JakeHartnell: vm.isPrivate( ) works, but doesn't change when you change the privacy control, only on save.

gif of this behavior

JakeHartnell: In other words, this scope of the help text should be tied to the current level of the Privacy directive. The way things are currently written, does it make sense that current state of the privacy directive is so separate from the annotation directive?

Should I try binding a property from the annotation controller to the isolated scope of the privacy directive?

Example:
```
scope:
    level: '='
```

Or is this also a form of 'reaching up' that I should avoid?

tilgovi:
You're absolutely right, of course, and I'm sorry to have lead you down that dead end.

Using the two-way binding in the isolated scope is fine, but then you have to pass the value through an attribute on the privacy directive, e.g. `<privacy level="vm.privacyLevel">`. That's a fine way to approach it, although I start to wonder whether the privacy directive is actually still usefully seen as an isolated component, but I suppose you needn't refactor that all now.

JakeHartnell:
Ok, it's now using two way binding.

tilgovi:
I think some might disagree, but since the value isn't actually used by the controller at all but is only ever used locally in the annotation template maybe just remove the line from the controller and skip the `vm.` namespace. I'm fine with template-local things. We don't get the testability that having an API on the controller would provide, but such is the nature of markup. One day UI tests or end-to-end tests could cover this.

JakeHartnell:
> I think some might disagree, but since the value isn't actually used by the controller at all but is only ever used locally in the annotation template maybe just remove the line from the controller and skip the vm. namespace.

I tried this, but couldn't get it to work.

tilgovi:
It was a bad suggestion on my part. All template variables should be namespaced because otherwise they get shadowed all over the place, even by things like `ng-if`.

I rebased it; moved the initializer to `null` to a controller property (you had it as `$scope.privacylevel` whereas the controller in the template is `vm`, or `this` in the controller); and camelCased `privacyLevel`.

JakeHartnell:
Thanks @tilgovi!

PR #2323, opened 2015-06-16
Fix document equivalence lookups

seanh:
These were broken by the recent search API refactoring

### DIFF
--- a/h/api/search.py
+++ b/h/api/search.py
@@ -8,9 +8,30 @@ import logging
 
 import webob.multidict
 
-from h.api.models import Annotation
+from h.api import models
 
 log = logging.getLogger(__name__)
+
+
+def _match_clause_for_uri(uri):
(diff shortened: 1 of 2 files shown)


h/api/search.py
```diff
@@ -7,12 +7,34 @@
 import logging
 
 import webob.multidict
+from annotator import document
```
nickstenning: This should be `from h.api import models`. The fact that the `Document` model comes from `annotator` is an implementation detail.

nickstenning:
One minor note about where you import your models from, and suggest you rebase into a single commit, but LGTM otherwise.

seanh:
@nickstenning Done.

If I have time later today I'll also investigate whether the search refactoring broke date-range searches with the before and after params, looks like it might have, and send a separate pr for that.

PR #2330, opened 2015-06-19
Small ux improvements

JakeHartnell:
There are 3 small things in this PR:
1. Kill note icon on card: *It was the same as the icon for make new note... which was confusing, and arguably did little to help people understand what was going on.* We obviously still need something else with regards to distinguishing page comments, but that's a separate issue.
Comment, without note icon

2. Slightly more padding on menus: *with mobile I like to have more padding on the menus so that it is easier to select things with my big fingers on a small screen.*
More padding on menus

3. Adding a tooltip to the highlight icon:
Tooltip on card highlight icon

Thoughts? Hopefully these aren't too controversial.

### DIFF
--- a/h/templates/client/annotation.html
+++ b/h/templates/client/annotation.html
@@ -10,7 +10,6 @@
        >{{vm.annotation.user | persona}}</a>
     <i class="h-icon-lock" ng-show="vm.isPrivate() && !vm.editing"></i>
     <i class="h-icon-border-color" ng-show="vm.isHighlight() && !vm.editing"></i>
-    <i class="h-icon-insert-comment" ng-show="vm.isComment()"></i>
     <span class="annotation-citation"
           ng-if="!vm.embedded"
           ng-show="vm.document.title">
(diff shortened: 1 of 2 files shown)


judell:
+1 for the tooltip. FWIW I'd like us to use them everywhere possible. Here, for example, we could say on the lock icon "Private Note"

h/templates/client/annotation.html
```diff
@@ -9,8 +9,7 @@
        ng-href="{{vm.baseURI}}u/{{vm.annotation.user}}"
        >{{vm.annotation.user | persona}}</a>
     <i class="h-icon-lock" ng-show="vm.isPrivate() && !vm.editing"></i>
-    <i class="h-icon-border-color" ng-show="vm.isHighlight() && !vm.editing"></i>
-    <i class="h-icon-insert-comment" ng-show="vm.isComment()"></i>
+    <i class="h-icon-border-color" ng-show="vm.isHighlight() && !vm.editing" title="This is a highlight, click 'edit' to add a note or tag."></i>
```
tilgovi: grammar note: should have a semicolon or a period, not a comma.

tilgovi: Of course I reply with no capitalization and with the same mistake. ;)

JakeHartnell:
Ok. Grammar updated. : )

> +1 for the tooltip. FWIW I'd like us to use them everywhere possible. Here, for example, we could say on the lock icon "Private Note"

I'm addressing this in: https://github.com/hypothesis/h/pull/2322

PR #2334, opened 2015-06-26
Don't unconditionally search for http://example.com/

### LINKED ISSUE #2326: Unexpected annotation found for unannotated URLs
Pick any URL with (presumably) no H annotations. Three examples:

https://hypothes.is/api/search?uri=http://www.benkler.org
https://hypothes.is/api/search?uri=http://www.whitehouse.gov
https://hypothes.is/api/search?uri=http://microsoft.com

The same spurious annotation (a page note) appears in each case:

```
{
    "total": 1, 
    "rows": [
        {
            "updated": "2015-03-29T13:29:23.388422+00:00", 
            "created": "2015-03-29T13:29:23.388406+00:00", 
            "text": "This is an example annotation.", 
            "tags": [], 
            "uri": "http://example.com/", 
            "user": "acct:lampholder@hypothes.is", 
            "document": {
                "eprints": {}, 
                "title": "Example Domain", 
                "twitter": {}, 
                "dc": {}, 
                "prism": {}, 
                "highwire": {}, 
                "facebook": {}, 
                "link": [
                    {
                        "href": "http://example.com/"
                    }
                ]
            }, 
            "consumer": "00000000-0000-0000-0000-000000000000", 
            "id": "3ro_gCA2TjuwkDXsArDLCA", 
            "permissions": {
                "admin": [
                    "acct:lampholder@hypothes.is"
                ], 
                "read": [
                    "group:__world__"
                ], 
                "update": [
                    "acct:lampholder@hypothes.is"
                ], 
                "delete": [
                    "acct:lampholder@hypothes.is"
                ]
            }
        }
    ]
}
```

nickstenning:
When we fail to find a document for the passed URI, we should probably
fall back to a naive match query for the passed URI, rather than a naive
match query for the literal string "http://example.com/".

...!

Fixes #2326.

### DIFF
--- a/h/api/search.py
+++ b/h/api/search.py
@@ -31,7 +31,7 @@ def _match_clause_for_uri(uri):
             }
         }
     else:
-        return {"match": {"uri": "http://example.com/"}}
+        return {"match": {"uri": uri}}
 
 
 def build_query(request_params):
--- a/h/api/test/search_test.py
+++ b/h/api/test/search_test.py
@@ -242,12 +242,17 @@ def test_build_query_for_uri(models):
     """
     models.Document.get_by_uri.return_value = None
 
-    query = search.build_query(
+    query1 = search.build_query(
         request_params=multidict.NestedMultiDict(
             {"uri": "http://example.com/"}))
+    query2 = search.build_query(
+        request_params=multidict.NestedMultiDict(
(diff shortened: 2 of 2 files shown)


tilgovi:
Good find :)

PR #2335, opened 2015-06-26
Suggestion to add uri.parts

bbarker:


tilgovi:
This is as good a place as any. How do we want to define "part".

Right now, it's tokenized on this regular expression: `[#+/:=?.-]|(?:%2[3BF])|(?:%3[ADF])`. The result is that we get tokens for each of the following, each further broken up by hyphens, etc.
- scheme
- domain component
- path component
- query key (further broken up by encoded delimetrs, like %2F)
- query value (further broken up by encoded delimetrs, like %2F)
- fragment (shouldn't be necessary because it should be stripped already)

I'm thinking that the fragment should be ignored and that the query string components should be tokenized intact, i.e. `key=foo%2Fbar` should just be the token `key=foo%2Fbar`, as has come up on the mailing list.

tilgovi:
@judell your guidance here would be appreciated. The use cases for `uri.parts` are (at least) an application user typing free text into the search bar and an API user trying to retrieve annotations based on tokens in the URI.

judell:
Major +1 for not decomposing query terms. But I'd like to know @bbarker's view.

tilgovi:
Superceded by #2344

PR #2337, opened 2015-06-29
Improve thread collapsing UX.

JakeHartnell:
- Threads collapse to single lines
- Use iconography consistent with collapsing / expanding UI in other applications
- Threads are fully expanded when you click replies on a top level annotation

It looks like this:
Thread collapsing


h/templates/client/thread.html
```diff
@@ -2,8 +2,8 @@
    class="threadexp"
    title="{{vm.collapsed && 'Expand' || 'Collapse'}}"
    ng-click="vm.toggleCollapsed()">
-   <span ng-class="{'h-icon-expand-more': !!vm.collapsed,
-                    'h-icon-expand-less': !vm.collapsed}"></span>
+   <span ng-class="{'h-icon-not-expanded': !!vm.collapsed,
```
tilgovi: Are these arrow-drop-down and arrow-drop-up?
https://www.google.com/design/icons/#ic_arrow_drop_down

If so, maybe we don't rename them.

tilgovi:
Looks good, but I think you've now got `arrow-drow-down`, `expanded` and `unexpanded` when we should have only two icons, `arrow-drow-down` and `arrow-drop-up`.

tilgovi:
You'll also have to fix the tests. It would be good to remove the failing test and to test the new behavior of the initial collapse state.

JakeHartnell:
Tests fixed, also cleaned up the code a bit and checked it out in different browsers.

tilgovi:
I like the collapsing. I think this should be down/up rather than right/down and that we shouldn't change it. The icons we have right now are "expand less" and "expand more" in the material icon set. This is what they're for. I don't think these arrows are any clearer. Although, I think it _may_ be _much_ clearer if we were to put them next to the reply count (and maybe not move the reply count).

tilgovi:
I might be wrong about the arrows. Where they're positioned now the way you've got it makes more sense.

lenazun:
Makes sense to me. Should we keep it as a pending detail for a future design review?

tilgovi:
Yep.

Issue #2329, opened 2015-06-18
atom feed 504 gateway timeout


judell:
Seems sensitive to the amount of data returned. A less-restrictive query:

```
curl -L "http://hypothes.is/stream.atom?user=judell"
<html>
 <head>
  <title>504 Gateway Timeout</title>
 </head>
 <body>
  <h1>504 Gateway Timeout</h1>
  The gateway has timed out.<br/><br/>
HTTPSConnectionPool(host='hypothes.is', port=443): Read timed out. (read timeout
=0.2)
```

A more-restrictive query:

```
curl -L "http://hypothes.is/stream.atom?user=judell&tags=reading" 
<? xml version="1.0" encoding="utf-8"?>
<feed xmlns="http://www.w3.org/2005/Atom">
````

(We saw this in stage, speculated it was a stage artifact, evidently not.)

tilgovi:
Probably deployment then. We may need to bump the proxy read timeout in nginx.

tilgovi:
Another thing would be to investigate whether we can generate a streaming response by templating each item separately, rather than iterating in the template, and returning a `Response` object with a generator as the body iterator.

seanh:
The app itself has a (configurable) timeout when waiting for the api to respond - I suspect that rather than the nginx timeout. Double that setting will fix it I think

tilgovi:
You're almost certainly right. The default proxy_upstream_timeout in nginx is 60s. The default API client timeout in h is 0.2s.

tilgovi:
I would probably do more than double it. The cost of having the connection open is low, especially with greenlets for each request. If the client disconnects, I think the greenlet exits anyway, so these aren't going to hang around. Something like 2s might be more appropriate.

seanh:
> I would probably do more than double it. The cost of having the connection open is low, especially with greenlets for each request. If the client disconnects, I think the greenlet exits anyway, so these aren't going to hang around. Something like 2s might be more appropriate.

Seems fine to me. I don't see why it should be set low. Rather wait a couple secs for a page to load, than get a 504. (Of course we should make the app fast as well.)

tilgovi:
Closed in 653f19e3b6a14db542432752ba3cc43e7c0f8ffe

PR #2338, opened 2015-06-30
Atom feed link tweaks

BigBlueHat:
Hopefully fixing some `<link rel="">` issues with the Atom feed.

I chose to use the URI for `oa:hasTarget` as the `rel` value for the annotated resource. It's the most accurate link relation I could find. :smile_cat: 

If someone could give this a quick test run too, I'd appreciate it...my `h` dev environs are still a tangle...

Thanks!

### DIFF
--- a/h/atom_feed.py
+++ b/h/atom_feed.py
@@ -113,6 +113,10 @@ def _feed_entry_from_annotation(
         entry["links"].append({"rel": "alternate", "type": "application/json",
                                "href": annotation_api_url(annotation)})
 
+    for target in annotation.get('target'):
+        entry["links"].append({"rel": "related",
+                               "href": target.get('source')})
+
     return entry
 
 
--- a/h/atom_feed.py
+++ b/h/atom_feed.py
@@ -82,12 +82,10 @@ def _feed_entry_from_annotation(
     }
 
     def get_selection(annotation):
-        targets = annotation.get("target")
-        if targets:
-            for target in targets:
-                for selector in target["selector"]:
-                    if "exact" in selector:
-                        return selector["exact"]
+        for target in annotation.get("target", []):
+            for selector in target["selector"]:
+                if "exact" in selector:
+                    return selector["exact"]
 
     content = ""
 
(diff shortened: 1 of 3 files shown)


tilgovi:
Ping this back after you change the `rel`.

BigBlueHat:
@tilgovi did you want me to switch to outputting all the values from `target.*.source`? or just stick with `annotation.uri` for now?

tilgovi:
@BigBlueHat that'd be ideal!

judell:
Thanks for this @BigBlueHat.

It'd be nice to also address https://github.com/hypothesis/h/issues/2329 so I can test both of these things together. The combo will unblock a blog post I've been meaning to write.

As a reminder, the user story here is RSS widgets (e.g. on WordPress) that want to link to annotation.uri instead of (or in addition to) the bare annotation. So from that perspective. annotation.uri alone should be fine for now.

BigBlueHat:
@judell for #2329 if
https://github.com/hypothesis/h/issues/2329#issuecomment-113298389 is the
solution, I can probably make that happen. If it's bumping the timeout, I'd
leave that with @tilgovi or @seanh.

How about I ship this with just `annotation.uri` for now, and then we
switch it to `target.*.source` once we've deprecated `annotation.uri`. Some
background: @tilgovi and I discussed the painfully incorrect semantics of
`annotation.uri` === the annotated page...and not (as you'd think) the URI
of the annotation. That key's also unlikely to stay around much longer as
we move toward the `target` array and all it's accurateness.

Once we move to `target.*.source` we're going to have to make a
determination about which is the "best"--as we'll need to add a rel value
for that to "draw it out" so WP and others can Do The Right Thing.
Otherwise...they'll have to link to all of them too. :frowning:

tilgovi:
Your wish is my command @judell: 653f19e3b6a14db542432752ba3cc43e7c0f8ffe

judell:
Excellent, thanks gents.

BigBlueHat:
K. That should be ready also. Switched from the `hasTarget` URL to `related` as the `rel`.

tilgovi:
@BigBlueHat I covered for you this time, but next time tests please :).

BigBlueHat:
Thanks! Sorry I forgot those. Next time! :smiley:

PR #2340, opened 2015-06-30
Fix guest embedding

JakeHartnell:
At the hacking the codex event, I found that unless I modified main.js, I couldn't get Epub.js to work as I couldn't instantiate the Annotator.Guest class, getting a nasty 'Annotator is undefined' error, and failing to create a Guest.

The docs currently suggest to assign the constructor function to 'window.hypothesisRole'. This first attempt at addressing the problem simply moves the constructor function into the 'window.hypothesisConfig' options. As such you would simply create a new guest annotator like so:

```
	return {
		Constructor: Annotator.Guest
	};
};
```

Thoughts? Other suggestions? This fixes the problem, but perhaps there is more to be done around our embed code?

I should note that Epub.js is not the only project that will be wanting to create Guest annotators for annotating iframes, so it would be nice if we could get this working so we can move forward with Readium.js integrations, etc.

### DIFF
--- a/h/static/scripts/annotator/main.js
+++ b/h/static/scripts/annotator/main.js
@@ -47,21 +47,12 @@ require('./plugin/textquote');
 require('./plugin/textposition');
 require('./plugin/textrange');
 
-var Klass = Annotator.Host;
-var docs = 'https://github.com/hypothesis/h/blob/master/README.rst#customized-embedding';
+var docs = 'https://h.readthedocs.org/en/latest/hacking/customized-embedding.html';
 var options = {
   app: jQuery('link[type="application/annotator+html"]').attr('href'),
   BucketBar: {container: '.annotator-frame'},
   Toolbar: {container: '.annotator-frame'}
 };
-
-if (window.hasOwnProperty('hypothesisRole')) {
-  if (typeof window.hypothesisRole === 'function') {
-    Klass = window.hypothesisRole;
-  } else {
-    throw new TypeError('hypothesisRole must be a constructor function, see: ' + docs);
-  }
-}
 
 // Simple IE autodetect function
 // See for example https://stackoverflow.com/questions/19999388/jquery-check-if-user-is-using-ie/21712356#21712356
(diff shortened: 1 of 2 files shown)


tilgovi:
Just lowercase `constructor` and I think we're good. Update the docs. I like it.
Also, as a good practice, let's delete the property from the options before passing it to the constructor. This is done because the option isn't used by the class that gets constructed, which doesn't need a handle to its own constructor. Best always not to pass superfluous things.

h/static/scripts/annotator/main.js
```diff
@@ -81,5 +73,6 @@ if (window.hasOwnProperty('hypothesisConfig')) {
 }
 
 Annotator.noConflict().$.noConflict(true)(function () {
-  window.annotator = new Klass(document.body, options);
+  window.annotator = new options.constructor(document.body, options);
+  delete options.constructor;
```
tilgovi: :ghost:

Use a temporary variable and delete it `before` it's passed. That was the whole point of deleting it ;).

h/static/scripts/annotator/main.js
```diff
@@ -47,22 +47,13 @@ require('./plugin/textquote');
...
 require('./plugin/textrange');
 
-var Klass = Annotator.Host;
-var docs = 'https://github.com/hypothesis/h/blob/master/README.rst#customized-embedding';
+var docs = 'https://h.readthedocs.org/en/latest/hacking/customized-embedding.html';
 var options = {
   app: jQuery('link[type="application/annotator+html"]').attr('href'),
   BucketBar: {container: '.annotator-frame'},
-  Toolbar: {container: '.annotator-frame'}
+  Toolbar: {container: '.annotator-frame'},
```
tilgovi: This is a syntax error.

JakeHartnell: My bad, fixed.

tilgovi:
Thanks. This is great. Thank you for the docs updates, too.

JakeHartnell:
:+1:

PR #2344, opened 2015-07-02
Expand the search API docs

seanh:
This adds docs for many undocumented params and features of the search
API.

Also removes docs of the (legacy) quote param. Annotations no longer
have this field.

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -107,12 +107,75 @@ search
           "total": 1
       }
 
-   :query limit: number of results to return
-   :query offset: index into the complete result set
-   :query uri: limit results to annotations of `uri` (must be urlencoded)
-   :query user: limit results to annotations created by `user`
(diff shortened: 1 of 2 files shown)


seanh:
- [ ] Some of the documented features don't have tests
- [ ] When searching for a uri (or uri.parts?) we expand the search to also cover uris of equivalent documents, should document this (may also affect "any" searches)
- [ ] Add todo for documenting the document param

tilgovi:
I don't think uri.parts is expanded and so it doesn't affect any searches either.

URI.parts currently hase `.` as a delimiter, so the example (`example.com`) won't do what it describes.

tilgovi:
Other than that second bit about `uri.parts` this is so solid, thank you.

seanh:
I can't remember what it was that I wanted to add a test for. Calling this done

seanh:
Rebased, the landscape problem should be fixed (ignored) now

PR #2345, opened 2015-07-03
Be slightly more defensive when processing annotations for the atom feed.

nickstenning:
When processing annotations for the Atom feed, we need to be slightly
more defensive -- the assumptions we'd like to make about annotation
structure aren't always true.

Fixes https://app.getsentry.com/hypothesis/prod/group/68815541/.


tilgovi:
Thanks. I saw this yesterday, too, but didn't get a chance to finish the patch. :beers:

PR #2346, opened 2015-07-03
Faster api

nickstenning:
The stream atom feed is currently timing out in production. There is a bunch of different stuff going on here (including the fact that the atom feed makes requests to our API by going out to the public internet and back in rather than directly to the service running on the same host...) but benchmarking suggests that the two biggest sources of slowness are:

1. JSON serialisation (specifically "pretty" output)
2. Making two identical queries to ElasticSearch instead of one...

In addition, I think 1000 annotations in the atom feed is probably a bit much -- an atom feed approaching a megabyte seems excessive. I've dropped the default number of annotations to 100.

With these three changes we go from about 1 request per second...

```
$ wrk -c1 -d10s -t1 'http://localhost:5000/stream.atom'
Running 10s test @ http://localhost:5000/stream.atom
  1 threads and 1 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency   727.94ms   31.91ms 772.22ms   53.85%
    Req/Sec     1.00      0.00     1.00    100.00%
  13 requests in 10.09s, 10.79MB read
Requests/sec:      1.29
Transfer/sec:      1.07MB
```

...to about 13:

```
$ wrk -c1 -d10s -t1 'http://localhost:5000/stream.atom'
Running 10s test @ http://localhost:5000/stream.atom
  1 threads and 1 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency    71.99ms   17.25ms 159.24ms   95.10%
    Req/Sec    13.58      5.06    20.00     57.29%
  139 requests in 10.03s, 12.61MB read
Requests/sec:     13.86
Transfer/sec:      1.26MB
```

### DIFF
--- a/h/app.py
+++ b/h/app.py
@@ -5,7 +5,6 @@ import logging
 import os
 
 from pyramid.config import Configurator
-from pyramid.renderers import JSON
 from pyramid.wsgi import wsgiapp2
 
 from h.api.middleware import permit_cors
@@ -97,7 +96,6 @@ def create_api(global_config, **settings):
     config.set_authorization_policy(acl_authz)
     config.set_root_factory('h.api.resources.create_root')
 
-    config.add_renderer('json', JSON(indent=4))
     config.add_subscriber('h.api.subscribers.set_user_from_oauth',
                           'pyramid.events.ContextFound')
     config.add_tween('h.api.tweens.auth_token')
--- a/h/api/search.py
+++ b/h/api/search.py
@@ -124,10 +124,13 @@ def search(request_params, user=None):
               request_params.get('uri'))
 
     query = build_query(request_params)
-    results = models.Annotation.search_raw(query, user=user)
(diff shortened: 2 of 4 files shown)


h/app.py
```diff
@@ -97,7 +97,7 @@ def create_api(global_config, **settings):
     config.set_authorization_policy(acl_authz)
     config.set_root_factory('h.api.resources.create_root')
 
-    config.add_renderer('json', JSON(indent=4))
+    config.add_renderer('json', JSON())
```
tilgovi: You can just remove this line then. The 'json' renderer is a default and I was just overriding it here.

tilgovi:
Other than my comment about the 'json' renderer this is great. Merge it after you fix that.

tilgovi:
Never mind. I've got it.

PR #2348, opened 2015-07-06
Search for one URI

nickstenning:
Previously, when loading Hypothesis into a page, the client would make
one search request for each and every URI found on the page, including
those found by the document metadata scanners (in the Annotator
"Document" and "PDF" plugins.)

This is wasteful of network bandwidth, and moreover should not be
necessary in most cases due to the ability to expand to equivalent URIs
on the server side.

The component responsible for coordinating communication across page
frames is the "crossframe" service. This services maintains a registry
of page frames (previously called "providers") and their associated
metadata. This commit reduces the metadata stored to simply the page
URL, or in the case of PDF documents, the fingerprint URN.

### DIFF
--- a/h/static/scripts/annotation-ui-sync.coffee
+++ b/h/static/scripts/annotation-ui-sync.coffee
@@ -1,4 +1,4 @@
-# Uses a channel between the sidebar and the attached providers to ensure
+# Uses a channel between the sidebar and the attached frames to ensure
 # the interface remains in sync.
 module.exports = class AnnotationUISync
   ###*
--- a/h/static/scripts/cross-frame.coffee
+++ b/h/static/scripts/cross-frame.coffee
@@ -1,7 +1,5 @@
 # Instantiates all objects used for cross frame discovery and communication.
 module.exports = class CrossFrame
-  providers: null
-
   this.inject = [
     '$rootScope', '$document', '$window', 'store', 'annotationUI'
     'Discovery', 'bridge',
@@ -12,7 +10,7 @@ module.exports = class CrossFrame
     Discovery, bridge,
     AnnotationSync, AnnotationUISync
   ) ->
-    @providers = []
+    @frames = []
 
     createDiscovery = ->
       options =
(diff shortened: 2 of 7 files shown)


nickstenning:
@tilgovi fancy reviewing this?

tilgovi:
If it pleases I'd like to wait until I merge anchoring-rewrite. Otherwise, that's going to get messy.

tilgovi:
I'll get a PR for it open today.

nickstenning:
Rebased.

tilgovi:
This is fine, but it will cause the first person to visit a new URL to not see any annotations. I'm not sure that's acceptable.

For instance, if one person annotates a page heavily and shares it with a friend, that friend might show up and find themselves redirected to the mobile version and see no annotations.

tilgovi:
Would love @judell thoughts here.

What's changing is that we won't be returning all the annotations on all the aliases for a page _until an annotation is saved that adds the alias to the server-side database_.

tilgovi:
I might prefer to see us send multiple URIs in a post body or something rather than just send one URI.

nickstenning:
So in the long run, I'm imagining that the submission of document metadata will simply be orthogonal. When you make a search query, a prior request will have submitted the metadata for the page you're on.

If we want to hold off on breaking this edge case until then, we'll need a way to submit `OR` queries for multiple URLs, which we don't currently have. I'm not overly keen to change the public API in this way without some careful thought.

nickstenning:
And just so we're clear, it doesn't necessarily mean the first person visiting that page sees no annotations, it means the first person visiting a page *for which we have no equivalence metadata, but where equivalence metadata exists on that page* sees no annotations.

So in the example you gave, I'm assuming it's pretty likely the original annotation(s) will have carried equivalence metadata that referenced the mobile site, and thus it will all work just fine.

The clear case is where annotations are first made on a PDF representation -- there the only metadata we *can* retrieve is the current URL and the PDF fingerprint -- so people who visit the HTML representation (even if that HTML representation contains a link to the PDF) will be missing out.

I'm sure you're aware of all this -- just adding it here for clarity for other readers.

tilgovi:
Is it typical to have, say, rel=alternate on a site that points at the
mobile version? I didn't think so. Maybe I'm wrong.

judell:
"So in the long run, I'm imagining that the submission of document metadata will simply be orthogonal. When you make a search query, a prior request will have submitted the metadata for the page you're on."

My understanding also.

"It means the first person visiting a page for which we have no equivalence metadata, but where equivalence metadata exists on that page sees no annotations."

Got it (I think), thanks.

"It's pretty likely the original annotation(s) will have carried equivalence metadata that referenced the mobile site"

Checking my understanding: that equivalence metadata was gathered server-side when the original annotation arrived, correct? (Or, perhaps, injected by manual or automated mechanisms as yet undefined.)

tilgovi:
I'm gonna merge this optimistically. Sites that are serious about SEO should have a rel=alternate to the mobile version just as they have a rel=canonical back to a canonical version.

We'll see how it goes.

PR #2349, opened 2015-07-06
Thread collapsing redux

JakeHartnell:
This reintroduces the thread collapsing improvements as described in #2337.

It appears that commit https://github.com/hypothesis/h/commit/cdddb16bdfd075b4268c8f5b883fa79b54f554b6 broke thread collapsing by hiding the bodies of top level annotations. This has been fixed by a small adjustment to annotations.scss.


tilgovi:
Cool. I'm poking at this and found a couple rough edges. Collapsing annotations while they are being edited results in awkwardly having save/cancel buttons without the annotation body. Hiding these results in a small, excess margin on annotations being edited. Hiding this results in new drafts disappearing completely when collapsed if the user is not logged in. I've mostly addressed all these and I'll push in the morning for your review.

PR #2354, opened 2015-07-08
Proper feature flags

nickstenning:
The way we had implemented feature flags was pretty inflexible, and
wasn't effectively allowing us to do the one thing that feature flags
are really intended to do, namely:

> Decouple deployment (when you push code to production) from release
> (when users see new features).

A truly effective feature flag system is one which achieves the above,
and also enables more dynamic options for "releasing" such as:

- release to admin users/beta users only
- release to a randomly selected 10% of users
- release only to one specific user

This requires that the feature flag system have some knowledge of the
accounts system.

Furthermore, in order to make the feature flag system truly atomic and
reversible (i.e. shit has gone wrong, let's flip that back) we need to
be able to toggle feature flags instantly, without needing to wait for
an application redeploy.

Both of these requirements indicate a feature flag system fully
integrated with the request-response cycle of the application, and one
where flag status can be more complex than a simple boolean.

In order to make this possible, this commit puts feature flags into a
"feature" table of the main application database. This allows us to
extend the definition of flag state as and when we see fit to enable
some or all of the release options denoted above.

The initial state of all feature flags is off, but in order to
facilitate an easy migration path for us the data migration with this
commit will read the current state of the existing flags from the system
environment and write appropriate rows into the table.

--

For what it's worth, I've done this work now because it's very important to be able to make an atomic flip from one data store to another (such as a migration from elasticsearch to postgres), which is a piece of work I'm currently planning out.

### DIFF
--- a/conf/development.ini
+++ b/conf/development.ini
@@ -22,11 +22,6 @@ es.host: http://localhost:9200
 
 h.autologin: True
 
-h.feature.claim: False
-h.feature.notification: False
-h.feature.queue: False
-h.feature.streamer: False
-
 mail.default_sender: "Annotation Daemon" <no-reply@localhost>
 
 pyramid.debug_all: True
--- a/conf/development.ini
+++ b/conf/development.ini
@@ -22,8 +22,6 @@ es.host: http://localhost:9200
 
 h.autologin: True
 
-h.feature.accounts: True
-h.feature.api: True
 h.feature.claim: False
 h.feature.notification: False
 h.feature.queue: False
--- a/conf/production.ini
+++ b/conf/production.ini
@@ -18,8 +18,6 @@ use: egg:h
 #h.client_secret:
 
 # Feature flags
-h.feature.accounts: True
-h.feature.api: True
 h.feature.claim: False
 h.feature.notification: True
 h.feature.queue: True
--- a/conf/production.ini
+++ b/conf/production.ini
@@ -17,12 +17,6 @@ use: egg:h
 #h.client_id:
 #h.client_secret:
 
-# Feature flags
-h.feature.claim: False
-h.feature.notification: True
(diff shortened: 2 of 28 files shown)


tilgovi:
LGTM. You can merge it yourself. I'll leave it to you in case you want to fix any of the lint messages.

tilgovi:
Tests don't run on travis, though. I don't know what that's about. The feature table not being created automatically?

nickstenning:
Yep, looks like I've screwed something up. Will take a look tomorrow.

nickstenning:
So, the issue with the tests was that building an extension requires templates to be rendered and adding support for dynamic feature toggles now implies that rendering templates requires a database. Joy.

Anyway, as it happens this all goes away with proper dynamic front-end feature flags, so I have extended this PR with the code needed to bring dynamic feature toggles to the client.

I'm not going to merge this myself now, as this is new code and I'd like it to be reviewed, but here's an overview:

- when this is merged, we can add feature toggles into code that spans frontend and backend, and deploy without any fear of this code being enabled before we're ready to release it
- we can toggle a feature in the database (and, in due course, in an admin UI) and it will be enabled across backend and frontend code without requiring any new deployment
- we can even toggle a feature without requiring a page reload on the frontend, as feature flags are only cached on the frontend for 5 minutes

nickstenning:
And all three lint errors reported by prospector now are spurious, IMO.

tilgovi:
Looks great. Only change I made was to assets.yaml so we're not rebuilding everything all the time in dev.

We should maybe move the final build artifacts about scripts. In other words, h/static/hypothesis.js and h/static/app.js and h/static/app.css or something.

JakeHartnell:
Very nice! :+1:

PR #2356, opened 2015-07-12
Update annotation.coffee

judell:


judell:
Until drafts persist, let's help people avoid losing their work.

nickstenning:
If we want to fix the problem, let's fix the problem. Until then, I think giving people workarounds just makes it look like we don't care.

tilgovi:
Agreed, @nickstenning

BigBlueHat:
Proposing #2361 over this one--given feedback that said updating text wasn't enough.

PR #2357, opened 2015-07-13
Fix token command for non default ports

seanh:
Also improve the docs

### DIFF
--- a/h/script.py
+++ b/h/script.py
@@ -104,9 +104,10 @@ def token(args):
     if not args.config_uri.endswith('#api'):
         args.config_uri += '#api'
 
+    request = Request.blank("/", base_url=args.base)
+
     paster.setup_logging(args.config_uri)
-    env = paster.bootstrap(args.config_uri)
-    request = env['request']
+    env = paster.bootstrap(args.config_uri, request=request)
     registry = env['registry']
 
     request.client = get_client(request, registry.settings['h.client_id'])
--- a/h/script.py
+++ b/h/script.py
@@ -4,6 +4,7 @@ from __future__ import print_function
 import argparse
 import os
 import sys
+import textwrap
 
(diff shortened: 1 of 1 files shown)


seanh:
@nickstenning

seanh:
Wondering about adding a mention of this command to our API docs in api.rst as well. It's only useful to developers working on their local instance as far as I can see (you have to pass the paster config file to the command), but at least those devs will be able to find it

tilgovi:
Is the indentation on that docstring strange?

h/script.py
```diff
@@ -118,12 +133,18 @@ def token(args):
 
     print(token)
 
-parser_token = subparsers.add_parser('token', help=token.__doc__)
+parser_token = subparsers.add_parser(
+    'token', description=token.__doc__,
```
nickstenning: If you could employ `textwrap.dedent` here ([c.f. `hypothesis-buildext`](https://github.com/hypothesis/h/blob/9f9496198e8177653dba5781fcbefed57f878be8/h/buildext.py#L267)) and reformat the docstring to look a little more conventional, that would make me happy :).

h/script.py
```diff
@@ -98,15 +103,25 @@ def reindex(args):
...
-    """Generate an OAuth bearer token for the specified principal."""
+    """Generate an OAuth bearer token for the specified principal.
+
+This token is suitable for passing to the h API as the value of an
+X-Annotator-Auth-Token header.
+
+For example, to authorize yourself as user seanh to your local dev instance
+of h do:
+
+    hypothesis token --base 'http://localhost:5000' --sub 'acct:seanh@127.0.0.1' conf/development.ini
```
nickstenning: Also this should probably read "localhost" not "127.0.0.1".

PR #2358, opened 2015-07-13
Add admin users

seanh:


h/accounts/views.py
```diff
@@ -574,6 +575,45 @@ def _update_subscription_data(request, subscription):
     request.session.flash(_('Changes saved!'), 'success')
 
 
+@view_config(attr='index', route_name='admin_users_index',
```
tilgovi: If there's no real common code for these, you're welcome to decide that you think simple functions are easier and clearer than a class. It saves having to dereference the request from `self`.

tilgovi:
I think the delete route is superfluous because you can use the button name an value to disambiguate within a single `POST` hanlder.

h/accounts/models.py
```diff
@@ -77,6 +77,8 @@ class User(Base):
                       server_default=sa.sql.expression.false(),
                       nullable=False)
 
+    admin = sa.Column(sa.BOOLEAN, default=False, nullable=False)
```
nickstenning: Can we get a `server_default` here?

h/accounts/views.py
seanh: Yeah I don't like the classes, but was just following the style of the rest of the views in this file. It does kind of make sense to group related views together in a file that has many

seanh:
TODO:

- [ ] Stop users from removing themselves from admins
- [ ] Nicer error when the user you try to add doesn't exist
- [ ] CLI for making admin users (needed to make the first admin user)
- [ ] Tests

seanh:
Do we really want admins to be able to do everything? I think the current implementation might let admins delete and edit other users' annotations and stuff (not tested). They don't necessarily need that power, we could just give them access to the admin and nipsa views specifically

tilgovi:
I don't think admins have any special permissions on annotations right now. See `h.models#Annotation.__acl__`.

h/accounts/models.py
nickstenning: To be clear, this *has* to have a `server_default` if it's not nullable, otherwise we'll break user registration in prod as soon as we run the migration for stage (they use the same database).

h/accounts/models.py
```diff
@@ -77,6 +78,10 @@ class User(Base):
                       server_default=sa.sql.expression.false(),
                       nullable=False)
 
+    admin = sa.Column(
```
nickstenning: I'm a pedant, I know, but could you format this in the same way as the others?

seanh:
> I don't think admins have any special permissions on annotations right now. See h.models#Annotation.__acl__.

I see. I guess the ACL on the annotation (lower down in the tree) overrides the one on the root node, not the other way round.

So the current implementation gives admins permission to do anything that there isn't a more-specific ACL stopping them from doing. That's probably fine, although we could tighten it up to just the admin and nipsa views specifically if we wanted.

seanh:
Notes:

* Adds an `admin` column to the `user` table (and a `.admins()` classmethod on the `User` class for convenience)

* `make_admin()` is implemented in `accounts/__init__.py`. This is called in two places (by the view code and the CLI code) so I wanted it shared and wanted the CLI code to just import h.accounts not h.accounts.something.

* I don't like the amount of patching necessary in `views_test.py`, but with

seanh:
@nickstenning @tilgovi Waiting for a landscape comment, but this can be reviewed now and then I'll rebase it after

Notes:

* Adds an `admin` column to the `user` table (and a `.admins()` classmethod on the `User` class for convenience)

* Simple HTML page for listing, adding and removing admins. Doesn't let you remove the last admin. Flash error if trying to add a user that doesn't exist. Uses usernames like "seanh" not "acct:seanh@hypothes.is". 

* Simple `hypothesis admin bob` CLI for adding admins as well. This is what we'll use to add the first admin (I'm assuming we can do this on our prod deployment).

* `effective_principals()` in `h/auth.py` accesses the accounts model to return `"group:admins"` if the user is an admin. If we want to do something more complicated (communicate this information via OAuth scopes or something), can we do that later?

* `"group:admins"` gets all permissions in the ACLs of both root resources (app and api). It was already there in the API one, and that's where the choice of `"group:admins"` as the string comes from (I think it might originate from Pyramid docs).

* So you just add an arbitrary permission that is never used as a principal to a view function, like `@view_config(..., permission='nipsa_user_index')` and this'll mean that only admins can use the view (and we could give that specific permission to specific other users later if we wanted). This is what's done for the admin management views in this PR. There's no end-to-end test with a test view like this and checking that Pyramid allows and denies access. Maybe there should be?

* `make_admin()` is implemented in `accounts/__init__.py`. This is called in two places (by the view code and the CLI code) so I wanted it shared and wanted the CLI code to just import h.accounts not h.accounts.something.

* I don't like the amount of patching necessary in `views_test.py`, but without breaking from our coding style for view functions and making these dependencies explicit and easily injectable I don't think it can be avoided.

* There's a simple db migration but (following our existing migrations) no tests for it (I have tested it manually but encourage you to do so as well, I have not tested it with postgres)

* New `User` factory for tests

nickstenning:
This looks pretty good. Thoughts:

- Consolidate the views into one permission, 'admin', for now. Having separate permissions for the three routes doesn't make a lot of sense to me -- why would I have permission to see a page that allowed me to remove admin users, but not to actually remove them?
- It would be nice to keep `script.py` mostly alphabetised.

I've tested the migration with postgres and it works fine.

If you make those changes and rebase this to clean it up, this'll go right in. Thanks @seanh!

seanh:
@nickstenning Done

nickstenning:
LGTM.

PR #2359, opened 2015-07-13
Four space js

tilgovi:
Standardize on four spaces for js files

### DIFF
--- a/h/browser/chrome/help/index.js
+++ b/h/browser/chrome/help/index.js
@@ -1,9 +1,9 @@
 // Detect the current OS and show approprite help.
 chrome.runtime.getPlatformInfo(function (info) {
-  var opts = document.querySelectorAll('[data-extension-path]');
-  [].forEach.call(opts, function (opt) {
-    if (opt.dataset.extensionPath !== info.os) {
-      opt.hidden = true;
-    }
-  });
+    var opts = document.querySelectorAll('[data-extension-path]');
+    [].forEach.call(opts, function (opt) {
+        if (opt.dataset.extensionPath !== info.os) {
+            opt.hidden = true;
+        }
+    });
 });
--- a/h/browser/chrome/lib/browser-action.js
+++ b/h/browser/chrome/lib/browser-action.js
@@ -1,62 +1,62 @@
 (function (h) {
-  'use strict';
+    'use strict';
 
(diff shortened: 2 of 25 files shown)


BigBlueHat:
All personal preferences aside, our style guide (or the one we point to) prescribes using 2 spaces...not 4:
https://google-styleguide.googlecode.com/svn/trunk/htmlcssguide.xml?showone=Indentation

BigBlueHat:
Actually...ignore that....it's for HTML/CSS...not JS.

The JS one apparently doesn't prescribe one.

tilgovi:
I'll update the style guide but let's keep this change set whitespace only. See any issues with it?

BigBlueHat:
I guess I don't completely understand why we're breaking rank with the style guide we've chosen. The "winning style" is apparently 2 spaces: https://github.com/Seravo/js-winning-style#user-content-indentation
> Two spaces, not longer and no tabs: Google, npm, Node.js, Idiomatic

Given those other style guides in the mix, I'd expect most JS developers do 2 spaces.

You've mentioned 4 spaces helps prevent (by force) too many callbacks and a preference for promises, function naming, etc.

Believe it or not this isn't meant to be a :bike: :house: (sorry...no :shed: emoji :stuck_out_tongue:), but rather an ask that we keep the barrier to entry on the project as low as possible. Prevailing winds and all that. :smile: 

:thought_balloon:'s?

tilgovi:
It was @nickstenning who had the preference for four IIRC.

nickstenning:
This isn't about changing anything, it's about standardising on one thing or another. Annotator is 4-space JS and will remain that way. Some of our code is 2-space, some 4.

I don't feel so strongly about this that I want to step on Benjamin's toes, but I do prefer 4-space.

What I prefer most of all is being consistent and clear.

tilgovi:
I prefer two for the reasons Benjamin listed, but I care about consistency more than anything.

I'm happy to close this, but if I do it'll mean I open one to make the stuff that is four already (the filters) into 2.

BigBlueHat:
2 has my vote for the reasons mentioned.

Once we pick one, I'm happy to do (or complete) the clean up.

Your call(s). :smile:

nickstenning:
Let's go with 2.

PR #2360, opened 2015-07-13
Add .editorconfig for (optional) code assist

BigBlueHat:
http://editorconfig.org/

**If** a developer is using EditorConfig, then all their future edits should be more inline with our style guides--assuming I mapped that all correctly. :grin: 

@tilgovi thoughts?

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,23 @@
+root = true
+
+[*]
+end_of_line = lf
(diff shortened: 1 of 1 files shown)


tilgovi:
Great. Can we strip out all the comments and the leading whitespace at the top? None of the comments really add anything.

tilgovi:
I did this.

PR #2361, opened 2015-07-14
Prevent annotation attempts when not signed in

BigBlueHat:
Simply hides the annotation form--which implied you could
use the thing...--and instead tells the user to sign in,
so they can do what they came to do.

Works like this: :movie_camera: https://youtu.be/KC4NmHZJea8

Obviously room for improvement, but so much better than what we've been doing.

Cheers!

### DIFF
--- a/h/templates/client/annotation.html
+++ b/h/templates/client/annotation.html
@@ -1,178 +1,183 @@
-<header class="annotation-header">
-  <!-- Deletion notice -->
-  <span ng-if="!vm.editing && vm.annotation.deleted">Annotation deleted.</span>
-
-  <!-- User -->
-  <span ng-if="vm.annotation.user">
-    <a class="annotation-user"
-       target="_blank"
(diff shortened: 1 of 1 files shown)


judell:
Thanks Benjamin!

judell:
BTW I'm happy to walk you through the dokku procedure if you want to spin one up for this.

JakeHartnell:
This a great quick fix! LGTM :+1:

judell:
Thanks!

judell:
The behavior on stage doesn't match the screencast. What changed? Now the annotation appears, with an embedded "You must sign in" message, but the Sign In link isn't live until the annotation is canceled.

BigBlueHat:
@judell doesn't look like stage has this patch yet.

tilgovi:
Correct

judell:
Checks out on stage now, thx.

judell:
I switched browsers today while testing. Thanks to this, I avoided losing data.

PR #2362, opened 2015-07-14
Anchoring rewrite

tilgovi:


tilgovi:
Discuss.

dwhly:
AWESOME.
screen shot 2015-07-14 at 3 35 11 pm

h/static/scripts/annotator/anchoring/pdf.coffee
```diff
@@ -0,0 +1,224 @@
...
+# It encapsulates the core anchoring algorithm, using the selectors alone or
+# in combination to establish the best anchor within the document.
+#
+# :param Element root: The root element of the anchoring context.
+# :param Array selectors: The selectors to try.
+# :param Object options: Options to pass to the anchor implementations.
+# :return: A Promise that resolves to a Range on success.
+# :rtype: Promise
+####
+exports.anchor = (root, selectors, options = {}) ->
```
tilgovi: I think this needs some cleaning up. The placeholder stuff is confusing and could use comments. The `anchorByPosition` function could be broken out for readability and given parameters rather than closing over the local scope.

h/static/scripts/annotator/anchoring/types.coffee
```diff
@@ -0,0 +1,55 @@
...
+
+
+###*
+# class:: RangeAnchor(range)
+#
+# This anchor type represents a DOM Range.
+#
+# :param Range range: A range describing the anchor.
+###
+class RangeAnchor
```
tilgovi: This class isn't broken out into a separate library. I'm actually not sure how useful it is in practice (the position anchors perform very well) and it doesn't map straightforwardly to any Web Annotations Data Model selector types, so I'm holding it here for the moment.

h/static/scripts/annotator/guest.coffee
```diff
@@ -1,7 +1,20 @@
-Promise = require('es6-promise').Promise
+raf = require('raf')
+Promise = global.Promise ? require('es6-promise').Promise
```
tilgovi: This pattern happens quite a bit. It's important because the es6-promise lib doesn't appear smart enough to delegate to the native implementation and using the polyfill has a negative performance impact due to setTimeout.

h/static/scripts/annotator/guest.coffee
```diff
@@ -10,21 +23,21 @@ module.exports = class Guest extends Annotator
...
-    DomTextMapper: {}
     TextSelection: {}
-    TextRange: {}
-    TextPosition: {}
-    TextQuote: {}
-    FuzzyTextAnchors: {}
-    FragmentSelector: {}
+
+  # Anchoring module
+  anchoring: require('./anchoring/html')
```
tilgovi: I made this a first-class citizen of the the `Annotator` instance rather than something for the options hash. It feels right. I don't want the options to grow forever.

h/static/scripts/annotator/guest.coffee
```diff
@@ -68,7 +70,6 @@ module.exports = class Guest extends Annotator
             this.publish(event, args)
       formatter: (annotation) =>
         formatted = {}
-        formatted.uri = @getHref()
```
tilgovi: This is orthogonal to these changes. I can restore this if you all want. On the backend, we're indexing this as an alias for `target.source` anyway.

h/static/scripts/annotator/guest.coffee
```diff
@@ -85,100 +86,49 @@ module.exports = class Guest extends Annotator
...
-
-  # Utility function to filter metadata and de-hash the URIs
-  getMetadata: =>
-    metadata = @plugins.Document?.metadata
-    metadata.link?.forEach (link) => link.href = @_removeHash link.href
-    metadata
+      uriPromise = Promise.reject()
+      metadataPromise = Promise.reject()
+
+    uriPromise = uriPromise.catch(-> decodeURIComponent(window.location.href))
```
tilgovi: Fallsbacks. There's a sensible result even without any plugin support for document metadata.

h/static/scripts/annotator/guest.coffee
```diff
@@ -211,21 +161,131 @@ module.exports = class Guest extends Annotator
 
     this.removeEvents()
 
-  setupAnnotation: ->
-    annotation = super
-    this.plugins.CrossFrame.sync([annotation])
-    annotation
+  setupAnnotation: (annotation) ->
```
tilgovi: Help me out and let me know what needs commenting here.

h/static/scripts/annotator/guest.coffee
```diff
@@ -287,40 +329,35 @@ module.exports = class Guest extends Annotator
...
-      event.stopPropagation()
-      this.selectAnnotations (event.data.getAnnotations event),
-        (event.metaKey or event.ctrlKey)
+  onHighlightMouseover: (event) ->
+    return unless @visibleHighlights
+    annotation = $(event.currentTarget).data('annotation')
+    annotations = event.annotations ?= []
+    annotations.push(annotation)
+    if event.target is event.currentTarget
+      setTimeout => this.focusAnnotations(annotations)
```
tilgovi: These might need a comment. I'm letting these events bubble now and passing along an `Array` as `event.annotations`. The innermost handler fires off an asynchronous task here. By the time it runs, the `Array` is filled.

In some sense I find this cleaner than walking up the tree. It doesn't stop the event from bubbling, which is always better if possible (less interference with the page).

h/static/scripts/annotator/host.coffee
```diff
@@ -39,8 +39,6 @@ module.exports = class Host extends Guest
 
     # Host frame dictates the toolbar options.
     this.on 'panelReady', =>
-      this.anchoring._scan() # Scan the document
```
tilgovi: :pizza:

h/static/scripts/annotator/main.js
```diff
@@ -13,54 +21,31 @@ Annotator.Plugin.CrossFrame.Bridge = require('../bridge')
...
 
-// Simple IE autodetect function
-// See for example https://stackoverflow.com/questions/19999388/jquery-check-if-user-is-using-ie/21712356#21712356
-var ua = window.navigator.userAgent;
-if ((ua.indexOf("MSIE ") > 0) ||     // for IE <=10
-    (ua.indexOf('Trident/') > 0) ||  // for IE 11
-    (ua.indexOf('Edge/') > 0)) {     // for IE 12
-  options["DomTextMapper"] = {"skip": true}
+// Document metadata plugins
+if (window.PDFViewerApplication) {
```
tilgovi: As much as possible I was trying to pull the document type switches out of the `Guest` and do it in one place, here. Probably, we can finish that by making `documentInfo` or something a first class citizen just as I did with the `anchoring` property.

h/static/scripts/annotator/main.js
```diff
@@ -13,54 +21,31 @@ Annotator.Plugin.CrossFrame.Bridge = require('../bridge')
...
 var options = {
   app: jQuery('link[type="application/annotator+html"]').attr('href'),
-  BucketBar: {container: '.annotator-frame'},
+  BucketBar: {container: '.annotator-frame', scrollables: ['body']},
   Toolbar: {container: '.annotator-frame'}
 };
 
-// Simple IE autodetect function
-// See for example https://stackoverflow.com/questions/19999388/jquery-check-if-user-is-using-ie/21712356#21712356
-var ua = window.navigator.userAgent;
```
tilgovi: Fuzzy anchoring should work on IE now!

h/static/scripts/annotator/monkey.coffee
```diff
@@ -1,153 +0,0 @@
-Annotator = require('annotator')
```
tilgovi: :monkey: :gun:

h/static/scripts/annotator/plugin/pdf.coffee
```diff
@@ -1,302 +1,113 @@
...
-    @_viewer.pages[index].textLayer.textLayerDiv
-
-  constructor: ->
-    # Set references to objects that moved around in different versions
-    # of PDF.js, and define a few methods accordingly
-    if PDFViewerApplication?
-      @_app = PDFViewerApplication
-      @_viewer = @_app.pdfViewer
+  pluginInit: ->
+    cache = {
```
tilgovi: This is less than ideal. The cache is stored here because I didn't want anything stateful in `anchoring/pdf`.

tilgovi: It's less than ideal in the sense that it splits responsibility across two places.

h/static/scripts/annotator/plugin/pdf.coffee
```diff
@@ -1,302 +1,113 @@
...
+      anchor: (root, selectors, options = {}) ->
+        options = extend({}, options, {cache})
+        return anchoring.anchor(root, selectors, options)
+      describe: (root, range, options = {}) ->
+        options = extend({}, options, {cache})
+        return anchoring.describe(root, range, options)
+    }
+
+    @pdfViewer = PDFViewerApplication.pdfViewer
+    @pdfViewer.viewer.classList.add('has-transparent-text-layer')
```
tilgovi: I'm not even sure we still support versions old enough not to have a transparent text layer. We may just want to change our CSS.

Also, we can look into using the annotation layer that PDF.js has, which would make a hell of a lot more sense.

tilgovi: But that would mean a separate highlighter implementation for PDF.js.

h/static/scripts/annotator/plugin/pdf.coffee
```diff
@@ -1,302 +1,113 @@
...
-  _finishScan: =>
-    # Do some besic calculations with the content
-    @_onHavePageContents()
+    for page in pdfViewer.pages when page.textLayer?.renderingDone
+      div = page.div ? page.el
+      placeholder = div.getElementsByClassName('annotator-placeholder')[0]
 
-    # OK, we are ready to rock.
-    @_pendingScanResolve()
+      switch page.renderingState
```
tilgovi: This probably needs explanations. It's interacting with the `anchoring/pdf` code. This code here is ensuring that when pages render/unrender that we're making sure that's detectable later by re-initializing the `textLayer` variable and yanking out the placeholder div that holds the highlight spans for un-rendered pages. Pulling the highlights out causes the annotations to re-anchor below.

h/static/scripts/annotator/plugin/pdf.coffee
```diff
@@ -1,302 +1,113 @@
...
+        when RenderingStates.INITIAL
+          page.textLayer = null
+        when RenderingStates.FINISHED
+          if placeholder?
+            placeholder.parentNode.removeChild(placeholder)
 
-    # Do whatever we need to do after scanning
-    @_onAfterScan()
+    for anchor in annotator.anchors when anchor.highlights?
+      if anchor.annotation in refreshAnnotations
```
tilgovi: We might consider having client-side only unique ids for annotation objects. Then we could do this sort of thing with a hash.

h/static/scripts/annotator/plugin/pdf.coffee
```diff
@@ -1,302 +1,113 @@
...
-    new Promise (resolve, reject) =>
-      if @anchoring.document.waitForInit?
-        @anchoring.document.waitForInit().then =>
-          try
-            resolve @_metadata()
-          catch error
-            reject "Internal error"
-      else
-        reject "Not a PDF dom mapper."
+    annotator.plugins.BucketBar?.update()
```
tilgovi: This can maybe be removed since `setupAnnotation` should take care of it.

tilgovi: Oh, but maybe I wanted the removed highlights to disappear from the bucket bar instantly.

h/static/scripts/annotator/plugin/textselection.coffee
```diff
@@ -21,80 +21,23 @@ class Annotator.Plugin.TextSelection extends Annotator.Plugin
...
-      selection.addRange(range.toRange()) if range
-      range
-
   # This is called when the mouse is released.
-  # Checks to see if a selection has been made on mouseup and if so,
+  # Checks to see if a selection  been made on mouseup and if so,
   # calls Annotator's onSuccessfulSelection method.
   #
   # event - The event triggered this. Usually it's a mouseup Event,
-  #         but that's not necessary. The coordinates will be used,
```
tilgovi: This comment was simply out of date.

h/static/scripts/cross-frame.coffee
```diff
@@ -20,7 +20,10 @@ module.exports = class CrossFrame
       new Discovery($window, options)
 
     createAnnotationSync = ->
-      whitelist = ['$highlight', '$orphan', 'target', 'document', 'uri']
+      whitelist = [
+        '$anchors', '$highlight', '$orphan',
+        'target', 'document', 'uri'
```
tilgovi: I guess if I removed `uri` in the other place I should remove it here.

h/static/scripts/vendor/diff_match_patch_uncompressed.js
```diff
@@ -1,2200 +0,0 @@
-/**
```
tilgovi: Yay! No more patched vendor libs. We'll have to see about restoring show differences without this.

package.json
```diff
@@ -85,9 +91,21 @@
       ]
     },
     "angular-mock": "global:angular.mock",
-    "diff-match-patch": "diff_match_patch",
-    "dom-text-mapper": "DomTextMapper",
-    "dom-text-matcher": "DomTextMatcher",
+    "dom-anchor-fragment": "domAnchorFragment",
```
tilgovi: Gross hackery. These are all UMD modules and we need to make sure they don't get tricked on pages that use AMD.

tilgovi: By that I mean, I made sure that they don't, but it's only through all this nasty stuff.

package.json
```diff
@@ -85,9 +91,21 @@
...
+    "dom-anchor-text-position": {
+      "depends": [
+        "dom-seek:seek"
+      ],
+      "exports": "domAnchorTextPosition"
```
tilgovi: I find this awkward. I should maybe rename the files. Even if the packages and repos are `dom-anchor-foo` maybe if I make the files `FooAnchor.js` then babel will make more sensible UMD wrappers that export `FooAnchor` to the global scope instead of `domAnchorFoo`.

tilgovi:
@dwhly yup!!!

It's a little misleading. 2200 lines just moved from a bundled diff-match-patch to one pulled in from npm.
Probably another 600 just moved out to the separate anchoring libraries.

Still, it's probably a net -2300 or so.

dwhly:
2300 hi5s

h/static/scripts/annotator/anchoring/html.coffee
```diff
@@ -0,0 +1,98 @@
...
+      resolve(range)
+    catch error
+      reject(error)
+  return new Promise(doQuery)
+
+
+###*
+# Anchor a set of selectors.
+#
+# This function converts a set of selectors into a document range using.
```
nickstenning: Minor comment blip here. Might want to complete the sentence!

h/static/scripts/annotator/anchoring/types.coffee
```diff
@@ -0,0 +1,55 @@
+Annotator = require('annotator')
+$ = Annotator.$
```
nickstenning: Unused AFAICT.

h/static/scripts/annotator/anchoring/html.coffee
```diff
@@ -0,0 +1,98 @@
...
+  types = [FragmentAnchor, RangeAnchor, TextPositionAnchor, TextQuoteAnchor]
+
+  selectors = for type in types
+    try
+      anchor = type.fromRange(root, range, options)
+      selector = anchor.toSelector(options)
+    catch
+      continue
+
+  return selectors
```
nickstenning: Just want to say awesome this file is. In particular, I really like how you've simplified the thing we discussed many times -- dependence of the anchor types on information from other selectors -- into optional additional arguments to `toRange/toSelector`.

h/static/scripts/annotator/anchoring/pdf.coffee
```diff
@@ -0,0 +1,224 @@
...
+      total += textContent.length
+      return getPageTextContent(index, cache).then(count)
+
+  return getPageTextContent(0, cache).then(count)
+
+
+###*
+# Anchor a set of selectors.
+#
+# This function converts a set of selectors into a document range using.
```
nickstenning: Ditto here with the slightly truncated comment.

h/static/scripts/annotator/anchoring/html.coffee
```diff
@@ -0,0 +1,98 @@
...
+
+  # Assert the quote matches the stored quote, if applicable
+  maybeAssertQuote = (range) ->
+    if quote?.exact? and range.toString() != quote.exact
+      throw new Error('quote mismatch')
+    else
+      return range
+
+  # Until we successfully anchor, we fail.
+  promise = Promise.reject('unable to anchor')
```
nickstenning: When reading this pattern:

```coffeescript
promise = Promise.reject('boom')

if foo?
  promise = promise.catch ->
    return myNewPromise()

if bar?
  promise = promise.catch ->
    return myOtherPromise()

return promise
```

My first thought was: "would this be simpler and clearer as..."

```coffeescript
promise = null

if foo?
  promise = myNewPromise()

if bar?
  promise = myOtherPromise()

if not promise?
  return Promise.reject('boom')
return promise
```

Of course the answer is "no", because the entire point is that you're chaining up the anchoring strategies so that if one of them fails to anchor the next one is tried.

I think adding a comment to this effect here, and pointing out that the chain goes, top to bottom, from "fast but brittle" to "slow but robust", would be useful.

h/static/scripts/annotator/anchoring/pdf.coffee
```diff
@@ -0,0 +1,224 @@
...
+  if cache[pageIndex]?
+    return Promise.resolve(cache[pageIndex])
+  else
+    joinItems = ({items}) ->
+      nonEmpty = (item.str for item in items when /\S/.test(item.str))
+      textContent = nonEmpty.join('')
+      cache[pageIndex] = textContent
+      return textContent
+
+    return PDFViewerApplication.pdfViewer.getPageTextContent(pageIndex)
```
nickstenning: It might be worth noting here that `pdfViewer.getPageTextContent` isn't obviously public API for PDF.js.

h/static/scripts/annotator/anchoring/pdf.coffee
```diff
@@ -0,0 +1,224 @@
...
+getPage = (pageIndex) ->
+  return PDFViewerApplication.pdfViewer.pages[pageIndex]
+
+
+getPageTextContent = (pageIndex, cache) ->
+  if cache[pageIndex]?
+    return Promise.resolve(cache[pageIndex])
+  else
+    joinItems = ({items}) ->
+      nonEmpty = (item.str for item in items when /\S/.test(item.str))
```
nickstenning: If you know more about the rationale behind doing this, that would be another candidate for a comment, IMO. It's not immediately clear to me when reading this code why we're doing this normalisation/sanitisation step, and it's not clear that it wouldn't potentially break if we got back something like

    <TextItem str="hello"><TextItem str=" "><TextItem str="world">

which would presumably be completely valid from PDF.js's point of view? (I don't know, which is why I'm asking...)

h/static/scripts/annotator/guest.coffee
nickstenning: Seems fine to me. It doesn't make sense for us to keep storing this field when the rest of the model explicitly says we have multiple targets.

nickstenning: Although, actually -- does the currently client use this field at all? Could removing it cause problems where annotations made with the new client don't work in older deployed versions?

h/static/scripts/annotator/plugin/pdf.coffee
nickstenning: I think it would be better -- certainly clearer -- to keep this entirely encapsulated in `anchoring/pdf`, and possibly expose a `reset` or `clearCache` function in the public interface of that module.

h/static/scripts/annotator/anchoring/pdf.coffee
nickstenning: Agreed. I think both `anchorByPosition` and `findInPages` probably need extracting. I think a lot of the utility functions that essentially abstract over the PDF.js interface could do with being in their own module with collaboration tests for how they exercise PDF.js.

nickstenning: And `anchorByPosition` itself could probably be reimplemented on top of a `isPageRendered` function in such a module.

h/static/scripts/annotator/anchoring/pdf.coffee
```diff
@@ -0,0 +1,224 @@
...
+exports.describe = (root, range, options = {}) ->
+  cache = options.cache ? {}
+  range = new xpathRange.BrowserRange(range).normalize()
+
+  startTextLayer = getNodeTextLayer(range.start)
+  endTextLayer = getNodeTextLayer(range.end)
+
+  # XXX: range covers only one page
+  if startTextLayer isnt endTextLayer
+    throw new Error('selecting across page breaks is not supported')
```
nickstenning: Is this a regression? How is this error relayed to the user?

h/static/scripts/annotator/anchoring/pdf.coffee
```diff
@@ -0,0 +1,224 @@
...
+          return findInPages(pages)
+      else
+        findInPages(pages)
+
+  return promise
+
+
+exports.describe = (root, range, options = {}) ->
+  cache = options.cache ? {}
+  range = new xpathRange.BrowserRange(range).normalize()
```
nickstenning: Do we even need to use `xpathRange` here? Given that a native browser range has `startContainer` and `endContainer`, and each of these are guaranteed not to span a text layer boundary, won't that do?

h/static/scripts/annotator/anchoring/types.coffee
```diff
@@ -0,0 +1,55 @@
...
+    @root = root
+    @range = xpathRange.sniff(range).normalize(@root)
+
+  @fromRange: (root, range) ->
+    return new RangeAnchor(root, range)
+
+  # Create and anchor using the saved Range selector.
+  @fromSelector: (root, selector) ->
+    data = {
+      start: selector.startContainer
```
nickstenning: Huh, you learn something every day. I didn't realise we'd diverged from the underlying implementation by renaming `{start, end}` to `{startContainer, endContainer}`.

h/static/scripts/annotator/anchoring/types.coffee
nickstenning: Let's leave it in for now but I'm totally up for removing it later. Certainly if it's not useful above and beyond the position selectors, then given that it's the cause of some of Annotator's more idiosyncratic (read: annoying) behaviour, such as splitting textNodes in the DOM when you make selections, I'd be all for removing it once we've checked the database to ensure it's not needed.

h/static/scripts/annotator/guest.coffee
nickstenning: Not sure I understand this. Why does using the polyfill have a negative performance impact, when [it's a no-op in environments that already define `Promise`](https://github.com/jakearchibald/es6-promise/blob/1e88f88bab1626911432a507613d63107f252e63/lib/es6-promise/polyfill.js#L21)?

h/static/scripts/annotator/anchoring/types.coffee
tilgovi: Yeah. That was a totally unnecessary thing I did once upon a time.

h/static/scripts/annotator/plugin/pdf.coffee
```diff
@@ -1,302 +1,113 @@
...
-  # Install watchers for various events to detect page rendering/unrendering
-  setEvents: ->
-    # Detect page rendering
-    addEventListener "pagerender", (evt) =>
+      @documentLoaded = Promise.resolve()
 
-      # If we have not yet finished the initial scanning, then we are
-      # not interested.
-      return unless @pageInfo?
+    @observer = new MutationObserver((mutations) => this.update())
```
nickstenning: :heart::green_heart::blue_heart:

h/static/scripts/annotator/guest.coffee
tilgovi: Oh, I totally understand now. I should be just getting `Promise` from the global and doing `require('es6-promise/polyfill')`. The entire problem is because I'm requiring the polyfill version directly. Thanks for catching that.

h/static/scripts/annotator/guest.coffee
```diff
@@ -35,21 +48,10 @@ module.exports = class Guest extends Annotator
...
-  constructor: (element, options, config = {}) ->
-    options.noScan = true
+  constructor: (element, options) ->
     super
-    delete @options.noScan
 
-    # Are going to be able to use the PDF plugin here?
-    if window.PDFTextMapper?.applicable()
-      # If we can, let's load the PDF plugin.
-      @options.PDF = {}
```
nickstenning: So what now actually causes the PDF plugin to be loaded on PDF documents? I can't atm see how/where this happens.

h/static/scripts/annotator/guest.coffee
nickstenning: Not sure we want/need to do a `decodeURIComponent` here in the client, do we?

h/static/scripts/annotator/guest.coffee
tilgovi: main.js

h/static/scripts/annotator/guest.coffee
```diff
@@ -211,21 +161,131 @@ module.exports = class Guest extends Annotator
...
 
-  setupAnnotation: ->
-    annotation = super
-    this.plugins.CrossFrame.sync([annotation])
-    annotation
+  setupAnnotation: (annotation) ->
+    self = this
+    root = @element[0]
+
+    anchors = []
```
nickstenning: A brief comment here explaining what these three containers are for ("what is a dead highlight? what is the difference between an anchor and an anchoredTarget?") would be good.

h/static/scripts/annotator/guest.coffee
```diff
@@ -211,21 +161,131 @@ module.exports = class Guest extends Annotator
...
+    anchoredTargets = []
+    deadHighlights = []
+
+    annotation.target ?= []
+
+    locate = (target) ->
```
nickstenning: "Docstring" style comments on these three functions.

h/static/scripts/annotator/guest.coffee
```diff
@@ -211,21 +161,131 @@ module.exports = class Guest extends Annotator
...
+      annotation.$orphan = anchors.length > 0
+      for anchor in anchors
+        if anchor.range?
+          annotation.$orphan = false
+
+      self.anchors = self.anchors.concat(anchors)
+      self.plugins.BucketBar?.update()
+      self.plugins.CrossFrame?.sync([annotation])
+
+    for anchor in self.anchors.splice(0, self.anchors.length)
```
nickstenning: `self.anchors.splice(0)` does the same thing.

nickstenning: Huh, apparently I'm wrong. It does, but it's non-spec. Ignore me.

h/static/scripts/annotator/guest.coffee
```diff
@@ -211,21 +161,131 @@ module.exports = class Guest extends Annotator
...
-    this.plugins.CrossFrame.sync([annotation])
-    annotation
+    return this.createAnnotation({$highlight: true})
+
+  deleteAnnotation: (annotation) ->
+    anchors = []
+    targets = []
+    unhighlight = []
+
+    for anchor in @anchors
```
nickstenning: I think `Array.prototype.filter` is supported for IE9 and up, so it might be clearer to use that here.

h/static/scripts/annotator/guest.coffee
nickstenning: I think Array.prototype.filter is supported for IE9 and up, so it might be clearer to use that here.

h/static/scripts/annotator/guest.coffee
```diff
@@ -211,21 +161,131 @@ module.exports = class Guest extends Annotator
...
+          anchors.push(anchor)
+          anchoredTargets.push(anchor.target)
+        else if anchor.highlights?
+          deadHighlights.push(anchor.highlights)
+          delete anchor.highlights
+          delete anchor.range
+      else
+        self.anchors.push(anchor)
+
+    deadHighlights = Array::concat(deadHighlights...)
```
nickstenning: Not sure I understand what the purpose of this line is...

h/static/scripts/annotator/guest.coffee
```diff
@@ -211,21 +161,131 @@ module.exports = class Guest extends Annotator
...
+    raf -> highlighter.removeHighlights(deadHighlights)
+
+    for target in annotation.target when target not in anchoredTargets
+      anchor = locate(target).then(highlight)
```
nickstenning: Schweet.

h/static/scripts/annotator/guest.coffee
```diff
@@ -211,21 +161,131 @@ module.exports = class Guest extends Annotator
...
+
+    getSelectors = (range) ->
+      options = {
+        cache: self.anchoringCache
+        ignoreSelector: '[class^="annotator-"]'
+      }
+      return self.anchoring.describe(root, range, options)
+
+    setDocumentInfo = ({metadata, uri}) ->
+      annotation.uri = uri
```
nickstenning: So you've removed one place where we set `annotation.uri`. We should probably be consistent, no?

h/static/scripts/annotator/guest.coffee
```diff
@@ -211,21 +161,131 @@ module.exports = class Guest extends Annotator
...
+      if metadata?
+        annotation.document = metadata
+
+    setTargets = ([info, selectors]) ->
+      source = info.uri
+      annotation.target = ({source, selector} for selector in selectors)
+
+    info = this.getDocumentInfo().then(setDocumentInfo)
+    selectors = Promise.all(ranges.map(getSelectors))
+    targets = Promise.all([info, selectors]).then(setTargets)
```
nickstenning: Does this work? `info` is the return value of a `.then` which doesn't return anything... or rather returns two different things depending on whether there's any metadata. Either an explicit return in `setDocumentInfo` or simply doing the `.then(setDocumentInfo)` would make this clearer.

h/static/scripts/annotator/guest.coffee
```diff
@@ -211,21 +161,131 @@ module.exports = class Guest extends Annotator
...
+
+    targets.then(=> this.setupAnnotation(annotation))
```
nickstenning: You've got `self` in scope here. Might as well use it...

h/static/scripts/annotator/guest.coffee
nickstenning: Simple and elegant, but *definitely* needs a comment.

h/static/scripts/annotator/highlighter.coffee
```diff
@@ -0,0 +1,43 @@
...
+  hl = $("<span class='#{cssClass}'></span>")
+
+  # Ignore text nodes that contain only whitespace characters. This prevents
+  # spans being injected between elements that can only contain a restricted
+  # subset of nodes such as table rows and lists. This does mean that there
+  # may be the odd abandoned whitespace node in a paragraph that is skipped
+  # but better than breaking table layouts.
+
+  nodes = $(normedRange.textNodes()).filter((i) -> not white.test @nodeValue)
+  r = nodes.wrap(hl).parent().show().toArray()
```
nickstenning: An explicit return here would be clearer. And what is `show()` doing here?

h/static/scripts/cross-frame.coffee
nickstenning: Yep.

h/static/scripts/annotator/main.js
nickstenning: Amazing.

h/static/scripts/annotator/monkey.coffee
nickstenning: Poor monkey.

h/static/scripts/annotator/plugin/pdf.coffee
nickstenning: A project for another day.

h/static/scripts/annotator/plugin/pdf.coffee
```diff
@@ -1,302 +1,113 @@
...
 
-        # Do some post-processing
-        content = @_parseExtractedText rawContent
+      return {title, link}
 
-        # Save the extracted content to our page information registery
-        @pageInfo[pageIndex] =
-          index: pageIndex
-          content: content
+  update: ->
```
nickstenning: A comment here about what exactly is being updated and when this function is called would be great.

h/static/scripts/annotator/plugin/pdf.coffee
nickstenning: Yes please comment.

h/static/scripts/annotator/plugin/pdf.coffee
nickstenning: It would also prevent the `O(N)` scanning that we do for normal pages when removing highlights. Again, though, one for another day.

package.json
nickstenning: Not sure I understand. Aren't we requiring the unbundled versions of these? At which point it's all up to browserify?

nickstenning: Ah, wait, ok. No, we're not, because they're written in ES6. Hmm. Ok.

h/static/scripts/annotator/guest.coffee
nickstenning: Thanks.

nickstenning:
Okay, there's obviously a lot here and I couldn't honestly say I've gone over it all with a fine-tooth comb, but I have looked over most of it now.

Most of my questions are basic comprehension issues and/or suggestions for inline commentary.

In general, it looks great. I'll give the branch a proper end-to-end test tomorrow. I don't see anything massive blocking merge, but I would like to get some of the more obscure parts of the code documented.

The only other reservation I have is about testing some of the functionality in `html.coffee` and `pdf.coffee` -- particularly the latter. Ultimately I am prepared to hold my tongue and save this for another day, but if the abstractions over PDF.js could make it into another module with collaboration tests against the PDF.js interface, that would make me double-extra-plus-happy.

h/static/scripts/annotator/guest.coffee
tilgovi: It's actually easier than that. `require('es6-promise')` runs the polyfill automatically but exports the shim. The solution is just to `require('es6-polyfill')` but ignore its exports rather than using them.

h/static/scripts/annotator/guest.coffee
tilgovi: I don't think so. I'll check.

h/static/scripts/annotator/guest.coffee
tilgovi: I don't know. If a user types a URL with reserved characters into the location bar the browser will encode it and `window.location.href` returns the encoded version. I thought it made some sense to store the user-facing version. Willing to consider pros and cons.

tilgovi:
Handled everything except for the cleanup of `anchoring/pdf` closures. I'm leaving `.uri` (though it's set by `createAnnotation` rather than the formatter now) because it's a supported facet for the client-side search filtering. We can delete it when we clean up or remove that code.

nickstenning:
I'd still like to see `pdf.coffee` cleaned up and tested, but :tada: :balloon: :fireworks: :sparkles: on this work.

PR #2363, opened 2015-07-14
Make the toolbar's note icon "New Page Note" if no selection

judell:
Story: A  Hypothesis director was unaware that the toolbar's note button is the way to do page-level tagging. This aims to clarify that option.

### DIFF
--- a/h/static/scripts/annotator/guest.coffee
+++ b/h/static/scripts/annotator/guest.coffee
@@ -265,7 +265,7 @@ module.exports = class Guest extends Annotator
     return confirm "You have selected a very short piece of text: only " + length + " chars. Are you sure you want to highlight this?"
 
   onSuccessfulSelection: (event, immediate) ->
-    document.getElementsByClassName('h-icon-insert-comment')[0].title = "New Note"
+    document.getElementsByClassName('toolbar-insert-comment')[0].title = "New Note"
     unless event?
       throw "Called onSuccessfulSelection without an event!"
     unless event.segments?
@@ -287,7 +287,7 @@ module.exports = class Guest extends Annotator
     true
 
   onFailedSelection: (event) ->
-    document.getElementsByClassName('h-icon-insert-comment')[0].title = "New Note"
+    document.getElementsByClassName('toolbar-insert-comment')[0].title = "New Page Note"
     @adder.hide()
     @selectedTargets = []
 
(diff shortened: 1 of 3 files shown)


h/static/scripts/annotator/guest.coffee
```diff
@@ -265,6 +265,7 @@ module.exports = class Guest extends Annotator
     return confirm "You have selected a very short piece of text: only " + length + " chars. Are you sure you want to highlight this?"
 
   onSuccessfulSelection: (event, immediate) ->
+    document.getElementsByClassName('toolbar-insert-comment')[0].title = "New Note"
```
tilgovi: Even if it duplicates some code I might prefer to do this in `Toolbar`. Unsure. I don't really know how to make this less ugly without refactoring the toolbar a bit. UI state is the worst.

h/static/scripts/directive/annotation.coffee
```diff
@@ -142,6 +142,7 @@ AnnotationController = [
     #              if they are open.
     ###
     this.view = ->
+      window.top.document.getElementsByName('insert-comment')[0].title = "New Page Note"
```
tilgovi: `toolbar-insert-comment`

h/static/scripts/directive/annotation.coffee
```diff
@@ -142,6 +142,7 @@ AnnotationController = [
     #              if they are open.
     ###
     this.view = ->
+      window.top.document.getElementsByName('toolbar-insert-comment')[0].title = "New Page Note"
```
nickstenning: `getElementsByName` isn't a thing.

nickstenning:
I don't think adding manipulation of the `title` property from arbitrary places in the code is the right way to do this. It would be very difficult for someone refactoring the toolbar code to find out that some code somewhere else is reaching in and fiddling with its state.

If we do this, it should an interface hook on the toolbar, which allows toggling the state.

judell:
"I don't think adding manipulation of the title property from arbitrary places in the code is the right way to do this" I didn't think so either, was hoping to a) find out what would be better, and b) how much effort better would be.

nickstenning:
I'm going to close this for now. If this is priority piece of work let's get it in the sprint and have a discussion there about how to fix it.

PR #2364, opened 2015-07-15
Misc styling improvements

JakeHartnell:
1.) Adjust vertical alignment of save an cancel icons.

Before:
selection_103

After:
selection_101

2.) Add extra space at the top of the "Sign in to create annotations" message.

Before:
selection_100

After:
selection_099

3.) Adjust font-sizes to be more visually appealing, and *slightly* reduce menu padding. Requested by @dwhly. : )

Before:
selection_105

After:
selection_104

### DIFF
--- a/h/static/styles/forms.scss
+++ b/h/static/styles/forms.scss
@@ -263,7 +263,7 @@
 // Positions the icon nicely within the button.
 .btn-icon {
   font-size: 1.4em;
-  vertical-align: top;
+  vertical-align: sub;
 }
 
 // Absolutely positions a message/icon to the left of a button.
--- a/h/templates/client/annotation.html
+++ b/h/templates/client/annotation.html
@@ -1,4 +1,4 @@
-<header class="annotation-header" ng-if="!vm.annotation.user">
+<header class="annotation-header" ng-if="!vm.annotation.user" style="margin-top:.75em">
   <strong>You must be signed in to create annotations.</strong>
 </header>
 
--- a/h/static/styles/common.scss
+++ b/h/static/styles/common.scss
@@ -188,7 +188,7 @@ html {
     a {
       display: block;
       line-height: 1;
-      padding: 1em;
+      padding: .8em;
       white-space: nowrap;
     }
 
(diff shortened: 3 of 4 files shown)


tilgovi:
Reduced font sizes? I think they should be bigger and the toolbar should be bigger.

h/templates/client/annotation.html
```diff
@@ -1,4 +1,4 @@
-<header class="annotation-header" ng-if="!vm.annotation.user">
+<header class="annotation-header" ng-if="!vm.annotation.user" style="margin-top:.75em">
```
tilgovi: Please, no inline styles.

h/static/styles/forms.scss
```diff
@@ -263,7 +263,7 @@
 // Positions the icon nicely within the button.
 .btn-icon {
   font-size: 1.4em;
-  vertical-align: top;
+  vertical-align: sub;
```
tilgovi: Is `sub` more appropriate than `middle`?

dwhly:
> Reduced font sizes? I think they should be bigger and the toolbar should be bigger.

I think this restores the font size to where we were before the change this last week (if not, it should-- that was the goal).  Moving forward I'd like to place these questions firmly in the camp of @conordelahunty and our new design function to reflect on and prioritize, particularly since these questions clearly evoke strong passions. i.e. How do we make these kinds of decisions, and on what basis?

tilgovi:
The font size did not change last week, so by your logic this is now in
Conor's court and I'll leave it to him to comment, merge, change, close,
etc.

dwhly:
This is the change between the last extension and this one:

screen shot 2015-07-15 at 9 27 24 am

tilgovi:
I promise you I did not change it, though I did make it explicit. The result likely depends on rounding error, DPI, browser and/or OS handling of fractional font sizes, etc. It was showing up at 14.something on my computer, and I made it an explicit 14.

Anyway.... Conor, yes?

conordelahunty:
I'd prefer if there were no further style changes until I've had time to do a proper audit and plan out the next iteration of the UI design.

JakeHartnell:
> I'd prefer if there were no further style changes until I've had time to do a proper audit and plan out the next iteration of the UI design.

Fair enough, closing for now.

PR #2373, opened 2015-07-19
Extract PDF anchoring closures

tilgovi:
Just pulling these functions out first.


tilgovi:
@nickstenning this should be a simple one, too

nickstenning:
LGTM. Would still be nice to see some tests for this code as and when you get a chance.

PR #2374, opened 2015-07-20
Straightforward API inclusion

tilgovi:
Clean up the boundaries between h.api and the rest of the application
so that it can be included in a straightforward manner.

The h.api module becomes more or less oblivious to authentication and
authorization, except that it still defines ACL properties for its
resources and it defines the token route.

The token route can be ripped out next and placed in another module,
since the h.api module needn't even know that OAuth tokens are in use
and in this way it can cease to rely on `request.create_token_response`.

This commit reintroduces pyramid_multiauth as a straightforward way to
support SessionAuthenticationPolicy and RemoteAuthenticationPolicy in
the same application.

The case of the default client in h.auth and the session view callable
are now oblivious to type of authentication policy in use.


nickstenning:
If we're using `pyramid_multiauth` in development and production, is there a reason we don't just add it to `setup.py` as an install requirement?

nickstenning:
You probably also need to update `script.py` -- the token command, specifically.

tilgovi:
Because it's not a code requirement anywhere, it's a deployment requirement
for our configuration.

The whole system would work equally well with just session authorization
and same origin annotating.

Or it could work with just a remote auth policy and middleware.

Or... you get the idea.

This is the reason I separated a create_app and includeme in app.py. The
former sets up the tweens and subscribers for our configuration, and this
could arguably be done in the .ini too, but I didn't tackle that for now.

But the purpose is to imagine that just as we have an h.api included in h,
so might h be included elsewhere, and we should not push unnecessary
dependencies on downstream consumers.

If you like, I'd be happy for you to put these in an extras, but I don't
have a problem leaving it like this for the moment.

nickstenning:
So currently `git clone ... && make dev` will fail, so we should add it somewhere that makes that not true!

h/subscribers.py
```diff
@@ -3,11 +3,8 @@
 
 def add_renderer_globals(event):
     request = event['request']
-    # Set the base url to use in the <base> tag
-    event['base_url'] = request.resource_url(request.root, '')
-    # Set the service url to use for API discovery
-    event['service_url'] = request.resource_url(request.root, 'api', '')
-    # Allow templates to check for feature flags
```
nickstenning: Why remove these comments?

tilgovi: They didn't add anything to me, but maybe i know too much?

The whole purpose of this event is to add globals for the renderer and the values themselves are pretty self describing. It didn't seem necessary to explain how the templates use these values. If one wonders, one can go see.

In particular, I didn't feel like it made sense for the comments to suggest only one particular use of each of the routes.

tilgovi:
Fixed token command. Updated dev extras to require `pyramid_multiauth`.

tilgovi:
Squashed. Rebased

conf/production.ini
```diff
@@ -22,9 +22,15 @@ mail.default_sender: "Annotation Daemon" <no-reply@localhost>
 #mail.host: localhost
 #mail.port: 25
 
+# Authentication configuration -- see the pyramid_multiauth documentation
+multiauth.policies: remote session
```
nickstenning: Presumably we need a `multiauth.groupfinder` setting here, too?

h/api/resources.py
```diff
@@ -43,7 +43,7 @@ class Root(Resource):
     ]
 
 
-def create_root(_):
+def create_root(request):
```
nickstenning: Seems a bit spurious to have this in this commit.

conf/production.ini
tilgovi: Yep. Good catch.

h/api/resources.py
tilgovi: I'll remove it, but I find it awkward not to name the variable when we're forced to receive it.

tilgovi:
Updated.

h/app.py
```diff
@@ -42,6 +22,28 @@ def create_app(global_config, **settings):
...
+    config.add_subscriber('h.subscribers.set_user_from_oauth',
+                          'pyramid.events.NewRequest')
+
+    config.add_tween('h.tweens.csrf_tween_factory')
+    config.add_tween('h.tweens.auth_token')
+
+    config.include(__name__)
+
+    app = config.make_wsgi_app()
+    app = permit_cors(app,
```
nickstenning: Let's not enable CORS for the entire application. That allows anyone to load any of our asset javascript cross-origin, which rather defeats the point of CORS in the first place. It shouldn't be too hard to add a PATH_INFO restriction to the `permit_cors` middleware.

tilgovi: I'd be happy to do that, but can we put it to a separate PR? It doesn't really defeat the point of CORS just because we use it liberally. We had the choice and I got to make it here, consciously. Our choice still doesn't include allowing credentials, which is the important vulnerability we'd like to avoid from cross origin requests. Allowing people to load our assets cross origin doesn't seem like a problem to me.

tilgovi: In fact our assets are explicitly listed as `Access-Control-Allow-Origin: *` independently of `permit_cors` usage.

tilgovi: See h/assets.py

nickstenning: Fair enough.

nickstenning:
LGTM.

PR #2377, opened 2015-07-20
Replace pyramid_layout with basic template inheritance

nickstenning:
This commit removes our use of `pyramid_layout` to assign assets for
particular rendered pages. Instead, we use the rather more
straightforward approach of rendering templates directly. These
templates can inherit from one another, and can render asset tags using
the integration between `webassets` and `jinja2`, configured in `h.app`.

This commit also renames all Jinja2 templates to have one extension,
".j2", as this simplifies the configuration of the Jinja2 renderer
considerably: we don't have to configure as many Jinja2 environments as
we have file extensions.


nickstenning:
Hold on one moment -- just spotted something that needs fixing.

nickstenning:
And that's now fixed.

h/app.py
```diff
@@ -54,10 +60,11 @@ def includeme(config):
     config.include('h.api_client')
 
     config.include('pyramid_jinja2')
-    config.add_jinja2_renderer('.js')
-    config.add_jinja2_renderer('.txt')
-    config.add_jinja2_renderer('.html')
-    config.add_jinja2_renderer('.xml')
+    config.add_jinja2_extension('h.jinja_extensions.IncludeRawExtension')
+    config.add_jinja2_extension('webassets.ext.jinja2.AssetsExtension')
```
tilgovi: Nice. I had forgotten this was there.

BigBlueHat:
So...why the move away from `pyramid_layout`? It seems this commit hard codes a bunch of stuff that was previously configurable.

It is to move the UX "thinking" out of Python-land and into the templates? If so, I'm generally in favor of that--especially if it gets us closer to a UI built from "static" markup.

Right now, though, I'm a bit concerned that we're giving up some flexibility...but maybe that's the point?
> Instead, we use the rather more straightforward approach of rendering templates directly.

tilgovi:
Super good.

tilgovi:
We don't really lose any flexibility that we were actually using. We can always override assets in Pyramid land so downstream consumers can replace templates and partials and whatnot. So it should be more accessible to most devs to override a jinja template than to override a layout as it requires less pyramid and python code.

tilgovi:
In any case, we'll see. We've never really been using pyramid_layout in a way that really leverages its purpose, so I'm happy to see it go. A lot of indirection and ambition about automatic asset handling, but it's not super clear it was helping with anything.

BigBlueHat:
Gotcha. Much thanks for the back story! :book:

PR #2379, opened 2015-07-20
Decaffeinate config/accounts.coffee & accounts/*.coffee

BigBlueHat:
This is the first of (hopefully) several PRs toward "decaffeinating" `h` (aka, getting off CoffeeScript).

The aim is to make `h` more accessible to more developers, lower the bar for UX changes, and generally help us find our way around our own code without having to reason about yet-another-build-system/code-style/etc.

If @hypothesis/developers could take a gander at this and comment on anything that needs tweaking, I'm happy to us that info to push ahead on the decaffeination. :heavy_minus_sign: :coffee: 

Thanks! :tea:


tilgovi:
Ugh. I went through and made it more idiomatic and sane.

@BigBlueHat if you want to drive this you're going to have to do do a bit of cleanup.

In this case, here are the things I did:

- Change function expressions into function statements so they have names in tracebacks.
- Clean up the IIFEs from the `do ->` syntax
- Get rid of variables-at-top style that we don't really follow anywhere else
- Preserve comments
- Remove the extra wrapper around CoffeeScript classes
- Whatever else you notice

I'm happy to accept PRs for these things, but they drive me mad like a scab I feel compelled to itch so please refrain unless you're going to go all the way or I'll go insane fixing them.

tilgovi:
And in this case I've now screwed up something in the auth...

tilgovi:
Alright, this one is good to go. :disappointed_relieved:

tilgovi:
Tests pass. Manual Q/A pass. Merging this thing.

BigBlueHat:
@tilgovi my apologies. I'm more than happy to handle all those scenarios (and more) and get things back toward "human" JS post decaf-ing. Thanks for that list. It's great to have key things to target and to know which things we don't care about (like var defs at the top). Thanks again.

vid:
Glad to see you're taking this step!

tilgovi:
Hope I wasn't too grumpy.

PR #2380, opened 2015-07-20
Add .jscrc for JavaScript Code Style help

BigBlueHat:
JSCS is like JSHint, but specifically for Code Style.

It comes with a Google Code Style (which we use currently)
pre-set. This config file sets that and can be extended
later for any departures we make (if any) from that style
guide.

This will be more valuable as we get off CoffeScript.

### DIFF
--- a/.jscsrc
+++ b/.jscsrc
@@ -1,0 +1,8 @@
+{
+  "preset": "google",
+  "excludeFiles": [
+    "node_modules/**",
+    "h/static/scripts/vendor"
+  ],
+  "maxErrors": 10
+}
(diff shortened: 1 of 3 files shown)


tilgovi:
Is JSHint made redundant by this?

BigBlueHat:
Don't think so. I found JSCS from the JSHint docs:
http://jshint.com/docs/options/

It seems JSHint is (now) focused on code quality (cyclomatic complexity, maxdepth, etc) vs. code style (indent, line length, etc). The JSHint docs say they're deprecating things like `indent` in favor of folks using JSCS.

I've got VIM setup to report from JSHint and it is finding things that the JSCS script isn't finding...and vice versa. Maybe someday they'll be "one thing" but right now, they're two...it seems.

tilgovi:
Okay.

tilgovi:
Thanks.

PR #2381, opened 2015-07-20
Stop two sources of dangling transactions

tilgovi:
Two sources of dangling transactions:

- WebSocket

Solved by releasing the transaction after checking features,
avoiding a transaction to get the user for every 'more_hits' call,
and the general hygiene of processing messages in a transaction.

- Notification worker

Process messages in a transaction.

### DIFF
--- a/h/streamer.py
+++ b/h/streamer.py
@@ -17,7 +17,6 @@ from jsonschema import validate
 from pyramid.config import aslist
 from pyramid.httpexceptions import HTTPBadRequest, HTTPForbidden
 from pyramid.threadlocal import get_current_request
-import transaction
 from ws4py.exc import HandshakeError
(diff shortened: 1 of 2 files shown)


h/streamer.py
```diff
@@ -533,10 +537,7 @@ def received_message(self, msg):
                 self.client_id = data.get('value')
         except:
             log.exception("Parsing filter: %s", msg)
-            transaction.abort()
```
nickstenning: This pathway previously aborted the transaction, and now doesn't. Reraise?

tilgovi:
@nickstenning good to go, thanks for the review

tilgovi:
There, that should get the tests passing.

tilgovi:
Should I do anything else here?

PR #2385, opened 2015-07-21
Reload annotations on auth change

tilgovi:
Just because the user has permission to read an annotation does not
imply that it should still be visible under the current search after
the authentication change. An example of this is when a NIPSA flagged
user logs out but has public annotations visible.

### DIFF
--- a/h/static/scripts/app-controller.coffee
+++ b/h/static/scripts/app-controller.coffee
@@ -62,12 +62,13 @@ module.exports = class AppController
       $scope.accountDialog.visible = false
 
     cleanupAnnotations = ->
-      # Clean up any annotations that need to be unloaded.
+      # Clean up all the annotations
       for id, container of $scope.threading.idTable when container.message
-        # Remove annotations the user is not authorized to view.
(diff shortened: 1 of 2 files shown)


tilgovi:
@seanh this should help you out a bit

nickstenning:
Holy crap. So when I said "surely we're not manually filtering out private annotations when auth scope changes"...

Anyway, it might be nice to take this opportunity to rename "cleanupAnnotations" to something a bit more expressive. And if we're removing everything is there a reason why we wouldn't just discard the entire threading table?

tilgovi:
Because we're not removing absolutely everything. We're keeping the drafts. We can choose to not do that, but that would be a regression vs current where highlights are immediately saved as soon as you log in (despite the change @BigBlueHat made making them say "You must sign in...").

tilgovi:
I can name it, export it, write a test, though.

nickstenning:
> I can name it, export it, write a test, though.

:+1:

tilgovi:
Now with test.

PR #2387, opened 2015-07-21
Getting picky with our code styles

BigBlueHat:
Doing this mostly to explore how close our code is to the Google code style we're using.

Most of this is JSCS driven. Some is from JSHint.

Let me know if you all think this is worth bothering with or whether you'd like to see these done "inline" with other changes.

### DIFF
--- a/.jscsrc
+++ b/.jscsrc
@@ -1,5 +1,7 @@
 {
   "preset": "google",
+  "requireSemicolons": true,
+
   "excludeFiles": [
     "node_modules/**",
     "h/static/scripts/vendor"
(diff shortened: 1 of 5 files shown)


nickstenning:
This is small enough to review by eye, so I'm happy to merge this, but in general I'd rather see incremental improvements to style consistency as we write code, rather than bulk updates (which have zero user value and carry risk).

If there's a way to do this, let's add style checking/auditing to our PR reviews if we can (much as we have for Python via @landscapeio, when it's working) so that we can improve it over time.

BigBlueHat:
:+1:

PR #2388, opened 2015-07-22
allow skinnier sidebar

judell:


tilgovi:
If you want this could you either fix the tests yourself or put it up as a sprint task for the next sprint, please?

Also, I'm not sure this is advisable. Might want to have @conordelahunty take a look. I picked a number based on what I observed to be the minimum size before the layout of annotation cards starts to break.

seanh:
Re-done with fixed tests https://github.com/hypothesis/h/pull/2509

PR #2394, opened 2015-07-22
Two front-end fixes

JakeHartnell:
#### Add extra space at the top of the "Sign in to create annotations" message.

Before:
selection_100

After:
selection_099

#### Hide tags when a thread is collapsed.
Tags weren't being collapsed and now everything collapses neatly to one line.

### DIFF
--- a/h/static/styles/annotations.scss
+++ b/h/static/styles/annotations.scss
@@ -184,6 +184,7 @@ privacy {
 
     .annotation-collapsed-replies {
       display: inline;
+      margin-left: .25em;
     }
   }
 }
--- a/h/static/styles/threads.scss
+++ b/h/static/styles/threads.scss
@@ -63,12 +63,9 @@ $thread-padding: 1em;
   }
 }
 
-.thread-message {
-  margin-bottom: .5em;
-}
-
+.thread-message,
 .thread-deleted {
-  margin: .8em 0;
+  margin: .5em 0;
 }
 
 .thread-load-more {
--- a/h/templates/client/annotation.html
+++ b/h/templates/client/annotation.html
@@ -103,7 +103,7 @@
     </tags-input>
   </div>
 
-  <div class="annotation-section tags tags-read-only"
+  <div class="annotation-body tags tags-read-only"
        ng-if="vm.annotation.tags.length && !vm.editing">
     <ul class="tag-list">
       <li class="tag-item" ng-repeat="tag in vm.annotation.tags">


h/static/styles/annotations.scss
```diff
@@ -178,12 +185,13 @@ privacy {
       margin: 0;
     }
 
-    .annotation-body, .annotation-footer {
+    .annotation-body, .annotation-footer, .tags-read-only {
```
tilgovi: Is this change necessary? I think the right change here is to make that an `annotation-body` instead of an `annotation-section`. I made the other tags section into "a body" recently, but I missed the read-only one.

h/static/styles/annotations.scss
```diff
@@ -178,12 +185,13 @@ privacy {
...
       display: none;
     }
 
     .annotation-collapsed-replies {
       display: inline;
+      margin-left: 3px;
```
tilgovi: Any reason this shouldn't be proportional, i.e., in ems?

h/static/styles/annotations.scss
```diff
@@ -34,7 +34,14 @@
   margin: .8em 0;
 }
 
-.annotation-header { margin-top: 0 }
+.annotation-header {
+  margin-top: 0;
+
+  &.please-sign-in {
+    margin-top: 0.75em;
```
tilgovi: drop the leading 0

tilgovi:
Okay, I need to take a deep look at this. The issue is that this `sub` shifts the icons in the top bar by a pixel.

While `sub` is more correct than `middle` for what we're using it for, I _think_, it's all sort of inappropriate and I think that comes back to some line height issues.

Our reset style sets `line-height` to be `normal` for buttons, causing it to use a browser default. The result is that the `1.4em` I chose for the button size doesn't end up making it as tall a single line block of text.

I'll play with it and decide what I think we should do.

tilgovi:
Or if you have ideas, throw em at me.

tilgovi:
If you could not bundle things together as "miscellaneous fixes" we could merge some of this, just not the button alignment.

JakeHartnell:
> If you could not bundle things together as "miscellaneous fixes" we could merge some of this, just not the button alignment.

I took out the button alignment.

h/static/styles/annotations.scss
```diff
@@ -34,7 +34,14 @@
   margin: .8em 0;
 }
 
-.annotation-header { margin-top: 0 }
+.annotation-header {
+  margin-top: 0;
+
+  &.please-sign-in {
```
tilgovi: Have you looked at removing or changing existing styles instead of adding this special case?

tilgovi: I have an elegant fix for this, but I'd like you to spend 30m to see if you can find it.

tilgovi: Any luck?

JakeHartnell: Sorry, I've been working on other projects the past 2 days. I can't read your mind but I took a stab at removing the extra class. Have a look!

tilgovi:
That's nicer. We can do even better, though, I think. The only change needed is to make the `.thread-message`top/bottom margin symmetrical. The adjacent margins collapse, so the result is that the top of the card has the same margin as the bottom.

PR #2399, opened 2015-07-23
Update Firefox extension for Jetpack Manager tool

tilgovi:


nickstenning:
It would have been nice to have a bit more context on what this is for! As it is I've worked out that `jpm` replaces `cfx`, and the changes look sensible, so...

PR #2402, opened 2015-07-24
Enable users to create groups

seanh:
@nickstenning suggests that (since it's behind a feature flag) this can be merged now even though the groups dropdown list in the sidebar is still a fake.

@tilgovi Want to review this? Since Nick and I have both worked on it.

What this does:

* Authorized users (anyone who's logged-in) can go to `/groups/new` to get a form for creating a new group
* The form has validation and re-rendering with inline errors and user data intact, if the group name is too short or too long
* On form submission group will be created, user is redirected to group's page including hashid as part of URL. Anyone (logged-in or not, member of group or not) can see this page if they have the link. No functionality on this page yet.
* In the database we save a bunch of generated stuff: created and updated times, creator, and make the creator the first member of the group


h/groups/views.py
```diff
@@ -0,0 +1,105 @@
...
+
+from h.groups import schemas
+from h.groups import models
+from h.accounts import models as accounts_models
+from h import security
+
+
+def _get_hashids(request):
+    salt = security.derive_key(
+        request.registry.settings["secret_key"], "h.groups.hashid", length=20)
```
nickstenning: As we discussed, I think this is a dangerous route to go down. Just add a config option for this.

setup.py
```diff
@@ -50,6 +51,7 @@ def run_tests(self):
     'gevent>=1.0.2,<1.1.0',
     'gnsq>=0.3.0,<0.4.0',
     'gunicorn>=19.2,<20',
+    'hashids>=1.1.0',
```
nickstenning: We should pin this to a specific version as recommended on hashids.org.

nickstenning: Maybe even add a comment explaining why (hashid format may change with updated hashid package, so always pin).

nickstenning:
So, just to be clear, this is butt ugly at the moment and definitely will need more work. My comment about the feature flags is merely observing that merging this doesn't block deployment, so if we're happy with what's here we can merge and carry on in additional PRs.

h/groups/models.py
```diff
@@ -0,0 +1,58 @@
...
+
+from h.db import Base
+from h.textutil import slugify
+
+
+class Group(Base):
+    __tablename__ = 'group'
+
+    id = sa.Column(sa.Integer, autoincrement=True, primary_key=True)
+    name = sa.Column(sa.Unicode(100), nullable=False)
```
tilgovi: How do we motivate a choice of group name size limit?

h/groups/models.py
```diff
@@ -0,0 +1,58 @@
...
+    def get_by_id(cls, id_):
+        try:
+            return cls.query.filter(
+                cls.id == id_).one()
+        except exc.NoResultFound:
+            return None
+
+
+user_group_table = sa.Table('user_group', Base.metadata,
+    sa.Column('user_id', sa.Integer, sa.ForeignKey('user.id')),
```
tilgovi: Should this be `nullable=False`?

h/groups/schemas.py
```diff
@@ -0,0 +1,10 @@
+# -*- coding: utf-8 -*-
+
+import colander
+
+from h.accounts.schemas import CSRFSchema
+
+
+class GroupSchema(CSRFSchema):
+    name = colander.SchemaNode(colander.String(),
+                               validator=colander.Length(min=4, max=100))
```
tilgovi: Again, I'd like to understand why we choose these minimums and maximums.

h/groups/views.py
```diff
@@ -0,0 +1,88 @@
...
+
+    group = models.Group.get_by_id(group_id)
+    if group is None:
+        raise exc.HTTPNotFound()
+
+    if slug is None or slug != group.slug:
+        canonical = request.route_url('group_read',
+                                      hashid=hashid,
+                                      slug=group.slug)
+        return exc.HTTPMovedPermanently(location=canonical)
```
tilgovi: What's the reasoning behind having the redirect?

h/textutil.py
```diff
@@ -0,0 +1,18 @@
+# -*- coding: utf-8 -*-
```
tilgovi: Maybe just use this? https://github.com/un33k/python-slugify

tilgovi:
Looks fine, just a few small questions.

h/groups/models.py
nickstenning: Honestly, I just figured that

    East Louisiana State Troopers Chinchilla Brevicaudata Appreciation Society - East Baton Rouge Branch

Was more than long enough for a group name.

h/groups/schemas.py
nickstenning: Minimum is so we have *something* reasonable for a slug. Maximum: see the chinchillas.

h/groups/models.py
nickstenning: Good call. There's no reason to allow NULLs here, so yes.

h/groups/views.py
nickstenning: Without the redirect, we have to choose between:

1. 404ing for requests with bad slugs
2. ignoring the slug entirely and serving the group page as identified solely by hashid

I don't particularly like 1) as it introduces more ways to break a group URL when copying and pasting it (e.g. in emails). I *really* don't like 2) as it has terrible search engine properties and encourages people to create nasty URLs for groups they don't like (i.e. taking `/groups/abc123/whale-lovers` and redistributing it as `/groups/abc123/zooplankton-haters`).

So, if we don't like either those options, the third option is to accept slug typos, but 301 them to the right place. This, incidentally, is the approach taken by Trello, among others. (e.g. `https://trello.com/b/o2ungX1n/current-springles` redirects to `https://trello.com/b/o2ungX1n/current-sprint`).

seanh: This is just to redirect `/groups/xKkqeD` (with or without trailing slash) to `/groups/xKkqeD/radiohead-fans`. I have to admit I'm not entirely sure what the use-case is (how would someone even have the link without slug, unless they typed it in manually?), but Trello does it!

It does also redirect `/groups/xKkqeD/the-wrong-slug` to `/groups/xKkqeD/radiohead-fans` which may be useful if the group's name and therefore its slug has changed (not that we support renaming groups yet but we probably will one day)

h/textutil.py
seanh: I think I prefer our own four line function over adding a lib

nickstenning: Good call. I evaluated a couple which all sucked for unicode, but I missed this one.

nickstenning: > I think I prefer our own four line function over adding a lib

Our own four line function depends on `unidecode` and `regex`...

seanh:
It might be nice to tweak some names to make them consistent. Our view callables are `create_group_form()`, `create_group()` and `read_group()`, yet the route names are `group_create` and `group_read1, and the templates are again `create_group` and `read_group`.

You could argue for verb first or noun first, or in the case of the callables just verb without noun (`create()`, `read()` etc) since they're already in a groups package.

We don't seem to be consistent about this in `views.py` files across the codebase.

h/groups/views.py
tilgovi: Okay. I would just think really hard about whether it should be a permanent or a temporary redirect. Can one change a group name? I'm still not even convinced of the idea that groups should "have" intrinsic names vs allowing people to rename their groups as they please.

tilgovi: Actually, that last part is _really_ important to me. Can we have a real discussion about intrinsic names before we merge this?

seanh: Intrinsic names?

My suggestion is this:

- Each group has a unique (auto-generated, not user-visible) id which is what we actually use to identify it

- We can encode the id to get the hashid which is what we show to the user in the URL, and decode the hashid to get the id which is how we know what group object a given http request is for

- The group's _name_ is for display purposes only, not for identification. Well, the _users_ may identify the group by seeing its friendly name, but the system always uses the id/hashid.

- Name is slugified into URL to make URL friendlier

- We should allow users to edit the names of groups after creation (not required for v 1.0)

This does seem to mean that these ought to be temporary redirects maybe, since the group name may change and then the URL we want to redirect to would change.

tilgovi: What I'm trying to say is that the group name could be on the join table, and users would be able to personalize the way they name their groups.

You could imagine a class where a teacher has three groups: Section 1, Section 2, and Section 3. The students, however, regardless of which they are in, might call it simply "Philosophy" or maybe another student wants to call it "Mrs. Hermann's Class".

Whether or not these are use cases we care about, or if anyone else thinks personal names for things is a desirable property of objects generally to the extent that I do, is not really what I'm trying to ask here.

It's more like this:

- I've noticed an assumption baked into this code that groups _have_ a (singular) name.

- Let me just check in with everybody about whether it was ever considered that maybe they don't, that there exist other ways to go about this.

tilgovi: And probably a big part of why I'm urgently flagging this is because I see a permanent redirect, and those are hard to change. So, I want to understand whether there's any reason this might change, and if we were to adopt such a multiple/personal naming model for groups we would obviously no longer have one single, correct URL that contains a slug for a group. My gut is to say just skip the slug altogether for now.

nickstenning: Okay, so let's change it to a temporary redirect for now. 

For what it's worth, I don't see the permanent redirect as being a massive problem either. No intermediaries (proxies) or crawlers that I know of *actually* store them permanently (we had this discussion with the Google crawler team when working on GOV.UK), and if the group name/slug were to change the worst that would happen is that they'd go direct to a page that would then issue them another 301 to the correct final URL.

Groups having multiple names -- one per user, is definitely a feature we didn't discuss as part of this card. It's certainly interesting, and by all means let's revisit it in the future.

nickstenning: I've gone and reread Google's guidance on use of redirects, and actually I think a permanent redirect is correct here.

See, for example: https://support.google.com/webmasters/answer/139066?hl=en

It's worth emphasising: a "permanent redirect" isn't permanent, in this context it just means "this isn't the canonical URL for the page you're requesting, please update your records". Which is exactly the meaning these redirects are intended to convey.

conf/production.ini
```diff
@@ -15,6 +15,8 @@ use: egg:h
 #h.client_id:
 #h.client_secret:
 
+h.hashids.salt: production salt
```
nickstenning: I'm inclined to remove this so that if you don't set the environment var in production the application will explode rather than silently using a predictable salt.

seanh: Iirc correctly I put this in because Travis was crashing otherwise (it builds the Chrome extension with production.ini iirc). We could fix the Travis script instead

nickstenning:
- Rebased on master
- Schema migration rebased onto current head
- Switched to using `python-slugify`
- Updated view function names
- Made the FK fields in the user-group join table `NOT NULLABLE`
- Lightly squashed the commit history

seanh:
Just fixed most of landscape's issues, I'm happy to leaves it remaining complaints alone

seanh:
@tilgovi We're done with this, merge if you're happy with it

h/groups/views.py
tilgovi: Thanks for making it temporary. I'd like to keep it that way for now because I'm not convinced yet that the place its redirecting to is canonical, yet, until we do some thinking about it.

tilgovi:
Rebased over the simple conflict with the admins permissions PR. Waiting for the green light just to be paranoid, but it looks good and I'll hit merge when that's good.

tilgovi:
\o/

PR #2403, opened 2015-07-24
Simpler session service

nickstenning:
This PR does three things. In order of importance:

- wraps the ngResource for managing session state in a service, which
  allows us to expose current session state as `session.state` on that
  instance. This means that components of the application that don't
  want to alter or explicitly fetch session state, but wish to display
  some part of it (such as the current authenticated userid, etc.) can
  do so more easily.
- moves configuration of the session resources into the session service
  rather than requiring configuration from outside via a
  sessionProvider.
- translates the session service to JavaScript from CoffeeScript.
- rolls up the existing 'xsrf' service/value into the session state.


nickstenning:
Having done this, it's obvious that it would be nice, if possible, to make `session.state.userid` the canonical "what's the current userid".

nickstenning:
But there's a bit more to unpick there between the `auth` and the `identity` service, so that's probably worth talking about synchronously...

tilgovi:
I think you should take a read through my emails to the dev list about extensibility and take another shot at this.

I would like to make the session a storage interface. Let's add DELETE/PUT to the server for /session/{key} so the client can actually use it as a place to persist things.

I like `session.state` because I don't want to mix the storage keys with the API methods. But a nicer API might be `session.get()` that returns the state (optionally `session.get('key')`) and `session.put('key', 'value')`. These could be either asynchronous or they could be synchronous with companion `session.save()` and `session.load()`, etc. One appealing thing about that is that `get` and `put` can be implemented concretely while `save()` and `load()` can be implemented on the provider as no-ops by default. Then, customizing the sessionProvider becomes totally optional, but customizing it to deal with a server-side session is pretty easy.

I like that you took the session stuff out of the accounts config, but I don't want to move those actions into the session more formally. I would rather the accounts package just have its own ngResource for some authentication routes and maybe we can finally deprecate the hideous __formid__ stuff soon.

Let's definitely hold the config package line, though. I really want to keep as many assumptions about client<->server protocol, particularly around Hypothesis-specific stuff like auth, out of the main application. The only protocol specifics I'm really comfortable putting outside the config package right now are basic annotation CRUD things. The rest is too specific and subject to change.

tilgovi:
I'm merging this. Let's start with this. It will make some next steps easier, for sure. Thank you.

PR #2404, opened 2015-07-24
Enable toggling features for admins only

nickstenning:
Allow toggling features on for users set as admins in our accounts system. This means we can deploy features to production and allow admin users to test them before releasing them to the general population.


tilgovi:
Nice one. Liking this new Features system.

PR #2405, opened 2015-07-27
Our 404 error response should have a 404 status code.

nickstenning:
Before:

    $ curl -sD- https://hypothes.is/app/i/dont/exist | head -n1
    HTTP/1.1 200 OK

After:

    $ curl -sD- https://localhost:5000/app/i/dont/exist | head -n1
    HTTP/1.1 404 Not Found


seanh:
Surprised that `@notfound_view_config` doesn't do this by default

PR #2407, opened 2015-07-27
Consolidate and style the admin interface

nickstenning:
This commit brings together "administrator" views into one place -- the
Hypothesis admin console, found at /admin.

So far, the only to admin views we have are for flagging/unflagging
users as NIPSA, and for administering administrative user status. In the
near future, we may also have a feature flag console in here.

- I've moved all the admin views into `h.admin` so that we don't
  confuse business logic view functions with administrative view
  functions, and so that in future it is easier to treat the `/admin`
  tree as a single entity for the purposes of permissions, etc.
- I've added an "admin" page layout, which uses Twitter Bootstrap for
  basic styling.
- I've rejigged both the NIPSA and the administrator users pages to make
  use of bootstrap.


nickstenning:
FYI, ignoring Bootstrap:

```
 h/accounts/test/views_test.py             | 153 ------------------------------------
 h/accounts/views.py                       |  57 --------------
 h/admin.py                                | 108 ++++++++++++++++++++++++++
 h/app.py                                  |   2 +-
 h/assets.yaml                             |  23 ++++++
 h/templates/admin/admins.html.jinja2      |  51 ++++++++++++
 h/templates/admin/index.html.jinja2       |  11 +++
 h/templates/admin/nipsa.html.jinja2       |  55 +++++++++++++
 h/templates/admin_users.html.jinja2       |  33 --------
 h/templates/includes/flashbar.html.jinja2 |  10 +++
 h/templates/layouts/admin.html.jinja2     |  63 +++++++++++++++
 h/templates/nipsa.html.jinja2             |  31 --------
 h/test/admin_test.py                      | 232 +++++++++++++++++++++++++++++++++++++++++++++++++++++++
 h/test/nipsa_test.py                      |  70 -----------------
 14 files changed, 554 insertions(+), 345 deletions(-)
```

nickstenning:
The coveralls figure is bogus, I think. It hasn't picked up the `h.admin` and `h.test.admin_test` modules.

nickstenning:
Oh, and this is another PR where a checkout and `git diff -C` will be your friend.

seanh:
I think it'd be nice to have `h/admin/admin_users/views.py` and `h/admin/nipsa/views.py` instead of having all the views in one file, but we can always move to a layout like that later when we have more features.

seanh:
I'm also thinking it'd be nice  to implement the whole foo feature in h.foo, so it'd be nice if it could have h/foo/admin/views.py, and register a tab and views with the admin console from there. Again happy to leave that until later

seanh:
I might suggest `admin_users` instead of `admins`, to differentiate it more from `admin` begin used for something else now

nickstenning:
> I'm also thinking it'd be nice to implement the whole foo feature in h.foo, so it'd be nice if it could have h/foo/admin/views.py, and register a tab and views with the admin console from there.

Maybe, but I'm currently thinking of the admin pages as clients of the API of other modules, rather than thinking of admin pages as provided by other modules. There are dependencies either way you look at it:

- either admin pages depend on a preexisting API to get their work done
- or admin pages depend on an admin-pages-registration API and are blended with the code they provide admin tools for

I don't feel particularly strongly about it either way, but I went with the former for now. If we end up having more admin pages perhaps we'll feel that registration is a better way to go...

nickstenning:
As for "admins" vs "admin_users" -- personally, I prefer the former, and I don't think there are any contexts where it's particularly unclear, but if you feel strongly I'm happy to change it.

PR #2409, opened 2015-07-28
Move `request.es` property to h.api.db

nickstenning:
Given that `h.api.db` is already responsible for reading the settings
and configuring the ElasticSearch client, it seems sensible that this
request property is added here.

This simplifies `h.api.search` and will simplify getting access to the
ElasticSearch client from CLI tools I'm currently working on.


tilgovi:
Yay. I also like seeing the index parameterized.

PR #2411, opened 2015-07-29
Show orphans (with notice) and sort by selector

tilgovi:


nickstenning:
Can you put this behind a feature flag so we have control over it in production?

tilgovi:
Closing, will re-open with new code.

PR #2412, opened 2015-07-29
Share a group

seanh:
This is the second pull request for the groups feature.

What's in here:

* If you visit a group's page when you're not logged-in it tells you what group this is and gives you a link to login in order to join
* If you are logged-in it gives you a button to join the group. Clicking this button adds you to the group in the db and reloads the page.
* If you are already a member it tells you so and gives you a link to copy and share with others in order to invite them
* "You've joined" flash message
* Needless to say some design polish is needed for these pages.
* The login link just opens `/login` in a new tab. After logging in or registering and activating you have to go back to the original tab and refresh it. It doesn't attempt to take you to the login page within the one tab and then have it redirect back to the group's page after login, because that doesn't work for people who need to register, activate, possibly install the extension. More onboarding flow design work is needed here.
* There are two links to the group's page in the groups dropdown in the sidebar: both the group's name and the "link" icon link to it. This is because in the future the group's name will become a button for focusing the sidebar on that group instead.

### DIFF
--- a/h/groups/logic.py
+++ b/h/groups/logic.py
@@ -1,0 +1,7 @@
+from h import hashids
+
+
+def url_for_group(request, group):
+    """Return the URL for the given group's page."""
+    hashid = hashids.encode(request, "h.groups", number=group.id)
+    return request.route_url('group_read', hashid=hashid, slug=group.slug)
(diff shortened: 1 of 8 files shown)


seanh:
@tilgovi This one is a couple of days old, also ready for review, just did a massive rebase onto the latest master

seanh:
Rebased onto latest master

nickstenning:
LGTM.

PR #2414, opened 2015-07-29
Reorganise `h.api.search` into package

nickstenning:
This is prep work for #2413, probably easier to review and merge on its own.


tilgovi:
Looks good.

PR #2417, opened 2015-07-30
Fix pylint warnings about `id` name

seanh:


seanh:
@nickstenning

tilgovi:
Can we just fix the places where we use `id`? I saw you use `id_` somewhere recently and I would prefer that, since `id` is a builtin.

seanh:
I'm happy either way, Nick wanted to allow it rather than fix places

tilgovi:
Having never actually had occasion to use the built-in `id` function explicitly, I can understand that.

Nick, think twice about it and then make a decision?

nickstenning:
This is fine by me. The case in particular that Sean's fixing is a class attribute for an `id` column, where `id_` is just horrible. Django gets around this by using `pk` everywhere.

Anyway, while I think using `id_` is good practice for local variables, I too have only once actually had a valid use for the `id` builtin, so this is fine. Please update the comment in `.prospector.yaml`, though.

seanh:
Alright I've force-pushed this. New suggestion: please use `id_` unless it's a db column or you have some other good reason, in which case silence the pylint message per-line

seanh:
Ok, scratch that. Force-pushed, this allows `id` as a name so the warnings about class attrs are gone, but locals and params will still get a warning about shadowing the builtin so use `id_` for those

PR #2425, opened 2015-07-31
Switch out jschannel for frame-rpc

### LINKED ISSUE #2397: H activation fails at abcnews.go.com
URL: http://abcnews.go.com/Politics/lindsey-graham-destroyed-cellphone-spite-donald-trump/story?id=32621487

Symptom: Hypothesis does not activate

Console error in hypothesis.min.js:

Uncaught TypeError: Cannot read property 'Util' of undefined8.../annotation-sync @ hypothesis.min.js?bc659cac:2s @ hypothesis.min.js?bc659cac:1e @ hypothesis.min.js?bc659cac:1(anonymous function) @ hypothesis.min.js?bc659cac:1

tilgovi:
The frame-rpc module is a proper CommonJS module and it's footprint
is really tiny. Its protocol is simpler. It doesn't handle connection
and buffering, but with our onConnect callback we don't need buffering.
The connection handshake that jschannel did was so trivial it's been
re-implemented here (each side tries to connect to the other once).

We have to handle timeouts ourselves, but that's all hidden in the
bridge code. In exchange, we get a simpler API, we get rid of the
call/notify distinction in favor of just passing a callback or not
and we avoid the excess overhead of the recursion guard in the
serialization code that was giving us false positive issues with
the document title.

This is one step toward removing all the browserify-shim requiring
libraries from the injected bundle, which will eventually fix #2397.


tilgovi:
Okay. This is verified manually. The tests all pass. I just re-forced with having removed the 'notify' calls. I did a grep for "method: " to make sure I hadn't missed any calls somehow.

I'm pretty confident this is good to go, unless there are objections.

judell:
Nice!

h/static/scripts/annotator/guest.coffee
```diff
@@ -413,11 +400,11 @@ module.exports = class Guest extends Annotator
 
   # Open the sidebar
   showFrame: ->
-    @crossframe?.notify method: 'open'
+    @crossframe?.call('open')
 
   # Close the sidebar
   hideFrame: ->
-    @crossframe?.notify method: 'back'
+    @crossframe?.call('back')
```
nickstenning: "open"/"back" o_O

nickstenning: (Not suggesting that's an issue with this PR. Just... o_O...)

h/static/scripts/bridge.coffee
```diff
@@ -14,20 +14,28 @@ module.exports = class Bridge
...
-  createChannel: (source, origin, scope) ->
+  createChannel: (source, origin, token) ->
+    channel = null
+    connected = false
+
+    ready = =>
+      return if connected
+      connected = true
+      for cb in @onConnectListeners
+        cb.call(null, channel, source)
```
nickstenning: I'm guessing nobody uses the function binding value (which has changed from the bridge instance to null).

nickstenning:
LGTM.

PR #2424, opened 2015-07-31
Limit what admins can do

seanh:
Fixes #2422

PR #2430, opened 2015-08-03
Simplify handling of admin permissions

seanh:
@tilgovi I realised that https://github.com/hypothesis/h/pull/2424 was kind of silly. We don't need a custom context resource factory for admin-only views, we just need to give the 'group:admin' principal the 'admin' permission on the root resources, and add the 'admin' permission to admin views.

That is, assuming we don't already use a permission called 'admin' anywhere else.

### DIFF
--- a/h/admin.py
+++ b/h/admin.py
@@ -1,39 +1,11 @@
 from pyramid import view
 from pyramid import httpexceptions
-from pyramid import security
 
 from h.api import nipsa
 from h.i18n import TranslationString as _
 from h import accounts
 from h import models
 from h import util
-
-
-class AdminResource(object):
-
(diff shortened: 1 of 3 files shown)


nickstenning:
LGTM.

I think 'admin' as a permission name is fine for these views, as the permission really kind of is "behave as a site admin" quite explicitly, but we should be cautious about missing the main point of a capability-based access control system: namely decoupling names for authorisation groups from names of capabilities. (So, don't fall into the trap of thinking that everything an admin should be able to do is to be guarded by the "admin" permission...)

seanh:
I was worried that maybe we used 'admin' as a permission name on some of our annotations too

PR #2432, opened 2015-08-04
Publish to a group

seanh:


seanh:
- [ ] The groups menu needs to be behind the groups feature flag
- [ ] Need to check what happens to the group field when you edit or reply to an annotation

seanh:
Issues with the Public / Only Me dropdown on the annotation cards (in the [mockup](https://groups4.dokku.hypothes.is/) this dropdown also includes any groups the user belongs to):

- You can have a different group selected in the global groups dropdown and in the dropdown on the annotation you're editing. Then when you hit save your annotation disappears.

- The dropdowns also appear in replies so you can have a reply to an annotation in group foo, but the reply is in group bar. Or a private reply to a group annotation (group annotations can't be private).

- Just not showing groups in the dropdown on the cards doesn't work either. We don't allow private annotations in groups, so we would have to hide the public/private dropdown if the annotation being created or edited belongs to a group. In the case of annotations being created, whether the annotation will belong to a group depends on what's selected in the global groups dropdown when you click save. But the user can change what they've selected in this dropdown _while_ they are editing an unsaved annotation. So we'd have to hide/show the public/private dropdown as different things get selected in the groups dropdown.

Suggested solution:

- Remove the public/private dropdown menu from annotation cards and have _only_ a global dropdown containing these entries:

  - Public
  - Private
  - And entries any groups the user belongs to

  No _All_ entry as in the mockup.

- This means that you can't show public, private and group annotations in the sidebar at once, only one of these categories can be shown at a time.

- When creating a new annotation, whatever is selected in this public/private and groups dropdown is what becomes of the annotation when you click save.

- When replying to an annotation, the reply also goes into the same category as the original annotation (you have to have that category selected in the dropdown, otherwise you wouldn't be able to see the annotation to click reply on it)

- Private replies to public annotations (and vice-versa) will no longer be possible (suspect these are broken in current h anyway)

- It won't be possible to edit annotations and change them from public to private or vice-versa, or move them between groups. There are too many questions about this anyway: What happens if I make a reply to a public annotation, and then the original annotation's author makes the annotation private? Or vice-versa? What happens to an annotation's replies if the annotation is edited and moved into or out of or between groups?

dwhly:
Thanks for the detail you've provided.  I agree that this is some of the trickiest territory we have, and getting it right is important.

> No All entry as in the mockup, this entry does not work.

Of your proposals, this is the one that's the biggest functional change, and the one I'll argue most strenuously for.  

The All option allows me to see all the annotations on a page that I'm authorized to see, without having to guess which various groups or modes they might be present in, and allowing me, for instance as a teacher with multiple class sections to see any and all annotations on a document simultaneously.  It allows me to see any private annotations on a page at the same time I can see any other annotations.

For instance, when I first open the app on a page, what should our default be?  Public?  If so, I won't be able to see any private notes I've made without switching there first.  From what I understand, your argument for removing All is that it obviates the need for the visibility drop down on annotation creation, because without All, the visiblity can always be slaved to the master view.  

While I understand that, I think we need to work through the complexity to address the issues you've raised, perhaps with some of the suggestions you've made below and a few others.  Getting rid of All is a big change from the current sense (albeit with only public and private) where there is a top level view that shows me everything I'm authorized for. 

> When creating a new annotation, whatever is selected in this public/private and groups dropdown is what becomes of the annotation when you click save.

Makes sense.

> When replying to an annotation, the reply also goes into the same category as the original annotation (you have to have that category selected in the dropdown, otherwise you wouldn't be able to see the annotation to click reply on it)

Agree strongly that replies should default to visibility of their parent.  The big question with restricting them to only the visibility of the parent is in losing the draft capability and personal note-taking that private allows.  I agree with preventing replying to a group annotation w/ the visibility of a different group, or public.

> Private replies to public annotations (and vice-versa) will no longer be possible (suspect these are broken in current h anyway)

Sean, I've just replied privately to this annotation of yours (https://hypothes.is/a/3SYM0dqjSZ2iX1-j1yH2Xg).  Seems to work.  The strange case of public replies to private annotations are currently allowed, though a bit pointless since only you would ever be able to do that to one of your own, and then only you would ever be able to see it. I'd be fine w/ preventing this though I think the impact is minimal if at all.

> It won't be possible to edit annotations and change them from public to private or vice-versa, or move them between groups. There are too many questions about this anyway: What happens if I make a reply to a public annotation, and then the original annotation's author makes the annotation private? 

Well, we currently allow users to delete annotations, which from the perspective of others is the equivalent of making something that was once public, private.  We provide an admonishment that explains it, which could simply remain as is.  But losing this capability wouldn't bother me too much. i.e. If you want to remove it from view, delete it-- the negative is that you can't preserve what you originally said for your own purposes.

> Or vice-versa? 

i.e. what if I make a public reply to a private annotation, and then "the original author" i.e. me makes the annotation public?  Non-issue, right?

> What happens to an annotation's replies if the annotation is edited and moved into or out of or between groups?

Agree with you that annotations (especially ones w/ replies) shouldn't be able to be moved between groups.

> What happens to an annotation if I edit it and move it into another group? Given that the sidebar is filtered to show only annotations from the original group, the edited annotation vanishes on save (confusing).

As above, agree that allowing moving annotations once created to another group is probably unnecessary and confusing as you point out.  Happy to prevent that.

> A user can do this manually by deleting an annotation, change to a different public/private/group in the dropdown, and re-creating the annotation.

Reasonable.

> What about if I reply to an annotation, then the original author deletes the annotation? This has nothing in particular to do with groups. The current behaviour is that Message not available appears in the stream or sidebar where the annotation and its replies would be. User can't see the annotation or the replies, and of course this makes no sense. (This looks like a bug, I'm guessing it's supposed to say Message not available then show the replies beneath the missing message.)

Yes, it's a bug. Just verified.  https://github.com/hypothesis/h/issues/2436


My suggestions:

- Keep All.
- Default visibility is whatever the social view selector is set to.
- Prevent replying with a different group that what the parent is. 
- Try to keep private replies (which could include text or tags).  Otherwise the only content people can't annotate for personal use (such as organizing w/ personal tags) on the web are annotations.  We also lose our drafting mode.
- Prevent changing groups once created.
- Prevent public replies to non-public annotations.
- In any other situation where creating an annotation might result in it disappearing, switch the view to "All".

seanh:
Thanks @dwhly. I think my rant above could have been a lot clearer (in my defense: Saturday evening).

Ok, I think what I'm gonna do is just go ahead with my implementation here (fixing up the code and tests etc, getting it reviewed) and ignore these UI problems for now, as it needs more thought and discussion.

But in the meantime here's another suggestion for discussion, trying to take into account some of the points you raise - Keep the concepts of public/private and groups separate; Use the global sidebar groups dropdown to show a list of groups, allow filtering the view to annotations from one group only, and to decide which group new annotations go into; Use the dropdowns on the individual annotations to select between public and private. So:

1. In the global groups dropdown menu have:

  - All
  - No group
  - My Group 1
  - My Group 2
  - ...

   When a new annotation is created, whichever group is selected in this dropdown at the moment when you hit save is what group the annotation will go into.

2. Do have a second dropdown on annotations when creating or editing them, in that dropdown have:

   - Public
   - Private

I've used _No group_ instead of _Public_ in the global dropdown because the word public is used to mean something else in the per-annotation public/private dropdown and I didn't want to overload the word. Also creating a private annotation in a group called public seems odd. _No group_ may not be the best label here though, but I can't think of a better one yet.

This solution would mean:

1. The _All_ view is back

2. Which raises the question - which group does an annotation go into, if it's created while _All_ is selected? Into the _No group_ group? Will this be clear to users?

3. Annotations can be public or private regardless of what group (if any) they belong to. You can have private annotations in groups.

4. You can edit a private annotation and make it public, or vice-versa. Drafting capability is back, both in _No group_ annotations and within groups and in replies.

5. You can't edit an annotation and move it between groups, though.

6. Private replies to public annotations and vice-versa are back

7. Replies can only belong to the same group as their parent - the UI doesn't allow anything else to happen (nor do we have to worry about what to do with replies if their parent gets moved into a different group, the UI doesn't allow moving annotations between groups)

8. We don't have a _Private_ view to show only your private annotations. A private annotation either belongs to the _No group_ group or to some other group, and to see it you have to select that group (or select the _All_ view).

  Note that in production Hypothesis right now there's no _Private_ view (or any views at all) so we're not losing a feature from production here, just losing a (problematic) feature from the groups mockup.

9. The disappear-on-save problem is gone.

10. We _do_ still have the problem of replies whose annotation is gone for some reason: it has been deleted, made private, or nipsa'd (at least it can't have been moved into a group though, since moving annotations isn't supported in this design). But all three of these problems already exist in production, and we have a solution (showing _Message not found_ and then showing the replies underneath). So I think that's fine.

dwhly:
> I think my rant above could have been a lot clearer (in my defense: Saturday evening).

For me, it's the back and forth on this that's clarifying, so thanks for wading in.

> Ok, I think what I'm gonna do is just go ahead with my implementation here (fixing up the code and tests etc, getting it reviewed) and ignore these UI problems for now, as it needs more thought and discussion.

Sounds good.

> But in the meantime here's another suggestion for discussion, trying to take into account some of the points you raise. 

I like the direction you're headed.  Here's a modified form that I think addresses the key points.

1- In the social view dropdown have (If you don't mind I'm going to stick w/ the term that @JakeHartnell and I first coined because the potential utility of this filter goes beyond just selecting for groups IMO.  Here's the [original doc](https://docs.google.com/document/d/1bXRjm7rL8xlwsb-kXiRfxSeB8qLShSzmG1MP6cB7JI8/edit) he drafted around the idea):

- All
- Public
- Only Me
- My Group 1
- My Group 2

2- In the current visibility selector on annotations when creating or editing them have:

- [Currently selected group, or "Public" if it's set to All]
- [All other groups I have permission for if All is selected]
- Only Me

This solution would mean:

1. The All view is back / remains.
2. The term "Public" preserves the sense that an annotation is discoverable by anyone coming to that page.  I don't think we want to suggest that annotations made against a private group, perhaps just between two people, are "Public" when we really mean "visible to the current group".
3. Instead, the name of the group being published to is explicitly stated/reinforced in the editor.  There's no question what you're publishing to.
4. If an annotation is created while All is selected, it defaults to the Public group (or "no group" if you will). This is the behavior we have now-- we default to Public.  However, all other groups you are authorized for are also available, since publishing to one of them will not result in a disappear-on-save situation.
5. You can edit a private annotation and make it public, or vice-versa.
6. You can't edit an annotation and move it between groups, because if it's been published to a group, the other groups are no longer available in the visibility selector.  Only the group and "Only Me".
7. Private replies to public annotations and vice-versa exist.
8. Replies can only belong to the same group as their parent or "Only Me", because again the other groups are no longer available in the visibility selector.
9. We preserve the Only Me view to show only your private annotations.
10. If the groups filter is set to a group we show any annotations made to that group as well as any private annotations. This would eliminate the last of the disappear-on-save problems I think.  (Not sure how complicated this is to implement). Essentially your private annotations are always visible. The Only Me option just filters for *only* your private annotations.

seanh:
Okay, I'm gonna present two possible solutions now. Which do we want to go for as a first pass at groups?

1. The simplest, less features approach that involves losing the ability that we currently have on prod to use private annotations as drafts?

2. A more complex, more features approach that brings many more UX complexity and questions?

* * *

#### 1. Simple, less features

How important to us is our drafting capability? That is, the ability to create a private annotation (viewable only to the user who created it) inside the public group or a user-created group, and then later edit that annotation and make it viewable by the public or by the group?

I wonder if we can give up on the ability to use public/private in this way for now, and just say that we'll introduce a separate "save as draft" feature later.

Both the private and the drafts feature may be better served if kept as separate features, trying to combine them into one is imo overloading one thing with two meanings and makes it conceptually and UI more complicated.

Specifically, if we can give up the drafts capability for now then we can simply put the following entries in the global social view dropdown:

* All
* Public
* Private
* My Group 1
* My Group 2
* ...

And then we would not have any local dropdowns on the annotations themselves - just the one global dropdown on the sidebar.

Public, Private, My Group 1 and My Group 2 etc would be exactly the same as each other, both conceptually and in terms of the UI. They are groups. Each annotation belongs to exactly one group. To make an annotation you select a group from the dropdown and annotate, the annotation goes into the selected group. And that's it. Annotations do not have any separate public/private state that's independent of what group they belong to, Public and Private are simply groups. You can't move an annotation between groups or between public and private by editing it.

The only special thing is _All_ which is not a group but a view over all of the groups, and when creating an annotation in this view it goes into the _Public_ group (and we'll indicate that by showing it on the annotation card).

This is by far the simplest solution we have - conceptually, in terms of the UI, and to implement.


* * *

#### 2. Complex, more features

These are my latest thoughts on a keep-most-of-the-proposed-features design, trying to take @dwhly's latest suggestions into account.

In the global social views dropdown:

* All
* Public
* My Group 1
* My Group 2
* ...

And then in the local dropdowns on the individual annotations when creating or editing them:

* <The currently selected group from the global dropdown, or Public if All is selected>
* Private

So in this version:

* Public is now two things: it is a group in the global dropdown, and it's a piece of state (public/private) that any annotation belonging to any group has independent of what group it belongs to.

* Private is no longer a group at all but part of that group-independent public/private state.

* We get our drafts capability back: You can create a private annotation and then later on edit it and make it public or viewable by the group

* You can create a "private" annotation inside the group called "public" :(

I've h
I want to hold back on two things for now - 1. Allowing creating annotations in any group when _All_ is selected and 2. Having _Only Me_ in the global social view menu.

* * *

1. Allowing creating annotations in any group when _All_ is selected

This is extra code. Normally the local dropdown on an annotation will show two options: Private and  the name of the selected group. Now I'll need extra code to instead show the names of all groups when _All_ is selected.

I think there may be UI problems too - what about when I _edit_ an annotation with the _All_ view selected? Then I could use the annotation's local dropdown to move it between groups. Since _All_ is selected we won't have the vanish-on-save problem, but we will have the problem of deciding what to do with an annotation's replies when it gets moved - do they move with the annotation? This would mean potentially making public replies group-private, or group-private replies public or private to a different group. Do they not move with the annotation? Then the annotation loses it replies, and the replies lose their annotation.

I think that moving annotations between groups isn't part of the initial groups requirements, so I'd prefer to avoid it at first.

We could say that all the groups show when creating an annotation, but then when editing an annotation it's fixed to the chosen group (even if the _All_ view is selected). So moving between groups is avoided. But will the groups being available when creating, then disappearing when editing, confuse and/or annoy users?

Also is the sticky visibility selector a major source of user error? It trips me up all the time - I publish several annotations that I meant to be public as private. So if we add groups to that, it may result in people publishing multiple annotations to the wrong group then cursing.

* * *

2. Having _Only Me_ in the global social view menu

This is a new feature (being able to show only your own private annotations) that we don't have in production now, that isn't strictly part of the groups feature, and that I think may create some user issues:

1. What group does an annotation go into, if it's created with _Only Me_ selected in to the social view dropdown? It's gonna have to go into an _Only Me_ group, which means that _Only Me_ now becomes a group in the same sense that _Public_ and _My Group 1_ etc are groups. But _Only Me_ is also a view over multiple groups in the same sense that _All_ is, I can create an annotation in _My Group 2_ and set its local dropdown to _Only Me_. So _Only Me_ is both a group and a view over multiple groups at once, and I'm not sure if that's a good idea.

2. If _Only Me_ is selected in the social view dropdown, then what do you show in the local annotation dropdowns when creating or editing? We would have to show _Only Me_ only, otherwise we'll get the vanish-on-save problem. But for _some_ of those _Only Me_ annotations I can also go over to the group they were created in and edit them there, and _then_ I'll get the group as an option in the local dropdown. So two different behaviors when editing the same annotation, potentially confusing.

* * *

So I'm thinking maybe in the global social view dropdown, as a first-pass:

* All
* Public
* My Group 1
* My Group 2
* ...

And then in the local dropdowns:

* <The currently selected group, or Public if All is selected>
* Private

I think this is nice and simple, easy both to implement and to understand as a user, because what shows on the local dropdowns is always the same whenever creating or editing (apart from the first entry changing to show the name of the current group).

It still seems _slightly_ odd to me that you can create a _Private_ annotation inside a group named _Public_. My preferred solution to this would be the one I initially suggested - move _Private_ up into the global social view dropdown and don't have any local dropdowns on the annotations at all. Public, Private, and user-created groups are then all the same conceptually and in terms of UI: they are groups, an annotation can belong to exactly one of them, and annotations don't have any other public/private state that's independent of which group they belong to. The only special thing is _All_

But as 

I think we may also get user confusion when they have _All_ selected and want to create an annotation in a group but it won't let them, they may not realise that they have to select the group in the global social view first. Again I think wait and see and then maybe tweak it

seanh:
Okay, I'm gonna present two possible solutions now. Which do we want to go for as a first pass at groups?

1. The simplest, less features approach that involves losing the ability that we currently have on prod to use private annotations as drafts?

2. A more complex, more features approach that brings many more UX complexity and questions?

* * *

#### 1. Simple, less features

How important to us is our drafting capability? That is, the ability to create a private annotation (viewable only to the user who created it) inside the public group or a user-created group, and then later edit that annotation and make it viewable by the public or by the group?

I wonder if we can give up on the ability to use public/private in this way for now, and just say that we'll introduce a separate "save as draft" feature later.

Both the private and the drafts feature may be better served if kept as separate features.

Specifically, if we can give up the drafts capability for now then we can simply put the following entries in the global social view dropdown:

* All
* Public
* Private
* My Group 1
* My Group 2
* ...

And then we would not have any local dropdowns on the annotations themselves - just the one global dropdown on the sidebar.

Public, Private, My Group 1 and My Group 2 etc would be exactly the same as each other, both conceptually and in terms of the UI. They are groups. Each annotation belongs to exactly one group. To make an annotation you select a group from the dropdown and annotate, the annotation goes into the selected group. And that's it. Annotations do not have any separate public/private state that's independent of what group they belong to, Public and Private are simply groups. You can't move an annotation between groups or between public and private by editing it.

The only special thing is _All_ which is not a group but a view over all of the groups, and when creating an annotation in this view it goes into the _Public_ group (and we'll indicate that by showing it on the annotation card).

This is by far the simplest solution we have - conceptually, in terms of the UI, and to implement.


* * *

#### 2. Complex, more features

These are my latest thoughts on a keep-most-of-the-proposed-features design, trying to take @dwhly's latest suggestions into account.

In the global social views dropdown:

* All
* Public
* My Group 1
* My Group 2
* ...

And then in the local dropdowns on the individual annotations when creating or editing them:

* The currently selected group from the global dropdown, or Public if All is selected
* Private

So in this version:

* Public is now two things: it is a group in the global dropdown, and it's a piece of state (public/private) that any annotation belonging to any group has independent of what group it belongs to.

* Private is no longer a group at all but part of that group-independent public/private state.

* We get our drafts capability back: You can create a private annotation and then later on edit it and make it public or viewable by the group

* You can create a "private" annotation inside the group called "public" :(

* We may get user confusion when they have the _All_ view selected and try to create an annotation. The user can _see_ annotations from all groups in this view, but they can only _create_ annotations in the _Public_ group. In the annotation's dropdown they can only choose between _Public_ and _Private_, they may not realise that to create an annotation in a group they have to first select the group in the global dropdown.

In this design I've held back on two additional features that @dwhly suggested above:

1. Allowing creating annotations in any group when _All_ is selected
   - More code and UX complexity
   - What about when I _edit_ an annotation with the _All_ view selected? Then I could use the annotation's local dropdown to move it between groups. We will have the problem of deciding what to do with an annotation's replies when it gets moved - do they move with the annotation? This would mean potentially making public replies group-private, or group-private replies public or private to a different group. Do they not move with the annotation? Then the annotation loses it replies, and the replies lose their annotation. Or do we allow users in the _All_ view to _create_ annotations in any group, but not allow them to move existing annotations between groups when editing them? Maybe we only remove the moving capability when the annotation has replies? More UX complexity.
   - I think that moving annotations between groups isn't part of the initial groups requirements, so I'd prefer to avoid it at first.
   - Is the sticky visibility selector a major source of user error? It trips me up all the time - I publish several annotations that I meant to be public as private. So if we add groups to that, it may result in people publishing multiple annotations to the wrong group then cursing.

2. Having _Private_ in the global social view menu.
   - What group does an annotation go into, if it's created with _Private_ selected in to the social view dropdown? It's gonna have to go into an _Private_ group, which means that _Private_ now becomes a group in the same sense that _Public_ and _My Group 1_ etc are groups. But _Private_ is also a view over multiple groups in the same sense that _All_ is, I can create an annotation in _My Group 2_ and set its local dropdown to _Private_. So _Private_ is both a group and a view over multiple groups at once, and I'm not sure if that's a good idea.
   - If _Private_ is selected in the social view dropdown, then what do you show in the local annotation dropdowns when creating or editing? We would have to show _Private_ only, otherwise we'll get the vanish-on-save problem. But for _some_ of those _Private_ annotations I can also go over to the group they were created in and edit them there, and _then_ I'll get the group as an option in the local dropdown. So two different behaviors when editing the same annotation, potentially confusing.

seanh:
Alright, I think this is the approach I'm going to go for: fewest possible changes from what we currently have in prod. Don't remove any features that we currently have in prod (e.g. the ability to use the public/only-me visibility selector to have drafts) but it also don't add any features that aren't strictly required to implement the groups requirements I've been given on the Trello cards (e.g. no _All_ or _Only Me_ views, no moving annotations between groups). Imo this is the best and quickest way to get to workable groups implementation, and then we can add additional features later, one by one. So:

1. We'll have a global _groups dropdown_ menu at the top of the sidebar, with these entries:

   - Everyone
   - My Group 1
   - My Group 2
   - + New Group

   This menu shows all the groups the user is a member of.
   In this case the user has created (or joined) two groups named My Group 1 and My Group 2, and all users are always members of the default Everyone group.

   When the user selects a group from this dropdown then 1. The sidebar is filtered to show only annotations from that group and 2. Any new annotations that the user creates will go into that group.

   When loading a new page Everyone is selected at first.

2. If the user hasn't joined any groups this menu would only have the one Everyone item, so maybe we wouldn't show the menu at all in that case.

3. As we currently have in production, we will have also have a separate _visibility dropdown_ menu on the individual annotations when creating or editing them. These will have two entries:

   - Name of current group from groups dropdown
   - Only Me

   So if Everyone is selected in the groups dropdown and you create or edit an annotation, then on the annotation's visibility dropdown you'll see:

   - Everyone
   - Only Me

   If My Group 1 was selected in the groups dropdown then you would see My Group 1 in place of Everyone in the annotation's visibility dropdown.

Notes:

- This is conceptually clean. We have two separate concepts: groups and visibility. We have two separate dropdown menus: the groups menu and the visibility menu. We don't (yet) introduce the third concept of a "view" over multiple groups (_All_ or _Only Me_ views), we don't have things that show up with the groups in the social view menu but that are not (quite) the same thing as groups.

- We retain our drafting capability. Whether in the Everyone group or another group an annotation can be made private and then later edited and made group-viewable.

- There's no way to move annotations between groups, so we don't have to deal with all the questions that raises

- There's no _All_ view. I'm arguing that being able to see all your annotations (public, private, any group...) at once would be a new feature (since currently in prod we don't have groups at all) and not part of the spec I've been given for this first version of groups (see Trello cards), and raises lots of complexities that I want to avoid for now. Consider adding it later as a separate feature.

- There's no _Only Me_ view that shows you all of your private annotations from any group. To see your private annotations, view whichever group you made them in. Again: new feature, out of spec, brings complexities. Consider adding as a separate feature later.

jeremydean:
in no way questioning your thinking here. seems sensible. just want to flag one thing:

from my personal experience and in conversation with folks, the "only me"/ private feature is *not* primarily a drafting feature.

user POV: i want to annotate this URL/page/doc but just for my own research/inquiry purposes with no intent of ever making public.

i think the drafting is an important sub-feature/hack, but at some point we need to deal with the use of a user wanting to annotate an entire page for themselves.

seanh:
This pull request is going haywire because of some git confusions we're having, technical issues

seanh:
> from my personal experience and in conversation with folks, the "only me"/ private feature is not primarily a drafting feature.
> 
> user POV: i want to annotate this URL/page/doc but just for my own research/inquiry purposes with no intent of ever making public.
> 
> i think the drafting is an important sub-feature/hack, but at some point we need to deal with the use of a user wanting to annotate an entire page for themselves.

To my mind this argues for:

1. _Only Me_ should become a group, move into the groups dropdown, then the visibility dropdown would no longer be needed at all. Great UI and conceptual simplification.

2. We add a separate _Save as draft_ feature

For this first groups implementation, in the interests of minimal changes from what we have currently, I'm not gonna do either of those things though

seanh:
Technical issues mostly resolved now..

dwhly:
> Alright, I think this is the approach I'm going to go for: fewest possible changes from what we currently have in prod. Don't remove any features that we currently have in prod (e.g. the ability to use the public/only-me visibility selector to have drafts) but it also don't add any features that aren't strictly required to implement the groups requirements I've been given on the Trello cards (e.g. no All or Only Me views, no moving annotations between groups). 

To be fair, the All capability has been a part of the design since the earliest mockups, and is implemented in the https://groups4.dokku.hypothes.is/ prototype and is described in the narratives that Jake and I worked on.  Can we continue to explore it?  If it was somehow left off the trello card then that's a shame, but it's a key element of this functionality.

> This is conceptually clean. We have two separate concepts: groups and visibility. We have two separate dropdown menus: the groups menu and the visibility menu. We don't (yet) introduce the third concept of a "view" over multiple groups (All or Only Me views), we don't have things that show up with the groups in the social view menu but that are not (quite) the same thing as groups.

So, I'd argue again from a user perspective that we have a view over multiple things now, which is the default view.  Both Public and Private annotations are visible there.  "All" my things that I've created are visible in the default view, I don't have to guess where to look for stuff I've written.

Without All folks will now have to guess whether there are annotations in a given group, and check each one manually.  This is a huge negative IMO.  They won't know when they come to a page whether they've made an annotation without hunting around.

> We retain our drafting capability. Whether in the Everyone group or another group an annotation can be made private and then later edited and made group-viewable.

Great.

> There's no way to move annotations between groups, so we don't have to deal with all the questions that raises

Per your earlier suggestion, in your Option 2 above.

> Or do we allow users in the All view to create annotations in any group, but not allow them to move existing annotations between groups when editing them?

I think this is the elegant solution.  There are no additional controls or options, when you go to edit an annotation you simply cannot choose a different group than the one the annotation already belongs to.

> There's no Only Me view that shows you all of your private annotations from any group. To see your private annotations, view whichever group you made them in. Again: new feature, out of spec, brings complexities. Consider adding as a separate feature later.

Again, this was in the groups4 mockup and has been in our thinking all along, and I'd like us to consider it.   At least top level annotations that are Only Me would be visible here (I'd be fine if we don't attempt to show threads of annotations under groups that happen to have Only Me replies).

I guess my ask is that we try to implement the prototype more or less as is, with the exception of being able to move annotations between groups once created?

lenazun:
In conversation this morning with @dwhly, @judell and @nickstenning we agreed it may be better if for now we stick to the implementation we scoped in the cards.
The All view and other improvements can be added as new cards, we will follow up with those once we've implemented what we've scoped.

seanh:
@tilgovi This is (finally) ready for review. I've squashed it all into one giant commit. Sorry about this - it should probably have been a few separate commits or even separate pull requests, but I didn't realise how big this was gonna get until it was too late. I think untangling it into a few commits that makes sense now would be too time consuming. I've written a very detailed commit message explaining what it does and why though: https://github.com/hypothesis/h/commit/06929522350604eda8e019721b4641b7ac40430b

seanh:
Ah, I just discovered a bug that the visibility controls say Public / Private (instead of Foobar / Private) when creating a new annotation in group Foobar (but they work correctly editing annotations). Will fix that...

seanh:
That's the bug above fixed.

seanh:
I noticed another issue that you can change the selected group in the groups dropdown _while_ creating or editing an annotation, and the UI doesn't behave good. Since the next pull request is gonna be changing what happens when you do this anyway, I'm gonna suggest leaving it as-is until then.

dwhly:
Assume we'll match the "only me" phrasing used elsewhere for private, or make a conscious decision to change?

seanh:
It's the same "Only Me", "visible only to you", etc

seanh:
In some places it says _Private_ as well, e.g. after a private annotation is saved it shows a lock icon and _Private_ on the annotation (again, same as in current master version). Not proposing to fix it yet but I think this might be a little confusing since we now have "private" groups (meaning visible only to group members) and "private" annotations (visible only to you) and you can have a private annotation within a private group ... Making _Only You_ into a top-level group and adding _Save as draft_ as a separate feature might improve this.

dwhly:
> In some places it says Private as well, e.g. after a private annotation is saved it shows a lock icon and Private on the annotation (again, same as in current master version).

Great catch.  We should fix.  https://github.com/hypothesis/h/issues/2452

> Not proposing to fix it yet but I think this might be a little confusing since we now have "private" groups (meaning visible only to group members)

Avoiding the use of the word private for "visible only to you" annotations was a conscious change we made for exactly this reason-- private just means restricted from public view, but it doesn't == singularly private.  (IIRC We lifted this phrasing from facebook whom we assumed might have gone through the same thought process, and also therefore conditioned millions or billion(s) of folks to this phrasing for us).

tilgovi:
Rebased.

h/api/groups/auth.py
```diff
@@ -0,0 +1,50 @@
+def _authorized_for_group(effective_principals, group_hashid):
```
tilgovi: I find this a little confusing to read. Is the `if` statement supposed to test for empty string? If it's `None`, should the argument be optional? Why is the result `True` if there's no argument?

h/api/groups/logic.py
```diff
@@ -0,0 +1,30 @@
+from h.api import models
+
+
+def set_group_if_reply(annotation):
```
tilgovi: I'm not sure if this is something that needs enforcing at all on the backend. The only thing that needs to be enforced is that an annotation must only be saved to a group for which the current request has write access.

I don't see a technical reason why a reply must be in the same group. Maybe the client should be responsible for copying the field from the parent the same way it copies the `uri` and creates `references` already.

h/api/groups/models.py
```diff
@@ -65,3 +83,8 @@ def get_by_id(cls, id_):
               sa.ForeignKey('group.id'),
               nullable=False)
 )
+
+
+def includeme(config):
+    assert config.registry.settings.get("h.hashids.salt"), (
```
tilgovi: Maybe raise `pyramid.errors.ConfigurationError`?

h/api/groups/search.py
```diff
@@ -0,0 +1,44 @@
...
+            }
+        }
+
+    :param request: The Pyramid request object
+    :type request: pyramid.request.Request
+
+    """
+    # Annotations that have no 'group' field or a null value for the group
+    # field should pass through the filter.
+    should_clauses = [{'missing': {'field': 'group'}}]
```
tilgovi: Why should they pass through?

h/api/resources.py
```diff
@@ -39,7 +39,7 @@ def __getitem__(self, key):
 class Root(Resource):
     __acl__ = [
         (Allow, Authenticated, 'create'),
-        (Allow, 'group:admin', 'admin'),
+        (Allow, 'group:__admin__', 'admin'),
```
tilgovi: Huh. I thought we had removed this.

h/api/search/core.py
```diff
@@ -10,10 +10,13 @@
 log = logging.getLogger(__name__)
 
 
-def search(request_params, user=None, search_normalized_uris=False):
+def search(request, request_params, user=None, search_normalized_uris=False):
```
tilgovi: If we're passing the request through now, is there any reason to have `request_params`? We can just take `request.params`.

tilgovi: Same with `user`.

h/static/scripts/directive/annotation.coffee
```diff
@@ -163,6 +163,12 @@ AnnotationController = [
         this.render()
         this.view()
 
+    this.group = ->
+      if model.group
+        group.getGroup(model.group)
+      else
+        group.focusedGroup()
```
tilgovi: This feels wrong to me. We need a clearer upgrade path here. Right now, old annotations will appear to switch groups to match whatever the focused group is.

h/static/scripts/directive/annotation.coffee
```diff
@@ -202,6 +208,10 @@ AnnotationController = [
       switch @action
         when 'create'
           updateDomainModel(model, @annotation)
+
+          if group.focusedGroup().hashid != '__public__'
```
tilgovi: This is confusing. How could the hashid be a readable string like `__public__`. Should the client even know these are hashids? Shouldn't they just be opaque identifiers here, and the `hash` bit is an implementation detail hidden within the server?

tilgovi: And shouldn't we set the group regardless?

h/static/scripts/group-service.js
```diff
@@ -0,0 +1,62 @@
...
+    name: 'Public',
+    // We need a sentinel value to identify the public group (which is not
+    // really a group). '__public__' won't clash with any of the hashids that
+    // we use to identify real groups because the hashids never contain _.
+    hashid: '__public__'
+  };
+
+  // Return the list of available groups.
+  var groups = function() {
+    return [publicGroup].concat(session.state.groups || []);
```
tilgovi: Can we just get the public group into the session group list and then this doesn't need to be a function but just a property? Even if it meant an extra web request maybe we should mutate the current group in the session and just use that as the source of all this information. It would mean the privacy directive is a bit clearer, too, because it wouldn't need to use a getter to fetch the current group, it could just be a binding. Maybe.

tilgovi:
This is looking nice in some ways but with some rough edges. Most troubling is the authorization functions that do their own principal comparisons whereas we could have a traversal resource for a group's annotations that has an __acl__ and let Pyramid protect the views. Instead, we have to pass the request down through to logic below the view functions, code more defensively than we otherwise would need to, and re-implement access control logic that the framework should handle for us. It also means that we're not raising the usual HTTP Exception class for unauthorized and therefore the global exception view for that doesn't apply.

I think we need to talk about whether all annotations are expected to have a non-empty group field or not. We may want the `prepare` function to add the public group by default for non-private annotations or something. There's a couple places we're contorting around the uncertainty of that field.

h/static/scripts/group-list-controller.js
```diff
@@ -0,0 +1,20 @@
+'use strict';
+
+angular.module('h').controller(
```
tilgovi: I would move this line to app.coffee and just export the controller. Also, use `// @ngInject` on the constructor.

tilgovi: You can probably write the tests without any angular at that point.

h/templates/client/group_list.html
```diff
@@ -1,16 +1,29 @@
...
+      <div class="pull-left group-name">
+        <a ng-click="ctrl.focusGroup(group.hashid)">
+          <i ng-if="group.hashid == '__public__'" class="h-icon-public"></i>
+          <i ng-if="group.hashid != '__public__'" class="h-icon-group"></i>
+          {{group.name}}
+        </a>
+      </div>
+      <a ng-href="{{group.url}}" target="_blank" class="h-icon-link pull-right"
+         title="Share this group"></a>
+      <div style="clear:both;"></div>
```
tilgovi: No inline styles, please.

h/api/groups/auth.py
seanh: It's supposed to test for the `group_hashid` being `None`. The idea is that the calling code just passes in `group_hashid=annotation.get('group')`, and the annotation may not contain any 'group' and the calling code doesn't need to worry about it.

I could have the calling code pass in the whole annotation instead of just the hashid? Make this code deal with how to get the annotation's group out of the annotation

h/api/groups/logic.py
seanh: Hmm. I prefer to enforce this server-side. If I remove it then people could use the API (or other clients etc) to create replies in different groups from the parents, even if our own UI doesn't let you do that.

I'd ask the question the other way round: do you see a reason why it _should_ be possible for a reply to belong to a different group? I don't see any benefit to us enabling that, it's not in any of our user stories.

The more consistent we can make our data, and the more we can enforce the consistency server-side, the easier it's gonna be to write code to handle that data, both our own server-side code and any API client code by us or third parties. There's just less variation and different possibilities to have to deal with. Surprising configurations like a reply that has a different group from its parent are especially beneficial to avoid for this reason.

Not having to worry about replies whose parent is in a different group and how to handle those (when the view is currently filtered to one group or the other, when the user has permission to read one group but not the other...) makes code easier to write.

I know we have a way of showing a reply without its parent already (the _Message not available_) but I'd like to minimise how possible it is for that to happen, users see it as a bug.

h/api/groups/search.py
seanh: Annotations that don't belong to any group (or they are in the "Public" group if you like) have no group field, and all users are always allowed to read them (unless they're private of course but that's dealt with separately) so we always want to let them pass.

h/api/resources.py
seanh: We removed it then we put it back https://github.com/hypothesis/h/pull/2430

h/static/scripts/directive/annotation.coffee
seanh: No upgrade path needed. Pre-groups annotations have no group field and will continue not to have one, and the code treats this as meaning that they belong to the "Public" group.

Right now you can select a group and then edit a pre-groups annotation and save it and it will get the group added, true. But once the next group request is completed then when a group is selected only annotations that already have that group will be shown. Pre-groups annotations will only show when the "Public" group is selected, and when that's selected and you save an annotation it doesn't get any groups field added.

h/static/scripts/directive/annotation.coffee
seanh: We shouldn't set it. Server-side, an annotation that belongs to the "Public" group is an annotation that has no group field. Client-side this is represented as belong to a group called "Public" with hashid "__public__" but that's only because Public looks and functions like a group in the UI (in the groups dropdown etc), it has no meaning server-side (and in fact the server wouldn't let you create an annotation in this group since the group doesn't exist and you aren't auth'd to write to it) so the client strips it before sending the annotation.

There's some explanation of why this was done (and what other possibilities were considered and why they were rejected) in this commit message: <https://github.com/hypothesis/h/commit/ef1dc7741310d55783247f323ac99cd6957b20fb>:

> - Annotations that don't belong to any group / that are public simply have no
>   `group` field.
> 
>   Using `group: "public"` (actually it would have to be `__public__` so not to
>  clash with any generated hashid) and `group: "__none__"` were considered. Of
>   these I prefer `none` because it doesn't imply the existence of a "public
>   group" that doesn't really exist, it's just a sentinel value used to mark
>   annotations with no group. The advantage was that every annotation always
>   having a group field with a string value is more consistent and would make
>   coding (both h code and API client code) simpler. But:
> 
>   1. We would have to write an Elasticsearch migration script to add `group:
>   "__none__"` to all pre-existing annotations.
> 
>   2. We'd have to add code to the API to ensure that even if the client sends
>   no group field when creating an annotation, `group: "__none__"` gets
>   inserted.
> 
>   3. You still have to write code to handle annotations with group `"__none__"`
>   differently all over the place.
> 
>   I decided it was simpler just to write code to handle annotations with no
>   group field.

seanh: > This is confusing. How could the hashid be a readable string like __public__. Should the client 
> even know these are hashids? Shouldn't they just be opaque identifiers here, and the hash bit 
> is an implementation detail hidden within the server?

The client-side code needs some kind of unique way to identify the group (including the pseudo-group _Public_) that the annotation belongs to. For example when the user clicks on a group in the groups dropdown, the controller needs to communicate which group was clicked on to the service. Or when we display various bit of the annotations and their editing controls, different directives need to know what to show in certain parts of the UI. We can't use group names because they're not unique.

Hashids are the obvious choice and then we just use the fake hashid `"__public__"` for when there's no group selected / the public group is selected.

I _could_ try and find some other way to represent this. Like for example when _Public_ is selected, maybe `group.focusedGroup()` just returns `null`. But there are various different places in the code where we need to display the name _Public_ for this group and I'd want to encode that name in one central place.

seanh: > Should the client even know these are hashids? Shouldn't they just be opaque identifiers here, and the hash bit is an implementation detail hidden within the server?

At least two bits of the client code need to know about hashids:

1. When the client receives an annotation from the server it contains `group: "<hashid>"` and the client needs to take that hashid and transform it into the full group dict (including name).

2. When the client saves an annotation it has to send `group: "<hashid>"` to the server, so it needs to pull just the hashid part out of whatever representation of the group it's using.

I could have a look and see if the rest of the code could return and pass full `{name: "<name>", hashid: "<hashid>"} group objects around, and most of the code could just work with `.name` and never know about `.hashid` (but blindly pass the whole object around to the hashid does stay with it).

seanh: We could also just use the word `id` instead of `hashid` wherever the hashid appears in the client code, probably true that it doesn't need to know that it's a hashid.

h/static/scripts/group-service.js
seanh: Hmm. Are you saying the server-side (`session.py`) we should send the public group to the client? Or are you saying that the client-side (`session.js`) would insert this?

Currently the public group doesn't exist server-side / the server doesn't know anything about it. It's just a fake group used in the client-side code to deal with the fact that we have this _Public_ item in the groups dropdown but that isn't really a group.

I'm starting to doubt my decision now. I decided that annotations with no group / belonging to the public group should just have no group field, as this made the server-side code simpler (see discussion in commit message). But now I'm realising that if there were a public group server-side that was always returned in the session and that every annotation belonged to (unless it was in some other group), that might make the _client-side_ code simpler.

seanh:
> Most troubling is the authorization functions that do their own principal comparisons whereas we could have a traversal resource for a group's annotations that has an acl and let Pyramid protect the views.

I'm not sure I understand this. The views we'd need to protect are create, update, delete and read annotations. So the context object is an annotation, not a group. Maybe I could make `Annotation.__acl__` into a callable that checks the annotation's group?

But a _traversal_ resource for a group's annotations I'm not sure I understand. I could understand this if the URL were something like `/api/groups/<group_id>/annotation/<annotation_id>`. Then we can have a traversal tree with a groups resource whose keys are individual group objects with an "annotation" key whose value is a dict of all that group's annotations by ID (something like that, anyway). But we don't have that `/groups/<group_id>` in the URL so I guess I don't see how we'd use traversal here?

seanh:
> It also means that we're not raising the usual HTTP Exception class for unauthorized and therefore the global exception view for that doesn't apply.

On this I followed the existing pattern when updating annotations: you aren't allowed to change an annotation's permissions unless you have the admin permission for the annotation. If you don't, the logic function raises `RuntimeError` and the view returns one of these `_api_error()`s:

https://github.com/hypothesis/h/blob/master/h/api/logic.py#L66

https://github.com/hypothesis/h/blob/master/h/api/views.py#L184

But I could have the view functions raise some other kind of exception instead.

But I see that you're trying to get the traversal / view selection system to raise unauthorized before the view callable is even called.

seanh:
> I think we need to talk about whether all annotations are expected to have a non-empty group field 
> or not. We may want the prepare function to add the public group by default for non-private 
> annotations or something. There's a couple places we're contorting around the uncertainty of that 
> field.

Yeah, this may simplify the client-side code. As mentioned above I made annotations with no group have no group field because it meant I didn't have to write code to migrate pre-existing annotations, to stop API clients from creating annotations with no group field, and to handle `group: "__none__"` specially all over the place. Thought it was easier just to handle annotations with no group field at all specially all over the place. This probably made the server-side code simpler but the client-side code more difficult, it looks like. So we could flip that trade-off around, deal with it server-side so the client doesn't need to know anything.

h/api/groups/models.py
seanh: I've changed this but it's confusing me, if I delete the h.hashids.salt line from development.ini then when it gets to this includeme() there _is_ a h.hashids.salt in the settings! And it's not coming from any environment variable. Where does this come from??

h/api/search/core.py
seanh: The thing is it shouldn't need to pass the request, it should just pass the specific thing that it needs from the request (the hashids salt or whatever), but I wanted to tidy this up in a follow-on pull request 'cos it's gonna make a lot of noise (lot of tests to tweak I think)

h/static/scripts/group-list-controller.js
seanh: This sounds good but I might need to see an example, having trouble getting it to work..

seanh:
I've found another UI bug: Make a public annotation (in the public group) then change the groups dropdown to another group, parts of the annotation's card change to make it look like it's in the group not public.

This should go away when we do the next pr as well so gonna ignore it

h/api/groups/logic.py
JakeHartnell: > Not having to worry about replies whose parent is in a different group and how to handle those (when the view is currently filtered to one group or the other, when the user has permission to read one group but not the other...) makes code easier to write.

> I know we have a way of showing a reply without its parent already (the Message not available) but I'd like to minimise how possible it is for that to happen, users see it as a bug.

+1

seanh:
@tilgovi Take a look at this commit: <https://github.com/hypothesis/h/commit/92967ab17226efa0740d5ce5095b9d11851c5104>. It sets `group: '__none__'` before returning an annotation from the API, if the annotation has no group.

There's a bit of complexity added server-side to make this work. The group's auth functions need to allow reading/writing annotation with `group: '__none__'`, and the group's search filter needs to let them through. Also `session.py` has to always return `{'name': 'Public', 'hashid': '__none__'}` in the groups list. This is the one place where the group name Public and the hashid `__none__` are encoded, which is a bit weird, but they have to come from somewhere.

But this allows us to remove most of the special handling of annotations with no group from the client side. The templates still need to treat group `__none__` specially because it has a different icon etc.

seanh:
@tilgovi I think I've responded to all your comments, except for changing the way the auth is done. I'll start looking into Pyramid auth now, see if I can find a more Pyramid-y way to do it.

seanh:
At first glance it's not clear to me how we'd implement the permissions for groups using ACLs. To create, update or delete an annotation I need to be the annotation's creator _and_ be a member of its group, I need two principals `user:seanh` and `group:foobar`. Without testing it, just reading the docs, I don't see how to represent this as an ACL. The docs say that each ACE has to have _one_ principal and one-or-more permissions, and each view can have _one_ permission.

So I can say you need the _write_ permission to access this view, and then grant the _write_ permission to the `user:seanh` principal or to the `group:foobar` principal or to both, but I can't say only grant this permission if the request has _both_ principals.

Similarly I can't just grant a `write_annotation` permission to the user principal and a `write_group` permission to the group principal and say that the view needs both permissions, it can only need one or the other.

tilgovi:
But `__acl__` can be a callable, so you can just grant or not grant the write permission based on the conjunction.

seanh:
Will that work? From the docs it looks like callable `__acl__'s don't receive the request or effective principals in their arguments, they just have `self` (which is the context object)

seanh:
Maybe creating new annotations and updating/deleting existing ones are two different things. To create an annotation in a group you need the group's principal. To edit or delete an annotation, you need the user principal of the annotation's creator.

This doesn't seem quite right as the user could be kicked out of a group and still be able to edit their annotations within the group.

seanh:
Ah, I got a suggestion from the #pyramid irc channel that sounds like it should work: put a callable `__acl__()` on the `Annotation` resource that grants the "write" permission (for example) to the principal "user~group" where user is the acct URI of the annotation's creator and group is the hashid of the annotation's group. Then just have `effective_principals()` return "user~group" principals for each group that the user is a member of. That seems like it would work?

I have to say this whole thing seems pretty tricky compared to just having the view callable start by calling an authorization function that allows or denies, and that is just a simple Python function so can contain whatever logic you want. I'm sure it has its advantages though.

seanh:
So it seems like doing this via ACLs is within grasp. Traversal I haven't considered yet. The thing is we require two principals (which I'm now suggesting we combine into one user~group principal). The user part belongs to the annotation - it's the annotation's creator. The group part belongs to the group. As I understand it traversal will start at the context resource and work its way up through the tree until it finds the first resource that has an `__acl__`. So it _either_ uses the annotation's acl or the group's (if we organized the group object as the annotation's parent in a traversal tree). Putting the ACL on the group doesn't seem any more right to me than putting it on the annotation ... Really it belongs half to each.

So does introducing traversal into the mix make this any better? Or is just using ACLs enough?

seanh:
Actually I guess we don't even need `__acl__` to be callable. They are persisted in Es currently, right? We just need to calculate the correct `user~group` ACEs at annotation create and update time and store them.

tilgovi:
The authorization system traverses up the tree until it finds an ACE that explicitly allows or denies the permission, not the first resources that _has_ an ACL.

seanh:
Hmm, this is the bit I was reading: <http://docs.pylonsproject.org/projects/pyramid//en/latest/narr/security.html#acl-inheritance-and-location-awareness> but I guess it doesn't say what happens when it does find an ACL.

But anyway, if it does work like you're saying then I guess we could make `Annotation` deny requests but never allow them, and then have `Group` either deny or allow _if_ the request gets that far.

It seems like maybe a lot of work for no gain though?

One thing I'm still wondering about, what about when a user tries to change an annotation's group? Or change its permissions? Currently we have checks for both in the logic function itself:

* https://github.com/hypothesis/h/blob/publish-to-a-group/h/api/logic.py#L90
* https://github.com/hypothesis/h/blob/publish-to-a-group/h/api/logic.py#L94

These seem like the kinds of thing you'd call authorization errors, but we're doing them outside of the Pyramid auth system. They depend on the submitted request params not on the annotation object as it currently exists in the model, so I don't see right away how they could be done with Pyramid auth

seanh:
> But anyway, if it does work like you're saying then I guess we could make Annotation deny requests but never allow them, and then have Group either deny or allow if the request gets that far.

Actually, would that work? Seems we'd want the annotation to deny a request to edit it if the annotation's creator _isn't_ in the effective principals, but I don't think you can do that.

tilgovi:
I think the annotation ACL is straightforward. It comes right from the annotation permissions. Dealing with things like the group, but not the author, being able to delete annotations is out of scope, I think, since we don't currently have a way to leave or remove people from groups anyway.

That leaves creation, because the permission is at the collection resource. I'd be happy to see us move to a multiple-endpoint setup, but I'm curious what you think.

`POST /api/annotations` could maybe be redirected to `POST /api/public/annotations`
`POST /api/group/xyzabc/annotations` could be the group annotation collections
`POST /api/user/alice/annotations` could be private annotations

etc

tilgovi:
I'm leaning toward the multiple-endpoint solution because I've been thinking a lot about some of the trickier use cases and I think moving toward representing permissions in a more capability sort of way by enforcing authorizations during the traversal will make them easier. These sorts of use cases are things like sharing annotations with anyone who has the link, inviting read-only participants in a group, publishing an annotation to multiple groups, etc.

Fetching the right set of annotations might become conceptually simpler for the client, too.

I can try to hack on this soon, but I'd be happy to help you merge this first before going deep on these permissions changes. To that end, if we can get ACLs for modifications enforced by permissions right now (I think that should already work with the `Annotation` resource as we have it) and enforce that the groups listed are a subset of the effective principals in our view callable for create/update that will suffice.

In the future, I think these checks won't go away anyway. Even if we were to have multiple endpoints that dictate where we're writing we'd still have to sanitize the group field unless we start reserving the top level keys of our Elasticsearch documents for internal use and nesting the user input one level down.

seanh:
Ok, agreed on just moving to permissions with checks in the view functions for now, and proper auth refactor separately. I'll get on that.

Ugh, just noticed that the context resource for `create()` is `Annotations` not `Annotation`. That means that the context resource doesn't have access to the group, which is only in the request params, so the `__acl__` can't allow or deny based on the group.

We have a similar problem when someone edits an annotation and tries to change its `.group`. We want to deny the request if it contains a group in the request params that's different from the annotation's current group but the ACL system doesn't have access to the request params. The best we can do is make this a validation error rather than auth, or do an additional auth check _inside_ the view/logic function as we currently do.

I think you're right that moving to different API endpoints for each group would get around these issues, and just seems more in tune with Pyramid's auth system.

So the way I imagine this working would be:

1. Creating an annotation is posting to `/api/group/xyzabc/annotations`. In this case there's no `Annotation` object and the `Annotations` collection object has no ACL so it's a group resource corresponding to the `xyzabc` part of the URL that has the ACL that ends up being called. The view would have permission "write" on it and the group's ACL would grant the write permission to the principal for each user who's in the group.

2. So the group would not be part of the request params as it is currently, what group you're creating the annotation in is taken from the URL. If you have a group in the request params, that's a validation error.

3. Editing an annotation is posting to `/api/group/xyzabc/annotations/<id>`. In this case the ACL is on the `Annotation` object, the view has a "write" permission on it and the ACL gives the write permission to the "user~group" principal so you have to both be the user who created it and be in its group.* Again, a group in the request params is a validation error.

4. Since there's no group in the params, moving an annotation isn't possible (or rather, an API client would have to do it in two requests with a delete then a create)

5. Read

6. Delete

* So if the user leaves the group, then no one can edit that annotation. I'm not sure what behaviour we want here. I don't think we want users who're no longer in groups to still be able to edit or delete their group annotations (what about read?). But I don't think we want other members of the group to be able to do so either. Maybe in the future we'll have group admins or moderators who can edit or delete the annotations of users who're no longer in the group. (We currently could treat the user who created the group as its admin/moderator.)

seanh:
This whole thing takes the idea that we have groups and each annotation belongs to one group and encodes it write into the URLs of our API. That worries me a little, if we end up wanting other ways of organizing things in the future instead of or as well as groups.

tilgovi:
Why is the group in the request params? I thought it was in the annotation payload. Or maybe that's what you mean.

Also, we can have access to the request in the resource. The root factory receives it as a parameter. Our `create_root` function just doesn't pass it along to the sub-resources.

seanh:
Annotation payload is what I meant, yes.

tilgovi:
If we have multiple endpoints I don't think we'd even need the request object. We would factor out common logic for the CRUD but implement separate view callables with context predicates for the different groups or collections. I think if multiple are defined, the most specific takes precedence.

So, for new annotations, we just set the group to the `context.__name__` derived group. For existing annotations, we would nott let the value be mutated. Unfortunately, that might imply a load/save instead of a single update call :(.

Hmmm.... this is all very tricky.

seanh:
If we can access the request in the `Annotations` resource then maybe we can do creates according to the ACL on `Annotations` and reads, edits and writes according to the ACL on `Annotation`.

Agree that if we had multiple endpoints we would not need the request object.

seanh:
@tilgovi Here's what I've come up with:

##### Reading annotations

* On annotation create or update we set `annotation['permissions']['read']` to `['group:<hashid>']` (only for shared annotations in groups, public annotations remain `['group:__world__']` and private annotations (in groups or in public) remain `['<userid>']`).

* The existing `__acl__` on `Annotation` handles the rest for us. No auth check in the read view code anymore.


##### Search filtering

* Pass `authorization_enabled=False` to `Annotation.search_raw()`. Otherwise annotator-store's filtering means that annotations with the above group read permissions are never returned. This now means we need to do our own filtering of both group and private annotations.

* Add our own auth filter to searches. This is a filter that says that any one of the principals in `annotation['permissions']['read']` must be either `'group:__world__'` or any of `request.effective_principals`. This filters out both private annotations and annotations in groups that the user isn't a member of.

* Remove filtering on the top-level group field. We will need to add filtering on the group field back in when we want to implement focusing on a group.


##### Creating annotations

* Add `'group:__none__'` to `effective_principals` so anyone (who is logged-in) can create annotations in the none group. `'group:<hashid>'` is already there for each group the user is a member of.

* Add an `__acl__()` to the `Annotations` resource, it looks up the group from the request's JSON body and then grants the "create" permission to the group's `'group:<hashid>'` principal. So this will allow the request if the request has a `'group:<hashid>'` principal.

* Remove auth check from create view code


##### Updating and deleting annotations

* On annotation create and update we set `annotation['permissions']['update']` to `['<userid>~group:<hashid>']` (FIXME: need a separator that we know will never be in the userid or hashid) (FIXME: need to do the same for the delete permission as well I think)

* We add `'<userid>~group:<hashid>'` principals to `effective_principals` for each group the user is a member of. So we have both `'group:<hashid>'` and `'<userid>~group:<hashid>'` effective principals for each group.

* No more auth check inside the update and delete view callables


##### Client

* Hacked `permissions.coffee` in a terrible way to make the edit and delete buttons show when they should.

##### Todo

Lots of new tests need to be written.

I've realised that `'group:__none__'` means the same thing as `'group:__world__'` which is already in our annotations and in annotator-store, so we should change `__none__` to `__world__` everywhere.

I think this is reasonably close to doing things the Right Way as far as Pyramid's auth system is concerned?

To play devil's advocate I'll make an alternative suggestion, for example for editing a group annotation:

```
def update_annotation_auth(request, annotation):
    if request.authenticated_userid != annotation['user'] or annotation['group'] not in user.groups:
        raise httpexceptions.HTTPNotAuthorized()

def update_annotation(request, annotation):
    update_annotation_auth(request, annotation)
    ...
```

seanh:
Found another UI bug: if you make a shared annotation in a group, then when you edit the annotation the visibility dropdown is initially set to Only Me. Should be initially set to the group. Works fine for public annotations.

seanh:
I've just force-pushed this. About to give the code another once-over. I've done the rename of `__none__` to `__world__` everywhere, and the removing of the "user~group" principals for updating annotations.

Randall, about your commits:

1. <https://github.com/hypothesis/h/commit/6d329a69bbc9f5f997804b4907720f8acc0607f0> (moving some nipsa stuff around): can I suggest a separate pr for this? It's not groups.

2. <https://github.com/hypothesis/h/commit/19c8c524386d769d62dea529de5a529926cad47f> (removing `set_group_if_reply()`): I'd like to see some more thought go into this. Removing

* I've cherry-picked the `# -*- coding: utf-8 -*-` one
* I've manually taken some stuff from <https://github.com/hypothesis/h/commit/49f164a8e25dbc9910055d282a490287079dca56> (the -wip commit): renaming hashid to id on the client-side, simplifying template code around public vs group vs private. There's more client-side and template stuff in this -wip commit but it seemed to break the UI / I couldn't get it to work so I've left it out, at least for now.
* I've moved the groups validation stuff (set group if reply and set group permissions) into `prepare()`.
* Added all the missing tests

seanh:
I've just force-pushed this. About to give the code another once-over. I've done the rename of `__none__` to `__world__` everywhere, the removing of the "user~group" principals for updating annotations, moved the groups validation stuff (set group if reply and set group permissions) into `prepare()`, and added all the missing tests.

Randall, about your commits:

1. <https://github.com/hypothesis/h/commit/6d329a69bbc9f5f997804b4907720f8acc0607f0> (moving some nipsa stuff around): can I suggest a separate pr for this? It's not groups.

2. <https://github.com/hypothesis/h/commit/6768496f3114fbc95b51eb319b23b5106dfef2ad> (remove the `Annotations.__acl__()`) - see my comment on the commit, this just seems to let any user create annotations in any group.

2. <https://github.com/hypothesis/h/commit/19c8c524386d769d62dea529de5a529926cad47f> (removing `set_group_if_reply()`): I'd like to see some more thought go into this. Removing it means that bugs in our client or a third-party client etc could create replies whose group is different from their annotation. I think we need to consider the consequences of this first from the user's point of view, and then make sure and test what our frontend's behaviour is when this happens. Even if we get this done, it's just adding complexity to the client. Not a requirement for v1 of groups. I suggest a separate pr.

3. I've cherry-picked the `# -*- coding: utf-8 -*-` one

4. <https://github.com/hypothesis/h/commit/5b16d9dd52fd53ef5c595c670756740a436289de> (add missing `__all__()` declaration) I'm happy to cherry-pick this one back in if you want, but see my comment on the commit, I am curious to know what it's for.

5. I've manually taken some stuff from <https://github.com/hypothesis/h/commit/49f164a8e25dbc9910055d282a490287079dca56> (the -wip commit): renaming hashid to id on the client-side, simplifying template code around public vs group vs private. There's more client-side and template stuff in this -wip commit but it seemed to break the UI / I couldn't get it to work so I've left it out, at least for now.

seanh:
@tilgovi so to summarize:

1. Remove the the top-level group field. Consequences:

   - Server-side auth check when creating a new annotation needs to work by looking at the principals in the permissions field instead (the `validate_permissions()` function from here: <https://github.com/hypothesis/h/commit/49f164a8e25dbc9910055d282a490287079dca56#diff-94d385b4ed9f7e113d86771720d8c074R6>).

   - Server-side code can no longer set the correct permissions according to the group (the `set_permissions()` function in api.group.auth). This will have to be removed, and we have to rely on the  client to send the right permissions.

   - Client-side code needs to look at permissions.read to see what group the annotation is in.

2. Remove the `set_group_if_reply()` function, server just allows replies in different groups.

3. I think you also wanted to move the server-side check that prevents you from changing an annotation's group. (Just make sure you can only change it to a group you are a member of.)

I think that was it.

dwhly:
> I think you also wanted to move the server-side check that prevents you from changing an annotation's group. (Just make sure you can only change it to a group you are a member of.)

Not sure if this is related, but Sean you and I had discussed a restriction here as a way of keeping cards from disappearing from view when their group permissions were changed.  Is this a decision to drop that check?  I'm ok w/ that, but just want to voice it.

seanh:
This means we're dropping that check because Randall doesn't want it on the server, yes. But note that our user interface still won't let people move annotations, so we won't have the problem (unless users post to our api directly and move the annotations that way)

h/api/views.py
```diff
@@ -152,7 +154,10 @@ def create(request):
     user = get_user(request)
 
     # Create the annotation
-    annotation = logic.create_annotation(fields, user)
+    try:
+        annotation = logic.create_annotation(fields=fields, user=user)
+    except RuntimeError as err:
+        return _api_error(request, err.args[0], status_code=err.args[1])
```
seanh: try except maybe not needed anymore

tilgovi:
Still in flux, @dwhly.

seanh:
I'm finished with this for today, take it away @tilgovi

tilgovi:
I didn't get to a point where I can commit anything today, but I'm working on it.

I've got to a point where h.api.logic is gone and the code is pretty simple. I have multiple collection URLs for annotations.

There are a few more details to work out. If it pleases, I'll work on it one more day and if I can't make enough progress we'll merge an interim solution with a group property.

seanh:
@tilgovi Yes keep working on it. I'll find other things to do today. Just please don't go too crazy - I started working on the code in this pull request 27 days ago and it was feature complete 9 days ago, want to get this feature out soon.

With multiple collection URLs, just remember that we're inevitably gonna have features that have to retrieve the annotations from all groups (that the request is authorized to read) at once some time. We could of course just add a new API endpoint for this when we need it, but each annotation that comes out is gonna need a group field or something to identify which group it came from so that the UI can display this to the user.

tilgovi:
I made the changes [we discussed](https://github.com/hypothesis/h/pull/2432#issuecomment-135110279).

The `validate_permissions` behavior isn't tested and I haven't fixed the front-end tests, yet. I suspect the front-end may not do exactly what you want for now, but I'm also happy to let that slide by and merge since at this point I think everything is effectively behind a feature flag. Existing clients shouldn't see any difference with this merged.

seanh:
I manually tested the frontend. The only problem that I found is that if I create a private annotation in a group, then the _to <icon> <group_name>_ string from the annotation is missing. This is the only thing that indicates that the annotation is in the group (until you edit it) so I think it should be there. When we do the next pull request this will matter less, because group annotation's will only show when that group is selected in the dropdown, but then when we add things like the _All_ view this will matter more again. So this is broken currently and should be fixed I'd say

seanh:
> I manually tested the frontend. The only problem that I found is that if I create a private annotation in a group, then the to string from the annotation is missing. This is the only thing that indicates that the annotation is in the group (until you edit it) so I think it should be there. When we do the next pull request this will matter less, because group annotation's will only show when that group is selected in the dropdown, but then when we add things like the All view this will matter more again. So this is broken currently and should be fixed I'd say

Oh, wait this is more broken than I realised. If I make a private annotation in a group then:

1. The `to <icon> <group_name>` string on the annotation's card is missing
2. If I edit the annotation and I have public or a different group selected in the groups menu, then the two options in the annotation's visibility menu will be _Only Me_ and either _Public_ or whatever group is selected in the groups menu, _not_ the group that the annotation was created in.

Since the `group` field has been removed from the annotations, it seems to me like we can no longer create private annotations in groups? You can do it in the UI, but what will actually happen server-side is that we create an annotation with no `group` field and with these permissions:

```
 u'permissions': {u'admin': [u'acct:seanh@127.0.0.1'],
                  u'delete': [u'acct:seanh@127.0.0.1'],
                  u'read': [u'acct:seanh@127.0.0.1'],
                  u'update': [u'acct:seanh@127.0.0.1']},
```

Which don't record that the annotation is in the group.

The fact that the annotation is in the group isn't saved anywhere, is lost when the user reloads the page.

This is why I had a group field - as I discussed with Nick they are two separate concepts: 1. Users in group foo have permission to read this annotation, and 2. This annotation is in group foo. Features such as user-private annotations in groups, or say draft annotations in groups, demonstrate why the two concepts need to be recorded separately.

I'm not sure where this leaves us at this point? Do we need the group field back?

h/api/auth.py
```diff
@@ -13,3 +18,13 @@ def get_user(request):
                 consumer = Consumer(key)
                 return User(userid, consumer, False)
     return None
+
+
+def validate_permissions(request, permissions):
+    """Validate that the annotation's permissions are within the principals."""
```
seanh: This needs a better docstring. Maybe something more specific, like `"""Raise HTTPUnauthorized if any of the principals in permissions['read'] are not in effective_principals."""` And maybe a further description part that describes the purpose of this, that it prevents any user from making an annotation in a group that they're not a member of.

h/api/logic.py
```diff
@@ -51,20 +48,12 @@ def _anonymize_deletes(annotation):
...
         fields.pop(field, None)
 
-    # If the user is changing access permissions, check if it's allowed.
-    changing_permissions = (
-        'permissions' in fields and
-        fields['permissions'] != annotation.get('permissions', {})
-    )
-    if changing_permissions and not has_admin_permission:
-        raise RuntimeError("Not authorized to change annotation permissions.",
-                           401)  # Unauthorized
```
seanh: Nice that `validate_permissions()` makes this check, which didn't really fit into the Pyramid auth system, no longer needed.

seanh:
It looks to me like you deleted a lot of Python tests and didn't replace them with new ones. Some of these are tests of behaviours that you've deleted (e.g. the server-side validation that a reply belongs to the same group as the parent, that an update doesn't change the group of an annotation, and that an update doesn't change the permissions of an annotation), so it makes sense to delete these. But removing these assurances from the server just means that we're now relying on the client to do these things correctly, so tests for them need to be added for the relevant parts of the client-side code.

In other places it looks to me like you've deleted some Python tests that shouldn't have been deleted.

And in places where Python code behaviour has been changed or added, there needs to be tests for these.

I haven't gone through an enumerated everything that's needed, but just flagging that the tests need a fair bit of sorting out here.

h/api/resources.py
```diff
@@ -37,13 +37,14 @@ def __getitem__(self, key):
...
 
 class Root(Resource):
-    __acl__ = [
-        (Allow, Authenticated, 'create'),
-        (Allow, 'group:admin', 'admin'),
-    ]
+    def __acl__(self):
+        return [
+            (Allow, Authenticated, 'create'),
+            (Allow, Everyone, 'search'),
```
seanh: I'm not quite sure why the `(Allow, Everyone, 'search')` needed to be added?

nickstenning:
I've had a good look at this entire PR and most of the comment thread. Given that it's been open for more than a month now, and there doesn't seem to be anything substantially functionally wrong with 1e3c819, I'm going to propose we roll this PR back to that point, merge that, and start testing with the rest of the staff.

The commits following 1e3c819, to remove the need for a "group" field, may well ultimately form part of a better solution, but I can't see any particular harm in merging what we have and seeking to make improvements in later pull requests. At the moment, cbc2ade has a number of problems -- it's possible to create annotations via the API in ways that cause undefined (or at least unintended) behaviour on the frontend, and the relationship between "private" and group annotations is confusing.

tilgovi:
I think we should re-consider all of this. The separation of this work into "publish to a group" and then "focus on a group" could have been useful had the first part been _only_ the server, but given that the user experience of a group is modal this should have been done with routing. That would provide a clear environmental definition of the group membership. I don't think we should merge this, unfortunately.

nickstenning:
I'm not sure I understand the objection. Why don't you want to merge it?

nickstenning:
For the sake of clarity, I'm going to open a new PR for the original work done here, and a new branch for the additions.

PR #2433, opened 2015-08-05
User.get_by_id() refactor

seanh:


nickstenning:
This makes me a bit nervous because previously `User.get_by_id(request, "acct:seanh@someotherdomain.com")` would return `None` and now it will return your `@hypothes.is` user.

nickstenning:
I'm not entirely clear on the implications of that... and certainly, we don't have any non-`@hypothes.is` accounts in our database at the moment... so perhaps it doesn't matter.

seanh:
Agreed, this does change the behaviour in that way.

I don't know what that behaviour is for, it was obviously put there on purpose but there's no tests for it and no code that uses it (afaik), and the git logs aren't enlightening me either. I think the behaviour must have something to do with looking ahead to future situations where there are accounts from different domains, and for some reason it was thought that an app running on domain foo should only see the accounts from domain foo. I might suggest that we don't implement code for handling multiple domains yet, wait until we actually have that problem and know what we want the behaviour in that situation to be. Do we know that we're gonna have different apps running on different domains sharing the same users table and each app wanting to see only its own users? Seems a bit speculative to me, although I may be missing the point of the behaviour anyway.


The username on its own is unique'd in the database, as well. All we have in the user table is "seanh", I don't think we have any way of knowing whether that is "acct:seanh@hypothes.is" or "acct:seanh@foobar.com". It just returns the "seanh" user _if_ the domain in the given userid matches the domain in the given request object? Seems fairly odd.

And the request object really doesn't belong in model code - it can make for some pretty awkward code where we have to pass the request through multiple functions in multiple modules just so one line of code can pass it to get_by_id().


@tilgovi What do you think? Can we make this change?

tilgovi:
We can't be expecting an acct URI in this function without considering the request.

This is leftover stuff from when old auth code had an `effective_principals` callback that passed the authenticated userid to this method and from a time when our horus stuff was so muddled that sometimes we'd wind up with a `User` primary key as our authenticated userid and sometimes we'd have an acct URI.

We should maybe start by assuming the userid is an acct URI and places like `effective_principals` should compare the authority part to the request host and call `get_by_username` directly. That leaves `get_by_id` to do exactly one thing: get a user by their primary key. It would probably only be used within the accounts package and related packages.

The only other way to go about this that is sensible to me is to add an authority to the `User` database and assume that in the future we'll have rows for users whose identities we don't issue.

tilgovi:
I think the right thing to do here for the time being is to punt on any of the hard questions and just duplicate some code. Thankfully, `split_user` is a helpful utility that makes this a little easier. Everywhere we call `User.get_by_id` we should instead be calling `split_user` and then `User.get_by_username`. Possibly, we could add an `is_local_user` helper, too.

```python
userid = request.authenticated_userid
if util.is_local_account(request, userid):
  username, _ = util.split_user(userid)
  user = User.get_by_username(username)
```

seanh:
That makes sense to me - since there's no acct: URIs in the database, the model code shouldn't use them, it can work with usernames only.

We could just delete get_by_id() then - I don't think there's any code actually using it to get a user by pkey.

Do we need to check whether it's a local user like that though? That means passing the request (or at least request.domain) down to every bit of code that needs to get a user. If they just call split_user() and get_by_username() then they don't need the request.

tilgovi:
For now, we will need to pass the request to these places (and probably already do) because without it we have no way to know whether the user is local or not. The user in our database with the same user portion as the account URI is a totally different user.

tilgovi:
If we added an authority column to our `User` table we could have `get_by_account_uri` or some such thing.

nickstenning:
Under what circumstances can `request.authenticated_userid` (or any other way of getting a userid, for that matter) currently contain anything other than a userid from the current domain...?

AIUI, `request.authenticated_userid` can only be set from:

1. the cookie-keyed session, which is set [here](https://github.com/hypothesis/h/blob/9f5674f0a5c6bc0ba09e838ad18c08e726c63463/h/accounts/views.py#L128-L129).
2. the JWT, which contains a userid as defined by... the cookie-keyed session when issued.

I think that dealing in `acct:` URIs externally is good, as it means we're namespaced in the event that we want to support different identity providers.

But I don't think there's a good reason to propagate that internally while we don't have any other identity providers, so I can't see the problem with just dropping the domain part on the floor and using `User.get_by_username`.

nickstenning:
And yes, my instinct is that eventually the right way to handle this is to store rows in the user table for accounts whose identities we don't manage. But we're not there yet, so we don't need to worry about that prematurely.

seanh:
It might be interesting (one day) to track all the places where we accept or expose userids or usernames from the external and 1. Make sure those are always full acct: userids and 2. Transform them to/from userids/usernames as close to the edges as possible, work with simple usernames in as much of our code as possible. Until a time when we actually have user accounts from different domains in the db.

seanh:
I can see an argument for doing the opposite: deal in full account URIs everywhere throughout our code, and just strip the domain part off immediately before/after reading/writing from/to the database. We wouldn't need to actually check that the request.domain matches the domain of the requested userid (since this check is always `True` currently). We simply strip the domain part off before writing to the db, stick it on again after reading from the db, don't pay any attention to what the domain value actually is. So all our code deals in full userids, the only part that would know about the small usernames would be the accounts model code (and possibly only one method in there). That would make for less refactoring, when it comes time go full userids with domains in the database.

_But_ as long as we don't have domains in the db, that would mean the model code has to just use the current domain from the request, which means we have to pass the request object all the way from the very top (view callable) to the very bottom (model class) of our code, all the functions in that chain need a request or domain param, which is a pita.

tilgovi:
Under what circumstances can request.authenticated_userid (or any other way of getting a userid, for that matter) currently contain anything other than a userid from the current domain...?

Never. I said that "this is leftover stuff ... from a time when our horus stuff was so muddled that sometimes we'd wind up with a User primary key as our authenticated userid and sometimes we'd have an acct URI."

It doesn't happen that way anymore.

@seanh said:

   I can see an argument for doing the opposite: deal in full account URIs everywhere throughout our code, and just strip the domain part off immediately before/after reading/writing from/to the database.

That's exactly what we're already doing and I'd like to continue doing that. I'm just advising us not to assume all accounts are local because currently we don't so that would be a step backward. I understand it's not something we need right now but that's fine. Let's not overhaul all this right now.

seanh:
I wonder if there's an easier way for `get_by_id()` to have access to the domain, without the calling code having to pass the `request` to it?

tilgovi:
There is `pyramid.threadlocal.get_current_request()`. It's advised against in the documentation as bad practice. I think that's largely because it makes testing harder since one has to manage and set up state around the test rather than just testing the function itself more purely. Our model code already uses thread local state because the SQLAlchemy session is thread scoped and the query property uses that.

seanh:
Hmm, yeah, I also found `get_current_registry()` and the docs seem to recommend it as the way to access config settings without the request: http://docs.pylonsproject.org/projects/pyramid//en/latest/narr/environment.html?highlight=get_current_registry#adding-a-custom-setting But then as you say in multiple other places the docs strongly recommend against using it at all :(

I noticed this bit in particular (emphasis mine):

> get_current_request function should never be called because it's "easier" or "more elegant" to think about calling it than to pass a request through a series of function calls when creating some API design. Your application should instead almost certainly **pass data derived from the request around rather than relying on being able to call this function to obtain the request in places that actually have no business knowing about it**. Parameters are meant to be passed around as function arguments, this is why they exist. Don't try to "save typing" or create "nicer APIs" by using this function in the place where a request is required; this will only lead to sadness later.

This seems to say to me like we ought to be passing the domain (data derived from the request) into get_by_id(), not the whole request object. I feel that model code has no business knowing about the request.

So most of the benefit would be lost but I'd still recommend refactoring get_by_id() like so:

- Take domain param, not request
- Call util.split_user() instead of doing it manually
- Then check the domain == the domain from the userid
- Return a user or None
- Remove unused support for getting user by pkey
- The code will be simpler, in particular no weird use of regex, and no direct use of sqlalchemy either
- We'll still have to pass domain through to every function that wants to call get_by_id though, which is a pain

How does that sound?

tilgovi:
Okay, but we should rename this function. get_by_id sounds like a lookup by
primary key.

seanh:
Closing in favor of https://github.com/hypothesis/h/pull/2439

PR #2435, opened 2015-08-07
Feature flags admin

nickstenning:
Currently, the only way to change the values of feature flags is through the database. This is non-ideal.

This PR adds a very basic web interface for modifying feature flag configuration.

### DIFF
--- a/h/admin.py
+++ b/h/admin.py
@@ -1,3 +1,4 @@
+from pyramid import session
 from pyramid import view
 from pyramid import httpexceptions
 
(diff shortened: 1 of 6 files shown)


h/admin.py
```diff
@@ -16,6 +17,31 @@ def index(_):
...
+                  permission='admin')
+def features_index(_):
+    return {"features": models.Feature.all()}
+
+
+@view.view_config(route_name='admin_features',
+                  request_method='POST',
+                  permission='admin')
+def features_save(request):
+    session.check_csrf_token(request)
```
tilgovi: You can just use the `check_csrf=True` predicate on the view config. Not sure you need to test it at that point.

nickstenning: You can, but that results in a 404 for the POST, which is a bit weird.

nickstenning:
screen shot 2015-08-07 at 18 23 26

lenazun:
nice 🎌

h/admin.py
tilgovi: Ah, fair enough. That's unfortunate.

PR #2437, opened 2015-08-10
Load PDF.js viewer for PDF media types

### LINKED ISSUE #1900: [Chome] [Extension] If a sites serves us PDF, but it doesn't declare it so in the filename, we fail to inject
That's because our extension code only recognizes a PDF file from the filename.

We need to add some better detection.

@RichardLitt suggested [this code](https://github.com/BeagleLab/beagle/blob/master/js/main.js#L71-L81).

BigBlueHat:
Fixes #1900

The PDF.js Viewer + our embed.js code will be used for
all PDFs--which means our extension does not currently
consider the toggle button state for PDFs.

Three files were copied (verbatim) from
http://github.com/mozilla/pdf.js
commit sha: f943b290e67bf27ae6cc27509b480ddd150f8ebc
https://raw.githubusercontent.com/mozilla/pdf.js/f943b290e67bf27ae6cc27509b480ddd150f8ebc/extensions/chromium/preserve-referer.js
https://raw.githubusercontent.com/mozilla/pdf.js/f943b290e67bf27ae6cc27509b480ddd150f8ebc/extensions/chromium/feature-detect.js
https://raw.githubusercontent.com/mozilla/pdf.js/f943b290e67bf27ae6cc27509b480ddd150f8ebc/extensions/chromium/pdfHandler.js

The previous URL sniffing and inject code for the
PDF viewer has been remoted from our code (in favor
of this upstream detection code).

The following permissions were added as need by the upstream code:
`fileBrowserHandler`, `webRequest`, `webRequestBlocking`, and `storage`.


BigBlueHat:
Yes...I'm fixing the tests...

BigBlueHat:
@judell if you want to "hand test" this, feel free. It works. I just need to update the test code, so the robots agree. :smile:

judell:
'if you want to "hand test" this, feel free.' By sideloading the extension you're testing?

BigBlueHat:
@judell correct.

BigBlueHat:
Just sharing some (hacky) progress for anyone curious. Commit log has more.

BigBlueHat:
k. Still no tests...sorry...new territory for me...

However! It's working. ;) Could probably still use some clean-up / refactoring, but wanted another set of fingers to test this out. @judell can you teach me how to package this for side loading? Alternatively, you could download this branch and put it in developer style. :smiley: 

Lastly, here's the scenarios this *works* in:
 - currently annotating and pasting in a URL to a PDF
 - browsing to a PDF while annotating
 - toggling annotating on while viewing a URL containing `.pdf`

Here's the one it doesn't:
 - toggling annotating while viewing a URL which does not contain `.pdf`, but indeed *is* a PDF

This is due to the fact that the PDF.js Viewer loader thing is using request info extraction to do its thing--which won't work later as there's no request being triggered.

The hack around is to toggle annotating on, then reload the PDF you're viewing. I explored trying to detect if the PDF was being viewed by Chrome's viewer, but that markup seems restricted somehow--which also means we probably can't (ever?) inject `embed.js` into that...even if we wanted to...

@judell thoughts appreciated.

tilgovi:
Couldn't the detection code that happens during the request store a flag?

tilgovi:
Part of why I nuked this stuff last time is that it seems like way overkill and very tricky to test.

Questions we should ask:

- Do we care about the referer header?

- Do we care about Chrome < 35?

- Do we care about iframes?

If we can answer "no" to all three of these questions then we can re-implement this in very few, very clear, clean lines of code.

Install handler, redirect to the viewer for PDFs if the extension is enabled, and that's it.

tilgovi:
I suggest we don't care about referer because it's not terribly reliable. It's not always visible when moving between HTTP and HTTPS or vice versa (I forget the rules). Some privacy addons and proxies block it altogether. I supposed _some_ publishers might think they can reliably use it to serve content selectively, but that seems like a stretch to me.

Chrome < 35 is basically not in use anymore. Source: https://clicky.com/marketshare/global/web-browsers/google-chrome/

With iframes, we don't currently support annotating them at all with the extension, so replacing a PDF that's in a frame with our PDF viewer that injects our code seems unnecessary and unwise. Unless we treat that even more specially but injecting the guest code only we'll wind up with a frame on the page that has its own sidebar, in addition to the top-level frame's sidebar. Gross.

I say nuke all the complexity and do something dumb simple.

BigBlueHat:
I'm cool with a Chrome > 35 only approach. I'll research the options there
and report back.

judell:
> @judell can you teach me how to package this for side loading? 

As I recall, the steps were:

1. hypothesis-buildext conf/development.ini chrome --base 'http://127.0.0.1:5000

2. cd build

3. zip chrome.zip chrome

All that in my cloud-hosted Linux dev environment. 

Then:

4. download and unzip chrome.zip

5. In Chrome: load unpacked extension

6. ssh -L 5000:localhost:5000 to tunnel to the dev instance

Pretty clunky, let's find out and document a better way.

tilgovi:
If you're building it yourself on the same machine where you're running it you can just load it unpacked right from the output directory and skip the zip/unzip.

BigBlueHat:
Yeah. I was looking for how best to get a packaged copy to others to test.

Zip file works for me of it works for you @judell. :smiley:

BigBlueHat:
@tilgovi stripped out a bunch of stuff. Still works! :wink:

Still needs to record PDF media types for future use (so turning on annotation works without re-requesting the current, not-file-extension'd PDF). That's next.

This, if you're happy with the code (and lack of tests :cry: ) would solve the majority case.

Your call, though. Feedback welcome. :mailbox_with_mail:

h/browser/chrome/lib/hypothesis-chrome-extension.js
```diff
@@ -37,6 +37,7 @@
     var store = new h.TabStore(localStorage);
     var state = new h.TabState(store.all(), onTabStateChange);
     var browserAction = new h.BrowserAction(chromeBrowserAction);
+    var pdfHandler = new h.pdfHandler(state);
```
tilgovi: Can you make this UpperCamelCase since it's using the `new` keyword and should match the others?

h/browser/chrome/lib/pdfHandler.js
```diff
@@ -0,0 +1,191 @@
...
+      return;
+    }
+    if (isPdfDownloadable(details)) {
+      // Force download by ensuring that Content-Disposition: attachment is set
+      return getHeadersWithContentDispositionAttachment(details);
+    }
+
+    var viewerUrl = getViewerURL(details.url);
+
+    // Implemented in preserve-referer.js
```
tilgovi: These lines should be removed, no?

h/browser/chrome/lib/pdfHandler.js
```diff
@@ -0,0 +1,191 @@
...
+    // if we're not an active tab, do nothing
+    if (!state.isTabActive(details.tabId)) {
+      return;
+    }
+
+    if (isPdfDownloadable(details)) {
+      return;
+    }
+
+    // NOTE: The manifest file has declared an empty content script
```
tilgovi: The relevant bug is marked fixed since March 2014. This comment can probably go. It's not accurate for our packaging, anyway.

h/browser/chrome/manifest.json
```diff
@@ -20,7 +20,11 @@
     "<all_urls>",
 
     "contentSettings",
-    "tabs"
+    "tabs",
+
+    "fileBrowserHandler",
+    "webRequest", "webRequestBlocking",
+    "storage"
```
tilgovi: Not sure we're using `storage` if we don't have the referer code.

h/browser/chrome/manifest.json
```diff
@@ -20,7 +20,11 @@
     "<all_urls>",
 
     "contentSettings",
-    "tabs"
+    "tabs",
+
+    "fileBrowserHandler",
+    "webRequest", "webRequestBlocking",
```
tilgovi: Can you sort and format these normally?

h/browser/chrome/manifest.json
```diff
@@ -28,6 +32,8 @@
   "background": {
     "persistent": true,
     "scripts": [
+      "lib/pdfHandler.js",
+
```
tilgovi: I'd kill this whitespace. No need to keep everything so separate. Just put it in line with the rest.

h/browser/chrome/lib/sidebar-injector.js
```diff
@@ -103,39 +109,14 @@
...
-    function injectIntoLocalPDF(tab) {
-      return new Promise(function (resolve, reject) {
-        isAllowedFileSchemeAccess(function (isAllowed) {
-          if (isAllowed) {
-            resolve(injectIntoPDF(tab));
-          } else {
-            reject(new h.NoFileAccessError('Local file scheme access denied'));
-          }
-        });
+    function reloadPDF(tab) {
```
tilgovi: Not sure I understand why this is necessary. Doesn't it "take over" the first time?

h/browser/chrome/lib/sidebar-injector.js
```diff
@@ -103,39 +109,14 @@
...
-      } else {
-        return Promise.reject(new h.LocalFileError('Local non-PDF files are not supported'));
-      }
-    }
-
-    function injectIntoRemoteDocument(tab) {
-      return isPDFURL(tab.url) ? injectIntoPDF(tab) : injectIntoHTML(tab);
-    }
-
-    function injectIntoPDF(tab) {
```
tilgovi: Where does this code live now? I'm not seeing where we ever inject into a PDF.

h/browser/chrome/manifest.json
BigBlueHat: Yeah. Left it there as I plan to use it for the PDF URL to media type "recording." Possibly going to upgrade TabState also...later...to use it. We'll see.

h/browser/chrome/lib/sidebar-injector.js
BigBlueHat: The extension no longer "takes over" all PDFs like it used to. This is currently handling the scenario (using the old URL sniffing method) for reloading a PDF when annotating is enabled.

h/browser/chrome/lib/sidebar-injector.js
BigBlueHat: We don't. ^_^ We simply load the viewer (see `pdfHandler.js`) in which we're already embedded.

h/browser/chrome/lib/hypothesis-chrome-extension.js
BigBlueHat: Done.

h/browser/chrome/lib/pdfHandler.js
BigBlueHat: Bother...I'd taken them out...commit fale? Sorry.

h/browser/chrome/lib/pdfHandler.js
BigBlueHat: Done.

h/browser/chrome/manifest.json
BigBlueHat: Done.

h/browser/chrome/manifest.json
BigBlueHat: Done.

BigBlueHat:
Think this is ready. :smile:

This is what it *doesn't* do...but maybe could later if it's actually deemed super important:
 - doesn't detect PDF's that were loaded prior to the extension being installed (because it lacks request data)
 - doesn't handle the back navigation well (leaves a viewer URL in the back chain) after annotation is turned off while viewing a PDF
    - flow: load PDF, toggle annotation on (should have a viewer now), toggle annotation off (should have the PDF again), click back...oh look! it's the viewer!...with no annotator :stuck_out_tongue_closed_eyes: 

Also, I've not written any new tests for this stuff, and there may be some old ones failing now...

BigBlueHat:
@tilgovi this is ready for someone to pick up who's up for writing tests for it. The Chrome event stuff will (likely) be the trickiest to test, but there are some smaller functions that could be unit tested directly (though they may need some refactoring to be more easily accessible for testing).

The only stuff "beyond" what's here is outlined in this comment (above):
https://github.com/hypothesis/h/pull/2437#issuecomment-134749220

nickstenning:
Sorry, I'm going to close this. Anyone should feel free to reopen when in it's in a mergeable state.

BigBlueHat:
@nickstenning @judell just don't let this fall off the priority map. Pretty sure @dwhly had in near the top for awhile, but I can't find the cards in Trello now.

Definitely should not get forgotten as it does solve real problems (unit test "cleaning" notwithstanding).

lenazun:
@BigBlueHat It is my understanding that this just needs tests.  I think generally the dev who wrote the code should write the tests, but if you need pairing with anyone to make this happen let me know.

BigBlueHat:
I'm reopening this so I can get love from Travis.

BigBlueHat:
@nickstenning any objections to my adding this to our dependencies for running tests?
https://github.com/vitalets/sinon-chrome

It would go along way toward making this stuff actually testable--without me writing that exact same library. The license is currently unspecified (so... All Right Reserved :frowning:), but I'm asking about that https://github.com/vitalets/sinon-chrome/issues/8

nickstenning:
> @nickstenning any objections to my adding this to our dependencies for running tests?
> https://github.com/vitalets/sinon-chrome

None at all!

BigBlueHat:
This is currently blocked by :two: things:
1. a bug in `sinon-chrome` (afaict): https://travis-ci.org/hypothesis/h/builds/84589030#L510-L512
2. my borked PhantomJS setup: `ERROR [launcher]: PhantomJS failed 2 times (cannot start). Giving up.`

If someone has cycles and/or interest in helping me get `sinon-chrome` integrated into our tests, I'd be most happy. :smile_cat: Once that's unblocked writing most of these tests looks fairly trivial (/me hopes).

robertknight:
Sorry this PR has been going unloved @BigBlueHat - will try to give you a hand with it once we get groups shipped!

BigBlueHat:
Thanks @robertknight! I've not been able to figure out the `sinon-chrome` stuff, and am just back from doing W3C stuff...and currently this is near the end of my :train: at present.

However, if you could get the `sinon-chrome` stuff working, I'd be happy to help finish up writing the tests...though they're trivial enough (afaik) that it would take someone more experienced...such as yourself :wink:...a minimal amount of time to write them.

I'm happy to answer any and all questions about the code either way, though. :smile: And again, much thanks for caring! :grin:

robertknight:
@BigBlueHat - Just so you know, I haven't forgotten this. Got pushed back behind the work to sort out Via this week. I hope to take a look at it in the next sprint.

robertknight:
I'm tentatively looking at replacing this with a simpler approach which looks at the content of the page to decide whether it is the native Chrome PDF viewer, as originally proposed in #1900 

I feel like I'm missing some past discussion that may have explained why the original approach proposed in #1900 was not pursued. Please clue me in if you know.

tilgovi:
Only that I didn't have any way to assess the likelihood of that being
reliable (triggered when it should, not when it shouldn't) and felt gross
about hitching this code path to a particular markup Chrome happens to have
today for this.

robertknight:
Thanks @tilgovi . I had similar concerns about false positives/negatives. I did an analysis of ~1800 recently annotated pages and found only a small number of <embed> tags, all of which were Flash.

At the moment, determining the mime type of a displayed document in Chrome either requires monitoring all network requests (via `chrome.webRequest`) or making another network request, so in the end I decide that it was worth trying the simpler approach of matching the PDF plugin <embed>. We can always change it if it provides not to work out. There is also an upstream issue to provide a better way to detect the content type which I'm monitoring.

PR #2439, opened 2015-08-11
Refactor User.get_by_id()

seanh:
- Take domain arg instead of request
- Use util.split_user(), remove code duplication
- Remove unused capability to get a user by integer primary key
- Also fix some variable names in split_user()

### DIFF
--- a/h/accounts/models.py
+++ b/h/accounts/models.py
@@ -13,6 +13,7 @@ from sqlalchemy.ext.declarative import declared_attr
 from sqlalchemy.ext.hybrid import hybrid_property
 from sqlalchemy.sql import expression
 
+from h import util
 from h.db import Base
 
 CRYPT = cryptacular.bcrypt.BCRYPTPasswordManager()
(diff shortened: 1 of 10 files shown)


tilgovi:
Still think it should be `get_by_uri` or something that doesn't seem like it's a primary integer key lookup.

tilgovi:
Also, I find passing the domain in and then comparing it a little bit weird. How many places do we even call this, I wonder?

tilgovi:
There are several, to answer my own question.

We have had before, and could have again, simply a `user` property on the request. That property function could do `split_user` then compare to domain then return `None` or lookup by username.

Then, many of the places where we are doing `get_by_userid` might just become `request.user`?

seanh:
> Still think it should be get_by_uri or something that doesn't seem like it's a primary integer key lookup.

We consistently call these things `userid`s throughout the codebase. `get_by_userid()`?

> Also, I find passing the domain in and then comparing it a little bit weird. How many places do we even call this, I wonder?

It's bizarre, but it's what the current code does (just much less clearly). Didn't you object to me removing this check in <https://github.com/hypothesis/h/pull/2433>?

> We have had before, and could have again, simply a user property on the request. That property function could do split_user then compare to domain then return None or lookup by username.
> 
> Then, many of the places where we are doing get_by_userid might just become request.user?

I don't think  that `get_by_id()` is always used to get the authorized user though, so it might not replace all uses.

tilgovi:
I hear you on all of it. Let's do `get_by_userid`.

tilgovi:
Rename the function and feel free to merge it yourself. Everything else is fine.

tilgovi:
Okay, I'm going to merge this and do a consistency refactor.

PR #2442, opened 2015-08-11
Refactor h/api/views.py

seanh:
This code was getting a bit legacy and hard to change or write new tests
for. In particular:

- The view functions relied on some private helper functions like
  _create_annotation() and _update_annotation() that were becoming
  fairly complex in themselves but didn't have their own unit tests
  (because we don't like testing private helper functions directly)

- The code in views.py didn't have enough tests, lots of things untested

So:

- Move _create_annotation() and _update_annotation() helpers into a new
  logic.py module so they can be public functions and have their own
  unit tests

- Add lots of tests for views.py and logic.py

### DIFF
--- a/h/api/views.py
+++ b/h/api/views.py
@@ -6,20 +6,16 @@ import logging
 
 from pyramid.view import view_config
 
-from h.api import search as search_lib
 from h.api.auth import get_user
 from h.api.events import AnnotationEvent
 from h.api.models import Annotation
 from h.api.resources import Root
 from h.api.resources import Annotations
-from h.api import nipsa
+from h.api import search as search_lib
+from h.api import logic
 
 
 log = logging.getLogger(__name__)
-
-
-# These annotation fields are not to be set by the user.
-PROTECTED_FIELDS = ['created', 'updated', 'user', 'consumer', 'id']
 
 
 def api_config(**kwargs):
@@ -80,7 +76,7 @@ def search(request):
 
     # The search results are filtered for the authenticated user
     user = get_user(request)
-    results = search_lib.search(request.params,
+    results = search_lib.search(request_params=request.params,
                                 user=user,
                                 search_normalized_uris=search_normalized_uris)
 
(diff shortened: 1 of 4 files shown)


seanh:
I went for `logic.py` instead of `api.py` because we already have two `logic.py` files and it doesn't clash with `h.api`. I'm happy to see these all renamed though. I don't know what's going on with `h/api/search/core.py`.

h/api/views.py
```diff
@@ -80,7 +76,7 @@ def search(request):
 
     # The search results are filtered for the authenticated user
     user = get_user(request)
-    results = search_lib.search(request.params,
+    results = search_lib.search(request_params=request.params,
```
tilgovi: `request_params` is a positional argument.

h/api/views.py
```diff
@@ -219,7 +215,8 @@ def delete(context, request):
 
 def _publish_annotation_event(request, annotation, action):
     """Publish an event to the annotations queue for this annotation action"""
-    event = AnnotationEvent(request, annotation, action)
+    event = AnnotationEvent(
```
tilgovi: These are all positional arguments.

tilgovi:
Looks totally fine aside from the positional arguments that you're passing as keyword arguments. I dislike that because a reader might not realize that the order matters and might add a keyword argument before the first should-be-positional argument.

tilgovi:
Never mind. I may make a PR to refactor this, but I see that the tests and everything expect this to be a keyword parameter even if it's not given a default in the method definition.

tilgovi:
Okay, again, I think there's no reason not to merge this, so I'll do so. I have some tweaks, some ideas for the search queries to be improved, some thoughts about parameters, I maybe think the tests are excessively independent to the point of being painfully repetitive, but we'll talk about it all in time. I'll make some PRs after this.

seanh:
I think that if you use the names when calling a function, it doesn't matter what order the caller uses the names in, even if the function defines them as positional args? e.g. `def foobar(foo, bar, gar=None):` and then do `foobar(gar=gar, bar=bar, foo=foo)` will work correctly I think. Is there a problem I'm missing?

Anyway, yes the reason why I changed this is that it I think it makes the tests clearer and less brittle. I prefer to test that "it passes value X as the 'foo' argument to function Y" rather than "it passes value X as the third argument to function Y".

Agree the tests are pretty boilerplatey and repetitive, but not sure what to do about it. Normally you wouldn't be adding tests for so many functions all at once, at least.

seanh:
Moving more code (and therefore tests) out of these view functions and into the logic functions would probably help

PR #2446, opened 2015-08-14
Support for server sent events

tilgovi:
Add server-side support for server sent events, allowing clients to
subscribe to searches.

The implementation relies on Elasticsearch percolator documents and
ephemeral nsq topics. Each client creates an ephemeral topic and a
percolator. The percolator stores the query from the request parameters
and whenever writes occur the annotation document is percolated and
sent to all matching topics.

The TTL feature of Elasticsearch ensures that percolator documents
are cleaned up even in the event of a server fault.

### DIFF
--- a/h/api/models.py
+++ b/h/api/models.py
@@ -5,6 +5,7 @@ import cgi
 from dateutil import parser
 from annotator import annotation
 from annotator import document
+from annotator import es
 
 
 class Annotation(annotation.Annotation):
@@ -16,6 +17,7 @@ class Annotation(annotation.Annotation):
         'tags': {'type': 'string', 'analyzer': 'uni_normalizer'},
         'text': {'type': 'string', 'analyzer': 'uni_normalizer'},
         'deleted': {'type': 'boolean'},
+        'nipsa': {'type': 'boolean'},
         'uri': {
             'type': 'string',
             'index_analyzer': 'uri',
@@ -231,6 +233,14 @@ class Annotation(annotation.Annotation):
                 links.append(source)
         return links
 
+    def percolate(self):
+        # TODO: could pass 'id' instead of 'body' for efficiency.
+        # Beware that after issuing a delete, this would cause an error.
+        return self.es.conn.percolate(index=self.es.index,
+                                      doc_type=self.__type__,
+                                      body={'doc': self},
+                                      percolate_format='ids')
+
 
 class Document(document.Document):
     __analysis__ = {}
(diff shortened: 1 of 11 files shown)


tilgovi:
Rather than trying to make the streamer sane I found that an implementation from scratch was much easier. I opted for server sent events instead of websockets, but this could be extended to support websockets easily.

tilgovi:
This is all the code that's necessary for the server side of real-time search. All of streamer.py, query-parser.coffee, and stream-filter.coffee could disappear. Most of search-filter.coffee can disappear or be made vastly simpler; it just needs to parse the terms into key/value pairs.

judell:
"All of streamer.py, query-parser.coffee, and stream-filter.coffee could
disappear." !!!

seanh:
This looks _great_

seanh:
Server-sent events are really simple to implement both client- and server-side, nice.

BigBlueHat:
:heart: :exclamation: 

One question:
> The implementation relies on Elasticsearch percolator documents and ephemeral nsq topics.

Makes NSQ a development requirement, yeah? Not a problem. Just clarifying.

tilgovi:
That's not a change from the past.

BigBlueHat:
Oh. My bad. I'd thought NSQ (previously) was only required in production.
Sorry for the noise.

tilgovi:
It's required for the real-time "streamer" feature.

h/api/models.py
```diff
@@ -47,6 +48,7 @@ def __acl__(self):
         'tags': {'type': 'string', 'analyzer': 'uni_normalizer'},
         'text': {'type': 'string', 'analyzer': 'uni_normalizer'},
         'deleted': {'type': 'boolean'},
+        'nipsa': {'type': 'boolean'},
```
BigBlueHat: This is the only thing in this code that seemed out of place, but I guess this is just exposing the `nipsa` flag on the annotation so UI-side code can Do The Right Thing with it?

BigBlueHat:
Other than that one :question: about the new Annotation model `nipsa` field being in this commit, this gets a hearty :+1: from me. Please :shipit: :exclamation:

tilgovi:
It's not about exposing it to clients. It's about putting it on the mapping
explicitly rather than letting ElasticSearch infer it when the first
NIPSA'd annotation is saved.

Percolation can only work on fields that are mapped and I wanted to be sure
it was mapped so I put it in the mapping.

Whether that gets sent to clients or not is a separate question answered
elsewhere in the code. This is just about search and storage.

BigBlueHat:
@tilgovi much thanks for the explanation. What I should rebase this? It looks good to go, and I believe @seanh :+1:'d it.

tilgovi:
No tests.

seanh:
I think this looks like a great change but I wasn't under the impression that it was finished

tilgovi:
It's pretty much done. All that I want now is to check that there is enough linger time on the channel to allow reconnects without interrupting the stream. That's important since most browsers are terrible at streaming xhr and leak memory if you don't periodically reconnect when emulating EventSource.

seanh:
Doesn't this also need a matching client-side part?

tilgovi:
It can be done in two parts. This doesn't remove anything the current
client needs.

nickstenning:
Closing in favour of #2511.

PR #2451, opened 2015-08-18
Visually truncate long bodies and quotes

JakeHartnell:
- Introduce collapsing bodies and quotes which expand when you click more.
- This is done through the new excerpt directive which looks to see if text is overflowing its container and appends 'More' and 'Less' spans.
- Watch annotation controller to see if an annotation is being edited so that excerpts can be applied to newly created and updated annotation.
- Add event for threadToggleCollapse in the thread directive, and evaluate excerpts after thread expansion.
- Implements @conordelahunty's design: https://hypothes-is.slack.com/files/conor/F09H4G4LC/pasted_image_at_2015_08_24_06_38_pm.png

It looks like this:
selection_121

### DIFF
--- a/h/static/scripts/app.coffee
+++ b/h/static/scripts/app.coffee
@@ -105,6 +105,7 @@ module.exports = angular.module('h', [
 
 .directive('annotation', require('./directive/annotation'))
 .directive('deepCount', require('./directive/deep-count'))
+.directive('excerpt', require('./directive/excerpt').directive)
 .directive('formInput', require('./directive/form-input'))
 .directive('formValidate', require('./directive/form-validate'))
 .directive('groupList', require('./directive/group-list').directive)
--- a/h/static/scripts/directive/excerpt.js
+++ b/h/static/scripts/directive/excerpt.js
@@ -1,0 +1,95 @@
+'use strict';
+
+function ExcerptController() {
+  var collapsed = true;
+
+  // Enabled is a test seam: overwritten in link function.
+  this.enabled = function () { return true; };
+
+  // Overflowing is a test seam: overwritten in link function.
+  this.overflowing = function () { return false; };
+
+  // Is the excerpt collapsed? True if no-one has toggled the excerpt open
+  // and the element is overflowing.
+  this.collapsed = function () {
+    if (!collapsed) {
+      return false;
+    }
+    return this.overflowing();
+  };
+
+  this.uncollapsed = function () {
+    return !collapsed;
+  };
+
+  this.toggle = function () {
+    collapsed = !collapsed;
+  };
+
+  return this;
+}
+
+/**
+ * @ngdoc directive
+ * @name excerpt
+ * @restrict E
+ * @description This directive truncates its contents to a height specified in
+ *              CSS, and provides controls for expanding and collapsing the
+ *              resulting truncated element. For example, with the following
+ *              template HTML:
+ *
+ *                  <article class="post">
+ *                    <excerpt>
+ *                      <div class="body" ng-model="post.body"></div>
+ *                    </excerpt>
+ *                  </article>
+ *
+ *              You would need to define the allowable height of the excerpt in
+ *              CSS:
+ *
+ *                  article.post .excerpt {
+ *                    max-height: 10em;
+ *                  }
+ *
+ *              And the excerpt directive will take care of the rest.
+ *
+ *              You can selectively disable truncation by providing a boolean
+ *              expression to the `enabled` parameter, e.g.:
+ *
+ *                  <excerpt enabled="!post.inFull">...</excerpt>
+ */
+function excerpt() {
+  return {
+    controller: ExcerptController,
+    controllerAs: 'vm',
+    link: function (scope, elem, attrs, ctrl) {
+      // Test if the transcluded element is overflowing its container. We use
+      // clientHeight rather than offsetHeight because we assume you'll be using
+      // this with "overflow: hidden;" (i.e. no scrollbars) and it's usually
+      // much faster to calculate than offsetHeight (which includes scrollbars).
+      ctrl.overflowing = function overflowing() {
+        var excerpt = elem[0].querySelector('.excerpt');
+        if (!excerpt) {
+          return false;
+        }
+        return (excerpt.scrollHeight > excerpt.clientHeight);
+      };
+
+      // If the `enabled` attr was provided, we override the enabled function.
+      if (attrs.enabled) {
+        ctrl.enabled = scope.enabled;
+      }
+    },
+    scope: {
+      enabled: '&?',
+    },
+    restrict: 'E',
+    transclude: true,
+    templateUrl: 'excerpt.html',
+  };
+}
+
+module.exports = {
+  directive: excerpt,
+  Controller: ExcerptController,
+};
--- a/h/static/styles/annotations.scss
+++ b/h/static/styles/annotations.scss
@@ -37,6 +37,27 @@
 .annotation-header { margin-top: 0 }
 .annotation-footer { margin-bottom: 0 }
 
+.annotation-section {
+  .excerpt { max-height: 4.8em; }
+  .excerpt-control a {
+    font-style: italic;
+    font-family: $serif-font-family;
+    font-weight: normal;
+  }
+  .excerpt--collapsed:after {
+    height: $base-line-height;
+    @include background(linear-gradient(
+      to right,
+      $mask-start-color,
+      $mask-end-color
+    ));
+  }
+}
+
+.annotation-body {
+  .excerpt { max-height: 16.2em; }
+}
+
 .annotation-user {
   color: $text-color;
   font-weight: bold;
--- a/h/static/styles/common.scss
+++ b/h/static/styles/common.scss
@@ -5,6 +5,7 @@
 @import 'mixins/responsive';
 @import 'grid';
 @import 'annotations';
+@import 'excerpt';
 @import 'forms';
 @import 'markdown-editor';
 @import 'spinner';
--- a/h/static/styles/excerpt.scss
+++ b/h/static/styles/excerpt.scss
@@ -1,0 +1,31 @@
+@import "compass/css3/images";
+
+.excerpt {
+  position: relative;
+  overflow: hidden;
+}
+
+.excerpt--collapsed:after {
+  position: absolute;
+  bottom: 0;
+  height: $base-line-height * 2; // This controls the apparent height of the gradient.
+  width: 100%;
+  content: "";
+  pointer-events: none;
+  @include background(linear-gradient(
+    to bottom,
+    $mask-start-color,
+    $mask-end-color
+  ));
+}
+
+.excerpt--uncollapsed {
+  max-height: 100% !important;
+}
+
+.excerpt-control a {
+  display: block;
+  text-align: right;
+  font-weight: bold;
+  width: 100%;
+}
--- a/h/templates/client/excerpt.html
+++ b/h/templates/client/excerpt.html
@@ -1,0 +1,12 @@
+<div ng-transclude ng-if="!vm.enabled()"></div>
+<div ng-if="vm.enabled()">
+  <div class="excerpt"
+       ng-class="{'excerpt--uncollapsed': vm.uncollapsed(),
+                  'excerpt--collapsed': vm.collapsed()}"
+       ng-transclude></div>
+
+  <div class="excerpt-control">
+    <a ng-if="vm.collapsed()" ng-click="vm.toggle()">More</a>
+    <a ng-if="vm.uncollapsed()" ng-click="vm.toggle()">Less</a>
+  </div>
+</div>
--- a/h/features.py
+++ b/h/features.py
@@ -14,6 +14,7 @@ FEATURES = {
     'streamer': "Enable 'live streaming' for annotations via the websocket?",
     'search_normalized': "Assume all data has normalized URI fields?",
     'show_unanchored_annotations': "Show annotations that fail to anchor?",
+    'truncate_annotations': "Truncate long quotes and bodies in annotations?",
 }
 
 
--- a/h/static/scripts/directive/annotation.coffee
+++ b/h/static/scripts/directive/annotation.coffee
@@ -405,8 +405,8 @@ AnnotationController = [
 #
 ###
 module.exports = [
-  '$document',
-  ($document) ->
+  '$document', 'features'
+  ($document,   features) ->
     linkFn = (scope, elem, attrs, [ctrl, thread, threadFilter, counter]) ->
       # Observe the isSidebar attribute
       attrs.$observe 'isSidebar', (value) ->
@@ -418,6 +418,9 @@ module.exports = [
           event.preventDefault()
           scope.$evalAsync ->
             ctrl.save()
+
+      # Give template access to feature flags
+      scope.feature = features.flagEnabled
 
       scope.share = (event) ->
         $container = angular.element(event.currentTarget).parent()
--- a/h/static/scripts/directive/test/annotation-test.coffee
+++ b/h/static/scripts/directive/test/annotation-test.coffee
@@ -13,6 +13,7 @@ describe 'annotation', ->
   fakeAnnotationMapper = null
   fakeAnnotationUI = null
   fakeDrafts = null
+  fakeFeatures = null
   fakeFlash = null
   fakeGroups = null
   fakeMomentFilter = null
@@ -56,6 +57,9 @@ describe 'annotation', ->
       add: sandbox.stub()
       remove: sandbox.stub()
     }
+    fakeFeatures = {
+      flagEnabled: sandbox.stub().returns(true)
+    }
     fakeFlash = sandbox.stub()
 
     fakeMomentFilter = sandbox.stub().returns('ages ago')
@@ -92,6 +96,7 @@ describe 'annotation', ->
     $provide.value 'annotationMapper', fakeAnnotationMapper
     $provide.value 'annotationUI', fakeAnnotationUI
     $provide.value 'drafts', fakeDrafts
+    $provide.value 'features', fakeFeatures
     $provide.value 'flash', fakeFlash
     $provide.value 'momentFilter', fakeMomentFilter
     $provide.value 'permissions', fakePermissions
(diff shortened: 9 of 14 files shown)


BigBlueHat:
Needs angular wonk code review. :wink: I'm :+1: on this as it solves the > 2 year old #487 issue.

@conordelahunty could you take a look at this also? Cheers!

h/static/styles/excerpt.scss
```diff
@@ -0,0 +1,36 @@
...
+    position: absolute;
+    bottom: 0;
+    left: 87%;
+
+    &:before {
+      opacity: .8;
+      content: " ";
+      margin-left: -3em;
+      padding-right: 40px;
+      background: -webkit-gradient(linear, left top, right top, from(rgba(255, 255, 255, 0)), to(white), color-stop(50%, white));
```
tilgovi: http://compass-style.org/reference/compass/css3/images/

JakeHartnell: Good shout!

JakeHartnell:
Ok, after much ado, I'm hoping this is good enough to be merged. :pray:

h/static/scripts/app.coffee
```diff
@@ -92,6 +92,7 @@ module.exports = angular.module('h', [
 
 .directive('annotation', require('./directive/annotation'))
 .directive('deepCount', require('./directive/deep-count'))
+.directive('excerpt', require('./directive/excerpt'))
```
nickstenning: Big :+1: to trying to keep this functionality as a separate directive.

h/static/scripts/directive/excerpt.coffee
```diff
@@ -0,0 +1,35 @@
+###*
+# @ngdoc directive
+# @name excerpt
+# @restrict C
```
nickstenning: I'm not a fan of class-invoked Angular magic. Preferably this would be a attribute-invoked or even element-invoked directive (see later comments for how that might work), even if that means doubling up on class names.

h/static/scripts/directive/excerpt.coffee
```diff
@@ -0,0 +1,35 @@
...
+# @description Checks to see if text is overflowing its container.
+# If so, it prepends/appends expanders/collapsers. Note, to work
+# the element needs a max height.
+###
+module.exports = [ '$timeout', ($timeout) ->
+  link: (scope, elem, attr, ctrl) ->
+    # Check for excerpts if an annotation has been created / edited.
+    scope.$watch (-> scope.vm.editing), (editing) ->
```
nickstenning: Both this and the the `threadToggleCollapse` event are hints that this isn't really an independent directive. That means that the logic of whether or not to display the "More"/"Less" collapsers is rather complicated.

What I think we should find a way to do is include the `excerpt` directive *inside* the relevant `annotation` templates, and keep the conditional that says ("only if I'm not editing") in there and independent of the `excerpt` directive.

h/static/scripts/directive/excerpt.coffee
```diff
@@ -0,0 +1,35 @@
...
+      if editing
+        elem.find('.more').remove()
+        elem.find('.less').remove()
+      scope.excerpt()
+
+    # Check for excerpts on threadCollapseToggle event.
+    scope.$on 'threadToggleCollapse', (value) ->
+      scope.excerpt()
+
+    do scope.excerpt = ->
```
nickstenning: This is fine as far as it goes. It works! But it's not terribly "angulary" -- in that we're attaching our own events and classes manually using jQuery rather than changing some state in the directive scope and having the markup change in the Angular digest cycle.

nickstenning:
The styling/markup parts of this look absolutely fine to me. I'm a bit concerned about the business logic to do with `vm.editing` and thread collapse events making its way into the `excerpt` directive, so I'm going to try and put together a quick demo of how it might be possible to avoid that.

nickstenning:
[Here's an example](http://plnkr.co/edit/OmGhYFp1kDSiJaDPrpsW?p=preview) of what I mean. It's a bit rough around the edges but it shows how you might be able to keep the disable/enable state as a parameter to the `excerpt` directive rather than including logic about the rest of the application.

Hope that helps!

tilgovi:
:+1: @nickstenning, and I would only tweak it to prefer positive rather than negative conditionals as I think they're easier to parse: e.g. `excerpt-if` rather than `excerpt-disable-if` and default it to `true` so that it's an optional attribute.

nickstenning:
> tweak it to prefer positive rather than negative conditionals

:+1:

nickstenning:
@JakeHartnell If you want to walk through the plunker I linked above at some point just let me know. I'm happy to help in any way I can to get this finished.

JakeHartnell:
@nickstenning this is great, and I get why you like this approach. I'm having one problem that I had early on when I tried a similar approach that also used transclusion. The markdown converter also transcludes the annotation-body section and throws the following error:

```
"Error: [$compile:multidir] Multiple directives [excerpt, markdown] asking for template on: <section name="text" class="annotation-body" excerpt="" excerpt-if="!vm.editing" ng-model="vm.annotation.text" ng-readonly="!vm.editing" markdown="">
```
Any thoughts on that? I'll keep working on it, but I think if we can get past that we have a nice modular solution here. Thanks for the awesome comments and the plunker.

tilgovi:
Unfortunately, the way around that is just another wrapping element so that
these directives don't conflict by being used on the same element.

JakeHartnell:
> Unfortunately, the way around that is just another wrapping element so that these directives don't conflict by being used on the same element.

Cool, thanks.

h/static/scripts/directive/excerpt.coffee
```diff
@@ -0,0 +1,48 @@
...
+      if excerpt == null
+        return false
+      return elem[0].scrollHeight > elem[0].clientHeight
+
+    scope.toggle = ->
+      scope.collapsed = !scope.collapsed
+
+  scope:
+    enabled: '=excerptIf'
+    maxheight: '=maxheight'
```
nickstenning: This doesn't need to go here, as far as I can tell. You can add some rules like `.annotation-body .excerpt { max-height: 16.2em }` to the CSS.

h/static/scripts/directive/excerpt.coffee
```diff
@@ -0,0 +1,48 @@
...
+# @name excerpt
+# @restrict AE
+# @description Checks to see if text is overflowing its container.
+# If so, it prepends/appends expanders/collapsers. Note, to work
+# the element needs a max height.
+###
+module.exports = ->
+  link: (scope, elem, attr, ctrl) ->
+    scope.collapsed = true
+
```
nickstenning: Would suggest adding a check here that sets `scope.enabled` to true if it's undefined.

h/static/scripts/directive/excerpt.coffee
```diff
@@ -0,0 +1,48 @@
...
+      scope.collapsed = !scope.collapsed
+
+  scope:
+    enabled: '=excerptIf'
+    maxheight: '=maxheight'
+    bottomGradient: '=excerptBottomGradient'
+  restrict: 'AE'
+  transclude: true
+  template: '''
+    <ng-transclude></ng-transclude>
```
nickstenning: Not sure how this isn't breaking everything, but this should be conditional on `enabled`. i.e. `<ng-transclude ng-if="!enabled"></ng-transclude>`

h/static/scripts/directive/excerpt.coffee
```diff
@@ -0,0 +1,48 @@
...
+
+    <div class="excerpt-wrapper">
```
nickstenning: Ditto here -- this whole section, not just `.excerpt-control` -- should be `ng-if="enabled"`.

nickstenning:
Ok, looking a *lot* better. But checking it out there seem to be a few bugs -- excerpt controls seem to be displayed all the time for quotes, regardless of how long they are.

I've added a couple of comments inline, and I also would like to see the template moved into `h/templates/client/` and added to `h/templates/app.html` along with all our other client templates.

JakeHartnell:
Thanks @nickstenning for the quick review and comments! Very helpful! Sorry about the bugs, I just wanted to share what I had working so far.

h/static/scripts/directive/excerpt.coffee
JakeHartnell: Funny enough, that's what I was going to do, but worried it wouldn't be modular enough.

nickstenning:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;tkq2CkmI&#x2F;4-collapse-long-quotes-and-bodies"> Collapse long quotes and bodies</a>

tilgovi:
:-( cannot see Trello

nickstenning:
> :-( cannot see Trello

Fixed.

h/static/scripts/directive/excerpt.coffee
JakeHartnell: The reason I took it out was it's a bit to general for our use case, especially since we are transcluding the markdown editor as well:
hypothesis

In the latest version I've also added an optional `excerpt-show-controls` attribute, which I use on the annotation body to prevent this problem.

h/static/scripts/directive/excerpt.coffee
```diff
@@ -0,0 +1,35 @@
...
+# @ngdoc directive
+# @name excerpt
+# @restrict AE
+# @description Checks to see if text is overflowing its container.
+# If so, it prepends/appends expanders/collapsers. Note, to work
+# the element needs a max height.
+###
+module.exports = ->
+  link: (scope, elem, attr, ctrl) ->
+    if scope.enabled == undefined
```
nickstenning: It would be more idiomatic (and safer) to check `if typeof scope.enabled == 'undefined'`.

h/static/scripts/directive/excerpt.coffee
```diff
@@ -0,0 +1,35 @@
...
+      scope.enabled = true
+
+    if scope.showControls == undefined
```
nickstenning: Ditto.

h/static/scripts/directive/excerpt.coffee
```diff
@@ -0,0 +1,35 @@
...
+      excerpt = elem[0].querySelector('.excerpt')
+      if excerpt == null
+        return false
+      return elem[0].scrollHeight > elem[0].clientHeight
+
+    scope.toggle = ->
+      scope.collapsed = !scope.collapsed
+
+  scope:
+    excerptHeight: '=excerptHeight'
```
nickstenning: Not sure if you've finished reworking this, but as I mentioned in my previous comments any reference to max-height can come out of the directive entirely and you can add `.annotation-body .excerpt { max-height: ...; }` rules to the CSS.

nickstenning:
Other than my comment about typechecking for `undefined` and moving management of the height into CSS, this is looking functionally good.

@conordelahunty At
screen shot 2015-09-23 at 11 50 44

conordelahunty:
@nickstenning I talked to @JakeHartnell  about this yesterday. He was going to get things in a place that you were happy with and then look at the changes I requested. Hopefully ready to go soon.

nickstenning:
Okay, I've just force-pushed this branch. I've simplified the `excerpt` directive, translated it to JavaScript (although "translated" is probably a strong word in this case), and cleaned up the CSS.

@robertknight perhaps you could take a look at the general approach, and if you're happy I'll sort out the tests.

nickstenning:
Oh, and the whole thing is now behind a feature flag.

conordelahunty:
Cool, I don't know if Jake got around to make the design changes or not yet but maybe Robert can pick this up if not.

h/static/styles/excerpt.scss
```diff
@@ -0,0 +1,31 @@
+@import "compass/css3/images";
+@import "variables";
```
robertknight: IIRC this will pull in a second copy of 'variables.scss' rather than require them once. I want to use a system in future where requiring in CSS works the same way as JS modules.

h/templates/client/excerpt.html
```diff
@@ -0,0 +1,18 @@
+<div ng-transclude ng-if="!enabled()"></div>
```
robertknight: Should use `controllerAs` in the directive definition and then `vm.isEnabled()` here.

h/static/styles/annotations.scss
```diff
@@ -37,6 +37,26 @@
...
+  .excerpt { max-height: 4.8em; }
+  .excerpt-control a {
+    font-style: italic;
+    font-family: $serif-font-family;
+    font-weight: normal;
+  }
+  .excerpt--collapsed:after {
+    height: $base-line-height;
+    @include background(linear-gradient(to right,
+      rgba(255,255,255,0) 20%,
```
robertknight: Colors should always come from `variables.scss`

h/templates/client/excerpt.html
```diff
@@ -0,0 +1,18 @@
...
+<div ng-if="enabled()">
+  <div class="excerpt"
```
robertknight: Suggest using `ng-transclude` as an element rather than attribute directive here.

h/templates/client/excerpt.html
```diff
@@ -0,0 +1,18 @@
...
+       ng-class="{'excerpt--uncollapsed': !collapsed,
+                  'excerpt--collapsed': isOverflowing() && collapsed}"
+       ng-transclude></div>
+
+  <div class="excerpt-control">
+    <a ng-if="isOverflowing() && collapsed"
+       ng-click="toggle()"
+       class="more"
```
robertknight: Should use element class names here (`excerpt-control__more`)

h/static/scripts/directive/excerpt.js
```diff
@@ -0,0 +1,60 @@
...
+module.exports = function () {
+  return {
+    link: function (scope, elem, attr, ctrl) {
+      scope.collapsed = true;
+
+      if (typeof scope.enabled === 'undefined') {
+        scope.enabled = function () { return true; };
+      }
+
+      scope.isOverflowing = function () {
```
robertknight: Is there any way of changing the sidebar width at the moment? In other words, will it ever be necessary to re-evaluate whether there is an overflow?

h/static/scripts/directive/excerpt.js
```diff
@@ -0,0 +1,60 @@
...
+ *              And the excerpt directive will take care of the rest.
+ *
+ *              You can selectively disable truncation by providing a boolean
+ *              expression to the `enabled` parameter, e.g.:
+ *
+ *                  <excerpt enabled="!post.inFull">...</excerpt>
+ */
+module.exports = function () {
+  return {
+    link: function (scope, elem, attr, ctrl) {
```
robertknight: For consistency with other directives I think it would make sense to expose properties on a named object rather than directly on the scope. That can either be done with a controller or just a named object on $scope

robertknight:
@nickstenning This looks good to me.
@conordelahunty - What are the design changes you wanted?

h/static/scripts/directive/excerpt.js
nickstenning: Ok, I can refactor this to use a controller.

h/static/scripts/directive/excerpt.js
nickstenning: Yes, but that will happen -- this is deliberately a function, not a property, that will be recalculated at the digest cycle, no?

nickstenning: Hmm. It sort of works. Unfortunately resizing the sidebar doesn't currently kick off a digest cycle...

h/static/styles/annotations.scss
nickstenning: Even white and black? :stuck_out_tongue:

h/static/styles/excerpt.scss
nickstenning: I'm guessing that's not actually a problem because `variables.scss` is just variable definitions, but you're absolutely right -- it's not necessary. I'll remove it.

h/templates/client/excerpt.html
nickstenning: You know, I'd like to, but I don't think I can. I'm not entirely sure if this is because `ngIf` is a transcluding directive as well, but it simply doesn't work, and the contents of the directive are never transcluded.

h/templates/client/excerpt.html
nickstenning: These classes can actually be removed now, I think. Good catch.

nickstenning:
Thanks for the review @robertknight. I'll fix up these things, write some tests, and then I reckon we can probably merge this and fix up the design in a separate PR, now that it's feature-flagged.

h/templates/client/excerpt.html
nickstenning: I think it's this bug: https://github.com/angular/angular.js/issues/1809

h/static/styles/annotations.scss
robertknight: Yes, its a bit nit-picky. The advantage is that we'll be able to add a lint to check for colors only being defined in one place in future.

h/templates/client/excerpt.html
robertknight: OK. Let's leave as-is for now. I'll have to read up on that later.

nickstenning:
@robertknight Right, that's everything except the resize behaviour fixed. I reckon we can fix that another day. Let me know what you think.

h/static/styles/common.scss
```diff
@@ -5,6 +5,7 @@
 @import 'mixins/responsive';
 @import 'grid';
 @import 'annotations';
+@import 'excerpt';
```
robertknight: This should go under 'components' in `app.scss`, assuming it isn't used by the help page.

JakeHartnell:
Thanks @robertknight and @nickstenning!

tilgovi:
This is also not enabled in production, it seems.

dwhly:
@conordelahunty is still working w/ the team to get his tweaks to the style implemented.

JakeHartnell:
@conordelahunty sorry I didn't get around to those, but it was good to see it on stage being used. Definitely thing we need to make the annotation bodies a bit bigger, as images are cut off.
annotationbodiesneedabiggermaxheight

judell:
It's also true that for a textual annotation, More may only reveal a few
characters.

PR #2453, opened 2015-08-18
Fix #2452

### LINKED ISSUE #2452: Change "private" to "only me"
In the editor, we use "Only Me" as the indicator of an annotation that only I'm supposed to see.  Save that annotation and it's reflected as "Private" on the card itself.  

screen shot 2015-08-18 at 6 26 38 am

We should fix that.

JakeHartnell:
Meant to resolve #2452.

selection_116


JakeHartnell:
@conordelahunty, if this is ok feel free to push the button.

PR #2454, opened 2015-08-18
Lenient search

tilgovi:
This is a set of three commits designed to enabled a slightly more relaxed search experience. Commit messages reproduced in chronological order:

commit f2c65f8eefff80b78bd1345a8a956f73a4b8ffce
Author: Randall Leeds <tilgovi@hypothes.is>
Date:   Fri Aug 14 15:50:02 2015 -0700

    Treat the URI search match clauses as a filter
    
    These were being wrapped in their own boolean "should" so that only
    one URI match was required to match, despite the other clauses being
    joined under a "must".
    
    Moving these "match" clauses to a query filter makes it possible to be
    more lenient in the future by relaxing the "must" clauses into a
    "should".

commit 49dcbef846751a22cf92c1862c12030f32480260
Author: Randall Leeds <tilgovi@hypothes.is>
Date:   Fri Aug 14 15:56:14 2015 -0700

    Use "should" instead of "must" for search queries
    
    This relaxes the search queries to be allow results that match only
    some of the query. This is hopefully more intuitive for the user,
    since partial matches are returned in the absence of full matches.

commit ce4751509233f5c9bda7f9f6cf50bd81a3922cb9
Author: Randall Leeds <tilgovi@hypothes.is>
Date:   Fri Aug 14 16:00:11 2015 -0700

    Use a simple_query_string for search remainders
    
    Specifying multiple "any" terms in a search was resulting in only the
    last term being applied, since "multi_match" takes a string. Instead,
    change to the "simple_query_string" query and join the terms. This
    allows multiple terms to be specified and enables the user to use
    some advanced, but fairly safe, constructions.

### DIFF
--- a/h/api/search/query.py
+++ b/h/api/search/query.py
@@ -89,10 +89,9 @@ def build(request_params, effective_principals, userid=None,
 
     if "any" in request_params:
         matches.append({
-            "multi_match": {
+            "simple_query_string": {
                 "fields": ["quote", "tags", "text", "uri.parts", "user"],
-                "query": request_params.getall("any"),
-                "type": "cross_fields"
+                "query": ' '.join(request_params.getall("any"))
             }
         })
         del request_params["any"]
--- a/h/api/search/query.py
+++ b/h/api/search/query.py
@@ -103,7 +103,7 @@ def build(request_params, effective_principals, userid=None,
     query = {"match_all": {}}
 
     if matches:
-        query = {"bool": {"must": matches}}
+        query = {"bool": {"should": matches}}
 
     if filters:
         query = {
--- a/h/api/search/query.py
+++ b/h/api/search/query.py
@@ -73,16 +73,19 @@ def build(request_params, effective_principals, userid=None,
         }
     }]
 
-    filters = [auth_filter(effective_principals)]
+    filters = [
+        auth_filter(effective_principals),
+        nipsa.nipsa_filter(userid=userid),
+    ]
     matches = []
 
     uri_param = request_params.pop("uri", None)
     if uri_param is None:
         pass
     elif search_normalized_uris:
-        filters.append(_term_clause_for_uri(uri_param))
+        filters.append(_filter_for_uri_term(uri_param))
     else:
-        matches.append(_match_clause_for_uri(uri_param))
+        filters.append(_filter_for_uri_match(uri_param))
 
     if "any" in request_params:
         matches.append({
@@ -96,9 +99,6 @@ def build(request_params, effective_principals, userid=None,
 
     for key, value in request_params.items():
         matches.append({"match": {key: value}})
-
-    # Add a filter for "not in public site areas" considerations
-    filters.append(nipsa.nipsa_filter(userid=userid))
 
     query = {"match_all": {}}
 
@@ -121,29 +121,17 @@ def build(request_params, effective_principals, userid=None,
     }
 
 
-def _match_clause_for_uri(uristr):
+def _filter_for_uri_match(uristr):
     """Return an Elasticsearch match clause dict for the given URI."""
     uristrs = uri.expand(uristr)
-    matchers = [{"match": {"uri": u}} for u in uristrs]
+    clauses = [{"match": {"uri": u}} for u in uristrs]
 
-    if len(matchers) == 1:
-        return matchers[0]
-    return {
-        "bool": {
-            "minimum_should_match": 1,
-            "should": matchers
-        }
-    }
+    if len(clauses) == 1:
+        return {"query": clauses[0]}
+    return {"query": {"bool": {"should": clauses}}}
 
 
-def _term_clause_for_uri(uristr):
+def _filter_for_uri_term(uristr):
     """Return an Elasticsearch term clause for the given URI."""
-    uristrs = uri.expand(uristr)
-    filters = [{"term": {"target.scope": uri.normalize(u)}}
-               for u in uristrs]
-
-    if len(filters) == 1:
-        return filters[0]
-    return {
-        "or": filters
-    }
+    scopes = [uri.normalize(u) for u in uri.expand(uristr)]
+    return {"terms": {"target.scope": scopes}}
--- a/h/api/search/test/query_test.py
+++ b/h/api/search/test/query_test.py
@@ -3,7 +3,6 @@
 import pytest
 import mock
 from webob import multidict
-from pyramid import testing
 
 from h.api.search import query
 
--- a/h/api/search/test/query_test.py
+++ b/h/api/search/test/query_test.py
@@ -278,7 +278,7 @@ def test_build_with_combined_user_and_tag_query():
 
 @build_fixtures
 def test_build_with_keyword():
-    """Keywords are returned in the query dict in a "multi_match" clause."""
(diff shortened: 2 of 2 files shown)


tilgovi:
First commit changes no functionality, but just gets the unequivocally required clause out of the query and into the filter: filtering by URI.

Second commit leverages that to join the query clauses more loosely, allowing partial matches. For example, searching "user:alice user:bob" would return annotations by both Alice and Bob where before it would return no annotations.

The last commit should also not really affect existing searches but does enable "quoted phrases" and a few other features like "dogs + fleas" for annotations about fleas and dogs or "gmo -naturalnews" for annotations on GMOs that don't come from Natural News.

judell:
Sounds good, thanks. FYI there will be an upcoming requirement, related to controlled-vocabulary tagging, to be more strict on tag queries, specifically to treat tags as atomic units that don't break on white space or punctuation and only match exactly.

tilgovi:
@judell cool. I think that's orthogonal to these changes. These changes swap disjunction for intersection so it means the query

    tag:foo tag:bar

will now return things tagged 'foo' or 'bar'.

These changes have no effect on whether the individual tag values are subjected to any analysis.

h/api/search/query.py
```diff
@@ -97,29 +96,24 @@ def build(request_params, userid=None, search_normalized_uris=False):
...
-        return matchers[0]
-    return {
-        "bool": {
-            "minimum_should_match": 1,
-            "should": matchers
-        }
-    }
+    if len(clauses) == 1:
+        return {"query": clauses[0]}
+    return {"query": {"bool": {"should": clauses}}}
```
seanh: Why the "query" here? I thought this was a filter

seanh: Couldn't this just be a terms filter? `{"uri": uristrs}` <https://www.elastic.co/guide/en/elasticsearch/guide/current/_most_important_queries_and_filters.html#_terms_filter> (and in the cases where these only one uri, return a `term` filter)

h/api/search/query.py
```diff
@@ -97,29 +96,24 @@ def build(request_params, userid=None, search_normalized_uris=False):
     }
 
 
-def _match_clause_for_uri(uristr):
+def _filter_for_uri_match(uristr):
     """Return an Elasticsearch match clause dict for the given URI."""
```
seanh: `"""Return an Elasticsearch filter...` this shouldn't be using match clause dicts afaict

seanh:
The first pull request _Treat the URI search match clauses as a filter_ makes sense to me, but I think you use slightly different Elasticsearch stuff for filters than for queries, e.g. using term and terms instead of match clauses, these just do simple boolean exact matching without non-exact matches or ranking or anything, so they're fast in filters, the stuff that was match queries but is now filters is still looking too much like Es query language instead of filter language, I think.

I think I'm not understanding exactly what the `search_normalized_uris` is supposed to do. Docstring: "Whether or not to use the "uri" param to search against pre-normalized URI fields" is not entirely clear to me.

Is the idea that if I do a search for documents whose uri is "http://www.example.com" and we know (from our `uri.expand()`) that "http://www.example.com/index.html" is an equivalent URI. then my search will also find annotations of that equivalent URI?

But then it seems that we do `_filter_for_uri_term()` if `search_normalized_uris` is `True`, and `_filter_for_uri_match()` if not, but _both_ `_filter_for_uri_term()` and `_filter_for_uri_match()` are doing URI expansion? So now I'm confused as to 1. What the difference between `_filter_for_uri_term()` and `_filter_for_uri_match()` is intended to be? and 2. What `search_normalized_uris` is intended to mean?

seanh:
Let me go and read the earlier URI normalization pull request, then this'll make more sense to me...

h/api/search/query.py
```diff
@@ -56,16 +56,15 @@ def build(request_params, userid=None, search_normalized_uris=False):
     if uri_param is None:
         pass
     elif search_normalized_uris:
-        filters.append(_term_clause_for_uri(uri_param))
+        filters.append(_filter_for_uri_term(uri_param))
     else:
-        matches.append(_match_clause_for_uri(uri_param))
+        filters.append(_filter_for_uri_match(uri_param))
```
seanh: So hold on, I thought `search_normalized_uris` was a feature flag just so that we don't necessarily have to deploy the new (less tested) URI normalized search yet? But `search_normalized_uris = True` is definitely the intended behaviour on the long-term, no? i.e. I thought we'd one day remove that feature flag and remove the old `_match_clause_for_uri()` code?

But now I'm confused because here you're changing the _old_ behaviour.

seanh: Oh I see, you needed to make this change to the old URI matching in order to enable the change (in the second commit) to combining multiple search arguments with should instead of must? That makes sense

h/api/search/query.py
```diff
@@ -97,29 +96,24 @@ def build(request_params, userid=None, search_normalized_uris=False):
...
+    clauses = [{"term": {"target.source_normalized": uri.normalize(u)}}
                for u in uristrs]
 
-    if len(filters) == 1:
-        return filters[0]
+    if len(clauses) == 1:
+        return clauses[0]
     return {
-        "or": filters
+        "or": clauses
```
seanh: I know it's from the original URI normalization pr, but use of or here seems off to me. Why not use terms? Rreturn {"terms": {"target.source_normalized": [<uri_strings>]}}. This does or logic. And when there's only one URI string then just {"term": {"target.source_normalized": <uri_string>}}

seanh:
Ok, I was pretty confused at first, but now that I understand this it looks good to me :shipit:

seanh:
This is finished, right? I'm happy to merge it, but I want to wait until after <https://github.com/hypothesis/h/pull/2432> is merged cos it looks like it clashes a bit and this one is smaller / will be easier to merge

tilgovi:
Thanks for the review. Glad you figured it all out.

The query in the filter is because match is not a valid filter, but query let's you use any query as a filter.

Both code paths do expansion. The old one does match, uses a lot of analysis. The new one does term, uses python code to normalize.

I'll review your PR first thing today. Happy to keep this until that merges.

h/api/search/query.py
tilgovi: I'll change that in a separate commit.

tilgovi:
Published to a group merged. This is rebased and ready to go.

tilgovi:
Okay. I fixed the conflicts and made the improvement @seanh suggested to use a "terms" clause.

I think this is good to go. To summarize once again this change:

- Makes both code paths toggled by `search_normalized` use filters to get the annotations for the correct URI. Before, only `search_normalized` did this.
- Having freed the query clause from the need to strictly enforce URI matches, it is made more lenient by changing from "must" to "should", making relevance-ordered partial matches possible from the search API (search for an author or a tag, for instance).
- Switches the un-keyed free text input for search to use the simple query string syntax, handling quoted phrases, negation, fuzzy searches, etc. https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-simple-query-string-query.html

nickstenning:
This looks good to me. I'm actually in the process of drastically cleaning up the search package, so when I get back from lunch I'll evaluate whether it's going to be easier to rebase that on top of this or vice versa.

PR #2455, opened 2015-08-19
Get username from auth not from form

### LINKED ISSUE #2441: The change email form is broken
It's not submitting the username field, getting an error back from the server.

seanh:
When editing a user's profile or disabling a user, use the user from
request.authenticated_userid not the username submitted in the form.

Remove username from the schema so the form submission doesn't need to
contain it.

The change email form was getting a validation error back from the
server because it didn't submit the username in the form - now it
doesn't need to. Fixes #2441.

### DIFF
--- a/h/accounts/views.py
+++ b/h/accounts/views.py
@@ -467,12 +467,10 @@ class ProfileController(object):
         if err is not None:
             return err
 
-        username = appstruct['username']
-        pwd = appstruct['pwd']
+        user = User.get_by_userid(
+            self.request.domain, self.request.authenticated_userid)
 
-        # Password check
-        user = User.get_user(username, pwd)
-        if user:
+        if User.validate_user(user, appstruct['pwd']):  # Password check.
             # TODO: maybe have an explicit disabled flag in the status
             user.password = User.generate_random_password()
             self.request.session.flash(_('Account disabled.'), 'success')
--- a/h/accounts/views.py
+++ b/h/accounts/views.py
@@ -463,6 +463,9 @@ class ProfileController(object):
 
     def disable_user(self):
         """Disable the user by setting a random password."""
+        if self.request.authenticated_userid is None:
+            return httpexceptions.HTTPUnauthorized()
+
         err, appstruct = validate_form(self.form, self.request.POST.items())
         if err is not None:
             return err
--- a/h/accounts/schemas.py
+++ b/h/accounts/schemas.py
@@ -193,8 +193,6 @@ class ProfileSchema(CSRFSchema):
     This form is broken into multiple parts, for updating the email address,
     password, and subscriptions, so multiple fields are nullable.
     """
-
-    username = colander.SchemaNode(colander.String())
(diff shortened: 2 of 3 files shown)


h/accounts/views.py
```diff
@@ -467,12 +467,10 @@ def disable_user(self):
...
 
-        username = appstruct['username']
-        pwd = appstruct['pwd']
+        user = User.get_by_userid(
+            self.request.domain, self.request.authenticated_userid)
 
-        # Password check
-        user = User.get_user(username, pwd)
-        if user:
+        if User.validate_user(user, appstruct['pwd']):  # Password check.
```
seanh: Actually this password check is unnecessary now, isn't it? It looks like the disable user form once allowed users to disable any user account that they know the password of? But now, it only lets you disable your own account which you're already logged-in to

h/accounts/views.py
```diff
@@ -467,12 +467,10 @@ def disable_user(self):
         if err is not None:
             return err
 
-        username = appstruct['username']
-        pwd = appstruct['pwd']
+        user = User.get_by_userid(
```
tilgovi: Don't we need to handle a case where the user is `None` / there is no authenticated user?

tilgovi:
I thought there was a reason we had the username in the form :-/.

h/accounts/views.py
seanh: Oh yeah, good catch

seanh:
> I thought there was a reason we had the username in the form :-/.

There doesn't seem to be any need for it (and disabling accounts is currently broken on master, it gets a 400 b/c the username is required but isn't in the form).

seanh:
I'll handle the case where they're not logged-in and add a test for that, gotta run for today though

tilgovi:
Okay, thanks.

seanh:
@tilgovi I just added `@view_defaults(permission=security.Authenticated)`, I think that does it but we don't have any way to test it without functional tests

h/accounts/views.py
```diff
@@ -400,6 +400,7 @@ class AjaxRegisterController(RegisterController):
 
 
 @view_auth_defaults
+@view_defaults(permission=security.Authenticated)
```
tilgovi: This isn't right. `security.Authenticated` is a principal, not a permission. You need an ACL that allows a permission to `Authenticated` on the context and then protect the view with that permission.

seanh:
@tilgovi Ok, this does it the same way as the other views in `ProfileController` do

h/accounts/views.py
```diff
@@ -497,6 +498,9 @@ def profile(self):
         return {'model': model}
 
     def unsubscribe(self):
+        if not self.request.authenticated_userid:
+            return
```
tilgovi: Should we return unauthorized here, too?

seanh: Oops, i think i added that by accident

seanh:
Ok, done again

PR #2456, opened 2015-08-19
Add an RSS feed of /stream

seanh:
Gonna factor out `h.api_client` (which is only used by the Atom feed, afaik) and then add an RSS feed alongside the Atom feed.


seanh:
@tilgovi This is ready for review now.

The Atom code is a little odd, it was the first thing I wrote when I joined and hadn't really adopted the h style yet. There's a lot of private functions, testing of private functions, and testing of helper functions (including in separate `util.py` module) indirectly by calling the main functions instead of mocking everything.

I've written the RSS code to match the Atom code's style since they're sharing a lot of code.

This should be refactored at some point just for consistency probably but I don't think it's a priority right now.

The Atom and RSS code look really similar and there is some actual duplication of boilerplate stuff that probably isn't worth removing, but a lot of it is code that's similar but not actually the same. I'm sure it could probably be refactored into a more generic feed generator that can output either Atom or RSS and not be so much two separate code paths. But again doesn't seem too important to me right now. I've removed the most obvious straight duplication.

h/feeds/util.py
```diff
@@ -0,0 +1,93 @@
...
+    return links
+
+
+def description_from_annotation(annotation):
+    """Return a description of the given annotation.
+
+    Suitable for use in the <description> field of an <item> of an RSS feed.
+
+    """
+    def get_selection(annotation):
```
tilgovi: Seems like several parts of our code might benefit from pulling this out and just having a normalization/framing function somewhere that ensures everything is a list that should be a list.

h/feeds/test/rss_feed_test.py
```diff
@@ -0,0 +1,160 @@
...
+    """Return a mock annotation_url() function.
+
+    It just returns a hard-coded URL, enough to make the code that calls this
+    function not crash.
+
+    """
+    return mock.Mock(return_value='https://hypothes.is/a/id')
+
+
+def test__feed_from_annotations_pubDate():
```
tilgovi: You're testing a lot of private functions in this file. Is there something that could be changed to make that less nasty?

h/views.py
```diff
@@ -150,15 +157,10 @@ def stream_atom(request):
...
-    try:
-        annotations = request.api_client.get(
-            "/search", params=params)["rows"]
-    except api_client.ConnectionError as err:
-        raise httpexceptions.HTTPServiceUnavailable(err)
-    except api_client.Timeout as err:
-        raise httpexceptions.HTTPGatewayTimeout(err)
-    except api_client.APIError as err:
-        raise httpexceptions.HTTPBadGateway(err)
+    annotations = api.search_annotations(
```
tilgovi: Rather than factoring out things into a logic function maybe just call the view from `h.api.views` directly.

tilgovi:
Something feels very, very wrong to me and I didn't say so when we did Atom because I was afraid I was being too difficult.

More or less, the search API and both feeds return the same data. That suggests to me that they should have the same view with a different renderer.

Could we be creating some jinja2 extensions and inserting some renderer globals so that the templates can do all the transformation they need on the same view result?

nickstenning:
Overall this looks pretty good to me.

I don't really see a big difference between same view and different renderer versus a factored-out library function and different views, which is what we have here. I'm not too bothered either way if the code is easy enough to read.

My one complaint would be the structure of the `h.feeds` package. Specifically, I am (broken record) going to object to the testing of private functions. It's either public interface to something (in this case, the renderer) or you shouldn't be testing it. I don't think fixing this is particularly difficult:

Make `h.feeds.atom_feed` and `h.feeds.rss_feed` standalone utility modules, where the functions you need to call (e.g. `feed_from_annotations`) are the public interface, and test that. If you break it up into helper functions that make sense on their own (`feed_item_from_annotation`?) I still think that's fine to leave as public interface, even if it's only `feed_from_annotations` that uses it. Then, move the logic currently in both `render_annotations` functions into `h.feeds.renderers`, taking the opportunity to deduplicate.

(I'd also suggest the simpler `h.feeds.atom` and `h.feeds.rss` for names for those modules.)

I think the end result of that will be something where you probably *can* just have a single view with different renderers, if that's clearer, or three two-line views, if that is.

seanh:
When we did Atom the app and the API were communicating over HTTP and the plan was to split them out into separate processes, and the search view was in the API but the Atom view was in the app, that's why the app had a separate view callable for the Atom stream.

Now that the plan is for the two to continue running in the one Pyramid process and we've said that the app can import and call the API's code, though - you're right, I looked through the Pyramid docs and there's a way to wrap view callables in less lines of code by using the `wrapper` argument to `add_view()`. See https://github.com/hypothesis/h/commit/06fc63660e3b3437970cd50871df2e2511ee0a9c

I've also moved a few of the helper functions from `h.feeds.util` into `h.util`. Tests for these should be done - the util functions were not being tested directly, only indirectly by testing the feed functions, but now that they've become top-level util functions I feel they should have their own unit tests, and the feed function's unit tests should be refactored to mock them out.

I'll also look into restructuring the module to remove the private functions.

seanh:
Ok, ran out of time for today, I finished rewriting all the tests after moving things around, only thing left to do is deduplicate the two `render_annotations_to_*()` functions

h/api/logic.py
```diff
@@ -12,6 +12,17 @@
...
+    results = search_lib.search(request_params=params,
+                                user=user,
+                                search_normalized_uris=search_normalized_uris)
+
+    return {
+        'total': results['total'],
+        'rows': [search_lib.render(a) for a in results['rows']],
+    }
+
+
```
seanh: The feed code isn't actually using this anymore, but we may as well leave it in because it matches the pattern that the rest of the API views on master use currently

seanh:
I'm gonna say this is ready for review again.

I have to admit, the way feeds work ended up being a bit circuitous, this is mostly because of the desire to implement Atom and RSS as Pyramid renderers (which we introduced when we first did Atom), and the desire to not have any duplication. Here's how it works currently:

1. `h/api/views.py`: The search API returns the annotations (named "rows" in the response) as JSON.

2. `h/feeds/views.py`: The `stream_atom_wrapper()` and `stream_rss_wrapper()` functions
   wrap the response from `h/api/views.py:search()`, they return the stuff
   needed to render the Atom or RSS feed: a dict with
   annotations (`request.wrapped_response.json_body['rows']`), a couple of
   routes, and a couple of config settings.

   By wrapping the search API view we might actually be rendering the annotations to JSON here and then immediately parsing them back into dicts again, but I hope Pyramid is smart enough not to.

   The wrapper views also specify the annotations_atom and annotations_rss renderers in
   their view config.

   Some `add_view()` calls in `includeme()` set these functions up as wrapper
   views of the search API view, and then `add_route()` calls setup routes to
   them.

3. `h/feeds/renderers.py`: `AnnotationsRSSRendererFactory`,
   `AnnotationsAtomRendererFactory`, and their helper function `_render()`,
   are basically just wrapper code that make
   `h.feeds.rss.feed_from_annotations` and `h.feeds.atom.feed_from_annotations`
   available as Pyramid renderers. They just pass the values returned by the
   view wrappers into these `feed_from_annotations()` functions, and also
   insert a couple of arguments of their own (the template filenames, and
   the `annotation_url()` and `annotation_api_url()` functions).

4. `h/feeds/atom.py` and `h/feeds/rss.py` are code for turning a list of
   annotations into dicts suitable for rendering to Atom or RSS using the
   Atom or RSS templates.

It _does_ mean that the view callables are really simple, and if we want any more Atom or RSS feeds for any other lists of annotations we can add them just by adding more trivial view callables. And there's no code duplication.

seanh:
So we still have view functions for each RSS and Atom feed here, they just use Pyramid view config's wrapper argument instead of wrapping the search API view manually. I don't think there's any way to cut this down further, without giving up on the Atom and RSS renderers as generic annotations - > Atom|RSS renderers. What the view wrapper functions do essentially is get values specific to these particular RSS and Atom feeds from config and routing and pass them to the RSS and Atom renderers. I don't see any other way for `config.add_view()`to do this other than using `wrapper`, nor do I want to move getting those values into the Atom and RSS code itself because that would make the code specific to these particular feeds, not generic,

tilgovi:
By wrapping the search API view we might actually be rendering the annotations to JSON here and then immediately parsing them back into dicts again, but I hope Pyramid is smart enough not to.

It's not. If you want that you want the `decorator` argument not the `wrapper` argument.

nickstenning:
So I like the changes you've made, and I would happily merge this as-is, but as mentioned on Slack, I'm dubious of the value of implementing the Atom/RSS formatters as renderers.

AIUI, the value of a renderer is in its reusability -- specifically, that you can implement a particular format transformation once, and then use it in multiple places.

Here, the renderers are generalisable only insofar as they're generate Atom/RSS from a list of annotations. The same would be true of a simple function, but with substantially reduced cognitive overhead.

I appreciate that you probably want to close out this piece of work, and so to reiterate -- I'm happy to merge as-is -- but unless I'm missing something significant I don't see the value of using a framework-specific mechanism here when a simple function returning a string would work as well.

seanh:
I think I missed that on Slack somehow. The renderer(s) weren't introduced by this pr but I'm happy to have a go at removing them and see how much simpler that makes this

seanh:
Alright, I want to test this against a dump of production data that Nick's got me, and also run it through a feed validator, I'm sure there'll be bugs to fix. But as far as the structure of the code goes, how it is now is how I'm planning to submit it for review. Recent changes:

1. Just use functions instead of Pyramid renderers. I think Nick was right that these are not generic enough to benefit from being renderers, they can be just as useful being functions which are simpler to understand. I _did_ want to keep these as generic feed functions though, that can render any feed of annotations, not just the two feeds that we have. So to that end we have `h.feeds.render.py` which is the functional equivalent of the old `renderers.py` and is exported as the public interface of the package.

2. For the same reason, I've undone the change that I did in response to Randall to make it use Pyramid view config's `wrapper` argument instead of having the view functions as simple wrappers. They're easier to understand and just as good as simple wrappers. Looking at the diff of the commit that did this change <https://github.com/hypothesis/h/commit/3edb7ea35ea4924ed782092c12f18023b639bb70> I think it's clear that it's a (small) simplification. I don't see any problem with simple wrapper functions, or any better way to do it.

3. I've moved the util functions for getting the title, description, etc from an annotation into properties on the `Annotation` class. Now that API and app are in one process instead of over HTTP the app's feed code can (and does) deal with (dict-like) Annotation objects not just plain dicts, and I think it's much nicer to have these as properties of the objects.

seanh:
Ok, I've tested it on our most recent 750 annotations from prod (the most that I seem to be able to get my local instance to render at once) and made it valid (enough) according to the w3c validator. Ready for review, @nickstenning if you have time

h/templates/rss.xml.jinja2
```diff
@@ -0,0 +1,28 @@
+{# Generic template for rendering an RSS feed given a feed dict.
+
+The feed dict should be a logical representation of an RSS feed as a Python
+dict, including a list of dicts for the feed's items. This template will
+render the feed to RSS XML.
+
+-#}
+<?xml version="1.0"?>
+<rss version="2.0" xmlns:atom="http://www.w3.org/2005/Atom">
```
nickstenning: I don't think `xmlns:atom` is supposed to be in an RSS feed :)

h/feeds/rss.py
```diff
@@ -0,0 +1,72 @@
...
+    Return a pubDate string like 'Tue, 03 Jun 2003 09:39:21 GMT' from a
+    Hypothesis API 'created' datetime string like
+    '2015-03-11T10:43:54.537626+00:00'.
+
+    Suitable for use as the contents of a <pubDate> element in an <item>
+    element of an RSS feed.
+
+    """
+    return parser.parse(annotation['created']).strftime(
+        '%a, %d %b %Y %H:%M:%S %Z')
```
nickstenning: This is really anal, so sorry, but you probably want `%z` here, not `%Z`. Dates in RSS should be RFC822, and `%Z` spits out stuff like "UTC" which isn't strictly RFC822-compliant.

h/feeds/util.py
```diff
@@ -0,0 +1,18 @@
...
+
+    Suitable for use as the value of the <id> element of an <entry> in an
+    Atom feed, or the <guid> element of an <item> in an RSS feed.
+
+    :returns: A tag URI (RFC 4151) for use as the ID for the Atom entry.
+    :rtype: string
+
+    """
+    return u"tag:{domain},{day}:{id_}".format(
+        domain=urlparse.urlparse(annotation_url(annotation)).netloc,
```
nickstenning: Use `.hostname` not `.netloc`, as `.netloc` can contain colons (for port numbers), thus messing up the tag format.

h/feeds/util.py
```diff
@@ -0,0 +1,18 @@
...
+        day=annotation.created_day_string, id_=annotation["id"])
```
nickstenning: The date in an RSS tag is the date at which the tag format was first used by the RSS generator, not the date the annotation was created. I'd suggest just shoving a constant in this file with today's date in it and using that.

h/templates/rss.xml.jinja2
nickstenning: But I'm wrong, so ignore me. Sorry.

nickstenning:
Other than a few minor niggles, this looks good!

h/templates/rss.xml.jinja2
seanh: Yeah, the validator told me to put that in, I thought it was wrong at first too!

h/feeds/rss.py
seanh: Ah, yes the validator did say my dates were invalid, but then they looked completely the same as their example valid date! Wasn't sure exactly what was wrong

h/feeds/util.py
seanh: Another one that I saw in the validator but decided to ignore (since we wouldn't have port numbers in prod), happy to fix it though

nickstenning:
I've gone ahead and addressed those minor issues.

Issue #2167, opened 2015-04-16
Remove "click to close" feature for sidebar


tilgovi:
I removed the click to close feature in 04504dca0b36cffd57d3fd7fe071fd1ef52079b0.

I added it back in c90df05fc940587b3f4aef871538ad3e9f226b8d at the request of @judell.

However, I thought I had done it at _someone_'s request. I just can't seem to find record of that now.

In any case, I think it might be a good idea. The chevron is the obvious interaction point for open/close and it's not clear we need another way to close the sidebar. It may even annoy users who don't understand why the sidebar closes seemingly randomly while they do other things.

csillag:
IIRC it was mentioned by Sean, but it was just a "first impression" thing.
2015.04.16. 18:47 ezt írta ("Randall Leeds" <notifications@github.com>):

tilgovi:
Ah, you may be right. Thanks, @csillag.

nickstenning:
Hi there! I'm going to close this as part of a clean-up of all issues currently open on this repo that represent ideas or features rather than reports of bugs or technical chores. 

I want to be clear that **this isn't intended to say anything at all about the content of this issue**—it certainly doesn't mean we're no longer planning to do the work discussed here—just that we don't want to use GitHub issues to track feature requests or ideas, because the threads can get long and somewhat unwieldy.

If you're interested in what we are working on at the moment, you can check out [our Trello board](https://trello.com/b/o2ungX1n/current-sprint) and, for a longer-term view, [our roadmap](https://hypothes.is/roadmap/).

And, if you're interested in following up on this issue, please do continue the discussion on our [developer community mailing list](https://groups.google.com/a/list.hypothes.is/forum/#!forum/dev). You might also want to check out [our contributing guide](https://github.com/hypothesis/h/blob/master/CONTRIBUTING.rst).

PR #2457, opened 2015-08-21
Support epub integrations

JakeHartnell:
There are two things I absolutely need for the epub integrations (Readium and Epub.js):
1. An event that fire when the sidebar opens and closes.
2. An option to disable click to close. 

For the record I don't support killing "click to close" off entirely as suggested in #2167 (not enough data it's a problem IMHO), but for a dedicated reading app it's much nicer without it. Other integrators may also want to disable it.

### DIFF
--- a/h/static/scripts/annotator/guest.coffee
+++ b/h/static/scripts/annotator/guest.coffee
@@ -69,6 +69,13 @@ module.exports = class Guest extends Annotator
     this._connectAnnotationSync(@crossframe)
     this._connectAnnotationUISync(@crossframe)
 
+    if !@options.disableClickToClose
+      @wrapper = @element
+      .on 'click', (event) =>
+        if !@selectedTargets?.length
+          this.hideFrame()
+      this
+
     # Load plugins
     for own name, opts of @options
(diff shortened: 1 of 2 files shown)


JakeHartnell:
I will fix the failing test...

tilgovi:
- Let's do these separately. I might have some separate thoughts on the open/close event but the click-to-close can probably merge.
- Can you change it to simply `options: {clickToClose: true/false}}` and default it to true?
- What's with the `try` around the event dispatch?

BigBlueHat:
:+1: to splitting it up. I do think the open/close events are useful in the short term for integrations like this epub.js one to happen sooner than later.

@tilgovi we spoke about coming at this from a different "setup"-focused direction (rather than a reactionary event-based one). For now, though, I think this a) solves the immediate need to get an epub.js integration (and soon redium.js from what the @JakeHartnell says) and b) doesn't tangle us up in a big refactor or blocked on Doing It Right.

Would love to hear your thoughts on this, though.

JakeHartnell:
Closing, will post these separately.

JakeHartnell:
https://github.com/hypothesis/h/pull/2467

PR #2459, opened 2015-08-21
Pin webassets to 0.10

seanh:
webassets 0.11 seems to break hypothesis-buildext


tilgovi:
Oh, excellent! It's released!

tilgovi:
I've been wanting it for a while.

tilgovi:
`libsass` filter is something we should really look at.

tilgovi:
I'll merge this now, and then check out what's wrong

PR #2467, opened 2015-08-27
Add events for sidebar opening and closing.

JakeHartnell:


JakeHartnell:
I'm not sure why the following test is failing:
```
describe 'panend event', ->
      it 'enables pointer events and transitions on the widget', ->
        host.gestureState = {final: 0}
        host.onPan({type: 'panend'})
        assert.isFalse(host.frame.hasClass('annotator-no-transition'))
        assert.equal(host.frame.css('pointer-events'), '')
```
It throws the following error: `TypeError: '[object EventConstructor]' is not a constructor (evaluating 'new Event('hypothesisSidebarClosed')')`. In my reading trying to fix it I came across this open issue in the phantomjs repo: https://github.com/ariya/phantomjs/issues/11289

Weirdly enough if I comment out just the test mentioned above everything passes. :hear_no_evil:

tilgovi:
If you read that issue from phantomjs, I think your code should read:

```js
var event = document.createEvent('CustomEvent');
event.initCustomEvent('hypothesisSidebarOpen', false, false, null)
```

Although, I bet it can be done with jQuery more easily.

However, I wanted to separate these PRs and I'm sorry if you've spent time on this one because I would prefer to take the same approach as `clickToClose`. In other words, rather than firing events that let the publisher react to what the applications does, give them options to control what it does.

In a sense, we already have this in the form of the Toolbar being an optional plugin. Without the toolbar, you don't have the chevron that opens the sidebar. With `clickToClose: false`, there isn't a default action to close the sidebar. If you combine those things, then you almost have a sidebar that doesn't open or close itself. I think we're still opening it automatically when you create an annotation, so we should decide what to do about that.

JakeHartnell:
>  I think we're still opening it automatically when you create an annotation, so we should decide what to do about that.

And that's my problem, which is why I've wanted a custom event for this. : )

BigBlueHat:
@tilgovi @JakeHartnell would it make sense to relate the "automatic open" stuff to the `clickToClose` setting in #2468? Such that if `"clickToClose": false` then we don't automatically open the sidebar--essentially pinning it open?

Do we need a setting to default it to open? and prevent `click` closing it?

To me, the scenario seems to be:
 - As a viewer author (hi @JakeHartnell!), I want to control the visibility of the annotation sidebar, so that I control when it's open, closed, where it's positioned, etc. to avoid it messing up the loverly UI I'm building. :smile_cat:

JakeHartnell:
To restate the problem:

The sidebar opens automatically (without telling me the integrator) when a new annotation is made. THAT is what I am trying to react to with an event.

> In other words, rather than firing events that let the publisher react to what the applications does, give them options to control what it does.

I'm for this in spirit. An alternative would be if I as an integrator could create my own custom adder, but my gut tells me this seems a bit too much work at the moment... though I could look into it.

JakeHartnell:
Trying a new approach.

PR #2468, opened 2015-08-27
Add clickToClose option

JakeHartnell:
Adds an option to enable/disable click to close. Needed for epub reader integrations.

### DIFF
--- a/h/static/scripts/annotator/guest.coffee
+++ b/h/static/scripts/annotator/guest.coffee
@@ -33,10 +33,11 @@ module.exports = class Guest extends Annotator
     ".annotator-hl mouseout":            "onHighlightMouseout"
     "setVisibleHighlights": "setVisibleHighlights"
 
-  # Plugin configuration
+  # Plugin / Options configuration
   options:
     TextHighlights: {}
     TextSelection: {}
+    clickToClose: true
 
   # Anchoring module
   anchoring: require('./anchoring/html')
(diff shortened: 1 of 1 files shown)


h/static/scripts/annotator/guest.coffee
```diff
@@ -71,6 +71,13 @@ module.exports = class Guest extends Annotator
     this._connectAnnotationSync(@crossframe)
     this._connectAnnotationUISync(@crossframe)
 
+    if @options.clickToClose
+      @wrapper = @element
+      .on 'click', (event) =>
+        if !@selectedTargets?.length
+          this.hideFrame()
+      this
```
tilgovi: The `this` here was at the bottom of the function in order to make this interface fluent. That would have been clearer if it said `return this`.

Could you move this line to the end of the function again and write it as `return this`?

tilgovi: Oh, never mind. I see this is because `_setupWrapper` is gone entirely. That's no good. I think the base `Annotator` class calls `_setupWrapper`. It might be better if you keep this code in there, otherwise add `_setupWrapper` to the list of no-op functions that follows: `_setupDynamicStyle` and such.

h/static/scripts/annotator/main.js
```diff
@@ -29,7 +29,8 @@ var docs = 'https://h.readthedocs.org/en/latest/hacking/customized-embedding.htm
 var options = {
   app: jQuery('link[type="application/annotator+html"]').attr('href'),
   BucketBar: {container: '.annotator-frame', scrollables: ['body']},
-  Toolbar: {container: '.annotator-frame'}
+  Toolbar: {container: '.annotator-frame'},
+  clickToClose: true
```
tilgovi: Can you move this into the `options` object on line 37 of `guest.coffee` instead? We should maybe move all the defaults into there, but for the moment let's just not add more noise here.

JakeHartnell:
Thanks for the comments @tilgovi!

PR #2470, opened 2015-08-28
Redirect to /welcome after claim

seanh:


tilgovi:
Complication: /welcome was designed as the landing page for the extension users, and the extension opens itself there. Users who follow the claim flow are going to end up on this page without the sidebar, but with instructions saying that they can open the sidebar! What sidebar!?

seanh:
Oof

conordelahunty:
@tilgovi "the claim flow" refers to what in this case? All installs of the extension?

seanh:
There's a certain number of users (thousands, I think) who reserved their Hypothesis usernames some time ago before we'd launched anything, and gave us their email addresses. Every now and then we send out some invite emails to some of these users, telling them they can now claim their account.

I don't know what the text of these emails is (it's in our Mandrill account, I think) but it asks you to click on a link like this to claim your account: `https://hypothes.is/claim_account/n9JFQQGKBsKIOJGCkR5c7abg_FFECU09OI496R00meXzCrHMC-emWn9SgWgU2Ccy7nNTOQ8DG-qPVcmAdjmDy3sidXNlcmlkIjogImFjY3Q6c2VhbmhAMTI3LjAuMC4xIn0`.

The link takes them to this page:

screenshot from 2015-08-31 13 06 55

They enter a password, click _Claim account_, and they're redirected to ... Actually I'm not sure where they end up in production. In development I get redirected to the front page (`http://127.0.0.1:5000`) which shows the same content as the `/welcome` page except that the sidebar _does_ open, even though I don't have the Chrome extension installed. On production, the front page is a WordPress page so I think the result would be different. @tilgovi ?

Anyway if I change it to redirect to `/welcome` instead then (even in development) they get the welcome page with no sidebar as Randall says, which is wrong.

Also, it does not automatically log me in when I claim my account, at least not in development. Is that happening in prod as well?

seanh:
So I'm not sure what the solution is here. Do we have an easy way to detect whether the user has the Chrome extension installed or not, and if not embed the sidebar into the page for them?

And how does this all fit into Conor's new onboarding flows work? Presumably we want them to hear about the extension and install it, if they don't already have it.

conordelahunty:
Can we use /welcome as our sandbox page? Surely if the problem is to do with words then let's change the words? This isn't a technical issue. Unless I'm missing something, the issue is that the user arrives at the page and it says that something will be there but it isn't there. Can't we solve this by just removing the problematic copy and replacing it with something useful (Go here to get the tools or whatever?)

dwhly:
> There's a certain number of users (thousands, I think) who reserved their Hypothesis usernames some time ago before we'd launched anything, and gave us their email addresses. Every now and then we send out some invite emails to some of these users, telling them they can now claim their account.

This particular issue will go away in a week or two when the last of these emails are sent.

tilgovi:
I think we have the same problem upon any activation.

seanh:
I'm just gonna close this for now, and discuss further on the originating Trello card. Even if I get the `/welcome` page working (e.g. I get the sidebar to open, even though these users don't necessarily have the Chrome extension installed) that page doesn't make sense for these users. It says for example that step 1 is _Create an account..._ when they have just claimed and logged in to one. As Randall says the welcome page is for people who've just installed the Chrome extension but not created an account yet.

PR #2472, opened 2015-08-28
Add targets to replies

### LINKED ISSUE #2440: Replied on stream, didn't save.
So, I replied to an annotation (https://hypothes.is/a/GJzuhwo1SHeyP6rvzMQLKw) in the stream.  

This one:

screen shot 2015-08-11 at 10 30 46 am

The reply appears to take as you can see above.  
It even has an ID, this one: https://hypothes.is/a/hfsMEcGeRcuv9Uwi148ZAQ

Then I went to the page, looked at the annotation, and my reply is not showing:

screen shot 2015-08-11 at 10 31 52 am

I went back to the stream, and my reply isn't showing there now either

screen shot 2015-08-11 at 10 39 35 am

seanh:
Denormalize the 'target' from the parent annotation into all of its
reply annotations in prepare().

This means that search_normalized URI searches (which search against
annotation.target.source_normalized) will now match replies to
annotations of the searched-for URI.

Fixes #2440.

### DIFF
--- a/h/api/search/transform.py
+++ b/h/api/search/transform.py
@@ -6,6 +6,7 @@ import copy
 
 from h.api import nipsa
 from h.api import uri
+from h.api import models
 
 
 def prepare(annotation):
@@ -38,11 +39,22 @@ def render(annotation):
 
 
 def _normalize_annotation_target_uris(annotation):
-    if 'target' not in annotation:
+    targets = None
+
+    if 'references' in annotation:
+        # This annotation is a reply, use the target from its parent instead
+        # of any a target that may already have been denormalized into the
+        # reply itself.
+        parent = models.Annotation.fetch(annotation['references'][0])
+        targets = parent.get('target')
+    else:
+        # This annotation is an original.
+        targets = annotation.get('target')
+
+    if not isinstance(targets, list):
         return
-    if not isinstance(annotation['target'], list):
-        return
-    for target in annotation['target']:
+
+    for target in targets:
         if not isinstance(target, dict):
             continue
         if 'source' not in target:
@@ -50,6 +62,11 @@ def _normalize_annotation_target_uris(annotation):
         if not isinstance(target['source'], basestring):
             continue
         target['source_normalized'] = uri.normalize(target['source'])
+
+    if 'references' in annotation:
+        # Denormalize the parent's target into the reply, overwriting any
+        # existing target that was previously denormalized into the reply.
+        annotation['target'] = targets
 
 
 def _filter_target_normalized_uris(data):


seanh:
Needs tests.

It's weird to be denormalizing annotation targets into replies in a function called `_normalize_annotation_target_uris()`, should probably do it in a different function.

It looks like `hypothesis annotool conf/development.ini  prepare` should do this denormalization for preexisting replies but this doesn't seem to work for me.

seanh:
@tilgovi Thoughts on this?

tilgovi:
I've had a similar idea, but I think this doesn't work as is.

It's not semantically correct to say a reply has the same targets as its parent. More appropriate would be to say that the reply would target the parent!

Ideal, from a model perspective, might be if fetching replies were an iterative process. That would allow parents to be re-targeted without losing their replies.

From a performance perspective, that's perhaps not viable, but there are ways. We could start in-lining replies -- or a subset of replies, possibly cached -- into our annotations so as to return them in one request without complicating the pagination parameters.

But for the immediate need of solving this issue, I think I'm fine with the approach if we choose a different field than `target` to store this data.

seanh:
How about if we keep them in target fields, but we remove the targets from replies again, in `render()`? We already remove the `source_normalized` fields from all annotations in `render()`, for replies we could remove the whole `target` field.

In general I think that in our thinking about this we need to separate out two different things:

1. The data model for our web annotations API. This should be semantically correct, it should be normalized, etc.

2. The search index for our search functions. This should positively not be semantically correct or normalized, it should be convenient - whatever it needs to be, to make our search functions efficient and easy to write. The one constraint seems to be that we need to be able to recreate the normalized copy from the denormalized copy in `render()` (in the future we will have the normalized copy in postgres, but for speed I don't think we'd want to be referring to that in `render()`).

So: denormalize the parents targets into replies in `prepare()` (doing this kind of denormalization in your app is very much in line with how Elasticsearch recommend that you should use it), use them in searches, and remove them from replies again in `render()`? Users would never see them.

For pagination inlining replies does seem like it could be a good idea, agree.

tilgovi:
This is what I was using the `uri` field for: a place to keep the list of relevant URIs for search so that the target field could stay semantically pure.

### Proposal:

1. Rename `source_normalized` to `scope`. This is the Web Annotations Data Model `hasScope`.
1. Create targets for replies.
  1. Set the `scope` to the value of `uri`. This accurately reflects what the client saw.
  1. Set the `source` to the parent `URL`.
  1. Keep `references` (for now) since I think it may still be useful.

For those with the `source_normalized` feature disabled, the `uri` field, still being sent by the client, will still be indexed in the old manner. However, it will also be indexed into `scope` under the new rules.

Finally, we can change the new `source_normalized` search to use the `target.scope` field rather than `target.source_normalized`.

I think it should be safe to be destructive with `scope` and `source`, but in case we don't want to do that yet we can keep a separate, normalized field, which we may wish to rename as it would include both `scope` and `source`.

seanh:
Is that a correct usage of the scope field, in the web annotations data model? Here's the docs: <http://www.w3.org/TR/annotation-model/#scope-of-a-resource>. It defines the scope of an annotation as _the context in which it was made, in terms of the resources that the annotator was viewing or using at the time_ and gives the example of an image file that may appear on more than one HTML page. If someone annotates the image file, the target should be the image, and the particular HTML page that the image happened to be shown on would be the scope.

The way I interpreted that, the image example is the same as the example of say a news article, tweet, or blog post that may appear in multiple pages: The news site, twitter or blog front page, the article/tweet/post's own individual permalink page, tag and search results pages, etc. In all cases (if you could implement it) you'd want the target of an annotation to be the article/tweet/post itself and the scope to be the particular page that it was being viewed on at the time.

So in the case of a Hypothesis annotation, the annotation may appear in the sidebar opened on a page, in the stream, in search results, on tag pages, user profile pages, group pages, etc. Those would be the scopes.

So if I'm looking at someone's annotation on the stream and I click reply, and we consider replying to be annotating the annotation (which it seems we do, in Hypothesis), then the scope of the reply annotation would be the stream, not the document that was annotated.

It seems that the concept that we're trying to encode into the reply annotations here to fix this bug is the (normalized) URL(s) of the parent annotation's target(s). Semantically I think that's something different from the scope.

The concept of a _reply_ isn't mentioned in the Web Annotations Data Model, except as that you can encode _replying_ as the _motivation_ for an annotation, to say that the annotation _represents a reply to a previous statement, either an Annotation or another resource_, but it doesn't give you any way that I can see to encode _which_ annotation or resource it's a reply to, and this has nothing to do with targets or scopes.

**Anyway** can I suggest that moving our normalized URLs from internal data (they're currently scrubbed from the annotations in `render()`, before we send the annotations to clients) into our public data model is something for a separate pull request? It might be good to keep this one small and just fix the bug that replies aren't showing in the sidebar.

Are you against the idea of denormalizing the parent annotation's targets into the reples in `prepare()`, and removing them again in `render()`? Search will work, and the public will never see it.

seanh:
> This is what I was using the uri field for: a place to keep the list of relevant URIs for search so that the target field could stay semantically pure.

This sort of thing seems like a good idea to me. The data model of our public API should be semantically pure and conform to the web annotations data model, but the data model for our internal storage (both in postgres and in elasticsearch) should not. We just need to be able to easily generate the semantically pure version from the elasticsearch document.

What do you think about making this explicit? We could do something like add a top-level `"__denormalized__"` key to the annotation documents and stuff whatever we want to make search work conveniently in there, and just delete that key from the document before returning it to clients as an annotation?

seanh:
> It's not semantically correct to say a reply has the same targets as its parent. More appropriate would be to say that the reply would target the parent!

I'm not sure this is actually correct either. What if I want to highlight some text and annotate, but also signal that my annotation is a reply to your annotation (where you had highlighted some different text, possibly even from a different document)? Would both my highlighted text and your annotation be targets of my reply annotation?

As I said the only mention of replying I can find in the data model is that you can encode it as an annotation's _motivation_, so that way my annotation can have a different target to yours and still be a reply to yours.

These kind of discussions could easily become an endless distraction.

tilgovi:
Recapping what @seanh and I discussed earlier today.

Using `target` doesn't feel semantically correct for replies as annotation-on-annotation. The target should point to the parent. Deciding what field and usage might be semantically appropriate for exposing to clients is going to take more thought. For now, we will copy the normalized source of each target to `somekey` (tbd, possibly a top-level key) and, in the case of replies, include the parent's values.

seanh:
Replaced with https://github.com/hypothesis/h/pull/2477

PR #2475, opened 2015-09-01
Signal when there are public annotations on a page in the Chrome extension icon

seanh:


seanh:
Issues:

- [ ] Don't show anything when num. annotations is 0
- [ ] If the number is > 999 (or maybe > 99?) we need to truncate it to "999+"
- [ ] What about the user's own annotations? Should they count?
- [ ] We're fetching the page's annotations from the server to do this. After the sidebar is injected, it fetches the same annotations again to display them. Is there a good way to pass the annotations that we're just fetched into the sidebar? Alternatively, add a new API endpoint to get just the number of annotations but not the annotations themselves. It's still two HTTP requests but it's less wasteful.
- [ ] What about when the num. annotations changes as the user is viewing the page? e.g. they receive live update annotations via the websocket, or they make annotations of their own. Do we want the number on the badge to update? This may be difficult. If the badge just showed a circle when there are any annotations instead of the number, we wouldn't have this problem
- [ ] Needs tests

seanh:
Screenshot of the current look:

screenshot from 2015-09-01 13 11 31

conordelahunty:
@seanh 

"Don't show anything when num. annotations is 0"
– Agree

conordelahunty:
@seanh

"Don't show anything when num. annotations is 0"
– Agree

conordelahunty:
@seanh

"Don't show anything when num. annotations is 0"
– Agree

"If the number is > 999 (or maybe > 99?) we need to truncate it to "999+"
– Love your optimism ;) but yeah, agree

"What about the user's own annotations? Should they count?"
– I think so. If I have annotated a document and return to it, I should see that the annotations are there.

"What about when the num. annotations changes as the user is viewing the page?"
– Punt for now

dwhly:
> Don't show anything when num. annotations is 0

Agree.  I think mostly it's the red dot / square that will draw the eye, so nothing when zero.

> If the number is > 999 (or maybe > 99?) we need to truncate it to "999+"

Agree.  Whichever fits.

> What about the user's own annotations? Should they count?

Yes.  

> Do we want the number on the badge to update?  Possibly ignore this for a first version?

+1

dwhly:
@conordelahunty Didn't see your answers till just now!  Love that we agreed on all four points. :)

conordelahunty:
@dwhly A good sign!

package.json
```diff
@@ -59,7 +59,7 @@
     "proxyquire": "^1.6.0",
     "proxyquire-universal": "^1.0.8",
     "proxyquireify": "^3.0.0",
-    "sinon": "^1.15.4"
+    "sinon": "1.16.1"
```
seanh: This is to fix tests that were failing on Travis but not locally, due to Sinon differences

h/browser/chrome/karma.config.js
```diff
@@ -51,7 +63,7 @@ module.exports = function(config) {
 
 
     // web server port
-    port: 9876,
+    port: 9877,
```
seanh: Our other Karma tests (the ones for the Angular app) run on port 9876, which means that if you start these tests at the same time they will actually end up on port 9877 (but if you start them on their own they'll be on 9876).. Since I now need to know the port number for the proxy above, I had to fix this to make it constant.

h/browser/chrome/karma.config.js
```diff
@@ -32,6 +41,9 @@ module.exports = function(config) {
       '../../static/scripts/blocklist.js'
     ],
 
+    proxies: {
+      "/settings.json": "http://localhost:9877/base/test/settings.json"
```
seanh: The code expects to find `/settings.json` (which is where the file is in production), but Karma serves it at the wrong location, so proxy it.

seanh:
Ready for review. @robertknight Maybe you want to review this one?

- Sends a request to our search API on each page load

- The request is for the number of annotations only (server does not return or
  render the actual annotations)

- If > 0 then the number is shown on Chrome browser action badge. If > 999 then
  it's truncated to "999+". The button's tooltip is also set.

- If the button is showing "!" (because an error occurred) then the number of
  annotations will not overwrite this. You can create an error like this for
  testing by adding a site to the [blocklist](https://github.com/hypothesis/h/pull/2157)
  then trying to activate the extension on that site.

- The "Hypothesis is active/inactive" tooltips that are set when the user
  activates/deactivates still work, but if a number of annotations tooltip is
  set they do not override it. My reasoning is that if a badge is shown on the
  button, then the tooltip should always be an explanation of what the badge
  means, not something unrelated.

- The Chrome extension code now needs access to the API URL. To provide this
  buildext now writes a settings.json file into the extension, and the new
  settings.js module wraps it.

h/browser/chrome/karma.config.js
nickstenning: You don't need to specify the hostname. `/base/test/settings.json` appears to work, and that relieves you from the need to change the port, too.

h/browser/chrome/karma.config.js
```diff
@@ -18,6 +18,15 @@ module.exports = function(config) {
       // Polyfills for PhantomJS
       '../../../node_modules/js-polyfills/polyfill.js',
 
+      {
+        pattern: 'test/settings.json',
+        watched: true,
+        included: false,
+        served: true,
+        nocache: false
```
nickstenning: The values of `watched`, `served`, and `nocache` are the defaults and can be omitted.

h/browser/chrome/lib/hypothesis-chrome-extension.js
```diff
@@ -137,6 +137,10 @@
         browserAction.deactivate(tabId);
       }
 
+      h.Settings.then(function(settings) {
```
nickstenning: A capital letter usually implies a constructor in JavaScript land. I'd prefer it if this were `h.settings`.

h/browser/chrome/lib/browser-action.js
```diff
@@ -36,23 +36,88 @@
...
+     * @param {string} tabUrl The URL of the current tab.
+     * @param {string} serviceUrl The URL of the Hypothesis API.
+     */
+    this.updateBadge = function(tabId, tabUrl, serviceUrl) {
+      // Fetch the number of annotations of the current page from the server,
+      // and display it as a badge on the browser action button.
+      var xhr = new XMLHttpRequest();
+      xhr.onload = function() {
+        var total = 0;
+        total = JSON.parse(this.response).total;
```
nickstenning: What happens if for some reason the response contains data that isn't valid JSON, or the JSON isn't correctly formatted (e.g. has no `total` field)? Does the background.html crash? Does the extension continue to work?

If it's possible that bad data will crash the extension, this function needs to deal with that possibility.

h/browser/chrome/lib/browser-action.js
```diff
@@ -36,23 +36,88 @@
...
+                title = _('There are ' + totalString + ' annotations on ' +
+                          'this page');
+              }
+              setTitleAndBadgeText(tabId, title, totalString);
+            }
+          });
+        }
+      };
+
+      if (serviceUrl.charAt(serviceUrl.length - 1) !== '/') {
```
nickstenning: This shouldn't be dealt with here. It should either be normalised in the settings module when loaded, or it should just be defined to omit the trailing slash.

h/browser/chrome/lib/browser-action.js
```diff
@@ -36,23 +36,88 @@
...
+      xhr.onload = function() {
+        var total = 0;
+        total = JSON.parse(this.response).total;
+        if (total > 0) {
+          var totalString = total.toString();
+          if (total > 999) {
+            totalString = '999+';
+          }
+          chromeBrowserAction.getBadgeText({tabId: tabId}, function(text) {
+            if (!text) {
```
nickstenning: You've noted in the PR that this condition deals with the possibility that there's an error badge ("!") on the extension. I'd add a comment here to that effect, because it's not obvious, and I could easily see a future reorganisation losing this behaviour.

nickstenning:
Overall looks pretty good. My main concern is handling errors in `browser-action.js`.

h/browser/chrome/lib/browser-action.js
seanh: I've added a comment, though I think I already had better - there's a test for it.

h/browser/chrome/lib/hypothesis-chrome-extension.js
seanh: Done

h/browser/chrome/karma.config.js
seanh: Done

h/browser/chrome/karma.config.js
seanh: I've changed this to a relative URL, but left the port change in. Even with a relative URL it still uses the port from the Karma config file, not the port that Karma is actually running on, to fetch the file from.

h/browser/chrome/lib/browser-action.js
seanh: If there's no total, then var total ends up `undefined` which is not `> 0` and so all that happens is that no badge is shown, there's no error or crash. I bet there are ways to make it crash though..

seanh: Done

h/browser/chrome/lib/browser-action.js
seanh: Done

seanh:
Rebased, @nickstenning all done

nickstenning:
Tried it out, looks great!

My only concern is that these alert badges make me want to click on them, which of course instantly deactivates the extension...

...well, I guess we'll see if that's a problem for others!

seanh:
> My only concern is that these alert badges make me want to click on them, which of course instantly deactivates the extension...

Heh, yeah

conordelahunty:
Ooh, that's an interesting one. Didn't think of that, I must admit. The extensions button (currently) switches the annotations on/off. Now that the extension is “always on” perhaps the behaviour should be to hide/show the sidebar?

judell:
> Now that the extension is “always on” perhaps the behaviour should be to
hide/show the sidebar?

We need to decide if there will be a menu associated with the icon as is
often the case with extensions. If so, show/hide would be one of a few
possible options we can discuss.

PR #2477, opened 2015-09-02
2440 replies dont match search normalized url queries 2

### LINKED ISSUE #2440: Replied on stream, didn't save.

seanh:
Fixes #2440.

### DIFF
--- a/h/api/models.py
+++ b/h/api/models.py
@@ -69,9 +69,9 @@ class Annotation(annotation.Annotation):
                     'search_analyzer': 'uri',
                     'copy_to': ['uri'],
                 },
-                # We store the 'source_normalized' unanalyzed and only do term
-                # filters against this field.
-                'source_normalized': {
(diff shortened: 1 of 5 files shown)


seanh:
Ready for review

tilgovi:
I merged this because with our current web client everything should work as expected. I opened #2479 and #2480.

PR #2483, opened 2015-09-03
Always return parent annotations of replies

### LINKED ISSUE #1916: Too many message not available on stream

seanh:
Problem:

The search API sometimes returns reply annotations without their
corresponding parent annotations being in the same query results.

This happens when it's returning, say, the 20 most recent annotations
and there are some reply annotations in those 20 but the parent annotations of
those replies are not among the 20 most recent annotations.

This means the client receives some reply annotations but can't display
them properly (as a threaded discussion) because it doesn't have their
parent annotations.

Solution:

Make sure that the search API always returns all of the parent
annotations, for any reply annotations that it returns.

This is done by splitting the search API's Elasticsearch query into two
queries:

1. The original query, but with an additional filter so that it returns
only top level annotations and not replies.

2. An additional query for all of the replies to the annotations from
the first query.

Then merge the results of the two queries and return them to the
client.

This fixes the "Message not available" cards that are appearing on our
/stream page in the frontend. These occur when the client receives
replies from the search API but does not receive the parent annotations
of those replies in the same query.

Fixes #1916.

### DIFF
--- a/h/static/scripts/widget-controller.coffee
+++ b/h/static/scripts/widget-controller.coffee
@@ -29,7 +29,7 @@ module.exports = class WidgetController
         if offset < total
           _loadAnnotationsFrom query, offset
 
-        annotationMapper.loadAnnotations(results.rows)
+        annotationMapper.loadAnnotations(results.rows.concat(results.replies))
 
     loadAnnotations = (frames) ->
       for f in frames
--- a/h/static/scripts/widget-controller.coffee
+++ b/h/static/scripts/widget-controller.coffee
@@ -29,7 +29,7 @@ module.exports = class WidgetController
         if offset < total
           _loadAnnotationsFrom query, offset
 
-        annotationMapper.loadAnnotations(results.rows.concat(results.replies))
+        annotationMapper.loadAnnotations(results.rows, results.replies)
 
     loadAnnotations = (frames) ->
       for f in frames
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -84,6 +84,11 @@ search
 
    **Example response**:
 
+   Top-level annotations (but not replies) that match the search query are
+   returned in the "rows" field. The "total" field counts the number of these
+   top-level annotations. The "replies" field always contains all of the reply
+   annotations to the top-level annotations in "rows".
+
    .. sourcecode:: http
 
       HTTP/1.1 200 OK
@@ -104,25 +109,28 @@ search
                   "user": "acct:gluejar@hypothes.is"
               }
           ],
-          "total": 1
+          "total": 1,
+          "replies": [
+              {...}
+          ]
       }
 
-   :query limit: The maximum number of annotations to return, for example:
-       ``/api/search?limit=30``. (Default: 20)
+   :query limit: The maximum number of top-level annotations to return in
+       "rows", for example: ``/api/search?limit=30``. (Default: 20)
 
-   :query offset: The minimum number of initial annotations to skip. This is
-       used for pagination. For example if there are 65 annotations matching
-       our search query and we're retrieving up to 30 annotations at a time,
-       then to retrieve the last 5 do: ``/api/search?limit=30&offset=60``.
-       (Default: 0)
+   :query offset: The minimum number of initial top-level annotations to skip.
+       This is used for pagination. For example if there are 65 annotations
+       matching our search query and we're retrieving up to 30 annotations at a
+       time, then to retrieve the last 5 do:
+       ``/api/search?limit=30&offset=60``.  (Default: 0)
 
-   :query sort: Specify which field the annotations should be sorted by. For
-       example to sort annotations by the name of the user that created them,
-       do: ``/api/search?sort=user`` (default: updated)
+   :query sort: Specify which field the top-level annotations should be sorted
+       by. For example to sort annotations by the name of the user that created
+       them, do: ``/api/search?sort=user`` (default: updated)
 
-   :query order: Specify which order (ascending or descending) the annotations
-       should be sorted in. For example to sort annotations in ascending
-       order of created time (i.e. oldest annotations first) do:
+   :query order: Specify which order (ascending or descending) the top-level
+       annotations should be sorted in. For example to sort annotations in
+       ascending order of created time (i.e. oldest annotations first) do:
        ``/api/search?sort=created&order=asc``. (Default: desc)
 
    :query uri: Search for annotations of a particular URI, for example
--- a/h/api/search/core.py
+++ b/h/api/search/core.py
@@ -25,9 +25,15 @@ def search(request_params, user=None, search_normalized_uris=False):
         search against pre-normalized URI fields.
     :type search_normalized_uris: bool
 
-    :returns: a dict with keys "rows" (the list of matching annotations, as
-        dicts) and "total" (the number of matching annotations, an int)
+    :returns: a dict with keys:
+        - "rows": The list of matching annotations, as dicts.
+                  Note: only top-level annotations are returned here, not
+                  reply annotations.
+        - "total": The number of matching top-level annotations, an int.
+        - "replies": The list of all reply annotations to the matching
+                     top-level annotations, as dicts.
     :rtype: dict
+
     """
     userid = user.id if user else None
     log.debug("Searching with user=%s, for uri=%s",
@@ -38,11 +44,27 @@ def search(request_params, user=None, search_normalized_uris=False):
                        search_normalized_uris=search_normalized_uris)
     results = models.Annotation.search_raw(body, user=user, raw_result=True)
 
-    total = results['hits']['total']
-    docs = results['hits']['hits']
-    rows = [models.Annotation(d['_source'], id=d['_id']) for d in docs]
+    ids = [h['_id'] for h in results['hits']['hits']]
+    replies = models.Annotation.search_raw(
+        {
+            'query': {
+                'terms': {'references': ids}
+            },
+            'size': 10000,
+        }, user=user, raw_result=True)
 
-    return {"rows": rows, "total": total}
+    if len(replies['hits']['hits']) < replies['hits']['total']:
+        log.warn("The number of reply annotations exceeded the page size of "
+                 "the Elasticsearch query. We currently don't handle this, "
+                 "our search API doesn't support pagination of the reply set.")
+
+    total = results['hits']['total']
+    rows = [models.Annotation(hit['_source'], id=hit['_id'])
+            for hit in results['hits']['hits']]
+    replies = [models.Annotation(hit['_source'], id=hit['_id'])
+               for hit in replies['hits']['hits']]
+
+    return {"rows": rows, "total": total, "replies": replies}
 
 
 def index(user=None, search_normalized_uris=False):
--- a/h/api/search/query.py
+++ b/h/api/search/query.py
@@ -10,6 +10,9 @@ def build(request_params, userid=None, search_normalized_uris=False):
     Translates the HTTP request params accepted by the h search API into an
     Elasticsearch query dict.
 
+    Always inserts a filter so that only top-level annotations will be returned
+    by Elasticsearch, never reply annotations.
+
     :param request_params: the HTTP request params that were posted to the
         h search API
     :type request_params: webob.multidict.NestedMultiDict
@@ -24,6 +27,7 @@ def build(request_params, userid=None, search_normalized_uris=False):
     :returns: an Elasticsearch query dict corresponding to the given h search
         API params
     :rtype: dict
+
     """
     # NestedMultiDict objects are read-only, so we need to copy to make it
     # modifiable.
@@ -50,7 +54,7 @@ def build(request_params, userid=None, search_normalized_uris=False):
         }
     }]
 
-    filters = []
+    filters = [{'missing': {'field': 'references'}}]
     matches = []
 
     uri_param = request_params.pop("uri", None)
--- a/h/api/search/test/core_test.py
+++ b/h/api/search/test/core_test.py
@@ -41,6 +41,91 @@ def test_search_does_pass_userid_to_build(build, _):
     assert build.call_args[1]["userid"] == "test_id"
 
 
+@mock.patch("h.api.search.core.models.Annotation")
+@mock.patch("h.api.search.core.query.build")
+def test_search_queries_for_replies(_, Annotation):
+    """It does a second Es query for replies to the results of the first."""
+    ids = ['foo', 'bar', 'gar']
+    Annotation.search_raw.side_effect = [
+        # The first time search_raw() is called it returns the result of
+        # querying for the top-level annotations only.
+        {
+            'hits': {
+                'total': 3,
+                'hits': [
+                    {'_id': ids[0], '_source': 'source'},
+                    {'_id': ids[1], '_source': 'source'},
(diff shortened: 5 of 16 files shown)


seanh:
@tilgovi @nickstenning Either one of you. Proof of concept here, not final code. I want agreement among us first. But this seems like a simple and effective solution to me.
 
To reproduce _Message not available_ try forcing `size` to, say, 5 in `query.py:build()` and then in the sidebar make an annotation with 6 replies and then look at the stream. Then checkout this branch and see it fixed.

seanh:
One thing we should probably do before merging this is enforce a maximum value on `size`, because the more results the first query returns the bigger the second query becomes. Suggest 200 as that seems to be the most our client requests.

seanh:
Another problem is that the size of our server's reply is now out of control. It returns at most `size` top-level annotations, plus all replies to all of those which is an unbounded number of replies. We could of course enforce a generous limit on the second query as well, though that may lead to the client displaying incorrect results in extreme cases.

seanh:
Oh, and this means that when the client would previously receive n annotations (counting both top-levels and replies) it now receives n top-levels plus all of their replies. The frontend doesn't seem to mind though

tilgovi:
The calculation for `total` is a bit problematic isn't it? If I used an offset that was greater than the number of top level annotations and less than `total` I would get no results.

seanh:
Hmm, yeah. I guess if the behaviour of the search API is now to return n top-level annotations plus all their replies, then the total, limit and offset should be counts of the top-level annotations only and not the replies. So I'll just take that bit out.

It does feel a bit wrong that we have a search API that's returning these objects, both top-level annotations and reply annotations merged together in a single stream, and it supports counts like total, limit and offset, but basically the actual number of objects returned by any query is now a bit disjointed from those pagination numbers :(

It might be more correct if the server did a little processing to inline those replies into the top-level annotations. Or even just put them in a separate key in the results, "replies" instead of "hits", or something.

This first pass is the easiest quick fix for the bug, since it doesn't require any client changes, but maybe we can do something to make it less weird without requiring too heavy client changes.

nickstenning:
If I understand correctly, then if I set `?limit=10`, and the first annotation in the resultset has 5 replies, I'm now going to get a response that contains 15 annotations. Personally I think I'd be super confused by such an API.

I don't think this is your fault, @seanh, because I think the semantics around replies are fundamentally confusing.

There are (as I can see it at the moment) only two scenarios that are consistent:

1. Replies are annotations and are treated just like any other annotations. A search query doesn't treat them specially. The stream contains a list of annotations, in order, with links that *dynamically* load thread parents or thread children, if they exist. This implies knowing whether or not an annotation has parents/children when you do the initial load. (You'll note that this is essentially how Twitter conversations work.)

2. Replies are not annotations, but, uh, replies. A distinct domain object. At this point we have many options open to us. We can inline replies as properties of annotations and render them server-side. We can return them in the API call for annotations, either always or optionally.

While ultimately I would lean towards interpretation 2, I think that's probably a little way off and a lot more work that interpretation 1, so that would be the preferred direction to move in.

Here are some compromise options which be preferable in the short term:

1. Add a top-level field (`references`) to the search API response which contains all annotations referenced by the resultset (parents and children). That way the semantics of `total` with respect to `rows` is preserved, but you don't need to perform a second API call to get the replies.
2. Add another API call which accepts a set of IDs and returns the referenced annotations. Call this when the stream loads and dynamically insert the replies.

tilgovi:
Couldn't we also solve this, in the direction of option 1, by just improving the display?

Right now we say "Message not available" but that's only there because we _do_ have a reply available. Rather than nesting that under an empty root, we could put it in the stream itself as a top level thing. This is very much the way Twitter is doing it. Their stream is flat-ish.

To do that we would need to ditch the way we're currently leaning on load events to populate a tree construction in the threading service, but the result would probably be nicer and simpler to reason about and evolve in the code, though it'd be a little more initial investment.

Of course, all of this makes me want to ask why we're spending so much mental energy on something that could be 80% solved for the moment by just hiding these things.

judell:
When I was exploring how threading works I noticed the client making a call to (I think, can't find it now) /api/references. What was that about? Is it gone?

tilgovi:
Never existed.

You may be thinking of the standalone page which does perform a search for
the annotation and then another for any descendants.

The same could be done in the stream but it's potentially lots of requests.

seanh:
I'd like to suggest limiting this to fixing the bug at hand, and hold off on redesigning the stream (at least for this pull request).

I think solutions involving re-designing the stream to be more like Twitter's stream, and/or to make additional requests when you click to expand replies, sound like more work to me.

The current intended behaviour of the stream appears to be that it shows a list of threads (thread = an annotation and all of its replies). Each thread is collapsed at first but expands when you click on the replies link. This is done without making additional requests to the server at click time - the annotations and their replies are all fetched together in one request. (Or at least, the frontend code seems to falsely assume that they are.) The stream is sorted newest top-level annotations first, new replies do not cause old annotations to jump back to the top.

> Replies are not annotations, but, uh, replies. A distinct domain object. At this point we have many 
> options open to us. We can inline replies as properties of annotations and render them 
> server-side. We can return them in the API call for annotations, either always or optionally.

We may not be as far off from this as you think. On this branch we have the list of n top-level annotations and the list of all replies to those n as two separate lists. Instead of concatenating them, we could easily inline the replies into the top-levels server-side now.

> Add a top-level field (references) to the search API response which contains all annotations 
> referenced by the resultset (parents and children). That way the semantics of total with respect to 
> rows is preserved, but you don't need to perform a second API call to get the replies.

That would also be really easy to do from what we have now.

seanh:
In general if we want replies to be a separate domain object from annotations, then a first step towards it might be to change APIs like the search API to return top-level annotations only. This can easily be done by adding a filter, letting the replies continue to exist alongside annotations in the same search index, but never retrieving them at the same time.

We can then make the APIs always inline replies into the annotations, or add additional endpoints for getting all the replies of a given annotation, etc. Still all without changing the fact that the search index contains both annotations and replies mixed together.

seanh:
Alright, following Nick's suggestion and just putting the replies in a separate key. I've fixed the stream page to handle this, should probably check other places in the frontend that get responses from the search API..

seanh:
I think the sidebar (widget-controller) is the only place left to do, and fix tests

seanh:
I just want to manually test each of the different parts of h with this before it's merged, but I just force-pushed this and it's the proposed final solution with tests so feel free to take a look

tilgovi:
Some thoughts in case we think we want to do better:

- Might be easier for developers if the replies were a dict keyed by top-level annotation ID. It would allow people to construct some threads without doing the full-on JWZ algorithm we've got.

- It may make more long term sense if we were to limit these to just first-order replies and flag when things are deeper, so they could be fetched on demand. That might have been hard before, when `references` was all we had to go on for searching replies, but now that replies have targets it's simple enough to get only first order replies by requesting annotations that target the top level ones.

Thoughts on those?

tilgovi:
Just to say a bit more about that, I've already said for many months now that we can solve the issue of there being cards that just say "Message not available" by simply hiding them client-side and it would be a tiny, one-line change. So, I think if we are shipping any more interesting or complete solution like this we should do so in a way that we really like and want to support, not as something we feel like we rushed out to address that issue, since there are less involved ways to do that.

tilgovi:
So, ask ourselves, if we haven't, "What would a good API for fetching annotations and replies that we would want to use as developers look like?"

seanh:
Ok, as well as the unit tests I've manually tested this on the stream, in the sidebar, on individual annotations, and with the live updates and infinite scroll - all seems to be working.

I think (if code review is happy with it) we should merge this for now (it fixes the bug) and consider further improvements separately.

To respond to Randall's suggestions:

> Might be easier for developers if the replies were a dict keyed by top-level annotation ID. It would allow people to construct some threads without doing the full-on JWZ algorithm we've got.

If we're going to do that, then I think we might as well inline replies into the annotations themselves, instead of them having their own top-level key. And we might as well do the full jwz threading server-side as well.

> It may make more long term sense if we were to limit these to just first-order replies and flag when things are deeper...

I don't see why we would limit it to first-order replies. I think we _do_ need to limit the number of replies that we send at some point, just to keep the size of our server responses under control, and provide some way for clients to make additional requests to get the rest. But my first thought would be to limit it to the most recent n replies, first-order or nested.

> What would a good API for fetching annotations and replies that we would want to use as developers look like?

Alright, so limiting this to the use-cases that I know about:

1. A page showing the most recent n annotations and all the replies to those n, i.e. the stream page

2. A page showing all of the annotations of a given document and all their replies, i.e. the sidebar

3. A page showing just one annotation and all of its replies, i.e. the individual annotation page

In every case, not making any assumption about whether the annotations and replies are fetched in one request or the replies are retrieved with additional requests.

Also, forgetting about our current client-side code. If we were starting a new client from scratch, and we had the above 3 use-cases, what would be the most convenient API for us as developers?

- We'd want to be able to do queries that get the most recent _top level_ annotations, all of the top level annotations of a given document, or a single top-level annotation by ID.
- In all cases we're querying against top-level annotations only, we're not querying against replies. 
- But in all cases, we also want to get all the replies to all of the annotations that matched our queries, since we always want to display those with the annotations.
- It'd be most convenient if those replies just came back in the first request, rather than us having to make additional requests.
- And given that we want to show the replies threaded, it'd be most convenient if they came back already threaded for us and we just had to render them.

So the changes I'd make to the API that we currently have on this pull request would be:

1. Inline the replies into the annotations, instead of returning them in their own top-level key
2. Inline them with the nested threading done, moving the JWZ threading algorithm from client to server-side

I'd suggest that those are both changes that can be made in follow-up pull requests to this one. Put them under technical debt, probably, and prioritise them that way. Or you could argue that they go under product, since this is making the API better for our developer users, either way.

There's one other problem: We need to deal with the situation when there are so many replies that we can't send them all at once because the response would be too big. The current solution is just to truncate them at 10,000 and log a warning. The best API for this would probably be an additional endpoint for getting all the replies to a given annotation, with pagination. I'm not sure whether we'd want our client-side code to actual use such an API though, maybe we should just make our individual annotation page (which I think should be reimplemented as a server-side page) show all the replies, and our client just needs to know when there are more replies and link the user to that page to read them.

Anyway, _that_ one would be a third follow-up pr on top of this one and I'd say a lower priority one, since I don't think we're gonna hit this problem often (and we have the logging in place to monitor whether we do).

seanh:
Just remembered that with this change we don't need normalized URLs in replies anymore, so I should remove that from `prepare()`...

tilgovi:
> I'd suggest that those are both changes that can be made in follow-up pull requests to this one. Put them under technical debt, probably, and prioritise them that way.

This is a pernicious line of thinking that erodes the quality of our work.

When something is critical and needs to be fixed immediately at the expense of technical debt that's one thing, but in this case we're talking about a minor issue that, in the user's experience, is cosmetic and has been open since February.

If there are ways we know it can be much better, let's do them when we have time and acknowledge the costs here.

We are not just a public web service where we can ship updates any time we want. We have partners that are in the process of deploying our application with their own backends that we don't write or control. It's time to change our mentality.

h/static/scripts/widget-controller.coffee
```diff
@@ -29,7 +29,7 @@ module.exports = class WidgetController
         if offset < total
           _loadAnnotationsFrom query, offset
 
-        annotationMapper.loadAnnotations(results.rows)
+        annotationMapper.loadAnnotations(results.rows.concat(results.replies))
```
nickstenning: I think a comment here to explain what's going on would be useful.

seanh:
@nickstenning Ok, I've clarified and deduplicated the concatenating of annotations and replies arrays.

One other thing that came to mind: with this change, a thread will only come up in search results if its top-level annotation matches the search. If you search for something that's in one of the replies, you won't find the thread.

dwhly:
> If you search for something that's in one of the replies, you won't find the thread.

I assume that means if I search for q=user:dwhly that none of my replies will match?

seanh:
> > If you search for something that's in one of the replies, you won't find the thread.
> 
> I assume that means if I search for q=user:dwhly that none of my replies will match?

If you were searching using our search API then no, they wouldn't match. You'd find your top-level annotations only.

Based on a quick test this means that if you search on the stream page then your replies don't match (I guess the stream page calls our search API) but if you search in the sidebar then your replies _do_ match (I guess because the sidebar uses its own, client-side search implementation).

I believe that the current behaviour on master / production is that if you search for `user:dwhyly` in the stream, and `dwhyly` has made a reply to an annotation that was made by someone else, then all you'll see in place of that thread is a _Message not available_ card.

Thoughts on ways that we can get rid of Message not available _and_ make sure that threads will show up in search results if a reply (not just a top-level annotation) matches the query _and_ support pagination, at the same time:

1. Inline replies into annotations at search index time. Es documents might get too big if a thread has too many replies though.

2. Is there a fancy Es query we can do that will return top-level annotations only, but that will return them whether the annotation itself or any of its replies matched the query? I think there may not be a good way to do this unfortunately.

3. If _threads_ (as opposed to the individual annotations and replies) had their own unique IDs that were denormalized into the annotations and replies,  then we could do it in two Es queries: 1. Query for all annotations and replies that match the user's query parameters. 2. Query for all annotations and replies whose thread ID matches any of the thread IDs of the results of the first query.

   What would be a unique thread ID that's denormalized into all of the thread's annotations and replies? We don't have thread IDs as such in Hypothesis (as far as I know), but I believe that all replies contain the top-level annotation's ID in `references` and the top-level annotation contains it in `id` so we could use that.

   But there may not be a good way to reconcile this with pagination.

dwhly:
> I believe that the current behaviour on master / production is that if you search for user:dwhyly in the stream, and dwhyly has made a reply to an annotation that was made by someone else, then all you'll see in place of that thread is a Message not available card.

If this is the case, and thus this PR doesn't make it worse, then I'm fine with it as a fix for the Message Not available problem.  But then we'll need to prioritize enabling folks to search for whatever they've contributed, whether it be a top level annotation or a reply.

seanh:
I think it's a good user story, but I think there's a design question here, too. Currently the stream always shows the matching annotations with the replies initially collapsed. So let's say we did make the API so that a thread would come in the results if one of its replies matched the query, what you'd actually see is the annotation (which is not by user dwhyly) and to find the reply that caused that thread to be in the results you'd have to expand the replies, in which case you may see a bunch of other stuff not by dwhyly mixed in as well, there'd be no indication that the one matching reply was the thing that actually matched / the reason why the whole thread was in the search results.

So I'd say this comes under the general project of a redesigned user experience for the stream? Which I think is in the backlog somewhere.

dwhly:
> So I'd say this comes under the general project of a redesigned user experience for the stream? Which I think is in the backlog somewhere.

Yeah.  A while back @gergely-ujvari implemented a nice collapsed view to search results in the sidebar.  So that everything above the result was collapsed under a "Load n more annotations" and everything below was collapsed under a "Load n more replies" (more or less).  Something like this is probably necessary here.  (note to @conordelahunty)

lenazun:
Comments on closing:
"message not available only appears in the stream currently (because the sidebar makes multiple requests until it has all the annotations), I don't expect to come back to this problem until we reimplement the stream with new use-cases, a new design, and a complete rewrite as purely server-side code
I think the only way to make it appear in the sidebar would be to delete an annotation that has replies, or somehow manage to use the API to make a reply to an annotation that is on a different document or in a different group etc"

PR #2485, opened 2015-09-04
Swagger.io API docs

BigBlueHat:
Super early work on using [Swagger](http://swagger.io/) for our API docs.

You can paste it into the [Swagger Editor](http://editor.swagger.io/#/edit) to get a sense of what an auto-generated docs site / testing-console-thing would look like once published.

All told I found Swagger very intuitive in it's format and layout.

If there are no objections to this work / approach, I'll continue it over the coming conference weeks (yeah :airplane: rides!) and get a more official PR sent in when it's "done."

Just wanting feedback for now. @judell @robertknight @tilgovi


nickstenning:
I'm not sure exactly why this appears to have diverged from master so long ago, but it would need at the very least a rebase now.

BigBlueHat:
@nickstenning I'm going to close this actually in favor of building a proper developers site and API Console--which is in progress here (for now): https://github.com/BigBlueHat/developers.hypothes.is

Once that's beyond experimental, I'll move it under the `hypothesis` org, and get it published at `developers.hypothes.is`...or that's the plan at least.

The Swagger file itself doesn't help much unless we publish it with our API (for tools that can use it) or publish it with the tools that can use it (like an API console, test scripts, etc)--so, heading this that direction.

PR #2487, opened 2015-09-08
Publish to a group

nickstenning:
I'm opening a new pull request. The last one (#2432) was getting unwieldy and we were having conversations about revisions on revisions.


seanh:
Did a quick test, found a couple of details that seem to have gotten broken in the UI:

1. When creating a shared annotation in a group, the editor says _This annotation will be visible to everyone_, should be _to group Foobar_.

2. Create a shared annotation in a group, then edit it, the visibility dropdown is set to _Only Me_ initially. Should be the group.

tilgovi:
Rebased against master, with the changes that @seanh did to move the search logic out of the views.

nickstenning:
So that should now be both of the bugs you've identified fixed, @seanh.

For bonus points, the author's default group is also persisted across page loads via localStorage.

h/api/groups/auth.py
```diff
@@ -0,0 +1,45 @@
...
+    # client sends just the user's ID in the read permissions.
+    is_private = (annotation['permissions']['read'] == [annotation['user']])
+
+    if is_private:
+        # The groups feature doesn't change the permissions for private
+        # annotations at all.
+        return
+
+    group = annotation.get('group')
+    if group is None or group == '__world__':
```
seanh: `if group in (None, '__world__'):`

seanh: Now I'm nit-picking my own old code

PR #2490, opened 2015-09-08
Add configurable callbacks for sidebar open/close.

JakeHartnell:
A retry of https://github.com/hypothesis/h/pull/2467

The purpose is the same, implement the Epub reader integrations and enabling 3rd party integrators to better adapt to annotation UI. 

This PR uses configurable callbacks rather than events.

### DIFF
--- a/h/static/scripts/annotator/host.coffee
+++ b/h/static/scripts/annotator/host.coffee
@@ -68,6 +68,11 @@ module.exports = class Host extends Guest
     @frame.css 'margin-left': "#{-1 * @frame.width()}px"
     @frame.removeClass 'annotator-collapsed'
 
+    if typeof @options.showFrameCallback == 'function'
+      # Only callback once.
+      if not @toolbar.find('[name=sidebar-toggle]').hasClass('h-icon-chevron-right')
+        @options.showFrameCallback()
+
     if @toolbar?
       @toolbar.find('[name=sidebar-toggle]')
       .removeClass('h-icon-chevron-left')
@@ -77,6 +82,11 @@ module.exports = class Host extends Guest
     @frame.css 'margin-left': ''
     @frame.removeClass 'annotator-no-transition'
     @frame.addClass 'annotator-collapsed'
+
+    if typeof @options.hideFrameCallback == 'function'
+      # Only callback once.
+      if not @toolbar.find('[name=sidebar-toggle]').hasClass('h-icon-chevron-left')
+        @options.hideFrameCallback()
 
     if @toolbar?
       @toolbar.find('[name=sidebar-toggle]')


BigBlueHat:
I like the way that's looking. Do you have the epub.js code using this? It'd be great to see what the integration looks like now. Hopefully that's also simpler.

JakeHartnell:
```
            var onSidebarToggle = function() {
              if (!$("#main").hasClass("single")) {
                $("#main").addClass("single");
              } else {
                $("#main").removeClass("single")
              }
            }

              return {
                // app: "https://hypothes.is/app.html"
                app: "http://localhost:5000/app.html",
                Toolbar: {container: '#annotation-controls'},
                clickToClose: false,
                showFrameCallback: onSidebarToggle,
                hideFrameCallback: onSidebarToggle
              }
            };
```

JakeHartnell:
@tilgovi if this is an acceptable approach I can write tests for this. Once this is done I can get a version of Epub.js and Hypothes.is up for people to play with and start work on the Readium integration!

tilgovi:
This seems to be the same approach but with callbacks instead of events. The problem I have with it is that it enshrines the current behavior (creating annotations opens the sidebar) rather than making it configurable.

The sidebar opens in these ways:

- Starting the flow to create a new annotation
- Clicking the chevron in the toolbar

The sidebar closes in these ways:

- Clicking on the document
- Clicking on the chevron in the toolbar

We added `clickToClose` to toggle whether it closes when you click on the document because it felt nicely simple, at the time. It's an off/on feature. I'm not sure I love it, because I hesitate to ever introduce any options. I prefer construction by composition of static components rather than having lots of switches. It's easier to test, easier to reason about, easier to re-use in unexpected ways, when code is built from small pieces that can be composed. But, anyway, we have `clickToClose` for now.

I'm less into callbacks because it's reactive. I don't think the developer should be reacting to our application. I think the application should be behaving according to the wishes of the developer.

So, let's talk practically about what could change that might be much better.

The chevron is 2/3 of the related interactions (leaving aside `clickToClose`). For your use case, are you going to want to keep the toolbar or make your own buttons for toggling highlights and creating page notes? I suspect you'll want to tap into the page notes, at least, because you may need to store some page epub-related page information. Or maybe pages don't even make sense because it's reflowable? I don't know. So, my question is do you really want the toolbar at all? If you turn the toolbar off, you eliminate 2/3 of the times when the sidebar does something (opens/closes) that you feel you need to be able to react to.

The last interaction is the initiation of the annotation create flow. That's a little trickier because the adder isn't something that's optional right now, so you don't have the opportunity to direct the sidebar to open and create a new annotation at that moment. That behavior is currently fixed.

Do you want the adder? Would you want to be able to customize it? Would you want to be able to, for instance, decide that it only shows up when the sidebar is open, when you're in some annotation mode? Or maybe highlights are something you always want to make?

I'm not opposed to all events and callbacks, I just like to keep things nicely separate. A button can emit an event when it's clicked. A sidebar can emit an event when it opens. But our sidebar isn't a sidebar with an annotator inside it, it's an all-in-one-annotation-sidebar. That's where I start to cringe a little bit and feel like it's growing too many knobs.

Can we do better? Can we talk more about the needs rather than the implementation? What would really be ideal here? If you could get an annotation "widget", and I use that word deliberately instead of "sidebar", and place it wherever you want and control when it's hidden or visible and decide when and how annotations are created, would you prefer that?

JakeHartnell:
> This seems to be the same approach but with callbacks instead of events. 

It is.

> The problem I have with it is that it enshrines the current behavior (creating annotations opens the sidebar) rather than making it configurable.

Does it? I don't see how sidebar open and close events / callbacks enshrine that (creating annotations opens the sidebar).

> I'm not opposed to all events and callbacks, I just like to keep things nicely separate. A button can emit an event when it's clicked. A sidebar can emit an event when it opens. But our sidebar isn't a sidebar with an annotator inside it, it's an all-in-one-annotation-sidebar.

But it's still a sidebar and *it covers up content*. I don't think what I'm proposing is crazy: either two events or two callbacks.

> Do you want the adder? Would you want to be able to customize it? Would you want to be able to, for instance, decide that it only shows up when the sidebar is open, when you're in some annotation mode? Or maybe highlights are something you always want to make?

Certainly, but one thing at a time. I would really like to get integrations of Epub.js and Readium working before October 1st (you can imagine why), and until I have something merged which will allow me to not have the sidebar cover up content, I am blocked.

tilgovi:
Closed in favor of #2494.

@JakeHartnell I'm going to push to stage imminently and if things look cool throughout the day the newest will get deployed to production. At that point, I'd be happy to take a look at a branch of the epub sample application and see if I can't help you get it going.

PR #2493, opened 2015-09-11
Refactor Pyramid routing and view configuration

tilgovi:
When resources are backed by database models, keep the two separate
to improve extensibility and package isolation. By separating the
resource tree from the persistence models, the resource tree can be
extended without dealing dealing with database models and mappers.
With sub-classing of the resource tree, explicit containment of the API
is no longer needed for any view other than the exception views. The
annotation model also becomes simpler, no longer containing Pyramid
permissions details. In this manner, more details about groups and
users are lifted out of the annotation API package.

Remove the "admin party" mode for annotations since otherwise this
would introduce "upsert" operations using PUT for anyone in any group.
Simply removing this unused functionality allows the annotation resource
to be behave as if its model always exists, even if empty and not
persisted, which avoids having to deal with that case within the acl code
and the view callables.

Factor out a generic Collection resource and implement the stream and the
search in terms of that. Remove most of the logic from the stream view
by making the tag and user search shortcuts implemented as a collection of
collections of annotations so that the view itself need only inspect its
context for the presence of a query. Future work can remove the need for
a redirect by teaching the client about these routes, thereby beautifying
the stream URLs for real user and tag pages.

Make all the view configuration decorators and helpers more concise and
consistent.

### DIFF
--- a/h/api/__init__.py
+++ b/h/api/__init__.py
@@ -5,8 +5,11 @@ from h.api.logic import search_annotations
 
 
 def includeme(config):
-    config.set_root_factory(create_root)
-    config.add_route('api', '/')
+    config.add_route('api', '/', factory=create_root)
+    config.add_route('access_token', '/access_token')
+
+    # XXX: Client should be using /access_token, isn't yet.
+    config.add_route('token', '/token')
 
     config.include('h.features')
     config.include('h.api.db')
--- a/h/api/resources.py
+++ b/h/api/resources.py
@@ -1,8 +1,10 @@
 # -*- coding: utf-8 -*-
 
-from pyramid.security import Allow, Deny, Authenticated, Everyone
+from pyramid.decorator import reify
+from pyramid.security import Allow, Deny
+from pyramid.security import Authenticated, Everyone
 
-from .models import Annotation
+from h.api import models
 
 
 class Resource(dict):
@@ -22,12 +24,69 @@ class Resource(dict):
         self[name] = obj
 
 
-class Annotations(object):
+class Collection(Resource):
     """
-    Annotations is a container resource that exposes annotations as its
-    children in the tree.
+    A collection of other resources.
     """
-    def __init__(self, request):
+
+    def __getitem__(self, key):
+        if key not in self and callable(self.factory):
+            self.add(key, self.factory(key))
+        return super(Collection, self).__getitem__(key)
+
+    def factory(self, key):
+        raise NotImplementedError
+
+
+class Annotation(Resource):
+    """
+    A Resource representing an annotation.
+    """
+
+    def __acl__(self):
+        acl = []
+
+        model = self.model
+
+        # Convert annotator-store roles to pyramid principals
+        for action, roles in model.get('permissions', {}).items():
+            for role in roles:
+                if role.startswith('system.'):
+                    raise ValueError('{} is a reserved role.'.format(role))
+                elif role.startswith('group:'):
+                    if role == 'group:__world__':
+                        principal = Everyone
+                    elif role == 'group:__authenticated__':
+                        principal = Authenticated
+                    elif role == 'group:__consumer__':
+                        raise NotImplementedError("API consumer groups")
+                    else:
+                        principal = role
+                else:
+                    principal = role
+
+                # Append the converted rule tuple to the ACL
+                rule = (Allow, principal, action)
+                acl.append(rule)
+
+        return acl
+
+    @reify
+    def model(self):
+        if 'id' in self:
+            instance = models.Annotation.fetch(self['id'])
+            return instance or models.Annotation(id=self['id'])
+        else:
+            return models.Annotation()
+
+
+class Annotations(Collection):
+    """
+    A collection of Annotation resources.
+    """
+
+    def __init__(self, request, **kwargs):
+        super(Annotations, self).__init__(**kwargs)
         self.request = request
 
     def __acl__(self):
@@ -40,20 +99,12 @@ class Annotations(object):
             aces.append((Allow, 'group:' + group, 'create'))
         return aces + [(Deny, Everyone, 'create')]
 
-    def __getitem__(self, key):
-        instance = Annotation.fetch(key)
-        if instance is None:
-            raise KeyError(key)
-        instance.__name__ = key
-        instance.__parent__ = self
-        return instance
+    def factory(self, key):
+        return Annotation(id=key)
 
 
 class Root(Resource):
-    __acl__ = [
-        (Allow, Authenticated, 'create'),
-        (Allow, 'group:__admin__', 'admin'),
-    ]
+    pass
 
 
 def create_root(request):
--- a/h/accounts/views.py
+++ b/h/accounts/views.py
@@ -4,12 +4,12 @@ import json
 
 import deform
 from pyramid import httpexceptions
+from pyramid.exceptions import BadCSRFToken
 from pyramid.view import view_config, view_defaults
 from pyramid.security import forget, remember
 from pyramid_mailer import get_mailer
 from pyramid_mailer.message import Message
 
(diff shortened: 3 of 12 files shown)


tilgovi:
Ready for review.

seanh:
> By separating the resource tree from the persistence models, the resource tree can be extended without dealing dealing with database models and mappers.

This seems like a very good idea to me.

seanh:
So `json_view()` was added in `h/views.py`, and then the changes to `h/accounts/views.py` are just refactoring it to use `json_view()`? This seems like an independent change from the rest of the changes in this commit, would be good to split that out into its own commit I think.

h/api/__init__.py
```diff
@@ -5,8 +5,11 @@
 
 
 def includeme(config):
-    config.set_root_factory(create_root)
-    config.add_route('api', '/')
+    config.add_route('api', '/', factory=create_root)
+    config.add_route('access_token', '/access_token')
+
+    # XXX: Client should be using /access_token, isn't yet.
+    config.add_route('token', '/token')
```
seanh: I'm not sure what `XXX` means, should this be `FIXME` or `TODO`?

seanh:
I think this is pretty hard to understand as one big commit. It would help me a lot if it was split into I think at least four: the view config changes, separating resource tree and model, factoring out Collection and refactoring tag and user search based on it, and other independent changes e.g. the changes to `api/__init__.py`.

I'm behind all of the changes though. I think separating resource tree and model, and the idea of the app subclassing the API's resource tree classes. seems pretty good, also think we should be putting more code (helper methods, validation, etc) as methods on these resource classes, make them the extension point.

Not sure I totally understand the benefits of the Collection/tag/search changes yet.

seanh:
The only thing that I'm not sure about with extending the resource tree, is that the resource tree classes seem fairly fundamental/core to how the API's code works, and we're essentially exposing these as the API's API to the app to extend it. Does this tightly couple the app to the API? If we want to do refactoring of the API's resource tree or resource classes, and the app is extending those... I just wonder if we can define a smaller contact area between app and API.

h/api/resources.py
```diff
@@ -22,12 +24,69 @@ def add(self, name, obj):
...
+                rule = (Allow, principal, action)
+                acl.append(rule)
+
+        return acl
+
+    @reify
+    def model(self):
+        if 'id' in self:
+            instance = models.Annotation.fetch(self['id'])
+            return instance or models.Annotation()
```
nickstenning: What's the use case for returning a blank annotation if we have an id?

h/api/resources.py
```diff
@@ -22,12 +24,69 @@ def add(self, name, obj):
...
+            for role in roles:
+                if role.startswith('system.'):
+                    raise ValueError('{} is a reserved role.'.format(role))
+                elif role.startswith('group:'):
+                    if role == 'group:__world__':
+                        principal = Everyone
+                    elif role == 'group:__authenticated__':
+                        principal = Authenticated
+                    elif role == 'group:__consumer__':
+                        raise NotImplementedError("API consumer groups")
```
nickstenning: I know you didn't change this in this PR, but it strikes me that this should be `continue` and not `raise NotImplementedError`. Ditto the `raise` above. I shouldn't be able to crash h by creating an annotation with one of these keys in the permissions field.

h/api/resources.py
nickstenning: Yeah, this is pretty odd. It means that the only reason I get a 404 at `/a/gibberish` is because a blank annotation doesn't give the `read` permission.

nickstenning:
> The only thing that I'm not sure about with extending the resource tree, is that the resource tree classes seem fairly fundamental/core to how the API's code works, and we're essentially exposing these as the API's API to the app to extend it. Does this tightly couple the app to the API? If we want to do refactoring of the API's resource tree or resource classes, and the app is extending those... I just wonder if we can define a smaller contact area between app and API.

I don't think this is an issue. The application refers to `h.resources.Annotation`, which can be anything. It just happens to be a subclass of `h.api.resources.Annotation` but that fact isn't visible to the rest of the application.

---

Looks good to me overall, but for my comment about the behaviour when `Annotation.fetch(id)` returns `None`.

h/api/resources.py
tilgovi: Yep. That's the intention.

tilgovi:
Fixed the mistake where I forgot the `id=self['id']` in the resource model. Thanks, @nickstenning. Gonna merge this.

seanh:
> > The only thing that I'm not sure about with extending the resource tree, is that the resource tree classes seem fairly fundamental/core to how the API's code works, and we're essentially exposing these as the API's API to the app to extend it. Does this tightly couple the app to the API? If we want to do refactoring of the API's resource tree or resource classes, and the app is extending those... I just wonder if we can define a smaller contact area between app and API.
> 
> I don't think this is an issue. The application refers to `h.resources.Annotation`, which can be anything. It just happens to be a subclass of `h.api.resources.Annotation` but that fact isn't visible to the rest of the application.

I guess I don't understand why this isn't an issue. The API has a resource tree involving Resource, Root, Collection, Annotation and Annotations. The app then has its own resource tree with its own Annotation and Annotations classes among others. The app's resource tree subclasses Annotation, Annotations and Resource from the API's resource tree.

That's what I was getting at - it seems like the interface between app and API is the API's whole resource tree. Plus various parts of the API that the app calls, including view and logic functions and some helper functions etc.

The contact area between app and API seems quite large and ill-defined, so I'm just worrying whether it's gonna be possible to refactor API code (for example) without necessitating a lot of refactoring of app code.

Not yet convinced that we aren't creating a problem here..

tilgovi:
Two different things:

1) The API of the service
2) The public interface of the package providing it

For a Pyramid package, I would say the resource tree is a critical part of its public interface. It's the resources you get as your view context. The root factory is how you control routing, etc.

You can duplicate code, if you really want to, so that the HTML site uses different resources, but I wouldn't do it just because you're worried about the resources in the API package shifting and changing. They are public. It's not even separately versioned and packaged (much to my dismay) so I'm not sure why so much concern.

I would have just imported the resources directly and not sub-classed them, but sub-classing makes the applications view predicates match more specifically against these rather than the API routes.

What's attractive about having the resource tree be public is that you can extend the application in many ways just by changing the root factory. It would be a worthwhile goal to extract groups and nipsa from the api package entirely. Build the query and the permissions in the resource context and then the views don't have to know anything about these features and they can be re-used for a simple, single-user, admin-only annotation store or a multi-user, multi-group store.

The api package is really just an "annotations" package. It happens to ship with JSON views and so it's called "api" right now, but it shouldn't be.

If it really bothers you, I would say pull the views out of `h.api` and just export the models and resources as the public, somewhat abstract, interface for the package.

tilgovi:
I moved this way very deliberately because I was thinking _exactly_ of the goal of moving groups and nipsa out of the `h.api` package. I wanted to construct a resource tree where the leaves hold enough information to create the proper search queries, factoring in group membership and nipsa, without exposing the implementation details there.

A sketch of the `h.api.views#search` view would be something like this, then:

```python
@api_config(context=Annotations)
def search(context, request):
    """Search the database for annotations matching with the given query."""
    query = context.query(request)
    results = models.Annotation.search_raw(body=query, raw_result=True,
                                           authorization_enabled=False)

    total = results['hits']['total']
    docs = results['hits']['hits']
    rows = [models.Annotation(d['_source'], id=d['_id']) for d in docs]

    return {"rows": rows, "total": total}
    return {
        'total': total,
        'rows': rows,
    }
```

Here, `context.query` would use the query builder to assemble the query for this context, but there is no longer any single `search` library code that knows how to build a query from a bunch of filters like `AuthFilter`, and `UriFilter`, because whether those apply is specific to the collection being queried.

PR #2494, opened 2015-09-12
Split host / sidebar and clean up plugin loading

tilgovi:
Clean up the way we decide how to load and configure the application and provide more example application classes.

In other words, sub-classing instead of piling on options for things.

From the documentation change:

```rst
``Annotator.Guest`` expects to connect to an annotator widget running in a
different frame. Any number of instances can communicate with a single widget
in order to provide annotation of many frames.

``Annotator.Host`` is an extended version of ``Annotator.Guest`` that will
instantiate an annotator widget by loading the location given by the ``app``
option in an iframe and appending it to the document.

``Annotator.Sidebar`` is an extended ``Annotator.Host`` that puts the widget
in a sidebar interface. It loads additional plugins that show a bar of bucket
indicators, each providing the ability to select a cluster of highlights, and a
toolbar that can be used to resize the widget and control other aspects of the
user interface.

``Annotator.PdfSidebar`` is a custom version of ``Annotator.Sidebar`` with
defaults tailored for use in a PDF.js viewer.
```

This is proposed as an alternative to #2490.

By using `Annotator.Host` instead of `Annotator.Sidebar` it is trivial to load a version of the application that creates the annotator widget but doesn't load the toolbar or bucket bar and doesn't open or close. This leaves the integrator to control hiding, showing or resizing the widget and providing the controls normally on the toolbar.

### DIFF
--- a/h/static/scripts/annotator/guest.coffee
+++ b/h/static/scripts/annotator/guest.coffee
@@ -5,8 +5,6 @@ raf = require('raf')
 scrollIntoView = require('scroll-into-view')
 
 Annotator = require('annotator')
-Annotator.Plugin.BucketBar = BucketBar = require('./plugin/bucket-bar')
-Annotator.Plugin.CrossFrame = CrossFrame = require('./plugin/cross-frame')
 $ = Annotator.$
 
 highlighter = require('./highlighter')
@@ -33,11 +31,9 @@ module.exports = class Guest extends Annotator
     ".annotator-hl mouseover":           "onHighlightMouseover"
     ".annotator-hl mouseout":            "onHighlightMouseout"
 
-  # Plugin / Options configuration
   options:
-    TextHighlights: {}
+    Document: {}
     TextSelection: {}
-    clickToClose: true
 
   # Anchoring module
   anchoring: require('./anchoring/html')
@@ -131,24 +127,20 @@ module.exports = class Guest extends Annotator
 
   _setupWrapper: ->
     @wrapper = @element
-    if @options.clickToClose
-      @wrapper.on 'click', (event) =>
-        if !@selectedTargets?.length
-          this.hideFrame()
-    return this
+    this
 
   # These methods aren't used in the iframe-hosted configuration of Annotator.
-  _setupDynamicStyle: -> this
   _setupViewer: -> this
   _setupEditor: -> this
   _setupDocumentEvents: -> this
+  _setupDynamicStyle: -> this
 
   destroy: ->
     $('#annotator-dynamic-style').remove()
 
     @adder.remove()
 
-    @wrapper.find('.annotator-hl').each ->
+    @element.find('.annotator-hl').each ->
       $(this).contents().insertBefore(this)
       $(this).remove()
 
@@ -331,7 +323,7 @@ module.exports = class Guest extends Annotator
     else
       # Show the adder button
       @adder
-        .css(Annotator.Util.mousePosition(event, @wrapper[0]))
+        .css(Annotator.Util.mousePosition(event, @element[0]))
         .show()
 
     true
@@ -388,14 +380,6 @@ module.exports = class Guest extends Annotator
 
     @visibleHighlights = shouldShowHighlights
 
-  # Open the sidebar
-  showFrame: ->
-    @crossframe?.call('open')
-
-  # Close the sidebar
-  hideFrame: ->
-    @crossframe?.call('back')
-
   onAdderMouseup: (event) ->
     event.preventDefault()
     event.stopPropagation()
@@ -412,5 +396,4 @@ module.exports = class Guest extends Annotator
         this.createHighlight()
       when 'comment'
         this.createAnnotation()
-        this.showFrame()
     Annotator.Util.getGlobal().getSelection().removeAllRanges()
--- a/docs/hacking/customized-embedding.rst
+++ b/docs/hacking/customized-embedding.rst
@@ -1,44 +1,35 @@
 Customized embedding
 ####################
 
-To customize the plugins that are loaded, define a function ``window.hypothesisConfig``
-which returns an options object::
+To customize the application, define a function ``window.hypothesisConfig``
+which returns an options object.
 
+The ``constructor`` property should be used to select an annotation
+application. Four are provided: ``Annotator.Guest``, ``Annotator.Host``,
+``Annotator.Sidebar`` and ``Annotator.PdfSidebar``.
+
+``Annotator.Guest`` expects to connect to an annotator widget running in a
+different frame. Any number of instances can communicate with a single widget
+in order to provide annotation of many frames.
+
+``Annotator.Host`` is an extended version of ``Annotator.Guest`` that will
+instantiate an annotator widget by loading the location given by the ``app``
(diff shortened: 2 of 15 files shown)


JakeHartnell:
> By using Annotator.Host instead of Annotator.Sidebar it is trivial to load a version of the application that creates the annotator widget but doesn't load the toolbar or bucket bar and doesn't open

I need to open the sidebar when someone clicks on the adder or a highlight. Am I supposed to add my own event listeners for clicks on the adder and highlights? This is a pain in the ass since we are dealing with iframes and onAdderClick calls `event.preventDefault()` and `event.stopPropagation?()`.

> This leaves the integrator to control hiding, showing or resizing the widget and providing the controls normally on the toolbar.

Again, I think the problem is when some clicks on either a highlight or the adder. The sidebar should open but I don't see this PR as giving me (the integrator) full control over hiding / showing the sidebar in this area.

Thoughts? Am I missing something? I've spent an hour playing with this, and didn't see how it could accomplish what I need.

> In other words, sub-classing instead of piling on options for things.

Ok, so is this what we'll have integrators doing, making sub-classes? Would you rather I just made a subclass for the Epub integrations?

tilgovi:
I think subclassing is a better approach for now. Eventually, there should be no base class, because the components are isolated enough and easy enough to import and use that you just construct it from scratch.

If you look at the new contents of sidebar.coffee is it not clear how you would make your integration? You have full control just by overriding a couple methods. Check out the definitions of create annotation and selectAnnotations there.

BigBlueHat:
@tilgovi so I like where this is headed, but I do understand (having just tried this myself) that it's not clear (yet) without lots of digging into the CoffeeScript what all needs extending, preventing, etc.

I think this does pave the way for making it all more obvious, and I'm :+1: for this refactoring regardless of whether it's the immediate solution @JakeHartnell needs.

However, I would like to see if we can help @JakeHartnell close the loop on this epub.js integration with some more explanation / sample code and/or a chat with the three of us.

Here's some code @JakeHartnell posted earlier: https://github.com/futurepress/epub.js/blob/3286aabddaa47db8ad87536aaad2312acfe24944/examples/hypothesis.html#L24-L30

Given that code (pasted here):
```javascript
              return {
                app: "http://localhost:5000/app.html",
                Toolbar: {container: '#annotation-controls'}
              }
            };
```

I'd assume we need to make it something like this?
```javascript
function EPubSidebar() {
  // ...extend Sidebar (or Guest even) to extend / prevent the events to match your use case
}

  return {
    app: "http://localhost:5000/app.html",
    constructor: EpubSidebar
  }
};
```

What's unclear (to me still) is what all @JakeHartnell would need to customize in `EpubSidebar` to get the showing and hiding the way he wants.

Obviously the configuration based way is more "obvious" but I agree that this method is far more scalable, customizable (considering all the scenarios we can't conceivably offer config options for...), and is the right direction. A little bit more help, and I think both @JakeHartnell can pick the :football: up and run with it the rest of the way. :smile:

tilgovi:
Start with Host, not Sidebar, as the constructor and there is no hide/show/resize/toolbar/bucketbar.

Subclass Host and do what you need to. Use Sidebar as an example code. For instance, Sidebar overrides createAnnotation to open the sidebar at that moment. I'm happy to help, but I need more information about what I can help with.

BigBlueHat:
Ah. That makes more sense. :smiley: 

@JakeHartnell just copy `sidebar.coffee` into your project (turn it into JS if you want), and customize it to extend `Annotator.Host` and remove/add the bits you need.

Seems pretty straightforward and obviously very flexible in what you could produce--vs. just "Look out! A Sidebar's incoming!!!" :wink:

tilgovi:
If I can do anything else to help, just let me know.

h/static/scripts/annotator/guest.coffee
```diff
@@ -33,11 +31,9 @@ module.exports = class Guest extends Annotator
     ".annotator-hl mouseover":           "onHighlightMouseover"
     ".annotator-hl mouseout":            "onHighlightMouseout"
 
-  # Plugin / Options configuration
   options:
-    TextHighlights: {}
```
nickstenning: Does this mean that `Guest` now won't display highlights by default?

nickstenning:
Only one little question about the changes to `Guest`, otherwise this looks pretty nice.

h/static/scripts/annotator/guest.coffee
tilgovi: No. Sorry that was a confusing diff. That plugin hasn't existed for a while, since no-dtm maybe, and this line had no effect.

tilgovi:
Okay, I took out the language about plugins because I think we might well wish to steer away from that, similar to how Annotator has moved from very object oriented "plugins" to very functional "extensions".

BigBlueHat:
So. It isn't as straight forward as I'd hoped... :frowning: #2584 

This new setup only works for customizing / extending the `Annotator.Host` object from *within* the `h` code base--there's no way (afaict) to do this from "outside" by including something like `embed.js` and then configuring / setting that up in a custom way--which I'm pretty sure is what @JakeHartnell was after.

@tilgovi am I misunderstanding a piece to this puzzle? If I'm extending `Annotator.Host` from outside the `h` code base, I have to get access to it in some way--and currently it's not in the global scope nor include-able directly. :confused: 

Here's what Jake was attempting earlier:
https://github.com/futurepress/epub.js/commit/3286aabddaa47db8ad87536aaad2312acfe24944#diff-30ac643a3215fbb8d838c689e14d3823R18

 - include `embed.js`
 - customize with `window.hypothesisConfig`
 - ...profit?

@nickstenning / @judell this is essentially blocking @JakeHartnell's work on EPUB integration--unless we move that into the core of the product. Regardless, your input(s) here are also welcome.

judell:
> This new setup only works for customizing / extending the Annotator.Host object from within the h code base

That's a tradeoff resulting from the switch from configurable callbacks (https://github.com/hypothesis/h/pull/2490) to subclassing, right? Should we reevaluate the tradeoffs in light of current requirements? I know several would-be integrators who would rather not have to extend from within.

> unless we move [epub] into the core of the product

That seems worth discussing separately from all-up front-end extensibility. EPUB is (or should be) a core capability.

BigBlueHat:
It's less a tradeoff and more that there's still a step (or two) to complete it--the big one being making the "parent" objects accessible to children begat outside the `h` codebase. That's currently not possible, but could be with relative ease--the decision to do it, however, being the one hard part.

The EPUB discussion can certainly be had in parallel and likely doesn't need to be blocked by this discussion / decision.

tilgovi:
If you need to subclass, the constructor should be globally available during the configuration callback. It's not the cleanest thing in the world but it does work.

To improve this, we could have annotator/main.js export the `Annotator` symbol and move the actual bootstrapping into embed.js.

tilgovi:
Also, we should use the mailing list rather than closed issues for talking about improvements to the status quo.

JakeHartnell:
> Also, we should use the mailing list rather than closed issues for talking about improvements to the status quo.

+1, there's already an actual issue open for this as well: https://github.com/hypothesis/h/issues/2584

PR #2495, opened 2015-09-15
Improve factoring of search/filtering

nickstenning:
Testing and changing the the `h.api.query.build` function was getting somewhat unwieldy.

This commit refactors the query builder into a class, `query.Builder`, the behaviour of which can be extended by filters and matchers.

Filters return match conditions for an Elasticsearch filter which restrict the search results to ONLY those matching the returned filter object.

Matchers return match conditions for an Elasticsearch query, and form part of a match clause in which annotations SHOULD match the filter to be returned from the query.

Both filters and matchers are defined to be callables which accept the query parameters only.


tilgovi:
I like this a lot. I wonder if there should be `NipsaFilter` as a class in the same way and make that consistent, but that doesn't block this for me.

tilgovi:
good stuff

PR #2498, opened 2015-09-16
Enable annotations on H documentation

robertknight:
Add Hypothes.is to H's documentation pages so
users can provide feedback or ask questions
on issues when deploying or hacking H.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -98,6 +98,11 @@ pygments_style = 'tango'
 # A list of ignored prefixes for module index sorting.
 #modindex_common_prefix = []
 
+# Hypothes.is customizations using extension API
+# see http://sphinx-doc.org/latest/extdev/index.html
+def setup(app):
+    # Add annotation to docs pages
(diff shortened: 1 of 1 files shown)


nickstenning:
Nice.

tilgovi:
I'm not seeing it on http://h.readthedocs.org/en/latest/INSTALL.html

robertknight:
@tilgovi - The documentation currently on h.readthedocs.org was last updated from commit 90026ddb86917b4b534d0124eea37e6664d7f5c1 (see the READTHEDOCS_DATA var near the top of the page source). AIUI RTD generation should be triggered via a web hook on commit, but I don't know if there are any filters - eg. does it only regenerate if particular files are modified by a commit?

nickstenning:
Fixed.

tilgovi:
Weird. It's still not consistently on all pages.

tilgovi:
Oh, never mind. I don't know what I'm seeing. Maybe it was just slow.

PR #2500, opened 2015-09-16
Fix groups views -- check user's auth status directly

nickstenning:
Rather than relying on the rather oddly-named[^1] "authenticated"
permission, which was removed (probably by accident) in 998347a, just
check the user's authentication status directly in these views.

[^1]: Permissions are usually verbs, not adjectives.

### DIFF
--- a/h/groups/views.py
+++ b/h/groups/views.py
@@ -15,11 +15,13 @@ _ = i18n.TranslationString
 
 @view_config(route_name='group_create',
              request_method='GET',
-             renderer='h:groups/templates/create.html.jinja2',
-             permission="authenticated")
+             renderer='h:groups/templates/create.html.jinja2')
 def create_form(request):
     """Render the form for creating a new group."""
     if not request.feature('groups'):
+        raise exc.HTTPNotFound()
+
+    if request.authenticated_userid is None:
         raise exc.HTTPNotFound()
 
     schema = schemas.GroupSchema().bind(request=request)
@@ -30,11 +32,13 @@ def create_form(request):
 
 @view_config(route_name='group_create',
              request_method='POST',
-             renderer='h:groups/templates/create.html.jinja2',
-             permission="authenticated")
+             renderer='h:groups/templates/create.html.jinja2')
 def create(request):
     """Respond to a submission of the create group form."""
     if not request.feature('groups'):
+        raise exc.HTTPNotFound()
+
+    if request.authenticated_userid is None:
         raise exc.HTTPNotFound()
 
     form = deform.Form(schemas.GroupSchema().bind(request=request))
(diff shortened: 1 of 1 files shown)


robertknight:
So a [Pyramid ACL](http://docs.pylonsproject.org/docs/pyramid/en/1.0-branch/narr/security.html) as used in the previous code looks like a logical place to define the check that the user has permission to create a group, except that as noted in the PR, permissions should be verbs rather than state adjectives, eg. a 'create-group' permission which happens to be granted to all signed in users.

h/groups/views.py
```diff
@@ -15,13 +15,15 @@
...
-             renderer='h:groups/templates/create.html.jinja2',
-             permission="authenticated")
+             renderer='h:groups/templates/create.html.jinja2')
 def create_form(request):
     """Render the form for creating a new group."""
     if not request.feature('groups'):
         raise exc.HTTPNotFound()
 
+    if request.authenticated_userid is None:
+        raise exc.HTTPNotFound()
```
seanh: Do you know what was happening before (when the 'authenticated' permission still worked)? Was it returning a 404 or was it a 401 or something else?

nickstenning: It's a match predicate for the route, so it would have been a 404.

seanh: Ok. I think I kind of agree with @robertknight that it's more pure in a Pyramid sense to always let the authz system handle authz so that in the case of an authz error the view never even gets called (and the solution he suggests sounds right to me). It means for example that we never have to test any auth stuff when writing tests for view callables. But I'm also fine with doing it the way you have it, I don't think it's a big problem.

nickstenning:
If the 'group-create' permission were used in other places, I might agree. As it is, I don't see any value in splitting the logic across two different locations and requiring someone to understand the pyramid auth system when a simple "is this user authenticated?" check in the relevent view functions suffices.

robertknight:
Fair enough - good to merge on that basis, we can refactor as/when permissions get more complex.

h/groups/views.py
tilgovi: > It's a match predicate for the route, so it would have been a 404.

You're wrong about that, @nickstenning. If the route is protected with a permission predicate it causes an unauthorized response. The only reason we show a 404 is because we have a catch-all for unauthorized that turns it into a 404.

nickstenning: Fair enough. Same result, though. This commit doesn't change behaviour from before it broke.

judell: "we have a catch-all for unauthorized that turns it into a 404." Ah. Why?
(This relates to https://github.com/hypothesis/h/issues/2502.)

tilgovi: It's a fairly common approach that we've been taking on the HTML representations all along. It keeps information about whether something exists or not private.

judell: Fair enough. Note, FWIW, that the situation Ron Synder ran into -- which I
just checked myself -- is that 404 is also the response when failing to
create an annotation with bogus permissions. That complicates debugging
auth problems.

PR #2504, opened 2015-09-17
Alias 'tag' to 'tags' in search API params

seanh:


seanh:
@nickstenning

nickstenning:
Gotta run, but this could easily be implemented as a separate matcher class, say `TagMatcher`, which would prevent us from making the core `Builder` class complicated again.

tilgovi:
Why not just change it from `tags` to `tag`?

seanh:
Force pushed a new version of this as a matcher class.

I assume we still want to support tags because people might have code that's using it, the request was to alias it not change it.

seanh:
Added some search API docs updates to this.
@nickstenning What's the status of the `TODO` about the `document` search API param here? file:///home/seanh/Projects/hypothesis/h/docs/_build/html/api.html#get--api-search

tilgovi:
You could change from `tags` to `tag` pretty easily.

- Change the model mapping
- Set up render/prepare to flip it back and forth
- Run `annotool prepare`

Then upgrade the client.

- Set up `annotationsLoaded` to change `tags` into `tag`
- Update the template and code to use `tag`

At some point it's safe to drop the transformation on the server, at which point it's safe to drop the transformation at the client.

h/api/search/query.py
```diff
@@ -163,3 +163,17 @@ def __call__(self, params):
...
         del params["any"]
         return result
+
+
+class TagsMatcher(object):
+
+    """Matches the tags field against 'tag' or 'tags' parameters."""
+
+    def __call__(self, params):
+        tags = set(params.getall('tag') + params.getall('tags'))
```
nickstenning: Sorry to come back to this but I have one minor nitpick. Could you change this line to: `tags = set(v for k, v in params.items() if k in ['tag', 'tags'])`?

That way it's consistent with the rest of this file, which assumes that `params` is a dict-like iterable, not specifically a `MultiDict`.

PR #2507, opened 2015-09-17
Bump gevent version

tilgovi:
This should stop the logging that occurs when clients disconnect.
https://github.com/gevent/gevent/pull/377

### DIFF
--- a/h/server.py
+++ b/h/server.py
@@ -1,5 +1,8 @@
 # -*- coding: utf-8 -*-
+from datetime import datetime
+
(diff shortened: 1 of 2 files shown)


tilgovi:
Also should fix the SSL problems @robertknight found.
Want to try it?

nickstenning:
I'm a bit reluctant to depend on a beta release when this isn't causing us issues in production, but I'll try it out.

tilgovi:
It's not actually important. I just wanted to share it since I did the work.

nickstenning:
Closing -- thanks for the work, and we'll revisit this when gevent 1.1 is released.

PR #2508, opened 2015-09-18
Clean up socket, threading, mapper with routing

tilgovi:
Move all the logic required to set up a clean environment for the
client views into the dependency resolution mechanism of ngRoute.

The annotation mapper gains an unloadAnnotations method to make the
isolated code cleaner.

Use angular-websocket to remove the need for our own socket class.

The AppController becomes much more sane.


nickstenning:
Adds ~5K to `app.min.js`, unfortunately, but I think that's a price worth paying at the moment.

PR #2509, opened 2015-09-18
Allow the sidebar to be resized thinner

seanh:
Reduce the MIN_RESIZE size at which the sidebar will snap shut, when the
user is resizing the width of the sidebar and making it skinnier.

This is helpful for example when trying to relate a set of annotations
in the sidebar to their highlights (or lack thereof) in a long PDF.

### DIFF
--- a/h/static/scripts/annotator/sidebar.coffee
+++ b/h/static/scripts/annotator/sidebar.coffee
@@ -5,7 +5,7 @@ Hammer = require('hammerjs')
 Host = require('./host')
 
 # Minimum width to which the frame can be resized.
-MIN_RESIZE = 280
+MIN_RESIZE = 80
 
 
 module.exports = class Sidebar extends Host
--- a/h/static/scripts/annotator/test/sidebar-test.coffee
+++ b/h/static/scripts/annotator/test/sidebar-test.coffee
@@ -78,7 +78,7 @@ describe 'Sidebar', ->
         assert.calledOnce(show)
 
       it 'calls `hide` if the widget is not fully visible', ->
-        sidebar.gestureState = {final: -100}
+        sidebar.gestureState = {final: -50}
         hide = sandbox.stub(sidebar, 'hide')
         sidebar.onPan({type: 'panend'})
         assert.calledOnce(hide)


tilgovi:
Be careful with this. Being that thin causes the layout to break in the annotation cards, IIRC.

Another way to handle this that I've played with and just never finished is to make the container div of the iframe double wide and absolute when open and then fix the frame within it. That makes the browser think there is content extending off the right hand side and then you get a horizontal scroll bar you can use to move the page _underneath_ the sidebar rather than moving the sidebar.

seanh:
> Be careful with this. Being that thin causes the layout to break in the annotation cards, IIRC.

Yep, it does

> Another way to handle this that I've played with and just never finished is to make the container div of the iframe double wide and absolute when open and then fix the frame within it. That makes the browser think there is content extending off the right hand side and then you get a horizontal scroll bar you can use to move the page underneath the sidebar rather than moving the sidebar.

I guess I don't totally understand what the use-case for this request is, so not sure whether that would serve it just as well. @judell ?

judell:
@seanh as mentioned [here](https://github.com/hypothesis/h/pull/2509) it can be useful to shrink the sidebar beyond the halfway point in order to visually correlate cards in the sidebar with their targets in the document. It was formerly possible, then went away.

tilgovi:
Separate the problem from the solution, @judell. I can be difficult to correlate cards in the sidebar with the targets in the document because they are covered by the sidebar.

Shrinking the sidebar is only one way to fix that.

nickstenning:
I'm really not sure about this. Do we really want to allow people to make the thing look like this?

screen shot 2015-09-21 at 18 24 10

@judell?

nickstenning:
This also has non-obvious implications for what happens when the groups feature (with its "scope selector" dropdown in the top bar) is turned on.

judell:
At one time we did allow it, and I relied on it. Given that some users are surprised to find that the sidebar is drag-resizeable at all, this is by definition a power user feature. As a power user I find this handy and am more than willing to suffer a transient degradation of sidebar formatting in order to reveal a critical piece of the document that's otherwise hidden.

There are, of course, other solutions, including ones that might dovetail with a need to make editing and viewing happen in a mobile-responsive way. And that's the much bigger fish we need to fry. So if this smells too bad to tolerate then kill it and we'll track this as one of the aspects of a mobile/responsive strategy.

dwhly:
Compromise suggestion: 

How about we make it a little bit smaller than currently, to perhaps 30-40% of the default width (not sure what the value is now).  

Whatever the number is that we settle on it's somewhere on a continuum of tradeoffs between these various concerns.  Agree w/ @judell that this is for power users, which means hardly anyone is ever going to discover unless we advertise a bit more.  Generally tho, the good news is that folks can clearly see that making it smaller has a consequence, and can figure out where to release their finger based on what they're seeing.

> This also has non-obvious implications for what happens when the groups feature (with its "scope selector" dropdown in the top bar) is turned on.

Agree.  Lets sanity check this against that coming control.

tilgovi:
> Whatever the number is that we settle on it's somewhere on a continuum of tradeoffs between these various concerns.

Otherwise known as the process I used to arrive at the number that's there now.

dwhly:
> Otherwise known as the process I used to arrive at the number that's there now.

Sure.  Now there are a couple votes for slightly thinner is all.

> Nick: Do we really want to allow people to make the thing look like this?

No. (I don't).

tilgovi:
Sure, you can all vote to have it slightly smaller. My point was that your choice is probably not going to be that significantly different from my own because it's balancing the same concerns. While you can all nevertheless convene a vote for a dubiously marginal improvement, I don't think it's worth the time spent discussing it.

nickstenning:
Sorry, Sean. We've decided to punt on this for now.

PR #2510, opened 2015-09-18
Show orphan annotations with a notice

tilgovi:


tilgovi:
Still needed:

- [ ] Maybe some tests change, I haven't checked yet.
- [ ] Use the feature flag in `shouldShowThread`, test that better

nickstenning:
screen shot 2015-09-21 at 17 55 43

dwhly:
I'd remove the apology.  It's not necessarily our fault.  

"We cannot find the exact position of this annotation".

Also, doublechecking that this will be feature flagged.

nickstenning:
> Also, doublechecking that this will be feature flagged.

:heavy_check_mark:

seanh:
I've rebased this after a pretty nasty merge with master, hope I didn't break anything, tests are still passing and the diff against master looks the same as it did before (if I didn't miss anything)

tilgovi:
Looks right to me at a glance.

seanh:
It doesn't seem to be working though, I'm not seeing the footer on orphaned annotations :( Trying to figure it out..

tilgovi:
Did you enable the feature? Are you seeing orphan annotations at all? Have you made a sufficiently large change to the content to actually cause an annotation to be orphaned?

seanh:
Feature is enabled, I'm testing by annotating a page then completely replacing the contents of the page, I am seeing the orphaned annotation just without footer

h/templates/client/thread.html
```diff
@@ -45,3 +45,7 @@
       ng-show="vm.shouldShowAsReply()">
   </li>
 </ul>
+
+<footer class="thread-anchor-notice" ng-if="vm.container.message.$orphan">
```
tilgovi: Here's the problem. This got messed up in some rebasing or refactoring. I may have pushed it with this problem to begin with.

The flag is `$anchored` not `$orphan`, and the meaning is inverted.

seanh: Yeah, I had found that but it seemed to be that way in your original commit, ok, that fixes it, thanks!

seanh:
Alright, this looks good to me now, will merge when Travis passes

judell:
@nickstenning I'm set up to test this on stage but since this didn't merge until after you updated stage earlier today I don't think I can observe the new behavior there.

seanh:
Oops, https://github.com/hypothesis/h/issues/2521

tilgovi:
Seems this feature was never turned on in production. Was that intentional?

dwhly:
Yep. Prefer to have a real home for orphans.

PR #2511, opened 2015-09-18
Percolator

tilgovi:
This builds on the percolator work.

The event-stream interface is great, but browser support and the polyfill situation are not. Meanwhile, we already have a working websocket.

As a way to make a clean break from the existing websocket code, the idea here is to first refactor the front end to remove all the stuff it does to construct the filter objects it currently sends over the websocket. This makes room for a new protocol between the client and the server that can just send the query string.


tilgovi:
I built this on top of a merge of percolator-sse and cleanup-socket-threading-routes, but the server sent events view doesn't need to stay.

tilgovi:
This is WIP. Everything except the sidebar search and realtime search work. The new websocket protocol is non-existent, and so the client never sends anything over the socket.

The next step is to have the client send the query and have the server use the percolator. That's probably as easy as using the new `parse-query` module to construct the query string, sending it over the socket, and parsing that into a `MultiDict` on the backend, then hooking it up to stream events back from the percolator.

I would make a decorator for registering a `WebSocket` class as a view so the logic around the handshake in `streamer.py` is re-usable. Then, I would add a _new_ websocket endpoint to the backend, switch the client to use that, and build out the protocol.

Bonus points if you clear/ignore the cookies from the request and set the authorization based on a websocket message. It might even be possible to hold the socket open across authorization changes, if that's desirable at all.

tilgovi:
Probably the easiest, fastest way to do the new websocket would be to take an auth token sent over the wire, if any, and the client id, and construct a derived `request` object that has parameters taken from the websocket, no cookies, an authorization and client-id headers taken from the messages. Then, just invoke the server sent events view callable directly, which returns a generator.

So, this is mostly all done, it just needs a new WebSocket route, handshake, and setup messages.

nickstenning:
I'll keep the branch around but close this for now. We'll probably get round to merging this work (or a derivative thereof) once we've got groups out of the way.

Issue #1292, opened 2014-07-09
Remove compass dependency


aron:
This is the first step to moving towards using lib libsass[1][2] and removing the ruby dependency in the project. We only use a small (but useful) subset of the compass functions, mainly for polyfilling, so it seems like we can easily include them directly in the project.

I've listed them all out here:

- Clearfix (pie-clearfix)
- Background images (background, background-image)
- Responsive (respond-to)
- Box shadow (box-shadow)
- Transition (single-transition, transition)
- Box model (box-sizing)
- Reset (reset-font, reset-box-model)
- Transition (transition, transition-duration, transition-property, transition-timing-function)
- Transform (scaleX, scale, rotateX)
- Wrapping (force-wrap)
- SVG (stretch-y)

[1] https://github.com/sass/libsass
[2] https://github.com/sass/node-sass

nickstenning:
Hi there! I'm going to close this as part of a clean-up of all issues currently open on this repo that represent ideas or features rather than reports of bugs or technical chores. 

I want to be clear that **this isn't intended to say anything at all about the content of this issue**—it certainly doesn't mean we're no longer planning to do the work discussed here—just that we don't want to use GitHub issues to track feature requests or ideas, because the threads can get long and somewhat unwieldy.

If you're interested in what we are working on at the moment, you can check out [our Trello board](https://trello.com/b/o2ungX1n/current-sprint) and, for a longer-term view, [our roadmap](https://hypothes.is/roadmap/).

And, if you're interested in following up on this issue, please do continue the discussion on our [developer community mailing list](https://groups.google.com/a/list.hypothes.is/forum/#!forum/dev). You might also want to check out [our contributing guide](https://github.com/hypothesis/h/blob/master/CONTRIBUTING.rst).

nickstenning:
For what it's worth, we finally managed to do this: #2958.

PR #2514, opened 2015-09-18
T87 - Improve group scope dropdown ui

robertknight:
This is the initial cut of the improved group link sharing UI.

Remaining tasks:
 * Implement suitable tests
 * Clarify design when there are many groups in the list. Currently it will grow ad-infinitum

group-share-ui-update

### DIFF
--- a/h/static/scripts/directive/group-list.js
+++ b/h/static/scripts/directive/group-list.js
@@ -1,17 +1,64 @@
 'use strict';
 
+var assert = require('assert');
+
+function GroupsListController($scope) {
+  $scope.expandedGroupId = undefined;
+
+  // used to keep the dropdown from closing when the user
+  // interacts with the inline link share pane within the groups list
+  $scope.stopClickPropagation = function($event) {
+    $event.stopPropagation();
+  }
+
+  // show the share link for the specified group or clear it if
+  // null
+  $scope.toggleShareLink = function (groupId) {
+    if (!groupId || $scope.expandedGroupId === groupId) {
+      $scope.expandedGroupId = undefined;
+    } else {
+      $scope.expandedGroupId = groupId;
+    }
+  };
+
+  $scope.shouldShowShareLink = function (groupId) {
+    return $scope.expandedGroupId === groupId;
+  }
+
+  $scope.linkForGroup = function (groupId) {
+    return $scope.baseURI + 'groups/' + groupId;
+  }
+}
+
 /**
  * @ngdoc directive
  * @name groupList
  * @restrict AE
  * @description Displays a list of groups of which the user is a member.
  */
-
 // @ngInject
 module.exports = function (groups) {
   return {
-    link: function (scope, elem, attrs) {
-      scope.groups = groups;
+    controller: ['$scope', GroupsListController],
+    link: function ($scope, elem, attrs) {
+      $scope.groups = groups;
+
+      // set the base URI used later to construct the sharing
+      // link for the group
+      $scope.baseURI = elem[0].ownerDocument.baseURI;
+
+      $scope.$watch('expandedGroupId', function (activeGroupId) {
+        if (activeGroupId) {
+          // wait for the share link field to be revealed and then select
+          // the link's text
+          setTimeout(function() {
+            var activeShareLinkField = elem[0].querySelector('.share-link-field[data-group-id=' + activeGroupId + ']');
+            assert(activeShareLinkField);
+            activeShareLinkField.focus();
+            activeShareLinkField.select();
+          }, 0);
+        }
+      });
     },
     restrict: 'AE',
     scope: {},
--- a/h/static/scripts/directive/group-list.js
+++ b/h/static/scripts/directive/group-list.js
@@ -32,7 +32,7 @@ function GroupsListController($scope) {
  * @description Displays a list of groups of which the user is a member.
  */
 // @ngInject
-module.exports = function (groups, $window) {
+function GroupsListDirective(groups, $window) {
   return {
     controller: GroupsListController,
     link: function ($scope, elem, attrs) {
@@ -58,4 +58,9 @@ module.exports = function (groups, $window) {
     scope: {},
     templateUrl: 'group_list.html'
   };
+};
+
+module.exports = {
+  directive: GroupsListDirective,
+  _Controller: GroupsListController
 };
--- a/h/static/scripts/directive/group-list.js
+++ b/h/static/scripts/directive/group-list.js
@@ -23,6 +23,15 @@ function GroupsListController($scope) {
   $scope.shouldShowShareLink = function (groupId) {
     return $scope.expandedGroupId === groupId;
   }
+
+  $scope.sortedGroups = function () {
+    return $scope.groups.all().concat().sort(function (a, b) {
+      if (a.public !== b.public) {
+        return a.public ? -1 : 1;
+      }
+      return a.name.localeCompare(b.name);
+    });
+  }
 }
 
 /**
--- a/h/static/scripts/directive/group-list.js
+++ b/h/static/scripts/directive/group-list.js
@@ -1,14 +1,8 @@
 'use strict';
 
 // @ngInject
-function GroupsListController($scope) {
+function GroupListController($scope) {
   $scope.expandedGroupId = undefined;
-
-  // used to keep the dropdown from closing when the user
-  // interacts with the inline link share pane within the groups list
-  $scope.stopClickPropagation = function($event) {
-    $event.stopPropagation();
-  }
 
   // show the share link for the specified group or clear it if
   // null
@@ -41,9 +35,9 @@ function GroupsListController($scope) {
  * @description Displays a list of groups of which the user is a member.
  */
 // @ngInject
-function groupsList(groups, $window) {
+function groupList(groups, $window) {
   return {
-    controller: GroupsListController,
+    controller: GroupListController,
     link: function ($scope, elem, attrs) {
       $scope.groups = groups;
 
@@ -70,6 +64,6 @@ function groupsList(groups, $window) {
 };
 
 module.exports = {
-  directive: groupsList,
-  _Controller: GroupsListController
+  directive: groupList,
+  Controller: GroupListController
 };
--- a/h/static/scripts/directive/group-list.js
+++ b/h/static/scripts/directive/group-list.js
@@ -1,7 +1,6 @@
 'use strict';
 
-var assert = require('assert');
-
+// @ngInject
 function GroupsListController($scope) {
   $scope.expandedGroupId = undefined;
 
@@ -24,10 +23,6 @@ function GroupsListController($scope) {
   $scope.shouldShowShareLink = function (groupId) {
     return $scope.expandedGroupId === groupId;
   }
-
-  $scope.linkForGroup = function (groupId) {
-    return $scope.baseURI + 'groups/' + groupId;
-  }
 }
 
 /**
@@ -37,15 +32,15 @@ function GroupsListController($scope) {
  * @description Displays a list of groups of which the user is a member.
  */
 // @ngInject
-module.exports = function (groups) {
+module.exports = function (groups, $window) {
   return {
-    controller: ['$scope', GroupsListController],
+    controller: GroupsListController,
     link: function ($scope, elem, attrs) {
       $scope.groups = groups;
 
-      // set the base URI used later to construct the sharing
-      // link for the group
-      $scope.baseURI = elem[0].ownerDocument.baseURI;
+      $scope.createNewGroup = function() {
+        $window.open('/groups/new', '_blank');
+      }
 
       $scope.$watch('expandedGroupId', function (activeGroupId) {
         if (activeGroupId) {
@@ -53,7 +48,6 @@ module.exports = function (groups) {
           // the link's text
           setTimeout(function() {
             var activeShareLinkField = elem[0].querySelector('.share-link-field[data-group-id=' + activeGroupId + ']');
-            assert(activeShareLinkField);
             activeShareLinkField.focus();
             activeShareLinkField.select();
           }, 0);
--- a/h/static/scripts/directive/group-list.js
+++ b/h/static/scripts/directive/group-list.js
@@ -32,7 +32,7 @@ function GroupsListController($scope) {
  * @description Displays a list of groups of which the user is a member.
  */
 // @ngInject
-function GroupsListDirective(groups, $window) {
+function groupsList(groups, $window) {
   return {
     controller: GroupsListController,
     link: function ($scope, elem, attrs) {
@@ -61,6 +61,6 @@ function GroupsListDirective(groups, $window) {
 };
 
 module.exports = {
-  directive: GroupsListDirective,
+  directive: groupsList,
   _Controller: GroupsListController
 };
--- a/h/static/scripts/directive/test/group-list-test.js
+++ b/h/static/scripts/directive/test/group-list-test.js
@@ -1,0 +1,46 @@
+'use strict';
+
+var groupsList = require('../group-list');
+
+describe('GroupsListController', function () {
+  var controller;
+  var $scope;
+
+  beforeEach(function () {
+    $scope = {};
+    controller = new groupsList._Controller($scope);
+  });
+
+  it('toggles share links', function () {
+    $scope.toggleShareLink('group-a');
+    assert.equal($scope.expandedGroupId, 'group-a');
+    $scope.toggleShareLink('group-a');
+    assert.equal($scope.expandedGroupId, undefined);
+
+    $scope.toggleShareLink('group-b');
+    assert.equal($scope.expandedGroupId, 'group-b');
+    $scope.toggleShareLink('group-c');
+    assert.equal($scope.expandedGroupId, 'group-c');
+  });
+
+  it('shows share link for selected group', function () {
+    assert.equal($scope.shouldShowShareLink('group-a'), false);
+    $scope.toggleShareLink('group-a');
+    assert.equal($scope.shouldShowShareLink('group-a'), true);
+    $scope.toggleShareLink('group-b');
+    assert.equal($scope.shouldShowShareLink('group-a'), false);
+    assert.equal($scope.shouldShowShareLink('group-b'), true);
+  });
+});
+
+
+// <groups-list> directive
+// - check that it renders all visible groups
+// - check that share links visible for non-public groups
+// - check that share link is focused after toggling share link
+
+// TODO
+// - read Angular unit testing guide
+// - read Angular E2E testing guides
+// - grok an existing directive test
+// - get a trivial <groups-list> directive test failing, then make it work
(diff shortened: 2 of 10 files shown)


dwhly:
Nice!

I kind of like the copy URL inlined in the card.  Agree that dealing w/ long lists is important.

robertknight:
Perhaps we should do the same for creating new groups as well. Currently the process of creating a new group looks like this:

1. Click 'New group'
2. That opens up a very spartan tab that asks for a single piece of data (the group name) which shows a 'Create Group' button.
3. Clicking 'Create Group' creates the group and gives you a link, but when you go back to the original tab to make use of the new group, it doesn't appear until you refresh the page.

Inlining the 'Create Group' action into the above menu would make the process slicker. The potential downside is that you wouldn't have room to offer any explanations about groups in that menu, but would have to link out to a separate page for that.

dwhly:
> The potential downside is that you wouldn't have room to offer any explanations about groups in that menu, but would have to link out to a separate page for that.

That in and of itself might be a DQ.

dwhly:
So, been going back over @conordelahunty's designs after my somewhat off the cuff remarks earlier.  Follow up thoughts:

1- We should stick to a design process here, and respect our designer (not that you were intending otherwise!) -- giving him feedback and letting him evolve the design in partnership with us, but ultimately having him own it.

2- On this particular suggestion, from a consistency POV, it's definitely important to maintain consistency between creating and sharing a group.  Also, the designs as made provide for the social sharing icons in addition to just the form field, and as you suggest adding the rational / description for a group is difficult in the smaller space and something we've discussed before.  (Card for it: https://trello.com/c/DByfGdQt/264-add-descriptions-for-groups)

robertknight:
> So, been going back over @conordelahunty's designs after my somewhat off the cuff remarks earlier.

To clarify, there have been several rounds of design, the one I'm working to is https://trello.com/c/08dCx7gd/87-group-scope-dropdown-ui-updates which came out of the meeting on Tuesday this week.

> 1- We should stick to a design process here, and respect our designer (not that you were intending otherwise!) -

Yes, of course. What is actually implemented in this PR follows the [original spec](https://trello.com/c/08dCx7gd/87-group-scope-dropdown-ui-updates) (at least, assuming I've correctly understood all aspects of it).

dwhly:
Ack. Forgive, I must have been looking at the completely wrong card. Apologies!  Feeling like an idiot now.

h/static/scripts/directive/group-list.js
```diff
@@ -1,17 +1,64 @@
...
  * @restrict AE
  * @description Displays a list of groups of which the user is a member.
  */
-
 // @ngInject
 module.exports = function (groups) {
   return {
-    link: function (scope, elem, attrs) {
-      scope.groups = groups;
+    controller: ['$scope', GroupsListController],
```
nickstenning: If you add an `// @ngInject` comment before the definition of `GroupsListController`, the array-form function definition isn't necessary.

h/static/scripts/directive/group-list.js
```diff
@@ -1,17 +1,64 @@
...
+    link: function ($scope, elem, attrs) {
+      $scope.groups = groups;
+
+      // set the base URI used later to construct the sharing
+      // link for the group
+      $scope.baseURI = elem[0].ownerDocument.baseURI;
```
nickstenning: You shouldn't need to construct group URLs. The data sent from the server for groups [includes a prerendered URL](https://github.com/hypothesis/h/blob/4a324db9c3495e71a4a1c3752617d7b9e957e195/h/session.py#L40-L42).

h/static/styles/app.scss
```diff
@@ -82,27 +82,82 @@ ol {
...
-}
+    .group-name {
+      overflow: hidden;
+      text-overflow: ellipsis;
+      width: $group-list-width - 30px;
+    }
+  }
+
+  .group-item {
+    display: flex;
```
nickstenning: Nice :)

h/templates/app.html.jinja2
```diff
@@ -29,10 +29,8 @@
            on-clear="search.clear()"></div>
       <!-- / Searchbar -->
 
-      <div class="group-list"
-           ng-if="auth.user && feature('groups')",
-           group-list="">
-      </div>
+      <group-list class="group-list" ng-if="auth.user && feature('groups')">
```
nickstenning: :+1: To element directives.

nickstenning:
Okay, this is looking really good. Just a couple of notes in the code, and a couple of notes on visuals:

- Comparing with @conordelahunty's mockup, I think the line-spacing between the group name and the "Invite others" link needs tightening up a bit.
- The "New group" link should have a lowercase 'g', and the "add" icon needs a heavier weight to match up with the other icons in the dropdown.

conordelahunty:
@robertknight (catching up on your comments from a couple of days ago, sorry) I wanted to keep the group creation out of the dropdown for exactly the reason you mention, not enough space to explain what to do. I'd like it to be more inline with the group invitation page, I will have to provide you with an updated design for that.

@dwhly Sorry about the minor confusion, we made a couple of quick calls on things last week when we were working through the details and I didn't flag them with you.

conordelahunty:
@robertknight Looking good btw. How would you like me to feedback on visual bugs? 
We can do it on a case by case or do a sweep of them later?

robertknight:
> How would you like me to feedback on visual bugs? 

Annotate a screenshot of the bug with a big arrow pointing out the error :)

>  We can do it on a case by case or do a sweep of them later?

I prefer feedback earlier rather than later, so I think case-by-case is preferred.

robertknight:
scope-dropdown-impl

dwhly:
> Sorry about the minor confusion, we made a couple of quick calls on things last week when we were working through the details and I didn't flag them with you.

No worries.

So, we don't need to necessarily address now, but I *do* think we should generally maintain consistency across our similar interfaces.   As @robertknight observes, the invite and create interfaces are two sides of the same coin.

For instance, the group create dialog gives you a URL to copy-paste and then a set of social shortcuts which create stub emails/tweets/fb posts for you.  On mobile this can be a particular advantage (by decreasing the number of compound interactions required to do copy, and then hunt for a given app to paste the clipboard to).  If that's a useful paradigm, we should carry it throughout?

(Personally I'm not bothered by the dedicated card for group creation.  It's the same number of clicks to reveal / dismiss it as the invite UI?)

conordelahunty:
@robertknight @dwhly Agreed on the consistency. Take a look at a revised design of group creation below. This last step could also be how we share the group.

groups-creation-process

nickstenning:
LGTM. @conordelahunty?

h/static/scripts/directive/group-list.js
```diff
@@ -1,17 +1,61 @@
...
 module.exports = function (groups) {
   return {
-    link: function (scope, elem, attrs) {
-      scope.groups = groups;
+    controller: GroupsListController,
+    link: function ($scope, elem, attrs) {
+      $scope.groups = groups;
+
+      $scope.createNewGroup = function() {
+        window.open('/groups/new', '_blank');
```
nickstenning: I'm not sure I understand why this became JS rather than an href?

conordelahunty:
@robertknight As for the shot you pasted above - move the invite link closer to the title by 3-4px. Also, can you reduce the line height of the "You can invite other people to..." copy. And add a little more top and bottom padding to the "+ New group" cell.

dropdown-feedback

dwhly:
We should be consistent in the share options we support.   From the share dialog for pages, we have:

![screen shot 2015-09-21 at 9 00 13 am]

(https://cloud.githubusercontent.com/assets/1413074/9997276/1f03b474-6040-11e5-99b7-d07cabd0394c.png)

Also, should we be consistent in the UI style between these various actions?  I'm sensing that there's a potential shift to a new aesthetic paradigm here, but it's not complete.  Should we keep the current style until we're ready to make a uniform shift?

h/static/scripts/directive/group-list.js
nickstenning: Ah, understood. Because the "New group" button is different to the rest.

nickstenning:
In the light of @conor's comments, it might make more sense to special-case the "New group" button in the CSS.

nickstenning:
> Should we keep the current style until we're ready to make a uniform shift?

I don't think so. Incremental improvement is good.

h/static/scripts/directive/group-list.js
```diff
@@ -1,17 +1,61 @@
...
+      }
+
+      $scope.$watch('expandedGroupId', function (activeGroupId) {
+        if (activeGroupId) {
+          // wait for the share link field to be revealed and then select
+          // the link's text
+          setTimeout(function() {
+            var activeShareLinkField = elem[0].querySelector('.share-link-field[data-group-id=' + activeGroupId + ']');
+            assert(activeShareLinkField);
```
nickstenning: Not sure how I feel about assertions in production code. I'd rather we had a `if (!activeShareLinkField) { console.warn(...); }`.

conordelahunty:
> Should we keep the current style until we're ready to make a uniform shift?
> I don't think so. Incremental improvement is good.

Incremental improvement is definitely preferred, but I understand the concerns. The changes I am trying to make will hopefully feel harmonious, but are very much intentional baby steps forward. I can simplify some of the things I am proposing. @dwhly which elements worry you the most?

h/static/scripts/directive/group-list.js
robertknight: The problem with that is that you end up introducing untested code paths. The assert() is redundant here really though since the call to `activeShareLinkField.focus()` would produce a more useful error in this context. I'll remove it.

dwhly:
> Incremental improvement is good.

There are so many places and situations in which that's the case, and I strongly endorse the philosophy generally-- particularly with respect to underlying functionality.  But I *do* think with the UI the application should look and feel consistent.

>  I can simplify some of the things I am proposing. @dwhly which elements worry you the most?

I'm not sure it's simplification per-se that I'm getting at.  It's that these two cards are serving the same type of function with completely different design styles.

Sharing the page (w/ via) and providing social links.

screen shot 2015-09-21 at 9 00 13 am

Sharing a group and providing social links.

screen shot 2015-09-21 at 9 46 44 am

I happen to *like* the look that you're headed towards.  I guess my ask is that we convert things over we do it more or less simultaneously, and put together a dictionary for our overall design vocabulary.

(I'm really not trying to be an obstructionist ass, I swear.)

robertknight:
t87-scope-dropdown

conordelahunty:
@robertknight Thanks Robert, looks tight

h/static/styles/app.scss
tilgovi: You're going to want to use the Compass mixins for this, or I think you'll miss out on IE10 support.

tilgovi: See, for example, https://github.com/hypothesis/h/blob/master/h/static/styles/topbar.scss#L20

conordelahunty:
> (I'm really not trying to be an obstructionist ass, I swear.)

Not at all. It's a fair point. Let me take another quick look and see if we can do something we're both happy with.

h/static/styles/app.scss
robertknight: > You're going to want to use the Compass mixins for this, or I think you'll miss out on IE10 support.

Ah, indeed. Actually Safari will be the bigger issue, they only dropped the prefix in Safari 9.

Has post-processing with [autoprefixer](https://github.com/postcss/autoprefixer) been considered for handling vendor prefixes automatically?

tilgovi: I considered it but wasn't sure it accounted for all the usage.

tilgovi: In other words, until we're not using compass anyway, just use compass.

h/static/scripts/directive/group-list.js
```diff
@@ -1,20 +1,66 @@
...
     },
     restrict: 'AE',
     scope: {},
     templateUrl: 'group_list.html'
   };
 };
+
+module.exports = {
+  directive: GroupsListDirective,
+  _Controller: GroupsListController
```
nickstenning: Not sure about the leading underscore here. If we're exporting we're exporting it, so it should just be `Controller`.

h/static/scripts/directive/group-list.js
```diff
@@ -1,20 +1,66 @@
...
 /**
  * @ngdoc directive
  * @name groupList
  * @restrict AE
  * @description Displays a list of groups of which the user is a member.
  */
-
 // @ngInject
-module.exports = function (groups) {
+function GroupsListDirective(groups, $window) {
```
nickstenning: The directive is a factory function rather than a strict constructor, so for the sake of consistency I'd say htis should be `groupsList` or similar, rather than `GroupsListDirective`.

robertknight: Agreed. There are a couple of other mistakes in this commit, I'll ping on Slack when its ready for review.

h/static/scripts/directive/group-list.js
robertknight: I'm using the underscore to denote a 'private' export just for use in the tests, which I think is a fairly common convention.

nickstenning: I've had this back and forth with Sean a few times in Python land. My view is that if it's private, it's an implementation detail and shouldn't be tested. If it's tested, it's public interface, and shouldn't have its name mangled.

conordelahunty:
@dwhly Hey Dan, took a look at some of the consistency issues we have and put together a little proposal.

I tried to get the group creation, invite, and success/invite others pages look the same consistent. I then tried to get the sharing a document function in the sidebar consistent with the rest of the options. Take a look and let me know what you think.

https://docs.google.com/presentation/d/1Fr681lSNkQCxRUnLsi8WGFocpRQ-n5kaR05dDAwc3lc/edit?usp=sharing

h/static/scripts/directive/group-list.js
robertknight: I think would consider the controller an implementation detail but the advantage of some white-box testing of units is that it makes debugging easier if something fails and the tests run faster (in this case, just some trivial function calls instead of rendering HTML in a browser), so you can have more of them.

I've added tests for the directive now which test the toggle at the component level as well.

h/static/scripts/directive/group-list.js
```diff
@@ -1,20 +1,75 @@
...
 /**
  * @ngdoc directive
  * @name groupList
  * @restrict AE
  * @description Displays a list of groups of which the user is a member.
  */
-
 // @ngInject
-module.exports = function (groups) {
+function groupsList(groups, $window) {
```
nickstenning: Everywhere else this directive is referred to as `groupList`, so both this and the controller should probably be `group*` (singular) rather than `groups*` (plural).

h/static/scripts/directive/group-list.js
```diff
@@ -1,20 +1,75 @@
...
     },
     restrict: 'AE',
     scope: {},
     templateUrl: 'group_list.html'
   };
 };
+
+module.exports = {
+  directive: groupsList,
+  _Controller: GroupsListController
```
nickstenning: I hear your explanation, but I'm not convinced. Let's just call this `Controller` and be done with it. There's no harm done by exporting it properly.

h/static/scripts/directive/test/privacy-test.js
```diff
@@ -1,6 +1,6 @@
 'use strict';
 
-var PrivacyController = require('../privacy').PrivacyController;
+var PrivacyController = require('../privacy')._Controller;
```
nickstenning: I guess you're changing this in order to be consistent with the changes you've made to the `groupList` directive. That's fine, but ideally we'd keep changes like this in separate commits, if not in an entirely separate PR.

But I'm not a great fan of the change to start with, so perhaps just back this out?

h/static/styles/app.scss
nickstenning: You still need to address this. For an example, see [this commit](https://github.com/hypothesis/h/commit/3039032).

h/static/scripts/directive/group-list.js
```diff
@@ -1,20 +1,75 @@
...
+    } else {
+      $scope.expandedGroupId = groupId;
+    }
+  };
+
+  $scope.shouldShowShareLink = function (groupId) {
+    return $scope.expandedGroupId === groupId;
+  }
+
+  $scope.sortedGroups = function () {
```
nickstenning: I'd rather we just sorted them on the server side and had the client render them in the order given.

h/static/scripts/directive/test/group-list-test.js
```diff
@@ -0,0 +1,135 @@
...
+// returns true if a jQuery-like element has
+// been hidden directly via an ng-show directive.
+//
+// This does not check whether the element is a descendant
+// of a hidden element
+function isElementHidden(element) {
+  return element.hasClass('ng-hide');
+}
+
+describe('GroupsListDirective', function () {
```
nickstenning: `s/GroupsListDirective/groupList/`

h/templates/client/group_list.html
```diff
@@ -2,29 +2,60 @@
...
+        <!-- the group name and share link !-->
+        <div class="group-details">
+          <div class="group-name-container">
+            <a class="group-name-link"
+               href=""
+               title="{{ group.public ? 'Show public annotations' : 'Show and create annotations in ' + group.name }}">
+               {{group.name}}
+            </a>
+          </div>
+          <div class="share-link-container" ng-click="stopClickPropagation($event)" ng-if="!group.public">
```
nickstenning: Not sure the indirection of `stopClickPropagation` adds anything here. `ng-click="$event.stopPropagation()"` would achieve the same thing and is equally clear, I think.

h/static/scripts/directive/group-list.js
robertknight: > If you need a hand making that change, just shout.

Yes - I haven't looked into the backend that serves the API yet at all having a few pointers would be useful.

h/templates/client/group_list.html
robertknight: Agreed - done.

h/static/scripts/directive/test/group-list-test.js
robertknight: Done

h/static/scripts/directive/group-list.js
robertknight: Done

h/static/styles/app.scss
robertknight: @tilgovi - Is there a previous discussion about removing compass somewhere or does anyone else know what needs to be done? Ideally I'd like all the client deps to be handled via npm

nickstenning: https://github.com/hypothesis/h/issues/1292

nickstenning:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;08dCx7gd&#x2F;87-group-scope-dropdown-ui-updates"> Group scope dropdown UI updates</a>

nickstenning:
The only thing blocking merge here is updates to the CSS. Let me know if I'm out of date on that.

robertknight:
@nickstenning - That's correct and the CSS issue has been addressed.

nickstenning:
LGTM. Nice work. :cake:

lenazun:
🎉

PR #2517, opened 2015-09-21
Add docs for how authenticate to the api

seanh:


tilgovi:
Please, please, no!! The cookie authentication is _not_ something that should be considered public API of the annotation service.

tilgovi:
Or, to be more positive in my response, add this to a new file that documents the application internals. That the cookie and XSRF flow has found its way into some client libraries is a sad result of having not put forward a better authentication mechanism for the API, but I would definitely not enshrine it in documentation or encourage it for API consumers. It is a flow meant for browsers. Recommending anything but our official client use it at all is to violate the spirit of what it does. For instance, XSRF is cross site request forgery. The whole point is that things using this flow should be served by the application to the browser and never used by a third party. Of course, it's to protect users from third party sites and scripts and stand-alone applications can certainly pretend to be making a same-origin request with cookies, but that's really against the spirit of what's supposed to be happening here.

Please, please move away from cookies and not toward them.

seanh:
Is there any other way for a client to authenticate to our API currently, other than this stuff involving XSRF tokens and cookies? Or does one need to be developed?

tilgovi:
One needs to be developed.

tilgovi:
hypothesis/vision#182 has relevant discussion.

seanh:
Ok, good to know. What do you think about documenting the current situation (as I've done) but adding a big warning to the docs, to the effect of this isn't really meant for third-party clients to use and will break in the future?

The fact is I need a way to do this (it's really useful, when working on h) and there are already a few h API clients out there that have reverse-engineered this and hacked together their own solutions, so I think just helping people out with an example and a warning for now, until we develop something better, might be good?

tilgovi:
With a huge personal :+1: to basic auth.

tilgovi:
<shrug> I don't know. HTTP Basic Auth would probably not have taken you much more time and the benefits are much larger.

tilgovi:
Those libraries didn't reverse engineer this, by the way. I think we offered them personal instruction. That was time we should have spent solving this problem. I'm worried that even with a fat warning in the documentation this still contributes to the sense that this problem isn't urgent and I think it is and has been and should have been prioritized much higher.

seanh:
Reading <https://github.com/hypothesis/vision/issues/182> it looks like HTTP Basic Auth over https was suggested as an interim, "experimental" / "highly subject to change" solution until we get something better based on OAuth.

My vote is that given that the docs in this pull request are already written with a working example, we just publish them as an interim / subject to change solution instead. Implementing a better kind of auth for the API should be on a Trello card in the backlog and prioritised that way.

I don't feel particularly strongly though, this can also just be closed

@judell @nickstenning ?

judell:
"Ok, good to know. What do you think about documenting the current situation (as I've done) but adding a big warning to the docs, to the effect of this isn't really meant for third-party clients to use and will break in the future?

The fact is I need a way to do this (it's really useful, when working on h) and there are already a few h API clients out there that have reverse-engineered this and hacked together their own solutions, so I think just helping people out with an example and a warning for now, until we develop something better, might be good?"

+1

"My vote is that given that the docs in this pull request are already written with a working example, we just publish them as an interim / subject to change solution instead. Implementing a better kind of auth for the API should be on a Trello card in the backlog and prioritised that way."

+1

"Those libraries didn't reverse engineer this, by the way."

I watched @rdhyee do exactly that at our post-I Annotate hack day, I picked up on his work and extended it for my own use, @bbarker followed Raymond's example to create a Perl version.  I'm not sure if @rdmpage followed one of those examples or worked it out for himself. But anyway, this is the de facto mechanism used by a growing number of developers, partners, and integrators, Until we supplant it with dev tokens and/or basic auth we should absolutely provide documentation, and a working example as @seanh has done, with appropriate caveats.

tilgovi:
I don't think basic auth is interim. It's a very sensible authentication form, although it's not a bad idea to limit it to the token endpoint to discourage it if you want.

I helped Raymond at that hackathon, IIRC, but it's not important.

Anyway, I've said my opinion. I don't think this should be done.

tilgovi:
It's not a good mechanism. It's a terrible API. Even with a caveat in the documentation the documentation still encourages its use by virtue of existing.

rdhyee:
Re https://github.com/hypothesis/h/pull/2517#issuecomment-142368777 --> yes, @tilgovi helped me to get my code working.  Without that help, I wouldn't have spent much time trying to reverse engineer hypothes.is.  FWIW, I considered it a great positive that Randall helped me.

tilgovi:
And I was a ball of anxiety and guilt then over having not finished a proper authentication story. I'm downright embarrassed at this point and I'd like to make it right.

rdhyee:
Let me know if there's something I can do to help.

judell:
> yes, @tilgovi helped me to get my code working

I stand corrected, thanks @tilgovi!

> And I was a ball of anxiety and guilt then over having not finished a proper authentication story.

I've seen worse, and we will do better, but the cat's out of the bag on this one, so let's own up to the reality of the current situation.

tilgovi:
I don't think the cat is out, but it will be if you merge this. We know of only two libraries depending on this and we have direct contact with both maintainers. I'm serious about doing the work to fix this. It's important to me.

seanh:
Ok, I'm closing this, I'll send a separate PR with just the syntax highlighting fixes.

@tilgovi A pull request for basic auth or other solution would be great

tilgovi:
Thanks, @seanh. Coming soon.

PR #2518, opened 2015-09-21
Add make target to run karma tests with auto-watch enabled

robertknight:
This provides a convenient way to start Karma tests and auto-run
them when test files are changed, shortening the test cycle by ~10s (time to do a full vs incremental Browserify build).

Because the test files are incorporated into a Browserify bundle
by karma-browserify which watches files for changes via watchify(),
we can turn off Karma's own watching of the test files.

This fixes a problem in watch mode where changing a test spec file would result
in Karma running all the tests twice, once after the test spec file
changed and then again afterwards due to the browserify bundle
being updated a second or two after the test spec file changes.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,6 @@
 SHELL := bash
 PATH := bin:${PATH}
+NPM_BIN = "$$(npm bin)"
 
 # Unless the user has specified otherwise in their environment, it's probably a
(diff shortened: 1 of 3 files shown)


tilgovi:
:+1:

nickstenning:
LGTM.

PR #2520, opened 2015-09-22
Don't crash if annotation document's link is None

### LINKED ISSUE #2519: Crash when creating an annotation with document with "links": None
If you post some JSON like this to annotation create:

```
{
  "document": {"link": None},
  "permissions": {"read": ["group:__world__"]}
}
```

then it crashes:

```
  File "/home/seanh/.virtualenvs/h/local/lib/python2.7/site-packages/pyramid/tweens.py", line 21, in excview_tween
    response = handler(request)
  File "/home/seanh/.virtualenvs/h/local/lib/python2.7/site-packages/pyramid_tm/__init__.py", line 99, in tm_tween
    reraise(*exc_info)
  File "/home/seanh/.virtualenvs/h/local/lib/python2.7/site-packages/pyramid_tm/__init__.py", line 80, in tm_tween
    response = handler(request)
  File "/home/seanh/.virtualenvs/h/local/lib/python2.7/site-packages/pyramid/router.py", line 163, in handle_request
  File "/home/seanh/.virtualenvs/h/local/lib/python2.7/site-packages/pyramid/config/views.py", line 596, in __call__
    return view(context, request)
  File "/home/seanh/.virtualenvs/h/local/lib/python2.7/site-packages/pyramid/config/views.py", line 329, in attr_view
    return view(context, request)
  File "/home/seanh/.virtualenvs/h/local/lib/python2.7/site-packages/pyramid/config/views.py", line 305, in predicate_wrapper
    return view(context, request)
  File "/home/seanh/.virtualenvs/h/local/lib/python2.7/site-packages/pyramid/config/views.py", line 287, in _authdebug_view
    return view(context, request)
  File "/home/seanh/.virtualenvs/h/local/lib/python2.7/site-packages/pyramid/config/views.py", line 245, in _secured_view
    return view(context, request)
  File "/home/seanh/Projects/hypothesis/h/h/api/cors.py", line 20, in wrapper
    response = wrapped(context, request)
  File "/home/seanh/.virtualenvs/h/local/lib/python2.7/site-packages/pyramid/config/views.py", line 355, in rendered_view
    result = view(context, request)
  File "/home/seanh/.virtualenvs/h/local/lib/python2.7/site-packages/pyramid/config/views.py", line 501, in _requestonly_view
    response = view(request)
  File "/home/seanh/Projects/hypothesis/h/h/api/views.py", line 153, in create
    annotation = logic.create_annotation(fields=fields, user=user)
  File "/home/seanh/Projects/hypothesis/h/h/api/logic.py", line 33, in create_annotation
    annotation.save()
  File "/home/seanh/.virtualenvs/h/local/lib/python2.7/site-packages/annotator/annotation.py", line 53, in save
    d.save()
  File "/home/seanh/.virtualenvs/h/local/lib/python2.7/site-packages/annotator/document.py", line 148, in save
    self._remove_deficient_links()
  File "/home/seanh/.virtualenvs/h/local/lib/python2.7/site-packages/annotator/document.py", line 117, in _remove_deficient_links
    filtered_list = [l for l in links if 'href' in l]
TypeError: 'NoneType' object is not iterable
```

seanh:
Fixes #2519

### DIFF
--- a/h/api/test/views_test.py
+++ b/h/api/test/views_test.py
@@ -3,6 +3,7 @@
 
 import mock
 import pytest
+
 from pyramid import testing
 
 from h.api import views
@@ -129,7 +130,7 @@ def test_annotations_index_renders_results(search_lib):
 
 # The fixtures required to mock all of create()'s dependencies.
 create_fixtures = pytest.mark.usefixtures(
-    'logic', 'AnnotationEvent', 'search_lib')
+    'logic', 'AnnotationEvent', 'search_lib', 'validators')
 
 
 @create_fixtures
@@ -154,6 +155,30 @@ def test_create_calls_create_annotation(logic):
     logic.create_annotation.assert_called_once_with(
         request.json_body,
         userid=request.authenticated_userid)
+
+
+@create_fixtures
+def test_create_calls_validator(validators):
+    request = mock.Mock()
+
+    views.create(request)
+
+    validators.Annotation.return_value.validate.assert_called_once_with(
+        request.json_body)
+
+
+@create_fixtures
+def test_create_returns_api_error_for_validation_error(validators):
+    class Error(Exception):
+        pass
+    validators.Error = Error
+    validators.Annotation.return_value.validate.side_effect = (
+        validators.Error(mock.sentinel.reason))
+
+    response = views.create(mock.Mock())
+
+    assert response['status'] == 'failure'
+    assert response['reason'] == mock.sentinel.reason
 
 
 @create_fixtures
@@ -233,7 +258,8 @@ def test_read_does_not_crash_if_annotation_has_no_group():
 
 
 # The fixtures required to mock all of update()'s dependencies.
-update_fixtures = pytest.mark.usefixtures('logic', 'Annotation', 'search_lib')
+update_fixtures = pytest.mark.usefixtures(
+    'logic', 'Annotation', 'search_lib', 'validators')
 
 
 @update_fixtures
@@ -245,6 +271,30 @@ def test_update_returns_error_if_json_parsing_fails():
     error = views.update(mock.Mock(), request)
 
     assert error['status'] == 'failure'
+
+
+@update_fixtures
+def test_update_calls_validator(validators):
+    request = mock.Mock()
+
+    views.update(mock.Mock(), request)
+
+    validators.Annotation.return_value.validate.assert_called_once_with(
+        request.json_body)
+
+
+@update_fixtures
+def test_update_returns_api_error_for_validation_error(validators):
+    class Error(Exception):
+        pass
+    validators.Error = Error
+    validators.Annotation.return_value.validate.side_effect = (
+        validators.Error(mock.sentinel.reason))
+
+    response = views.update(mock.Mock(), mock.Mock())
+
+    assert response['status'] == 'failure'
+    assert response['reason'] == mock.sentinel.reason
 
 
 @update_fixtures
@@ -398,5 +448,12 @@ def _publish_annotation_event(request):
 @pytest.fixture
 def access_token(request):
     patcher = mock.patch('h.api.views.access_token', autospec=True)
+    request.addfinalizer(patcher.stop)
+    return patcher.start()
+
+
+@pytest.fixture
+def validators(request):
+    patcher = mock.patch('h.api.views.validators', autospec=True)
     request.addfinalizer(patcher.stop)
     return patcher.start()
--- a/h/api/test/validators_test.py
+++ b/h/api/test/validators_test.py
@@ -1,0 +1,17 @@
+# -*- coding: utf-8 -*-
+import pytest
+
+from h.api import validators
+
+
+def test_Annotation_returns_None_for_valid_data():
+    data = {"document": {"link": ["http://example.com/example"]}}
+
+    assert validators.Annotation().validate(data) is None
+
+
+def test_Annotation_raises_if_document_link_is_None():
+    data = {"document": {"link": None}}  # Invalid link.
+
+    with pytest.raises(validators.Error):
+        validators.Annotation().validate(data)
--- a/h/api/validators.py
+++ b/h/api/validators.py
@@ -1,0 +1,20 @@
+# -*- coding: utf-8 -*-
+"""Classes for validating data passed to the annotations API."""
+
+
+class Error(Exception):
+
+    """Base exception class for all exceptions raised by this module."""
+
+    pass
+
+
+class Annotation(object):
+
+    """A validator for annotations."""
+
+    def validate(self, data):
+        """Raise h.api.validation.Error if the data is invalid."""
+        if 'document' in data and 'link' in data['document']:
+            if not isinstance(data['document']['link'], list):
+                raise Error("document.link must be an array")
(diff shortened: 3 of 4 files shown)


seanh:
This raises the question of where we want to do validation in the API:

- In the view function? These are currently just wrappers for logic functions.
- In the logic function? That's where I've put it. But this logic layer only needs to exist so that the app can call it, and it's been suggested that this isn't necessary because views can just call other views. Not sure.
- In the search lib's `prepare()`? But this seems like some validation that we'd want to do before accepting the data at all (e.g. before saving it in postgres, when we have that), not just for search
- Do we need an additional layer of validation functions? Probably good to separate validation from logic

seanh:
Also weird for logic to be raising httpexceptions just for views to catch them and turn them into something else, if logic's gonna do this it should probably define its own exception classes

tilgovi:
> Also weird for logic to be raising httpexceptions just for views to catch them and turn them into something else, if logic's gonna do this it should probably define its own exception classes

:+1:

nickstenning:
As you have already noted yourself, a logic function raising an HTTP exception is more than odd -- it defeats the point of having a logic layer at all.

I'd suggest that if you want to do this kind of thing, then you add some validation functionality that accepts a dict and returns a validated dict but may throw a validation exception (which might contain a message). Shove it in `h.api.schemas` or similar. Then the view function can do something like:

```python
try:
     data = schemas.Annotation(request.json_body).validate()
except ValidationError as e:
    raise pyramid.httpexceptions.HTTPBadRequest(e.message)
```

Either way, we should pull all validation of input data into one place, and the thing that validates data shouldn't know anything about HTTP exceptions.

seanh:
Ok, I've added `h.api.validators` (I didn't want to call it schemas as there aren't any explicit schemas as such).

The logic function calls the validation but does not catch the validation error - this is caught by the view function which turns it into an API error. Which now that I say it makes me think maybe it should just be the view function that calls the validation.

I had the validation function simply raise an error or return `None`, it does not return the validated data at least for now. That isn't necessary yet, and also it might be good to separate validation from converting.

seanh:
Moved calling the validation from logic into view, which is simpler. I looked back at <https://github.com/hypothesis/h/pull/2442>, which created `logic.py`, the reasoning for it was that the view callables were getting too complex and were too difficult to test / would need too many tests each. By that reasoning I think it makes sense to have things like logic and validation in separate modules that can be tested separately and have them be pure not integration functions, then views is the one integration module that needs tests like "it calls validation correctly" and "it calls logic correctly".

tilgovi:
I would consider just using `colander` from the view functions before wading into custom validation.

tilgovi:
In other words, as @nickstenning said, put it in `h.api.schemas`, but I would go further and say "and use `colander`". That will let you define each validation function separately, test them as units, and then just assemble fairly declaratively and have a single integration point with the view.

tilgovi:
I also like that because I think the `logic` layer, if it exists at all, should be assuming more preconditions about its input since it's called by application code, not by the router. Validate everything before it hits the logic and the logic can simpler.

seanh:
To use colander here wouldn't I have to add a complete schema for the whole annotation object? I'm happy to go away and read a first version of colander validation for the API if you guys want, but I'm just trying to fix one little thing here (reject annotation's with a document with a link that is `None`) so maybe the minimal validation here can be accepted, and we can move it to colander later?

I can already define each validation function separately and test them as units with this code, and it already uses the view as the single integration point. Maybe `Annotation` isn't a good name for my validator class though, should be `AnnotationCreate`.

This pull request doesn't touch the logic code at all (anymore).

seanh:
Rebased with some renaming to clarify things

tilgovi:
To use colander you would have to specify a full schema for the `Document`, but that's not so hard and doesn't remove any flexibility from `Annotation`.

But really there's just a simple bug in annotator-store, it seems, that could be fixed here, rather than introducing a whole custom validation module?

tilgovi:
And you actually don't have to use the deserialization from `colander`, that would drop missing fields, you could just use it to check validation. So you could validate the annotation with only a single `document` schema node that has a validator that does the custom validation.

It's exactly what you have except it re-uses a library we're already depending on that provides a sensible framework for expanding validation.

```python
# in h.api.schemas
class AnnotationSchema(colander.MappingSchema):
    document = colander.SchemaNode(colander.Mapping(), validator=validate_document)

# in h.api.views
schema = AnnotationSchema().bind(request=request)
try:
    schema.deserialize(annotation)  # return value ignored
except colander.Invalid:
    raise httpexceptions.HTTPBadRequest
```

nickstenning:
Two comments:

1. I don't like `AnnotationCreate`. We're validating the object, not the action.
2. The whole point of making it a class was so that the constructor takes needed context objects (such as the request, or whatever), and `validate` takes the data to be validated. That way one validator can be used on multiple objects, which makes them easier to use and test.

As for whether you go all out and use colander for this single validation, I really don't care. But if this gets more complicated then absolutely we should switch to using colander.

seanh:
> I don't like AnnotationCreate. We're validating the object, not the action.

Well, we might want to use one validation function when creating a new annotation and another when updating an annotation, that's what the name was intended to clarify. It's not actually an annotation object that we pass to it, but the form fields that the user submitted, so arguably we are validating the action.

> The whole point of making it a class was so that the constructor takes needed context objects (such as the request, or whatever), and validate takes the data to be validated. That way one validator can be used on multiple objects, which makes them easier to use and test.

Done, although now it doesn't need an `__init__()` because it doesn't (yet) need any context objects, it could just be a function but we might as well leave it a class for when it does need some context

seanh:
@nickstenning This is ready again

nickstenning:
As I said before, this doesn't make any sense as `AnnotationCreate`. This validation rule surely needs to be applied both on create and update, otherwise I can simply create an annotation with `"link": []` and then update it with `"link": null`, no?

seanh:
@nickstenning Ok, I've pushed a commit that fixes it for updates as well. Needs tests, but let me know what you think first.

I do think we need separate validators for create and update. For example we don't allow changing an annotation's group when updating the annotation. That rule is currently implemented somewhere else. But in general I think we might want different validation rules for creates and updates.

This could also be done as one `Annotation` class with `validate_create()` and `validate_update()` methods..

seanh:
Or a common base class, since currently the shared `_validate()` helper function doesn't have `self`

seanh:
Ok, on second thought I've force-pushed this to just call it `Annotation` again, as it looks a little silly instead. If/when we do want to validate creates and updates differently, I'd suggest an `Annotation` base class with `AnnotationCreate` and `AnnotationUpdate` sub-classes

seanh:
This got all messed up by rebase, fixing..

seanh:
Ok, that's fixed. Don't rebase before coffee

h/api/test/views_test.py
```diff
@@ -257,7 +307,7 @@ def test_update_calls_update_annotation(logic):
     logic.update_annotation.assert_called_once_with(
         context.model,
         request.json_body,
-        request.authenticated_userid)
+        userid=request.authenticated_userid)
```
nickstenning: This is an unrelated change that you can remove.

nickstenning: This test failure was fixed in 269139e.

nickstenning:
Okay, merging. I will just note that if this gets any more complicated (as it undoubtedly will) we absolutely *should* be using colander here rather than reinventing the wheel.

tilgovi:
Still don't understand why preventing annotations from changing groups is
important.

Just remove that restriction and create and update become the same.

Issue #2391, opened 2015-07-22
Cannot Start h:  No module named h.server


soundobj:
Hi there, following the installation guide: https://h.readthedocs.org/en/latest/hacking/install.html

when trying to start the service via make dev I have been getting some errors:

(h) make dev
bash: gunicorn: command not found
make: *** [dev] Error 127

so  I installed gunicorn:

(h) pip install gunicorn
Collecting gunicorn
  Using cached gunicorn-19.3.0-py2.py3-none-any.whl
Installing collected packages: gunicorn
Successfully installed gunicorn-19.3.0
(h)dans-MacBook-Air:h dan$ make dev

Error: No module named paste.deploy

so I installed pasteDeploy:

pip install pasteDeploy
Collecting pasteDeploy
/Users/xxx/.virtualenvs/h/lib/python2.7/site-packages/pip/_vendor/requests/packages/urllib3/util/ssl_.py:90: InsecurePlatformWarning: A true SSLContext object is not available. This prevents urllib3 from configuring SSL appropriately and may cause certain SSL connections to fail. For more information, see https://urllib3.readthedocs.org/en/latest/security.html#insecureplatformwarning.
  InsecurePlatformWarning
  Using cached PasteDeploy-1.5.2-py2.py3-none-any.whl
Installing collected packages: pasteDeploy
Successfully installed pasteDeploy-1.5.2

then
(h) make dev

Error: class uri 'h.server.Worker' invalid or not found:  

[Traceback (most recent call last):
  File "/Users/xxx/.virtualenvs/h/lib/python2.7/site-packages/gunicorn/util.py", line 140, in load_class
    mod = import_module('.'.join(components))
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/importlib/__init__.py", line 37, in import_module
    __import__(name)
ImportError: No module named h.server
]

Any help would be greatly appreciated,
regards.

PS when trying to setup the python dependencies for testing I am getting this error

python setup.py test
/Users/xxx/.virtualenvs/h/lib/python2.7/site-packages/setuptools/dist.py:294: UserWarning: The version specified ('0.5.1-1-gbc636bf') is an invalid version, this may not work as expected with newer versions of setuptools, pip, and PyPI. Please see PEP 440 for more details.

.....

No local packages or download links found for annotator<0.15,>=0.14.2
error: Could not find suitable distribution for Requirement.parse('annotator<0.15,>=0.14.2')
[...]

PR #2523, opened 2015-09-22
Auto re-install Python, Node deps if out of date

robertknight:
Add a marker file in node_modules/ and .eggs/ to record
whether the installed dependencies are out of date
w.r.t. package.json or requirements.txt and auto-install them
if necessary when running 'make dev'.

The marker file is not created until node / Python deps
have been successfully installed, which avoids the problem
mentioned in install.rst

For the Python dependencies, what other files need to be considered besides 'requirements.txt'?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,9 +11,15 @@ export PIP_REQUIRE_VIRTUALENV
 
 default: deps
 
-deps:
+deps: .eggs/.uptodate node_modules/.uptodate
+
+.eggs/.uptodate: setup.py requirements.txt
 	pip install --use-wheel -e .[dev,testing,YAML]
+	touch $@
+
+node_modules/.uptodate: package.json
 	npm install
+	touch $@
 
 clean:
 	find . -type f -name "*.py[co]" -delete
@@ -26,8 +32,9 @@ clean:
 	rm -f h/static/scripts/hypothesis.*js
 	rm -f h/static/styles/*.css
 	rm -f .coverage
+	rm -f node_modules/.uptodate .eggs/.uptodate
 
-dev:
+dev: deps
 	@gunicorn --reload --paste conf/development.ini
 
 test: client-test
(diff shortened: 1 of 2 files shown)


nickstenning:
This is nice magic, when it works, but unfortunately it sometimes doesn't work, and results in [very confused users](https://github.com/hypothesis/h/issues/2391).

As a result, we've committed almost [exactly the opposite change](https://github.com/hypothesis/h/commit/3039032) less than 3 months ago so I'm reluctant to flip flop again!

robertknight:
Ah - this PR is different from what we had previously in that it only creates the markers to indicate successful installation once `npm install` / `pip install` have fully completed. So if `npm install` failed half-way through then 'make deps' again would repeat the pip/npm install.

The reason I added this was because I ran into some confusing errors a couple of times after pulling updates from master and then trying to run `make dev` without having re-run `make deps`. An alternative approach would be to have a target which unconditionally installs dependencies and include that in the setup docs and then one which is conditional that is a prerequisite of `make dev`.

nickstenning:
> An alternative approach would be to have a target which unconditionally installs dependencies and include that in the setup docs and then one which is conditional that is a prerequisite of make dev.

You mean like `make deps` ;)

nickstenning:
So, this isn't different from the previous behaviour except in one regard: the timestamp files you're using are solely timestamp files, and hopefully that means they can't get updated by other means. So, provided you update this pull request to:

1. Add `setup.py` to the list of deps for the python install.

2. Remove the `.uptodate` files in the `clean` target.

I'm happy to merge this.

robertknight:
> 1. Add setup.py to the list of deps for the python install.
> 2. Remove the .uptodate files in the clean target.

Done

nickstenning:
Thanks. We'll try this again and hope nobody gets confused by it this time.

PR #2526, opened 2015-09-22
Fix HTTP syntax highlighting in API docs

seanh:


nickstenning:
I applaud your attention to detail, but I'm not sure what I think about turning `{ ... }` into `{ }` -- yes it breaks syntax highlighting, but it's also not at all as clear. The ellipsis clearly indicates that there should be some data there, whereas without it the examples seem to imply that you can post this thing with an empty permission object, an empty targets array, etc, and all will be fine.

seanh:
But, the syntax highlighting! :) I don't think there's a great way to do this:

- `" ... "` would work but that looks like it suggests that a string is valid when it isn't.
- Put in full example permissions, document and target data.
- Accept that the syntax highlighting will be broken for some of the example requests and responses on that page
- Turn off syntax highlighting for all of them, at least it's consistent

seanh:
I think maybe turn it off for all of them?

nickstenning:
If you can do so, sure!

seanh:
Done

PR #2532, opened 2015-09-23
Consolidate .gitignore files

nickstenning:
Having these scattered about across the repository makes it hard to work
out where something is ignored, and most tools/editors that read
.gitignore files don't pay attention to those not in the repository
root.


robertknight:
LGTM. I propose in future that we try to have as much of the compiled/generated outputs in the `build/` directory which will enable simplifying `.gitignore`, `make clean` and other tools - eg. grepping the source tree but ignoring compiled outputs.

PR #2533, opened 2015-09-23
Fix document title and domain for local file PDF annotations

seanh:


seanh:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;9sqXWtNi&#x2F;98-bug-file-href-and-empty-for-domain-on-some-pdf-annotations"> Bug: file:&#x2F;&#x2F; href and empty () for domain on some PDF annotations</a>

seanh:
Ok, this moves the logic that creates the annotation card titles (`<user>` on `<document title>` (`<document domain>`) out of the template and into AngularJS filters, and it deals with the of the local PDF files that this bug is for.

There may be other cases where the titles are still displayed wrong, but at least now when we find them we can add a test for them and fix them and know we won't break it again in the future and we didn't break any of the other tested cases.

These filters are only being used in one place and the point of filters is to be reusable, but in this case the reason that I'm using filters is to move the code from template to js so that it becomes testable. I could just move the code into `AnnotationController` instead, but I don't want to add to `AnnotationController` because it's a mess.

By the way I think that what I'm doing here (concatenating user values into strings and then rendering them) is safe because [ng-bind-html](https://docs.angularjs.org/api/ng/directive/ngBindHtml) sanitizes the HTML by default if you've included ngSanitize, [which we have](https://github.com/hypothesis/h/blob/master/h/static/scripts/app.coffee#L121).

tilgovi:
Lovely.

h/static/scripts/filter/document-domain.js
```diff
@@ -0,0 +1,21 @@
+'use strict';
+
+module.exports = [function() {
```
nickstenning: There's no need to wrap this in an array.

h/static/scripts/filter/document-title.js
```diff
@@ -0,0 +1,30 @@
+'use strict';
+
+module.exports = [function() {
```
nickstenning: Ditto. The array injection whatnot a) only matters if you want to inject something, and b) isn't necessary anymore because you can just mark a function with `// @ngInject`.

h/static/scripts/filter/document-title.js
```diff
@@ -0,0 +1,30 @@
...
+  function documentTitleFilter(document) {
+    var title = document.title;
+    var uri = document.uri;
+    if (uri && !(uri.startsWith('http://') || uri.startsWith('https://'))) {
+      // We only link to http(s) URLs.
+      uri = null;
+    }
+
+    if (title && uri) {
+      return ('on &ldquo;<a target="_blank" href="' + uri + '">' + title +
```
nickstenning: This looks like an XSS vulnerability to me. What's stopping me submitting an annotation with `document.title = '</a><script>alert("Maliciousness!")</script>';`?

I'm guessing this probably needs to use `$interpolate` or similar to actually render these strings.

h/static/scripts/filter/test/document-domain-test.js
```diff
@@ -0,0 +1,44 @@
+'use strict';
+
+describe('documentDomain', function() {
+
+  before(function() {
+    angular.module('h', []).filter(
```
nickstenning: Given the simplicity of these modules, it might be better to test them without any reference to angular at all...

h/static/scripts/filter/document-title.js
tilgovi: `ngSanitize` will strip out the script.

nickstenning: Ah, I see. It won't strip out other HTML, though, so there's still unintended behaviour here. For example: `document.title = '</a><a href="http://buycialisnow.com">BUY CIALIS NOW</a>';` will still work.

seanh: Hmm, yeah, maybe it should get passed through something that just strips all HTML tags

tilgovi: Or if you could somehow know another way, in the template, whether the thing is linkable, you could put the `a` tag in the template.

seanh: Ok I've pushed a commit that fixes this (the tests will be broken, I'll fix them).

This does make me think two things though:

1. We should be escaping HTML in annotation fields such as title server-side
2. This filters approach smells - we shouldn't be creating HTML by concatenating strings. On the other hand, I wanted to make the logic testable. Maybe @tilgovi is right that `AnnotationController` should just add some helpful booleans like `isLinkable` to make the logic in the templates really simple to write.

seanh:
Ok, all the comments addressed, the one thing that's still on my mind about this is whether the the two filter should be combined into one annotationTitle filter that includes the username as well, as that might be more useful.

package.json
```diff
@@ -60,7 +60,8 @@
     "proxyquire": "^1.6.0",
     "proxyquire-universal": "^1.0.8",
     "proxyquireify": "^3.0.0",
-    "sinon": "1.16.1"
+    "sinon": "1.16.1",
+    "escape-html": "^1.0.3"
```
nickstenning: This is a production dep, not a dev dependency.

nickstenning: Oh, and when you move it, please respect the alphabetisation of the file!

nickstenning:
Happy to merge as-is if you put the dep in the right place and rename the function.

h/static/scripts/filter/document-domain.js
```diff
@@ -0,0 +1,26 @@
+'use strict';
+
+var escape = require('escape-html');
```
nickstenning: Should probably be called something else (`escapeHtml`?) as JS has a builtin called `escape` (which doesn't do what you want).

seanh:
Done

PR #2534, opened 2015-09-23
Use autoprefixer to add vendor prefixes to CSS

robertknight:
Use postcss and the autoprefixer plugin to post-process
CSS to add vendor prefixes and remove use of compass
mixins which did the same thing.

This moves compatibility concerns out of our styling
and down to a post-processor whose options can be adjusted
as necessary and moves us a step towards removing the
build dependency on compass.

Looking at the generated output, it fixes several issues
where we were inconsistent about which prefixes were used.

Pyramid webassets does have an autoprefixer filter but
it was written for an older version of autoprefixer which
was a standalone tool rather than a postcss plugin. In future
this also paves the way to replacing the cleancss asset filter
with an optional step in the postcss-filter.js script.

One thing I'm not sure about is where this filter script belongs in the source tree,
as there isn't a top level `scripts` or `utils` dir.

### DIFF
--- a/h/static/styles/base.scss
+++ b/h/static/styles/base.scss
@@ -1,4 +1,4 @@
-@import "compass/css3";
+@import 'compass/css3';
 @import 'compass/reset/utilities';
 
 @import "./variables";
@@ -6,15 +6,13 @@
 /* Style input placeholders */
 @mixin placeholder {
   &.placeholder { @content; }
-  &:-moz-placeholder { @content; }
-  &::-moz-placeholder { @content; }
-  &:-ms-input-placeholder { @content; }
-  &::-webkit-input-placeholder { @content; }
+  &:placeholder { @content; }
+  &::placeholder { @content; }
 }
 
 /* Shadow mixins */
 @mixin smallshadow($a: 0, $b: 1px, $c: .1) {
-  @include box-shadow($a $b 1px hsla(0, 0%, 0%, $c));
+  box-shadow: $a $b 1px hsla(0, 0%, 0%, $c);
 }
 
 @mixin quote {
--- a/scripts/postcss-filter.js
+++ b/scripts/postcss-filter.js
@@ -1,0 +1,27 @@
+#!/usr/bin/env node
+
+// post-process CSS using PostCSS
+// (https://github.com/postcss/postcss)
+//
+// This adds vendor prefixes using autoprefixer
+// https://github.com/postcss/autoprefixer
+
+require('es6-promise').polyfill();
+
+var autoprefixer = require('autoprefixer');
+var postcss = require('postcss');
+
+var inputCss = '';
+
+process.stdin.on('data', function (chunk) {
+  inputCss += chunk;
+});
+
+process.stdin.on('end', function () {
+  postcss([autoprefixer])
+    .process(inputCss)
+    .then(function (result) {
+      console.log(result.css);
+    });
+});
+
--- a/h/assets.py
+++ b/h/assets.py
@@ -73,6 +73,26 @@ class CleanCSS(ExternalTool):
 
 register_filter(CleanCSS)
 
+class PostCSS(ExternalTool):
+    """ Add vendor prefixes using postcss and autoprefixer.
+
+        webassets does ship with an 'autoprefixer' filter but
+        it does not support the current version of autoprefixer
+        which ships as a postcss plugin rather than standalone
+        tool.
+
+        Using postcss directly from JS will also enable transitioning
+        to a single JS-based tool for all CSS processing.
+    """
+
+    name = 'postcss'
+    max_debug_level = None
+
+    def output(self, _in, out, **kw):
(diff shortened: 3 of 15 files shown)


robertknight:
I set up an IE 10 VM and did some brief testing to verify that it looks as expected.

robertknight:
By "set up" I mean, "downloaded from dev.modern.ie" - Microsoft are making life a lot easier now.

tilgovi:
I would suggest the use of `postcss-cli` and drop the script. Check out how the `extra_args` flags work in conf/*.ini with `webassets` and use one of the built-in `webassets` filters as an example for that.

The only thing to watch for is that you may want to use `pyramid.config.aslist()` on the value of `extra_args`. Our `Browserify` filter in `assets.py` does its own version of this in its input function, where it checks the type of `extra_args`. That `extra_args` is such a pervasive parameter in `webassets` filters has already had me wonder more than once if I shouldn't just handle that type coercion in `pyramid_webassets`, which I help maintain.

At that point, this would be a great filter to submit upstream, too.

tilgovi:
Which we should also do with Browserify, I think.

tilgovi:
> Which we should also do with Browserify, I think.

Submit upstream, that is.

h/static/styles/base.scss
```diff
@@ -6,15 +6,13 @@
 /* Style input placeholders */
 @mixin placeholder {
   &.placeholder { @content; }
-  &:-moz-placeholder { @content; }
-  &::-moz-placeholder { @content; }
-  &:-ms-input-placeholder { @content; }
-  &::-webkit-input-placeholder { @content; }
+  &:placeholder { @content; }
```
nickstenning: `.placeholder`?

robertknight:
> Check out how the extra_args flags work in conf/*.ini with webassets and
> use one of the built-in webassets filters as an example for that.

Where I was intending to go with this is ultimately to have a single script which implements the full CSS processing pipeline with just a flag for dev/production. This can then easily be reused outside of the context of the main Python web app.

The other issue is that postcss-cli is not part of the main postcss project, so it doesn't implement command-line mappings for all options that might be useful (eg. does not currently support [source maps](https://github.com/code42day/postcss-cli/issues/3)). I could of course write a patch for that and push it upstream, but it doesn't save much over just using the API directly.

scripts/postcss-filter.js
```diff
@@ -0,0 +1,27 @@
...
+var postcss = require('postcss');
+
+var inputCss = '';
+
+process.stdin.on('data', function (chunk) {
+  inputCss += chunk;
+});
+
+process.stdin.on('end', function () {
+  postcss([autoprefixer])
```
nickstenning: Could we be explicit about what browsers we're supporting, perhaps with a `browserslist` config file?

scripts/postcss-filter.js
```diff
@@ -0,0 +1,27 @@
+#!/usr/bin/env node
+
+// post-process CSS using PostCSS
+// (https://github.com/postcss/postcss)
+//
+// This adds vendor prefixes using autoprefixer
+// https://github.com/postcss/autoprefixer
+
+require('es6-promise').polyfill();
```
nickstenning: Not a blocker for this PR, but we should probably just require Node v0.12 at this point.

nickstenning:
LGTM. I'll let you wonder about the two suggestions I've made asynchronously.

robertknight:
> Not a blocker for this PR, but we should probably just require Node v0.12 at this point.

I'd be happy to submit a PR for that. Will any configuration need changing besides package.json's engines field and possibly the Travis config that you know of?

> Could we be explicit about what browsers we're supporting, perhaps with a browserslist config file?

Makes sense. I'll look into it.

tilgovi:
:-( I feel like this adds more fragmentation to the pipeline configuration
and misses a nice opportunity to help webassets.

tilgovi:
There seems to be quite a bit of coordination between PostCSS maintainers and postcss-cli.

https://github.com/postcss/postcss/issues/154#issuecomment-78456147
https://github.com/code42day/postcss-cli/issues/3#issuecomment-94269493

If you want to get rid of webassets, that would be a fine direction to go. I would like to see the client package-able without Python at all. However, in that world I would expect to see a single task running in JavaScripts, something like a Gulpfile or similar. I'm not a big fan of a proliferation of little maintainer scripts. It's yet another thing a new contributor has to learn and scatters configuration across a bunch of files instead of a few.

Can I make any PR to ameliorate this?

robertknight:
> If you want to get rid of webassets, that would be a fine direction to go.
> I would like to see the client package-able without Python at all. 

I would like to see it be possible to build/run/edit the front-end without needing to run an instance of the Python server.

> Can I make any PR to ameliorate this?

If we can cut down the proliferation of tools to a single postcss-based pipeline I think that would be useful and a step in this direction. That could either mean chipping away at the compass dependency or perhaps replacing cleancss with a postcss plugin.

PR #2535, opened 2015-09-23
Translate annotation mapper

nickstenning:
I'm planning on doing a fair bit of moving things around in `annotationMapper`, and I'd rather that didn't mean adding more CoffeeScript to the codebase. Thus, in preparation, I'm translating it to JavaScript.

The only intended functional change is the removal of the `setupAnnotation` method which I can't find anything using.


robertknight:
Does this relate to a particular GitHub issue or card or is it purely refactoring?

nickstenning:
It's preparatory work for the "focus on a group" card.

h/static/scripts/annotation-mapper.js
```diff
@@ -0,0 +1,74 @@
+'use strict';
+
+
+// Fetch the container object for the passed annotation from the threading
```
robertknight: What is a 'container' for an annotation and what's the structure of annotations in the threading service (if the answer to this is already in the docs somewhere that I missed, feel free to just point me at it)?

nickstenning: It's not in the docs. The threading service is a wrapper around [this implementation](https://github.com/maxogden/conversationThreading-js) of the [venerable JWZ threading algorithm](https://www.jwz.org/doc/threading.html).

I will note, however, that I haven't changed any of that in this pull request. This is a straight translation of CoffeeScript to JavaScript, and this is a small helper function to make the code easier to read.

h/static/scripts/annotation-mapper.js
```diff
@@ -0,0 +1,74 @@
...
+// service, but only return it if it has an associated message.
+function getContainer(threading, annotation) {
+  var container = threading.idTable[annotation.id];
+  if (container === null || typeof container === 'undefined') {
```
robertknight: Ah, I see why you asked :) - I would recommend just using the falsey-ness of null/undefined etc. so that null/undefined are treated equivalently and also for consistency with logic that interfaces with third party code which may use either interchangeably.

Otherwise LGTM.

nickstenning: Right. But here I'm trying to preserve as near as possible the behaviour of the previous code. So I'd rather not change this in this case.

tilgovi:
Take a look at this, which I had in my stash from long, long ago:

https://gist.github.com/tilgovi/59e74d9c5dd0b3c4b8a7

Not every change is relevant, but you might consider whether that API for the mapper class is sensible. Just the other day I was thinking that you could make `mapper` a into a generic service by having it be a factory function for mapper classes, taking an `ngResource` as a parameter. The events could become more generic just by dropping the `annotation` prefixes. You would probably then still inject `annotationMapper` into the view controllers, but it could be resolved on view change as an injectable for the controller that uses the factory function. That would give each view a fresh mapper instance. The `identity_map` (as SQLAlchemy calls it) is one of the functions we're using `threading.idTable` now. If the mapper were encapsulated like this, and the threading driven off the mapper events, that would be a win for sanity.

Just some food for thought.

tilgovi:
But yes, merge this, I was just dropping what I had in my head for you to consider as you rewrite.

h/static/scripts/annotation-mapper.js
robertknight: Makes sense

PR #2536, opened 2015-09-23
Hide unanchored warnings when not sidebar

### LINKED ISSUE #2521: "We can't find the exact position of this annotation" shown on every annotation in the stream

seanh:
Don't show the "We can't find the exact position of this annotation"
warnings on unanchored annotations when on the stream or individual
annotation pages - only show them in the sidebar.

Fixes #2521


nickstenning:
Nice.

PR #2537, opened 2015-09-23
Remove remnants of annotator store auth

tilgovi:
The consumer field is removed entirely because with authorization
disabled in annotator-store the field serves no purpose at present.

Claims used by the Annotator Auth plugin on the front end are removed
from the JWT since this client has been deprecated for some time.

The User and Consumer model stubs from annotator-store are removed
and the authenticated userid from Pyramid is passed to the logic
functions of h.api now.


tilgovi:
- `h.streamer` builds its own auth filtered query, so does not need authorization enabled
- `h.api.search.core` was already overriding the default to disable authorization (thanks @seanh)

So, I am fairly certain this is all safe and good to go, but please, please double check my reasoning.

h/api/logic.py
```diff
@@ -66,13 +63,11 @@ def update_annotation(annotation, fields, has_admin_permission):
...
 
     # If the user is changing access permissions, check if it's allowed.
-    changing_permissions = (
-        'permissions' in fields and
-        fields['permissions'] != annotation.get('permissions', {})
-    )
-    if changing_permissions and not has_admin_permission:
-        raise RuntimeError(
-            _('Not authorized to change annotation permissions.'), 401)
+    permissions = annotation.get('permissions', {})
```
tilgovi: This should probably move to some validation logic.

h/api/logic.py
```diff
@@ -66,13 +63,11 @@ def update_annotation(annotation, fields, has_admin_permission):
...
+    if fields.get('permissions', {}) != permissions:
+        if user not in permissions.get('admin', []):
+            raise RuntimeError(
+                _('Not authorized to change annotation permissions.'), 401)
```
tilgovi: BTW, instead of this weird leakage of HTTP codes upward, these functions, if we really want to keep this logic layer, could throw custom exception classes that we could just catch as exception views and turn into HTTP exceptions.

nickstenning:
Merged as 11135fd (and 95bdc76).

PR #2539, opened 2015-09-23
Remove api logic

tilgovi:
WIP just to see what the view functions look like without `h.api.logic`.

I feel like whatever separation the logic layer provides is not really helpful right now and things seem kind of confused.

Things that are strange:

- `h.api.logic` functions raise exceptions with HTTP status codes, even though these are supposed to be "below" the HTTP layer
- The `prepare` function is called in logic, but `render` is called in the view, where it seems like those should be on one side of the same interface, perhaps
- We have to pass user information along that was readily available on the request
- Sanitization of input is happening in a logic function where preconditions could instead be assumed.

There are individual parts of this that it would be nice to simplify. I think the instinct to pull things out into a logic layer was a nice one, but the problem wasn't that there was view code and logic code together. I think it's just that the logic itself is too complicated.

Things could get much simpler if:

- We remove the curious restriction that the `group` field can't be changed, which seems unmotivated.
- We remove the anonymization / retraction support. It's of questionable utility and relies on the client to know more than it can/should (such as whether there are replies, especially in a federated world).
- We could remove the ACLs altogether, really easily, just by saying that only the owner and admins can do anything. The client would then just decide "can I write/update/delete" by checking "am I the author".

At that point, none of these view functions would be more than about eight lines.


tilgovi:
This is on top of #2537, so check the last commit if you want the cleaner diff.

tilgovi:
Tests not changed yet. I open this mainly for discussion, since talk about the logic module seems to be getting scattered across a few issues right now.

nickstenning:
I'm not averse to this change, but for what it's worth, I don't think the presence of a logic module is (in and of itself) necessarily the problem. The problem is as you rightly identify that we don't cleanly distinguish between sanitizing input and updating model data.

If you want to fix the tests I'm happy to merge something that looks like this.

nickstenning:
Closing. Feel free to reopen if you update it.

PR #2543, opened 2015-09-24
Limit group names to 25 characters

seanh:


seanh:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;2NiiPelT&#x2F;94-limit-length-of-group-names-to-25-chars"> Limit length of group names to 25 chars</a>

tilgovi:
25 seems pretty small. What's the reasoning?

dwhly:
Space in the UI.  Start small-ish, expand if we need to.

tilgovi:
Where is there a space problem in the UI? Should we consider whether there is a better way to deal with that?

25 feels really short to me.

    $ echo "Mrs. Blake's 9th Period English" | wc -c
    32

tilgovi:
And of course, non-English group names, too.

Would it not be better to re-evaluate the UI where this is challenging and either make it accommodating or truncate the display intelligently?

nickstenning:
Ignoring what the right magic number is for the moment, I think we should take this opportunity, @seanh, to fix this properly and replace *all* uses of `Unicode(n)/String(n)/VARCHAR(n)` with `UnicodeText()/Text()`.

In both Postgres and SQLite the length-limited and non-length-limited versions use identical storage types ([TEXT in SQLite](https://www.sqlite.org/datatype3.html#affinity), and [varlena in Postgres](http://www.postgresql.org/docs/current/static/datatype-character.html)), and the only difference is a slight performance hit from length checking.

Usually I'm in favour of pushing validation and schema enforcement as far back in the stack as possible, but this is one exception. The amount of pain that's caused by having to do data migrations for these kind of magic numbers is ridiculous.

So:

- [ ] ensure all length-limited strings have validation enforced elsewhere, either with colander or ideally with [`sqlalchemy.orm.validates`](http://docs.sqlalchemy.org/en/rel_0_9/orm/mapped_attributes.html) methods on the model objects
- [ ] translate all length-limited string fields to text-type fields, with appropriate data migration

seanh:
@nickstenning OK, that all sounds like a good idea and I'll get on to working on it in a separate pull request.

In the interest of not blocking groups though, I've force-pushed a new version of this pull request that just changes 100 to 25 in colander only not in the db. I think we can merge this so that it's not blocking the groups feature, and I can work on changing how we validate string lengths globally in an unblocking thread.

nickstenning:
:+1:

PR #2544, opened 2015-09-24
Add bundle dependency on annotation-mapper.js

tilgovi:
Until the build outputs are in a separate directory from the source
modules we need to maintain the .js files here without globbing or
webassets won't re-build the bundle when annotation-mapper.js changes
during development.


robertknight:
> Until the build outputs are in a separate directory
> from the source modules

Is there anything in particular standing in the way of that other than changing the output paths for the various assets?

tilgovi:
Nope!

PR #2545, opened 2015-09-25
Don't crash if no parent is found

nickstenning:
We can't always guarantee the parent's existence (its owner might delete it) so we must handle this case.


tilgovi:
TBH I'm still not even sure why the server does this.

The client can easily implement this behavior and the server, IMO, has no reason to assume it isn't the client's intention to reply elsewhere, especially given that replying is just annotating.

PR #2546, opened 2015-09-25
Allow missing permissions

nickstenning:
We can't always guarantee the permissions field is present, as our API does not currently enforce this. Don't crash in this case.

This PR also fixes another test that was passing due to a couple of bugs in the test, rather than because the code was correct.

### DIFF
--- a/h/api/groups/auth.py
+++ b/h/api/groups/auth.py
@@ -13,7 +13,7 @@ def set_permissions(annotation):
         return
 
     group = annotation.get('group')
-    if group is None or group == '__world__':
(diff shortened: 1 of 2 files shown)


h/api/groups/auth.py
```diff
@@ -13,7 +19,7 @@ def set_permissions(annotation):
         return
 
     group = annotation.get('group')
-    if group is None or group == '__world__':
+    if group is None or group == '' or group == '__world__':
```
seanh: `if group in (None, '', '__world__')`

PR #2548, opened 2015-09-25
Remove character limits from database columns

seanh:


seanh:
In this case I opted to simply remove the limit and not add any validation of it in Python. I'm not sure why a 256-character limit was needed here, and also there didn't appear to be any handling of it in the code, if sqlalchemy raised an exception.

seanh:
The general approach I'm taking here for a string that _does_ actually need to have a max length validated:

- Remove max length from database column
- Add the max length as a constant in the relevant `models.py`
- Add a sqlalchemy `@validates` for it
- Also validate it in the relevant Colander schema

A problem with this is if the Colander schema doesn't validate it, then when you try to create the model object the `@validates` will raise an error that isn't caught, server will crash. Do I want to add error handling in the views twice? Once for Colander errors then once for exceptions raised by the model. But since the Colander schema _does_ validate it, this should never happen, so I'm going to go for not handling the error.

seanh:
Notes:

* This requires the database to be migrated and the new code deployed at the same time. The new code will not work against an old version of the db, nor will the old code work against a new version of the db. This is because https://github.com/hypothesis/h/commit/22f6b65c051d72e71cbd130fad60dc038b72035f removes the `created_by` and `valid_until` columns, and these columns had `NOT NULL` constraints on them.

* I've done this as lots of separate migration scripts, I think this is better if one migration script fails 'cos it'll just re-run that one (and ones after it that were not run) whereas if it was one big migration it'd re-run the whole thing but some of the changes would already be made so it would crash. (We should probably always use transactions in our migrations, if that's possible.)

* For `uid` I removed the max length from the db table but did not replace it with Python validation. It's only ever set as a side-effect of setting `username` and is generated from `username`. User code _can_ create a `User` with a long uid though, but it seems like the right fix here is to prevent user code from setting uid at all or to not even save uid, just generate it from username when requested.

* `salt` is a similar story, it only ever seems to be generated by the `_generate_random_string()` function so it doesn't seem to be necessary to validate its length although technically user code _could_ create a `User` with long salt.

* I didn't bother to change the `VARCHAR`s in the notification package since I think that code needs to go anyway. Otherwise, this removes all string length limits from database columns.

seanh:
@nickstenning This can be reviewed now

nickstenning:
> Do I want to add error handling in the views twice? Once for Colander errors then once for exceptions raised by the model. But since the Colander schema does validate it, this should never happen, so I'm going to go for not handling the error.

That sounds like the right approach to me. We should never hit model validation errors in views. If we do, it's our error, not bad input, and should propagate so it ends up in Sentry.

nickstenning:
> to not even save uid, just generate it from username when requested.

FWIW that defeats the entire point of that field, which is to enforce uniqueness of the dot-normalized username at the database layer.

nickstenning:
> We should probably always use transactions in our migrations, if that's possible.

All alembic migrations are transactional.

h/accounts/models.py
```diff
@@ -42,16 +45,16 @@ class Activation(Base):
...
                      nullable=False,
                      unique=True,
                      default=_generate_random_string)
 
-    # FIXME: remove these unused columns
-    created_by = sa.Column(sa.Unicode(30), nullable=False, default=u'web')
-    valid_until = sa.Column(sa.DateTime,
-                            nullable=False,
-                            default=datetime.utcnow() + timedelta(days=3))
+    @sa.orm.validates('code')
```
nickstenning: This isn't necessary. We never set the code manually.

h/accounts/models.py
```diff
@@ -87,6 +90,8 @@ def _get_username(self):
         return self._username
 
     def _set_username(self, value):
+        if not USERNAME_MIN_LENGTH <= len(value) <= USERNAME_MAX_LENGTH:
+            raise ValueError
```
nickstenning: Raise an instance, not the type, and give it a meaningful message:

```python
raise ValueError('username must be between {min} and {max} characters in length'.format(min=USERNAME_MIN_LENGTH, max=USERNAME_MAX_LENGTH))
```

h/accounts/models.py
```diff
@@ -112,6 +117,12 @@ def username(self):
     activation_id = sa.Column(sa.Integer, sa.ForeignKey(Activation.id))
     activation = sa.orm.relationship('Activation', backref='user')
 
+    @sa.orm.validates('email')
+    def validate_email(self, key, email):
+        if len(email) > EMAIL_MAX_LENGTH:
+            raise ValueError
```
nickstenning: Again, instance not type, and a meaningful message.

h/accounts/models.py
```diff
@@ -136,6 +147,8 @@ def _get_password(self):
         return self._password
 
     def _set_password(self, raw_password):
+        if not PASSWORD_MIN_LENGTH <= len(raw_password) <= PASSWORD_MAX_LENGTH:
+            raise ValueError
```
nickstenning: Same again: instance not type, meaningful message.

h/accounts/models.py
```diff
@@ -136,6 +147,8 @@ def _get_password(self):
         return self._password
 
     def _set_password(self, raw_password):
+        if not PASSWORD_MIN_LENGTH <= len(raw_password) <= PASSWORD_MAX_LENGTH:
```
nickstenning: We didn't enforce a maximum password length before, so let's not introduce that here.

h/accounts/views.py
```diff
@@ -124,7 +124,7 @@ def login(self):
 
         self.request.registry.notify(LoginEvent(self.request, user))
 
-        userid = 'acct:{}@{}'.format(user.username, self.request.domain)
+        userid = u'acct:{}@{}'.format(user.username, self.request.domain)
```
nickstenning: This seems like an unrelated change. No?

h/groups/models.py
```diff
@@ -37,6 +41,12 @@ def __init__(self, name, creator):
         self.creator = creator
         self.members.append(creator)
 
+    @sa.orm.validates('name')
+    def validate_name(self, key, name):
+        if not GROUP_NAME_MIN_LENGTH <= len(name) <= GROUP_NAME_MAX_LENGTH:
+            raise ValueError
```
nickstenning: Another one where we should have a proper message.

h/accounts/views.py
seanh: Yeah I think it is. I noticed a "non-unicode bind param for unicode column" warning from sqlalchemy, and I had changed the type of the username column, so thought it was my fault and fixed it. Looking at the diff though the column was already unicode so the warning should already have been there. We may as well fix it though! It's in its own commit.

seanh:
> > We should probably always use transactions in our migrations, if that's possible.
> 
> All alembic migrations are transactional. (On decent databases...)

Ah, so not on sqlite then! We could probably merge all those revisions into one then

h/accounts/models.py
seanh: Ok. there was a 256-character limit on the _hashed_ password (for some reason), but I guess these hashes don't generally change in length as the password does. I'll just remove this

h/accounts/models.py
seanh: I need to get one of my linters to complain about this, none of them do

seanh:
@nickstenning Rebased with requested fixes, I'll look at https://github.com/hypothesis/h/pull/2565 now

seanh:
Ok, I've merged https://github.com/hypothesis/h/pull/2565 and rebased this on it. I changed <https://github.com/hypothesis/h/blob/remove-256-character-limit-from-uri-column/h/migrations/versions/32efc6033dd1_remove_created_by_and_valid_until_.py>'s `downgrade()` to have `nullable=True` and no default or server_default, as it should now. I also changed the `down_revision` in <https://github.com/hypothesis/h/blob/remove-256-character-limit-from-uri-column/h/migrations/versions/426c77c1b30c_change_subscriptions_uri_to_unicodetext.py> (the first migration in this pr) to be the revision of the migration you added in <https://github.com/hypothesis/h/pull/2565>.

One thing I've noticed though, if there are rows in the activation table then the downgrade in <https://github.com/hypothesis/h/pull/2565> will fail, it needs a server_default, I'll fix that...

seanh:
Ok gonna test upgrading and downgrading..

seanh:
Ok, that was a hassle. I've tested this by:

1. Checking out `ff10dc686d04a41531d59c7a6b29f902c76410a4` (from before#2565)
2. `rm .h.db`, `make dev`, register a user but don't activate them, register another user, activate them, then do a password reset but don't complete it. Got a couple of rows in the activation table now, including created_on and valid_until values.
3. Checked out this branch
4. stamp ef3059e0396, upgrade head, downgrade -12, upgrade head
5. Run h, activate the account, login to it, complete the password reset, login to that account, register, activate and log into another account.

PR #2549, opened 2015-09-25
Fix comments (AKA "page notes") with search_normalized

### LINKED ISSUE #2484: Store page annotations with a selector-less target
Right now page annotations have no targets. Instead, they should have a single target with the page source and no selectors.

- [ ] Update the client to create this target properly
- [ ] Update the client to consider such annotations as anchored. Current logic would consider them orphaned.
- [ ] Add logic to `prepare` on the server to migrate annotations coming from older clients.

Nothing else is needed to take incoming annotations in the new format and make them work for older clients since the ES mapping for `target.source` already copies to `uri`. We can leave that for the moment and maybe add a comment about cleaning it up.

nickstenning:
The last thing blocking full rollout of the new normalized URI searches is #2484. This pull request address and closes that issue. From the commit messages:

---

In order for comments (AKA "page notes") to be discoverable with the search logic implemented in #2413 (cc4ca35), they need to have at least one target with a source field.

Previously a comment was defined as an annotation

- with a `uri`
- without any targets
- without any references (no targets but with references == a reply)

Now comments have a single target with a single field, `source`.

---

When receiving data from older clients, this commit will ensure that we rewrite old-style comments to new ones with a target with a source field.

This will also serve as a data migration to be run with `hypothesis annotool`.

---

Closes #2484.

### DIFF
--- a/h/api/search/test/transform_test.py
+++ b/h/api/search/test/transform_test.py
@@ -56,6 +56,40 @@ def test_prepare_adds_scope_field(_, ann_in, ann_out, uri_normalize):
 
 @mock.patch('h.api.search.transform.models')
 @mock.patch('h.api.search.transform.groups')
+@pytest.mark.usefixtures("uri_normalize")
+@pytest.mark.parametrize("ann_in,ann_out", [
+    ({"uri": "giraffe"},
+     {"uri": "giraffe",
+      "target": [{"source": "giraffe", "scope": ["*giraffe*"]}]}),
+    ({"uri": "giraffe", "target": []},
+     {"uri": "giraffe",
+      "target": [{"source": "giraffe", "scope": ["*giraffe*"]}]}),
+    ({"uri": "giraffe", "references": []},
+     {"uri": "giraffe",
+      "references": [],
+      "target": [{"source": "giraffe", "scope": ["*giraffe*"]}]}),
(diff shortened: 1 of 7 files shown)


nickstenning:
Suggested manual testing strategy if you want to verify locally that this works:

- on master, create a comment ("page note")
- turn on `search_normalized`, and observe that the comment doesn't appear when you reload the page (it's not returned in search)
- check out this branch
- run the migration of past data: `hypothesis annotool conf/development.ini prepare`
- refresh and observe the page note has returned, and isn't marked as an orphan either in the sidebar or anywhere else.

seanh:
Looks good to me, just need annotool to be fixed so I can test

nickstenning:
> just need annotool to be fixed

Done!

judell:
Thank you. Shall I plan to verify this when it hits stage?

nickstenning:
:+1:

tilgovi:
I don't think you can remove the `done`calls in these tests. Now they are synchronous and they might pass before the promise resolves, no? Then, the promise resolves and maybe fails an assertion and it's swallowed silently.

nickstenning:
You can return a promise from a test.

tilgovi:
Great. That's news to me. I like that. Thanks.

tilgovi:
You may want to make the return explicit, though.

PR #2550, opened 2015-09-26
T89 combined scope save btn

robertknight:
This implements the combined privacy scope selector and post
button from the [mocks](https://trello.com/c/hkmaJW4t/89-combined-scope-selection-save-button).

- [X] Investigate the case of the missing group icons (`h-icon-group`, `h-icon-group-add` `h-icon-expand-less`, `h-icon-expand-more`) which were added in 873f1930653605939217d6946bbec93935ab662a . _Update: Of these icons only `h-icon-group` was actually used and that has been replaced with an updated design from Conor_
- [x] Fix spacing between Post and Cancel buttons
- [x] Fix bug where clicking visibility drop-down arrow in Chrome saves annotation changes instead of showing dropdown menu

**Initial disabled state**
t89-disabled-state

**Enabled state**
t89-enabled-state

**Dropdown state**
t89-dropdown-state

The high level change is that the `<privacy>` directive and 'Save' buttons have been removed and replaced with a combined post/scope button implemented by `<publish-annotation-btn>`. The dropdown button itself is implemented in `<dropdown-menu-btn>`.

The dropdown menu is implemented using a newer version of the Angular UI Bootstrap dropdown component that we were previously using. The newer version adds keyboard accessibility and the ability to control the dropdown state via bindings. The ui.bootstrap module is a custom build from the Angular UI Bootstrap website with a couple of minor changes for Angular 1.2 compatibility.

Couple of other things to note:

 * Rather than passing the entire annotation to the `<publish-annotation-btn>` directive, I pass
   in just the state needed to render it and callbacks to save the annotation and set the privacy level.
   This is the 'props/attributes down, events up' pattern.

 * Each of the new directives has its own CSS file for styling - this will make it easier
   to group all the files that make up a component together in future as part of
   the larger effort to separate out the different frontends.

 * I added some helpers to reduce the boilerplate when testing directives.


nickstenning:
If it's ok with you I'm going to note things here as I find them.

nickstenning:
I think this is a bug:

1. Make a highlight (i.e. with the highlight tool)
2. Edit the highlight
3. Click on the dropdown's arrow to change its visibility.

Expected: I'm allowed to change the visibility or given an explanation as to why I can't.

Actual: The annotation appears to save.

nickstenning:
I'm not sure about "Post to Only Me" as copy. Is that approved by the @conordelahunty?

nickstenning:
Another bug, I think:

1. Create an annotation, with some text.
2. Edit the annotation, and change the text
3. Click on the dropdown arrow and change the visibility of the annotation

Expected: my edits to the text are unchanged.

Actual: my edits are discarded.

nickstenning:
I reckon the cancel button is too close to the dropdown button at the moment. Again that is of course @conordelahunty's call, so paging him here.

screen shot 2015-09-29 at 19 30 55

JakeHartnell:
> I reckon the cancel button is too close to the dropdown button at the moment. Again that is of course @conordelahunty's call, so paging him here.

Looks a little close to me too.

robertknight:
> Expected: I'm allowed to change the visibility or given an explanation as to why I can't.
> Actual: The annotation appears to save.

I can reproduce this in Chrome but not Firefox.

I also found a pre-existing bug where editing the blank text for a highlight and then clicking _Cancel_ does not reset the text.

> I reckon the cancel button is too close to the dropdown button at the moment. 

Indeed - that is a bug, its a regression introduced by the last refactoring I did.

robertknight:
@nickstenning so both the bugs you reported have the same cause.

Given the following simplified version of the DOM in the button:

```html
<div onclick="console.log('click event propagated')">
	<button disabled><div>Click me</div></button>
</div>
```

Clicks on the inner `<div>` propagate _through_ the disabled button in Chrome but not Firefox. The fix was to only install a `click` event handler on the appropriate `<button>` element rather than the parent.

nickstenning:
Okay, this is looking pretty good and I'm happy to merge, but I'm still a bit nervous about: a) modifying vendor libs and b) our seemingly ever-increasing page weight. This PR adds 10K to compressed `app.min.js` before we even count CSS. If there's a way we can consolidate our use of `angular-bootstrap` so we don't have two distinct versions of it that would make me very happy, even if it comes in a later commit.

nickstenning:
So -- let me know what you think to the above, @robertknight, and then we'll see whether we want to merge this now or later.

robertknight:
I'll look into consolidating angular-bootstrap now - that might not be too difficult. I'll have a think about page weight and look through our bundle using [disc](https://www.npmjs.com/package/disc) or a similar tool.

robertknight:
To recap the discussion on Slack about `angular-bootstrap`, there is a newer `tabs` directive but the API changed substantially so replacing it is best left to a separate PR I think.

What I have done for now is to avoid editing the custom ui-bootstrap build from upstream and instead created a wrapper which shims the necessary APIs from Angular 1.3 only within the `ui.bootstrap` module. When/if we upgrade to a newer version, the shim can just be removed.

nickstenning:
LGTM.

tilgovi:
Nice, @robertknight

PR #2554, opened 2015-09-28
Implement focusing on a group

seanh:
- [ ] Don't break the stream or individual annotation pages
- [x] Change the shown annotations when the selected group changes
- [ ] Also filter annotations from other groups out from live updates
- [ ] Focus the public group on sign out

### DIFF
--- a/h/static/scripts/annotation-mapper.js
+++ b/h/static/scripts/annotation-mapper.js
@@ -18,11 +18,15 @@ function getContainer(threading, annotation) {
 
 // Wraps the annotation store to trigger events for the CRUD actions
 // @ngInject
-function annotationMapper($rootScope, threading, store) {
+function annotationMapper($rootScope, threading, store, groups) {
   function loadAnnotations(annotations) {
     var loaded = [];
 
     annotations.forEach(function (annotation) {
+      if (annotation.group !== groups.focused().id) {
+        return;
+      }
+
       var container = getContainer(threading, annotation);
       if (container !== null) {
         angular.copy(annotation, container.message);
--- a/h/static/scripts/annotation-mapper.js
+++ b/h/static/scripts/annotation-mapper.js
@@ -66,7 +66,13 @@ function annotationMapper($rootScope, threading, store) {
     loadAnnotations: loadAnnotations,
     unloadAnnotations: unloadAnnotations,
     createAnnotation: createAnnotation,
-    deleteAnnotation: deleteAnnotation
+    deleteAnnotation: deleteAnnotation,
+    threads: function() {
+      return threading.root.children;
+    },
+    thread: function(id) {
+      return threading.idTable[id];
+    }
   };
 }
 
--- a/h/static/scripts/annotation-mapper.js
+++ b/h/static/scripts/annotation-mapper.js
@@ -19,11 +19,25 @@ function getContainer(threading, annotation) {
 // Wraps the annotation store to trigger events for the CRUD actions
 // @ngInject
 function annotationMapper($rootScope, threading, store, groups) {
-  function loadAnnotations(annotations) {
-    var loaded = [];
 
-    annotations.forEach(function (annotation) {
-      if (annotation.group !== groups.focused().id) {
+  // All of the annotations that annotationMapper has received from calls to
+  // its loadAnnotations() method.
+  var received = [];
+
+  // The annotations that are currently loaded into the page context.
+  var loaded = [];
+
+  function loadAnnotations(annotations) {
+    received = received.concat(annotations);
+    loadAnnotationsFromGroup(annotations, groups.focused().id);
+  }
+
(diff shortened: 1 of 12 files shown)


h/static/scripts/annotation-mapper.js
```diff
@@ -18,11 +18,29 @@ function getContainer(threading, annotation) {
...
+  // All of the annotations that annotationMapper has received from calls to
+  // its loadAnnotations() method.
+  var received = [];
+
+  // The annotations that are currently loaded into the page context.
+  var loaded = [];
+
   function loadAnnotations(annotations) {
-    var loaded = [];
+    received = received.concat(annotations);
```
seanh: This assumes that the same annotation never gets passed to `loadAnnotations()` twice, which may not be true. `received` may need to become a set, or a dict keyed by annotation id, or something,

h/static/scripts/annotation-mapper.js
```diff
@@ -18,11 +18,29 @@ function getContainer(threading, annotation) {
...
 // @ngInject
-function annotationMapper($rootScope, threading, store) {
+function annotationMapper($rootScope, threading, store, groups) {
+
+  // All of the annotations that annotationMapper has received from calls to
+  // its loadAnnotations() method.
+  var received = [];
+
+  // The annotations that are currently loaded into the page context.
+  var loaded = [];
```
seanh: We could maybe avoid having to keep track of this list here, by just getting it from `threading`. I sort of feel like it's `annotationMapper`'s job to keep track of this though, and `threading` is just a dump.

h/static/scripts/annotation-mapper.js
```diff
@@ -18,11 +18,29 @@ function getContainer(threading, annotation) {
...
+
   function loadAnnotations(annotations) {
-    var loaded = [];
+    received = received.concat(annotations);
+    loadAnnotationsFromGroup(annotations, groups.focused().id);
+  }
+
+  function loadAnnotationsFromGroup(annotations, groupId) {
```
seanh: Loads the given annotations into the page context, but only the ones from the given group. Docstring needed.

h/static/scripts/annotation-mapper.js
```diff
@@ -45,6 +70,10 @@ function annotationMapper($rootScope, threading, store) {
 
       $rootScope.$emit('annotationDeleted', annotation);
     });
+
+    annotations.slice().forEach(function(annotation) {
+      loaded.splice(loaded.indexOf(annotation, 1));
+    });
```
seanh: This crazy dance is because when our `$rootScope.$on('groupFocused', ...` above calls `unloadAnnotations()`, the passed `annotations` array _is_ the `loaded` array, and you can't modify an array while you iterate over it in JavaScript, so we need to iterate over a copy of it.

h/static/scripts/annotation-mapper.js
```diff
@@ -66,7 +95,13 @@ function annotationMapper($rootScope, threading, store) {
...
     unloadAnnotations: unloadAnnotations,
     createAnnotation: createAnnotation,
-    deleteAnnotation: deleteAnnotation
+    deleteAnnotation: deleteAnnotation,
+    threads: function() {
+      return threading.root.children;
+    },
+    thread: function(id) {
+      return threading.idTable[id];
+    }
```
seanh: These need docstrings

seanh:
Getting a `TypeError: Cannot read property 'children' of undefined` in the console when switching back and forth between groups, from `threading.annotationDeleted()`

seanh:
> Getting a TypeError: Cannot read property 'children' of undefined in the console when switching back and forth between groups, from threading.annotationDeleted()

You have to sign out the sign in then switch back and forth between groups, to get this

seanh:
> > Getting a TypeError: Cannot read property 'children' of undefined in the console when switching back and forth between groups, from threading.annotationDeleted()
> 
> You have to sign out the sign in then switch back and forth between groups, to get this

Fixed this, but there's still a problem that the numbers in the bucket bar get all wrong, if you sign out then sign in again then switch between groups multiple times

nickstenning:
So I've had a quick look at this and I think we're going to get ourselves into all sorts of confusion if we try and maintain all the state in the client.

For now, there's a much simpler (if marginally less efficient) way to do this. Have a look at [this branch](https://github.com/hypothesis/h/compare/focus-on-a-group). It is by no means perfect, and has some of the same issues as this one (we need to forcibly switch to the public group on logout, for example), but requires substantially fewer changes to the code. The only downside is that we're reloading annotations from the server every time we change focus, but I think we can optimize that later.

nickstenning:
I'd be happy to pair on this with you at some point tomorrow, @seanh?

seanh:
@nickstenning I'm free now, but can also pair tomorrow if you want

seanh:
Closing in favour of https://github.com/hypothesis/h/pull/2566

PR #2556, opened 2015-09-29
Make auth.effective_principals do what it says

### LINKED ISSUE #2555: Reply notification emails not going out
Mandrill logs show no reply email notifications are going out at the moment.
The last outgoing reply email in the system is Sept 15. 
Reported by user.

nickstenning:
Previously `h.auth.effective_principals` was a "groupfinder", returning the list of groups of which a given userid is a member.

This commit renames that function to `groupfinder`, and updates the two configuration files to use it in that capacity. It also adds a new helper function to `h.auth`, which actually does return the list of effective principals for a given userid.

This is extracted from a work-in-progress branch to fix #2555 in a way that doesn't involve reimplementing the pyramid auth system in multiple places in our code.

### DIFF
--- a/conf/development.ini
+++ b/conf/development.ini
@@ -11,7 +11,7 @@ es.host: http://localhost:9200
 
 mail.default_sender: "Annotation Daemon" <no-reply@localhost>
 
-multiauth.groupfinder: h.auth.effective_principals
+multiauth.groupfinder: h.auth.groupfinder
 multiauth.policies: remote session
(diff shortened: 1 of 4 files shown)


nickstenning:
Please do check this out and actually verify that the permissions system works as expected. Making changes to auth code is always risky, so it's worth being extra careful.

seanh:
LGTM, I'll check it out and test reading and writing public/private/group annotations..

PR #2558, opened 2015-09-29
Fix email notifications

### LINKED ISSUE #2555: Reply notification emails not going out

nickstenning:
When we receive an annotation in the notification mailer, we need to perform a set of checks:

- to determine if the annotation is a reply
- to determine who the author of the thread parent is
- to determine if we should send a reply notification to that author

In particular, we only want to send a reply notification email if the author is going to be able to read the reply. Previously, this was done by checking if the principals allowed by the 'read' permission included the system principal "system.Everyone".

Unfortunately, 998347a broke this by removing normal Annotation instances' status as a pyramid "context object", meaning that no annotations were passing the test and no reply notification emails were being sent.

This commit updates that permissions check. Now we:

1. Compute the set of principals of the parent annotation's author, using `h.auth.effective_principals`.
2. Compute the set of principals that are allowed to 'read' the reply annotation.
3. Abort unless these two sets have a non-null intersection.

Fixes #2555.

----

I apologise for the size of the diff, but it was necessary to shuffle some code around in order to be able to effectively test that this change correctly solves the problem.

### DIFF
--- a/h/api/auth.py
+++ b/h/api/auth.py
@@ -1,5 +1,7 @@
 # -*- coding: utf-8 -*-
 
+from pyramid import security
+
 from annotator.auth import Consumer, User
 
 
@@ -13,3 +15,21 @@ def get_user(request):
                 consumer = Consumer(key)
                 return User(userid, consumer, False)
     return None
+
+
+def translate_annotation_principals(principals):
+    """
+    Translate a list of annotation principals to a list of pyramid principals.
+    """
+    result = set([])
+    for principal in principals:
+        # Ignore suspicious principals from annotations
+        if principal.startswith('system.'):
+            continue
+        if principal == 'group:__world__':
+            result.add(security.Everyone)
+        elif principal == 'group:__authenticated__':
+            result.add(security.Authenticated)
+        else:
+            result.add(principal)
+    return list(result)
--- a/h/api/models.py
+++ b/h/api/models.py
@@ -217,6 +217,19 @@ class Annotation(annotation.Annotation):
         return parser.parse(self["created"]).strftime("%Y-%m-%d")
 
     @property
+    def parent(self):
+        """
+        Return the thread parent of this annotation, if it exists.
+        """
+        if 'references' not in self:
+            return None
+        if not isinstance(self['references'], list):
+            return None
+        if not self['references']:
+            return None
+        return Annotation.fetch(self['references'][-1])
+
+    @property
     def target_links(self):
         """A list of the URLs to this annotation's targets."""
         links = []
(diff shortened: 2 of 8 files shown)


nickstenning:
Oh, and it's worth noting that this branch includes the work in #2556, so we should merge that and rebase this branch afterwards.

nickstenning:
Rebased on top of master. Good to go (I hope).'

nickstenning:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;80lS1f2Z&#x2F;114-2555-reply-notification-emails-not-going-out"> #2555 Reply notification emails not going out</a>

h/api/auth.py
```diff
@@ -13,3 +15,21 @@ def get_user(request):
...
+
+
+def translate_annotation_principals(principals):
+    """
+    Translate a list of annotation principals to a list of pyramid principals.
+    """
+    result = set([])
+    for principal in principals:
+        # Ignore suspicious principals from annotations
+        if principal.startswith('system.'):
```
seanh: What's the problem here? We're trying to avoid letting people make annotations writable by system.Everyone so that they can come back later when not logged in and edit them?

h/api/auth.py
```diff
@@ -13,3 +15,21 @@ def get_user(request):
...
+            continue
+        if principal == 'group:__world__':
+            result.add(security.Everyone)
```
seanh: Hmm. This means that if I use the API to make an annotation with `'update': ['group:__world__']` then even people who aren't logged-in can edit the annotation. If we changed it to `security.Authenticated` here then they'd at least need to have an account. Or maybe this should be done at create time - don't allow `'group:__world__'` in the update permissions, or don't have the client sending the permissions at all!

seanh: Also, you could already do this before this pr was merged, so never mind

seanh:
I've taken a closer look at this now and I don't see any problems with it.

I think the way that we persist permissions on the annotations, and let the clients supply those permissions, should probably be re-considered but that's not an issue with this pull request.

h/api/auth.py
nickstenning: That's the definition of `group:__world__` ... and yes, `'update': ['group:__world__']` would allow anyone to update the annotation, just as `'read': ['group:__world__']` allows anyone to read the annotation, even if they're not logged in.

Also, as you note, this isn't a change in behaviour. This function has just been factored out.

h/api/auth.py
nickstenning: The problem is that `translate_annotation_principals` could be used in many different contexts, and I can't predict what all of those will be. Allowing data from outside to map to pyramid "special" principals is an accident waiting to happen.

PR #2560, opened 2015-09-30
When transforming old-style comments, handle null targets

nickstenning:
There are loads of old-style comments, mostly from arXiv, which have explicit null targets. These should be treated as "empty" for the purposes of mapping to new-style comments.


nickstenning:
c.f. https://groups.google.com/a/list.hypothes.is/forum/#!topic/eng/4SKGcHDqlrI

PR #2562, opened 2015-09-30
Fix a bug where canceling changes did not reset the text

### LINKED ISSUE #2561: Canceling edit does not reset annotation text field if it was previously empty
**Steps to reproduce**

1. On a page with H enabled, select some text and click the Highlight action
2. Select the highlight and click the 'Edit' link in the sidebar
3. Add some text for the annotation
4. Click the 'Cancel' button.

**Expected result**

Highlight reverts to previous state and the text is cleared

**Actual result**

The text added to the annotation appears to have been saved. Refreshing the page however shows that it hasn't actually been updated.

robertknight:
AnnotationController.revert() reverted changes to the
annotation in the editor by copying across properties
from the saved annotation to the draft.

Any properties which were _added_ in the draft but not
set in the saved version, such as the text when the text
is empty, were not reverted.

Fixes #2561

### DIFF
--- a/h/static/scripts/directive/annotation.coffee
+++ b/h/static/scripts/directive/annotation.coffee
@@ -281,7 +281,7 @@ AnnotationController = [
     this.render = ->
       # Extend the view model with a copy of the domain model.
       # Note that copy is used so that deep properties aren't shared.
-      angular.extend @annotation, angular.copy model
+      @annotation = angular.extend {}, angular.copy model
 
       # Set the URI
       @annotationURI = new URL("/a/#{@annotation.id}", this.baseURI).href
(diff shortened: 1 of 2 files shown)


nickstenning:
LGTM.

h/static/scripts/directive/annotation.coffee
```diff
@@ -281,7 +281,7 @@ AnnotationController = [
     this.render = ->
       # Extend the view model with a copy of the domain model.
       # Note that copy is used so that deep properties aren't shared.
-      angular.extend @annotation, angular.copy model
+      @annotation = angular.extend {}, angular.copy model
```
tilgovi: This could just be `@annotation = angular.copy(model)` I think.

robertknight: Yes, indeed. Hadn't read the docs for `angular.copy()` - it creates the destination object implicitly.

Issue #2563, opened 2015-09-30
Intermittent Bridge.createChannel() test failure


robertknight:
The `Bridge.createChannel()` test is intermittently timing out. eg. See https://travis-ci.org/hypothesis/h/builds/82894887

nickstenning:
I can't remember the last time I saw this happen. Closing.

PR #2564, opened 2015-09-30
Warn about slow tests and add extended timeout for flakey test

robertknight:
Enable reporting of all tests that take longer than 500ms
to execute. In combination with high or disabled timeouts
for individual tests, this can be used to determine whether
tests that intermittently time out within the 2000ms threshold
are slow or hanging or failing to trigger the async callback.

Add a high timeout for the Bridge.createChannel() test which
is intermittently failing, eg. in https://travis-ci.org/hypothesis/h/builds/82894887

A few things to note about that run:
 - The overall completion time for the test suite was higher
   than average by more than 2000ms, so other tests may have
   taken a lot longer to run than normal
 - The test which timed out is pretty simple and synchronous,
   but there are a number of other async tests in the same suite so
   perhaps one of those tests is interfering with it

See #2563

### DIFF
--- a/h/static/scripts/karma.config.js
+++ b/h/static/scripts/karma.config.js
@@ -90,6 +90,13 @@ module.exports = function(config) {
     // possible values: config.LOG_DISABLE || config.LOG_ERROR || config.LOG_WARN || config.LOG_INFO || config.LOG_DEBUG
     logLevel: config.LOG_INFO,
 
+    // warn about tests that take a long time to execute,
+    // in combination with per-test timeouts this can also be used
+    // to discover whether tests that intermittently timeout are merely slow,
+    // in which case they will trigger a warning,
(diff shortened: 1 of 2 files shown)


nickstenning:
Very happy to add a warning for slow tests, but I don't really want to bump the timeout for the flaky test. Looking briefly at the code I can see no reason at all why that test should be timing out — as you say, it looks entirely synchronous — which would make me worry that something much weirder is going on.

robertknight:
I've realized that if this test does exceed the default timeout again (2s) it will very likely unnoticed because the slow-tests check is only a warning and doesn't cause the test to actually fail. What is actually needed is for the test to still fail, but give an indication of whether it took longer than expected (say 3s) to run or "never" completes at all.

nickstenning:
Okay, I'm going to close this for now.

PR #2566, opened 2015-09-30
Implement focus on a group

seanh:
When a user clicks on a group name in the "scope selector" dropdown
menu, we want to filter the visible annotations by group. This commit
implements a rather inefficient but functional approach to this, simply
reloading the angular view.

In addition, we make sure that the initial annotation load includes a
filter on the currently-focused group, and we discard non-focused-group
annotations coming in over the websocket.

### DIFF
--- a/h/api/test/views_test.py
+++ b/h/api/test/views_test.py
@@ -5,6 +5,7 @@ import mock
 import pytest
 
 from pyramid import testing
+from pyramid import httpexceptions
 
 from h.api import views
 
@@ -216,6 +217,16 @@ def test_create_returns_render(search_lib):
 
 # The fixtures required to mock all of read()'s dependencies.
 read_fixtures = pytest.mark.usefixtures('search_lib', 'AnnotationEvent')
+
+
+@ read_fixtures
+def test_read_404s_for_group_annotations_if_groups_feature_is_off():
+    context = mock.Mock(model={'group': 'foo'})
+    request = mock.Mock()
+    request.feature.return_value = False
+
+    with pytest.raises(httpexceptions.HTTPNotFound):
+        views.read(context, request)
 
 
 @read_fixtures
--- a/h/api/views.py
+++ b/h/api/views.py
@@ -167,6 +167,15 @@ def read(context, request):
     """Return the annotation (simply how it was stored in the database)."""
     annotation = context.model
 
+    # Don't show group annotations for users who don't have the groups feature
+    # enabled, even if they are members of the group (this can happen if a user
+    # joins a group and then the groups feature is later disabled for that
+    # user).
+    if not request.feature('groups'):
+        if annotation.get('group') not in (None, '__world__'):
+            raise httpexceptions.HTTPNotFound()
+
+
     # Notify any subscribers
     _publish_annotation_event(request, annotation, 'read')
 
--- a/h/static/scripts/app-controller.coffee
+++ b/h/static/scripts/app-controller.coffee
@@ -36,6 +36,10 @@ module.exports = class AppController
     # Default sort
     $scope.sort = name: 'Location'
 
+    $scope.$on('groupFocused', (event) ->
+      $route.reload()
+    )
+
     identity.watch({
       onlogin: (identity) -> $scope.auth.user = auth.userid(identity)
       onlogout: -> $scope.auth.user = null
--- a/h/static/scripts/app-controller.coffee
+++ b/h/static/scripts/app-controller.coffee
@@ -32,8 +32,9 @@ module.exports = class AppController
     $scope.accountDialog = visible: false
     $scope.shareDialog = visible: false
 
-    # Check to see if we are on the stream page so we can hide share button.
-    $scope.isEmbedded = $window.top isnt $window
+    # Check to see if we're in the sidebar, or on a standalone page such as
+    # the stream page or an individual annotation page.
+    $scope.isSidebar = $window.top isnt $window
 
     # Default sort
     $scope.sort = name: 'Location'
--- a/h/static/scripts/app-controller.coffee
+++ b/h/static/scripts/app-controller.coffee
@@ -4,11 +4,13 @@ angular = require('angular')
 module.exports = class AppController
   this.$inject = [
     '$controller', '$document', '$location', '$route', '$scope', '$window',
-    'annotationUI', 'auth', 'drafts', 'features', 'identity', 'session'
+    'annotationUI', 'auth', 'drafts', 'features', 'groups', 'identity',
+    'session'
   ]
   constructor: (
      $controller,   $document,   $location,   $route,   $scope,   $window,
-     annotationUI,   auth,   drafts,   features,   identity,  session
+     annotationUI,   auth,   drafts,   features,   groups,   identity,
+     session
   ) ->
     $controller('AnnotationUIController', {$scope})
 
@@ -42,7 +44,14 @@ module.exports = class AppController
 
     identity.watch({
       onlogin: (identity) -> $scope.auth.user = auth.userid(identity)
-      onlogout: -> $scope.auth.user = null
+      onlogout: ->
+        $scope.auth.user = null
+
+        # Currently all groups are private so when the user logs out they can
+        # no longer see the annotations from any group they may have had
+        # focused. Focus the public group instead, so that they see any public
+        # annotations in the sidebar.
+        groups.focus('__world__')
       onready: -> $scope.auth.user ?= null
     })
 
--- a/h/static/scripts/app.coffee
+++ b/h/static/scripts/app.coffee
@@ -12,8 +12,8 @@ socket = null
 resolve =
   store: ['store', (store) -> store.$promise]
   streamer: [
-    '$websocket', 'annotationMapper'
-    ($websocket,   annotationMapper) ->
+    '$websocket', 'annotationMapper', 'groups'
+    ($websocket,   annotationMapper,   groups) ->
       # Get the socket URL
       url = new URL('/ws', baseURI)
       url.protocol = url.protocol.replace('http', 'ws')
@@ -32,6 +32,14 @@ resolve =
         action = message.options.action
         annotations = message.payload
         return unless annotations?.length
+
+        # Discard annotations that aren't from the currently focused group.
+        # FIXME: Have the server only send us annotations from the focused
+        # group in the first place.
+        annotations = annotations.filter((ann) ->
+          return ann.group == groups.focused().id
+        )
+
         switch action
           when 'create', 'update', 'past'
             annotationMapper.loadAnnotations annotations
--- a/h/templates/app.html.jinja2
+++ b/h/templates/app.html.jinja2
@@ -17,7 +17,7 @@
     <div class="inner content">
       <button class="btn btn-clean"
               ng-click="shareDialog.visible = !shareDialog.visible"
-              ng-if="isEmbedded"
+              ng-if="isSidebar"
               title="Share this page">
         <i class="h-icon-share btn-icon"></i>
       </button>
@@ -29,7 +29,7 @@
            on-clear="search.clear()"></div>
       <!-- / Searchbar -->
 
-      <group-list class="group-list" ng-if="auth.user && feature('groups') && isEmbedded">
+      <group-list class="group-list" ng-if="auth.user && feature('groups') && isSidebar">
       </group-list>
 
       <div ng-switch="auth.user">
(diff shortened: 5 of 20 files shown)


seanh:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;F26qrUFw&#x2F;49-focus-on-a-group"> Focus on a group</a>

seanh:
- [x] Change the shown annotations when the selected group changes
- [ ] Don't break the stream or individual annotation pages
- [x] Also filter annotations from other groups out from live updates
- [ ] Focus the public group on sign out
- [ ] Make sure that everything works properly when the focused group is changed when an annotation is being created or edited
- [ ] Tests
- [ ] Make sure everything is fine when the groups feature flag is off

seanh:
I think it can be left for a later refactoring in a separate pull request, but couldn't this all be done in terms of routing? Instead of  the groups service broadcasting "groupFocused", have the items in the groups dropdown just be links to different Angular "pages" (`#` URLs), so that clicking on them will trigger a load of a new route.

seanh:
Start creating a new annotation (annotation card in sidebar, editor open), don't save it yet, and change the group in the groups dropdown. The current behaviour is definitely broken, but what _should_ happen?

1. The in-progress annotation remains visible, but its group changes with the groups dropdown?

2. Or the in-progress annotation disappears, but if you switch back to the original group it is there again with the editor still open and any text you had typed still there?

If 2, that raises the question of the user losing data if they go to a different web page. Currently h won't stop you from going to a different web page with annotation edits open, you'll just lose those edits. But at least those edits were "on screen", although they may have been scrolled out of view in the sidebar. Groups just makes this situation worse.

Start editing an annotation but don't save it, and change the group in the groups dropdown. Again the current behaviour on this branch is broken. In this case I think the annotation should definitely disappear along with all the another annotations from the group you're leaving, but if you go back to that group the annotation should probably still be open with any edits you had done still there.

Both of these just point to how broken inline (rather than modal) creating and editing is.

seanh:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;C74bnlsL&#x2F;110-do-not-show-group-selector-in-the-stream-view"> Do not show Group selector in the Stream view</a>

seanh:
I'm seeing a group's annotations instead of the public ones in the sidebar, when running it with the groups feature disabled. Think it's because I had a group focused when it was enabled, and it's still getting it from localStorage.

h/templates/app.html.jinja2
```diff
@@ -29,7 +29,7 @@
            on-clear="search.clear()"></div>
       <!-- / Searchbar -->
 
-      <group-list class="group-list" ng-if="auth.user && feature('groups')">
+      <group-list class="group-list" ng-if="auth.user && feature('groups') && isEmbedded">
```
nickstenning: I'm not sure that's what you want. You yourself added `isSidebar` not so long ago: 255a820.

seanh:
From <https://github.com/hypothesis/h/commit/3a3115047d1da5abe4a4e343d08300fd3eb4381b>, the new `isSidebar` variable replaces several existing variables and is used to hide/show:

The share dialog (share this page on twitter etc) is shown in the top bar,
only when in the sidebar (and not on the `/stream` page or on individual
annotation pages).

screenshot from 2015-10-01 12 23 18


The group list dropdown menu, shown in the top bar only when in the sidebar.

screenshot from 2015-10-01 12 23 44

The "location" entry in the sort menu (sorted by newest, oldest, location)
shown only when in the sidebar.

screenshot from 2015-10-01 12 24 14

Note that a separate variable `isStream` (which is true when in the sidebar or on the stream, but not when on individual annotation pages) is used to show/hide this menu, `isSidebar` is then used to show/hide the "Location" item in the menu.

The "citation" on annotations that says which page was annotated
(the 'on "Example Document" (example.com)' in the annotation's title),
is shown only when _not_ in the sidebar.

screenshot from 2015-10-01 12 24 41

seanh:
Group annotations are showing in the stream when the groups feature is off. Another edge case - the groups feature will have had to be on for the user previously, so that they could join a group in order to be authorised to see its annotations. But we should probably change the search API to always insert a `group:__public__` filter when the groups feature is off

seanh:
Apart from that everything seems fine when the groups feature is turned off.

Actually, I think we probably shouldn't bother hiding group annotations from the stream when groups is off, it's not a big problem if it does occur, and would only occur for users who joined a group then had the groups feature turned off again afterwards.

seanh:
Hmm, actually when you see group annotations in the stream or on individual pages when the groups feature is off and you reply to one of those annotations, then it gets weird. So I've pushed a couple commits to hide them

h/static/scripts/app-controller.coffee
```diff
@@ -4,11 +4,13 @@ angular = require('angular')
 module.exports = class AppController
   this.$inject = [
     '$controller', '$document', '$location', '$route', '$scope', '$window',
-    'annotationUI', 'auth', 'drafts', 'features', 'identity', 'session'
+    'annotationUI', 'auth', 'drafts', 'features', 'identity', 'session',
+    'groups'
```
nickstenning: Could you re-alphabetise this?

seanh:
@robertknight Ok, this is ready to be reviewed. What this should do:

- [ ] https://trello.com/c/F26qrUFw/49-focus-on-a-group When you select a group in the groups dropdown, the sidebar changes to show only annotations from that group
- [ ] https://trello.com/c/C74bnlsL/110-do-not-show-group-selector-in-the-stream-view The groups dropdown does not appear on the stream or on individual annotation pages
- [ ] When you logout it should show the annotations from the public group (even if you had another group focused)
- [ ] "Live update" annotations (that come into the client via the WebSocket) should also be filtered to only ones from the currently focused group
- [ ] The stream page still shows annotations from all groups
- [ ] When the groups feature is disabled, everything should continue working as before

* It's totally broken if you change between groups while editing an annotation, we're aware of that and want to merge it anyway

Let me know if you want any help with how to test it

robertknight:
@seanh Went through your checklist above, I'll note bugs as I go:

>  The stream page still shows annotations from all groups

The stream view does not show the group dropdown selector but only shows annotations from whichever group I last selected in another H tab.

**Steps to reproduce**

1. Open `http://localhost:5000`
2. Add a public annotation and a group annotation to that page. Select the 'Public' view
3. Go to http://localhost:5000/stream in another tab
4. Go back to the first tab and select the Group from the dropdown
5. Go back to the stream tab and refresh the page

**Expected**

The stream tab displays all annotations regardless of what is selected in the other tab.

**Actual**

The stream tab displays annotations from the group that is selected in the dropdown in the other tab.

robertknight:
> It's totally broken if you change between groups while editing an annotation, we're aware of that and want to merge it anyway

Can you clarify what 'totally broken' means and the plan to fix it after merge?

h/api/test/views_test.py
```diff
@@ -218,6 +219,16 @@ def test_create_returns_render(search_lib):
 read_fixtures = pytest.mark.usefixtures('search_lib', 'AnnotationEvent')
 
 
+@ read_fixtures
+def test_read_404s_for_group_annotations_if_groups_feature_is_off():
```
robertknight: From a client POV I would prefer more descriptive error codes if a feature is not enabled. I think that is a convention to discuss separately though.

h/api/views.py
```diff
@@ -167,6 +167,15 @@ def read(context, request):
     """Return the annotation (simply how it was stored in the database)."""
     annotation = context.model
 
+    # Don't show group annotations for users who don't have the groups feature
+    # enabled, even if they are members of the group (this can happen if a user
+    # joins a group and then the groups feature is later disabled for that
+    # user).
+    if not request.feature('groups'):
+        if annotation.get('group') not in (None, '__world__'):
+            raise httpexceptions.HTTPNotFound()
```
robertknight: Likewise, if a user unexpectedly gets a 404 during the testing phase because of this, it won't be obvious why.

seanh:
Sounds like the stream page is more broken than I realised, I'll investigate.

> > It's totally broken if you change between groups while editing an annotation, we're aware of that and want to merge it anyway
> 
> Can you clarify what 'totally broken' means and the plan to fix it after merge?

If you create a new annotation but don't save it, leaving the annotation editor / textarea open, and change to a different in the groups dropdown, then the annotation you were editing disappears along with all the other annotations from the previous group (arguably the correct behaviour so far), but then if you change back to the original group the unsaved annotation is still there but anything you'd typed is gone. Screencast:

untitled screencast 1

If you do the same with editing an existing annotation you also lose your edits but in this case the editor closes as well:

untitled screencast 2

h/static/scripts/app-controller.coffee
```diff
@@ -30,15 +32,27 @@ module.exports = class AppController
...
-    # Check to see if we are on the stream page so we can hide share button.
-    $scope.isEmbedded = $window.top isnt $window
+    # Check to see if we're in the sidebar, or on a standalone page such as
+    # the stream page or an individual annotation page.
+    $scope.isSidebar = $window.top isnt $window
 
     # Default sort
     $scope.sort = name: 'Location'
 
+    $scope.$on('groupFocused', (event) ->
```
robertknight: I don't think the approach of using events is bad, but I do think it would help to have all the events that are used across the application defined in one place. They can then be referenced elsewhere.

eg:

```js
// event-types.js
module.exports = {
  /** Description of when event is used and the associated payload data */
  GROUP_FOCUSED: 'groupFocused'
};

// some-other-file.js
var events = require('path/to/event-types');

$scope.$on(events.GROUP_FOCUSED, function (event) {
 ...
});
```

h/static/scripts/app-controller.coffee
```diff
@@ -30,15 +32,27 @@ module.exports = class AppController
...
+      $route.reload()
```
robertknight: Nice simple approach.

h/static/scripts/app-controller.coffee
```diff
@@ -30,15 +32,27 @@ module.exports = class AppController
...
       onlogin: (identity) -> $scope.auth.user = auth.userid(identity)
-      onlogout: -> $scope.auth.user = null
+      onlogout: ->
+        $scope.auth.user = null
+
+        # Currently all groups are private so when the user logs out they can
+        # no longer see the annotations from any group they may have had
+        # focused. Focus the public group instead, so that they see any public
+        # annotations in the sidebar.
+        groups.focus('__world__')
```
robertknight: This is a reasonable approach. An alternative would be to broadcast a LOGGED_OUT event, which the groups service listens for and in turn updates the focused group. The advantage would be that all groups logic would be contained in that service, and writing a test would be very easy.

h/static/scripts/app.coffee
```diff
@@ -32,6 +32,14 @@ resolve =
         action = message.options.action
         annotations = message.payload
         return unless annotations?.length
+
+        # Discard annotations that aren't from the currently focused group.
+        # FIXME: Have the server only send us annotations from the focused
+        # group in the first place.
```
robertknight: As a separate task, I think it would be very useful to display counts of annotations across different groups in the groups dropdown, so even if the server sends only the annotation payloads for the focused group, having facet counts from ES for the other groups would be very useful.

h/static/scripts/app-controller.coffee
robertknight: ... and it looks like we don't currently have a test for this in `app-controller-test`?

h/static/scripts/app-controller.coffee
seanh: Is this a refactor that can be done in a separate pr?

h/static/scripts/app-controller.coffee
seanh: To be honest, I couldn't be bothered to figure out how to fake events in tests just to test this. I'll look into it..

h/static/scripts/app.coffee
seanh: Yeah that does sound useful

robertknight:
* I've tested the behavior when annotation push updates (ie. the web socket) is enabled, that works as expected.

* Did a review of the code, I have a question about a test and a possible alternative approach to clearing the focused group on logout: https://github.com/hypothesis/h/pull/2566/files#r41011272 . Otherwise, LGTM

* With regards to changing the selected group whilst an annotation is being edited, I agree we can address that separately. Priority for V1 is to make it so that you can't trivially unsaved changes this way.

h/static/scripts/app-controller.coffee
seanh: Test added

PR #2569, opened 2015-10-01
Consolidate thread show/hide logic into ThreadController

nickstenning:
This template conditional is notorious for accumulating extra terms, so
this commit rolls it into ThreadController where it can be more easily
tested.

### DIFF
--- a/h/static/scripts/directive/thread.coffee
+++ b/h/static/scripts/directive/thread.coffee
@@ -13,7 +13,8 @@ uuid = require('node-uuid')
 # the collapsing behavior.
 ###
 ThreadController = [
-  ->
+  '$scope',
+  ($scope) ->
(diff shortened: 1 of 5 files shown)


h/static/scripts/directive/thread.coffee
```diff
@@ -13,7 +13,8 @@ uuid = require('node-uuid')
 # the collapsing behavior.
 ###
 ThreadController = [
-  ->
+  '$scope',
```
robertknight: Is `// @ngInject` supported in CoffeeScript sources (ie. does that transform run before or after CS processing)?

nickstenning: Sadly not. Only for JS at the moment.

PR #2570, opened 2015-10-01
Show filenames of local documents with titles

seanh:
When the annotated document is a local (file://) document and it has a
title, then show the filename in place of the domain on the annotation
card. For example:

  seanh on "Document Title" (filename.pdf)

In cases where the document has no title then the filename would already
have been used in place of "Document Title" so don't repeat it in
(...)'s.

For remote documents the domain name is shown in (...)'s instead and the
document title is hyperlinked so the people can see the full URL by
hovering over the link.

### DIFF
--- a/h/static/scripts/filter/test/document-domain-test.js
+++ b/h/static/scripts/filter/test/document-domain-test.js
@@ -29,12 +29,32 @@ describe('documentDomain', function() {
     assert(domain === '');
   });
 
+  it('returns the filename for local documents with titles', function() {
+    var domain = documentDomainFilterProvider()({
+      title: 'example.com',
+      uri: 'file:///home/seanh/MyFile.pdf'
+    });
+
+    assert(domain === '(MyFile.pdf)');
+  });
+
   it('escapes HTML in the document domain', function() {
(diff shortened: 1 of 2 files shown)


seanh:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;Kj8vWkme&#x2F;106-show-the-filename-on-cards-for-local-files"> Show the filename on cards for local files.</a>

h/static/scripts/filter/test/document-domain-test.js
```diff
@@ -29,6 +29,15 @@ describe('documentDomain', function() {
     assert(domain === '');
   });
 
+  it('returns the filename for local documents with titles', function() {
+    var domain = documentDomainFilterProvider()({
+      title: 'example.com',
+      uri: 'file:///home/seanh/MyFile.pdf'
+    });
+
+    assert(domain === '(MyFile.pdf)');
```
robertknight: Use `assert.equal(actual, expected)` rather than `assert(actual === expected)` and you'll get more helpful error messages when the test fails.

Issue #2571, opened 2015-10-01
Broken URL references in production mode CSS


robertknight:
URL rebasing is disabled when merging and minifying CSS files for production, so the relative URLs in `styles/vendor/icomoon.css` and `styles/vendor/katex.css` are not correctly rewritten when they are 
concatenated into `styles.hypothesis.min.css` and `styles/app.min.css`.

The fonts in icomoon.css happen to work because the TTF version is inlined into the file as a data URL.

This should be fixed as part of moving the asset building pipeline out of Python.

nickstenning:
I'm pretty sure this was fixed by #3209. Feel free to reopen if not.

PR #2572, opened 2015-10-01
Fix icon fonts in production builds

robertknight:
As described in GH #2571, URL rebasing is broken
in the CSS asset pipeline, so relative URLs in
CSS files under styles/vendor are incorrect.

Previously icomoon.css worked because one of the
font variants (TTF) was inlined as a data URL.
This broken in the previous update which had only
the relative .woff URL.

This commit patches the issue by using an absolute URL
in the icomoon.css file. It will be fixed properly in
the CSS build as part of moving that out of the Python
web app.

See #2571

### DIFF
--- a/h/static/styles/vendor/icomoon.css
+++ b/h/static/styles/vendor/icomoon.css
@@ -1,6 +1,13 @@
 @font-face {
 	font-family: 'h';
-	src: url('fonts/h.woff?e944zd') format('woff');
+	/* WARNING - the URL below was modified from the generated URL
+	 * to use an absolute instead of relative URL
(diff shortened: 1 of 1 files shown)


tilgovi:
I removed the cssrewrite filter from the webassets bundles. I thought it wasn't necessary anymore since I started using `img-url`, but forgot about the plain CSS files. You should just be able to add cssrewrite to the filters again.

59aeb9222b378b90abf080235dd8558940d46776

robertknight:
@tilgovi Thanks. Once we have the initial iteration of groups out, I'd like to spend some time re-doing the asset build pipeline. In the process all of the CSS build can be moved to postcss.

PR #2573, opened 2015-10-01
Add a Chrome badge blacklist

seanh:
Add a blacklist of URLs that the Chrome extension's badge will never show a number for.

### DIFF
--- a/h/admin.py
+++ b/h/admin.py
@@ -190,7 +190,7 @@ def users_index(request):
                   renderer='h:templates/admin/badge.html.jinja2',
                   permission='admin')
 def badge_index(_):
-    return {"uris": models.BadgeBlocklist.all()}
+    return {"uris": models.Blocklist.all()}
 
 
 @view.view_config(route_name='admin_badge',
@@ -200,7 +200,7 @@ def badge_index(_):
                   permission='admin')
 def badge_add(request):
     try:
-        request.db.add(models.BadgeBlocklist(uri=request.params['add']))
+        request.db.add(models.Blocklist(uri=request.params['add']))
     except ValueError as err:
         request.session.flash(err.message, 'error')
     return badge_index(request)
@@ -213,7 +213,7 @@ def badge_add(request):
                   permission='admin')
 def badge_remove(request):
     uri = request.params['remove']
-    request.db.delete(models.BadgeBlocklist.get_by_uri(uri))
+    request.db.delete(models.Blocklist.get_by_uri(uri))
     return badge_index(request)
 
 
--- a/h/admin.py
+++ b/h/admin.py
@@ -185,6 +185,38 @@ def users_index(request):
     return {'username': username, 'user': user}
 
 
+@view.view_config(route_name='admin_badge',
+                  request_method='GET',
+                  renderer='h:templates/admin/badge.html.jinja2',
+                  permission='admin')
+def badge_index(_):
+    return {"uris": models.BadgeBlocklist.all()}
+
+
+@view.view_config(route_name='admin_badge',
+                  request_method='POST',
+                  request_param='add',
+                  renderer='h:templates/admin/badge.html.jinja2',
+                  permission='admin')
+def badge_add(request):
+    try:
+        request.db.add(models.BadgeBlocklist(uri=request.params['add']))
+    except ValueError as err:
+        request.session.flash(err.message, 'error')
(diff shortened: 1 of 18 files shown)


seanh:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;xgknj2RP&#x2F;109-don-t-show-chrome-badge-numbers-on-certain-pages"> Don&#39;t show Chrome badge numbers on certain pages</a>

nickstenning:
Is there a reason why this blacklist would be distinct from the blacklist that already exists to prevent hypothesis from activating on certain sites?

nickstenning:
Worth noting that blacklist or no blacklist, I'd be very happy to see a PR that adds a badge-specific endpoint, because it would make it a lot easier to then cache that endpoint separately from our search API.

seanh:
I think so, yes. They're two different things: h is completely borked on this site, disable it and show the user an error/apology vs just don't show the annotation count on this page.

If a page is completely blocked though, that ought to imply that it's automatically badge-blocked as well.

They might be able to share some implementation mechanism. iirc the existing blocklist is implemented as a config file setting, built into the Chrome extension at build time. With the badge blocklist the Chrome extension is sending a request to the server to get the badge number so all we need to do is make the server respond with 0 for blocked sites, the list of sites can be server-side and editable by admins. So we could maybe put the blocked or not flag in this server response as well.

seanh:
- [ ] Move blacklist into database
- [ ] Add admin page for blacklist
- [ ] Add server-side tests
- [ ] Support regexes in blacklist

h/badge/models.py
```diff
@@ -0,0 +1,52 @@
...
+
+    @classmethod
+    def all(cls):
+        return cls.query.all()
+
+    @classmethod
+    def is_blocked(cls, uri):
+        """Return True if the given URI is blocked."""
+        for pattern in cls.all():
+            if fnmatch.fnmatch(uri, unicode(pattern)):
```
seanh: This might be a lot of work for the server? If we have a lot of patterns. Let's see I guess

h/badge/views.py
```diff
@@ -0,0 +1,33 @@
+from pyramid import httpexceptions
+
+from h import models
+from h.api.views import api_config
+from h.api.resources import Root
+from h.api import search as search_lib
```
seanh: Hmmmm

seanh:
@nickstenning Can be reviewed

seanh:
@nickstenning Moved it to `LIKE` as requested

seanh:
@nickstenning Done the table rename

h/badge/views.py
nickstenning: Ah. Well here's the thing. I don't think this should live under `/api`. It's not part of our public API. In keeping with where stuff like the session JSON blob lives, I think it should live at `/app/badge` or whatever.

I'm not going to block merging on this but we should change this before we push it out.

seanh: Fine, I'll make this change now as a separate pr

h/models.py
```diff
@@ -28,6 +29,7 @@
 NipsaUser = nipsa_models.NipsaUser
 Subscriptions = notification_models.Subscriptions
 User = accounts_models.User
+Blocklist = badge_models.Blocklist
```
nickstenning: Alphabetisation!

nickstenning: And you need to add it to `__all__`.

h/admin.py
```diff
@@ -185,6 +185,38 @@ def users_index(request):
...
+    try:
+        request.db.add(models.Blocklist(uri=request.params['add']))
+    except ValueError as err:
+        request.session.flash(err.message, 'error')
+    return badge_index(request)
+
+
+@view.view_config(route_name='admin_badge',
+                  request_method='POST',
+                  request_param='remove',
```
nickstenning: This is inconsistent with all the other remove views in this file. I'm sorry if this is nitpicking but inconsistency is just avoidable overhead when trying to understand the code.

seanh: Alright, but the existing ones are wrong, so I've added a commit to change them as well

h/models.py
seanh: Done

seanh:
Rebased

nickstenning:
Looks good to me.

PR #2574, opened 2015-10-02
Avoid String.prototype.startsWith() for compatibility Safari and IE

### LINKED ISSUE #2568: Safari doesn't display links to the original article in annotations
When visiting a stream of annotations or a single annotation in Safari 8.0, the link to the original article is not displayed.  See for example:

nolinks

vs Chrome

links

robertknight:
Links to source document URIs were broken under Safari <= 8 and
other browsers as well due to reliance on the
String.prototype.startsWith() API

This issue went unnoticed in Karma/PhantomJS tests because
Karma was primed with the complete set of polyfills from
the js-polyfills library, including many (all?) ES6 polyfills.

This same set of polyfills was_not_ used in the browser build however.

This commit fixes the original issue by using 'indexOf()' instead of
'startsWith()' and reduces the likelihood of such problems going
unnoticed during testing by using a much more minimal set of polyfills
under PhantomJS.

The alternative approach would be to go in the opposite direction
and polyfill any ES6 APIs in the browser. However, we need to be
careful about code bloat, so the approach taken here is the more
conservative option.

Fixes #2568

### DIFF
--- a/h/static/scripts/filter/document-domain.js
+++ b/h/static/scripts/filter/document-domain.js
@@ -17,7 +17,7 @@ module.exports = function() {
     var domain = escapeHtml(document.domain || '');
     var title = escapeHtml(document.title || '');
 
-    if (uri.startsWith('file://') && title) {
+    if (uri.indexOf('file://') === 0 && title) {
       var parts = uri.split('/');
       var filename = parts[parts.length - 1];
       if (filename) {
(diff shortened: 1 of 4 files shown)


robertknight:
CC @seanh

seanh:
Thanks, @robertknight !

judell:
"However, we need to be careful about code bloat, so the approach taken here is the more
conservative option." 

+1

PR #2577, opened 2015-10-02
Group landing pages redesign

robertknight:
This implements the new landing pages for groups. I still want to reorganize and tidy up the commits in this PR but I'm posting the PR now to review the design/flow and also to get feedback on the proposed mechanism to optimize the 'Join a Group' flow.

 * The 'Join Group X' and 'Login to Join Group X' pages have been merged into a single template,
   as the designs are the same.
 * ~~To improve the common flow where a user who is not signed in and quite likely not signed up
    receives a link, the 'Login to Join Group X' page subscribes to an event service. The event
    service is currently a basic queue which polls a JSON array stored in localStorage
    at a slow interval. When the user signs in, an event is added to this queue and the "Login to Join"
    tab picks this up and automatically refreshes. I was thinking of using the same mechanism to
    refresh the groups list in any open tabs after the user creates a new group.~~ _Edit: We decided
    to approach this differently. Currently the user must refresh the 'Sign in to join' page to see
    the join link_
 * The basic functionality of the groups pages does not require JS, but for nicities like auto-refreshing
    the page, client-side validation of the group name form etc, there is a new bundle `site-bundle.js`
    which contains site (rather than app)-specific code. Likewise there is also a new `site.scss`
    file which contains styling that is only used for pages on the site, rather than pulling in all of the
    app styling
 * To help achieve consistency in the design, I've extended the color palette in `variables.scss`. Since the grays
     were previously named 'gray-light', 'gray-lighter' etc. trying to fit new names into this namespace
     would have been awkward. To solve this, I used [this handy tool](http://chir.ag/projects/name-that-color) to come up with descriptive and grep-able names for the new colors from the design.

The flow now looks like this.

**1. User clicks 'New Group'**

t93-1-create-group-initial

**2. User fills out the group name and possibly clicks the info link**

t93-2-create_group_filled
t93-3-create-group-explained

**3. User clicks the 'Create Group' button and gets taken to the share page**

t93-4-share-group

**4. A user who is not logged in receives and visits the link**

t93-5-join-group-not-logged-in

**5. The 'Sign in' button opens the [login page](https://hypothes.is/login) in another tab. Once the user has signed in, (or created an account and then signed in), the original 'Join Group' tab will automatically refresh:**

t93-6-join-group-signed-in

**6. Once the user clicks 'Join Group', they get taken to the share page (step 3 above)**


judell:
Beautiful.

h/assets.yaml
```diff
@@ -214,6 +214,26 @@ inject_css:
         - h:static/styles/variables.scss
         - h:static/styles/mixins/icons.scss
 
+# the H website
+site_js:
+  output: scripts/site-bundle.js
```
nickstenning: You haven't included a minified output for this bundle. Is this intentional?

nickstenning:
Functionally this looks just fine to me. The only question I have (other than the unminifiedness of the site bundle) is about testing strategy.

The swig/extracting sections of the server templates strategy seems to me to be unnecessarily involved and rather brittle. I think it might be clearer to just include some example HTML fixtures in the tests and load those, either just as raw strings in the test suite files or via ajax.

robertknight:
> I think it might be clearer to just include some example HTML fixtures in the tests and
> load those, either just as raw strings in the test suite files or via ajax.

That was my initial approach but the problem then is that you can break the page with a change to the template and the test wouldn't fail. I'd like to set up some basic E2E testing using Selenium or a similar tool that would cover the whole flow.

h/assets.yaml
robertknight: I left it out as the bundle is very small at this stage so I didn't see much point. Happy to change that for consistency if you'd prefer.

nickstenning: Okay, that's fine for now. I'm not entirely sure we even need to be minifying/obfuscating our scripts, because we serve them gzipped, so the delivered size is barely improved. Stripping comments is still valuable, though, so ¯\_(ツ)_/¯. We can revisit this in future.

PR #2578, opened 2015-10-02
Deform form rendering

nickstenning:
So, this is (half of) what I got up to on the train.

----

This PR removes custom form code from the templates for account claim and groups packages, in favour of using generated form code from deform.

The resulting templates *should* be pixel-for-pixel identical with the ones they replace, with the added bonus of some ARIA annotations on form fields, to help navigating forms with a screenreader.

### DIFF
--- a/h/groups/templates/create.html.jinja2
+++ b/h/groups/templates/create.html.jinja2
@@ -13,25 +13,7 @@
     <div class="group-form">
       <i class="h-icon-group group-form__heading-icon"></i>
       <div class="group-form__heading">Create a new group</div>
-      <form class="group-form__form" action="{{ form.action }}" method="{{ form.method }}" novalidate>
-        {{ form['csrf_token'].render() | safe }}
-        <div class="form-field{% if form['name'].error %} form-field-error{% endif %}">
-          <label class="group-form__name-label" for="group-form__name-input">
-            What do you want to call the group?
-          </label>
-          <input class="group-form__name-input js-group-name-input" id="group-form__name-input"
-                 type="text" name="name" placeholder="Group Name" value="{{ data.name }}"
-                 autofocus required maxlength="25" />
-          {% if form['name'].error %}
-            <ul class="form-error-list">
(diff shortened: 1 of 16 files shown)


h/groups/templates/create.html.jinja2
```diff
@@ -12,29 +12,7 @@
   <div class="content paper">
     {% include "h:templates/includes/header.html.jinja2" %}
     <div class="form-vertical">
-      <form class="form" action="{{ form.action }}" method="{{ form.method }}" novalidate>
```
robertknight: This is going to conflict with the changes in https://github.com/hypothesis/h/pull/2577 . We'll have to figure out how to reconcile those tomorrow.

nickstenning:
Okay, rebased and ready to go. @robertknight if you wouldn't mind looking at this (especially for the groups create form) that would be fab.

h/templates/deform/form.jinja2
```diff
@@ -0,0 +1,38 @@
...
+    {{ field.renderer(field.widget.item_template, field=f, cstruct=cstruct.get(f.name, null)) }}
+  {% endfor -%}
+
+  <div class="form-actions">
+    <div class="form-actions-buttons">
+      {%- for button in field.buttons -%}
+        <button id="{{ field.formid + button.name }}"
+                name="{{ button.name }}"
+                type="{{ button.type }}"
+                class="btn{% if button.css_class %} {{ button.css_class }}{% endif %}"
```
robertknight: Mixing the existing CSS classes for forms ('btn') with those for the new-style groups forms might be vulnerable to unexpected styling depending on the order in which the declarations occur.

The forms that I checked looked correct but this is something we might need to revisit.

nickstenning: I did consider setting this up so that it would simply override it, i.e.:

```jinja2
{% if button.css_class %}{{ button.css_class }}{% else %}btn{% endif %}
```

The only reason I didn't do that is it's inconsistent with how `css_class` works everywhere else...

PR #2581, opened 2015-10-05
Server-rendered login form

nickstenning:
This PR replaces the client-side rendered login form with one rendered by deform on the server.

While it would appear that this commit just adds code, that's because it's only an incremental improvement. It actually removes dependencies on a bunch of components we want to remove, and will be able to when the rest of the account forms are rewritten in a similar way. This changeset:

- sets us on the way to a drastic simplification of the account form handling code both on the server and in the client.
- makes a couple of small improvements to user experience (e.g. there is no longer a one second pause after logging in)
- prepares the way for doing "login in a popup," which solves some difficult Same Origin Policy problems.
- makes it much easier to understand what's going on in the login/logout ajax requests sent by the client: the login/logout controller no longer uses AjaxFormViewMapper

### DIFF
--- a/h/accounts/schemas.py
+++ b/h/accounts/schemas.py
@@ -116,9 +116,14 @@ class CSRFSchema(colander.Schema):
 
 
 class LoginSchema(CSRFSchema):
-    username = colander.SchemaNode(colander.String())
+    username = colander.SchemaNode(
+        colander.String(),
+        title=_('Username or email address:'),
+        widget=deform.widget.TextInputWidget(autofocus=True),
+    )
     password = colander.SchemaNode(
         colander.String(),
+        title=_('Password:'),
         widget=deform.widget.PasswordWidget()
     )
 
--- a/h/accounts/test/views_test.py
+++ b/h/accounts/test/views_test.py
@@ -14,6 +14,7 @@ from h.conftest import DummySession
 
 from h.accounts.views import ajax_form
 from h.accounts.views import validate_form
+from h.accounts.views import AjaxAuthController
 from h.accounts.views import AuthController
 from h.accounts.views import ForgotPasswordController
 from h.accounts.views import RegisterController
@@ -45,6 +46,23 @@ class FakeInvalid(object):
 
     def asdict(self):
         return self.errors
+
+
+def form_validating_to(appstruct):
+    form = Mock()
+    form.validate.return_value = appstruct
+    form.render.return_value = 'valid form'
+    return form
+
+
+def invalid_form(errors=None):
+    if errors is None:
+        errors = {}
+    invalid = FakeInvalid(errors)
+    form = Mock()
+    form.validate.side_effect = deform.ValidationFailure(None, None, invalid)
+    form.render.return_value = 'invalid form'
+    return form
 
 
 def test_ajax_form_handles_http_redirect_as_success():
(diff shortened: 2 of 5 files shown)


nickstenning:
Rebased and ready for review.

robertknight:
There are some visual side effects to this PR that you might want to ask Conor about. 
Specifically, the site currently has custom styling for validation errors:

client-rendered-form

With this PR, we get the built-in browser UI for showing form validation errors:

deform-forms

robertknight:
Also, if you click 'Create an account' and then go back to the 'Sign in' tab, it switches back from the server-side rendered form to the client-side variant.

seanh:
Not part of this pr, but is the `bad_csrf_token()` function in `accounts/views.py` actually used?

seanh:
This all looks good to me apart from the issue that Robert pointed out that if you leave the fields empty you get browser validation.

The issue with `/register` linking back to the Angular version of `/login` might not be important, since I assume we're planning to replace `/register` with a server-side implementation as well.

nickstenning:
> Also, if you click 'Create an account' and then go back to the 'Sign in' tab, it switches back from the server-side rendered form to the client-side variant.

Yep, that's going to be addressed by #2582. In the meantime, it's harmless.

nickstenning:
I'm not sure what the issue is with having client-side validation done by the browser. It's always baffled me that we chose to implement our own custom client-side validation rather than using the controls and the feedback that the user is already familiar with from the browser.

But, as you say, it *is* a visual change (albeit one I intended) so paging @conordelahunty to ask if he has strong feelings one way or the other...

conordelahunty:
Well I obviously would prefer to be able to 1) at very least change the copy of the validation and 2) would also like to be able to control the styling but am I right in understanding that that goes against the work you just did?

seanh:
I think mixing browser and custom validation error messages is a problem, these are both from the new version of the form:

screenshot from 2015-10-08 11 49 22

screenshot from 2015-10-08 11 50 03

robertknight:
> I'm not sure what the issue is with having client-side validation done by the browser.
> It's always baffled me that we chose to implement our own custom client-side validation
> rather than using the controls and the feedback that the user is already familiar with from the browser.

I didn't say it was necessarily a _bad_ thing. I'm not fully aware of the trade-offs involved and that's something I/we need to learn about. I just wanted to flag it as a user visible change here.

conordelahunty:
> I think mixing browser and custom validation error messages is a problem

Yep. For sure.

nickstenning:
Okay, understood. My (admittedly biased) preference would be to merge this and schedule bringing back client-side form validation at a later date. The existing client-side validation is bound up in a crapload of other stuff we don't need and are looking to remove.

conordelahunty:
Okay, my understanding (design perspective) is that we are temporarily removing the ability for us to customise what we say to users when forms don't validate. Happy to merge this now as long as we look for a solution to how to this in a sustainable manner in the near future.

dwhly:
Can I ask are the validations the same?  Particularly around username length and password length on create.

nickstenning:
> Can I ask are the validations the same? Particularly around username length and password length on create.

Yes.

robertknight:
As far as I checked this yesterday, everything was working as expected given the discussion above.

PR #2582, opened 2015-10-05
Server-rendered accounts forms

nickstenning:
Replaces the client-rendered account lifecycle forms, namely:

- registration
- forgot password
- reset password

with forms rendered on the server with deform and validated with colander. The ajax endpoints for the associated forms are removed by this PR, and the associated client code for form handling is removed.

The server-side forms should be visually identical to their predecessors with a couple of small exceptions:

- Make the use of "Sentence case" (as opposed to "Initial Caps") consistent across form labels: so, "Email address" rather than "Email Address".
- Render the email form inputs as HTML5 `<input type="email">` fields. This allows mobile clients to customize the keyboard, and triggers automatic client-side validation in most modern browsers.
- The "reset password" form will now tell you explicitly if your reset code is invalid. In addition, you can visit "/reset_password" (without a code) in order to get a form where you can paste the code as sent in your email.
- The "forgot password" form now redirects you to the code-less version of the "reset password" form by default, rather than to the home page.
- The "forgot password" form no longer flashes you a "check your email" message. Instead, the "reset password" form will remind you to do so if you didn't provide a code.
- The client registration template had out-of-date information on the allowable length of usernames. The server-side form now uses the `USERNAME_MAX_LENGTH` and `USERNAME_MIN_LENGTH` constants so this is unlikely to happen again.


nickstenning:
Rebased and ready for review. We'll also need signoff from @conordelahunty on this one: perhaps next week we could sit together and review the changes?

conordelahunty:
Sure thing @nickstenning - let's do it first thing on Monday

nickstenning:
Ok, I've run @conordelahunty through these changes and he's happy with them, provided we get a card in the backlog to bring back support for validation where we have control over the appearance and copy of the validation messages in the near term.

seanh:
Reset code field disappearing:

untitled screencast - cropped 1

seanh:
Reset password form is pre-filled with email address and old password:

screenshot from 2015-10-12 14 47 57

seanh:
Create account form is pre-filled with previous email address and password (not sure if there's an easy way to stop browsers from doing this):

screenshot from 2015-10-12 14 49 27

seanh:
Reset password tab disappears:

untitled screencast - cropped 2

seanh:
The code looks good to me, so it's just the UX problems above

nickstenning:
Okay, so:

- disappearing reset code field should be fixed (it still disappears if it's correct, but that's fine, I think)
- I've added `autocomplete="off"` to the reset password form fields
- I don't think autocomplete on the registration form is a bug. That's what autocomplete is for!
- Reset password tab disappearing... I'm not sure what else this should do?

seanh:
I'd argue that the registration form should not autocomplete, you can't register with the same username or email twice and shouldn't be using the same password, so it doesn't seem useful to me? Not a big deal though.

nickstenning:
> I'd argue that the registration form should not autocomplete, you can't register with the same username or email twice and shouldn't be using the same password, so it doesn't seem useful to me? Not a big deal though.

AIUI autocomplete will usually prefill registration forms with the user's saved data (i.e. "here's my email address, browser, use it to fill in any registration form") if it exists in preference to using "whatever you filled in on this page last". Disabling autocomplete would disable that useful behaviour.

PR #2583, opened 2015-10-05
Extract the WebSocket client into its own module and add tests

robertknight:
In preparation for adding additional push notification types,
separate out the client into its own module and add tests.

T-105


nickstenning:
Needs a rebase. I'd be happy to do it myself but it's in your fork :)

nickstenning:
Otherwise this looks great!

PR #2585, opened 2015-10-06
Do not show vendored PDF.js in language stats

berkerpeksag:
GitHub also marks some PDF.js files as Java because it
looks for Emacs mode lines and some files have the
following mode line:

    /* -*- Mode: Java; tab-width: 2; indent-tabs-mode: nil; c-basic-offset: 2 -*- */

See h/h/browser/chrome/content/build/pdf.js for example.

I use the same method at python.org: https://github.com/python/pythondotorg/blob/master/.gitattributes

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,3 @@
 h/_version.py export-subst
+h/browser/chrome/content/build/* linguist-vendored
+h/browser/chrome/content/web/* linguist-vendored


robertknight:
I would suggest we instead give the `build/` directory a standard name for 3rd-party code (eg. `vendor/`) to make it more obvious that this is third-party code. linguist will then ignore code in such directories automatically.

robertknight:
Actually, it looks like everything in `browser/chrome/content` is PDF.js? If so then it would make sense to rename that entire directory. Thoughts @seanh ?

seanh:
I think you can rename the `content` directory, yes. No idea why it was called content, but sounds like we have a good reason to call it vendor, and looks like there's no reason you can't rename it (and all references to it)

nickstenning:
Regardless of how we decide to move code around in our repository, I didn't know it was even possible to do this, and it will be very nice to have GitHub not report our repository as Java! Thanks you very much @berkerpeksag!

PR #2587, opened 2015-10-06
WIP - Group membership change push notifications

robertknight:
This implements the client and server side parts of push notifications for group membership changes for a user.

This adds a new class of push notification, 'user-status-notification',
which is broadcast to logged-in clients when changes to the user's
state, including the list of groups that they are a member of,
changes.

----

- [ ] Figure out how to get the correct URL for the group
- [ ] Avoid duplicating the logic that serializes groups into JSON in session.py and streamer.py
- [ ] Write tests for `groups.{add,remove}` on the client
- [ ] Write tests for event queuing when a user joins a group on the server
- [ ] Write tests for push notifications of group join / leave events

----

When the user creates or joins a group, a notification is
added to the event queue with an associated user ID
which is then broadcast via the web socket.

On the client side, the streamer service listens for the new
class of notification and updates the groups part of the session
state in response, firing off 'group-joined' or 'group-left'
notifications which trigger a reload of the view.

When the user leaves a group, this may trigger an implicit change
of focus if the user was a member of the group.

### DIFF
--- a/h/api/events.py
+++ b/h/api/events.py
@@ -1,8 +1,33 @@
 # -*- coding: utf-8 -*-
-class AnnotationEvent(object):
+
+# Types representing events stored in the notification queue
+# (see queue.py)
+
+class Event(object):
+    def __init__(self, request):
+        self.request = request
+
+class AnnotationEvent(Event):
     """An event representing an action on an annotation."""
 
     def __init__(self, request, annotation, action):
-        self.request = request
+        Event.__init__(self, request)
         self.annotation = annotation
         self.action = action
+
+class UserStatusEvent(Event):
+    """An event representing a change in a user's status."""
+
+    def __init__(self, request, user_id, type, group_id = None):
+        assert type in ['group-joined', 'group-left']
+
+        Event.__init__(self, request)
+
+        self.user_id = user_id
+        """ The ID of the user affected by the event """
+
+        self.type = type
+        """ The type of action (eg. joined a group, left a group) """
+
+        self.group_id = group_id
+        """ Optional. For group membership events, the hash ID of the group """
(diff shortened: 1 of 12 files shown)


h/api/events.py
```diff
@@ -1,8 +1,33 @@
...
+class AnnotationEvent(Event):
     """An event representing an action on an annotation."""
 
     def __init__(self, request, annotation, action):
-        self.request = request
+        Event.__init__(self, request)
         self.annotation = annotation
         self.action = action
+
+class UserStatusEvent(Event):
```
nickstenning: Hopefully you'll be glad to hear that very little of this is actually needed. Sorry if I didn't get the message across clearly enough -- you asked in Slack and I tried to explain that you didn't need to follow the `AnnotationEvent` pattern.

So, I'd revert all changes to this file and `h/api/queue.py`, and just push a JSON blob onto the queue in the group view functions. I'll annotate those below.

h/groups/views.py
```diff
@@ -5,6 +5,7 @@
 from pyramid.view import view_config
 from pyramid import renderers
 
+from h.api.events import UserStatusEvent
```
nickstenning: No need for this at all.

h/groups/views.py
```diff
@@ -29,6 +30,14 @@ def create_form(request):
 
     return {'form': form, 'data': {}}
 
+def _send_group_join_notification(request, hashid):
```
nickstenning: You can simply change this function to do something like:

```python
queue = request.get_queue_writer()
data = {
    'user_id': request.authenticated_userid,
    'type': 'group-joined',
    'group_id': hashid
}
queue.publish('user', json.dumps(data))
```

robertknight:
Closing this PR and re-opening a new one from a branch in `hypothesis/h`

PR #2588, opened 2015-10-06
T88 - Add ability to leave groups

robertknight:
This adds a 'leave' button to groups which prompts to leave when clicked. The confirmation dialog is currently a default browser confirm dialog whose aesthetic quality will depend on the platform - I think we're going to want to re-do those in future.

The groups list does not update itself after removing yourself from a group, you have to refresh the page. Automatically updating the groups list when creating, joining or leaving a group will happen as part of https://github.com/hypothesis/h/pull/2587

t88-leave-group

### DIFF
--- a/h/groups/test/views_test.py
+++ b/h/groups/test/views_test.py
@@ -390,6 +390,21 @@ def test_join_redirects_to_group_page(Group):
     assert isinstance(result, httpexceptions.HTTPRedirection)
 
 
+leave_fixtures = pytest.mark.usefixtures('User', 'Group')
+
+
+@leave_fixtures
+def test_leave_removes_user_from_group_members(Group, User):
+    group = mock.Mock()
+    Group.get_by_hashid.return_value = group
+    User.get_by_userid.return_value = mock.sentinel.user
+
+    request = _mock_request(matchdict=_matchdict())
+    result = views.leave(request)
+
+    group.members.remove.assert_called_once_with(mock.sentinel.user)
+
+
 @pytest.fixture
 def Form(request):
     patcher = mock.patch('h.groups.views.deform.Form', autospec=True)
--- a/h/groups/views.py
+++ b/h/groups/views.py
@@ -28,6 +28,7 @@ def create_form(request):
 
     return {'form': form.render()}
 
+
 @view_config(route_name='group_create',
              request_method='POST',
              renderer='h:groups/templates/create.html.jinja2')
@@ -170,8 +171,32 @@ def _group_form(request):
     return form
 
 
+@view_config(route_name='group_leave',
+             request_method='POST')
+def leave(request):
+    if not request.feature('groups'):
+        raise exc.HTTPNotFound()
+    if request.authenticated_userid is None:
+        raise exc.HTTPNotFound()
+
+    hashid = request.matchdict["hashid"]
+    group = models.Group.get_by_hashid(hashid)
+
+    if group is None:
+        raise exc.HTTPNotFound()
+
+    user = models.User.get_by_userid(request.domain,
+                                     request.authenticated_userid)
+
+    group.members.remove(user)
+
+    return exc.HTTPNoContent()
+
+
 def includeme(config):
     config.add_route('group_create', '/groups/new')
+    config.add_route('group_leave', '/groups/{hashid}/leave')
+
     # Match "/groups/<hashid>/": we redirect to the version with the slug.
     config.add_route('group_read', '/groups/{hashid}/{slug:[^/]*}')
     config.add_route('group_read_noslug', '/groups/{hashid}')
--- a/h/static/scripts/directive/group-list.js
+++ b/h/static/scripts/directive/group-list.js
@@ -1,7 +1,7 @@
 'use strict';
 
 // @ngInject
-function GroupListController($scope) {
+function GroupListController($scope, $window) {
   $scope.expandedGroupId = undefined;
 
   // show the share link for the specified group or clear it if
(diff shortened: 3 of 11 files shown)


nickstenning:
This looks great -- fast work! But there's one niggle here.

I don't think we should let the creator of a group leave, for the time being. Creators of the group [are special](https://github.com/hypothesis/h/blob/c3eae4744ab1302b310c49aef02e6f47cb819b85/h/groups/models.py#L41-L42), and we're recording this information so that if (when?) we build out group administration features, we have someone to be the first administrator.

When a group creator leaves a group, that causes problems, because we need to either:

- know who the new administrator should be, or
- remove the group entirely -- all members and all annotations within that group

So for now, the simplest thing to do would be to not allow leaving a group if you are that group's creator.

dwhly:
> Creators of the group are special, and we're recording this information so that if (when?) we build out group administration features, we have someone to be the first administrator.

Reasonable point.  However, as someone who will create a bunch of them, I can guarantee right now I will want to be able to leave them.  

Question: If I click on the join link again, will I be able to?  I assume so, right?  Would it remember me as the creator?

If so, I'd argue that the administrator should get an email that they've created a group, and if they want to join again, they can click the email link again.  

Otherwise.... leave at your own peril?  (The peril of losing a feature we haven't announced yet).

nickstenning:
> Reasonable point. However, as someone who will create a bunch of them, I can guarantee right now I will want to be able to leave them.

Why? Because you're testing the groups feature? If that's the reason, then we can delete them manually.

dwhly:
> Why? Because you're testing the groups feature?

I think it was more because "As a user that will want to create a bunch of groups, including probably very temporary ones to do brief copy-editing on a document, I want to be able to delete them instead of have them accumulate in my pull down menu".  In fact, I'd go a step further and say that if I discover in the UI that I *can't* delete ones that I create, it would probably be a strong disincentive to me (as a normal user) to create groups for more trivial or fleeting uses, such as copy-editing.

I guess my preference is that people can remove them, and potential re-add themselves using the original link for now.  But that we not apriori prohibit removal.

IIRC our notion was always that this initial implementation was more or less Etherpad-for-annotation.  i.e. simple disposable groups via a link with absolutely no administration.  If you lose the link you lose the group (although we give them a nice list of ones they've saved in the dropdown at present).

@judell, thoughts?

conordelahunty:
Agree that creating a group and then being stuck with it is a bit of a pain. That being said, I think that not allowing the group creator to leave a group they created is probably the best solution for this v1 of Groups. We know we will need to build out some basic admin features soon (we definitely need them for classrooms) and as Nick pointed out, it could leave some nasty technical issues hanging around.

If we allow them to leave the group they created then what happens to any annotations created in the group? Do they stay in the group? I guess the same question could be asked of anyone leaving any group. Do your annotations stay or go? If they are the only member of the group will the annotations be deleted if they leave the group or does it stay around as a phantom group? How do we decide who the group admin is when we launch those features? The second person to join the group? 

Anyway, there are probably 100+ other questions like these once you start to pick at it. Saying that the owner can't leave a group they create for this v1 is probably correct.

dwhly:
> If we allow them to leave the group they created then what happens to any annotations created in the group? Do they stay in the group? I guess the same question could be asked of anyone leaving any group. 

Yes, exactly.  I *thought* the decision we made was that if I left the group, the group could still see my annotations because they were still tagged with the group's ID.  In other words, someone leaving a group doesn't change any annotations they will have made, it just changes their ability to create or view a group's annotations..  Of course they can always see their own annotations, regardless, even if they were made in a group.

> If they are the only member of the group will the annotations be deleted if they leave the group or does it stay around as a phantom group? 

My understanding is that if I click on the join group link again, assuming it's in an invite, or that I've saved it, I can join again.  And then I'd have access to the annotations again.

> How do we decide who the group admin is when we launch those features?

I think by default it'd be the person who created it.  If they've left, then tough luck till they join it again.  I don't have a problem with that.  For me, this initial go round we're not making any promises at all about administration.

Can I get a clarification about the way what we have now has been implemented?  Is my user ID as the person that created a group stored along w/ the group?  So, if I leave, and then I click an old Join link, is it like I never left?

nickstenning:
> Can I get a clarification about the way what we have now has been implemented? Is my user ID as the person that created a group stored along w/ the group? So, if I leave, and then I click an old Join link, is it like I never left?

Yep that's correct. We currently store the identity of the creator independently from the group members list.

h/groups/views.py
```diff
@@ -159,9 +159,31 @@ def join(request):
     url = request.route_url('group_read', hashid=group.hashid, slug=group.slug)
     return exc.HTTPSeeOther(url)
 
+@view_config(route_name='group_leave',
```
nickstenning: Nitpicky McNitPick here: two blank lines before a function definition in Python land.

h/groups/test/views_test.py
```diff
@@ -401,6 +401,18 @@ def test_join_redirects_to_group_page(Group):
 
     assert isinstance(result, httpexceptions.HTTPRedirection)
 
+leave_fixtures = pytest.mark.usefixtures('User', 'Group')
+
+@leave_fixtures
```
nickstenning: Ditto here with the two blank lines: both before and after this function.

nickstenning:
On that basis I'm going to merge this. We can decide later if we want to prevent group creators from leaving.

conordelahunty:
Just got in there before me!

Talked to Dan and Jon about this yesterday and I think creators should be able to leave the group. I think it's a language issue more than anything else. A creator leaving a group is a bit of an odd one as it could mean lot's of things - that they are trying to remove/delete the group, archive it, that it was only intended to be temporary, etc. So when they leave they should receive an email explaining that if they want to rejoin it/re-open it/unarchive it then they can do so by clicking on this link (which is just the group invitation link). We still store who created a group so down the line when we add admin rights for groups we can parcel all of these into a list of old/archived groups.

Anyway, I'll make a card for this in Trello.

PR #2591, opened 2015-10-07
Push notifications of group membership changes

robertknight:
Implement push notifications for when a user joins or leaves a group

This adds a new class of push notification, 'session-change'
which is broadcast to logged-in clients when changes to the user's
state, including the list of groups that they are a member of,
changes.

When the user creates or joins a group, a notification is
added to the event queue with an associated user ID
which is then broadcast via the web socket.

On the client side, the streamer service listens for the new
class of notification and triggers an update of the session
state in response.

When the user leaves a group, this may trigger an implicit change
of focus if the user was a member of the group.
Implement push notifications for when a user joins or leaves a group

This adds a new class of push notification, 'session-change'
which is broadcast to logged-in clients when changes to the user's
state, including the list of groups that they are a member of,
changes.

When the user creates or joins a group, a notification is
added to the event queue with an associated user ID
which is then broadcast via the web socket.

On the client side, the streamer service listens for the new
class of notification and triggers an update of the session
state in response.

When the user leaves a group, this may trigger an implicit change
of focus if the user was a member of the group.

----
Remaining Tasks:

- [X] Write tests for event queuing when a user joins a group on the server
- [X] Write tests for push notifications of group join / leave events

### DIFF
--- a/h/groups/test/views_test.py
+++ b/h/groups/test/views_test.py
@@ -155,6 +155,20 @@ def test_create_with_non_ascii_name():
     views.create(_mock_request(params={"name": u"☆ ßüper Gröup ☆"}))
 
 
+@create_fixtures
+def test_create_broadcasts_join_event(Group):
+    group = mock.Mock(hashid=mock.sentinel.hashid)
+    Group.return_value = group
+    request = _mock_request()
+
+    views.create(request)
+
+    request.get_queue_writer().publish.assert_called_once_with('user', {
+        'type': 'group-joined',
+        'userid': request.authenticated_userid,
+        'group': group.hashid,
+    })
+
 # The fixtures required to mock all of read()'s dependencies.
 read_fixtures = pytest.mark.usefixtures('Group', 'User', 'renderers')
 
@@ -389,6 +403,19 @@ def test_join_redirects_to_group_page(Group):
 
     assert isinstance(result, httpexceptions.HTTPRedirection)
 
+@join_fixtures
+def test_join_broadcasts_join_event(Group):
+    group = mock.Mock(hashid = mock.sentinel.hashid)
+    Group.get_by_hashid.return_value = group
+    request = _mock_request(matchdict=_matchdict())
+
+    views.join(request)
+
+    request.get_queue_writer().publish.assert_called_once_with('user', {
+        'type': 'group-joined',
+        'userid': request.authenticated_userid,
+        'group': mock.sentinel.hashid,
+    })
 
 leave_fixtures = pytest.mark.usefixtures('User', 'Group')
 
@@ -404,6 +431,20 @@ def test_leave_removes_user_from_group_members(Group, User):
 
     group.members.remove.assert_called_once_with(mock.sentinel.user)
 
+
+@leave_fixtures
+def test_leave_broadcasts_leave_event(Group, User):
+    group = mock.Mock(hashid = mock.sentinel.hashid)
+    Group.get_by_hashid.return_value = group
+    User.get_by_userid.return_value = mock.sentinel.user
+
+    request = _mock_request(matchdict=_matchdict())
+    views.leave(request)
+    request.get_queue_writer().publish.assert_called_once_with('user', {
+        'type': 'group-left',
+        'userid': request.authenticated_userid,
+        'group': mock.sentinel.hashid,
+    })
 
 @pytest.fixture
 def Form(request):
--- a/h/groups/test/views_test.py
+++ b/h/groups/test/views_test.py
@@ -156,7 +156,7 @@ def test_create_with_non_ascii_name():
 
 
 @create_fixtures
-def test_create_broadcasts_join_event(Group):
+def test_create_publishes_join_event(Group):
     group = mock.Mock(hashid=mock.sentinel.hashid)
     Group.return_value = group
     request = _mock_request()
@@ -164,7 +164,7 @@ def test_create_broadcasts_join_event(Group):
     views.create(request)
 
     request.get_queue_writer().publish.assert_called_once_with('user', {
-        'type': 'group-joined',
+        'type': 'group-join',
         'userid': request.authenticated_userid,
         'group': group.hashid,
     })
@@ -404,7 +404,7 @@ def test_join_redirects_to_group_page(Group):
     assert isinstance(result, httpexceptions.HTTPRedirection)
 
 @join_fixtures
-def test_join_broadcasts_join_event(Group):
+def test_join_publishes_join_event(Group):
     group = mock.Mock(hashid = mock.sentinel.hashid)
     Group.get_by_hashid.return_value = group
     request = _mock_request(matchdict=_matchdict())
@@ -412,7 +412,7 @@ def test_join_broadcasts_join_event(Group):
     views.join(request)
 
     request.get_queue_writer().publish.assert_called_once_with('user', {
-        'type': 'group-joined',
+        'type': 'group-join',
         'userid': request.authenticated_userid,
         'group': mock.sentinel.hashid,
     })
@@ -447,7 +447,7 @@ def test_leave_returns_not_found_if_user_not_in_group(Group, User):
 
 
 @leave_fixtures
-def test_leave_broadcasts_leave_event(Group, User):
+def test_leave_publishes_leave_event(Group, User):
     group = mock.Mock(hashid=mock.sentinel.hashid,
                       members=[mock.sentinel.user])
     Group.get_by_hashid.return_value = group
@@ -456,7 +456,7 @@ def test_leave_broadcasts_leave_event(Group, User):
     request = _mock_request(matchdict=_matchdict())
     views.leave(request)
     request.get_queue_writer().publish.assert_called_once_with('user', {
-        'type': 'group-left',
+        'type': 'group-leave',
         'userid': request.authenticated_userid,
         'group': mock.sentinel.hashid,
     })
--- a/h/groups/test/views_test.py
+++ b/h/groups/test/views_test.py
@@ -422,19 +422,34 @@ leave_fixtures = pytest.mark.usefixtures('User', 'Group')
 
 @leave_fixtures
 def test_leave_removes_user_from_group_members(Group, User):
+    user = mock.sentinel.user
     group = mock.Mock()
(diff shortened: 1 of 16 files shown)


h/streamer.py
```diff
@@ -25,7 +26,8 @@
 from h.api import uri
 from h.api.search import query
 from .models import Annotation
-
+from h.groups.models import Group
```
nickstenning: `from h.models import Group`

h/streamer.py
```diff
@@ -439,6 +441,8 @@ def match(self, target, action=None):
 class WebSocket(_WebSocket):
     # Class attributes
     event_queue = None
+    """ Tuples of (topic, message) pulled from the message queue """
```
nickstenning: This should probably just be a line comment.

h/streamer.py
```diff
@@ -464,15 +468,23 @@ def start_reader(cls, request):
...
         cls.event_queue = gevent.queue.Queue()
         reader_id = 'stream-{}#ephemeral'.format(_random_id())
-        reader = request.get_queue_reader('annotations', reader_id)
-        reader.on_message.connect(cls.on_queue_message)
-        reader.start(block=False)
+
+        for topic in ['annotations', 'user']:
+            reader = request.get_queue_reader(topic, reader_id)
+            reader.on_message.connect(
+                receiver=partial(cls.on_queue_message, topic),
```
nickstenning: You don't need to do this. The `gnsq.Reader` instance is already passed into the message handler, and it has a `topic` attribute.

h/groups/test/views_test.py
```diff
@@ -155,6 +155,20 @@ def test_create_with_non_ascii_name():
     views.create(_mock_request(params={"name": u"☆ ßüper Gröup ☆"}))
 
 
+@create_fixtures
+def test_create_broadcasts_join_event(Group):
```
nickstenning: Minor niggle, but `broadcasts` is a bit misleading here, I think. `publishes` might be a bit clearer. I'm guessing this is a hangover from when it was in `h.api`.

h/groups/test/views_test.py
```diff
@@ -155,6 +155,20 @@ def test_create_with_non_ascii_name():
...
+    group = mock.Mock(hashid=mock.sentinel.hashid)
+    Group.return_value = group
+    request = _mock_request()
+
+    views.create(request)
+
+    request.get_queue_writer().publish.assert_called_once_with('user', {
+        'type': 'group-joined',
```
nickstenning: For the sake of simplicity (and consistency with route names, method names, etc.), perhaps this event type should be an imperative rather than a past-participle: `group-join`, `group-leave`.

h/queue.py
```diff
@@ -11,6 +13,8 @@ def __init__(self, namespace, *args, **kwargs):
     def publish(self, topic, data):
         if self.namespace is not None:
             topic = '{0}-{1}'.format(self.namespace, topic)
+        if not isinstance(data, str):
```
nickstenning: I'm not sure about this change. Either the `h.queue` module should be responsible for serialisation and deserialisation, or it shouldn't. With this change it's asymmetrical, with the library responsible for serialisation but subscribers responsible for deserialisation.

h/static/scripts/groups.js
```diff
@@ -45,41 +51,63 @@ function groups(localStorage, session, $rootScope, features, $http) {
...
+     focusedGroup = g;
+     localStorage.setItem(STORAGE_KEY, g.id);
+     $rootScope.$broadcast(events.GROUP_FOCUSED, g.id);
+   }
+  }
+
+  // reset the focused group if the user leaves it
+  $rootScope.$on(events.SESSION_CHANGED, function () {
+    if (focusedGroup) {
+      var match = session.state.groups.filter(function (group) {
```
nickstenning: I think you could probably simplify this to:

```javascript
if (focusedGroup) {
  var match = get(focusedGroup.id);
  if (!match) {
    ...
  }
}
```

h/static/scripts/session.js
```diff
@@ -59,7 +61,7 @@ function sessionActions(options) {
  *
  * @ngInject
  */
-function session($document, $http, $resource, flash) {
+function session($document, $http, $resource, flash, $rootScope) {
```
nickstenning: I try and keep these ridiculous angular parameter lists alphabetised, so `$resource, $rootScope, flash`.

h/static/scripts/streamer.js
```diff
@@ -20,7 +20,7 @@ var socket;
  * @return An angular-websocket wrapper around the socket.
  */
 // @ngInject
-function connect($websocket, annotationMapper, groups) {
+function connect($websocket, annotationMapper, groups, session) {
```
nickstenning: Docstring needs an update.

h/static/scripts/streamer.js
```diff
@@ -69,6 +64,26 @@ function connect($websocket, annotationMapper, groups) {
...
     }
+  }
+
+  function handleSessionChangeNotification(message) {
+    session.update(message.model);
+  }
+
+  // Listen for updates
+  socket.onMessage(function (event) {
+    message = JSON.parse(event.data)
```
nickstenning: This should probably be wrapped in a `try {  } catch { }`. I see it wasn't before, but invalid JSON over the websocket shouldn't explode the frontend. Happy to see this in a future PR instead of here.

h/static/scripts/test/app-controller-test.coffee
```diff
@@ -137,7 +138,11 @@ describe 'AppController', ->
     createController()
     assert.isFalse($scope.shareDialog.visible)
 
-  it 'calls $route.reload() when the focused group changes', ->
+  it 'calls $route.reload() when the session state changes', ->
```
nickstenning: This would be clearer as two different tests, without the `forEach`, I think.

h/static/scripts/test/session-test.js
```diff
@@ -30,9 +34,10 @@ describe('h:session', function () {
   }));
 
 
-  beforeEach(mock.inject(function (_$httpBackend_, _session_) {
+  beforeEach(mock.inject(function (_$httpBackend_, _session_, _$rootScope_) {
```
nickstenning: I know it's pernickety, but same again with the alphabetisation of parameters.

h/static/scripts/app-controller.coffee
```diff
@@ -38,9 +40,14 @@ module.exports = class AppController
...
 
-    $scope.$on('groupFocused', (event) ->
-      $route.reload()
-    )
+    # Reload the view when the focused group changes or the
+    # list of groups that the user is a member of changes
+    groupChangeEvents = [events.SESSION_CHANGED, events.GROUP_FOCUSED];
+    groupChangeEvents.forEach((eventName) ->
+      $scope.$on(eventName, (event) ->
+        $route.reload()
```
nickstenning: So the `SESSION_CHANGED` event will fire on login/logout, which means I think we are probably calling `$route.reload()` twice on login/logout now. Or, if we're not, it's only because of a race... You might be able to remove [some if not all of this `$watch`?](https://github.com/hypothesis/h/pull/2591/files#diff-e394c98a9a7311a619855172ca940fa6R79).

nickstenning:
Overall this is looking really good. I've left a bunch of line comments to address, and you should take a quick [gander over Landscape's feedback on the Python](https://landscape.io/diff/222024) because some of that looks like signal.

h/queue.py
robertknight: If we're using JSON for message bodies everywhere then I think it would be easier to do both serialization and de-serialization in the queue. The motivation was to make it easier to mock `publish()` in testing, since 'data' can now be an dict where the values may be `mock.Mock` instances - if the caller has to do de-(serialization) then it runs into problems trying to JSON serialize dicts containing `mock.Mock` instances in tests.

h/static/scripts/session.js
robertknight: Makes sense.

h/static/scripts/app-controller.coffee
robertknight: Yes indeed, it looks like that is the case.

robertknight:
I'm currently investigating a few issues I can see in testing:

* [ ] The WebSocket appears to get torn down and re-opened on the client whenever the `route.reload()` is called. Since this happens in response to 'group focused' and 'session change' events, it means that the WS connection is closed and re-opened every time the user switches groups or joins/leaves a group
* [ ] Annotation push notifications appear to be broken in this branch despite the tests passing

h/queue.py
robertknight: I started looking into adding JSON deserialization to the reader yesterday. I think it makes sense but is a large enough change that it would be better in a separate PR

h/static/scripts/test/app-controller-test.coffee
robertknight: Done

robertknight:
The aforementioned issues have been resolved. I think this is good to go, with some follow up work in separate PRs to:

* Handle exceptions in the WS message processing loop on the server
* Avoid reloading the whole view when the groups list or focused group changes
* Make the queue reader/writer handle JSON (de-)serialization symmetrically

nickstenning:
LGTM. I think I'd rather  `isInitialLoad` were an internal latch in the session service, rather than being clever with the CSRF token, but I'll merge this anyway.

tilgovi:
If I read this right you removed the websocket reconnection on session updates. Are you sure you don't need to reconnect the websocket if the group membership or other session information changes? The server-side request object associated with the websocket is used for authorization tests. I guess group membership changes might work fine, assuming `effective_principals` isn't cached, but it makes me nervous.

robertknight:
@tilgovi - The websocket does reconnect on session updates. What happens now is that all session updates - eg. via `session.logout()` will result in a `SESSION_CHANGED` event being emitted when the session state is updated and that is handled by reloading the view, which resets the WS connection.

tilgovi:
Most excellent. Sorry for the noise.

robertknight:
What I'd really like is some better logging infrastructure to make it possible to trace these events when they happen in the app and verify this kind of thing.

Events are good for decoupling parts of the app but I must admit I had to re-read a few files to check that everything was as expected.

PR #2592, opened 2015-10-07
Fix app initialization race when issuing annotation search query

### LINKED ISSUE #2590: Race condition in app initialization
The current master branch has a race condition in initialization of the app. When the main document's frame connects to the sidebar, `WidgetController.loadAnnotations()` is invoked which tries to get the current focused group in order to construct a search query.

However, if the session state API call to `/app` has not yet completed, the groups service's list of groups is empty and there is no focused group so `groups.focused()` returns null resulting in an exception when constructing the search query.

The actual search API call needs to be authenticated so there is no point constructing or dispatching the query until auth completes and the app state has been retrieved. The 'dispatching' part of this is already deferred until auth completes because of the JWT interceptor that is setup on the $http service - deferring all API requests until auth has completed. However the construction of the query in `WidgetController._loadAnnotationsFrom()` should also be deferred until the app state is ready.

Although it is an error in this case to try and use the focused group before the groups list has been loaded, it would be worth considering making groups.focused() non-null by hardcoding the existence of the Public group in the client. In other words, setting the initial state of the session in `session.js` to `{ groups: [<public group>] }` rather than `{}`. However, this would mean that if we add any additional attributes to the public group in future, we would also need to update the object in the client.

robertknight:
If the initial annotation search query was constructed
before the user's session state had been retrieved,
the groups list was empty and groups.focused() returned null
in WidgetController._loadAnnotationsFrom

This commit resolves the issue by adding a 'groups.ready()'
function which returns a promise once the groups list is ready.

It currently just returns the session.load().$promise.

Fixes #2590

### DIFF
--- a/h/static/scripts/groups.js
+++ b/h/static/scripts/groups.js
@@ -31,7 +31,15 @@ function groups(localStorage, session, $rootScope, features) {
     }
   };
 
+  /** Returns a promise which resolves once the list of groups
+   * has been loaded.
+   */
+  function ready() {
+    return session.load().$promise;
+  }
+
   return {
+    ready: ready,
     all: all,
     get: get,
 
--- a/h/static/scripts/test/widget-controller-test.coffee
+++ b/h/static/scripts/test/widget-controller-test.coffee
@@ -1,5 +1,7 @@
 {module, inject} = angular.mock
 
+Promise = require('core-js/library/es6/promise')
+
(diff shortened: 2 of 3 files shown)


robertknight:
Talking with nick we decided on an alternative approach which is to make the main content view in the sidebar dependent on the session state being ready before it loads.

PR #2593, opened 2015-10-07
Document the search API's _id parameter

seanh:


seanh:
Alternatively we could add `id` to the Es source documents

robertknight:
Does the underscore denote 'private' fields in the index or is there another reason for that?

seanh:
It's because it's the document ID generated for us by Elasticsearch, as opposed to a field that we passed in as part of the source data, and Es underscores all its own generated fields

nickstenning:
I don't think we should document this. It's "accidental" in the sense that it's a fall-through match query. We already have an API endpoint for retrieving annotations by ID, and "[t]here should be one-- and preferably only one --obvious way to do it."

seanh:
Alright, but the docs do say that you can search by any field that the API returns and this confused me, so I've changed this to document that the `id` field doesn't work instead.

nickstenning:
Great!

tilgovi:
Why doesn't `id` work?
https://github.com/openannotation/annotator-store/blob/master/annotator/elasticsearch.py#L111

It should be searchable, even if it's not in `_source`.

PR #2594, opened 2015-10-07
Fix race for session data on startup

### LINKED ISSUE #2590: Race condition in app initialization

nickstenning:
Several parts of the application require access to the session data fetched from the server when rendering. On first page load we don't have this data (something we can fix in the future by inlining the data with the viewer HTML) and this resulted in a race between the session data load triggered by `checkAuthentication` (in config/identity.js) and the query sent to the server by `WidgetController`, which needs to know the current focused group id.

Fix this by ensuring that the first load of session data is complete before switching views -- achieved by adding an item to the list of dependencies passed to `$routeProvider`.

This works, but results in multiple requests for session data in quick succession, because `checkAuthentication` runs while the view is loading. To resolve this problem, we add a straightforward TTL check to the `session.load()` method, and only make a new request if the data is more than 5m old.

Session data will be overwritten as normal whenever session mutating methods (`session.login()`, `session.logout()`) etc. are called.

- [x] needs tests for the changes to the session service.

Fixes #2590.

### DIFF
--- a/h/static/scripts/app.coffee
+++ b/h/static/scripts/app.coffee
@@ -7,6 +7,9 @@ require('angular-jwt')
 streamer = require('./streamer')
 
 resolve =
+  # Ensure that we have available a) the current authenticated userid, and b)
+  # the list of user groups.
+  sessionState: ['session', (session) -> session.load().$promise]
   store: ['store', (store) -> store.$promise]
   streamer: streamer.connect
   threading: [
--- a/h/static/scripts/session.js
+++ b/h/static/scripts/session.js
@@ -1,6 +1,9 @@
 'use strict';
 
+var Promise = require('core-js/library/es6/promise');
 var angular = require('angular');
+
+var CACHE_TTL = 5 * 60 * 1000; // 5 minutes
 
 var ACCOUNT_ACTIONS = [
   ['login', 'POST'],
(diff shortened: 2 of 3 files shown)


nickstenning:
Okay, tests added.

PR #2595, opened 2015-10-07
Insert 'group': '__world__" if none

### LINKED ISSUE #2589: Incorrect results returned from search query when logged-in
blog.jonudell.net has 5 annotations with these group properties:

            "group": "__world__",
            "group": "__world__",
            "group": "__world__",
            "group": "MEpy2d",
            "group": "__world__",

I'm logged into 0.7.5.70 (g562d537.dirty), it reports 5 as I reckon it should because I'm a member of MEpy2d.

Badge query: GET /api/search?limit=0&uri=http://blog.jonudell.net/ 
Result: {"rows": [], "total": 5}

But no annotations anchor because the extension makes this query:

https://stage.hypothes.is/api/search?group=__world__&limit=200&offset=0&order=asc&sort=created&uri=http:%2F%2Fblog.jonudell.net%2F

And gets no results: {"rows": [], "total": 0}

seanh:
When creating or updating an annotation if the annotation has no 'group'
field then insert `'group': '__world__'`.

This fixes a bug where people calling our API directly can create
annotations with no group field, then for example when the sidebar
searches for annotations with `'group': '__world__'` it won't find these
annotations, but the Chrome extension badge (which doesn't put any group
filter in its searches) will find them.

This could be fixed client-side but it's better to make the annotations
in our search index consistent, than to try and always remember that
public annotations may either have `group: '__world__'` or no group at
all.

This means that it's no longer necessary for us to insert
`group: '__world__'` on the way out in render().

Annotations pre-dating this commit with no group can be migrated with:

    hypothesis annotool conf/production.ini prepare

Fixes #2589.

### DIFF
--- a/h/api/groups/__init__.py
+++ b/h/api/groups/__init__.py
@@ -3,9 +3,11 @@
 from h.api.groups.auth import group_principals
 from h.api.groups.auth import set_permissions
 from h.api.groups.logic import set_group_if_reply
+from h.api.groups.logic import insert_group_if_none
 
 __all__ = (
     'group_principals',
     'set_permissions',
     'set_group_if_reply',
+    'insert_group_if_none'
 )
--- a/h/api/groups/logic.py
+++ b/h/api/groups/logic.py
@@ -37,3 +37,8 @@ def set_group_if_reply(annotation):
     else:
         if 'group' in annotation:
             del annotation['group']
+
+
+def insert_group_if_none(annotation):
+    if not annotation.get('group'):
+        annotation['group'] = '__world__'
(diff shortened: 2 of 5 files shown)


nickstenning:
I guess you've noticed that the tests are broken. If it helps, I'd suggest a separate PR to separate the code that remains in transform into independently tested functions, followed by this one.

seanh:
Ah, no transform is already using separate functions (as far as this pr goes anyway), the problem was just that I removed a line from render, should be fixed now

nickstenning:
LGTM.

tilgovi:
Ignore my comment on the other PR. I see that this ensures that the other part of the code sees only group world and never empty group.

PR #2596, opened 2015-10-07
Inline the H icon font

robertknight:
The previous fix to use an absolute URL to the icon
font did not work in the Chrome extension because
the base path for assets was different.

The proper solution here will involve refactoring
the CSS build pipeline and possibly automatically
inlining font URLs in the process. I did not just
add a preprocessor to inline all URLs because
that would also have affected fonts in katex.min.css
and added significantly to the size of the CSS bundle.

This commit reverts back to the previous solution of
inlining just the icon font.

See #2571


h/static/styles/vendor/icomoon.css
```diff
@@ -1,13 +1,13 @@
...
...MkHFAtLEw5IQMCQXtuYCcNDzMqGjAVYUgOGw8KFAkUa0U2hUoNGQwjTFFWLJHfmE4BHzUVAv4NEQMTPCUSGgceJCE5TCwMGAwDIjlOMBg1HTdcHAEOCwJNcw4EBAICPlICKjACARYjGA1trNVoBg4HFjcgAAAAAAEAAAABgACj/yc1Xw889QALBAAAAAAA0jFYbAAAAADSMVhsAAD/qwQAA8AAAAAIAAIAAAAAAAAAAQAAA8D/wAAABAAAAAAABAAAAQAAAAAAAAAAAAAAAAAAADIEAAAAAAAAAAAAAAACAAAABAAA1QQAAJIEAACrBAAAgAQAAGQEAAArBAAAKwQAAVUEAADVBAAA1QQAAFUEAAAABAAA1QQAAIAEAAAABAABKwQAAQAEAABrBAAAVQQAANUEAAEABAAAVQQAAIAEAABVBAABbgQAAAAEAABVBAAAkgQAANUEAAGrBAAAJgQAAAgEAAAIBAAAAwQAAFUEAACABAAA1QQAANUEAACABAAAgAQAAFUEAAErBAAAVQQAAAAEAAAABAAAAAAAAAAACgAUAB4AOABKAJYA1AFgAa4CHAIwAkoCZAKYAtQC6gMIAzIDagOCA9IEFAQwBEwEkgS6BOwE/gVOBYYFmAWyBb4GGgamB6AIbgjICRYJOglKCW4JkgnYCeYKEAqgCtQLPAAAAAEAAAAyAK8ABgAAAAAAAgAAAAAAAAAAAAAAAAAAAAAAAAAOAK4AAQAAAAAAAQABAAAAAQAAAAAAAgAHACoAAQAAAAAAAwABACQAAQAAAAAABAABAD8AAQAAAAAABQALAAMAAQAAAAAABgABACcAAQAAAAAACgAaAEIAAwABBAkAAQACAAEAAwABBAkAAgAOADEAAwABBAkAAwACACUAAwABBAkABAACAEAAAwABBAkABQAWAA4AAwABBAkABgACACgAAwABBAkACgA0AFxoAGhWZXJzaW9uIDEuNQBWAGUAcgBzAGkAbwBuACAAMQAuADVoAGhoAGhSZWd1bGFyAFIAZQBnAHUAbABhAHJoAGhGb250IGdlbmVyYXRlZCBieSBJY29Nb29uLgBGAG8AbgB0ACAAZwBlAG4AZQByAGEAdABlAGQAIABiAHkAIABJAGMAbwBNAG8AbwBuAC4AAAADAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA')
```
tilgovi: https://github.com/borodean/postcss-assets/tree/3168ab2b07cf537240ecc20e0eb7688434987614#inlining-files might make this easier

robertknight: @tilgovi Thanks - I looked at both postcss-assets and postcss-url for this and I think we'll end up using one or the other in future.

PR #2597, opened 2015-10-07
Replace %20 with space when displaying filenames etc

seanh:


seanh:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;9sqXWtNi&#x2F;98-bug-file-href-and-empty-for-domain-on-some-pdf-annotations"> Bug: file:&#x2F;&#x2F; href and empty () for domain on some PDF annotations</a>

nickstenning:
This seems a bit off to me. Couple of questions:

- why are we doing this in the title? are there cases where the title is URI encoded too? if so, we should probably fix that at source rather than here.
- why are we doing this with the URI? do we display the raw URI? If not, I don't think there's any need to apply this transformation.
- for the "domain"... shouldn't we at least be using `decodeURIComponent` rather than just replacing one character?

seanh:
Okay, force-pushed this using `decodeURIComponent()` and not decoding the title. Also instead of decoding the entire URI, just decode the filename or domain right before it's returned.

nickstenning:
Nice! :symbols:

PR #2600, opened 2015-10-08
Remove special-handling of pre-groups annotations

seanh:
Now that we always insert 'group': '__world__' server side on annotation
create or update if the annotation has no group field, and we've run a
migration on staging/prod, we no longer need special-handling for
pre-groups annotations that have no groups field.


nickstenning:
LGTM.

tilgovi:
Wait, but what if I submit an annotation that doesn't have a group now?

PR #2601, opened 2015-10-09
Enable privacy dropdown even when saving an annotation is disabled

robertknight:
* Implement design change to allow a user to change the privacy
   setting for an annotation even if the annotation cannot
   yet be published because no tags or text have been entered.

 * Disable the hover state for the main section of the button
   when it is disabled.

 * Improve consistency between the 'Post to' button style and the
   buttons on the groups pages. This fixes the vertical alignment
   of the 'Post to' button label in the process.

T-125


nickstenning:
LGTM. :cake:

PR #2602, opened 2015-10-09
Remove the 'search_normalized' feature flag

nickstenning:
This is now deployed and toggled on everywhere. Removing it simplifies the relevant query code and reduces the number of possible configuration states of our application.


seanh:
lgtm but needs merge conflicts resolved

PR #2603, opened 2015-10-09
Remove the 'queue' feature flag

nickstenning:
This is also deployed and toggled on everywhere.

The fact that this data is relayed through a pyramid event (`h.api.events.AnnotationEvent`) already provides a suitable decoupling between the our web application's use of NSQ, and the `h.api` package's
interface. By moving the code that actually publishes to NSQ into `h.subscribers` we remove one point of dependency between `h.api` and NSQ (the `h.api.nipsa` module still needs to be addressed).


seanh:
lgtm but needs merge conflicts resolved

PR #2604, opened 2015-10-09
Remove the 'notification' feature flag

nickstenning:
This is deployed and toggled on everywhere, so we can simplify our code by removing this feature flag.

### DIFF
--- a/h/accounts/views.py
+++ b/h/accounts/views.py
@@ -547,7 +547,6 @@ class ProfileController(object):
         model = {}
         if userid:
             model["email"] = User.get_by_userid(request.domain, userid).email
-        if request.feature('notification'):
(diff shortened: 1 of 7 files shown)


h/accounts/views.py
```diff
@@ -482,7 +482,6 @@ def profile(self):
         model = {}
         if userid:
             model["email"] = User.get_by_userid(request.domain, userid).email
-        if request.feature('notification'):
```
nickstenning: This isn't strictly a no-op change, but the original code looks wrong to me. There's an `if userid:` check two lines earlier which the subscriptions lookup blithely ignores.

seanh:
`from pyramid import httpexceptions as exc` no longer needed in `notification/views.py`. Otherwise looks good

PR #2606, opened 2015-10-09
Let the annotation post button resize to fit its content

robertknight:
Remove the fixed width from the annotation post button
and let the dropdown menu's width adjust to fit its
content rather than being the same width as the button.

To avoid the menu disappearing into the left edge of
the sidebar if "Only Me" is selected and the alternative is a long group name,
the menu has been re-aligned so that it is centered underneath the button's dropdown arrow.

 * Simplify the styling used to hide the dropdown menu
   to use just the visibility property instead of
   setting pointer events to none, applying a rotation transform
   _and_ setting opacity to 0.

   I'm unclear why the combination of a transform, pointer events
   and opacity was used originally but the dropdown menus
   continue to work with the simplified styling.

post-button-width5
post-button-width6

T-112

### DIFF
--- a/h/static/styles/app.scss
+++ b/h/static/styles/app.scss
@@ -153,6 +153,12 @@ $group-list-width: 270px;
   }
 }
 
+// the name of a group in the groups drop-down list
+// and 'Post to <Group>' button for saving annotations
+.group-name-link {
+  white-space: nowrap;
+}
+
 /* The user account dropdown menu */
 .user-picker {
   .avatar {
--- a/h/static/styles/common.scss
+++ b/h/static/styles/common.scss
@@ -164,14 +164,12 @@ html {
 }
 
 .dropdown-menu {
-  @include rotateX(90deg);
+  visibility: hidden;
   background: $white;
   border: solid 1px $gray-lighter;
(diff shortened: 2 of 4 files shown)


dwhly:
What happens on long group names when the sidebar width is reduced by half?

robertknight:
> What happens on long group names when the sidebar width is reduced by half?

How do you do that? (_I'm sure I could find out looking through the code, but looking purely as a user isn't obvious_)

dwhly:
Grab the tab at upper right that opens the sidebar, click it to open, and then grab it and move right/left to resize the panel.

robertknight:
When the sidebar is at the minimal width, the menu flows off the right edge of the screen but is still usable. The bigger issue at the moment is the top-bar.

post-button-width-narrow2

tilgovi:
https://github.com/hypothesis/h/pull/2509#issuecomment-142034157
https://github.com/hypothesis/h/pull/2543#issuecomment-143028577

I also raised this issue when we were reviewing the prototype, but I was largely ignored.

nickstenning:
LGTM.

PR #2608, opened 2015-10-10
UI tweaks for the group creation form

robertknight:
* Increase the spacing between the label and the group name

 * Disable focus outline for the group name field, since
   the input has no border

 * Turn off autocomplete for the group name input field

### DIFF
--- a/h/groups/schemas.py
+++ b/h/groups/schemas.py
@@ -25,6 +25,7 @@ class GroupSchema(CSRFSchema):
         widget=deform.widget.TextInputWidget(
             autofocus=True,
             css_class="group-form__name-input js-group-name-input",
+            disable_autocomplete=True,
             label_css_class="group-form__name-label",
             max_length=GROUP_NAME_MAX_LENGTH,
             placeholder=_("Group Name")))
--- a/h/static/styles/group-form.scss
+++ b/h/static/styles/group-form.scss
@@ -53,6 +53,7 @@
 
     &__name-label {
       font-size: 18px;
+      margin-bottom: 10px;
     }
 
     &__name-input {
@@ -62,6 +63,7 @@
       text-align: center;
       margin-bottom: 40px;
       width: 350px;
+      outline: none;
 
       // TODO - Cyan cursor color
 


dwhly:
> Turn off autocomplete for the group name input field

For someone creating multiple groups with only slight variations in name autocomplete could be handy.   e.g. English 224 Section 1, English 224 Section 2 ...

robertknight:
> For someone creating multiple groups with only slight variations in
> name autocomplete could be handy. e.g. English 224 Section 1, English 224 Section 2 ... 

It felt a little odd to me to see a prompt to create another group with the same name and Jon echoed similar comments in his recent narrative.

I don't have particularly strong views on this though. I'll run it by conor on Monday and let him make a call.

tilgovi:
@dwhly that's not an application feature, it's a browser feature. If your browser is set to autocomplete form inputs and this is the same input field then the browser should handle that. A bonus is that the browser handles this in a way that the user can control in their settings.

dwhly:
> It felt a little odd to me to see a prompt to create another group with the same name and Jon echoed similar comments in his recent narrative.

That's fine.  Whatever you guys think best.

> @dwhly that's not an application feature, it's a browser feature.

@tilgovi: He's got it as a checklist item remaining to be done, so whatever he thought he was going to do is I guess what I'm remarking on.

tilgovi:
Oh, I see. Turn *off* auto-complete.

nickstenning:
Yep, I'd like a @judell call on this one.

h/groups/schemas.py
```diff
@@ -23,7 +23,7 @@ class GroupSchema(CSRFSchema):
             min=GROUP_NAME_MIN_LENGTH,
             max=GROUP_NAME_MAX_LENGTH),
         widget=deform.widget.TextInputWidget(
-            autofocus=True,
```
nickstenning: I'm a bit confused by this. You didn't disable the focus outline, you just disabled autofocus...

robertknight: Indeed - WTF moment. That's a mistake. The change to disable the outline is in the SCSS. I'll fix and re-test.

robertknight: Fixed

nickstenning:
Okay, well, I'm going to merge this, but @judell if you decide you would like autocomplete on on the group name field, please shout here or somewhere.

judell:
I much prefer autocomplete off in this case so thanks for this.

PR #2609, opened 2015-10-11
Convert Chrome extension to Browserify + CommonJS modules

robertknight:
This replaces the ad-hoc module system in the Chrome extension based around a `window.h` global with CommonJS + Browserify.

Rationale:

* Consistency with the main app
* Clearer dependency structure
* Simplifies the `manifest.json` and Karma config files
* A step towards sharing code between the Firefox, Chrome and Safari extensions in future

Most of the tests didn't need that much alteration because they used function arguments for dependency injection. The one test that did stub out other modules now uses proxyquire.

I suggest looking at the [diff ignoring whitespace change](https://github.com/hypothesis/h/pull/2609/files?w=1) (add '?w=1 to URL') for a clearer view of the changes.

### DIFF
--- a/h/browser/chrome/lib/tab-state.js
+++ b/h/browser/chrome/lib/tab-state.js
@@ -1,96 +1,94 @@
-(function (h) {
-  'use strict';
+'use strict';
 
-  var states = {
-    ACTIVE:   'active',
-    INACTIVE: 'inactive',
-    ERRORED:  'errored',
+var states = {
+  ACTIVE:   'active',
+  INACTIVE: 'inactive',
+  ERRORED:  'errored',
+};
+
+/* Manages the state of the browser action button to ensure that it displays
+ * correctly for the currently active tab. An onchange callback will be
+ * called when the extension changes state. This will be provided
+ * with the tabId, the current and previous states.
+ *
+ * Each state has a method to enable it such as activateTab() and a method
+ * to query the current state such as isTabActive().
+ *
+ * initialState - An Object of tabId/state keys. Used when loading state
+ *   from a persisted store such as localStorage.
+ * onchange     - A function that recieves onchange(tabId, current, prev).
+ */
+function TabState(initialState, onchange) {
+  var _this = this;
+  var currentState;
+  var previousState;
+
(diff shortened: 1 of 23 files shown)


tilgovi:
\o/

h/browser/chrome/lib/tab-state.js
```diff
@@ -1,96 +1,94 @@
...
-    ACTIVE:   'active',
-    INACTIVE: 'inactive',
-    ERRORED:  'errored',
-  };
-
-  /* Manages the state of the browser action button to ensure that it displays
-   * correctly for the currently active tab. An onchange callback will be
-   * called when the extension changes state. This will be provided
-   * with the tabId, the current and previous states.
+//'use strict';
```
nickstenning: Any reason this got commented out?

h/buildext.py
```diff
@@ -179,10 +180,12 @@ def build_chrome(args):
     copytree('h/browser/chrome/content', 'build/chrome/content')
     copytree('h/browser/chrome/help', 'build/chrome/help')
     copytree('h/browser/chrome/images', 'build/chrome/images')
-    copytree('h/browser/chrome/lib', 'build/chrome/lib')
     copytree('h/static/images', 'build/chrome/public/images')
-    shutil.copyfile(
-        'h/static/scripts/blocklist.js', 'build/chrome/lib/blocklist.js')
+
+    os.makedirs('build/chrome/lib')
```
nickstenning: Perhaps just remove the lib directory now and dump the bundle straight in `build/chrome/`?

h/browser/chrome/lib/tab-state.js
robertknight: No, that's a mistake. In future I can get the build tooling to add `'use strict'` automatically rather than having to add it to each file manually.

h/buildext.py
robertknight: Hmm, I still have a slight preference for putting the bundles in their own dir separate from the manifest and separate from any vendor code which is not bundled (eg. PDF.js)

nickstenning:
LGTM.

PR #2610, opened 2015-10-12
Refactor blocklist

seanh:
- [x] Doesn't send request when loading first document in new tab
- [x] Fix bad naming


h/browser/chrome/lib/blocklist.js
```diff
@@ -0,0 +1,74 @@
...
+ */
+'use strict';
+
+(function(h) {
+
+  /**
+   * Return a Promise that resolves to whether the URI is blocked and how many
+   * annotations it has.
+   *
+   * @function
```
robertknight: AFAIK  `@function` is redundant above a function declaration

h/browser/chrome/lib/blocklist.js
```diff
@@ -0,0 +1,74 @@
...
+
+        try {
+          blocklist = JSON.parse(this.responseText);
+          total = blocklist.total;
+          blocked = blocklist.blocked;
+        } catch (e) {
+          errorMsg = 'Received invalid JSON from the /blocklist endpoint: ';
+          errorMsg = errorMsg + this.responseText;
+          console.error(errorMsg);
+          reject(errorMsg);
```
robertknight: I think it is a good practice to make arguments to `reject()` always an `Error` instance or an object with a compatible interface. This has two advantages:
 * It makes it easier to find out where the Error came from, because it captures its stack
 * `catch()` blocks will always have the same type of object to work with (Error or some subclass)

h/browser/chrome/lib/blocklist.js
```diff
@@ -0,0 +1,74 @@
...
+          return;
+        }
+
+        if (typeof total !== 'number') {
+          errorMsg = 'Received invalid total from the /blocklist endpoint: ';
+          errorMsg = errorMsg + total;
+          console.error(errorMsg);
```
robertknight: I'd suggest moving the error handling (eg. console logging) up a level into the `h.blocklist.then()...` call so that you can handle the different classes of error in one place.

robertknight:
So to summarize, the endpoint that used to be called `badge` which returned one piece of information, the annotation count for a given URL, has now been replaced by another endpoint called `blocklist` which returns two pieces of information: The annotation count and whether the to load H on that URL or not.

I think the name `blocklist` is confusing if the primary, happy path use for this endpoint is to get the annotation count for the badge. I'd perhaps call it 'summary' or 'status' instead?

I'm still slightly dubious about calling it a blocklist at all, or even using a blocklist of URL patterns to deal with the problem. AIUI the main use case is for URLs that primarily show content that is dependent on some context (user account, time, location etc.) which is not reflected in the URL. We can save that for a separate discussion though since we can change how URLs are "blocked"/don't show badge counts without needing to change the API.

seanh:
@nickstenning @robertknight If either of you gets a chance tomorrow, I could do with a second pair of eyes on this one. I'm trying to refactor this Chrome extension refactoring on top of the recent refactoring of the Chrome extension to use browserify and proxyquire. The error I'm getting is:

```
eval (npm bin)/karma start h/browser/chrome/karma.config.js --no-single-run --auto-watch --port 9877 
13 10 2015 19:21:46.960:INFO [framework.browserify]: registering rebuild (autoWatch=true)
13 10 2015 19:21:48.918:INFO [framework.browserify]: 514428 bytes written (1.21 seconds)
13 10 2015 19:21:48.925:INFO [framework.browserify]: bundle built
13 10 2015 19:21:48.930:WARN [karma]: No captured browser, open http://localhost:9877/
13 10 2015 19:21:48.937:INFO [karma]: Karma v0.13.10 server started at http://localhost:9877/
13 10 2015 19:21:48.943:INFO [launcher]: Starting browser PhantomJS
13 10 2015 19:21:49.212:INFO [PhantomJS 1.9.8 (Linux 0.0.0)]: Connected on socket gy617XQHZOe53lX0AAAA with id 63874887
PhantomJS 1.9.8 (Linux 0.0.0) h.getBlocklist sends the correct XMLHttpRequest to the server FAILED
        SyntaxError: Fake server request processing threw exception: Unable to parse JSON string (/tmp/31ab750cbc734d88123966d3cf60c1ab.browserify:611 <- lib/settings.js:25:0)
PhantomJS 1.9.8 (Linux 0.0.0) h.getBlocklist sends the correct XMLHttpRequest to the server FAILED
        SyntaxError: Fake server request processing threw exception: Unable to parse JSON string (/tmp/31ab750cbc734d88123966d3cf60c1ab.browserify:611 <- lib/settings.js:25:0)
        AssertionError: Unspecified AssertionError
            at /home/seanh/Projects/hypothesis/h/node_modules/chai/chai.js:210
            at /home/seanh/Projects/hypothesis/h/node_modules/chai/chai.js:2092
            at onResolved (/tmp/31ab750cbc734d88123966d3cf60c1ab.browserify:1007 <- ../../../test/blocklist-test.js:38:0)
            at lib$es6$promise$$internal$$tryCatch (/tmp/31ab750cbc734d88123966d3cf60c1ab.browserify:2946 <- ../../../node_modules/es6-promise/dist/es6-promise.js:327:0)
            at lib$es6$promise$$internal$$invokeCallback (/tmp/31ab750cbc734d88123966d3cf60c1ab.browserify:2958 <- ../../../node_modules/es6-promise/dist/es6-promise.js:339:0)
            at lib$es6$promise$$internal$$publish (/tmp/31ab750cbc734d88123966d3cf60c1ab.browserify:2929 <- ../../../node_modules/es6-promise/dist/es6-promise.js:310:0)
            at lib$es6$promise$asap$$flush (/tmp/31ab750cbc734d88123966d3cf60c1ab.browserify:2740 <- ../../../node_modules/es6-promise/dist/es6-promise.js:121:0)
PhantomJS 1.9.8 (Linux 0.0.0) h.getBlocklist "after each" hook FAILED
        TypeError: 'undefined' is not a function (evaluating 'console.error.restore()')
            at /tmp/31ab750cbc734d88123966d3cf60c1ab.browserify:996 <- ../../../test/blocklist-test.js:27:0
PhantomJS 1.9.8 (Linux 0.0.0): Executed 3 of 123 (3 FAILED) ERROR (0.061 secs / NaN secs)
```

I'm _not_ getting the error in Chrome, it seems to work fine, apart from one test failing.

The problem is ine blocklist-test.js <https://github.com/hypothesis/h/commit/8ba6c57da433db3b7f9eec2edd9262e4690c896e>, it seems to be throwing up a JSON parsing error from settings.js, but settings.js should be patched by proxyquire.

I'll take a fresh look at it tomorrow as well.

seanh:
Same error on Travis

robertknight:
@seanh My bad for landing the Browserify-ication at an inconvenient time - I'll look into this tomorrow, along with getting a better understanding of how Proxyquire works.

h/browser/chrome/lib/blocklist.js
seanh: Removed

h/browser/chrome/lib/blocklist.js
seanh: Done

seanh:
Ok, this is ready for review now. Summary:

- The blocklist in the config file that prevents h from launching on some pages (https://github.com/hypothesis/h/pull/2157) is removed, and we now use the blocklist in the database that prevents the Chrome extension from showing numbers of annotations on its badge (https://github.com/hypothesis/h/pull/2573) for both purposes

- `/api/badge` is now `/app/uriinfo` and it now returns the number of annotations of the URI _and_ whether or not the URI is blocked. It no longer returns number 0 for blocked URIs since it's returning blocked as a boolean, up to the extension to decide not to show a number if the URI is blocked

- The previous `blocklist.js` (which read a list of blocklist patterns from the config, and tested a URI against them) is deleted, replaced with a new `uri-info.js` that makes HTTP requests to the uriinfo endpoint

- There's some refactoring in `browser-action.js` and `sidebar-injector.js`, they no longer query the blocklist (now uriinfo) themselves, they get called or not called and passed the right info by `hypothesis-chrome-extension.js`

- The bookmarklet no longer supports blocklisting at all, launches on any page. I think this is okay 'cos it'd be a pain to implement and because the main reasons for needing a blocklist are the Chrome extension showing badge numbers on blocked pages and the Chrome extension auto-launching on blocked pages when browsing in a tab with it activated. All the bookmarklet does is launch when requested by the user.

- Badge admin page is renamed to blocklist

- h.badge Python package renamed h.blocklist

There's some naming inconsistency in that the HTTP endpoint is called uriinfo and it's called uriinfo in the Chrome extension code, but inside the Python code it's the h.blocklist package. I think this is okay for now.

seanh:
This isn't flagged, should it be?

seanh:
For the reviewer, I'll just point out this commit message https://github.com/hypothesis/h/commit/6b3055fad60d0738aa42f98228ed7d1186ddaca0 which might explain some parts

seanh:
Quick note on how the blocklist works in the Chrome extension:

- As soon as the user loads a new page we send a request to `/app/uriinfo`
- As soon as this request comes back we show the num. annotations on the badge (if the uri isn't blocked, and the number is > 0)
- When the user clicks on the button to launch the sidebar (or when it auto-launches if browsing in a tab with the sidebar activated) we: 1. Send the uriinfo request if it hasn't been sent already (since in some cases Chrome doesn't seem to call us when the page loads, see commit message I linked to above), 2. Wait for the uriinfo request to come back or to time out, 3. Launch or don't launch depending on whether the uri is blocked
- The uriinfo request times out after one second, if that happens (or it fails for any reason) we just launch

h/browser/chrome/lib/uri-info.js
```diff
@@ -0,0 +1,120 @@
...
+var settings = require('./settings');
+var uriInfoPromise;
+
+/**
+  * Make an HTTP request to the uriinfo endpoint and return a Promise that
+  * resolves to the value of the response.
+  *
+  */
+function getUriInfo(uri) {
+  if (uri === undefined) {
```
robertknight: For consistency, I'd like to use falsey-ness everywhere rather than explicitly checking for a particular falsey value, except where the difference between an empty string / null and undefined matters.

robertknight:
Landscape reported some unused imports in the Python code which I'd suggest checking.

robertknight:
> When the user clicks on the button to launch the sidebar (or when it auto-launches if
> browsing in a tab with the sidebar activated) we: 1. Send the uriinfo request if it hasn't been
> sent already (since in some cases Chrome doesn't seem to call us when the page loads, see 
>commit message I linked to above), 2. Wait for the uriinfo request to come back or to time
> out, 3. Launch or don't launch depending on whether the uri is blocked

So to clarify, in the _happy_ case, the `uriinfo` request has already completed and this adds no latency to opening the sidebar?

h/browser/chrome/lib/hypothesis-chrome-extension.js
```diff
@@ -147,10 +148,20 @@ function HypothesisChromeExtension(dependencies) {
...
+    //    call uriinfo() again later (when the user clicks on the browser
+    //    button to activate the sidebar) it will return the already-resolved
+    //    Promise.
+    uriinfo(tab.url).then(function(info) {
+      if (!info.blocked) {
+        browserAction.updateBadge(info.total, tab.id);
+      }
+    }).
+    catch(function() {
+      // Silence console error message about uncaught exception here.
```
robertknight: If there is a specific error that you want to ignore for good reason, I'd suggest silencing that one type rather than ignoring all errors.

h/browser/chrome/lib/hypothesis-chrome-extension.js
```diff
@@ -169,15 +180,33 @@ function HypothesisChromeExtension(dependencies) {
...
       });
     }
-    else if (state.isTabInactive(tab.id)) {
+
+    if (state.isTabActive(tab.id)) {
+      return uriinfo(tab.url).then(function(info) {
+        if (info.blocked) {
+            tabErrors.setTabError(
+              tab.id, new errors.BlockedSiteError(
+                "Hypothesis doesn't work on this site yet."));
```
robertknight: Can I see a screenshot of what this looks like for the user?

h/browser/chrome/lib/uri-info.js
```diff
@@ -0,0 +1,120 @@
+/**
+ * A wrapper for the server's uriinfo endpoint.
+ *
+ * Exports a function that makes an HTTP request to the server's uriinfo
+ * endpoint and returns a Promise that resolves to the response. Usage:
+ *
+ *     var uriInfo = require('uri-info');
```
robertknight: In the code above `uri-info` is imported as `uriinfo`. For consistency we should use `camelCase` there.

h/browser/chrome/lib/uri-info.js
```diff
@@ -0,0 +1,120 @@
...
+ *     uriInfo(uri).then(function(info) {
+ *       ...
+ *     });
+ *
+ * This function can be called with the same uri multiple times consecutively
+ * and it'll just keep returning the same Promise - won't make multiple HTTP
+ * requests.
+ *
+ * But if you call the function with a different uri, then call it with the
```
robertknight: So if a user opens a bunch of tabs at once - eg. after a session restore, this will likely result in the same URI being fetched multiple times because `uri-info` has singleton state that is shared between all users.

If the return value was constructor or factory function then each tab could hold its own state.

h/browser/chrome/lib/uri-info.js
```diff
@@ -0,0 +1,120 @@
...
+        blocked = info.blocked;
+      } catch (e) {
+        errorMsg = 'Received invalid JSON from the server: ';
+        errorMsg = errorMsg + this.responseText;
+        console.error(errorMsg);
+        reject({reason: errorMsg});
+        return;
+      }
+
+      if (typeof total !== 'number') {
```
robertknight: If in future we have more complex JSON schemas, we might want to consider something like [validate](https://www.npmjs.com/package/validate) for checking the object structure on the client. In this case the schema is trivial so it isn't necessary.

h/browser/chrome/test/hypothesis-chrome-extension-test.js
```diff
@@ -67,11 +69,16 @@ describe('HypothesisChromeExtension', function () {
...
       deactivate: sandbox.spy(),
+      updateBadge: sandbox.spy(),
     };
     fakeSidebarInjector = {
       injectIntoTab: sandbox.stub().returns(Promise.resolve()),
       removeFromTab: sandbox.stub().returns(Promise.resolve()),
     };
+    fakeUriInfo = sandbox.stub().returns(
+        Promise.resolve({total: 3, blocked: false}));
+    // Fix a proxyquire crash due to a PhantomJS bug.
```
robertknight: If you need a workaround for a bug, we should always reference the upstream issue in the bug tracker or explicitly describe the issue. In this case, you can reference https://github.com/thlorenz/proxyquireify/issues/44

h/browser/chrome/test/hypothesis-chrome-extension-test.js
```diff
@@ -198,44 +205,74 @@ describe('HypothesisChromeExtension', function () {
       it('injects the sidebar if the tab is active', function () {
         var tab = createTab();
         fakeTabState.isTabActive.withArgs(1).returns(true);
-        onUpdatedHandler(tab.id, {status: 'complete'}, tab);
-        assert.calledWith(fakeSidebarInjector.injectIntoTab, tab);
+        return onUpdatedHandler(tab.id, {status: 'complete'}, tab).then(
```
robertknight: You can use something like [chai-as-promised](https://github.com/domenic/chai-as-promised/) to reduce this boilerplate.

h/browser/chrome/test/uri-info-test.js
```diff
@@ -0,0 +1,261 @@
...
+  beforeEach(function() {
+    server = sinon.fakeServer.create({
+      autoRespond: true,
+      respondImmediately: true
+    });
+
+    sinon.stub(console, 'error');
+
+    var settingsPromise = Promise.resolve({serviceUrl: serviceUrl});
+    settingsPromise['@noCallThru'] = true;
```
robertknight: Might be worth moving this to a small helper function (eg. in a `./test/util.js` module) which wraps the input function. That way you only have to document the reason for using `@noCallThru` in one place.

h/browser/chrome/test/uri-info-test.js
```diff
@@ -0,0 +1,261 @@
...
+    server.respondWith(
+      'GET', serviceUrl + '/app/uriinfo?uri=' + uri,
+      [200, {}, '{"total": 3, "blocked": false}']
+    );
+
+    return uriInfo(uri).then(
+      function onResolved() {
+        assert(server.requests.length === 1);
+        var request = server.requests[0];
+        assert(request.method === 'GET');
```
robertknight: Use `assert.equal()` rather than `assert(x === y)`. Unlike Python's testing suite, chai won't automatically show the arguments if the assertion fails.

h/blocklist/test/views_test.py
```diff
@@ -0,0 +1,51 @@
...
+
+
+@uriinfo_fixtures
+def test_uriinfo_raises_if_no_uri():
+    with pytest.raises(httpexceptions.HTTPBadRequest):
+        views.uriinfo(mock.Mock(params={}))
+
+
+@pytest.fixture
+def models(config, request):  # pylint:disable=unused-argument
```
robertknight: Any reason to use this pylint pragma as opposed to using a dummy variable which pylint will ignore?

nickstenning: I'm not sure I understand why the `config` argument is here at all... If it's unused you can just omit it.

nickstenning:
I've checked this out and run it and it seems to work well. The only slight issue I found is that by keeping the old promise around you're guaranteeing that stale data will be displayed even if the user refreshes the page. That is: I go to a page, make three annotations, and refresh, expecting the badge count to be updated. It isn't, because the old promise URI matches the current URI.

This is even true if the promise rejects.

Perhaps get rid of the old promise after some timeout?

```javascript
uriInfoPromise.then(
    function () { setTimeout(function () { uriInfoPromise = null }, 10000); }),
    function () { uriInfoPromise = null; }
)
```

h/blocklist/views.py
```diff
@@ -0,0 +1,36 @@
...
+
+from h import models
+from h.api import search as search_lib
+
+
+@view_config(route_name='uriinfo',
+             request_method='GET',
+             accept='application/json',
+             renderer='json',
+             )
```
nickstenning: Tiny thing, but putting this paren on the preceding line would be more consistent with the rest of the codebase.

h/blocklist/views.py
```diff
@@ -0,0 +1,36 @@
+from pyramid import httpexceptions
+from pyramid.view import view_config
+
+from h import models
+from h.api import search as search_lib
```
nickstenning: Not sure there's any reason to rename this, is there? It's renamed in `h.api.views` simply because there's already a function in that module called `search`.

h/browser/chrome/lib/hypothesis-chrome-extension.js
```diff
@@ -169,15 +180,33 @@ function HypothesisChromeExtension(dependencies) {
...
+            state.errorTab(tab.id);
+        } else {
+          inject(tab);
+        }
+      }).catch(function() {
```
nickstenning: It might be clearer to use the `.then(success, failure)` form here. `.then(success).catch(failure)` is always a bit weird because if you return a promise from `success` then you'll start catching things you didn't intend to catch. No big deal, though.

h/browser/chrome/lib/uri-info.js
```diff
@@ -0,0 +1,120 @@
...
+      var errorMsg;
+
+      try {
+        info = JSON.parse(this.responseText);
+        total = info.total;
+        blocked = info.blocked;
+      } catch (e) {
+        errorMsg = 'Received invalid JSON from the server: ';
+        errorMsg = errorMsg + this.responseText;
+        console.error(errorMsg);
```
nickstenning: Here and below you're both logging to the console and rejecting the promise. It probably makes more sense to leave any console logging to the caller. That'll make the console when testing this module a bit less icky, and will reduce duplication.

h/browser/chrome/lib/uri-info.js
```diff
@@ -0,0 +1,120 @@
...
+  * resolves to the value of the response.
+  *
+  */
+function getUriInfo(uri) {
+  if (uri === undefined) {
+    return Promise.reject({reason: 'uri was undefined'});
+  }
+
+  var p = new Promise(function(resolve, reject) {
+    var request = new XMLHttpRequest();
```
nickstenning: Not now -- I'm happy to merge this as is -- but I note that a lot of the boilerplate in this function is associated with wrapping `XMLHttpRequest` in a promisey API. Another way of doing this would be to use the [`fetch` API](https://fetch.spec.whatwg.org/) which has been available in Chrome since v42 or something like that.

h/browser/chrome/lib/uri-info.js
seanh: Done

h/browser/chrome/lib/uri-info.js
robertknight: Indeed. The [isomorphic-fetch](https://github.com/matthew-andrews/isomorphic-fetch) package is a polyfill for `fetch()` which works in both Node and Browserify and will use the native `fetch()` implementation if available.

You can then write code as:

```js
var fetch = require('isomorphic-fetch');
fetch(url).then(function (resp) {
  return resp.json();
}).then(function (json) {
  // process response
}).catch(function (err) {
  // handle error
});
```

In the Chrome and Firefox extensions, we could just shim this to use the `fetch()` global.

seanh:
> Landscape reported some unused imports in the Python code which I'd suggest checking.

Done

> So to clarify, in the happy case, the uriinfo request has already completed and this adds no latency 
> to opening the sidebar. In the worst-case scenario, it will add one second of latency?

Correct. Actually, running `curl 'https://hypothes.is/api/badge?uri=foo' -w '%{time_total}'` now makes me worry that one second isn't a long enough timeout.

The worst-case scenario may occur more often than you think - when the user is browsing with the extension activated in a tab then it injects the sidebar immediately, now it will have to wait for this request or timeout.

robertknight:
> The worst-case scenario may occur more often than you think - when the user is browsing
> with the extension activated in a tab then it injects the sidebar immediately, now it will
> have to wait for this request or timeout.

Right - I'm quite concerned about adding any delay between the point where the user activates H and the extension showing up.

How about this for an alternative approach - _always_ inject the sidebar as soon as the user activates H, but disable it when the check resolves if H is disabled.

We could potentially use the sidebar as a place to display a more useful message about H and relevant links to pages with more info.

seanh:
> How about this for an alternative approach - always inject the sidebar as soon as the user activates H, 
> but disable it when the check resolves if H is disabled.

Is there a good way to communicate this state (that the uriinfo request has come back and said blocked) between the Chrome extension's content scripts and the sidebar code running in page context?

My other concern is with launching h and having it working momentarily, and then suddenly disabling itself while the user may be interacting with it.

If we're worried about the delay here, I wonder if we should go ahead with the approaches we discussed (Bloom filter etc) to avoid having to send a request for every page?

h/browser/chrome/lib/hypothesis-chrome-extension.js
seanh: @robertknight How's this for the error handling? https://github.com/hypothesis/h/commit/6dabcb63be7213b2395a03ff51f19fa51c25a220

I'm a bit sad that I had to use `this`, but I think the alternative is just `return {name: 'UriInfoError', ...}` and then in the catch check the value of `.name`.

robertknight:
> Is there a good way to communicate this state (that the uriinfo request has come back and
> said blocked) between the Chrome extension's content scripts and the sidebar code running
> in page context?

If you can get a reference to the sidebar's iframe you can use `window.postMessage()`. In the sidebar you could listen for that via `window.addEventListener('message')`

> My other concern is with launching h and having it working momentarily,
> and then suddenly disabling itself while the user may be interacting with it.

It will likely take several seconds for the user to start actually doing anything much - by which time the timeout would have expired anyway.

h/browser/chrome/lib/hypothesis-chrome-extension.js
seanh: untitled screencast 4

Note that I'm clicking the browser button, but Screencastify shows the mouse cursor in the wrong location

h/browser/chrome/lib/hypothesis-chrome-extension.js
robertknight: >  I'm a bit sad that I had to use this, but I think the alternative is
just return {name: 'UriInfoError', ...} and
> then in the catch check the value of .name

Checking the value of .name is actually a reasonable approach. Subclassing
`Error` is indeed awkward in JS. The important thing I think is to return
an object which conforms to the interface of 'Error' interface - namely
having name and message properties.

h/browser/chrome/lib/uri-info.js
seanh: I'm not sure what you want me to change here? This is consistent with other modules, eg. `sidebar-injector.js` is imported as `SidebarInjector`.

I've corrected some other places where `uri-info.js` was imported as `uriinfo` (lowercase) though

h/browser/chrome/lib/uri-info.js
seanh: Because the promise is a singleton shared by all tabs, if I open multiple tabs to the same URL consecutively it'll only send one request. If I open multiple tabs to different URLs it'll send multiple requests.

I think I might have to investigate what happens when we send lots of simultaneous requests for different tabs with different URLs, that should be showing different badge numbers or be blocked/not blocked...

h/browser/chrome/test/hypothesis-chrome-extension-test.js
seanh: This bit is actually gone (factored out) now anyway

h/browser/chrome/test/hypothesis-chrome-extension-test.js
seanh: We don't have this lib yet do we? I can add it and replace lots of places where we're doing this in tests, but suggest separate pr

h/browser/chrome/test/uri-info-test.js
seanh: Done

h/blocklist/test/views_test.py
seanh: It is helpful to have it there if you come along to edit the code later and need it though

h/blocklist/views.py
seanh: Done

h/blocklist/views.py
seanh: Done

h/browser/chrome/lib/hypothesis-chrome-extension.js
seanh: Done

seanh:
Even if the user doesn't interact with the sidebar during the time that it's waiting to hear back about whether the site is blocked, the sidebar would have to be shown in some sort of "not ready yet" UI state, or else it would appear ready and then suddenly change to not ready, wouldn't it?

I think the alternative solution we discussed on Slack/Hangout (moving the blocklist pattern matching client side so that it's really fast) is gonna be better - don't launch the sidebar at all if the URL is blocked, but find out whether it's blocked instantly so there's no delay.

seanh:
Ok, I've responded to most of the comments, just a couple of small ones from Nick still to do.

I'm concerned about Nick's comment about it keeping old stale promises around, and Robert's comment about what happens when lots of tabs are loading at once, I need to investigate those.

h/browser/chrome/lib/uri-info.js
robertknight: > I've corrected some other places where uri-info.js was imported as
uriinfo (lowercase) though

That's what I meant :)

h/browser/chrome/test/hypothesis-chrome-extension-test.js
robertknight: > We don't have this lib yet do we? I can add it and replace lots of places
where we're doing this in tests, but suggest separate pr

Makes sense

h/blocklist/test/views_test.py
nickstenning: What? If typing `config, ` is a problem, then surely typing `  # pylint:disable=unused-argument` must be an almost insurmountable one :stuck_out_tongue:.

I'd rather not have code (and pylint pragmas) in our repository that isn't necessary.

h/browser/chrome/lib/uri-info.js
seanh: I think I'd rather leave these here. If I move the error logging into a catch or onRejected in the code that's using the promise, then each error will get logged twice (once when the promise is called when the page loads, again when the same promise is called again when the user clicks the button to launch h), and if you refresh the page the error will also be logged again even though the http request wasn't made again. With the logging where it is, it's one failed http request = one logged error.

seanh:
I've fixed the stale promise issue by adding a `force_refresh` option.

h/blocklist/test/views_test.py
seanh: I mean if someone comes to edit that function, they can see that it can accept a config param, it's sitting there just waiting to be used, inviting change! I've removed it anyway

h/browser/chrome/lib/uri-info.js
seanh: Ok, there seems to be a problem with a race condition between tabs:

- Open tab 1 to uri 1, uriinfo request is sent
- Open tab 2 to uri 2, uriinfo request is sent
- Tab 2's uriinfo response is received
- Tab 1's uriinfo response is received

Both tabs will show the correct badge number, but if you click to activate tab 1 it will unnecessarily send the uriinfo request again. That's the only thing that goes wrong as far as I can tell. It does seem like the Promise should be per tab instead of global...

seanh:
Rebased on master and squashed a bit

seanh:
Ok, I just need to fix this issue https://github.com/hypothesis/h/pull/2610#discussion_r42152148

Possibilities are:

1. Keep a mapping of tab IDs to stored Promises in uri-info.js (storing one Promise per tab, instead of one global Promise)

2. Enhance tab-state.js, which already maintains a mapping of tab IDs to tab states (active, inactive, errored) to allow other modules to call it and add arbitrary key value pairs to a given tab's state. Use that to store per tab Promises.

robertknight:
I just did some initial testing of this. The `uri-info` requests failed with a 404 for reasons I haven't yet debugged. Ignoring that local issue however, if the 'uriinfo' request fails for any reason other than the site being blocked then H doesn't load at all and the user doesn't see any indication why.

robertknight:
As a general point, throwing exceptions from the H extension is completely invisible to users and also, until we get Sentry logging set up, to us as well.

h/browser/chrome/lib/uri-info.js
```diff
@@ -0,0 +1,138 @@
...
+ * @param {String} uri The URI to be checked
+ * @param {Boolean} forceRefresh Make a new HTTP request, even if the
+ *   previously made request was for the same uri (optional, default: false)
+ *
+ * @returns {Promise} A Promise that resolves to an object with two properties:
+ *   1. total:   the number of annotations there are of this URI (Number)
+ *   2. blocked: whether or not this URI is blocklisted (Boolean)
+ *
+*/
+function get(uri, forceRefresh) {
```
robertknight: As a general coding style comment, I would suggest preferring options objects to booleans unless the meaning is very obvious or performance critical. Looking at a caller - `uriInfo.get(uri, true)` I have no idea what that second parameter means.

robertknight:
The UX for this is currently as follows. If I'm on a blocked site, firstly no badge count will be displayed and an error indicator appears ('!'). Hovering the mouse over the error indicator displays this:

blocked-site-toolbar

If I then click on the badge, I get this:

blocked-site-help

I really don't think we can ship this. My current thinking is that as a first iteration, we shouldn't actually block H from loading but rather just not display a badge count. As a next step, get proper design input on the UX for this.

dwhly:
Maybe we can skip the "!" and warning message too?

robertknight:
@dwhly - Based on Jon's response to my mail on dev@ and earlier discussion with sean and nick today it doesn't look like we need this for the time being. We'll review tomorrow in case someone _does_ come up with a good reason to keep it.

seanh:
> My current thinking is that as a first iteration, we shouldn't actually block H from loading but rather just not display a badge count.

This is the currently deployed behaviour

PR #2611, opened 2015-10-13
Handle exceptions broadcasting WS messages gracefully

robertknight:
If an exception occurs whilst processing an NSQ message
in the WebSocket server, log the exception and continue
on to the next message.

Since there is a single greenlet iterating over all
NSQ messages that might result in a WebSocket broadcast,
an error processing one message would previously cause all
subsequent WS broadcasts to cease.

### DIFF
--- a/h/queue.py
+++ b/h/queue.py
@@ -18,7 +18,7 @@ class NamespacedNsqd(object):
         return self.client.publish(topic, data)
 
 
-def get_reader(request, topic, channel):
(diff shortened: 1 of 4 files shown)


nickstenning:
Thank you for this! Unfortunately, I've just realised there's a good reason not to do this -- I'll explain tomorrow. It *does* need fixing, it's just that this isn't quite how to fix it. So, no merging while I'm not looking :stuck_out_tongue:

PR #2612, opened 2015-10-13
Don't endlessly re-request features on connectivity issues

### LINKED ISSUE #2547: Features fetcher busy-polls ajax requests when the connection fails
## Steps to reproduce

1. Open a page with the sidebar active (e.g. the development homepage)
2. Stop the development server
3. Wait for `CACHE_TTL` from `h/static/scripts/features.js` (5m by default)

## What should happen

The features fetcher should exponentially back off on trying to fetch the new features data.

## What does happen

It makes a request every time the digest cycle happens.

nickstenning:
If, for whatever reason, the request to fetch features data fails, simply retry a little later (with a randomised exponential backoff). This commit uses the [retry module](https://www.npmjs.com/package/retry) to provide the backoff logic.

N.B. If we fail 10 times (over a period spanning between ~15 and ~30 minutes, using the defaults from the retry module) then we will simply stop looking for new features data. There's no point burning client CPU forever in the hope that the server reappears.

Fixes #2547.

---

FYI:

```shell
$ browserify -r retry | uglifyjs -m | wc -c
    3315
```

### DIFF
--- a/h/static/scripts/features.js
+++ b/h/static/scripts/features.js
@@ -21,28 +21,38 @@
  */
 'use strict';
 
+var retry = require('retry');
+
 var CACHE_TTL = 5 * 60 * 1000; // 5 minutes
 
 function features ($document, $http, $log) {
   var cache = null;
-  var pending = false;
+  var operation = null;
   var featuresUrl = new URL('/app/features', $document.prop('baseURI')).href;
 
   function fetch() {
     // Short-circuit if a fetch is already in progress...
-    if (pending) {
+    if (operation) {
       return;
     }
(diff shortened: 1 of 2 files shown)


h/static/scripts/features.js
```diff
@@ -21,28 +21,38 @@
  */
 'use strict';
 
+var retry = require('retry');
+
 var CACHE_TTL = 5 * 60 * 1000; // 5 minutes
 
 function features ($document, $http, $log) {
   var cache = null;
```
robertknight: Could use a comment here noting the type of `cache`

PR #2613, opened 2015-10-13
Sort groups server-side

robertknight:
Sort groups server-side by name, then by ID

 * Sort groups on the server in a useful order on the server
   rather than the client.

   This reduces the need for logic which is otherwise likely
   to be duplicated across clients.

 * Sort groups by name, _then by ID_. This ensures that if
   multiple groups have the same name they are sorted in
   a consistent order in the client.

### DIFF
--- a/h/session.py
+++ b/h/session.py
@@ -18,6 +18,15 @@ def pop_flash(request):
             for k in ['error', 'info', 'warning', 'success']}
 
 
+def _group_sort_key(group):
+    """Sort private groups for the session model list"""
+
+    # groups are sorted first by name but also by ID
+    # so that multiple groups with the same name are displayed
+    # in a consistent order in clients
+    return (group.name.lower(), group.hashid)
+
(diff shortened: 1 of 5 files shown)


h/session.py
```diff
@@ -18,6 +18,15 @@ def pop_flash(request):
             for k in ['error', 'info', 'warning', 'success']}
 
 
+def _group_sort_key(group):
+    """Sort private groups for the session model list"""
+
+    # groups are sorted first by name but also by ID
+    # so that multiple groups with the same name are displayed
+    # in a consistent order in clients
+    return '%s:%s' % (group.name.lower(), group.hashid)
```
nickstenning: It would probably be more pythonic (and not dependent on the sorting properties of `':'`) to return a tuple here: `return (group.name.lower(), group.hashid)`

nickstenning:
LGTM.

PR #2615, opened 2015-10-14
T93 New 'Clear selection' button design

robertknight:
Implement new design for search status bar

This provides a clearer design for the search status
button.

The new design is part of the groups roll-out and
is only shown when the groups feature is enabled.

t92-new_clear_btn

t92-new_clear_btn_2

### DIFF
--- a/h/static/scripts/app.coffee
+++ b/h/static/scripts/app.coffee
@@ -124,6 +124,7 @@ module.exports = angular.module('h', [
 .directive('windowScroll', require('./directive/window-scroll'))
 .directive('dropdownMenuBtn', require('./directive/dropdown-menu-btn'))
 .directive('publishAnnotationBtn', require('./directive/publish-annotation-btn'))
+.directive('searchStatusBar', require('./directive/search-status-bar'))
 
 .filter('converter', require('./filter/converter'))
 .filter('moment', require('./filter/moment'))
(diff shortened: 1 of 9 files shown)


conordelahunty:
@robertknight Thanks for this

1. Can you improve the alignment of the X. It looks like it needs to be pushed up 2-3px.

2. How is the hover state defined? Another 10-15% darker?

3. I think might combine the button text (Clear selection) and the text that follows (1 selected annotation) in the future, but this is fine for now.

I think we may be able to ship this one without going behind the Groups feature flag.

robertknight:
> How is the hover state defined? Another 10-15% darker?

Same as the other places where this button style is used. The only difference for this variant is that it is 5px shorter

> I think we may be able to ship this one without going behind the Groups feature flag.

If you're happy to ship this unflagged that would simplify things for me.

robertknight:
t92-new_clear_btn_3

t92-new_clear_btn_4

conordelahunty:
@robertknight I assume the styling for that tooltip is defined by your browser?

robertknight:
@conordelahunty - Yes, that the browser default styling under Ubuntu.

nickstenning:
Looks schweeeeeet! :heart: :present:

dwhly:
:+1:

PR #2617, opened 2015-10-14
streamer: Don't silently stop processing on exceptions

nickstenning:
Prior to this commit, if an exception is thrown by a message handler (i.e. `h.streamer.broadcast_annotation_message` or `h.streamer.broadcast_session_change_message`) then the worst possible combination of events will occur:

1. NSQ will not be notified, and the message will thus not be requeued.
2. The greenlet processing the internal event queue will crash. There is no support in the code for restarting this greenlet, and thus all WebSocket clients connected to this worker (or that connect subsequently) will stop receiving updates.
3. The readers connected to NSQ will continue to receive messages and push them onto the internal event queue. These messages will never be handled, and will continue to stack up in memory until, presumably, the process OOMs.

All in all, not a terribly happy situation. It's very likely that this is what was causing our realtime update server to stop working occasionally, so thank you to @robertknight for spotting that!

This commit completely alters how we handle queue messages, fixing all three issues. The main changes are as follows:

- The internal queue is gone. Instead, we spin up a greenlet on app boot which is responsible, in turn, for starting the queue reader processes.
- The handling of all messages is done synchronously in the `on_message` handlers of the readers. In the event of an exception, `gnsq` will handle requeueing the message with NSQ.
- If any reader greenlet terminates (an event which should never occur) then all reader greenlets are killed and restarted. A message is also written to the logs to indicate that this happened.

There are numerous further simplifications which can now be made to this code, but I've omitted them from this PR in the interests of making the changes as clear as possible.

### DIFF
--- a/h/streamer.py
+++ b/h/streamer.py
@@ -11,16 +11,18 @@ import unicodedata
 import weakref
 
 import gevent
-import gevent.queue
 from jsonpointer import resolve_pointer
 from jsonschema import validate
 from pyramid.config import aslist
+from pyramid.events import ApplicationCreated
+from pyramid.events import subscriber
 from pyramid.httpexceptions import HTTPBadRequest, HTTPForbidden
 from pyramid.threadlocal import get_current_request
 from ws4py.exc import HandshakeError
 from ws4py.websocket import WebSocket as _WebSocket
 from ws4py.server.wsgiutils import WebSocketWSGIApplication
 
+from h import queue
 from h.api import nipsa
 from h.api import uri
 from h.api.search import query
@@ -443,11 +445,7 @@ USER_TOPIC = 'user'
 
 
 class WebSocket(_WebSocket):
-    # Class attributes
-
-    # Tuples of (topic, message) pulled from the message queue
-    event_queue = None
-
+    # All instances of WebSocket, allowing us to iterate over open websockets
     instances = weakref.WeakSet()
     origins = []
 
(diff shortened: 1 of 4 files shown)


h/test/streamer_test.py
```diff
@@ -283,23 +268,26 @@ def setUp(self):
     def tearDown(self):
         self.should_patcher.stop()
 
+    def msg(self, n):
```
robertknight: `msg` appears to be unused

robertknight:
One minor nit but otherwise LGTM.

h/test/streamer_test.py
nickstenning: Oops! Thank you.

nickstenning:
Fixed. @seanh I think you should probably have a look at this one, too.

h/streamer.py
```diff
@@ -655,12 +620,82 @@ def should_send_annotation_event(socket, annotation, event_data):
...
+    """
+    Process a message from the queue.
+
+    Deserializes and distributes messages from the queue to the appropriate
+    handler functions.
+    """
+    if reader.topic == ANNOTATIONS_TOPIC:
+        broadcast_annotation_message(message, WebSocket.instances)
+    elif reader.topic == USER_TOPIC:
+        broadcast_session_change_message(message, WebSocket.instances)
```
seanh: Do we assume reader.topic will never be anything other than these two? Should we crash or log something if it is?

h/streamer.py
```diff
@@ -655,12 +620,82 @@ def should_send_annotation_event(socket, annotation, event_data):
     return True
 
 
+def process_message(reader, message):
+    """
+    Process a message from the queue.
+
+    Deserializes and distributes messages from the queue to the appropriate
```
seanh: Is this function actually doing any deserializing? I don't see any

h/streamer.py
```diff
@@ -655,12 +620,82 @@ def should_send_annotation_event(socket, annotation, event_data):
...
+
 def _random_id():
     """Generate a short random string"""
     data = struct.pack('Q', random.getrandbits(64))
     return base64.urlsafe_b64encode(data).strip(b'=')
 
 
+def _wait_for_readers(readers):
+    """Wait for one or more of the readers to exit."""
+    while 1:
```
seanh: `while True`?

seanh:
Ok, I've tested this by making `broadcast_annotation_message()` crash if the annotation contains the word crash. It seems to work as expected: live updates will (eventually) start working again between the two web browsers, which they won't on master.

The minor comments I made aren't important, the code is a simplification and is nice and easy to follow. @nickstenning happy to see this merged with or without those changes

h/streamer.py
nickstenning: Not yet. If we merge this I'll promise I'll make it do so very soon :)

h/streamer.py
nickstenning: We certainly shouldn't crash, as that will requeue the message, which will result in another crash, etc, etc, up to `MAX_ATTEMPTS`, whatever that is at the moment...

I can certainly add a `log.warn` here.

h/streamer.py
nickstenning: Meh. Once upon a time I was told that `while 1:` was more Pythonic (for what are probably highly dubious reasons... did you know you can reassign `True`.... `True = 0`?), but tbh you're probably right. I'll change it.

PR #2620, opened 2015-10-15
Revert "Use "should" instead of "must" for search queries"

### LINKED ISSUE #2540: unexpected result for api/rss/atom query that combines tag and user facets
https://hypothes.is/api/search?tags=wisdom -> 8

https://hypothes.is/api/search?tags=wisdom&uri=https://posts.postlight.com/this-is-postlight-fe49148b014 -> expect 1, got 1

https://hypothes.is/api/search?tags=wisdom&user=judell -> expect 1, got 1242

(https://hypothes.is/api/search?user=judell -> 1235)

nickstenning:
This reverts commit 7a9c8c7d54d109ad245047790c4de157b054141c, as discussed in #2540. At the moment the search endpoint is being used primarily as a filter endpoint, and default-AND semantics are more useful for this purpose.

Fixes #2540.

### DIFF
--- a/h/api/search/query.py
+++ b/h/api/search/query.py
@@ -38,7 +38,7 @@ class Builder(object):
         query = {"match_all": {}}
 
         if matchers:
-            query = {"bool": {"should": matchers}}
+            query = {"bool": {"must": matchers}}
 
         if filters:
             query = {
--- a/h/api/search/test/query_test.py
+++ b/h/api/search/test/query_test.py
@@ -145,7 +145,7 @@ def test_builder_default_param_action():
 
     q = builder.build({"foo": "bar"})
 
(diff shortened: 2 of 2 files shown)


nickstenning:
It's worth noting that a default-OR for fallthrough keywords is inconsistent: some keywords are implemented as filters (which are required, and thus behave like they're ANDed with the rest of the query). I take the point made in #2540 that a true "search" endpoint would default to OR, but we're not implementing a search endpoint, no matter what it's called. It's a mechanism for filtering and retrieving annotations matching some criteria. When we implement a more useful search interface, we can reconsider the organisation of the various bits of this API.

judell:
Thanks!

PR #2622, opened 2015-10-15
New design for top bar

robertknight:
This implements the new top-bar design from https://trello.com/c/DZEDxU4C/90-top-bar-ui-updates

Parts of this are feature flagged and parts will be visible for everyone:

For **all users**:
 * The top bar corners have been squared off, the height increased slightly and the icons now line up vertically with the annotation cards below

For **groups-enabled users only**:
 * The top bar has been re-arranged with the group name on the left, and search/sort/share/account icons on the right
 * The account name / provider has been replaced by a head-n-shoulders icon
 * The search bar is now collapsed by default. It expands on hover, when focused and when the search query is non-empty

When testing this, please be aware that the new design only appears if you are logged in when the app initially loads.

**New design as seen by non-groups enabled users**

t90-new_top_bar_signed_out

**New design as seen by groups-enabled users**

t90-new_top_bar_new_account_menu

----
**TODO**

- [X] Fix the spectacularly funky effect when resizing the sidebar. I suspect this is due to the way the search icon expander is implemented
- [X] Group list label tweaks: Make the group name bold, correct the group icon color, remove the spacing between the group label and the drop-down arrow, increase the spacing between the group type indicator and the name
- [X] Account icon tweaks: Correct the drop-down indicator color
- [X] Update the labels and order of items in the Account menu to match the mocs: "My Annotations" should be moved to the top and the name should be the username. "Account" => "Account settings", swap Help / Feedback order
- [X] Correct the vertical alignment of the search icon (see conor's comments below)
- [X] Fix the top bar not being centered when viewing it in the stream
- [X] Fix the group name heading being pushed over to the left when the search icon is expanded


conordelahunty:
@robertknight This is starting to look good. One small thing, the search icon is slightly off, can you move it up 1-2px.

I think my main feedback will come once I can click around with it and see how it feels, so if this could make it's way to staging or somewhere like that in the not too distant future, I'd appreciate it.

Nice one!

robertknight:
Okay, this is ready to roll. There are some additional changes on the [original card](https://trello.com/c/DZEDxU4C/90-top-bar-ui-updates) relating to moving the filter icon into the top bar, moving the loading icon into the bucket bar and changing the signed-out design. These will be moved to separate cards.

seanh:
If I'm signed-in as a user who doesn't have the groups feature enabled, I still get the old username dropdown not the new head-n-shoulders icon, is that intentional?

seanh:
Search doesn't open on hover for me, I have to click (which is fine if that's intended)

seanh:
Slightly weird interaction if I:

1. Open the search, type a search term, hit enter
2. Edit the text in the search box, deleting it all. But don't hit enter/submit it.
3. Move the keyboard focus out of the search box, it closes.
4. Click to open the search box again, _now_ it seems to submit my new (empty) search text for me

untitled screencast - cropped 3

seanh:
> Search doesn't open on hover for me, I have to click (which is fine if that's intended)

Ah, intended

dwhly:
Question: Is it possible to have the search box move as far left as it can given how wide the scope text is at that moment?  Or is that too complicated?

seanh:
Other than the issues mentioned above, lgtm

robertknight:
> If I'm signed-in as a user who doesn't have the groups feature enabled,
> I still get the old username dropdown not the new head-n-shoulders icon, is that intentional?

Yes. The new UX feature flagged as part of groups except for the change of shape to the top bar.

> Search doesn't open on hover for me, I have to click (which is fine if that's intended)

Yes, that's the behavior we settled on after some testing this morning.

> Is it possible to have the search box move as far left as it can given how wide the scope text is at that moment?

Yes, I think that should be straightforward.

robertknight:
@seanh - Which tool are you using to make the videos out of interest?

seanh:
@robertknight Using this: https://www.screencastify.com/ Record, crop, then download as gif, and you can drag the gif from Chrome's downloads bar straight into github

nickstenning:
This looks good to me, acknowledging the issues raised by others above. Given that the issues raised are only with the feature-flagged version of the bar, I'm going to go ahead and merge this. Fixes can come along in a later PR.

robertknight:
> Click to open the search box again, now it seems to submit my new (empty) search text for me

@seanh  - So this is happening because the search icon is a `<button>` and by default, `<button>` elements nested in a `<form>` will submit the form. Easy to fix, just understanding a few issues around focus handling and unit tests.

PR #2624, opened 2015-10-16
Simplify streamer

nickstenning:
Some more simplifications for `h.streamer`. These changes shouldn't change any functionality, and they don't fix any bugs, but they do make the code easier to read and easier to test. Specifically, our per-topic message handlers are no longer responsible for:

- serialisation/deserialisation
- writing to the websocket
- iterating over open websockets

All these responsibilities are now in one place, in `h.streamer.process_message`.

### DIFF
--- a/h/streamer.py
+++ b/h/streamer.py
@@ -523,51 +523,68 @@ class WebSocket(_WebSocket):
             raise
 
 
-def _annotation_packet(annotations, action):
+def handle_annotation_event(message, socket):
     """
-    Generate a packet suitable for sending down the websocket that represents
-    the specified action applied to the passed annotations.
+    Get message about annotation event `message` to be sent to `socket`.
+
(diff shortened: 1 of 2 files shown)


h/streamer.py
```diff
@@ -590,76 +605,48 @@ def _authorized_to_read(effective_principals, annotation):
...
-    """
-    if reader.topic == ANNOTATIONS_TOPIC:
-        broadcast_annotation_message(message, WebSocket.instances)
-    elif reader.topic == USER_TOPIC:
-        broadcast_session_change_message(message, WebSocket.instances)
-    else:
-        log.warn("streamer: don't know how to handle message from topic '%s'",
-                 reader.topic)
+    for socket in WebSocket.instances:
+        reply = handler(data, socket)
```
seanh: `reply` seems an odd thing to call this (it's not being sent back to the thing that sent the original message)

PR #2625, opened 2015-10-16
Improve flow when joining a group from logged-out

nickstenning:
Now, when joining a group from a logged-out state, you'll first see a "Sign in to join $GROUP" button, then you'll log in, then you'll be redirected back to the group page where you'll see a "Join $GROUP" button.

### DIFF
--- a/h/groups/templates/join.html.jinja2
+++ b/h/groups/templates/join.html.jinja2
@@ -21,7 +21,8 @@
           </button>
         </form>
       {% else %}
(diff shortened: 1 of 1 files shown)


nickstenning:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;nIuDYNAj&#x2F;127-improve-flow-for-users-who-are-not-signed-in-when-visiting-group-join-page"> Improve flow for users who are not signed in when visiting group join page</a>

PR #2626, opened 2015-10-16
Add userid_domain config setting

seanh:
Add a config setting that overrides the domain in userids like
"acct:seanh@domain.com". If no h.userid_domain config file setting or
USERID_DOMAIN environment variable is given it falls back on the old behavior:
use request.domain.

This new setting is useful for our staging and production sites.
request.domain is hypothes.is on production but stage.hypothes.is on staging.
The two sites share the same database and search index. It's actually the same
user account in the database that you're logging in to when you log in to
either staging or production, since the user accounts in the database are
keyed by just the username (eg. seanh) and don't use the domain part. But for
example the permissions access control lists in the annotations use the full
userids, acct:seanh@stage.hypothes.is or acct:seanh@hypothes.is. This means
that if you logged into your account via stage.hypothes.is you won't be able to
edit your annotations made while logged into production hypothes.is, and
vice-versa.

With this new config setting, we can override the domain on the staging site
so that you get acct:seanh@hypothes.is when you log into either site and can
always edit all of your annotations.

This commit also removes a feature from the get_by_userid() class method.
Previously you would pass it a domain like hypothes.is and a userid like
acct:seanh@hypothes.is, and if the two domains didn't match it would return
None. This could never happen because everywhere this method is called it's
passed request.domain and request.authenticated_userid so the domains always
match.

But with the new config setting, it's possible for request.authenticated_userid
to be acct:seanh@hypothes.is while request.domain is stage.hypothes.is, causing
get_by_userid() to return None, which breaks things, for example you can no
longer login.

I could change every place that calls get_by_userid() to pass in the
userid_domain setting instead of request.domain. But that's just silly:
userid_from_username() is fetching userid_domain, and the code that calls
get_by_userid() if fetching the same userid_domain setting, and then
get_by_userid() is verifying that the setting is the same as itself and
having a `return None` line that can never get called.

Instead, just remove this feature from get_by_userid().

### DIFF
--- a/h/accounts/models.py
+++ b/h/accounts/models.py
@@ -200,14 +200,8 @@ class User(Base):
         return valid
 
     @classmethod
-    def get_by_userid(cls, domain, userid):
+    def get_by_userid(cls, userid):
         """Return the user with the given ID, or None.
-
-        :param domain: The domain for the current request, for example:
-            u'hypothes.is'. This must match the domain part of the userid
-            argument - you can't retrieve users who don't belong to the given
-            domain.
-        :type domain: unicode
 
         :param userid: A userid unicode string, for example:
             u'acct:kim@hypothes.is'
@@ -221,9 +215,6 @@ class User(Base):
             username, userdomain = util.split_user(userid)
         except TypeError:
             # userid didn't match the pattern that split_user() expects.
-            return None
-
-        if userdomain != domain:
             return None
 
         return cls.get_by_username(username)
--- a/h/accounts/test/models_test.py
+++ b/h/accounts/test/models_test.py
@@ -20,7 +20,7 @@ def test_get_by_userid_calls_split_user(util):
     """It should call split_user() once with the given userid."""
     util.split_user.return_value = ('fred', 'hypothes.is')
 
-    models.User.get_by_userid('hypothes.is', 'acct:fred@hypothes.is')
+    models.User.get_by_userid('acct:fred@hypothes.is')
 
     util.split_user.assert_called_once_with('acct:fred@hypothes.is')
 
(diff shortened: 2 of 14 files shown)


seanh:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;JkzDo3l1&#x2F;119-enable-staff-to-always-use-the-stage-extension-and-site"> Enable staff to always use the stage extension and site</a>

seanh:
Oh, I should add a "migration" script to connect to Es directly and delete all stage.hypothes.is annotations, I'll do that in a gist now

seanh:
One thing to note: once this gets deployed, if you're already logged-in to stage I think you'll need to log out and log in again

seanh:
@nickstenning And here's a script to delete all `@stage.hypothes.is` annotations. Do run a backup first! https://gist.github.com/seanh/c8492ff44166cf4a0f7e

tilgovi:
Long overdue! Yay.

h/accounts/models.py
```diff
@@ -223,9 +217,6 @@ def get_by_userid(cls, domain, userid):
             # userid didn't match the pattern that split_user() expects.
             return None
 
-        if userdomain != domain:
```
nickstenning: The thing is, if we're not paying any attention to the domain, I don't think we should have a `get_by_userid` at all, because it has rather misleading behaviour -- `models.User.get_by_userid('acct:foo@microsoft.com')` will happily return the user whose userid is `acct:foo@hypothes.is`.

This isn't necessarily a reason to block merging this PR. Just flagging something that needs cleaning up.

nickstenning:
Other than my comment about the weirdness of having a `get_by_userid` that just ignores part of its argument, this looks good. My only other request is for `AUTH_DOMAIN` rather than `USERID_DOMAIN`, which is overly specific, I think.

nickstenning:
I'm just going to merge this and tweak the name of the config var on master.

Thanks, @seanh! :cake:

PR #2629, opened 2015-10-17
Upgrade to Angular 1.4.7

robertknight:
This is mostly @nickstenning 's work upgrading Angular to v1.4.7

Aside from much, err, fun debugging what happens when you mix native and Angular Promises in tests, due to Angular Promises having support for synchronous fulfillment/rejection in tests and native ones not, the fixes required are mostly minor edge-case behavior changes.

All of the changes look backwards compatible to me so if we want, we could merge the fixes first and do a little more testing with Angular 1.4.x in a branch.


nickstenning:
I can't find anything wrong with this. I'm not going to merge it straight away simply because that will complicate pushing to production tomorrow. In addition, depending on how straightforward getting #2636 is, it might be easier to rebase this on that than the reverse, as it simply removes code that is changed here (the match directive and the angular-bootstrap-tabbable dependency).

nickstenning:
@robertknight Did you want to test this in Safari and IE before merging?

nickstenning:
Meh. Seems to work in Safari. If the worst comes to the worst we can back it out.

tilgovi:
Nice!

PR #2631, opened 2015-10-19
Fix an issue where clicking the search icon would submit the form

robertknight:
Clicking the search icon should only focus the search field,
not submit the search.

This addresses the issue mentioned at https://github.com/hypothesis/h/pull/2622#issuecomment-148808509


nickstenning:
LGTM.

PR #2632, opened 2015-10-19
Update npm / python modules before running tests

robertknight:
Ensure npm/python modules are up to date before running tests


nickstenning:
:+1:

PR #2633, opened 2015-10-19
Make settings loader CSP-compatible

nickstenning:
I forgot about CSP! We disallow inline scripts in the Chrome extension, so 18c47ba broke the extension. This commit replaces an inline script with some JSON data, parsed by the settings factory.


tilgovi:
Huh? This still has an inline script.

PR #2635, opened 2015-10-19
Remove get_by_userid()

seanh:
Replace it with a request.authenticated_user property for the authenticated
User object.

### DIFF
--- a/h/accounts/__init__.py
+++ b/h/accounts/__init__.py
@@ -44,8 +44,11 @@ def authenticated_user(request):
     if not request.authenticated_userid:
         return None
 
-    # pylint: disable=unpacking-non-sequence
-    username, _ = util.split_user(request.authenticated_userid)
+    try:
+        username = util.split_user(request.authenticated_userid)['username']
+    except ValueError:
+        return None
+
     return models.User.get_by_username(username)
 
 
--- a/h/accounts/__init__.py
+++ b/h/accounts/__init__.py
@@ -44,24 +44,35 @@ def auth_domain(request):
     return request.registry.settings.get('h.auth_domain', request.domain)
 
 
+def get_user(userid, request):
+    """Return the User object for the given userid, or None.
+
+    This will also return None if the given userid is None, if it isn't a valid
+    userid, if its domain doesn't match the site's domain, or if there's just
+    no user with that userid.
+
+    """
+    if userid is None:
+        return None
+
+    try:
+        parts = util.split_user(userid)
+    except ValueError:
+        return
+
+    if parts['domain'] != request.auth_domain:
+        return None
+
+    return models.User.get_by_username(parts['username'])
+
+
 def authenticated_user(request):
     """Return the authenticated user or None.
 
     :rtype: h.accounts.models.User or None
 
     """
-    if not request.authenticated_userid:
-        return None
-
-    try:
-        parts = util.split_user(request.authenticated_userid)
-    except ValueError:
-        return None
-
-    if parts['domain'] != request.auth_domain:
-        return None
-
-    return models.User.get_by_username(parts['username'])
+    return get_user(request.authenticated_userid, request)
 
 
 def includeme(config):
--- a/h/accounts/__init__.py
+++ b/h/accounts/__init__.py
@@ -45,11 +45,17 @@ def authenticated_user(request):
         return None
 
     try:
-        username = util.split_user(request.authenticated_userid)['username']
+        parts = util.split_user(request.authenticated_userid)
     except ValueError:
         return None
 
-    return models.User.get_by_username(username)
+    auth_domain = request.registry.settings.get(
+        'h.auth_domain', request.domain)
+
+    if parts['domain'] != auth_domain:
+        return None
+
+    return models.User.get_by_username(parts['username'])
 
 
 def includeme(config):
--- a/h/accounts/__init__.py
+++ b/h/accounts/__init__.py
@@ -1,4 +1,5 @@
 # -*- coding: utf-8 -*-
+from h import util
 from h.accounts import models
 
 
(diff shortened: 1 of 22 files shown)


h/accounts/test/views_test.py
```diff
@@ -830,30 +829,28 @@ def test_activate_event_when_validation_succeeds(event,
 
 
 @pytest.mark.usefixtures('subscriptions_model')
-def test_profile_looks_up_by_logged_in_user(authn_policy, user_model):
-    """
-    When fetching the profile, look up email for the logged in user.
+def test_profile_looks_up_by_logged_in_user():
```
nickstenning: I'd just remove this test. It's no longer testing a collaboration with another component.

h/accounts/__init__.py
```diff
@@ -34,8 +35,25 @@ def make_staff(username):
...
+def authenticated_user(request):
+    """Return the authenticated user or None.
+
+    :rtype: h.accounts.models.User or None
+
+    """
+    if not request.authenticated_userid:
+        return None
+
+    # pylint: disable=unpacking-non-sequence
```
nickstenning: Hmm. Pylint is right -- `util.split_user` can return None, so we should probably check for that eventuality.

h/accounts/__init__.py
```diff
@@ -34,8 +35,25 @@ def make_staff(username):
...
+    username, _ = util.split_user(request.authenticated_userid)
```
nickstenning: I'd like us to check that the second half of this tuple matches `request.domain`.

h/auth.py
```diff
@@ -136,9 +137,14 @@ def groupfinder(userid, request):
     """
     principals = set()
 
-    user = models.User.get_by_userid(userid)
+    username, _ = util.split_user(userid)
+    if username is None:
```
nickstenning: Once this implements all the same checks as `h.accounts.authenticated_user`, it's going to be a fair bit of duplication. Again, I'd suggest implementing the logic in a function which accepts a userid, not a request, and returns a user object or None, and put that in `h.accounts`. That function can then be used by this module, and by the function that you use as a request method:

##### `accounts/__init__.py`:
```python
def get_user(userid):
    ...

def includeme(config):
    config.add_request_method(lambda req: get_user(req.authenticated_userid), 
                                                   name='authenticated_user', 
                                                   reify=True)
```

##### `auth.py`:
```python
from h import accounts

def groupfinder(userid, request):
    ...
    user = accounts.get_user(userid)
    ...
```

h/accounts/__init__.py
seanh: Wait, I thought you wanted us to check that when `request.authenticated_userid` gets set, though? Was gonna do that separately

h/accounts/test/views_test.py
seanh: It's testing that the new `.authenticated_user` property gets called, previously it was testing that `get_by_userid()` got called, it's equivalent no?

nickstenning: It was testing that the correct arguments were passed to `get_by_userid`. `.authenticated_user` has no signature so there's no collaboration to test. Put another way, if this function passed all the other tests, but failed this one because you'd refactored it to get the user another way, that would be silly.

h/accounts/__init__.py
nickstenning: `request.authenticated_userid` is set by the authentication policy, which is the pyramid default, `pyramid.authentication.SessionAuthenticationPolicy`. You could provide a custom authentication policy which ensured the userid was in the right domain if you want. That might not be a bad idea. Regardless, any code which takes a string userid and turns it into a user object should ensure that it's not just throwing away a domain which doesn't match.

I promise I'm not this retentive about everything. But for security components... better safe than sorry.

h/accounts/test/views_test.py
seanh: Done

h/accounts/__init__.py
seanh: Done, I changed it to raise an exception: https://github.com/hypothesis/h/commit/a2025ccac881ed4e8990c9e5ffa1a8241445ea4f

h/accounts/__init__.py
seanh: Done, and also added `request.auth_domain` to avoid duplication: <https://github.com/hypothesis/h/commit/c22efde798c59039cb3dd2313dee54a33abdaf0b>, <https://github.com/hypothesis/h/commit/f5d54b9627945186f362cb325766512f12ccb515>

h/auth.py
seanh: Done

h/accounts/test/__init___test.py
```diff
@@ -85,3 +85,77 @@ def test_make_staff_raises_if_user_does_not_exist(get_by_username):
...
     with pytest.raises(accounts.NoSuchUserError):
         accounts.make_staff("fred")
+
+
+# The fixtures required to mock all of user()'s dependencies.
+authenticated_user_fixtures = pytest.mark.usefixtures('util', 'get_by_username')
+
+
+@authenticated_user_fixtures
+def test_authenticated_user_returns_None_if_authenticated_userid_is_None():
```
nickstenning: These tests should probably be testing `get_user`, not `authenticated_user`.

h/auth.py
```diff
@@ -42,6 +42,8 @@
 
 from .interfaces import IClientFactory
 from .oauth import JWT_BEARER
+from h import accounts
+from h import util
```
nickstenning: Unused import here.

nickstenning:
Fix the tests and the import and we're merged :)

seanh:
Done

nickstenning:
:space_invader: Thank you!

PR #2636, opened 2015-10-19
Replace client-side profile form

nickstenning:
This PR moves the last major piece of sidebar form handling -- namely the forms for changing email, password, and notification settings -- to the server, removing a fair chunk of client complexity, and opening the way to dramatic improvements to the form appearance and behaviour in the future.

Unlike the previous changes I've made to form-handling, these are not pixel-for-pixel translations of the forms that existed in the sidebar, for a couple of different reasons:

- The profile update form no longer has a delete account button. The delete account functionality doesn't really do anything: it sets a random password and logs the user out. We haven't removed any of their personally identifying information, and we can't even tell by looking at the database who has attempted to delete their account. We've [agreed to remove this button](https://hypothes-is.slack.com/archives/design/p1445270969000256) until such time as we can implement the feature properly.

  In its place is a paragraph inviting users who wish to delete their accounts to email support.

- The wording and layout of the notifications update form has been tweaked to read more clearly.

There is one concern I have about this as it stands, which is that there's little to no confirmation on successful submission of these forms, as both simply take you back to the form, and are fast enough that it's difficult to spot that the form submission succeeded. I want to work with @robertknight and @conordelahunty to agree an interim solution which improves this.

Please do check this out to see how the new forms behave. Reviewing is probably easier a commit at a time. @conordelahunty perhaps we can sit and go through this tomorrow or Wednesday?


robertknight:
Some UX issues from initial testing:

**1. The 'Account settings' link from the client needs to be updated**

Account -> Account Settings in the app shows a blank form rather than opening the settings page.

account-settings

**2. Scary content in tooltips and validation error messages**

field-tooltip

validation-error

nickstenning:
1. That was a rebase error -- I didn't notice that there are now two versions of the top bar!
2. I'm pretty sure you have "Show autofill predictions" turned on at chrome://flags/ -- that's nothing to do with this change.

robertknight:
These forms are currently bundling `app.css` which consists of all the styling used in the app itself. It should use `site_css` instead which is the new root stylesheet for the website.

To do that, you'll need to import `forms` into the `site.scss`.

This can be done as a separate PR if it turns out to be non-trivial.

robertknight:
>  to the server, removing a fair chunk of client complexity, and opening the way
> to dramatic improvements to the form appearance and behaviour in the future.

Aside from having more space than we had in the sidebar for the form, what other improvements did you have in mind?

Something that concerns me with this direction is that we'll have two very different ways of writing pieces of the front-end depending on whether we're dealing with the website or the sidebar - different templating languages, widgets etc. We can re-use CSS but that's about it. The result of that could be that moving functionality from the app to the site or vice-versa becomes a complete rewrite. I'm thinking about this in the context of something you mentioned yesterday about amount of work involved in presenting annotation cards on the groups landing page, if we had wanted to do that, for example.

This isn't a blocker to merging these changes now, and I don't expect to have a concrete plan about where to go in future at the moment, simply because we need a better understanding of what functionality we're likely to add to the site and sidebar in the coming months before making any technical calls. However we should definitely think through this in the November meetup.

h/accounts/views.py
```diff
@@ -465,116 +404,111 @@ def _register(self, username, email, password):
         self.request.registry.notify(RegistrationEvent(self.request, user))
 
 
+@view_defaults(route_name='profile',
+               renderer='h:templates/accounts/profile.html.jinja2')
+@view_config(attr='profile_form', request_method='GET')
+@view_config(attr='profile', request_method='POST')
```
seanh: I've never understood this pattern in h of putting `@view_config`s on the class. It just means they're far away from the method they refer to. The way the docs show to do this is with `@view_defaults` on the class but `@view_config` on the methods: <http://docs.pylonsproject.org/projects/pyramid/en/1.5-branch/narr/viewconfig.html#view-defaults-class-decorator>. I guess there's an argument to be made for following the existing style in h though.

h/accounts/test/views_test.py
```diff
@@ -829,232 +727,280 @@ def test_activate_event_when_validation_succeeds(event,
...
-        GET={'subscription_id': 'subscription_id'}
-    )
+@notifications_fixtures
+def test_notifications_form_with_valid_data_updates_subscriptions(authn_policy,
+                                                                  subscriptions_model):
+    request = DummyRequest(post={})
+    authn_policy.authenticated_userid.return_value = 'fiona'
+    subs = [
+        FakeSubscription('reply', True),
+        FakeSubscription('foo', False),
```
robertknight: Do we validate that the subscription type one from a predefined set anywhere?

robertknight: ie. Is the subscription type required to be a value from an enum or can it be arbitrary?

h/accounts/test/views_test.py
```diff
@@ -829,232 +727,280 @@ def test_activate_event_when_validation_succeeds(event,
...
-def test_unsubscribe_not_authorized(Subscriptions):
-    """If you try to unsubscribe someone else's subscription you get a 401."""
-    Subscriptions.get_by_id.return_value = Mock(
-        uri='acct:bob@hypothes.is', active=True)
-    request = MagicMock(
-        authenticated_userid='acct:fred@hypothes.is',
-        GET={'subscription_id': 'subscription_id'}
-    )
+@notifications_fixtures
+def test_notifications_form_with_valid_data_updates_subscriptions(authn_policy,
```
robertknight: There are two tests with the same name `test_notifications_form_with_valid_data_updates_subscriptions`. The second test doesn't appear to actually check if the subscription is updated.

seanh:
This error message seems odd to me (I'm already logged in):

screenshot from 2015-10-21 12 08 49

seanh:
I think that for _confirm_ fields (this would apply to the email one as well) we should _only_ check whether they're the same as the other field. There's no point in correcting errors like this one, unless you happen to correct it to be exactly the same as the other field:

screenshot from 2015-10-21 12 14 12

I also wouldn't show the "emails/passwords don't match error" unless the first email/password passed all other validation errors

h/templates/accounts/profile.html.jinja2
```diff
@@ -0,0 +1,50 @@
...
+      </h2>
+      {{ password_form | safe }}
+
+      <h2 class="form-heading">
+        <span>{% trans %}Delete account{% endtrans %}</span>
+      </h2>
+      <legend>
+        {% trans %}
+        If you would like to delete your account, please email us at <a
+        href="mailto:support@hypothes.is">support@hypothes.is</a> from your
```
seanh: For me this is opening up an email compose window to `///support@hypothes.is`, though I'm not sure why, the `href` looks correct to me

seanh:
The code looks good to me - another huge simplification. The UI concerns I raised can be dismissed as not changes from the current behaviour if you want. I'm not seeing the scary content in tooltips and error messages that Robert's getting.

seanh:
Just quickly tested with groups disabled - top bar link seems to work fine now.

seanh:
On the notifications tab, it'd be nice if the _Save changes_ button was disabled until I have actually made some changes to be saved. Also this button could just be _Save_ maybe. But better - just submit the form when I click on the checkbox? Then you don't need a _Save_ button.

h/accounts/test/views_test.py
nickstenning: Whoops! Fixed.

nickstenning:
Rebased onto the `get_by_userid` changes now in master.

h/accounts/views.py
nickstenning: Yep, happy to change this but I'm not going to do it in this already-huge PR.

h/accounts/test/views_test.py
nickstenning: In practice there's only one type. You can POST whatever you like to the form handler endpoint and it'll just ignore subscription types it doesn't recognise.

robertknight:
The scary tooltips I mentioned were indeed due to the `chrome://flags` setting that Nick mentioned.

From my side, this looks good except for:
 * Try to use site_css rather than app_css for the page if we can
 * The duplicated test I mentioned

We had a fairly lengthy discussion on Slack around my earlier comments about sharing components between the sidebar and the site. See https://hypothes-is.slack.com/archives/public/p1445421946003215 . There was a consensus that we want to focus the sidebar on its core purpose of annotations and make parts of the site that don't require rich functionality to be much lighter-weight. We also need to establish clear user stories and metrics around the experience without JS, with a slow network connection, with a screen reader etc. so that these changes form part of a bigger story.

nickstenning:
That's great feedback, both, thank you:

- [ ] Use `site.css` rather than `app.css`
- [ ] Only show "doesn't match" errors on the confirmation fields.
- [ ] Fix "email already in use" validation message.

As for the "Save changes" button on the notifications page, I think we should save that for a progressive enhancement we can make to that form at a later date.

seanh:
> As for the "Save changes" button on the notifications page, I think we should save that for a progressive enhancement we can make to that form at a later date.

Happy with that (we'd need to us js to make the checkbox auto submit when checked, right?)

nickstenning:
> Happy with that (we'd need to us js to make the checkbox auto submit when checked, right?)

Yep.

nickstenning:
I've added some confirmation messages for when these forms are successfully submitted. Pending approval of these from Conor (awaited in Slack) this is good to go, I think.

screen shot 2015-10-21 at 16 41 14

judell:
From the referenced chat: "The understanding I think we're slowly coming to is that (broadly) if it's not to do with a specific page I'm annotating, it shouldn't be in the sidebar." 

+1 from me and I think from @conordelahunty as well.

"Something that concerns me with this direction is that we'll have two very different ways of writing pieces of the front-end depending on whether we're dealing with the website or the sidebar"

Indeed. We have enough such duplication elsewhere. Sounds like you're walking down this path with eyes open, though. And if we've got the right Occam's razor for deciding what's in or out of the sidebar, then out-migration from the sidebar should be a one-time event with little or no subsequent boundary-crossing.

PR #2638, opened 2015-10-20
Extension badge refactor

robertknight:
This PR refactors and simplifies the BrowserAction class in two ways, in preparation for a future PR to customize the badge depending on the extension build type (dev, staging, production).

* Makes it actually stateless (as the documentation claims it should be). Its responsibility is just to take the logical H state of a tab (active/inactive/errored, annotation count) and update the badge's label, tooltip and text. Previously it didn't actually store any state directly in BrowserAction but it used to instead read state by querying the Chrome browser action API, which made it effectively stateful. Having state flow only in one direction (H model -> badge) will make it easier to add new things which might affect the badge's ultimate appearance and also re-use the code across extensions which use different APIs to update the badge.

* Separates out the logic for querying the badge count and updating the model state from the logic that updates the presentation of the badge. This simplifies a bunch of tests.

### DIFF
--- a/h/browser/chrome/test/browser-action-test.js
+++ b/h/browser/chrome/test/browser-action-test.js
@@ -1,3 +1,5 @@
+var assign = require('core-js/modules/$.assign');
+
 describe('BrowserAction', function () {
   'use strict';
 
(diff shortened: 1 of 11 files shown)


h/browser/chrome/test/browser-action-test.js
```diff
@@ -8,341 +10,134 @@ describe('BrowserAction', function () {
...
-      var originalGetBadgeTextFunc = fakeChromeBrowserAction.getBadgeText;
-
-      fakeChromeBrowserAction.getBadgeText = function(args, func) {
-        func('some badge text that is already showing');
-      };
-      action.updateBadge("tabId", "tabUrl", "http://example.com");
-      fakeChromeBrowserAction.getBadgeText = originalGetBadgeTextFunc;
-
-      assert(fakeChromeBrowserAction.setTitle.notCalled);
+      action.update(1, {
```
nickstenning: Hahaha. Nice.

nickstenning: (In reference to how much simpler these tests became).

nickstenning:
LGTM. Checked it out and built it -- all seems fine and dandy.

PR #2640, opened 2015-10-20
Extract sort dropdown into own component

robertknight:
Extract the sort dropdown into a reusable component
and add tests in preparation for moving it from the annotation list into the top bar later.

### DIFF
--- a/h/static/scripts/directive/sort-dropdown.js
+++ b/h/static/scripts/directive/sort-dropdown.js
@@ -1,0 +1,11 @@
+module.exports = function () {
+  return {
+    restrict: 'E',
+    scope: {
+      sortBy: '=',
+      sortOptions: '=',
+      onChangeSortBy: '&',
+    },
+    templateUrl: 'sort_dropdown.html',
+  }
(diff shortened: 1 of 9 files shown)


h/static/scripts/directive/sort-dropdown.js
```diff
@@ -0,0 +1,11 @@
+module.exports = function () {
+  return {
+    restrict: 'E',
+    scope: {
+      sortBy: '=',
+      sortOptions: '=',
+      onChangeSortBy: '&',
```
nickstenning: Is this necessary? `=` is a two-way binding, so shouldn't you simply be able to assign to `sortBy` within the directive?

robertknight: @nickstenning - I want a one-way binding here for data-in, events-out. Unfortunately Angular lacks convenient syntax for that in Angular 1.x.

nickstenning: Ok, no problem.

PR #2643, opened 2015-10-20
Move sort dropdown to top bar

robertknight:
This depends/builds on https://github.com/hypothesis/h/pull/2640 . For groups-enabled users, this moves the filter dropdown to the top bar.

For all users, there are visual changes to the dropdown menu. The heights have been made consistent across all menus and the radio button indicator is shown for everyone as well. To accommodate the wider menu, the dropdown has been re-aligned for non-groups users.

**Non-groups users**

non-groups-dropdown

**Groups users**

groups-enabled-dropdown


conordelahunty:
@robertknight Can you make the font sizes in that "Groups users" menu be the same size as they are in the original "Non-groups users" screenshot.

That icon is weird, it shouldn't have that stroke issue. I amended it 20mins or so after I first uploaded them to the trello card, did you grab it before then?

robertknight:
Discussed the font size issue - the change in font size is because the menu items in the top bar dropdowns all need to be consistent with one another. The text size was revised down slightly form those screenshots from 15px to 14px.

nickstenning:
Looks good, although I have one question. The choice of what to display (both the sort control and the overall design of the sidebar) doesn't seem to respect the 'groups' feature flag if you're logged out. It's perfectly possible to have the flag set to 'true' if you're logged out, indeed that's what releasing the groups feature will look like.

Specifically for this PR, if I'm logged out and the 'groups' feature is toggled on, I get the old-style top bar but I don't see a sort control.

So I think there's some work to do, either in this PR or a subsequent one, to ensure that the things that are supposed to be toggled by the 'groups' feature flag are always toggled, whether you're logged in or not.

robertknight:
@nickstenning Thanks for raising that. It needs looking into. I can put a card for this in up next and handle it as part of a separate PR?

nickstenning:
I've rebased this for you and will be merging it if it goes green...

PR #2650, opened 2015-10-21
Set group and visibility of reply to that of parent

seanh:
Set the group and visibility of replies to that of their parents, when
replies are first created.

Also prevent the visibility setting of a reply from overwriting the
cached-in-local-storage visibility level. This prevents the following
problem:

- I'm annotating publicly (my cached-in-local-storage visibility
  setting is public)
- I reply to a private annotation
- The visibility of my reply is set to that of its parent, private
- This visibility is then cached in local storage
- I make a new top-level annotation, and now it's defaulting to private
  instead of public, even though the user never did anything to change
  the visibility from public to private, it was done automatically
  because they replied to a private annotation.

Obviously changing the mode from private to public when replying to a
public annotation was also possible.

This means that if the user _does_ change the visibility of a reply, we
_don't_ cache that either:

- I'm annotating publicly (my cached-in-local-storage visibility
  setting is public)
- I reply to a public annotation
- The visibility of my reply is set to that of its parent, public
- This visibility is not cached in local storage because it's a reply
  (and the value cached in the local storage is already the same anyway)
- I then deliberately change the visibility of my reply from public to
  private
- Again this visibility is not cached in local storage because it's a reply
- The next time I make a new top-level annotation, it will still be
  defaulting to public, will not have changed to private mode

### DIFF
--- a/h/static/scripts/directive/test/annotation-test.coffee
+++ b/h/static/scripts/directive/test/annotation-test.coffee
@@ -26,6 +26,7 @@ describe 'annotation', ->
   fakeTags = null
   fakeTime = null
   fakeUrlEncodeFilter = null
+  fakeLocalStorage = null
   sandbox = null
 
   createDirective = ->
@@ -87,6 +88,10 @@ describe 'annotation', ->
       nextFuzzyUpdate: sandbox.stub().returns(30)
     }
     fakeUrlEncodeFilter = (v) -> encodeURIComponent(v)
+    fakeLocalStorage = {
+      setItem: sandbox.stub()
+      getItem: sandbox.stub()
+    }
 
     fakeGroups = {
       focused: -> {}
@@ -108,6 +113,7 @@ describe 'annotation', ->
     $provide.value 'tags', fakeTags
     $provide.value 'time', fakeTime
     $provide.value 'urlencodeFilter', fakeUrlEncodeFilter
+    $provide.value 'localStorage', fakeLocalStorage
     $provide.value 'groups', fakeGroups
     return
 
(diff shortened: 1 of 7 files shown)


seanh:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;YJNH2Lop&#x2F;143-when-i-reply-to-an-annotation-my-reply-should-go-to-the-same-group-and-have-the-same-visibility"> When I reply to an annotation, my reply should go to the same group and have the same visibility</a>

seanh:
This is done, who wants to review it? @nickstenning @robertknight

seanh:
This should also fix the problem with replying in the stream. No matter what group you had focused last time you used the sidebar, replies in the stream should get the same group as their parents

nickstenning:
This looks great. Nice and clean. One minor thing: the `publish-annotation-btn` directive no longer needs an `isNew` param, so that can be removed (from the directive, its tests, and `annotation.html`).

h/static/scripts/directive/test/annotation-test.coffee
```diff
@@ -691,6 +751,44 @@ describe("AnnotationController", ->
     it("creates the directive without crashing", ->
       createAnnotationDirective({})
     )
+
+    it("sets the permissions of new annotations from local storage", ->
```
robertknight: here are only two cases here so it doesn't make much difference, but in general I prefer that when you have several tests with very similar logic, they be expressed as:

```
var cases = [...]
cases.forEach(function (case) {
it('does something for case ' + case.name, function () {
  // create variant using the case
});
}
```

robertknight:
LGTM - Just the trivial comment nick mentioned.

PR #2657, opened 2015-10-22
Fix extension clobbering embedded H

robertknight:
This PR refactors the way that the extension determines whether or not it needs to install the sidebar into the current tab and fixes an issue where the extension would automatically remove embedded H on a page even when the user did not interact with it.

The approach is to introduce two additional pieces of per-tab state in `TabState`. `extensionSidebarInstalled` tracks whether the sidebar was installed by the extension into this tab and `ready` tracks whether the tab is ready for the extension to be loaded.

Along the way, I discovered that the documentation in `TabStore` about persisting H state across browser sessions is completely wrong because it uses tab IDs as keys and those are not persistent across sessions. The only use case for `TabStore` currently therefore is persisting H tab state across extension updates and re-installation. I have simplified it and kept it doing that for now.

Another user-visible side-effect of this refactoring is that the badge state can now be updated as soon as a new page load starts, even though the actual injection of the sidebar does not happen until the tab is fully loaded. The result is that badge numbers now appear significantly faster when opening a new page. Actual injection of the sidebar remains deferred until the tab, including _all_ external resources are fully loaded. Ideally we would inject the sidebar as soon as the page receives the 'DOMContentLoaded' event and it would show up much faster. That can wait for a separate PR however.

### DIFF
--- a/h/browser/chrome/lib/hypothesis-chrome-extension.js
+++ b/h/browser/chrome/lib/hypothesis-chrome-extension.js
@@ -107,7 +107,7 @@ function HypothesisChromeExtension(dependencies) {
     });
   };
 
-  function onTabStateChange(tabId, current, previous) {
+  function onTabStateChange(tabId, current) {
     if (current) {
       browserAction.update(tabId, current);
 
@@ -197,6 +197,8 @@ function HypothesisChromeExtension(dependencies) {
     state.clearTab(tabId);
   }
 
+  // installs or uninstalls the sidebar from a tab when the H
+  // state for a tab changes
   function updateTabDocument(tab) {
     // If the tab has not yet finished loading then just quietly return.
     if (tab.status !== TAB_STATUS_COMPLETE) {
--- a/h/browser/chrome/lib/hypothesis-chrome-extension.js
+++ b/h/browser/chrome/lib/hypothesis-chrome-extension.js
@@ -81,9 +81,9 @@ function HypothesisChromeExtension(dependencies) {
     chromeTabs.query({}, function (tabs) {
       tabs.forEach(function (tab) {
         if (state.isTabActive(tab.id)) {
-          state.activateTab(tab.id, {force: true});
+          state.activateTab(tab.id);
         } else {
-          state.deactivateTab(tab.id, {force: true});
+          state.deactivateTab(tab.id);
         }
       });
     });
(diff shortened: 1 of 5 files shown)


dwhly:
> Ideally we would inject the sidebar as soon as the page receives the 'DOMContentLoaded' event and it would show up much faster.

I have asked about that before, and had always thought it wasn't possible.  Incredible if so!

h/browser/chrome/lib/hypothesis-chrome-extension.js
```diff
@@ -162,20 +187,33 @@ function HypothesisChromeExtension(dependencies) {
...
+      return sidebar.injectIntoTab(tab)
+        .catch(function (err) {
+          tabErrors.setTabError(tab.id, err);
+          state.errorTab(tab.id);
+        });
     }
-    else if (state.isTabInactive(tab.id)) {
-      return sidebar.removeFromTab(tab);
+    else if (state.isTabInactive(tab.id) && isInstalled) {
+      return sidebar.removeFromTab(tab).then(function () {
```
seanh: Why do we do `removeFromTab()` in this `updateTabDocument()` function anyway? We already do it elsewhere when the browser action is clicked

robertknight: This is the only place that `removeFromTab()` is called. Conceptually the approach is that when you click the browser action, a page load completes or some other event happens it changes the H state of the tab. This function is called in response to an H tab state change and if the tab is inactive but the sidebar is currently installed then it will be removed.

robertknight:
> I have asked about that before, and had always thought it wasn't possible. Incredible if so!

Its fiddly but it can be done. For future reference - https://productforums.google.com/forum/#!topic/chrome/xy6JEAZrjFg

h/browser/chrome/lib/hypothesis-chrome-extension.js
```diff
@@ -50,43 +51,33 @@ function HypothesisChromeExtension(dependencies) {
...
   /* Sets up the extension and binds event listeners. Requires a window
    * object to be passed so that it can listen for localStorage events.
    */
   this.listen = function (window) {
     chromeBrowserAction.onClicked.addListener(onBrowserActionClicked);
     chromeTabs.onCreated.addListener(onTabCreated);
+
+    // when a user navigates within an existing tab,
+    // onUpdated is fired in most cases
     chromeTabs.onUpdated.addListener(onTabUpdated);
```
seanh: Think this comment would be good as a jsdoc comment on the `onTabUpdated()` function.

h/browser/chrome/lib/hypothesis-chrome-extension.js
```diff
@@ -50,43 +51,33 @@ function HypothesisChromeExtension(dependencies) {
...
-      if (isState && isUpdate && event.newValue !== JSON.stringify(store.all())) {
-        store.reload();
-        state.load(store.all());
-      }
-    });
+    // ... but when a user navigates to a page that is loaded
+    // via prerendering or instant results, onTabReplaced is
+    // fired instead. See https://developer.chrome.com/extensions/tabs#event-onReplaced
+    // and https://code.google.com/p/chromium/issues/detail?id=109557
+    chromeTabs.onReplaced.addListener(onTabReplaced);
```
seanh: This one too

seanh:
Wait, is `ready` actually used anywhere?

robertknight:
`Wait, is ready actually used anywhere?` good question - it is currently used implicitly in `updateTabDocument` in that it triggers invocation of `onTabStateChanged` when a page load completes, causing the sidebar to be injected if the user activated the tab whilst the page was loading.

It would make sense to check it explicitly in `updateTabDocument` I think.

seanh:
Yeah, currently if I make it so that `ready` is always `false`, then I can still inject the sidebar. So it seems to make sense to check it explicitly.

But before we didn't have this ready variable at all. Why do we need it now? We're trying to prevent h from being injected before the page is fully loaded?

h/browser/chrome/lib/hypothesis-chrome-extension.js
robertknight: I prefer to keep jsdoc comments for public API and inline '//' comments for implementation notes. It lets you see at a glance what you need to know if you're using the code vs. hacking on it.

robertknight:
> But before we didn't have this ready variable at all. Why do we need it now?

I originally added it so that a state change would occur (and thus `onTabStateChange` would be triggered) going from a state where the tab was active but not yet loaded (and thus the sidebar cannot yet be installed) to a state where it is active and loaded.

It does also have the benefit though of making it easier to change in future when we consider a tab to be sufficiently loaded to inject the sidebar and thus load the sidebar quicker. See my comments at the end of the PR above.

PR #2661, opened 2015-10-22
Disable caching for features/ AJAX responses

### LINKED ISSUE #2658: Groups features not visible in IE, Edge
**Steps to reproduce**

Tested on a VM with Windows 10, IE 11 and Edge:

1. Go to a site with H embedded
2. Sign in as a user who has the groups feature enabled
3. Refresh the page

**Expected result**

Groups features and design changes visible in UI

**Actual**

Groups features and design changes not visible.
I tested on the same VM with the same user and Firefox and was able to access groups features.

robertknight:
This fixes an issue where the response for /app/features would
be cached by IE 10, 11 resulting in the groups features not
appearing after signing in to a groups-enabled account
and refreshing the page.

I've verified that this fixes the problem I saw but haven't evaluated where other calls might be affected in the same way yet and if so, is there a more generic configuration to disable caching for AJAX responses.

Fixes #2658

### DIFF
--- a/h/features.py
+++ b/h/features.py
@@ -100,7 +100,8 @@ def flag_enabled(request, name):
 @view_config(route_name='features_status',
              request_method='GET',
              accept='application/json',
-             renderer='json')
+             renderer='json',
+             http_cache=0)
 def features_status(request):
     """Report current feature flag values."""
     return {k: flag_enabled(request, k) for k in FEATURES.keys()}


robertknight:
Related - http://stackoverflow.com/questions/32261000/how-to-avoid-ajax-caching-in-internet-explorer-11-when-additional-query-string-p .

Should we add `Cache-Control: no-cache` to all `application/json` responses by default?

robertknight:
Slack discussion: https://hypothes-is.slack.com/archives/public/p1445600110003769

nickstenning:
Okay, so here `http_cache=0` goes way over the top, adding the following headers:

```
Cache-Control: max-age=0, must-revalidate, no-cache, no-store
Pragma: no-cache
Expires: ...
```

This is pretty silly, as `max-age=0, must-revalidate` == `no-cache`, and there's no need to specify `no-store`, really.

So perhaps change this to `http_cache=(0, {'no_store': False})` and we'll merge it.

I suspect in future we may wish to write our own view decorator in order to have a bit more control over C-C values.

nickstenning:
> Should we add Cache-Control: no-cache to all application/json responses by default?

I don't think that's necessary at this point. Almost all calls to `/api` URLs made by the client will include an `Authorization` header, which should mean that IE won't cache them. We may, however, need to to add a similar declaration to the `/app` view.

robertknight:
Separate to this PR, I wonder if it still makes sense to have separate `/app` and `/app/features` endpoints, given that we'll want the feature data at the same time as the other session data.

robertknight:
> Almost all calls to /api URLs made by the client will include an Authorization header, which should mean that IE won't cache them.

So loading up H at the moment, when signed in, I see the following requests:

`/api` - No Authorization header in request, no cache control headers in response
`GET /app/features` - No Authorization header in request, no cache control headers in response (without this PR)
`GET /app` - No Authorization header in request, Cache-Control "max-age=0, must-revalidate, no-cache, no-store" in response
`GET /api/token` - No Auth header in request (obviously), no cache-control in response
`GET /api/search` - Auth header in request, no cache-control in response (not needed)

I'm going to check which of those IE is apparently caching atm.

robertknight:
> We may, however, need to to add a similar declaration to the /app view.

Ah, the `/app` endpoint is already declared with:

`@json_view(route_name='session', http_cache=0)`

So it already includes all the cache-control headers mentioned earlier. Would it make sense for `/features` to use `@json_view` and then move the caching controls into `@json_view`?

nickstenning:
Merged as 882648a. I don't think coupling output format to cache behaviour at the moment is a good idea, so let's keep `json_view` as is for now. We will need to revisit caching across the application in the near future anyway.

PR #2662, opened 2015-10-22
Replace hashids with pubids

nickstenning:
This PR removes all use of the `Group.hashid` property, and then removes all the supporting infrastructure for hashids. It replaces hashids with a column called `pubid` on the group table, which is filled by default using the output of the `h.pubid:generate` function.

For posterity, it's worth noting the reason for this change: at the time of writing, we are using hashids for two reasons:

1. To prevent users from guessing the URL for the "join a group" pages of groups they have not been invited to.

2. To avoid leaking details of the how many groups have been created through the use of an integer primary key in URLs.

Both of these are ultimately security requirements, and the cryptographic properties of hashids are [questionable at best](https://carnage.github.io/2015/08/cryptanalysis-of-hashids/).

In order to avoid the possibility of having to remove hashids at a later date, thus breaking many URLs, we are adopting a much simpler, much more "stupid" (i.e. less "clever"), way of generating this "public IDs", which now relies for its cryptographic properties on the system CSPRNG[^1] rather than the algorithm or implementation of hashids. And, frankly, if /dev/urandom is broken, we have bigger problems...

[^1]: "Cryptographically Secure Pseudo-Random Number Generator". See also: https://en.wikipedia.org/wiki/Cryptographically_secure_pseudorandom_number_generator


nickstenning:
This is dependent on #2660 being merged first.

nickstenning:
Well that was fast! Rebased.

nickstenning:
For the migration that adds pubids to existing groups, I've taken the opportunity to write an example of a combined schema and data migration that's safe to run against a live database. Specifically, that means that it never locks the whole table for any operation that needs to read or write data from every row in that table.

For the case of the groups table, right now, this is almost certainly overkill. Given that only our own staff are using it, the impact of a full-table lock (such as would be taken if we added the 'pubid' column as NOT NULLABLE, or we created the unique constraint without setting `postgresql_concurrently=True`) would be minimal. However, it can serve as an example for future migrations that need to touch, say, the user table.

nickstenning:
I don't imagine anyone was going to sprint to merge this, but hold on just in case. I want to double check my maths w.r.t. collision probabilities for the pubids.

nickstenning:
Unsurprisingly I had got my maths wrong. Now corrected. See the docstring in `h.pubid` for details.

nickstenning:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;SA3ICrMr&#x2F;142-replace-hashids-with-fixed-random-ids"> Replace hashids with fixed random ids</a>

robertknight:
> Both of these are ultimately security requirements, and the cryptographic properties of hashids are questionable at best.

The hashids.org site itself puts it much more bluntly: https://hypothes.is/a/atAtQnBjRC6F3Pm_nLbFeA

h/migrations/versions/4ed50e7ac940_add_pubid_column_to_group_table.py
```diff
@@ -0,0 +1,61 @@
...
+    sa.column('pubid', sa.Text),
+)
+
+
+def upgrade():
+    bind = op.get_bind()
+    session = Session(bind=bind)
+
+    # Add the pubid column to the group table, but add it as nullable and
+    # without the unique constraint. This is an O(1) operation.
```
robertknight: What reference are you using the costs of various DDL operations?

robertknight:
LGTM - all pretty clear and sensible. @seanh - Did you have any comments?

From a security POV, you have to consider not only the probability of a collision when creating a new group but also the number of requests that a user would have to make to find a valid group ID which they had not been invited to via brute force. That number is very high (~6 million gives a 90% probability of finding a match). It is much more likely that the URL leaks via another means.

h/migrations/versions/4ed50e7ac940_add_pubid_column_to_group_table.py
nickstenning: Some combination of [the PostgreSQL docs](www.postgresql.org/docs/9.4/static/sql-altertable.html), a couple of great talks about PostgreSQL migrations ([1](russ.garrett.co.uk/talks/postgres-gds/), [2](https://vimeo.com/83756644)), and experience :stuck_out_tongue:.

nickstenning:
There's actually one thing I want to change before this merges. Currently we have one migration that does both schema and data migration, but that will break production because we're adding a non-nullable field for which production won't be setting defaults. I need to split this into three (schema, data, schema) to fix this issue.

nickstenning:
Ok, that's now done. This should be good to go.

seanh:
Not that it really needed to be tested like this, but I ran `h.pubid.generate()` _lots_ of times and verified no collisions! (On Linux)

seanh:
And on a quick test groups seem to work, the db migration seems to work

seanh:
@nickstenning I take it we're not bothered about migrating existing annotations?

nickstenning:
@seanh That's a really good point. I think I'll migrate them, but I'll hack together a couple of scripts to do that separately from this PR. There's no good way of doing it unfortunately -- either we make the annotations disappear in prod when we deploy to stage, or we make the annotations disappear in stage until we've deployed to prod. The latter is probably what we'll have to do.

nickstenning:
And for reference, here's that migration script: https://gist.github.com/nickstenning/189735f51ce49341fea1

PR #2667, opened 2015-10-23
Trello 144 add a list of most recently annotated by the group urls to the group page

seanh:


seanh:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;Nj6M1pMF&#x2F;144-add-a-list-of-up-to-most-recently-annotated-by-the-group-urls-to-the-group-page"> Add a list of up to most recently annotated (by the group) URLs to the group page</a>

h/api/models.py
```diff
@@ -166,6 +176,76 @@ def title(self):
             return ""
 
     @property
+    def filename(self):
+        if self.uri and self.uri.startswith("file://"):
+            return self.uri.split("/")[-1] or None
+
+    @property
+    def document_link(self):
+        """Return a link to this annotation's document.
```
robertknight: It seems odd to try and render the HTML outside of the template. You could just always render the `<a>` tag in the markup and only include the `href` attribute if a URL is available.

h/api/test/models_test.py
```diff
@@ -78,6 +79,74 @@ def test_title_annotation_that_has_no_document():
...
+     'title (example.pdf)'),
+
+    # If the filename is the same as the title it shouldn't show it.
+    (('file:///home/bob/Documents/example.pdf', 'example.pdf'),
+     'example.pdf'),
+
+    # If the annotation has an http(s) uri but no document title it should
+    # return the uri linked to the uri.
+    (('http://example.com/example.html', None),
+     '<a href="http://example.com/example.html">http://example.com/example.html</a>'),
```
robertknight: Given the limited space for the displayed URL, I suggest we should strip uninformative parts such as the scheme and leading 'www.' from what is rendered.

h/groups/templates/share.html.jinja2
```diff
@@ -18,6 +18,18 @@
...
+
+      {% if document_links %}
+      Recently annotated documents:
+      <ul>
+        {% for document_link in document_links %}
+          <li>{{ document_link }}</li>
+        {% endfor %}
+      </ul>
+      {% else %}
+        <p>This group hasn't shared any annotations yet.</p>
```
robertknight: This sounds slightly odd. The group doesn't share annotations - people in the group do. "No annotations have been shared in this group yet"?

h/groups/views.py
```diff
@@ -91,7 +94,35 @@ def _read_group(request, group):
 
     """
     url = request.route_url('group_read', hashid=group.hashid, slug=group.slug)
-    template_data = {'group': group, 'group_url': url}
+
+    # TODO: Get shared annotations only.
+    result = search.search(request, params={
+        "group": group.hashid, "limit": 1000})
```
robertknight: This will need adjusting for Nick's hashid -> pubid conversion

h/groups/views.py
```diff
@@ -91,7 +94,35 @@ def _read_group(request, group):
...
+    annotations = result['rows']
+
+    # Group the annotations by URI.
+    annotations_by_uri = collections.OrderedDict()
+    for annotation in annotations:
+        if annotation.uri in annotations_by_uri:
```
robertknight: Does URI normalization need to be handled here or elsewhere?

h/api/models.py
seanh: Yeah, that would simplify things :) Main reason I'm doing this in Python instead of in Jinja2 is that the logic is complex enough that I want unit tests for it. But if we can put the `<a>` tag in either way and the only logic Jinja2 needs is to output the `href` only if there's something to link to, that should be fine

h/api/test/models_test.py
seanh: Good idea!

h/groups/views.py
seanh: I was thinking that was out of scope for this version, and that this whole thing will work much better once we have annotations and documents in a proper database, but I could probably just pass each uri through our `normalize()` function though

seanh:
Screenshot of how the page looks currently:

screenshot from 2015-10-26 11 14 49

seanh:
- [ ] Show the full document title in the tooltip

h/groups/templates/share.html.jinja2
seanh: I think I prefer my wording as it's simpler, and I think users will understand "the group" to mean both the group (i.e. the thing that exists in our database) and the users who're in that group (this seems like typical usage of the word group to me). I just prefer direct language, I'm sick of seeing needlessly indirect things like _this annotation is viewable only by you_ instead of _only you can see this annotation_ etc etc in Hypothesis.

h/api/models.py
seanh: Ok, having looked into this, I want to keep it how it is. I understand that rendering HTML tags in Python instead of in the template is weird, but:

- This Python function is actually pretty complicated logic, and it's nice to keep it all in one place where it can be tested
- This logic is duplicated in JavaScript code for the annotation cards and now in Python for the page. With the code in Python, we can potentially add the link to the annotation on its way out of our API (in our search lib's `render()` function) and have the client-side code just use it. The problem with this though is annotations that are being created but haven't been saved to the server yet.

If we wanted to move this into the template, I think we'd need four separate Python functions to get the href, title, link_text, and domain, and then template code that looked like this:

    <a {% if href %}href="{{ href }}"{% endif %} title={{ title }}>{{ link_text }}</a>{% if domain %}({{ domain }}){% endif %}

And then even if we added the href, title, link_text, and domain to the annotations in `render()` we'd still have to duplicate that bit of template code containing (admittedly simple) logic client-side.

And splitting this one function into four would not be that simple as the values are inter-dependent (although having four smaller functions would probably benefit the code).

I'm willing to be overridden here though

h/groups/views.py
seanh: Done this

h/groups/views.py
seanh: Done

seanh:
This is how this looks currently on production data:

screenshot from 2015-10-26 17 36 45

I'm just planning to add some tests now and then this is done.

I added `(hostname|filename)` after each annotation (if it has a hostname or filename that's different from the link text) because that's what we have in the sidebar/stream, and it was specifically requested in those places because it was useful info for some documents. It's not in Conor's mockup though, and it does sometimes wrap onto a new line.

I also added an arbitrary amount of vertical space between `<li>s` to make it a little more readable.

seanh:
Tests added and rebased

robertknight:
The initial view after you've just created a new group looks somewhat bare and the most important thing we want users to do next, share the group, is tucked down at the bottom of the page.

@conordelahunty - Thoughts?

empty-group-view

robertknight:
@seanh Couple of UX notes:

* I'm not sure if was on the design, but I generally avoid contractions ("hasn't") in UI text
* The link should be focused when the page loads so the user can just press Ctrl+C to copy or tap on mobile

seanh:
> I'm not sure if was on the design, but I generally avoid contractions ("hasn't") in UI text

I generally use them on purpose! Makes it more casual/friendly imo. But afaik we don't have any style guide for this stuff. We need one. Conor's not here this week either so we can't really defer to him on it. In this case I'll just go ahead and change it

seanh:
> The link should be focused when the page loads so the user can just press Ctrl+C to copy or tap on mobile

This already happens for me in Chrome and Firefox. Is it not happening for you?

seanh:
Changed hasn't -> has and rebased on latest master.

Shall I go ahead and add 1em of left and right padding to `.group-form`? I think it needs it, as some document titles go right up to the very edge of the white box

robertknight:
> Shall I go ahead and add 1em of left and right padding to .group-form?
> I think it needs it, as some document titles go right up to the very edge of the white box

I'm using `px` measurements for everything new as it is much easier to match up directly with the designs.

robertknight:
I can see room for a bunch of design tweaks but I think it is important to get this landed and give the others a chance to play around with it.

h/api/models.py
```diff
@@ -166,6 +170,112 @@ def title(self):
...
+        for this annotation's document.
+
+        User-supplied values are escaped so the string is safe for raw
+        rendering (the returned string is actually a jinja2.Markup object and
+        won't be escaped by Jinja2 when rendering).
+
+        """
+        uri = jinja2.escape(self.uri) or ""
+
+        if uri.startswith("http://") or uri.startswith("https://"):
```
robertknight: Is the uri normalized here so that the scheme is always lower-case?

seanh:
> > Shall I go ahead and add 1em of left and right padding to .group-form?
> > I think it needs it, as some document titles go right up to the very edge of the white box
> 
> I'm using `px` measurements for everything new as it is much easier to match up directly with the
> designs.

10px?

h/api/models.py
seanh: It's not, just pushed a fix and tests for this

seanh:
> I can see room for a bunch of design tweaks but I think it is important to get this landed and give the others a chance to play around with it.

Agreed

h/api/models.py
```diff
@@ -166,6 +170,112 @@ def title(self):
...
+        if self.title:
+            title = jinja2.escape(self.title)
+            link_text = title
+            if uri.startswith("file://"):
+                domain = jinja2.escape(self.filename)
+            else:
+                domain = urlparse.urlparse(uri).hostname
+        else:
+            if uri.startswith("file://"):
+                title = urllib2.unquote(jinja2.escape(self.filename))
```
robertknight: I'd strongly advise not mixing up logic for determining field values with logic for performing HTML escaping.

There is an HTML injection vulnerability here:

```
1. Create a file with the title `"><script>window.alert("Pwned");<%2Fscript>.html` and no content (especially, no <title>)
2. Service this via a local HTTP server and annotate it in a group
3. Refresh groups page
```

robertknight: The way I would approach this is to split the rendering into two parts - one which determines the different parts of the link (href, title etc.) and another which renders the appropriate link, escaping the values accordingly.

seanh: Is that injection really possible? I'm escaping the title, and I have a test for it

seanh: > The way I would approach this is to split the rendering into two parts - one which determines the different parts of the link (href, title etc.) and another which renders the appropriate link, escaping the values accordingly.

It's probably a good idea to break this up, as it's a long function

h/api/search/query.py
```diff
@@ -96,8 +96,18 @@ class AuthFilter(object):
...
+        :param private: whether or not to include private annotations in the
+            search results
+        :type private: bool
+
+        """
         self.request = request
+        self.private = private
 
     def __call__(self, params):
         groups = list(self.request.effective_principals)
```
robertknight: This initially confused me because 'groups' is not in fact, just a list of groups but a set of `principals` which can include both groups and user IDs. I suggest renaming the `groups` variable here accordingly (eg. to `principals`)

h/api/search/test/core_test.py
```diff
@@ -0,0 +1,30 @@
...
+search_fixtures = pytest.mark.usefixtures('query', 'models')
+
+
+@search_fixtures
+def test_search_passes_private_to_AuthFilter(query):
+    request = mock.Mock()
+
+    core.search(request, mock.Mock(), private=True)
+
+    query.AuthFilter.assert_called_once_with(request, True)
```
robertknight: Could this be written as `assert_called_once_with(request, private=True)`? If so, that would be clearer.

h/api/test/models_test.py
```diff
@@ -78,6 +79,108 @@ def test_title_annotation_that_has_no_document():
...
+    assert annotation.document_link == out
+
+
+def test_document_link_escapes_HTML_in_the_document_title():
+    spam_link = '<a href="http://example.com/rubies">Buy rubies!!!</a>'
+    annotation = models.Annotation(
+        uri='http://example.com/example.html',
+        document={'title': '</a>' + spam_link})
+
+    assert spam_link not in annotation.document_link
```
robertknight: This doesn't verify that the link was escaped properly, only that the injected link doesn't occur _exactly_ as in the input text. I'd suggest checking for the presence of any escaped chars - eg. by checking that `escape(input) == input`

h/static/styles/group-form.scss
```diff
@@ -73,6 +74,14 @@
     }
   }
 
+  .group-form-document-list {
+    text-align:center;
+  }
+
+  .group-form-document-list li {
+    margin-top: .5em;
```
robertknight: I'm generally preferring `px` units in new styling as this makes it easier to ensure fidelity with mockups.

h/api/search/test/core_test.py
seanh: Done

h/api/models.py
```diff
@@ -166,6 +175,113 @@ def title(self):
...
 
     @property
+    def filename(self):
+        if is_file(self.uri):
+            return self.uri.split("/")[-1] or ""
+        else:
+            return ""
+
+    @property
+    def document_link(self):
```
nickstenning: This is starting to look increasingly like presenter code, and it makes me a bit uneasy to have it in the model. I think for now we merge this as it is, but in future I think we should perhaps consider having an `AnnotationPresenter` class which takes an `Annotation` in its constructor and is responsible for all the mapping between the data model and the presentation model. This would presumably also make the logic in these things a bunch easier to test as it wouldn't require setting up the database connection, etc.

h/api/models.py
```diff
@@ -166,6 +175,113 @@ def title(self):
...
+            else:
+                return content[:length] + jinja2.Markup("&hellip;")
+
+        if link_text:
+            link_text = truncate(link_text)
+
+        if domain:
+            domain = truncate(domain)
+
+        assert title, "The title should never be empty"
```
nickstenning: Perhaps it's just a matter of preference, but I'd rather see the code structured in a way that makes it self-evident that the title is not empty, rather than using assertions like this.

That is, flip the logic on its head and take a "progressive enhancement" approach rather than the "graceful degradation" one you're employing here:

```
uri = jinja2.escape(self.uri) or ""
title = urllib2.unquote(uri)

if self.title:
    title = ...
elif is_file(uri):
    title = ...
```

h/api/models.py
```diff
@@ -166,6 +175,113 @@ def title(self):
...
+            link = ('<a title="{title}">{link_text}</a> ({domain})'.format(
+                title=title, link_text=link_text, domain=domain))
+        elif href and not domain:
+            link = '<a href="{href}" title="{title}">{link_text}</a>'.format(
+                href=href, title=title, link_text=link_text)
+        elif (not href) and (not domain):
+            link = '<a title="{title}">{link_text}</a>'.format(
+                title=title, link_text=link_text)
+        else:
+            assert False, "We should never get here"
```
nickstenning: This assertion is unnecessary if you just reduce the `if` clause to its minimal form.

```python
if href and domain:
    ...
elif domain: # domain and no href
    ...
elif href: # href and no domain
    ...
else: # no domain and no href
    ...
```

h/api/search/test/core_test.py
```diff
@@ -0,0 +1,30 @@
...
+
+
+@pytest.fixture
+def query(request):
+    patcher = mock.patch('h.api.search.core.query', autospec=True)
+    request.addfinalizer(patcher.stop)
```
nickstenning: Small request. Can you make sure not to add the finalizer until the patcher is actually started? Otherwise a failure to patch (perhaps because the module moved) will also spew an error about "called stop on unstarted patcher" and the real error gets hidden further up. So:

```python
patcher = mock.patch(...)
result = patcher.start()
return result
```

h/groups/templates/create.html.jinja2
```diff
@@ -18,8 +18,7 @@
     <div class="group-form-footer">
       <a href="" class="group-form-footer__explain-link js-group-info-link">Tell me more about groups</a>
       <div class="group-form-footer__explain-text js-group-info-text is-hidden">
-        <p class="group-form-footer__heading">About Groups</p>
-        <p>Groups let you annotate and discuss content on the web privately.</p>
+        {% include "about-groups.html" %}
```
nickstenning: This file should have a `.jinja2` extension.

h/groups/test/views_test.py
```diff
@@ -349,6 +349,125 @@ def test_read_if_already_a_member_returns_response(Group, renderers):
     assert views.read(request) == mock.sentinel.response
 
 
+@read_fixtures
+def test_read_calls_search(Group, search, renderers):
```
nickstenning: Small quibble on the name. We're not testing that it calls `search`. We're testing the collaboration between the view function and the `search` function -- that is, we're testing that it calls `search` with the correct arguments. Might be nice to make that clearer in the method name or the docstring, or both.

h/groups/test/views_test.py
```diff
@@ -349,6 +349,125 @@ def test_read_if_already_a_member_returns_response(Group, renderers):
...
+
+    views.read(request)
+
+    search.search.assert_called_once_with(
+            request, private=False, params={"group": g.pubid, "limit": 1000})
+
+
+@read_fixtures
+def test_read_calls_normalize(Group, search, renderers, uri):
+    """It shold call normalize() with each of the URIs from search()."""
```
nickstenning: I'm not really sure about this test. Again, what I think you actually want to test is that if `search` returns N annotations whose URIs differ only in normalisation, you only get one back, no? There are many ways of doing this, some of which don't involve calling `uri.normalize` (e.g. partition on `target.source` rather than `uri.normalize(annotation.uri)`).

h/groups/test/views_test.py
```diff
@@ -349,6 +349,125 @@ def test_read_if_already_a_member_returns_response(Group, renderers):
...
+
+    views.read(request)
+
+    assert (
+        renderers.render_to_response.call_args[1]['value']['document_links']
+        == ["document_link_1", "document_link_2", "document_link_3"])
+
+
+@read_fixtures
+def test_read_duplicate_documents_are_removed(Group, search, renderers, uri):
```
nickstenning: So yes, if you've got this test, then the test I mentioned in [this comment](https://github.com/hypothesis/h/pull/2667/files#r43128239) seems redundant and overly brittle.

h/groups/test/views_test.py
```diff
@@ -349,6 +349,125 @@ def test_read_if_already_a_member_returns_response(Group, renderers):
...
+    user.groups = [g]  # The user is a member of the group.
+    annotations = [
+        mock.Mock(uri="uri_{n}".format(n=n),
+                  document_link="document_link_{n}".format(n=n))
+        for n in range(0, 50)
+    ]
+    search.search.return_value = {"rows": annotations}
+
+    def normalize(uri):
+        # All three annotations' URIs normalize to the same URI.
```
nickstenning: `s/three//`

h/groups/views.py
```diff
@@ -101,7 +105,28 @@ def _read_group(request, group):
...
-    template_data = {'group': group, 'group_url': url}
+
+    result = search.search(request, private=False, params={
+        "group": group.pubid, "limit": 1000})
+    annotations = result['rows']
+
+    # Group the annotations by URI.
+    # Create a dict mapping the (normalized) URIs of the annotated documents
+    # to the most recent annotation of each document.
+    annotations_by_uri = collections.OrderedDict()
```
nickstenning: I'm not sure if you'd say this is easier or harder to read, but it makes a bit more sense to me:
```
buckets = itertools.groupby(annotations, lambda a: uri.normalize(a.uri))
document_links = itertools.islice(next(anns).document_link for _, anns in buckets), 25)
```

nickstenning:
I'll also note that we should probably remove the autofocus from the group share field as part of this work, as this causes the viewport to jump down the page on load when there are more than a few "recently annotated documents."

nickstenning:
I'm going to merge this as-is, @seanh. It works, and it can serve as the basis for some design iteration. I would like it if you could go through the remaining comments on this nonetheless. Cheers.

seanh:
Unfortunately testing this will be a little pointless as I'm about to push a complete refactor of the document link rendering :(

seanh:
Continuing here https://github.com/hypothesis/h/pull/2683

h/groups/views.py
seanh: Matter of opinion I guess, I find my way easier to read. Mine also breaks as soon as it has 25 unique normalized URIs.

h/api/models.py
seanh: Do I need to setup the database model to test these?

But yes we could move it out of the model. If we did so it might be convenient if `search.search()` returned `AnnotationPresenter` objects, or maybe it's better if user code (such as view code) wraps `Annotation`s in `AnnotationPresenter`s itself.

h/api/models.py
seanh: I think these are gone in [the refactor](https://github.com/hypothesis/h/pull/2683)

h/api/models.py
seanh: Also gone in [the refactor](https://github.com/hypothesis/h/pull/2683)

h/api/search/query.py
seanh: Renamed

h/api/test/models_test.py
seanh: This checking is done differently in [the refactor](https://github.com/hypothesis/h/pull/2683)

h/groups/templates/create.html.jinja2
seanh: Done on [the other pull request](https://github.com/hypothesis/h/pull/2683)

h/groups/test/views_test.py
seanh: Done

seanh:
Ok, I've gone through the comments, I think they've all been replied to here or else done on <https://github.com/hypothesis/h/pull/2683>, a few are marked as TODOs on <https://github.com/hypothesis/h/pull/2683>, will continue this work over on <https://github.com/hypothesis/h/pull/2683>.

h/api/models.py
seanh: This has been broken up here https://github.com/hypothesis/h/pull/2683

h/api/search/test/core_test.py
seanh: To be done here https://github.com/hypothesis/h/pull/2683

h/groups/test/views_test.py
seanh: This is literally just supposed to be testing that it calls `normalize()` once for each of the URIs that `search()` returns. But I think you're right that the test below (the one that tests that duplicate normalized URIs are removed) covers this, so this one can just be deleted

PR #2669, opened 2015-10-26
Do not change focused group when clicking 'leave' icon

robertknight:
This PR adds a missing test that clicking 'Leave' and rejecting the dialog does not result in leaving the group and fixes an issue where leaving a group other than the one you had focused resulted in the focused group changing.

There is an issue I'm seeing locally where join/leave actions initially update the H groups list but at some point the push notifications appear to stop working.


seanh:
I've reproduced the issue on master and confirmed that this branch fixes it

PR #2672, opened 2015-10-26
Move signin control into its own directive

nickstenning:
This commit moves the signin and account control (and dropdown) into its own directive.

To make it easier to understand what's going on at load, I've also changed the signalling of user signed-in/signed-out state to an explicit `status` field on the `auth` object, rather than relying on a tristate of {undefined, null, user object}. This results in templates that are marginally more verbose, but substantially clearer in intent.

Lastly, I've moved the computation of the `username` and `provider` properties into AppController so that it doesn't need a `$scope.$watch`.

This is work in preparation for [sorting out the top bar for signed-out users](https://trello.com/c/T1D2ijhF/138-top-bar-design-changes-for-signed-out-users).

### DIFF
--- a/h/static/scripts/app-controller.coffee
+++ b/h/static/scripts/app-controller.coffee
@@ -1,6 +1,7 @@
 angular = require('angular')
 
-events = require('./events');
+events = require('./events')
+parseAccountID = require('./filter/persona').parseAccountID
 
 module.exports = class AppController
   this.$inject = [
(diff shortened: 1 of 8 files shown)


h/static/scripts/app-controller.coffee
```diff
@@ -53,9 +56,27 @@ module.exports = class AppController
...
-      onlogin: (identity) -> $scope.auth.user = auth.userid(identity)
-      onlogout: -> $scope.auth.user = null
-      onready: -> $scope.auth.user ?= null
+      onlogin: (identity) ->
+        # Hide the account dialog
+        $scope.accountDialog.visible = false
+        # Update the current logged-in user information
+        userid = auth.userid(identity)
+        parsed = parseAccountID(userid)
+        angular.copy({
```
robertknight: In the Chrome extension where we're not using Angular I've started to use [Object.assign](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Object/assign) or a polyfill for it for this kind of thing. We might want to consider doing the same in the main app in future.

robertknight:
> To make it easier to understand what's going on at load, I've also changed the signalling
> of user signed-in/signed-out state to an explicit status field on the auth object, rather
> than relying on a tristate of {undefined, null, user object}. 

:thumbsup: - Relying on the difference between different falsey values in JS is bound to lead to errors.

PR #2673, opened 2015-10-26
Deliver up-to-date groups list in groups leave/join notifications

### LINKED ISSUE #2671: Groups list does not always update correctly when joining and leaving groups
**Steps to reproduce**

1. With a groups-enabled account, open a tab with H
2. From one of the tabs, use the groups dropdown to create or leave a group
3. Repeat step (2) a few times

Extended version: Use several tabs with H instead of just one

**Expected result**

All tabs show a consistent and correct groups list

**Actual result**

The groups list does not always update properly when joining and leaving groups.

robertknight:
When pushing a session state change notification to clients
after joining or leaving a group, the session model could
be out of date because the groups view published the notification
prior to committing the database transaction.

If the websocket server picked up the message from NSQ and queried
the DB for the session state prior to the commit happening,
the client would receive a notification but the payload
would be out of date.

We considered two possible approaches to fix this - one
was to commit the transaction in the groups views before
publishing the notification, the other was to serialize the
session model as part of the notification.

This commit takes the latter approach for consistency with how
annotation notifications work and also as slightly more
explicit to read. We may want to reconsider in future.

Fixes #2671

### DIFF
--- a/h/groups/test/views_test.py
+++ b/h/groups/test/views_test.py
@@ -62,7 +62,8 @@ def test_create_form_returns_form(Form):
 
 
 # The fixtures required to mock all of create()'s dependencies.
-create_fixtures = pytest.mark.usefixtures('GroupSchema', 'Form', 'Group')
+create_fixtures = pytest.mark.usefixtures('GroupSchema', 'Form', 'Group',
+                                          'session_model')
 
 
 @create_fixtures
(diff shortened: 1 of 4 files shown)


seanh:
This also appears to fix an issue that if you leave the currently focused group it remains in the groups dropdown until you reload.

seanh:
This looks good to me, I did a lot of leaving and joining groups with the sidebar open in multiple tabs, I found one issue: <https://github.com/hypothesis/h/issues/2677> which I don't think should block launching groups

Issue #2641, opened 2015-10-20
Broken UI, lost work, when creating a new group while creating or editing an annotation


seanh:
1. Begin creating a new annotation, don't save it
2. Create a new group (opens in new tab)
3. Return to first tab, which refreshes with the new group

Result: UI goes a bit broken, typed annotation text is lost

![untitled screencast 5](https://cloud.githubusercontent.com/assets/22498/10609014/8869fa9e-7739-11e5-90fb-d94d527dffa4.gif)

robertknight:
Current status: The UI broken-ness appears to have been fixed by the Angular 1.2.x -> Angular 1.4.x upgrade.

The issue with the annotation text being lost persists. Switching groups currently results in a route-reload which loses any unsaved changes to annotations.

PR #2674, opened 2015-10-26
Avoid route reload on groups list change

robertknight:
This builds on https://github.com/hypothesis/h/pull/2669 and fixes the issue described in https://github.com/hypothesis/h/issues/2641 where the user would lose any unsaved annotation text if they joined or left a group whilst an annotation was being created in another tab.

If the user leaves the group which is currently focused in a tab, that will still result in unsaved annotations in that tab being lost.

### DIFF
--- a/h/static/scripts/directive/group-list.js
+++ b/h/static/scripts/directive/group-list.js
@@ -35,6 +35,7 @@ function GroupListController($scope, $window, groups) {
 
   $scope.allGroups = groups.all();
   $scope.$on(events.GROUPS_CHANGED, function () {
+    $scope.focusedGroup = groups.focused();
     $scope.allGroups = groups.all();
   });
 
--- a/h/static/scripts/directive/group-list.js
+++ b/h/static/scripts/directive/group-list.js
@@ -1,7 +1,9 @@
 'use strict';
 
+var events = require('../events');
+
 // @ngInject
-function GroupListController($scope, $window) {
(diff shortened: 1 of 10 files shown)


h/static/scripts/directive/group-list.js
```diff
@@ -19,13 +21,27 @@ function GroupListController($scope, $window) {
...
     }
   }
+
+  $scope.focusGroup = function (groupId) {
+    groups.focus(groupId);
+  }
+
+  $scope.allGroups = groups.all();
+  $scope.$on(events.GROUPS_CHANGED, function () {
+    $scope.allGroups = groups.all();
```
nickstenning: Is this really necessary? Seems like a bit of a performance microoptimisation over just passing the `groups` object into the template.

nickstenning: Ah, wait, I see. It is necessary because a digest cycle won't automatically be triggered when the session data updates.

h/static/scripts/test/session-test.js
```diff
@@ -158,10 +158,11 @@ describe('h:session', function () {
   });
 
   describe('#update()', function () {
-    it('broadcasts an event when the session is updated', function () {
+    it('broadcasts SESION_CHANGED when the session changes', function () {
```
nickstenning: s/SESION/SESSION/

nickstenning:
LGTM.

PR #2676, opened 2015-10-26
Fix display of the (new) top bar for signed-out users

nickstenning:
- Show the focused group ("Public") when signed-out, without allowing the user to change it.
- Show a "Sign in" link in the right place, with the correct padding.
- Minor tweak to dropdown menu so dividers don't show in the wrong places.

### DIFF
--- a/h/static/scripts/directive/group-list.js
+++ b/h/static/scripts/directive/group-list.js
@@ -60,7 +60,9 @@ function groupList(groups, $window) {
       });
     },
     restrict: 'E',
-    scope: {},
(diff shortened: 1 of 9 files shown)


nickstenning:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;T1D2ijhF&#x2F;138-top-bar-design-changes-for-signed-out-users"> Top bar design changes for signed-out users</a>

robertknight:
LGTM - Just needs a rebase.

PR #2678, opened 2015-10-27
Show "Only Me" when editing a private annotation

seanh:
Currently on master when you create a private annotation in a group, after saving it, the top of the annotation card looks like this:

screenshot from 2015-10-27 12 01 25

But when creating or editing an annotation, the _Only Me_ part is not shown even if _Only Me_ is selected in the _Post_ dropdown, I consider this a bug:

screenshot from 2015-10-27 12 03 19

_Only Me_ will appear after saving it.

This pull request fixes this to dynamically show and hide the _Only Me_ part as the _Post_ dropdown changes, when creating/editing:

untitled screencast - cropped 5

I think this accurately reflects what is happening - an annotation is being created _in the group_ but it is _private_. For better or for worse, you can create private annotations in groups currently, so not showing the group name would be wrong.

### DIFF
--- a/h/templates/client/annotation.html
+++ b/h/templates/client/annotation.html
@@ -21,7 +21,7 @@
         </a>
       </span>
       <i class="h-icon-border-color" ng-show="vm.isHighlight() && !vm.editing" title="This is a highlight. Click 'edit' to add a note or tag."></i>
-      <span ng-show="vm.isPrivate() && !vm.editing"
+      <span ng-show="vm.isPrivate()"
             title="This annotation is visible only to you.">
         <i class="h-icon-lock"></i> Only Me
       </span>


seanh:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;q9bfOR5J&#x2F;135-while-drafting-an-annotation-the-group-name-or-lack-thereof-should-reflect-its-current-visibility"> While drafting an annotation, the group name (or lack thereof) should reflect its current visibility</a>

robertknight:
Merged. We can always change this later if we decide we want to take a different approach.

judell:
This looks beautiful! And thanks for making the screencast to illustrate, that's super-helpful.

PR #2679, opened 2015-10-27
Adjust annotation card style to fit groups design

robertknight:
This updates the presentation of annotation cards as per the designs at https://trello.com/c/RGZfxwTa/148-update-the-styling-of-the-groups-name-on-annotations-cards

**Public**

public-group

**Private groups**

private-group


h/static/styles/annotations.scss
```diff
@@ -31,13 +38,38 @@
...
+
 .annotation-footer { margin-bottom: 0 }
 
+.u-flex-spacer {
+  flex-grow: 1;
+}
+
 .annotation-section {
+  margin-top: 14px;
+  margin-bottom: 14px;
```
nickstenning: This is problematic for replies, which have targets but no selectors (q.v. `annotation.html`):

screen shot 2015-10-27 at 16 04 37

robertknight:
Related annotated screenshot from @dwhly 
This screenshot is not from this branch but it might be worth making these changes at the same time:

screen_shot_2015-10-21_at_10_31_18_am

robertknight:
This is how it currently looks with the tweaks @dwhly asked for and a fix for the spacing at the top of replies mentioned by @nickstenning above:

group-name-styling

I think the annotation actions add too much clutter myself. I have a small variation which shows actions only on hover that I could do as a separate PR.

fade actions on hover

@conordelahunty - Thoughts on the above?

dwhly:
> I think the annotation actions add too much clutter myself. I have a small variation which shows actions only on hover that I could do as a separate PR.

We had it this way before for exactly the same reason, loved it.  Strong support here.  We'd probably need to suppress this on mobile though, because no hover.

dwhly:
screen shot 2015-10-28 at 7 25 19 am

I've just realized that trimming the end of the thread line on any but the last reply in a sequence means that the ones above no longer join.  Cancel my stupid suggestion!  (Or just trim the last.)

robertknight:
>  Cancel my stupid suggestion! (Or just trim the last?)

I've changed the styling so that the dashed line is only trimmed for the last item in the list.

nickstenning:
Sorry @robertknight but this still has issues with replies. This time it's that the dotted line now breaks next to the "n replies" link for nested replies:

screen shot 2015-10-29 at 14 24 39

nickstenning:
The behaviour of the margin line in this context also looks wrong to me:

screen shot 2015-10-29 at 14 26 25

robertknight:
@nickstenning - I've reverted the changes to the dashed line for the moment. I'll get conor's feedback about how it should look in different states another time.

dwhly:
:+1:

nickstenning:
LGTM.

PR #2681, opened 2015-10-27
Remove the remnants of a once-functional visual diff feature

nickstenning:
Once upon a time, we could show a diff between what the annotation has anchored to in the page and the original selected text. That feature has been non-functional since at least 4f60030.

No rush on this one. Just noticed some dead code while reviewing other things.


robertknight:
There's no code like deleted code :+1:

PR #2682, opened 2015-10-27
Remove the remnants of a defunct annotation pulse feature

nickstenning:
Another feature that used to work but now doesn't. Remove the code for now, as it's just bloat that is, at best, doing nothing.

As with #2681, no rush here. Just noticed some dead code while reviewing other things.


tilgovi:
\o/ this was one of the few things that made "thread" annotation-specific.

judell:
So that explains it! @dwhly spotted it in the wild recently. I had never encountered the mythical creature.

PR #2683, opened 2015-10-27
Document link refactor

seanh:


seanh:
TODO:

- [ ] Fix patchers, see [Nick's comment](https://github.com/hypothesis/h/pull/2667/files#r43126490)

seanh:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;Nj6M1pMF&#x2F;144-add-a-list-of-up-to-most-recently-annotated-by-the-group-urls-to-the-group-page"> Add a list of up to most recently annotated (by the group) URLs to the group page</a>

seanh:
One thing I've noticed is that since the text in the `<input>` on the group page is automatically selected to make copying easier, and the text is at the bottom of the page, the browser window will scroll to the bottom of the page on page load. This only really matters if the list of documents is long enough to make a scrollbar appear of course, but with our current length (max 25 documents) it can be.

seanh:
Ok, I've refactored this, and done some more work to make it more bullet proof against unexpected data.

@nickstenning I'll do the patcher thing in a separate pr, since we have that patcher problem all over our code

h/api/models.py
```diff
@@ -11,9 +11,45 @@
...
+            return content[:length] + jinja2.Markup("&hellip;")
+
+    hostname = truncate(hostname)
+    link_text = truncate(link_text)
+
+    if href and hostname:
+        link = ('<a href="{href}" title="{title}">{link_text}</a> '
+                '({hostname})'.format(href=href, title=title,
+                                      link_text=link_text, hostname=hostname))
+    elif hostname and not href:
```
nickstenning: The second half of these conditionals isn't needed... it's implied by the fact we're in an `elif` clause and we'd never of got here if it weren't true.

h/api/models.py
```diff
@@ -11,9 +11,45 @@
...
+        link = ('<a title="{title}">{link_text}</a> ({hostname})'.format(
+            title=title, link_text=link_text, hostname=hostname))
+    elif href and not hostname:
+        link = '<a href="{href}" title="{title}">{link_text}</a>'.format(
+            href=href, title=title, link_text=link_text)
+    else:
+        link = '<a title="{title}">{link_text}</a>'.format(
+            title=title, link_text=link_text)
+
```
nickstenning: This whole function would be a fair bit clearer if you just assigned the format strings and did the `.format` once at the end:

```
if href and hostname:
    tpl = '<a href="{href}" title="{title}">{link_text}</a> ({hostname})'
elif hostname:
    tpl = '...'
elif href:
   ...

link = tpl.format(href=href, title=title, link_text=link_text, hostname=hostname)
return jinja2.Markup(link)
```

h/api/models.py
```diff
@@ -166,28 +202,173 @@ def get_analysis(cls):
...
+        of it, otherwise return "".
+
+        The filename is escaped and safe to be rendered.
+
+        If it contains escaped characters then the filename will be a
+        Markup object so it won't be double-escaped.
+
+        """
+        if self.uri.lower().startswith("file:///"):
+            # self.uri is already escaped so we don't need to escape it again
```
nickstenning: There's a bunch of places you've got comments like this. I don't think this is the right approach. The entire point of a taintedness approach to string escaping is that you don't need to examine the whole call stack to understand if what you're doing is safe. If this function returns something that needs to be escaped, it should escape it. Using `jinja2.escape(self.uri.split('/')[-1])` will not double escape anything.

h/api/search/query.py
```diff
@@ -118,14 +118,14 @@ def __call__(self, params):
...
-        if 'group:__world__' not in groups:
-            groups.insert(0, 'group:__world__')
+        if 'group:__world__' not in principals:
+            principals.insert(0, 'group:__world__')
 
         if not self.private:
-            groups = [g for g in groups
-                      if not g == self.request.authenticated_userid]
+            principals = [p for p in principals
+                          if not p == self.request.authenticated_userid]
```
nickstenning: s/`not p ==`/`p !=`/

h/api/test/models_test.py
```diff
@@ -65,120 +66,486 @@ def groups(pattern, text):
...
+    assert isinstance(
+        models.Annotation(uri="http://foo.com").uri, jinja2.Markup)
+
+
+def test_uri_when_uri_is_not_a_string():
+    for uri in (True, None, 23, 23.7, {"foo": False}, [1, 2, 3]):
+        assert isinstance(models.Annotation(uri=uri).uri, unicode)
+
+
+@patch("h.api.models.Annotation.uri", new_callable=PropertyMock)
```
nickstenning: As I mentioned the other day, there's really no need to use `PropertyMock` here. `@patch.object(models.Annotation, 'uri', 'http://example.com/example.html')` would work just fine and would be less to understand.

nickstenning:
I have a bunch of comments, but none of them are blockers to merging. Take a chance to read them over and feel free to either submit PRs addressing the comments or roll that work into whatever else we do with this later.

Issue #2530, opened 2015-09-23
Flush feature flags on login/logout


nickstenning:
Currently feature flags are cached for about 5 minutes on the client, and persist across logins/logouts.

Seeing as logging in/out can chance your permissions level, and thus your feature flag values, we should probably flush the feature flags when you do so that the next time something asks after a feature flag it gets the new value.

nickstenning:
Closed by #2684.

PR #2684, opened 2015-10-27
Wait for feature flag data before loading sidebar view

robertknight:
This fixes an issue where the app could show a private group as focused in the top bar but annotations from the public group.

This happened when:
 1. The feature flag data was loaded _after_ the initial search for annotations had been performed.
 2. The user has groups enabled
 3. In the user's previous H session, they had a private group selected.

In this scenario, when the initial search queries the focused group, it will get Public because until the feature flag data is returned, private groups is disabled.

Once the feature flag data loads, the groups feature is enabled, the focused group implicitly changes to the user's last selected private group but the annotation search is not updated.

The approach taken here is to request features at the same time as session data on sidebar load and wait for both before loading the view and requesting annotations.

`features.fetch()` now returns a promise which resolves when the features are fetched or the retrieval fails. I left out the exponential retry logic for the moment. If fetching feature data fails, the client will simply behave as if all features are turned off until the cache expires. I'm happy to reconsider this if we think it is important enough to keep.

----

* [X] Update the reference to `SESSION_CHANGED` once https://github.com/hypothesis/h/pull/2674 is merged

### DIFF
--- a/h/static/scripts/app.coffee
+++ b/h/static/scripts/app.coffee
@@ -7,6 +7,9 @@ require('angular-jwt')
 streamer = require('./streamer')
 
 resolve =
+  # Ensure that we have feature flags available before we load the main
+  # view as features such as groups affect which annotations are loaded
+  featuresLoaded: ['features', (features) -> features.fetch()]
   # Ensure that we have available a) the current authenticated userid, and b)
   # the list of user groups.
   sessionState: ['session', (session) -> session.load().$promise]
--- a/h/static/scripts/features.js
+++ b/h/static/scripts/features.js
@@ -33,10 +33,7 @@ function features ($document, $http, $log, $rootScope) {
   var featuresUrl = new URL('/app/features', $document.prop('baseURI')).href;
   var fetchOperation;
 
-  // FIXME - This should be changed to clear the feature flag cache
-  // only in response to the USER_CHANGED event once
-  // https://github.com/hypothesis/h/pull/2674 lands
-  $rootScope.$on(events.SESSION_CHANGED, function () {
+  $rootScope.$on(events.USER_CHANGED, function () {
     cache = null;
   });
 
--- a/h/static/scripts/features.js
+++ b/h/static/scripts/features.js
@@ -21,58 +21,69 @@
  */
 'use strict';
 
-var retry = require('retry');
+var assign = require('core-js/modules/$.assign');
+
+var events = require('./events');
 
 var CACHE_TTL = 5 * 60 * 1000; // 5 minutes
 
-function features ($document, $http, $log) {
+// @ngInject
+function features ($document, $http, $log, $rootScope) {
   var cache = null;
-  var operation = null;
   var featuresUrl = new URL('/app/features', $document.prop('baseURI')).href;
+  var fetchOperation;
+
+  // FIXME - This should be changed to clear the feature flag cache
+  // only in response to the USER_CHANGED event once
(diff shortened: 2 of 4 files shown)


h/static/scripts/features.js
```diff
@@ -21,58 +21,66 @@
...
+      };
+    }).catch(function (err) {
+      // if for any reason fetching features fails, we behave as
+      // if all flags are turned off
+      $log.warn('failed to fetch feature data', err);
+      cache = assign({}, cache, {
+        updated: Date.now(),
+      });
+    }).then(function () {
+      fetchOperation = null;
```
nickstenning: I'm not sure you can do this any more. Using the presence or absence of the operation to determine if an operation was already in-flight is fine if you're not returning the promise, but there's a race condition here now -- you'll be returning null if this completes before the function returns.

nickstenning: This might be a valid use for `$resolved`.

h/static/scripts/app.coffee
```diff
@@ -7,6 +7,9 @@ require('angular-jwt')
 streamer = require('./streamer')
 
 resolve =
+  # Ensure that we have feature flags available before we load the main
+  # view as features such as groups affect which annotations are loaded
+  features: ['features', (features) -> features.fetch()]
```
nickstenning: This is a funny one. The object you pass to `resolve` is [supposed to provide additional injectable objects](https://code.angularjs.org/1.4.7/docs/api/ngRoute/provider/$routeProvider)... so I would expect what you return from this function to overwrite the `features` service.

Clearly this doesn't happen, otherwise you'd have noticed (and indeed I've just verified that `'features'` still injects the features service) but perhaps to be on the safe side we could call this `featuresLoaded` or something.

h/static/scripts/features.js
```diff
@@ -21,58 +21,66 @@
  */
 'use strict';
 
-var retry = require('retry');
```
nickstenning: I'm pretty sure you can now remove this from `package.json`.

h/static/scripts/features.js
```diff
@@ -21,58 +21,66 @@
...
+        flags: response.data,
+      };
+    }).catch(function (err) {
+      // if for any reason fetching features fails, we behave as
+      // if all flags are turned off
+      $log.warn('failed to fetch feature data', err);
+      cache = assign({}, cache, {
+        updated: Date.now(),
+      });
+    }).then(function () {
```
nickstenning: Probably clearer and more idiomatic to use `finally` here.

h/static/scripts/features.js
robertknight: To recap a discussion on Slack, there isn't a race condition here because promise rejection/fulfillment handlers are always invoked on subsequent turns of the event loop.

I'm not too keen on promise.$resolved because it isn't part of the [$q API](https://code.angularjs.org/1.4.7/docs/api/ng/service/$q) or A+/ES6 promises.

h/static/scripts/app.coffee
robertknight: Done

h/static/scripts/features.js
robertknight: Done. Again, I'm not super keen on the use of methods that are not part of the promises spec but `finally()` is part of the spec of `$q` and the `q` API it is based on and I think that does make things clearer.

nickstenning:
LGTM.

nickstenning:
This also fixes #2530.

PR #2687, opened 2015-10-28
Set permissions of new annotations on login

### LINKED ISSUE #2686: Creating annotation when signed out, then signing in and saving results in annotations with invalid permissions
**Steps to reproduce**

1. Create an annotation (_not highlight_) when signed out
2. Sign-in
3. Add text and save. 
4. Click on the 'Link' action for the annotation and open the annotation in a new tab

**Expected**
 * Annotation card for selected annotation shown
 * Annotation has valid permissions in DB

**Actual**
Error:

```
  File "/home/robert/.virtualenvs/h/local/lib/python2.7/site-packages/pyramid/tweens.py", line 21, in excview_tween
    response = handler(request)
  File "/home/robert/hypothesis/repos/h/h/tweens.py", line 24, in conditional_http_tween
    response = handler(request)
  File "/home/robert/.virtualenvs/h/local/lib/python2.7/site-packages/pyramid_tm/__init__.py", line 99, in tm_tween
    reraise(*exc_info)
  File "/home/robert/.virtualenvs/h/local/lib/python2.7/site-packages/pyramid_tm/__init__.py", line 80, in tm_tween
    response = handler(request)
  File "/home/robert/.virtualenvs/h/local/lib/python2.7/site-packages/pyramid/router.py", line 163, in handle_request
  File "/home/robert/.virtualenvs/h/local/lib/python2.7/site-packages/pyramid/config/views.py", line 276, in _authdebug_view
    permission))
  File "/home/robert/.virtualenvs/h/local/lib/python2.7/site-packages/pyramid/authorization.py", line 84, in permits
    acl = acl()
  File "/home/robert/hypothesis/repos/h/h/api/resources.py", line 54, in __acl__
    principals = auth.translate_annotation_principals(roles)
  File "/home/robert/hypothesis/repos/h/h/api/auth.py", line 13, in translate_annotation_principals
    if principal.startswith('system.'):
AttributeError: 'NoneType' object has no attribute 'startswith'
```

In the DB, the annotation's permissions are saved as "[null]" for all permissions except read.

seanh:
While signed out create a new annotation. You'll see a "You must sign in
to create an annotation" card. This card is in fact an annotation
in the client-side code, but one that has no user, no permissions, etc
etc. If you sign in, you'll then see a full annotation card with editor
open in place of the previous card, you can type an annotation and save
it.

Except that this will send invalid permissions to the server, which will
happy save them (no validation) and then crash.

The fix:

On sign in, set the model.permissions and @annotation.permissions of
such annotations. This means that the Post dropdown will set itself
correctly to shared or private and on posting the annotation the correct
permissions will be sent to the server.

Also fixed permissions.coffee to return null instead of invalid
permissions when not signed in.

Fixes #2686.

### DIFF
--- a/h/static/scripts/directive/annotation.coffee
+++ b/h/static/scripts/directive/annotation.coffee
@@ -61,19 +61,24 @@ AnnotationController = [
 
     model = $scope.annotationGet()
 
+    # Set model.permissions to the default permissions for new annotations,
+    # _if_ model.permissions isn't already set.
+    setDefaultPermissions = ->
+      if not model.permissions?.read
+        defaultLevel = localStorage.getItem(STORAGE_KEY);
+        if (defaultLevel != 'private') and (defaultLevel != 'shared')
+          defaultLevel = 'shared';
+        if defaultLevel == 'private'
+          model.permissions = permissions.private()
+        else
+          model.permissions = permissions.shared(model.group)
+
     # Set the group of new annotations.
     if not model.group
       model.group = groups.focused().id
 
     # Set the permissions of new annotations.
-    if not model.permissions
-      defaultLevel = localStorage.getItem(STORAGE_KEY);
-      if (defaultLevel != 'private') and (defaultLevel != 'shared')
-        defaultLevel = 'shared';
-      if defaultLevel == 'private'
-        model.permissions = permissions.private()
-      else
-        model.permissions = permissions.shared(model.group)
+    setDefaultPermissions()
 
     highlight = model.$highlight
     original = null
@@ -378,6 +383,8 @@ AnnotationController = [
           highlight = false # Prevents double highlight creation.
         else
           drafts.add model, => this.revert()
+
+      setDefaultPermissions()
 
       updateTimestamp(model is old)  # repeat on first run
       this.render()
--- a/h/static/scripts/directive/annotation.coffee
+++ b/h/static/scripts/directive/annotation.coffee
@@ -1,5 +1,7 @@
 ### global -validate ###
 
+events = require('../events')
+
 # Validate an annotation.
 # Annotations must be attributed to a user or marked as deleted.
 # A public annotation is valid only if they have a body.
@@ -58,6 +60,8 @@ AnnotationController = [
     @timestamp = null
 
     model = $scope.annotationGet()
+
+    model.user ?= session.state.userid
 
     # Set the group of new annotations.
     if not model.group
@@ -374,17 +378,15 @@ AnnotationController = [
       this.render()
     , true
 
-    # Watch the current user
-    # TODO: fire events instead since watchers are not free and auth is rare
-    $scope.$watch (-> session.state.userid), (userid) ->
-      model.user ?= userid
+    $scope.$on(events.USER_CHANGED, ->
+      model.user ?= session.state.userid
 
       # Set model.permissions on sign in, if it isn't already set.
       # This is because you can create annotations when signed out and they
       # will have model.permissions = null, then when you sign in we set the
       # permissions correctly here.
       model.permissions = model.permissions or permissions.default(model.group)
-
+    )
 
     # Start editing brand new annotations immediately
     unless model.id? or (this.isHighlight() and highlight) then this.edit()
--- a/h/static/scripts/directive/annotation.coffee
+++ b/h/static/scripts/directive/annotation.coffee
@@ -1,7 +1,5 @@
 ### global -validate ###
 
-STORAGE_KEY = 'hypothesis.privacy'
-
 # Validate an annotation.
 # Annotations must be attributed to a user or marked as deleted.
 # A public annotation is valid only if they have a body.
@@ -46,10 +44,10 @@ errorMessage = (reason) ->
 AnnotationController = [
   '$scope', '$timeout', '$q', '$rootScope', '$document',
   'drafts', 'flash', 'permissions', 'tags', 'time',
-  'annotationUI', 'annotationMapper', 'session', 'groups', 'localStorage'
+  'annotationUI', 'annotationMapper', 'session', 'groups',
   ($scope,   $timeout,   $q,   $rootScope,   $document,
    drafts,   flash,   permissions,   tags,   time,
-   annotationUI,   annotationMapper,   session,   groups,   localStorage) ->
+   annotationUI,   annotationMapper,   session,   groups) ->
 
     @annotation = {}
     @action = 'view'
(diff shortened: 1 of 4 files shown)


seanh:
This fixes the bug but I haven't tested it extensively, and it could do with adding some tests and some comments / docstrings.

seanh:
What we should really do here (when we have more time) is have valid permissions (and probably many other fields besides) be generated server-side instead of client-side, and also strictly validate any fields that do come from the client

h/static/scripts/directive/annotation.coffee
```diff
@@ -371,6 +379,8 @@ AnnotationController = [
         else
           drafts.add model, => this.revert()
 
+      setPermissions()
```
robertknight: setPermissions() is called just prior to `render()` which recreates the view model (`@annotation`) from the domain model - so I don't think it is necessary to modify `@annotation` in `setPermissions()`

robertknight:
Needs tests as you mentioned, otherwise LGTM. As you also mentioned, server-side validation is needed - that can come in a separate PR. I tentatively created a Trello card: https://trello.com/c/UKq1Zi0v/156-validate-annotation-permissions

One thought which occurred to me is whether we need to expose the full richness of the permissions model to clients or whether we could change the permissions field to just an enum `{shared, private}`. The server would then fill out the rest of the details. I think the current approach is probably better to avoid API changes in future and for consistency between the model that we work with on the server and the model that clients see, but just a thought.

h/static/scripts/directive/annotation.coffee
```diff
@@ -61,19 +61,27 @@ AnnotationController = [
 
     model = $scope.annotationGet()
 
+    # Set the permissions of a new anonotation.
+    setPermissions = ->
```
robertknight: I think the naming could be a little clearer - `setDefaultPermissions`. In functions in this class you also need to be explicit about whether the changes apply to the _domain_ model (`model`) or the _view_ model (`@annotation`). In the context I think what we want to do is apply the changes to `model` and then let `render()` update the view.

h/static/scripts/directive/annotation.coffee
seanh: You're right. In an earlier version of this fix it was necessary but now not

seanh:
> change the permissions field to just an enum {shared, private}

Or just a `private: true|false`. +1

h/static/scripts/directive/annotation.coffee
```diff
@@ -61,19 +61,24 @@ AnnotationController = [
 
     model = $scope.annotationGet()
 
+    # Set model.permissions to the default permissions for new annotations,
+    # _if_ model.permissions isn't already set.
+    setDefaultPermissions = ->
+      if not model.permissions?.read
```
nickstenning: Is there a good reason to check the `read` key rather than just `if not model.permissions`? The comment says `if model.permissions isn't already set`.

h/static/scripts/permissions.coffee
```diff
@@ -37,10 +37,15 @@ module.exports = ['session', (session) ->
   # Typical use: annotation.permissions = permissions.private()
   ###
   private: ->
-    read: [session.state.userid]
-    update: [session.state.userid]
-    delete: [session.state.userid]
-    admin: [session.state.userid]
+    if not session?.state?.userid
```
nickstenning: This is overly defensive. If `session` doesn't exist something is badly wrong. And the `session` service sets `state` to an empty object by default. `if not session.state.userid` should be fine here.

nickstenning:
Looks good. Just a couple of minor comments.

h/static/scripts/permissions.coffee
seanh: It does reduce the amount of mocking needed everywhere though...

h/static/scripts/directive/annotation.coffee
seanh: This is because `model.permissions` sometimes gets set to `{}`. But I'm refactoring this, so I'll check whether this is still necessary

seanh:
I'm still working on refactoring the code to move more responsibility into permissions.coffee and then refactoring the tests to follow...

h/static/scripts/permissions.coffee
nickstenning: If the service depends on the session service, then we need to supply a session service mock when running the tests. I don't think adding defensive conditionals to the code just to support testing is a good idea.

nickstenning: Also, looking at the tests, that's simply not true. A fake session object is already injected for all the `permissions` tests.

h/static/scripts/directive/annotation.coffee
nickstenning: > This is because model.permissions sometimes gets set to {}

It does, where? I ask because the previous condition was `if not model.permissions`.

seanh: Previously it was broken, and this was the reason why.

I'll dig into it further

seanh:
Ok, I've added tests, and also refactored the code to move permissions-generating responsibility into permissions.coffee, and rebased onto master.

@nickstenning Sure, I can do that TODO now

h/static/scripts/directive/annotation.coffee
seanh: Ok I managed to get rid of this check

h/static/scripts/permissions.coffee
seanh: Yeah, I know the existing tests will pass without this.

One thing that I'm not happy with about our client-side tests is the amount of detail faking needed, though. The AnnotationController tests need a fake session service with a state, and a fake permissions serve with several methods, etc etc. If we make a change to the interface of any of these services then all our test fakes become out of date.

nickstenning:
There were two instances of `session?.state?.userid`. 0a0e990 only addresses one, fwiw.

seanh:
Ok, I recorded [this](https://github.com/hypothesis/h/issues/2692) elsewhere. I think this pull request is ready now

seanh:
I've tested this but I recommend the reviewer test it carefully as well:

- Create private annotations while signed in
- Create public annotations while signed in
- Check the public/private stickiness
- With privacy stuck on public, sign out, create an annotation, sign in, and complete it
- With privacy stuck on private, sign out, create an annotation, sign in, and complete it
- Create highlights, both when signed in and when signed out (then sign in to have them saved), and with the privacy stuck to public and to private
- Create group annotations
- Edit annotations

dwhly:
@judell: can we capture the above as a part of a draft testing narrative?

h/static/scripts/permissions.coffee
```diff
@@ -63,6 +72,29 @@ module.exports = ['session', (session) ->
       admin: [session.state.userid]
     }
 
+  private: private_
```
nickstenning: This is mainly CoffeeScript weirdness, but I find this a bit harder to read now that we switch apparently without warning from defining functions to an (implicit) returned object.

I think it would be less confusing if you stuck with the old way of just defining the object inline, and using `this.private` and `this.shared` in `default`.

h/static/scripts/permissions.coffee
```diff
@@ -63,6 +72,29 @@ module.exports = ['session', (session) ->
...
+
+  shared:shared
```
nickstenning: Missing a space.

nickstenning:
Every once in a while with this when creating a highlight when signed out and subsequently signing in, this branch sends *two* POSTs and thus creates two highlights.

nickstenning:
I don't seem to be able to reproduce this reliably, so I guess it's a race condition of some sort or other.

seanh:
Done

seanh:
Hmm. race condition, that's annoying. But I bet it's because the code for saving highlights on sign in is in a watch() that's actually called way more often than just on sign in, moving that into the USER_CHANGED event might fix it, but I have to be able to reproduce it first...

nickstenning:
I can reproduce the bug I describe on master, so it's not this PR's problem. Even if it were, frankly -- it's a bug that occurs some of the time and is easily resolved by the user, while the PR fixes a bug that occurs all of the time and is impossible for the user to resolve. Merging.

Thanks Sean!

PR #2688, opened 2015-10-29
Fix exception when GROUPS_CHANGED event is broadcast during route load

robertknight:
The <group-list> directive attempted to update itself in
response to a group change notification without triggering
a full digest cycle by using `$scope.$apply`.

This was based on the incorrect understanding that $apply only
dirty-checks the current scope downwards. In fact, in dirty-checks
the root scope. Additionally, the logic was pointless since
group list/focus changes happen in response to two types of events,
both of which are triggered in the context of $apply:

 * An event handler when the user selects a group

 * A callback from angular-websocket when a WebSocket message
   is received.


nickstenning:
LGTM.

PR #2689, opened 2015-10-29
Avoid losing unsaved changes to new or existing annotations when switching groups

robertknight:
This PR fixes two issues related to creating or editing annotations and switching groups:

 * I create a new annotation, realize that I have the wrong group selected and then switch to the intended group. In master, the annotation remains in the previous group and I have to (a) delete it and (b) re-create it in the new group
 * I start editing an annotation, then switch to another group. In master, the unsaved changes are lost with no warning. At least when you logout you get a prompt as to whether you want to keep or discard changes.

The end-user behavior of this PR is:

 * When switching groups, new annotations are moved to the newly selected group and the text and tags that has been entered so far are preserved.
 * When switching groups, in-progress edits to existing annotations are kept. The draft will not be shown when viewing the new group but will re-appear with unsaved edits intact when switching back to the original group.

The implementation involves:

 * Replacing the drafts service with one which stores not just the identities of annotations
   which are being edited also the changes to text and tags of the draft.
 * Not discarding drafts when an annotation is unloaded
 * Moving new annotations to the newly focused group when the focused group changes
 
I also added some documentation on what the threading service is for whilst I was grokking the code.

----
TODO
* [X] Check if not reloading the view when the focused group changes is something that can be split out into a separate PR. I originally thought it was essential but now I'm not so sure. _Update: Yes, it looks like this is not critical and can be moved into a separate PR_
* [X] Look into an issue where, when switching groups, new annotations are posted to the currently selected group, but the 'Post to' button does not update. _Update: This is an existing bug in the Post button. I'll fix it and clean up the code for that in a separate PR: #2691_
* [X] Add tests to check that the annotation controller saves and restores drafts as expected
* [X] Check that the behavior of discarding changes on logout is consistent with what we had before. _This was broken but is fixed by the last commit_


h/static/scripts/drafts.js
```diff
@@ -0,0 +1,84 @@
+/**
+ * The drafts service provides temporary storage for unsaved edits
+ * to new or existing annotations.
+ *
+ * A draft consists of a 'model' which is the original annotation
+ * which the draft is associated with and `changes' which is
+ * a set of edits to the original annotation.
+ */
+function DraftStore() {
```
robertknight: The name of the service is inconsistent with the name of the module. Perhaps rename `drafts` to `draft-store`?

seanh:
This one:

1. Create an annotation in a group, type some text and tags, but don't save it
2. Change to another group
3. Change back to the first group
4. Result: draft annotation it still there, but its contents are lost

is also fixed by <https://github.com/hypothesis/h/pull/2693>. Although I see that in _this_ pull request, the behaviour when there is an unsaved draft annotation and you change to another group is that the annotations goes with you and changes between groups, instead of staying behind in the first group and waiting for you to return.

That's a UX change, is it something we want? (My vote is yes.)

seanh:
> Although I see that in _this_ pull request, the behaviour when there is an unsaved draft annotation and you change to another group is that the annotation goes with you and changes between groups, instead of staying behind in the first group and waiting for you to return.

Hmm, but when _editing_ and already saved annotation, if I change between groups without saving, the annotation remains behind in the original group. My edits are not lost, they remain when I return. But this means that the behaviour when editing is different than when creating. Is that ok?

seanh:
Small issue:

1. Edit an annotation, change its privacy e.g. from Only Me to Group
2. Change the focus to another group
3. Change the focus back to the original group. The annotation that is being edited has its privacy set to Only Me again

seanh:
> Hmm, but when editing and already saved annotation, if I change between groups without saving, the annotation remains behind in the original group. My edits are not lost, they remain when I return. But this means that the behaviour when editing is different than when creating. Is that ok?

This is may be particularly confusing when doing both at once.

1. Focus group one.
2. Create annotation one and save it.
3. Edit annotation one, but don't save it.
4. Focus group two. Annotation one is _not_ shown, it remains behind in group one
5. Create annotation two, but don't save it.
6. Focus group one. You now see annotation one and annotation two, both being edited, both say they're in group one.
7. Focus group two. You see annotation two only, and now it says that it's in group two.

seanh:
<https://github.com/hypothesis/h/pull/2693> will fix the problem with unsaved changes being lost when switching groups while creating or editing an annotation. At least if the user remembers to go back to the original group, they won't lose their work.

I wonder whether that's enough? And introducing different behaviours for creating and editing as this pull request does is complicating things a little too much and making it a little too confusing.

I know that the UX here is still far from perfect and it's still quite easy to lose unsaved work by navigating to another page, but suspect that a proper design iteration of the sidebar to make creating and editing annotations more modal, and prevent moving between groups while creating or editing, is what's really needed to sort this out.

robertknight:
>  I wonder whether that's enough? 

For me, and based on the comments of other staff members I think the most important issue to fix is the case where I start creating an annotation in the main page - which is the natural starting point for the activity, then go to the sidebar and realise that I want to post this annotation to a different group.

I think the behaviour for editing vs. creating makes sense - new annotations are always posted to a currently visible group but once saved, annotations don't move groups. We can revisit that easily enough in future.

robertknight:
> 1. Edit an annotation, change its privacy e.g. from Only Me to Group
> 2. Change the focus to another group
> 3. Change the focus back to the original group. The annotation that is being edited has its privacy set to Only Me again

I noticed this as well but I think it would be better to fix this after your permissions handling refactoring is done.

nickstenning:
LGTM!

judell:
This is wonderful!

I noticed the following. On http://journals.plos.org/plosone/article?id=10.1371/journal.pone.0132735 if I begin an annotation in a group (where none exist) and then switch it to Public (where many exist) the new annotation does not have focus, it's necessary to scroll down to find it.

PR #2690, opened 2015-10-29
Query tags field with a match query (use mapping analyzer)

### LINKED ISSUE #2655: api query for tag with uppercase succeeds only if lowercase
Tags in use for that URI:

{
  "tags": [
    "LT2015"
  ]
},
{
  "tags": [
    "LT2015",
    "JJJS"
  ]
}

https://hypothes.is/api/search?tags=LT2015

{"rows": [], "total": 0}

https://hypothes.is/api/search?tags=lt2015

{ "rows":[...], "total": 22 }

I did not think we were case-sensitive at all, never mind oppositely so. I'll add some tests to the query correctness suite.

nickstenning:
The "tags" field is stored in Elasticsearch analysed (by our custom "uni_normalizer" analyzer). This means, broadly, that in the inverted index, individual terms in a tag will be stripped of punctuation and case-normalised.  As such, we need to do the same normalisation when querying, and the easiest way to do this is with a match query.

This commit changes "tag" queries so that:

- all tag parameters must match
- multiple tokens within one tag query are all required to match

By way of example, for an annotation with tags:

    ["hello", "#THERE", "big world"]

These queries will match:

    tag=hello
    tag=HELLO
    tag=#hello
    tag=hello&tag=there
    tag=hello&tag=there&tag=big+world
    tag=hello&tag=there&tag=world+big

Whereas these will not:

    tag=he+llo
    tag=hello&tag=monkeys
    tag=hello&tag=world

Fixes #2655.

### DIFF
--- a/h/api/search/query.py
+++ b/h/api/search/query.py
@@ -197,4 +197,6 @@ class TagsMatcher(object):
             del params['tags']
         except KeyError:
             pass
-        return {'terms': {'tags': [tag for tag in tags]}} if tags else None
+        matchers = [{'match': {'tags': {'query': t, 'operator': 'and'}}}
+                    for t in tags]
+        return {'bool': {'must': matchers}} if matchers else None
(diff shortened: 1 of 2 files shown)


seanh:
Lgtm, just waiting for Travis

PR #2691, opened 2015-10-29
Fix 'Post to' button not updating when an annotation's group changed & cleanup code

robertknight:
The 'group' property on the scope used by the 'Post to' button
was set when the component was instantiated but not updated
if the group subsequently changed. This is not directly visible in master because
an annotation's group never changes, but relates to #2689 

Fix this and simplify the code by using the 'bindToController'
facility from Angular 1.3.x to avoid copying properties
from the directive's scope to the controller.

This means that 'vm.group' expressions in the controller now
update when the corresponding attribute changes.

### DIFF
--- a/h/static/scripts/directive/publish-annotation-btn.js
+++ b/h/static/scripts/directive/publish-annotation-btn.js
@@ -1,43 +1,5 @@
 'use strict';
 
-// @ngInject
-function PublishAnnotationBtnController($scope) {
-  var vm = this;
-
-  vm.group = {
-    name: $scope.group.name,
-    type: $scope.group.public ? 'public' : 'group'
-  };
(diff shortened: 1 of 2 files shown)


h/static/scripts/directive/publish-annotation-btn.js
```diff
@@ -47,7 +9,23 @@ function PublishAnnotationBtnController($scope) {
 // @ngInject
 module.exports = function () {
   return {
-    controller: PublishAnnotationBtnController,
+    bindToController: true,
```
robertknight: This is a new feature in Angular 1.3 which makes properties in a directive's isolate scope accessible on the controller.

nickstenning:
LGTM.

PR #2693, opened 2015-10-29
Avoid route reload on change

robertknight:
Avoid reloading whole view when focused group changes

When the focused group changes, instead of reloading the whole
view just unload the currently displayed annotations and then
reload the annotations for the newly selected group(s) and
any drafts of unsaved annotations.

The only immediately 'visible' effect of the change is
that switching groups now only results in a call to /search
rather than network requests to fetch session state and features.

In future this should make it easier to preserve state
for the area below the top bar when switching groups.

### DIFF
--- a/h/static/scripts/app-controller.coffee
+++ b/h/static/scripts/app-controller.coffee
@@ -46,8 +46,7 @@ module.exports = class AppController
       options: ['Newest', 'Oldest', 'Location']
     }
 
-    # Reload the view when the focused group changes or the
-    # list of groups that the user is a member of changes
+    # Reload the view when the user switches accounts
     reloadEvents = [events.USER_CHANGED];
     reloadEvents.forEach((eventName) ->
       $scope.$on(eventName, (event, data) ->
--- a/h/static/scripts/app-controller.coffee
+++ b/h/static/scripts/app-controller.coffee
@@ -48,7 +48,7 @@ module.exports = class AppController
 
     # Reload the view when the focused group changes or the
     # list of groups that the user is a member of changes
-    reloadEvents = [events.USER_CHANGED, events.GROUP_FOCUSED];
+    reloadEvents = [events.USER_CHANGED];
     reloadEvents.forEach((eventName) ->
       $scope.$on(eventName, (event, data) ->
         if !data || !data.initialLoad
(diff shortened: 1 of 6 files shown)


dwhly:
This makes me really happy.

h/static/scripts/test/widget-controller-test.coffee
```diff
@@ -1,16 +1,20 @@
...
   fakeAnnotationUI = null
   fakeAuth = null
   fakeCrossFrame = null
+  fakeDrafts = null
   fakeStore = null
   fakeStreamer = null
   fakeStreamFilter = null
   fakeThreading = null
   fakeGroups = null
+  lastSearchResult = null
```
seanh: This is only used by one test, does it need to be at such a high scope? I think a lot of our tests (especially client-side ones) are pretty hard to understand because to understand a given test method I need to understand not just that test method but lots of other bits of code spread throughout the test module that it depends on. And sometimes these shared fake objects that are shared between tests include fairly detailed fake behaviour that couples tests together, since changing the fake can break many tests at once.

If just one test needs to save the last search result it might be clearer if that test included the code for doing that, rather than adding it to the shared fake.

seanh:
Looks good to me, I've tested it and it seems to work well

seanh:
> The only immediately 'visible' effect of the change is
> that switching groups now only results in a call to /search
> rather than network requests to fetch session state and features.

It also means that the contents of draft annotations are not lost, when you switch to another group and back again :)

h/static/scripts/test/widget-controller-test.coffee
robertknight: I'll have a look and see if I can refactor along these lines. I haven't felt the pain you mentioned in the client-side tests myself yet but I can see the hazard.

h/static/scripts/app-controller.coffee
```diff
@@ -46,9 +46,8 @@ module.exports = class AppController
       options: ['Newest', 'Oldest', 'Location']
     }
 
-    # Reload the view when the focused group changes or the
-    # list of groups that the user is a member of changes
-    reloadEvents = [events.USER_CHANGED, events.GROUP_FOCUSED];
+    # Reload the view when the user switches accounts
+    reloadEvents = [events.USER_CHANGED];
```
nickstenning: I'd probably go the whole hog and get rid the `forEach` below, as well.

nickstenning:
This is mergeable. Please have a look at following up on the comments here. Good work, @robertknight!

PR #2696, opened 2015-10-30
Add AnnotationHTMLPresenter

seanh:
This splits h.api.models.Annotation into two classes:

1. h.api.models.Annotation, which contains only model code and no presentation
   code.

   This _does_ contain code for papering over the variability of our annotation
   data. For example it has a .uri property that always returns a unicode
   string even if the underlying annotation dict has no "uri" key or has a key
   whose value isn't a string.

2. h.AnnotationHTMLPresenter, a wrapper class for h.api.models.Annotation that:

   a. Overrides some properties with HTML-specific behaviour.
      For example it overrides .uri to provide an HTML-escaped copy.

   b. Adds some new HTML-specific properties. For example it adds a
      .document_link property that returns an escaped, HTML-formatted link
      for the annotation's document.

Many Annotation tests are now moved to AnnotationHTMLPresenter tests.

feed/atom.py, feeds/rss.py and groups/views.py are updated to use
AnnotationHTMLPresenter.

Tests for these modules need to be updated to mock AnnotationHTMLPresenter
behaviour, which is unfortunately quite awkward and requires a mock class.


seanh:
The tests are still using PropertyMock because they're copied over from existing tests, and I'm not refactoring this aspect of the tests in this pr.

seanh:
This can be reviewed now, but since it adds no user-visible features and potentially introduces bugs, I'd consider not merging refactors like this one until after the groups release next week.

h/annotation_html_presenter.py
```diff
@@ -0,0 +1,284 @@
...
+
+import jinja2
+
+
+def _format_document_link(href, title, link_text, hostname):
+    """Return a document link for the given components.
+
+    Helper function for the .document_link property below.
+
+    The given href, title, link_text and hostname are assumed to be already
```
tilgovi: If this comment is true, why are they escaped at the end of this function?

seanh: It's an out of date comment, fixing..

seanh:
Rebased

h/__init__.py
```diff
@@ -1,4 +1,9 @@
 # -*- coding: utf-8 -*-
+from h.annotation_html_presenter import AnnotationHTMLPresenter
```
nickstenning: I don't think there's any need to expose this here.

h/annotation_html_presenter.py
```diff
@@ -0,0 +1,284 @@
+"""A class that wraps h.api.models.Annotation and adds some HTML properties."""
```
nickstenning: I'm not too keen on ridiculously long module names. I'd be inclined to call this `h.presenters`. That way it can turn into a package at some point in the future without having to rename imports all over the codebase.

h/feeds/atom.py
```diff
@@ -56,6 +58,8 @@ def feed_from_annotations(
     :rtype: dict
 
     """
+    annotations = [h.AnnotationHTMLPresenter(a) for a in annotations]
```
nickstenning: It doesn't make sense to be using something called `AnnotationHTMLPresenter` in a module that isn't rendering to HTML. I"m also not sure at all about having to then dereference the original annotation with `annotation.annotation` earlier.

What's the need to use the presenter at all in this module?

h/feeds/rss.py
```diff
@@ -58,6 +59,8 @@ def feed_from_annotations(annotations, annotation_url, rss_url, html_url,
     :rtype: dict
 
     """
+    annotations = [h.AnnotationHTMLPresenter(a) for a in annotations]
```
nickstenning: Ditto as per my comment on the Atom renderer.

h/feeds/atom.py
nickstenning: Oh, I think I see why you've done this. It's because `.title` has moved to the presenter class. In which case this is easily solved by moving the logic sections of `title` back onto the model, leaving the `title` property on the presenter to be something like:

```python
@property
def title(self):
    return jinja2.escape(self.annotation.title)
```

nickstenning:
Looks pretty good overall. Main issues are:

- [ ] Put this in an `h.presenters` module or similar. `h.annotation_html_presenter` is a bit of a mouthful.
- [ ] There shouldn't be logic in something called "*HTMLPresenter" that isn't fundamentally to do with escaping for/rendering to HTML. As such, there shouldn't be any reason to use "AnnotationHTMLPresenter" when rendering RSS or Atom feeds (although I guess there might conceivably be a need for an RSSPresenter or an AtomPresenter).

seanh:
Ok, I've renamed the module. The RSS and Atom feeds actually do contain HTML - they both contain an HTML-formatted description of the annotation (which now comes from `AnnotationHTMLPresenter.description`).

RSS and Atom feeds would also want fields such as title to be escaped, wouldn't they? Although in this case we're escaping it for output to XML not HTML I guess, not sure exactly what the difference would be off hand.

It could be factored into AnnotationRSSPresenter and AnnotationHTMLPresenter..

nickstenning:
LGTM. I do still have some niggles about the way `AnnotationHTMLPresenter` is used in `h.feeds`, but they're not a big deal.

PR #2697, opened 2015-10-30
Add request context to errors caught by Sentry

nickstenning:
At the moment errors caught and processed by Sentry contain almost no useful information about where they came from save the stack trace.

This commit adds basic request information (URL, method, etc.) to errors thrown from within the application.

In addition, we now log 404s for informational purposes.

I've also added an explanatory comment to the middleware stack configured in `production.ini` so that the role of the raven middleware is understood.


nickstenning:
For client-side logging, this makes it trivial to expose the DSN in `app.html`'s settings object. The relevant value is `request.sentry.get_public_dsn()`.

PR #2699, opened 2015-10-30
Fix search button not expanding search bar in Safari

### LINKED ISSUE #2654: Clicking on search icon in new "top bar" does nothing
In Safari 9.0 on Mac OS X 10.11, clicking on the magnifying glass in the top bar,

image

doesn't do anything. No errors logged to the console.

robertknight:
The search bar expands from 0 to 150px when the input is focused
but in Safari the <input> will not accept focus via
input.focus() if its max-width is _exactly_ 0px.

Setting it to a near-zero +ve value instead avoids the problem.

Fixes #2654

### DIFF
--- a/h/static/styles/simple-search.scss
+++ b/h/static/styles/simple-search.scss
@@ -32,7 +32,13 @@
     padding: 0px;
     width: 100%;
 
-    max-width: 0px;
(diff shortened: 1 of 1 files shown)


nickstenning:
LGTM.

Issue #1412, opened 2014-08-18
Collapsed annotations still show tags.


dwhly:
Tags are no longer collapsed.
![screen shot 2014-08-18 at 3 58 04 pm](https://cloud.githubusercontent.com/assets/1413074/3959359/412899ec-272b-11e4-97b9-9a593402bd20.png)

PR #2704, opened 2015-11-02
Fix display of num replies on collapsed replies

### LINKED ISSUE #2702: "n replies" text displayed right next to annotation date on collapsed reply
A small visual bug -- if you collapse a reply with replies of its own you get this:

screen shot 2015-10-30 at 18 52 55

seanh:
When you collapse a reply that itself has replies an "n replies" link
appears next to the username on the collapsed reply. A recent commit
broke the display/alignment of this. Correct it.

Fixes #2702.

### DIFF
--- a/h/templates/client/annotation.html
+++ b/h/templates/client/annotation.html
@@ -15,6 +15,14 @@
            ng-href="{{vm.baseURI}}u/{{vm.annotation.user}}"
            >{{vm.annotation.user | persona}}</a>
       </span>
+
+      <span class="annotation-collapsed-replies">
+        <a class="reply-count small" href=""
+          ng-click="replyCountClick()"
+          ng-pluralize count="replyCount"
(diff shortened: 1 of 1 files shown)


seanh:
What I'm seeing now:

untitled screencast - cropped 6

nickstenning:
Great! It may not ultimately be what we want it to look like but it's certainly better than what it does right now.

seanh:
It's the same as how it used to work as well

nickstenning:
Is it? I though "n replies" was on the next line. Oh well, bad memory. Even better :)

dwhly:
> I though "n replies" was on the next line.

What we had before collapsed to a single line.  With the introduction of the group line, we've now lost that.  I'd argue to bring it back, for the same reasons here when some thread work introduced a second line as well: https://github.com/hypothesis/h/issues/1412

Will make a trello card for it.

PR #2706, opened 2015-11-02
Python 3 compatibility

nickstenning:
This PR doesn't make h work on Python 3 (for one thing, we'll need to wait for gevent 1.1...) but it takes the first few steps towards compatibility.

Not to be merged just yet.

### DIFF
--- a/gunicorn.conf.py
+++ b/gunicorn.conf.py
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
 import os
-import urlparse
+from h._compat import urlparse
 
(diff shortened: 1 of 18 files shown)


tilgovi:
Do you need _compat? I would just use the `six` library. It's already a transitive dependency anyway.

seanh:
Ooh, glad to see the start of this

seanh:
@nickstenning Rebased this and fixed a couple more things. Take a look at h/api/models.py and h/presenters.py in particular, and if you're happy then merge

nickstenning:
Looks good.

PR #2708, opened 2015-11-02
Delete unsaved drafts on sign out

### LINKED ISSUE #2701: Phantom highlight/bucket bar entries created when signing out with unsaved annotations
Reported by Nick whilst testing the change that moves new, unsaved annotations between groups when switching focus.

1. create a draft annotation while signed in,
2. sign out (accepting the dialog that asks if you want to discard drafts)
the highlight and bucket bar entry remains, although the annotation isn't visible in the sidebar
signing back in doesn't bring the annotation back
looks like the "annotationDeleted" event either isn't being fired or isn't propagating to the hosting page
separately, I note that for some reason the drafts service says "reply"/"replies" when it means any kind of annotation: "You have an unsaved reply." 

Related issue: The dialog prompting to discard drafts says "reply"/"replies" when it means any kind of annotation: "You have an unsaved reply."

seanh:
Fixes #2701.

### DIFF
--- a/h/static/scripts/app-controller.coffee
+++ b/h/static/scripts/app-controller.coffee
@@ -5,14 +5,14 @@ parseAccountID = require('./filter/persona').parseAccountID
 
 module.exports = class AppController
   this.$inject = [
-    '$controller', '$document', '$location', '$route', '$scope', '$window',
-    'annotationUI', 'auth', 'drafts', 'features', 'groups', 'identity',
-    'session'
+    '$controller', '$document', '$location', '$rootScope', '$route', '$scope',
+    '$window', 'annotationUI', 'auth', 'drafts', 'features', 'groups',
+    'identity', 'session',
   ]
   constructor: (
-     $controller,   $document,   $location,   $route,   $scope,   $window,
-     annotationUI,   auth,   drafts,   features,   groups,   identity,
-     session
+     $controller,   $document,   $location,   $rootScope,   $route,   $scope,
+     $window,   annotationUI,   auth,   drafts,   features,   groups,
+     identity,   session
   ) ->
     $controller('AnnotationUIController', {$scope})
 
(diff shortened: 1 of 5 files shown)


h/static/scripts/drafts.js
```diff
@@ -19,6 +19,14 @@ function DraftStore() {
   }
 
   /**
+   * Returns a list of all drafts - both unsaved new annotations, and unsaved
+   * edits to saved annotations.
+   */
+  this.drafts = function drafts() {
+    return this._drafts.slice();
```
nickstenning: AFAICT you only ever use this to count the drafts. Perhaps just make a `count` function that returns `this._drafts.length` instead of copying the entire array?

h/static/scripts/app-controller.coffee
```diff
@@ -7,12 +7,12 @@ module.exports = class AppController
...
     '$controller', '$document', '$location', '$route', '$scope', '$window',
     'annotationUI', 'auth', 'drafts', 'features', 'groups', 'identity',
-    'session'
+    'session', '$rootScope'
   ]
   constructor: (
      $controller,   $document,   $location,   $route,   $scope,   $window,
      annotationUI,   auth,   drafts,   features,   groups,   identity,
-     session
+     session, $rootScope
```
nickstenning: I'm a pain I know, but `$rootScope` should go after `$location`.

h/static/scripts/app-controller.coffee
```diff
@@ -101,11 +101,25 @@ module.exports = class AppController
...
+    # Prompt to discard any unsaved drafts.
+    promptToLogout = ->
+      # TODO - Replace this with a UI which doesn't look terrible.
+      if drafts.drafts().length == 1
+        text = 'You have an unsaved annotation.\n
+                Do you really want to discard this draft?'
+      else if drafts.drafts().length > 1
+        text = 'You have ' + drafts.drafts().length + ' unsaved annotations.\n
+                Do you really want to discard these drafts?'
+      return (drafts.drafts().length == 0 or window.confirm(text))
```
nickstenning: s/`window`/`$window`/g

h/static/scripts/test/app-controller-test.coffee
```diff
@@ -156,3 +161,67 @@ describe 'AppController', ->
     fakeRoute.reload = sinon.spy()
     $scope.$broadcast(events.USER_CHANGED, {initialLoad: false})
     assert.calledOnce(fakeRoute.reload)
+
+  describe 'logout()', ->
+    it 'prompts the user if there are drafts', ->
+      callCount = window.confirm.callCount
```
nickstenning: How does this even work? Where is `window.confirm` stubbed...?

The answer is [in `annotation-test.coffee`](https://github.com/hypothesis/h/blob/e970fa8bc94f75229bf7b988f67108375360521e/h/static/scripts/directive/test/annotation-test.coffee#L401), and this only works because:

- that test doesn't clean up after itself properly (it should be using `sandbox.stub` not `sinon.stub`)
- this test runs later than that one

So, in short:

- Use `$window`, not `window`, in the controller.
- Use an angular-injected fake to provide a fake `$window` object, not a global stub.
- If you don't mind, fix up the other test (perhaps in a different PR) to not leave `window.confirm` stubbed after the test has run.

nickstenning:
LGTM.

PR #2709, opened 2015-11-02
Appropriately encode the tab URL for a query string

### LINKED ISSUE #2648: UnicodeDecodeError: 'utf8' codec can't decode byte 0xd3 in position 118: invalid continuation byte
https://app.getsentry.com/hypothesis/prod/group/81642856/

```
UnicodeDecodeError: 'utf8' codec can't decode byte 0xd3 in position 118: invalid continuation byte
(19 additional frame(s) were not displayed)
...
  File "webob/request.py", line 835, in GET
    vars = GetDict(data, env)
  File "webob/multidict.py", line 287, in __init__
    MultiDict.__init__(self, data)
  File "webob/multidict.py", line 38, in __init__
    items = list(args[0])
  File "webob/compat.py", line 125, in parse_qsl_text
    yield (x.decode(encoding), y.decode(encoding))
  File "encodings/utf_8.py", line 16, in decode
    return codecs.utf_8_decode(input, errors, True)
```

This is quite mysterious and I'm not entirely sure what it refers to, but it would seem to be thrown by a view handling JSON as the actual logging comes from `h.views:json_error`.

It seems likely to have been introduced by 998347a, given the time when it was first seen.

nickstenning:
Tab URLs can contain all kinds of weird things, including:

- query strings (`?foo=bar`)
- percent-encoded strings that do not correspond to valid UTF-8 sequences as mandated by RFC3986 (recently seen: `/ac%FD-pul-biber`, an example of ISO-8859-9 percent-encoding)

We need to ensure that when we pass the URL to the badge endpoint we pass a valid encoded string.

Fixes #2648.

### DIFF
--- a/h/browser/chrome/lib/tab-state.js
+++ b/h/browser/chrome/lib/tab-state.js
@@ -25,6 +25,11 @@ var DEFAULT_STATE = {
   ready: false,
 };
 
+/** encodeUriQuery encodes a string for use in a query parameter */
+function encodeUriQuery(val) {
+  return encodeURIComponent(val).replace(/%20/g, '+');
+}
+
 /** TabState stores the H state for a tab. This state includes:
  *
  * - Whether the extension has been activated on a tab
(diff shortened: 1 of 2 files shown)


h/browser/chrome/lib/tab-state.js
```diff
@@ -25,6 +25,11 @@ var DEFAULT_STATE = {
   ready: false,
 };
 
+/** encodeUriQuery encodes a string for use in a query parameter */
+function encodeUriQuery(val) {
+  return encodeURIComponent(val).replace(/%20/g, '+');
```
tilgovi: Why the `.replace()`?

PR #2714, opened 2015-11-03
Simplify the group share links

seanh:


seanh:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;7xTKdVNH&#x2F;155-link-to-group-page-from-scope-selector"> Link to group page from scope selector</a>

h/static/scripts/directive/test/group-list-test.js
```diff
@@ -122,20 +102,9 @@ describe('groupList', function () {
...
-    var toggleLink = element.find('.share-link-toggle');
-    var expander = element.find('.share-link-expander');
-    assert.ok(isElementHidden(expander));
-    toggleLink.click();
-    assert.ok(!isElementHidden(expander));
-    toggleLink.click();
-    assert.ok(isElementHidden(expander));
+    var link = element.find('.share-link');
+    assert.equal(link.length, 1);
+    assert.equal(link[0], GROUP_LINK);
```
nickstenning: This is a neat trick... I was looking at this thinking "How on earth can that possibly pass? `link[0]` is a DOM Element, and `GROUP_LINK` is a string...!?"

Well, it turns out `String(anchor_element)` coerces to the `href` of the element.

While this does indeed allow this test to pass, I think it might be good to be a bit more explicit with what you're testing here.

seanh: The diff here is confusing. This pr deletes some no longer relevant tests, does not add any test code.

The neat trick here is used a few times throughout this test file

nickstenning:
LGTM.

judell:
Thanks Sean!

PR #2718, opened 2015-11-04
Change perms when moving annotations btwn groups

### LINKED ISSUE #2713: Post control not following scope change.
Go to a page, with scope as Public.
Select text.
Initiate an annotation.
Editor shows.
Now change scope to a group.
Post control shows "Only Me", instead of group name.

Video here:
https://drive.google.com/file/d/0B0yW-dSdEYAcZnZheFlEZm5yeG8/view?usp=drivesdk

Screenshot:
screenshot 2015-11-02 16 21 06

seanh:
Fixes #2713.

Needs test.

### DIFF
--- a/h/static/scripts/directive/annotation.coffee
+++ b/h/static/scripts/directive/annotation.coffee
@@ -425,17 +425,22 @@ AnnotationController = [
       if !vm.editing
         return
 
+      # move any new annotations to the currently focused group when
+      # switching groups. See GH #2689 for context
+      if !model.id
+        newGroup = groups.focused().id
+        if permissions.isShared(vm.annotation.permissions, vm.annotation.group)
+          model.permissions = permissions.shared(newGroup)
+          vm.annotation.permissions = model.permissions
+        model.group = newGroup
+        vm.annotation.group = model.group
+
       # if we have a draft, update it, otherwise (eg. when the user signs out)
       # do not create a new one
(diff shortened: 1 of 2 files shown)


seanh:
Test done.

h/static/scripts/directive/annotation.coffee
```diff
@@ -429,7 +429,11 @@ AnnotationController = [
       # move any new annotations to the currently focused group when
       # switching groups. See GH #2689 for context
       if !model.id
-        model.group = groups.focused().id
+        debugger
```
nickstenning: Probably not :)

seanh:
Rebased on latest master with debugger statement removed

seanh:
Ok this doesn't seem to fix the issue anymore...

seanh:
This is ready now

seanh:
This is ready now. It's a bit ugly (having to change both `model` and `@annotation` together) but it does actually seem to be necessary to preserve all the functionality, and at least we know that this pr isn't introducing any `model`/`@annotation` sync issues.

To be not horrible I think it needs this refactoring https://github.com/hypothesis/h/issues/2728

nickstenning:
Good stuff. Sorting out AnnotationController needs to happen sooner rather than later. It's a disaster area right now.

seanh:
Yep

PR #2719, opened 2015-11-06
Replace angular-websocket with a minimal WebSocket wrapper & fix push notifications after reconnection

robertknight:
This PR fixes two issues:
 * H does not build with the current version of Node JS (>= 4.x)
 * Clients never received annotation push notifications if they lost the connection to the server and then automatically reconnected.

The first issue with the build is due to an indirect and unnecessary dependency on an old version of `bufferutil` via `angular-websocket` -> `ws 0.7.x`. The old version of bufferutil requires a native compile build step and is not compatible with the version of V8 shipped with current NodeJS

I'm going to submit a PR upstream to `angular-websocket` to update the dependency, but it was completely pointless in our case since that code was never used in a browser context.

The commit message has more details, but essentially angular-websocket provided us with 3 things:
 * Automatic reconnection with exponential backoff
 * Queuing of messages sent prior to connection
 * Wrapping `WebSocket.onmessage` callbacks with `$rootScope.$apply`

This commit replaces `angular-websocket` with a minimal wrapper which accomplishes the first using the well-established `retry` module, the second via a trivial queue system and the third by wrapping the on-message handler in `streamer.js` with `$rootScope.$apply`. In addition to removing the dependency, it also shows better error messages.

The second issue occurred because after an automatic reconnection, the configuration messages specifying which annotations the client should receive was not re-sent to the server. This commit stores the configuration data as part of the service client and resends configuration messages after a reconnection.


robertknight:
Marking this as WIP as I'm currently doing some more extensive testing of this first.

h/static/scripts/test/streamer-test.js
```diff
@@ -67,26 +73,28 @@ describe('streamer', function () {
   });
 
   it('should send a client ID', function () {
-    assert.equal(socket.messages.length, 1);
-    assert.equal(socket.messages[0].messageType, 'client_id');
-    assert.equal(socket.messages[0].value, streamer.clientId);
+    var fakeWebSocket = socket._socket;
```
nickstenning: If we're proxyquiring the `Socket` constructor I'm not sure I see why we need to expose `_socket` like this. Can't we just reference the return value of `new FakeSocket`?

nickstenning:
I have one question about exposing the socket singleton for use in tests, and I'm also a bit concerned that the dependency on EventEmitter is another 5K of code in our client.

`app.min.js` on `master` is 541K minified, and on this branch is 555K.

Thoughts?

nickstenning:
https://www.npmjs.com/package/tiny-emitter might be worth considering...

robertknight:
I've slimmed down the dependencies by replacing EventEmitter with 'tiny-emitter' and `util.inherits` with 'inherits' directly (util.inherits is just an alias for the "inherits" module internally). `app.min.js` measures 546KB in my build.

We could introduce JS/CSS bundle size checks as part of the Travis CI build, although that would be easier to do once the JS build chain is separated out from Python.

I've replaced the private `_socket` export as suggested.

nickstenning:
LGTM.

PR #2725, opened 2015-11-18
Don't update timestamps of unsaved annotations

seanh:
Fix a bug where the updateTimestamp() function in AnnotationController would
try to update the @timestamp property of new annotations that have no
model.updated timestamp because they haven't been saved to the server yet.
This result in nonsense values being generated and updateTimestamp() being
called multiple times per second for each unsaved annotation.


nickstenning:
Ta!

PR #2731, opened 2015-11-20
Faster database tests

nickstenning:
Tests that touched the database were doing a full drop and recreation of the entire schema on every test. This meant that even a no-op test took an appreciable fraction of a second to run.

Slow tests mean sad developers, so this commit changes how we run such tests: the database schema is dropped and recreated only once -- at the start of the test session -- and test isolation is instead guaranteed using PostgreSQL's SAVEPOINT feature.

For each test function, we run the following in the database:

1. BEGIN; -- Open a new transaction. This is implicit.
2. SAVEPOINT some_savepoint_name; -- Establish a savepoint within the current transaction. For more on savepoints, see [the PostgreSQL docs][1].
3. Run the test function.
4. ROLLBACK TO SAVEPOINT some_savepoint_name; -- Roll the state of the database back to that at the savepoint.

This results in database tests that are between 5 and 10x faster, and our overall test suite runs in less than half the time. (On my machine a full `py.test` run is down to 11-12s from 30-35s).

[1]: http://www.postgresql.org/docs/current/static/sql-savepoint.html

### DIFF
--- a/h/accounts/test/models_test.py
+++ b/h/accounts/test/models_test.py
@@ -6,20 +6,21 @@ import re
 import pytest
 from sqlalchemy import exc
 
+from h import db
 from h.accounts import models
 from h.test import factories
 
 
-def test_activation_has_asciinumeric_code(db_session):
+def test_activation_has_asciinumeric_code():
     act = models.Activation()
 
-    db_session.add(act)
-    db_session.flush()
+    db.Session.add(act)
+    db.Session.flush()
 
     assert re.match(r'[A-Za-z0-9]{12}', act.code)
 
 
-def test_cannot_create_dot_variant_of_user(db_session):
+def test_cannot_create_dot_variant_of_user():
     fred = models.User(username='fredbloggs',
                        email='fred@example.com',
                        password='123')
@@ -27,13 +28,13 @@ def test_cannot_create_dot_variant_of_user(db_session):
                         email='fred@example.org',
                         password='456')
 
-    db_session.add(fred)
-    db_session.add(fred2)
+    db.Session.add(fred)
+    db.Session.add(fred2)
     with pytest.raises(exc.IntegrityError):
-        db_session.flush()
+        db.Session.flush()
 
 
-def test_cannot_create_case_variant_of_user(db_session):
+def test_cannot_create_case_variant_of_user():
     bob = models.User(username='BobJones',
                       email='bob@example.com',
                       password='123')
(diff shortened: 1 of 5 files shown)


h/conftest.py
```diff
@@ -69,6 +69,36 @@ def settings():
...
+    Prepare the SQLAlchemy session object.
+
+    We enable fast repeatable database tests by setting up the database only
+    once per session (see :func:`setup_database`) and then wrapping each test
+    function in a SAVEPOINT/ROLLBACK TO SAVEPOINT within the transaction.
+    """
+    savepoint = transaction.savepoint()
+
+    def _cleanup():
+        db.Session.remove()
```
robertknight: Since `db.Session.remove` has been monkey-patched to be a no-op below, I think this line is a no-op?

h/accounts/test/models_test.py
```diff
@@ -6,20 +6,21 @@
 import pytest
 from sqlalchemy import exc
 
+from h import db
 from h.accounts import models
 from h.test import factories
 
 
-def test_activation_has_asciinumeric_code(db_session):
```
robertknight: I'm not sure about the change to autouse the `database_session` fixture. Passing it in as an argument made it easier IMO to trace the setup/teardown code that happens around each test.

h/conftest.py
nickstenning: The honorable gentleman makes an excellent point. It's worth noting that `monkeypatch`es are automatically undone by a finalizer, but in this case there's no guarantee that they'll be undone in the right order. Additionally, `savepoint.rollback()` will ensure that the session is restored to the correct state anyway, so it's not necessary here.

h/accounts/test/models_test.py
seanh: So we can have _every_ test (even those that don't even use the db) do the savepoint and rollback with _no_ noticeable slowdown? Nice.

Not sure how I feel about the `db_session` being `autouse` vs explicit, I could go with it either way. I understand Robert's point but at the same time not having it `autouse` just introduces the possibility of forgetting to put the fixture on one test and introducing a test bug. Also this fixture is very generic code that should rarely need to be changed, or be the cause of a test failure.

seanh:
The tests are much faster for me too, nice :)

h/accounts/test/models_test.py
nickstenning: The database session is not injected or configured explicitly in application code, so I'm not sure why we'd necessarily want it to be so in the tests. I agree that there's a "too much magic" line for test setup/teardown code, but I don't think this is on the far side of it -- there's a bunch of other code that is `autouse`d and I don't think we'd like what our tests looked like if we had to explicitly invoke those fixtures for every test.

PR #2733, opened 2015-11-20
Remove old feature flag data

nickstenning:
When a feature flag is removed from the codebase, it will remain in the database. This could potentially cause very surprising issues in the event that a feature flag with the same name (but a different meaning) is added at some point in the future.

This commit adds a function which removes unknown feature flags from the database, and configures it to run at application startup.

### DIFF
--- a/h/features.py
+++ b/h/features.py
@@ -1,10 +1,15 @@
 # -*- coding: utf-8 -*-
 from __future__ import unicode_literals
 
+import logging
+
 from pyramid.view import view_config
 import sqlalchemy as sa
+import transaction
 
 from h.db import Base
+
+log = logging.getLogger(__name__)
 
 FEATURES = {
     'claim': "Enable 'claim your username' web views?",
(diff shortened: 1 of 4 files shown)


h/features.py
```diff
@@ -108,8 +142,8 @@ def features_status(request):
 
 
 def includeme(config):
+    # Remove old feature flags from the database on startup
+    config.action(None, remove_old_flags, order=90)
```
robertknight: Is there a central place that the order values for configuration actions could be defined, to make it more explicit what the groupings of actions are and in what order those groups run?

h/features.py
```diff
@@ -66,6 +72,9 @@ def get_by_name(cls, name):
         """Fetch a flag by name."""
         return cls.query.filter(cls.name == name).first()
 
+    def __repr__(self):
+        return '<Feature %s everyone=%s>' % (self.name, self.everyone)
```
seanh: I'm curious why you still use `%s` string formatting? It's supposed to be [going away eventually in Python 3](https://docs.python.org/3.2/tutorial/inputoutput.html#old-string-formatting), isn't it?

nickstenning: Old habits die hard.

h/features.py
nickstenning: It's not supposed to be an absolute ordering, and pyramid modules are supposed to be (mostly) independent of one another, so no. I'm not sure the value of such a thing would outweigh the indirection costs at the moment.

PR #2737, opened 2015-11-23
Fix the missing threads in the stream

### LINKED ISSUE #1916: Too many message not available on stream

seanh:
- [x] The search API now returns top-level annotations only, not replies.
- [x] If an undocumented `_include_replies: True` argument is passed then it returns an additional `"replies"` key containing all replies to the top-level annotations in the `"rows"` key
- [x] Up to 100 replies are returned. There is not pagination of the `"replies"` list. If an annotation set has more than 100 replies, log a warning.
- [x] No longer copy `scope` from parent annotations into replies, not used anymore
- [x] Revert <https://github.com/hypothesis/h/commit/5a5b4e345c050a71e619fb5732426b2306f29e64> (which replaced the _Message not available_ cards in the stream with simply hiding the threads with missing annotations)

This just does exactly what <https://github.com/hypothesis/h/pull/2483> did except it also makes returning the replies optional and off by default. I re-did it manually because rebasing the three month old branch was too messy.

Some discussion about how we might iterate on this API later in this comment: <https://github.com/hypothesis/h/pull/2483#issuecomment-138887155>. I'd just add that when searching, we probably want search queries to be matched against annotations and their replies and if an annotation or any of its replies matches a query then return the whole thread.

Fixes #1916.

### DIFF
--- a/h/api/search/core.py
+++ b/h/api/search/core.py
@@ -9,7 +9,7 @@ from h.api.search import query
 log = logging.getLogger(__name__)
 
 
-def search(request, params, private=True):
+def search(request, params, private=True, separate_replies=False):
     """
     Search with the given params and return the matching annotations.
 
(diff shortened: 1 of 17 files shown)


seanh:
Renamed `include_replies` -> `separate_replies` and made it so that if this option is _not_ given then it reverts to the old behaviour: just the one rows list containing both annotations and replies. Rebased.

h/api/search/core.py
```diff
@@ -18,28 +23,62 @@ def search(request, params, private=True):
...
+    if separate_replies:
+        builder.append_filter(query.TopLevelAnnotationsFilter())
+    results = models.Annotation.search_raw(builder.build(params),
                                            raw_result=True,
                                            authorization_enabled=False)
 
     total = results['hits']['total']
     docs = results['hits']['hits']
     rows = [models.Annotation(d['_source'], id=d['_id']) for d in docs]
+    return_value = {"rows": rows, "total": total}
```
nickstenning: Minor niggle, but `return_value` isn't a particularly helpful name. I can see it's a return value because it follows the `return` keyword, but that doesn't tell me what it is.

nickstenning:
Looks good. Let's circle back once we've got a better grip on the annotation/reply domain model and improve this then.

PR #2738, opened 2015-11-24
Enable staff to look up user's email addresses

seanh:
Allow staff to access the /admin index page and to use some of the admin pages
within (so far only the /admin/users page).

Assign a different permission to each admin page ('admin_features',
'admin_nipsa' etc) then on the root resource assign some of these permissions
to admins and some to staff (or to both admins and staff).

This enables staff users to access the admin index page and some of the admin
pages within.

Then tweak the admin template to only show links to the admin pages that the
user has permission to use.

The "admin" permission, previously assigned to admin users, is removed since
it's no longer used.

### DIFF
--- a/h/admin.py
+++ b/h/admin.py
@@ -13,7 +13,7 @@ from h import util
 @view.view_config(route_name='admin_index',
                   request_method='GET',
                   renderer='h:templates/admin/index.html.jinja2',
-                  permission='admin')
+                  permission='admin_index')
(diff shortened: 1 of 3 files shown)


seanh:
All the code changes here are either declarative or template code, so I didn't add any tests

nickstenning:
Very nice. Thank you!

lenazun:
🙌

PR #2739, opened 2015-11-24
Enable highlights by default everywhere

### LINKED ISSUE #2695: Enable highlights by default for embedded H
**Update: Decision here is that we should enable highlights by default for embedded H, but keep the config option to disable it. _The homepage documentation will need to be updated accordingly_**

Hypothesis is embedded in our documentation at `h.readthedocs.org` but highlights are off by default, which means that useful comments for newcomers will likely go unnoticed.

We should turn them on by default.

That also raises the question of perhaps whether the default setting should be to have highlights enabled by default and require them to be turned _off_ if users don't want that. There might already have been a long discussion on that subject that I wasn't party to. @judell ?

seanh:
Enable highlights by default in the Chrome extension, bookmarklet and embed.

They were already enabled by default in the Chrome extension and
bookmarklet so this only changes the embed.

Highlights are turned on by default in annotator/host.coffee if no
showHighlights option is found.

The extension and bookmarklet no longer need to set showHighlights: true as
this is now the default. The embed (which was not setting showHighlights: true
and so was getting the default setting of false) will now get the new default
setting of true.

For people embedding h into their own pages, they can turn off highlights by
default by including:

    <script>
      window.hypothesisConfig=function(){return{showHighlights:false}};
    </script>

fixes #2695

### DIFF
--- a/h/static/extension/config.js
+++ b/h/static/extension/config.js
@@ -4,6 +4,5 @@ function hypothesisConfig() {
   var hypothesisIntent = document.querySelector('[name="hypothesis-intent"]');
   return {
     firstRun: hypothesisIntent && hypothesisIntent.content === 'first-run',
-    showHighlights: true
   };
 }
--- a/h/static/scripts/annotator/host.coffee
+++ b/h/static/scripts/annotator/host.coffee
@@ -26,7 +26,10 @@ module.exports = class Host extends Guest
 
     this.on 'panelReady', =>
       # Initialize tool state.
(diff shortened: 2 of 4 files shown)


seanh:
Note the embed instructions on the hypothes.is front page need updating when this is deployed.

nickstenning:
> Note the embed instructions on the hypothes.is front page need updating when this is deployed.

Although I'm guessing that you can take care of this with #2740?

nickstenning:
LGTM.

PR #2740, opened 2015-11-25
Move front page from WordPress into app

seanh:
- [x] Deal with `main.js`, probably just figure out exactly what js stuff is actually used on the front page and re-write it
- [x] Put a `rel="feed"` link for the blog feed in the `<head>`
- [x] Make the login etc links in the top-right work
- [x] Google Analytics?
- [x] Integrate home.html.jinja2 with base.html.jinja2


seanh:
@robertknight Want to review this?

The minified CSS file needs to be recreated as a nice, minimal one but I suggest doing that as a separate PR. When we do that we can also migrate the `front_page_js` and `front_page_css` bundles into the site one (or maybe some stuff will remain that's only needed on the front page).

In the meantime I think we can go ahead and publish this front page as-is if we need to?

I'm getting a couple of errors in the console that it's failing to load some fonts, but I'm also seeing these in production, I'm guessing it's from the minified CSS file and will disappear when we sort that out.

h/templates/home.html.jinja2
```diff
@@ -0,0 +1,305 @@
...
+      <link rel="stylesheet" href="{{ ASSET_URL }}">
+  {% endassets %}
+{% endblock %}
+
+{% block content %}
+  <header class="banner hiring">
+    <div class="container">
+      <span>
+        We're hiring
+        <a href="{{ base_url }}jobs/#developer">developers</a>.
```
nickstenning: Is there a good reason to have all these links as `{{ base_url }}...`? Can't they just be relative links?

h/templates/layouts/base.html.jinja2
```diff
@@ -29,6 +29,35 @@
     {% endfor -%}
     {% block styles %}{% endblock %}
 
+    <link rel="apple-touch-icon" sizes="57x57"
```
nickstenning: Yeesh. If @robertknight could look at this and decide which of these we actually need, that would be great. And if it still ends up being loads, I don't think a `{% for size in [57, 60, 72, 76...] %}` would go amiss here.

h/views.py
```diff
@@ -95,13 +95,11 @@ def widget(context, request):
     return {}
 
 
-@view_config(renderer='h:templates/help.html.jinja2', route_name='index')
 @view_config(renderer='h:templates/help.html.jinja2', route_name='help')
 @view_config(renderer='h:templates/help.html.jinja2', route_name='onboarding')
 def help_page(context, request):
     current_route = request.matched_route.name
     return {
-        'is_index': current_route == 'index',
```
nickstenning: There's a bunch of code you can now remove from `templates/help.html.jinja2`.

h/assets.yaml
```diff
@@ -220,6 +220,7 @@ site_css:
   filters: cleancss
   contents:
     - h:static/styles/vendor/icomoon.css
+    - h:static/styles/vendor/css/font-awesome.css
```
nickstenning: Perhaps I'm missing something, but I can't see a reason why fontawesome needs to be included in `site_css`?

nickstenning:
Not sure about copying the compiled `h/static/styles/front-page/main.css` into this repository. [This](https://github.com/hypothesis/wordpress-theme-hypothesis/blob/master/assets/styles/main.scss) is where it's built from, but I highly doubt we need all of that for just the front page. Even if we do, we should probably be using `bootstrap_css` as already provided rather than adding another copy of bootstrap to the repository.

h/templates/home.html.jinja2
seanh: It `{{ base_url }}` isn't the root URL of a domain? But not sure if we care about that

h/views.py
seanh: Done

h/assets.yaml
seanh: Oops, that slipped in by accident, rebased it out

seanh:
> Not sure about copying the compiled h/static/styles/front-page/main.css into this repository. This is 
> where it's built from, but I highly doubt we need all of that for just the front page. Even if we do, we 
> should probably be using bootstrap_css as already provided rather than adding another copy of
> bootstrap to the repository.

Yes, my suggestion was that we could fix that in a separate pr though, possibly at the same time as implementing Conor's new design, and possibly to be done by Robert instead of me

seanh:
I think this is just waiting to hear from Robert about Apple icons now

h/templates/layouts/base.html.jinja2
robertknight: Just the 152x152 (or 144x144) icons should be sufficient - see https://mathiasbynens.be/notes/touch-icons and https://github.com/h5bp/html5-boilerplate/issues/1367

The tiny downside is that older devices download the larger image and then rescale it.

seanh: Rebased with just the one apple-touch-icon. The Android icon and the three different sizes of favicon are still there.

PR #2744, opened 2015-11-26
Remove unused `deleted` property handling

nickstenning:
At some point in the distant past (7f36f3a) annotation deletes were handled by adding the "deleted" property to annotations and anonymising them. Now we actually delete them.

This commit removes code that was intended to handle such annotations.

There were a grand total 11 such annotations remaining in the database. I have taken a backup copy of these and deleted them.


robertknight:
This cleared up a couple of pending questions I had.

PR #2745, opened 2015-11-26
Validate incoming annotation data with jsonschema

nickstenning:
As part of my spike looking into migrating to Postgres, I'm investigating what a minimal interface with which the `h.api` package could interact with to save and fetch annotation data, moving that responsibility out of `h.api`.

This PR replaces some custom validation logic with a completely extensible use of `jsonschema` to validate incoming annotation data, and moves some of the validation from `h.api.logic` into the schema.

There's obviously a lot more we can do in terms of validating annotation data, and `jsonschema` also potentially allows us to return much better error messages when things do go wrong. For now, this is about the smallest thing we could do to get started along that road.

Feedback welcome.

### DIFF
--- a/h/api/schemas.py
+++ b/h/api/schemas.py
@@ -2,7 +2,7 @@
 """Classes for validating data passed to the annotations API."""
 
 
-class Error(Exception):
+class ValidationError(Exception):
 
     """Base exception class for all exceptions raised by this module."""
 
@@ -14,7 +14,7 @@ class Annotation(object):
     """A validator for annotations."""
 
     def validate(self, data):
-        """Raise h.api.validation.Error if the data is invalid."""
+        """Raise ValidationError if the data is invalid."""
         if 'document' in data and 'link' in data['document']:
             if not isinstance(data['document']['link'], list):
-                raise Error("document.link must be an array")
+                raise ValidationError("document.link must be an array")
--- a/h/api/schemas.py
+++ b/h/api/schemas.py
@@ -6,6 +6,10 @@ import jsonschema
 from jsonschema.exceptions import best_match
 
 
+# These annotation fields are not to be set by the user.
+PROTECTED_FIELDS = ['created', 'updated', 'user', 'id']
+
+
 class ValidationError(Exception):
     pass
 
@@ -59,6 +63,15 @@ class AnnotationSchema(JSONSchema):
             },
         },
     }
+
+    def validate(self, data):
+        appstruct = super(AnnotationSchema, self).validate(data)
+
+        # Some fields are not to be set by the user, ignore them
+        for field in PROTECTED_FIELDS:
+            appstruct.pop(field, None)
+
+        return appstruct
 
 
 def _format_jsonschema_error(error):
--- a/h/api/schemas.py
+++ b/h/api/schemas.py
@@ -1,20 +1,70 @@
 # -*- coding: utf-8 -*-
 """Classes for validating data passed to the annotations API."""
 
+import copy
+import jsonschema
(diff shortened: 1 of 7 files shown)


nickstenning:
For the sake of consistency with the rest of the codebase, I've moved `h.api.validators` to `h.api.schemas`, so this diff is probably easier to read if you check out the branch and use `git diff -C`.

h/api/schemas.py
```diff
@@ -0,0 +1,79 @@
...
+
+    def __init__(self):
+        self.validator = jsonschema.Draft4Validator(self.schema)
+
+    def validate(self, data):
+        """
+        Validate `data` according to the current schema.
+
+        :param data: The data to be validated
+        :return: valid data
```
seanh: Does this always return the input data as it was given? Or does it sometimes modify it?

h/api/schemas.py
```diff
@@ -0,0 +1,79 @@
...
+            result.pop(field, None)
+
+        return result
+
+
+def _format_jsonschema_error(error):
+    """Format a :py:class:`jsonschema.ValidationError` as a string."""
+    if error.path:
+        dotted_path = '.'.join(error.path)
+        return '{0}: {1}'.format(dotted_path, error.message)
```
seanh: Note that `{1}` is not as good as `{error_message}` when a translator comes to localize a string. Of course this string will not be internationalized but it might be good for consistency reasons to always format strings in the same way everywhere?

seanh:
I'll have to familiarize myself with JSON Schema, but this approach looks good to me

seanh:
Was reading up on JSON Schema, the [Space Telescope book](http://spacetelescope.github.io/understanding-json-schema/) is good. It looks pretty good.

It might be nice to be explicit about the [two phases of validation](http://spacetelescope.github.io/understanding-json-schema/about.html). Maybe the `JSONSchema` class's `validate()` can call a `validate_structure()` method that validates against `self.schema` then if that succeeds call an empty `validate_semantics()` method. Subclasses like `AnnotationSchema` could then just set `self.schema` and override `validate_semantics()`.

Also noticed that you're popping keys in `validate()` which is mutating the data, not validating it. It might be nice to also explicitly split this into a separate phase. Maybe an empty `prepare()` method that gets called before the validation methods and subclasses override it. Arguably mutating data isn't part of validation at all and should be done by the calling code before passing the data into validation. If it is part of the validation code just probably good to separate it out explicitly, and also to be explicit in the docstrings: the returned data is valid _and may have been modified_. Also does it return a modified copy of the given data, or modify the given data in place? Probably either is fine but good to document which it is.

h/api/schemas.py
nickstenning: My intent was that validate should be referentially transparent, so I suppose it should make a copy and return that.

h/api/schemas.py
nickstenning: Sure thing.

nickstenning:
So you've made a very good point about modifying the input data to `validate()`. That isn't intentional and I'll fix this here.

As for splitting `validate_structure` and `validate_semantics`, I'd suggest we reconsider that once we have a bit more going on in here and understand what the code is actually going to look like.

We have a very similar pattern of code in our Colander form validators, where they do boring data structure and format validations in the superclass's `validate`, and then check model validity afterwards. See [`EmailChangeSchema`](https://github.com/hypothesis/h/blob/34aa9390f76ae5906c62257d3eb10d1e82cb3b2a/h/accounts/schemas.py#L226-L249) for an example of this.

judell:
Thanks _so_ much for this! It's an important first step.

PR #2746, opened 2015-11-27
Fix sending of WebSocket messages enqueued during a disconnection

robertknight:
Any WebSocket messages submitted whilst the WS is disconnected
should be re-sent when the WS reconnects.

This was broken because upon a disconnect, the 'onclose'
event handler first synchronously called reconnect(), resulting
in a new WebSocket instance being created and then set 'socket'
to null. This caused an error in sendMessages() after the 'onopen'
event was triggered.

Fix the problem by not using a null 'socket' var to represent
a disconnected socket. Instead just use the WebSocket's readyState
property.


nickstenning:
:horse:

PR #2747, opened 2015-11-27
Keyboard focus the login form when shown

seanh:
Automatically put the keyboard focus on the username field in the login
form when it's shown.

### DIFF
--- a/h/static/scripts/app.coffee
+++ b/h/static/scripts/app.coffee
@@ -109,6 +109,7 @@ module.exports = angular.module('h', [
 .directive('formInput', require('./directive/form-input'))
 .directive('formValidate', require('./directive/form-validate'))
 .directive('groupList', require('./directive/group-list').directive)
+.directive('hAutofocus', require('./directive/h-autofocus'))
 .directive('markdown', require('./directive/markdown'))
 .directive('simpleSearch', require('./directive/simple-search'))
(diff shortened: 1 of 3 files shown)


h/static/scripts/directive/h-autofocus.js
```diff
@@ -0,0 +1,22 @@
+'use strict';
+
+/** An attribute directive that focuses an <input> when it is shown.
```
robertknight: Need to re-word this slightly to clarify when it is triggered. s/shown/"added to the DOM".

If the `<input>` was shown with `ng-show`, the current implementation would not work. The current implementation is nice and simple so rather than changing it to work with `ng-show`, I'd just re-word this sentence slightly.

h/static/scripts/app.coffee
```diff
@@ -123,6 +123,7 @@ module.exports = angular.module('h', [
 .directive('signinControl', require('./directive/signin-control'))
 .directive('sortDropdown', require('./directive/sort-dropdown'))
 .directive('topBar', require('./directive/top-bar'))
+.directive('hAutofocus', require('./directive/h-autofocus'))
```
robertknight: I'd like to keep this list sorted alphabetically (there are a couple of other items here out of place, I suggest just re-sort the complete list)

seanh:
Done and rebased

PR #2756, opened 2015-11-30
Translate annotation.coffee to JavaScript

seanh:


h/static/scripts/directive/annotation.js
```diff
@@ -0,0 +1,730 @@
...
+    message = reason.status + ' ' + reason.statusText;
+    if (reason.data.reason) {
+      message = message + ': ' + reason.data.reason;
+    }
+  }
+  return message;
+}
+
+/**
+  * @ngdoc type
```
robertknight: Does anyone at H actually use ngdoc documentation? If not, I'd be tempted to get rid of it - just keep the JSDoc comments.

h/static/scripts/directive/annotation.js
```diff
@@ -0,0 +1,730 @@
...
+  // Export the baseURI for the share link.
+  vm.baseURI = $document.prop('baseURI');
+
+  $scope.$on('$destroy', function() {
+    updateTimestamp = angular.noop;
+  });
+
+  // Watch for changes to the domain model and recreate the view model when it
+  // changes.
+  // XXX: TODO: don't clobber the view when collaborating.
```
robertknight: This comment looks obsolete to me

h/static/scripts/directive/annotation.js
```diff
@@ -0,0 +1,730 @@
...
+        updateDraft(model);
+      }
+    }
+
+    updateTimestamp(model === old);  // Repeat on first run.
+    vm.render();
+  }, true);
+
+  $scope.$on(events.USER_CHANGED, function() {
+    if (model.user === null) {
```
robertknight: Use `if (!model.user)` instead of explicit checks for null or undefined.

h/static/scripts/directive/annotation.js
```diff
@@ -0,0 +1,730 @@
...
+
+  vm.action = 'view';
+  vm.document = null;
+  vm.editing = false;
+  vm.isSidebar = false;
+  vm.preview = 'no';
+  vm.timestamp = null;
+
+  model = $scope.annotationGet();
+  if (model.user === undefined) {
```
robertknight: Use `if (!model.user)` instead of checks for undefined/null equality.

h/static/scripts/directive/annotation.js
```diff
@@ -0,0 +1,730 @@
...
+  /**
+    * @ngdoc method
+    * @name annotation.AnnotationController#edit
+    * @description Switches the view to an editor.
+    */
+  vm.edit = function() {
+    if (!drafts.get(model)) {
+      updateDraft(model);
+    }
+    vm.action = model.id !== null ? 'edit' : 'create';
```
robertknight: Use `model.id ?` rather than check against null.

h/static/scripts/directive/annotation.js
```diff
@@ -0,0 +1,730 @@
...
+
+  /**
+  * Update the given annotation domain model object with the data from the
+  * given annotation view model object.
+  */
+  updateDomainModel = function(domainModel, viewModel) {
+    var i;
+    var tagTexts = [];
+    for (i = 0; i < viewModel.tags.length; i++) {
+      tagTexts.concat(viewModel.tags[i].text);
```
robertknight: Array.concat() does not mutate its argument, it returns a new array.

This could be written more simply as:

```js
var tagTexts = viewModel.tags.map(function (tag) { return tag.text });
```

Also, there is a typo in the argument to angular.extend. Should be `tags` not `tag`.

h/static/scripts/directive/annotation.js
```diff
@@ -0,0 +1,730 @@
...
+    * @description
+    * Creates a new message in reply to this annotation.
+    */
+  vm.reply = function() {
+    var id = model.id;
+    var references = model.references || [];
+    var reply;
+    var uri = model.uri;
+
+    if (typeof references === 'string') {
```
robertknight: Can 'references' ever actually be a string?

h/static/scripts/directive/annotation.js
```diff
@@ -0,0 +1,730 @@
...
+    }
+  };
+
+  /**
+    * @ngdoc method
+    * @name annotation.AnnotationController#render
+    * @description Called to update the view when the model changes.
+    */
+  vm.render = function() {
+    var documentTitle;
```
robertknight: I'm guessing that the reason the 'var' statements were hoisted to the top of the function is a result of the CoffeeScript compiler since CoffeeScript doesn't differentiate between declaration and definition.

In JS I'd try to put the declaration and definition together (ie. move 'var domain' down to the point where it is actually used and change to 'var domain ='). This will make it easier to use let/const (from ES2015) in future - const in particular is useful because it prevents unintended re-assignment of variables. It also makes it easier for type checkers and linters to infer the type of a variable.

h/static/scripts/directive/annotation.js
```diff
@@ -0,0 +1,730 @@
...
+        title: domain
+      };
+    }
+
+    if (vm.document.title.length > 30) {
+      vm.document.title = vm.document.title.slice(0, 30) + '…';
+    }
+
+    // Form the tags for ngTagsInput.
+    tagsAsObjects = [];
```
robertknight: This could be written more simply as:
```js
vm.annotation.tags = (vm.annotation.tags || []).map(function (tag) {
  return { text: tag };
});
```

Its a bit confusing though that we first set 'annotation.tags' to a string array and then change it to an array of `{ text: string }`

h/static/scripts/directive/annotation.js
```diff
@@ -0,0 +1,730 @@
...
+          link = model.document.link[i];
+          if (!(link.href.indexOf('urn'))) {
+            continue;
+          }
+          uri = link.href;
+          break;
+        }
+      }
+
+      documentTitle = Array.isArray(
```
robertknight: Hmm, I wonder how many annotations in the DB have multiple titles and how many store the title as string vs. array?

h/static/scripts/directive/annotation.js
```diff
@@ -0,0 +1,730 @@
...
+    if (!repeat) {
+      return;
+    }
+
+    fuzzyUpdate = time.nextFuzzyUpdate(model.updated);
+    nextUpdate = (1000 * fuzzyUpdate) + 500;
+
+    $timeout(function() {
+      updateTimestamp(true);
+      $scope.$digest();
```
robertknight: OK, so presumably the idea here is to set `$timeout`'s 'apply' parameter to false to avoid a `$rootScope.$apply` and instead only digest the current scope. That's a sensible thing to do from a performance point of view, but could use a comment.

In general, I think it is a good idea to put a comment after any non-obvious boolean parameters to note what they are for, since JS doesn't have keyword arguments (eg. `$timeout(..., false /* apply */)`)

h/static/scripts/directive/annotation.js
```diff
@@ -0,0 +1,730 @@
...
+      draftDomainModel = {};
+      updateDomainModel(draftDomainModel, vm.annotation);
+      updateDraft(draftDomainModel);
+    }
+  });
+
+  return vm;
+}
+
+// @ngInject
```
robertknight: Does `// @ngInject` work if it isn't directly above the function expression? Even if it does, it would be more consistent to put it just above the function or method declaration.

h/static/scripts/directive/test/annotation-test.js
```diff
@@ -0,0 +1,1377 @@
...
+        'finally': angular.noop
+      });
+    });
+
+    it('persists upon login', function() {
+      delete annotation.id;
+      delete annotation.user;
+      fakeSession.state.userid = null;
+      createDirective();
+      $scope.$digest();
```
robertknight: This `$scope.$digest()` call could be hoisted into `createDirective()`, if it isn't there already.

h/static/scripts/directive/test/annotation-test.js
```diff
@@ -0,0 +1,1377 @@
...
+      controller.reply();
+      match = sinon.match({
+        references: [annotation.id],
+        uri: annotation.uri
+      });
+      assert.calledWith(fakeAnnotationMapper.createAnnotation, match);
+    });
+
+    it('makes the annotation shared if the parent is shared', function() {
+      var reply;
```
robertknight: Again, I'm guessing this comes from the CoffeeScript compiler but I'd suggest going through and merging definitions and declarations where this is easy to do.

h/static/scripts/directive/test/annotation-test.js
```diff
@@ -0,0 +1,1377 @@
...
+          read: [group]
+        };
+      };
+      controller.reply();
+      assert(reply.permissions.read.indexOf('my group') !== -1);
+    });
+
+    it(
+      'does not add the world readable principal if the parent is private',
+      function() {
```
robertknight: Small formatting nit. For all hand-written JS we have a space between the `function` keyword and the parenthesies in function expressions. We should be be consistent in these files as well.

h/static/scripts/directive/test/annotation-test.js
```diff
@@ -0,0 +1,1377 @@
...
+
+    it(
+      'calls annotationMapper.delete() if the delete is confirmed',
+      function(done) {
+        sandbox.stub($window, 'confirm').returns(true);
+        fakeAnnotationMapper.deleteAnnotation.returns($q.resolve());
+        controller['delete']().then(function() {
+          assert(fakeAnnotationMapper.deleteAnnotation.calledWith(annotation));
+          done();
+        });
```
robertknight: Missing a `.catch()`  block here. Alternatively, return the promise from the function instead of using `done` and the test harness should handle this for you.

h/static/scripts/directive/test/annotation-test.js
```diff
@@ -0,0 +1,1377 @@
...
+
+    it(
+      'emits annotationCreated when saving an annotation succeeds',
+      function(done) {
+        sandbox.spy($rootScope, '$emit');
+        annotation.$create.returns(Promise.resolve());
+        controller.save().then(function() {
+          assert($rootScope.$emit.calledWith('annotationCreated'));
+          done();
+        });
```
robertknight: Missing a `.catch()` block here, or returning the promise from the function - which is the method I would recommend.

h/static/scripts/directive/test/annotation-test.js
```diff
@@ -0,0 +1,1377 @@
...
+      function(done) {
+        annotation.$create.returns(Promise.reject({
+          status: 0
+        }));
+        controller.save().then(function() {
+          assert(fakeFlash.error.calledWith(
+            'Service unreachable.', 'Saving annotation failed'));
+          done();
+        });
+        $timeout.flush();
```
robertknight: Hmm, does `$timeout.flush()` actually work here given that we're using a native Promise, not `$q`. In general, I would be inclined to use native promises rather than `$q` going forwards. Given that we're going to have native promises (or polyfills) anyway since they are part of the web platform, we might as well have only one implementation instead of two.

h/static/scripts/directive/test/annotation-test.js
```diff
@@ -0,0 +1,1377 @@
...
+      id: 'new-group'
+    });
+    $rootScope.$broadcast(events.GROUP_FOCUSED);
+    assert.deepEqual(
+      annotation.permissions.read, ['acct:bill@localhost'],
+      'The annotation should still be private');
+  });
+});
+
+describe('AnnotationController', function() {
```
robertknight: It looks like this set of tests might be duplicating a bunch of stuff from the annotation directive tests. I presume fixing that is part of the refactoring?

h/static/scripts/directive/annotation.js
seanh: We don't ever compile it afaik, but we do have ngdoc (not just jsdoc) comments all over our code currently and it does have some [Angular-specific directives](https://github.com/angular/angular.js/wiki/Writing-AngularJS-Documentation#standard-supported-jsdoc-directives), particularly ones like `service`, `filter` and `directive`, that I think it might be worth keeping it for as long as we're using Angular.

h/static/scripts/directive/annotation.js
seanh: Hmm, also I have no idea what that comment is talking about, I'll remove it

h/static/scripts/directive/annotation.js
seanh: Done

h/static/scripts/directive/annotation.js
seanh: Done. Incidentally this is how I write in Python as well - making use of truthiness and falsiness in boolean expressions, but @nickstenning often requests me to change it to make it explicit (`if foo is False` etc)

h/static/scripts/directive/annotation.js
seanh: How do you feel about these ternary expressions anyway? I find them really hard to read, would rather have a bigger `if` / `else`

h/static/scripts/directive/annotation.js
seanh: A bit disturbing that this didn't cause any tests to fail. On the other hand `updateDomainModel()` doesn't have any unit tests, so it's not surprising.

h/static/scripts/directive/annotation.js
seanh: Dunno, but this is in the original `CoffeeScript` so I'm gonna vote that this pull request shouldn't change it. It likely has to do with old annotations in our "database" from back in a day when references was a string.

h/static/scripts/directive/annotation.js
seanh: Ah, I was following Douglas Crockford's advice to put all your `var`s at the top of the function in JavaScript, because that's where they're scoped to. Do you want all variables to be defined where they're first used instead, then?

h/static/scripts/directive/annotation.js
seanh: Ah, good to know about `map()`. As for why we do this nonsense in the first place - yes it is confusing :)

h/static/scripts/directive/annotation.js
seanh: Yeah. Stuff like this in our client-side code is really bad. Even if we do have documents in the db with array titles, hopefully we'll soon be able to handle that server-side and always provide annotations in a consistent format to API clients

h/static/scripts/directive/annotation.js
seanh: Done

h/static/scripts/directive/test/annotation-test.js
seanh: Hmm, gonna leave this as I'm translating the CoffeeScript not refactoring it

h/static/scripts/directive/test/annotation-test.js
seanh: Really? Our linter jscs complains if you do that

h/static/scripts/directive/test/annotation-test.js
seanh: Good catch, but this is from the CoffeeScript. I think I'll do a separate commit to get rid of all the `done()`s

h/static/scripts/directive/test/annotation-test.js
seanh: Hmm. It doesn't appear to be necessary, but it was there in the CoffeeScript for some reason (presumably). Might as well remove it

h/static/scripts/directive/test/annotation-test.js
seanh: Possibly. This set of tests was meant to demonstrate a better way of writing Angular tests. Instead of having dozens of global variables and shared `beforeEach()` behaviour at the top of the `describe()` and having to jump up and down in the file to understand what's happening, it allows each test to be much more independently understandable by calling helper functions to get just the things that it needs instead. Basically all of the variables in a given test are actually defined in the test function itself (though they may be defined by calling a helper function and getting values that it returns).

If we wanted to pick this new style though, we're left with either translating all our client tests to it or putting up with having two different styles of Angular tests, so I'm tempted to ditch it.

h/static/scripts/directive/annotation.js
seanh: Done

h/static/scripts/directive/annotation.js
seanh: Done

h/static/scripts/directive/annotation.js
seanh: Done

h/static/scripts/directive/annotation.js
robertknight: Agree - perhaps you could make a note to check though. We might be able to simplify this post-migration.

h/static/scripts/directive/annotation.js
robertknight: Generally yes, except for variables first used inside loops where Crockford's point is very valid.

h/static/scripts/directive/test/annotation-test.js
robertknight: It does? In that case our linter is definitely not consistent with most of the code we've been writing recently.

h/static/scripts/directive/test/annotation-test.js
robertknight: The solution to this I think is simply to have smaller units with fewer dependencies.

h/static/scripts/directive/annotation.js
```diff
@@ -0,0 +1,701 @@
...
+  /**
+    * @ngdoc method
+    * @name annotation.AnnotaitonController#hasContent
+    * @returns {boolean} True if the currently edited annotation has
+    *          content (ie. is not just a highlight)
+    */
+  vm.hasContent = function() {
+    var textLength = (vm.annotation.text || '').length;
+    var tagsLength = (vm.annotation.tags || []).length;
+    return (textLength > 0 || tagsLength > 0);
```
robertknight: Skimming this now I can see that the code is inconsistent about whether it relies on implicit conversions from numbers to booleans. Despite what I said about `if (<expr>)` being idiomatic JS, I still think it is preferable to use explicit tests for `number > 0` rather than relying on 0 being falsey.

Don't need to make changes now though.

h/static/scripts/directive/annotation.js
```diff
@@ -0,0 +1,701 @@
...
+    }
+  };
+
+  /**
+  * Update the given annotation domain model object with the data from the
+  * given annotation view model object.
+  */
+  function updateDomainModel(domainModel, viewModel) {
+    var i;
+    var tagTexts = [];
```
robertknight: I'd still like to see this refactored to use a simple map over `viewModel.tags`

h/static/scripts/directive/annotation.js
```diff
@@ -0,0 +1,701 @@
...
+  /**
+    * @ngdoc method
+    * @name annotation.AnnotationController#authorize
+    * @param {string} action The action to authorize.
+    * @returns {boolean} True if the action is authorized for the current user.
+    * @description Checks whether the current user can perform an action on
+    * the annotation.
+    */
+  vm.authorize = function(action) {
+    if (model === null) {
```
robertknight: Reading the code above, it doesn't look possible for `model` to be null here, that check should be removed.

h/static/scripts/directive/annotation.js
```diff
@@ -0,0 +1,701 @@
...
+    }
+
+    if (!validate(vm.annotation)) {
+      return flash.info('Please add text or a tag before publishing.');
+    }
+
+    // Update stored tags with the new tags of this annotation.
+    var newTags = vm.annotation.tags.filter(function(tag) {
+      var tags = model.tags || [];
+      return (tags.indexOf(tag.text) === -1);
```
robertknight: Wrapping the expression in parentheses is unnecessary here

h/static/scripts/directive/annotation.js
```diff
@@ -0,0 +1,701 @@
...
+
+    // Extract the document metadata.
+    var uri = model.uri;
+    var domain = new URL(uri).hostname;
+    if (model.document) {
+      if (uri.indexOf('urn') === 0) {
+        var i;
+        for (i = 0; i < model.document.link.length; i++) {
+          var link = model.document.link[i];
+          if (!(link.href.indexOf('urn'))) {
```
robertknight: This looks wrong. AIUI the purpose of this block of code is to extract a URL (as opposed to a URI) from the model to provide a clickable link.

This is a case where I would definitely not rely on number -> boolean conversions. The idea here is to find the first `<link>` which has a scheme other than `urn`, so this `if` expression will evaluate to _true_ and skip the link if it starts with "urn". This is correct, and actually the original CoffeeScript had the opposite incorrect behavior:

```coffee
          # This URI is not clickable, see if we have something better
          for link in model.document.link when link.href.indexOf("urn")
            uri = link.href
            break
```

In this case, it would be much clearer if written as:

```js
if (link.href.indexOf('urn') === 0) {
  continue;
}
```

robertknight: Strictly speaking, that could still be incorrect since the protocol could start with 'urn' but have something else after it, we should actually be checking for 'urn:' instead.

h/static/scripts/directive/annotation.js
seanh: I'd say let's do this as part of the refactoring to come after. This pr is just supposed to be cs -> js translation, and this check exists in the cs

h/static/scripts/directive/annotation.js
seanh: Done

h/static/scripts/directive/annotation.js
seanh: Leaving this for now (and for my opinion: when dealing with `0` I'd rather consistently use `if (number)` than `if (number > 0)`)

h/static/scripts/directive/annotation.js
seanh: Done

h/static/scripts/directive/annotation.js
seanh: I've changed it to `if (link.href.indexOf('urn:') === 0)` (note the `:`)

PR #2758, opened 2015-12-01
Add 'Annotating All Knowledge' banner to the homepage

robertknight:
Add a temporary banner to the legacy homepage design
until the new homepage is ready to roll.

### DIFF
--- a/h/templates/home.html.jinja2
+++ b/h/templates/home.html.jinja2
@@ -17,6 +17,29 @@
 {% endblock %}
 
 {% block content %}
+  <style>
+    .press-release-banner {
+      background-color: #f7f7f7;
+      border: 1px solid #ccc;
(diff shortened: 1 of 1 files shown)


h/templates/home.html.jinja2
```diff
@@ -239,6 +262,9 @@
 
           </div>
         </div>
+        <div class="press-release-banner">
+            <p>Hypothes.is announces a coalition of over 40 scholarly organizations bringing annotation to all knowledge. <a href="https://hypothes.is/annotating-all-knowledge/"><b>Learn more</b></a></p>
```
nickstenning: "Hypothesis" (no dot) if it's not the logotype.

robertknight: Done

PR #2761, opened 2015-12-02
Refactor annotation controller, part 1

seanh:
Here I'm just fixing various low-hanging fruits / tidying up before attempting more serious refactoring

### DIFF
--- a/h/static/scripts/directive/annotation.js
+++ b/h/static/scripts/directive/annotation.js
@@ -328,9 +328,6 @@ function AnnotationController(
     * the annotation.
     */
   vm.authorize = function(action) {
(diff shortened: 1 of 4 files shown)


h/static/scripts/directive/annotation.js
```diff
@@ -162,9 +163,13 @@ function AnnotationController(
 
   vm.action = 'view';
   vm.document = null;
-  vm.editing = false;
-  vm.isSidebar = false;
-  vm.preview = 'no';
+  if ($scope.isSidebar === 'true') {
```
robertknight: Why is `$scope.isSidebar` a string?

PR #2762, opened 2015-12-03
Refactor saving highlights in AnnotationController

seanh:
This started out as me trying to do a small fix to `AnnotationController`, I just wanted to move the line of code that automatically saves new highlights to the server on login out of a `$watch()`  function and to remove some duplicated state (`highlight`, `model.$highlight`, `vm.isHighlight()`, `vm.hasContent()` ...). But it ended up being a can of worms.

The changes I've ended up with are mostly just refactoring the code for clarity but this does also fix at least one broken behaviour (creating highlights while logged out, then logging in). See list of use-cases below.

Code changes:

* Rename local variable `highlight` to `newlyCreatedByHighlightButton` and add
  a docstring - clarify what this piece of state actually means.

   Yes I know this is a really long variable name but this state variable has a very particular meaning and is only used in a couple of places - I think the explicitness is justified here for clarity.

* Fix `vm.isHighlight()`: It now returns `false` for new annotations that the
  user hasn't entered any text or tags for yet (instead of `true`).

* Move the code for saving new highlights to the server into a
  `saveNewHighlight()` function that's called on AnnotationController
  instantiation instead of having it in a `$watch()` function that's called
  every time the local variable `model` changes.

  Also use the presence of `model.id` to avoid re-saving highlights that have
  already been saved - instead of abusing the `highlight`
  (now `newlyCreatedByHighlightButton`) variable and confusing what it
  represents.

* Don't automatically open the annotation editor on Annotationcontroller
  instantiation if the annotation is a highlight. This fixes this bug:

  1. Create a new highlight while logged out
  2. Log in
  3. Your highlight is saved to the server, but also the annotation editor is
     open on your highlight.

  The intention is that highlights are simply saved to the server, not opened
  for editing.

  _Annotations_ created while logged out, on the other hand, are _not_ saved
  to the server on login but _are_ opened for editing on login.

Cases to test (these are all things that are either broken on master, or were broken at one point by the earlier versions of the code changes in this pull request:

* During all cases, make sure that bucket bar numbers are correct

* Create new highlights while logged out, then login. The intended behaviour is that your highlights are silently saved to the server on login. They aren't double saved, their annotation editors aren't opened.

* Create new highlights while logged in. Intended behaviour: highlight is created and saved to server straight away when highlight button is clicked. Annotation editor is never opened.

* Load highlights from the server: reload the page and/or move between groups and/or logout and in again. Highlights should reappear as they were. Should not be re-saved again creating more highlights.

* Create new _annotations_ while logged out, then login. The intended behaviour is that your annotations are _not_ automatically saved to the server on login, and their annotation editors _are_ opened.

* Create new _annotations_ while logged in. The intended behaviour is that your annotations are _not_ automatically saved to the server, and their annotation editors _are_ opened.

* Load annotations from the server: reload the page and/or move between groups and/or logout and in again.

* Create a new annotation but do not save it, change to another group, change back to original group. Intended behaviour is that the annotation moves with you between groups and any privacy, text or tag edits are preserved.

* Edit an already saved annotation but do not save it, change to another group, change back to original group. The annotation does not go with you, but when you return to the group it is still there with its editor open and any edits preserved.

* **This one is still broken** (but is broken on master too): edit an already saved annotation and set its text to empty, change to another group, change back to original group - annotations original text has re-appeared. It should be empty.

* Create a highlight while logged in. Edit the highlight. Do not type any text or tags yet and do not save it. Move to another group. Move back. Highlight should still be there with its editor opened. Now enter some text, move to another group and back again - edits should be preserved.

P.S. In my opinion most of the complexity here is caused by using one directive / controller for many things (annotations, highlights, replies, page notes).

### DIFF
--- a/h/static/scripts/directive/annotation.js
+++ b/h/static/scripts/directive/annotation.js
@@ -218,9 +218,7 @@ function AnnotationController(
       // has no text or tags.
       var isPageNote = (model.target || []).length === 0;
       var isReply = (model.references || []).length !== 0;
-      var hasText = (model.text || '').length !== 0;
-      var hasTags = (model.tags || []).length !== 0;
-      return (!isPageNote && !isReply && !hasText && !hasTags);
+      return (!isPageNote && !isReply && !vm.hasContent());
     }
   };
 
--- a/h/static/scripts/directive/annotation.js
+++ b/h/static/scripts/directive/annotation.js
@@ -364,8 +364,8 @@ function AnnotationController(
   /**
     * @ngdoc method
     * @name annotation.AnnotaitonController#hasContent
-    * @returns {boolean} True if the currently edited annotation has
-    *          content (ie. is not just a highlight)
+    * @returns {boolean} `true` if this annotation has content, `false`
+    *   otherwise.
     */
   vm.hasContent = function() {
     var textLength = (vm.annotation.text || '').length;
--- a/h/static/scripts/directive/annotation.js
+++ b/h/static/scripts/directive/annotation.js
@@ -216,10 +216,11 @@ function AnnotationController(
       // example there's no model.highlight: true.  Instead a highlight is
       // defined as an annotation that isn't a page note or a reply and that
       // has no text or tags.
-      var targetLength = (model.target || []).length;
-      var referencesLength = (model.references || []).length;
-      var tagsLength = (model.tags || []).length;
-      return (targetLength && !referencesLength && !(model.text || tagsLength));
+      var isPageNote = (model.target || []).length === 0;
+      var isReply = (model.references || []).length !== 0;
+      var hasText = (model.text || '').length !== 0;
+      var hasTags = (model.tags || []).length !== 0;
+      return (!isPageNote && !isReply && !hasText && !hasTags);
     }
   };
 
--- a/h/static/scripts/directive/annotation.js
+++ b/h/static/scripts/directive/annotation.js
@@ -186,7 +186,82 @@ function AnnotationController(
   // Set the permissions of new annotations.
   model.permissions = model.permissions || permissions['default'](model.group);
 
-  var highlight = model.$highlight;
+  /**
+   * `true` if this AnnotationController instance was created as a result of
+   * the highlight button being clicked.
+   *
+   * `false` if the annotation button was clicked, or if this is a highlight or
+   * annotation that was fetched from the server (as opposed to created new
+   * client-side).
+   */
+  var newlyCreatedByHighlightButton = model.$highlight || false;
+
+  /**
+    * @ngdoc method
+    * @name annotation.AnnotationController#isHighlight.
+    * @returns {boolean} true if the annotation is a highlight, false otherwise
+    */
+  vm.isHighlight = function() {
+    if (newlyCreatedByHighlightButton) {
+      return true;
+    } else if (!model.id) {
+      // If an annotation has no model.id (i.e. it has not been saved to the
+      // server yet) and newlyCreatedByHighlightButton is false, then it must
+      // be an annotation not a highlight (even though it may not have any
+      // text or tags yet).
+      return false;
+    } else {
+      // Once an annotation has been saved to the server there's no longer a
+      // simple property that says whether it's a highlight or not.  For
+      // example there's no model.highlight: true.  Instead a highlight is
+      // defined as an annotation that isn't a page note or a reply and that
+      // has no text or tags.
+      var targetLength = (model.target || []).length;
+      var referencesLength = (model.references || []).length;
+      var tagsLength = (model.tags || []).length;
+      return (targetLength && !referencesLength && !(model.text || tagsLength));
+    }
+  };
+
+  /** Save this annotation if it's a new highlight.
+   *
+   * The highlight will be saved to the server if the user is logged in,
+   * saved to drafts if they aren't.
+   *
+   * If the annotation is not new (it has already been saved to the server) or
+   * is not a highlight then nothing will happen.
+   *
+   */
+  function saveNewHighlight() {
+    if (model.id) {
+      // Already saved.
+      return;
+    }
+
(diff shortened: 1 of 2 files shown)


seanh:
Needs tests

seanh:
Ready for review

judell:
"But it ended up being a can of worms." No good deed goes unpunished.

h/static/scripts/directive/annotation.js
```diff
@@ -186,7 +186,81 @@ function AnnotationController(
...
+      // Not a highlight,
+      return;
+    }
+
+    if (model.user) {
+      // User is logged in, save to server.
+      // Highlights are always private.
+      model.permissions = permissions.private();
+      model.$create().then(function() {
+        $rootScope.$emit('annotationCreated', model);
```
nickstenning: So I don't think this should block merging this PR, but this smells really awful to me. `AnnotationController` shouldn't be responsible for firing these lifecycle events. Indeed, most of the rest of them have moved into `annotationMapper`, which makes me wonder why `AnnotationController` isn't using `annotationMapper.createAnnotation(object)` rather than `ngResource.$create`...

h/static/scripts/directive/test/annotation-test.js
```diff
@@ -470,6 +470,177 @@ describe('annotation.js', function() {
...
+        annotation.$create = sandbox.stub().returns({
+          then: function() {}
+        });
+
+        createDirective();
+
+        assert.notCalled(annotation.$create);
+      });
+
+      it('does not save old highlights on initialization', function() {
```
robertknight: I'd suggest we call these "existing" rather than "old" for consistency with comments elsewhere

h/static/scripts/directive/test/annotation-test.js
```diff
@@ -470,6 +470,177 @@ describe('annotation.js', function() {
...
+
+      it('edits annotations with drafts on initialization', function() {
+        // This is not a new annotation.
+        annotation.id = 'annotation_id';
+        // This is not a highlight.
+        annotation.$highlight = undefined;
+        annotation.text = 'Annotation text';
+        annotation.tags = ['tag_1', 'tag_2'];
+        // The drafts service has some draft changes for this annotation.
+        fakeDrafts.get.returns('foo');
```
robertknight: This should return an object that conforms to the interface for `drafts.get` (`{text?: string, tags?: string[], ...}`). It happens to work with the current implementation of AnnotationController, but could incorrectly break in future.

h/static/scripts/directive/test/annotation-test.js
```diff
@@ -470,6 +470,177 @@ describe('annotation.js', function() {
...
+
+        var controller = createDirective().controller;
+
+        assert.isFalse(controller.editing());
+      });
+
+      it('edits highlights with drafts on initialization', function() {
+        // You can edit a highlight, enter some text or tags, and save it (the
+        // highlight then becomes an annotation). You can also edit a highlight
+        // and then change focus to another group and back withou saving the
```
robertknight: s/withou/without

h/static/scripts/directive/test/annotation-test.js
```diff
@@ -470,6 +470,177 @@ describe('annotation.js', function() {
...
+        // highlight, in which case the highlight will have draft edits.
+        //
+        // This is not a new highlight.
+        annotation.id = 'annotation_id';
```
robertknight: This is duplicated in a bunch of tests, it could be refactored into a helper.

h/static/scripts/directive/annotation.js
robertknight: Indeed, we definitely don't want to have multiple places handling the actual creation of the object on the server if possible.

robertknight:
So overall:

1. This is looking much clearer, the comments clarify things especially
2. I'd agree with Nick's comments about `annotationMapper` - it would be preferable to have the submission of annotations to the server only happen in one place
3. There is a lot of duplication in the tests (eg. in terms of repeated comments explaining the same thing about the consequences of setting different attributes on a new annotation) which is going to make refactoring harder in future. Extracting these into a helper could simplify that.

eg:
```js
it('...', function () {
  var annotation = createAnnotation({
    // properties customizing the new annotation, everything
    // else is set to a default
  });
});
```

h/static/scripts/directive/test/annotation-test.js
seanh: Hmm. If I do detailed mock of the `drafts.get()` interface in the `AnnotationController` tests then if I changed the drafts interface I'd have to go through the `AnnotationController` tests and update those mocks, otherwise it'd be a bit weird. No chance I would remember to do this.

All that AnnotationController cares about here is the truthiness so I think that's all that we should mock, and `'foo'` makes it clear that it just has to return _something_, we don't care what.

If we change AnnotationController's init to care about more of the drafts interface then this test might break, yes, and we'd have to improve the mocking, but AnnotationController tests are exactly what I'd expect to break if I changed AnnotationController's behaviour so I think that's okay.

h/static/scripts/directive/test/annotation-test.js
seanh: Done

seanh:
I've factored out the `annotation` global from the tests to pave the way for removing duplication by adding helper functions for creating different kinds of annotation. Going to lunch now, will finish after..

h/static/scripts/directive/annotation.js
seanh: Agreed, let's remember to do this one later

h/static/scripts/directive/test/annotation-test.js
seanh: Done

seanh:
@robertknight As long as you're happy that I ignored the old -> existing comment (I can't really be bothered to go through both files and change old -> existing everywhere unless you really want it), I think that's everything responded to

PR #2763, opened 2015-12-03
Refactor annotation controller tests

seanh:
It'll be easier to look at the individual commits than the whole diff (because everything got indented).

In trying to write tests for <https://github.com/hypothesis/h/pull/2762/> I discovered some problems with the `AnnotationController` tests. In particular:

* `controller` and some other globals are shared between all tests and not reset before some tests. This was actually causing some tests to pass or fail depending on order or selection of tests.

* Because the directive and controller are initialized in `beforeEach()` functions (in many of the tests, anyway) it's not possible to test what happens on initialization under different circumstances (without writing some tests that are completely different than all of the others)

* Some tests were inside the `describe()` for `AnnotationController`s tests which implies a _lot_ of shared variables and setup, when they didn't need any of that stuff

I've resisted the urge to get rid of all the global variables in these tests (although I'd like to), and just got rid of the ones that are set by `createDirective()` and which were actually causing problems (plus a couple more that were only used in one or two places and really had no excuse being a global).

### DIFF
--- a/h/static/scripts/directive/test/annotation-test.js
+++ b/h/static/scripts/directive/test/annotation-test.js
@@ -6,6 +6,15 @@ var events = require('../../events');
 var module = angular.mock.module;
 var inject = angular.mock.inject;
 
+/** Return Angular's $compile service. */
+function compileService() {
+  var $compile;
+  inject(function(_$compile_) {
+    $compile = _$compile_;
+  });
+  return $compile;
+}
+
 describe('annotation.js', function() {
 
   describe('extractDocumentMetadata()', function() {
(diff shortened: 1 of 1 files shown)


robertknight:
I reviewed this locally with `git diff -w`. From a quick search, it appears that GitHub's diff viewer doesn't have such an option built-in, or did I miss it.

A good improvement, although I do think some of the tests (eg. the explicit tests for `updateDomainModel`) are verging on being too white-box rather than testing from the POV of the class user.

seanh:
The point of unit-testing something like `updateDomainModel()` is that the tests provide some documentation for the function / we actually know what the function does, and we can refactor the code inside the function and know we haven't broken it. In other words - the tests are for us, not the user. It's not about end-to-end behaviour.

PR #2764, opened 2015-12-04
Add groups report to admin dashboard

finitud:
This ticks the first two items in https://trello.com/c/iO3YJnyW/191-staff-dashboard-groups-report

### DIFF
--- a/h/admin.py
+++ b/h/admin.py
@@ -207,6 +207,12 @@ def badge_remove(request):
     request.db.delete(models.Blocklist.get_by_uri(uri))
     return badge_index(request)
 
+@view.view_config(route_name='admin_groups',
+                  request_method='GET',
+                  renderer='h:templates/admin/groups.html.jinja2',
+                  permission='admin_groups')
+def groups_index(request):
+    return {"groups": models.Group.all(), "request": request}
 
 def includeme(config):
     config.add_route('admin_index', '/admin')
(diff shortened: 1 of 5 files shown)


h/admin.py
```diff
@@ -207,6 +207,12 @@ def badge_remove(request):
     request.db.delete(models.Blocklist.get_by_uri(uri))
     return badge_index(request)
 
+@view.view_config(route_name='admin_groups',
+                  request_method='GET',
+                  renderer='h:templates/admin/groups.html.jinja2',
+                  permission='admin_groups')
+def groups_index(request):
+    return {"groups": models.Group.all(), "request": request}
```
nickstenning: You shouldn't need to pass `request` in explicitly, as it's one of the [system values provided to renderers by Pyramid](http://docs.pylonsproject.org/projects/pyramid/en/latest/narr/templates.html#renderer-system-values).

nickstenning:
This looks great. Only one minor point about template context, and perhaps you could just add a sentence to the top of the page to explain its purpose (see the other admin pages for examples)?

nickstenning:
Oh, and one more thought: perhaps instead of separate columns for the username and user email, it might make sense to have one column which links the creator's username to `/admin/users?username={{ group.creator.username }}`?

h/admin.py
finitud: Oh, thanks, will fix! :+1:

nickstenning:
LGTM! :cake:

PR #2766, opened 2015-12-04
Remove USER_CHANGED event from AnnotationController

seanh:
When the user logs out or logs in we throw away all the
AnnotationController instances and create new ones. So everything it
needs to do "when the user logs in" (or out) AnnotationController can do
on Initialization, it doesn't need the event.

### DIFF
--- a/h/static/scripts/directive/annotation.js
+++ b/h/static/scripts/directive/annotation.js
@@ -174,14 +174,12 @@ function AnnotationController(
   /** The domain model, contains the currently saved version of the annotation
    * from the server. */
   var model = $scope.annotationGet();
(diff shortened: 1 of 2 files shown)


seanh:
This is presumably the reason why all the `AnnotationController`s are recreated on `USER_CHANGED`: https://github.com/hypothesis/h/blob/master/h/static/scripts/app-controller.coffee#L50

seanh:
Test for it https://github.com/hypothesis/h/blob/master/h/static/scripts/test/app-controller-test.coffee#L165

nickstenning:
LGTM. Thanks for the additional context.

PR #2770, opened 2015-12-07
New homepage design

robertknight:
This implements the new homepage design. Since the design is still in flux, it is served at `/new-homepage` and the existing design is served at "/"

The only user-dependent content is the "Login" link which is replaced with the user's name if they are logged in.

**Implementation Notes**
- The new front-page now shares the same `site.css` bundle as the other static pages (eg. groups). This will make it easier to roll out a consistent design across all of those pages.
- This implementation still uses the Bootstrap modal and alerts. Rather than including the full bootstrap distribution on the front page, I've replaced it with a minimal build that component, plus a config file which can be used with the custom build tool on Bootstrap's website to rebuild in future. The Bootstrap config, styling and scripts are in `h/static/styles/vendor/bootstrap`
- The old logo-only header used on several pages has been renamed to `logo-header`
- I've CommonJS-ified the controller code for the installer buttons and moved it into the `site.js` bundle to make it consistent with the rest of the site.


nickstenning:
I know this is still WIP, but I do have one comment at this early stage: I'm not super keen for us to roll our own RSS feed caching mechanism just to render the homepage. We probably want to set up a generally-usable cache based on [dogpile](https://dogpilecache.readthedocs.org/en/latest/) or [beaker](https://beaker.readthedocs.org/en/latest/modules/cache.html).

As such, perhaps it would be better to drop the feed section of the page entirely at this point (as previously discussed in sprint planning) and add that as a separate piece of work. That way we can discuss the approach we're going to take to caching the feed, how we want this to work in development, etc., without blocking the main work of the redesign.

robertknight:
> As such, perhaps it would be better to drop the feed section of the page entirely at this
> point (as previously discussed in sprint planning) and add that as a separate piece of work. 

Another alternative, given the relatively low frequency of new blog updates, would simply be to cheat in the initial PR and hardcode the contents of the front page stories. That way I can deliver an initial PR which matches the design. We can then quickly follow up with a more generic caching mechanism.

judell:
Code of conduct has the link https://hypothes.is/code-of-conduct, should it be https://hypothes.is/community-guidelines/?

robertknight:
This is good to go from my side.

h/assets.yaml
```diff
@@ -93,7 +97,7 @@ admin_css:
 admin_js:
   contents:
     - jquery
-    - bootstrap
+    - bootstrap_js
```
nickstenning: I'm guessing this should be `bootstrap_full_js`?

h/static/scripts/installer-controller.js
```diff
@@ -0,0 +1,34 @@
...
+  // check for Chrome extension support. We also check here for !mobile
+  // so that the page is correct when testing in Chrome dev tools with
+  // mobile device simulation enabled
+  var offerChromeInstall = uaDetect.chromeExtensionsSupported &&
+                           !uaDetect.isMobile;
+  var offerBookmarkletInstall = !offerChromeInstall &&
+                                !uaDetect.isMobile &&
+                                !uaDetect.isMicrosoftEdge;
+
+  showIf('.js-install-chrome', offerChromeInstall);
```
nickstenning: It make make more sense to take a Modernizr-like approach here and add "feature" classes to `document.body` which can then be used to show/hide parts of the page from CSS, as opposed to having to explicitly run `showIf` for each element.

If you need to hide things until you can detect features, but want them shown by default if JS fails to load, you can use the old chestnut of putting `document.documentElement.classList.add('js')` before any other scripts in the `<head>` to avoid a FOUC.

h/assets.yaml
robertknight: Yes, indeed.

nickstenning:
Merging behind a feature flag.

PR #2772, opened 2015-12-07
Generate a downloadable groups report in CSV format

finitud:
This finishes story https://trello.com/c/iO3YJnyW/191-staff-dashboard-groups-report

The CSV renderer has been taken from the Pyramid documentation, with only very minor modifications: https://pyramid-cookbook.readthedocs.org/en/latest/templates/customrenderers.html

### DIFF
--- a/h/admin.py
+++ b/h/admin.py
@@ -207,12 +207,37 @@ def badge_remove(request):
     request.db.delete(models.Blocklist.get_by_uri(uri))
     return badge_index(request)
 
+
 @view.view_config(route_name='admin_groups',
                   request_method='GET',
                   renderer='h:templates/admin/groups.html.jinja2',
                   permission='admin_groups')
 def groups_index(request):
     return {"groups": models.Group.all()}
+
+
+@view.view_config(route_name='admin_groups_csv',
+                  request_method='GET',
+                  renderer='csv',
+                  permission='admin_groups')
+def groups_index_csv(request):
+    groups = models.Group.all()
+
+    header = ['Group name', 'Group URL', 'Creator username',
+              'Creator email', 'Number of members']
+    rows = [[group.name,
+             request.route_url('group_read',
+                               pubid=group.pubid,
+                               slug=group.slug),
+             group.creator.username,
+             group.creator.email,
+             len(group.members)] for group in groups]
+
+    filename = 'groups.csv'
+    request.response.content_disposition = 'attachment;filename=' + filename
+
+    return {'header': header, 'rows': rows}
+
 
 def includeme(config):
     config.add_route('admin_index', '/admin')
@@ -222,5 +247,6 @@ def includeme(config):
     config.add_route('admin_staff', '/admin/staff')
     config.add_route('admin_users', '/admin/users')
     config.add_route('admin_groups', '/admin/groups')
+    config.add_route('admin_groups_csv', '/admin/groups.csv')
     config.add_route('admin_badge', '/admin/badge')
     config.scan(__name__)
--- a/h/renderers.py
+++ b/h/renderers.py
@@ -1,0 +1,34 @@
+# -*- coding: utf-8 -*-
+# Taken from:
+# https://pyramid-cookbook.readthedocs.org/en/latest/templates/customrenderers.html
+# with minor modifications
+import csv
+from h._compat import StringIO
+
+
+class CSV(object):
+    def __init__(self, info):
+        pass
+
+    def __call__(self, value, system):
+        """ Returns a plain CSV-encoded string with content-type
+        ``text/csv``. The content-type may be overridden by
+        setting ``request.response.content_type``."""
+
+        request = system.get('request')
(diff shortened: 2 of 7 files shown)


h/admin.py
```diff
@@ -214,6 +214,27 @@ def badge_remove(request):
 def groups_index(request):
     return {"groups": models.Group.all()}
 
+@view.view_config(route_name='admin_groups',
```
nickstenning: I reckon we should make the URL for this `/admin/groups.csv`. That would mean giving it its own route name (`admin_groups_csv`) and setting the URL in `includeme`.

h/renderers.py
```diff
@@ -0,0 +1,31 @@
+# Taken from https://pyramid-cookbook.readthedocs.org/en/latest/templates/customrenderers.html
+# with minor modifications
+import csv
+try:
+    import StringIO as io # python 2
```
nickstenning: Two minor comments here. 

1. We try to centralise all the Py2/Py3 switching we do in one place, namely `h/_compat.py`, so this should move there.
2. This actually imports different tokens in Python 2 and Python 3 (`io` in Python 2 and `StringIO` in Python 3.)

So I'd suggest something like this in `h/_compat.py`:

```python
try:
    from io import StringIO
except ImportError:
    from StringIO import StringIO
```

which you can then use as:

```python
from h._compat import StringIO
```

h/renderers.py
```diff
@@ -0,0 +1,31 @@
...
+except ImportError:
+    from io import StringIO # python 3
+
+class CSVRenderer(object):
+   def __init__(self, info):
```
nickstenning: Looks good. I think I'd be inclined to call this simply `CSV` given that it's in a module called `renderers`, as that leads to rather more fluent code:

```python
from h.renderers import CSV
```

or 

```python
from h import renderers

...

renderers.CSV
```

h/renderers.py
```diff
@@ -0,0 +1,31 @@
...
+         if ct == response.default_content_type:
+            response.content_type = 'text/csv'
+
+      fout = io.StringIO()
+      writer = csv.writer(fout, delimiter=',', quotechar=',', quoting=csv.QUOTE_MINIMAL)
+
+      writer.writerow(value.get('header', []))
+      writer.writerows(value.get('rows', []))
+
+      return fout.getvalue()
```
nickstenning: I think we should probably add some tests for this class. I'd expect it to test at least:

- the default value of `content_type`
- the string output of the renderer for a simple example

h/templates/admin/groups.html.jinja2
```diff
@@ -8,6 +8,12 @@
     On this page you can see a list of all the groups and their details.
   </p>
 
+  <p>
+    <a href="/admin/groups?fmt=csv">
```
nickstenning: This URL should probably be generated using `request.route_url`. It would probably look something like `request.route_url('admin_groups_csv')``` if you make the changes suggested earlier in this PR.

nickstenning:
Looking great! A few comments above.

h/renderers.py
```diff
@@ -0,0 +1,31 @@
+# Taken from https://pyramid-cookbook.readthedocs.org/en/latest/templates/customrenderers.html
```
nickstenning: Could you just add a `# -*- coding: utf-8 -*-` comment to the top of this new file?

finitud: :+1:

finitud:
Many thanks for your comments, @nickstenning! I have added a few more commits that add your suggestions.

h/test/renderers_test.py
```diff
@@ -0,0 +1,21 @@
...
+import pytest
+
+from pyramid.testing import DummyRequest
+
+from h import renderers
+
+def test_response_content_type():
+    renderer = renderers.CSV({})
+    req = DummyRequest()
+    renderer({}, { 'request': req })
```
nickstenning: Only one *tiny* niggle here: good Python style doesn't have these extra spaces. You might want to install [`flake8`](https://pypi.python.org/pypi/flake8) to help catch things like this. If you run `flake8 h/test/renderers_test.py` that will tell you what needs fixing.

h/test/renderers_test.py
```diff
@@ -0,0 +1,21 @@
+# -*- coding: utf-8 -*-
+import pytest
+
+from pyramid.testing import DummyRequest
+
+from h import renderers
+
+def test_response_content_type():
```
nickstenning: Another tiny stylistic detail: two blank lines before function definitions.

h/test/renderers_test.py
```diff
@@ -0,0 +1,21 @@
+# -*- coding: utf-8 -*-
+import pytest
```
nickstenning: This import looks unused to me.

finitud:
This has been nicely rebased and is ready to merge.

nickstenning:
LGTM. :cake:

PR #2773, opened 2015-12-07
Don't crash if parent has no text

### LINKED ISSUE #2771: Notifications worker crashes on reply to annotation with no text
I'm not sure how it was created but this is a [public annotation with no text field](https://hypothes.is/a/Iqvi3yueRBGRwidlaI8buw) which, when replied to, will cause the notifications worker to crash:

```
KeyError: 'text'
  File "h/notification/reply_template.py", line 146, in generate_notifications
    parent)
  File "h/notification/reply_template.py", line 159, in render_reply_notification
    tmap = create_template_map(request, annotation, parent)
  File "h/notification/reply_template.py", line 55, in create_template_map
    'parent_text': parent['text'],
```

seanh:
Fix a crash in the notification worker when the parent annotation that has been
replied to has no `text` item.

Fixes #2771.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,6 +1,12 @@
 Unreleased
 ==========
 
+Bug fixes
+---------
+
+- Fix a crash that meant that reply notification emails would not be sent if
+  the annotation replied to had no text (#2771).
+
 Features
 --------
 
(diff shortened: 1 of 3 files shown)


seanh:
I wanted to fix this properly by adding a `.text` property to `h.api.models.Annotation` so that the notifications worker and other code doesn't need to worry about not crashing on annotations with no text. But I didn't want to spend all day on it, and `reply_template.py` and its test are pretty tightly coupled to the idea of annotations being dicts.

nickstenning:
Cheers!

PR #2774, opened 2015-12-07
2728 misc annotation controller refactors part 2

seanh:


seanh:
This contains a lot of moving code around, so the diff will be easier if you look at the individual commits

nickstenning:
LGTM.

PR #2779, opened 2015-12-08
Remove "at least two characters" copy from password field

finitud:
On account registration, claim, and password reset.
There are other related messages (such as the error popup warning the user for password too short) which still warn about this limit, but aren't visible except as error messages.


nickstenning:
Thank you!

PR #2780, opened 2015-12-08
Move API exception handling into exception views

nickstenning:
With a view to making the API view functions easier to read and modify, this commit moves API exception handling for a number of different error types into standalone exception views. This eliminates the need for repetitive boilerplate in view functions, which can simply raise an exception of the appropriate type.

### DIFF
--- a/h/api/test/views_test.py
+++ b/h/api/test/views_test.py
@@ -5,7 +5,6 @@ import mock
 import pytest
 
 from pyramid import testing
-from pyramid import httpexceptions
 
 from h.api import views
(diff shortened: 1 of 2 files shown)


seanh:
As discussed in Slack there's a 500 if you post JSON without a top-level object to the create API, but that bug also exists on master.

I think the 404 response if you post to the create API without a JSON Content-Type could be more helpful if it were an error saying Content-Type whatever isn't supported. I'm not sure what REST/HTTP has to say about this though

PR #2781, opened 2015-12-08
Annotation controller domain model view model refactor

seanh:
This is, finally, the big AnnotationController refactoring that we talked about
in Berlin, after [several](https://github.com/hypothesis/h/pull/2756) [pull](https://github.com/hypothesis/h/pull/2761) [requests](https://github.com/hypothesis/h/pull/2762) [to](https://github.com/hypothesis/h/pull/2763) [prepare](https://github.com/hypothesis/h/pull/2766) [the](https://github.com/hypothesis/h/pull/2774) ground for it. There's
still much to be done to improve this code but I think it's a good idea to
stop with this pull request for now, in part so I can work on something else
and in part so that we can let all these recent changes settle and see what's
been broken.

For what it's worth though, some things it would be good to do in the future:

* Go over `annotation-test.js`. Replace all the
  `describe('when the annotation is a highlight')`-type tests with unit tests
  for particular functions and methods. It's fine to have some end-to-end style
  tests in here but we need unit tests as well, a lot of the tests in
  `annotation-test.js` are pretty fuzzy about what code they're actually
  testing. In general, fill out the number of unit tests and fix/get rid of
  dodgy tests. There's a lot of room for improvement here.

* Separate saving and editing (currently one long method with two cases).

* Make saving and editing use ngResource correctly. It should not be necessary to modify `domainModel`, `updateDomainModel()` shouldn't be needed at all, just send the changes (only) to the server and `domainModel` will automatically update itself on a successful response from the server, then in an `onFulfilled` method call `updateViewModel()`.

* Separate code paths for creating new highlights, annotations, replies and
  page notes and possibly separate out creating new ones vs instantiating
  AnnotationController instances for existing ones, too. This should enable
  some complexity to be removed from AnnotationController, such as the use of
  `$highlight`.

* Consider having separate directives for highlights, replies, page notes and
  annotations, instead of one complex directive that tries to do it all.

* Move functions at the top of annotation.js into their own module. Right now
  these have to be added to the public API of the annotation module in order to
  unit test them.

* A lot of complexity and logic could be removed from `annotation.js` and
  friends by having the server do it for us instead and just send us a simpler
  and more useful annotation object. Things like deciding whether a given
  annotation is a highlight, page note, reply or annotation, like extracting
  metadata from the annotation object, like dealing with ACL objects (these
  could be kept entirely server-side), validating data, etc etc.

* Replace the "deep counting" of replies with something _much_ simpler
  (and possibly computed server-side).

* The annotation create and edit APIs respond with a JSON object of the new or
  updated annotation as it is on the server, but afaik `AnnotationController`
  just ignores this response and continues working with its `domainModel`
  object. Maybe it should pay attention to the server.

Summary of the changes that _are_ in this pull request:

`model` (now `domainModel`) is almost entirely read-only. The only thing that modifies it is `updateDomainModel()` which is only called in `vm.save()` to update the domain model with the unsaved changes from the view model before sending the domain model to the server to be saved. In Berlin we mentioned that when saving to the server the `save()` method would take `domainModel` and `vm`, combine them into a local object of everything to be sent to the server, send it and then replace `domainModel` wholesale with the response from the server (since the annotation create and edit APIs respond with a JSON object of the new/updated annotation). I haven't done this change, but it would now be a change to the `vm.save()` method only.

I've focused on eliminating duplication between `domainModel` and `vm`, and on simplifying
the control and data flows, and making `domainModel` read-only everywhere other than in `updateDomainModel()`/`vm.save()`.

The changes:

Some small changes to just remove simple code duplication, make code more readable,
etc:

* Added `isNew()` helper function instead of testing for the existence of
  `model.id` (which is duplication, but it's also not clear without a code
  comment what the semantics of such a check are intended to be).

* Similarly `isReply()` helper function added.

* `model` is now called `domainModel` (and this name is used consistently
  everywhere, the previous name was not).

* `vm` is still `vm`, but `vm.annotation` is now called `vm.form` and there are
  comments to explain that `vm.form` contains the read-write/`ng-model`
  portions of `vm` and the rest of `vm` is intended to be read-only to the
  templates. Again these names are now used consistently everywhere.

Removing data duplication, particularly duplication of the annotation domain model
values in multiple places which was causing all sorts of confusion and
problems:

* `vm.annotation.group` is gone, the templates just use the `vm.group()` method
  (which returns `groups.focused().id`). The annotation's current group is read-only for the templates, and is always the currently focused group.

* `vm.annotation.permissions` is gone, replaced with a simple boolean `vm.isPrivate`.

* Several other fields that were on `vm.annotation` (`id`, `target`, `updated`,
  `user`) are also replaced with helper methods for accessing the values from
  `domainModel`.

  Using accessor methods like this avoids duplicating the data in two places,
  makes it clear that the data is intended to be read-only for the templates,
  and (particularly in the case of `permissions`) allows the templates to see a
  simplified API to the model.

* `vm.document` has been removed, replaced with `vm.document()` method,
  again replaces managing a variable and duplicated data with just calling a
  read-only accessor method.

* We used to have `domainModel.tags`, `vm.tags` and `vm.annotation.tags`. This
  is now just `domainModel.tags` and `vm.form.tags`.

Making the data flow explicit:

* We're explicit about exactly what properties are copied from `vm` to
  `domainModel` (now encapsulated in the `updateDomainModel()` function)
  and vice-versa (`updateViewModel()`) and when, by having helper functions to
  do this and having them copy the individual fields explicitly instead of
  calling `angular.extend()` on entire objects.

* We're explicit about exactly what properties are copied from `vm` and
  `domainModel` into the `drafts` service (`saveToDrafts()`) and vice-versa
  (`restoreFromDrafts()`) and when, by having helper functions and having them
  copy the individual fields explicitly instead of copying "black box"
  `changes` objects into drafts and using `angular.extend()` on them.

* How tags are translated between `domainModel.tags` format (an array of
  strings) and `vm.form.tags` format (an array of objects with `.text`
  properties) and when is now made explicit with two helper functions
  `domainModelTagsFromViewModelTags()` and
  `viewModelTagsFromDomainModelTags()`.

* The two distinct and separate processes of "updating the view model from
  the domain model" (now `updateDomainModel()`) and "restoring this instance of
  AnnotationController from drafts" (now `restoreFromDrafts()`) have been
  separated (previously a single method did both of these).

Making the control flow more comprehensible:

* `onDomainModelChange()` is gone, this nasty little method would run whenever
  `domainModel` changed and would copy values from `domainModel` onto
  `vm.annotation` (now `vm.form`) sometimes overwriting values on `vm` with
  older values, destroying user data and creating a confusing control / data
  flow. It was also relied on to run when an `AnnotationController` instance
  was initialized, since `domainModel` would be changed by the initialization
  code, then `onDomainModelChange()` would run and do some more initialization
  work.

There are now two ways that user changes to the annotation flow through
`AnnotationController`:

1. Text and tags are duplicated as `domainModel.text` and `domainModel.tags`
  and `vm.form.text` and `vm.form.tags` (the two copies of tags are in
  different formats, as well).

  On initialization of a new `AnnotationController` instance
  `domainModel.text` and `domainModel.tags` are used to initialize `vm.form`.
  This also happens when we `revert()` user changes (when the user clicks
  _Cancel_ while editing). These are the only two times when we copy text and
  tags from `domainModel` to `vm` (which you can now verify by searching for
  calls to `updateViewModel()`).

  `vm.form.text` and `vm.form.tags` are written by the templates as the user
  types, via `ng-model`. These changes are then copied from `vm.form` to
  `domainModel` when the user clicks _Post_, before we send the data to the
  server (again, you can now verify this by searching for calls to
  `updateDomainModel()`).

2. Permissions and group are duplicated as `domainModel.permissions` and `domainModel.group` and `vm.group()` (which just returns `groups.focused().id` and `vm.isPrivate`.

   There's `ng-model` or Angular two-way binding involved for permissions and group.

   Events are triggered by the templates that cause `AnnotationController` to
   make changes:

   1. When the user changes the selection in the _Post_ dropdown
      `vm.setPrivacy()` runs and this changes the value of `vm.isPrivate`.

   2. When the user changes the focused group (this is not even the
      `annotation.html` template that does this) the templates will see the change to `groups.focused()`
      reflected by the `vm.group()` method that they use.


h/static/scripts/directive/annotation.js
```diff
@@ -72,50 +89,98 @@ function extractDocumentMetadata(model) {
...
-*
-*/
-function updateDomainModel(domainModel, viewModel) {
-  angular.extend(domainModel, viewModel);
-  domainModel.tags = viewModel.tags.map(function(tag) {
-    return tag.text;
-  });
+/** Return `true` if the given annotation is a reply, `false` otherwise. */
+function isReply(domainModel) {
+  if ((domainModel.references || []).length > 0) {
```
robertknight: The `if` statement is redundant here, you can just write this as `return <boolean expression>`, ie. `return domainModel.references && domainModel.references.length > 0`

h/static/scripts/directive/annotation.js
```diff
@@ -72,50 +89,98 @@ function extractDocumentMetadata(model) {
...
-  // Extend the view model with a copy of the domain model.
-  // Note that copy is used so that deep properties aren't shared.
-  vm.annotation = angular.extend({}, angular.copy(model));
+/** Return `true` if the given annotation is new, `false` otherwise.
+ *
+ * "New" means this annotation has been newly created client-side and not
+ * saved to the server yet.
+ */
+function isNew(domainModel) {
+  if (domainModel.id) {
```
robertknight: This can be simplified to `return !!domainModel.id`

h/static/scripts/directive/annotation.js
```diff
@@ -72,50 +89,98 @@ function extractDocumentMetadata(model) {
...
+}
 
-  vm.annotationURI = new URL(
-    '/a/' + vm.annotation.id, vm.baseURI).href;
+/**
+  * Save the given annotation to the drafts service.
+  *
+  * Any existing drafts for this annotation will be overwritten.
+  *
+  * @param {object} domainModel - The full domainModel object of the
```
robertknight: The docs don't match the actual arguments here.

h/static/scripts/directive/annotation.js
```diff
@@ -327,38 +394,17 @@ function AnnotationController(
...
 
-    if (model.user) {
+    if (domainModel.user) {
       // User is logged in, save to server.
       // Highlights are always private.
-      model.permissions = permissions.private();
-      model.$create().then(function() {
-        $rootScope.$emit('annotationCreated', model);
+      domainModel.permissions = permissions.private();
+      domainModel.$create().then(function() {
```
robertknight: Missing a `catch` here. What happens if `$create()` fails, does anything get logged anywhere?

h/static/scripts/directive/annotation.js
```diff
@@ -660,13 +713,13 @@ function AnnotationController(
     // creating or editing, we cache that and use the same privacy level the
     // next time they create an annotation.
     // But _don't_ cache it when they change the privacy level of a reply.
-    if (!model.references) {  // If the annotation is not a reply.
+    if (!isReply(domainModel)) {
```
robertknight: :+1: - Much clearer

h/static/scripts/directive/test/annotation-test.js
```diff
@@ -1502,27 +1472,36 @@ describe('annotation.js', function() {
 
       it('moves new annotations to the focused group', function() {
         var annotation = defaultAnnotation();
+        // id must be null so that AnnotationController considers this a new
+        // annotation.
         annotation.id = null;
```
robertknight: We're mixing up use of null and undefined in difference places, I need to do a little more background reading but I think it would make sense to recommend exclusively using one or the other in the style guide in future.

h/static/scripts/drafts.js
```diff
@@ -31,47 +39,53 @@ function DraftStore() {
...
+  };
 
   /** Retrieve the draft changes for an annotation. */
   this.get = function get(model) {
-    for (var i=0; i < this._drafts.length; i++) {
+    for (var i = 0; i < this._drafts.length; i++) {
       if (match(this._drafts[i], model)) {
-        return this._drafts[i].changes;
+        return {
+          private: this._drafts[i].private,
```
robertknight: Perhaps rename to `isPrivate` to make it clear that this is a boolean?

h/static/scripts/drafts.js
```diff
@@ -31,47 +39,53 @@ function DraftStore() {
...
+          tags: this._drafts[i].tags,
+          text: this._drafts[i].text
```
robertknight: Use trailing commas for the last element in multi-line object literals.

h/static/scripts/drafts.js
```diff
@@ -31,47 +39,53 @@ function DraftStore() {
...
+  };
 
   this.discard = function discard() {
     this._drafts = [];
-  }
+  };
 }
 
-module.exports = function () {
+module.exports = function() {
```
robertknight: Should have a space after the `function` keyword for consistency with the rest of our code.

h/static/scripts/test/drafts-test.js
```diff
@@ -11,23 +11,25 @@ describe('drafts', function () {
     it('should save changes', function () {
       var model = {id: 'foo'};
       assert.notOk(drafts.get(model));
-      drafts.update(model, {text: 'edit'});
-      assert.deepEqual(drafts.get(model), {text: 'edit'});
```
robertknight: I actually think it was easier to work with the previous versions of the API methods where the parameters were passed via an object. `drafts.update(model, {isPrivate: true, ...})` seems clearer than `drafts.update(model, true, ...)` to me.

h/templates/client/annotation.html
```diff
@@ -71,7 +71,7 @@
   <!-- Body -->
   <section name="text" class="annotation-body">
     <excerpt enabled="vm.feature('truncate_annotations') && !vm.editing()">
-      <markdown ng-model="vm.annotation.text"
+      <markdown ng-model="vm.form.text"
```
robertknight: Definitely helpful to have this distinction between the properties of an annotation that are editable in the card and those that are not.

nickstenning:
Looking good but as usual I a suggestion about naming. The AnnotationController instance is indeed a view model, in that it is a model of the view state. It makes perfect sense to move RW properties to something like `form` as you've done, and expose other relevant viewmodel state as properties on the instance.

But renaming `annotation` to `domainModel` everywhere makes less sense to me. For starters it's not descriptive ("the domain model for what?"), but perhaps more importantly it's just an odd name, as in common usage "domain model" refers to the whole set of abstractions relating annotations to documents, users, etc. The annotation object isn't *the* domain model, it's a "domain model object", as in, an object from the domain model as opposed to the view model.

Anyway, it's not hugely important, but it's worth thinking about what a new developer might make of this code. I think "viewModel" speaks for itself in this context, but "domainModel" raises more questions than it answers. Perhaps just stick with calling it "annotation" and if you're concerned about people making changes to it directly then stick a big comment up at the top explaining the data flow in this component.

h/static/scripts/directive/annotation.js
seanh: This is me getting nervous about apparent boolean expressions in JavaScript actually returning non-boolean (but truthy or falsy) values, and wanting to be explicit about returning `true` of `false`. It does appear to be unnecessary in this case though.

h/static/scripts/directive/annotation.js
seanh: Hmm, this one I'm not so sure about it, that would work but I don't think `!!foo` is very clear

h/static/scripts/directive/annotation.js
seanh: This _is_ missing a catch but is the same as on master, so I'd like to leave it to a separate pr

seanh: (And I think if it fails it will fail silently, apart from something in the browser console)

h/static/scripts/directive/test/annotation-test.js
seanh: Sounds good

h/static/scripts/drafts.js
seanh: Really? I know this isn't JSON, but trailing commas are invalid in JSON, and I figured that for consistency with JSON we should do the same in JavaScript object literals

h/static/scripts/directive/annotation.js
seanh: Done

h/static/scripts/directive/annotation.js
seanh: Done, it's actually just `!domainModel.id`

h/static/scripts/directive/annotation.js
seanh: Done

h/static/scripts/drafts.js
seanh: Done

h/static/scripts/drafts.js
seanh: Done

h/static/scripts/drafts.js
seanh: As discussed, leaving this to a future pull request that updates our jscs config

h/static/scripts/test/drafts-test.js
seanh: Done

seanh:
I've responded to most of the comments but I discovered a nasty bug in this that requires major changes. :( I've got a working fix, just need to tidy it up and then fix the tests.

seanh:
- [ ] The annotation permalinks of new annotations are broken, until you reload them

seanh:
- [x] Create an annotation and save it, reload the page, check that it looks right
- [x] Edit an annotation and save it, reload the page, check that it looks right
- [x] Create a highlight and save it, reload the page, check that it looks right
- [x] Edit a highlight and save it, reload the page, check that it looks right
- [x] Create a page note and save it, reload the page, check that it looks right
- [x] Edit a page note and save it, reload the page, check that it looks right
- [x] Create a reply and save it, reload the page, check that it looks right
- [x] Edit a reply and save it, reload the page, check that it looks right
- [ ] Check that when creating a new annotation the initial privacy setting is private or shared depending on what you last selected in the Post dropdown
- [ ] Check that when editing an annotation the initial privacy is as saved on the server, even if this differs from the "sticky" value
- [ ] Check that when replying the initial privacy is that of the annotation being replied to, even if this differs from the "sticky" value
- [x] Create a new annotation. Don't save it. Make it shared. Change the focused group. Save the annotation. Check that it was correctly saved as a shared annotation in the newly focused group.
- [x] Create a new annotation. Don't save it. Make it private. Change the focused group. Save the annotation. Check that it was correctly saved as a private annotation in the newly focused group.
- [x] Edit an existing annotation changing the text, tags and privacy. Change the focused group. Change it back. Check that the annotation's editor is open and your edits to all three fields are still there. Do this with both editing a private annotation to be shared and vice-versa.
- [x] Edit an existing annotation changing the text, tags and privacy. Cancel your edits. Check that the annotation's original values are restored. Do this with both editing a private annotation to be shared and vice-versa.
- [x] Create an annotation and a highlight while logged out. Login. Check that they're saved correctly.
- [x] Test that annotations look correct on the stream
- [x] Test that annotations look correct on individual annotation pages.

seanh:
Ok, I'm gonna say that this should be reviewed again.

In the interests of drawing a line under this I haven't fixed the weird way that `vm.save()` works (see pull request description, I added it to the todo list).

But I have fixed the reverting of privacy when you hit _Cancel_ while editing, and I've removed all writing to `domainModel` outside of `vm.save()`, tested it a lot (see checklists) and fixed a couple of bugs.

It'd be good if the reviewer could also test this carefully. There's a lot of complexity here and I don't trust `annotation-test.js` to find bugs.

seanh:
Rebased-in some changes to the tests for how `AnnotationController` uses drafts.

I'll work on something else now, this can be reviewed.

robertknight:
Updates of annotation tags and text in response to WS notifications appears broken:

**STR**
 1. Create annotation on example.com using bookmarklet in tab A
 2. Open tab B displaying example.com
 3. Edit annotation tags and text in tab B
 4. Go back to tab A
 
**Expected**
 Annotation updated in tab A
 
**Actual**
 Annotation unchanged in tab A
 
**Notes**
 Using the WebSocket frames view in Chrome Dev Tools,
 I can see that when the annotation is updated, the client receives
 the annotation payload but does not display the update.

robertknight:
Issue #2 - Creating annotations whilst signed out:

 1. Sign out
 2. Create annotation
 3. Sign in
 
Expected:
 - Annotation appears in sidebar, ready for editing
 
Actual:
 * Annotation does not appear in sidebar
 * Selecting annotation in page results in sidebar displaying '1 Selected annotation' message but no annotation card is visible

h/static/scripts/directive/annotation.js
```diff
@@ -72,50 +89,93 @@ function extractDocumentMetadata(model) {
...
+ *
+ * Copy any properties from vm that might have been modified by the user into
+ * domainModel, overwriting any existing properties in domainModel.
+ *
+ * Additionally, the `tags` property of vm - an array of objects each with a
+ * `text` string - will become a simple array of strings in domainModel.
+ *
+ * @param {object} domainModel The object to copy properties to
+ * @param {object} vm The object to copy properties from
+ * @returns undefined
```
robertknight: If there is no return value then just leave off the '@returns'

h/static/scripts/directive/annotation.js
```diff
@@ -72,50 +89,93 @@ function extractDocumentMetadata(model) {
...
+}
 
-  // Form the tags for ngTagsInput.
-  vm.annotation.tags = (vm.annotation.tags || []).map(function(tag) {
-    return {text: tag};
-  });
+/** Update the view model from the domain model changes. */
+function updateViewModel(domainModel, vm, permissions) {
+  // Extend the view model with a copy of the domain model.
+  // Note that copy is used so that deep properties aren't shared.
```
robertknight: This comment is out of date - `copy()` is no longer being used here.

h/static/scripts/directive/annotation.js
```diff
@@ -72,50 +89,93 @@ function extractDocumentMetadata(model) {
...
+  vm.form = {
+    text: domainModel.text,
+    tags: viewModelTagsFromDomainModelTags(domainModel.tags),
+  };
+  vm.annotationURI = new URL('/a/' + domainModel.id, vm.baseURI).href;
```
robertknight: Why are we constructing the annotation link ourselves here, doesn't the domainModel include a property which is the link? (ie. `domainModel.link`?)

robertknight: Just checked again - looks like it doesn't.

h/static/scripts/directive/annotation.js
```diff
@@ -426,23 +445,26 @@ function AnnotationController(
...
-          annotationMapper.deleteAnnotation(model).then(
+          annotationMapper.deleteAnnotation(domainModel).then(
             null, onRejected);
         });
       }
     }, true);
   };
 
+  /** Return some metadata extracted from the annotated document. */
+  vm.document = function() {
```
robertknight: Something to be aware of here is that any functions which are used in Angular expressions will be run every time the application does a full digest cycle. If this kind of function were to do something expensive, it could have a performance impact.

h/static/scripts/directive/annotation.js
```diff
@@ -598,41 +609,50 @@ function AnnotationController(
...
+    var newTags = vm.form.tags.filter(function(tag) {
+      var tags = domainModel.tags || [];
       return tags.indexOf(tag.text) === -1;
     });
     tags.store(newTags);
 
     switch (vm.action) {
       case 'create':
-        updateDomainModel(model, vm.annotation);
+        updateDomainModel(domainModel, vm, permissions, groups);
```
robertknight: Something to be aware of here, which might explain the bug I reported earlier, is that the domain model can be updated outside of this directive in response to push notifications received from the WS server.

h/static/scripts/directive/test/annotation-test.js
```diff
@@ -149,196 +149,116 @@ describe('annotation.js', function() {
...
-
-    it('uses the first title when there are more than one', function() {
-      var vm = {};
-      var model = {
-        uri: 'http://example.com/example.html',
-        document: {
-          title: ['first title', 'second title']
-        }
-      };
+    it('copies text from viewModel into domainModel', function() {
```
robertknight: I'm not asking for any further changes here for the moment, but if I were new to this code I think I'd find some of this quite difficult to understand, because it is very much expressed in terms of the language of the implementation, rather than the API of the component.

h/static/scripts/directive/test/annotation-test.js
```diff
@@ -1460,137 +1353,42 @@ describe('annotation.js', function() {
...
-        assert.calledWith(fakeDrafts.remove, annotation);
+        return controller.save().then(function() {
+          assert.calledWith(fakeDrafts.remove, annotation);
+        });
       });
     });
 
-    describe('when the focused group changes', function() {
-      it('updates the current draft', function() {
+    describe('onGroupFocused()', function() {
```
robertknight: I actually preferred the description of this method as "when the focused group changes". The fact that this event is handled by a function called `onGroupFocused()` internally is an implementation detail.

h/templates/client/annotation.html
```diff
@@ -26,17 +26,17 @@
...
         </a>
-        <span ng-show="vm.isPrivate()"
+        <span ng-show="vm.isPrivate"
               title="This annotation is visible only to you.">
           <i class="h-icon-lock"></i><span class="annotation-header__group-name" ng-show="!vm.group().url">Only me</span>
         </span>
         <i class="h-icon-border-color" ng-show="vm.isHighlight() && !vm.editing()" title="This is a highlight. Click 'edit' to add a note or tag."></i>
         <span class="annotation-citation"
-              ng-bind-html="vm.document | documentTitle"
+              ng-bind-html="vm.document() | documentTitle"
```
robertknight: The use of a filter here is unnecessary, this can just be `vm.document().title`, `vm.document().domain`. I think that would enable removing the `documentTitle` and `documentDomain` filters from the app entirely.

seanh:
I think the commit I just pushed https://github.com/hypothesis/h/commit/95921897aaed64177241d3c0ca4816bab21cd2b5 fixes the live updates issue without breaking anything else

h/static/scripts/directive/annotation.js
seanh: Indeed, and `updateViewModel()` was not being called in that case, so the live updates were not working. Should be fixed now.

seanh:
https://github.com/hypothesis/h/commit/7c083bc57893dfc1fc0211f61d45805a349a62f2 fixes the annotating-while-signed-out issue.

h/static/scripts/directive/annotation.js
seanh: Done

h/static/scripts/directive/annotation.js
seanh: Done

h/static/scripts/directive/annotation.js
seanh: Ok, but for now I prefer to remove data duplication, until we've got this code under control, then we can optimise as and when necessary. Good to know, though

h/static/scripts/directive/test/annotation-test.js
seanh: Disagree, these are unit tests for `updateDomainModel()` whose parameters are `viewModel` (well, `vm`) and `domainModel`, so this is in the language of the API of the component (where the component in this context is `updateDomainModel()`).

h/static/scripts/directive/test/annotation-test.js
seanh: This was a deliberate change, and one I'd like to do to the rest of `annotation-test.js` as well. These "when the group changes" and "when it is a highlight" etc etc tests are a problem imo. It's not clear which part of the code each one of them is supposed to be testing, there are lots of tests that duplicate each other or that are no longer valid (but for obscure reasons still passing) or that test several disparate parts of the code at once etc as a result.

In the case of event listeners though, I think I'd be okay with using the name of the event rather than the name of the listener function, so `describe('when the GROUP_FOCUSED event is emitted')` or something.

Although just using the function name would be consistent with tests for other functions and methods.

h/templates/client/annotation.html
seanh: We'd have to move the code from the filters into `vm.document()` (or rather `extractDocumentData()`) though, wouldn't we? But yes, it would be a simplification, since the only time that `vm.document()` is used it's passed through these filters.

Not in this pr though!

PR #2784, opened 2015-12-09
Consolidate API validation

nickstenning:
With a view to turning `h.api.logic` into `h.api.storage` -- and as a result the central point of indirection for all `h.api` storage accesses, I've moved some of the annotation payload prep and validation out of `h.api.logic` and alongside the other validation logic in `h.api.schemas`.

### DIFF
--- a/h/api/schemas.py
+++ b/h/api/schemas.py
@@ -4,7 +4,9 @@
 import copy
 import jsonschema
 from jsonschema.exceptions import best_match
+from pyramid import i18n
 
+_ = i18n.TranslationStringFactory(__package__)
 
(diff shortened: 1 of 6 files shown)


h/api/schemas.py
```diff
@@ -78,13 +80,69 @@ class AnnotationSchema(JSONSchema):
...
+
+class CreateAnnotationSchema(object):
+
+    """
+    Validate the payload from a user when creating an annotation.
+    """
+
+    def __init__(self, request):
+        self.request = request
+        self.structure = AnnotationSchema()
```
seanh: Is `schema` a more obvious name here?

PR #2790, opened 2015-12-10
Move all templates into `templates/` directory

nickstenning:
It would be nice if each module could be responsible for their own templates, but in reality nearly every template forms part of a large tree of templates, extending base layouts, and so on.

In the interests of consistency, it makes sense to have all templates in the `h/templates/` directory rather than some being in `h/templates/` and some being in their respective module directories.


h/claim/views.py
```diff
@@ -28,7 +28,7 @@ def claim_account(request):
 
 @view_config(route_name='claim_account',
              request_method='POST',
-             renderer='h:claim/templates/claim_account.html.jinja2')
+             renderer='h:teampltes/claim/claim_account.html.jinja2')
```
seanh: Typo

nickstenning: Wow. I have no idea how I managed that -- I didn't do these by hand. Anyway, thank you for the eagle eyes.

PR #2795, opened 2015-12-11
Make the Via form on the homepage work as a plain HTML form

robertknight:
This enables the Via link to do something useful even
if the homepage JS has not loaded.

A downside is that this introduces an additional redirect on the happy path. We could remove that but I'm unsure whether it is going to matter.

### DIFF
--- a/h/home.py
+++ b/h/home.py
@@ -1,5 +1,14 @@
 """The home/front page of the //hypothes.is/ site."""
 from pyramid import view
+from pyramid import httpexceptions
+
+
+@view.view_config(route_name='via_redirect',
+                  request_method='GET')
+def via_redirect(context, request):
+    url = request.params['url']
+    via_link = 'https://via.hypothes.is/{}'.format(url)
+    return httpexceptions.HTTPFound(location=via_link)
 
 
 @view.view_config(route_name='index',
@@ -24,4 +33,5 @@ def index(context, request):
 
 def includeme(config):
     config.add_route('index', '/')
+    config.add_route('via_redirect', '/via')
     config.scan(__name__)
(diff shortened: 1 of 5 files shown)


nickstenning:
Why not keep the `onsubmit` handler (although possibly put it somewhere more sensible) so that the current behaviour is retained if JS loads and it still works if it doesn't.

robertknight:
Relevant discussion from Slack: https://hypothes-is.slack.com/archives/public/p1450352404000009
TL;DR - We'll retain the client-side form handling to avoid the redirect for the majority of users.

h/home.py
```diff
@@ -1,5 +1,14 @@
 """The home/front page of the //hypothes.is/ site."""
 from pyramid import view
+from pyramid import httpexceptions
+
+
+@view.view_config(route_name='via_redirect',
+                  request_method='GET')
+def via_redirect(context, request):
+    url = request.params['url']
```
nickstenning: This is a `KeyError` if someone visits without the `url` param. Raise a 404 or perhaps a 400 instead.

h/static/scripts/site.js
```diff
@@ -23,7 +22,7 @@ document.addEventListener('DOMContentLoaded', function () {
   // setup route
   var route = document.location.pathname;
   if (route.match('^/(new-homepage)?$')) {
```
nickstenning: Could you remove the `new-homepage` option from this. Happy for that to be a separate commit/PR.

judell:
This is /really/ helpful, thanks!

h/static/scripts/site.js
robertknight: Done

nickstenning:
Great!

PR #2796, opened 2015-12-12
Small tweaks to the header for re-use in the blog

robertknight:
* Add notes in the header and footer about updating the blog's
   navigation when updating the navigation on the main site.

 * Increase the Z-index for dropdown menus in the header
   to avoid a problem where the search box on the blog
   appeared over the top of the dropdown.

   Add a central place for a Z-Index scale in variables.scss
   to avoid such problems in future.

 * Set the horizontal padding explicitly for the nav dropdown
   menus. Previously the site relied on padding set by
   a compass reset which was not present on the blog which
   uses a much more minimal reset.

### DIFF
--- a/h/static/styles/header.scss
+++ b/h/static/styles/header.scss
@@ -107,8 +107,10 @@
 
     padding-top: 2px;
     padding-bottom: 2px;
+    padding-left: 0px;
+    padding-right: 0px;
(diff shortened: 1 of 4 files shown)


nickstenning:
LGTM.

PR #2801, opened 2015-12-14
Enable showdown's literalMidWordUnderscores

### LINKED ISSUE #2598: link autoactivation turns _ into <em>
### Steps to reproduce:

1. Create an annotation
2. Paste a URL into the annotation body which contains underscores, such as `https://hypothes.is/stream?q=tag:group_test_needs_card`
3. Save and view

### Expected result:

An unmodified URL in the annotation card, e.g.:

> https://hypothes.is/stream?q=tag:group_test_needs_card

### Actual result:

The underscores in the URL get turned into emphasis tags, e.g.:

> https://hypothes.is/stream?q=tag:group%3Cem%3Etest%3C/em%3Eneeds%3Cem%3Ecard

seanh:
This may be a controversial way of fixing this bug, since it changes Showdown's output in ways other than fixing underscores in link as well, but it seems to be what the showdown authors recommend: <https://github.com/showdownjs/showdown/issues/211>.

Also, I haven't looked, but I suspect it might be hard to fix the behaviour otherwise, it looks to me like by the time showdown's auto-link sub-parser runs the underscores in links have already been turned into `<em>`s by other showdown code. I'm guessing this may also happen before any showdown extension we write runs. So it _may_ not be simple to fix this in another way:

This stops showdown from interpreting underscores in the middle of words
as `<em>` and `<strong>`, instead it treats them as literal underscores.

Example:

    some text with__underscores__in middle

will be parsed as

    <p>some text with__underscores__in middle</p>

Fixes #2598 (URLs with `_`'s in them were getting `<em>` tags inserted into
them, breaking the links).

### DIFF
--- a/h/static/scripts/filter/converter.js
+++ b/h/static/scripts/filter/converter.js
@@ -11,7 +11,12 @@ module.exports = function () {
   // see https://github.com/showdownjs/showdown#valid-options
   var converter = new showdown.Converter({
     extensions: [targetBlank],
-    simplifiedAutoLink: true
+    simplifiedAutoLink: true,
+    // Since we're using simplifiedAutoLink we also use
+    // literalMidWordUnderscores because otherwise _'s in URLs get
+    // transformed into <em>'s.
+    // See https://github.com/showdownjs/showdown/issues/211
+    literalMidWordUnderscores: true,
   });
   return converter.makeHtml.bind(converter);
 };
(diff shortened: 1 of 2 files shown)


robertknight:
I think the approach taken here makes sense, but can you add a reference to the Showdown issue as a code comment with a brief note on why we're enabling the option, in case it turns out to have unforeseen consequences in future.

Also, even though this is just enabling an option to showdown, I think it might still be useful to have a test on our side to verify that the behavior is preserved after any future changes.

nickstenning:
:+1: to the solution but also :+1: to including a comment explaining why the option is being used.

seanh:
Docs and test added

PR #2802, opened 2015-12-16
T158 - Update truncated annotation card design

robertknight:
This updates the design of truncated annotation cards following https://trello.com/c/Jg0NWyF3/158-update-design-for-truncated-annotations and discussion with Conor.

**User-visible changes**

 * The card has been made more compact by moving the expansion controls for quotes inline with
   the last line of the quote itself, showing only two lines of the quote when collapsed and
   moving the expansion control for the body into the tags line
 * Added an animation when expanding and collapsing the card
 * Make expanding annotation bodies easier by providing a large clickable area at the bottom
   of the card which expands the body when clicked

t158-updated-trunc-design

**Implementation changes**
 * The changes to the `<excerpt>` component are to give it two modes - one which provide inline controls to handle expansion/collapse and another which allows the containing directive to provide
those controls.
 * I've removed the controller tests and intend to only implement tests at the directive level,
   based on the fact that I went through about 3 different approaches to the implementation
   which had almost the same external interface but changed the internals completely.

---

**TODO**
- [x] Make expanding the quote easier but adding a hit target across the whole of the bottom of the quote, for ease of use and consistency with how the annotation body behaves.
- [x] 'More' link for annotation body should be grey when the card lacks focus
- [x] 'More' / 'Less' link state for annotation body not correct immediately after switching from edit to view mode
- [x] Fix cut-off when excerpt expands from one height to another and results in the document gaining a vertical scrollbar as a result (_Fixed in the sense that the height will be corrected after the next digest cycle after the scrollbar appears_)

**Needs additional testing**
(These issues I haven't yet been able to reproduce exactly as in https://github.com/hypothesis/h/pull/2802#issuecomment-165434711 , but a couple of the above issues were due to not using data-binding properly, and the second/third of these might be related to that)
- Investigate drop shadow not spanning full width in [Nick's screenshot](https://cloud.githubusercontent.com/assets/3602/11869132/48838c36-a4bd-11e5-85a8-7652b797756b.png)
- Investigate 'More' link being shown redundantly on a body that didn't need collapsing
- Investigate body not being collapsed (even though it should have been) and 'More' link still showing

### DIFF
--- a/h/static/scripts/directive/annotation.js
+++ b/h/static/scripts/directive/annotation.js
@@ -587,6 +587,11 @@ function AnnotationController(
     }
   };
 
+  vm.toggleCollapseBody = function(event) {
+    event.stopPropagation();
+    vm.collapseBody = !vm.collapseBody;
+  };
+
   /**
     * @ngdoc method
     * @name annotation.AnnotationController#reply
--- a/h/static/scripts/directive/annotation.js
+++ b/h/static/scripts/directive/annotation.js
@@ -304,7 +304,7 @@ function AnnotationController(
     /** Determines whether controls to expand/collapse the annotation body
      * are displayed adjacent to the tags field.
      */
-    vm.canCollapseBody = true;
+    vm.canCollapseBody = false;
 
     /** Determines whether the annotation body should be collapsed. */
     vm.collapseBody = true;
--- a/h/static/scripts/directive/annotation.js
+++ b/h/static/scripts/directive/annotation.js
@@ -301,6 +301,14 @@ function AnnotationController(
     /** A callback for resetting the automatic refresh of vm.timestamp */
     vm.cancelTimestampRefresh = undefined;
 
+    /** Determines whether controls to expand/collapse the annotation body
+     * are displayed adjacent to the tags field.
+     */
+    vm.canCollapseBody = true;
+
+    /** Determines whether the annotation body should be collapsed. */
+    vm.collapseBody = true;
+
     /** The domain model, contains the currently saved version of the
       * annotation from the server (or in the case of new annotations that
       * haven't been saved yet - the data that will be saved to the server when
@@ -749,6 +757,14 @@ function AnnotationController(
 
   vm.user = function() {
     return domainModel.user;
+  }
+
+  /** Sets whether or not the controls for
+   * expanding/collapsing the body of lengthy annotations
+   * should be shown.
+   */
+  vm.setBodyCollapsible = function(canCollapse) {
+    vm.canCollapseBody = canCollapse;
   };
 
   init();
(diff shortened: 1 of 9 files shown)


nickstenning:
There's all kinds of stuff broken here, I'm afraid.

- [ ] Drop shadow seems wrong (and not as displayed in your screenshot):

  screen shot 2015-12-17 at 12 52 32

- [ ] Not collapsed on stream, but "More" link shown regardless:

  screen shot 2015-12-17 at 12 55 04

- [ ] "More" link shown on annotation that clearly doesn't need collapsing:

  screen shot 2015-12-17 at 12 56 05

- [ ] "More" link shown on a newly created reply that clearly doesn't need collapsing:

  screen shot 2015-12-17 at 12 56 44

robertknight:
An additional bug I see in the sidebar. If clicking the 'More' link results in the sidebar expanding from a height at which a scrollbar _is not required_ to one at which a scrollbar _is required_, the bottom of the annotation body or quote can get cut off:

bottom-cutoff-1

bottom-cutoff-2

robertknight:
I haven't been able to reproduce the clipped shadow from the first screenshot, but I can reproduce a variation of the issue with the 'More' link showing redundantly as follows:

1. Start creating an annotation with a short quote
2. Set the body text to [this text](https://gist.github.com/robertknight/dea5db004985c7fe38de)
3. Click 'Post to Public'

Expected: Truncated annotation body shown, with 'More' link to expand it
Actual: Full annotation body shown, clicking 'More' link once does nothing, clicking 'Less' link then collapses it to the correct height

robertknight:
Another way to reproduce https://github.com/hypothesis/h/pull/2802#issuecomment-165887193 is to turn on the feature flag after the page loads. The result is that the 'More' link appears when the flag is enabled but the annotation card height is not adjusted.

This issue appears different than the screenshots above however since in that context, the 'More' link is visible but the drop shadow at the bottom of the annotation is not present.

robertknight:
I was not able to reproduce all of the original issues in @nickstenning 's comment but I've simplified the way the `<excerpt>` content's state is updated and fixed the issues that I could reproduce. I suggest we give this another round of testing and see whether any of the original reported problems can still be reproduced.

I've tested on FF/Chrome/Edge under Linux, Mac and Windows 10. Re. the drop shadow, one small difference is that my Mac does not have a retina display, although I can't immediately see how that would affect the issue at hand.

nickstenning:
This is looking very good to me. Only minor niggle is I think you need to stop propagation of the click event somewhere because clicking on "more/less" links scroll the hosting page to the highlight at the moment, which I don't think is desired.

robertknight:
> Only minor niggle is I think you need to stop propagation of the click event somewhere...

Indeed. Fixed.

nickstenning:
Thank you for all your hard work on this @robertknight. It's looking great.

PR #2803, opened 2015-12-16
Decouple password reset flow from account activation

finitud:
# Migration notes

This pull request contains three migrations:
* The first one will add a `password_updated` column to the `user` table without a default value (to prevent locking the production database), and then adds a `server_default` which will provide a default value of `now()` for all new entries.
* The second one will fill the missing `password_updated` fields with the default value.
* The third one one will make the `password_updated` field non-nullable, and needs to be run after the pull request is deployed to production. Otherwise Bad Things Will Happen :tm: 

Migrations #1 and #2 are safe to run on the production database even if this pull request hasn't been deployed to the production environment. Migration #3 isn't.

# Implementation notes

Password reset was dealt with by re-using the account activation mechanism, and the password reset token was thus an activation token. This had the unpleasant side effect that if a user created an account, and asked for a password reset before activating the account, the password reset token would overwrite the activation token, effectively locking them out of their account.

The solution implemented here uses a capability URL, where the information needed for the password reset is encoded in the password reset token itself. In particular, a password reset token contains the username of the requestor, and it then becomes URL-safe encoded, timestamped, and signed with a secret key.

In addition, we add a new `password_updated` column to the user table in the database, which gets set to the current time every time the user's password is set.

When the user attempts to reset their password by using a password reset token, the signature of the provided token is verified against our secret key, and the timestamped is checked. If this verification succeeds, we reset the user's password, and update the `password_updated` field in the corresponding database entry.

The password reset fails if the token is invalid, older than 72h, or the user has reset their password after the token creation.

### DIFF
--- a/h/accounts/__init__.py
+++ b/h/accounts/__init__.py
@@ -1,4 +1,8 @@
 # -*- coding: utf-8 -*-
+from itsdangerous import URLSafeTimedSerializer
+
+from ..security import derive_key
+
 from h import util
 from h.accounts import models
 
@@ -96,3 +100,8 @@ def includeme(config):
     config.include('.schemas')
     config.include('.subscribers')
     config.include('.views')
+
+    secret = config.registry.settings['secret_key']
+    derived = derive_key(secret, b'h.accounts')
+    serializer = URLSafeTimedSerializer(derived)
+    config.registry.password_reset_serializer = serializer
--- a/h/accounts/__init__.py
+++ b/h/accounts/__init__.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 from itsdangerous import URLSafeTimedSerializer
 
-from ..security import derive_key
+from h.security import derive_key
 
 from h import util
 from h.accounts import models
--- a/h/accounts/models.py
+++ b/h/accounts/models.py
@@ -131,10 +131,10 @@ class User(Base):
     # Password salt
     salt = sa.Column(sa.UnicodeText(), nullable=False)
     # Last password update
-    last_password_update = sa.Column(sa.DateTime(timezone=False),
+    last_password_update = sa.Column(sa.DateTime(),
                                      default=sa.sql.func.now(),
                                      server_default=sa.func.now(),
-                                     nullable=True)
+                                     nullable=False)
 
     @hybrid_property
     def password(self):
--- a/h/accounts/models.py
+++ b/h/accounts/models.py
@@ -131,10 +131,10 @@ class User(Base):
     # Password salt
     salt = sa.Column(sa.UnicodeText(), nullable=False)
     # Last password update
-    last_password_update = sa.Column(sa.DateTime(),
-                                     default=sa.sql.func.now(),
-                                     server_default=sa.func.now(),
-                                     nullable=False)
+    password_updated = sa.Column(sa.DateTime(),
+                                 default=sa.sql.func.now(),
+                                 server_default=sa.func.now(),
+                                 nullable=False)
 
     @hybrid_property
     def password(self):
@@ -152,7 +152,7 @@ class User(Base):
             raise ValueError('password must be more than {min} characters '
                              'long'.format(min=PASSWORD_MIN_LENGTH))
         self._password = self._hash_password(raw_password)
-        self.last_password_update = sa.sql.func.now()
+        self.password_updated = sa.sql.func.now()
 
     def _hash_password(self, password):
         if not self.salt:
--- a/h/accounts/models.py
+++ b/h/accounts/models.py
@@ -133,7 +133,7 @@ class User(Base):
     # Last password update
     password_updated = sa.Column(sa.DateTime(),
                                  default=sa.sql.func.now(),
-                                 server_default=sa.func.now(),
+                                 server_default=sa.sql.func.now(),
                                  nullable=False)
 
     @hybrid_property
--- a/h/accounts/models.py
+++ b/h/accounts/models.py
@@ -32,7 +32,7 @@ def _generate_random_string(length=12):
 class Activation(Base):
 
     """
-    Handles activations/password reset items for users.
+    Handles activations for users.
 
     The code should be a random hash that is valid only once.
     After the hash is used to access the site, it'll be removed.
@@ -130,6 +130,11 @@ class User(Base):
     _password = sa.Column('password', sa.UnicodeText(), nullable=False)
     # Password salt
     salt = sa.Column(sa.UnicodeText(), nullable=False)
+    # Last password update
+    last_password_update = sa.Column(sa.DateTime(timezone=False),
+                                     default=sa.sql.func.now(),
+                                     server_default=sa.func.now(),
+                                     nullable=True)
 
     @hybrid_property
     def password(self):
@@ -147,6 +152,7 @@ class User(Base):
             raise ValueError('password must be more than {min} characters '
                              'long'.format(min=PASSWORD_MIN_LENGTH))
         self._password = self._hash_password(raw_password)
+        self.last_password_update = sa.sql.func.now()
 
     def _hash_password(self, password):
         if not self.salt:
--- a/h/accounts/test/views_test.py
+++ b/h/accounts/test/views_test.py
@@ -47,6 +47,14 @@ class FakeUser(object):
             setattr(self, k, kwargs[k])
 
 
+class FakeSerializer(object):
+    def dumps(self, obj):
+        return 'faketoken'
+
+    def loads(self, token):
+        return {'username': 'foo@bar.com'}
+
+
 # A fake version of colander.Invalid
 class FakeInvalid(object):
     def __init__(self, errors):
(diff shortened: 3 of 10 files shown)


h/accounts/__init__.py
```diff
@@ -1,4 +1,8 @@
 # -*- coding: utf-8 -*-
+from itsdangerous import URLSafeTimedSerializer
```
seanh: This should just be `import itsdangerous`, see <https://google.github.io/styleguide/pyguide.html#Imports>

h/accounts/__init__.py
```diff
@@ -1,4 +1,8 @@
...
+
+from ..security import derive_key
```
seanh: Should be `from h import security`. Also for the rest of the imports

h/accounts/__init__.py
```diff
@@ -96,3 +100,8 @@ def includeme(config):
     config.include('.schemas')
     config.include('.subscribers')
     config.include('.views')
+
+    secret = config.registry.settings['secret_key']
+    derived = derive_key(secret, b'h.accounts')
+    serializer = URLSafeTimedSerializer(derived)
+    config.registry.password_reset_serializer = serializer
```
seanh: We try not to modify the config registry at runtime, instead add a method to the request object, for example see <https://github.com/hypothesis/h/blob/master/h/accounts/__init__.py#L92> (or grep for `add_request_method`)

h/accounts/__init__.py
```diff
@@ -96,3 +100,8 @@ def includeme(config):
     config.include('.schemas')
     config.include('.subscribers')
     config.include('.views')
+
+    secret = config.registry.settings['secret_key']
+    derived = derive_key(secret, b'h.accounts')
```
seanh: I don't think the `b` is necessary here? Byte strings are the default in Python 2

h/accounts/test/views_test.py
```diff
@@ -47,6 +47,14 @@ def __init__(self, **kwargs):
             setattr(self, k, kwargs[k])
 
 
+class FakeSerializer(object):
+    def dumps(self, obj):
+        return 'faketoken'
+
+    def loads(self, token):
+        return {'username': 'foo@bar.com'}
```
seanh: I think all these fake classes are fine, but they could also be done with the `mock` library, for example this one might be something like:

```
fakeSerializer = mock.Mock()
fakeSerializer.dumps.return_value = 'faketoken'
fakeSerializer.loads.return_value = {'username': 'foo@bar.com'}
```

h/accounts/__init__.py
nickstenning: Alice is just copying the existing pattern for adding these serializers. I think this is fine.

h/accounts/__init__.py
nickstenning: But we have a codebase that we're trying to prepare for Python 3 compatibility. These are bytestrings and must remain bytestrings when interpreted by Python 3, so this is correct.

h/accounts/models.py
```diff
@@ -130,6 +130,11 @@ def is_activated(self):
     _password = sa.Column('password', sa.UnicodeText(), nullable=False)
     # Password salt
     salt = sa.Column(sa.UnicodeText(), nullable=False)
+    # Last password update
+    last_password_update = sa.Column(sa.DateTime(timezone=False),
```
nickstenning: `timezone=False` is [the default](http://docs.sqlalchemy.org/en/latest/core/type_basics.html#sqlalchemy.types.DateTime) for `sa.DateTime`, so you can omit it here.

h/accounts/models.py
```diff
@@ -130,6 +130,11 @@ def is_activated(self):
...
+                                     default=sa.sql.func.now(),
+                                     server_default=sa.func.now(),
+                                     nullable=True)
```
nickstenning: This shouldn't be nullable here. The idea is that you can add the metadata in the model as non-nullable, and then add a nullable column with the migration.

h/accounts/views.py
```diff
@@ -293,9 +285,11 @@ def reset_password_with_code_form(self):
         # If valid, we inject the supplied it into the form as a hidden field.
         # Otherwise, we 404.
         try:
-            user = schemas.ResetCode().deserialize(None, code)
+            user = schemas.ResetCode().deserialize(self.schema, code)
         except colander.Invalid:
             raise httpexceptions.HTTPNotFound()
+        except BadSignature:
```
nickstenning: This should be caught in the the validator and reraised as `colander.Invalid`, partially because `Invalid` contains additional metadata about what was invalid that's used by `deform`, but mainly because this is exposing implementation details of the validator to its client.

setup.py
```diff
@@ -64,6 +64,7 @@ def run_tests(self):
     'requests>=2.7.0',
     'ws4py>=0.3,<0.4',
     'zope.sqlalchemy>=0.7.6,<0.8.0',
+    'itsdangerous>=0.24',
```
nickstenning: Sorry to be a pain, but these are in alphabetical order, so if we could keep them that way that would be great :)

h/migrations/versions/530268a1937c_add_password_last_updated_column_to_.py
```diff
@@ -0,0 +1,26 @@
...
+down_revision = '43645baa68b2'
+
+from alembic import op
+import sqlalchemy as sa
+
+
+def upgrade():
+    op.add_column('user', sa.Column('last_password_update',
+                                    sa.DateTime(timezone=False),
+                                    default=sa.sql.func.now(),
```
nickstenning: `default` and `server_default` aren't doing anything here, so you can omit them.

h/migrations/versions/530268a1937c_add_password_last_updated_column_to_.py
```diff
@@ -0,0 +1,26 @@
...
+revision = '530268a1937c'
+down_revision = '43645baa68b2'
+
+from alembic import op
+import sqlalchemy as sa
+
+
+def upgrade():
+    op.add_column('user', sa.Column('last_password_update',
+                                    sa.DateTime(timezone=False),
```
nickstenning: Ditto about `timezone=False`.

h/accounts/views.py
```diff
@@ -293,9 +285,11 @@ def reset_password_with_code_form(self):
         # If valid, we inject the supplied it into the form as a hidden field.
         # Otherwise, we 404.
         try:
-            user = schemas.ResetCode().deserialize(None, code)
+            user = schemas.ResetCode().deserialize(self.schema, code)
```
nickstenning: I'm pretty sure passing in `None` for the colander node is fine. Is there a reason you're passing in the schema now?

seanh:
What should happen if I:

- Register a new account, but don't activate it
- Request a password reset and use it
- Try to login with the new password

I think I should be allowed to login, which implies that resetting my password should implicitly activate my account. (Current behaviour is that I can reset my password, but still can't login until I've activated.)

h/accounts/test/views_test.py
```diff
@@ -339,37 +347,33 @@ def test_forgot_password_creates_no_activations_when_validation_fails(activation
 
 
 @forgot_password_fixtures
-def test_forgot_password_creates_activation_for_user(activation_model,
-                                                     authn_policy):
+def test_forgot_password_create_no_activation_for_user(activation_model,
```
nickstenning: Just delete this test. We're not going to check that every component that's not supposed to be used by `forgot_password` isn't used by `forgot_password` :)

seanh:
Related (but no this branch's fault): https://github.com/hypothesis/h/issues/2810

nickstenning:
> I think I should be allowed to login, which implies that resetting my password should implicitly activate my account. (Current behaviour is that I can reset my password, but still can't login until I've activated.)

Agreed, but that can be a separate piece of work from this PR.

nickstenning:
Two more things have occurred to me:

1. For consistency with other timestamp fields (see, for example, the groups model) I think I'd prefer that the DB column was called `password_updated`.
2. In case you were wondering, it's perfectly ok to add the data migration and the constraint migration to this PR. We can run them one at a time when deploying.

h/accounts/__init__.py
finitud: :+1:

h/accounts/__init__.py
finitud: :+1:

finitud:
> For consistency with other timestamp fields (see, for example, the groups model) I think I'd prefer that the DB column was called password_updated.

I have updated the column name to reflect this.

h/accounts/test/views_test.py
finitud: Oh, this looks nice! I went for the fake classes since that was the approach that I found already there in the code, but this one looks quite neat. Which one would be the preferred way going forward?

h/accounts/models.py
```diff
@@ -130,6 +130,11 @@ def is_activated(self):
     _password = sa.Column('password', sa.UnicodeText(), nullable=False)
     # Password salt
     salt = sa.Column(sa.UnicodeText(), nullable=False)
+    # Last password update
+    password_updated = sa.Column(sa.DateTime(),
+                                 default=sa.sql.func.now(),
```
nickstenning: `sa.sql.func` == `sa.func`, so the difference between the `default` and `server_default` lines is a bit confusing here.

h/migrations/versions/530268a1937c_add_password_last_updated_column_to_.py
```diff
@@ -0,0 +1,25 @@
...
+down_revision = '43645baa68b2'
+
+from alembic import op
+import sqlalchemy as sa
+
+
+def upgrade():
+    op.add_column('user', sa.Column('password_updated',
+                                    sa.DateTime(),
+                                    server_default=sa.func.now(),
```
nickstenning: As discussed in slack, don't add the `server_default` on creation, as that makes this a rather unfriendly migration (locks table, rewrites every row). So:

1. Add a line to this migration which does `op.alter_column('user', 'password_updated', server_default=sa.func.now())`.
2. Add an intermediate data migration which adds the data to rows with missing values for this column. See [this migration](https://github.com/hypothesis/h/blob/0d4a6a0449ffd7983adfef945afae04f8e37cc11/h/migrations/versions/1f82507200d7_fill_in_missing_pubids.py) for an example.
3. Update the last migration (the one that does `ALTER COLUMN password_updated NOT NULL`) to reference the data migration as its parent.

finitud: :+1:

h/accounts/models.py
finitud: I've made these consistent.

nickstenning:
Great stuff, @wyan. You certainly dived straight into the deep end here. Sorry it's taken so long to get this reviewed and merged, but here we go...!

lenazun:
🎉

PR #2804, opened 2015-12-16
jscs: Require `function () {}`

seanh:
Change jscs's required style from `function()` to `function ()`.

This deviates from the Google style guide and jscs presets that we use.

### DIFF
--- a/.jscsrc
+++ b/.jscsrc
@@ -6,5 +6,19 @@
     "node_modules/**",
     "h/static/scripts/vendor"
   ],
-  "maxErrors": 10
+  "maxErrors": 10,
+  "disallowSpacesInAnonymousFunctionExpression": null,
+  "disallowSpacesInFunctionDeclaration": {
+    "beforeOpeningRoundBrace": true
+  },
+  "disallowSpacesInFunctionExpression": null,
+  "disallowSpacesInNamedFunctionExpression": {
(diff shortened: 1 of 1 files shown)


robertknight:
The rule doesn't quite match the coding style that has been used for the hand-written JS. The rule we have in our current code is that there is always a space after the function keyword, not that there is a space after the function name. ie. `function foo() {}`, `function () {}`. This `jscs` rule differs slightly in that it requires braces after function _names_.

The rule we want to set I think is: http://jscs.info/rule/requireSpacesInAnonymousFunctionExpression

seanh:
Ok, after a great deal of fiddling, I now have:

```
var foo1 = function () {};  // Right.
var foo2 = function() {};  // Wrong.
var foo3 = function (){};  // Wrong.

var foo4 = function foo4() {};  // Right.
var foo5 = function foo5 () {};  // Wrong.
var foo6 = function foo6(){};  // Wrong.

function foo7() {}  // Right.
function foo8 () {}  // Wrong.
function foo9(){}  // Wrong.
```

robertknight:
LGTM. FWIW, I don't have particularly strong views about any particular style and in future I'd be happy to have the entire codebase use a style that requires minimal configuration of any linter, but I'd like to have a tool in place to do the formatting automatically before that.

PR #2805, opened 2015-12-16
Add media embeds feature

seanh:
https://trello.com/c/TPUsXCk4/192-enable-users-to-display-videos-within-annotations-instead-of-as-links

- <del>[ ] SoundCloud</del> ([see comment](https://github.com/hypothesis/h/pull/2805#issuecomment-165463456))
- [x] CSS: The iframes need to be the right width
- <del>[ ] What attrs should the iframes have on them? Fullscreen etc</del> (YouTube and Vimeo just set `frameborder="0"` which I've added, and 
`allowfullscreen`, [which doesn't work](https://github.com/hypothesis/h/pull/2805#issuecomment-165477549))
- [x] Replace URI.js with something more minimal like <https://github.com/sindresorhus/query-string>
- [x] Feature flag
- [x] Changelog

---

TODO Post merge:
 - [ ] Scan our existing data for YouTube, Vimeo and SoundCloud links, see if we missed any link formats

### DIFF
--- a/h/static/scripts/media-embedder.js
+++ b/h/static/scripts/media-embedder.js
@@ -1,0 +1,155 @@
+'use strict';
+
+/**
+ * Return an iframe DOM element with the given src URL.
+ */
+function iframe(src) {
+  var iframe_ = document.createElement('iframe');
+  iframe_.src = src;
+  iframe_.classList.add('annotation-media-embed');
+  iframe_.setAttribute('frameborder', '0');
+  return iframe_;
+}
+
+/**
+ * Return a YouTube embed (<iframe>) DOM element for the given video ID.
+ */
+function youTubeEmbed(id) {
+  return iframe('https://www.youtube.com/embed/' + id);
+}
+
+function vimeoEmbed(id) {
+  return iframe('https://player.vimeo.com/video/' + id);
+}
+
+/**
+ * A list of functions that return an "embed" DOM element (e.g. an <iframe>)
+ * for a given link.
+ *
+ * Each function either returns `undefined` if it can't generate an embed for
+ * the link, or a DOM element if it can.
+ *
+ */
+var embedGenerators = [
+
+  // Matches URLs like https://www.youtube.com/watch?v=rw6oWkCojpw
+  function iframeFromYouTubeWatchURL(link) {
+    if (link.hostname !== 'www.youtube.com') {
+      return;
+    }
+
+    if (!/\/watch\/?/.test(link.pathname)) {
+      return;
+    }
+
+    var groups = /[&\?]v=([^&#]+)/.exec(link.search);
+    if (groups) {
+      return youTubeEmbed(groups[1]);
+    }
+  },
+
+  // Matches URLs like https://youtu.be/rw6oWkCojpw
+  function iframeFromYouTubeShareURL(link) {
+    if (link.hostname !== 'youtu.be') {
+      return;
+    }
+
+    var groups = /^\/([^\/]+)\/?$/.exec(link.pathname);
+    if (groups) {
+      return youTubeEmbed(groups[1]);
+    }
+  },
+
+  // Matches URLs like https://vimeo.com/149000090
+  function iFrameFromVimeoLink(link) {
+    if (link.hostname !== 'vimeo.com') {
+      return;
+    }
+
+    var groups = /^\/([^\/\?#]+)\/?$/.exec(link.pathname);
+    if (groups) {
+      return vimeoEmbed(groups[1]);
+    }
+  },
+
+  // Matches URLs like https://vimeo.com/channels/staffpicks/148845534
+  function iFrameFromVimeoChannelLink(link) {
+    if (link.hostname !== 'vimeo.com') {
+      return;
+    }
+
+    var groups = /^\/channels\/[^\/]+\/([^\/?#]+)\/?$/.exec(link.pathname);
+    if (groups) {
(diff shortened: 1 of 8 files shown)


h/static/scripts/media-embedder.js
```diff
@@ -0,0 +1,134 @@
...
+ * Each function either returns `undefined` if it can't generate an embed for
+ * the URL, or a DOM element if it can.
+ */
+var embedGenerators = [
+  function iframeFromYouTubeWatchURL(url) {
+    if (!/^https:\/\/www.youtube.com\/watch(.*)$/.test(url)) {
+      return;
+    }
+
+    var id = URI(url).search(true).v;
```
nickstenning: I'm not entirely sure a 61KiB external dependency is justified for something that could be done with a single regular expression: `[?&]v=([^&#]+)`

seanh:
We can't do SoundCloud embeds in the simple way that we've done YouTube and Vimeo ones here. The SoundCloud URL (e.g. <https://soundcloud.com/chrislake/chris-lorenzo-chris-lake-the-calling>) doesn't contain the video ID (e.g. `237101415`) needed for the embed src.

Suggest we stick to just YouTube and Vimeo for now, we can add SoundClould later.

We could:

1. Fetch the SoundCloud URL's HTML page, the video ID is in the HTML source in lots of places, e.g. in lots of meta headers like this: `<meta property="twitter:player" content="https://w.soundcloud.com/player/?url=https%3A%2F%2Fapi.soundcloud.com%2Ftracks%2F237101415&amp;auto_play=false&amp;show_artwork=true&amp;visual=true&amp;origin=twitter">`.

2. [Use SoundCloud's client-side JavaScript library](https://developers.soundcloud.com/docs/api/guide#playing) (requires adding a script to our page)

3. [Use their HTTP API's /resolve endpoint](https://developers.soundcloud.com/docs/api/reference#resolve) to resolve the URL to the ID.

1 and 3 seem to me like they should be done server-side, fetching the ID for a given URL once only and keeping our own cache of them, as opposed to for example every client fetching every ID every time they render an annotation.

seanh:
Not sure why, but fullscreen doesn't work, `<iframe src="..." class="..." allowfullscreen mozallowfullscreen webkitallowfullscreen>` doesn't work (created by doing `iframe_.setAttribute('allowfullscreen', ''); iframe_.setAttribute('mozallowfullscreen', ''); iframe_.setAttribute('webkitallowfullscreen', '');` in JavaScript).

seanh:
@robertknight I haven't gone looking for more kinds of YouTube and Vimeo links in our production data yet, but I think what's here can be reviewed now.

I ended up using the DOM API more to parse the URLs. If you have a DOM element for an `<a>` you can use `el.hostname`, `el.pathname` and `el.search`. This avoids the need for URI.js or any URL-parsing lib, or for long regexes. I still use regexes to match against parts of the path but I've tried to be pretty careful with them.

One potential issue still remaining is that I've set the width and height of the videos to fit nicely in the sidebar, but this makes them too thin for the stream. Doesn't look great on the stream.

seanh:
If there's no other issues with this I think we could merge it now (it's behind a feature flag) and do the search of production data and fixing the video sizes on the stream later.

robertknight:
@seanh - My preference re. SoundCloud would be to do that server-side, I'm very wary of loading 3rd party scripts into our app, except when isolated in an iframe.

h/static/scripts/media-embedder.js
```diff
@@ -0,0 +1,155 @@
...
+ *
+ * If the given link element is a link to an embeddable media then it will be
+ * replaced in the DOM with an embed (e.g. an <iframe>) of the same media.
+ *
+ * If it's not an embeddable link the link will be left untouched.
+ *
+ */
+function replaceLinkWithEmbed(link) {
+
+  // If the user gives a custom link text then we don't replace the link with
```
robertknight: This comment should be part of the function documentation and I think we should also provide a summary of the rationale for it in the context of replacing links that originated from markdown.

seanh: This was done in a follow-up pr

seanh:
> @seanh - My preference re. SoundCloud would be to do that server-side, I'm very wary of loading 3rd party scripts into our app, except when isolated in an iframe.

Yeah, that's my feeling too. Since I doubt that SoundCloud page -> video ID mappings ever change, caching them should be very easy. Maybe something like this:

1. We have a worker process that takes annotations off a queue, scans them for SoundCloud links, consults our SoundCloud cache or calls the SoundCloud API if it has to, adds the SoundCloud ID to the annotation. Probably want some throttling of our SoundCloud API requests.

2. Send the annotation to this worker on every create / update.

3. Since the worker is asynchronous, after the client has sent an annotation create or update to the server and received the response the SoundCloud embeds may not have updated yet. The client may need to receive an update down the WebSocket a moment later before it can display the embed.

The exact details of how it'd work would depend on whether we end up replacing SoundCloud links inline, or showing embeds at the bottom of the card.

I also think it'd be cool if the API it calls to put an annotation onto the worker's queue was also accessible as an admins-only HTTP/JSON API. That way we can make it re-parse annotations manually easy, and it's easy to do a "migration" of production data by running a script on your laptop. Maybe it'd even be handy to have an API to feed SoundCloud URLs onto the SoundCloud fetcher's queue, just to play with / test / pre-fill / make corrections to our SoundCloud cache.

I'm just into having HTTP/JSON "maintenance hatches" for this sort of thing.

PR #2807, opened 2015-12-17
Consolidate /app and /app/features endpoints

robertknight:
Experience has taught us that the client needs the session
data (current user, list of groups) and set of enabled feature
flags at the same time, and also needs to invalidate them
in the same scenarios (eg. account switching).

Fetching this data via two separate requests made it more complicated
to ensure the client had a consistent view of session and feature-flag
data. To simplify things and also same a network request:

 * On the server, put the feature flag data into the session data payload.

 * On the client, use the existing central storage and cache management
   for the session data payload to manage feature flags as well.

   The features service now becomes a thin wrapper around part of
   the session state.

### DIFF
--- a/h/features.py
+++ b/h/features.py
@@ -137,14 +137,25 @@ def remove_old_flags_on_boot(event):
     remove_old_flags()
 
 
+def all(request):
+    """
+    Returns a dict mapping feature flag names to enabled states
+    for the user associated with a given request.
+    """
+    return {k: flag_enabled(request, k) for k in FEATURES.keys()}
+
+
+# Deprecated dedicated endpoint for feature flag data,
+# kept for compatibility with older clients (<= 0.8.6).
+# Newer clients get feature flag data as part of the session data
+# from the /app endpoint.
 @view_config(route_name='features_status',
              request_method='GET',
              accept='application/json',
              renderer='json',
              http_cache=(0, {'no_store': False}))
 def features_status(request):
-    """Report current feature flag values."""
-    return {k: flag_enabled(request, k) for k in FEATURES.keys()}
+    return all(request)
 
 
 def includeme(config):
--- a/h/session.py
+++ b/h/session.py
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 from pyramid.session import SignedCookieSessionFactory
 
+from h import features
 from h import models
 from h.security import derive_key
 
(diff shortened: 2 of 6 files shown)


h/features.py
```diff
@@ -137,13 +137,17 @@ def remove_old_flags_on_boot(event):
...
+# via the /app endpoint. The /app/features endpoint is exposed as a separate
+# service for compatibility with older clients (<= 0.8.6)
 @view_config(route_name='features_status',
              request_method='GET',
              accept='application/json',
              renderer='json',
              http_cache=(0, {'no_store': False}))
 def features_status(request):
-    """Report current feature flag values."""
+    """Returns a dictionary of feature flag keys and their current status"""
```
nickstenning: I'm not sure about this change to the docstring. A "dictionary" isn't a thing in JSON.

h/session.py
```diff
@@ -10,6 +11,7 @@ def model(request):
     session['csrf'] = request.session.get_csrf_token()
     session['userid'] = request.authenticated_userid
     session['groups'] = _current_groups(request)
+    session['features'] = features.features_status(request)
```
nickstenning: I'd rather we didn't use view functions as helper functions. Move `return {k: flag_enabled(request, k) for k in FEATURES.keys()}` into a helper function called something like `all` and then have both the view function in `h.features` and this call `features.all(request)`.

h/static/scripts/features.js
```diff
@@ -1,90 +1,34 @@
 /**
  * Feature flag client.
  *
- * This is a small utility which will periodically retrieve the application
```
nickstenning: Good job on removing outdated docs, but it would be nice to replace them with a one or two line summary of what this does actually now do :)

nickstenning:
Mostly looks good. A few small comments here and there.

h/test/session_test.py
```diff
@@ -26,3 +27,21 @@ def test_sorts_groups(User):
...
+
+
+@patch('h.session.features')
+def test_includes_features(features, fake_user):
+    feature_dict = {
+        'feature_one': True,
+        'feature_two': False,
+    }
+    features.all = Mock(return_value=feature_dict)
+    request = Mock(authenticated_user=fake_user)
```
nickstenning: I'm still going to merge this but I will just add that I'm not sure the indirection of a fixture was really needed for this `fake_user` arg. You could just as easily have done `fake_user = Mock(groups=[])` on the preceding line.

nickstenning:
LGTM. Thanks, Robert.

PR #2808, opened 2015-12-17
Remove Moment.js

robertknight:
Moment.js is only used in a single place, for converting ISO8601 date-time strings
from the annotation's 'updated' field to user-friendly
date strings (eg. 'Wednesday 17 Dec, 18:59') for the tooltips
for the last-updated link in the annotation card.

Replacing this with functions from the built-in Date object saves
167KB from the generated bundle size (see
moment/min/moment-with-locales.min.js) and also speeds up the digest
cycle. For 20 annotation cards in the /stream view, generating the
date string used to take 6-7ms with Moment. It takes <2ms with
the built in functions in a current build of Chrome.

This still shows up as the most expensive watch expression, so
we might want to add caching as well.

### DIFF
--- a/h/static/scripts/directive/annotation.js
+++ b/h/static/scripts/directive/annotation.js
@@ -1,6 +1,7 @@
 /* jshint node: true */
 'use strict';
 
+var dateUtil = require('../date-util');
 var events = require('../events');
 
 /** Return a domainModel tags array from the given vm tags array.
@@ -744,7 +745,7 @@ function AnnotationController(
       return '';
     }
     var date = new Date(domainModel.updated);
-    return date.toDateString() + ', ' + date.toLocaleTimeString();
+    return dateUtil.format(date);
   }
 
   vm.user = function() {
--- a/h/static/scripts/directive/annotation.js
+++ b/h/static/scripts/directive/annotation.js
@@ -739,6 +739,14 @@ function AnnotationController(
     return domainModel.updated;
   };
 
+  vm.updatedString = function () {
+    if (!domainModel.updated) {
+      return '';
+    }
+    var date = new Date(domainModel.updated);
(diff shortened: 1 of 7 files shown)


h/static/scripts/directive/annotation.js
```diff
@@ -739,6 +739,14 @@ function AnnotationController(
     return domainModel.updated;
   };
 
+  vm.updatedString = function () {
+    if (!domainModel.updated) {
+      return '';
+    }
+    var date = new Date(domainModel.updated);
+    return date.toDateString() + ', ' + date.toLocaleTimeString();
```
nickstenning: Should this not be `toLocaleDateString` too?

robertknight: `toDateString()` actually generates the most similar output to the previous result. I've replaced this now with `toLocaleDateString()` which provides customization of the output where supported (IE 11, Edge, Firefox, Chrome) and fall back to `toDateString()` elsewhere (IE 10, Safari).

A downside of not using Moment.js is that the output in this case is implementation dependent but I think it is the right trade-off vs. the page weight of Moment.

nickstenning:
I'm a bit confused. Why are we formatting the dates with `toLocaleString`'s options? Surely using the user's default locale to format the date is not only fine, but exactly what we want in this case.

robertknight:
> Surely using the user's default locale to format the date is not only fine,
> but exactly what we want in this case.

The default format that `toLocaleString()` produces in English is 'DD/MM/YY, HH:MM:SS' in Firefox/Chrome which is usable but I think the output with the custom options above (which includes the day and month name and omits the seconds) is more readable and thoughtful (eg. "Friday, 18 Dec 2015, 17:03")

PR #2809, opened 2015-12-17
Use Angular 1.3.x+ one-time bindings

robertknight:
Since the upgrade to Angular 1.4.x, we have a new facility available to us to reduce the number of watches created by expressions in templates. Prefixing an expression with `::` causes it to be evaluated only once and the watch is removed thereafter.

This PR uses '::' for a bunch of expressions which do not change during the lifetime of the app (eg. `isSidebar`) or an annotation card (eg. the annotation quote).

~~In combination with https://github.com/hypothesis/h/pull/2808 this reduces the length of a typical `$digest` cycle in the default view of the `/stream` from ~30ms to ~20ms and the number of watches from ~1690 to ~1530.~~

**Update** - I revised the PR after the discussion. It is now much more conservative and only cuts the number of watchers down from 1690 to 1649 in the same scenario, but that's still a small win and I think it is also a useful record of the approach.

Longer term we could do much better by taking advantage of the fact that the majority of expressions in an annotation card can only change when the annotation data itself is modified, which never happens for the majority of annotations. This PR is a quick and easy win though.

### DIFF
--- a/h/templates/client/annotation.html
+++ b/h/templates/client/annotation.html
@@ -9,7 +9,7 @@
       <span>
         <a class="annotation-user"
            target="_blank"
-           ng-href="{{vm.baseURI}}u/{{vm.user()}}"
+           ng-href="{{::vm.baseURI}}u/{{vm.user()}}"
            >{{vm.user() | persona}}</a>
       </span>
 
@@ -33,11 +33,11 @@
         <i class="h-icon-border-color" ng-show="vm.isHighlight() && !vm.editing()" title="This is a highlight. Click 'edit' to add a note or tag."></i>
         <span class="annotation-citation"
               ng-bind-html="vm.document() | documentTitle"
-              ng-if="!vm.isSidebar">
+              ng-if="::!vm.isSidebar">
         </span>
         <span class="annotation-citation-domain"
               ng-bind-html="vm.document() | documentDomain"
-              ng-if="!vm.isSidebar">
+              ng-if="::!vm.isSidebar">
         </span>
       </span>
     </span>
@@ -49,7 +49,7 @@
        target="_blank"
        title="{{vm.updated() | moment:'LLLL'}}"
        ng-if="!vm.editing() && vm.updated()"
-       ng-href="{{vm.baseURI}}a/{{vm.id()}}"
+       ng-href="{{::vm.baseURI}}a/{{vm.id()}}"
        >{{vm.timestamp}}</a>
   </header>
 
--- a/h/templates/app.html.jinja2
+++ b/h/templates/app.html.jinja2
@@ -25,7 +25,7 @@
     on-logout="logout()"
     account-dialog="accountDialog"
     share-dialog="shareDialog"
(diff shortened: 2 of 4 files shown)


h/templates/client/annotation.html
```diff
@@ -96,8 +96,8 @@
   <div class="annotation-body tags tags-read-only"
        ng-if="vm.form.tags.length && !vm.editing()">
     <ul class="tag-list">
-      <li class="tag-item" ng-repeat="tag in vm.form.tags">
-        <a href="/stream?q=tag:'{{tag.text|urlencode}}'" target="_blank">{{tag.text}}</a>
+      <li class="tag-item" ng-repeat="tag in ::vm.form.tags">
+        <a href="/stream?q=tag:'{{::tag.text|urlencode}}'" target="_blank">{{::tag.text}}</a>
```
seanh: This breaks live-update of tags, when the annotation is edited by another client

h/templates/client/annotation.html
```diff
@@ -1,16 +1,16 @@
-<header class="annotation-header" ng-if="!vm.user()">
+<header class="annotation-header" ng-if="::!vm.user()">
```
seanh: This does change when you create an annotation or highlight while logged out, then log in. That feature seems to still work correctly on this branch though which surprises me a little, but I guess the AnnotationController instance is destroyed and a new one created for it from the drafts service.

h/templates/client/annotation.html
```diff
@@ -110,7 +110,7 @@
               class="dropdown-menu-btn"><i class="h-icon-check btn-icon"></i> Delete</button>
       <publish-annotation-btn
         class="publish-annotation-btn"
-        group="vm.group()"
+        group="::vm.group()"
```
seanh: I think this one breaks the UI when creating a new annotation and changing between groups without saving it

seanh:
On a quick look, I think `vm.form` and `vm.group()` are not safe to do this to.

What are all the things that a user (either the current user, or someone simultaneously using a different client) can edit?

I think it's group, permissions/isPrivate, text and tags.

So none of those are ever safe to do this to, I think.

For example you might think that group can only ever change for new annotations not saved to the server yet, which I think is true so far, but I wouldn't be surprised if we end up enabling moving saved annotations between groups at some point.

Text, tags and perms can always change because of live update unfortunately.

h/templates/client/annotation.html
robertknight: Good catch. I'd not considered live updates by other clients. I'll revise this.

nickstenning:
Overall I'm not entirely convinced that any of this except `isSidebar` is safe. As @seanh points out, these annotations can be updated in their entirety by the realtime update feature.

robertknight:
I'll change this so it just makes `isSidebar` and `baseURI` constant for the moment. That unfortunately loses most (but not all) of the performance wins but as seanh pointed out, my assumptions about what could change were wrong.

PR #2814, opened 2015-12-18
Enable video embeds to be shown fullscreen

robertknight:
The 'allowfullscreen' attribute must be set on both
the embed's <iframe> and the parent sidebar <iframe>
when the embed's main script is loaded to enable
fullscreen support.


seanh:
Cheers

PR #2816, opened 2015-12-18
Fix video embeds in Firefox < 45

seanh:
innerText is non-standard and not supported in Firefox < 45.

Use textContent instead.


seanh:
@robertknight

PR #2817, opened 2015-12-18
Clear selection before creating new annotation

seanh:
Clear any selection that exists in the sidebar before creating a new
annotation. Otherwise the new annotation with its form open for the user
to type in won't be visible because it's not part of the selection.

https://trello.com/c/EqRZwdZC/210-clear-selected-annotations-when-a-user-initiates-a-new-annotation

### DIFF
--- a/h/static/scripts/test/app-controller-test.coffee
+++ b/h/static/scripts/test/app-controller-test.coffee
@@ -168,6 +168,30 @@ describe 'AppController', ->
     $scope.$broadcast(events.USER_CHANGED, {initialLoad: false})
     assert.calledOnce(fakeRoute.reload)
 
+  describe 'on "beforeAnnotationCreated"', ->
+
+    ###*
+    #  It should clear any selection that exists in the sidebar before
+    #  creating a new annotation. Otherwise the new annotation with its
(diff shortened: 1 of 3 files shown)


h/static/scripts/test/app-controller-test.coffee
```diff
@@ -168,6 +168,28 @@ describe 'AppController', ->
...
+  describe 'on "beforeAnnotationCreated"', ->
+    it 'calls $scope.clearSelection()', ->
+      createController()
+      sandbox.spy($scope, 'clearSelection')
+
+      $rootScope.$emit('beforeAnnotationCreated', {})
+
+      assert(
+        $scope.clearSelection.called
+        'It should clear any selection that exists in the sidebar before '
```
nickstenning: Could you move this text into a comment and do `assert.called($scope.clearSelection)`. That'll be easier to read if the test fails and the detail will still be available in the test function for debugging.

seanh: Done

robertknight:
This is a great and long overdue usability improvement.

judell:
What @robertknight said. Thanks!

PR #2821, opened 2015-12-21
Update the displayed annotation timestamp immediately after creation/…

### LINKED ISSUE #2819: timestamp missing from newly created annotation
image

The missing timestamp appears after a page refresh. Minor fallout from https://github.com/hypothesis/h/pull/2808?

robertknight:
…update

The call to updateTimestamp() for new annotations is a no-op because
the annotation has no timestamp yet.

After saving an annotation, we first need to update the timestamp
and also trigger the periodic refresh of the timestamp.

 * Remove unnecessary 'repeat' parameter to updateTimestamp()
   since the function was always called with this set to true

 * Keep track of whether a repeat has already been scheduled
   and avoid kicking of multiple timestamp update cycles

Fixes #2819

### DIFF
--- a/h/static/scripts/directive/annotation.js
+++ b/h/static/scripts/directive/annotation.js
@@ -240,6 +240,7 @@ function AnnotationController(
   var vm = this;
   var domainModel;
   var newlyCreatedByHighlightButton;
+  var timestampUpdateScheduled = false;
 
   /**
     * Initialize this AnnotationController instance.
@@ -281,6 +282,10 @@ function AnnotationController(
     /** A "fuzzy string" representation of the annotation's last updated time.
      */
     vm.timestamp = null;
+
+    // keeps track of whether the timer that periodically refreshes
+    // vm.timestamp is active
+    timestampUpdateScheduled = false;
 
     /** The domain model, contains the currently saved version of the
       * annotation from the server (or in the case of new annotations that
@@ -330,7 +335,7 @@ function AnnotationController(
     // log in.
     saveNewHighlight();
 
-    updateTimestamp(true);
+    updateTimestamp();
     updateViewModel(domainModel, vm, permissions);
 
     // If this annotation is not a highlight and if it's new (has just been
@@ -406,9 +411,7 @@ function AnnotationController(
 
   // We use `var foo = function() {...}` here instead of `function foo() {...}`
   // because updateTimestamp gets redefined later on.
-  var updateTimestamp = function(repeat) {
-    repeat = repeat || false;
-
+  var updateTimestamp = function() {
     // New (not yet saved to the server) annotations don't have any .updated
     // yet, so we can't update their timestamp.
     if (!domainModel.updated) {
@@ -417,7 +420,7 @@ function AnnotationController(
 
     vm.timestamp = time.toFuzzyString(domainModel.updated);
 
-    if (!repeat) {
+    if (timestampUpdateScheduled) {
       return;
     }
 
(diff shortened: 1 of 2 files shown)


h/static/scripts/directive/annotation.js
```diff
@@ -282,6 +282,8 @@ function AnnotationController(
      */
     vm.timestamp = null;
 
+    vm.timestampUpdateScheduled = false;
```
seanh: It looks like this doesn't need to be on `vm`? (It's not used by the templates)

seanh:
Quick test, the timestamp doesn't seem to get updated when I edit and save an annotation?

Also an obscure one: Create a highlight while signed out, sign in, highlight is automatically saved but shows no timestamp.

robertknight:
> Quick test, the timestamp doesn't seem to get updated when I edit and save an annotation?

Indeed. That looks like a different, existing bug. When changes are saved to an annotation, a copy of the domain model is made and $update() is called on the copy. The changes from the updated annotation returned by the server are saved to the _copy_ and `domainModel.updated` is not updated on the original annotation. This also means that any other properties that are modified or set on the modified annotation are also not saved to the local instance.

h/static/scripts/directive/annotation.js
robertknight: I've moved this to a variable in the enclosing scope.

robertknight:
There was a recent change in 95921897aaed64177241d3c0ca4816bab21cd2b5 which removed logic that copied properties from the model instance which was saved to the server (`updatedModel`) back to the local instance (`domainModel`)

h/static/scripts/directive/annotation.js
```diff
@@ -282,6 +283,10 @@ function AnnotationController(
      */
     vm.timestamp = null;
 
+    // keeps track of whether the timer that periodically refreshes
+    // vm.timestamp is active
+    timestampUpdateScheduled = false;
```
seanh: Nitpicky, but in this file I've gone for jsdoc '/**' comments for the docstrings of methods/function/variables, and I've used capital letters and full stops!

h/static/scripts/directive/annotation.js
```diff
@@ -240,6 +240,7 @@ function AnnotationController(
   var vm = this;
   var domainModel;
   var newlyCreatedByHighlightButton;
+  var timestampUpdateScheduled = false;
```
seanh: Did you mean to initialize this to `false` in two places?

seanh:
> Also an obscure one: Create a highlight while signed out, sign in, highlight is automatically saved but shows no timestamp.

This issue can be fixed by adding a call to `updateTimestamp()` in the `saveNewHighlight()` function, after the `$rootScope.$emit('annotationCreated', domainModel)`. (It's a remaining weakness of `AnnotationController` that when you create a new highlight while logged out then log in, there's a completely separate code path for saving those, it doesn't go through `vm.save()`).

h/static/scripts/directive/annotation.js
```diff
@@ -436,6 +442,10 @@ function AnnotationController(
     */
   function view() {
     vm.action = 'view';
+
+    // trigger an immediate update of the timestamp when switching
+    // back to the view mode just after creating or updating an annotation
+    updateTimestamp();
```
seanh: Should we call this in `updateViewModel()`, rather than here? `updateViewModel()` is called both after saving a new annotation and after updating an annotation. `updateViewModel()` is also called on `init()`, so you could remove the call to `updateTimestamp()` in `init()` as well. The idea for `updateViewModel()` was that it would become the _only_ function that updates any of the values on `vm`, and it would be called at just the right times.

robertknight:
I'm going to close this and open a new PR which handles the logic for updating the timestamp in a single place, which fixes several issues together but turned out to necessitate some quite different refactoring.

PR #2824, opened 2015-12-21
Add a dismissable tutorial in the sidebar

seanh:
https://trello.com/c/PfK5vbcM/212-sidebar-tutorial

### DIFF
--- a/h/accounts/models.py
+++ b/h/accounts/models.py
@@ -79,6 +79,11 @@ class User(Base):
                       default=False,
                       server_default=sa.sql.expression.false())
 
+    sidebar_tutorial_dismissed = sa.Column(sa.Boolean,
+                                           default=False,
+                                           server_default=(
+                                                sa.sql.expression.false()))
+
     def _get_username(self):
         return self._username
 
--- a/h/accounts/test/views_test.py
+++ b/h/accounts/test/views_test.py
@@ -223,7 +223,7 @@ def test_logout_response_has_forget_headers(authn_policy):
     assert result.headers['x-erase-fingerprints'] == 'on the hob'
 
 
-@pytest.mark.usefixtures('routes_mapper')
+@pytest.mark.usefixtures('routes_mapper', 'session')
 def test_login_ajax_returns_status_okay_when_validation_succeeds():
     request = DummyRequest(json_body={}, auth_domain='hypothes.is')
     controller = AjaxAuthController(request)
@@ -261,7 +261,7 @@ def test_login_ajax_raises_JSONError_on_non_object_json():
                                       'top-level object')
 
 
-@pytest.mark.usefixtures('routes_mapper')
+@pytest.mark.usefixtures('routes_mapper', 'session')
 @mock.patch('h.accounts.schemas.check_csrf_token')
 def test_login_ajax_converts_non_string_usernames_to_strings(_):
     for input_, expected_output in ((None, ''), (23, '23'), (True, 'True')):
@@ -278,7 +278,7 @@ def test_login_ajax_converts_non_string_usernames_to_strings(_):
             [('username', expected_output), ('password', 'pass')])
 
 
-@pytest.mark.usefixtures('routes_mapper')
+@pytest.mark.usefixtures('routes_mapper', 'session')
 @mock.patch('h.accounts.schemas.check_csrf_token')
 def test_login_ajax_converts_non_string_passwords_to_strings(_):
     for input_, expected_output in ((None, ''), (23, '23'), (True, 'True')):
@@ -306,7 +306,7 @@ def test_login_ajax_raises_ValidationFailure_on_ValidationFailure():
     assert exc_info.value.error.asdict() == {'password': 'too short'}
 
 
-@pytest.mark.usefixtures('routes_mapper')
+@pytest.mark.usefixtures('routes_mapper', 'session')
 def test_logout_ajax_returns_status_okay():
     request = DummyRequest()
 
@@ -1001,6 +1001,14 @@ def pop_flash(request):
     func = patcher.start()
     request.addfinalizer(patcher.stop)
     return func
+
+
+@pytest.fixture
+def session(request):
+    patcher = patch('h.accounts.views.session', autospec=True)
+    session = patcher.start()
+    request.addfinalizer(patcher.stop)
+    return session
 
 
 @pytest.fixture
--- a/h/accounts/views.py
+++ b/h/accounts/views.py
@@ -587,6 +587,16 @@ def reset_password_link(request, reset_code):
     return request.route_url('reset_password_with_code', code=reset_code)
 
 
+@view_config(route_name='dismiss_sidebar_tutorial', request_method='POST',
+             renderer='json')
+def dismiss_sidebar_tutorial(request):
+    if request.authenticated_userid is None:
+        raise accounts.JSONError()
+    else:
+        request.authenticated_user.sidebar_tutorial_dismissed = True
+        return ajax_payload(request, {'status': 'okay'})
+
+
 def includeme(config):
     config.add_route('login', '/login')
     config.add_route('logout', '/logout')
@@ -597,4 +607,6 @@ def includeme(config):
     config.add_route('reset_password_with_code', '/reset_password/{code}')
     config.add_route('profile', '/profile')
     config.add_route('profile_notifications', '/profile/notifications')
+    config.add_route('dismiss_sidebar_tutorial',
+                     '/app/dismiss_sidebar_tutorial')
     config.scan(__name__)
--- a/h/features.py
+++ b/h/features.py
@@ -15,9 +15,10 @@ log = logging.getLogger(__name__)
 
 FEATURES = {
     'claim': "Enable 'claim your username' web views?",
-    'new_homepage': "Show the new homepage design?",
-    'truncate_annotations': "Truncate long quotes and bodies in annotations?",
     'embed_media': "Replace YouTube, Vimeo etc links in annotations with embeds",
+    'new_homepage': "Show the new homepage design?",
+    'sidebar_tutorial': "Show a tutorial to new users in the sidebar",
+    'truncate_annotations': "Truncate long quotes and bodies in annotations?",
 }
 
 
--- a/h/migrations/versions/1ef80156ee4_add_sidebar_tutorial_dismissed_column_.py
+++ b/h/migrations/versions/1ef80156ee4_add_sidebar_tutorial_dismissed_column_.py
@@ -1,0 +1,23 @@
+"""Add the sidebar_tutorial_dismissed column to user table.
+
+Revision ID: 1ef80156ee4
+Revises: 43645baa68b2
+Create Date: 2015-12-21 18:49:15.688177
+
+"""
+
+# revision identifiers, used by Alembic.
+revision = '1ef80156ee4'
+down_revision = '43e7c4ed2fd7'
+
+from alembic import op
+import sqlalchemy as sa
+
+
+def upgrade():
+    op.add_column(
+        'user',
+        sa.Column('sidebar_tutorial_dismissed', sa.Boolean(), nullable=True))
+
+def downgrade():
+    op.drop_column('user', 'sidebar_tutorial_dismissed')
--- a/h/session.py
+++ b/h/session.py
@@ -12,6 +12,11 @@ def model(request):
     session['userid'] = request.authenticated_userid
     session['groups'] = _current_groups(request)
     session['features'] = features.all(request)
+    session['preferences'] = {}
+    if request.feature('sidebar_tutorial'):
+        user = request.authenticated_user
+        if user and not user.sidebar_tutorial_dismissed:
+            session['preferences']['show_sidebar_tutorial'] = True
     return session
 
 
--- a/h/static/scripts/directive/sidebar-tutorial.js
+++ b/h/static/scripts/directive/sidebar-tutorial.js
@@ -1,0 +1,32 @@
+'use strict';
+
+// @ngInject
+function controller(session) {
+  /*jshint validthis:true */
+  var vm = this;
+
+  vm.showSidebarTutorial = function () {
+    return session.state.preferences.show_sidebar_tutorial;
+  };
+
+  vm.dismiss = function () {
+    session.dismiss_sidebar_tutorial();
+  };
+}
+
+/**
+ * @ngdoc directive
+ * @name sidebarTutorial
+ * @description Displays a short tutorial in the sidebar.
+ */
+// @ngInject
+module.exports = function () {
+  return {
+    bindToController: true,
+    controller: controller,
+    controllerAs: 'vm',
+    restrict: 'E',
+    scope: {},
+    templateUrl: 'sidebar_tutorial.html'
+  };
+};
--- a/h/static/styles/sidebar-tutorial.scss
+++ b/h/static/styles/sidebar-tutorial.scss
@@ -1,0 +1,50 @@
+.sidebar-tutorial__header {
+  color: $color-cardinal;
+  font-weight: $title-font-weight;
+}
+
+/* We want an ordered list in which the numbers are aligned with the text
+   above (not indented or dedented), and wrapped lines in list items are
+   aligned with the first character of the list item (not the number, i.e
+   they're indented):
+
+       This is a list:
+
+       1. This is a list item.
+       2. This is a long list item
+          that wraps.
+       3. This is another list item.
+
+   What's more, we want the numbers to be a different color to the text of the
+   list items, which means we need to use ::before content for them.
+
+   This appears to be much harder than you'd think.
+
+   The code below comes from this Stack Overflow answer:
+   http://stackoverflow.com/questions/10428720/how-to-keep-indent-for-second-line-in-ordered-lists-via-css/17515230#17515230
+*/
+.sidebar-tutorial__list {
+  counter-reset: sidebar-tutorial__list;
+  display: table;
+  padding: 0;
+}
+.sidebar-tutorial__list-item {
+  list-style: none;
+  counter-increment: sidebar-tutorial__list;
+  display: table-row;
+}
+.sidebar-tutorial__list-item::before {
+  content: counter(sidebar-tutorial__list) ".";
+  display: table-cell; /* aha! */
+  text-align: right;
+  padding-right: .3em;
+  color: $gray-light;
+}
+
+.sidebar-tutorial__list-item-content {
+  margin-top: 1em; /* Put vertical space in-between list items, equal to the
+                      space between normal paragraphs.
+                      Note: This also puts the same amount of space above the
+                      first list item (i.e. between the list and whatever's
+                      above it). */
+}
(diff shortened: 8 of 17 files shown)


seanh:
I haven't added any tests for this - each of the individual units is so simple that it doesn't seem to require testing. Most of the complexity here is how several units fit together to make the feature work, and tjat isn't amenable to unit testing.

seanh:
I have to say, this feature doesn't make sense as long as the <https://hypothes.is/welcome> page still exists in its current form. When the user installs th

seanh:
Screencast:

Here I'm using Chrome and have the Chrome extension installed. I browse to the <https://hypothes.is/docs/help> page (which, incidentally, is the same as the <https://hypothes.is/welcome> page, dunno why) and the sidebar opens. The sidebar tutorial is not shown, because I'm not logged in. It is never shown when you're not logged in. Then I log in, and the tutorial appears at the top of the sidebar. It'll appear like this for new user accounts, the first time you log in to them. You can't see my mouse cursor in this screencast but I click the x / close button in the top-right of the tutorial card and it just disappears. I reload the page and the tutorial does not come back - it'll never come back for this user and there's no way to get it back.

untitled screencast 9

seanh:
screenshot from 2015-12-21 19-27-45

h/migrations/versions/1ef80156ee4_add_sidebar_tutorial_dismissed_column_.py
```diff
@@ -0,0 +1,28 @@
...
+import sqlalchemy as sa
+
+
+def upgrade():
+    with op.batch_alter_table('user') as batch_op:
+        batch_op.add_column(sa.Column('sidebar_tutorial_dismissed',
+                                      sa.Boolean,
+                                      nullable=False,
+                                      default=False,
+                                      server_default=sa.sql.expression.true()))
```
seanh: The `server_default` here is `true()` so that on a db migration, the tutorial will not be shown to all existing users. But the app default is `False` so it will be shown to new users who register thereafter.

h/static/scripts/directive/sidebar-tutorial.js
```diff
@@ -0,0 +1,32 @@
...
+function controller($http, session) {
+  var vm = this;
+
+  vm.showSidebarTutorial = function () {
+    return session.state.show_sidebar_tutorial;
+  };
+
+  vm.dismiss = function () {
+    $http({method: 'POST', url: '/sidebar_tutorial/dismiss'});
+    session.state.show_sidebar_tutorial = false;
```
robertknight: I suggest that we shouldn't modify the session state directly here but instead provide methods on the session service to do that. That keeps all modification of the session data in one place.

More generally, what we are doing here is effectively an optimistic update of the session state. It seems quite possible that we might want to do that for other things in future. That will be easier if all modifications happen in one place.

robertknight:
Since this is a feature seen only on first use of the app, it is something that we (as staff) or regular existing users (who know how the app usually works) are not likely to see often and wouldn't notice if it broke. Consequently I think having some basic level of integration testing to verify that "the tutorial sidebar appears when a new user starts the app" (you could split that into the client and server-side parts, we don't need to cover both with one test), is valuable.

Another way we can approach this would be to track metrics about user retention/engagement and get an alert if that drops.

Neither of these block this PR, but perhaps put a card in the infrastructure backlog and we can talk about it in the new year.

seanh:
OK, I think we should hold back on merging this anyway, at least until we have a new design or go-ahead for the existing design and have implemented that design in CSS, but probably also until we've considered this a bit more carefully from the user's point of view imo (showing the sidebar tutorial right next to the welcome page tutorial).

I'll leave it here for now but will come back to it at some point and add tests

h/static/scripts/directive/sidebar-tutorial.js
seanh: I've added `session.dismissSidebarTutorial()`

seanh:
I've added a card to discuss how we might want to write integration tests for things like this: https://trello.com/c/Xu8hXqFj/52-basic-integration-testing-for-backend-and-frontend

seanh:
@nickstenning Would be good to get your comment on this comment https://github.com/hypothesis/h/pull/2824#discussion_r48181242

seanh:
I've put this behind a feature flag because of the CSS, so it can be merged now.

@conordelahunty @robertknight What'a the deal with the CSS for this one? Shall I go ahead and write new CSS to make it look like this:



Or is the design going to change or are there CSS rules coming soon that I could reuse?

robertknight:
@seanh - Use colors and font sizes from `variables.scss`. The rest of the styling for this component should go in its own `.scss` file - similar to the other directives that have recently been added.

h/migrations/versions/1ef80156ee4_add_sidebar_tutorial_dismissed_column_.py
nickstenning: That's not how `server_default` works. `server_default` sets the schema for the table to include a default, so this migration will actually create a table with a schema that doesn't match the model.

But there is another issue with this, which is that adding a column with a `DEFAULT` will lock the table and write every row. So:

1. Get rid of `op.batch_alter_table` -- it's not necessary for Postgres and was only a workaround for the fact that SQLite doesn't support `ALTER TABLE ... ADD COLUMN`.
2. Add the column without a `default` (which is a no-op here) or a `server_default`.
3. Alter the column to set the (correct) `server_default`: `op.alter_column('user', 'sidebar_tutorial_dismissed', server_default=sa.sql.expression.false())`.
4. Add a second migration to update the data for preexisting users.

h/resources.py
```diff
@@ -44,6 +45,7 @@ class Root(Resource):
         (Allow, 'group:__admin__', 'admin_badge'),
         (Allow, 'group:__admin__', 'admin_groups'),
         (Allow, 'group:__staff__', 'admin_groups'),
+        (Allow, security.Authenticated, 'authenticated'),
```
nickstenning: This is a bit of an abuse of the ACL system. Just raise a 404 (or whatever) from within the view if `request.authenticated_user` is `None`.

h/static/scripts/directive/sidebar-tutorial.js
```diff
@@ -0,0 +1,34 @@
+'use strict';
+
+// @ngInject
+function controller($http, session, features) {
+  /*jshint validthis:true */
+  var vm = this;
+
+  vm.showSidebarTutorial = function () {
+    return (features.flagEnabled('sidebar_tutorial') &&
```
nickstenning: I don't see there's any reason the client-side component needs to test the feature. Just don't add the field to the session server-side if the flag isn't enabled.

h/accounts/test/views_test.py
```diff
@@ -259,7 +260,7 @@ def test_login_ajax_converts_non_string_usernames_to_strings(_):
     for input_, expected_output in ((None, ''), (23, '23'), (True, 'True')):
         request = DummyRequest(
             json_body={'username': input_, 'password': 'pass'},
-            auth_domain='hypothes.is')
+            auth_domain='hypothes.is', authenticated_user=None)
```
nickstenning: It's probably better to add a fixture which mocks `session.model` and use that consistently across all the `AjaxAuthController` tests. At the moment the implementation details of `session.model` are leaking into tests which are supposed to be testing entirely unrelated things.

h/features.py
```diff
@@ -18,6 +18,7 @@
     'new_homepage': "Show the new homepage design?",
     'truncate_annotations': "Truncate long quotes and bodies in annotations?",
     'embed_media': "Replace YouTube, Vimeo etc links in annotations with embeds",
+    'sidebar_tutorial': "Show a tutorial to new users in the sidebar",
```
nickstenning: Could we please keep stuff like this alphabetised?

h/static/scripts/directive/sidebar-tutorial.js
```diff
@@ -0,0 +1,34 @@
...
+            session.state.show_sidebar_tutorial);
+  };
+
+  vm.dismiss = function () {
+    $http({method: 'POST', url: '/sidebar_tutorial/dismiss'});
```
nickstenning: I'm not entirely sure about this. The session is an `ngResource` already, so I think it would make more sense to add a method on the session which will result in a request to the server to modify the server state.

I'd rather see this implemented as an `action` on the `session` resource than as an HTTP call hiding in a directive.

h/migrations/versions/1ef80156ee4_add_sidebar_tutorial_dismissed_column_.py
seanh: Done

h/static/scripts/directive/sidebar-tutorial.js
seanh: As per Nick's suggestion this has now been changed again, to use an ngResource action

h/resources.py
seanh: Done

h/static/scripts/directive/sidebar-tutorial.js
seanh: Done

h/accounts/test/views_test.py
seanh: Done

h/features.py
seanh: Done

h/static/scripts/directive/sidebar-tutorial.js
seanh: Done

seanh:
@robertknight @nickstenning All comments responded to. It would be nice to merge this now so that other prs don't stomp on it, I can do the CSS separately (it's behind a feature flag)

h/accounts/models.py
```diff
@@ -78,6 +78,9 @@ class User(Base):
                       default=False,
                       server_default=sa.sql.expression.false())
 
+    sidebar_tutorial_dismissed = sa.Column(
+        sa.Boolean, default=False, server_default=sa.sql.expression.false())
```
nickstenning: This isn't going to stop me merging, but the formatting of this attribute is completely different from all the ones above it. Consistency in formatting in cases like this can really help with readability.

h/accounts/test/views_test.py
```diff
@@ -306,13 +306,9 @@ def test_login_ajax_raises_ValidationFailure_on_ValidationFailure():
...
 
-@pytest.mark.usefixtures('routes_mapper')
+@pytest.mark.usefixtures('routes_mapper', 'session')
 def test_logout_ajax_returns_status_okay():
-    request = DummyRequest()
-
-    result = AjaxAuthController(request).logout()
-
-    assert result['status'] == 'okay'
+    AjaxAuthController(DummyRequest()).logout()['status'] == 'okay'
```
nickstenning: You've managed to remove the `assert` from this test.

nickstenning: Also the three line layout is deliberate. The three sections are the canonical [Arrange, Act, Assert](http://c2.com/cgi/wiki?ArrangeActAssert), so removing them isn't necessarily helpful.

h/accounts/views.py
```diff
@@ -587,6 +587,16 @@ def reset_password_link(request, reset_code):
     return request.route_url('reset_password_with_code', code=reset_code)
 
 
+@view_config(route_name='dismiss_sidebar_tutorial', request_method='POST',
+             renderer='json')
+def dismiss_sidebar_tutorial(request):
+    if request.authenticated_userid is None:
+        raise accounts.JSONError()
+    else:
```
nickstenning: Minor point of style (which isn't going to stop me merging) but:

```
if condition:
    raise Something()
else:
    something_else()
```

is usually more readable without the `else:`, i.e.

```
if condition:
    raise Something()
something_else()
```

seanh: I put the else in because I thought it was more readable that way

h/accounts/test/views_test.py
```diff
@@ -306,13 +306,9 @@ def test_login_ajax_raises_ValidationFailure_on_ValidationFailure():
...
 
-@pytest.mark.usefixtures('routes_mapper')
+@pytest.mark.usefixtures('routes_mapper', 'session')
 def test_logout_ajax_returns_status_okay():
-    request = DummyRequest()
-
-    result = AjaxAuthController(request).logout()
-
-    assert result['status'] == 'okay'
+    assert AjaxAuthController(DummyRequest()).logout()['status'] == 'okay'
```
nickstenning: This is a no-op change. Can you remove it from this PR?

h/accounts/views.py
nickstenning: Well I'm not going to bikeshed with you about "readability." But in general, IMO, less indentation is usually better.

h/accounts/test/views_test.py
seanh: Done

h/session.py
```diff
@@ -12,6 +12,10 @@ def model(request):
     session['userid'] = request.authenticated_userid
     session['groups'] = _current_groups(request)
     session['features'] = features.all(request)
+    user = request.authenticated_user
+    if features.flag_enabled(request, 'sidebar_tutorial'):
```
nickstenning: `request.feature('sidebar_tutorial')` would be more conventional.

nickstenning:
As noted in Slack this now LGTM, but the migrations will need to be reordered or this will need to wait until the migrations in #2803 are deployed to production.

> so each of password reset (A) and sidebar tutorial (B) have three migrations: currently they're ordered A1 A2 A3 B1 B2 B3
> ​we can merge this, but only if they're ordered A1 B1 A2 B2 A3 B3

I also think we should consider whether `/sidebar_tutorial/dismiss` is a sensible URL endpoint for this. We probably want to put something under `/app` rather than adding an entirely new URL hierarchy.

h/session.py
```diff
@@ -12,6 +12,10 @@ def model(request):
...
+        if user and not user.sidebar_tutorial_dismissed:
+            session['show_sidebar_tutorial'] = True
```
robertknight: I think we should probably put this in a suitable namespace under settings rather than at the root of the session object. Perhaps "prefs" or "settings"?

h/accounts/views.py
robertknight: I'm going to agree with nick here on style. In client code the recommended style is:

```js
if (condition) {
  // early exit
   throw SomeError() // or 'return x';
}
// happy path
```

It avoids unnecessary indentation.

h/static/scripts/directive/sidebar-tutorial.js
```diff
@@ -0,0 +1,32 @@
+'use strict';
+
+// @ngInject
+function controller($http, session) {
```
robertknight: The `$http` dependency here is unused.

h/test/session_test.py
```diff
@@ -1,47 +1,90 @@
...
+@mock.patch('h.session.features', autospec=True)
+def test_model_includes_features(features, fake_user):
     feature_dict = {
         'feature_one': True,
         'feature_two': False,
     }
-    features.all = Mock(return_value=feature_dict)
-    request = Mock(authenticated_user=fake_user)
-    assert model(request)['features'] == feature_dict
+    features.all = mock.Mock(return_value=feature_dict)
```
robertknight: I find this very verbose. We essentially have a matrix of conditions and expected outputs:

```
tutorial flag enabled? | tutorial dismissed? | expect show_tutorial?
N | N | N
N | Y | N
Y | N | Y
Y | Y | N
```
Is there a facility that we could use to define data driven tests with a truth table like the above, do we have any existing instances of this?

robertknight: I should add that I wouldn't block the merge on this

nickstenning: > Is there a facility that we could use to define data driven tests with a truth table like the above, do we have any existing instances of this?

[`@pytest.mark.parametrize`](https://pytest.org/latest/parametrize.html). [For example](https://github.com/hypothesis/h/blob/d83349155dd622825d4284153c9284cb2070ecb8/h/api/test/uri_test.py#L10).

h/session.py
seanh: Done

h/static/scripts/directive/sidebar-tutorial.js
seanh: Done

h/test/session_test.py
seanh: Done

h/session.py
seanh: Done

h/accounts/models.py
seanh: Done. Fwiw I originally fell back to this style rather than introduce a `\` or `(...)`'s because if I follow the style used above this line then `server_default=sa.sql.expression.false()` doesn't fit on one line.

seanh:
> I also think we should consider whether /sidebar_tutorial/dismiss is a sensible URL endpoint for this. We probably want to put something under /app rather than adding an entirely new URL hierarchy.

I'm not really sure what to do about this. I've changed it to `/app/dismiss_sidebar_tutorial`:

* Putting it under `/app/` makes sense, it's an app thing
* In general I guess `/` should denote navigation or structure in URLs. For example we have `/annotation` and under that we have `/annotation/<id>` etc. In this case there isn't really any structure, I could do `/sidebar/dismiss_tutorial` or `/sidebar_tutorial/dismiss` etc but it's artificial, there aren't any other URLs under those `/`s
* I left the long URL with `_`'s in it because I prefer explicit over some terse and cryptic thing and because in existing URLs we've used `_` to separate words

seanh:
screenshot from 2016-01-08 19-12-42

seanh:
Ok, CSS done, and I think I've responded to all the comments, and I've rebased it on master and re-tested

seanh:
This requires a look from @conordelahunty before enabling the feature flag for everyone, but can be merged now

conordelahunty:
@seanh Thanks Sean, I think this is fine for now.

h/static/styles/sidebar-tutorial.scss
```diff
@@ -0,0 +1,50 @@
...
+
+   What's more, we want the numbers to be a different color to the text of the
+   list items, which means we need to use ::before content for them.
+
+   This appears to be much harder than you'd think.
+
+   The code below comes from this Stack Overflow answer:
+   http://stackoverflow.com/questions/10428720/how-to-keep-indent-for-second-line-in-ordered-lists-via-css/17515230#17515230
+*/
+.sidebar-tutorial__list {
```
robertknight: I must admit, looking again this is cleaner that I thought on first glance. :+1:  for the clear explanation of the desired result, this is really useful.

seanh: The reason I ended up putting this back on, btw, was that the two-liner solution we were discussing does not work when you _also_ need the numbers to be a different colour (as far as I can tell). You're talking to someone with limited CSS knowledge but I think this may be the only way to get all 3 requirements at once.

robertknight: The obvious approach that you'd hope you could take would be to specifically style the list marker. There _is_ a '::marker' pseudo-element for this but it isn't supported by any browsers AFAIK.

PR #2825, opened 2015-12-22
Update annotation timestamp when creating and editing annotations

robertknight:
This is a revised approach to https://github.com/hypothesis/h/pull/2821 after discovery of two related bugs and following the comments in that PR.

This PR fixes three recent issues:

 * The timestamp for an annotation card did not appear after saving a new annotation
 * Timestamps did not appear on cards for newly added highlights
 * Timestamps did not update when editing an existing annotation

The root cause of the issue is that the function which updates the timestamp and triggers a periodic refresh depending on its age was not invoked in these three cases.

This PR fixes the issue by:

 * Extracting out the logic for managing the decaying interval (ie. an interval whose refresh rate depends on the age of an input date) into a separate function in `time.coffee`
 * Triggering a refresh of the timestamp and periodic timer in `updateViewModel()` which is the central function that handles updates to the view model when the domain model changes
 * Triggering an update of the view model when a new highlight is successfully saved to the server

### DIFF
--- a/h/static/scripts/directive/annotation.js
+++ b/h/static/scripts/directive/annotation.js
@@ -349,7 +349,7 @@ function AnnotationController(
     // log in.
     saveNewHighlight();
 
-    updateViewModel($scope, time, domainModel, vm, permissions);
+    updateView(domainModel);
 
     // If this annotation is not a highlight and if it's new (has just been
     // created by the annotate button) or it has edits not yet saved to the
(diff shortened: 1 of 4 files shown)


robertknight:
This would be a good time to translate `time(-test)?.coffee` to JS but that can be done as part of a separate PR.

h/static/scripts/directive/annotation.js
```diff
@@ -343,14 +361,20 @@ function AnnotationController(
     }
   }
 
+  function updateView(domainModel) {
+    updateViewModel($scope, time, domainModel, vm, permissions);
```
seanh: Ok, but everything in this file is carefully alphabetized!

PR #2826, opened 2015-12-22
Improve stream / sidebar date formatting performance

### LINKED ISSUE #2820: Perf regression when updating annotation last-displayed timestamps
Moment.js was recently removed in favor of using the browser's `Date` APIs in https://github.com/hypothesis/h/pull/2808 . The initial version of this was much faster than Moment. However, the revised version which actually got merged turns out to be much ~3x slower in Chrome than the Moment formatting - accounting for ~20ms of digest time (vs ~7ms for the old Moment implementation and ~2ms for the initial version of that PR) on the default `/stream` view.

A quick scan of JS profiling suggests that behind the scenes the browser is creating a new formatting object for each call to `toLocaleDateString` which is very expensive. Couple of things we can do here: 1) Cache the timestamp string, 2) Create a date formatter once and re-use it.

robertknight:
This fixes a regression introduced by https://github.com/hypothesis/h/pull/2808 . `Date.toLocaleDateFormat` is much faster than Moment without options but much slower with. This PR follows the advice on MDN to create a single `Intl.DateTimeFormat` instance and re-use it. It also caches the resulting string.

The end result is that a digest cycle for `/stream` is now ~20ms faster than master and ~6ms faster than before the Moment.js removal on my system.

Fixes #2820

### DIFF
--- a/h/static/scripts/date-util.js
+++ b/h/static/scripts/date-util.js
@@ -1,29 +1,24 @@
-var DATE_SUPPORTS_LOCALE_OPTS = (function () {
-  try {
-    // see http://mzl.la/1YlJJpQ
-    (new Date).toLocaleDateString('i');
-  } catch (e) {
-    if (e instanceof RangeError) {
-      return true;
-    }
-  }
-  return false;
-})();
+// cached date formatting instance.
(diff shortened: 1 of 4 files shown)


nickstenning:
LGTM.

PR #2831, opened 2016-01-06
Fix inline Chrome extension installation

### LINKED ISSUE #2830: Inline Chrome extension installation is broken
The `<link rel="chrome-webstore-item">` is missing from the homepage since it was moved into the app, so inline installation fails and the user gets redirected to the Chrome web store instead, which makes for a less smooth experience.

robertknight:
Add the missing <link rel="chrome-webstore-item"> link
required for inline installation of Chrome extensions.

Previously the install link JS was failing due to this
and the result was that the user got directed to the
Chrome Web Store.

Fixes #2830

### DIFF
--- a/h/home.py
+++ b/h/home.py
@@ -6,7 +6,10 @@ from pyramid import view
                   request_method='GET',
                   renderer='h:templates/old-home.html.jinja2')
 def index(context, request):
-    context = {}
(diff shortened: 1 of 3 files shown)


nickstenning:
LGTM.

PR #2834, opened 2016-01-06
Use content rather than URL-based PDF detection where possible

### LINKED ISSUE #1900: [Chome] [Extension] If a sites serves us PDF, but it doesn't declare it so in the filename, we fail to inject

robertknight:
This fixes #1900 by detecting the native Chrome PDF viewer directly where possible, rather than relying on the URL including ".pdf". This mechanism could also be extended in future as a way to add custom handling for other embedded viewers.

The Chrome PDF viewer is an HTML page consisting of a single `<embed>` tag for the native PDF viewer plugin. Third party PDF viewers such as PDF.js which redirect the PDF downloads to a chrome-extension:// URL were supported previously and are not supported now. I'm checking whether there are any 3rd-party viewers that might be affected by this approach.

An alternative approach that sean explored was to check the 'Content-Type' header of the request for the page's main resource. The downside of that is that we have to watch all incoming network requests and have permissions to do that, since the viewer cannot check after the page is received - unless it makes another network request to fetch the original resource (see related upstream issue at https://code.google.com/p/chromium/issues/detail?id=242575 )

### DIFF
--- a/h/browser/chrome/lib/sidebar-injector.js
+++ b/h/browser/chrome/lib/sidebar-injector.js
@@ -97,20 +97,42 @@ function SidebarInjector(chromeTabs, dependencies) {
     return PDF_VIEWER_URL + '?file=' + encodeURIComponent(url);
   }
 
+  // returns true if the extension is permitted to inject
+  // a content script into a tab with a given URL.
+  function canInjectScript(url) {
+    var canInject;
+    if (isSupportedURL(url)) {
+      canInject = Promise.resolve(true);
+    } else if (isFileURL(url)) {
+      canInject = util.promisify(isAllowedFileSchemeAccess)();
+    } else {
+      canInject = Promise.resolve(false);
+    }
+    return canInject;
+  }
+
   function detectTabContentType(tab) {
     if (isPDFViewerURL(tab.url)) {
       return Promise.resolve(CONTENT_TYPE_PDF);
     }
 
-    if (!isSupportedURL(tab.url)) {
-      return Promise.resolve(CONTENT_TYPE_HTML);
-    }
-
-    return executeScriptFn(tab.id, {
-        code: toIIFEString(detectContentType)
-      }).then(function (frameResults) {
-        return frameResults[0].type;
-      });
+    return canInjectScript(tab.url).then(function (canInject) {
+      if (canInject) {
+        return executeScriptFn(tab.id, {
+            code: toIIFEString(detectContentType)
+          }).then(function (frameResults) {
+            return frameResults[0].type;
+          });
+      } else {
+        // we cannot inject a content script in order to determine the
+        // file type, so fall back to a URL-based mechanism
+        if (tab.url.indexOf('.pdf') !== -1) {
+          return Promise.resolve(CONTENT_TYPE_PDF);
+        } else {
+          return Promise.resolve(CONTENT_TYPE_HTML);
+        }
+      }
+    });
   }
 
   function isPDFViewerURL(url) {
@@ -125,7 +147,7 @@ function SidebarInjector(chromeTabs, dependencies) {
     // Injection of content scripts is limited to a small number of protocols,
     // see https://developer.chrome.com/extensions/match_patterns
     var parsedURL = new URL(url);
-    var SUPPORTED_PROTOCOLS = ['http:', 'https:', 'ftp:', 'file:'];
+    var SUPPORTED_PROTOCOLS = ['http:', 'https:', 'ftp:'];
     return SUPPORTED_PROTOCOLS.some(function (protocol) {
       return parsedURL.protocol === protocol;
     });
--- a/h/browser/chrome/lib/sidebar-injector.js
+++ b/h/browser/chrome/lib/sidebar-injector.js
@@ -3,6 +3,29 @@
 var blocklist = require('../../../static/scripts/blocklist');
 var errors = require('./errors');
 var settings = require('./settings');
+var util = require('./util');
+
+var CONTENT_TYPE_HTML = 'HTML';
+var CONTENT_TYPE_PDF = 'PDF';
+
+// a function which is executed as a content script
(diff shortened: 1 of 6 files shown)


robertknight:
The original issue (#1900) actually links to an expanded version of this approach at https://github.com/BeagleLab/beagle/blob/master/js/main.js#L82  . In the discussion in #1900 I didn't see an explanation of why the much more complex strategy from https://github.com/hypothesis/h/pull/2437 was used. @BigBlueHat - did I miss something critical?

h/browser/chrome/lib/util.js
```diff
@@ -0,0 +1,44 @@
...
+  if (typeof chrome !== 'undefined' && chrome.extension) {
+    return chrome.extension.lastError;
+  } else {
+    return undefined;
+  }
+}
+
+/**
+ * Converts an async Chrome API into a function
+ * which returns a promise.
```
seanh: I am just going to trust you that this works! :)

h/browser/chrome/test/sidebar-injector-test.js
```diff
@@ -7,10 +7,30 @@ describe('SidebarInjector', function () {
   var fakeChromeTabs;
   var fakeFileAccess;
 
+  // the content type that the detection script injected into
+  // the page should report ('HTML' or 'PDF')
+  var contentType;
+  // the return value from the content script which checks whether
+  // the sidebar has already been injected into the page
+  var isAlreadyInjected;
```
seanh: I'd prefer to avoid these globals, but I suppose to do so you'd have to refactor existing tests?

h/browser/chrome/lib/sidebar-injector.js
```diff
@@ -3,6 +3,32 @@
...
+var util = require('./util');
+
+var CONTENT_TYPE_HTML = 'HTML';
+var CONTENT_TYPE_PDF = 'PDF';
+
+// a function which is executed as a content script
+// to determine the type of content being displayed in a tab
+function detectContentType(element) {
+  // check if this is the Chrome PDF viewer
+  var element = element || document;
```
seanh: Is `element` actually used? Is it good to have both a param and a var named `element`?

h/browser/chrome/lib/sidebar-injector.js
```diff
@@ -67,13 +95,52 @@ function SidebarInjector(chromeTabs, dependencies) {
...
+    } else if (isFileURL(url)) {
+      canInject = util.promisify(isAllowedFileSchemeAccess)();
+    } else {
+      canInject = Promise.resolve(false);
+    }
+    return canInject;
+  }
+
+  function detectTabContentType(tab) {
+    if (isPDFViewerURL(tab.url)) {
```
seanh: What cases does this apply to? Could do with a docstring on `isPDFViewerURL()`

h/browser/chrome/lib/sidebar-injector.js
```diff
@@ -67,13 +95,52 @@ function SidebarInjector(chromeTabs, dependencies) {
     }
   };
 
+  /**
+   * Returns the type of content in a DOM element or document
+   */
+  this.detectContentType = detectContentType;
```
seanh: Putting it on `this`? It's a shame that `detectContentType` is now "defined" in two different places (with two different docstrings).

h/browser/chrome/lib/sidebar-injector.js
```diff
@@ -3,6 +3,32 @@
 var blocklist = require('../../../static/scripts/blocklist');
 var errors = require('./errors');
 var settings = require('./settings');
+var util = require('./util');
+
+var CONTENT_TYPE_HTML = 'HTML';
+var CONTENT_TYPE_PDF = 'PDF';
+
+// a function which is executed as a content script
+// to determine the type of content being displayed in a tab
```
seanh: Maybe it's just me, but I prefer to distinguish "docstrings" (i.e. the docs for a function or something) from normal inline code comments, by writing them as a jsdoc or ngdoc comment (or in Python as a docstring). Python is probably where this comes from actually: you wouldn't put a normal code comment before a Python function saying what it does, because that's what a docstring is for. Does the same apply to jsdoc comments? (I know we're never going to want to use jsdoc to extract it, but it's just for clarity.)

The other great thing about Python docstrings is that PEP257 actually describes how they should be written to be clear and consistent. Basically sentences that start with "Return " are good. With this function for example, you would write "Return 'PDF' if the document is a PDF, 'HTML' otherwise. This function is executed as a content script in a tab's document". I think that's much clearer. With the comment as currently written, I actually have to read the code to figure out what the function does / returns.

robertknight: I agree on both counts. I originally only used JSDoc comments for the public API of a module but after doing some testing with a couple of tools which should documentation for functions at the point where you use them (Visual Studio Code, Atom), both will show the first sentence of JSDoc but not inline comments.

h/browser/chrome/lib/sidebar-injector.js
robertknight: That's a mistake. It should be an assignment to the default parameter. ie. If we were using ES6 I would have written `function detectContentType(element = document)`

seanh: As in `element = element || document`?

If you can pass in the element that this function queries against I think you can remove the `el` and `document` globals in the tests for it too

h/browser/chrome/test/sidebar-injector-test.js
robertknight: Yes, I think that would be best done in a separate PR to avoid conflating it with the other changes here.

PR #2838, opened 2016-01-07
Refactor chrome tab error handling

robertknight:
This simplifies management of the error information for a tab when the sidebar fails to inject by storing it in the same object as the rest of the tab state. It also makes it easier to ensure that all the state for a tab is consistent - eg. enforcing that the 'error' property is null when the tab did not error.

It also fixes a bug where the tab was not always correctly re-activated after a page navigation within a tab if activation was attempted but failed on the previous page.


seanh:
> It also fixes a bug where the tab was not always correctly re-activated after a page navigation within a tab if activation was attempted but failed on the previous page.

I don't like this behaviour. If I visit one page and try to activate the extension and it errors, I now can't deactivate the extension while on that page (because clicking the button again opens the error tab). Also the button does not turn dark to show that it's activated. No matter what I do the extension will auto activate when I navigate to the next page.

On the other hand, I don't like the behaviour on master: visit a page and successfully activate the extension, browse to a page where the extension errors, browse to a third page -> extension has been deactivated.

seanh:
I don't think it's worth trying to fix the above though - the dual use of the button for activate/deactivate and see error is the problem, better design needed.

robertknight:
@seanh - I agree, the design needs a re-think - best done separately from the refactoring.

PR #2840, opened 2016-01-07
Refactor badge count fetching in Chrome extension

robertknight:
Separate out the logic for querying the server for annotation
counts and processing the response from the logic for updating
the tab state.

 * Replaced code for using XHR and returning the result via a promise
   with the `fetch()` API

 * Move the client code for the Hypothesis /badge endpoint out
   of TabState into its own module and divide up the tests
   accordingly.

 * Ensure the annotation count for a tab is reset to zero if
   querying the /badge endpoint fails.


h/browser/chrome/lib/fetch.js
```diff
@@ -0,0 +1,20 @@
+/**
+ * Performs a network fetch for JSON data and returns a Promise
+ * for the result.
+ */
+function fetchJSON(url) {
```
nickstenning: You know you're in Chrome here. Can we just use [fetch](https://developer.mozilla.org/en/docs/Web/API/Fetch_API)?

robertknight: `fetch()` was made stable in Chrome 42 (April 2015). I figured that's still a little too recent. We could use the 'isomorphic-fetch' polyfill (http://caniuse.com/#feat=fetch) and then just remove that at some point in future.

robertknight: Done - Added the 'whatwg-fetch' polyfill (isomorphic-fetch is a wrapper around that, which is also Node-compatible, but we only need the client-side one)

h/browser/chrome/lib/tab-state.js
```diff
@@ -151,31 +148,13 @@ function TabState(initialState, onchange) {
    * @param {string} apiUrl The URL of the Hypothesis API.
    */
   this.updateAnnotationCount = function(tabId, tabUrl, apiUrl) {
```
nickstenning: `apiUrl` is now an unused parameter -- you need to update the function signature and the docstring.

robertknight: Thanks. I will add a lint for this kind of thing in future.

nickstenning:
Otherwise LGTM.

h/browser/chrome/lib/tab-state.js
robertknight: Done

PR #2843, opened 2016-01-08
Add "create account or sign in" banner

seanh:
https://trello.com/c/NUoaiUO4/213-create-account-banner

### DIFF
--- a/h/static/styles/app.scss
+++ b/h/static/styles/app.scss
@@ -51,6 +51,24 @@ body {
   }
 }
 
+.create-account-banner {
+  background-color: $gray-dark;
+  border-radius: 2px;
+  color: $color-silver-chalice;
+  font-weight: bold;
+  height: 34px;
+  line-height: 34px;
+  margin-bottom: .72em;
+  margin-left: auto;
+  margin-right: auto;
+  text-align: center;
+  width: 100%;
+}
+
+.create-account-banner a {
+  color: $white;
+}
+
 .sheet {
   border: solid 1px $gray-lighter;
   border-radius: 2px;
--- a/h/templates/app.html.jinja2
+++ b/h/templates/app.html.jinja2
@@ -33,6 +33,14 @@
   />
   </top-bar>
 
+  <div class="create-account-banner" ng-if="auth.status === 'signed-out'">
+    To annotate this document
+    <a href="{{ request.route_path('register') }}" target="_blank">
+      create a free account
+    </a>
+    or <a href="" ng-click="login()">sign in</a>
+  </div>
+
   <div class="content" ng-cloak>
     <!-- Account Dialog -->
     <div class="sheet" ng-if="accountDialog.visible">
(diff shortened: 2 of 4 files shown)


seanh:
screenshot from 2016-01-08 20-49-52

screenshot from 2016-01-08 20-50-31

seanh:
untitled screencast - cropped 11

seanh:
I haven't changed "Sign in" to "Login" in the top bar, as you can partially see in this mockup:



Because doing so wasn't listed on the Trello card.

seanh:
Do we want a feature flag for this? If not, it's done.

h/static/styles/app.scss
```diff
@@ -50,6 +50,24 @@ body {
   }
 }
 
+.create-account-banner {
```
robertknight: Styles for individual components should go in their own .scss files which are then imported at the top.

robertknight: On second thoughts, I think we can leave this as-is for the moment. As long as the component has its own class name, the markup and styling can be easily extracted later if necessary.

seanh:
This one has no feature flag and open design questions on Trello, requires a look from @conordelahunty before merge

seanh:
I've rebased this on latest master.

This **cannot be feature flagged** because the banner is only shown when you're not signed in. If it were feature flagged it'd either be on for everyone or for no one, there's no way to allow staff only to preview it.

Remaining questions from Trello (the checked ones are answered imo):

- [x] Should the banner continue to show when the sign-in form is shown (as it does currently)? Conor says yes (Sean's reason why: it's the only thing on screen with a link to the register form)

- [x] Should the banner appear above or below the sign in form? Conor, Jon and current implementation say above.

- [x] Login vs sign in and register vs create an account vs sign up language - I've used the language used in the mockup on the Trello card, but have not changed the language elsewhere (such as just above this new banner, in the top bar of the sidebar). Suggest general language fixing / making consistent should be done separately.
 
- [ ] Do we use full stops at the ends of sentences, in UI strings? Jon says yes. The mockup and current implementation do not have one.

I think this can be merged now unless we want to add a full stop, or unless Conor wants to see it (e.g. screenshare?) first.

@conordelahunty @judell

robertknight:
@seanh One way you could feature flag this is by making it host-dependent, ie. only setting the feature flag on staging or dev.

robertknight:
... for example, by adding a virtual "staging" group which every user is a member of, but only in non-production environments.

nickstenning:
I think the risk associated with deploying this is so small that I'm not worried about feature flagging it.

robertknight:
I agree, just highlighting a possible way we could have done it.

nickstenning:
@conordelahunty says he's happy.

seanh:
> @seanh One way you could feature flag this is by making it host-dependent, ie. only setting the feature flag on staging or dev.

I don't _think_ this is possible with the current ff system. I could investigate adding it in a separate pr? Might be useful for future when-signed-out features.

PR #2846, opened 2016-01-11
Fix warnings during client test execution

robertknight:
* Wrap error reporting of sidebar injection failures
   so that we can stub it during tests and in future
   get them reported back to us

 * Silence warnings about WebSocket disconnections during
   the socket test.

 * Disallow console warnings during tests. In future it is likely
   that we may end up wrapping console reporting from the main app
   in the same way as in the Chrome extension but for the moment
   it is still useful to log exceptional issues


nickstenning:
LGTM.

PR #2847, opened 2016-01-11
Fix #2845 unsafe eval on GitHub

### LINKED ISSUE #2845: Chrome extension doesn't work on Github
image

seanh:


h/static/scripts/vendor/frame-rpc.js
```diff
@@ -0,0 +1,85 @@
+var isarray = require('isarray');
```
robertknight: The upstream license needs to be reproduced at the top of this file - see https://github.com/substack/frame-rpc/blob/master/LICENSE

seanh: Done

nickstenning:
Looks good, but for two issues:

1. If it's been modified by us, it shouldn't be in `vendor/`. `vendor/` implies unmodified code that has been "vendored" into that directory. If we've modified it put it directly in our source tree, possibly with a comment at the top of the file to explain the purpose of the code.

2. You need to replace the use of `isarray` with `angular.isArray`.

seanh:
Both done, tests should pass now

robertknight:
`angular.isArray()` is fine for the moment but we can now use the standard library's `Array.isArray` function.

judell:
Happy dance!

PR #2850, opened 2016-01-11
Use raven-js for client-side error capture and reporting

robertknight:
This adds reporting of client-side JS errors to Sentry using Raven JS. It covers:
 * The part of the sidebar that lives inside the `<iframe>`
 * The Chrome extension's background page
 * New pages on the site (the new homepage, groups pages)

It does not currently cover Annotator code that runs in the context of the host page. That is likely to generate a lot more reports and I think we should attack that once we've got Sentry/Raven working well in environments that we can control.

For the Chrome extension and site pages we use the basic Raven JS client. For the sidebar, we additionally use the Angular integration which wraps Angular's `$exceptionHandler` service.

Errors are associated with the same Sentry client DSN as the main service and will have the same release version as the 'h' package. If the user ID has been received at the point when the error is logged then it will be included.

----

- [X] ~~Investigate what we need to do to make source maps available in Sentry. We'll likely tackle this work as a separate task but it should be filed for future work~~ (I'll do this separately from this PR)
- [X] Do another round of testing for errors from the main site, Chrome extension and sidebar. Verify that they are all reported to Sentry correctly


judell:
Really looking forward to having this data, thanks Robert.

robertknight:
This is currently producing some spam in the Travis CI logs because the SENTRY_DSN env var is not defined: https://travis-ci.org/hypothesis/h/builds/101828936#L545

robertknight:
This contains all the functionality that I want to include for the initial Sentry integration. I'll do another round of testing tomorrow but the code is ready for review.

robertknight:
Related to https://hypothesis.zendesk.com/inbox/tickets/65 , we are currently capturing the signed in username for errors in the sidebar but not errors that occur in pages on the site, which should be easy, or errors that occur in the background page of the Chrome extension.

Capturing the username for errors on the site at least should be easy to do. For the Chrome extension we could do something like have the app post a message to the background script after login and the Chrome extension's background page would record the last-logged-in user and send that as context with error messages.

h/buildext.py
```diff
@@ -12,6 +12,7 @@
 import subprocess
 import textwrap
 import json
+import raven
```
nickstenning: Sorry to be a pain, but this should go in the second block of imports, as these are organised as:

```
system imports

vendor imports

local imports
```

h/buildext.py
```diff
@@ -20,6 +21,8 @@
 
 import h
 from h._compat import urlparse
+from h.sidebar_app_config import app_config
```
nickstenning: I'm not sure about the name of this module. I'd be inclined to call the module `h.sidebar`: `from h.sidebar import app_config` is just as clear and doesn't stutter like `from h.sidebar_app_config import app_config`.

h/buildext.py
```diff
@@ -149,6 +167,8 @@ def get_env(config_uri, base_url):
     env = paster.bootstrap(config_uri, request)
     request.root = env['root']
 
+    request.sentry = raven.Client(release=raven.fetch_package_version('h'))
```
nickstenning: Do you need to specify `release` here?

h/buildext.py
```diff
@@ -215,8 +235,12 @@ def build_chrome(args):
     # Render the sidebar html.
     if webassets_env.url.startswith('chrome-extension:'):
         build_extension_common(env, bundle_app=True)
+        request = env['request']
+        app_dict = {
```
nickstenning: Not sure what `app_dict` means. `context` would be a more conventional name for a template context object :)

h/test/sidebar_app_config_test.py
```diff
@@ -0,0 +1,25 @@
...
+
+
+@pytest.mark.parametrize("url_in,url_out", [
+    ("http://example.com", "ws://example.com"),
+    ("https://example.com", "wss://example.com"),
+    ("http://monkey.com:80/foo", "ws://monkey.com:80/foo"),
+    ("https://giraffe.com:1443/ws?hello", "wss://giraffe.com:1443/ws?hello"),
+])
+def test_websocketize(url_in, url_out):
+    result = sidebar_app_config._websocketize(url_in)
```
nickstenning: As I've said many times before, please let's not test private functions. Test `app_config` here, not `websocketize`, as there are no external callers of `websocketize` so it's perfectly possible to change the implementation of `app_config` in a way that ensures the app still works correctly, but the tests break, or vice versa.

h/test/views_test.py
```diff
@@ -12,9 +13,10 @@
 from h import views
 
 
+@patch('h.views.app_config', autospec=True)
 class TestAnnotationView(unittest.TestCase):
 
-    def test_og_document(self):
+    def test_og_document(self, app_config):
```
nickstenning: If you're not using these anywhere, I'd suggest you just add a single `autouse` fixture at the bottom of the module, like so:

```python
@pytest.fixture(autouse=True)
def app_config(request):
    patcher = mock.patch('h.views.app_config', autospec=True)
    patcher.start()
```

That way you won't need to modify the rest of the file at all.

robertknight: Done

h/test/sidebar_app_config_test.py
robertknight: Done.

h/buildext.py
robertknight: Done

h/buildext.py
robertknight: It isn't essential, but it is consistent with how we create the client in `sentry.py`.

nickstenning:
Very nice work. LGTM.

PR #2851, opened 2016-01-12
Fix "angular not defined" crash

seanh:
This was preventing the sidebar from loading.


seanh:
Oops! Should have listened to @robertknight the first time

robertknight:
@seanh - Thanks. As a general note going forwards, I think we should prefer utility functions that are not part of Angular because that makes the result more usable across a variety of contexts. In particular, JSON parsing/serialization, copying object properties and testing object types are now covered either by ES5 or ES6 standard library functions - `Object.assign()` instead of `angular.copy()` or `angular.extend()`

PR #2854, opened 2016-01-12
Show the annotation's group not the focused group

### LINKED ISSUE #2839: Stream shows wrong group for annotations
The stream shows the name of the currently focused group as the group for all annotations, regardless of which group the annotation actually came from.

**Steps to reproduce**
1. Visit a page with H embedded, Via or use the bookmarklet to open H
2. Select a private group G from the dropdown other than Public
3. Visit the stream

**Expected**
Recently posted public annotations are shown without a group name next to them

**Actual**
All annotations in the stream are shown as being created in group G

Slack discussion: https://hypothes-is.slack.com/archives/public/p1452189656000444

seanh:
In AnnotationController use domainModel.group instead of
groups.focused() to determine which group to display on the annotation.

groups.focused() works fine as long as the group that the annotation
belongs to is always the currently focused group. This is true when in
the sidebar, but not on the stream or on individual annotation pages. On
those pages the most recently focused group the last time the sidebar
was shown is displayed on every annotation, instead of the actual groups
of the annotations.

This means that we now need to update the domainModel.group of new
annotations when the focused group changes, to support this use:

1. Create a new annotation but don't click save yet
2. Change the focused group
3. Save the annotation

The annotation is correctly saved with the currently (and not the
previously) focused group, and also after the focused group the name of
the newly focused group is displayed on the annotation before it's
saved.

Fixes #2839.

### DIFF
--- a/h/static/scripts/directive/annotation.js
+++ b/h/static/scripts/directive/annotation.js
@@ -398,6 +398,11 @@ function AnnotationController(
     if (vm.editing()) {
       saveToDrafts(drafts, domainModel, vm);
     }
+
+    // New annotations move to the new group, when a new group is focused.
+    if (isNew(domainModel)) {
+      domainModel.group = groups.focused().id;
+    }
   }
 
   function onUserChanged(event, args) {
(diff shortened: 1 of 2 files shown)


seanh:
I'll add a couple of tests for this..

seanh:
Tests added

PR #2855, opened 2016-01-12
Fix annotation thread collapsing

### LINKED ISSUE #2823: Comments should not be visable/unfurled by default
Comments on annotations are now vusible/unfurled by default. They should only be viewable once you have clicked the "x replies" link.

seanh:
This was broken by all the recent annotation directive refactoring:
annotation threads are now shown uncollapsed by default (should be
collapsed) and the buttons to collapse/uncollapse them do nothing.

vm/ctrl.editing became a method vm/ctrl.editing() but some code in the
link function wasn't updated.

This fixes them to be collapsed by default and gets the
collapse/uncollapse buttons working but I'm not sure if the collapse
behaviour is correct with nested threads.

Fixes #2823.

### DIFF
--- a/h/static/scripts/directive/annotation.js
+++ b/h/static/scripts/directive/annotation.js
@@ -789,22 +789,26 @@ function link(scope, elem, attrs, controllers) {
 
   elem.on('keydown', ctrl.onKeydown);
 
-  // FIXME: Replace this counting code with something more sane, and
-  // something that doesn't involve so much untested logic in the link
-  // function (as opposed to unit-tested methods on the AnnotationController,
-  // for example).
-  // Keep track of edits going on in the thread.
+  // FIXME: Replace this counting code with something more sane.
   if (counter !== null) {
-    // Propagate changes through the counters.
     scope.$watch((function() {return ctrl.editing();}), function(editing, old) {
-      if (editing) {
+      if (editing) {  // The user has just started editing this annotation.
+
+        // Keep track of edits going on in the thread.
+        // This 'edit' count is for example to uncollapse a thread if one of
+        // the replies in the thread is currently being edited when the
+        // annotations are first rendered (this can happen when switching
+        // focus to a different group then back again, for example).
         counter.count('edit', 1);
-        // Disable the filter and freeze it to always match while editing.
+
+        // Always show an annotation if it is being edited, even if there's an
+        // active search filter that does not match the annotation.
         if ((thread !== null) && (threadFilter !== null)) {
           threadFilter.active(false);
           threadFilter.freeze(true);
         }
-      } else if (old) {
+
+      } else if (old) {  // The user has just finished editing this annotation.
         counter.count('edit', -1);
         if (threadFilter) {
           threadFilter.freeze(false);
@@ -812,7 +816,6 @@ function link(scope, elem, attrs, controllers) {
       }
     });
 
-    // Clean up when the thread is destroyed.
     scope.$on('$destroy', function() {
       if (ctrl.editing() && counter) {
         counter.count('edit', -1);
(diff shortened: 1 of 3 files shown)


seanh:
This is the basic fix, but I'm not sure that collapse/uncollapse of nested threads is working correctly. Would be nice to add a test too but this part of the code remains poorly understood and not very testable.

seanh:
Actually I think the collapsing works fine, the UI is just quite unclear

seanh:
Tests coming..

seanh:
Ok, tests added.

Also note this commit: <https://github.com/hypothesis/h/commit/e58dc5d5f02538c7d3c84ed290ab47a4f2a31096>, I removed a `$watch()` that (as far as I can tell) isn't doing anything. This is not necessary to fix the bug at hand, but since I was understanding and writing unit tests for the link function and discovered this bit of weirdness, I removed it. This is the commit that the watch came from: <https://github.com/hypothesis/h/commit/9f94f2953b477ed889d59348deff2c4090aca27f>.

The tests that I've added are unit tests for the link function on its own (and I had to export the link function to make doing this reasonable), not unit tests for the directive as a whole. These tests wouldn't actually have caught the bug at hand, since the bug was caused by `AnnotationController`s interface changing and the link function not being updated, you would need tests that test `AnnotationController` and the link function together. These tests do serve to provide documentation for what the link function does though, and safety when making changes to the link function. I might have a look and see if I can write whole-directive tests for this too.

h/static/scripts/directive/annotation.js
```diff
@@ -781,6 +781,49 @@ function AnnotationController(
...
+function link(scope, elem, attrs, controllers) {
+  var ctrl = controllers[0];
+  var thread = controllers[1];
+  var threadFilter = controllers[2];
+  var counter = controllers[3];
+
+  elem.on('keydown', ctrl.onKeydown);
+
+  // FIXME: Replace this counting code with something more sane.
+  if (counter !== null) {
```
robertknight: Should be `if (counter) {`

h/static/scripts/directive/annotation.js
```diff
@@ -781,6 +781,49 @@ function AnnotationController(
...
+        // Keep track of edits going on in the thread.
+        // This 'edit' count is for example to uncollapse a thread if one of
+        // the replies in the thread is currently being edited when the
+        // annotations are first rendered (this can happen when switching
+        // focus to a different group then back again, for example).
+        counter.count('edit', 1);
+
+        // Always show an annotation if it is being edited, even if there's an
+        // active search filter that does not match the annotation.
+        if ((thread !== null) && (threadFilter !== null)) {
```
robertknight: Again, should test falsiness rather than against null

h/static/scripts/directive/test/annotation-test.js
```diff
@@ -355,6 +355,202 @@ describe('annotation', function() {
...
+    }
+
+    function mockDeepCountController() {
+      return {
+        count: sinon.stub()
+      };
+    }
+
+    it('binds AnnotationController.onKeydown to "keydown"', function () {
+      var annotationController = mockAnnotationController();
```
robertknight: There is an awful lot of boilerplate at the start of each of these tests. Can you try to extract it out and more DRY?

h/static/scripts/directive/annotation.js
seanh: This is pre-existing code (also the ones below), I didn't refactor it

h/static/scripts/directive/test/annotation-test.js
seanh: I've made them a little more DRY

h/static/scripts/directive/annotation.js
seanh: By the way it looks to me like this guard (and the one below it as well) is probably only hear because the required controllers are `null` when running the tests!

nickstenning:
Could you add a CHANGES.txt entry for this?

seanh:
Changelog added

nickstenning:
LGTM. Great stuff.

PR #2857, opened 2016-01-12
Paginate groups admin

nickstenning:
Just a small PR to address a minor niggle of my own.

Rather than loading and displaying all groups on one page, use a pagination decorator to provide a slightly more enjoyable experience (for both user and rendering engine).

This also sorts the groups so that the most recently created groups are displayed first.

### DIFF
--- a/h/paginator.py
+++ b/h/paginator.py
@@ -1,0 +1,83 @@
+# -*- coding: utf-8 -*-
+
+from __future__ import division
+
+import functools
+import math
+
+PAGE_SIZE = 20
+
+
+def paginate(wrapped=None, page_size=PAGE_SIZE):
+    """
+    Decorate a view function, providing basic pagination facilities.
+
+    Wraps a view function that returns a :py:class:`sqlalchemy.orm.query.Query`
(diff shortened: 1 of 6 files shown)


h/test/paginator_test.py
```diff
@@ -0,0 +1,161 @@
...
+def test_paginate_returns_resultset():
+    request = DummyRequest()
+    query, view = fake_view(total=10)
+
+    wrapped = paginate(view)
+    result = wrapped(context=None, request=request)
+
+    assert result['results'] == query
+    # Assert that .all() has been called to turn the query into a list.
+    assert query.all_called
```
robertknight: Minor quibble. I generally prefer to verify only what ultimately gets returned and any side-effects and not have the tests dependent on exactly how the data gets fetched.

h/paginator.py
```diff
@@ -0,0 +1,83 @@
+# -*- coding: utf-8 -*-
+
+from __future__ import division
```
robertknight: /me reads https://www.python.org/dev/peps/pep-0238/ - that must have been a fun discussion

PR #2858, opened 2016-01-13
Show correct hostname in 'Add to your site' instructions

robertknight:
Instead of hardcoding the hypothes.is domain in the instructions
for adding Hypothes.is to a site, use request.resource_url() to
get the correct host and scheme for the host serving embed.js

Also remove the instructions for adding 'showHighlights: true'
since this is no longer needed as highlights are shown by default.

(This follows up from a support request where a user was confused about how to use the embed from a private instance of H)


nickstenning:
LGTM.

PR #2859, opened 2016-01-13
Implement a hysteresis threshold in excerpts

robertknight:
Allow the excerpt's contents to exceed the collapsed
height by a threshold set with the 'overflow-hystersis' attribute
before the excerpt's contents are truncated.

This fixes a problem where the contents of an annotation
body could exceed the threshold by only a couple of pixels,
resulting in a 'More' link which resulted in only half a line
of extra text being revealed.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -4,6 +4,9 @@ Unreleased
 Bug fixes
 ---------
 
+- Avoid truncating annotation cards that exceed the collapsed height
+  by only a small amount (feature flagged: truncate_annotations)
+
 - Fix a bug where threads were being shown uncollapsed instead of collapsed
   initially, and the collapse/uncollapse buttons didn't work (#2855)
 
--- a/h/static/scripts/directive/excerpt.js
+++ b/h/static/scripts/directive/excerpt.js
@@ -65,16 +65,16 @@ function excerpt() {
           return {};
         }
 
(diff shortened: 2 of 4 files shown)


nickstenning:
Needs a rebase for the CHANGES.txt addition now I've merged #2855.

robertknight:
Discussed what's remaining for go-live with this with @conordelahunty . Once this is merged and pushed to production, we can turn this on for everyone.

conordelahunty:
@robertknight As long as it's tested and @judell has seen it

nickstenning:
LGTM, although I think we may want to up the threshold value at some point in the future -- 20 pixels is still "not much" extra to display.

judell:
"20 pixels is still "not much" extra to display." Last night's SOTU experience made that very clear. Try searching the sidebar for "More" at https://www.whitehouse.gov/the-press-office/2016/01/12/remarks-president-barack-obama-%E2%80%93-prepared-delivery-state-union-address. For quotes in particular, expanding a 2-line snippet to reveal a 3rd line is common and annoying.

robertknight:
@judell - Ah. The threshold currently only applies to the annotation body. I hadn't considered annotation quotes carefully enough. I can follow up with a separate PR to address that.

judell:
Thanks!

PR #2860, opened 2016-01-13
Build and upload Chrome extensions as build artifacts

nickstenning:
Travis supports uploading named files as "build artifacts" after the tests have run.

This commit adds tasks to the Makefile specifically to build our stage and production extensions (the extension IDs are public data) and sets up Travis to upload zipped extension bundles as artifacts.

At the same time it:

a) removes the development builds of the extension from testing to save
   time, as it seems unlikely that we will break the development
   extension and not the production extensions.

b) removes the build of the Firefox extension from testing, as it seems
   very likely that this iteration of the Firefox extension will soon be
   discarded in its entirety in favour of one built on the WebExtensions
   API.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,10 @@
 # use Travis container build infrastructure
 sudo: false
 addons:
+  artifacts:
+    paths:
+      # Upload all built zipfiles (browser extensions)
+      - $(ls build/*.zip | tr "\n" ":")
   postgresql: "9.4"
 language:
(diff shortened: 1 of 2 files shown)


robertknight:
An issue I ran into is that Travis S3 upload was a Ruby based tool that sometimes resulted in lengthy delays (eg. a minute or so) whilst installing dependencies. Might be something to keep an eye out for.

PR #2861, opened 2016-01-13
Strip Via prefixes in URL normalization

seanh:
https://trello.com/c/xwyKzlwp/223-strip-via-as-part-of-normalization

### DIFF
--- a/h/api/test/uri_test.py
+++ b/h/api/test/uri_test.py
@@ -8,6 +8,13 @@ from h.api import uri
 
 
 @pytest.mark.parametrize("url_in,url_out", [
+    # Should strip https://via.hypothes.is/ from the start of URIs
+    ("https://via.hypothes.is/https://example.com", "https://example.com"),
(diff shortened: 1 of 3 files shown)


h/api/test/uri_test.py
```diff
@@ -8,6 +8,10 @@
 
 
 @pytest.mark.parametrize("url_in,url_out", [
+    # Should strip https://via.hypothes.is/ from the start of URIs.
+    ("https://via.hypothes.is/https://en.wikipedia.org/wiki/Pirate_game",
+     "https://en.wikipedia.org/wiki/Pirate_game"),
```
seanh: Ok I should have used example.com..

seanh: Done

nickstenning:
I think we may want to limit this to URLs that start `https://via.hypothes.is/http{,s}:`, otherwise we get nonsense results if someone tries to annotate the front page of via. Don't worry about this @seanh, I'll do it.

nickstenning:
LGTM.

PR #2862, opened 2016-01-13
Add WebTrends ?WT.mc_id= to URL params blacklist

seanh:
https://trello.com/c/hSxsjvim/222-add-webtrends-to-set-of-url-params-removed-in-normalization

### DIFF
--- a/h/api/uri.py
+++ b/h/api/uri.py
@@ -62,6 +62,7 @@ This package is responsible for defining URI normalization routines for use
 elsewhere in the Hypothesis application. URI expansion is handled by
 :py:function:`h.api.storage.expand_uri`.
 """
+import re
 
 from h._compat import urlparse
 from h._compat import url_quote, url_quote_plus, url_unquote, url_unquote_plus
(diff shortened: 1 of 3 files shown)


nickstenning:
Could you find the WebTrends documentation and check if that's the only relevant query param? Also, having a reference to the documentation in a comment (as with the Google Analytics params) would be very useful.

seanh:
Looking into it

seanh:
Ok, I've added a comment with a link to the relevant WebTrends docs, and I've changed the query param blacklisting to use regexes and blacklisted `WT.*`.

All of the leading `^`s in the regexes are not strictly necessary because of what `re.match()` does but I figured it was good to be explicit and thorough with them.

nickstenning:
I've now fiddled with this enough that I'd appreciate a second set of eyes. @finitud?

h/api/uri.py
```diff
@@ -244,3 +248,8 @@ def _normalize_queryname(name):
 
 def _normalize_queryvalue(value):
     return url_quote_plus(url_unquote_plus(value), safe=UNRESERVED_QUERY_VALUE)
+
+
+def _blacklisted_query_param(s):
+    """Return True if the given string matches any BLACKLISTED_QUERY_PARAMS."""
+    return any(re.match(patt, s) for patt in BLACKLISTED_QUERY_PARAMS)
```
seanh: Oh, nice!

PR #2863, opened 2016-01-14
Add a small hysteresis threshold for annotation quotes

robertknight:
If an annotation quote requires 2 or 3 lines of text, it will
not be truncated, otherwise it will be truncated to 2 lines.

CC @judell


nickstenning:
LGTM.

PR #2865, opened 2016-01-15
Use our own Elasticsearch client when searching

nickstenning:
This PR adds `h.api.search.client.Client`, which is a thin wrapper around elasticsearch-py's Elasticsearch class and is intended to have some of the same attributes as [annotator-store's ElasticSearch class][1] so it can serve as a direct replacement in our codebase.

Then, rather than using `Annotation.search_raw` to perform searches, we use the ElasticSearch client now provided as `request.es`.

The motivation here is to decouple the search code from the persistence code so that we can swap out the persistence code (which would include the Annotation model) without affecting search.

[1]: https://github.com/openannotation/annotator-store/blob/45267b0/annotator/elasticsearch.py#L18

### DIFF
--- a/h/api/search/test/core_test.py
+++ b/h/api/search/test/core_test.py
@@ -1,7 +1,17 @@
 import mock
 import pytest
 
+from pyramid.testing import DummyRequest
+
 from h.api.search import core
+
+
+def dummy_request():
+    """Return a mock request with a faked out ElasticSearch connection."""
+    es = mock.Mock(spec_set=['conn', 'index', 't'])
+    es.conn.search.return_value = {'hits': {'total': 0, 'hits': []}}
+
+    return DummyRequest(es=es)
 
 
 search_fixtures = pytest.mark.usefixtures('query', 'nipsa', 'models', 'log')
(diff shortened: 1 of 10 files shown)


h/api/search/test/core_test.py
```diff
@@ -77,20 +91,20 @@ def test_search_queries_for_replies(query, models):
...
+    assert last_call_kwargs['body'] == builder.build.return_value
 
 
 @search_fixtures
 def test_search_returns_replies(models):
-    """It should return an Annotation for each reply from search_raw()."""
-    models.Annotation.search_raw.side_effect = [
-        # The first time search_raw() is called it returns the result of
-        # querying for the top-level annotations only.
+    """It should return an Annotation for each reply from search()."""
```
robertknight: So I understand, this means that the result is a dict which matches the annotation schema, rather than an instance of a class called "Annotation"?

nickstenning: Yep. I'll update the comment.

finitud:
:+1:

h/api/search/test/core_test.py
nickstenning: Ah, no, wait. This is still correct. The switch to returning dicts happens in #2866.

PR #2866, opened 2016-01-15
Remove model dependence from search

nickstenning:
This commit removes all remaining mentions of `models.Annotation` from `h.api.search`, in order that changes to `models.Annotation` do not affect the behaviour of search.

Clients of `h.api.search` are now responsible for turning the dicts returned from the search methods into `Annotation` objects if necessary.

**NB:** This branch is based on #2865 and will need to be rebased once that has merged.

### DIFF
--- a/h/groups/views.py
+++ b/h/groups/views.py
@@ -102,9 +102,10 @@ def _read_group(request, group):
     """
     url = request.route_url('group_read', pubid=group.pubid, slug=group.slug)
 
-    result = search.search(request, private=False, params={
-        "group": group.pubid, "limit": 1000})
-    annotations = [presenters.AnnotationHTMLPresenter(a)
+    result = search.search(request,
+                           private=False,
+                           params={"group": group.pubid, "limit": 1000})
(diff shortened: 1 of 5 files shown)


finitud:
Looks good to me.

h/groups/views.py
```diff
@@ -102,10 +102,11 @@ def _read_group(request, group):
...
 
-    result = search.search(request, private=False, params={
-        "group": group.pubid, "limit": 1000})
-    annotations = [presenters.AnnotationHTMLPresenter(a)
-                   for a in result['rows']]
+    result = search.search(request,
+                           private=False,
+                           params={"group": group.pubid, "limit": 1000})
+    annotations = [models.Annotation(a) for a in result['rows']]
+    annotations = [presenters.AnnotationHTMLPresenter(a) for a in annotations]
```
finitud: Such an embrace of mutable state makes me slightly uncomfortable, but it's probably the most clear way to do this.

nickstenning: Fair point. I can update this.

PR #2871, opened 2016-01-18
Move configuration of ES index into search

nickstenning:
This PR finishes the work that was started in #2866 to make the search package alone responsible for management of the search index. It moves the autocreation-on-startup of the search index into the `h.api.search` package, and removes that package's dependency on annotator-store.


h/api/db.py
```diff
@@ -140,15 +22,14 @@ def bind_engine(engine, base=Base, should_create=False, should_drop=False):
 
 
 def includeme(config):
-    """Configure and possibly initialize Elasticsearch and its models."""
-    registry = config.registry
-    settings = registry.settings
+    """Configure Elasticsearch client."""
+    settings = config.registry.settings
+
+    if 'es.host' in settings:
```
robertknight: What happens if 'es.host' and 'es.index' are not specified in the settings?

h/api/models/elastic.py
```diff
@@ -0,0 +1,64 @@
...
+            return text_type(uri_)
+        else:
+            return ""
+
+    @property
+    def parent(self):
+        """
+        Return the thread parent of this annotation, if it exists.
+        """
+        if 'references' not in self:
```
robertknight: Could this be simplified to something like `if not isinstance(self.get('references'), list): return None`?

h/api/models/test/elastic_test.py
```diff
@@ -0,0 +1,86 @@
...
+def test_uri():
+    assert Annotation(uri="http://foo.com").uri == "http://foo.com"
+
+
+def test_uri_with_no_uri():
+    assert Annotation().uri == ""
+
+
+def test_uri_when_uri_is_not_a_string():
+    for uri in (True, None, 23, 23.7, {"foo": False}, [1, 2, 3]):
```
robertknight: Aside from programming errors, are there legitimate cases when uri will not be a string?

h/api/models/elastic.py
```diff
@@ -0,0 +1,64 @@
...
+            return None
+        if not isinstance(self['references'], list):
+            return None
+        if not self['references']:
+            return None
+        return Annotation.fetch(self['references'][-1])
```
robertknight: So this means that annotations can, or at least could at some point in the past, have multiple parents?

h/api/models/test/elastic_test.py
```diff
@@ -0,0 +1,86 @@
...
+
+@patch.object(Annotation, 'fetch', spec=True)
+def test_parent_returns_thread_parent(fetch):
+    annotation = Annotation(references=['abc123', 'def456'])
+    parent = annotation.parent
+    assert parent == fetch.return_value
+
+
+@pytest.fixture
+def link_text(request):
```
robertknight: What triggers the use of these fixtures? I expected to see references to `link_text` as arguments to the `test_` functions.

h/api/search/config.py
```diff
@@ -0,0 +1,334 @@
...
+            'filter': ['icu_folding']
+        }
+    }
+}
+
+DOCUMENT_MAPPING = {
+    '_id': {'path': 'id'},
+    '_source': {'excludes': ['id']},
+    'analyzer': 'keyword',
+    'date_detection': False,
```
robertknight: What is this for?

h/api/search/config.py
```diff
@@ -0,0 +1,334 @@
...
+    """Ensure that the ICU analysis plugin is installed for ES."""
+    # Pylint issue #258: https://bitbucket.org/logilab/pylint/issue/258
+    #
+    # pylint: disable=unexpected-keyword-arg
+    names = [x.strip() for x in conn.cat.plugins(h='component').split('\n')]
+    if 'analysis-icu' not in names:
+        message = ("ICU Analysis plugin is not installed for Elasticsearch\n"
+                   "  See the installation instructions for more details:\n"
+                   "  https://github.com/hypothesis/h/blob/master/"
+                   "INSTALL.rst#installing")
```
robertknight: This file no longer exists. Could either link to http://h.readthedocs.org/en/latest/hacking/install.html or http://h.readthedocs.org/en/latest/INSTALL.html

h/api/search/config.py
```diff
@@ -0,0 +1,334 @@
...
+def _update_index_mappings(conn, name, mappings):
+    """Attempt to update the index mappings."""
+    name = _resolve_alias(conn, name)
+    try:
+        for doc_type, body in mappings.items():
+            conn.indices.put_mapping(index=name,
+                                     doc_type=doc_type,
+                                     body=body)
+    except elasticsearch.exceptions.RequestError as e:
+        if not e.error.startswith('MergeMappingException'):
```
robertknight: Do we only get the error back in the form of a string or could we check with `isinstance`?

h/api/search/test/config_test.py
```diff
@@ -0,0 +1,59 @@
+# -*- coding: utf-8 -*-
+
+import itertools
+import re
+import urllib
+
+from h.api.search.config import ANNOTATION_ANALYSIS
+
+
+def test_strip_scheme_char_filter():
```
robertknight: I presume the purpose of this filter is to remove tokens like "http(s)" that would otherwise appear in almost every URI?

robertknight:
I've looked through this now and had a few minor questions, which were mostly about understanding the annotation model. Otherwise, all good from my side.

h/api/db.py
nickstenning: Then the underlying code will default to `http://localhost:9200/annotator`: https://github.com/openannotation/annotator-store/blob/45267b04047210d26ab39b40570e3dd28179a49d/annotator/elasticsearch.py#L28-L29

h/api/models/elastic.py
nickstenning: It could. I'm not sure that would be more readable. And you'd still have to check if it was empty or not to preserve behaviour.

Also, in case it's not clear: this code is unchanged. I'm just moving it.

h/api/models/elastic.py
nickstenning: As discussed this morning, `references` is just list of known ancestors, not a list of multiple parents. The tree structure of the conversation is derived from a set of annotations, each of which has a list of annotations known to be its ancestors.

h/api/models/test/elastic_test.py
nickstenning: No. This is all to deal with the fact that we basically accept anything POSTed to `/api/annotations`. [As yet](https://github.com/hypothesis/h/blob/move-es-config-into-search/h/api/schemas.py#L55) there is no validation that the `uri` field or the `target.source` is a string.

h/api/models/test/elastic_test.py
nickstenning: Correct. This is unused. Let's leave it for now as again this just moving code.

h/api/search/config.py
nickstenning: `git log -S date_detection` says:

```
commit ebfae16
Author: Randall Leeds <tilgovi@hypothes.is>
Date:   2015-02-12 14:28:23 -0800

    Safer document metadat indexing

    Disable indexing of the document field inside the annotation type and
    don't detect dates in the document type. This should help prevent
    errors when saving annotations on documents that have sketchy metadata.
```

h/api/search/config.py
nickstenning: This is indeed out of date. I've updated it with a link directly to the elasticsearch-icu repo, as our install document now just says the same thing -- that you must have it installed.

h/api/search/config.py
nickstenning: Nope, this is the only way of checking it. These errors simply wrap the semi-structured error data which comes back from ElasticSearch.

h/api/search/test/config_test.py
nickstenning: Your guess is as good as mine. This is just moved code, and `git log -S test_strip_scheme_char_filter` reveals the not-altogether-enlightening:

```
commit b0ed4a4
Author: Randall Leeds <tilgovi@hypothes.is>
Date:   2015-04-28 16:05:46 -0700

    Test elasticsearch filters and tokenizers
```

nickstenning: It's mainly to make a search for "example.com/foo/bar" match annotations where the URI field is "https://example.com/foo/bar," I expect.

Issue #2870, opened 2016-01-18
Use Docker Compose for Production Deployment of Services


JJediny:
Wondering if there was any reason docker-compose isn't being used to setup the required services and port binding for use in production?

JJediny:
Errors from attempt: https://gist.github.com/JJediny/3501a2cccb44463c2bfd

- [ ] Seems the final command from the [main Dockerfile is not functioning](https://github.com/hypothesis/h/blob/master/Dockerfile#L61)
- [ ] Docker image of nsqd "nsqio/nsq" in [installation guide](https://h.readthedocs.org/en/latest/INSTALL.html) is crashing on startup.
- [ ] PR shows using nginx for mail server but only becuase this image also pull correctly](https://hub.docker.com/r/svendowideit/ambassador/) I see the comment section directs to use the source Dockerfile... not sure what all this micro-services encompasses or how abstracted it is from other container processes. What all is this container comprised of?

nickstenning:
Hi @JJediny. I think adding a docker-compose file for use in development is a great idea.

I've taken a look at the compose file you used in #2873 and at the trace above, and I think I know what most of the problems might be:

> Seems the final command from the main Dockerfile is not functioning

I'm pretty sure it works fine. The problem you might be having is that you're overriding it in your `docker-compose.yml` -- see [here](https://github.com/hypothesis/h/pull/2873/files#diff-4e5e90c6228fd48698d074241c2ba760R5) (and duplicated [here](https://github.com/hypothesis/h/pull/2873/files#diff-4e5e90c6228fd48698d074241c2ba760R19)). You might want to try removing the `command:` line entirely as the default command for the image should work.

> Docker image of nsqd "nsqio/nsq" in installation guide is crashing on startup.

Here you have the inverse problem! The installation guide shows starting the `nsqio/nsq` container with a command of `/nsqd`, but your `docker-compose.yml` doesn't specify a `command:` for that image. Adding `command: /nsqd` should do the trick.

> PR shows using nginx for mail server...

I'm not sure what the question is here. Nginx isn't a mail server although you can use it as an SMTP reverse proxy, so perhaps that's what you were trying? It's not clear from the pull request as you seem to have got the nginx Dockerfile where `nginx.conf` should be.

For running in development I'd suggest using something like `python -m smtpd -n -c DebuggingServer localhost:25` running in a container to print mails to stdout for inspection.

[...]

PR #2873, opened 2016-01-18
Attempt to translate/automate installation using docker-compose

JJediny:
Attempt to translate/automate installation using docker-compose… Errors/Issues on #2870

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,46 @@
+hypthosis:
+  build: .
+  ports:
+   - "5000:5000"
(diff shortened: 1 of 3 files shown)


JJediny:
Only meant as a *placeholder* for discussion around errors and approach

nickstenning:
I've added some comments on the parent issue: #2870.

nickstenning:
I'm going to close this PR. We can continue to use the parent issue for comments, but there are a number of issues raised there which would need to be addressed before we could merge this.

PR #2874, opened 2016-01-19
Decouple sidebar app HTML / embed JS template rendering from Pyramid

robertknight:
This is a first step towards making it possible to build the browser extension and sidebar without running the website and more importantly, all the services that it depends upon.

This PR removes all references to the Pyramid request object from the app.html and embed.js templates. The view callables for the sidebar, single annotation page and stream now pass in the URLs that the template requires directly, plus a webassets environment which provides the URLs for scripts and styles.

I also removed a redundant X-UA-Compatible  `<meta>` tag from 'base.html.jinja2' which is no longer needed since IE adopts the same behavior in the presence of the `<!DOCTYPE html>` tag.

----
* [x] - Pass Google Analytics tracking ID into app HTML


h/templates/app.html.jinja2
```diff
@@ -1,127 +1,93 @@
...
-      "apiUrl": "{{ request.route_url('api') }}",
-      "websocketUrl": "{{ request.route_url('ws') | websocketize }}"
-    }
-  </script>
-  {% assets "app_js" %}
-    <script src="{{ ASSET_URL }}"></script>
-  {% endassets %}
-{% endblock %}
+    <!-- Analytics !-->
+    {% if ga_tracking_id %}
```
robertknight: I'm not passing in 'ga_tracking_id' here. I'm just checking whether we still need this.

h/client_assets.py
```diff
@@ -0,0 +1,153 @@
...
+        ga_cookie_domain = 'none'
+    else:
+        ga_cookie_domain = 'auto'
+
+    # the serviceUrl parameter must contain a path element
+    if urlparse(base_url).path == '':
+        base_url += '/'
+
+    app_config = {
+      'apiUrl': api_url,
```
nickstenning: Always 4-space indents in Python.

h/client_assets.py
```diff
@@ -0,0 +1,153 @@
...
+      'serviceUrl': base_url,
+      'websocketUrl': websocket_url,
+    }
+
+    if sentry_dsn:
+        app_config.update({
+          'raven': {
```
nickstenning: Ditto. 4-space indents.

h/client_assets.py
```diff
@@ -0,0 +1,153 @@
...
+            'dsn': sentry_dsn,
+            'release': __version__
+          }
+        })
+
+    return {
+      'app_config': json.dumps(app_config),
```
nickstenning: 4-space indents.

h/client_assets.py
```diff
@@ -0,0 +1,153 @@
...
+                         create an account
+    :param sentry_dsn: The _public_ Sentry DSN for client-side crash reporting
+    :param ga_tracking_id: The Google Analytics tracking ID
+    :param extra: A dict of optional properties specifying link tags and
+                  meta attributes to be included on the page, passed through to
+                  app.html.jinja2
+    """
+    template = Template(_read_resource('templates/app.html.jinja2'))
+    assets_dict = _app_html_context(
+      api_url=api_url,
```
nickstenning: 4-space indents.

h/buildext.py
```diff
@@ -225,15 +225,32 @@ def build_chrome(args):
         subprocess_args.append('--debug')
     subprocess.call(subprocess_args)
 
-    # Render the sidebar html.
+
+    # Render the sidebar html
+    base_url = env['request'].route_url('index')
```
nickstenning: You define `base_url` but then inconsistently use `args.base` and `base_url` below. This needs to be clear and consistent either way.

h/buildext.py
```diff
@@ -225,15 +225,32 @@ def build_chrome(args):
...
     if webassets_env.url.startswith('chrome-extension:'):
         build_extension_common(env, bundle_app=True)
-        request = env['request']
-        context = {
-          'app_config': app_config(request)
-        }
         with codecs.open(content_dir + '/app.html', 'w', 'utf-8') as fp:
-            data = render('h:templates/app.html.jinja2', context, request)
+            data = client_assets.render_app_html(
+              api_url=api_url,
```
nickstenning: 4-space indents.

h/client_assets.py
```diff
@@ -0,0 +1,153 @@
...
+    :param websocket_url: The WebSocket URL which the client should connect to
+    :param register_url: The URL which new users should be sent to in order to
+                         create an account
+    :param sentry_dsn: The _public_ Sentry DSN for client-side crash reporting
+    :param ga_tracking_id: The Google Analytics tracking ID
+    :param extra: A dict of optional properties specifying link tags and
+                  meta attributes to be included on the page, passed through to
+                  app.html.jinja2
+    """
+    template = Template(_read_resource('templates/app.html.jinja2'))
```
nickstenning: It would probably be neater to define a module-level template environment:

```python
from jinja2 import Environment, PackageLoader
jinja_env = Environment(loader=PackageLoader(__package__, 'templates'))
```

and then load templates from it later as:

```python
jinja_env.get_template('app.html.jinja2')
```

h/buildext.py
```diff
@@ -20,8 +20,8 @@
 import raven
 
 import h
+from h import client_assets
```
nickstenning: Could this module just be called `h.client`?

h/buildext.py
```diff
@@ -53,11 +53,13 @@ def build_extension_common(env, bundle_app=False):
...
         if bundle_app:
-            app_uri = request.webassets_env.url + '/app.html'
+            app_html_url = request.webassets_env.url + '/app.html'
         else:
-            app_uri = request.route_url('widget')
-        value = {'app_uri': app_uri}
-        data = render('h:templates/embed.js.jinja2', value, request=request)
+            app_html_url = request.route_url('widget')
+        data = client_assets.render_embed_js(
+          webassets_env=request.webassets_env,
```
nickstenning: 4-space indents.

nickstenning: Or, in cases like this, PEP8 style would be preferable:

```python
data = client_assets.render_embed_js(webassets_env=request.webassets_env,
                                     app_html_url=app_html_url)
```

h/views.py
```diff
@@ -32,6 +32,20 @@ def _handle_exc(request):
         raise
 
 
+def _render_app(request, extra={}):
+    request.response.text = client_assets.render_app_html(
+      api_url=request.route_url('api'),
+      base_url=request.route_url('index'),
+      extra=extra,
+      ga_tracking_id=request.registry.settings.get('ga_tracking_id'),
+      register_url=request.route_url('register'),
```
nickstenning: For now I guess I can live with this, but we probably shouldn't have two different ways of passing URLs into the client application. They should probably all live in the client settings object.

h/views.py
```diff
@@ -32,6 +32,19 @@ def _handle_exc(request):
         raise
 
 
+def _render_app(request, extra={}):
+    request.response.text = client.render_app_html(
+      api_url=request.route_url('api'),
```
nickstenning: 4-space indents here!

h/views.py
```diff
@@ -84,22 +95,22 @@ def annotation(context, request):
...
 
-@view_config(route_name='embed', renderer='h:templates/embed.js.jinja2')
+@view_config(route_name='embed')
 def embed(context, request):
     request.response.content_type = b'text/javascript'
-    return {
-        'app_config': app_config(request),
-    }
+    request.response.text = client.render_embed_js(
+      webassets_env=request.webassets_env,
```
nickstenning: Ditto!

nickstenning:
Ok, good stuff. Sorry for all the nitpicking.

PR #2877, opened 2016-01-20
Remove the incomplete Firefox extension

robertknight:
A recent successful prototype of a Firefox extension
based on the new WebExtensions API highlighted that as
the way forwards for Hypothesis on Firefox, as it lets
us offer feature parity with much easier maintenance.

If we decide that we want to support the extension for Firefox pre-45, we may
do that by shimming the Chrome APIs using the JPM SDK APIs, similar
to what https://github.com/einaregilsson/Redirector have done. Either way, we
clearly want to share the bulk of the code and tests as far as possible.

In the meantime, removing the legacy extension code makes the current
front-end build refactoring easier.


nickstenning:
LGTM.

PR #2878, opened 2016-01-20
Decouple Chrome extension build from Pyramid

robertknight:
**This builds on https://github.com/hypothesis/h/pull/2874 and will need to be rebased once that is merged**

This PR removes all usage of Pyramid and Pylons from the Chrome extension build, except for a utility function in pyramid.path.

Instead, the script creates a minimal webassets environment, adds in the bundles defined in `assets.yaml` and uses that to generate the output resources.

To simplify creation of this webassets environment, the configuration parameters for Browserify and CleanCSS which used to be defined in the paster config file are now just listed directly in `assets.py`.

As a result, the hypothesis-buildext script no longer requires a paster config file as input. Instead it just takes the name of the browser to build for, plus --debug, --base and --assets flags.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -73,7 +73,7 @@ extensions: build/$(ISODATE)-$(BUILD_ID)-chrome-prod.zip
 build/%-chrome-stage.zip:
 	@rm -rf build/chrome $@
 	hypothesis-buildext chrome \
-		--base 'https://stage.hypothes.is' \
+		--service 'https://stage.hypothes.is' \
 		--sentry-dsn $(SENTRY_DSN_STAGE) \
 		--assets 'chrome-extension://iahhmhdkmkifclacffbofcnmgkpalpoj/public'
 	@zip -qr $@ build/chrome
@@ -81,7 +81,7 @@ build/%-chrome-stage.zip:
 build/%-chrome-prod.zip:
 	@rm -rf build/chrome $@
 	hypothesis-buildext chrome \
-		--base 'https://hypothes.is' \
+		--service 'https://hypothes.is' \
 		--sentry-dsn $(SENTRY_DSN_PROD) \
 		--assets 'chrome-extension://bjfhmglciegochdpefhhlphglcehbmek/public'
 	@zip -qr $@ build/chrome
--- a/Makefile
+++ b/Makefile
@@ -74,7 +74,7 @@ build/%-chrome-stage.zip:
 	@rm -rf build/chrome $@
 	hypothesis-buildext chrome \
 		--service 'https://stage.hypothes.is' \
-		--sentry-dsn "$(SENTRY_DSN_STAGE)" \
+		--sentry-public-dsn '$(SENTRY_DSN_STAGE)' \
 		--assets 'chrome-extension://iahhmhdkmkifclacffbofcnmgkpalpoj/public'
 	@zip -qr $@ build/chrome
 
@@ -82,7 +82,7 @@ build/%-chrome-prod.zip:
 	@rm -rf build/chrome $@
 	hypothesis-buildext chrome \
 		--service 'https://hypothes.is' \
-		--sentry-dsn "$(SENTRY_DSN_PROD)" \
+		--sentry-public-dsn '$(SENTRY_DSN_PROD)' \
 		--assets 'chrome-extension://bjfhmglciegochdpefhhlphglcehbmek/public'
 	@zip -qr $@ build/chrome
 
--- a/Makefile
+++ b/Makefile
@@ -74,7 +74,7 @@ build/%-chrome-stage.zip:
 	@rm -rf build/chrome $@
 	hypothesis-buildext chrome \
 		--service 'https://stage.hypothes.is' \
-		--sentry-dsn $(SENTRY_DSN_STAGE) \
+		--sentry-dsn "$(SENTRY_DSN_STAGE)" \
 		--assets 'chrome-extension://iahhmhdkmkifclacffbofcnmgkpalpoj/public'
 	@zip -qr $@ build/chrome
 
@@ -82,7 +82,7 @@ build/%-chrome-prod.zip:
 	@rm -rf build/chrome $@
 	hypothesis-buildext chrome \
 		--service 'https://hypothes.is' \
-		--sentry-dsn $(SENTRY_DSN_PROD) \
+		--sentry-dsn "$(SENTRY_DSN_PROD)" \
 		--assets 'chrome-extension://bjfhmglciegochdpefhhlphglcehbmek/public'
 	@zip -qr $@ build/chrome
 
--- a/Makefile
+++ b/Makefile
@@ -72,19 +72,17 @@ extensions: build/$(ISODATE)-$(BUILD_ID)-chrome-prod.zip
 
 build/%-chrome-stage.zip:
 	@rm -rf build/chrome $@
-	SENTRY_DSN=$(SENTRY_DSN_STAGE) hypothesis-buildext \
-		conf/production.ini \
-		chrome \
+	hypothesis-buildext chrome \
 		--base 'https://stage.hypothes.is' \
+		--sentry-dsn $(SENTRY_DSN_STAGE) \
 		--assets 'chrome-extension://iahhmhdkmkifclacffbofcnmgkpalpoj/public'
 	@zip -qr $@ build/chrome
 
 build/%-chrome-prod.zip:
 	@rm -rf build/chrome $@
-	SENTRY_DSN=$(SENTRY_DSN_PROD) hypothesis-buildext \
-		conf/production.ini \
-		chrome \
+	hypothesis-buildext chrome \
 		--base 'https://hypothes.is' \
+		--sentry-dsn $(SENTRY_DSN_PROD) \
 		--assets 'chrome-extension://bjfhmglciegochdpefhhlphglcehbmek/public'
 	@zip -qr $@ build/chrome
 
--- a/h/buildext.py
+++ b/h/buildext.py
@@ -14,13 +14,16 @@ import subprocess
 import textwrap
 
 from jinja2 import Environment, PackageLoader
-from pyramid import paster
 from pyramid.path import AssetResolver
-from pyramid.request import Request
 import raven
+import webassets
+from webassets.loaders import YAMLLoader
 
 import h
+# h.assets provides webassets filters for JS and CSS compilation
+import h.assets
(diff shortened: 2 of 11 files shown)


h/buildext.py
```diff
@@ -35,13 +39,32 @@
...
+       package-relative paths (eg. 'h:path/to/asset.css') that are used
+       in assets.yaml
+    """
+
+    def search_for_source(self, ctx, item):
+        if item.startswith('../'):
+            # relative URLs in assets.yaml are relative to h/static
+            # Note that the return path here is not canonicalized
+            # (ie. '../' is not removed) for consistency with the way
+            # webassets resolves paths in the main app.
```
robertknight: I had a little difficulty explaining this here. The issue is that when building assets outside of the root 'h/static' directory, webassets puts the results in `h/static/webassets-external/[MD5_HASH_OF_PATH]_filename` where `MD5_HASH_OF_PATH` is a hash of the absolute, non-canonicalized path (eg. "/home/robert/hypothesis/h/h/static/../../node_modules/.../angular.js").

When building in development mode where embed.js is included in the Chrome extension but other JS files are not, the webassets environment in buildext.py needs to generate the same URLs for vendor JS files referenced in embed.js as the '/embed.js' route in the main app.

nickstenning:
Ah, I hadn't noticed that `/docs/help` got broken by #2874. Is there a solution to this? Can we still trigger a build of the client for use in a page like that? How is that supposed to work now?

robertknight:
Fixed `/docs/help` with #2893 - we embed H using the "public API" for that task.

h/buildext.py
```diff
@@ -35,13 +39,32 @@
 resolve = AssetResolver().resolve
 
 
-def build_extension_common(env, bundle_app=False):
+class Resolver(webassets.env.Resolver):
+    """A custom resolver for webassets which can resolve
```
nickstenning: I'll fix this separately, but docstrings should be formatted like one of the following examples:

```
"""A single line docstring."""
```

```
"""
The first line of the docstring.

Further description in a following paragraph...
...blah blah blah.
"""
```

There is a third style which some people use but which I have a mild preference for avoiding:

```
"""The first line of the docstring.

Further description in a following paragraph...
...blah blah blah.
"""
```

This one is none of these three and doesn't format correctly.

h/buildext.py
nickstenning: Perhaps you could update this comment with the explanation you've given here. It's a lot clearer than the one currently in the source.

h/buildext.py
robertknight: Thanks. Sorry about these issues. I've actually read the relevant PEPs now, I should have done that much sooner.

h/buildext.py
```diff
@@ -93,21 +117,14 @@ def chrome_manifest(request):
         version = h.__version__
         version_name = 'Official Build'
 
-    src = request.resource_url(request.context)
-
-    # We need to use only the host and port for the CSP script-src when
```
nickstenning: You've removed this behaviour. Unless I'm misunderstanding something that will mean that if you now build a development extension (i.e. one where assets are loaded from localhost) then the CSP checks for the extension will fail when loading assets.

Or is loading assets from the application no longer supported? Or perhaps this was a bug in Chrome that has since been fixed?

h/buildext.py
```diff
@@ -188,22 +198,22 @@ def build_chrome(args):
 
         chrome-extension://<extensionid>/public
     """
-    paster.setup_logging(args.config_uri)
-
-    os.environ['WEBASSETS_BASE_DIR'] = os.path.abspath('./build/chrome/public')
-    if args.assets is not None:
-        os.environ['WEBASSETS_BASE_URL'] = args.assets
+    base_url = args.service_url
```
nickstenning: This is now mightily confusing, because we have two names for the same thing and it's not always clear which one is supposed to have a trailing slash and which one isn't.

If this is being passed around internally as `base_url`, I'm not sure we should have renamed the external argument to ``--service``.

If the only thing we really need is the service URL, then we probably shouldn't be chopping off the trailing slash here and then adding it back later with `url_with_path`.

h/buildext.py
```diff
@@ -219,80 +236,78 @@ def build_chrome(args):
...
-    sentry_dsn = env['request'].sentry.get_public_dsn('https')
-    settings = settings_dict(base_url=base_url,
-                             api_url=api_url,
-                             sentry_dsn=sentry_dsn)
+    api_url = '{}/api'.format(base_url)
+    websocket_url = websocketize('{}/ws'.format(base_url))
+    sentry_dsn = None
+
+    if args.sentry_dsn:
+        sentry_dsn = raven.Client(args.sentry_dsn).get_public_dsn()
```
nickstenning: As previously discussed, this won't work if the passed DSN is a public DSN, which it will have to be on Travis:

```
$ python -c 'import raven; raven.Client("https://e06fdcb78fcd46e48ba87129ae142180@app.getsentry.com/37293").get_public_dsn()'

  File "<string>", line 1, in <module>
  File "/Users/borior/.virtualenvs/h/lib/python2.7/site-packages/raven/base.py", line 140, in __init__
    self.set_dsn(dsn, transport)
  File "/Users/borior/.virtualenvs/h/lib/python2.7/site-packages/raven/base.py", line 194, in set_dsn
    transport_registry=self._registry,
  File "/Users/borior/.virtualenvs/h/lib/python2.7/site-packages/raven/conf/remote.py", line 91, in from_string
    raise InvalidDsn('Invalid Sentry DSN: %r' % url.geturl())
raven.exceptions.InvalidDsn: Invalid Sentry DSN: 'https://e06fdcb78fcd46e48ba87129ae142180@app.getsentry.com/37293'
```


I'd suggest we explicitly name the argument `--sentry-public-dsn` and then use it directly, without validation, if passed.

Makefile
```diff
@@ -72,19 +72,17 @@ extensions: build/$(ISODATE)-$(BUILD_ID)-chrome-prod.zip
 
 build/%-chrome-stage.zip:
 	@rm -rf build/chrome $@
-	SENTRY_DSN=$(SENTRY_DSN_STAGE) hypothesis-buildext \
-		conf/production.ini \
-		chrome \
-		--base 'https://stage.hypothes.is' \
+	hypothesis-buildext chrome \
+		--service 'https://stage.hypothes.is' \
+		--sentry-dsn "$(SENTRY_DSN_STAGE)" \
```
nickstenning: It's a tiny thing, I know, but the other arguments use single quotes, so these should as well.

h/buildext.py
robertknight: Indeed. I'll do that.

h/buildext.py
```diff
@@ -219,80 +236,78 @@ def build_chrome(args):
...
 
     # Render the sidebar html
-    base_url = env['request'].route_url('index')
-    api_url = env['request'].route_url('api')
-    websocket_url = client.websocketize(env['request'].route_url('ws'))
-    sentry_dsn = env['request'].sentry.get_public_dsn('https')
-    settings = settings_dict(base_url=base_url,
-                             api_url=api_url,
-                             sentry_dsn=sentry_dsn)
+    api_url = '{}/api'.format(base_url)
```
nickstenning: This previously had a trailing slash on it. I've had a quick look and I don't think this change should make any difference to how the client works, but wanted to call that out in case you knew better than me.

h/buildext.py
```diff
@@ -219,80 +236,78 @@ def build_chrome(args):
...
-                           help='A path (relative to base) or URL from which '
-                           'to load the static assets',
-                           default=None,
-                           metavar='PATH/URL')
+parser.add_argument('--sentry-dsn',
+                    default='',
+                    help='Specify the Sentry DSN for crash reporting',
+                    metavar='SENTRY_DSN')
+parser.add_argument(
+    '--service',
```
nickstenning: Not sure why this argument is formatted completely differently from all the others. If it could not be that would be lovely.

h/buildext.py
```diff
@@ -219,80 +236,78 @@ def build_chrome(args):
...
 
 BROWSERS = {'chrome': build_chrome}
 
 
 def main():
-    # Set a flag in the environment that other code can use to detect if it's
-    # running in a script rather than a full web application. See also
-    # h/script.py.
-    #
-    # FIXME: This is a nasty hack and should go when we no longer need to spin
```
nickstenning: Hooray!

h/buildext.py
robertknight: You're quite right. This should have a trailing slash on it. Otherwise that breaks the resolution of the 'token' URL.

h/buildext.py
robertknight: Fixed

h/buildext.py
robertknight: This flag is still being set in `script.py` at the moment. Was it only needed for the extension build or did we need it for other purposes as well?

Makefile
robertknight: Done. I incorrectly assumed that make wouldn't substitute variables in single-quoted strings.

h/buildext.py
robertknight: Actually this is preserving the previous behavior, the "script_host_url" passed as the context when rendering the manifest is the host + port.

What has changed is that now we always include just the origin of the URL serving the assets, never the path. Previously if you tried to run your local H instance on a host other than localhost, the CSP entry in the manifest would end up including the path as well.

h/buildext.py
robertknight: Done

h/buildext.py
```diff
@@ -219,80 +235,77 @@ def build_chrome(args):
...
     # Write build settings to a JSON file
     with codecs.open('build/chrome/settings-data.js', 'w', 'utf-8') as fp:
+        settings = settings_dict(service_url, api_url, args.sentry_public_dsn)
         fp.write('window.EXTENSION_CONFIG = ' + json.dumps(settings))
 
 
-parser = argparse.ArgumentParser('hypothesis-buildext')
+def check_sentry_dsn_is_public(dsn):
+    parsed_dsn = urlparse.urlparse(dsn)
+    if parsed_dsn.password:
```
nickstenning: Nice.

h/buildext.py
nickstenning: It's still needed in the other places, I'm afraid. It's stopping app spin-up things happening when they're not meant to.

nickstenning:
LGTM. Thank you so much for all this awesome work.

PR #2880, opened 2016-01-20
Extract the login form into its own directive

robertknight:
This makes the login form consistent with other recently
written components by having the controller and directive in a single
file and using an element directive in the main app.

Having all the logic in one place should make it a little easier when we come to move the login form to a popup to avoid problems with third-party cookie restrictions and enable OAuth login flows in future.


h/static/scripts/directive/login-form.js
```diff
@@ -0,0 +1,106 @@
...
+      controller: Controller,
+      controllerAs: 'vm',
+      restrict: 'E',
+      scope: {
+        onClose: '&',
+      },
+      templateUrl: 'login_form.html',
+    };
+  },
+  controller: Controller,
```
nickstenning: The naming here should be consistent. If the thing is a constructor, it should always be `Controller`. If it's not, it should always be `controller`.

nickstenning:
I think you also want to commit the removal of `static/scripts/auth-controller.js` here.

robertknight:
> I think you also want to commit the removal of static/scripts/auth-controller.js here.

Done

nickstenning:
LGTM.

PR #2881, opened 2016-01-20
Decouple the Hypothesis service URL from the app's base URI

robertknight:
Previously the URL set as the <base> tag was used indirectly
to specify the URL of the Hypothesis service that various
links referred to as well as for other purposes.

Make things clearer by adding an explicit 'serviceUrl' attribute
to the Hypothesis client settings and use that in place of the
baseURI. This also gets us a step closer towards making it possible
to build a development front-end which uses the production service for debugging etc.

The custom template whitelist configuration has been removed
because all of the templates are embedded in app.html, so the
default value of ['self'], allowing templates to be loaded
only from the same domain as the page, is fine - unless anyone is aware of any scenarios where this would not be the case.

---

* [x] I see a local issue with this branch where the Chrome extension fails to connect to the websocket. Need to investigate whether this is related to the changes in this branch. (_This was due to be not having whitelisted requests from the chrome-extension://$ID URL of my local extension build_)
* [x] The 'New Group' link in the Chrome extension should use the service URL explicitly.
* [x] Links in the 'Sign In' menu should use the service URL explicitly.
* [x] Forgot Password and 'Create an account' links need to include the service URL / be absolute
* [x] Share links are broken in the Chrome extension

### DIFF
--- a/h/sidebar.py
+++ b/h/sidebar.py
@@ -11,5 +11,6 @@ def app_config(request):
     """
     return {
         'apiUrl': request.route_url('api'),
+        'serviceUrl': request.route_url('index'),
         'websocketUrl': _websocketize(request.route_url('ws')),
     }
--- a/h/static/scripts/app.coffee
+++ b/h/static/scripts/app.coffee
@@ -5,7 +5,6 @@ if window.RAVEN_CONFIG
 
 
 require('autofill-event')
-baseURI = require('document-base-uri')
 angular = require('angular')
 require('angular-jwt')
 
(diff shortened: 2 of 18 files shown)


nickstenning:
Just noting that if I merge #2880 first this will require a rebase as it changes `auth-controller.js` which turns into a directive in that PR.

robertknight:
Yes. If #2880 can be merged first I'll rebase this afterwards.

nickstenning:
Done. Now needs a rebase.

h/buildext.py
```diff
@@ -146,6 +146,7 @@ def settings_dict(env):
 
     config.update({
         'buildType': build_type_from_api_url(api_url),
+        'serviceUrl': base_url,
```
nickstenning: Doesn't `config` already contain `serviceUrl`? https://github.com/hypothesis/h/pull/2881/files#diff-065076192547b184e5a6d61622292028R14

robertknight: Ah, yes indeed.

nickstenning:
LGTM.

PR #2882, opened 2016-01-20
Recognize 'group' as a valid parameter name in search queries

robertknight:
The backend already supports searches for 'group:hashid' but
this was not supported on the stream because 'group'
was not recognized as a valid parameter so it was passed
to the backend in the 'any' field.

This trivial PR only exposes the minimal group search functionality that the backend already had. It doesn't by any means implement group search in a friendly way - eg. allowing the user to search for "group:_part of the name_". It does make it possible to link to a stream of annotations from a particular group however.

### DIFF
--- a/h/static/scripts/search-filter.coffee
+++ b/h/static/scripts/search-filter.coffee
@@ -12,7 +12,8 @@ module.exports = class SearchFilter
       # The whole term is data
       return [null, term]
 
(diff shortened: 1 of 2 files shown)


judell:
Thanks!

nickstenning:
LGTM.

PR #2883, opened 2016-01-20
Remove the static blocklist

robertknight:
The configuration option for defining a static blocklist of sites
on which Hypothesis should not load has never been used and
we're not anticipating doing so.


nickstenning:
LGTM.

PR #2884, opened 2016-01-21
Fix canceling edits reverting to the first-loaded version of an annotation

robertknight:
When updating an annotation, the update is currently applied
to a copy of the domain model. AIUI the rationale for this is to ensure
that 'domainModel' always reflects what is actually stored on the
server, so it should not be updated until after the update is
successfully committed.

However, after the update is successfully commited, the view was
updated with the updated domain model, but the local domain model
instance was never replaced. Therefore when reverting changes,
the card reverted to the first-loaded text for the annotation
instead of the last-saved version.

This commit fixes the problem by replacing the domain model
reference whenever an annotation update is received.

---

Addendum, as Sean has pointed out a couple of times, the tests for `annotation.js` are a mess and the directive does far too much. I'm going to create some follow up cards to split it into smaller pieces.

### DIFF
--- a/h/static/scripts/directive/annotation.js
+++ b/h/static/scripts/directive/annotation.js
@@ -384,6 +384,7 @@ function AnnotationController(
 
   function onAnnotationUpdated(event, updatedDomainModel) {
     if (updatedDomainModel.id === domainModel.id) {
+      domainModel = updatedDomainModel;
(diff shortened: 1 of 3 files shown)


nickstenning:
DWISOTT. :wink:

dwhly:
TIL DWISOTT

PR #2887, opened 2016-01-22
Prevent double-posting of new annotations

### LINKED ISSUE #2864: Annotation form allows duplicate annotation creation
#### Steps to reproduce

1. Start creating a new annotation.
2. Add some text to the annotation.
3. Instead of clicking once on the "Post to Public" button, click as fast as possible as many times as possible.

#### Expected

Only one annotation should be created.

#### Actual

As many annotations are created as times you clicked until the form derenders.

----

This interacts particularly badly with users who have buggy hardware or software that consistently clicks multiple times. Case in point: https://hypothes.is/stream?q=user:acct:reinahondo@hypothes.is

robertknight:
When the user clicks the 'Post' button to create an annotation,
optimistically switch the card back to View mode but display
a 'Saving...' indicator in place of the Reply/Edit/Delete links.

This makes the UI appear more responsive when the user clicks
the Post button and also prevents an issue where the user could
click 'Post' multiple times during the save and create multiple
annotations.

 * Fix a possible inconsistency between the 'Post' button's enabled
   state and whether or not the save() function can succeed.

   The hasContent() and isShared() methods also already have tests,
   so this lets us remove several redundant tests.

 * Fix inconsistency in the return type of the save() function -
   always return a promise.

 * Treat negative status values as network errors as well as 0.
   If the server is unreachable, the real status value may be -1.

Fixes #2864

----

Current Design:

save-indicator

### DIFF
--- a/h/static/scripts/directive/annotation.js
+++ b/h/static/scripts/directive/annotation.js
@@ -27,11 +27,11 @@ function domainModelTagsFromViewModelTags(viewModelTags) {
  */
 function errorMessage(reason) {
   var message;
-  if (reason.status === 0) {
+  if (reason.status <= 0) {
     message = 'Service unreachable.';
   } else {
     message = reason.status + ' ' + reason.statusText;
-    if (reason.data.reason) {
+    if (reason.data && reason.data.reason) {
       message = message + ': ' + reason.data.reason;
     }
   }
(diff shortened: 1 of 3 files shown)


robertknight:
Design feedback: Disable the button whilst posting for now. Consider a more optimistic-update style UI as part of forthcoming annotation card design updates.

nickstenning:
Really nice fix.

PR #2889, opened 2016-01-25
Do Not Pull - another stab at docker-compose

JJediny:
Per issue #2870


docker-compose.yml
```diff
@@ -0,0 +1,44 @@
...
+    #CLIENT_SECRET: #a unique API secret for signing authentication requests
+    #ELASTICSEARCH_INDEX: #the Elasticsearch index name for annotation storage
+    #MAIL_DEFAULT_SENDER: #a sender address for outbound mail
+    #WEBASSETS_BASE_DIR: #the base directory for static assets
+    #WEBASSETS_BASE_URL: #the base URL for static asset routes
+  links:
+   - redis
+   - postgres
+   - elasticsearch
+  command: python -m smtpd -n -c DebuggingServer localhost:25
```
nickstenning: Perhaps I'm missing something, but this means that you're not running Hypothesis at all. I was suggesting earlier that you might want to run the debugging server in another container, but probably not in the hypothesis container.

docker-compose.yml
```diff
@@ -0,0 +1,44 @@
+hypthosis:
```
nickstenning: `s/hypothosis/hypothesis/`

docker-compose.yml
```diff
@@ -0,0 +1,44 @@
...
+  build: .
+  ports:
+   - "5000:5000"
+  environment:
+    DATABASE_URL: "postgresql://hypothesis:mysecretpassword@localhost:5432/hypothesis" #username:password@host:port/database
+    #APP_URL: http://127.0.0.1  #Replace with IP or DNS Name
+    SECRET_KEY: TCtkLQ3k5kNkezsFiPDvq04yG4olCwU6QF7XT6gktGDaEiBZAnxkDJXYCLMfF96W1j+A1PMO+vyfQUgMboEnoQ
```
nickstenning: In development it's probably ok to omit this entirely.

docker-entrypoint-initdb.d/initdb.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
+POSTGRES="gosu postgres"
```
nickstenning: I'm not sure I understand the purpose of this file. Everything that's done in here should be done already by [the startup script in `docker-library/postgres`](https://github.com/docker-library/postgres/blob/master/9.5/docker-entrypoint.sh), except possibly creation of the test database, which we could probably document separately...

...or am I misunderstanding something?

nickstenning:
Closing due to inactivity.

PR #2891, opened 2016-01-25
Don't open "welcome page" when installed by admin policy

### LINKED ISSUE #2890: Opening "welcome" page is unhelpful/confusing in school/office install settings
When the Chrome extension is installed by an administrative policy, as in a school or office setting, the fact that we auto-open the "/welcome" page on install can be unhelpful.

From a user report of this issue:

> The Chromebooks I am working with are shared among multiple users - sometimes 5 or 6 different students in a day. Part of our configuration for these CBs is that all account information is wiped off of the device when a user signs out. As a result, Hypothesis sees every time any user signs into any CB as the "first" time for that user. Suppressing that page load on "first" installation will help immensely.

It seems reasonable that we should default to *not* opening the welcome page when the Chrome extension is installed as a result of an administrative policy rather than user action. The alternative is that a tab will appear to open without any user action.

nickstenning:
If the extension has been installed as the result of an administrative policy rather than user action, it doesn't make a lot of sense (and appears to cause some frustration) to open a tab to our "welcome page."

This PR suppresses the opening of the "welcome page" on first run if the install type is `admin`. Chrome documentation references:

- https://developer.chrome.com/extensions/management#event-onInstalled
- https://developer.chrome.com/extensions/management#type-ExtensionInfo
- https://developer.chrome.com/extensions/management#type-ExtensionInstallType

Fixes #2890.

### DIFF
--- a/h/browser/chrome/lib/hypothesis-chrome-extension.js
+++ b/h/browser/chrome/lib/hypothesis-chrome-extension.js
@@ -79,7 +79,17 @@ function HypothesisChromeExtension(dependencies) {
   };
 
   /* Opens the onboarding page */
-  this.firstRun = function () {
+  this.firstRun = function (installDetails) {
+    // If we've been installed because of an administrative policy, then don't
+    // open the welcome page in a new tab.
+    //
+    // It's safe to assume that if an admin policy is responsible for installing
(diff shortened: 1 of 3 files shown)


robertknight:
LGTM

PR #2892, opened 2016-01-25
Add YAPF config and docs for automatic formatting of Python source code

robertknight:
This adds the (one-line) config setting and instructions in the documentation for using [yapf](https://github.com/google/yapf) to reformat code according to PEP8 rules.

Running `yapf -i <Python source>` will take care of all the basic formatting, although it won't reorder or group imports in PEP8 style. Relevant [upstream issue](https://github.com/google/yapf/issues/209)

From brief testing, YAPF may not eliminate all style violations but it shouldn't introduce any new ones.

### DIFF
--- a/docs/hacking/code-style.rst
+++ b/docs/hacking/code-style.rst
@@ -23,10 +23,15 @@ via the ``lint`` make command::
 Python
 ------
 Strict PEP8_. The project also adheres closely to the
-`Google Python Style Guide`_.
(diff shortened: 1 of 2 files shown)


nickstenning:
This is find, but I'll just note that as previously discussed YAPF isn't perfect, and sometimes it reformats code in pretty stupid ways. By all means treat YAPF output as a guideline, but it's not good enough to be considered a canonical formatting at the moment.

PR #2893, opened 2016-01-25
Fix the /docs/help page

robertknight:
The embed.js template now requires additional data which was
not provided when rendered via the /docs/help route.

Resolve the issue by including the embed in the same way
that we advise users to include it in their own pages.


robertknight:
Might be worth setting up a Behave test for this, speaking of which, there is a fix for the flakey smoke tests [pending review](https://github.com/hypothesis/smokey/pull/7).

nickstenning:
LGTM.

PR #2894, opened 2016-01-26
Double the truncation threshold height for annotations

robertknight:
As per https://trello.com/c/7QtpfMB5/ increase the threshold
so that only very long annotation bodies are truncated.


nickstenning:
CONTROVERSIAL.

nickstenning:
Not.

PR #2901, opened 2016-02-01
Remove the need to specify the Chrome extension ID when building with bundled assets

robertknight:
This simplifies building the Chrome extension and should make things simpler when we have extensions for other browsers in future by removing the need to specify the 'chrome-extension://<ID>' URL when bundling assets.

This URL was previously used to reference scripts and styles in `app.html` and `embed.js`. In `app.html` this is no longer needed following the decoupling of the service URL from the `<base>` URL, so assets can just use root-relative paths (eg. `/public/scripts/...`)

For embed.js, the extension add a `<meta>` tag to the document containing the root URL of the Hypothesis app and embed.js constructs asset URLs relative to this. The reason for using a `<meta>` tag is that this is accessible by scripts running on the page but in isolated worlds.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -74,16 +74,14 @@ build/%-chrome-stage.zip:
 	@rm -rf build/chrome $@
 	hypothesis-buildext chrome \
 		--service 'https://stage.hypothes.is' \
-		--sentry-public-dsn '$(SENTRY_DSN_STAGE)' \
-		--assets 'chrome-extension://iahhmhdkmkifclacffbofcnmgkpalpoj/public'
+		--sentry-public-dsn '$(SENTRY_DSN_STAGE)'
 	@zip -qr $@ build/chrome
 
 build/%-chrome-prod.zip:
 	@rm -rf build/chrome $@
 	hypothesis-buildext chrome \
 		--service 'https://hypothes.is' \
-		--sentry-public-dsn '$(SENTRY_DSN_PROD)' \
-		--assets 'chrome-extension://bjfhmglciegochdpefhhlphglcehbmek/public'
+		--sentry-public-dsn '$(SENTRY_DSN_PROD)'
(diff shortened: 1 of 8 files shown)


nickstenning:
This looks great in principle, but I get `embed.js:17 Uncaught TypeError: Failed to construct 'URL': Invalid base URL` on `/docs/help` now.

nickstenning:
There's also a slight problem (?) where if you build the extension without `--debug` and with `--no-bundle-sidebar` then it will request `.min.js` URLs from the server, which aren't built in development.

Although I'm happy to accept that in this case "build the extension with --debug" is a valid solution.

nickstenning:
Last problem. If I build the extension with:

```
hypothesis-buildext --debug --service https://localhost:5000 --no-bundle-sidebar chrome
```

and then force the assets build with:

```
hypothesis assets conf/development.ini
```

then I still get a 404 from `https://localhost:5000/assets/webassets-external/ad0808dd0f264894e56caee80ae26d21_icomoon.css` when loading the extension.

robertknight:
> This looks great in principle, but I get embed.js:17 Uncaught TypeError:
> Failed to construct 'URL': Invalid base URL on /docs/help now

Fixed

> There's also a slight problem (?) where if you build the extension without --debug 
> ...
> Last problem. If I build the extension with:
> hypothesis-buildext --debug --service https://localhost:5000 --no-bundle-sidebar chrome
> and then force the assets build with:
> hypothesis assets conf/development.ini

Fixed. The problem in both cases was that the URLs generated by webassets in the webassets environment created in buildext.py to render embed.js didn't match those used in the web service.

For now I've changed the build so that embed.js will always reference bundled versions of hypothesis.(css|js) and dependencies regardless of whether app.html is loaded from the Chrome extension or from the service.

nickstenning:
Super! LGTM.

PR #2903, opened 2016-02-02
Send activation emails from a worker process

seanh:
Send new user account activation emails from a worker process, instead of from
the main application thread.

This avoids a problem when the register view is called and sending the email
(mailer.send(message)) raises an exception *but the email was actually sent*.
The transaction would be rolled back and the user account not created, and the
user would receive an unusable "please activate your account" email.

Changes:

- The register view callable now puts a message on a new "activations"
  nsq topic, instead of actually sending the email.

- The new h/accounts/worker.py is a hypothesis-worker that subscribes to the
  new "activations" topic and actually sends an email for each message
  published to the toipic.

- Update accounts/test/views_test.py to fix broken tests.
  These were mainly broken because _register() now accesses
  request.get_queue_writer() and Pyramid's DummyRequest() objects don't have
  that method. Just use a mock.Mock() instead.
  In a couple of places assertions had to be updated as well.

In sending the email fails (if pyramid_mailer's send_immediately() raises) then
the message will be requeued (this is gnsq's default behavior when a worker
function raises an exception) and the worker will try again to send it.

Note that we don't attempt to prevent sending the same activation email
multiple times if a message published to the "activations" topic is received by
the worker multiple times. This may happen, since nsq guarantees
"at least once" delivery. If this becomes a problem we may need to set a flag
in the database to record when a particular activation email has already been
sent successfully.


seanh:
I didn't add any tests for the new worker.py because it's really simple

seanh:
I've added a bit more to this: instead of an "activations" worker for sending account activation emails I added an "email" worker that can be used to send any emails. There's an `h.email.send()` convenience function. Both activation and forgot password emails use it.

Notification emails are different. Activation and forgot password were on the main thread, so I've changed them to use the new worker. Notifications already use their own notifications worker, which sends email in response to the annotations nsq topic. I've left this as is.

h/accounts/test/views_test.py
```diff
@@ -349,7 +349,7 @@ def test_forgot_password_creates_no_activations_when_validation_fails(activation
 @patch('h.accounts.views.reset_password_link')
 @forgot_password_fixtures
 def test_forgot_password_generates_reset_link(reset_link, authn_policy):
-    request = DummyRequest(method='POST')
```
nickstenning: Using a dummy request object is almost always a better idea than using a dumb Mock, because it will raise if we try and access attributes on it that aren't supposed to exist.

You appear to have done this so you can set `request.authenticated_userid` to `None`, but that's already done by the `authn_policy` fixture.

h/accounts/test/views_test.py
```diff
@@ -413,7 +420,7 @@ def test_forgot_password_redirects_on_success(authn_policy):
 
 @pytest.mark.usefixtures('routes_mapper')
 def test_forgot_password_form_redirects_when_logged_in(authn_policy):
-    request = DummyRequest()
+    request = mock.Mock(method="POST", authenticated_userid="jane")
```
nickstenning: You now have two lines next to one another saying different things: `authenticated_userid="jane"` and `authn_policy.authenticated_userid.return_value = "acct:jane@doe.org"`.

h/accounts/test/views_test.py
```diff
@@ -391,10 +393,15 @@ def test_forgot_password_sends_mail(reset_mail, authn_policy, mailer):
     controller = ForgotPasswordController(request)
     controller.form = form_validating_to({"user": user})
     message = reset_mail.return_value
+    reset_mail.return_value = {
+        'recipients': ['giraffe@thezoo.org'], 'subject': 'subject',
```
nickstenning: This looks suspiciously like YAPF-formatted code to me ;)

I find this really hard to read. Any of

1) all on one line
2) 
```python
foo = {
    'bar': 'baz',
    'wibble': 'nonsense',
}
```
3)
```python
foo = {'bar': 'baz',
       'wibble': 'nonsense'}
```
is absolutely fine, but not some mixture of the three!

h/accounts/test/views_test.py
```diff
@@ -1033,3 +1045,11 @@ def activation_model(config, request):
     model = patcher.start()
     request.addfinalizer(patcher.stop)
     return model
+
+
+@pytest.fixture
+def email(config, request):
+    patcher = patch('h.accounts.views.email', autospec=True)
+    model = patcher.start()
```
nickstenning: `s/model/module/` -- it's not a model.

h/accounts/test/views_test.py
```diff
@@ -1033,3 +1045,11 @@ def activation_model(config, request):
     model = patcher.start()
     request.addfinalizer(patcher.stop)
     return model
+
+
+@pytest.fixture
+def email(config, request):
```
nickstenning: `config` isn't used here.

h/accounts/views.py
```diff
@@ -348,7 +345,7 @@ def register(self):
             return {'form': self.form.render()}
 
         self._register(username=appstruct['username'],
-                       email=appstruct['email'],
+                       email_address=appstruct['email'],
```
nickstenning: The field in the appstruct is called `email`, and the keyword argument to the user constructor is called `email`, so passing it through a thing called `email_address` just looks weird.

Perhaps we could call the module `h.mailer`?

nickstenning:
Overall looks pretty good to me! Comments inline.

h/accounts/test/views_test.py
seanh: I've put DummyRequest back

seanh:
Done

h/accounts/test/views_test.py
nickstenning: You still don't need the additional `authenticated_userid` keyword. Or at least you shouldn't. As mentioned above, that's what the `authn_policy` fixture is supposed to be doing.

h/accounts/test/views_test.py
```diff
@@ -413,8 +426,7 @@ def test_forgot_password_redirects_on_success(authn_policy):
 
 @pytest.mark.usefixtures('routes_mapper')
 def test_forgot_password_form_redirects_when_logged_in(authn_policy):
-    request = DummyRequest()
-    authn_policy.authenticated_userid.return_value = "acct:jane@doe.org"
+    request = DummyRequest(authenticated_userid="acct:jane@doe.org")
```
nickstenning: Why is this change needed? If the `authn_policy` fixture isn't working somehow, then let's fix it. But it should definitely work here...

nickstenning:
Only the confusion about use of the `authn_policy` fixture to resolve and then I'm very happy for this to merge.

nickstenning:
*N.B.* There will be a merge resolution needed between this PR and #2913.

nickstenning:
LGTM!

PR #2904, opened 2016-02-02
Don't 404 if a user visits an activation link twice

seanh:
When a user activates their account we delete the row from the activations
database table. If the user then tries to visit an activation link from an
activation email that we sent them, after their account has already been
activated, then we can no longer find the activation row in the db so we 404.

This confuses some users.

Instead, redirect them to the front page (as we do after a successful
activation) with a flash message suggesting that they may already be activated
and should try signing in.

Note that if the user tries to visit a genuinely wrong activation link they
will get this same redirect and suggestion - we don't distinguish between old
activation links and incorrect ones.

### DIFF
--- a/h/accounts/test/views_test.py
+++ b/h/accounts/test/views_test.py
@@ -648,22 +648,6 @@ activate_fixtures = pytest.mark.usefixtures('ActivationEvent',
 
 
 @activate_fixtures
-def test_activate_404s_if_code_missing():
-    request = DummyRequest(matchdict={'id': '123'})  # No 'code' in matchdict.
-
-    with pytest.raises(httpexceptions.HTTPNotFound):
-        RegisterController(request).activate()
-
-
-@activate_fixtures
-def test_activate_404s_if_id_missing():
-    request = DummyRequest(matchdict={'code': 'abc123'})  # No 'id'.
-
-    with pytest.raises(httpexceptions.HTTPNotFound):
-        RegisterController(request).activate()
-
-
-@activate_fixtures
 def test_activate_404s_if_id_not_int():
     request = DummyRequest(matchdict={
         'id': 'abc',  # Not an int.
--- a/h/accounts/test/views_test.py
+++ b/h/accounts/test/views_test.py
@@ -640,49 +640,64 @@ def test_register_form_redirects_when_logged_in(authn_policy):
         controller.register_form()
 
 
-activate_fixtures = pytest.mark.usefixtures('activation_model',
+activate_fixtures = pytest.mark.usefixtures('ActivationEvent',
+                                            'activation_model',
                                             'notify',
                                             'routes_mapper',
                                             'user_model')
 
-@activate_fixtures
-def test_activate_returns_not_found_if_code_missing():
-    request = DummyRequest(matchdict={'id': '123'})
-
-    result = RegisterController(request).activate()
-
-    assert isinstance(result, httpexceptions.HTTPNotFound)
-
 
 @activate_fixtures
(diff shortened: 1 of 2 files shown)


judell:
This will be really helpful, thank you.

h/accounts/models.py
```diff
@@ -167,6 +167,10 @@ def _hash_password(self, password):
         return text_type(CRYPT.encode(password + self.salt))
 
     @classmethod
+    def get_by_id(cls, id_):
+        return cls.query.filter(cls.id == id_).first()
```
nickstenning: I'm not really convinced we need a `get_by_id` class method. But even if we do, it should do `return cls.query.get(id_)`.

h/accounts/test/views_test.py
```diff
@@ -634,14 +634,13 @@ def test_register_form_redirects_when_logged_in(authn_policy):
...
-                                            'user_model')
+activate_fixtures = pytest.mark.usefixtures(
+    'activation_model', 'ActivationEvent', 'user_model')
+
 
 @activate_fixtures
-def test_activate_returns_not_found_if_code_missing():
-    request = DummyRequest(matchdict={'id': '123'})
+def test_activate_404s_if_code_missing():
+    request = mock.Mock(matchdict={'id': '123'})  # No 'code' in matchdict.
```
nickstenning: What on earth do you have against `DummyRequest`!? :stuck_out_tongue:

nickstenning:
Ok, a couple of comments:

1. I don't think you need to distinguish between users that exist and users that don't -- indeed I'd rather we didn't, because currently this code allows us to trivially enumerate user ids and determine whether users exist or not (at least if they're activated).

2. This isn't your fault, as you're only doing what's already there, but you need to `raise` 404s, not `return` them, otherwise you'll get the ugly 404 styling instead of the correct exception-view-handled 404 styles.

3. I think "Perhaps you've already activated your account?" is a bit abrupt in this context -- they don't know that, almost by definition. Maybe something like:

  > We didn't recognise that activation link. Perhaps you've already activated your account? If so, try signing in] using...

4. If we're going to say "sign in" then perhaps take this opportunity to tweak the flash message displayed for successful activations, too. It currently says "login".

h/accounts/models.py
seanh: Lazy copy-paste, but it looks like this might be getting deleted anyway

h/accounts/test/views_test.py
seanh: Ok, I've put DummyRequest back and done the necessary stuff to get it working

seanh:
Ok, changes made and rebased.

A lot of what I did was changing the `activate()` logic to do the friendly redirect only if the activation code is wrong but the id does match an activated user, and rewriting the tests to match that. With the check for the id matching an activated user now removed the logic change is much simpler and much of the test changes I had done not necessary. I've split it up into new commits that should be logical enough.

One thing I've noticed: we show the _"We didn't recognize that activation link. Perhaps you've already activated your account? If so, try signing in using the username and password that you provided"_ message even if they visit a bad activation link while already signed in. This can be confusing because the _signing in_ link in the message will just take them straight to the stream! We could add logic to deal with these cases:

- User is already signed in to the account matching the id in the URL. Redirect to front page with "you're activated and logged in" message.
- User is signed in to a different account. Redirect to front page with "you're signed in to a different account, log out and try again."

seanh:
Handled those two cases.

I've tested it but recommend you test manually as well, lots of cases here.

seanh:
Done

nickstenning:
LGTM.

PR #2905, opened 2016-02-03
Group page update

lenazun:
This includes:
- Adding a link to a search by group ID under the group name
- Creating a new div to display the document list, aligned to the left
- Style changes to do all of that.

screenshot from 2016-02-02 17 36 13

screenshot from 2016-02-02 17 36 45

### DIFF
--- a/h/static/styles/group-form.scss
+++ b/h/static/styles/group-form.scss
@@ -19,7 +19,6 @@
     @include flex-center-column();
 
     font-size: $body2-font-size;
-    min-height: 280px;
     background-color: $white;
     border: $border;
     color: $color-dove-gray;
@@ -50,6 +49,16 @@
       margin-bottom: 40px;
     }
 
+    &__heading--short {
+      font-size: $title-font-size;
+      font-weight: $title-font-weight;
+      margin-bottom: 5px;
+    }
+
+    &__nocontent-text {
+      padding-top: 10px;
+    }
+
     .form-field {
       @include flex-center-column();
     }
@@ -76,12 +85,30 @@
     }
   }
 
-  .group-form-document-list {
-    text-align:center;
-  }
+  .group-document-list {
 
-  .group-form-document-list li {
-    margin-top: 1em;
+    justify-content: left;
+    font-size: $body2-font-size;
+    line-height: $body2-line-height;
+    background-color: $white;
+    color: $color-dove-gray;
+
+    border: $border;
+    border-top: none;
+    padding-top: 20px;
+    padding-bottom: 10px;
+    padding-left: 20px;
+    padding-right: 20px;
+
+    &__heading {
+      font-weight: bold;
+      margin-bottom: 1em;
+    }
+
+    &__list li {
+      margin-bottom: 1.5em;
+    }
+
   }
 
   .group-form-footer {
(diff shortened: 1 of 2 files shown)


h/templates/groups/share.html.jinja2
```diff
@@ -12,19 +12,24 @@
   <div class="content content--narrow">
     <div class="group-form is-member-of-group">
       <i class="h-icon-group group-form__heading-icon"></i>
-      <div class="group-form__heading">{{ group.name }}</div>
+      <div class="group-form__heading-short">{{ group.name }}</div>
 
       {% if document_links %}
-        Recently annotated documents:
-        <ul class="group-form-document-list">
+        <a href="//hypothes.is/stream?q=group:{{ group.pubid }}"
```
robertknight: This URL will point to the wrong host when using a development server. You can get the correct host using `{{ request.route_url('stream') }}?q=group:{{ group.pubid }}`

h/static/styles/group-form.scss
```diff
@@ -50,6 +49,12 @@
       margin-bottom: 40px;
     }
 
+    &__heading-short {
```
robertknight: If you introduce a variation of an element, the convention is to use '--modifier-name', so `&__heading--short` in this case.

robertknight:
Couple of minor corrections but looking good :)

conordelahunty:
Looking good Lena, I have one small thing which is can you left align the 'group documents' with the 'About Groups'. There seems to be a difference of 20px or so.

lenazun:
Thank you both!  I've fixed the issues. @conordelahunty this is what it looks like now:

screenshot from 2016-02-03 10 09 12

h/templates/groups/share.html.jinja2
```diff
@@ -12,19 +12,24 @@
...
+      <div class="group-form__heading--short">{{ group.name }}</div>
 
       {% if document_links %}
-        Recently annotated documents:
-        <ul class="group-form-document-list">
+        <a href="{{ request.route_url('stream') }}?q=group:{{ group.pubid }}"
+           target="_blank"
+           title="Recent annotations in this group">
+           View recent group annotations</a>
+    </div>
```
robertknight: Closing an already open tag inside the `{% if document_links %}` block makes this a bit confusing to read. I'd suggest restructuring this so that tags opened and closed inside `{%if %} ... {% endif %}` paris are balanced - that might mean having two `{%if document_links %}` blocks.

lenazun: Thanks @robertknight ! I made some changes to make it more readable.  Note that the `{% if %}...{% endif %}` tags are outside the second div, since if there's no documents, that div shouldn't appear at all.  Let me know if there are better ways to do it!

robertknight:
This is looking great @lenazun - One last minor thing to do is squash the commit into one or two commits with `git rebase` to get a clean history, where commit messages take the form:

```
<Short one line summary>

<Brief description of what changed and more importantly why.
  In this case since it is a design change you don't need a "why"
  section. A link to the Trello card is useful if you have one>
```

Then I can get this merged.

conordelahunty:
Thank you Lena

nickstenning:
I've taken the liberty of rebasing this into a single commit so we can merge it and get it out later today.

nickstenning:
:tada: NICE WORK LENA.

PR #2906, opened 2016-02-03
Make connecting the WebSocket optional

nickstenning:
In preparation for splitting the hosting of the WebSocket server into a separate process, make it possible to run the client without attempting to connect to the WebSocket. This may make developing locally easier when not working with the realtime parts of the client code.

If the `websocketUrl` setting is missing, we will not attempt to connect.


robertknight:
LGTM

PR #2907, opened 2016-02-03
Admin account deletion

chdorner:
As discussed with @nickstenning and @lenazun, this is the minimal viable solution for deleting a user from the admin interface.

It does not allow to delete a user that is a group creator, because of the foreign key constraint on `group.creator_id`, and there are plans on using a group creator as a group admin in the future.

This will delete the following objects in order:

1. Group memberships
2. Annotations
3. User

Since this is the minimal viable solution, we are deleting all the objects during the request and not in a background worker. This means that users with lots of annotations should still be deleted in a shell on the server.
However, to make that easier I've put the relevant logic into its own function, so you would only need to run the following:

```python
>>> from h.accounts.models import User
>>> from h import admin
>>> user = User.get_by_username(u'bobthebuilder')
>>> admin.delete_user(request, user)
>>> request.tm.commit()
```

_As I'm new to H and Python, please point out even the smallest things, as it is the best way to get used to Pythonland._

### DIFF
--- a/h/admin.py
+++ b/h/admin.py
@@ -290,9 +290,8 @@ def delete_user(request, user):
     message.
     """
 
-    for group in user.groups:
-        if group.creator == user:
-            raise UserDeletionError('Cannot delete user who is a group creator.')
+    if models.Group.created_by(user).count() > 0:
+        raise UserDeletionError('Cannot delete user who is a group creator.')
 
     user.groups = []
 
--- a/h/admin.py
+++ b/h/admin.py
@@ -172,8 +172,10 @@ def users_index(request):
     user_meta = {}
     username = request.params.get('username')
 
-    if username is not None:
+    if username:
         user = models.User.get_by_username(username)
+        if user is None:
+            user = models.User.get_by_email(username)
 
     if user is not None:
         # Fetch information on how many annotations the user has created
--- a/h/admin.py
+++ b/h/admin.py
@@ -3,12 +3,19 @@ from pyramid import session
 from pyramid import view
 from pyramid import httpexceptions
 
+from elasticsearch import helpers as es_helpers
+
 from h.api import nipsa
+from h.api import storage
 from h.i18n import TranslationString as _
 from h import accounts
 from h import models
 from h import paginator
 from h import util
+
+
+class UserDeletionError(Exception):
+    pass
 
 
 @view.view_config(route_name='admin_index',
@@ -251,6 +258,28 @@ def groups_index_csv(request):
     request.response.content_disposition = 'attachment;filename=' + filename
 
     return {'header': header, 'rows': rows}
+
+
+def delete_user(request, user):
+    """
+    Deletes a user with all their group memberships and annotations.
+
+    Raises UserDeletionError when deletion fails with the appropriate error
+    message.
+    """
+    for group in user.groups:
+        if group.creator == user:
+            raise UserDeletionError('Cannot delete user who is a group creator.')
+
+    user.groups = []
+
+    userid = util.userid_from_username(user.username, request)
+    query = _all_user_annotations_query(userid)
+    annotations = es_helpers.scan(client=request.es.conn, query={'query': query})
+    for annotation in annotations:
+        storage.delete_annotation(annotation['_id'])
+
+    request.db.delete(user)
 
 
 def _all_user_annotations_query(userid):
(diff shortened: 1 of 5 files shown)


h/templates/admin/users.html.jinja2
```diff
@@ -43,8 +43,34 @@
           </tr>
         </tbody>
       </table>
+
+      <form id="users-delete-form" method="POST" action="{{request.route_path('admin_users_delete')}}" class="form-inline">
+        <input type="hidden" name="username" value="{{user.username}}">
+
+        {% if user_meta['annotations_count'] > 100 %}
```
chdorner: This is quite an arbitrarily set number, since I'm going through the `api.storage` module, deleting each annotation results in two requests to Elasticsearch.

h/test/admin_test.py
```diff
@@ -492,9 +506,55 @@ def test_users_index_user_found(User):
...
+    'delete_user')
 
-badge_index_fixtures = pytest.mark.usefixtures('models')
+@users_delete_fixtures
+def test_users_delete_redirect(User):
+    request = DummyRequest(params={"username": "bob"})
+    User.get_by_username.return_value = None
 
+    result = admin.users_delete(request)
+    assert result.__class__ == httpexceptions.HTTPFound
```
chdorner: It would be nice to test the location of the redirect as well, but it seems like the `routes_mapper` fixtures makes all `route_path` calls return a dummy path. The other solution would be to assert calls to `route_path`. Any thoughts?

h/templates/admin/users.html.jinja2
nickstenning: :+1: looks fine to me.

h/admin.py
```diff
@@ -186,6 +191,24 @@ def users_index(request):
 
     return {'username': username, 'user': user, 'user_meta': user_meta}
 
+@view.view_config(route_name='admin_users_delete',
```
nickstenning: Convention is 2 blank lines between functions, as here, although it's 1 blank line between methods in classes, perversely.

h/admin.py
```diff
@@ -186,6 +191,24 @@ def users_index(request):
...
+                 request_method='POST',
```
nickstenning: If I were a pernickety asshole, I'd suggest lining up the first character of `request` with the first character of `route` on the line above. As it happens, I'm a pernickety asshole. :)

h/admin.py
```diff
@@ -250,6 +273,31 @@ def groups_index_csv(request):
 
     return {'header': header, 'rows': rows}
 
+class UserDeletionError(Exception):
```
nickstenning: Shove this up at the top of the file. Typically I'd try and keep a file ordered something like this:

```
imports
CONSTANTS
Exceptions
Classes
functions
_private_functions
```

h/admin.py
```diff
@@ -260,7 +308,6 @@ def _all_user_annotations_query(userid):
         }
     }
 
-
```
nickstenning: As mentioned earlier, this two-blank-lines things is weird, but it is the convention...

h/admin.py
```diff
@@ -250,6 +273,31 @@ def groups_index_csv(request):
...
+    """
+    Deletes a user with all their group memberships and annotations.
+
+    Raises UserDeletionError when deletion fails with the appropriate error
+    message.
+    """
+
+    for group in user.groups:
+        if group.creator == user:
+            raise UserDeletionError('Cannot delete user who is a group creator.')
```
nickstenning: Ah, now I think about it, there's a slight problem with this, which is that at the moment it's possible for someone to be a creator of a group that they've subsequently left. That is, they can be a creator of a group that isn't in `user.groups`.

So the condition probably wants to be `Group.query.filter(Group.creator == user).count() > 0`.

chdorner: This is fixed in 98291693671f154ee4e119ab9948d0e661b014f1

chdorner:
I made the style changes into their respective commits. The new way of figuring out if a user is a group creator deserves its own commit, that is in 98291693671f154ee4e119ab9948d0e661b014f1.

nickstenning:
LGTM.

nickstenning:
First PR :cake: :sparkles: :rocket:

chdorner:
:tada:

judell:
I annotated the 2 lines of code that implement lookup by email address, and
added a thank-you note from the support team!

PR #2908, opened 2016-02-03
Fix flash of 'Create Account' banner in stream

robertknight:
Add 'ng-cloak' attribute so that 'Create Account' banner
is hidden until the app is loaded.


nickstenning:
LGTM.

PR #2910, opened 2016-02-03
Fix client-side router on site

### LINKED ISSUE #2867: "Tell me more about groups" link is broken
1. Go to https://hypothes.is/groups/new
2. Click "Tell me more about groups"

Expected: Details about how to use groups appears
Actual: Nothing happens

The bug is something I found when running the TypeScript compiler over our JS code, which spotted an incorrect use of `RegExp.match()` in https://github.com/hypothesis/h/blob/master/h/static/scripts/site.js#L26

robertknight:
The JS for the group share page failed to load
because of an incorrect comparison of the result
of String.match() against a number and also
because of an attempt to load Bootstrap
unnecessarily on this pages.

This commit fixes the issue and prepares us for
adding new pages in future by:

 * Adding a client-side micro-router to simplify running
   page-specific JS

 * Only loading the Bootstrap components on pages that
   actually need it.

Fixes #2867

### DIFF
--- a/h/static/scripts/share-group-form.js
+++ b/h/static/scripts/share-group-form.js
@@ -1,4 +1,7 @@
 function ShareGroupFormController(element) {
+  if (!element.querySelector('.is-member-of-group')) {
+    return;
+  }
   var shareLink = element.querySelector('.js-share-link');
   shareLink.focus();
   shareLink.select();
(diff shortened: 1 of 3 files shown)


nickstenning:
LGTM.

PR #2913, opened 2016-02-03
Standalone websocket server

nickstenning:
Make it possible to run the websocket server on its own, separate from the rest of the application code.

This adds a paste entrypoint called "websocket" and a pair of config files (for production and development) at

    conf/websocket.ini
    conf/development-websocket.ini

These allow you to run a standalone websocket server (listening by default on port 5001) using, for example:

    gunicorn --paste conf/websocket.ini

This PR also changes the Dockerfile build to make it easier to deploy this in production -- by default the Dockerfile will run just the web application, rather than the web application and a selection of workers.


robertknight:
I've tested this locally with both the extension and the stream and verified that the app does not error if no websocket URL is provided in the config.

Server-side code looks fine, although I've only limited familiarity with our Docker setup and Pyramid configuration.

nickstenning:
At @chdorner's suggestion I've switched to `app.ini`, `development-app.ini` and `websocket.ini`, `development-websocket.ini` as being a bit more tab-completion friendly.

PR #2916, opened 2016-02-04
Add automated retry with backoff for session loads

robertknight:
If a session load fails, subsequent calls to
session.load() would trigger another request immediately.
Together with feature flag checks triggering session.load(),
this could result in a cycle if no network connection was present:

 1. App loads during initial digest cycle, 'flag.featureEnabled()'
    is called to test whether to show certain UI elements
 2. session.load() is triggered
 3. session.load() fails and the HTTP response error triggers
    a digest cycle
 4. GOTO 1

This commit resolves the problem by automatically retrying
session.load() calls with an exponential backoff.

A side _benefit_ of this is that if you activate the Chrome extension when you have no
network connectivity, the app will defer loading until the network becomes available.
Previously if the `/app` request failed, the sidebar loaded but displayed in the default logged-out state.

A downside is that no error indication is currently shown to indicate why the app remains in a loading state.

### DIFF
--- a/h/browser/chrome/test/uri-info-test.js
+++ b/h/browser/chrome/test/uri-info-test.js
@@ -1,20 +1,7 @@
 var uriInfo = require('../lib/uri-info');
 var settings = require('./settings.json');
 
-/**
- * Takes a Promise<T> and returns a Promise<Result>
- * where Result = { result: T } | { error: any }.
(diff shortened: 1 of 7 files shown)


nickstenning:
LGTM.

PR #2918, opened 2016-02-05
Don't autofocus the group share link

### LINKED ISSUE #2917: Share link autofocus on group homepage is confusing
We currently autofocus the share link on group homepages. This is fine when the page is short, but rather confusing when the group has been busy annotating lots of pages -- the page will jump down to the bottom of the page when loading.

We probably want to reconsider this behaviour.

(If you need an example groups page, ask me.)

nickstenning:
Unfortunately, on group pages for groups who have annotated lots of documents, this results in a rather confusing jump to the bottom of the page.

Fixes #2917.


robertknight:
LGTM. I think it would still be useful to auto-select the text to make copying the link easier, but only when the input fields is focused, not on page load. That can be done separately though and rather than making it part of a page-specific controller, it could just be generic functionality enabled by a "js-select-on-focus" class or attribute on an element.

PR #2920, opened 2016-02-08
Always count annotations with the correct userid

nickstenning:
Searching by email address would result in a query of the form

   ...{'term': {'user': {'foo@gmail.com@hypothes.is'}}}...

This commit ensures that we always count annotations using the correct userid.

I nominate @chdorner for review seeing as he touched this code last.


chdorner:
That is a very good point! Looks good to me.

PR #2922, opened 2016-02-08
Add a PENDING_REMOVAL bucket for features on their way out

nickstenning:
This commit adds a FEATURES_PENDING_REMOVAL bucket into which we can put feature flags which are in the process of being removed from the codebase. This:

- makes it immediately obvious to programmers whether or not a feature is in use when looking at `h/features.py`
- hides such features from the admin console
- allows the application to error or warn when feature flag lookups haven't been correctly removed from the application


finitud:
LGTM :+1:

PR #2923, opened 2016-02-08
Simplify server side auth code

seanh:


h/api/views.py
```diff
@@ -157,8 +162,28 @@ def search(request):
...
+    request.
+
+    """
+    try:
+        session.check_csrf_token(request, token='assertion')
+    except exceptions.BadCSRFToken:
+        raise httpexceptions.HTTPUnauthorized
+
+    expires_in = 3600
+    response = pyramid.response.Response(
```
nickstenning: The `api_config` decorator implies a JSON response type already, so you can skip the explicit construction of the response, and the `json.dumps` and just return a dictionary.

nickstenning: That is, if you just move this whole thing into `annotator_token` and get rid of `access_token`.

h/auth.py
```diff
@@ -158,9 +76,8 @@ def groupfinder(userid, request):
     return list(principals)
 
 
-def effective_principals(userid, request, groupfinder=groupfinder):
-    """
-    Return the list of effective principals for the passed userid.
+def effective_principals(userid, request, groupfinder_=groupfinder):
```
nickstenning: There's now no need for separate `groupfinder` and `effective_principals` functions. You can just merge the functionality of `groupfinder` directly into `effective_principals`, and make the `groupfinder` tests into `effective_principals` tests.

requirements.txt
```diff
@@ -2,7 +2,7 @@
 certifi
 cffi
 gunicorn
-pyramid_multiauth
+honcho
```
nickstenning: This looks like a rebase error. There should be no need for `honcho` in `requirements.txt` any more.

h/api/views.py
seanh: Done

seanh:
Ok, after a lot of rebasing the diff and commit history are looking reasonable.

I still have to fix the `hypothesis token` command.

seanh:
Ok, done. I think that's everything, but I'll give it another check tomorrow morning before review,

h/api/auth.py
```diff
@@ -19,3 +22,96 @@ def translate_annotation_principals(principals):
...
+
+    :returns: the userid authenticated by the request's JWT or None
+    :rtype: unicode or None
+
+    """
+    if 'Authorization' not in request.headers:
+        return None
+
+    try:
+        return jwt.decode(request.headers.get('Authorization')[7:],
```
nickstenning: It's not immediately clear what `[7:]` is doing. Perhaps better to be explicit here?

```
authz_header = request.headers.get('Authorization')

if authz_header is None:
    return None

if not authz_header.startswith('Bearer '):
    return None

token = authz_header[len('Bearer '):]
...
```

h/api/auth.py
```diff
@@ -19,3 +22,96 @@ def translate_annotation_principals(principals):
...
+                          key=request.registry.settings['h.client_secret'],
+                          audience=request.host_url,
+                          leeway=240,
+                          algorithm='HS256').get('sub')
```
nickstenning: This should be `algorithms=['HS256']` I think.

h/api/auth.py
```diff
@@ -19,3 +22,96 @@ def translate_annotation_principals(principals):
...
+    except jwt.InvalidTokenError:
+        return None
+
+
+def unauthenticated_userid_from_bearer_token(request):
```
nickstenning: As mentioned earlier I'm not sure there's much point in doing this. I don't think we actually need or use `unauthenticated_userid` anywhere. I'm happy for it to delegate to `authenticated_userid` in the authentication policy.

h/api/test/auth_test.py
```diff
@@ -28,3 +31,209 @@ def test_translate_annotation_principals(p_in, p_out):
...
+        'h.client_id': 'id',
+        'h.client_secret': 'secret'
+    }
+    if token:
+        request.headers = {'Authorization': token}
+    return request
+
+
+@pytest.fixture
+def encode(request):
```
nickstenning: I think these tests would probably be clearer if you just had one fixture for the `jwt` package rather than separate fixtures for `encode` and `decode`.

h/api/test/auth_test.py
```diff
@@ -28,3 +31,209 @@ def test_translate_annotation_principals(p_in, p_out):
...
+    request = mock_request()
+
+    auth.generate_bearer_token(request, 3600)
+
+    assert encode.call_args[0][0]['aud'] == request.host_url
+
+
+@generate_bearer_token_fixtures
+def test_generate_bearer_token_passes_algorithm_to_encode(encode):
+    """It should encode request.host_url as 'aud'."""
```
nickstenning: This docstring doesn't match the test. 

Overall I'm not sure we need to test the passing of the various claims quite so laboriously. One test should suffice to verify the claim dictionary.

h/api/test/auth_test.py
```diff
@@ -28,3 +31,209 @@ def test_translate_annotation_principals(p_in, p_out):
...
+
+@unauthenticated_userid_from_bearer_token_fixtures
+def test_unauthenticated_userid_from_bearer_token_when_decode_fails(decode):
+    request = mock_request(token='Bearer abc123')
+    decode.side_effect = jwt.InvalidTokenError
+
+    result = auth.unauthenticated_userid_from_bearer_token(request)
+
+    assert result is None
+
```
nickstenning: It might be nice to have a test of the one important invariant:

```
userid == decode(encode(userid))
```

h/script.py
```diff
@@ -197,19 +198,15 @@ def token(args):
...
     request = bootstrap(args)
-    registry = request.registry
-
-    request.client = get_client(request, registry.settings['h.client_id'])
-    request.user = args.sub
-    request.expires_in = args.ttl
-    request.extra_credentials = {}
-
-    token = generate_signed_token(request)
+    mock_request = mock.Mock(
```
nickstenning: Using `mock` here is a bit weird. If you have to use a fake request object, perhaps use a `namedtuple` instead.

h/auth.py
```diff
@@ -171,116 +67,30 @@ def effective_principals(userid, request, groupfinder=groupfinder):
...
-                                 config.object_description(factory),
-                                 'client factory')
-    intr['factory'] = factory
-    config.action(IClientFactory, register, introspectables=(intr,))
+    return list(principals)
 
 
-validator = RequestValidator()
+def is_api_request(request):
+    return request.path.startswith('/api') and request.path != '/api/token'
```
nickstenning: We probably also need to exclude `/api/badge` here for the time being.

nickstenning:
Looking pretty good! A few comments inline.

h/api/auth.py
seanh: Done, although I didn't see your sample code, but I did something similar.

h/api/auth.py
seanh: Done

h/api/auth.py
seanh: Removed

h/api/test/auth_test.py
seanh: Normally I would do that (normally I prefer to patch top-level names in the module under test, and name the fixture exactly the same as the name being patched in the module) but in this case the tests need to `import jwt` in order to get `jwt.InvalidTokenError` so a fixture named `jwt` would clash. Rather than make up a name like `jwt_fixture` or something I just patched `encode` and `decode` instead.

h/api/test/auth_test.py
seanh: Fixed

h/api/test/auth_test.py
seanh: Yeah, I'll add that. Actually originally with all the `encode()` tests, instead of testing that `encode()` was called with the right 'sub' for example, I would take the result of `encode()` and decode it (by calling jwt directly in the tests) and test that it decoded to the right sub.

Similarly with the `decode()` tests, I called jwt directly to encode something and pass it in.

In the way that seemed like a better way to test this to me as it tests that it really succeeds and fails when we think it does. But it's not how most of our unit tests work. and it was testing the jwt lib.

I still feel a bit unsure about this kind of "it should call library foo with the params that I think it should" kind of tests. Yes, but what if I misread the foo docs and the params that I think it should be using don't actually cause the foo lib to do what I think it does?

Anyway, I'll add the one invariant test.

h/api/test/auth_test.py
nickstenning: Doing `from jwt import InvalidTokenError` won't actually make the world explode ;)

h/script.py
seanh: But I love mock! _So_ easy and useful. :) (Done)

h/auth.py
seanh: Done

seanh:
@nickstenning Comments responded to

h/api/test/auth_test.py
seanh: Done

seanh:
@nickstenning Ok I squashed those tests. Finished now

h/api/test/auth_test.py
```diff
@@ -28,3 +31,150 @@ def test_translate_annotation_principals(p_in, p_out):
...
+def test_generate_bearer_token_calls_encode(jwt):
+    """It should pass the right arguments to encode()."""
+    before = datetime.datetime.utcnow()
+    request = mock_request()
+
+    auth.generate_bearer_token(request, 3600)
+
+    assert jwt.encode.call_args[0][0]['sub'] == 'acct:testuser@hypothes.is', (
+        "It should encode the userid as 'sub'")
+    after = datetime.datetime.utcnow() + datetime.timedelta(seconds=3600)
```
nickstenning: This is fine, and I'll merge it, but here's another way of writing this test that personally I think would be much clearer and easier to maintain:

```python
class ApproximateMatcher(object):
    def __init__(self, value, delta):
        self.value = value
        self.delta = delta
    def __eq__(self, other):
        return self.value - self.delta < other < self.value + self.delta

def approximate_datetime(when):
    return ApproximateMatcher(when, delta=datetime.timedelta(seconds=2))
...

jwt.encode.assert_called_once_with({
    'sub': 'acct:testuser@hypothes.is',
    'iss': 'expected-client-id',
    'aud': 'expected-audience',
    'iat': approximate_datetime(now),
    'exp': approximate_datetime(now + datetime.timedelta(seconds=3600)),
}, secret, algorithm='HS256')
```

seanh: Hmm. I like the reusability of `approximate_matcher()`. But if `generate_bearer_token()` took longer than two seconds to execute that would fail. Not that I think it's likely to take that long, but I like the fact that `before` is definitely a time before it was called, and `after` + 3600 seconds should definitely be a time after the expiry time, it's not dependent on any timing of how long things take to run.

What's hard to maintain about it?

nickstenning: It's not about the exact implementation of `ApproximateMatcher` -- you could easily implement it to do exactly the same assertion you're doing in your version.

It's about the difference between "assert it made this call" and `assert ...call_args[0][0].get('sub') == 'foo'`, which doesn't map easily onto how you're thinking about that call when writing the code under test.

nickstenning:
This is absolutely brilliant.

For those of you following at home, the diff says +402 -832, but that's only half the story. The majority of code added is in tests, some of which are testing things that weren't tested before.

But even more impressively, we've just dropped close to 8000 lines of dependencies, depending on how you count, without adversely affecting the behaviour of Hypothesis. In fact, we're now in a much happier place where the API doesn't mysteriously accept cookie authentication without ever being documented to do so.

Great work, Sean!

h/api/test/auth_test.py
seanh: Yeah, that would probably be better

PR #2925, opened 2016-02-08
Fix base URL not being specified for /app/dismiss_sidebar_tutorial route

robertknight:
The dismiss_sidebar_tutorial session action overrode the whole
URL, including replacing the base URL with just '/app'. Therefore
it would have failed when app.html's origin did not match the service
URL - as in the Chrome extension.

 * Pass the complete URL for the dismiss_sidebar_tutorial action

 * Remove several unused actions from session.js that have been
   replaced by server-rendered forms

 * Add test for dimiss_sidebar_tutorial action

### DIFF
--- a/h/static/scripts/session.js
+++ b/h/static/scripts/session.js
@@ -1,5 +1,6 @@
 'use strict';
 
+var assign = require('core-js/library/fn/object/assign');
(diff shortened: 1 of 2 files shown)


nickstenning:
LGTM.

PR #2926, opened 2016-02-09
Specify the minimum Chrome version in the manifest

robertknight:
We've had several reports of users who were confused by ungraceful
behavior when the extension failed to load under old versions
of Chrome.

This sets the minimum version in the manifest so that users will
get a clear indication if they try to install the extension on an unsupported
version of Chrome.

The current minimum version that our extension will actually work
on is Chrome 34 (that I've tested) but in the interests of defining a simple ongoing
support policy which maps to a clear number of regular releases
and a well-defined time frame, I've rounded that to
'the last 10 stable releases' which makes the minimum version Chrome 38.

We can revise this if necessary.

Card CXkq74c7


nickstenning:
LGTM.

PR #2933, opened 2016-02-10
Fail more gracefully if user tries to activate H on a page served by a Chrome extension

robertknight:
If a user tries to activate H on a page served by a Chrome extension, such as a custom PDF viewer, the embedded viewer was loaded and displayed an unhelpful error.

This PR restores the intended behavior where the user sees an exclamation mark in the badge and clicking on that displays a page explaining the limitations of the extension with respect to annotating non-HTTP pages. I've also added a specific mention of this case to the help text.

In the specific case of a custom PDF viewer, we could often do better by extracting the original URL from the tab URL, which is typically of the form `chrome-extension://<id>/viewer?file=<original URL>` and opening that directly in the Hypothesis PDF viewer. That will require some additional work and does not cover all cases, so I've left that as a separate task.

### DIFF
--- a/h/browser/chrome/test/sidebar-injector-test.js
+++ b/h/browser/chrome/test/sidebar-injector-test.js
@@ -12,6 +12,10 @@ function createTestFrame() {
   return frame;
 }
 
+// The root URL for the extension returned by the
+// extensionURL(path) fake
+var EXTENSION_BASE_URL = 'chrome-extension://hypothesis';
+
 describe('SidebarInjector', function () {
   'use strict';
 
@@ -60,7 +64,7 @@ describe('SidebarInjector', function () {
     injector = new SidebarInjector(fakeChromeTabs, {
       isAllowedFileSchemeAccess: fakeFileAccess,
       extensionURL: sinon.spy(function (path) {
-        return 'CRX_PATH' + path;
+        return EXTENSION_BASE_URL + path;
       })
     });
   });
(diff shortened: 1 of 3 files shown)


judell:
Thanks for this!

I'll confess that while I've seen the exclamation mark in the badge a few times, I may never have clicked it, and certainly did not realize the resulting page was configurable to show different explanations for different circumstances. I've elsewhere noted that may be a useful Q&A for the FAQ.

h/browser/chrome/test/sidebar-injector-test.js
```diff
@@ -78,21 +82,33 @@ describe('SidebarInjector', function () {
...
+            assertReject, function (err) {
+              assert.instanceOf(err, errors.RestrictedProtocolError);
+              assert.notCalled(spy);
+            }
+          );
+        });
     });
 
+    it('succeeds if the tab is already displaying the embedded PDF viewer',
+      function () {
```
nickstenning: Small thing, but we should probably be consistent about:

```javascript
it('does a thing', function () {
  ...
});
```

vs

```javascript
it('does a thing', 
  function () {
    ...
  }
);
```

I'm not sure if you did this because of line length? If so, *seriously* don't worry about it. Especially in tests.

robertknight: Will do. Line length was the only reason.

robertknight: Done

nickstenning:
LGTM.

PR #2936, opened 2016-02-10
Deduplicate the profile pages

seanh:
Add layouts/profile.html.jinja2 for each profile template in
templates/accounts/ to extend, instead of duplicating code in each
template.


nickstenning:
LGTM. Thanks for that!

PR #2937, opened 2016-02-10
Postgres Annotations

chdorner:
This PR includes changes for the first part of migrating annotation data from Elasticsearch to Postgres.

__Mapping Elasticsearch annotation data to Postgres annotations__

- The `target` field from Elasticsearch is being split up into different columns on the Postgres side, namely `target_uri`, `target_uri_normalized`, `target_selectors`. This also means that going forward we only support annotations with _one_ target.
- We discard the `annotation[document]` field from Elasticsearch with plans to regenerate them from scratch based off of Postgres annotation data.
- Any data that is not mapped onto Postgres columns will end up in the `extra` JSONB column.

__`AnnotationSelectorJSONB` SQLAlchemy type__
We found some annotations that include the NULL byte (`\u0000`) which is allowed in JSON, but not inside Postgres in any field that is stored as a text, including JSONB.
We evaluated multiple solutions for escaping NULL bytes, a global custom JSON de/serializer, special SQLAlchemy type annotation for JSONB that recursively un/escapes NULL bytes in all keys and values. But ultimately decided to go with the minimally invasive solution of only un/escaping NULL bytes in the `prefix`/`exact`/`suffix` fields of the `TextQuoteSelector` types. Since those are the only places where we currently know that NULL bytes exist. This also means that if we try to insert strings containing NULL bytes in other places it will explode and we get notified about it.

@nickstenning Please add some more information about the changes you've been doing before I took over.


nickstenning:
I'm going to close this and open a new PR so that we can get this and #2945 lined up in the right order.

PR #2941, opened 2016-02-10
Improve capture of sidebar injection errors

robertknight:
* Capture the tab URL as context when injecting the sidebar
   fails, for use in debugging.

 * Switch tests to use the toResult() helper instead of
   Promise.catch() to handle assert errors inside
   catch blocks properly. Previously the log tests passed
   even if the assertions inside `injectError.catch()` failed.

 * Handle the Error-like objects used by chrome.extension.lastError
   better.

   When a Chrome async API fails, it sets chrome.extension.lastError
   to an Error-like object with a 'message' property. Since this
   object is not an instance of Error, RavenJS stringified it
   with .toString(), resulting in an unhelpful '[object Object]'
   message in Sentry.

   This commit adds logic to extract out the message property
   of such objects when capturing exceptions.

### DIFF
--- a/h/static/scripts/raven.js
+++ b/h/static/scripts/raven.js
@@ -10,6 +10,8 @@
  * as a dependency.
  */
 
+require('core-js/fn/object/assign')
+
 var Raven = require('raven-js')
 
 var enabled = false;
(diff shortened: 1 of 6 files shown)


h/static/scripts/raven.js
```diff
@@ -54,11 +54,29 @@ function angularModule() {
...
+    // use that instead.
+    if (typeof error === 'object' && error.message) {
+      error = error.message;
+    }
+  }
+
   Raven.captureException(error, {
     extra: {
       context: context,
+      details: details,
```
nickstenning: I think this additional nesting is possibly a bit messy? Perhaps the signature of this could just be `report(error, [context])`, e.g. 

```javascript
report(error, {when: 'injecting sidebar', url: tab.url});
```

which maps to:

```javascript
Raven.captureException(error, {extra: Object.assign({}, context)});
```

or if you really do want a required message, then possibly:

```javascript
extra = Object.assign({}, {when: when}, context);
Raven.captureException(error, {extra: extra});
```

What do you think?

robertknight: Agreed. It would be better without the nesting.

robertknight: Done. I've gone for the third option above because I do think having a string description of the context is useful. Its how I knew where to start looking when fixing the errors here despite the Sentry reports missing useful information otherwise.

nickstenning:
LGTM.

PR #2942, opened 2016-02-10
Improve Chrome sidebar injection error detection and handling

robertknight:
This PR fixes several aspects of sidebar injection errors in the Chrome extension:

1. If a 'known' error happened, clicking on the badge would show an explanation.
    For an unknown error, nothing happened. I've changed this to be consistent
    and open the help page.
2. In several cases async Chrome APIs were called without checking for an error
    in the callback. I've altered the code to use the `util.promisify` wrapper which
    wraps async Chrome APIs into promises and handles checking for errors properly.
3. Speculative fix for https://app.getsentry.com/hypothesis/prod/issues/111098657/ ,
    where if the content script injected into the page to determine the content type threw
    an exception, the script returned an undefined result instead of the expected
    `Array<{type: string}>`.

### DIFF
--- a/h/browser/chrome/help/index.html
+++ b/h/browser/chrome/help/index.html
@@ -64,6 +64,12 @@
         position: relative;
         margin-top: 1.5em;
       }
+
+      .error-details {
+        padding: 10px;
+        background-color: #eaeaea;
+        border-radius: 2px;
+      }
     </style>
   </head>
   <body>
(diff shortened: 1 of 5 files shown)


h/browser/chrome/help/index.html
```diff
@@ -105,6 +105,13 @@ <h1 class="help-item__heading">We’re sorry, Hypothesis doesn’t work on this
         <img class="help-item__icon" src="sad-annotation.svg" />
         <h1 class="help-item__heading">We’re sorry, Hypothesis doesn't work on this site yet.</h1>
       </section>
+      <section id="other-error" class="help-item">
+        <img class="help-item__icon" src="sad-annotation.svg" />
+        <p class="center">
+          There was a problem loading Hypothesis into this page:
```
nickstenning: The other pages make use of an `<h1>` which looks slightly smarter.

In addition, I think we need to gear this copy towards less technical users. So perhaps something like:

```
Oh no! There was a problem loading Hypothesis into this page.

This isn't meant to happen, and we're sorry it's happened to you. Our team has been notified of the error and will try and fix it as soon as possible. If it keeps happening, please get in touch with support@hypothes.is, and pass them these technical details of the error:

<code><pre>
...
</pre></code>
```

We are reporting these errors to Sentry, yes?

robertknight: I agree about the `<h1>` and adding a note that the error has been reported to us. As far as copy for errors go, I'll have a re-think but I do think it is important to keep the message simple and short as well as humble. Making the user read long sentences after you've screwed up is annoying. Also, if the user relays that message onto someone else (eg. Jeremy, Jon) they are less likely to lose the gist if the message is short.

These errors are reported to Sentry.

nickstenning: By all means make it shorter! But it does need to give them guidance on what they can do next, not just "There was an error: TRACEBACK" ;)

robertknight:
something-went-wrong-v2

nickstenning:
LGTM.

PR #2943, opened 2016-02-11
Fix auth{entication,orization} policies for websocket server

nickstenning:
The removal of pyramid_multiauth and friends missed the websocket server. This commit:

- moves configuration of the authentication and authorization policies back into `h.auth` so it is correctly picked up by the websocket server
- removes pyramid_multiauth configuration from `conf/*websocket.ini`.

@seanh this currently blocks deployment, so if you could find a time to review this soon, that would be great!

### DIFF
--- a/conf/development-websocket.ini
+++ b/conf/development-websocket.ini
@@ -1,18 +1,12 @@
 [app:main]
 use: egg:h#websocket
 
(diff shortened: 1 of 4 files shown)


seanh:
Thanks!

PR #2945, opened 2016-02-11
Finish indirection through storage

nickstenning:
This PR completes the indirection of all access to the Annotation and Document models through `h.api.storage`. In some places this results in less than pleasant code, but fear not: in due course all this work will be reversed.

Unfortunately, in the mean time it is important that we have a single place in the code where access to the models is mediated so that we can switch between the ElasticSearch and PostgreSQL models objects on the basis of a feature flag.

In order to achieve this it made most sense to move the various annotation transformers into a single place -- `h.api.transform`. Keeping `h.api.groups.logic` didn't make a lot of sense when none of the code in there actually depended on anything specific to the `h.api.groups` package. It also makes the tests for `h.api.transform` much clearer.


nickstenning:
This is a joint @nickstenning/@chdorner effort. Tagging @seanh, @finitud for review.

seanh:
Really nice to have all the transforming in one place.

This looks great to me. @finitud?

finitud:
LGTM :+1:

PR #2946, opened 2016-02-11
Postgres annotations

nickstenning:
This PR includes changes for the first part of migrating annotation data from Elasticsearch to Postgres.

__Mapping Elasticsearch annotation data to Postgres annotations__

- The `target` field from Elasticsearch is being split up into different columns on the Postgres side, namely `target_uri`, `target_uri_normalized`, `target_selectors`. This also means that going forward we only support annotations with _one_ target.
- We discard the `annotation[document]` field from Elasticsearch with plans to regenerate them from scratch based off of Postgres annotation data.
- Any data that is not mapped onto Postgres columns will end up in the `extra` JSONB column.

__`AnnotationSelectorJSONB` SQLAlchemy type__

We found some annotations that include the NULL byte (`\u0000`) which is allowed in JSON, but not inside Postgres in any field that is stored as a text, including JSONB.

We evaluated multiple solutions for escaping NULL bytes, a global custom JSON de/serializer, special SQLAlchemy type annotation for JSONB that recursively un/escapes NULL bytes in all keys and values. But ultimately decided to go with the minimally invasive solution of only un/escaping NULL bytes in the `prefix`/`exact`/`suffix` fields of the `TextQuoteSelector` types. Since those are the only places where we currently know that NULL bytes exist. This also means that if we try to insert strings containing NULL bytes in other places it will explode and we get notified about it.


nickstenning:
*N.B.* I've set the base for this PR to be the finish-indirection-through-storage, which should make it a bit easier to review. I can never remember if this will do the right thing when that branch merges. I guess we'll find out.

nickstenning:
This is a joint @nickstenning/@chdorner effort. Tagging @seanh, @finitud for review.

h/api/db/types.py
```diff
@@ -0,0 +1,186 @@
...
+# version and variant fields respectively. The four bits M can take values {1,
+# 2, 3, 4, 5} in specified UUID types, and the first three bits of N can take
+# the values {8, 9, 0xa, 0xb} in specified UUID types.
+#
+# In order to expand a 15-byte ElasticSearch flake ID into a value that can be
+# stored in the UUID field, we insert the magic nibbles 0xe, 0x5 into the
+# version and variant fields respectively. These values are disjoint with any
+# specified UUID so the resulting UUID can be distinguished from those
+# generated by, for example, PostgreSQL's uuid_generate_v1mc(), and mapped back
+# to a 20-char ElasticSearch flake ID.
```
seanh: I'll take your word for it!

h/api/db/mixins.py
```diff
@@ -0,0 +1,21 @@
...
+class Timestamps(object):
+    created = sa.Column(sa.DateTime,
+                        default=datetime.datetime.utcnow,
+                        server_default=sa.func.now(),
+                        nullable=False)
+    updated = sa.Column(sa.DateTime,
+                        server_default=sa.func.now(),
+                        default=datetime.datetime.utcnow,
+                        onupdate=datetime.datetime.utcnow,
+                        nullable=False)
```
seanh: Nice!

h/api/models/annotation.py
```diff
@@ -0,0 +1,79 @@
...
+                        server_default='__world__',
+                        nullable=False,
+                        index=True)
+
+    #: The textual body of the annotation.
+    text = sa.Column(sa.UnicodeText)
+    #: The tags associated with the annotation.
+    tags = sa.Column(pg.ARRAY(sa.UnicodeText, zero_indexes=True))
+
+    #: A boolean indicating whether this annotation is shared with members of
```
seanh: Yay!

h/api/models/annotation.py
```diff
@@ -0,0 +1,79 @@
...
+    target_uri_normalized = sa.Column(sa.UnicodeText)
+    #: The serialized selectors for the annotation on the annotated page.
+    target_selectors = sa.Column(types.AnnotationSelectorJSONB,
+                                 default=list,
+                                 server_default=sa.func.jsonb('[]'))
+
+    #: An array of annotation IDs which are ancestors of this annotation.
+    references = sa.Column(pg.ARRAY(types.URLSafeUUID),
+                           default=list,
+                           server_default=sa.text('ARRAY[]::uuid[]'))
```
seanh: This can't be a proper one-many relationship with foreign key constraint?

scripts/migrate.py
```diff
@@ -0,0 +1,324 @@
...
+    # base64-encoded UUIDs or base64-encoded flake IDs. Instead, they're
+    # base64-encoded random numbers between 0 and 1, in ASCII...
+    #
+    # So, we filter out things that couldn't possibly be valid IDs.
+    references = [r for r in references if len(r) in [20, 22]]
+
+    return references
+
+
+def _permissions_allow_sharing(user, group, perms):
```
seanh: My god

scripts/migrate.py
```diff
@@ -0,0 +1,324 @@
...
+    #
+    # 1. Missing 'read' permissions field. Fix: set the permissions to private.
+    if not read_perms:
+        read_perms = [user]
+
+    # 2. 'read' permissions field is [None]. Fix: as in 1).
+    elif read_perms == [None]:
+        read_perms = [user]
+
+    # 3. Group 'read' permissions that don't match the annotation group. I
```
seanh: I'm glad that this duplication of the group id will no longer exist

seanh:
This looks great to me. I'm happy to see it merged. @finitud?

When the feature flag for switching this on comes I'll test it out some.

chdorner:
Closing this one since the target branch should be `master`. Recreated the pull request as #2955.

PR #2948, opened 2016-02-11
Add developer API tokens

seanh:
TODO:

- [x] You can't actually use a developer token to authenticate a request yet
- [x] Some inline CSS in the templates needs to be moved
- [x] Add created and updated datetimes to the token model
- [x] <del>Use a sqlalchemy uuid type</del> (Used `random()` and `hexlify()` instead.)
- [x] Update the API docs
- [x] Add `"dev-"` prefix to tokens
- [x] Tests
- [x] DB migration
- [x] Index on `token`


h/accounts/views.py
```diff
@@ -565,6 +565,36 @@ def _user_notifications(self):
...
+        token = models.Token.get(self.request.authenticated_userid)
+        if token:
+            return {'token': token}
+        else:
+            return {}
+
+    @view_config(request_method='POST')
+    def post(self):
+        """Process notifications POST data."""
+        old_token = models.Token.get(self.request.authenticated_userid)
```
nickstenning: It would be clearer I think to have a `.refresh()` method on the token model which generates a new token. That way there's no need to faff with the database session.

docs/api.rst
```diff
@@ -6,6 +6,29 @@ developers interested in integrating functionality from Hypothesis into their
...
+
+2. Put the API token in the ``Authorization`` header in your requests to the
+   API.
+
+**Example request**::
+
+   GET /api
+   Host: hypothes.is
+   Accept: application/json
+   Authorization: dev-45326b75-b998-431e-a306-70ec121bdf2a
```
nickstenning: There's a bit of confusion here about the nature of the `Authorization` header. Both JWTs and test tokens should be sent with `Authorization: Bearer <token>`.

h/api/auth.py
```diff
@@ -90,3 +92,26 @@ def userid_from_bearer_token(request):
...
+    :type request: pyramid.request.Request
+
+    :returns: the userid authenticated by the request's dev token, or None
+    :rtype: unicode or None
+    """
+    if not request.headers.get('Authorization', '').startswith('dev-'):
+        return None
+
+    token = unicode(request.headers['Authorization'])
+    return models.Token.userid_for_token(token)
```
nickstenning: Not sure about adding this helper method to the Token model. This seems like it would be just as clear:

```python
tok = models.Token.query.get(value)
if tok is None:
    return None
return tok.userid
```

h/api/models/token.py
```diff
@@ -0,0 +1,38 @@
...
+from h.api import db
+
+
+class Token(db.Base):
+
+    """A long-lived API token for a user."""
+
+    __tablename__ = 'token'
+    userid = sqlalchemy.Column(sqlalchemy.UnicodeText(),
+                               primary_key=True)
```
nickstenning: I think it would make more sense for the token, possibly called `id`, to be the PK. There might be good reasons in the future for a user to have multiple types of token associated with their account, but there is never going to be a good reason to have multiple tokens with the same id.

h/api/models/token.py
```diff
@@ -0,0 +1,38 @@
...
+    token = sqlalchemy.Column(sqlalchemy.UnicodeText(),
+                              unique=True,
+                              nullable=False)
+
+    def __init__(self, userid):
+        self.userid = userid
+        self.token = 'dev-' + str(uuid.uuid4())
```
nickstenning: 1. There's no particular reason to use a UUID here. 
2. Most API tokens are "developer" tokens, I think we probably want to call these something else, and probably leave the name internal.

I'd be more inclined to go with something like:

```python
'6879-' + binascii.hexlify(os.urandom(16))
```

(The type field doesn't matter, particularly, but [0x68, 0x79] == ['h', 'y'])

h/api/models/token.py
```diff
@@ -0,0 +1,38 @@
...
+
+    @classmethod
+    def get(cls, userid):
+        return cls.query.get(userid)
+
+    @classmethod
+    def userid_for_token(cls, token):
```
nickstenning: As mentioned earlier, class methods on models are usually getters/searchers which return model objects. Having one return a property is a bit odd.

seanh:
Note to self: add integer primary key. Rename token to value.

seanh:
Move Bearer stripping into auth policy.

seanh:
* "Bearer " prefix is now stripped by auth policy, not by `h.api.auth`'s validator functions. (Also extracting the Authorization header value in the first place.)
* Both types of API token now require a "Bearer " prefix.
* The new API tokens now start with `u"6879-"`.
* The new API token are now generated by `random()` and `hexlify()`, not `uuid`.
* The `Token` model now has an integer primary key
* `Token.token` is now `Token.value`
* The `Token` model now has a `renegerate()` method that changes `value` in place, instead of deleting the row and making a new one

Still a couple of things to do...

h/accounts/views.py
```diff
@@ -565,6 +565,39 @@ def _user_notifications(self):
...
+        """Render the developer page, including the form."""
+        token = models.Token.get_by_userid(self.request.authenticated_userid)
+        if token:
+            return {'token': token.value}
+        else:
+            return {}
+
+    @view_config(request_method='POST')
+    def post(self):
+        """Process notifications POST data."""
```
robertknight: This method docstring looks wrong.

h/api/auth.py
```diff
@@ -90,3 +88,29 @@ def userid_from_bearer_token(request):
...
+    Note: this assumes that the prefix "Bearer " has already been stripped
+    from token.
+
+    :param token: the token to check
+    :type request: unicode
+
+    :returns: the userid authenticated by the token, or None
+    :rtype: unicode or None
+    """
+    if not token.startswith('6879-'):
```
robertknight: I'd prefer magic numbers to be stripped out into a `USEFULLY_NAMED_CONSTANT`

h/api/test/auth_test.py
```diff
@@ -102,64 +94,92 @@ def test_userid_from_bearer_token_calls_decode(jwt):
...
+        models):
+    assert auth.userid_from_api_token(u'abc123') is None
+    assert not models.Token.get_by_value.called
+
+
+@userid_from_api_token_fixtures
+def test_userid_from_api_token_calls_get_by_value(models):
+    auth.userid_from_api_token(u'6879-abc123')
+
+    models.Token.get_by_value.assert_called_once_with(u'6879-abc123')
```
robertknight: Where at all possible, tests should check only the results of a function not how it is implemented. In this case, I think it suffices to test that the result of `userid_from_api_token` matches whatever the database mock's lookup function returns.

h/static/styles/developer-page.scss
```diff
@@ -0,0 +1,8 @@
+// The developer-page block implements the styles used by the "Developer" page
+// in the account/profile pages.
+
+.developer-page__api-token-input {
```
robertknight: Something that isn't clearly defined in our docs at the moment is when something should be a block vs an element.
The basic principle is that anything which is reusable on its own should be a block.

In the frontend code this is usually straightforward because the same applies to a component (or Angular directive), so often component == block.

In the website templates this is less obvious. Here I would suggest that `api-token-input` should be its own block.

h/templates/accounts/developer.html.jinja2
```diff
@@ -0,0 +1,51 @@
...
+    <h2 class="form-heading">
+      <span>{% trans %}Generate your API token{% endtrans %}</span>
+    </h2>
+
+    {% if token %}
+      <label for="token">{% trans %}Your API token is:{% endtrans %}</label>
+      <input id="token" type="text" value="{{ token }}" readonly="readonly"
+             class="developer-page__api-token-input">
+      <span class="help-block">
+        Please keep your API token safe as it can be used to access your
```
robertknight: I think this should be a bit more explicit about what capabilities the API token has. eg. "... can be used to read, modify and delete all data in your account"

h/templates/accounts/developer.html.jinja2
```diff
@@ -0,0 +1,51 @@
...
+    {% endif %}
+          </div>
+        </div>
+      </form>
+      <h2 class="form-heading">
+        <span>{% trans %}API documentation{% endtrans %}</span>
+      </h2>
+      <legend>
+        You can learn more about the Hypothesis API at
+        <a href="http://h.readthedocs.org/en/latest/api.html">h.readthedocs.org/en/latest/api.html</a>.
```
robertknight: Should be an HTTPS link

h/accounts/views.py
seanh: Yep, good spot

h/api/auth.py
seanh: Done

h/api/test/auth_test.py
seanh: I'm not sure I agree with that, in the case of integration functions like this one. We use integration tests like the one above, testing that module A calls function foo from module B the right number of times with the right arguments, all over the place in our Python code. This is testing that it calls `get_by_value()` once and passes in the right argument. The next test tests that it returns the value from `get_by_value()`.

If I _only) tested that the result of `userid_from_api_token()` matches whatever the `get_by_value()` returns, then `userid_from_api_token()` could be calling `get_by_value()`with the wrong argument, getting the wrong result, and returning it, and the test would still pass.

h/static/styles/developer-page.scss
seanh: Done

h/templates/accounts/developer.html.jinja2
seanh: Hmm. But it's not true that it can edit _all_ data. You change your user profile settings over the API, for example. It's just the annotations and I think maybe indirectly groups too. Also you can _create_ new annotations and groups. I'm not sure I want to keep this statement up to date to reflect exactly what an API token can and can't do as the API changes, so I went for something generic.

h/templates/accounts/developer.html.jinja2
seanh: Done

h/api/test/auth_test.py
nickstenning: I'm with Sean here, and it might be useful if I try and explain a little of the why. You're certainly right that tests should test the interface of a unit, and not its implementation. Why? Because it makes the tests fragile -- meaning that you could refactor the unit and break the tests, even if the code actually works. That's a surefire way to make refactoring more painful (and thus less likely to happen).

But don't forget that many (but not all) units have *two* interfaces.

1. The interface the unit exposes to others -- its API contract. For a function that's basically its call signature and perhaps some aspects of the business logic it implements.
2. The dependencies your "unit under test" has on other units. Many units collaborate in order to achieve their ends -- if you don't control this second interface when testing, you're writing integrated tests, not unit test.

So, you want to verify *both* that the unit is upholding its API contract, *and* that it is correctly collaborating with other components.

This naturally leads to two kinds of unit tests: *contract* and *collaboration* tests. I'd suggest you always endeavour to keep them separated into distinct tests. This one is a collaboration test, because `userid_from_api_token` collaborates with `Token.get_by_value`, so we want to ensure that it is calling that function correctly.

I hope that goes some way to understanding the rationale. You can read more about contract/collaboration tests [here](http://bartling.blogspot.de/2012/01/understanding-power-of-isolated-object.html) and [here](http://blog.thecodewhisperer.com/permalink/stub-your-data-access-layer-it-wont-hurt). The original idea comes from [this talk](https://vimeo.com/80533536).

docs/api.rst
```diff
@@ -6,6 +6,30 @@ developers interested in integrating functionality from Hypothesis into their
...
+
+2. Put the API token in the ``Authorization`` header in your requests to the
+   API.
+
+**Example request**::
+
+   GET /api
+   Host: hypothes.is
+   Accept: application/json
+   Authorization: Bearer 6879-45326b75-b998-431e-a306-70ec121bdf2a
```
nickstenning: Could you replace the example tokens with something that looks like the token you're generating (i.e. not a UUID)?

h/api/models/__init__.py
```diff
@@ -10,3 +10,7 @@
 direct to the submodules of this package, but rather through the helper
 functions in `h.api.storage`.
 """
+from h.api.models.elastic import Annotation, Document
```
nickstenning: This looks like a rebase error. These are deliberately gone on master so that nobody tries to use them while we migrate to Postgres.

h/api/models/token.py
```diff
@@ -0,0 +1,54 @@
...
+    id = sqlalchemy.Column(sqlalchemy.Integer,
+                           autoincrement=True,
+                           primary_key=True)
+
+    userid = sqlalchemy.Column(sqlalchemy.UnicodeText(),
+                               nullable=False,
+                               unique=True)
+
+    value = sqlalchemy.Column(sqlalchemy.UnicodeText(),
+                              index=True,
```
nickstenning: `unique=True` implies `index=True`. You don't need `index=True` here.

h/api/models/token.py
```diff
@@ -0,0 +1,54 @@
...
+
+    __tablename__ = 'token'
+
+    id = sqlalchemy.Column(sqlalchemy.Integer,
+                           autoincrement=True,
+                           primary_key=True)
+
+    userid = sqlalchemy.Column(sqlalchemy.UnicodeText(),
+                               nullable=False,
+                               unique=True)
```
nickstenning: I don't think we want a unique constraint on the `userid` field. A user may have multiple tokens.

h/api/models/token.py
seanh: Done

h/api/models/token.py
seanh: But as things stand currently they only have one. Can we leave this constraint here for now and remove it later if/when needed?

h/api/models/__init__.py
seanh: Done

docs/api.rst
seanh: Done

h/api/models/__init__.py
```diff
@@ -10,3 +10,6 @@
 direct to the submodules of this package, but rather through the helper
 functions in `h.api.storage`.
 """
+from h.api.models.token import Token
+
+__all__ = ('Annotation', 'Document', 'Token')
```
seanh: Looks like another rebase error.

seanh: Fixed

h/api/models/token.py
nickstenning: Ok, sure.

nickstenning:
Great scott, we have API tokens!

:tada:

seanh:
Usage:

```bash
http 'https://hypothes.is/api/annotations' Authorization:'Bearer 6879-31d62c13b0099456de5379de90f90395' foo=bar 
```

(That's using [httpie](https://github.com/jkbrzt/httpie).)

tilgovi:
Thanks, @seanh!

JakeHartnell:
Nice!

PR #2950, opened 2016-02-12
Simpler version management

nickstenning:
We were using [versioneer](https://github.com/warner/python-versioneer) to provide version numbers automatically from source control. It's a great tool, and has allowed us to mostly ignore manual versioning. We simply tag a commit and then the built version of h has the right version.

Unfortunately, due to limitations in git, versioneer can't easily generate correct version numbers for a `git archive` tarball when not on a tag commit. The most straightforward way I can see to get correct version numbers for "off-tag" builds of h is to use the tarball output by

    python setup.py sdist

when building the Docker image.

But when we impose that restriction, it no longer makes an awful lot of sense to keep versioneer (and all its many and various ways of calculating the version) around. We can replicate its functionality for our simpler and less general setup with a few dozen lines in `setup.py`.


h/_version.py
```diff
@@ -1,460 +1,59 @@
...
+    # Git SHA
+    re.compile(r'(?P<sha>g[0-9a-f]{7})'),
+    # Dirty flag
+    re.compile(r'(?P<dirty>dirty)'),
+)
+VERSION_UNKNOWN = '0+unknown'
+
+
+def parse_version(ver):
+    """Parse a `git describe` version into a dictionary."""
```
seanh: I didn't know about `git describe`, neat!

PR #2952, opened 2016-02-12
Fix #2810: no error on sign in to unactivated account

### LINKED ISSUE #2810: No error message when trying to sign in to an unactivated account
1. Register a new user account, but don't activate it yet
2. Try to sign in to the account using the correct username or email and password

It doesn't log you in, just re-renders the login form with the same username/email and password filled in and no error message.

seanh:
screenshot from 2016-02-12 12-18-39


nickstenning:
LGTM.

PR #2954, opened 2016-02-12
Don't crash out if username blacklist doesn't load

nickstenning:
If, for whatever reason, we can't load the username blacklist, then don't fail to register the user. Instead, log a warning about the problem and carry on.

This isn't necessarily ideal, as it means that (at least in theory) a user can register a blacklisted username while this problem exists. But it's probably better than throwing a 500 and failing to register people.

### DIFF
--- a/h/accounts/schemas.py
+++ b/h/accounts/schemas.py
@@ -1,5 +1,6 @@
 # -*- coding: utf-8 -*-
 from pkg_resources import resource_stream
+import logging
 
 import colander
 import deform
@@ -10,6 +11,7 @@ from h import i18n
 from h.accounts import models
 
 _ = i18n.TranslationString
+log = logging.getLogger(__name__)
 
 USERNAME_BLACKLIST = None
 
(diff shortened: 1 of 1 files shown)


robertknight:
Any idea how this might have happened? This is a static resource that ships with the 'h' package, so presumably quite a lot else could be wrong if files from the package cannot be read.

nickstenning:
Yep! I missed it out: https://github.com/hypothesis/h/commit/e3cf844dfffa9266ee7ec2ddaf627922dfa711a5

seanh:
Would it be better to crash at startup if the blacklist file isn't found?

nickstenning:
> Would it be better to crash at startup if the blacklist file isn't found?

Not in production, no :)

seanh:
Was just thinking that we'd notice a crash right away, whereas we might not notice blacklisted usernames being silently taken.

Is this blacklist just hand-written to a file that's committed to git, then read by just the one line of code?
Couldn't we just hardcode the list in the code then?

nickstenning:
That's why it emits a warning, which will be picked up by Sentry and would precipitate an audit of the user table.

I'd always rather that we start up and start serving requests if we possibly can.

> Is this blacklist just hand-written to a file that's committed to git, then read by just the one line of code?
> Couldn't we just hardcode the list in the code then?

Sure, but I don't know that we'd want to. It's 400 lines of data. 

I don't expect to be playing around with the MANIFEST file too often and I while I don't expect this problem to recur, I just don't want it to take out user registration if it does.

PR #2956, opened 2016-02-12
Fix broken promises in IE

nickstenning:
We support IE10/11, neither of which define Promise globals, so make jshint warn if we use Promise without explicitly defining it (by importing a polyfill), and fix the places where we currently use Promise as an implied global.


nickstenning:
https://www.youtube.com/watch?v=QFtvT1dWfwM

PR #2958, opened 2016-02-13
New front-end build system based on Gulp

robertknight:
This implements a new build system for the front-end assets based on Gulp, Browserify and node-sass.

The benefits of this are:

 * A better front-end development experience through faster rebuilds. Code changes typically
   result in 0.5-1.5s update times when running `gulp watch`.
 * Source maps for production builds which should enable more accurate roll-ups in Sentry.
 * Avoid cluttering the source tree with compiled assets. Everything is now built in `build/{scripts,styles,images,fonts}`
 * Replace compass with node-sass which simplifies build setup
 * Easier maintenance by using CommonJS for all dependencies, rather than a mix of CommonJS for
   some dependencies and window globals imported via `<script>` tags for others.
 * Reduction of page weight for the user by making it easier to track where dependencies are used
   and easier factoring of vendor code into separate bundles from the main app. `unorm` and
   `showdown` are two notable examples that were easy to separate.

The changes in this PR fall into two groups, changes before e4c677a are a series of steps that make the subsequent introduction of a new build system easier e4c677a and later are the Gulp build system itself. Commit-by-commit review is probably easier and I could separate these into separate PRs.

At a high level, the changes are:
 * Add a 'gulpfile.js' to bundle the scripts, styles, images and fonts
 * Use CommonJS requires to pull in Angular and Katex instead of using script tags, which simplifies
   keeping track of which vendor libs are needed by which modules. The vendor libs are
   factored into separate bundles using Browserify's `require` and `externals` options.
 * Replace compass with node-sass plus compass-mixins which are the SCSS files from
   compass extracted into a separate lib. Many of our uses of compass mixins were purely
   for vendor prefixes and several of these needed to be replaced
 * Remove jQuery from the sidebar app bundle. It turned out that with a few trivial changes,
   we only need Angular's minimal `jqLite`. Note that the injector still needs the full jQuery
 * Factor out Raven JS, unorm, showdown into their own bundles
 * Cleanup of .gitignore following the separation of source and build directories for client assets

----

*TODO*
* [x] Log errors during incremental rebuilds
* [x] Investigate why unnecessary large polyfills for Node fs and crypto libs are being pulled into the app bundle.
* [x] Check the admin areas of the site
* [x] Do another pass over `gulpfile.js`, polish and add additional documentation
* [x] Re-test app and site in IE 10/11
* [x] Re-test the app in Edge, Safari
* [x] Investigate why the injector bundle contains inline sourcemaps for Annotator which bloat the bundle
* [x] Check that cache-busting is still working properly (fixed by replacing webassets with a simple static view that uses a path -> URL mapping manifest for appending cache-busting query strings.
* [x] Remove the `hypothesis assets` command and update the Dockerfile to use the new static asset build system
* [x] Update the front-end developer documentation to explain the changed workflows for building client-assets
* [x] Cleanup package.json such that everything that is included in code actually delivered to users is in `dependencies` and everything else is in `devDependencies`.
* [x] Unit tests for h.assets.Environment
* [x] Remove the large inline sourcemaps in the 'annotator.js' and 'annotator.document.js' from production builds of `injector.js` (Fixed by upgrading Browserify)


lenazun:
🎉

h/buildext.py
```diff
@@ -218,45 +187,84 @@ def build_chrome(args):
...
+    shutil.copyfile('build/scripts/extension.bundle.js',
+                    'build/chrome/lib/extension.bundle.js')
     if args.debug:
-        subprocess_args.append('--debug')
-    subprocess.call(subprocess_args)
+        shutil.copyfile('build/scripts/extension.bundle.js.map',
+                        'build/chrome/lib/extension.bundle.js.map')
+
+    webassets_env = AssetEnvironment('/public')
+    webassets_env.register('app_js', [
```
robertknight: This duplicates a bunch of configuration from `h/assets.yaml`, which is now essentially a dictionary mapping names of bundle collections (eg. `app_js` to bundle names). I think it would probably make sense to load in the existing `assets.yaml` file and use that. Might even make sense to convert `h/assets.yaml` to a simple JSON file and implement a trivial webassets loader for it.

h/buildext.py
```diff
@@ -295,11 +303,6 @@ def check_sentry_dsn_is_public(dsn):
                     '"wss://hypothes.is/ws")',
                     dest='websocket_url',
                     metavar='URL')
-parser.add_argument('--no-bundle-sidebar',
```
robertknight: I've taken out the option to skip bundling the sidebar for the moment to simplify things. After this is merged, it should be a fairly straightforward addition to move to a workflow where gulp can incrementally build the assets directly into the `build/chrome/{lib, public}` folders, post a notification to Chrome when an update happens and the extension will auto-reload itself.

h/static/scripts/app.coffee
```diff
@@ -4,9 +4,8 @@ if window.RAVEN_CONFIG
   require('./raven').init(window.RAVEN_CONFIG)
 
 
-require('autofill-event')
 angular = require('angular')
-require('angular-jwt')
+require('autofill-event')
```
robertknight: TODO - Add a note here that `autofill-event` needs to be included _after_ `window.angular` is defined by `require('angular')`

h/static/scripts/directive/annotation.js
```diff
@@ -707,7 +707,10 @@ function AnnotationController(
 
   vm.share = function(event) {
     var $container = angular.element(event.currentTarget).parent();
-    $container.addClass('open').find('input').focus().select();
+    $container.addClass('open');
```
robertknight: The changes to the directives here are about compatibility with `jqLite` instead of the full version of `jQuery`

h/static/scripts/test/bootstrap.js
```diff
@@ -1,2 +1,14 @@
 // Expose the sinon assertions.
 sinon.assert.expose(assert, {prefix: null});
+
+// Load Angular libraries required by tests.
+//
+// The tests for annotator currently rely on having
```
robertknight: TODO: s/annotator/Annotator

scripts/gulp/createBundle.js
```diff
@@ -0,0 +1,141 @@
...
+ *   require?: string[];
+ *   transforms: Transform[];
+ *
+ *   watch?: boolean;
+ *   minify?: boolean;
+ * }
+ */
+
+/**
+ * Bundles the JavaScript for an application.
```
robertknight: TODO: Add docs explaining that this contains the shared logic for building a Browserify bundle with source maps and suitable transforms.

scripts/gulp/createBundle.js
```diff
@@ -0,0 +1,141 @@
...
+  if (opts.watch) {
+    bundle.plugin(watchify);
+    bundle.on('update', function (ids) {
+      var start = Date.now();
+
+      log('Source files changed', ids);
+      build().then(function () {
+        log('Updated %s (%d ms)', bundleFileName, Date.now() - start);
+      }).catch(function (err) {
+        console.error('Building updated bundle failed:', err);
```
robertknight: TODO: This logic for reporting build errors does not work properly. Replace it with a corrected version which uses the `error` event from the `b.bundle()` result.

vendor-bundles.js
```diff
@@ -0,0 +1,38 @@
+// Defines a set of vendor bundles which are
```
robertknight: TODO: This should be a JSDoc module comment

vendor-bundles.js
```diff
@@ -0,0 +1,38 @@
...
+// libraries of 3rd-party code referenced by
+// one or more bundles of Hypothesis client/frontend
+// code
+
+/**
+ * Decorate a module ID with a 'doNotParse' property
+ * indicating that the module bundler should not try
+ * to parse it for require() statements.
+ */
+function doNotParse(id) {
```
robertknight: TODO: `module.exports` currently has this type:

```ts
interface BundleMap {
  [index: string]: Array<string | { name: string; noParse: boolean }>
}
```

It would be easier to understand if this was just `[index: string]: Array<{name: string, noParse?: boolean}>` instead.

vendor-bundles.js
```diff
@@ -0,0 +1,38 @@
+// Defines a set of vendor bundles which are
+// libraries of 3rd-party code referenced by
+// one or more bundles of Hypothesis client/frontend
+// code
+
+/**
+ * Decorate a module ID with a 'doNotParse' property
+ * indicating that the module bundler should not try
+ * to parse it for require() statements.
```
robertknight: TODO: Add a comment explaining why we would want to disable parsing for a module in the first place. In the case of 'katex', it is because `katex.min.js` is itself a Browserify bundle and consequently contains 'require()' statements which cannot be resolved outside of the original KaTeX source tree. For the other modules, it is because that speeds up the build by avoiding parsing huge JS source files which contain no `require()` statements.

scripts/gulp/createBundle.js
robertknight: Done

h/assets.py
```diff
@@ -112,7 +39,14 @@ def asset_response_subscriber(event):
 
 
 def includeme(config):
-    config.registry.settings.setdefault('webassets.bundles', 'h:assets.yaml')
+    WEBASSETS_DEFAULTS = {
```
nickstenning: This would probably be better as simply:

```python
settings.setdefault('webassets.bundles', 'h:assets.yaml')
settings.setdefault('webassets.base_dir', 'h:../build')
settings.setdefault('webassets.base_url', '/assets')
```

h/buildext.py
nickstenning: I'm pretty sure you could just use `webassets` itself without too much trouble. Just use `webassets` directly without using the `pyramid_webassets` indirection layer that the application uses.

And yes, if possible let's stick with reading `h/assets.yaml`. I'm probably not in favour of switching it to JSON -- configuration files need comments and JSON doesn't have comments.

h/static/scripts/annotator/main.js
```diff
@@ -1,9 +1,11 @@
+require('../polyfills');
+
 var extend = require('extend');
 var Annotator = require('annotator');
 
 // Polyfills
 var g = Annotator.Util.getGlobal();
-if (g.wgxpath) {
+if (g.wgxpath && !g.document.evaluate) {
```
nickstenning: `wgxpath` [already does this check](https://github.com/google/wicked-good-xpath/blob/320441591b69b02b31a2d19bafcfc1b4de19c034/src/wgxpath.js#L237), for what it's worth.

h/static/scripts/polyfills.js
```diff
@@ -0,0 +1,15 @@
...
+
+// URL constructor, required by IE 10+,
+// Microsoft Edge.
+window.URL = require('js-polyfills/url').URL;
+
+// document.evaluate() implementation,
+// required by IE 10, 11
+//
+// This sets `window.wgxpath`
+if (!window.document.evaluate) {
```
nickstenning: Ditto to my comment above.

h/static/scripts/polyfills.js
```diff
@@ -0,0 +1,15 @@
+// ES2015 polyfills
+require('core-js/es6/promise');
+require('core-js/fn/object/assign');
+
+// URL constructor, required by IE 10+,
+// Microsoft Edge.
+window.URL = require('js-polyfills/url').URL;
```
nickstenning: This is also a somewhat more brutal approach than [the polyfill itself takes](https://github.com/inexorabletash/polyfill/blob/15de2d3e7f21abd20304940a19d2436c80290e7f/url.js#L22). Can we not use the feature detection built into the polyfill here?

robertknight: Indeed. The polyfill's own feature detection unfortunately does not work when loaded as a CommonJS module because it relies on `this === window`. However, we can do our own feature detection trivially and I've implemented that. Of the different levels of support listed in the comment you linked to, the important thing we need is the `URL` constructor.

I've also corrected the comment because it turns out that current versions of Edge do support the URL constructor.

nickstenning:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;R3s7yHus&#x2F;268-new-front-end-build-system"> New front-end build system</a>

Dockerfile
```diff
@@ -28,18 +27,26 @@ WORKDIR /var/lib/hypothesis
...
 # Copy the rest of the application files
 COPY gunicorn.conf.py ./
 COPY conf ./conf/
 COPY h ./h/
-COPY scripts ./scripts/
+
+# Build client and static assets
+
+## Remove dependencies required only for unit testing
+RUN sed -i -r '/karma|proxyquire|phantom|mocha|chai|jscs/d' package.json
```
seanh: I wonder if everything needed to build the docker image could go in dependencies, and then only the things being removed by sed here could go in devDependencies, then you could just use `npm install --production` instead of 'sed`?

Since our only officially supported "production" deployment is the docker one, it might make sense that `npm install --production` is for use in the Dockerfile.

robertknight: @seanh - Yes, that's what @nickstenning 's just persuaded me makes more sense. Just doing that now.

Dockerfile
```diff
@@ -9,15 +9,13 @@ RUN apk add --update \
     python \
     py-pip \
     nodejs \
-    ruby \
+    git \
```
robertknight: This is required by the node-sass build, which builds the native libsass C++ lib as part of its install process

nickstenning:
An absolutely massive improvement. There are small things here and there that will need addressing, but none that are going to stop me merging this.

.gitignore
```diff
@@ -15,23 +15,3 @@ node_modules/
...
 
 *.log
-
-# temp directories generated by webassets
-h/static/.webassets-cache/
-h/static/webassets-external/
-
-# temp directory generated by compass
-# during CSS preprocessing
-h/static/tmp*/
```
seanh: Glad to see these annoying directories gone

PR #2962, opened 2016-02-15
Annotation test cleanup

robertknight:
This is an initial cleanup of the annotation tests in preparation for upcoming changes for direct linking.
The changes here are:

* Refactor the two sets of tests for AnnotatonController into a single set. This removes a bunch of duplication.
* Reduce the number of globals in the annotation tests by moving declarations for fakes into `beforeEach()` where possible
* Extract out annotation fixtures into a separate module. These are likely to be useful in other tests
* Extract out helper functions for transforming document metadata into a separate module


h/static/scripts/directive/test/annotation-test.js
```diff
@@ -404,28 +255,25 @@ describe('annotation', function() {
...
     var sandbox;
 
     function createDirective(annotation) {
-      annotation = annotation || defaultAnnotation();
+      annotation = annotation || fixtures.defaultAnnotation();
+
+      // A new annotation won't have any saved drafts yet.
+      if (!annotation.id) {
+        fakeDrafts.get.returns(null);
+      }
```
seanh: How come this had to be added? (And shouldn't it go in `defaultAnnotation()`?)

robertknight: It is a side effect I extracted out of `defaultAnnotation` when I made it a simple function in a separate module (annotation-fixtures) which does not have access to fakeDrafts.

robertknight: Also, the logic was duplicated in a couple of different fixtures previously. Now it is centralized in one place.

h/static/scripts/directive/test/util.js
```diff
@@ -1,13 +1,40 @@
...
   return name.replace(uppercasePattern, '-$1').toLowerCase();
 }
 
 /**
+ * Helper for retrieving an Angular module in a test.
+ *
+ * Given the 'inject' function from the 'angular-mocks' module,
+ * retrieves an instance of the specified Angular module.
+ */
+function ngModule(inject, name) {
```
seanh: When I see `ngModule()` being called it looks to me like an Angular thing being called. Is this a good naming style for our own helper functions?

robertknight: It will be called as `util.ngModule` where `var util = require('./util')` so I hope it should be obvious enough that it isn't part of angular. All the angular builtins are accessed via `angular.<foo>`.

seanh:
I'm not if I like sharing fixtures between test modules, even though we do it a lot in our Python code as well. It just increases "travel" as to understand a test you now need to go and find a fixture that it uses in a different file. It tempts you to make fixtures more complex as new tests come along that just need fixture foo but with a bar property as well etc. And you can end up in a place where you can't change a given fixture without causing lots of tests all over the place to fail or, worse, causing them to still succeed but to no longer actually be testing what they were supposed to be. It might be preferable to have code duplication of fixtures in different test modules, and let each copy of a given fixture evolve in the direction its tests need it to.

If you disagree though, I'm fine to merge this.

On the other hand I _do_ like the idea of "utilities for dealing with Angular" being in a shared utils module for different test modules to use. I much prefer `scope = testUtils.getNewScope();` in a test, instead of `scope = util.ngModule(inject, '$rootScope').$new();`, which is completely incomprehensible Angular-babble to me. Helpers like `getRootScope()` etc are currently duplicated in a few different test modules I think. Would it be nice to move these into a shared utils module, rather than having the tests depend on Angular directly?

About moving globals like `fakeUrlEncodeFilter` etc etc into `beforeEach()` - isn't the point of these being globals that each test has access to them, so a test can modify `fakeUrlEncodeFilter.foo = bar` if it has to for just that test. Whereas if they're now locked inside the `beforeEach()` that provides them to Angular, then you can no longer do that, or at least you can't get access to them in any easy and obvious way?

Not defending globals though.

But a better pattern, that I've used elsewhere, might be to use a `createAnnotationDirective()` helper function that returns an object containing not only the directive, but also each of the services and any other bits that the helper function created in order to create the directive. That way any test that needs to can get a reference to the mock version of a particular dependency and query or modify it. At the same time there are no globals. Cost is one line per test: to call `createAnnotationDirective()`, but that means that if any particular test doesn't need `createAnnotationDirective()` at all then it simply doesn't call it, which is also nice. And it increases the "I can completely understand this test by looking at the test's code only" factor because if the test is calling helper functions then at least I can see that it's doing so, there aren't `before()` and `beforeEach()` functions being called implicitly.

h/static/scripts/directive/test/util.js
seanh: Is it possible to have `testUtils.newScope()` instead of `util.ngModule(inject, '$rootScope').$new()`?

robertknight:
> About moving globals like fakeUrlEncodeFilter etc etc into beforeEach() - isn't the point of
> these being globals that each test has access to them, so a test can
> modify fakeUrlEncodeFilter.foo = bar if it has to for just that test. 

None of the tests did this though, they just provided enough of a stub for the rest of the code to work. Now only the stubs which _are_ directly referenced by individual tests are globals.

h/static/scripts/directive/test/annotation-test.js
seanh: Ok

seanh:
> > About moving globals like fakeUrlEncodeFilter etc etc into beforeEach() - isn't the point of
> > these being globals that each test has access to them, so a test can
> > modify fakeUrlEncodeFilter.foo = bar if it has to for just that test. 
> 
> None of the tests did this though, they just provided enough of a stub for the rest of the code to work.
> Now only the stubs which _are_ directly referenced by individual tests are globals.

But now if I want to write a test that _does_ want a different return value from `features.flagEnabled()` for example, when it was a global I could have just done `features.flagEnabled.returns(false)`. Now that it's local to `beforeEach()` I can't do that, can I? I'd have to turn it back into a global again.

seanh:
I guess if you're sure that none of the tests will ever need to access these particular dependencies, then it would make sense to turn them into `beforeEach()` locals. Is that the idea?

robertknight:
> Now that it's local to beforeEach() I can't do that, can I? I'd have to turn it back into a global again.

Yes indeed, that's kind of the point. You can now _see_ which dependencies are modified by tests and therefore get a better idea of what is actually being tested and what is not. Here for example, we can now see that the formatting of the document title, domain etc. in the annotation header was not tested.

seanh:
Ok, do you think it's worth clarifying this with some comments? "Dependencies which are modified by tests" and "Dependencies which aren't modified by tests".

h/static/scripts/directive/test/annotation-test.js
```diff
@@ -1404,6 +1175,26 @@ describe('annotation', function() {
           assert.ok(controller.editing());
         });
       });
+
+      it(
```
nickstenning: As mentioned before, I think worrying about long lines in tests is a recipe for a tedious life.

```javascript
it('Passes group:<id> to the server when saving a new annotation', function () {
  ...
})
```

h/static/scripts/directive/test/util.js
nickstenning: I think the point is that this function is more general than just `newScope()` -- of course it would be possible to have `newScope` in here in addition, defined as:

```javascript
function newScope(props) {
    var scope = ngModule(inject, '$rootScope').$new();
    return Object.assign(scope, props);
}
```

nickstenning:
Really nice work. I've made a couple of comments but this is already so much better than what's there at the moment that I'm going to go ahead and merge it.

The point in here about line length is really one about readability. Don't pay attention to the line length limit if it actually makes the file less readable, and changing between:

```javascript
it('does stuff', function () {
});

it(
  'does stuff',
  function () {
  }
);
```

throughout the codebase makes it *harder* to read than just going over the line length limit, IMO.

Line length limits make more sense in application code, where there are few justifications for really long lines (although there are still exceptions to that rule), and overall the rule helps rather than hinders readability. In tests though, I'm not sure that's true. Hence my comments. :)

h/static/scripts/directive/test/util.js
seanh: Yes. I was asking for functions like that to be defined in `util` and used in the tests instead, because it's more readable. This pr has actually replaced `getRootScope()` etc in tests with `util.ngModule(inject, '$rootScope').$new()` which I think is far less readable.

PR #2963, opened 2016-02-15
Convert filters in the annotation directive to plain functions

robertknight:
The annotation directive used a set of filters for formatting
document domains, titles and usernames. These filters were only used once in the whole codebase.

Converting these to plain functions eliminates a bunch of boilerplate in the code and the tests
and as noted at https://www.binpress.com/tutorial/speeding-up-angular-js-with-simple-optimizations/135 is a small performance optimization.

Transforming the filters into plain functions means that we use proxyquire to mock them in the tests.


seanh:
Should these functions still be in the `filter` dir if they're no longer Angular filters?

robertknight:
They should be moved out. Our `scripts/` directory needs re-organization though, it currently contains most of the client code.

seanh:
Just rebased this

seanh:
Ok

PR #2964, opened 2016-02-15
Don't open "welcome page" when installed by admin policy

### LINKED ISSUE #2890: Opening "welcome" page is unhelpful/confusing in school/office install settings

nickstenning:
This commit is a second go at fixing #2890, after the first abortive attempt: 7d3369e.

If the extension has been installed as the result of an administrative policy rather than user action, it doesn't make a lot of sense (and appears to cause some frustration) to open a tab to our "welcome page."

This commit suppresses the opening of the "welcome page" on first run if the install type is admin. Chrome documentation references:

- https://developer.chrome.com/extensions/management#event-onInstalled
- https://developer.chrome.com/extensions/management#type-ExtensionInfo
- https://developer.chrome.com/extensions/management#type-ExtensionInstallType

Fixes #2890.

### DIFF
--- a/h/browser/chrome/lib/hypothesis-chrome-extension.js
+++ b/h/browser/chrome/lib/hypothesis-chrome-extension.js
@@ -79,14 +79,19 @@ function HypothesisChromeExtension(dependencies) {
   };
 
   /* Opens the onboarding page */
-  this.firstRun = function (installDetails) {
+  this.firstRun = function (extensionInfo) {
     // If we've been installed because of an administrative policy, then don't
     // open the welcome page in a new tab.
     //
     // It's safe to assume that if an admin policy is responsible for installing
     // the extension, opening the welcome page is going to do more harm than
     // good, as it will appear that a tab opened without user action.
-    if (installDetails.installType === 'admin') {
+    //
+    // See:
+    //
+    //   https://developer.chrome.com/extensions/management#type-ExtensionInstallType
+    //
+    if (extensionInfo.installType === 'admin') {
       return;
     }
 
--- a/h/browser/chrome/lib/install.js
+++ b/h/browser/chrome/lib/install.js
@@ -20,8 +20,18 @@ chrome.runtime.requestUpdateCheck(function (status) {
 });
 
 function onInstalled(installDetails) {
+  // The install reason can be "install", "update", "chrome_update", or
+  // "shared_module_update", see:
+  //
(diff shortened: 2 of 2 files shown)


robertknight:
Code changes LGTM, I'm just reading the docs on setting up an environment to test this.

robertknight:
I've confirmed this works as intended. To test it, I followed the steps at https://www.chromium.org/administrators/linux-quick-start

On a Linux system:

* Build the Hypothesis extension
* Package the extension using 'Pack Extension' in chrome://extensions/
* Drop the packed .crx file into Chrome. Verify that the welcome page is opened and get the extension ID displayed in the extensions details field. 
* Uninstall the extension
* Close Google Chrome
* Create a directory containing two files `h.crx`, the .crx file generated in step 3 and `update.xml`, with the following content and the extension ID filled in:
```xml
<?xml version='1.0' encoding='UTF-8'?>
<gupdate xmlns='http://www.google.com/update2/response' protocol='2.0'>
  <app appid='INSERT_EXTENSION_ID_HERE'>
    <updatecheck codebase='http://localhost:8000/h.crx' version='2.0' />
  </app>
</gupdate>
```
* Serve the directory created in step 6 using a suitable HTTP file server, eg. `python -m SimpleHTTPServer 8000`
* Create the following file: `/etc/opt/chrome/policies/managed/install_hypothesis.json`:
```json
{
	"ExtensionInstallForcelist":  [
		"INSERT_EXTENSION_ID_HERE;http://localhost:8000/update.xml"
	]
}
```
* Start Google Chrome. Verify that the extension is installed and that uninstallation is disabled in chrome://extensions, but the welcome page is not shown.

nickstenning:
Sweeeeet! Thank you Robert.

PR #2966, opened 2016-02-15
Make annotation directive more consistent with other directives

robertknight:
* Make it an element directive and use the existing utility
   functions for instantiating them in tests
 * Use '=' bindings for input properties and 'on<EventName>' names
   for events
 * Remove unused 'showReplyCount' input property and the
   shouldShowNumReplies() function which was only used
   as the argument to this property.

### DIFF
--- a/h/static/scripts/directive/annotation.js
+++ b/h/static/scripts/directive/annotation.js
@@ -254,11 +254,6 @@ function AnnotationController(
     /** Whether or not this annotation is private. */
     vm.isPrivate = false;
 
-    /** Copy isSidebar from $scope onto vm for consistency (we want this
-      * directive's templates to always access variables from vm rather than
-      * directly from scope). */
-    vm.isSidebar = $scope.isSidebar;
-
     /** A fuzzy, relative (eg. '6 days ago') format of the annotation's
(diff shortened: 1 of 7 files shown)


h/static/scripts/directive/annotation.js
```diff
@@ -254,11 +254,6 @@ function AnnotationController(
     /** Whether or not this annotation is private. */
     vm.isPrivate = false;
 
-    /** Copy isSidebar from $scope onto vm for consistency (we want this
-      * directive's templates to always access variables from vm rather than
-      * directly from scope). */
-    vm.isSidebar = $scope.isSidebar;
```
robertknight: This is no longer needed because of the use of `bindToController` in the annotation directive

h/static/scripts/directive/annotation.js
```diff
@@ -290,7 +285,7 @@ function AnnotationController(
       * haven't been saved yet - the data that will be saved to the server when
       * they are saved).
       */
-    domainModel = $scope.annotationGet();
+    domainModel = vm.annotation;
```
seanh: Yay! Always hated `annotationGet()`.

PR #2967, opened 2016-02-15
Upgrade to Angular 1.5.0

robertknight:
This was a straightforward update to the recent major release. The only fix needed was to correctly mark a couple of optional properties to `<excerpt>` as optional by using '?'.

The other change I've made here is to take advantage of one-way bindings ('<' instead of '=') to indicate properties which are only _inputs_ to directives and differentiate them from properties which have two-way bindings ('='). Note that this is distinct from one-_time_ bindings ('::') which are only ever evaluated once.


nickstenning:
Rebased on master.

nickstenning:
LGTM.

PR #2968, opened 2016-02-16
Add "bouncer" support to Chrome extension

seanh:
This doesn't necessarily need to be merged yet, can wait until <https://github.com/hypothesis/bouncyhouse/pull/1> is finished, but <https://github.com/hypothesis/bouncyhouse/pull/1> needs this in order to work,

* * *

Add support for "bouncyhouse", the new Hypothesis direct-link bouncer
service, to the Chrome extension.

Some JavaScript from the bouncyhouse instance that will be running at
https://hpt.is/ needs to be able to tell whether the user has our Chrome
extension installed or not. This means that our Chrome extension needs
to:

1. Declare in its manifest.json that https://hpt.is/ can send messages
   to it.
2. Listen for and reply to those messages.

Development builds of the Chrome extension will need to accept messages
from dev instances of bouncyhouse, so add an argument to the
hypothesis-buildext command so that developers can specify their dev
instance's local URL instead of https://hpt.is/.

### DIFF
--- a/docs/hacking/chrome-extension.rst
+++ b/docs/hacking/chrome-extension.rst
@@ -20,7 +20,7 @@ To build and install the Chrome extension:
 
    .. code-block:: bash
 
-      hypothesis-buildext chrome --debug --service 'http://127.0.0.1:5000'
+      hypothesis-buildext chrome --debug --service 'http://127.0.0.1:5000' --bouncer 'http://127.0.0.1:8000/*'
 
    .. note::
 
--- a/h/browser/chrome/lib/install.js
+++ b/h/browser/chrome/lib/install.js
@@ -14,7 +14,19 @@ var browserExtension = new HypothesisChromeExtension({
 });
 
 browserExtension.listen(window);
+
 chrome.runtime.onInstalled.addListener(onInstalled);
+
+// Respond to messages sent by the JavaScript from https://hpt.is.
+// This is how it knows whether the user has this Chrome extension installed.
+chrome.runtime.onMessageExternal.addListener(
+  function (request, sender, sendResponse) {
+    if (request.type === 'ping') {
+      sendResponse({type: 'pong'});
+    }
+  }
+);
+
 chrome.runtime.requestUpdateCheck(function (status) {
   chrome.runtime.onUpdateAvailable.addListener(onUpdateAvailable);
 });
--- a/h/browser/chrome/manifest.json.jinja2
+++ b/h/browser/chrome/manifest.json.jinja2
@@ -46,5 +46,9 @@
     "lib/*",
     "help/*",
     "content/web/viewer.html"
-  ]
+  ]{% if bouncer_url -%},
+  "externally_connectable": {
+    "matches": ["{{ bouncer_url }}"]
+  }
+  {%- endif %}
 }
--- a/h/buildext.py
+++ b/h/buildext.py
@@ -75,7 +75,7 @@ def copytree(src, dst):
             shutil.copyfile(s, d)
 
 
-def chrome_manifest(script_host_url):
+def chrome_manifest(script_host_url, bouncer_url):
     # Chrome is strict about the format of the version string
     if '+' in h.__version__:
         tag, detail = h.__version__.split('+')
@@ -88,7 +88,8 @@ def chrome_manifest(script_host_url):
 
     context = {
         'version': version,
-        'version_name': version_name
+        'version_name': version_name,
+        'bouncer_url': bouncer_url,
     }
 
     if script_host_url:
@@ -204,7 +205,8 @@ def build_chrome(args):
 
     # Render the manifest.
     with codecs.open('build/chrome/manifest.json', 'w', 'utf-8') as fp:
-        data = chrome_manifest(script_host_url=None)
+        data = chrome_manifest(script_host_url=None,
+                               bouncer_url=args.bouncer_url)
         fp.write(data)
 
     # Write build settings to a JSON file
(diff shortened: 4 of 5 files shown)


h/browser/chrome/lib/install.js
```diff
@@ -14,7 +14,18 @@ var browserExtension = new HypothesisChromeExtension({
 });
 
 browserExtension.listen(window);
+
 chrome.runtime.onInstalled.addListener(onInstalled);
+
+chrome.runtime.onMessageExternal.addListener(
+  /** Respond to messages sent by the JavaScript from https://hpt.is.
```
robertknight: This indentation is confusing. It should just be an inline comment above the 'addListener' line.

h/buildext.py
```diff
@@ -303,6 +304,13 @@ def check_sentry_dsn_is_public(dsn):
 parser.add_argument('browser',
                     help='Specifies the browser to build an extension for',
                     choices=['chrome'])
+parser.add_argument('--bouncer-url',
```
nickstenning: The existing parameters which take urls are `--service` and `--websocket`, without a `-url` suffix. So this should be called `--bouncer`, and mapped to `args.bouncer_url` using the `dest` keyword argument.

It should also be defined with the `--options`, not down here with the positional arguments.

h/browser/chrome/lib/install.js
```diff
@@ -14,7 +14,18 @@ var browserExtension = new HypothesisChromeExtension({
...
+    * This is how it knows whether the user has this Chrome extension
+    * installed.
+    */
+  function (request, sender, sendResponse) {
+    sendResponse('Hi there!');
```
nickstenning: `PONG` would be more conventional ;)

docs/hacking/chrome-extension.rst
```diff
@@ -20,7 +20,7 @@ To build and install the Chrome extension:
 
    .. code-block:: bash
 
-      hypothesis-buildext chrome --debug --service 'http://127.0.0.1:5000'
+      hypothesis-buildext chrome --debug --service 'http://127.0.0.1:5000' --bouncer-url "http://127.0.0.1:8000/*"
```
nickstenning: Don't bother adding this here. If you want to document it in a subsequent section that explains what the bouncer is and why you might want to build the extension with support for it, that's fine, but not everyone needs to build the extension with support for the bouncer.

h/browser/chrome/manifest.json.jinja2
```diff
@@ -45,5 +45,9 @@
     "lib/*",
     "help/*",
     "content/web/viewer.html"
-  ]
+  ],
+
+  "externally_connectable": {
```
nickstenning: I reckon the default behaviour should be to generate an extension that is *not* externally connectable, rather than to generate one which allows connections from `https://hpt.is`.

h/buildext.py
```diff
@@ -303,6 +304,13 @@ def check_sentry_dsn_is_public(dsn):
...
+                    help="A Chrome extension match pattern that matches the "
+                         "direct-link URLs of the Hypothesis direct-link "
+                         "bouncer service. JavaScript from this site is "
+                         "allowed to send messages to the Chrome extension "
+                         '(optional, default: "https://hpt.is/*")',
+                    default='https://htp.is/*')
```
nickstenning: Maybe `htp` -> `hpt`, but maybe omit the default entirely, as suggested above.

nickstenning:
You could also add the relevant args to the `extensions` task in the Makefile.

h/browser/chrome/lib/install.js
seanh: Done

h/browser/chrome/lib/install.js
seanh: Done

docs/hacking/chrome-extension.rst
seanh: This is for development builds only. I don't think there's any reason not to include local bouncer support in dev builds? Sure not everyone is gonna use it, but when you do need to, it'll work. And putting this here is simpler for the docs and the developer experience than having to find a separate section somewhere else.

h/buildext.py
seanh: Done

h/buildext.py
seanh: Done

h/browser/chrome/manifest.json.jinja2
seanh: Done

seanh:
Added it to the Makefile

h/browser/chrome/lib/install.js
```diff
@@ -14,7 +14,16 @@ var browserExtension = new HypothesisChromeExtension({
 });
 
 browserExtension.listen(window);
+
+// Respond to messages sent by the JavaScript from https://hpt.is.
+// This is how it knows whether the user has this Chrome extension installed.
 chrome.runtime.onInstalled.addListener(onInstalled);
+
+chrome.runtime.onMessageExternal.addListener(
+  function (request, sender, sendResponse) {
```
robertknight: We're not doing any kind of validation here that the message is actually a request from the bouncer. That is the only external thing that is allowed to connect right now, but I think it would make sense to check the message type (if the request is an object and has a particular 'type' property) and send a suitable type in response.

robertknight:
One minor comment. Otherwise LGTM.

docs/hacking/chrome-extension.rst
nickstenning: I'm not sure I agree, but either way it needs to be updated to reflect the correct parameter now.

h/buildext.py
```diff
@@ -274,6 +275,11 @@ def check_sentry_dsn_is_public(dsn):
 
 
 parser = argparse.ArgumentParser('hypothesis-buildext')
+parser.add_argument('--bouncer',
```
nickstenning: For consistency please do what I originally asked, which was to use `dest='bouncer_url'` so that it's referenced as `bouncer_url` internally. It's probably also worth setting `metavar='URL'`.

This should also be moved down with the other parameters which represent service URLs. I know you're only trying to keep things alphabetised as I always moan about that, but this is different, as it's the order in which people see the arguments. Generic args like `--debug` should come first, I think.

h/browser/chrome/lib/install.js
seanh: Added a check for `type`

nickstenning:
LGTM.

PR #2973, opened 2016-02-17
Scrollable groups list

### LINKED ISSUE #2971: Groups list, when long, overflows page and isn't scrollable.
If you are a member of enough groups, the groups dropdown becomes unusable, because some groups end up below the viewport bottom and the list isn't scrollable.

screen shot 2016-02-17 at 17 47 37

nickstenning:
This commit ensures that the groups list is scrollable when it is forced to be smaller than the viewport due to the number of groups in it.

It's not an ideal solution, as in some contexts there isn't much of a hint to the user that the widget is scrollable. It probably beats the existing behaviour, though.

Fixes #2971.

Tagging @robertknight for review.

### DIFF
--- a/h/static/styles/app.scss
+++ b/h/static/styles/app.scss
@@ -109,6 +109,7 @@ body {
 /* The groups dropdown list. */
 
 $group-list-width: 270px;
+$group-list-spacing-below: 50px;
 
 .group-list {
   .dropdown {
@@ -117,6 +118,9 @@ $group-list-width: 270px;
 
   .dropdown-menu {
     width: $group-list-width;
+    max-height: 500px;  // fallback for browsers lacking support for vh/calc
+    max-height: calc(100vh - #{$top-bar-height} - #{$group-list-spacing-below});
+    overflow-y: auto;
 
     .group-name {
       overflow: hidden;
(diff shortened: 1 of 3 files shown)


h/static/styles/app.scss
```diff
@@ -117,6 +118,9 @@ $group-list-width: 270px;
 
   .dropdown-menu {
     width: $group-list-width;
+    max-height: 500px;  // fallback for browsers lacking support for vh/calc
```
robertknight: All of our supported browsers (IE >= 10) support vh/calc according to caniuse

h/static/styles/app.scss
```diff
@@ -117,6 +118,9 @@ $group-list-width: 270px;
...
+    max-height: calc(100vh - #{$top-bar-height} - #{$group-list-spacing-below});
+    overflow-y: scroll;
```
robertknight: I think only showing scrollbars when the group list needs them, using `overflow-y: auto` looks better in desktop browsers. On mobile you'll get the same behavior either way.

robertknight:
@nickstenning - Couple of minor comments. Using 'overflow: auto' being the main one. Otherwise LGTM.

I think the ideal behavior would be to display a shadow at the top/bottom when the content is scrollable similar to excerpts. That's not trivial so I'd leave that for another time.

h/static/styles/app.scss
nickstenning: Do you want me to remove this? Seems like a pretty harmless fallback in any case.

nickstenning:
Updated.

PR #2974, opened 2016-02-17
Move /api/token endpoint into application

nickstenning:
This starts the process of moving all auth-related code out of the `h.api` package and into the application. Token issuance is the domain of the web application, as it will eventually depend on more than just the current authenticated session.

### DIFF
--- a/h/app.py
+++ b/h/app.py
@@ -79,16 +79,6 @@ def includeme(config):
 
     config.include('h.api', route_prefix='/api')
     config.include('h.api.nipsa')
-    config.include('h.db')
-
-    # Override the traversal path for the api index route.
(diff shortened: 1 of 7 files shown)


chdorner:
Looks good to me. It's probably helpful to wait for @seanh's review, since I can't say (yet) in detail if these changes don't break anything.

h/app.py
```diff
@@ -79,16 +79,6 @@ def includeme(config):
 
     config.include('h.api', route_prefix='/api')
     config.include('h.api.nipsa')
-    config.include('h.db')
```
seanh: Wait, why don't we include h.db anymore?

nickstenning: We do. It was duplicated. See line 56.

PR #2975, opened 2016-02-17
Simplify routing

nickstenning:
This PR removes most of the Pyramid traversal routing machinery used by the app and the API. Hopefully this makes it somewhat easier to understand where routes are defined, and it makes URL generation cleaner throughout the application.

Of note:

- All the API views are now named routes, such as `api.annotation` or
  `api.search`. We do still use traversal, but in a limited way, in
  order to provide context objects (and hence access control lists) for
  particular routes. See `h/api/__init__.py` for where the API's named
  routes are connected up to URLs.
- /api now redirects to /api/ by means of the `append_slash=True`
  predicate on our 404 handler. This means that the 404 handler will
  automatically redirect requests lacking a trailing slash to the
  "slashed" version of the URL if it would result in a route match,
  across the application.
- Our API no longer responds to arbitrary additions to the request path
  for its routes. That is, /api/search/foo/bar/baz is now a 404 rather
  than matching the search endpoint. Strictly I suppose this is a
  breaking change, but if you were depending on that behaviour I suspect
  you deserve what's coming to you ;)
- All use of `request.resource_url(...)` throughout the application has
  been replaced by use of `request.route_url(...)`. This results in
  easier-to-understand URL generation, but also means we don't need to
  declare `__name__` and `__parent__` attributes on every traversal context
  factory.
- Rejection of annotation creation requests that try to create
  annotations in groups of which the authenticated user is not a member
  has moved into the annotation validators in `h.api.schemas` rather
  than through a magic ACL.

**N.B.** this PR is based on #2974 to ease review, but it will require a rebase once that's merged.


h/api/models/elastic.py
```diff
@@ -58,7 +64,28 @@ def target_links(self):
     def document(self):
         return self.get("document", {})
 
+    def __acl__(self):
```
chdorner: Is this something we need to add to the Postgres model as well? (Either now or soon)

nickstenning: At some point, yes. Don't worry about it for now. We're going to need to go through and do an audit of all the properties used by the web application and replicate them on the SQLA model at some point.

PR #2978, opened 2016-02-18
Do not clear selection and reset search when creating a reply

robertknight:
This fixes the Reply action reloading the view on the stream if search is non-empty, losing the new reply in the process.

The logic for clearing the selection and search in the sidebar
when creating a new annotation was incorrectly being triggered:

1. On the stream, where creating new annotations is not possible except for replies and
2. For replies. The selection should only be cleared in the sidebar when creating a new annotation with a location on the page.

This PR fixes the issue by moving the logic for clearing the selection from the stream's controller to the sidebar's controller and also but not clearing the selection if the new annotation is a reply

### DIFF
--- a/h/static/scripts/app-controller.coffee
+++ b/h/static/scripts/app-controller.coffee
@@ -125,12 +125,6 @@ module.exports = class AppController
       $scope.search.query = ''
       annotationUI.clearSelectedAnnotations()
 
-    $rootScope.$on('beforeAnnotationCreated', (event, data) ->
-      if data.$highlight
-        return
-      $scope.clearSelection()
-    )
-
     $scope.search =
       query: $location.search()['q']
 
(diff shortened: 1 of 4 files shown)


nickstenning:
Rebased on master.

nickstenning:
So this fixes one issue, which is that the search is reset. But it doesn't fix the fact that clicking on "Reply" seems to be missing a `event.preventDefault()` or something, because it's appending some long random ID to the URL hash. I don't think this serves any purpose and should probably fixed at the same time.

robertknight:
@nickstenning - That "random ID" is the annotation ID, taken from the ID attribute of the `<li>` element in the thread.

The hash change is being triggered by the logic in `thread.coffee` to scroll the sidebar to new annotations: https://github.com/hypothesis/h/blob/master/h/static/scripts/directive/thread.coffee#L232 . I can address that but I'd first like to understand the surrounding code better. Perhaps leave that for a separate PR?

nickstenning:
Possibly, but I'm nearly certain it didn't always behave like this.

robertknight:
You're right, though it was some time ago that the change was introduced in e58add0eb1130b90b97409a974727baf549c6a63 (April 2015)

nickstenning:
Okay. Let's revisit that weird hashchange stuff soon.

PR #2979, opened 2016-02-18
Convert <markdown> component to JS and add basic tests

robertknight:
Whilst triaging the `$sanitize.badparse()` issues in Sentry yesterday I had a brief look into the markdown editor component, which is in a pretty sorry state.

As a step to making changes to it possible, this PR adds some basic tests and converts the existing code to JS.

The conversion was automated using [decaffeinate](https://github.com/decaffeinate/decaffeinate)
with a small amount of subsequent cleanup.

The tests are very basic at the moment and cover
just the read-only/editing states and rendering
of markdown and LaTeX


nickstenning:
Yeah. A bit of cleaning up to do here :)

nickstenning:
There's no way I can review the code, but I've checked it out and Markdown editing seems to work just fine. Merging.

PR #2980, opened 2016-02-18
Annotation card & groups list style cleanup

robertknight:
This is a cleanup of the styles related to the annotation component and groups list, in preparation for direct linking changes.

Commit-by-commit review will probably be easiest.

* Extract group-list styles out into its own component file
* Rename files so that the style file name is consistent with the directive/component name 'annotations' -> 'annotation', 'threads' -> 'thread', 'markdown-editor' -> 'markdown'.
* Remove unused selectors (.privacy, .user)
* Rename 'user-picker' to 'signin-control-menu' and put it with the rest of the signin-control styles
* Use a mixin and @include for dropdown-menu-btn so that the final output styles are predictable. Previously @extends was used but that can silently (without a compile error) give different results depending on whether the extended/extending class is @import-ed first.
* Move styles which are only used in the app out of common.scss into app.scss


nickstenning:
LGTM.

PR #2981, opened 2016-02-18
Simplify routing

nickstenning:
This PR removes most of the Pyramid traversal routing machinery used by the app and the API. Hopefully this makes it somewhat easier to understand where routes are defined, and it makes URL generation cleaner throughout the application.

Of note:

- All the API views are now named routes, such as `api.annotation` or
  `api.search`. We do still use traversal, but in a limited way, in
  order to provide context objects (and hence access control lists) for
  particular routes. See `h/api/__init__.py` for where the API's named
  routes are connected up to URLs.
- /api now redirects to /api/ by means of the `append_slash=True`
  predicate on our 404 handler. This means that the 404 handler will
  automatically redirect requests lacking a trailing slash to the
  "slashed" version of the URL if it would result in a route match,
  across the application.
- Our API no longer responds to arbitrary additions to the request path
  for its routes. That is, /api/search/foo/bar/baz is now a 404 rather
  than matching the search endpoint. Strictly I suppose this is a
  breaking change, but if you were depending on that behaviour I suspect
  you deserve what's coming to you ;)
- All use of `request.resource_url(...)` throughout the application has
  been replaced by use of `request.route_url(...)`. This results in
  easier-to-understand URL generation, but also means we don't need to
  declare `__name__` and `__parent__` attributes on every traversal context
  factory.
- Rejection of annotation creation requests that try to create
  annotations in groups of which the authenticated user is not a member
  has moved into the annotation validators in `h.api.schemas` rather
  than through a magic ACL.


h/api/test/schemas_test.py
```diff
@@ -91,6 +91,44 @@ def test_createannotationschema_ignores_input_user(data, authn_policy):
...
+    # World group
+    ({'group': '__world__'}, [], False),
+    ({'group': '__world__'}, ['group:__world__'], True),
+
+    # Other group
+    ({'group': 'abcdef'}, [], False),
+    ({'group': 'abcdef'}, ['group:__world__'], False),
+    ({'group': 'abcdef'}, ['group:__world__', 'group:abcdef'], True),
+])
+def test_createannotationschema_rejects_annotations_to_other_groups(data,
```
seanh: I've been naming these like `test_CreateAnnotationSchema_*` 'cos I don't think that lowercasing it helps readability (even though pylint will complain). (Or put the tests for a class in a test class.)

nickstenning: This name is consistent with other names in this file.

seanh:
This is a big change, but I think a huge improvement. I love the removal of "hidden magic buried in subtleties of ACL definitions" for stopping people from posting to groups they don't belong to. And just general getting rid of traversal is great, much easier to trace from a URL to a  view callable, and those hierarchies of traversal resource classes gone.

PR #2985, opened 2016-02-19
Allow disabling URI equivalence lookups in the streamer

nickstenning:
This is a distasteful temporary hack to allow us, under heavy load, to prevent the websocket server from spewing errors due to connection pool exhaustion.

Longer-term fixes for the root cause of the problem are in the works, but having this tool available to us in production will be very useful.


chdorner:
Code looks good. Let's hope we don't have to keep this in the code base for long.

PR #2986, opened 2016-02-19
Postgres documents

chdorner:
_tl;dr_ This pull request adds the models and migrations for the new `Document` / `DocumentMeta` / `DocumentURI` models, and generates the data out of each annotation in the migration script.

## New SQLAlchemy models
**Document**
The Document itself that only has an `id` column and timestamps, acts as
the master model to wich both uris and metadata is attached.

**DocumentURI**
A logical document can be accessed by different URIs, all of these URIs
are stored individually as a DocumentURI. Along with other data like a
type and content type.
We use the annotation `target_uri` as the `claimant` and then store each
URI the documents lists as an equivalent URI into the `uri` field, along
with its potential type (e.g. self-claim, or rel attributes) and content type.

**DocumentMeta**
In addition to document uris, we store extra metadata of the document.
There is currently no use of that data yet except for generating
specific DocumentURI models for both dublincore and highwire doi uris.

## Migration changes

Instead of migrating Elasticsearch documents of type `document`, we
decided to simulate the way document-related data will be handled in the
future, by creating or updating Document models after each annotation
has been handled.

Things worth noting:

* Similar to the current Elasticsearch solution, existing documents
  deemed equivalent will be merged automatically without an audit
  process (this will probably change in the future).

* Document links that only have an `href` field and its value is the
  same as the annotation `target_url` are discarded in favour of a
  generated `self-claim` DocumentURI

* Document links starting with `doi:` are discarded in favour of
  generated `dc-doi` and `highwire-doi` from the respective meta data.

* Document links with only an href and a content_type of
  `application/pdf` will get a type of `highwire-pdf`.

* Document links with a `rel` value will be prefixed with `rel-` and
  used as the type.

* The type of DocumentMeta models is the path of nested keys.
    `{ "facebook": { "site_name": "..." } }` vs `facebook.site_name`

### DIFF
--- a/h/api/models/document.py
+++ b/h/api/models/document.py
@@ -2,6 +2,8 @@
 
 from __future__ import unicode_literals
 
+from datetime import datetime
+
 import sqlalchemy as sa
 from sqlalchemy.dialects import postgresql as pg
 from sqlalchemy.ext.hybrid import hybrid_property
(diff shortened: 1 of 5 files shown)


nickstenning:
LGTM.

PR #2987, opened 2016-02-19
Randomize websocket reconnects

nickstenning:
Currently, if the websocket connection fails, then we will always immediately attempt a reconnect. This can easily contribute to a thundering herd problem if all WebSocket clients (or even a large portion of them) are disconnected simultaneously.

This PR ensures that an aborted connection always results in at least a 1s-2s (randomized) delay before initial reconnect, and subsequently continuing randomized exponential backoff.

This PR also changes behaviour -- if, after 10 connection attempts, we still haven't successfully connected, we simply give up. 10 connection attempts with the current parameters corresponds to about 30 minutes.

### DIFF
--- a/h/static/scripts/websocket.js
+++ b/h/static/scripts/websocket.js
@@ -34,7 +34,14 @@ function Socket(url) {
 
   function reconnect() {
     var didConnect = false;
-    var connectOperation = retry.operation();
+    var connectOperation = retry.operation({
+      // Wait 2s before attempting to reconnect
+      minTimeout: 2000,
+      // Don't retry forever
+      retries: 10,
+      // Randomize retry times to minimise the thundering herd effect
+      randomize: true
+    });
     connectOperation.attempt(function (currentAttempt) {
       socket = new WebSocket(url);
       socket.onopen = function (event) {
--- a/h/static/scripts/websocket.js
+++ b/h/static/scripts/websocket.js
@@ -7,6 +7,9 @@ var inherits = require('inherits');
 // see https://developer.mozilla.org/en-US/docs/Web/API/CloseEvent
 var CLOSE_NORMAL = 1000;
 
+// Minimum delay, in ms, before reconnecting after an abnormal connection close.
+var RECONNECT_MIN_DELAY = 1000;
+
 /**
  * Socket is a minimal wrapper around WebSocket which provides:
  *
@@ -25,6 +28,9 @@ function Socket(url) {
   // the current WebSocket instance
   var socket;
 
+  // a pending operation to connect a WebSocket
+  var operation;
+
   function sendMessages() {
     while (messageQueue.length > 0) {
       var messageString = JSON.stringify(messageQueue.shift());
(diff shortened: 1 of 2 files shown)


h/static/scripts/websocket.js
```diff
@@ -32,44 +38,80 @@ function Socket(url) {
...
+
+    operation.attempt(function () {
       socket = new WebSocket(url);
       socket.onopen = function (event) {
-        // signal successful connection
-        connectOperation.retry();
-        didConnect = true;
-        sendMessages();
-
+        onOpen();
```
robertknight: I find it a little confusing here that some of the logic for the open event is in `onOpen()` and some is in the 'onopen' handler.

h/static/scripts/websocket.js
```diff
@@ -32,44 +38,80 @@ function Socket(url) {
...
+  }
+
+  // onAbnormalClose is called when a websocket connection closes abnormally.
+  // This may be the result of a failure to connect, or an abnormal close after
+  // a previous successful connection.
+  function onAbnormalClose(error) {
+    // If we're already in a reconnection loop, trigger a retry...
+    if (operation) {
+      if (!operation.retry(error)) {
+        console.error('reached max retries attempting to reconnect websocket');
```
robertknight: Hmm. We're not clearing `operation` here, even though an operation is no longer in progress after this. Strictly speaking it doesn't matter though because `Socket` doesn't provide a way to initiate a reconnection externally.

h/static/scripts/websocket.js
```diff
@@ -32,44 +38,80 @@ function Socket(url) {
...
       };
     });
-  };
+  }
+
+  // Reconnect the websocket after a short delay. This is used when the
+  // websocket has closed abnormally. If reconnection is already in progress, do
+  // nothing.
+  function reconnect() {
+    if (operation) {
```
robertknight: We're checking if an operation is in progress both here and in `onAbnormalClose`. Given that we only call `reconnect()` in one place, I think it would probably be clearer just to put this logic in `onAbnormalClose`.

robertknight:
The test to verify that no auto-reconnection happens after a normal disconnection will need adjusting given that the timeout is now randomized: https://github.com/hypothesis/h/blob/randomize-websocket-reconnects/h/static/scripts/test/websocket-test.js#L68

h/static/scripts/websocket.js
nickstenning: I'm not sure I understand what's confusing. I've just tried to keep all the references to emitted events in one place, and all the references to `operation` in another. What's your proposal?

h/static/scripts/websocket.js
nickstenning: What should happen if `onclose` happens multiple times after the maximum number of retries has occurred? Printing multiple errors to the console is harmless, while starting a reconnection loop again isn't. This seems to me the safer behaviour.

robertknight: After the maximum number of retries has occurred, presumably we won't be initiating any further connection attempts and so 'onclose' won't be called again?

h/static/scripts/websocket.js
nickstenning: :+1:

nickstenning:
> The test to verify that no auto-reconnection happens after a normal disconnection will need adjusting 

Good catch ✔

dwhly:
Question: At some point should we think about messaging this to the user ("Attempting to reconnect... next try in XX seconds") ending with an appropriate message ("Tried to reconnect for 30 minutes, click here to try again", or "Refresh the page to reconnect" or whatever).

robertknight:
@dwhly - I agree, if real-time comms are not working I'd like to surface some indicator of that in the app, as Trello, Slack and others do. Need design input on how best to do that.

dwhly:
@conordelahunty ^

robertknight:
A technical constraint on this is that we have limited information about the nature of the problem beyond "we are connected/connecting" vs. "the connection was closed or couldn't be established".

conordelahunty:
@nickstenning @robertknight Should we break messaging down into two seperate things. 1) how to display feedback so we can get this piece shipped and the 2) a new card for feedback/system messaging design?

robertknight:
@conordelahunty - I think we just need the one card to present a notice to the user about the status of real-time updates. This PR is essentially a small adjustment to the existing system for automatically reconnecting the WebSocket, so we can ship it already.

PR #2992, opened 2016-02-22
Decouple transaction manager from database session

nickstenning:
In preparation for making some substantial changes to the websocket server...

Move the configuration of the transaction integration into `h.app` rather than the config files, while also ensuring that

- Each request gets a brand new transaction manager rather than reusing
  the transaction manager one per thread.
- The transaction integration ignores assets requests.
- The transaction manager will attempt to retry retryable exceptions.
- The transaction manager will not annotate the transaction with the
  value of request.unauthenticated_user -- we don't use this
  functionality.

The important change comes in f0c96f2, where we make it possible to use the database session without `pyramid_tm` being part of the application. For the websocket server, where there is no request/response cycle, it doesn't makes sense to have a transaction manager responsible for committing the session.

### DIFF
--- a/conf/app.ini
+++ b/conf/app.ini
@@ -30,7 +30,6 @@ mail.default_sender: "Annotation Daemon" <no-reply@localhost>
 pyramid.includes:
     pyramid_mailer
     pyramid_redis_sessions
-    pyramid_tm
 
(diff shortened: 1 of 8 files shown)


chdorner:
Looking good! Great that we get the retry behaviour in :+1: 

:rocket:

PR #2993, opened 2016-02-22
Fix annotation tag links

robertknight:
The ng-href directive does not treat its contents as
an Angular expression.

Fixes a regression introduced in e334fea0241282 and add a test.


nickstenning:
LGTM.

PR #2994, opened 2016-02-22
Add a first pass at a releaser script

nickstenning:
This script automates some of the process of releasing new versions of Hypothesis, including:

- creating the tagged release commit
- pushing the release to GitHub
- adding CHANGES information to the GitHub release
- uploading a source distribution tarball to the GitHub release

### DIFF
--- a/scripts/release
+++ b/scripts/release
@@ -1,0 +1,127 @@
+#!/bin/sh
+
+# GitHub configuration for uploading release tarball. The GitHub token needs at
+# least the `public_repo` OAuth scope.
+: ${GITHUB_TOKEN:=}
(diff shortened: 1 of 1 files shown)


robertknight:
LGTM. If the logic starts to get more complex this might be easier to grok if written in Python.

seanh:
I totally would have written this in Python, not a fan of shell scripts

nickstenning:
It's a script whose primary job is to call out to other executables. That's what shell scripts are good at.

Also, let's work on the basis that automation is better than no automation. This didn't exist before, and now it does. Telling me off for not writing it in Python when I'm the person a) doing the releases and b) writing the script is... less than helpful.

PR #2996, opened 2016-02-23
Streamer work queue

nickstenning:
This is a rewrite of the core message handling routines of the streamer to address a fundamental problem with the previous implementation.

Prior to this commit, WebSocket clients shared a thread-local transaction manager provided by `pyramid_tm`. Access to the database was wrapped by this transaction manager, and under high load it was possible for incoming messages from two different websocket clients (A and B) to trigger the following series of events:

1. Message from A arrives, streamer starts a database transaction.
2. Message from B arrives.
3. Gevent context switches from the greenlet associated with client
        A to the greenlet associated with client B.
4. Message from B queries the database, and reuses the transaction
        originally opened for client A.
5. Processing for B's message completes, and the transaction
        commits.
6. Gevent context switches back to the greenlet associated with
        client A.
7. Processing for A's message completes, and the transaction
        attempts to commit.
8. Code inside `pyramid_tm` explodes because the transaction in
        question no longer exists.

This PR ensures that all incoming messages (whether from WebSocket clients or subscribed NSQ topics) are pushed into a `gevent.queue.Queue`, with a single greenlet responsible for pulling messages out of the queue. This single greenlet is also responsible for ensuring that the database session is appropriately committed/rolled back and closed after each message has been handled.

Since all database access from the streamer is intended to be read-only, this is enforced with an appropriate `SET TRANSACTION` call before starting message processing.

Previously, even relatively light traffic (10 connections opened at once) was enough to precipitate the race condition described above. It is now possible to open thousands of websocket connections to the server with no problems.

----

Notable other changes in this PR:

- `h.streamer` has been split up and is now a package
- I've replaced `python-statsd` with `stats` (see d0c2f65 for an explanation)
- We report stats on number of connected clients and work queue length.

----

This should address most of the issues we've been seeing with elasticsearch connection pool exhaustion in the websocket server.


chdorner:
Awesome work! I find it quite easy to follow the nsq handlers putting messages on the gevent queue, and how those are then getting processed. :star2:

robertknight:
Good work. I'm happy for this to be merged once it has been rebased.

robertknight:
I'm seeing an issue with live updates for group annotations not working properly:

**STR:**
1. Open two browser tabs at example.com
2. Load the bookmarklet on both of them and switch to the same group in both tabs
3. Create a new annotation in one tab, then switch to the other

**Expected:** New annotation appears in both tabs
**Actual:** New annotation does not appear in both tabs

I'm also not seeing session update notifications being delivered via the WS when creating or joining a new group

robertknight:
I'm seeing the above two issues on master as well as in this branch.

robertknight:
I can also reproduce the issue with group join/leave notifications not being received on the production service at https://hypothes.is

robertknight:
@dwhly reported this as #2999

robertknight:
The above problems were caused by me accessing the web service via `localhost:5000` but setting H_WEBSOCKET_URL to point to an IP address instead of a domain, so cookies were not sent when establishing the WS connection. After resolving that, everything worked as expected.

PR #2997, opened 2016-02-23
Fix form button background gradient

robertknight:
Remove unused '...' which were originally variable
arguments to the compass 'linear-gradient()' mixin.

Now we are just using the linear-gradient() CSS function
however so the '...' was never expanded, resulting in
invalid CSS.


nickstenning:
LGTM.

PR #2998, opened 2016-02-23
Remove unnecessary dependency on YAML

nickstenning:
The bundle config file has got immensely simpler as a result of recent work on the front-end build. This means that there's now no justification for using YAML, which requires a 3rd-party library and shared library support.

This commit replaces the YAML-format bundle config file with an INI-format file, for which parsing support exists in the Python standard library.

Why not use JSON for this file? Comments.


robertknight:
> Why not use JSON for this file? Comments.

ConfigParser works fine for me. For this specific use case there is also [commentjson](https://commentjson.readthedocs.org/). If we ever needed config files that were readable from both JS and Python that might be useful.

PR #3000, opened 2016-02-23
Remove angular-animate dependency

robertknight:
There are no references to ng-{enter, leave, move} or
the other hooks that I could find.

angular-animate was added in fdf132efe7d56de34633451c159b05460769a259
but the mentioned CSS classes no longer exist.

This reduces the Angular bundle size by about 30KB.


finitud:
This looks good! :+1:

PR #3002, opened 2016-02-23
Fix the `hypothesis admin` command

seanh:
Call h.db.Session.commit() directly, as request.tm.commit() no longer
works.

### DIFF
--- a/h/script.py
+++ b/h/script.py
@@ -16,6 +16,7 @@ from pyramid.request import Request
 
 from h import __version__
 from h import accounts
+from h import db
 from h.api.search import config as search_config
 from h._compat import PY2, text_type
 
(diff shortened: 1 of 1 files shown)


seanh:
I don't actually know what's going on here but this seems to fix it. I'll investigate a bit as I'd like to understand

seanh:
Giving up for tonight, I think it's something to do with `pyramid_tm` but haven't figured exactly what yet

nickstenning:
Oh, nice catch. This is an unfortunate side-effect of f0c96f29fdff9152b0b264a7062ddcb5be3cf9c6. That commit changed the relationship between the session object and the transaction manager so that it was possible to get a session that wasn't hooked up to the transaction manager (for use in the websocket server).

Unfortunately, that's exactly what's happening here, because nobody accesses the session using `request.db`, which is [where the hookup to the transaction manager happens](https://github.com/hypothesis/h/commit/f0c96f29fdff9152b0b264a7062ddcb5be3cf9c6#diff-a71822fc38f906c83720165de3dca59fR99).

This should be fixed for the whole application by moving the association between the session and the transaction manager into a NewRequest subscriber so that it's guaranteed to happen on every request.

nickstenning:
Closing in favour of #3005.

PR #3004, opened 2016-02-24
Fix docker build

nickstenning:
Some minor packaging cleanup, along with a couple of related improvements to how we build the Dockerfile:

1. Don't invalidate the Python dependencies layer just because
   `package.json`, `gulpfile.js` or `scripts/` changed.
2. Use a prebuilt node-sass binary. The prebuilt binaries supplied by
   the node-sass project aren't compatible with Alpine Linux (which uses
   musl-libc rather than glibc).
3. Make some small tweaks to the gulpfile so we can build just the
   assets needed for the web application (i.e. skipping the extension).


robertknight:
I think it would be useful to have a README alongside the prebuilt node-sass binary with instructions on how to generate a new version for future reference. Otherwise LGTM.

robertknight:
@nickstenning Will add the README after merge.

PR #3005, opened 2016-02-24
Ensure request session is correctly committed in non-web contexts

### LINKED ISSUE #3001: The `hypothesis admin` command no longer works
It seems to run fine (no crash) but the db is not changed.

nickstenning:
f0c96f2 changed the relationship between the transaction manager and the SQLAlchemy session so that it became possible to fetch a session that wasn't managed by the transaction manager. (This is useful for the websocket server, which manages its own database session.)

Unfortunately, this introduced a rather unpleasant bug, whereby if during the course of a request nobody accessed the session as `request.db` (using the base model `query` property exclusively, for example) then the session would never get tied to the transaction manager and would be dropped when the request ended.

This commit fixes that by moving the code that registers the session with the transaction manager into a NewRequest subscriber, which ensures that it happens on every request. This has the downside of requiring a session be constructed for every request, but there's no easy way around this while we use the magic session query property.

Pyramid doesn't run NewRequest subscribers when creating a request manually using `Request.blank` as we currently do in `h.script` and `h.worker`, so these environments must manually commit the open database transaction.

Fixes #3002.
Fixes #3001.

### DIFF
--- a/h/db.py
+++ b/h/db.py
@@ -92,13 +92,11 @@ def make_engine(settings):
     return engine_from_config(settings, 'sqlalchemy.')
 
 
-def _get_request_session(request):
-    # Create a session and register it with a transaction manager, if present.
-    session = Session()
+def register_session_with_tm(request):
+    """Register the current session with the request transaction manager."""
     tm = getattr(request, 'tm', None)
     if tm is not None:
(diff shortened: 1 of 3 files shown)


finitud:
Looks good. Merging.

PR #3008, opened 2016-02-24
Ensure the streamer works when NSQ is namespaced

nickstenning:
This is a bit of a nasty hack to ensure that we correctly process messages which have come from a namespaced NSQ topic (such as in stage, where all topic names are prefixed with 'stage-').

### DIFF
--- a/h/queue.py
+++ b/h/queue.py
@@ -11,8 +11,7 @@ class NamespacedNsqd(object):
         self.namespace = namespace
 
     def publish(self, topic, data):
-        if self.namespace is not None:
-            topic = '{0}-{1}'.format(self.namespace, topic)
+        topic = resolve_topic(topic, namespace=self.namespace)
         if not isinstance(data, str):
             data = json.dumps(data)
         return self.client.publish(topic, data)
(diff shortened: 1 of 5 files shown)


h/streamer/nsq.py
```diff
@@ -79,10 +75,7 @@ def handle_message(message, topic_handlers=None):
     incoming messages and messages to be sent out over the websockets.
     """
     if topic_handlers is None:
```
robertknight: The `topic_handlers` argument should be made non-optional, since the function doesn't do anything useful without it.

h/queue.py
```diff
@@ -48,6 +45,27 @@ def get_writer(settings):
...
+    Return a resolved name for the requested topic.
+
+    This uses the passed `namespace` to resolve the topic name, or,
+    alternatively, a pyramid settings object.
+    """
+    if namespace is not None and settings is not None:
+        raise ValueError('you must provide only one of namespace or settings')
+
+    if settings is not None:
+        ns = settings.get('nsq.namespace')
```
robertknight: Could this just be passed into the function as an argument?

nickstenning: I'm not sure what you mean.

h/streamer/nsq.py
nickstenning: :+1: Pushed.

h/queue.py
nickstenning: Oh, I think I see what you're asking -- no, not really, because the 'nsq.namespace' path is an implementation detail of this module. Asking `h.streamer` to pull it out of the application settings object by name is violating that encapsulation. That's the whole point of this function -- if I wasn't concerned with encapsulating the logic associated with namespacing within this module I could just pull `nsq.namespace` out of the settings object in `h.streamer` and prepend it to the topic names.

finitud:
:+1:

PR #3009, opened 2016-02-24
Fix tags autocomplete dropdown CSS

### LINKED ISSUE #3007: Tags autocomplete dropdown menu is broken
screenshot from 2016-02-24 13-04-01

seanh:
7fff01e changed the annotation cards from `<article>` to `<annotation>`.
This seems to break the CSS of the tags autocomplete dropdown
(particularly: the width is too short) unless the `<annotation>` has
`display: block;`

Fixes #3007.

### DIFF
--- a/h/static/styles/annotation.scss
+++ b/h/static/styles/annotation.scss
@@ -17,6 +17,7 @@ $annotation-card-left-padding: 10px;
 }
 
 .annotation {
+  display: block;
   font-family: $sans-font-family;
   font-weight: 300;
   position: relative;


seanh:
@robertknight I think only you can decide whether this is how you want this fixed

robertknight:
@seanh Yes, you're on the right lines and this is the right fix. So what is happening is this:

1. `<article>` is a block level element by default, ie. it has `display: block` set. `<annotation>` is a custom element (not part of the HTML spec) so it is an inline level element by default (ie. `display:inline`, like `<span>`)
2. The tags autocomplete dropdown (the `<div>` inside the `<auto-complete` element) is an _absolute positioned_ element (ie. `position: absolute`) which means that it is positioned and sized relative to its nearest ancestor that also has `position` set, in this case `<annotation>`, which has `position: relative`. It has `width: 100%` so that means it should take up 100% of the width of the `<annotation>` element.
3. _However_ the `position` attribute only applies to block level elements. Without `display: block`, the `position` property of the `<annotation>` element does nothing. Adding back `display:block` therefore restores the previous behavior.

The fundamental problem here is that there are two UI trees in the application, the tree of Angular components (`<annotation>`, `<markdown>`, `<auto-complete>` etc.) and the resulting HTML which that tree generates (with `<div>`, `<span>` etc.). IMO those should be two separate things and only the second tree should actually appear in the DOM. The first is a conceptual thing that we're concerned about as developers. However, the actual DOM includes both of these. Most of the time you can ignore the elements created by components (eg. `<annotation>`) because they are inline and have no styling applied directly to them. In this case, that's not true however.

seanh:
Cool, thanks for the explanation!

PR #3011, opened 2016-02-24
Postgres convergent evolution + Annotation presentation layer

chdorner:
I didn't anticipate this many changes, so the branch name doesn't really reflect the code changes anymore.

Notable changes:

## Recreate Postgres model interface in the Elasticsearch models

To be able to use models regardless whether the data originally came from Elasticsearch or Postgres, I've re-created the same class structure in the Elasticsearch models. Which means we're having in-memory-only representations of `Document` / `DocumentMeta` / `DocumentURI` classes for Elasticsearch.

An Elasticsearch annotation, which is base structure is a `dict` gets automatically parsed into the aforementioned objects as soon as these properties are accessed.

The normalisation code that used to live in he Postgres migration script (`scripts/migrate.py`) has also moved into the Elasticsearch model, with added tests now.

## API Presenter layer

There is a new module `h.api.presenters` with JSON presenters for the four model classes (`Annotation`, `Document`, `DocumentMeta`, `DocumentURI`).
These are built to return the same JSON structure as we do now (minus the few occasions where we normalise the data, see above).

**Important:** This means that regardless of feature flag, as soon as these code changes get deployed, the `/api/annotations/:id` endpoint will go through the new presentation layer.

## Feature flag for reading from Postgres

There is a feature flag that for reading from Postgres. This currently only has an effect on the `h.api.storage.fetch_annotation` function, all other places (e.g. searching, or loading the created annotation before returning it) will still go Elasticsearch.

### DIFF
--- a/h/api/models/annotation.py
+++ b/h/api/models/annotation.py
@@ -2,6 +2,7 @@
 
 from __future__ import unicode_literals
 
+from pyramid import security
 import sqlalchemy as sa
 from sqlalchemy.dialects import postgresql as pg
 
@@ -74,6 +75,26 @@ class Annotation(Base, mixins.Timestamps):
 
     #: Any additional serialisable data provided by the client.
     extra = sa.Column(pg.JSONB, nullable=True)
+
+    def __acl__(self):
+        """Return a Pyramid ACL for this annotation."""
+        acl = []
+        if self.shared:
+            group = 'group:{}'.format(self.groupid)
+            if self.groupid == '__world__':
+                group = security.Everyone
+
+            acl.append((security.Allow, group, 'read'))
+        else:
+            acl.append((security.Allow, self.userid, 'read'))
+
+        for action in ['admin', 'update', 'delete']:
+            acl.append((security.Allow, self.userid, action))
+
+        # If we haven't explicitly authorized it, it's not allowed.
+        acl.append(security.DENY_ALL)
+
+        return acl
 
     def __repr__(self):
         return '<Annotation %s>' % self.id
--- a/h/api/models/annotation.py
+++ b/h/api/models/annotation.py
@@ -5,6 +5,7 @@ from __future__ import unicode_literals
 from pyramid import security
 import sqlalchemy as sa
 from sqlalchemy.dialects import postgresql as pg
+from sqlalchemy.ext.associationproxy import association_proxy
 from sqlalchemy.ext.hybrid import hybrid_property
 
 from h.api import uri
@@ -79,6 +80,12 @@ class Annotation(Base, mixins.Timestamps):
     #: Any additional serialisable data provided by the client.
     extra = sa.Column(pg.JSONB, nullable=True)
 
+    document_uris = sa.orm.relationship('DocumentURI',
+                                        viewonly=True,
+                                        foreign_keys='DocumentURI.uri_normalized',
+                                        primaryjoin='DocumentURI.uri_normalized == Annotation.target_uri_normalized')
+    documents = association_proxy('document_uris', 'document')
+
     @hybrid_property
     def target_uri(self):
         return self._target_uri
@@ -91,6 +98,11 @@ class Annotation(Base, mixins.Timestamps):
     @hybrid_property
     def target_uri_normalized(self):
         return self._target_uri_normalized
+
+    @hybrid_property
+    def document(self):
+        if self.documents:
+            return self.documents[0]
 
     def __acl__(self):
         """Return a Pyramid ACL for this annotation."""
--- a/h/api/models/annotation.py
+++ b/h/api/models/annotation.py
@@ -5,10 +5,13 @@ from __future__ import unicode_literals
 from pyramid import security
 import sqlalchemy as sa
 from sqlalchemy.dialects import postgresql as pg
+from sqlalchemy.ext.hybrid import hybrid_property
 
+from h.api import uri
 from h.api.db import Base
 from h.api.db import mixins
 from h.api.db import types
+from h._compat import text_type
 
 
 class Annotation(Base, mixins.Timestamps):
@@ -60,9 +63,9 @@ class Annotation(Base, mixins.Timestamps):
                        server_default=sa.sql.expression.false())
 
     #: The URI of the annotated page, as provided by the client.
-    target_uri = sa.Column(sa.UnicodeText)
+    _target_uri = sa.Column('target_uri', sa.UnicodeText)
     #: The URI of the annotated page in normalized form.
-    target_uri_normalized = sa.Column(sa.UnicodeText)
+    _target_uri_normalized = sa.Column('target_uri_normalized', sa.UnicodeText)
     #: The serialized selectors for the annotation on the annotated page.
     target_selectors = sa.Column(types.AnnotationSelectorJSONB,
                                  default=list,
@@ -75,6 +78,19 @@ class Annotation(Base, mixins.Timestamps):
 
     #: Any additional serialisable data provided by the client.
     extra = sa.Column(pg.JSONB, nullable=True)
+
+    @hybrid_property
+    def target_uri(self):
+        return self._target_uri
+
+    @target_uri.setter
+    def target_uri(self, value):
+        self._target_uri = value
+        self._target_uri_normalized = text_type(uri.normalize(value), 'utf-8')
+
+    @hybrid_property
+    def target_uri_normalized(self):
+        return self._target_uri_normalized
 
     def __acl__(self):
         """Return a Pyramid ACL for this annotation."""
--- a/h/api/models/document.py
+++ b/h/api/models/document.py
@@ -27,25 +27,33 @@ class Document(Base, mixins.Timestamps):
         return '<Document %s>' % self.id
 
     @classmethod
-    def find_or_create_by_uris(cls, claimant_uri, uris,
+    def find_by_uris(cls, session, uris):
+        """Find documents by a list of uris."""
+        query_uris = [text_type(uri.normalize(u), 'utf-8') for u in uris]
+
+        matching_claims = (
+            session.query(DocumentURI)
+                   .filter(DocumentURI.uri_normalized.in_(query_uris))
+                   .distinct(DocumentURI.document_id)
+                   .subquery()
+        )
+
+        return session.query(Document).join(matching_claims)
+
+    @classmethod
+    def find_or_create_by_uris(cls, session, claimant_uri, uris,
                                created=None, updated=None):
         """
-        Find or create documents from a claimant uri and a set of uris.
+        Find or create documents from a claimant uri and a list of uris.
 
-        It tries to find a document based on he claimant and the set of uris.
+        It tries to find a document based on the claimant and the set of uris.
         If none can be found it will return a new document with the claimant
-        uri as its only document uri as a self-claim.
+        uri as its only document uri as a self-claim. It is the callers
+        responsibility to create any other document uris.
         """
 
-        normalized_claimant = text_type(uri.normalize(claimant_uri), 'utf-8')
-        query_uris = ([normalized_claimant] +
-                      [text_type(uri.normalize(u), 'utf-8') for u in uris])
-
-        matching_claims = (DocumentURI.query.filter(
-                DocumentURI.uri_normalized.in_(query_uris))
-                .distinct(DocumentURI.document_id).subquery())
-        matching_claims_aliased = aliased(DocumentURI, matching_claims)
-        documents = Document.query.join(matching_claims_aliased)
+        finduris = [claimant_uri] + uris
+        documents = cls.find_by_uris(session, finduris)
 
         if documents.count() == 0:
             doc = Document(created=created, updated=updated)
@@ -55,7 +63,7 @@ class Document(Base, mixins.Timestamps):
                         type='self-claim',
                         created=created,
                         updated=updated)
-            Document.query.session.add(doc)
+            session.add(doc)
 
         return documents
 
--- a/h/api/models/document.py
+++ b/h/api/models/document.py
@@ -26,6 +26,13 @@ class Document(Base, mixins.Timestamps):
     def __repr__(self):
         return '<Document %s>' % self.id
 
+    @property
+    def title(self):
+        titles = [m.value for m in self.meta if m.type == 'title']
+
+        if titles:
+            return titles[0]
+
     @classmethod
     def find_by_uris(cls, session, uris):
         """Find documents by a list of uris."""
--- a/h/api/models/test/annotation_test.py
+++ b/h/api/models/test/annotation_test.py
@@ -1,0 +1,40 @@
+# -*- coding: utf-8 -*-
+
+from __future__ import unicode_literals
+
+from pyramid import security
+
+from h.api.models.annotation import Annotation
+
+
+def test_acl_private():
+    ann = Annotation(shared=False, userid='saoirse')
+    actual = ann.__acl__()
+    expect = [(security.Allow, 'saoirse', 'read'),
+              (security.Allow, 'saoirse', 'admin'),
+              (security.Allow, 'saoirse', 'update'),
+              (security.Allow, 'saoirse', 'delete'),
+              security.DENY_ALL]
+    assert actual == expect
+
+
+def test_acl_world_shared():
+    ann = Annotation(shared=True, userid='saoirse', groupid='__world__')
+    actual = ann.__acl__()
+    expect = [(security.Allow, security.Everyone, 'read'),
+              (security.Allow, 'saoirse', 'admin'),
+              (security.Allow, 'saoirse', 'update'),
+              (security.Allow, 'saoirse', 'delete'),
+              security.DENY_ALL]
+    assert actual == expect
+
+
+def test_acl_group_shared():
+    ann = Annotation(shared=True, userid='saoirse', groupid='lulapalooza')
+    actual = ann.__acl__()
+    expect = [(security.Allow, 'group:lulapalooza', 'read'),
+              (security.Allow, 'saoirse', 'admin'),
+              (security.Allow, 'saoirse', 'update'),
+              (security.Allow, 'saoirse', 'delete'),
+              security.DENY_ALL]
+    assert actual == expect
--- a/h/api/models/test/annotation_test.py
+++ b/h/api/models/test/annotation_test.py
@@ -3,8 +3,34 @@
 from __future__ import unicode_literals
 
 from pyramid import security
+import pytest
 
+from h import db
 from h.api.models.annotation import Annotation
+from h.api.models.document import Document, DocumentURI
+
+
+annotation_fixture = pytest.mark.usefixtures('annotation')
+
+
+@annotation_fixture
+def test_document(annotation):
+    document = Document(uris=[DocumentURI(claimant=annotation.target_uri,
+                                          uri=annotation.target_uri)])
+    db.Session.add(document)
+    db.Session.flush()
+
+    assert annotation.document == document
+
+
+@annotation_fixture
+def test_document_not_found(annotation):
+    document = Document(uris=[DocumentURI(claimant='something-else',
+                                          uri='something-else')])
+    db.Session.add(document)
+    db.Session.flush()
+
+    assert annotation.document is None
 
 
 def test_acl_private():
@@ -38,3 +64,12 @@ def test_acl_group_shared():
               (security.Allow, 'saoirse', 'delete'),
               security.DENY_ALL]
     assert actual == expect
+
+
+@pytest.fixture
+def annotation():
+    ann = Annotation(userid="testuser", target_uri="http://example.com")
+
+    db.Session.add(ann)
+    db.Session.flush()
+    return ann
--- a/h/api/models/test/document_test.py
+++ b/h/api/models/test/document_test.py
@@ -1,5 +1,7 @@
 # -*- coding: utf-8 -*-
 
+from __future__ import unicode_literals
+
 import pytest
 
 from h import db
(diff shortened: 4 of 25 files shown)


h/api/storage.py
```diff
@@ -140,3 +146,7 @@ def _prepare(annotation):
     transform.normalize_annotation_target_uris(annotation)
 
     transform.add_nipsa(annotation)
+
+
+def _postgres_enabled():
+    return pyramid.threadlocal.get_current_request().feature('postgres')
```
seanh: It's a shame that you have to do the `threadlocal.get_current_request()` thing (I think of that as an advanced/internal Pyramid detail, better to avoid if you can), but I guess that is the right way, unless we want to add `request` as a param to each function in `storage`

chdorner: I agree. According to @nickstenning, it's bad but okay in this case as it will only be in the code base for a short amount of time. As soon as we migrated and switched over the production service, this will be removed (along with a lot of other code).

h/api/models/document.py
```diff
@@ -26,17 +26,18 @@ class Document(Base, mixins.Timestamps):
...
+    def title(self):
+        meta = DocumentMeta.query.filter(DocumentMeta.document_id == self.id,
+                                         DocumentMeta.type == 'title').first()
+        if meta:
+            return meta.value
+
     @classmethod
-    def find_or_create_by_uris(cls, claimant_uri, uris,
-                               created=None, updated=None):
+    def find_by_uris(cls, claimant_uri, uris=[]):
```
nickstenning: Is there any point making a distinction between `claimant_uri` and `uris` in this method? It's needed in the `find_or_create` version, but I don't think it's necessary here.

h/api/models/document.py
```diff
@@ -26,17 +26,18 @@ class Document(Base, mixins.Timestamps):
     def __repr__(self):
         return '<Document %s>' % self.id
 
+    @property
+    def title(self):
+        meta = DocumentMeta.query.filter(DocumentMeta.document_id == self.id,
```
nickstenning: I'd suggest just doing this kind of thing Python-side using `self.meta`.

h/api/models/document.py
```diff
@@ -26,17 +26,18 @@ class Document(Base, mixins.Timestamps):
...
+                                         DocumentMeta.type == 'title').first()
+        if meta:
+            return meta.value
+
     @classmethod
-    def find_or_create_by_uris(cls, claimant_uri, uris,
```
nickstenning: As mentioned in Slack, I'm going to suggest we make a hard break with the way that these kinds of methods are implemented in the rest of the application, and switch to having the convention be that classmethod helpers like these accept the database session as their first argument, rather than using the base class query property. As a result:

- they'll be easier to test
- there won't be any problem with making the session and/or transaction manager entirely request-scoped (rather than thread-local)
- we won't have to do any silliness like [this](https://github.com/hypothesis/h/blob/141e61148fedbbdfb964af884384569569ac5c39/h/db.py#L106) or [this](https://github.com/hypothesis/h/blob/141e61148fedbbdfb964af884384569569ac5c39/scripts/migrate.py#L64).

h/api/models/test/document_test.py
```diff
@@ -7,15 +9,66 @@
 from h.api.models.document import merge_documents
 
 
+def test_document_title():
+    doc = Document()
+    meta_title = DocumentMeta(type='title', value='The Title', document=doc, claimant='http://example.com')
```
nickstenning: This is absolutely fine, I just want to note FYI that you can also do stuff like

```python
doc.meta.append(DocumentMeta(type='title', value='The Title', claimant='http://example.com'))
```

if you prefer.

h/api/presenters.py
```diff
@@ -0,0 +1,136 @@
...
+"""
+
+import collections
+
+
+class AnnotationJSONPresenter(object):
+    def __init__(self, annotation):
+        self.annotation = annotation
+
+    def dict(self):
```
nickstenning: `asdict` would be the conventional/idiomatic name for this method.

chdorner: Fixed

h/api/models/document.py
chdorner: Good point. Fixed the method to only accept a list of uris.

h/api/presenters.py
```diff
@@ -0,0 +1,145 @@
...
+
+        return {'read': [read],
+                'admin': [self.annotation.userid],
+                'update': [self.annotation.userid],
+                'delete': [self.annotation.userid]}
+
+    @property
+    def target(self):
+        return [{'source': self.annotation.target_uri,
+                 'selector': self.annotation.target_selectors or {}}]
```
nickstenning: Should default to empty list rather than empty object.

h/api/presenters.py
```diff
@@ -0,0 +1,145 @@
...
+
+        return d
+
+
+class DocumentMetaJSONPresenter(object):
+    def __init__(self, document_meta):
+        self.document_meta = document_meta
+
+    def asdict(self):
+        reversed_path = self.document_meta.type.split('.')[::-1]
```
nickstenning: Maybe a comment to explain this implementation :stuck_out_tongue:

h/api/models/document.py
chdorner: Fixed.

h/api/models/document.py
chdorner: Fixed for both `find_by_uris` and `find_or_create_by_uris`.

What do you think about `api.storage`, should it be passed-in there as well? Especially now that we're querying the database in there.

h/api/presenters.py
chdorner: Fixed.

h/api/presenters.py
chdorner: Fixed.

h/api/models/document.py
```diff
@@ -26,26 +26,40 @@ class Document(Base, mixins.Timestamps):
...
+    @property
+    def title(self):
+        titles = [m.value for m in self.meta if m.type == 'title']
+
+        if titles:
+            return titles[0]
+
     @classmethod
-    def find_or_create_by_uris(cls, claimant_uri, uris,
+    def find_by_uris(cls, db, uris):
```
nickstenning: Sorry to be a PITA, but I think we should call this argument `session`.

h/api/models/document.py
nickstenning: I'd suggest that given they're going to need access to the features fetcher, perhaps we should just add `request` to the start of every `storage` API function's argument list?

h/api/storage.py
chdorner: @seanh we now pass the `request` into the storage functions because we also want to query database specifically from the passed-in DB session :+1:

h/api/models/document.py
chdorner: Fixed.

h/api/models/document.py
chdorner: No worries, good naming is important. Fixed.

h/api/storage.py
```diff
@@ -37,10 +36,13 @@ def fetch_annotation(id):
     :returns: the annotation, if found, or None.
     :rtype: dict, NoneType
     """
-    return models.Annotation.fetch(id)
+    if _postgres_enabled(request):
```
nickstenning: Indirection through `_postgres_enabled` probably isn't worth it any more. Just do `if request.feature('postgres_read'):`.

nickstenning: Hmm, although I see you're using `_postgres_enabled` as a test seam. Happy to leave this as-is for now.

h/api/models/annotation.py
```diff
@@ -75,5 +80,53 @@ class Annotation(Base, mixins.Timestamps):
...
+    def target_uri(self, value):
+        self._target_uri = value
+        self._target_uri_normalized = text_type(uri.normalize(value), 'utf-8')
+
+    @hybrid_property
+    def target_uri_normalized(self):
+        return self._target_uri_normalized
+
+    @hybrid_property
+    def document(self):
```
nickstenning: This method is introduced in 60edf9c but I reckon it should arrive in 4690f3a.

h/api/models/document.py
```diff
@@ -26,26 +26,40 @@ class Document(Base, mixins.Timestamps):
...
+
     @classmethod
-    def find_or_create_by_uris(cls, claimant_uri, uris,
+    def find_by_uris(cls, session, uris):
+        """
+        Find documents by a list of uris.
+        """
+        query_uris = [text_type(uri.normalize(u), 'utf-8') for u in uris]
+
+        matching_claims = (session.query(DocumentURI).filter(
```
nickstenning: It's a small thing, but this could potentially be made a little clearer, formatting-wise:

```python
matching_claims = (
    session.query(DocumentURI)
        .filter(DocumentURI.uri_normalized.in_(query_uris))
        .distinct(DocumentURI.document_id)
        .subquery()
    )
```

h/api/models/annotation.py
```diff
@@ -75,5 +80,53 @@ class Annotation(Base, mixins.Timestamps):
...
+        uris = self.document_uris
+        if uris:
+            return uris[0].document
+
+    def __acl__(self):
+        """
+        Return a Pyramid ACL for this annotation.
```
nickstenning: One-line docstrings should go on one line:

```
def __acl__(self):
    """Return a Pyramid ACL for this annotation."""
```

h/api/models/document.py
```diff
@@ -26,26 +26,40 @@ class Document(Base, mixins.Timestamps):
...
+                DocumentURI.uri_normalized.in_(query_uris))
+                .distinct(DocumentURI.document_id).subquery())
+        matching_claims_aliased = aliased(DocumentURI, matching_claims)
```
nickstenning: As discussed, we don't need to make an aliased object here because we're not expecting items from the subquery in the resultset (i.e. we're not calling `session.query(Document, matching_claims_aliased)`).

h/api/models/document.py
```diff
@@ -26,26 +26,40 @@ class Document(Base, mixins.Timestamps):
...
+        return session.query(Document).join(matching_claims_aliased)
+
+    @classmethod
+    def find_or_create_by_uris(cls, session, claimant_uri, uris,
                                created=None, updated=None):
         """
-        Find or create documents from a claimant uri and a set of uris.
+        Find or create documents from a claimant uri and a list of uris.
 
         It tries to find a document based on he claimant and the set of uris.
```
nickstenning: s/he/the/

h/api/models/document.py
```diff
@@ -26,26 +26,40 @@ class Document(Base, mixins.Timestamps):
...
         If none can be found it will return a new document with the claimant
         uri as its only document uri as a self-claim.
```
nickstenning: It might be worth explicitly noting here that it is up to the caller to create appropriate DocumentURI claims for the *other* URIs (`uris`) if necessary.

h/api/models/annotation.py
```diff
@@ -75,5 +80,53 @@ class Annotation(Base, mixins.Timestamps):
...
+        self._target_uri = value
+        self._target_uri_normalized = text_type(uri.normalize(value), 'utf-8')
+
+    @hybrid_property
+    def target_uri_normalized(self):
+        return self._target_uri_normalized
+
+    @hybrid_property
+    def document(self):
+        uris = self.document_uris
```
nickstenning: Given the association proxy we have above, this might be clearer as:

```python
def document(self):
    if self.documents:
        return self.documents[0]
```

h/api/storage.py
chdorner: Being able to mock it out easily in test without affecting other features (as we sometimes do in tests) is the reason why I did it as a private method.

h/api/models/test/annotation_test.py
```diff
@@ -0,0 +1,75 @@
...
+    expect = [(security.Allow, 'group:lulapalooza', 'read'),
+              (security.Allow, 'saoirse', 'admin'),
+              (security.Allow, 'saoirse', 'update'),
+              (security.Allow, 'saoirse', 'delete'),
+              security.DENY_ALL]
+    assert actual == expect
+
+
+@pytest.fixture
+def annotation(request):
```
nickstenning: `request` isn't used here.

h/api/models/test/elastic_test.py
```diff
@@ -7,117 +10,463 @@
...
-                    new_callable=PropertyMock)
-    request.addfinalizer(patcher.stop)
-    return patcher.start()
+from h.api.models.elastic import Document, DocumentMeta, DocumentURI
+
+
+class TestAnnotation(object):
+    @pytest.mark.parametrize('annotation', [
+        Annotation({'created': '2016-02-25T16:45:23.371848+00:00'}),
+        Annotation({'created': '2016-02-25T16:45:23.371848+00:00'})])
```
nickstenning: One of these shouldn't have a timestamp on it, I guess?

h/api/models/test/elastic_test.py
```diff
@@ -7,117 +10,463 @@
...
+    @pytest.mark.parametrize('annotation', [
+        Annotation({'created': 'invalid'}),
+        Annotation({'created': ''}),
+        Annotation({'created': None})])
+    def test_created_invalid(self, annotation):
+        assert annotation.created is None
+
+    @pytest.mark.parametrize('annotation', [
+        Annotation({'updated': '2016-02-25T16:45:23.371848+00:00'}),
+        Annotation({'updated': '2016-02-25T16:45:23.371848+00:00'})])
```
nickstenning: One of these shouldn't have a timestamp on it, I guess?

h/api/models/annotation.py
chdorner: Fixed.

h/api/models/annotation.py
chdorner: Fixed.

h/api/models/test/elastic_test.py
```diff
@@ -7,117 +10,463 @@
...
+                    DocumentMeta({'type': 'og.url', 'value': ['http://example.com'],
+                                  'claimant': 'http://example.com',
+                                  'created': datetime.datetime(2016, 2, 25, 16, 45, 23, 371848),
+                                  'updated': datetime.datetime(2016, 2, 25, 16, 45, 23, 371849)}),
+
+                    DocumentMeta({'type': 'title', 'value': ['Example Page'],
+                                  'claimant': 'http://example.com',
+                                  'created': datetime.datetime(2016, 2, 25, 16, 45, 23, 371848),
+                                  'updated': datetime.datetime(2016, 2, 25, 16, 45, 23, 371849)})]
+        assert doc.meta.sort() == expected.sort()
```
nickstenning: You're looking for `assert sorted(doc.meta) == sorted(expected)`.

h/api/models/test/elastic_test.py
```diff
@@ -7,117 +10,463 @@
...
+                                 'type': 'highwire-pdf',
+                                 'content_type': 'application/pdf',
+                                 'created': None, 'updated': None}),
+
+                    DocumentURI({'claimant': 'http://example.com',
+                                 'uri': 'http://example.com',
+                                 'type': 'self-claim',
+                                 'created': None, 'updated': None})]
+
+        assert doc.uris.sort() == expected.sort()
```
nickstenning: Ditto with `sorted(...)`.

h/api/models/test/elastic_test.py
```diff
@@ -7,117 +10,463 @@
...
+                                 'uri': 'https://example.com',
+                                 'type': 'rel-canonical',
+                                 'created': None, 'updated': None}),
+
+                    DocumentURI({'claimant': 'http://example.com',
+                                 'uri': 'http://example.com',
+                                 'type': 'self-claim',
+                                 'created': None, 'updated': None})]
+
+        assert doc.uris.sort() == expected.sort()
```
nickstenning: Ditto with `sorted(...)`.

h/api/models/document.py
chdorner: Fixed.

h/api/models/test/elastic_test.py
```diff
@@ -7,117 +10,463 @@
...
+    def test_claimant_normalized(self):
+        meta = DocumentMeta({'claimant': 'http://example.com'})
+        assert meta.claimant_normalized == 'http://example.com'
+
+    def test_type(self):
+        meta = DocumentMeta({'type': 'title'})
+        assert meta.type == 'title'
+
+    def test_type_normalizes_multiple_dots(self):
+        meta = DocumentMeta({'type': 'dc..description'})
```
nickstenning: :tada:

h/api/models/test/elastic_test.py
```diff
@@ -7,117 +10,463 @@
...
+        assert meta.type == 'dc.description'
+
+    def test_type_normalizes_case(self):
+        meta = DocumentMeta({'type': 'dc.Contributor.Sponsor'})
```
nickstenning: :balloon:

h/api/models/test/elastic_test.py
```diff
@@ -7,117 +10,463 @@
...
+        assert meta.type == 'dc.contributor.sponsor'
+
+    def test_type_normalizes_colons(self):
+        meta = DocumentMeta({'type': 'facebook.book:isbn'})
```
nickstenning: :sparkles:

h/api/models/document.py
chdorner: Fixed.

h/api/models/annotation.py
chdorner: Fixed.

h/api/models/test/elastic_test.py
```diff
@@ -7,117 +10,463 @@
...
+    def test_updated(self):
+        meta = DocumentMeta({'updated': datetime.datetime(2016, 2, 25, 16, 45, 23, 371848)})
+        assert meta.updated == datetime.datetime(2016, 2, 25, 16, 45, 23, 371848)
+
+    def test_claimant(self):
+        meta = DocumentMeta({'claimant': 'http://example.com'})
+        assert meta.claimant == 'http://example.com'
+
+    def test_claimant_normalized(self):
+        meta = DocumentMeta({'claimant': 'http://example.com'})
```
nickstenning: Want to check that it actually gets normalised? (Don't bother mocking out the normaliser... just stick a trailing slash on it or something.)

h/api/models/test/elastic_test.py
```diff
@@ -7,117 +10,463 @@
...
+    def test_updated(self):
+        docuri = DocumentURI({'updated': datetime.datetime(2016, 2, 25, 16, 45, 23, 371848)})
+        assert docuri.updated == datetime.datetime(2016, 2, 25, 16, 45, 23, 371848)
+
+    def test_claimant(self):
+        docuri = DocumentURI({'claimant': 'http://example.com'})
+        assert docuri.claimant == 'http://example.com'
+
+    def test_claimant_normalized(self):
+        docuri = DocumentURI({'claimant': 'http://example.com'})
```
nickstenning: Want to check that it actually gets normalised? (Don't bother mocking out the normaliser... just stick a trailing slash on it or something.)

h/api/models/test/elastic_test.py
```diff
@@ -7,117 +10,463 @@
...
+        assert docuri.claimant_normalized == 'http://example.com'
+
+    def test_uri(self):
+        docuri = DocumentURI({'uri': 'http://example.com'})
+        assert docuri.uri == 'http://example.com'
+
+    def test_uri_normalized(self):
+        docuri = DocumentURI({'uri': 'http://example.com'})
```
nickstenning: Ditto -- normalisation.

h/api/models/document.py
chdorner: Fixed.

h/api/models/document.py
chdorner: Fixed.

h/api/models/test/annotation_test.py
chdorner: Fixed.

h/api/models/test/elastic_test.py
chdorner: Yeah, one of them should not have the timezone. Fixed.

h/api/models/test/elastic_test.py
chdorner: Same as above. And fixed.

h/api/presenters.py
```diff
@@ -0,0 +1,148 @@
...
+        return d
+
+
+class DocumentURIJSONPresenter(object):
+    def __init__(self, document_uri):
+        self.document_uri = document_uri
+
+    def asdict(self):
+        return {'href': self.document_uri.uri,
+                'rel': self.rel,
```
nickstenning: Probably worth omitting `rel` and `type` altogether if they're missing or a weird type, as that'll be more consistent with the existing output of the API.

h/api/models/test/elastic_test.py
chdorner: Fixed.

h/api/models/test/elastic_test.py
chdorner: Fixed.

h/api/models/test/elastic_test.py
chdorner: Fixed.

h/api/models/test/elastic_test.py
chdorner: Fixed.

h/api/models/test/elastic_test.py
chdorner: Fixed.

h/api/models/test/elastic_test.py
chdorner: Fixed.

h/api/presenters.py
chdorner: Fixed.

nickstenning:
LGTM. *So* many things we have coming up will be made easier by the presenter layer work you've done here. Awesome stuff!

PR #3012, opened 2016-02-24
Add extra context to sentry logging

finitud:
https://trello.com/c/GlY7VGl6/266-add-data-about-message-queue-etc-to-sentry-context-in-workers

### DIFF
--- a/h/queue.py
+++ b/h/queue.py
@@ -17,7 +17,7 @@ class NamespacedNsqd(object):
         return self.client.publish(topic, data)
 
 
-def get_reader(settings, topic, channel):
+def get_reader(req, topic, channel):
     """
     Get a :py:class:`gnsq.Reader` instance configured to connect to the
     nsqd reader addresses specified in settings. The reader will read from
@@ -26,18 +26,20 @@ def get_reader(settings, topic, channel):
     The caller is responsible for adding appropriate `on_message` hooks and
     starting the reader.
     """
+    settings = req.registry.settings
     addrs = aslist(settings.get('nsq.reader.addresses', 'localhost:4150'))
     topic = resolve_topic(topic, settings=settings)
     reader = gnsq.Reader(topic, channel, nsqd_tcp_addresses=addrs)
     return reader
 
 
-def get_writer(settings):
+def get_writer(req):
     """
     Get a :py:class:`gnsq.Nsqd` instance configured to connect to the nsqd
     writer address configured in settings. The writer communicates over the
     nsq HTTP API and does not hold a connection open to the nsq instance.
     """
+    settings = req.registry.settings
     ns = settings.get('nsq.namespace')
     addr = settings.get('nsq.writer.address', 'localhost:4151')
     hostname, port = addr.split(':', 1)
@@ -68,8 +70,8 @@ def resolve_topic(topic, namespace=None, settings=None):
 
 def includeme(config):
     config.add_request_method(
-        lambda req, t, c: get_reader(req.registry.settings, t, c),
+        lambda req, t, c: get_reader(req, t, c),
         name='get_queue_reader')
     config.add_request_method(
-        lambda req: get_writer(req.registry.settings),
+        lambda req: get_writer(req),
         name='get_queue_writer')
--- a/h/queue.py
+++ b/h/queue.py
@@ -92,9 +92,5 @@ def resolve_topic(topic, namespace=None, settings=None):
 
 
 def includeme(config):
-    config.add_request_method(
-        lambda req, t, c: get_reader(req, t, c),
-        name='get_queue_reader')
-    config.add_request_method(
-        lambda req: get_writer(req),
-        name='get_queue_writer')
+    config.add_request_method(get_reader, name='get_queue_reader')
+    config.add_request_method(get_writer, name='get_queue_writer')
--- a/h/queue.py
+++ b/h/queue.py
@@ -42,7 +42,8 @@ def get_reader(request, topic, channel):
             sentry.captureException(exc_info=True, extra=extra)
 
         def _capture_error(error):
-            sentry.captureException(exc_info=True, extra=extra)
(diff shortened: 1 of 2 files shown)


conf/development-app-sentry.ini
```diff
@@ -0,0 +1,109 @@
+[pipeline:main]
```
nickstenning: I guess you only used this to test Sentry in development? If so it can probably be omitted from these changes.

h/queue.py
```diff
@@ -17,7 +17,14 @@ def publish(self, topic, data):
         return self.client.publish(topic, data)
 
 
-def get_reader(settings, topic, channel):
+def extra_context_data(req, topic=''):
```
nickstenning: The request object should usually be called `request` in full, for the sake of consistency with other APIs that have a similar structure.

h/queue.py
```diff
@@ -17,7 +17,14 @@ def publish(self, topic, data):
...
+    return {
+        'extra_context': True,
+        'req': req,
+        'topic': topic,
```
nickstenning: Probably just want `topic` here?

h/queue.py
```diff
@@ -26,18 +33,22 @@ def get_reader(settings, topic, channel):
     The caller is responsible for adding appropriate `on_message` hooks and
     starting the reader.
     """
+    req.sentry.extra_context(extra_context_data(req, topic))
```
nickstenning: Two things:

1. Let's check if `request.sentry` exists before using it. Something like:

    ```python
sentry = getattr(request, 'sentry', None)
if sentry is not None:
        sentry.extra_context(...)
    ```
2. Given we just have the topic and channel, here you can probably just inline this:
  
  ```python
sentry.extra_context({'topic': topic, 'channel': channel})
```

File: h/queue.py [hunk 5]
```diff
@@ -26,18 +33,22 @@ def get_reader(settings, topic, channel):
...
 
 
-def get_writer(settings):
+def get_writer(req):
     """
     Get a :py:class:`gnsq.Nsqd` instance configured to connect to the nsqd
     writer address configured in settings. The writer communicates over the
     nsq HTTP API and does not hold a connection open to the nsq instance.
     """
+    req.sentry.extra_context(extra_context_data(req))
```
nickstenning: We don't really have any context in this scenario, because a writer is ephemeral and can be used to write to any topic.

h/sentry.py
```diff
@@ -31,6 +31,11 @@ def user_context_data(request):
         'ip_address': request.client_addr,
     }
 
+def extra_context_data(request):
```
nickstenning: Not sure what this is doing here. Does it do anything or was it just for testing?

conf/development-app-sentry.ini
finitud: Was thinking of keeping it around to save pain if this needs to be debugged again, but you're right, probably by the time we need this file again it'll have diverged from the normal `development` config file.

nickstenning:
Need any more input here? Here's what I'd suggest:

1. write a function called `_capture` that closes over the sentry client and the topic when a reader is created, and does something like:

```python
...
sentry = getattr(request, 'sentry', None)
topic = ...
reader = ...

if sentry is not None:
    function _capture(message, error):
        extra = {'topic': topic}
        if message is not None:
            extra['message'] = message.body
        sentry.captureException(exc_info=True, extra=extra)
    reader.on_exception.connect(_capture, weak=False)
```

Make sense?

h/queue.py
finitud: Should this one go then?

h/sentry.py
finitud: I'm just unsure of whether adding the extra context in `h/queue.py` suffices or it's needed here as well.

finitud:
I have added @nickstenning's suggestions and done some clean up on this PR.

h/queue.py
```diff
@@ -26,18 +26,43 @@ def get_reader(settings, topic, channel):
...
+    if sentry is not None:
+        extra = {'topic': topic}
+
+        def _capture_exception(message, error):
+            if message is not None:
+                extra['message'] = message.body
+            sentry.captureException(exc_info=True, extra=extra)
+
+        def _capture_error(error):
+            sentry.extra_context(extra)
```
nickstenning: Does this actually capture the error? From the Sentry docs it looks like it wouldn't.

h/queue.py
```diff
@@ -26,18 +26,43 @@ def get_reader(settings, topic, channel):
...
+
+        def _capture_message(message):
+            if message is not None:
+                extra['message'] = message.body
+            sentry.extra_context(extra)
```
nickstenning: Same question here. Would this actually capture a message to Sentry?

h/queue.py
```diff
@@ -68,8 +93,8 @@ def resolve_topic(topic, namespace=None, settings=None):
 
 def includeme(config):
     config.add_request_method(
-        lambda req, t, c: get_reader(req.registry.settings, t, c),
+        lambda req, t, c: get_reader(req, t, c),
```
nickstenning: Now that these functions take a request as their first argument, this can be written more straightforwardly as:

```python
config.add_request_method(get_reader, name='get_queue_reader')
```

h/test/queue_test.py
```diff
@@ -77,11 +97,12 @@ def test_get_writer_namespace(fake_nsqd):
...
+    req = DummyRequest()
+    req.registry.settings = {
         'nsq.namespace': 'abc',
     }
-    writer = queue.get_writer(settings)
+    fake_client = fake_nsqd.return_value
+    writer = queue.get_writer(req)
     writer.publish('sometopic', {
         'key': 'value',
     })
```
nickstenning: So you've updated the tests to account for taking `request` instead of a settings object, but it would be good to test that the sentry integration actually does what we think, no?

nickstenning:
LGTM.

PR #3015, opened 2016-02-24
Enable admins to activate users

seanh:
Add an "Activate" button next to users on the /admin/users page so that
admins can activate user accounts.

Some refactoring to avoid code duplication between activating by users
clicking on an activation link and admins clicking the new activate button:
move the code that actually activates the user onto the User model where
both views can call it.

### DIFF
--- a/h/accounts/models.py
+++ b/h/accounts/models.py
@@ -132,6 +132,11 @@ class User(Base):
 
         return False
 
+    def activate(self):
+        """Activate the user by deleting any activation they have."""
+        session = sa.orm.object_session(self)
+        session.delete(self.activation)
+
     # Hashed password
     _password = sa.Column('password', sa.UnicodeText(), nullable=False)
     # Password salt
--- a/h/accounts/test/models_test.py
+++ b/h/accounts/test/models_test.py
@@ -132,3 +132,17 @@ def test_staff_members_does_not_return_non_staff_users():
 
     for non_staff in non_staff:
         assert non_staff not in staff
+
+
+def test_User_activate_activates_user():
+    user = models.User(username='kiki', email='kiki@kiki.com',
+                       password='password')
+    activation = models.Activation()
+    user.activation = activation
+    db.Session.add(user)
+    db.Session.flush()
+
+    user.activate()
+    db.Session.commit()
+
+    assert user.is_activated
--- a/h/accounts/test/views_test.py
+++ b/h/accounts/test/views_test.py
@@ -691,10 +691,9 @@ class TestActivateController(object):
         with pytest.raises(httpexceptions.HTTPNotFound):
             views.ActivateController(request).get_when_not_logged_in()
 
-    def test_get_when_not_logged_in_successful_deletes_activation(
+    def test_get_when_not_logged_in_successful_activates_user(
             self,
-            user_model,
-            activation_model):
+            user_model):
         request = DummyRequest(matchdict={'id': '123', 'code': 'abc456'})
         request.db.delete = mock.create_autospec(request.db.delete,
                                                  return_value=None)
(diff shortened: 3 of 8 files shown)


seanh:
screenshot from 2016-02-24 17-50-02

* * *

screenshot from 2016-02-24 17-50-17

lenazun:
👏

h/accounts/models.py
```diff
@@ -132,6 +133,11 @@ def is_activated(self):
 
         return False
 
+    def activate(self, request):
+        """Activate the user by deleting any activation they have."""
+        request.db.delete(self.activation)
```
nickstenning: This is probably better done as:

```python
session = sa.orm.object_session(self)
session.delete(self.activation)
```

h/accounts/models.py
```diff
@@ -132,6 +133,11 @@ def is_activated(self):
...
+        request.registry.notify(events.ActivationEvent(request, self))
```
nickstenning: This needs to go somewhere else. SQLAlchemy model objects shouldn't be triggering Pyramid events, not least because they require the registry to do so. If that means that this one line is duplicated in both `admin.py` and `accounts/views.py` for now, then that's absolutely fine.

h/accounts/test/models_test.py
```diff
@@ -132,3 +134,61 @@ def test_staff_members_does_not_return_non_staff_users():
...
+
+
+@activate_fixtures
+def test_User_activate_deletes_activation():
+    request = mock_request()
+    user = factories.User()
+
+    user.activate(request)
+
+    request.db.delete.assert_called_once_with(user.activation)
```
nickstenning: I'd suggest not attempting to mock out the database session in model tests. Just use the database. Set up some data, run some functions, and then verify the state of the database.

nickstenning:
Looks good, although I don't think moving the triggering of the `ActivationEvent` into the model is the right thing to do. Move that out and you can just have an `.activate()` method on the user that doesn't require any arguments.

h/accounts/models.py
seanh: Although we need `request` anyway to send the notification, but see below...

h/accounts/models.py
seanh: I can see the ugliness of doing this in the model class, but I weighed that against duplicating it and chose this one, as it's only this one method that needs to have request passed into it.

The benefit if having a simple `user.activate()` method seems lost to me if you have to call `request.registry.notify(events.ActivationEvent(request, self))` after every time you call `activate()` (and have to add a test for that each time too).

So it seems to come down to wanting model classes to be pure vs wanting to avoid the duplication.

I guess the way to have both things (no duplication, and keep the model class pure) would be to introduce something like an `h.accounts.logic` module to hold an `activate()` function that could be called from multiple different places. But I figured the chances of me getting away with adding that were low :)

I guess th

nickstenning: We're talking about duplication of a trivial piece of code in two places. It's not even a disaster if someone forgets to call it. I'd much rather have duplication than the wrong abstraction, and having a model class which accepts `request` objects is definitely the wrong abstraction :).

seanh: And a logic module is out of the question? I guess you could argue that just the one thing isn't enough to make a logic module worth it. But otherwise, _that_ seems like the right abstraction to me.

(Have to admit I'm not seeing what's definitely wrong about adding one param and one line to the model class, that seems equally as wrong as two lines and two tests of duplication to me.)

nickstenning: When did I give you the impression that a logic module was out of the question? I'm not sure it's *worth* it in this case, but I'd be fine with it.

seanh: I just got myself caught in an unfortunate cycle, when trying to anticipate what would pass review. I can see pros and cons of all three solutions (including clear cons of putting it on the model as I decided to) but to me they seem equally weighed. Initially I had it duplicated, but thought that wouldn't get through review (turns out it would have). Next I thought of a logic module, but guessed that adding a logic module just for one function wouldn't get passed review (turns out, it would have). So I ended up adding it to the model class, thinking that had the best chance of making it through the review (turns out, it didn't!).

I'll put it back to duplication...

seanh:
Ok, this is done again

nickstenning:
Looks fab. Sorry for the misunderstanding earlier!

PR #3017, opened 2016-02-24
Refactor and add tests for annotation editor toolbar commands

robertknight:
This is part of a cleanup of the markdown editor to make fixing the related  UX bugs for this sprint easier.

This commit refactors the markdown editor's toolbar command handling by separating out the logic for applying commands (make selection bold, insert link, convert selected lines to a list etc.) and the logic for actually applying that to the input field in the markdown editor.

I also removed a couple of `$timeout` calls which trigger an unnecessary full $digest cycle. As a temporary measure this will make the UI more responsive when typing in the editor on pages with large numbers of annotations.

### DIFF
--- a/h/static/scripts/directive/markdown.js
+++ b/h/static/scripts/directive/markdown.js
@@ -5,6 +5,7 @@
 var angular = require('angular');
 var katex = require('katex');
 
+var commands = require('../markdown-commands');
 var mediaEmbedder = require('../media-embedder');
 
 var loadMathJax = function() {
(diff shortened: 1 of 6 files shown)


nickstenning:
I'm not sure if this is a bug introduced by this code, but the "quote"/blockquote button doesn't update the selection appropriately to account for the new characters at the start of the line.

nickstenning:
The same is true of the `ul`/`ol` buttons. I guess none of these are really intended to be used with a selection, but it seems like it might be nice to preserve it, regardless.

robertknight:
> The same is true of the ul/ol buttons. I guess none of these are really intended
> to be used with a selection, but it seems like it might be nice to preserve it, regardless.

Agreed. I've rewritten the block formatting command to preserve the selection.

nickstenning:
Great! FWIW, I really like the `parseState`/`formatState` helpers in the tests. Making it easy to write good tests makes me happy.

nickstenning:
LGTM.

robertknight:
>  Making it easy to write good tests makes me happy.

This is one of the reasons I find functional approaches to front-end architecture so interesting. The logic which can be expressed in a functional way becomes very easy to test.

PR #3018, opened 2016-02-24
Packaging cleanups

nickstenning:
This is two related things rolled into one. First, a commit that does the following:

Run the Python test suite in an isolated virtualenv, against the
installed package.

Using tox to run tests is best practice within the Python community, as
it ensures that the tests are run against the code that will be
installed by a distribution package rather than the code in the current
directory. This helps defend against manifest omissions and other common
errors.

This will also make it substantially easier to run the tests against
multiple versions of Python in the future.

One breaking change here is precipitated by the `--pyargs` option to
py.test. This causes py.test to interpret its arguments as Python module
paths rather than filesystem paths and is necessary to ensure isolation
between the tests and the checked-out repository files. In practice this
just means that if you've been used to running

    py.test h/foo/bar/

or similar, you should now run

    tox h.foo.bar

instead.

----

Second part:

This commit moves a couple of helper tasks for developing the frontend
into the gulpfile, and also improves the automatic installation of
dependencies and building of assets.

All that is now required to get a running development server (with built
frontend assets) from a fresh checkout of the repository is:

    make dev

Subsequent runs of `make dev` will ensure the JavaScript and Python
dependencies are up-to-date but will not rebuild the assets unless they
are either explicitly cleaned,

    make clean dev

or the developer runs the gulp `watch` task independently (recommended).

Similarly, from a fresh checkout, it should be possible to run the test
suite simply by running:

    make test

Note that this commit removes `gulp-cli` and `karma-cli` from the
package dependencies -- these packages are not intended to be installed
in a software distribution, although they can both be installed globally
to assist with easier dispatch of gulp/karma tasks if needed:

    npm install -g gulp-cli karma-cli


Makefile
```diff
@@ -11,25 +10,24 @@ PIP_REQUIRE_VIRTUALENV = 1
...
 clean:
 	find . -type f -name "*.py[co]" -delete
 	find . -type d -name "__pycache__" -delete
-	rm -f .coverage
-	rm -f node_modules/.uptodate .eggs/.uptodate
+	rm -f node_modules/.uptodate h.egg-info/.uptodate
 	rm -rf build dist
 
+.PHONY: dev
+dev: build/manifest.json h.egg-info/.uptodate
```
robertknight: The `dev` target depends on the Python deps and the JS assets but not the JS deps (`node_modules/.uptodate`), so changing package.json and running `make dev` doesn't trigger an npm install.

robertknight:
On my system something is causing gulp to continue waiting for about 20 seconds after the test tasks complete.

I'm just having a look with https://github.com/thlorenz/active-handles to see what is keeping the process alive.

robertknight:
Looks like the `ws` node_module is setting up a 30-second timeout in `cleanupWebsocketResources` and that keeps the process open. I haven't yet tested different node 

It doesn't look like this delay is happening on the CI system so it might be related to the version of Node (v5.5) I'm using or another dependency. I'll keep digging into it but I don't think it should block the merge.

Makefile
nickstenning: Oops! An oversight on my part. Fixed. (`build/manifest.json` should have had a dependency on `node_modules/.uptodate` anyway, otherwise this wouldn't have worked on a clean checkout.

robertknight:
LGTM. @seanh - perhaps you want to take a look as you're more familiar with Python's test tooling ecosystem than I am.

seanh:
We could also [use tox to create the development virtualenv](https://tox.readthedocs.org/en/latest/example/devenv.html)

seanh:
If I understand this correctly (and I just met tox for the first time, so...) then `tox` will get tox to run the tox test environment defined by the `[testenv]` section in `tox.ini`, so it'll run `coverage run -m pytest ...`.

We then define additional test environments for cleaning and combining and reporting the coverage data, so `tox -e clean` will run `coverage erase` and `tox -e coverage` will run `coverage combine` and `coverage report`, neither of these tox test environments actually run the tests, they just work with the covera

seanh:
If I understand this correctly (and I just met tox for the first time, so...) then `tox` will get tox to run the tox test environment defined by the `[testenv]` section in `tox.ini`, so it'll run `coverage run -m pytest ...`.

We then define additional test environments for cleaning and combining and reporting the coverage data, so `tox -e clean` will run `coverage erase` and `tox -e coverage` will run `coverage combine` and `coverage report`, neither of these tox test environments actually run the tests, they just work with the coverage data output by `tox` (running the default tox test env).

You actually end up with three virtualenvs in the `.tox` dir: py27, clean and coverage.

This is just a slightly surprising use of tox test environments to me. From a first read of the tox docs, I understood that the reason you would use multiple test environments was if you wanted to run your tests with different versions of Python and/or with different versions of some dependencies. So each test environment is for running your tests in, against different environment contents.

We seem to be using the clean and coverag

seanh:
I guess my question is, couldn't you just run `coverage erase` if you wanted to do that, instead of `tox -e clean`? And similarly for running `coverage combine report`. Why do this with tox envs and is that not an unusual thing to use a tox env for?

seanh:
Aside from that question this looks cool to me, I'm happy to see tox and interested in using it for our Python dev envs as well

seanh:
I guess one answer to my question is that you'd have to do `pip install coverage` on Travis or in your dev env, whereas `tox -e clean` creates a new virtualenv and pip installs coverage into it for you. Is that the only motivation for this?

nickstenning:
You've basically got it. It's not at all an uncommon way of using Tox, and in future I'd envision adding tox tasks for running flake8, checking the docs build, etc. I suppose it doesn't make much difference if it's just `pip install coverage`, but when it gets to be `pip install coverage flake8 sphinx sphinx-template-foo ...` it's easier to just encode all that in tox and not worry about it.

Plus, you can now just do `tox -e clean,py27,coverage` in one shot.

seanh:
> It's not at all an uncommon way of using Tox

Ok, then this looks good to me, tox looks really useful

seanh:
>  tox tasks for running flake8

Surely prospector?

PR #3020, opened 2016-02-25
Move NIPSA to app

nickstenning:
The only integration points between the API code and NIPSA were:

1. Ensuring that newly-created annotations have {"nipsa": True} if the
   annotation's owner is marked NIPSA'd in `h.api.transform`.

2. Modifying searches made by `h.api.search` so that NIPSA'd annotations
   are excluded when the search is being made by a user other than the
   annotation owner.

Both of these integration points have in this commit been turned into more general ones:

1. The API code now triggers a `h.api.events.AnnotationBeforeSaveEvent`
   which can be used by subscribers that wish to modify the annotation
   dictionary before it is saved.

2. The API search module now adds two configuration directives,
   `add_search_filter` and `add_search_matcher`, which can be used to
   register additional filters or matchers to be used when making
   requests against Elasticsearch.

This change makes it possible for reusers of `h.api` to use a simpler annotations API by default, and implement their own NIPSA-like functionality if so desired.


seanh:
Looks good, I'll give it a quick test just be sure...

PR #3022, opened 2016-02-25
Use Node v4.3 on Travis

robertknight:
Use Node v4 for consistency with what we use to
build the production assets in the Docker build
and also to enable some ES2015 goodness in Gulp scripts (eg. template strings)

Travis does not support multiple languages
in the 'language' field, so non-default Node
versions need to be installed with nvm as
part of the install process.

Rebuilding node-sass was needed to get a version
of the native binding that matched the Node version,
even when after clearing the cache for the branch. I'm not clear
at the moment on why this was necessary.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,8 +6,7 @@ addons:
       # Upload all built zipfiles (browser extensions)
       - $(ls build/*.zip | tr "\n" ":")
   postgresql: "9.4"
-language:
-  - python
+language: python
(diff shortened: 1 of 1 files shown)


nickstenning:
Merged as 7ae8c78. The rebuild wasn't necessary after clearing the Travis cache, and I've moved the installation of Node to before the first `npm install` in the file. :+1:

PR #3023, opened 2016-02-25
Make hound run jscs

seanh:


robertknight:
Before merging this, I think it would be a good idea to do a one-off fix across the codebase using `jscs --fix` to correct many of the existing stylistic inconsistencies and avoid cluttering future PRs with fixes to existing code.

nickstenning:
@robertknight these two things (merging this and running an autofixer on our code) can happen in either order. In the meantime, Hound will AIUI only comment on files changed in a PR, so hopefully it won't be too noisy.

PR #3024, opened 2016-02-25
Add Content-Security-Policy-Report-Only header

finitud:
https://trello.com/c/r2MISk7g/267-deploy-csp-for-web-application

### DIFF
--- a/h/config.py
+++ b/h/config.py
@@ -77,6 +77,7 @@ SETTINGS = [
 
     # Content Security Policy configuration
     EnvSetting('csp.report_uri', 'CSP_REPORT_URI'),
+    EnvSetting('csp.report_only', 'CSP_REPORT_ONLY'),
 
     # Debug/development settings
     EnvSetting('debug_query', 'DEBUG_QUERY'),
--- a/h/config.py
+++ b/h/config.py
@@ -75,6 +75,9 @@ SETTINGS = [
     EnvSetting('h.db.should_drop_all', 'MODEL_DROP_ALL', type=asbool),
     EnvSetting('h.websocket_url', 'H_WEBSOCKET_URL'),
 
+    # Content Security Policy configuration
+    EnvSetting('csp.report_uri', 'CSP_REPORT_URI'),
+
     # Debug/development settings
     EnvSetting('debug_query', 'DEBUG_QUERY'),
 ]
--- a/h/test/tweens_test.py
+++ b/h/test/tweens_test.py
@@ -39,3 +39,25 @@ def test_tween_csp_uri():
     response = tween(request)
 
     assert 'report-uri localhost' in response.headers['Content-Security-Policy']
+
+
+def test_tween_csp_header():
+    request = DummyRequest()
+    request.registry.settings = {
+        "csp": {
+            "font-src": ["'self'", "fonts.gstatic.com"],
+            "report-uri": ['localhost'],
+            "script-src": ["'self'"],
+            "style-src": ["'self'", "fonts.googleapis.com"],
+        },
+    }
+
+    tween = tweens.content_security_policy_tween_factory(
+        lambda req: req.response,
+        request.registry)
+
+    response = tween(request)
+    expected = "font-src 'self' fonts.gstatic.com; report-uri localhost; " \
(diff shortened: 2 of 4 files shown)


h/tweens.py
```diff
@@ -72,3 +72,19 @@ def csrf_tween(request):
...
+    policy = registry.settings.get('csp', {})
+    policy = "; ".join([
+        " ".join([k] + [v2 for v2 in v if v2 is not None])
+        for k, v in sorted(policy.items())
+        if [v2 for v2 in v if v2 is not None]
+    ])
+
+    def content_security_policy_tween(request):
+        resp = handler(request)
+        resp.headers["Content-Security-Policy-Report-Only"] = policy.format(request=request)
```
nickstenning: I'd suggest adding a `csp.report_only` setting that can be set with a header -- `CSP_REPORT_ONLY`, say, and then toggling the name of the header (`Content-Security-Policy` vs `Content-Security-Policy-Report-Only`) on the basis of that setting.

nickstenning:
Looks good. Let's start with something like:

```python
            "font-src": ["'self'", "fonts.gstatic.com"],
            "report-uri": [config.registry.settings.get("csp.report_uri")],
            "script-src": ["'self'"],
            "style-src": ["'self'", "fonts.googleapis.com"],
```

and see what we get. @robertknight sound about right?

h/config.py
```diff
@@ -24,6 +24,7 @@ def settings_from_environment():
     _setup_client(settings)
     _setup_statsd(settings)
     _setup_websocket(settings)
+    _setup_csp(settings)
```
nickstenning: The config module has changed substantially in #3035 so this will need rebasing. Looks like the options you've added will map straightforwardly, though.

h/tweens.py
```diff
@@ -72,3 +72,24 @@ def csrf_tween(request):
         return response
 
     return csrf_tween
+
+
+def content_security_policy_tween_factory(handler, registry):
```
nickstenning: Looks like we need some tests for this function?

finitud: Added some.

h/test/tweens_test.py
```diff
@@ -0,0 +1,57 @@
+# -*- coding: utf-8 -*-
+
+from pyramid.testing import DummyRequest as _DummyRequest
+from h import tweens
+
+
+class DummyRegistry():
```
nickstenning: `pyramid.testing.DummyRequest` should already have a registry, so you probably don't need this.

h/test/tweens_test.py
```diff
@@ -0,0 +1,57 @@
...
+        self.settings = {}
+
+
+class DummyRequest(_DummyRequest):
+    def __init__(self, *args, **kwargs):
+        self.registry = DummyRegistry()
+        super(DummyRequest, self).__init__(*args, **kwargs)
+
+
+class DummyHeader():
```
nickstenning: DummyHeader? It looks like this is supposed to be a dummy handler? If so, I'd suggest you just use `handler = lambda request: None` in the tests.

nickstenning: Oh I see, you need the response object. So again, perhaps use a real request/response object:

```python
handler = lambda request: request.response
```

h/test/tweens_test.py
```diff
@@ -0,0 +1,41 @@
...
+def test_tween_csp_uri():
+    request = DummyRequest()
+    request.registry.settings = {'csp': {'report-uri': ['localhost']}}
+    tween = tweens.content_security_policy_tween_factory(
+        lambda req: req.response,
+        request.registry)
+
+    response = tween(request)
+
+    assert 'report-uri localhost' in response.headers['Content-Security-Policy']
```
chdorner: It might be good to have a test for the value of the header as well. Just to mitigate future bugs in the code that generates the value string of this header. As far as I understand the user experience could completely break when CSP is enabled and the header value is not what we expect it to be.
The easiest would be to set some specific CSP policy in the test and compare it to the value string we expect it to generate.

finitud: I've made this one more strict ('==' instead of 'in') and added one more that tests for a more complex header.

nickstenning:
LGTM.

PR #3028, opened 2016-02-26
Implement tools to enable Chrome extension sourcemaps

robertknight:
When the H client is served from a local URL (eg.
chrome-extension://ID/scripts/foo.bundle.js), the original
source URLs and source maps are not accessible
to Sentry.

Enabling source maps for extensions is done in three steps,
the first two of which are implemented in this PR:

 1. Changes to the crash reporting to transform local URLs
    into filenames when reporting exceptions.

    This results in URLs for source files in Sentry
    reports which are independent of where the files
    are served from
    (eg. chrome-extension://ID/scripts/bundle.js -> bundle.js)

 2. A Gulp task that uploads JS bundles and their associated
    sourcemaps to Sentry using the Sentry API.

    The assets are uploaded to Sentry using just the filename as
    the URL, so the source files and source maps are correctly
    matched up with the transformed URLs produced by the changes
    in (1).

 3. Configuring CI builds to run the Gulp task from step (2).

This commit makes use of ES2015 template strings in
upload-to-sentry.js, so the JSHint file has been adjusted
accordingly.

### DIFF
--- a/.jshintrc
+++ b/.jshintrc
@@ -9,6 +9,7 @@
   "strict": "global",
   "undef": true,
   "unused": true,
+  "esversion": 6,
   "globals": {
     "chrome": false,
     "h": false,
--- a/.jshintrc
+++ b/.jshintrc
@@ -10,6 +10,7 @@
   "undef": true,
   "unused": true,
   "esversion": 6,
+  "esnext": true,
   "globals": {
     "chrome": false,
     "h": false,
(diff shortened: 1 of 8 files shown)


.jshintrc
```diff
@@ -9,6 +9,8 @@
   "strict": "global",
   "undef": true,
   "unused": true,
+  "esversion": 6,
```
nickstenning: Is this wise? I know we're using some ES6 features in the tooling, but we don't currently rewrite ES6 in the application source code, do we?

robertknight: We don't yet, no. I'll have a look to see what the options are for enabling this only for the Gulp scripts.

nickstenning: I'm just concerned that while this silences warnings for the toolchain, it also potentially suppresses warnings about the actual client code.

nickstenning: I'm going to merge this for now. Let's just be aware of this problem.

PR #3029, opened 2016-02-26
Add a test and fix a parameter for queries

lenazun:
Added a test for when we can't parse a query. 
Also added the `strict_parsing=True` parameter to make this possible.


nickstenning:
LGTM!

PR #3031, opened 2016-02-29
Only warn once about Sentry not being configured

robertknight:
In development environments where no Sentry DSN is set,
requests to /app.html produce a warning on each request
due request.sentry.public_dsn() instantiating raven.Client
without a DSN.

This resolves the issue by warning only the first time
that request.sentry is requested in a given run of the app
and muting further warnings.

### DIFF
--- a/h/sentry.py
+++ b/h/sentry.py
@@ -8,9 +8,38 @@ log exceptions from within the application with a useful complement of
 diagnostic data.
 """
 
+import logging
+from os import getenv
+
 import raven
 from raven.transport import GeventedHTTPTransport
 from raven.utils.wsgi import get_environ
+
+log = logging.getLogger(__name__)
+
+
+class SentryStub(object):
+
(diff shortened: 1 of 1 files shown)


nickstenning:
Morning! TBH I think this is probably better done just be setting the log level for the raven package to be `WARN` rather than `INFO`. We don't even need the warning once in development...

nickstenning:
And Sod's Law say that this is currently difficult because Sentry configures its own loggers.

Somewhat amusingly, a fix for this is the head commit on the Raven repository: https://github.com/getsentry/raven-python/commit/110344408c38c3d642191f46a97b7cba0d7aefc9.

nickstenning:
Here's an alternative which has a similar effect (shuts up raven in development) #3032.

robertknight:
Thanks. I'll close this for the moment and we can continue the discussion on the other PR.

PR #3032, opened 2016-02-29
Standardise and improve log output for all servers

nickstenning:
This commit tidies up our logging configuration and ensures the following:

- log format is the same for all loggers, for all servers
- the root log level is WARNING in all cases.
- the log level for `gunicorn.error` (which prints gunicorn
  startup/status messages) is INFO in all cases.
- production servers have a Sentry log handler configured which reports
  all messages at WARNING or above.
- development servers log `h` (our code) at INFO.
- until Raven 5.11.0 is released (which will stop Raven from
  automatically configuring its own logger) we override the `raven`
  logger explicitly and set the log level to WARNING in development

### DIFF
--- a/conf/app.ini
+++ b/conf/app.ini
@@ -7,7 +7,6 @@ pipeline:
   raven
   h
 
-
 [app:h]
 use: egg:h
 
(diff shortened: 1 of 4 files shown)


robertknight:
> the log level for gunicorn.error (which prints gunicorn startup/status messages) is INFO in all cases.

Is `gunicorn.error` used for anything else which should be logged as a warning or error?

conf/app.ini
```diff
@@ -40,67 +39,57 @@ redis.sessions.timeout: 604800
...
 [handlers]
 keys = console, sentry
 
-
 [formatters]
 keys = generic
 
-
 [logger_root]
+level = WARNING
```
robertknight: Are `WARN` and `WARNING` aliases or actually different?

nickstenning: They're the same, but [`WARNING` is the canonical name](https://docs.python.org/2/library/logging.html#levels).

PR #3033, opened 2016-02-29
[requires.io] dependency update on master branch

nickstenning:


nickstenning:
I'm going to submit for these separately.

PR #3035, opened 2016-02-29
Clean up application config/settings

nickstenning:
The application bootstrap was a bit of a mess, and understanding where a particular application setting came from and why was difficult.

This commit splits up the responsibilities for the different parts of this process and hopefully makes it much easier to add/remove/change settings and configuration in the future.

The main changes are:

- A new module, `h.settings`, provides helpers for reading configuration from the system environment.

- `h.config` now exposes a single function, `configure`, which returns a `pyramid.config.Configurator` object configured correctly for the application. In future I imagine this module will take responsibility for more of the application bootstrap.

- All inclusion of app modules and so on is moved into `h.app`'s `includeme` function.

Smaller changes include:

- It is now possible to configure all of the external services using *either* docker links *or* straightforward environment variables. This means it's no longer necessary to set all kinds of crazy env vars in dev to emulate Docker.

- Removed support for the deprecated `SESSION_SECRET` config variable.


h/settings.py
```diff
@@ -0,0 +1,79 @@
...
+    """A setting read from Docker link environment variables."""
+
+    def __init__(self, setting, link, pattern):
+        self.setting = setting
+        self.link = link.upper()
+        self.pattern = pattern
+
+        # Determine the settings that need to be present
+        formatter = string.Formatter()
+        self.placeholders = [x for _, x, _, _ in formatter.parse(pattern)
```
robertknight: I think it would be clearer to name this `field_name` or something along those lines rather than `x` for clarity.

h/test/settings_test.py
```diff
@@ -0,0 +1,83 @@
...
+
+    with pytest.raises(settings.SettingError):
+        func({'PORT': 'notanint'})
+
+
+@pytest.mark.parametrize('setting,link,pattern,environ,expected', (
+    # Should return None if any of the required vars aren't set
+    ('database_url', 'db', 'db://{addr}', {}, None),
+    ('database_url', 'db', 'db://{addr}', {'DB_HOST': 'foo'}, None),
+    ('database_url', 'db', 'db://{host}:{port}', {'DB_ADDR': 'foo'}, None),
```
robertknight: Might be worth raising an error if only some of the required vars are set, eg. in case a dev makes a typo when setting one of the settings.

h/settings.py
nickstenning: Fair call. I'll update this.

h/test/settings_test.py
nickstenning: These aren't going to be set manually (or shouldn't be) so I figured I'd leave that out.

PR #3036, opened 2016-02-29
Remove undocumented /api/annotations endpoint

chdorner:
As discussed with @nickstenning, it is not being used, nor is it documented. Plus it is pretty useless, as it only returns the first 20 search results, without any way of specifying a search term. There is a proper search endpoint for that.

### DIFF
--- a/h/api/test/views_test.py
+++ b/h/api/test/views_test.py
@@ -85,22 +85,6 @@ def test_search_returns_search_results(search_lib):
     assert result == search_lib.search.return_value
 
 
-def test_annotations_index_searches(search_lib):
-    request = testing.DummyRequest()
-
-    views.annotations_index(request)
-
-    search_lib.search.assert_called_once_with(request, {"limit": 20})
-
-
-def test_annotations_index_returns_search_results(search_lib):
-    request = testing.DummyRequest()
-
-    result = views.annotations_index(request)
-
-    assert result == search_lib.search.return_value
-
-
 create_fixtures = pytest.mark.usefixtures('AnnotationEvent',
                                           'schemas',
                                           'storage')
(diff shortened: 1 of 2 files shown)


nickstenning:
Pinging @seanh, @wyan, @robertknight just so you see this and know that it's happened. I don't imagine this is going to cause any problems at all, but it's a change to our public API so you should be aware of it at least.

lenazun:
should we tweet about this from @HypothesisEng?

nickstenning:
@lenazun TBH I think it's fine. I can't find any record in the logs of anyone using this.

judell:
I've used it a lot, it's a nice convenience, but I'm sure it won't be missed by the wider world.

nickstenning:
@judell: really!? it's literally identical to just calling `/api/search` with no parameters, no?

judell:
I have a bookmark I use to quickly get to the JSON for an URL behind our Link element, i.e. of the form:

https://hypothes.is/a/AVL-Z2xVH9ZO4OKSkz2U

Formerly it turned that into:

https://hypothes.is/api/annotations/AVL-Z2xVH9ZO4OKSkz2U

Now I've changed it to:

https://hypothes.is/api/search?id=AVL-Z2xVH9ZO4OKSkz2U

As I said I am quite possibly the only person who ever did this. Can't even remember why, it may have been documented somewhere at some point, but no worries, a trivial thing.

nickstenning:
Right, I figured it was something a bit different.

URLs of this form:

https://hypothes.is/api/annotations/AVL-Z2xVH9ZO4OKSkz2U

aren't going anywhere. This PR only affects the singular URL

https://hypothes.is/api/annotations

which is a hangover from Annotator days and pretty much completely pointless.

PR #3037, opened 2016-03-01
Split Travis build up using matrix

nickstenning:
Now that the frontend build no longer depends quite so heavily on the Python application, we can split the build up a bit so it executes in parallel on Travis.


robertknight:
Great stuff!

PR #3038, opened 2016-03-01
Live reload development tool for Hypothesis client

robertknight:
This implements a live reload server and client support which enables the Hypothesis client
to automatically reload stylesheets when they change and reload the app if scripts
or templates are changed.

Live reloading when scripts change currently requires support from the hosting page, which must
respond to a `reloadrequest` message from the sidebar app.

There are existing tools such as LiveReload and BrowserSync which do this but I've opted for something custom at the moment because of the differences between Hypothesis and a normal web page. On the client, it is also quite useful that we can re-use the existing WebSocket wrapper.

There are two parts to this:

 - A server which serves test pages with the Hypothesis client embedded and notifies
   it when front-end assets (styles, scripts) are changed.

   Test pages are served at http://localhost:3000/<any path>
   by default.

 - A small client script which connects to the server and listens for notifications of asset changes.
   It will then reload style tags etc.

----
**TODO**
- [x] Only start the live reload server when running `gulp watch`
- [x] Investigate why the sidebar does not always reload when a script or template change triggers a full page reload

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,3 +1,12 @@
+Unreleased
+==========
+
+Miscellanea
+-----------
+
+- Added a live reload facility for development of the Hypothesis
+  front-end (#3038)
+
 0.16.0 (2016-04-04)
 ===================
 
--- a/docs/hacking/install.rst
+++ b/docs/hacking/install.rst
@@ -227,6 +227,10 @@ will automatically rebuild the assets whenever the source files change.
     npm install -g gulp-cli
     gulp watch
 
+When `gulp watch` is running, you can visit http://localhost:3000
+to see a page with an embedded Hypothesis client which will automatically reload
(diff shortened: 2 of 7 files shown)


robertknight:
There are still optimizations that could be made here to make the reload quicker when JS files and templates change, but I think this is in decent shape and very useful when tweaking styles, text etc. during development.

CHANGES
```diff
@@ -94,6 +94,12 @@ Miscellanea
 - Early support for serving Content-Security-Policy headers from the
   application (#3024).
 
+Miscellanea
```
nickstenning: This is in completely the wrong part of the file now.

robertknight: Good spot. This happened during the rebase.

PR #3039, opened 2016-03-01
Fix WebSocket server initialization

robertknight:
Adapt the WS server initialization logic following the
changes to app configuration in c55615fb6a0


finitud:
:+1:

PR #3040, opened 2016-03-01
Clean up deleted Elasticsearch annotations in Postgres

chdorner:
After each run of the migration script, identify all the annotations
that were previously imported but have been deleted in Elasticsearch and
delete them from Postgres as well.

### DIFF
--- a/scripts/migrate.py
+++ b/scripts/migrate.py
@@ -65,6 +65,8 @@ def main():
         logging.getLogger('sqlalchemy.engine').setLevel(logging.INFO)
 
     migrate_annotations(request.es)
+    print('')
+    delete_nonexisting_annotations(request.es)
 
 
 def migrate_annotations(es_client):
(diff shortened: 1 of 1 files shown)


scripts/migrate.py
```diff
@@ -283,5 +313,22 @@ def _batch_iter(n, iterable):
...
+    annotations = scan(es)
+    for batch in _batch_iter(BATCH_SIZE, annotations):
+        for ann in batch:
+            ids.add(ann['_id'])
+
+    return ids
+
+
+def collect_pg_annotations():
+   ids = set()
```
nickstenning: Seeing as this returns annotations, not IDs, I suspect you might as well call `Session.query(Annotation).all()` above and do away with this function.

chdorner: Does `.all()` use the Postgres streaming functionality automatically behind the scenes?

nickstenning:
Well, that's one way of doing it...!

Loading all 300K annotations into memory does seem like a pretty brutal approach to take here. Perhaps something like this could be a slightly less blunt approach:

```python

in_postgres = set((a.id for a in Session.query(Annotation.id)))

for batch in _batch_iter(BATCH_SIZE, scan(es)):
    in_postgres.difference_update((a['_id'] for a in batch))

Session.query(Annotation).filter(Annotation.id.in_(in_postgres)).remove()
Session.commit()
```

scripts/migrate.py
nickstenning: You're collecting them all in one big set anyway, so I `yield_per` doesn't buy you anything. I suspect you previously had this function returning a set of ids, no?

nickstenning:
LGTM.

PR #3041, opened 2016-03-02
Focus input field after creating a new annotation

robertknight:
This fixes two issues that resulted in the input field not being focused when creating a new annotation:

1. In addition to focusing the input field for the annotation, the sidebar iframe itself needs to be focused.
2. In order to ensure that the new annotation was visible in the sidebar, the client set the window's location to '#<annotation ID>', which relies on the annotation elements having their ID set as the `<id>` attribute. This also has the effect of focusing the whole annotation and losing focus from the input field. The solution is to use Element.scrollIntoView() instead.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,3 +1,12 @@
+Unreleased
+==========
+
+Features
+--------
+
+- The annotation text input is focused automatically when
+  new annotations are created.
+
 0.9.7 (2016-03-01)
 ==================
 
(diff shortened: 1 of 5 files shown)


robertknight:
Testing this locally I've seen that this is sometimes called before the element is created.

seanh:
Looks good, testing it now..

seanh:
Seems to work. This is a very nice bit of polish to the UX

PR #3042, opened 2016-03-02
Remove the MathJax math rendering fallback

robertknight:
The MathJax fallback was inadvertently broken during recent
refactoring of the <markdown> component and has long been untested.

This removes the fallback for the moment and in doing so
provides a clearer definition of what math we support -
the subset supported by KaTeX.

Looking at our public annotations, there is very little use
of complex math currently.

When we find a clear use case for it, we can revisit supporting
additional math beyond what KaTeX supports.

See https://trello.com/c/9KkjsO6v/66-remove-the-mathjax-fallback
for more background.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,6 +1,13 @@
 Unreleased
 ==========
 
+Bug fixes
+---------
+
+- Fixed error when rendering invalid LaTeX markup. As a side effect,
+  the fallback to MathJax rendering for markup that is not supported
+  by KaTeX has been removed for the time being (#3042).
+
 0.10.0 (2016-03-07)
 ===================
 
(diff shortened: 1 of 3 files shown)


CHANGES
```diff
@@ -7,6 +7,13 @@ Features
 - The annotation text input is focused automatically when
   new annotations are created.
 
+Bug fixes
+---------
+
+- Fixed error when rendering invalid LaTeX markup. As a side effect,
+  the fallback to MathJax rendering for markup that is not supported
+  by KaTeX has been removed for the time being.
```
nickstenning: Can you include the PR number in this? One way to do this consistently is to add the CHANGES entry in a commit pushed after you first open the PR.

nickstenning: This will also need a rebase so it gets added to the right place.

nickstenning:
LGTM.

PR #3044, opened 2016-03-02
Adds an issue template

lenazun:
This aims to make our github issues more consistent


ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,30 @@
...
+2. Click on '....'
+3. Scroll down to '....'
+
+### Expected results
+
+
+### Actual results
+
+
+### On which platforms did you notice this:
```
robertknight: I would change the text to "In which browsers did you notice this?". "Platform" suggests operating system to me which is not relevant most of the time. I would also leave out the list of browsers - we're missing some (for example, "Chrome on Android" and "iOS Safari" are both mobile but very different targets) and if we mention specific versions, they are going to get out of date. The current version of  Safari is 9.x for example.

robertknight:
The build failed because of a check that makes sure files are either included in MANIFEST.in for inclusion in source distributions of H or explicitly excluded. I think we probably want to exclude this since it is a GitHub specific artifact, but check with @nickstenning

nickstenning:
This needs to go in [a folder called `.github`](https://github.com/blog/2111-issue-and-pull-request-templates), and when it's there the test will pass, as `check-manifest` ignores dotfiles.

nickstenning:
:+1:

PR #3045, opened 2016-03-03
Fix syntax error in gulpfile.js

seanh:
It looks like backticks ` were used instead of ', resulting in:

    /.../h/gulpfile.js:33
        throw new Error(`Environment variable ${key} is not set`);
                        ^
    SyntaxError: Unexpected token ILLEGAL


seanh:
Turns out this is ES6 template strings which require node 4.0+, the above error appears on node < 4.0

PR #3046, opened 2016-03-03
Fix ubuntu dev install docs

seanh:


seanh:
I also had to do `npm rebuild node-sass` but I don't think you'd have to do that if following these instructions on a clean Ubuntu install

PR #3047, opened 2016-03-03
Fix elastic.Document initializer

chdorner:
Allow any number of arguments and keyword arguments, as the super class
supports an `id` keyword argument.


nickstenning:
LGTM. Will merge when Travis is happy.

PR #3048, opened 2016-03-03
Use new Document API when expanding URIs

chdorner:
This fixes a bug we discovered on staging (https://app.getsentry.com/hypothesis/stage/issues/114915027/) that got introduced with #3011.


chdorner:
Closing this in favour of #3050 which is the superior fix, as even changing the our code that uses the `Document.uris()` method would not fix every possible case. The internal annotator-store code is using it as well.

PR #3049, opened 2016-03-03
Add direct linking feature flag

robertknight:
Add a feature flag for direct linking and generate
https://hpt.is/<ID> URLs for annotations in the feeds
and in the client when the feature flag is enabled.

Note that the second and third commits in this PR fix existing JSHint issues with annotation-test.js that are needed for the changes in the first commit to pass review.

### DIFF
--- a/h/static/scripts/directive/annotation.js
+++ b/h/static/scripts/directive/annotation.js
@@ -1,8 +1,6 @@
 /* jshint node: true */
 'use strict';
 
-var Promise = require('core-js/library/es6/promise');
-
 var annotationMetadata = require('../annotation-metadata');
 var dateUtil = require('../date-util');
 var documentDomain = require('../filter/document-domain');
--- a/h/static/scripts/directive/annotation.js
+++ b/h/static/scripts/directive/annotation.js
@@ -109,13 +109,22 @@ function updateDomainModel(domainModel, vm, permissions) {
 }
 
 /** Update the view model from the domain model changes. */
-function updateViewModel($scope, time, domainModel, vm, permissions) {
+function updateViewModel($scope, time, domainModel,
+                         vm, permissions) {
 
   vm.form = {
     text: domainModel.text,
     tags: viewModelTagsFromDomainModelTags(domainModel.tags),
   };
-  vm.annotationURI = new URL('/a/' + domainModel.id, vm.serviceUrl).href;
+
+  if (domainModel.links) {
+    vm.annotationURI = domainModel.links.incontext ||
+                       domainModel.links.html ||
+                       '';
+  } else {
+    vm.annotationURI = '';
+  }
+
   vm.isPrivate = permissions.isPrivate(
     domainModel.permissions, domainModel.user);
 
--- a/h/static/scripts/directive/test/annotation-test.js
+++ b/h/static/scripts/directive/test/annotation-test.js
@@ -347,7 +347,7 @@ describe('annotation', function() {
       };
 
       var fakeFeatures = {
-        flagEnabled: sandbox.stub().returns(true)
+        flagEnabled: sandbox.stub().returns(true),
       };
 
       fakeFlash = {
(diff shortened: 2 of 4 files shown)


h/feeds/render.py
```diff
@@ -1,9 +1,20 @@
...
 from pyramid import renderers
 
+from h.features import flag_enabled
 from h.feeds import atom
 from h.feeds import rss
 
 
+def annotation_url(request, annotation):
+    """Return the HTML permalink URL for the given annotation."""
+    if flag_enabled(request, 'direct_linking'):
```
nickstenning: You don't need to import `flag_enabled`. You can just ask `if request.feature('direct_linking'):`

nickstenning:
What does this imply for the sidebar as a potential consumer, in the future, of data from somewhere else? We're now hard-coding URLs and URL structures into the frontend, which we've been slowly moving away from. I'd rather we didn't go back there.

Is it possible to shim the direct link URI into the *model* rather than into the view code? At least until such time as it can be inserted by presenters server-side?

nickstenning:
To clarify what I mean by that, I'd much rather see view code checking properties on the annotation objects to see how to create a link for it rather than hard-coding that behaviour into the renderer.

robertknight:
Summary of Slack discussion: @chdorner is going to have a look at how much work it is to add the information as a property of the Annotation object in the JSON presenter layer. If that proves simple enough then we'll just do that, otherwise we might put a similar transform in the client.

robertknight:
I've now changed this PR to use links provided by the API, assuming the following structure discussed in Slack:

```ts
interface Annotation {
  ...
  links: {
    /// Link to standalone annotation page
    html: string;
    /// Link to view of annotation in context
    incontext: string;
    /// Link to annotation as JSON data
    json: string;
  }
}
```

h/feeds/render.py
robertknight: @nickstenning - I'm guessing we'll want to change how we get the link here completely in light of the discussion around adding a `links` property to annotations.

h/features.py
```diff
@@ -18,7 +18,8 @@
     'claim': "Enable 'claim your username' web views?",
     'new_homepage': "Show the new homepage design?",
     'ops_disable_streamer_uri_equivalence': "[Ops] Disable streamer URI equivalence support?",
-    'postgres_read': 'Use postgres to fetch annotations from storage'
+    'postgres_read': 'Use postgres to fetch annotations from storage',
+    'direct_linking': "Generate direct links to annotations in context in the client?",
```
nickstenning: Let's keep this list in alphabetical order.

h/feeds/render.py
nickstenning: With a view to getting this merged sooner rather than later, I'd suggest simply backing out the changes to the RSS feeds. We can always make them use "direct links" later.

h/static/scripts/directive/annotation.js
```diff
@@ -111,13 +109,21 @@ function updateDomainModel(domainModel, vm, permissions) {
...
+function updateViewModel($scope, time, domainModel,
+                         vm, permissions, features) {
 
   vm.form = {
     text: domainModel.text,
     tags: viewModelTagsFromDomainModelTags(domainModel.tags),
   };
-  vm.annotationURI = new URL('/a/' + domainModel.id, vm.serviceUrl).href;
+
+  if (features.flagEnabled('direct_linking') &&
```
nickstenning: You don't even need to check the feature flag here. We can simply include or omit the `incontext` link server-side when we toggle the flag, so would suggest making `domainModel.links && domainModel.links.incontext` your test.

robertknight:
@nickstenning - I have updated this to use the incontext (preferred) or HTML links from the Annotation object and removed the changes to the feeds package for the moment. This is ready for another look next week.

h/static/scripts/directive/annotation.js
```diff
@@ -111,13 +109,20 @@ function updateDomainModel(domainModel, vm, permissions) {
...
+                         vm, permissions) {
 
   vm.form = {
     text: domainModel.text,
     tags: viewModelTagsFromDomainModelTags(domainModel.tags),
   };
-  vm.annotationURI = new URL('/a/' + domainModel.id, vm.serviceUrl).href;
+
+  if (domainModel.links) {
+    vm.annotationURI = domainModel.links.incontext || domainModel.links.html;
```
nickstenning: If the default is to be empty-string, then this should probably have a `|| ''` at the end, too.

nickstenning:
I'm going to merge this as it's only a small change, but please do try and take the time to make sure commit messages are updated when rebasing branches. The commit message says "Add direct linking feature flag" when it doesn't do anything of the sort.

robertknight:
@nickstenning - Thanks for the reminder. That was a stupid oversight on my part.

PR #3050, opened 2016-03-03
Rename Document.uris relationship to `document_uris`

chdorner:
To not clash with the annotator-store Document.uris method. This can and
should be reverted as soon as annotator-store is no longer a dependency
anymore.

See: https://app.getsentry.com/hypothesis/stage/issues/114915027/


nickstenning:
LGTM.

PR #3051, opened 2016-03-03
Normalize elastic.DocumentMeta og types

chdorner:
The annotator switched the naming of Facebook meta tags from "og" to
"facebook" mid 2013. This will normalise them and always return the
current "facebook" type.

The above-mentioned commit in the client can be found [here](https://github.com/openannotation/annotator/commit/61c385da0e3ba1598ced82175c0689768b84d688).


nickstenning:
Naïsh.

PR #3053, opened 2016-03-04
Convert {app, app-controller, widget-controller}.coffee to JS

robertknight:
In preparation for upcoming changes to these files as part of the direct linking feature, convert them from CoffeeScript to JS.

For reference, the process I'm using is:
 1. Run the latest version of [decaffeinate](https://github.com/decaffeinate/decaffeinate) on the .coffee file and associated tests, this may require fixing up indentation ambiguities in the file first.
 2. Remove the .coffee file so that `require()` loads the JS version
 3. Convert any ES6-isms to ES5 - there are usually only a small handful so this doesn't particularly
     harm readability at the moment IMO.
 4. Run tests, check that relevant part of the app still works
 5. Check the JS file for any JSHint violations
 6. Run `tsfmt -r` on the JS files to do an initial set of formatting corrections
 7. Run `jscs -x` on the JS files to fix styling issues
 8. Do a pass over the code and fix any outstanding issues. Common issues include:
  - Using `@ngInject` instead of `[dep1, dep2, function (...) { ... }]` for specifying Angular dependencies
  - Removing unnecessary `return` statements from the last statement of a function. CoffeeScript
    implicitly returns the value of the last statement in a function and because it is untyped (sigh), the
    converter doesn't know whether the function actually intends to return a value or not,
    so it has to assume that it does.


nickstenning:
Mostly looks good but the indentation seems to be all over the shop. For example,

```javascript
$scope.scrollTo = function (annotation) {
    if (angular.isObject(annotation)) {
      return crossframe.call('scrollToAnnotation', annotation.$$tag);
    }
  };
```

which, if consistent with the rest of our codebase, would be: 

```javascript
$scope.scrollTo = function (annotation) {
  if (angular.isObject(annotation)) {
    return crossframe.call('scrollToAnnotation', annotation.$$tag);
  }
};
```

That seems to be a problem throughout.

robertknight:
Indeed. I'll get that fixed.

robertknight:
> Mostly looks good but the indentation seems to be all over the shop. For example,

I've fixed that now by running the JS from decaffeinate through two post-processing steps:

[tsfmt](https://github.com/vvakame/typescript-formatter) does an initial pass over the code, correcting indentation but with some differences from our coding style. `jscs -x` can then be used to tidy up the remainder.

nickstenning:
LGTM.

PR #3055, opened 2016-03-04
Travis Sentry sourcemaps

nickstenning:
This PR changes the `buildext` command so that it doesn't shell out to gulp itself. Instead, it assumes that prebuilt assets are available in build, and prints an error suggesting the user run `gulp build` if it can't find a file it expects to exist. I've verified that the resulting extension is byte-for-byte identical before and after this change.

This makes it possible to run `gulp build` just once, and then build multiple extensions from the built assets, and subsequent commits set that up for Travis tag builds.

### DIFF
--- a/.travis/install
+++ b/.travis/install
@@ -9,5 +9,7 @@ case "$ACTION" in
     extensions)
         pip install --use-wheel -e .
         npm install
+        # trick the Makefile into skipping npm install
+        touch node_modules/.uptodate
         ;;
 esac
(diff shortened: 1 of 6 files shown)


robertknight:
This looks good. One issue I found in the existing code whilst testing this is that the process would exit with zero status if the sourcemap upload failed because the exception was never passed to the callback in `through.obj`. I've fixed this so that the task will exit with a non-zero status and show a backtrace if the upload fails.

PR #3057, opened 2016-03-04
Sheetaluk/280 when a user replies focus on and expand that reply

sheetaluk:
calling toggleCollapsed against the parent of the thread to show the replies container after a user publishes a reply to an annotation.

### DIFF
--- a/h/static/scripts/directive/annotation.js
+++ b/h/static/scripts/directive/annotation.js
@@ -734,6 +734,12 @@ function link(scope, elem, attrs, controllers) {
         if (threadFilter) {
           threadFilter.freeze(false);
         }
+
+        // Ensure that the just-edited annotation remains visible.
+        if (thread.parent) {
+          thread.parent.toggleCollapsed(false);
+        }
+
       }
     });
 
--- a/h/static/scripts/directive/test/annotation-test.js
+++ b/h/static/scripts/directive/test/annotation-test.js
@@ -153,7 +153,10 @@ describe('annotation', function() {
       };
       mockThreadController = {
         collapsed: true,
-        toggleCollapsed: sinon.stub()
+        toggleCollapsed: sinon.stub(),
+        parent: {
+          toggleCollapsed: sinon.stub()
+        }
       };
       mockThreadFilterController = {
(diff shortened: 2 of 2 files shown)


h/static/scripts/directive/annotation.js
```diff
@@ -734,6 +734,10 @@ function link(scope, elem, attrs, controllers) {
         if (threadFilter) {
           threadFilter.freeze(false);
         }
+        if (thread.parent) {
+          thread.parent.toggleCollapsed(false);
```
robertknight: This could use a brief one-line comment here to explain what is going on, eg. `// Ensure that the just-edited annotation remains visible`.

robertknight:
First PR :+1: 

There are a couple of things to do in order to finish this off:

1. Use `git rebase` to squash the second commit with a trivial change into the first. When working on a branch it is common that you might have commits which are separate changes and then other commits which fix up minor syntactic issues, edge cases etc. Those secondary commits should generally be squashed.
2. Write a test case in `annotation-test.js`. See https://github.com/hypothesis/h/blob/master/h/static/scripts/directive/test/annotation-test.js#L133 for tests for related code. I started doing some exploration into refactoring the way threading works about a week ago, so the tests are useful to document that behavior and avoid breakage when that happens.

h/static/scripts/directive/test/annotation-test.js
```diff
@@ -263,6 +266,20 @@ describe('annotation', function() {
...
+      mockAnnotationController.editing.returns(true);
+      link(scope, mockElement, mockAttributes, mockControllers);
+      scope.$digest();
+
+      mockAnnotationController.editing.returns(false);
+      scope.$digest();
+
+      assert.equal(
+        true,
+        mockThreadController.parent.toggleCollapsed.lastCall.calledWithExactly(false));
```
robertknight: You can simplify this slightly to `assert.calledWith(mockThreadController.parent.toggleCollapsed, false)`

robertknight:
LGTM. Just the small simplification I suggested above and then this is good to be merged.

robertknight:
First PR :cake:

PR #3059, opened 2016-03-07
Fix the permissions key when rendering annotations

### LINKED ISSUE #3058: Edit/delete links active for other users' annotations
### Steps to reproduce

1. Focus on another users' annotation card

2. Edit/delete links are active

### Expected behaviour

They should not appear.

chdorner:
It used to be `permissions` (plural), the new API presenter rendered it as `permission` singular.

Fixes #3058, which showed the edit/delete button for other users' annotations, but that was a rendering bug only, it wouldn't allow anybody to delete other users' annotations.


nickstenning:
LGTM. Thanks for the quick turnaround.

PR #3063, opened 2016-03-08
Leave out default target selector from presented annotation

### LINKED ISSUE #3060: Page notes do not display in sidebar
### Steps to reproduce

Create a page note

It displays in the stream

But not in the sidebar

### Other info

A standalone client-side script also broke today. For empty text fields it was expecting `"text":""` but today instead began began receiving `"text":null`. Adding a null check fixed it.

Might that be the cause here? Might it also be related to https://github.com/hypothesis/h/issues/3062?

chdorner:
An empty list of target selectors will result in the client trying to
anchor the annotation, which will fail, so these orphaned annotations
will not be visible.
Page notes are a good example of annotations that don't have selectors
by design, but should not be filtered out.

Fixes #3060

### DIFF
--- a/h/api/presenters.py
+++ b/h/api/presenters.py
@@ -65,8 +65,11 @@ class AnnotationJSONPresenter(object):
 
     @property
     def target(self):
-        return [{'source': self.annotation.target_uri,
-                 'selector': self.annotation.target_selectors or []}]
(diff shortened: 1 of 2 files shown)


robertknight:
LGTM.

PR #3065, opened 2016-03-08
Fix installation of URL polyfill with js-polyfills v0.1.16

### LINKED ISSUE #63: Reg: Unify databases [f2]
We have 8000+ (?) usernames already reserved.  We'll have people that want to use them.  (JPbarlow comes to mind).  Thus, when someone is trying to login, we should check to see if that username is already reserved, check that the email addr they supply matches, and then record their (double entered) password.

### LINKED ISSUE #3064: Hypothesis v0.10.0 broken under IE
### Steps to reproduce
1. Visit a page with H embedded or use the bookmarklet with IE 11.0

### Expected behaviour
The sidebar should load

### Actual behaviour
Many `Object does not support this action` errors in the console.

Windows 10, IE 11.0

### Additional details
Sentry reports: https://app.getsentry.com/hypothesis/prod/issues/115604816/

robertknight:
js-polyfills v0.1.16 was recently published with a change in how the polyfill is installed.
This version of the lib was picked up by the recent v0.10.0 release.

js-polyfills v0.1.15 used to install the polyfill on
`this.URL`, where `this` was set to module.exports in
the context of a Browserify bundle. Consequently the result
was available as `require('js-polyfills/url').URL`.

As of v0.1.16 however, it installs the polyfill on `self`,
which is the Window object in the context of the browser
and so `require('js-polyfills/url').URL` is undefined.

Fixes #3064

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,6 +1,12 @@
 Unreleased
 ==========
 
+Bug Fixes
+---------
+
+- Fix compatibility with IE 10, 11.0 and early versions
+  of Microsoft Edge (#3064)
+
 0.10.0 (2016-03-07)
 ===================
 
--- a/h/static/scripts/polyfills.js
+++ b/h/static/scripts/polyfills.js
@@ -9,7 +9,7 @@ require('core-js/fn/object/assign');
 try {
   new window.URL('https://hypothes.is');
 } catch (err) {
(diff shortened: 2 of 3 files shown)


nickstenning:
:+1:

PR #3066, opened 2016-03-08
Fix confirmation of user deletion in admin panel

nickstenning:
83c4f48 broke this by removing the "scripts" block from the admin layout.


chdorner:
This actually doesn't work now because of the CSP policy refusing to run inline Javascript.

nickstenning:
I'd suggest that we merge this and #3068 and we worry about the inline JS later when doing an audit of CSP reports.

PR #3067, opened 2016-03-08
Configure gunicorn statsd

chdorner:
We were configuring the gunicorn statsd host/port the wrong way.

This fixes that, plus makes sure that we don't error when an environment variable setting is of the wrong type, but we emit a warning and then ignore its value.

### DIFF
--- a/gunicorn.conf.py
+++ b/gunicorn.conf.py
@@ -1,14 +1,21 @@
 # -*- coding: utf-8 -*-
 import os
-from h._compat import urlparse
 
 # Smart detect heroku stack and assume a trusted proxy.
 # This is a convenience that should hopefully not be too surprising.
 if 'heroku' in os.environ.get('LD_LIBRARY_PATH', ''):
     forwarded_allow_ips = '*'
 
-if 'STATSD_PORT' in os.environ:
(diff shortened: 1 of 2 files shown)


gunicorn.conf.py
```diff
@@ -1,14 +1,22 @@
...
+if not os.environ.get('GUNICORN_STATS_DISABLE', None):
+    if 'STATSD_PORT_8125_UDP_ADDR' in os.environ and \
+       'STATSD_PORT_8125_UDP_PORT' in os.environ:
+            _host = os.environ['STATSD_PORT_8125_UDP_ADDR']
+            _port = os.environ['STATSD_PORT_8125_UDP_PORT']
+            statsd_host = '{}:{}'.format(_host, _port)
+
+    elif 'STATSD_HOST' in os.environ:
+        _host = os.environ['STATSD_HOST']
+        _port = os.environ.get('STATSD_PORT', '8215')
```
nickstenning: s/8215/8125/

chdorner: Doh! Fixed.

nickstenning:
LGTM.

PR #3069, opened 2016-03-08
Allow sentry logging to work correctly from the websocket server

finitud:


nickstenning:
This looks good, and will stop the websocket server crashing, but you also need to update the `get_reader` call in `process_nsq_topic` to pass in a Sentry client of its own construction!

h/queue.py
```diff
@@ -92,6 +90,16 @@ def resolve_topic(topic, namespace=None, settings=None):
     return topic
 
 
+def _get_queue_reader(request, topic, channel):
+    get_reader(request.registry.settings,
```
nickstenning: Oops! You're missing a `return` here.

finitud: Whoops! Keep forgetting Python functions don't just return the last value!

nickstenning:
LGTM.

PR #3070, opened 2016-03-08
Render annotation text as empty string when missing

chdorner:
As mentioned on #3060 in addition to the fix that made the page notes appear again (#3063)


nickstenning:
LGTM.

PR #3072, opened 2016-03-08
Use the presenter's text property

chdorner:
Which was missing from #3070


nickstenning:
Sorry we missed you, 11 deleted bytes.

PR #3074, opened 2016-03-09
Convert 'annotation-ui' from CoffeeScript to JS

robertknight:
Convert the 'annotation-ui' module which manages the state of selected and focused annotations from CS to JS.

Given that the resulting module is a very simple service with no dependencies, I've also de-Angularized the unit test.


nickstenning:
This is just a straight-up translation?

robertknight:
Yes, plus a minor simplification in the test setup. In this and future PRs I'll keep anything that's not a translation in separate commits.

nickstenning:
LGTM.

PR #3075, opened 2016-03-09
Add a utility script to assist with CoffeeScript -> JS conversion

robertknight:
Add a utility script to assist with CoffeeScript -> JS conversion

It accepts a list of CoffeeScript file paths on stdin and for each <path>.coffee file in the input, it writes
a <path>.js output file containing the converted, reformatted source.

The result is ES2015 source so will require some fixups as long as we're using ES5 in the main app
codebase.

The current transformation process involves:

 1. Performing an initial CoffeeScript -> JS
    conversion with 'decaffeinate'
 2. Applying some fixups for common issues, currently done with dumb string modifications. In future I will probably look at doing this properly using a Babel or TypeScript transformer. [js-codemod](https://github.com/cpojer/js-codemod) is another tool that could be used.
 3. Parsing the source with Babylon to check that it is valid ES2015
    If not, the user is prompted to go and simplify the CoffeeScript
    so that decaffeinate can process it correctly.
 4. Reformatting the source with typescript-formatter, which
    takes care of indentation and high-level formatting.
 5. Running the source through JSCS' fix mode which deals
    with smaller details.


nickstenning:
:coffee: :computer: :+1:

PR #3076, opened 2016-03-09
Write annotations to postgres

seanh:
It's probably good to look at the individual commits for this one, since there's a few commits (e.g. grouping entire test modules into test classes) that create a lot of diff noise.

Ok, so what this pull request does is:

* Create annotations and create and update annotation metadata (`Document`, `DocumentURI` and `DocumentMeta` objects) in Postgres when the create annotation API is called if the postgres_write feature flag is on

* Adds some extra validation to the create annotation API (only when postgres_write is on) so that it's not easy to make it crash

What it does not do:

* Does not index annotations from Postgres into Es
* Does not do anything in Postgres when the annotation update or delete APIs are called, these still ignore the postgres_write feature flag and write to Es

In general I've tried not to change the legacy behaviour when postgres_write is off. So I've moved the existing code unmodified into `legacy_*()` functions that get called when postgres_write is off, and put the new code in place of the original functions. This involves some duplication but the legacy stuff is to be deleted.

### Parsing the annotation data from the client

Now that we have a real database we can't just throw any crap into it, either the db or our own code will crash if required fields are missing or if known fields are the wrong type. Minimal additional JSON Schema validation has been added (only to the create annotation API and only when postgres_write is on) to prevent these crashes and send proper validation errors back.

The format of annotation data that the client POSTs in annotation create and update requests is a little different from what we need to create `h.api.models.Annotation` objects in the database. Code for converting this dict from the client into a more correct/convenient format is now all in `h.api.schemas` so that when it comes out of schema validation it's in the right format. Some of this code is new, some of it has been moved from `storage` to `schemas`.

`schemas` only does "structural" validation and transformations, i.e. work that doesn't require looking in the db. Some further "semantic" validation and transformation that does require reading from the db is done in `storage`.

### Parsing the document metadata from the client

The format of the "document" sub-object that the client POSTs in annotation create and update requests is quite far from the format that we need in order to create/update the `Document`, `DocumentMeta` and `DocumentURI` objects in the database. The new `h.api.parse_document_claims` module handles transforming this dict into two lists that are in a more convenient format for `storage` to later use to create the db objects. The code in here is adapted from existing code in `migrate.py` and `elastic.py`.

`schemas` calls `parse_document_claims` to do this, so when the data comes out of schema validation it's in the right/convenient format.

### Storage / logic of creating and updating annotation data and document metadata

Storing an annotation is considerably more complicated than it is with the legacy Elasticsearch storage code in h, because of the need to create and update document equivalence data and document metadata.

`h.api.storage.create_annotation()` is really a "logic" function with some semantic validation/transformation mixed in, all the structural validation and transformation and the model handling is elsewhere. The logic of `h.api.storage.create_annotation()` is:

* Force replies to have the same group as their parent
* Fail if replying to an annotation ID that doesn't exist
* Fail if writing to a group that you don't have permission to (note this has to happen after setting the group of replies)
* Create an `Annotation` model object from the (already validated) data
* Find all existing `Document` objects referred to by any existing `DocumentURI` objects that match any of the document URI equivalence claims in the post
* If there are no such `Document`s create a new `Document` with no `DocumentURI`s or `DocumentMeta`s pointing to it (yet)
* If there is one such `Document` use it
* If there are multiple such `Documents` merge them into one (all of the `DocumentURI`s will now point to one `Document` and the other `Document`s will be deleted, this destroys information)
* Update the `.updated` time of the `Document`
* Create new `DocumentURI`s, or update existing ones, for all document URI equivalence claims in the request
* Create new `DocumentMeta`s, or update existing ones, for all document metadata claims in the request

### Creating and updating document metadata in Postgres

The logic "create an object in the db with these params, or update an equivalent one if one already exists" seems like it belongs in the model to me, so `create_or_update_document_meta()` and `create_or_update_document_uri()` are added to `models` to do this. Again these are adapted from existing code in `migrate.py`. `storage` calls them. `merge_documents()` which does similar work was already in `models`, is now called by `storage` as well.

### DIFF
--- a/h/api/storage.py
+++ b/h/api/storage.py
@@ -12,9 +12,6 @@ from functools import partial
 from h.api import transform
 from h.api import models
 from h.api.events import AnnotationBeforeSaveEvent
-from h.api.models import elastic
-from h.api.models.annotation import Annotation
-from h.api.models.document import Document
 
 
 def annotation_from_dict(data):
@@ -27,7 +24,7 @@ def annotation_from_dict(data):
     :returns: the created annotation
     :rtype: dict
     """
-    return elastic.Annotation(data)
+    return models.elastic.Annotation(data)
 
 
 def fetch_annotation(request, id):
@@ -44,13 +41,13 @@ def fetch_annotation(request, id):
     :rtype: dict, NoneType
     """
     if _postgres_enabled(request):
-        return request.db.query(Annotation).get(id)
+        return request.db.query(models.Annotation).get(id)
 
-    return elastic.Annotation.fetch(id)
+    return models.elastic.Annotation.fetch(id)
 
 
 def _legacy_create_annotation_in_elasticsearch(request, data):
-    annotation = elastic.Annotation(data)
+    annotation = models.elastic.Annotation(data)
     # FIXME: this should happen when indexing, not storing.
     _prepare(request, annotation)
     annotation.save()
@@ -134,7 +131,7 @@ def update_annotation(request, id, data):
     :returns: the updated annotation
     :rtype: dict
     """
-    annotation = elastic.Annotation.fetch(id)
+    annotation = models.elastic.Annotation.fetch(id)
     annotation.update(data)
 
     # FIXME: this should happen when indexing, not storing.
@@ -154,7 +151,7 @@ def delete_annotation(request, id):
     :param id: the annotation id
     :type id: str
     """
-    annotation = elastic.Annotation.fetch(id)
+    annotation = models.elastic.Annotation.fetch(id)
     annotation.delete()
 
 
@@ -177,9 +174,9 @@ def expand_uri(request, uri):
     """
     doc = None
     if _postgres_enabled(request):
-        doc = Document.find_by_uris(request.db, [uri]).one_or_none()
+        doc = models.Document.find_by_uris(request.db, [uri]).one_or_none()
     else:
-        doc = elastic.Document.get_by_uri(uri)
+        doc = models.elastic.Document.get_by_uri(uri)
 
     if doc is None:
         return [uri]
--- a/h/api/storage.py
+++ b/h/api/storage.py
@@ -10,6 +10,7 @@ assumed to be validated.
 from functools import partial
 
 from h.api import transform
+from h.api import models
 from h.api.events import AnnotationBeforeSaveEvent
 from h.api.models import elastic
 from h.api.models.annotation import Annotation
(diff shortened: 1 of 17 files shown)


nickstenning:

h/api/storage.py
```diff
@@ -44,6 +46,154 @@ def fetch_annotation(request, id):
...
 
+def _legacy_create_annotation_in_elasticsearch(request, data):
+    annotation = elastic.Annotation(data)
+    # FIXME: this should happen when indexing, not storing.
+    _prepare(request, annotation)
+    annotation.save()
+    return annotation
+
+
+def transform_links(document, claimant, created, updated):
```
chdorner: I added the `created` and `updated` since when we migrate the data over to Postgres we don't want to have the current timestamp as the value for document objects. I'm not sure if it makes sense to do this when we write annotations from Postgres directly from the views.

h/api/storage.py
```diff
@@ -44,6 +46,154 @@ def fetch_annotation(request, id):
...
+        # disregard doi links as these are being added separately from the
+        # highwire and dc metadata later on.
+        if link.keys() == ['href'] and link['href'].startswith('doi:'):
+            continue
+
+        uri_ = link['href']
+        type = None
+
+        # highwire pdf (href, type=application/pdf)
+        if set(link.keys()) == set(['href', 'type']) and len(link.keys()) == 2:
```
chdorner: I know it's not your fault, but as you pointed out this statement is wrong.

To add some more context to what I wrote in Slack:

We should certainly check that the value of `type` is `application/pdf`. But after thinking more about it, we should also discard that link and generate it server-side from the highwire meta data (`document[highwire][pdf_url]`) to keep it consistent with how we handle the doi links.
This also has the added benefit of us getting way more `highwire-pdf` document uris, since we have more documents with this document meta data vs uri data.

Thoughs? (plus @nickstenning)

h/api/storage.py
seanh: The one argument I can think of for passing in `created` and `updated` (rather than letting the db auto generate them) is that each `DocumentURI` will have the same `created` value as the annotation that created it, and the same `updated` value as the annotation that last updated. Also the `DocumentURI`'s `created` and `updated` will be exactly the same as each other, for ones that have never been updated. Rather than all of these times being milliseconds different to each other. Whether that's actually important at all I'm not sure but it does mean that comparisons like `created == updated` will work, if we ever need them.

I noticed that with the `Timestamps` mixin we're using, a new annotation gets `created` and `updated` times that are actually slightly different to each other.

seanh:
- [ ] Add minimal extra validation to schemas
- [ ] Move data transformation out of storage into schemas
- [ ] Only do new schema stuff if feature flag is on
- [ ] Use `normalize()` helper everywhere in `h.api.models.document`
- [ ] Move all `h.api.models.test.document_test` tests into classes
- [ ] Test manually with various different document metadata
- [ ] See TODOs and FIXMEs in diff

h/api/storage.py
seanh: Fixed this

seanh:
`storage.create_annotation()` could probably do with being broken up into multiple functions

seanh:
This ended up being really big. I'm away until Tuesday but happy to break this up into ~4 separate bits when I get back

seanh:
Closing this because I've broken it up into four separate pull requests: <https://github.com/hypothesis/h/pull/3150>, <https://github.com/hypothesis/h/pull/3151>, <https://github.com/hypothesis/h/pull/3152>, <https://github.com/hypothesis/h/pull/3153>

PR #3078, opened 2016-03-09
changing adder icons and note icon in toolbar.

sheetaluk:
Toolbar:
1. Changed comment icon to note icon.
2. When text selection is successful, change note icon to annotation icon and title on hover to "New Annotation".
3. When text selection has failed, change annotation icon to note icon and title on hover to "New Page Note".

Adder:
1. Changed comment icon to annotate icon and old highlight icon to the new highlight icon.
2. Added helper labels under each icon.
2. Icons are gray-dark and the labels below are gray-light when adder pops up. When user hovers on one, it stays gray-dark, the label below turn gray-dark. The other icon turns gray-light and it's label stays gray-light.


nickstenning:
Great work so far!

A couple of issues I think we need to resolve before this can merge.

- [ ] For me it looks like the new adder is missing some styles. Not sure if I failed to build something appropriately but this persists even after `make clean && gulp build`:

    screen shot 2016-03-10 at 09 16 40

- [ ] The icon of the toolbar button correctly changes, but it still creates page notes when a selection is present. It needs to trigger the creation of an annotation, rather than a page note, in this scenario.

- [ ] The positioning of the now-slightly-larger is a bit off. Specifically, it gets in the way when just idly selecting text and it's rather hard to avoid clicking on it, even if I don't want to. Ideally, I think the point of the little protruding triangle should end up next to the mouse pointer. Alternatively, we could take the Medium approach and have the adder appear in the middle at the top of the selection rather than next to the mouse pointer. @conordelahunty do you have thoughts on this issue?

h/static/styles/vendor/icomoon.css
```diff
@@ -1,175 +1,192 @@
...
...USGgceJCE5TCwMGAwDIjlOMBg1HTdcHAEOCwJNcw4EBAICPlICKjACARYjGA1trNVoBg4HFjcgAAAAAAEAAAABgAC9zfrvXw889QALBAAAAAAA0wXCbgAAAADTBcJuAAD/qwQAA8AAAAAIAAIAAAAAAAAAAQAAA8D/wAAABAAAAP/EBAAAAQAAAAAAAAAAAAAAAAAAADoEAAAAAAAAAAAAAAACAAAABAAA1QQAAJIEAACrBAAAAAQAAGQEAAArBAAAKwQAAVUEAADVBAAA1QQAAFUEAAAABAAA1QQAAIAEAAAABAABKwQAAQAEAABrBAAAVQQAANUEAAEABAAAVQQAAIAEAABVBAABbgQAAAAEAABVBAAAkgQAANUEAAGrBAAAJgQAAAgEAAAIBAAAAwQAAFUEAACABAAA1QQAANUEAACABAAAgAQAAFUEAAErBAAAVQOAAIADxAAABAAAAAQAAAAEAAAABAAAAAQAAMAEAABABAAAAAQAAAAEAAAAAAAAAAAKABQAHgA4AEoAlADaAWQBsAIeAjICTAJmApgC0ALmAwQDLANiA3oDyAQKBCYEQgSIBLAE4ATyBUAFeAWKBaQFsAYKBpYHkAheCLgJBAkmCTQJVgl4Cb4JzAn2CigKogscC14LjAvsDA4MSgzYDQwNdAAAAAEAAAA6AK8ACAAAAAAAAgAAAAAAAAAAAAAAAAAAAAAAAAAOAK4AAQAAAAAAAQABAAAAAQAAAAAAAgAHACoAAQAAAAAAAwABACQAAQAAAAAABAABAD8AAQAAAAAABQALAAMAAQAAAAAABgABACcAAQAAAAAACgAaAEIAAwABBAkAAQACAAEAAwABBAkAAgAOADEAAwABBAkAAwACACUAAwABBAkABAACAEAAAwABBAkABQAWAA4AAwABBAkABgACACgAAwABBAkACgA0AFxoAGhWZXJzaW9uIDEuNQBWAGUAcgBzAGkAbwBuACAAMQAuADVoAGhoAGhSZWd1bGFyAFIAZQBnAHUAbABhAHJoAGhGb250IGdlbmVyYXRlZCBieSBJY29Nb29uLgBGAG8AbgB0ACAAZwBlAG4AZQByAGEAdABlAGQAIABiAHkAIABJAGMAbwBNAG8AbwBuAC4AAAADAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA') format('woff');
```
robertknight: Ah, the `update-icon-font.py` script needs to be updated so that it doesn't inline the `h.woff` file automatically. Can you cherry-pick the version of `update-icon-font.py` from bfd41e1c5ecf518b082876662e376a9ca3ed931e and use that to generate the font.

h/static/scripts/annotator/guest.coffee
```diff
@@ -44,8 +44,14 @@ module.exports = class Guest extends Annotator
   html: extend {}, Annotator::html,
     adder: '''
       <div class="annotator-adder">
-        <button class="h-icon-insert-comment" data-action="comment" title="New Note"></button>
-        <button class="h-icon-border-color" data-action="highlight" title="Highlight"></button>
+        <div class="annotator-adder-action">
+          <button class="h-icon-annotate" data-action="comment"></button>
+          <label class="adder-label">Annotate</label>
```
robertknight: The label and the button elements are not connected here. You can either use the `for` attribute on the `<label>` or you could put the caption inside the `<button>` - have a look at the [MDN docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label) for details.

h/static/styles/annotator/inject.scss
```diff
@@ -34,28 +34,45 @@ $base-font-size: 14px;
...
+      width: 45px;
+      font-size: 18px;
+      margin: 2px 0px;
+      padding: 0;
+      text-align: center;
+      background: $white !important;
+      color: $gray-dark !important;
+    }
+
+    .adder-label {
```
robertknight: Try to keep nesting in CSS to a minimum - as this makes maintenance easier. If you can define a class at the top-level (ie. so you get just `.class` as the selector rather than `.parent .child`) that is preferable.

For newly written styling, we try to use [BEM](https://css-tricks.com/bem-101/) conventions. Legacy CSS in our codebase however isn't written this way and over-uses nesting.

conordelahunty:
Looking good Sheetal. I've added some design feedback, notes, and videos to the Trello card. Have also responded to @nickstenning 's question of positioning of the adder too.

nickstenning:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;w6UIP3xw&#x2F;281-update-the-toolbar-s-note-icon-tooltip-and-update-the-adder-design"> Update the toolbar&#39;s note icon, tooltip, and update the adder design</a>

h/static/styles/annotator/inject.scss
sheetaluk: Also on slack: I forgot to talk to you about this: saw your comment on using BEM stying. The card got a little big, are you open to creating a new card for it? Do we want to take this opportunity to refactor the styling in `static/styles/annotator/inject.scss`? It is only 290 lines long.

h/static/scripts/annotator/guest.coffee
```diff
@@ -44,8 +44,16 @@ module.exports = class Guest extends Annotator
   html: extend {}, Annotator::html,
     adder: '''
       <div class="annotator-adder">
-        <button class="h-icon-insert-comment" data-action="comment" title="New Note"></button>
-        <button class="h-icon-border-color" data-action="highlight" title="Highlight"></button>
+        <div class="adder-actions">
+          <div class="annotator-adder-action action-annotate">
+            <button id="adder-icon-annotate" class="h-icon-annotate" data-action="comment"></button>
+            <label for="adder-icon-annotate" class="adder-label">Annotate</label>
```
robertknight: OK, we're getting close :)

In discussion on Slack, Nick just mentioned an issue where there is a region of what looks like the button that is between the `<button>` and the `<label>`. Clicking that doesn't create a new annotation. I think the right way to address this is going to be to put the label just as text inside the button. eg. `<button ...><span class="h-icon-annotate"...></span>Annotate</button>`, then use `display: flex; flex-direction: column` to layout the contents of the button.

h/static/scripts/annotator/guest.coffee
```diff
@@ -417,6 +433,7 @@ module.exports = class Guest extends Annotator
     event.preventDefault?()
     event.stopPropagation?()
     @adder.hide()
+    console.log($(event.target).data('action'))
```
robertknight: This was left in by mistake. You might want to have a look at modifying our JSHint config file to catch such mistakes.

robertknight:
Hi Sheetal - The styling looks correct and its now much easier to tweak. I noticed a few more things in testing this, proposed fixes are in this commit: https://github.com/hypothesis/h/commit/39ee2a684574f28770b8ad216f03ad43ab85ee63 . Could you have a look at that and let me know if you spot anything that is off from Conor's designs?

* There is a deadzone between the buttons and the visible edge of the toolbar. As you mentioned in the standup yesterday, this can be avoided by moving the spacing from outside the buttons into padding inside the buttons
* The cursor for the button elements needs to be set to `pointer`
* The rounded border for the adder was appearing feint at the corners because it was being partially obscured by the white background of the toolbar button. In 39ee2a6845 this is resolved by giving the individual buttons a transparent background
* Since we're using flexbox to lay out the contents of the toolbar button, we might as well use it for the toolbar itself - this makes everything much easier.

Lastly, before merging this, we should squash it down to one commit.

nickstenning:
Looks good to me! Great work!

PR #3079, opened 2016-03-10
Upgrade gevent

nickstenning:
Finally, a Python 3 compatible release of gevent, and one that fixes various bugs with HTTPS support in 1.0.2.

This commit also updates the pin for gunicorn to ensure we get a version compatible with this gevent.


chdorner:
LGTM

PR #3081, opened 2016-03-10
Annotation links

nickstenning:
Expose a `links` property on annotations, which can be used to store and retrieve (from the client-side) hypermedia references to other representations or related URLs for the annotation.


robertknight:
LGTM. I have some misgivings about the way that the request object and its gigantic interface gets passed around everywhere but that isn't causing us problems right now.

PR #3082, opened 2016-03-10
Add homepage nav link to Bioscience portal

nickstenning:
This is synchronized with the equivalent change to the WordPress navigation.


chdorner:
LGTM

PR #3084, opened 2016-03-10
Simplify API authentication in the client and fix #3083, #2924

### LINKED ISSUE #2924: Client auth state out of sync
Related to #2876 , emvincent reported that their client got into a state where the top menu bar showed them as being logged in but when they created a new annotation, they were shown a message indicating that they needed to log in:

screen shot 2016-01-20 at 15 42 51

The signin control state reflects `$scope.auth.username` whereas the annotation control uses the `user` property of annotations, which is initialized from `session.state.userid`. If `auth.username` and `session.state.userid` could get out of sync, with `session.state.userid` being empty, this would explain the issue.

### LINKED ISSUE #3083: Group annotations not loaded on first request after switching user account
### Steps to reproduce
1. Open client on a URL `$URL` and sign in as user `$USER`
2. Switch to a private group `$GROUP` and create an annotation on the page
3. Sign out
4. Sign in again

### Expected behaviour
1. After signing in at step 4, the client should switch back to the last-used group `$GROUP`
2. The annotation created in step 2 should be visible 

### Actual behaviour
The client switches back to `$GROUP` but the annotation is not loaded.

### Additional details
On the first call to the `/api/search` endpoint after signing in at step 4, no `Authorization` header is present. Subsequent search calls, eg. those made when switching groups, are authenticated.

robertknight:
In order to make API requests, the client needs to
fetch a JWT token and then configure 'angular-jwt'
to provide it on subsequent HTTP requests to
API endpoints.

This fairly simple task was complicated by
the inclusion of an emulation of the deprecated
Mozilla IdentityManager API - see https://developer.mozilla.org/en-US/docs/Web/API/IdentityManager

This commit replaces the identity module with
a much simpler implementation that only does
what we actually need at present:

 1. Enable the 'angular-jwt' interceptor which
    adds 'Authorization: Bearer <Token>' headers
    to API HTTP requests.

 2. Provide the JWT interceptor with a function
    which fetches JWT tokens and caches them.

The new implementation fixes two bugs in the previous
implementation:

 1. Cached API tokens were not invalidated properly when
    signing out (#3083).

    (In the old code, 'authPromise' was set to a rejected promise
     after signing out, but 'checkAuthentication()' checked for
     'authPromise' being _null_ when deciding whether to retrieve
     a new token. Consequently API requests made immediately
     after signing in could end up being unauthenticated).

 2. The value of $scope.auth.username and session.state.userid
    could get out of sync (#2924).

    In the new implementation, $scope.auth.username is always
    updated whenever the USER_CHANGED event is emitted and that
    event is always emitted when session.state.userid changes.

Fixes #3083
Fixes #2924


h/static/scripts/session.js
```diff
@@ -153,7 +154,7 @@ function session($http, $resource, $rootScope, flash, raven, settings) {
     return model;
   };
 
-  function process(data, headersGetter) {
+  function process(data) {
```
robertknight: I'm guessing Hound is referring to Node's "process" object here.

robertknight:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;rIvxXODr&#x2F;288-group-annotations-not-loaded-on-first-request-after-signing-in"> Group annotations not loaded on first request after signing in</a>

h/static/scripts/app-controller.js
```diff
@@ -46,37 +60,21 @@ module.exports = function AppController(
...
   });
 
-  identity.watch({
-    onlogin: function (identity) {
-      // Hide the account dialog
-      $scope.accountDialog.visible = false;
-      // Update the current logged-in user information
-      var userid = auth.userid(identity);
-      var parsed = parseAccountID(userid);
-      angular.copy({
```
nickstenning: This was previously using `angular.copy`, presumably to preserve the identify of the `$scope.auth` object. The new code doesn't do that -- are we sure that this behaviour doesn't need to be preserved?

robertknight: Yes. `$scope.auth` is now purely data derived from `session.state.userid` which is passed down to `<signin-control>` and `<group-list>` for controlling which parts of those components are displayed. In `signin_control.js` and `group_list.js`, those are already (prior to this commit) using one-way (`<`) bindings to indicate that auth is only an input property which is not modified by those components.

package.json
```diff
@@ -50,6 +50,7 @@
     "node-uuid": "^1.4.3",
     "page": "^1.6.4",
     "postcss": "^5.0.6",
+    "query-string": "^3.0.1",
```
nickstenning: Is there a substantial difference between `query-string` and native node `querystring`?

nickstenning:
This looks good to me. I've checked it out and tested as much as I can -- and it seems to behave correctly.

I think the `identity` module was well-intentioned but premature. We will need to have some kind of generic authentication and identity layer in place in order to talk to multiple backends, but this change does at least make clear that we're reliant on cookie-authenticated sessions for now.

package.json
nickstenning: Oh, right, `query-string` is half the size.

robertknight: Its [simpler](https://github.com/sindresorhus/query-string/blob/master/index.js) and appears to be better maintained than the 'querystring' implementation which Browserify is using by default (`querystring-es3`).

PR #3085, opened 2016-03-10
(1/3) Open sidebar and set selection when '#annotations' URL fragment is present

robertknight:
This is the first part of adding support for '#annotations' URL fragments in the client.
This PR covers:

1. Detecting when a tab URL contains '#annotations' and automatically injecting H in the Chrome extension
2. Detecting when Annotator is loaded into a document (by the extension, bookmarklet or embedder) containing '#annotations' in the URL fragment, automatically opening the sidebar and passing the annotation ID through to the client
3. Initializing the annotation selection from the value passed through in step (2)

Subsequent PRs will cover:
 1. Switching the focused group to the one containing the annotation, if necessary
 2. Preserving the intent to select an annotation after signing in
 3. Displaying a suitable message if the user does not have permission to access the annotation

### DIFF
--- a/h/browser/chrome/lib/hypothesis-chrome-extension.js
+++ b/h/browser/chrome/lib/hypothesis-chrome-extension.js
@@ -10,6 +10,15 @@ var TabStore = require('./tab-store');
 var TAB_STATUS_LOADING = 'loading';
 var TAB_STATUS_COMPLETE = 'complete';
 
+/**
+ * Returns true if a tab URL contains a link to an annotation,
+ * which should result in Hypothesis being automatically injected into
+ * the tab.
+ */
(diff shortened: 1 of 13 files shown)


robertknight:

nickstenning:
As discussed in Slack, this is all looking good except that we need to ensure that multiple extensions don't all try to inject at the same time.

robertknight:
Previous comment addressed in a separate PR in #3097

nickstenning:
LGTM. One question to consider -- I assume we want to scroll to the annotated content as well?

robertknight:
> I assume we want to scroll to the annotated content as well?

Yes. That's an important addition that's needed. I'll make a note on the card.

PR #3086, opened 2016-03-11
(2/3) Switch to group containing selected annotation when sidebar loads

robertknight:
This PR is a follow up at #3085 and will need rebasing when that lands. It implements support for automatically switching to the group containing the selected annotation when the sidebar loads.

There are two parts to this:

 1. The first commit extracts the logic for paging through and processing API search results into a separate class
 2. The second commit adds the logic to load annotations from all groups when there is a selection, then find and focus the appropriate group and filter the results by that group.

The next step after this will be to persist the selection when switching accounts, so that when a user receives a direct-linked group annotation but is logged out, then signs in, they see that annotation.

### DIFF
--- a/h/static/scripts/widget-controller.js
+++ b/h/static/scripts/widget-controller.js
@@ -3,6 +3,7 @@
 var angular = require('angular');
 
 var events = require('./events');
+var SearchClient = require('./search-client');
 
 // @ngInject
 module.exports = function WidgetController(
@@ -12,9 +13,6 @@ module.exports = function WidgetController(
   $scope.threadRoot = threading.root;
   $scope.sortOptions = ['Newest', 'Oldest', 'Location'];
 
-  var DEFAULT_CHUNK_SIZE = 200;
-  var loaded = [];
-
   var _resetAnnotations = function () {
     // Unload all the annotations
     annotationMapper.unloadAnnotations(threading.annotationList());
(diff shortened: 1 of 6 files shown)


robertknight:

h/static/scripts/widget-controller.js
```diff
@@ -12,61 +28,107 @@ module.exports = function WidgetController(
...
-    for (var i = 0, f; i < frames.length; i++) {
-      f = frames[i];
-      var ref;
-      if (ref = f.uri, loaded.indexOf(ref) >= 0) {
-        continue;
+    searchClients.forEach(function (client) {
+      client.cancel();
+    });
+
+    var urls = frames.reduce(function (urls, frame) {
```
robertknight: FIXME - If new frames are added to `crossframe.frames`, this will load annotations for _all_ frames, not just the newly-added frames. That was the reason why the _loaded_ array was maintained previously to keep track of which frames annotations had been loaded for.

h/static/scripts/widget-controller.js
```diff
@@ -12,61 +28,107 @@ module.exports = function WidgetController(
...
-    var q = angular.extend(queryCore, query);
-    q._separate_replies = true;
-
-    store.SearchResource.get(q, function (results) {
-      var total = results.total;
-      offset += results.rows.length;
-      if (offset < total) {
-        _loadAnnotationsFrom(query, offset);
-      }
+  var searchClients = [];
```
robertknight: TODO: Add a comment here noting that `searchClients` is an `Array<SearchClient>` of active annotation searches.

h/static/scripts/widget-controller.js
```diff
@@ -12,61 +28,107 @@ module.exports = function WidgetController(
...
+        results = results.filter(function (result) {
+          return result.group === groupID;
+        });
+      }
+      if (results.length) {
+        annotationMapper.loadAnnotations(results);
+      }
+    });
+    searchClient.on('end', function () {
+      searchClients.splice(searchClients.indexOf(searchClient), 1);
```
robertknight: FIXME: SearchClient's `cancel` event does not currently emit the `end` event, so the `searchClients` array will not be cleared if in-flight searches are canceled below.

h/static/scripts/widget-controller.js
```diff
@@ -12,61 +28,107 @@ module.exports = function WidgetController(
...
+   *        to load annotations for.
+   */
   var loadAnnotations = function (frames) {
-    for (var i = 0, f; i < frames.length; i++) {
-      f = frames[i];
-      var ref;
-      if (ref = f.uri, loaded.indexOf(ref) >= 0) {
-        continue;
+    searchClients.forEach(function (client) {
+      client.cancel();
```
robertknight: FIXME: See the note about canceling searches above.

h/static/scripts/widget-controller.js
```diff
@@ -12,61 +28,107 @@ module.exports = function WidgetController(
...
+      null : groups.focused().id;
+
+    for (var i=0; i < urls.length; i++) {
+      _loadAnnotationsFor(urls[i], group);
     }
 
-    if (loaded.length > 0) {
-      streamFilter.resetFilter().addClause('/uri', 'one_of', loaded);
+    if (urls.length > 0) {
+      streamFilter.resetFilter().addClause('/uri', 'one_of', urls);
```
robertknight: NOTE: When fixing the above issue related to loading annotations, make sure that we still add all URLs to the filter here.

robertknight:
I'm going to close this and open a new PR which includes both selecting the right group and focusing the annotation on load.

PR #3087, opened 2016-03-11
Annotation links round 2

nickstenning:
This PR makes it possible for us to generate `incontext` links on rendered annotations for direct linking. See previous discussion in #3049.

It's split into a series of commits that probably make most sense to review one by one. In order, they do the following:

1. Fixes a bug in the annotation presenter class.
2. Makes the generation of annotation `links` properties extensible, so that we don't have to hard-code Hypothesis-specific links in the `h.api` module.
3. Makes the default `json` link use this "link generator" functionality.
4. Ensures the "link generator" configurator hooks are available to the rest of the application.
5. Adds an `html` link type that points to the "standalone annotation page" -- this can replace the manual construction of such URLs client-side.
6. Makes it possible for a link generator to opt out of generating a link (needed so that direct linking can be toggled using a feature flag).
7. Adds the `direct_linking` feature flag.
8. Adds `incontext` links to annotations if a) the `direct_linking` flag is toggled on and b) `BOUNCER_URL` is set in the application environment.

### DIFF
--- a/h/api/__init__.py
+++ b/h/api/__init__.py
@@ -3,6 +3,7 @@
 
 def includeme(config):
     config.include('h.api.db')
+    config.include('h.api.presenters')
     config.include('h.api.search')
     config.include('h.api.views')
 
--- a/h/api/presenters.py
+++ b/h/api/presenters.py
@@ -189,7 +189,14 @@ def deep_merge_dict(a, b):
             a[k] = v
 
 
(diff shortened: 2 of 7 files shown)


chdorner:
LGTM. I will let @robertknight have a look as well.

robertknight:
LGTM :+1:

PR #3089, opened 2016-03-11
Remove account claim package

nickstenning:
This package was used to allow users who registered to "claim a username" before our launch to do just that.

It is now 6 months since the last of the "claim your username" emails were sent out, and more than a year since the first. 8 people have used this facility in the last 2 months.

Out of an abundance of caution, I'd like to remove code that interacts with the accounts system if it's no longer needed. Given that we can handle the very small number of remaining username claims via user support, I think it's time to remove `h.claim`.

_No rush to review this one._


chdorner:
I've just rebased master. There was a merge conflict in `h/features.py`.

chdorner:
LGTM

PR #3090, opened 2016-03-11
Split out elastic.DocumentMeta type normalisation

### LINKED ISSUE #3062: User is unable to edit an annotation or page note in a specific web page
### Steps to reproduce
1.  Go here https://slate.adobe.com/cp/uOGoa/
2.  Make an annotation or page note.  
3.  Attempt to edit the annotation or page note

### Expected behaviour
Annotation is saved when clicking on post button.

### Actual behaviour
Internal server error message appears, annotation is not saved. 

I was able to repro in Mac OSx Chrome and Firefox.  User appears to be on Windows for Chrome.

### Additional details
The console shows: 
https://hypothes.is/api/annotations/AVNTLqTtH9ZO4OKSlGvq 
Failed to load resource: the server responded with a status of 500 (OK)

User error:
500error

chdorner:
Due to Elasticsearch's dynamic mapping creation, we can't normalise the
type string of an elastic DocumentMeta. Creating an annotation would
still work since it doesn't run this code, but updating an existing
annotation fails when it tries do create or update the Document ES
document because it uses the normalised version which then clashes with
the mapping that Elasticsearch automatically created based on old data.

I've moved the normalisation part into its own method
(`normalized_type`) in order to still be able to use it in the
migration, and thus automatically clean up our document metadata when we
migrate it from Elasticsearch to Postgres.

This bug was introduced when we merged #3011 and used the new
presentation layer for rendering Elasticsearch annotations.

Fixes #3062

### DIFF
--- a/h/api/models/elastic.py
+++ b/h/api/models/elastic.py
@@ -382,17 +382,26 @@ class DocumentMeta(dict):
 
     @property
     def type(self):
-        value = self.get('type')
+        return self.get('type')
+
+    @property
+    def value(self):
+        return self.get('value')
+
+    @property
+    def normalized_type(self):
+        """
+        Normalized version of the type string
+
+        This should only be used in the Postgres migration script.
+        """
+        value = self.type
         if value:
             value = value.lower().replace(':', '.')
             value = re.sub(r'\.{2,}', '.', value)
             value = re.sub(r'^og\.', 'facebook.', value)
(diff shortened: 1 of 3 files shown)


nickstenning:
The laws of unintended consequences strike again. Or, as @judell says: "no good deed goes unpunished."

PR #3093, opened 2016-03-14
Display message if a direct-linked annotation is not available

robertknight:
This is based on **#3108** and will need to be rebased and merged afterwards.

This adds a message in the sidebar which is displayed if a direct-linked annotation is selected but that annotation is not available, after fetching all annotations for the current page.

### DIFF
--- a/h/static/scripts/widget-controller.js
+++ b/h/static/scripts/widget-controller.js
@@ -3,12 +3,22 @@
 var events = require('./events');
 var SearchClient = require('./search-client');
 
+function firstKey(object) {
+  for (var k in object) {
+    if (!object.hasOwnProperty(k)) {
+      continue;
+    }
+    return k;
+  }
+  return null;
+}
+
 /**
  * Returns the group ID of the first annotation in `results` whose
  * ID is a key in `selection`.
  */
 function groupIDFromSelection(selection, results) {
-  var id = Object.keys(selection)[0];
+  var id = firstKey(selection);
   var annot = results.find(function (annot) {
     return annot.id === id;
   });
@@ -191,6 +201,12 @@ module.exports = function WidgetController(
       return false;
     }
     return annotation.$$tag in $scope.focusedAnnotations;
+  };
+
+  $scope.selectedAnnotationUnavailable = function () {
+    return searchClients.length === 0 &&
+           annotationUI.hasSelectedAnnotations() &&
+           !threading.idTable[firstKey(annotationUI.selectedAnnotationMap)];
   };
 
   $rootScope.$on(events.BEFORE_ANNOTATION_CREATED, function (event, data) {
--- a/h/static/styles/thread.scss
+++ b/h/static/styles/thread.scss
@@ -31,7 +31,10 @@ $thread-padding: $annotation-card-left-padding;
   }
 
   &__icon {
-    font-size: 50px;
+    background-image: url(../images/icons/lock.svg);
+    background-repeat: no-repeat;
+    width: 56px;
+    height: 48px;
   }
 }
 
--- a/h/static/styles/thread.scss
+++ b/h/static/styles/thread.scss
@@ -15,6 +15,26 @@ $thread-padding: $annotation-card-left-padding;
   }
 }
 
+.annotation-unavailable-message {
+  display: flex;
+  flex-direction: column;
+  border: 1px solid $gray-lighter;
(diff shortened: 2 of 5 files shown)


robertknight:

h/static/styles/thread.scss
```diff
@@ -15,6 +15,26 @@ $thread-padding: $annotation-card-left-padding;
   }
 }
 
+.annotation-unavailable-message {
+  display: flex;
+  flex-direction: column;
+  border: 1px solid $gray-lighter;
+  padding-top: 30px;
```
robertknight: TODO: The multiples of 10px here are chosen to align with the 10px grid which the new design is based off. Now would be a good time to introduce that as a variable in `variables.scss`

h/static/scripts/search-client.js
```diff
@@ -0,0 +1,98 @@
...
+ * SearchClient handles paging through results, canceling search etc.
+ *
+ * @param {Object} resource - ngResource class instance for the /search API
+ * @param {Object} opts - Search options
+ * @constructor
+ */
+function SearchClient(resource, opts) {
+  opts = opts || {};
+
+  var DEFAULT_CHUNK_SIZE = 200;
```
nickstenning: This has increased from 100 to 200 -- that is possibly ok, but it might be better to leave it as is for now because of [the rather ugly way replies are loaded on the server](https://github.com/hypothesis/h/blob/3e8d22814a4e3ce8e6e743d5ca1ae59820c5dbe0/h/api/search/core.py#L58).

h/static/scripts/groups.js
```diff
@@ -72,12 +72,15 @@ function groups(localStorage, session, settings, $rootScope, $http) {
...
-     focusedGroup = g;
-     localStorage.setItem(STORAGE_KEY, g.id);
-     $rootScope.$broadcast(events.GROUP_FOCUSED, g.id);
-   }
+    var prevFocused = focused();
+    var g = get(id);
+    if (g) {
+      focusedGroup = g;
+      localStorage.setItem(STORAGE_KEY, g.id);
+      if (prevFocused.id !== g.id) {
```
nickstenning: This change seems a bit out of place in this commit. It certainly looks harmless, but I don't see how it has anything to do with pagination of search results, so either an explanation in the commit message, or splitting it into its own commit wouldn't go amiss.

nickstenning:
Ah, just realised I'm reviewing these in the wrong order. Switching to #3108.

h/static/scripts/widget-controller.js
```diff
@@ -193,6 +203,12 @@ module.exports = function WidgetController(
     return annotation.$$tag in $scope.focusedAnnotations;
   };
 
+  $scope.selectedAnnotationUnavailable = function () {
+    return searchClients.length === 0 &&
```
nickstenning: Related to a comment I made on #3108, relying on `searchClients.length` to determine whether requests are in-flight to fetch annotations is pretty opaque. At the very least add a comment, or better add a helper function elsewhere in this file to hide the details of `searchClients` from other parts of the code.

h/templates/client/viewer.html
```diff
@@ -21,6 +21,13 @@
       on-change-sort-by="sort.name = sortBy">
     </sort-dropdown>
   </li>
+  <li class="annotation-unavailable-message"
+      ng-if="selectedAnnotationUnavailable()">
+    <span class="h-icon-lock annotation-unavailable-message__icon"></span>
```
nickstenning: This icon isn't the one on the Trello card. Is that agreed on?

h/static/scripts/groups.js
robertknight: I'll add an explanation here - the gist is that focusing a group triggers a reload of annotations in `widget-controller.js`, but we don't want to do that unnecessarily if `groups.focus()` is called with a group that is the same as the current group.

robertknight: Update - I split this into a separate commit in #3108 and added an explanation plus tests.

h/templates/client/viewer.html
robertknight: No. It needs to be updated to use the SVG that Conor provided.

nickstenning:
Ok, then checklist to get this merged:

- [ ] uses correct icon for the padlock

nickstenning:
LGTM.

PR #3094, opened 2016-03-14
Postgres read support for storage.expand_uri

chdorner:
I initially wanted to get the badge endpoint to be able to work with Postgres, later I've discovered that the badge endpoint doesn't take URI equivalence into account. The upside is that we now have Postgres support for `h.api.storage.expand_uri`.

### DIFF
--- a/h/streamer/websocket.py
+++ b/h/streamer/websocket.py
@@ -71,7 +71,7 @@ def handle_message(message):
             jsonschema.validate(payload, filter.SCHEMA)
 
             # Add backend expands for clauses
-            _expand_clauses(payload)
+            _expand_clauses(socket.request, payload)
 
             socket.filter = filter.FilterHandler(payload)
         elif msg_type == 'client_id':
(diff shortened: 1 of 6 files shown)


h/streamer/websocket.py
```diff
@@ -71,7 +71,7 @@ def handle_message(message):
             jsonschema.validate(payload, filter.SCHEMA)
 
             # Add backend expands for clauses
-            _expand_clauses(payload)
+            _expand_clauses(socket.request, payload)
```
chdorner: @nickstenning Is the `socket.request.db` the same session as the session created in [streamer.py](https://github.com/hypothesis/h/blob/master/h/streamer/streamer.py#L72)?

chdorner:
I was finally able to get the websockets to work locally (without SSL). And could verify that these changes do not break expanding the URI when a websocket connects.

h/streamer/websocket.py
nickstenning: I think it should be the same due to the way that `scoped_session` works, but I think we should probably check.

nickstenning: If it's not, that's a very good point and we should probably pass `session` all the way through.

chdorner: Now that I can actually run websockets. It is the same DB session.
Considering the short discussion on Slack regarding passing through the session all the way, which would unfortunately mean that the `h.api.storage` functions would have to accept the request and the session in addition to other arguments. Is it still worth it?

The other solution I could come up with is that the `h.api.storage` function could accept an optional DB session (as a keyword argument) for exactly these use cases where it's hard to follow.

Thoughts?

nickstenning:
LGTM.

PR #3097, opened 2016-03-16
Enable Chrome extension to detect when H is already present on page

robertknight:
This reworks the way that Hypothesis detects whether it is already present on the page in embed.js, in a way that enables it to report back the app.html URL for the existing instance to the Chrome extension.

The Chrome extension can then update its state for that tab to reflect the fact that it is not injected.

In this PR, that is done by simply setting the tab state to inactive. We could in future add an indicator that the user tried to activate H but that another instance is already active, or have the extension override the instance of H on the page.

 * Detect the <link> element added to the page by embed.js,
   rather than window.annotator. This enables detection to work
   when embed.js is run in an isolated world that shares the DOM
   but not the JS environment, as is the case for content scripts
   in Chrome and Firefox.

 * Change unloading to operate by firing an event at the <link>
   element. This enables a content script to trigger unloading
   of H from an isolated JS environment.

 * Change injection of embed.js in the Chrome extension to
   execute embed.js as a content script, rather than by
   adding it as a `<script>` tag. This enables embed.js to
   access extension APIs and also report its result back
   to the extension via the return value of the script.

   A side benefit is that it also avoids a bug in Firefox
   current WebExtensions implementation where `<script>` tags
   added to the page by a content script are not executed.

### DIFF
--- a/h/browser/chrome/lib/sidebar-injector.js
+++ b/h/browser/chrome/lib/sidebar-injector.js
@@ -2,7 +2,6 @@
 
 var detectContentType = require('./detect-content-type');
 var errors = require('./errors');
-var settings = require('./settings');
 var util = require('./util');
 
 var CONTENT_TYPE_HTML = 'HTML';
@@ -12,11 +11,23 @@ function toIIFEString(fn) {
   return '(' + fn.toString() + ')()';
 }
 
-function addMetaTagFn(name, content) {
-  var metaTag = document.createElement('meta');
-  metaTag.name = name;
-  metaTag.content = content;
-  document.head.appendChild(metaTag);
+/**
+ * Extract the value returned by a content script injected via
+ * chrome.tabs.executeScript() into the main frame of a page.
+ *
+ * executeScript() returns an array of results, one per frame which the script
+ * was injected into.
+ *
+ * See https://developer.chrome.com/extensions/tabs#method-executeScript
+ *
+ * @param {Array<any>} result
+ */
+function extractContentScriptResult(result) {
+  if (Array.isArray(result) && result.length > 0) {
+    return result[0];
+  } else {
+    return;
+  }
 }
 
 /* The SidebarInjector is used to deploy and remove the Hypothesis sidebar
@@ -120,8 +131,9 @@ function SidebarInjector(chromeTabs, dependencies) {
         return executeScriptFn(tab.id, {
             code: toIIFEString(detectContentType)
           }).then(function (frameResults) {
-            if (Array.isArray(frameResults)) {
-              return frameResults[0].type;
+            var result = extractContentScriptResult(frameResults);
+            if (result) {
+              return result.type;
             } else {
               // If the content script threw an exception,
               // frameResults may be null or undefined.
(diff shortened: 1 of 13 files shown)


h/browser/chrome/lib/sidebar-injector.js
```diff
@@ -12,11 +11,12 @@ function toIIFEString(fn) {
   return '(' + fn.toString() + ')()';
 }
 
-function addMetaTagFn(name, content) {
-  var metaTag = document.createElement('meta');
-  metaTag.name = name;
-  metaTag.content = content;
-  document.head.appendChild(metaTag);
+function extractContentScriptResult(result) {
```
nickstenning: A comment here linking to https://developer.chrome.com/extensions/tabs#method-executeScript and explaining (briefly) why the result is an array would be good.

h/browser/chrome/lib/sidebar-injector.js
```diff
@@ -120,8 +120,9 @@ function SidebarInjector(chromeTabs, dependencies) {
         return executeScriptFn(tab.id, {
             code: toIIFEString(detectContentType)
           }).then(function (frameResults) {
```
nickstenning: I know it's not new in this PR, but a nice feature of promises is that this kind of nesting isn't needed. This is what we've got here:

```javascript
foo(x).then(function (y) {
  return bar(y).then(function (z) {
    return baz(z);
  });
});
```

which is much clearer as:

```javascript
foo(x)
  .then(bar)
  .then(baz)
```

nickstenning:
I think this is a really nice approach and opens up all kinds of possibilities for improved messaging around such scenarios in the future. Update the tests and let's get this merged!

nickstenning:
This is a very welcome change. While there is huge room for improvement about what we communicate to the user in these scenarios, I think not blindly injecting twice is way better than the behaviour before, and correctly communicating the active/inactive state in the Chrome extension is great.

h/static/extension/config.js
```diff
@@ -1,8 +0,0 @@
-function hypothesisConfig() {
-  // Pages on our site can include a meta tag to trigger specific behaviour
-  // when the extension loads.
-  var hypothesisIntent = document.querySelector('[name="hypothesis-intent"]');
```
nickstenning: Sorry if this is a dumb question, but if we've removed this, then how do we now trigger the extension to open automatically on the welcome page after installation?

nickstenning: Also, if it's gone from here it should be gone from `h/templates/help.html.jinja2`

robertknight: Gah. I'd moved that function for detecting the first-run use case to config.js in the direct-linking branch. But when I separated out 291-detect-existing-h-client into a separate branch where config.js isn't present, I hadn't moved the code back to `annotator/main.js`.

robertknight: So there were two different mechanisms here that could trigger the sidebar to show the login dialog when it opened, one was to provide a `hypothesisConfig()` function with a 'firstRun' property, the other was to add a `hypothesis-intent` meta tag to the page.

The reason for this second option was presumably so that it would work if `annotator/main.js` was run in an isolated JS world (possibly in the old Firefox extension) which couldn't see the page's JS.

I think we may want to look at doing that again in future, but we're not currently using it so I've modified `help.html.jinja` to always use the `hypothesisConfig()` method for the moment.

PR #3098, opened 2016-03-16
Refactor and speed up client load in embed.js

robertknight:
embed.js created a public but undocumented window.hypothesisInstall()
function which had functionality ostensibly for customizing the
injection method. This was unused however.

What was much worse is that `<script>` and `<style>` tags were loaded
_serially_ by adding each `<script>` tag to the page and then waiting for
the `onload` callback to be invoked before adding the next. This added
needless latency to app startup.

This commit makes the install function private, simplifies it and adds
all script and style tags to the page immediately so that the browser
can load resources concurrently.

### DIFF
--- a/h/templates/embed.js.jinja2
+++ b/h/templates/embed.js.jinja2
@@ -35,6 +35,10 @@ function injectScript(src) {
   var script = document.createElement('script');
   script.type = 'text/javascript';
   script.src = resolve(src);
+
+  // Set 'async' to false to maintain execution order of scripts.
+  // See https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script
+  script.async = false;
   document.head.appendChild(script);
 };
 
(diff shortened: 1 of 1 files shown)


nickstenning:
LGTM.

nickstenning:
I'm retracting my "LGTM", because testing this locally it seems that there is order dependence to the bundles loaded. About 1 in 3 page loads I can get an error in the console such as:

```
_prelude.js:1 Uncaught Error: Cannot find module 'jquery'
```

or

```
main:1 Uncaught Error: Cannot find module '/h/static/scripts/polyfills.js'
```

robertknight:
Hmm, yes indeed. There is an order dependence when it comes to actually executing the code now I think about it. The bundles can be loaded concurrently but `injector.bundle.js` cannot start executing until the other bundles have all been loaded.

robertknight:
However, we are adding the bundles in the correct order, as defined in `assets.ini` and I thought that execution of the scripts would also be guaranteed to happen in that order. Looks like that assumption may have been faulty.

nickstenning:
LGTM.

PR #3099, opened 2016-03-16
Only add self-claim DocumentURI when claimant is set

chdorner:
When this is being used from a Document that has been loaded from
Elasticsearch (e.g. when expanding URIs), we don't have the claimant in
the Document, which means there is no point in adding a self-claim URI
where the `uri` field would be missing.


nickstenning:
LGTM.

PR #3104, opened 2016-03-17
Fix tab state being set to invalid value when opening a new tab

robertknight:
Fix type error - `newActiveState` was being initialized with
state.getState() which returns a TabState and then passed as the value
for the 'state' property of the updated TabState if the tab URL did not
contain a '#annotations:' fragment.

Also make import of Object.assign() polyfill consistent with how it is
done in the main application.


nickstenning:
LGTM.

PR #3105, opened 2016-03-17
Sheetaluk/new sharer

sheetaluk:


h/static/styles/share-link.scss
```diff
@@ -27,6 +33,30 @@
...
+
+.share-target__icon {
+  color: #7A7A7A;
+  text-decoration: none;
+  font-size: 20px;
+  cursor: pointer;
+}
+
+.share-target__icon:hover {
+  color: #202020 !important;
```
robertknight: What is `!important` for here? - That is something we should try to avoid as much as possible in the app.

h/static/styles/share-link.scss
```diff
@@ -27,6 +33,30 @@
...
+  font-size: $body1-font-size;
+  padding: 10px;
+}
+
+.share-target__label {
+  font-weight: bold;
+}
+
+.share-target__icon {
+  color: #7A7A7A;
```
robertknight: Are these colors from the new design palette? If so I'd suggest that we introduce variables for them in `variables.scss`, using Conor's naming for them.

h/templates/client/annotation.html
```diff
@@ -154,18 +154,50 @@
...
+          </a>
+          <input class="share-dialog__link" type="text" value="{{vm.annotationURI}}" readonly>
+        </span>
+      </span>
+      <span class="share-dialog-wrapper">
+        <button class="small btn btn-clean"
+          ng-click="vm.share($event)">
+          <i class="h-icon-share btn-icon"></i>
+        </button>
+        <span class="share-dialog" ng-click="$event.stopPropagation()">
```
robertknight: The new share dialog has a slight drop shadow in the design to make it stand out from the card a little. That needs to be added here - feel free to add it to both the new and old if that makes things easier.

h/templates/client/annotation.html
```diff
@@ -154,18 +154,50 @@
     <div class="annotation-actions" ng-if="!vm.isSaving && !vm.editing() && vm.id()">
       <div ng-show="vm.isSaving">Saving…</div>
       <button class="small btn btn-clean"
-              ng-click="vm.reply()"
-              ><i class="h-icon-reply btn-icon"></i> Reply</button>
+        ng-click="vm.reply()">
+        <i class="h-icon-reply btn-icon"></i>
+        Reply
+      </button>
       <span class="share-dialog-wrapper">
```
robertknight: I think now, or as a follow up PR (you decide) would be a good time to extract this out into its own component. See 'signin-control.{js|html|scss}' for an example of a similar simple component which renders some data and has a couple of events.

h/static/styles/share-link.scss
```diff
@@ -12,7 +12,21 @@
 }
 
 .share-link:hover {
-  text-decoration: underline;
+  width: 100%;
+  color: $gray-dark;
+}
+
+.share-dialog-link {
```
robertknight: This is starting to get confusing now because we're mixing styles for three different components (the groups list, the share pane triggered by the drop-down and the in-card share dialog) in the same file. I'll merge this now but I suggest as a follow up extracting the share dialog out into its own component (so we'd have `share-dialog.scss`, `share_dialog.html` and `share-dialog.js`)

PR #3108, opened 2016-03-18
Switch to appropriate group and scroll to direct linked annotation when sidebar loads

robertknight:
This PR implements switching to the appropriate group and scrolling to a direct-linked annotation when the sidebar loads.

In order to determine which group the annotation is in, the sidebar will load annotations from all groups rather than just the focused group if there is a selection when fetching annotations. In the common case where there are fewer than 200 annotations in total across all groups, this avoids any extra queries to the search endpoint. If there are _more_ than 200 annotations in total, the client will fetch all annotations before it displays anything if there is a selection. Once the annotations have been received, the client finds the group containing the selected annotation (if any), focuses that group and then filters the annotations by that group before displaying them.

The changes in this PR are:
 * Extracting pagination and processing of search results out of the sidebar's controller into a separate `SearchClient` class.
 * Changing the searching logic depending on whether or not there is a selection at the point when annotations are loaded
 * Finding and focusing the appropriate group once annotations are loaded
 * Scrolling to the annotation once the sidebar receives a 'sync' event from the client reporting that it has anchored the annotation in the page.

Commit-by-commit review might be easier here.

### DIFF
--- a/h/static/scripts/events.js
+++ b/h/static/scripts/events.js
@@ -1,16 +1,40 @@
+'use strict';
+
 /**
  * This module defines the set of global events that are dispatched
  * on $rootScope
  */
 module.exports = {
-  /** Broadcast when the currently selected group changes */
-  GROUP_FOCUSED: 'groupFocused',
-  /** Broadcast when the list of groups changes */
+  // Session state changes
+
+  /** The list of groups changed */
   GROUPS_CHANGED: 'groupsChanged',
-  /** Broadcast when the signed-in user changes */
+  /** The signed-in user changed */
   USER_CHANGED: 'userChanged',
-  /** Broadcast when the session state is updated.
-   * This event is NOT broadcast after the initial session load.
+  /**
+   * The session state was updated.
    */
   SESSION_CHANGED: 'sessionChanged',
+
+  // UI state changes
+
+  /** The currently selected group changed */
+  GROUP_FOCUSED: 'groupFocused',
+
+  // Annotation events
+
+  /** A new annotation has been created locally. */
+  BEFORE_ANNOTATION_CREATED: 'beforeAnnotationCreated',
+
+  /** Annotations were anchored in a connected document. */
+  ANNOTATIONS_SYNCED: 'sync',
+
+  /** An annotation was created on the server and assigned an ID. */
+  ANNOTATION_CREATED: 'annotationCreated',
+
+  /** An annotation was either deleted or unloaded. */
+  ANNOTATION_DELETED: 'annotationDeleted',
+
+  /** A set of annotations were loaded from the server. */
+  ANNOTATIONS_LOADED: 'annotationsLoaded',
 };
--- a/h/static/scripts/search-client.js
+++ b/h/static/scripts/search-client.js
@@ -1,0 +1,98 @@
+'use strict';
+
+var EventEmitter = require('tiny-emitter');
+var inherits = require('inherits');
+
+/**
+ * Client for the Hypothesis search API.
+ *
+ * SearchClient handles paging through results, canceling search etc.
+ *
+ * @param {Object} resource - ngResource class instance for the /search API
+ * @param {Object} opts - Search options
+ * @constructor
+ */
+function SearchClient(resource, opts) {
+  opts = opts || {};
+
+  var DEFAULT_CHUNK_SIZE = 200;
+  this._resource = resource;
+  this._chunkSize = opts.chunkSize || DEFAULT_CHUNK_SIZE;
+  if (typeof opts.incremental !== 'undefined') {
+    this._incremental = opts.incremental;
+  } else {
+    this._incremental = true;
+  }
+  this._canceled = false;
+}
+inherits(SearchClient, EventEmitter);
+
+SearchClient.prototype._getBatch = function (query, offset) {
+  var searchQuery = Object.assign({
+    limit: this._chunkSize,
+    offset: offset,
+    sort: 'created',
+    order: 'asc',
+    _separate_replies: true,
+  }, query);
+
+  var self = this;
+  this._resource.get(searchQuery).$promise.then(function (results) {
+    if (self._canceled) {
+      return;
+    }
+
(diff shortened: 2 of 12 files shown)


h/static/scripts/events.js
```diff
@@ -1,16 +1,40 @@
...
+
+  // UI state changes
+
+  /** The currently selected group changed */
+  GROUP_FOCUSED: 'groupFocused',
+
+  // Annotation events
+
+  /** A new annotation has been created locally. */
+  BEFORE_ANNOTATION_CREATED: 'beforeAnnotationCreated',
```
robertknight: These are all pre-existing events, I've just defined constants for them here so that they can all be documented in one place.

h/static/scripts/search-client.js
```diff
@@ -0,0 +1,98 @@
...
+
+var EventEmitter = require('tiny-emitter');
+var inherits = require('inherits');
+
+/**
+ * Client for the Hypothesis search API.
+ *
+ * SearchClient handles paging through results, canceling search etc.
+ *
+ * @param {Object} resource - ngResource class instance for the /search API
```
robertknight: As a separate task, I think we can look at replacing `ngResource` with a simple fetch here. I'll have to check exactly what it does for us at the moment that we still rely on.

h/static/scripts/search-client.js
```diff
@@ -0,0 +1,98 @@
...
+}
+inherits(SearchClient, EventEmitter);
+
+SearchClient.prototype._getBatch = function (query, offset) {
+  var searchQuery = Object.assign({
+    limit: this._chunkSize,
+    offset: offset,
+    sort: 'created',
+    order: 'asc',
+    _separate_replies: true,
```
robertknight: I'm passing `_separate_replies` here because that was what the previous code did. I haven't verified that we still need to do this though.

h/static/scripts/widget-controller.js
```diff
@@ -5,6 +5,21 @@ var angular = require('angular');
 var events = require('./events');
 var SearchClient = require('./search-client');
 
+/**
+ * Returns the group ID of the first annotation in @p results whose
```
nickstenning: `@p`?

h/static/scripts/widget-controller.js
```diff
@@ -67,8 +114,17 @@ module.exports = function WidgetController(
   };
 
   $scope.$on(events.GROUP_FOCUSED, function () {
+    if (searchClients.length) {
+      // If the current group changes as a _result_ of loading annotations,
```
nickstenning: This comment makes sense, but it's not clear how it relates to the code. Perhaps you could explain why `searchClients.length` would be zero if `the group changes because of a user action during loading of annotations`?

h/static/scripts/widget-controller.js
```diff
@@ -13,28 +28,46 @@ module.exports = function WidgetController(
...
 
   var searchClients = [];
 
   function _loadAnnotationsFor(uri, group) {
-    var searchClient = new SearchClient(store.SearchResource);
+    var searchClient = new SearchClient(store.SearchResource, {
+      // If no group is specified, we are fetching annotations from
+      // all groups in order to find out which group contains the selected
+      // annotation, therefore we need to load all chunks before processing
+      // the results
```
nickstenning: This is fine. What I supposed I had actually envisioned was that we would always fetch all annotations, and wouldn't then need to issue an additional query when the user switched between groups. I recognise that's probably a lot more work (and would need changes to the way the streamer works, too) but it's worth bearing in mind as a possible simplification in the future.

h/static/scripts/search-client.js
nickstenning: You do still need to do this in order to get the `replies` field back in the search results. It's a hack until we can rearrange the search API to return replies for annotations inline with those annotations.

nickstenning:
Okay, this looks mostly fine other than a few comments about readability and documentation.

Also, please see the two comments I made on #3093 before I realised I was commenting on the wrong PR. (In particular about bumping the default resultset size up to 200.)

robertknight:
> This is fine. What I suppose I had actually envisioned was that we would always
> fetch all annotations, and wouldn't then need to issue an additional query when
> the user switched between groups

Good point. This would also make it easy to display annotation counts in the groups list.

I've got a separate branch which refactors the threading system and in the process makes it easier to separate the annotations that are loaded from those that are actually displayed. We could definitely make this change quite easily in that branch. I'm not sure about how much work it would be here - would have to experiment.

nickstenning:
GitHub has also marked [two](https://github.com/hypothesis/h/pull/3108#discussion_r56810959), [comments](https://github.com/hypothesis/h/pull/3108#discussion_r56811535) as being made on outdated diffs, which I don't think is true.

robertknight:
> GitHub has also marked two comments as being made on outdated diffs, which I don't think is true.

@nickstenning Indeed - I haven't pushed any changes this morning and those comments are still valid. Unless you happened to have an old diff open in a tab from last week or something like that?

h/static/scripts/widget-controller.js
robertknight: This is me thinking of Doxygen where `@p` is a way of referencing a parameter in the description of a method. I'll look up the correct way to do this in JSDoc.

robertknight: There isn't a formal one that I could find. Angular's docs use markdown backticks to indicate when a word refers to a parameter name. That seems like a reasonable approach.

h/static/scripts/widget-controller.js
```diff
@@ -12,78 +26,165 @@ module.exports = function WidgetController(
...
+    });
+    searchClient.on('end', function () {
+      // Remove client from list of active search clients
+      searchClients.splice(searchClients.indexOf(searchClient), 1);
+    });
+    searchClient.get({uri: uri, group: group});
+  }
+
+  /**
+   * Load annotations for all URLs associated with @p frames.
```
nickstenning: Suggest replace `@p` with backticks again here.

nickstenning:
LGTM.

PR #3109, opened 2016-03-18
Architecture decision records!

nickstenning:
This PR adds a folder in which to store [Architecture Decision Records](http://thinkrelevance.com/blog/2011/11/15/documenting-architecture-decisions) for the Hypothesis software.

It also includes a first ADR, which we've already discussed at length and accepted, describing the motivation for moving annotation data into PostgreSQL.

/cc @lenazun

### DIFF
--- a/docs/arch/adr-001.rst
+++ b/docs/arch/adr-001.rst
@@ -1,0 +1,80 @@
+ADR 1: PostgreSQL persistence for annotations
+=============================================
+
+Context
+-------
+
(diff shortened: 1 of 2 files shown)


robertknight:
LGTM. I was initially going to complain that it would be nice to have the docs in a format that GitHub will render, but it turns out that it supports a [bunch of formats](https://github.com/github/markup) besides markdown, including rst.

PR #3110, opened 2016-03-18
Refactor feature flag to access through Client + caching

chdorner:
Which allows us to keep the same usage
`request.feature('new-feature-name')`, but the database access is much
more optimised.

As soon as `request.feature` is accessed, we go to the database and read
the state of all feature flags into memory, Pyramid then memoizes the
client and subsequent calls to check if a given feature flag is enabled
or not are going directly to the internal cache within the request
lifecycle.

### DIFF
--- a/h/test/features_test.py
+++ b/h/test/features_test.py
@@ -28,107 +28,90 @@ def features_pending_removal_override(request):
     request.addfinalizer(patcher.stop)
 
 
-def test_flag_enabled_raises_for_undocumented_feature():
-    request = DummyRequest()
+class TestClient(object):
+    def test_init_loads_features(self):
+        db.Session.add(features.Feature(name='notification'))
+        db.Session.flush()
 
-    with pytest.raises(features.UnknownFeatureError):
-        features.flag_enabled(request, 'wibble')
+        client = self.client()
+        assert client._cache.keys() == ['notification']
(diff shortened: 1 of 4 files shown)


h/test/features_test.py
```diff
@@ -28,107 +28,90 @@ def features_pending_removal_override(request):
...
+            client.enabled('abouttoberemoved')
+
+    def test_enabled_false_if_not_in_database(self, client):
+        assert client.enabled('notification') == False
+
+    def test_enabled_false_if_everyone_false(self, client):
+        client._cache['notification'] = features.Feature(everyone=False)
+        assert client.enabled('notification') == False
+
+    def test_enabled_true_if_everyoni_true(self, client):
```
robertknight: Typo - `everyoni`

robertknight:
I don't see any logic here to invalidate the cache if feature flags are changed via the admin dashboard. Am I missing something?

chdorner:
@robertknight each request will get its own client, so there's no need for updating the cache.
That means that we still go to the database for each request, but only once per request.

h/test/features_test.py
chdorner: Fixed.

nickstenning:
LGTM.

PR #3112, opened 2016-03-18
Cached features and workers

chdorner:
We need to manually reload the feature cache when handling new messages in the workers, since the Pyramid requests we use in the workers are long-lived. In fact, they live as long as the worker python process exists.

### DIFF
--- a/h/features.py
+++ b/h/features.py
@@ -131,8 +131,25 @@ class Client(object):
             raise UnknownFeatureError(
                 '{0} is not a valid feature name'.format(name))
 
-        feature = self._cache.get(name)
+        return self._cache.get(name, False)
 
+    def all(self):
+        """
+        Returns a dict mapping feature flag names to enabled states
+        for the user associated with a given request.
+        """
+        return self._cache
+
+    def reload(self):
+        """Reloads the internal cache"""
+        names = FEATURES.keys()
+        all_ = self.request.db.query(Feature).filter(
+            Feature.name.in_(names)).all()
+        features = {f.name: f for f in all_}
+
+        self._cache = {n: self._state(features.get(n, None)) for n in names}
+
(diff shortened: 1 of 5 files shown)


h/mailer.py
```diff
@@ -43,6 +43,8 @@ def worker(request):
     """
     def handle_message(_, message):
         """Receive a message from nsq and send it as an email."""
+        request.feature.reload()
+
```
chdorner: I did not verify this yet

h/nipsa/worker.py
```diff
@@ -57,6 +57,8 @@ def worker(request):
     """
     def handle_message(_, message):
         """Handle a message on the "nipsa_users_annotations" channel."""
+        request.feature.reload()
```
chdorner: I did not verify this yet

h/notification/worker.py
```diff
@@ -18,6 +18,8 @@ def run(request):
     round-robin fashion.
     """
     def handle_message(reader, message=None):
+        request.feature.reload()
```
chdorner: I did not verify this yet

h/features.py
```diff
@@ -123,8 +131,25 @@ def enabled(self, name):
...
+        return self._cache
+
+    def reload(self):
+        """Reloads the internal cache"""
+        names = FEATURES.keys()
+        all_ = self.request.db.query(Feature).filter(
+            Feature.name.in_(names)).all()
+        features = {f.name: f for f in all_}
+
+        self._cache = {n: self._state(features.get(n, None)) for n in names}
```
nickstenning: `features.get(n)` is the same as `features.get(n, None)` and more idiomatic.

chdorner:
Closing in favour of #3115

PR #3114, opened 2016-03-19
Fix list/quote toolbar command when selection is empty and cursor is at start of line

### LINKED ISSUE #3091: Incorrect behavior when clicking new list button with empty selection
### Steps to reproduce
1. Create a new annotation
2. In the annotation text input field, position the cursor at the start of a new line
3. Press any of the list buttons in the toolbar to create a new list

### Expected behaviour
Editor should insert the appropriate marker ('*', '1.' etc.) at the start of the line and position the cursor after it so I can start typing.

### Actual behaviour
Editor inserts the appropriate marker but leaves the cursor at the start of the line, so you have to press additional keys to get the cursor to the end of the line before starting to type.

robertknight:
This fixes the list/quote toolbar button so that if, prior to clicking the button, the editor is in this state:

```
Line one

^ (cursor pos)
```

Afterwards it ends up in this state:

```
Line one
> 
  ^ (cursor pos)
```

Ready for the user to start typing straight away.

Fixes #3091

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -22,6 +22,10 @@ Bug fixes
 - Fixed an issue where private group annotations were not loaded correctly
   after switching user accounts (#3083).
 
+- Fix cursor position after using editor toolbar buttons to create a new list
+  or quote, if the selection was previously empty and the cursor was positioned
+  at the start of the line.
+
 Miscellanea
 -----------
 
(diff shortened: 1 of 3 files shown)


nickstenning:
LGTM.

PR #3115, opened 2016-03-21
Refactor feature caching

chdorner:
Instead of loading all features from the database when the Client is
initialized, this will only load the features from the database when the
actual state of the flags is accessed through either `enabled` or `all`.

The reason for this refactoring is that we have long-lived requests in
the `streamer` and package, and for the workers. These requests live for
as long as the Python process does, which means that we have to manually
invalidate the cache when a new chunk of work is about to get handled.

With the old caching mechanism, this meant that the only way to
invalidate the cache is to load the state of the flags from the
database. This solution however adds more roundtrips to Postgres, since
most websocket work and worker jobs don't rely on features at the
moment, so there's no need to load them from the database. But to
prevent weird behaviour in the future, there is a need to invalidate the
cache.

This changes the caching mechanism to only load the state of the
flags from the database when the state of one particular feature is
accessed. Thus allowing us to have a `clear()` method that only resets
the internal caching variable to an empty dictionary.

The methods that rely on the cache being loaded will now first check if
the cache is already loaded, if not, they will call `load()` first, and
then access the flags through the fresh cache.

### DIFF
--- a/h/features.py
+++ b/h/features.py
@@ -101,15 +101,26 @@ class Feature(db.Base):
 
 
 class Client(object):
+    """
+    Determine if the named feature is enabled for the current request.
+    If the feature has no override in the database, it will default to
+    False. Features must be documented, and an UnknownFeatureError will be
+    thrown if an undocumented feature is interrogated.
+    """
+
     def __init__(self, request):
         self.request = request
-
-        all_ = request.db.query(Feature).filter(
-            Feature.name.in_(FEATURES.keys())).all()
-        self._cache = {f.name: f for f in all_}
+        self._cache = {}
 
     def __call__(self, name):
         return self.enabled(name)
+
+    def load(self):
+        """Loads the feature flag states into the internal cache."""
+        all_ = self._fetch_features()
+        features = {f.name: f for f in all_}
+        self._cache = {n: self._state(features.get(n))
+                       for n in FEATURES.keys()}
 
     def enabled(self, name):
         """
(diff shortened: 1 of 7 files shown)


h/features.py
```diff
@@ -118,13 +130,36 @@ def enabled(self, name):
...
         """
         if name not in FEATURES:
             raise UnknownFeatureError(
                 '{0} is not a valid feature name'.format(name))
 
-        feature = self._cache.get(name)
+        if not self._cache:
+            self.load()
+
+        return self._cache.get(name, False)
```
nickstenning: I think at this point this should be a `self._cache[name]`, no? If `name` is unknown we just raised an `UnknownFeatureError`, and `self.load` creates entries for everything in `FEATURES`.

h/features.py
```diff
@@ -118,13 +130,36 @@ def enabled(self, name):
...
+    def all(self):
+        """
+        Returns a dict mapping feature flag names to enabled states
+        for the user associated with a given request.
+
+        When the internal cache is empty, it will automatically load the
+        feature flags from the database first.
+        """
+        if not self._cache:
+            return self.load()
```
nickstenning: Oops! Stray `return` here means that on first load this returns `None`.

h/features.py
chdorner: Yes, `self._cache[name]` should be safe at that point.
Fixed.

h/features.py
chdorner: Of course. Fixed.

h/features.py
chdorner: Actually, a test started failing because of this. Investigating that now

nickstenning:
LGTM.

PR #3116, opened 2016-03-21
rearranging annotation icons

sheetaluk:
Change and rearrange annotation edit, reply, delete and share buttons.

Trello card:
https://trello.com/c/4ElKy8DX


h/static/styles/annotation.scss
```diff
@@ -98,19 +98,27 @@ $annotation-card-left-padding: 10px;
...
-    .annotation:hover & {
-      color: $link-color;
-    }
+  &:hover {
+    color: $dg-1;
   }
 }
 
+.annotation-action-btn__icon {
+  font-size: 13.4px;
```
robertknight: Why the sub-pixel precision on the font size?

h/static/styles/variables.scss
```diff
@@ -151,3 +151,15 @@ $score-width: 40px;
 $score-height: $score-width;
 $top-bar-height: 40px;
 $input-border-radius: 2px;
+
+// Variables for the new colour palette
```
robertknight: Should use American English spelling for 'color'.

One way we could iterate towards the new design would be to do a pass through the code which replaced all uses of the other grays with ones from this set.

h/templates/client/annotation.html
```diff
@@ -174,9 +192,12 @@
         </span>
       </span>
       <span class="share-dialog-wrapper" ng-if="vm.feature('direct_linking')">
-        <button class="small btn btn-clean"
+        <button class="small btn btn-clean annotation-action-btn"
```
robertknight: Can we get rid of the other CSS classes here and just add the minimal necessary set of formatting to `annotation-action-btn`?

h/static/styles/variables.scss
robertknight: (Changing the grays to use the new palette everywhere would _definitely_ be part of a separate card I should add. Needs discussion with @conordelahunty as well about how we split that up).

robertknight:
This is what I see with the current iteration of this PR:

264-new-card-actions

Couple of things to note:

1. Several of the icons have had their paths messed up by IcoMoon export. @conordelahunty - I solved this for the lock icon in #3093 by merging the two paths in the lock icon using Inkscape before adding the SVG to IcoMoon.
2. The margin at the bottom of the card [looks larger than in the mocks](https://trello-attachments.s3.amazonaws.com/56be0c93d1b130942a81b3b2/677x896/0cc39682b3f7782cd018108d159ee4ca/annotation_footer.png) for example.

robertknight:
LGTM. As a follow-up, could you please check the set of icons that we have in the IcoMoon font and remove any that are no longer needed.

PR #3117, opened 2016-03-21
Change icon size in the bucket bar

conordelahunty:
Changing the icon size from 18px to 16px to get rid of fuzzy icons.


robertknight:
LGTM :+1:

PR #3118, opened 2016-03-21
Use more Pythonic way for the argument default value

chdorner:
In response to @seanh's [comment](https://github.com/hypothesis/h/commit/587d546a2f3b175f7845b59e4a74d14d59179f80#commitcomment-16734605) and the mutable default argument values
trap in Python.


nickstenning:
:+1:

PR #3119, opened 2016-03-21
Replace '@p <ident>' with `ident` in JSDoc comments

robertknight:
Using '@p $ident' to refer to parameters in JSDoc comments is not legal
syntax. Use Markdown backticks instead, ala. Angular's docs.


nickstenning:
LGTM.

PR #3120, opened 2016-03-21
Remove user status bitfield

nickstenning:
Remove a bunch of unused legacy properties from the user model. In some cases (`subscriptions`) it's not even clear that they were ever meaningfully used.

Each of `email_confirmed`, `optout`, and `invited` appear nowhere else in the code. `subscriptions` is referenced (and set) by `h.notifications` but appears to serve no purpose.

This PR also adds a schema migration to drop the `status` column from the `user` table. `ALTER TABLE ... DROP COLUMN` is an `O(1)` operation so this is safe to run against production as soon as the code updates have gone all the way out to production.

### DIFF
--- a/h/migrations/versions/0d4755a0d88b_remove_status_column_from_user_table.py
+++ b/h/migrations/versions/0d4755a0d88b_remove_status_column_from_user_table.py
@@ -1,0 +1,24 @@
+"""Remove status column from user table
+
+Revision ID: 0d4755a0d88b
+Revises: 2494fea98d2d
+Create Date: 2016-03-21 20:07:07.002482
+
+"""
+
+# revision identifiers, used by Alembic.
+revision = '0d4755a0d88b'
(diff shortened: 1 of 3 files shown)


h/migrations/versions/0d4755a0d88b_remove_status_column_from_user_table.py
```diff
@@ -0,0 +1,24 @@
...
+# revision identifiers, used by Alembic.
+revision = '0d4755a0d88b'
+down_revision = '2494fea98d2d'
+
+from alembic import op
+import sqlalchemy as sa
+
+
+def upgrade():
+    # Dropping a column is O(1) so this is safe to run against a production
```
robertknight: Interesting. Apparently it doesn't actually remove the column - https://hpt.is/AVOfZ_5AH9ZO4OKSlWUH

robertknight:
LGTM

PR #3124, opened 2016-03-22
Add message when user logged out and directed to an annotation.

sheetaluk:
Added a message when a user is logged out and directed to an annotation.

Trello card:
https://trello.com/c/dR3qXaxx


h/static/styles/app.scss
```diff
@@ -102,3 +102,38 @@ body {
     top: 1em;
   }
 }
+
+.loggedout-message {
```
robertknight: I suggest we extract this block into its own .scss file and then `@include` it at the components section at the top.

h/templates/app.html.jinja2
```diff
@@ -52,6 +52,25 @@
       <main ng-view=""></main>
     </div>
 
+    <div class="loggedout-message" ng-if="isSidebar && auth.status === 'signed-out' && settings.annotations" ng-cloak>
+      This is a public annotation created with Hypothes.is
```
robertknight: Ah, I've realized that this might not be a public annotation. It could also be either a private annotation or a group annotation. As it currently stands, we'll only know that once the annotation finishes loading.

robertknight: Re-reading the description on the card, this message should only be shown if:

1. There is a selection
2. The user is not signed in AND
3. The annotation is available to the user

Together these imply that the annotation must exist and it must be public.
This can be implemented in a very similar way to the implementation of https://github.com/hypothesis/h/pull/3093 .

I suggest rebasing on master, then moving this to `viewer.html` and changing the `ng-if` condition to be similar to the way that the permission message is displayed.

nickstenning:
What's the status here? It's not clear to me whether you've addressed @robertknight's comments.

Also, could you possibly look into improving the commit messages here. "rebasing and stuff" isn't terribly helpful, and even the commit message on 3c0a728 is more of a description of "what it does" rather than "why it does it". See also my email from this morning to the `eng@` list.

h/static/styles/loggedout-message.scss
```diff
@@ -0,0 +1,36 @@
+.loggedout-message {
+  margin: 25px auto;
```
robertknight: Where do these margins come from? This also duplicates the margin for the `.loggedout-message-logo` class. If there are two pieces of styling that are the same for a reason, please make use of variables and mixins to indicate this. Have a look at `excerpt.scss` for examples of this. Note that in that file, the purpose of the `@at-root` at the top is to prevent the variables from being used in other files.

h/templates/client/viewer.html
```diff
@@ -40,5 +40,23 @@
       ng-repeat="child in threadRoot.children | orderBy : sort.predicate"
       ng-show="vm.shouldShow()">
   </li>
+  <li class="loggedout-message" ng-if="isSidebar && auth.status === 'signed-out' && settings.annotations && selectedAnnotationAvailable()" ng-cloak>
```
robertknight: I would suggest extracting the `ng-if` condition out of the template and into a separate function in the controller, where you can add comments to explain what is going on and tests that the result is correct. Otherwise we're going to forget what these conditions are for in future.

robertknight:
Further to @nickstenning 's question from last time, my comments from the previous round of review have been addressed but I had a couple more minor notes.

robertknight:
This does not show the message at the moment since the `settings.annotation` expression in the template is invalid and unfortunately, Angular just silently evaluates invalid expressions as falsey - which is a good reason to anything but trivial expressions in Angular templates.

See https://github.com/hypothesis/h/commit/2e6162e115ae1fff6825ddf1928063ab1c3200d0 for one possible way to fix this, although note that the tests will also need updating.

h/static/scripts/test/widget-controller-test.js
```diff
@@ -100,6 +101,10 @@ describe('WidgetController', function () {
       focus: sinon.stub(),
     };
 
+    fakeSettings = {
+      annotations: true
```
robertknight: `settings.annotations` is actually an optional string (the annotation ID), rather than a boolean. For the purpose of the tests this works the same way, but it is confusing to use a different data type in tests than what the actual code works with.

Conceptually:

```ts
interface Settings {
  annotations?: string;
}
```

PR #3125, opened 2016-03-22
Support filtering test suites to run using '--grep' argument to gulp

robertknight:
Add support for filtering the tests that are run by Gulp test tasks
using '--grep <pattern>'.

This avoids the need to use the `(describe|it).only` test modifiers to
filter the tests that are run.

The pattern argument is a JS regex expression which is passed directly
to mocha as the 'grep' option.

eg. To start the app test suite in auto-restart mode, but only run tests
for the markdown editor, run:

  gulp test-watch-app --grep markdown

### DIFF
--- a/docs/hacking/install.rst
+++ b/docs/hacking/install.rst
@@ -257,9 +257,12 @@ source code. To start the test runner in auto-watch mode, run:
 
     gulp test-watch-app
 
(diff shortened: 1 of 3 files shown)


nickstenning:
LGTM.

PR #3126, opened 2016-03-22
Update ISSUE_TEMPLATE.md

oliversauter:
Added question, if people checked, if the bug is already in the issue tracker.

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -6,6 +6,12 @@ instead.
 To make it easy for us to help you please fill out the information requested
 below.
 -->
+
+### I checked, if the bug is already listed
+
+- [ ] yes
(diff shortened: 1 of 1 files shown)


nickstenning:
I applaud the intent, but perhaps this is a little *too* passive-aggressive :stuck_out_tongue_winking_eye:?

I think just adding a sentence to the comment at the top which reminds people to check before submitting an issue would be great.

oliversauter:
Hehe, possibly. I am no native speaker. So sometimes I don't hit the right tone ;)
Thought about it to be rather a lil bit funny :smile:

But yeah sure, makes sense. Will change it and make another pull request.

robertknight:
@oliversauter - You don't need to make a new pull request, you can just replace the branch that was used in this PR using `git push --force` (though do have a read up on what that does first before using it, its not the default behavior for good reason).

PR #3129, opened 2016-03-23
Add Visual Studio Code project config file for JS

robertknight:
In combination with JSDoc annotations, this enables IntelliSense,
project navigation, inline documentation, refactoring and other goodies
using Visual Studio Code and other IDEs that integrate with TypeScript's
Language Service for JavaScript (aka.  'Salsa').

See https://code.visualstudio.com/docs/languages/javascript

The project file can be named either `jsconfig.json` without 'allowJs'
or `tsconfig.json` with the 'allowJs' config option enabled. I've opted
for `tsconfig.json` because that turns up better documentation on the
web and is recognized by more tools.

The project config file lives in h/static/scripts so that Visual Studio
only tries to parse and process .js files under that directory.  The
config file does support specifying the set of files to include, but
each file currently has to be listed individually. Glob support is
planned for the future. See
https://github.com/Microsoft/TypeScript/issues/1927

### DIFF
--- a/h/static/scripts/tsconfig.json
+++ b/h/static/scripts/tsconfig.json
@@ -1,0 +1,7 @@
+{
+  "compilerOptions": {
+    "target": "ES5",
+    "module": "commonjs",
+    "allowJs": true
+  }
+}


nickstenning:
I'm generally reluctant to merge support for specific editors/environments, as it just clutters up the root of the repository.

Editorconfig seems like a fair concession, as it's generic, but this --- I'm not so sure.

Thoughts?

nickstenning:
If you're adding it because it shows up in `git status`, then don't forget you can add your own custom ignores to the repository in `.git/info/exclude`.

robertknight:
@nickstenning - I think I the PR/commit title is probably misleading. This is about enabling the use of refactoring, code navigation etc. with IDEs and other tools that can leverage TypeScript's language services. I mentioned VS Code because that's the best tool for said job right now, though Atom also has a pretty capable add-in.

nickstenning:
Hmmm. Ok. I'm reminded, though, of why I cannot stand the use of JSON for config files. It would be really nice if we could include a comment at the top of this file explaining its purpose...

robertknight:
@nickstenning - Ah,you're in luck. tsconfig.json does actually support comments, despite being a JSON file: https://github.com/Microsoft/TypeScript/pull/5450

PR #3130, opened 2016-03-23
Add message to private and group annotations sharer.

sheetaluk:
1. Show nothing if it is a public annotation
2. Show "Only me... " if it is a private annotation
3. Show "Only me..." if it is a private annotation AND also a group annotation
3. Show "Group..." if it is a group annotation.

Trello:
https://trello.com/c/4cOUJ4gv

### DIFF
--- a/h/static/styles/annotation.scss
+++ b/h/static/styles/annotation.scss
@@ -180,7 +180,6 @@ $annotation-card-left-padding: 10px;
   background: $white;
   border: 1px solid $gray-lighter;
   border-radius: 2px;
-  white-space: nowrap;
   width: 200px;
   font-size: $body1-font-size;
   cursor: default;
@@ -222,6 +221,17 @@ $annotation-card-left-padding: 10px;
   width: 220px;
   border: none;
   color: $text-color;
+}
+
+.share-dialog-msg {
+  color: $gray-light;
+  margin: -5px 10px 10px 10px;
+  line-height: 15px;
+  font-size: 11px;
+}
+
+.share-dialog-msg__audience {
+  font-style: italic;
 }
 
 .annotation-collapsed-replies {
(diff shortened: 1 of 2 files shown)


h/templates/client/annotation.html
```diff
@@ -198,6 +198,18 @@
               class="share-dialog-target__icon h-icon-mail"></a>
           </span>
           <input class="share-dialog-link" type="text" value="{{vm.annotationURI}}" readonly>
+          <span class="share-dialog-msg" ng-if="vm.group() && !vm.group().public">
+            <span class="share-dialog-msg__audience">
+              Group.
+            </span>
+            Only group member will be able to view this annotation.
```
robertknight: s/member/member_s_

h/static/styles/annotation.scss
```diff
@@ -216,6 +215,17 @@ $annotation-card-left-padding: 10px;
   color: $text-color;
 }
 
+.share-dialog-msg {
+  color: $gray-light;
+  margin: -5px 10px 10px 10px;
+  line-height: 15px;
```
robertknight: Are these using typography elements from the new palette? If so, now would be a good time to introduce those in `variables.scss`.

h/templates/client/annotation.html
```diff
@@ -221,6 +221,18 @@
               class="share-dialog-target__icon h-icon-mail"></a>
           </span>
           <input class="share-dialog-link" type="text" value="{{vm.annotationURI}}" readonly>
+          <span class="share-dialog-msg" ng-if="vm.group() && !vm.group().public">
```
robertknight: If an annotation is both in a group and private, both messages show up. If an annotation is private, we should only show the second (Only me) one.

robertknight:
LGTM. I would strongly advocate moving this out into its own component though as a follow-up PR so we can more easily write tests for this.

PR #3131, opened 2016-03-23
Convert AnnotationUI, AnnotationUISync from CoffeeScript to JS

robertknight:
No intended logic changes here, just a straight conversion.


nickstenning:
LGTM.

PR #3132, opened 2016-03-23
Use specific Sentry DSN for the client part of H

chdorner:
Which lets us use the `prod-client`/`stage-client` Sentry projects for both the extension and the client part of hosted H.


chdorner:
Would be good to have @nickstenning's and @robertknight's :eyes: on this

robertknight:
LGTM. We might want to be slightly more explicit about the fact that this is a public DSN, so there shouldn't be a password component in the URL. That could just be a comment above the new `EnvSetting()`

chdorner:
Good point @robertknight, I added the comment.

robertknight:
LGTM. Is the env var configured for the production service?

PR #3133, opened 2016-03-23
Preserve selection when switching accounts

robertknight:
Preserve selection when switching accounts

When visiting a direct-linked annotation (ie. #annotations:<id> is
present in the URL) we want to preserve the selection when the user
signs in, whether the annotation was private or in a group, in which
case they will need to sign in to see it, or public, in which case they
might need to sign in to reply.

Previously public direct-linked annotations became deselected when
signing in because they were unloaded when switching accounts, and this
fired the same ANNOTATION_DELETED event as when an annotation is
removed, which results in the annotation being removed from the
selection.

This commit enables the selection to be preserved by introducing a
different event within the app when annotations are unloaded vs.
deleted. When an annotation is unloaded, it is not removed from the
selection.

### DIFF
--- a/h/static/scripts/annotation-sync.coffee
+++ b/h/static/scripts/annotation-sync.coffee
@@ -130,6 +130,15 @@ module.exports = class AnnotationSync
       return unless bodies.length
       @bridge.call('loadAnnotations', bodies)
 
+    'annotationsUnloaded': (annotations) ->
+      self = this
+      annotations.forEach (annotation) ->
+        # In the client, unloading an annotation is handled the same way as
+        # deleting an annotation. Within the app however, we handle the events
+        # differently in some cases
+        delete self.cache[annotation.$$tag]
+        self._mkCallRemotelyAndParseResults('deleteAnnotation')(annotation)
+
   _syncCache: (channel) ->
     # Synchronise (here to there) the items in our cache
     annotations = (this._format a for t, a of @cache)
--- a/h/static/scripts/annotation-mapper.js
+++ b/h/static/scripts/annotation-mapper.js
@@ -1,5 +1,8 @@
 'use strict';
 
+var angular = require('angular');
+
+var events = require('./events');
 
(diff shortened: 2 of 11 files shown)


robertknight:

nickstenning:
Needs a rebase now.

robertknight:
Summary of a discussion in Slack, the issue that this PR originally purported to fix was already working in master. However, the related problem where the selection was lost when signing in to reply to a public annotation, _is_ fixed by this PR.

See https://hypothes-is.slack.com/archives/public/p1459336449001709

I have updated the commit message and PR description to explain the issue that is actually fixed here.

seanh:
This branch doesn't seem to fix this problem:

1. While logged out, follow a direct link to a public annotation
2. Log in from the sidebar
3. The annotation is no longer selected

seanh:
Ok, after rebuilding the assets, this seems to work well to me. The only way I could get the selection to be lost is:

1. Visit a direct link to a group annotation while signed in
2. Sign out
3. Selection is lost

h/static/scripts/annotation-sync.coffee
```diff
@@ -130,6 +130,15 @@ module.exports = class AnnotationSync
       return unless bodies.length
       @bridge.call('loadAnnotations', bodies)
 
+    'annotationsUnloaded': (annotations) ->
+      self = this
+      annotations.forEach (annotation) ->
+        # In the client, unloading an annotation is handled the same way as
+        # deleting an annotation. Within the app however, we handle the events
+        # differently in some cases
```
seanh: I'm not sure what the client and the app refer to in this comment?

robertknight: Good point. By 'client' here I mean "the annotator guest that lives in the page" and by 'app' I mean, "the app running in the sidebar". These are referred to as "guest" and "server" in some places. Consistent vocabulary would be useful though.

PR #3134, opened 2016-03-23
Fix CSP violation on new homepage

chdorner:
Fixing these CSP violations is quite a good task to dabble around with some frontend code. So here is the first one.

We are not allowing inline styles in our CSP policy, this commit adds a new class `svg-symbols` to the `svg` tag containing the symbols for the hiring banner arrow and the navigation arrows, and moves the `display: none` into the `header.scss` file.

Please let me know about wrong naming, wrong CSS files, etc.

The CSP violations that are already reported are listed here: https://report-uri.io/account/csp_reports/

### DIFF
--- a/h/static/styles/util.scss
+++ b/h/static/styles/util.scss
@@ -22,3 +22,7 @@
 .u-strong {
   font-weight: bold;
 }
+
+.u-hidden {
+  display: none;
+}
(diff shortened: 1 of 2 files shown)


robertknight:
The approach looks fine, although since `.svg-symbols` is just a generic helper for hiding the element I would suggest moving it to `util.scss` and renaming it to `.u-hidden`

chdorner:
Thanks @robertknight, moved it into the util file and renamed the class.

robertknight:
LGTM.

PR #3138, opened 2016-03-24
List group memberships of user in admin interface

chdorner:
This just shows a table of groups the user belongs to underneath the user data in the admin interface. The whole block is omitted when the user is not a member of any group.

2016-03-24-105911_560x390_scrot

### DIFF
--- a/h/groups/models.py
+++ b/h/groups/models.py
@@ -24,7 +24,7 @@ class Group(Base, mixins.Timestamps):
                       default=pubid.generate,
                       unique=True,
                       nullable=False)
-    name = sa.Column(sa.UnicodeText(), nullable=False)
+    name = sa.Column(sa.UnicodeText(), nullable=False, index=True)
 
     # We store information about who created the group -- we don't use this
     # currently, but it seems careless to lose this information when in the
--- a/h/groups/models.py
+++ b/h/groups/models.py
@@ -34,9 +34,9 @@ class Group(Base, mixins.Timestamps):
     creator = sa.orm.relationship('User')
 
     # Group membership
-    members = sa.orm.relationship('User',
-                                  secondary='user_group',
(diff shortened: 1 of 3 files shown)


chdorner:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;jXlzp8Iz&#x2F;301-in-the-admin-interface-add-groups-this-user-belongs-to-to-the-user-data"> In the admin interface, add &quot;groups this user belongs to&quot; to the user data</a>

robertknight:
If the list of groups could be sorted by name (case-insensitive) that would be useful when the user is a member of many groups.

Also worth adding a note to the CHANGES file.

chdorner:
@robertknight added the ordering. Will add the `CHANGES` on master after the merge, that file is always a source of confusion when PR's get merged.

h/groups/models.py
```diff
@@ -34,9 +34,9 @@ class Group(Base, mixins.Timestamps):
     creator = sa.orm.relationship('User')
 
     # Group membership
-    members = sa.orm.relationship('User',
-                                  secondary='user_group',
-                                  backref='groups')
+    members = sa.orm.relationship(
+        'User', secondary='user_group', backref=sa.orm.backref(
+            'groups', order_by=sa.func.lower(name)))
```
nickstenning: using a reference to `name` here seems a bit magical to me. I think `'Group.name'` would probably be a bit clearer.

nickstenning: Although perhaps that doesn't work because you're lowercasing things first?

chdorner: that's what I wanted to do first, but I'm getting this: `NameError: name 'Group' is not defined`

robertknight:
LGTM

PR #3139, opened 2016-03-24
Preserve URL fragments when redirecting to PDF.js in extensions

robertknight:
Direct linking to annotations in PDFs was broken on systems with the
Chrome extension installed because the '#annotations' fragment was lost
when the Chrome extension redirected from the original URL to the
bundled PDF.js viewer.

This is fixed by extracting and URI-encoding the non-fragment parts of
the URL when generating the 'file' query string param for the PDF.js
viewer and then appending the original fragment to the resulting URL,
which becomes:

  chrome-extension://<ID>/content/viewer.html?file=<...>#annotations:ID

When the embedded version of H in the bundled PDF.js viewer loads, it
will then pick up the fragment and scroll to and focus on the
annotation.

### DIFF
--- a/h/browser/chrome/lib/sidebar-injector.js
+++ b/h/browser/chrome/lib/sidebar-injector.js
@@ -1,5 +1,7 @@
 'use strict';
 
+var queryString = require('query-string');
+
 var detectContentType = require('./detect-content-type');
 var errors = require('./errors');
 var util = require('./util');
@@ -50,6 +52,8 @@ function SidebarInjector(chromeTabs, dependencies) {
 
   var executeScriptFn = util.promisify(chromeTabs.executeScript);
 
+  var PDFViewerBaseURL = extensionURL('/content/web/viewer.html');
+
   if (typeof extensionURL !== 'function') {
     throw new TypeError('extensionURL must be a function');
   }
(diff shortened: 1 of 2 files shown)


robertknight:
Realized that when de-activating the extension, we _don't_ want to preserve the fragment if it was a '#annotations' fragment, since that will just result in the extension being re-activated.

robertknight:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;rhbTLLY1&#x2F;292-bug-fix-direct-links-on-pdfs"> BUG: Fix direct links on PDFs</a>

nickstenning:
Ok, this is clearly an improvement, but direct links to PDFs still seem not to work even with this change. Is that correct?

For example, the (public) annotation referenced in the linked Trello card is apparently not found -- the "you don't have permission to view this annotation" box is shown instead.

robertknight:
@nickstenning - The annotation on the GitHub issue does work, but you need to build the dev extension to point at the production Hypothes.is service, oh and also have the stage extension installed (then reload a couple of times if the stage extension wins the activation race - this isn't very tester friendly at the moment).

seanh:
I was able to test this by:

1. Running h locally
2. Running bouncer locally
3. Serving a PDF file at localhost (using `python -m SimpleHTTPServer`)
4. Open the localhost PDF file in the browser, activate h, make an annotation, get and follow the direct link to the annotation

As expected it works on this branch but not on master (have to rememeber to re-build assets and Chrome extension after changing branches).

Noticed the issue with the "You do not have permission" card appearing momentarily, not sure if that's to do with this PR.

untitled screencast 10

seanh:
The "You do not have permission" card appearing momentarily seems to happen on PDFs but not on HTML files

seanh:
Code looks good to me. I think the "You do not have permission" card appearing momentarily is a race between HTTP requests and not to do with this PR or with PDFs

PR #3141, opened 2016-03-24
CSP Violation: form input onfocus

chdorner:
I'm not sure if this is the correct way to handle this, so I've held off writing tests.

One thing I noticed is that this doesn't work right after a page load, presumably because the `focus` event is fired before all the JS code is loaded (as described [here](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#attr-autofocus)).

### DIFF
--- a/h/static/scripts/form-select-onfocus-controller.js
+++ b/h/static/scripts/form-select-onfocus-controller.js
@@ -1,0 +1,20 @@
+'use strict';
+
+function FormSelectOnFocusController(root) {
+  this._elements = root.querySelectorAll('.js-select-onfocus');
+
+  for (var i = 0; i < this._elements.length; i++) {
+    var element = this._elements[i];
+
+    // In case the `focus` event has already been fired, select the element
+    if (element === document.activeElement) {
+      element.select();
+    }
+
+    element.addEventListener('focus', function(event) {
+      event.target.select();
+    });
+  }
+}
+
+module.exports = FormSelectOnFocusController;
--- a/h/static/scripts/site.js
+++ b/h/static/scripts/site.js
@@ -1,3 +1,5 @@
+'use strict';
+
 // configure error reporting
 var settings = require('./settings')(document);
 if (settings.raven) {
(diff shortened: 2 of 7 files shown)


chdorner:
Any thoughts on this would be appreciated, this is all a bit out of my depth at the moment.

h/static/scripts/form-select-onfocus-controller.js
```diff
@@ -0,0 +1,17 @@
+function FormSelectOnFocusController(element) {
+  var self = this;
+
+  this._formInputs = element.querySelectorAll('form input.js-select-onfocus');
```
robertknight: Selectors should be minimally specific, so just `.js-select-onfocus` would be fine here.

h/static/scripts/site.js
```diff
@@ -24,6 +25,10 @@ page('/groups/new', function () {
   new CreateGroupFormController(document.body);
 });
 
+page('/login', function() {
+  new FormSelectOnFocusController(document.body);
```
robertknight: This issue and the other one just means there should be a `'use strict';` at the top of the file.

robertknight:
> One thing I noticed is that this doesn't work right after a page load,
> presumably because the focus event is fired before all the JS code is loaded

That might indeed be the case. In which case you can use `document.activeElement` to test whether the element has focus at the point when the JS code loads and select it if so.

h/static/scripts/form-select-onfocus-controller.js
```diff
@@ -0,0 +1,22 @@
...
+
+function FormSelectOnFocusController(element) {
+  this._elements = element.querySelectorAll('.js-select-onfocus');
+
+  this._onfocusHandler = function(event) {
+    event.target.select();
+  };
+
+  for (var i = 0; i < this._elements.length; i++) {
+    var element = this._elements[i];
```
chdorner: This is fixed now.

h/static/scripts/site.js
chdorner: This is fixed now.

chdorner:
Thanks @robertknight. I've fixed these issues, and added tests. This is now ready for a final review.

robertknight:
LGTM

PR #3142, opened 2016-03-24
CSP violation: admin confirm users delete

chdorner:
This pull request:

* Sets up sentry for the admin JS part
* Moves the confirm dialog when deleting users into it's own JS controller

I would need some help on how to properly test this, as this is all a bit out of my depth at the moment. Any other comments appreciated as well.

### DIFF
--- a/h/static/scripts/admin-users.js
+++ b/h/static/scripts/admin-users.js
@@ -1,0 +1,19 @@
+'use strict';
+
+function AdminUsersController(element, window_) {
+  this._form = element.querySelector('#js-users-delete-form');
+
+  function confirmFormSubmit() {
+    return window_.confirm('This will permanently delete all the user\'s data. Are you sure?');
+  }
+
+  if (this._form) {
+    this._form.addEventListener('submit', function (event) {
+      if (!confirmFormSubmit()) {
+        event.preventDefault();
+      }
+    });
+  }
+}
+
+module.exports = AdminUsersController;
--- a/h/static/scripts/test/admin-users-test.js
+++ b/h/static/scripts/test/admin-users-test.js
@@ -1,0 +1,47 @@
+'use strict';
+
+var AdminUsersController = require('../admin-users');
+
+// helper to dispatch a native event to an element
+function sendEvent(element, eventType) {
+  // createEvent() used instead of Event constructor
+  // for PhantomJS compatibility
+  var event = document.createEvent('Event');
+  event.initEvent(eventType, true /* bubbles */, true /* cancelable */);
+  element.dispatchEvent(event);
+
+  return event;
+}
+
+describe('AdminUsersController', function () {
+  var root;
+  var form;
+
+  beforeEach(function () {
+    root = document.createElement('div');
+    root.innerHTML = '<form id="js-users-delete-form">' +
+                     '<input type="submit" id="submit-btn">';
+    form = root.querySelector('form');
+    document.body.appendChild(root);
+  });
+
+  afterEach(function () {
+    root.parentNode.removeChild(root);
+  });
+
+  it('it submits the form when confirm returns true', function () {
+    var fakeWindow = {confirm: sinon.stub().returns(true)};
+    new AdminUsersController(root, fakeWindow);
+
+    var event = sendEvent(form, 'submit');
+    assert.isFalse(event.defaultPrevented);
+  });
+
+  it('it cancels the form submission when confirm returns false', function () {
+    var fakeWindow = {confirm: sinon.stub().returns(false)};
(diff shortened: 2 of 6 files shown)


h/static/scripts/admin-users.js
```diff
@@ -0,0 +1,15 @@
+'use strict';
+
+function AdminUsersController(element) {
+  this._form = element.querySelector('#js-users-delete-form');
+
+  function confirmFormSubmit() {
+    return window.confirm("This will permanently delete all the user's data. Are you sure?");
+  }
+
+  if(this._form !== null) {
```
robertknight: In JS code we just use truthiness tests rather than testing for exactly null vs. undefined etc., so just `if (this._form)`.

h/static/scripts/test/admin-users-test.js
```diff
@@ -0,0 +1,29 @@
...
+  var template;
+
+  before(function () {
+    template = '<form id="js-users-delete-form">' +
+               '<input type="submit">';
+  });
+
+  beforeEach(function () {
+    element = document.createElement('div');
+    element.innerHTML = template;
```
robertknight: If you want to interact with an element in tests, you need to add it to the page, eg. using `document.body.appendChild()` and then remove it afterwards using eg:

```js
afterEach(function () {
   element.parentNode.removeChild(element);
})
```

h/static/scripts/test/admin-users-test.js
```diff
@@ -0,0 +1,29 @@
...
+  });
+
+  it('it should confirm the delete user form submission', function () {
+    // click on submit button
```
robertknight: You can use `element.querySelector()` to get a reference to the button and then just call its `click()` method.

h/static/scripts/test/admin-users-test.js
```diff
@@ -0,0 +1,29 @@
...
+    // assert that confirm is being called
```
robertknight: There are two basic ways to stub functions like `window.confirm`:

Instead of relying on the `window` global in the implementation, inject it. Then in tests you can inject a stub, eg:

```
function AdminUsersController(element, window_) {
...
window_.confirm(...))
}
```

Then in the tests:

```js
var fakeWindow = {confirm: sinon.stub().returns(true)};
var controller = new AdminUsersController(element, fakeWindow)
...
assert.called(fakeWindow.confirm)
```

The other option is to rely on Sinon's ability to stub globals using `sinon.stub(object, function name)` in `beforeEach()` and then remove the stub in `afterEach()` using `object.functionName.restore();`

I generally prefer the first approach as it is a little more straightforward.

chdorner:
All of the comments have been addressed, I also changed the way it is cancelling the form submission by calling `event.preventDefault()`.

Ready for another review @robertknight, but as always with these CSP violations, no rush! :)

h/static/scripts/test/admin-users-test.js
```diff
@@ -0,0 +1,51 @@
...
+  afterEach(function() {
+    root.parentNode.removeChild(root);
+  });
+
+  it('it submits the form when confirm returns true', function() {
+    var fakeWindow = {confirm: sinon.stub().returns(true)};
+    new AdminUsersController(root, fakeWindow);
+
+    var event = initEvent('submit');
+    var spy = sinon.spy(event, 'preventDefault');
```
robertknight: Rather than spying on the `preventDefault` method, you can use `event.defaultPrevented` - https://developer.mozilla.org/en-US/docs/Web/API/Event/defaultPrevented , assuming that PhantomJS 1.x supports it. If not, this is fine.

h/static/scripts/admin-users.js
```diff
@@ -0,0 +1,19 @@
+'use strict';
+
+function AdminUsersController(element, window_) {
+  this._form = element.querySelector('#js-users-delete-form');
+
+  function confirmFormSubmit() {
+    return window_.confirm("This will permanently delete all the user's data. Are you sure?");
+  }
+
+  if(this._form) {
```
robertknight: Nit, should have a space between the `if` and `(`. You can use `jscs <filename>` to check the coding style for a file and `jscs --fix <filename>` to fix up trivial issues automatically.

robertknight:
Couple of trivial issues, but otherwise this looks good.

h/static/scripts/test/admin-users-test.js
chdorner: This is much nicer, thanks for that!

h/static/scripts/admin-users.js
chdorner: That's the command I was looking for, pushing and then waiting for Hound is not the best workflow :)

PR #3144, opened 2016-03-26
Show adder if user makes a selection and then launches the plugin.

sheetaluk:
If a user selects some text, and then realises that Hypothesis isn't
active, their first reaction will probably be to activate Hypothesis.
In this scenario, we should pop up the adder near the selection on launch.

* If event type is 'ready' and there is a selection, then show adder
in the top middle of the bounding box of the selection.
* Otherwise, show the adder in the mouse position.

Trello:
https://trello.com/c/1NElt1Oc/297-show-annotation-adder-on-text-selected-before-launching-the-tool


robertknight:
I'm afraid I won't get a chance to review this properly today, but from a quick glance one thing I'll note is that we'll need tests for this.

PR #3146, opened 2016-03-28
Show list of group members in the group page.

sheetaluk:
Users want to see other users in the groups they belong to
and annotations each of these users have made and shared with a
group. This will help them decide how to share annotations with
their groups.

https://trello.com/c/rSUQLuos/294-as-a-user-i-want-to-be-able-to-find-out-who-is-a-member-of-a-group-i-m-in

### DIFF
--- a/h/groups/views.py
+++ b/h/groups/views.py
@@ -125,7 +125,10 @@ def _read_group(request, group):
                       for annotation in annotations_by_uri.values()]
 
     template_data = {
-        'group': group, 'group_url': url, 'document_links': document_links}
+        'group': group,
+        'group_url': url,
+        'document_links': document_links
+    }
 
     return renderers.render_to_response(
         renderer_name='h:templates/groups/share.html.jinja2',
--- a/h/templates/groups/share.html.jinja2
+++ b/h/templates/groups/share.html.jinja2
@@ -32,17 +32,27 @@
         </ul>
     </div>
     {% endif %}
+    {% if group.members %}
+    <div class="group-members-list">
+      <ul class="group-members-list__list">
+        <p class="group-members-list__heading">Group Members:</p>
+        {% for member in group.members|sort(attribute='username')  %}
+          <li>
+            <a href="{{ request.route_url('stream') }}?q=group:{{ group.pubid }} user:{{ member.username }}"
+              target="_blank"
+              title="{{ member.username }}'s annotations in this group">
+              {{ member.username }}
+            </a>
+          </li>
+        {% endfor %}
+      </ul>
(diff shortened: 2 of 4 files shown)


h/groups/views.py
```diff
@@ -124,8 +124,15 @@ def _read_group(request, group):
...
 
+    member_usernames = [member.username
+                        for member in group.members]
+
     template_data = {
-        'group': group, 'group_url': url, 'document_links': document_links}
+        'group': group,
+        'group_url': url,
+        'document_links': document_links,
+        'member_usernames': sorted(member_usernames)
```
chdorner: This will be fine for the amount of users in a typical group. Not sure if it makes sense to optimize potential slowness when a group has lots of users by letting the database sort the group members.

I did the same thing recently for the inverse relationship (groups a user is member of) [here](https://github.com/hypothesis/h/blob/master/h/groups/models.py#L37-L39).

h/templates/groups/share.html.jinja2
```diff
@@ -32,17 +32,27 @@
         </ul>
     </div>
     {% endif %}
+    {% if member_usernames %}
+    <div class="group-members-list">
+      <ul class="group-members-list__list">
+        <p class="group-members-list__heading">Group Members:</p>
+        {% for member_username in member_usernames %}
+          <li>
+            <a href="{{ request.route_url('stream') }}?q=group:{{ group.pubid }} username: {{ member_username }}"
```
chdorner: The issue here is that the username you want to display is `chdorner` for example, but the username you want so search for is `chdorner@hypothes.is`.

You can call [this](https://github.com/hypothesis/h/blob/master/h/util/user.py#L24-L32) function for turning a username into a user id.

chdorner: Oh, and it should be `user:` (instead of `username:`), and without a space between the key and the value.

h/groups/views.py
nickstenning: I'm pretty sure you don't even need this change, do you? Just pass in group and then in the templates iterate over the users as:

```jinja2
{% for user in group.members|sort(attribute='username') %}
...
{% endfor %}
```

h/templates/groups/share.html.jinja2
nickstenning: A search for `user:chdorner` on the stream will work just fine.

chdorner: I would still do add the `@hypothes.is` part to keep it consistent though, no?

chdorner:
LGTM

PR #3151, opened 2016-03-29
Postgres Write 2/4: Model code for creating new annotations in Postgres

seanh:
This pull request adds the new model code that the `storage` module will use to create and update `DocumentMeta` and `DocumentURI` objects in the db (see <https://github.com/hypothesis/h/pull/3153>), specifically it adds two factory functions `create_or_update_document_meta()` and `create_or_update_document_uri()`.

The logic "create an object in the db with these params, or update an equivalent one if one already exists" seems like it belongs in the model to me, so `create_or_update_document_meta()` and `create_or_update_document_uri()` are added to `models` to do this.

`merge_documents()` which does similar work was already in `models`, is will be called by `storage` as well.

These are adapted from existing code in `migrate.py`. `storage` calls them.

### DIFF
--- a/h/api/models/document.py
+++ b/h/api/models/document.py
@@ -260,7 +260,6 @@ def create_or_update_document_uri(session,
 
 def create_or_update_document_meta(session,
                                    claimant,
-                                   claimant_normalized,
                                    type,
                                    value,
                                    document,
@@ -286,10 +285,6 @@ def create_or_update_document_meta(session,
         if a new DocumentMeta is created
     :type claimant: unicode
 
-    :param claimant_normalized: the value of the new or existing DocumentMeta's
-        claimant_normalized attribute
-    :type claimant_normalized: unicode
-
     :param type: the value of the new or existing DocumentMeta's type attribute
     :type type: unicode
 
@@ -311,19 +306,18 @@ def create_or_update_document_meta(session,
 
     """
     existing_dm = DocumentMeta.query.filter(
-        DocumentMeta.claimant_normalized == claimant_normalized,
+        DocumentMeta.claimant_normalized == uri_normalize(claimant),
         DocumentMeta.type == type).one_or_none()
 
     if existing_dm is None:
         session.add(DocumentMeta(
-                    _claimant=claimant,
-                    _claimant_normalized=claimant_normalized,
-                   type=type,
-                   value=value,
-                   document=document,
-                   created=created,
-                   updated=updated,
-        ))
+                    claimant=claimant,
+                    type=type,
+                    value=value,
+                    document=document,
+                    created=created,
+                    updated=updated,
+                    ))
     else:
         existing_dm.value = value
         existing_dm.updated = updated
--- a/h/api/models/document.py
+++ b/h/api/models/document.py
@@ -322,9 +322,9 @@ def create_or_update_document_meta(session,
         existing_dm.value = value
         existing_dm.updated = updated
         if not existing_dm.document == document:
-            log.warn('Found DocumentMeta with id %d does not match expected '
-                     'document with id %d', existing_dm.id,
-                     document.id)
+            log.warn("Found DocumentMeta (id: %d)'s document_id (%d) doesn't "
+                     "match given Document's id (%d)",
+                     existing_dm.id, existing_dm.document_id, document.id)
 
 
 def merge_documents(session, documents, updated=datetime.now()):
--- a/h/api/models/document.py
+++ b/h/api/models/document.py
@@ -251,8 +251,9 @@ def create_or_update_document_uri(session,
                              updated=updated)
         session.add(docuri)
     elif not docuri.document == document:
-        log.warn('Found DocumentURI with id %d does not match expected '
-                 'document with id %d', docuri.document_id, document.id)
+        log.warn("Found DocumentURI (id: %d)'s document_id (%d) doesn't match "
+                 "given Document's id (%d)",
+                 docuri.id, docuri.document_id, document.id)
 
     docuri.updated = updated
 
--- a/h/api/models/document.py
+++ b/h/api/models/document.py
@@ -289,7 +289,7 @@ def create_or_update_document_meta(session,
 
     :param value: the value to set the new or existing DocumentMeta's value
         attribute to
-    :type value: unicode
+    :type value: list of unicode strings
 
     :param document: the value to use for the DocumentMeta's document if a new
         DocumentMeta is created
--- a/h/api/models/document.py
+++ b/h/api/models/document.py
@@ -8,7 +8,6 @@ import logging
 import sqlalchemy as sa
 from sqlalchemy.dialects import postgresql as pg
 from sqlalchemy.ext.hybrid import hybrid_property
-from sqlalchemy.orm import aliased
 
 from h.api.db import Base
 from h.api.db import mixins
--- a/h/api/models/document.py
+++ b/h/api/models/document.py
@@ -234,7 +234,7 @@ def create_or_update_document_uri(session,
     :type updated: datetime.datetime
 
     """
-    docuri = DocumentURI.query.filter(
+    docuri = session.query(DocumentURI).filter(
         DocumentURI.claimant_normalized == uri_normalize(claimant),
         DocumentURI.uri_normalized == uri_normalize(uri),
         DocumentURI.type == type,
@@ -304,7 +304,7 @@ def create_or_update_document_meta(session,
     :type updated: datetime.datetime
 
     """
-    existing_dm = DocumentMeta.query.filter(
+    existing_dm = session.query(DocumentMeta).filter(
         DocumentMeta.claimant_normalized == uri_normalize(claimant),
         DocumentMeta.type == type).one_or_none()
 
--- a/h/api/models/document.py
+++ b/h/api/models/document.py
@@ -185,7 +185,7 @@ class DocumentMeta(Base, mixins.Timestamps):
         return '<DocumentMeta %s>' % self.id
 
 
-def create_or_update_document_uri(db,
+def create_or_update_document_uri(session,
                                   claimant,
                                   uri,
                                   type,
@@ -207,8 +207,8 @@ def create_or_update_document_uri(db,
     belongs to may be different. The claimant and uri are normalized before
     comparing.
 
-    :param db: the database session
-    :type db: sqlalchemy.orm.session.Session
+    :param session: the database session
+    :type session: sqlalchemy.orm.session.Session
 
     :param claimant: the .claimant property of the DocumentURI
     :type claimant: unicode
@@ -249,7 +249,7 @@ def create_or_update_document_uri(db,
                              document=document,
                              created=created,
                              updated=updated)
-        db.add(docuri)
+        session.add(docuri)
     elif not docuri.document == document:
         log.warn('Found DocumentURI with id %d does not match expected '
                  'document with id %d', docuri.document_id, document.id)
@@ -257,7 +257,7 @@ def create_or_update_document_uri(db,
     docuri.updated = updated
 
 
-def create_or_update_document_meta(db,
+def create_or_update_document_meta(session,
                                    claimant,
                                    claimant_normalized,
                                    type,
@@ -278,8 +278,8 @@ def create_or_update_document_meta(db,
     claimant and type, but its value, document and created and updated times
     needn't match the given ones.
 
-    :param db: the database session
-    :type db: sqlalchemy.orm.session.Session
+    :param session: the database session
+    :type session: sqlalchemy.orm.session.Session
 
     :param claimant: the value to use for the DocumentMeta's claimant attribute
         if a new DocumentMeta is created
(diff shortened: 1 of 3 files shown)


seanh:

seanh:

h/api/models/document.py
```diff
@@ -40,7 +44,7 @@ def title(self):
     @classmethod
     def find_by_uris(cls, session, uris):
         """Find documents by a list of uris."""
-        query_uris = [text_type(uri.normalize(u), 'utf-8') for u in uris]
+        query_uris = [normalize(u) for u in uris]
```
chdorner: This might be the right time to look at making `uri.normalize` return a unicode string.

h/api/models/document.py
```diff
@@ -182,6 +186,153 @@ def __repr__(self):
         return '<DocumentMeta %s>' % self.id
 
 
+def create_or_update_document_uri(db,
```
chdorner: @nickstenning mentioned in one of my previous pull requests that when passing in the database connection, we should call the parameter `session`.
The comment is [here](https://github.com/hypothesis/h/pull/3011#discussion_r54537119)

h/api/models/document.py
```diff
@@ -182,6 +186,153 @@ def __repr__(self):
...
+                             uri=uri,
+                             type=type,
+                             content_type=content_type,
+                             document=document,
+                             created=created,
+                             updated=updated)
+        db.add(docuri)
+    elif not docuri.document == document:
+        log.warn('Found DocumentURI with id %d does not match expected '
+                 'document with id %d', docuri.document_id, document.id)
```
chdorner: Looks like there is a mismatch the interpolated string and the passed in variables (DocumentURI id gets replaced with `docuri.document_id`). In any case, I think we should add all three ids: `docuri.id`, `docuri.document_id`, `document.id`.

h/api/models/document.py
```diff
@@ -182,6 +186,153 @@ def __repr__(self):
...
+
+    docuri.updated = updated
+
+
+def create_or_update_document_meta(db,
```
chdorner: Same as mentioned above, this parameter should be called `session`.

h/api/models/document.py
```diff
@@ -182,6 +186,153 @@ def __repr__(self):
...
+
+    """
+    existing_dm = DocumentMeta.query.filter(
+        DocumentMeta.claimant_normalized == claimant_normalized,
+        DocumentMeta.type == type).one_or_none()
+
+    if existing_dm is None:
+        db.add(DocumentMeta(
+            _claimant=claimant,
+            _claimant_normalized=claimant_normalized,
```
chdorner: You can skip having the `claimant_normalized` being passed-in to this function, and use the `claimant` (non-underscore) setter which takes care of setting the normalized claimant. (Similar to what you already did for the `create_or_update_document_uri` function).

h/api/models/document.py
```diff
@@ -182,6 +186,153 @@ def __repr__(self):
...
+            created=created,
+            updated=updated,
+        ))
+    else:
+        existing_dm.value = value
+        existing_dm.updated = updated
+        if not existing_dm.document == document:
+            log.warn('Found DocumentMeta with id %d does not match expected '
+                     'document with id %d', existing_dm.id,
+                     document.id)
```
chdorner: Adding the `existing_dm.document_id` might be good as well, as this `DocumentMeta` instance can move to another document in the future, so the log message might be more helpful when we include all three ids.

h/api/models/test/document_test.py
```diff
@@ -119,6 +122,416 @@ def test_document_find_or_create_by_uris_no_results():
...
+        @property
+        def id(self):
+            pass
+        @property
+        def created(self):
+            pass
+        @property
+        def updated(self):
+            pass
+    return mock.Mock(spec=Document)
```
chdorner: Is there a reason why you couldn't just use the `h.api.models.document.Document` class as a spec for this mock? Then we don't have to make sure to synchronize the API with this class here.

h/api/models/test/document_test.py
```diff
@@ -119,6 +122,416 @@ def test_document_find_or_create_by_uris_no_results():
...
+        content_type=None,
+        document=mock_document(),
+        created=datetime.datetime.now(),
+        updated=datetime.datetime.now())
+
+    # FIXME: We need to assert that this is called with the right arguments,
+    # but that's very awkward to do with the way sqlalchemy querying works.'
+    # Move this functionality onto the model class itself where the tests can
+    # use the actual database.
+    assert DocumentURI.query.filter.call_count == 1
```
chdorner: Like you said in the comment, I think testing the querying by creating a document with a document uri in the database and then making sure that this one is being used is probably the better way to test this.
What's the reason for the need to move this onto the model class itself to be able to use the database in the tests?

h/api/models/test/document_test.py
```diff
@@ -119,6 +122,416 @@ def test_document_find_or_create_by_uris_no_results():
...
+            document=mock.Mock(),
+            created=yesterday(),
+            updated=now(),
+        )
+
+        # FIXME: We need to assert that this is called with the right
+        # arguments, but that's very awkward to do with the way sqlalchemy
+        # querying works.' Move this functionality onto the model class itself
+        # where the tests can use the actual database.
+        assert DocumentMeta.query.filter.call_count == 1
```
chdorner: Similar to what I mentioned above, why not just create a document with a document meta and then check that it doesn't initialize a new one?

h/api/models/test/document_test.py
```diff
@@ -119,6 +122,416 @@ def test_document_find_or_create_by_uris_no_results():
...
+        document=mock_document(),
+        created=now - datetime.timedelta(days=3),
+        updated=now)
+
+    assert existing_document_uri.updated == now
+
+
+@pytest.mark.usefixtures('DocumentMeta',
+                         'log')
+class TestCreateOrUpdateDocumentMeta(object):
```
chdorner: Is there a reason why the tests for `create_or_update_document_meta` are in a test class, but the tests for `create_or_update_document_uri` are not?
Would be nice if we could keep them consistent (I personally prefer the test class, but I'm fine if you want to stick to the other one).

h/api/models/test/document_test.py
```diff
@@ -119,6 +122,416 @@ def test_document_find_or_create_by_uris_no_results():
...
+            def __init__(self):
+                self.claimant_normalized = None
+                self.type = None
+                self.value = None
+                self.created = None
+                self.updated = None
+                self.document = document
+                self.id = None
+                self.document_id = None
+        return mock.Mock(spec=DocumentMeta())
```
chdorner: Similar question to above. Is there a reason why you couldn't use the real `DocumentMeta` class as the spec for the mock?

h/api/models/test/document_test.py
chdorner: Ignore this comment, I was reviewing commit by commit.

h/api/models/test/document_test.py
chdorner: Ignore this comment, I was reviewing commit by commit.

h/api/models/test/document_test.py
chdorner: Ignore this comment, I was reviewing commit by commit.

chdorner:
I was reviewing commit by commit (shouldn't have done that!), so Github marks one comment that still applies as out-of-date ([this one](https://github.com/hypothesis/h/pull/3151#discussion_r58017621)), you can safely ignore the rest, because those issues you already fixed in later commits.

h/api/models/document.py
seanh: Do you know what the deal is with `uri.normalize()`? It seems to deliberately return a UTF-8 string

h/api/models/document.py
seanh: Makes sense

seanh: Done (once I rebase and push it..)

h/api/models/document.py
seanh: Oh, stupid old style formatting!

seanh: Done

h/api/models/document.py
seanh: Done

h/api/models/document.py
chdorner: I'll have a go at it. It should be in its own PR anyway, since almost all of the 4 PG pull requests to convert the URLs to a unicode string again.

h/api/models/document.py
seanh: Ok, done

h/api/models/document.py
seanh: Done

h/api/models/test/document_test.py
seanh: Good point, I don't see why not

h/api/models/test/document_test.py
seanh: Added a comment to clarify this. We can't use the real `DocumentMeta` as a spec because in the test that class is patched with a mock.

seanh:
Yeah, sorry, the commit history of this one could have been a little cleaner, the problem is that I did the "use db in tests" commits after a couple of "tidy up the tests" commits with big diffs, which might have made reordering and squashing the commits difficult

seanh:
@chdorner I think that's everything responded to (apart from the `uri.normalize()` thing which I think we can do separately)

h/api/models/document.py
chdorner: #3163 got merged, so all this `text_type(uri.normalize(..))` stuff can go :+1:

seanh: Ok, rebased it out

h/api/models/document.py
```diff
@@ -3,15 +3,18 @@
...
 
 import sqlalchemy as sa
 from sqlalchemy.dialects import postgresql as pg
 from sqlalchemy.ext.hybrid import hybrid_property
-from sqlalchemy.orm import aliased
 
-from h.api import uri
 from h.api.db import Base
 from h.api.db import mixins
+from h.api.uri import normalize as uri_normalize
```
seanh: There's an annoying problem here. `create_or_update_document_uri()` has a parameter called `uri` which shadows the `uri` imported here (if we did `from h.api import uri`) and means that `create_or_update_document_uri()` can't use the `uri` module.

I would just rename the parameter to `uri_` but I want `storage` to be able to do `create_or_update_document_uri(**data)` with the validated data from the client. And the field that the client sends is called `uri` not `uri_`.

So I changed this to import the function directly which is a bit naughty and aliased the function to make it clearer.

seanh:
I think this one is good to go now, once direct linking is released?

I'd like to get this merged before we merge 3/3 and 4/4 as it'll just make them easier to test

h/api/models/document.py
```diff
@@ -181,6 +184,148 @@ def __repr__(self):
...
+    :param created: the time that will be used as the .created time for the new
+        DocumentURI, if a new one is created
+    :type created: datetime.datetime
+
+    :param updated: the time that will be set as the .updated time for the new
+        or existing DocumentURI
+    :type updated: datetime.datetime
+
+    """
+    docuri = DocumentURI.query.filter(
```
chdorner: Sorry for seeing this so late. We should change that to `session.query(DocumentURI).filter`, so we don't rely on the database session being used by `Document.query` is the same as the passed-in session (for example in the websockets it would be different).

h/api/models/document.py
```diff
@@ -181,6 +184,148 @@ def __repr__(self):
...
+    :param created: the value to use for the DocumentMeta's created attribute
+        if a new DocumentMeta is created
+    :type created: datetime.datetime
+
+    :param updated: the value to set the new or existing DocumentMeta's updated
+        attribute to
+    :type updated: datetime.datetime
+
+    """
+    existing_dm = DocumentMeta.query.filter(
```
chdorner: Same as above, should be `session.query(DocumentMeta).filter`

h/api/models/document.py
```diff
@@ -181,6 +184,148 @@ def __repr__(self):
...
+    :param claimant: the value to use for the DocumentMeta's claimant attribute
+        if a new DocumentMeta is created
+    :type claimant: unicode
+
+    :param type: the value of the new or existing DocumentMeta's type attribute
+    :type type: unicode
+
+    :param value: the value to set the new or existing DocumentMeta's value
+        attribute to
+    :type value: unicode
```
chdorner: I've just got bitten by the same thing when I wanted to try out the code in a console. DocumentMeta values are a list of unicodes.

chdorner: Which actually is quite annoying because when you pass it a string/unicode SQLAlchemy/Postgres will just write the string as a list of individual characters :disappointed:

h/api/models/document.py
seanh: Done. I wonder if the `DocumentURI.query` property should be removed then? Seems like trouble waiting to happen

h/api/models/document.py
seanh: Done

h/api/models/document.py
seanh: Done

chdorner:
LGTM

PR #3152, opened 2016-03-29
Postgres Write 3/4: Schema validation and transformation for creating new annotations in Postgres

seanh:
This pull request adds:

1. Transformation of annotation data and document metadata from the client into the format that the `storage` module will need it in (see <https://github.com/hypothesis/h/pull/3153>)

   The format of the "document" sub-object that the client POSTs in annotation create and update requests is quite far from the format that we need in order to create/update the `Document`, `DocumentMeta` and `DocumentURI` objects in the database. The new `h.api.parse_document_claims` module handles transforming this dict into two lists that are in a more convenient format for `storage` to later use to create the db objects. The code in here is adapted from existing code in `migrate.py` and `elastic.py`.

   `schemas` calls `parse_document_claims` to do this, so when the data comes out of schema validation it's in the right/convenient format.

2. Additional validation of the data from the client so that `storage` and the db don't crash.

   Now that we have a real database we can't just throw any crap into it, either the db or our own code will crash if required fields are missing or if known fields are the wrong type. Minimal additional JSON Schema validation has been added (only to the create annotation API and only when postgres_write is on) to prevent these crashes and send proper validation errors back.

`schemas` only does "structural" validation and transformations, i.e. work that doesn't require looking in the db. Some further "semantic" validation and transformation that does require reading from the will be done in `storage`.

### DIFF
--- a/h/api/parse_document_claims.py
+++ b/h/api/parse_document_claims.py
@@ -1,0 +1,237 @@
+"""
+Functions for parsing document claims data from the client.
+
+Functions for parsing the document claims (document metadata claims and URI
+equivalence claims) that the client POSTS in the JSON "document" sub-object in
+annotation create and update requests.
+
+The data is parsed into a format suitable for storage in our database model,
+and returned.
+
+"""
+from __future__ import unicode_literals
+
+
+def document_uris_from_data(document_data, claimant):
+    """
+    Return one or more document URI dicts for the given document data.
+
+    Returns one document uri dict for each document equivalence claim in
+    document_data.
+
+    Each dict can be used to init a DocumentURI object directly:
+
+        document_uri = DocumentURI(**document_uri_dict)
+
+    Always returns at least one "self-claim" document URI whose URI is the
+    claimant URI itself.
+
+    :param document_data: the "document" sub-object that was POSTed to the API
+        as part of a new or updated annotation
+    :type document_data: dict
+
+    :param claimant: the URI that the browser was at when this annotation was
+        created (the top-level "uri" field of the annotation)
+    :type claimant: unicode
+
+    :returns: a list of one or more document URI dicts
+    :rtype: list of dicts
+
+    """
+    document_uris = document_uris_from_links(document_data.get('link', []),
+                                             claimant)
+
+    document_uris.extend(
+        document_uris_from_highwire_pdf(document_data.get('highwire', {}),
+                                        claimant)
+    )
+
+    document_uris.extend(
+        document_uris_from_highwire_doi(document_data.get('highwire', {}),
+                                        claimant)
+    )
+
+    document_uris.extend(
+        document_uris_from_dc(document_data.get('dc', {}),
+                              claimant)
+    )
+
+    document_uris.append(document_uri_self_claim(claimant))
+
+    return document_uris
+
+
+def document_metas_from_data(document_data, claimant):
+    """
+    Return a list of document meta dicts for the given document data.
+
+    Returns one document meta dict for each document metadata claim in
+    document_data.
+
+    Each dict can be used to init a DocumentMeta object directly:
+
+        document_meta = DocumentMeta(**document_meta_dict)
+
+    :param document_data: the "document" sub-object that the client POSTed to
+        the API as part of a new or updated annotation
+    :type document_data: dict
+
+    :param claimant: the URI that the browser was at when this annotation was
+        created (the top-level "uri" field of the annotation)
+    :type claimant: unicode
+
+    :returns: a list of zero or more document meta dicts
+    :rtype: list of dicts
+
+    """
+    def transform_meta_(document_meta_dicts, items, path_prefix=None):
+        """Fill document_meta_dicts with document meta dicts for the items."""
+        if path_prefix is None:
+            path_prefix = []
+
+        for key, value in items.iteritems():
+            keypath = path_prefix[:]
+            keypath.append(key)
+
+            if isinstance(value, dict):
+                transform_meta_(document_meta_dicts,
+                                value,
+                                path_prefix=keypath)
+            else:
+                if not isinstance(value, list):
+                    value = [value]
+
+                document_meta_dicts.append({
+                    'type': '.'.join(keypath),
+                    'value': value,
+                    'claimant': claimant,
+                })
+
+    items = {k: v for k, v in document_data.iteritems() if k != 'link'}
+    document_meta_dicts = []
+    transform_meta_(document_meta_dicts, items)
+    return document_meta_dicts
+
+
+def document_uris_from_links(link_dicts, claimant):
+    """
+    Return document URI dicts for the given document.link data.
+
+    Process a document.link list of dicts that the client submitted as part of
+    an annotation create or update request and return document URI dicts for
+    all of the document equivalence claims that it makes.
+
+    """
+    document_uris = []
+    for link in link_dicts:
+
+        # Disregard self-claim URLs as they're added separately later.
+        if link.keys() == ['href'] and link['href'] == claimant:
+            continue
+
+        # Disregard doi links as these are being added separately from the
+        # highwire and dc metadata later on.
+        if link.keys() == ['href'] and link['href'].startswith('doi:'):
+            continue
+
+        # Disregard Highwire PDF links as these are being added separately from
+        # the highwire metadata later on.
+        if set(link.keys()) == set(['href', 'type']):
+            if link['type'] == 'application/pdf':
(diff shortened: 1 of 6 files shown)


seanh:

h/api/schemas.py
```diff
@@ -90,9 +205,75 @@ class CreateAnnotationSchema(object):
 
     def __init__(self, request):
         self.request = request
-        self.structure = AnnotationSchema()
+        if self.request.feature('postgres_write'):
+            self.structure = AnnotationSchema()
+        else:
+            self.structure = LegacyAnnotationSchema()
 
     def validate(self, data):
```
chdorner: Please help me understand something here.
By moving the previous implementation of `validate` into `_legacy_validate` and depending on the `postgres_write` feature flag either return the dict that has been sanitized by `validate` (for flag on) or `_legacy_validate` (for flag off), doesn't that mean that we start storing things like `target_uri` instead of the `uri` for the Elasticsearch annotations as soon as we turn the feature flag on?
Since turning the feature flag on means we write to both Elasticsearch and to Postgres. Which means that the JSON structure we return in the API will render `target_uri` instead of `uri` which is what we need.
Or am I missing something here?

h/api/schemas.py
```diff
@@ -59,6 +61,119 @@ class AnnotationSchema(JSONSchema):
...
+            },
+            'text': {
+                'type': 'string',
+            },
+            'uri': {
+                'type': 'string',
+            },
+        },
+        'required': [
+            'permissions',
```
chdorner: We might want to add the `uri` to the required fields as well, since currently the column does not allow `NULL` values and it doesn't provide a default value.

h/api/schemas.py
```diff
@@ -90,9 +205,75 @@ class CreateAnnotationSchema(object):
...
+        # annotation.target_selectors whose value is the first selector in
+        # the client'ss target.selectors list.
+        # Anything else in the target dict, and any selectors after the first,
+        # are discarded.
+        target = appstruct.pop('target', [])
+        if target:  # Replies and page notes don't have 'target'.
+            target = target[0]  # Multiple targets are ignored.
+            new_appstruct['target_selectors'] = target['selector']
+
+        new_appstruct['groupid'] = appstruct.pop('group', '__world__')
```
chdorner: We already have `__world__` set as the default in the database, I like it better here, maybe we should remove the default value from the database? so we don't have to maintain the same value in two places?

h/api/schemas.py
```diff
@@ -90,9 +205,75 @@ class CreateAnnotationSchema(object):
...
+        # groupid is added to the reply annotation later by the storage code.
+        # Here we just delete any group sent by the client from replies.
+        if new_appstruct['references'] and 'groupid' in new_appstruct:
+            del new_appstruct['groupid']
+
+        new_appstruct['extra'] = appstruct
+
+        # Transform the "document" dict that the client posts into a convenient
+        # format for creating DocumentURI and DocumentMeta objects later.
+        document_data = appstruct.pop('document', {})
```
chdorner: Would it make sense to pass separate copies of `document_data`into the next two function calls? So we don't rely on these two functions not mutating this dictionary.

h/api/parse_document_claims.py
```diff
@@ -0,0 +1,215 @@
...
+                                path_prefix=keypath)
+            else:
+                if not isinstance(value, list):
+                    value = [value]
+
+                document_meta_dicts.append({
+                    'type': '.'.join(keypath),
+                    'value': value,
+                    'claimant': claimant,
+                    'claimant_normalized':
```
chdorner: There's no need to set the `claimant_normalized` manually, the `claimant` setter takes care of this.

h/api/parse_document_claims.py
```diff
@@ -0,0 +1,215 @@
...
+        if link.keys() == ['href'] and link['href'].startswith('doi:'):
+            continue
+
+        uri_ = link['href']
+        type_ = None
+
+        # Handle Highwire PDF links.
+        if set(link.keys()) == set(['href', 'type']):
+            if link['type'] == 'application/pdf':
+                type_ = 'highwire-pdf'
```
chdorner: I was thinking about this case recently. Instead of relying on the annotator doing the right thing ([this](https://github.com/openannotation/annotator/blob/v1.2.x/src/plugin/document.coffee#L122-L128)), wouldn't it be safer to get the `highwire-pdf` DocumentURI from the highwire metadata, similar to what we do for the doi uris?

h/api/schemas.py
seanh: Oh damn, I forgot that the feature flag was supposed to mean write to both storage backends together, this code is intended (when combined with the next pull request) to write to postgres only when the feature flag is on.

Hopefully it won't be too difficult to fix it to write different data to each instead though..

seanh: This is fixed now (in this pr at least)

h/api/schemas.py
seanh: This one is actually okay because the `validate()` method does `new_appstruct['target_uri'] = appstruct.pop('uri', '')`.

In the future we probably should add uri to the required fields I agree, because why would we want people posting annotations with no uri? But for this pr, I just wanted to put in the absolute minimum validation to stop users from being able to crash it.

h/api/schemas.py
seanh: Schema validation and transformation does seem like the correct place to insert default values, so I agree. Can I do that in a separate PR though? I would like to get postgres PRs 1-4 all merged so that the whole thing can be tested together, before making a lot more changes to it.

h/api/schemas.py
seanh: Sure, seems no harm in being defensive

h/api/parse_document_claims.py
seanh: Done

h/api/parse_document_claims.py
seanh: So you mean we'd just add read the `data["highwire"]["pdf_url"]` list directly, and add a `DocumentURI` for each PDF URL in that list? Can do..

seanh: Done

h/api/parse_document_claims.py
seanh: Done

seanh:
Ok, I'll see if I can tidy up the history a bit...

h/api/schemas.py
seanh: Created an issue to do this separately https://github.com/hypothesis/h/issues/3177

seanh:
@chdorner I think I've responded to all the comments now, and I've completely rewritten the history of this one so it's nicer

h/api/schemas.py
chdorner: Sounds good to me

h/api/parse_document_claims.py
```diff
@@ -0,0 +1,237 @@
...
+    """
+    Return PDF document URI dicts for the given 'highwire' document metadata.
+
+    Process a document.highwire dict that the client submitted as part of an
+    annotation create or update request and return document URI dicts for all
+    of the PDF document equivalence claims that it makes.
+
+    """
+    document_uris = []
+    hwpdfvalues = highwire_dict.get('pdf', [])
```
chdorner: I think this should be `highwire_dict.get('pdf_url', [])`, as that's what the [annotator does](https://github.com/openannotation/annotator/blob/v1.2.x/src/plugin/document.coffee#L122-L128).
Or did you find some data that had the PDF url with the key `pdf`?

h/api/parse_document_claims.py
```diff
@@ -0,0 +1,237 @@
...
+        if link.keys() == ['href'] and link['href'] == claimant:
+            continue
+
+        # Disregard doi links as these are being added separately from the
+        # highwire and dc metadata later on.
+        if link.keys() == ['href'] and link['href'].startswith('doi:'):
+            continue
+
+        # Disregard Highwire PDF links as these are being added separately from
+        # the highwire and dc metadata later on.
```
chdorner: Nit, the comment should not say `and dc`.

chdorner:
Looking great! Only thing need to resolve is how it reads the `highwire-pdf` url (unless I'm wrong about my assumption). Then it's good to merge

h/api/parse_document_claims.py
seanh: Good catch, fixed

h/api/parse_document_claims.py
seanh: Done

PR #3153, opened 2016-03-29
Postgres Write 4/4: Storage code for creating new annotations in Postgres

seanh:
This pull request depends on the model code in <https://github.com/hypothesis/h/pull/3151> and the schemas code in <https://github.com/hypothesis/h/pull/3152> (not for the tests to pass, but for the code to actually work if you turn on the postgres_write feature flag).

This pull requests makes the `h.api.storage` changes necessary to accept the newly validated and transformed data from `h.api.schemas` (see <https://github.com/hypothesis/h/pull/3152>) and do the necessary logic to create and update `Annotation`, `Document`, `DocumentMeta` and `DocumentURI` objects in the database, for some of this calling new `h.api.models` functions from <https://github.com/hypothesis/h/pull/3151>.

The new storage logic is only when the postgres_write feature flag is on and only for creating new annotations, not for updating or deleting annotations.

Storing an annotation is considerably more complicated than it is with the legacy Elasticsearch storage code in h, because of the need to create and update document equivalence data and document metadata.

`h.api.storage.create_annotation()` is really a "logic" function with some semantic validation/transformation mixed in, all the structural validation and transformation and the model handling is elsewhere. The logic of `h.api.storage.create_annotation()` is:

* Force replies to have the same group as their parent
* Fail if replying to an annotation ID that doesn't exist
* Fail if writing to a group that you don't have permission to (note this has to happen after setting the group of replies)
* Create an `Annotation` model object from the (already validated) data
* Find all existing `Document` objects referred to by any existing `DocumentURI` objects that match any of the document URI equivalence claims in the post
* If there are no such `Document`s create a new `Document` with no `DocumentURI`s or `DocumentMeta`s pointing to it (yet)
* If there is one such `Document` use it
* If there are multiple such `Documents` merge them into one (all of the `DocumentURI`s will now point to one `Document` and the other `Document`s will be deleted, this destroys information)
* Update the `.updated` time of the `Document`
* Create new `DocumentURI`s, or update existing ones, for all document URI equivalence claims in the request
* Create new `DocumentMeta`s, or update existing ones, for all document metadata claims in the request

### DIFF
--- a/h/api/storage.py
+++ b/h/api/storage.py
@@ -94,14 +94,13 @@ def create_annotation(request, data):
 
     # The user must have permission to create an annotation in the group
     # they've asked to create one in.
-    if data['groupid'] == '__world__':
-        group_principal = security.Everyone
-    else:
+    if data['groupid'] != '__world__':
         group_principal = 'group:{}'.format(data['groupid'])
-    if group_principal not in request.effective_principals:
-        raise schemas.ValidationError('group: ' +
-                                      _('You may not create annotations in '
-                                        'groups you are not a member of!'))
+        if group_principal not in request.effective_principals:
+            raise schemas.ValidationError('group: ' +
+                                          _('You may not create annotations '
+                                            'in groups you are not a member '
+                                            'of!'))
 
     annotation = models.Annotation(**data)
     request.db.add(annotation)
--- a/h/api/storage.py
+++ b/h/api/storage.py
@@ -34,7 +34,7 @@ def annotation_from_dict(data):
     return models.elastic.Annotation(data)
 
 
-def fetch_annotation(request, id):
+def fetch_annotation(request, id, _postgres=None):
     """
     Fetch the annotation with the given id.
 
@@ -47,7 +47,12 @@ def fetch_annotation(request, id):
     :returns: the annotation, if found, or None.
     :rtype: dict, NoneType
     """
-    if _postgres_enabled(request):
+    # If no postgres arg is passed then decide whether to use postgres based
+    # on the postgres feature flag.
+    if _postgres is None:
+        _postgres = _postgres_enabled(request)
+
+    if _postgres:
         return request.db.query(models.Annotation).get(id)
 
     return models.elastic.Annotation.fetch(id)
@@ -82,7 +87,8 @@ def create_annotation(request, data):
     if data['references']:
         top_level_annotation_id = data['references'][0]
         top_level_annotation = fetch_annotation(request,
-                                                top_level_annotation_id)
+                                                top_level_annotation_id,
+                                                _postgres=True)
         if top_level_annotation:
             data['groupid'] = top_level_annotation.groupid
         else:
@@ -226,7 +232,7 @@ def expand_uri(request, uri):
 
 def _prepare(request, annotation):
     """Prepare the given annotation for storage."""
-    fetcher = partial(fetch_annotation, request)
+    fetcher = partial(fetch_annotation, request, _postgres=False)
     transform.prepare(annotation, fetcher)
 
     # Fire an AnnotationBeforeSaveEvent so subscribers who wish to modify an
--- a/h/api/storage.py
+++ b/h/api/storage.py
@@ -128,7 +128,7 @@ def create_annotation(request, data):
 
     for document_uri_dict in document_uri_dicts:
         models.create_or_update_document_uri(
-            db=request.db,
+            session=request.db,
             document=document,
             created=annotation.created,
             updated=annotation.updated,
@@ -136,7 +136,7 @@ def create_annotation(request, data):
 
     for document_meta_dict in document_meta_dicts:
         models.create_or_update_document_meta(
-            db=request.db,
+            session=request.db,
             document=document,
             created=annotation.created,
             updated=annotation.updated,
--- a/h/api/storage.py
+++ b/h/api/storage.py
@@ -9,11 +9,19 @@ assumed to be validated.
 
 from functools import partial
 
+from pyramid import i18n
+from pyramid import security
+
+from h.api import schemas
 from h.api import transform
+from h.api import models
 from h.api.events import AnnotationBeforeSaveEvent
 from h.api.models import elastic
 from h.api.models.annotation import Annotation
 from h.api.models.document import Document
+
+
+_ = i18n.TranslationStringFactory(__package__)
 
 
 def annotation_from_dict(data):
@@ -48,6 +56,14 @@ def fetch_annotation(request, id):
     return elastic.Annotation.fetch(id)
 
 
+def legacy_create_annotation(request, data):
+    annotation = elastic.Annotation(data)
+    # FIXME: this should happen when indexing, not storing.
+    _prepare(request, annotation)
+    annotation.save()
+    return annotation
+
+
 def create_annotation(request, data):
     """
     Create an annotation from passed data.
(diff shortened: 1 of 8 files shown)


seanh:

h/api/storage.py
```diff
@@ -61,12 +74,77 @@ def create_annotation(request, data):
     :returns: the created annotation
     :rtype: dict
     """
-    annotation = elastic.Annotation(data)
-
-    # FIXME: this should happen when indexing, not storing.
-    _prepare(request, annotation)
+    if not request.feature('postgres_write'):
+        return _legacy_create_annotation_in_elasticsearch(request, data)
```
chdorner: I think this is similar to a comment on another one of these 4 pull requests. This would mean that when the `postgres_write` feature flag is on, we only write to Postgres. But we decided to write to both Postgres and the old Elasticsearch. Or is this being handled somewhere else that I don't understand yet?

h/api/storage.py
```diff
@@ -61,12 +74,77 @@ def create_annotation(request, data):
...
+            raise schemas.ValidationError(
+                'references.0: ' +
+                _('Annotation {annotation_id} does not exist').format(
+                    annotation_id=top_level_annotation_id)
+            )
+
+    # The user must have permission to create an annotation in the group
+    # they've asked to create one in.
+    if data['groupid'] == '__world__':
+        group_principal = security.Everyone
```
chdorner: I don't understand this part of Pyramid yet. Does `security.Everyone` mean everyone but only logged in? Or does it also include logged-out users?

h/api/test/storage_test.py
```diff
@@ -9,109 +11,515 @@
...
-        mock.Mock(uri='http://example.com/', type='rel-canonical'),
-    ]
-    assert storage.expand_uri(request, "http://example.com/") == [
-            "http://example.com/"]
 
+@pytest.mark.usefixtures('AnnotationBeforeSaveEvent',
+                         'models',  # Don't try to talk to real Elasticsearch!
+                         'partial',
+                         'transform')
+class TestCreateAnnotation(object):
```
chdorner: I would rename this test class and the Postgres one to `TestCreateAnnotationElasticsearch/Legacy` and the Postgres one to `TestCreateAnnotation`, makes it easier to get rid of the legacy code later on.

h/api/storage.py
seanh: It also includes logged-out users. Pyramid automatically adds this principal to every request:

http://docs.pylonsproject.org/projects/pyramid/en/latest/narr/security.html#special-principal-names

http://docs.pylonsproject.org/projects/pyramid/en/latest/api/security.html?highlight=security.everyone#pyramid.security.Everyone

h/api/test/storage_test.py
seanh: Yep, these have ended up `TestCreateAnnotation` and `TestLegacyCreateAnnotation`, as the functions they test are being renamed to `create_annotation()` and `legacy_create_annotation()`.

seanh:
I've fixed this to have separate top level `create_annotation()` and `legacy_create_annotation()` and expect `views` to call them as necessary (calling both when the postgres feature flag is on). `storage` itself doesn't read the feature flag anymore.

Waiting for https://github.com/hypothesis/h/pull/3152 to be merged to finish this.

I still need to actually write the `views` code that calls both storage functions when postgres is on, and to manually test that the whole pile (views, schemas, storage, model) still works together.

h/api/storage.py
seanh: This is fixed now, see https://github.com/hypothesis/h/pull/3153#issuecomment-205929233

seanh:
Ok, this can be reviewed again.

I think it's creating the right stuff in the database (but please double check it for me).

I made it use the same ID for annotations in Pg and in Es. Deleting annotations when postgres is on still doesn't work for other reasons, but that can wait for a separate pull request. Haven't tried updating, that can wait too.

h/api/views.py
```diff
@@ -156,15 +156,35 @@ def create(request):
...
-    annotation = storage.create_annotation(request, legacy_appstruct)
+    # When 'postgres' is on make sure that annotations in the legacy
+    # Elasticsearch database use the same IDs as the PostgreSQL ones.
+    if request.feature('postgres'):
+        assert annotation.id
+        legacy_appstruct['id'] = annotation.id
 
-    _publish_annotation_event(request, annotation, 'create')
+    legacy_annotation = storage.legacy_create_annotation(request,
+                                                         legacy_appstruct)
```
chdorner: Just found a bug when I tried this out creating a reply to an annotation.

So the problem is that `legacy_create_annotation` ends up calling `h.api.transform.prepare` with a partial func for `h.api.storage.fetch_annotation`. Then inside the transform it fetches the top level annotation with that passed-in partial func and then expects the result to be a dictionay (`if 'group' in top_level_annotation:`), but since the postgres feature flag is on, the fetcher returns a postgres model instance.

chdorner:
Tried it out locally and found a bug when creating replies with the feature flag turned on (see above). Other than that it looks great!

h/api/storage.py
seanh: PS for only logged-in users it's `security.Authenticated`

chdorner: Wouldn't it make sense then to set this to `security.Authenticated` then?
We probably return an error earlier when a logged-out request tries to create an annotation, so it's not a big deal. It just reads wrong to me.

h/api/views.py
seanh: Right. So I think it only crashes when creating replies with the postgres feature flag is on. There's a mismatch because `fetch_annotation()` always fetches from either postgres or elasticsearch depending on the feature flag. But when the feature flag is on, the view code actually creates annotations in _both_ elasticsearch and postgres. When `legacy_create_annotation()` calls `fetch_annotation()` we always want to fetch from elastic, and when `create_annotation()` calls `fetch_annotation()` we always want to fetch from postgres, but when something else (e.g. the `read()` view) calls `fetch_annotation()` then we want to do the current behaviour of fetching from either postgres or elastic depending on the feature flag.

I guess that the annotation _should_ always exist in both postgres and es when the postgres feature flag is on. So one way to fix this might be to change the legacy transform code so that it works when the fetched parent annotation is a postgres one. It could access properties on the annotation classes, instead of reading them as dicts. That might have been simpler, but it didn't come to me first and also I don't really like changing the legacy code, so I went for this instead <https://github.com/hypothesis/h/pull/3153/commits/1a4f37d95caf8eb0d2a8d4d5f8c2a22206e21b8e>. This introduces some undesirable logic complexity but at least it can be removed when we remove the legacy es db.

Deleting and updating annotations are gonna need to do the same thing as well.

h/api/storage.py
seanh: Yeah it is weird. Actually I think the whole `storage` is written to assume that the user is authenticated so I don't think this check is needed at all: <https://github.com/hypothesis/h/pull/3153/commits/a239bbd654048d6265a2b10f88bc3218a19764d0>

seanh:
@chdorner Done!

h/api/storage.py
nickstenning: You still need to check that the user is a member of the group to which you're about to write the annotations.

seanh: Yes, _if_ the group isn't `"__world__"` it still does that (and there's a test for it)

chdorner:
Looking good! Unfortunately the JSON-LD branch created some merge conflicts. Will merge as soon as those are fixed

seanh:
Fixed

chdorner:
LGTM :tada:

PR #3155, opened 2016-03-30
Reintegrate annotation/document migration

chdorner:
Because we now run PostgreSQL 9.4 in production (which supports the
JSONB data type), and we are soon going to merge postgres annotation
write support behing a feature flag.


nickstenning:
LGTM.

PR #3156, opened 2016-03-30
Fix the signature of NSQ error/exception handlers

nickstenning:
The exception handlers we were installing to handle errors and exceptions thrown by `gnsq.Reader` objects had the wrong signature, and as a result were causing the email notification worker to crash.

The `gnsq` source code reveals that `on_error`, `on_exception`, and `on_giving_up` are all called with the reader instance as the first argument.

This commit:

- Fixes the signature of the installed handler functions
- Moves the setup of the error handlers into a helper function to
  improve the readability of `get_reader`
- Rewrites the tests to minimise the possibility of reintroducing this
  bug in the future -- the dispatch of simulated errors is the
  responsibility of the FakeReader object, not of individual test
  functions.

### DIFF
--- a/h/queue.py
+++ b/h/queue.py
@@ -31,27 +31,7 @@ def get_reader(settings, topic, channel, sentry_client=None):
     reader = gnsq.Reader(topic, channel, nsqd_tcp_addresses=addrs)
 
     if sentry_client is not None:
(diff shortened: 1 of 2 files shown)


chdorner:
LGTM

PR #3157, opened 2016-03-30
Fix a unicode warning

seanh:
We're getting this:

    Unicode type received non-unicode bind param value 'acct:fred@localhost'`

when registering new user accounts. This commit fixes it.


robertknight:
LGTM

PR #3159, opened 2016-03-30
Use filtering to show direct-linked annotations instead of selection (and thereby support direct links to replies)

robertknight:
**Background**

The sidebar has two different mechanisms for filtering the list of displayed annotations, filters and selection. Filters are used when you enter a query via search in the top bar, selection is used when you click on an annotation in the page and it filters the sidebar to show only that annotation.

**Overview**

This PR changes direct linking to filter the sidebar to show only direct-linked annotations using the filtering system, via an "id:<annotation ID>" query, rather than using the selection mechanism.

The reason for this is that the filtering mechanism works with both top-level annotations and replies, whereas the selection mechanism only works with top-level annotations, something I hadn't originally considered.

The alternative approach I could have taken would have been to implement support for selecting replies, but this would have involved re-implementing the system that hides parts of the conversation that do not match the search filter. I explored this approach but not too surprisingly it ended up being cleaner and simpler to re-use the existing system.

**UX Issues**

Some UX issues to note:
* The search field is now populated with an "id:<ID>" query automatically when following a direct link, IMO this looks a little ugly, I could perhaps not display the actual search query in the top bar in this case?
* When direct linking to a non-top level annotation, you'll now see a "View N more in this conversation" link below the annotation body, matching the behavior when using the search tool

### DIFF
--- a/h/static/scripts/widget-controller.js
+++ b/h/static/scripts/widget-controller.js
@@ -253,6 +253,14 @@ module.exports = function WidgetController(
            !!threading.idTable[directLinkedAnnotationID()];
   };
 
+  $scope.topLevelThreadCount = function () {
+    return threading.root.children.length;
+  };
+
+  $scope.shouldShowTotalCount = function () {
+    return groups.focused().public && !!directLinkedAnnotationID();
+  };
+
   $scope.isLoading = isLoading;
 
   $rootScope.$on(events.BEFORE_ANNOTATION_CREATED, function (event, data) {
--- a/h/static/scripts/widget-controller.js
+++ b/h/static/scripts/widget-controller.js
@@ -138,6 +138,10 @@ module.exports = function WidgetController(
     searchClient.get({uri: uri, group: group});
   }
 
+  function isLoading() {
+    return searchClients.length > 0;
+  }
+
   /**
    * Load annotations for all URLs associated with `frames`.
    *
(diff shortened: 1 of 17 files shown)


h/static/scripts/widget-controller.js
```diff
@@ -160,28 +181,31 @@ module.exports = function WidgetController(
...
+      return tag.tag === annot.$$tag;
     });
     if (!matchesSelection) {
       return;
     }
-    focusAnnotation(selectedAnnot);
-    scrollToAnnotation(selectedAnnot);
+
+    // Only annotations (not replies) can be scrolled to and focused, so
+    // we need to the parent annotation of the direct-linked annotation
```
robertknight: Typo: "we need to the"

robertknight:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;O5KyU1DN&#x2F;309-bug-fix-direct-links-from-replies"> BUG: Fix direct links from replies</a>

robertknight:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;zZRS63jX&#x2F;312-the-clear-selection-button-has-new-copy-i-e-see-all-n-annotations"> The &quot;Clear selection&quot; button has new copy, i.e. &quot;See all N annotations.&quot;</a>

robertknight:
We've decided not to support direct links to replies in the first iteration, but there are some other worthwhile that I'd like to extract out into a separate PR.

robertknight:
Closing this for the moment. I will cherry-pick the relevant parts into other PRs.

PR #3161, opened 2016-03-31
Make client injection robust to pages that replace the URL fragment as they load

### LINKED ISSUE #3122: EBSCOhost direct link not working
Visit https://hyp.is/AVPIPxc2H9ZO4OKSldxe with 1) In in a browser without the Chrome extension, 2) Chrome with the extension

**Expected result**

Page loads, extension activated, page scrolls to annotation

**Actual result**

In scenario 1) Works as expected, in scenario 2) Page loads, extension not activated.

The "#annotations:<id>" fragment is getting rewritten and URL-escaped when the page loads.

### LINKED ISSUE #3137: Direct link fails on medium.com (and other sites that modify URL fragment on page load)
### Steps to reproduce
1. Visit https://hyp.is/L_OOCOt9RY6Eij3Q-fFiTQ using Chrome extension

### Expected behaviour
Sidebar opens, focuses and displays annotation

### Actual behaviour
Sidebar opens, focuses annotation, but displays "You do not have permission to see this annotation"

Chrome / Windows 10

### Additional details
The Medium URL, https://medium.com/intercom-inside/the-dribbblisation-of-design-406422ccb026#.f2jqk75gw, has its own fragment id which, in this case, we do not replace.

robertknight:
This is a set of changes that make direct-linking more robust to changes in the URL fragment by the page when it loads, such as on Medium and EBSCOHost.

There are two changes here which help with this:

1. Making the parsing of the '#annotations' fragment stricter by only matching characters after '#annotations:' which are valid annotation ID chars (the URL-safe Base64 charset)
2. Capturing the ID from the '#annotations' fragment when a tab starts loading in the Chrome extension and passing that ID through to the sidebar when it is injected, so that the direct link works even if client-side JS rewrites the fragment whilst it loads, or a client-side triggered redirect causes the URL fragment to be lost.

Fixes #3137
Fixes #3122

### DIFF
--- a/h/browser/chrome/lib/sidebar-injector.js
+++ b/h/browser/chrome/lib/sidebar-injector.js
@@ -14,6 +14,19 @@ function toIIFEString(fn) {
 }
 
 /**
+ * Adds a <script> tag containing JSON config data to the page.
+ *
+ * Note that this function is stringified and injected into the page via a
+ * content script, so it cannot reference any external variables.
+ */
+function addJSONScriptTagFn(name, content) {
+  var scriptTag = document.createElement('script');
+  scriptTag.className = name;
+  scriptTag.textContent = content;
+  document.head.appendChild(scriptTag);
+}
+
+/**
  * Extract the value returned by a content script injected via
  * chrome.tabs.executeScript() into the main frame of a page.
  *
@@ -62,18 +75,23 @@ function SidebarInjector(chromeTabs, dependencies) {
     throw new TypeError('isAllowedFileSchemeAccess must be a function');
   }
 
-  /* Injects the Hypothesis sidebar into the tab provided.
+  /**
+   * Injects the Hypothesis sidebar into the tab provided.
    *
-   * tab - A tab object representing the tab to insert the sidebar into.
+   * @param {Tab} tab - A tab object representing the tab to insert the sidebar
+   *        into.
+   * @param {Object?} config - An object containing configuration info that
+   *        is passed to the app when it loads.
    *
    * Returns a promise that will be resolved if the injection succeeded
    * otherwise it will be rejected with an error.
    */
-  this.injectIntoTab = function(tab) {
+  this.injectIntoTab = function(tab, config) {
+    config = config || {};
     if (isFileURL(tab.url)) {
       return injectIntoLocalDocument(tab);
     } else {
-      return injectIntoRemoteDocument(tab);
+      return injectIntoRemoteDocument(tab, config);
     }
   };
 
@@ -193,7 +211,7 @@ function SidebarInjector(chromeTabs, dependencies) {
     });
   }
 
-  function injectIntoRemoteDocument(tab) {
+  function injectIntoRemoteDocument(tab, config) {
     if (isPDFViewerURL(tab.url)) {
       return Promise.resolve();
     }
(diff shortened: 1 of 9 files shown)


seanh:
Nice

h/static/scripts/annotator/config.js
```diff
@@ -21,9 +35,9 @@ function config(window_) {
     }
   }
 
-  var annotFragmentMatch = window_.location.hash.match(/^#annotations:(.*)/);
-  if (annotFragmentMatch) {
-    options.annotations = annotFragmentMatch[1];
+  var directLinkedID = annotationIDs.extractIDFromURL(window_.location.href);
```
nickstenning: I'd suggest adding a comment here to explain that this is not the only place where direct link fragments are detected. If I were reading this code in the future, I'd be unlikely to guess that this is the "last resort" place to detect the fragment, and that the Chrome extension is already doing this.

h/browser/chrome/lib/sidebar-injector.js
```diff
@@ -288,6 +308,22 @@ function SidebarInjector(chromeTabs, dependencies) {
...
   }
+
+  /**
+   * Inject configuration information for the Hypothesis application
+   * into the page as JSON data via a <meta> tag.
+   *
+   * A <meta> tag is used because that makes it available to JS content
+   * running in isolated worlds.
+   */
+  function injectConfig(tabId, config) {
```
nickstenning: Is there any particular reason to use a `<meta>` tag rather than `<script class="js-hypothesis-settings" type="application/json">` as used in `app.html`? I ask only because it would seem sensible for there to be consistency between places we read settings data from the DOM.

h/static/scripts/annotator/config.js
robertknight: Ah, now that you mention "last resort", that makes me realize that in the context of Via, we could re-use the mechanism added here for the Chrome extension and have Via inject the same config tag into the page.

robertknight: That can come as a separate step though

h/browser/chrome/lib/sidebar-injector.js
robertknight: One difference is that these are a different set of settings. `js-hypothesis-settings` in `app.html` are the sidebar's private settings. This is another way to specify the same set of public settings as `window.hypothesisConfig`. A `<meta>` tag seemed the obvious way to do a key/value pair to me, although we could perhaps use a script tag with `<script class="js-hypothesis-config">` instead?

I think the biggest reason to use one vs. the other would be if there was some environment where using one or the other caused a website compatibility problem.

nickstenning: I know they're not the same settings. I'm merely making the point that we probably should use the same basic mechanism unless we have a good reason to do otherwise.

robertknight: I have reworked this now to use the same mechanism but look for a `js-hypothesis-config` class on the `<script>` tag instead of `js-hypothesis-settings`.

h/static/scripts/annotator/config.js
robertknight: Explanation added as suggested.

h/static/scripts/annotator/test/config-test.js
```diff
@@ -42,4 +63,12 @@ describe('annotator configuration', function () {
       firstRun: true,
     });
   });
+
+  it('merges the config from the "hypothesis-config" meta tag', function () {
```
nickstenning: Description needs updating.

h/static/scripts/settings.js
```diff
@@ -6,16 +6,20 @@ require('core-js/fn/object/assign');
...
+ * class `settingsClass` which contain JSON content.
  *
  * If there are multiple such tags, the configuration from each is merged.
  *
  * @param {Document|Element} document - The root element to search for
  *                                      <script> settings tags.
+ * @param {string} settingsClass - The class name to match on <script> tags.
  */
-function settings(document) {
+function settings(document, settingsClass) {
```
nickstenning: Let's add a test for the `settingsClass` param.

robertknight: Done

nickstenning:
LGTM.

PR #3162, opened 2016-03-31
Minor copy and style tweaks to the logged-out message

nickstenning:
- We prefer "Hypothesis" in prose text to "Hypothes.is"
- Remove unwanted padding at the end of anchors.
- Say "sign in" rather than "log in," as we do everywhere else in the
  sidebar.
- Slight improvement to readability of call-to-action sentence by
  separating the "X or Y" clauses.

#### Before

screen shot 2016-03-31 at 13 01 35

#### After

screen shot 2016-03-31 at 13 18 58

Ping @conordelahunty for review/approval/MERCILESS REJECTION.

### DIFF
--- a/h/static/styles/loggedout-message.scss
+++ b/h/static/styles/loggedout-message.scss
@@ -7,13 +7,7 @@
   flex-direction: column;
 }
 
-.loggedout-message-actions {
-  margin: auto;
-  width: 75%;
-  text-align: center;
-}
-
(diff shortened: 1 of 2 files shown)


conordelahunty:
Don't think it needs the gap between paragraphs, but otherwise all are fine.

nickstenning:
Cheers. Updated.

robertknight:
LGTM

PR #3163, opened 2016-03-31
Uri normalize unicode

chdorner:
We have quite a good chunk of code that turns the result of this function into a unicode type string all over the code base. So it is about time that this function starts returning a unicode string.

### DIFF
--- a/h/api/test/uri_test.py
+++ b/h/api/test/uri_test.py
@@ -5,7 +5,7 @@ from __future__ import unicode_literals
 import pytest
 
 from h.api import uri
-from h._compat import text_type
+from h.api._compat import text_type
 
 TEST_URLS = [
     # Should strip https://via.hypothes.is/ from the start of URIs
--- a/h/api/test/uri_test.py
+++ b/h/api/test/uri_test.py
@@ -5,9 +5,9 @@ from __future__ import unicode_literals
 import pytest
 
 from h.api import uri
+from h._compat import text_type
 
-
-@pytest.mark.parametrize("url_in,url_out", [
+TEST_URLS = [
     # Should strip https://via.hypothes.is/ from the start of URIs
     ("https://via.hypothes.is/https://example.com", "https://example.com"),
     ("https://via.hypothes.is/http://foo.com/bar/", "http://foo.com/bar"),
@@ -154,6 +154,13 @@ from h.api import uri
     ("http://example.com?gclid_foo=abcde", "http://example.com?gclid_foo=abcde"),
     ("http://example.com?bar_gclid=abcde", "http://example.com?bar_gclid=abcde"),
     ("http://example.com?WT=abcde", "http://example.com?WT=abcde"),
-])
+]
+
+
+@pytest.mark.parametrize("url_in,url_out", TEST_URLS)
 def test_normalize(url_in, url_out):
     assert uri.normalize(url_in) == url_out
+
+@pytest.mark.parametrize("url,_", TEST_URLS)
+def test_normalize_returns_unicode(url, _):
+    assert isinstance(uri.normalize(url), text_type)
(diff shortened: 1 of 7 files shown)


h/api/test/uri_test.py
```diff
@@ -154,6 +153,13 @@
...
     ("http://example.com?WT=abcde", "http://example.com?WT=abcde"),
-])
+]
+
+
+@pytest.mark.parametrize("url_in,url_out", TEST_URLS)
 def test_normalize(url_in, url_out):
     assert uri.normalize(url_in) == url_out
+
+@pytest.mark.parametrize("url,_", TEST_URLS)
```
nickstenning: Rather than duplicating the tests, I'd suggest you add `from __future__ import unicode_literals` to the top of the test file, and then add a specific test to ensure that URLs get mapped to text:

```
@pytest.mark.parametrize("url_in", [
    "http://example.com/text",
    b"http://example.com/bytes",
])
def test_normalize_returns_text(url_in):
    assert isinstance(uri.normalize(url_in), text_type)
```

And note that you should use `text_type`, which is `str` on Python 3 and `unicode` on Python 2. I'd suggest creating a `h/api/_compat.py` and copying the relevant parts of `h/_compat.py` into it.

h/api/uri.py
```diff
@@ -64,6 +64,7 @@
 """
 import re
 
+from h._compat import text_type
```
nickstenning: As touched on above, this shouldn't really be importing from `h._compat`. We should duplicate the relevant bits into `h.api._compat`. Also the below imports.

h/api/uri.py
```diff
@@ -140,11 +141,11 @@ def normalize(uristr):
 
     # If this isn't a URL, we don't perform any normalization
     if uri.scheme.lower() not in URL_SCHEMES:
-        return uristr
+        return text_type(uristr)
```
nickstenning: This will fail on Python 2 if `uristr` has non-ASCII characters in it. You should specify the encoding just as you've been doing inline.

h/api/test/uri_test.py
chdorner: The test file already imports `unicode_literals`.

Since the implementation has 3 different return statements, and each of those converts the value first to `text_type` before returning, I thought it is going to be safer to run all the test cases through the `isinstance` assertion. This will also ensure that any changes in the future (maybe another if statement with an early return) will be tested automatically.

Thoughts?

I will change the assertion to use `text_type` rather than `unicode` though, that's a good poing.

nickstenning:
I think overall I'd be more comfortable if this function dealt with text (and hence sequences of codepoints) internally rather than converting from text to bytes on the way in and then back to text on the way out.

That said, this will do for now, once the comments above are addressed.

h/api/test/uri_test.py
nickstenning: I'd rather we had custom cases that were guaranteed to test all return paths, in that case.

chdorner: What's the reasoning for not testing on all the URIs?
I just think that the ~240ms (just did a quick test, three cases vs. all the urls) speed penalty is justified because it makes it harder to introduce a bug which wouldn't return a unicode string.

nickstenning: Ok!

chdorner: Just pushed the change that tests for `text_type` instead of `unicode`.

h/api/uri.py
chdorner: :+1:
Duplicated the relevant bits (also looked for other `h._compat` occurrences in `h.api`) to `h.api._compat` and changed the import paths.

h/api/uri.py
chdorner: Fixed.

nickstenning:
LGTM.

PR #3166, opened 2016-03-31
Fix broken loggedout CTA.

sheetaluk:
* Create a loggedout-message directive
* Move related html to separate template

https://trello.com/c/dR3qXaxx/257-add-explanatory-text-below-public-focused-annotation-when-logged-out

### DIFF
--- a/h/client.py
+++ b/h/client.py
@@ -16,6 +16,7 @@ ANGULAR_DIRECTIVE_TEMPLATES = [
     'dropdown_menu_btn',
     'excerpt',
     'group_list',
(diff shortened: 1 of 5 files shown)


robertknight:
LGTM. I made a couple of minor changes:

* Use hyphen case in the templates, ie. `<loggedout-message>` rather than `<loggedout_message>`. The latter works, which surprised me (only a little) but we use hyphenated tag names everywhere elase.
* I sorted the directives in `app.js` and `client.py` alphabetically.

PR #3168, opened 2016-04-01
Remove incontext links for replies

### LINKED ISSUE #3127: direct link fails for reply
### Steps to reproduce
1.Visit http://www.nytimes.com/2016/02/22/us/politics/delegate-count-leaving-bernie-sanders-with-steep-climb.html?_r=0
2. Click bouncer link on the annotation: OK
3. Click bouncer link on its reply: Nada.

### Expected behaviour
Sidebar opens, annotation focused, reply expanded.

### Actual behaviour
Sidebar displays Clear Selection and says 1 selected annotation, but nothing displays.

robertknight:
We are currently not decided on how we want to support direct-linking to
replies. Removing the incontext links causes the client to show links to the standalone HTML page instead and avoids exposing incontext links to other clients.

Fixes #3127
https://trello.com/c/tWqgkw82/313-remove-direct-links-to-replies

### DIFF
--- a/h/links.py
+++ b/h/links.py
@@ -15,7 +15,8 @@ def html_link(request, annotation):
 
 def incontext_link(request, annotation):
     """Generate a link to an annotation on the page where it was made."""
-    if not request.feature('direct_linking'):
+    is_reply = bool(annotation.references)
+    if not request.feature('direct_linking') or is_reply:
         return None
 
     bouncer_url = request.registry.settings.get('h.bouncer_url')
--- a/h/links.py
+++ b/h/links.py
@@ -1,0 +1,35 @@
+# -*- coding: utf-8 -*-
+
+"""
+Provides links to different representations of annotations.
+"""
+
+
+from h._compat import urlparse
+
+
+def html_link(request, annotation):
+    """Generate a link to an HTML representation of an annotation."""
+    return request.route_url('annotation', id=annotation.id)
+
+
+def incontext_link(request, annotation):
+    """Generate a link to an annotation on the page where it was made."""
+    if not request.feature('direct_linking'):
+        return None
+
+    bouncer_url = request.registry.settings.get('h.bouncer_url')
+    if not bouncer_url:
+        return None
+    return urlparse.urljoin(bouncer_url, annotation.id)
+
+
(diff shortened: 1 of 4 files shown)


h/views/main.py
```diff
@@ -87,7 +87,10 @@ def _html_link(request, annotation):
 
 def _incontext_link(request, annotation):
     """Generate a link to an annotation on the page where it was made."""
-    if not request.feature('direct_linking'):
+
+    # Direct links are currently not supported for replies
+    is_reply = bool(annotation.get('references'))
```
robertknight: I'm not sure that this is a very Pythonic way to achieve this. We also have a similar `is_reply` helper in one other place in the code. Anyone have suggestions on a better way to do this?

h/views/main.py
```diff
@@ -87,7 +87,10 @@ def _html_link(request, annotation):
 
 def _incontext_link(request, annotation):
```
robertknight: I could use suggestions on how best to test link handling. This is registered via the `includeme()` function below, so one approach would perhaps be to call `includeme()` with a mock context object and intercept the link handler registration. Thoughts?

nickstenning: I'd suggest you move everything link-related into its own module, say `h.links`, where `incontext` and `html` can become public functions, with their own tests. Testing the `includeme` is a route to sadness and despair :)

h/views/main.py
nickstenning: This is fine.

h/links.py
```diff
@@ -0,0 +1,36 @@
...
+def incontext_link(request, annotation):
+    """Generate a link to an annotation on the page where it was made."""
+    is_reply = bool(annotation.references)
+    if not request.feature('direct_linking') or is_reply:
+        return None
+
+    bouncer_url = request.registry.settings.get('h.bouncer_url')
+    if not bouncer_url:
+        return None
+    return urlparse.urljoin(bouncer_url, annotation.id)
```
nickstenning: Yeah, I'm not sure I necessarily prefer this over using a named static route as the previous implementation did, but it's fine for now.

nickstenning:
LGTM.

PR #3169, opened 2016-04-01
Avoid showing 'Clear Selection' or 'Clear Search' buttons whilst loading

robertknight:
When loading annotations whilst a search query is active or selection is
present, avoid showing the 'Clear Search' or 'Clear Selection' buttons
until the search has actually completed and we know whether any
annotations match the query or selection.

This fixes the flash of the 'Clear Selection' button when viewing a
direct-linked annotation.


nickstenning:
LGTM.

PR #3170, opened 2016-04-01
Display 'Show all N public annotations' message when there is a selection

robertknight:
In the case when the selection has been set as a result of visiting a
direct-linked annotation, this makes it more obvious that there are
other annotations on the page to see.

 * When the public group is focused and there is a selection,
   change the wording of the 'Clear Selection' message.

https://trello.com/c/zZRS63jX/312-the-clear-selection-button-has-new-copy-i-e-see-all-n-annotations

### DIFF
--- a/h/static/scripts/directive/search-status-bar.js
+++ b/h/static/scripts/directive/search-status-bar.js
@@ -1,3 +1,5 @@
+'use strict';
+
 // @ngInject
 module.exports = function () {
   return {
@@ -8,6 +10,7 @@ module.exports = function () {
       onClearSelection: '&',
       searchQuery: '<',
       selectionCount: '<',
+      totalCount: '<',
     },
     templateUrl: 'search_status_bar.html',
   };
--- a/h/static/scripts/directive/test/search-status-bar-test.js
+++ b/h/static/scripts/directive/test/search-status-bar-test.js
@@ -1,5 +1,7 @@
 'use strict';
 
+var angular = require('angular');
+
 var util = require('./util');
 
 describe('searchStatusBar', function () {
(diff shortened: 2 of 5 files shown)


robertknight:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;zZRS63jX&#x2F;312-the-clear-selection-button-has-new-copy-i-e-see-all-n-annotations"> The &quot;Clear selection&quot; button has new copy, i.e. &quot;See all N annotations.&quot;</a>

nickstenning:
Looking at [the PRD](https://docs.google.com/document/d/1ug4ur81mgKzWyh9Juv1jHb2HXOYJM2BsQuGG5XZcOVo/edit) I don't think the intent was for this to become a special case for the public group. We just want to change the piece of UI which currently looks like:

>  [Clear selection]   M selected annotation(s)

to look like

> [See all N annotations]

Did I misunderstand the intent, @conordelahunty?

conordelahunty:
Yeah, I think can be solved by just saying "See all N annotations" rather than differentiating between public and private.

robertknight:
Following discussion with @conordelahunty on Slack, the behavior now is:

- Change the wording of the 'Clear Selection' button to 'Show all annotations' (if 0 or 1 annotations present in group) or 'Show all N annotations' (if many annotations in group).
- Show the same wording regardless of the selected group

nickstenning:
You'll need to update the tests.

nickstenning:
Now the tests pass but it needs a rebase because I merged #3169 :)

nickstenning:
Excellent. Much nicer and much simpler than the first version :)

PR #3171, opened 2016-04-01
Rename postgres_read feature flag to postgres

seanh:
This one feature flag is to be used to enable both reading from and
writing to PostgreSQL.


nickstenning:
LGTM.

PR #3172, opened 2016-04-01
Append target URIs to "in context" links in the API

nickstenning:
As originally agreed, these links should have "schemaless" versions of the annotation target URI appended to the end.


robertknight:
LGTM :+1:

PR #3173, opened 2016-04-01
Use require() for templates

robertknight:
This PR changes the way UI components fetch their HTML templates to use the same `require()` mechanism that is used for importing JS imports, instead of relying on the template being added to `app.html` as a `<script>` tag with the HTML as the content. Using `require()` for .html files is done via the `stringify` Browserify transform which converts HTML into JS strings.

There are a number of general reasons for doing this, plus it also helps specifically with fixing some issues with collapsible excerpts [1]:

* It makes components more self-contained, as each pulls in its own templates
* ~~It moves my drinks cabinet a few inches closer to Berlin~~ It simplifies the app.html build a little
* It makes behavior of all components the same, because they will all resolve their templates synchronously, whereas there were some components previously using `template` (which is sync) and others using `templateUrl` (which is async).
* It is faster, because it avoids going through a bunch of machinery internally in Angular that is used for fetching templates from the network (internally `templateUrl` fetches work exactly the same as an `$http` request, except that it uses a local cache which gets pre-populated from the `<script>` tags on the page).

[1] Specifically, it makes template fetching synchronous, which means that the `<excerpt>` post-link function is called, it can rely on its content having been fully processed and thus it can measure the height of that content. Subsequently the height can change, but in response to specific known events (eg. images loading, the sidebar being resized) which we can watch for separately.

### DIFF
--- a/scripts/gulp/create-bundle.js
+++ b/scripts/gulp/create-bundle.js
@@ -11,7 +11,6 @@ var coffeeify = require('coffeeify');
 var exorcist = require('exorcist');
 var gulpUtil = require('gulp-util');
 var mkdirp = require('mkdirp');
-var stringify = require('stringify');
 var uglifyify = require('uglifyify');
 var watchify = require('watchify');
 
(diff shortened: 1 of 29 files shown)


scripts/gulp/create-bundle.js
```diff
@@ -117,6 +118,12 @@ module.exports = function createBundle(config, buildOpts) {
 
   var bundle = browserify([], bundleOpts);
 
+  bundle.transform(stringify, {
```
nickstenning: All our other browserify transforms are defined in `package.json`. Should this not go there, too?

robertknight: Given that this transform is applied unconditionally, it probably should. I'll give it a go.

robertknight: Done.

nickstenning:
LGTM.

PR #3176, opened 2016-04-05
New annotation threading implementation

robertknight:
This PR is a spike of a reworking of how the client takes a list of annotations and replies from the server and renders them as a nested conversation thread, taking into account the current sort, filter and grouping settings.

**Goals**

This work aims to address a number of issues with the current threading system, make it easier to implement things that are already planned (such as the Annotations / Orphans / Pages Notes tabs) and experiment with new approaches to organizing conversation threads and grouping/sorting/filtering threads in future.

Specifically:
- Make it easier to make changes relating to sorting, filtering and grouping of annotations
- Improve perceived performance by eliminating the current flickering as annotations enter the UI over a series of animation frames.
- Improve actual performance by eliminating a lot of work related to filtering/sorting etc that currently happens in the `$digest` cycle, even if nothing changed.
- Make the UI scale better to large conversations and heavily annotated pages by making it possible to lazily load annotation cards (either automatically or possibly in response to the user tapping a 'Show More' button at the bottom of the list)
- Make it easier to test the threading/sorting/filtering logic by decoupling it from the visual presentation

**Design**

The design in this PR implements a [Redux](https://github.com/reactjs/redux)-esque model [1] where the set of loaded annotations and other state (sort, filter settings) that is needed to render the UI is stored in a single immutable structure. This is currently held by the `annotationUI` service. The `rootThread` service listens for changes in that state and in response, calls a `buildThread` function to generate an immutable `Thread` object which is the UI model. The `Thread` object's structure looks like the rendered UI:

```ts
interface Thread {
  annotation: Annotation
  collapsed: boolean;
  replyCount: number;
  children: Array<Thread>;
  ...
}
```

From that, the `viewer.html` root component renders the top-level annotation threads. For each top-level thread, an `<annotation-thread>` component renders the annotation card (the `<annotation>` component), the reply counts and if uncollapsed, any replies. When the user clicks links to expand annotations or changes the sort/filter state, a notification is dispatched which causes a new `Thread` object to be generated. Angular then takes care of syncing the UI to match the new `Thread`. Note that when the `Thread` object changes, Angular doesn't recreate the whole UI, it just re-orders the existing `<annotation-thread>` components (using the `track-by` option for `ng-repeat`) and creates new ones for newly visible annotations.

Implementing new sorting/filtering etc. involves adding the state to the central store and implementing support in the `buildThread()` function for taking that into account when generating the `Thread` structure.

The overall UI structure and data flow looks like this:

threading

**What is done/not-done in this PR**

This PR implements the basic design and should preserve all current sorting, filtering and other behavior. It has some initial tests for the `buildThread` and `rootThread` pieces to show how they would work. What's missing are a comprehensive set of tests for `buildThread`, `rootThread` and `<annotation-thread>` and fixups for the existing tests and complete documentation for all the new functions.

[1] [Redux] is a small, popular library for managing state in front-end apps which is designed around this architecture. There is a lot of useful tooling, patterns, documentation etc. around it. I think it would be worth considering as a way to manage state in the app, but as a separate step.


nickstenning:
TBQH this diff is too large to meaningfully review, but the intent and description sounds good. As previously mentioned I'm a big fan of making the threading system much more explicit and less magic (*cough* deepCount *cough*).

The only caveat I would have is to ensure you preserve the semantics of the JWZ threading algorithm, namely that the `references` field contains a list of ancestors, ordered oldest to newest. These are not all guaranteed to exist, and the list is not guarantee to be complete.

h/static/scripts/build-thread.js
```diff
@@ -0,0 +1,295 @@
...
+ * If the Annotation has been created on the server, it will have
+ * an ID assigned, otherwise we fall back to the local-only '$$tag'
+ * property.
+ */
+function id(annotation) {
+  return annotation.id || annotation.$$tag;
+}
+
+/**
+ * Creates a thread of annotations from a list of annotations.
```
nickstenning: This comment should certainly explain where the core of the algorithm comes from 😉

robertknight:
> TBQH this diff is too large to meaningfully review, but the intent and description sounds good.

OK, what I could do is split up the new implementation into different commits for the relevant chunks (model layer + tests, view layer + tests etc), land those in one PR and then have a separate PR which actually swaps out the old implementation for the new one.

robertknight:
> The only caveat I would have is to ensure you preserve the semantics of the
> JWZ threading algorithm, namely that the references field contains a list of ancestors,
> ordered oldest to newest. These are not all guaranteed to exist, and the list is not
> guarantee to be complete.

This matches my understanding of the inputs to the JWZ threading algorithm, I'll re-read the article in case I missed anything. BTW. The `threadAnnotations()` function in `build-thread.js` is the replacement for the original JWZ implementation.

robertknight:
Had some initial discussions about this now and I think we're good to proceed. The next steps here are to finish off the tests for the new code and split this up into smaller pieces that are easier to review.

h/static/scripts/build-thread.js
robertknight: TBH I think this is so much simpler than jwz that I don't think it makes sense to describe it as being the JWZ algorithm any more. Re-reading the JWZ docs though I realized that one difference introduced here is in what happens if an annotation with replies is deleted whilst the user has the sidebar open. In this branch, the reply becomes a top-level annotation and ends up behaving like a page note. In master, there is a placeholder "Message was deleted" annotation. In both cases the replies can be edited and new replies can be created, which are then immediately 'lost' when the page is reloaded. I filed https://trello.com/c/9boX9ToY/

robertknight:
OK. I've talked through the design with Nick and Sheetal. I'm closing this spike PR and opening a series of new ones for the implementation proper.

h/static/scripts/build-thread.js
nickstenning: I certainly agree that there is a design discussion to be had here. For now I think there are probably many improvements that can be made to the code without changing existing behaviour. One question: I'm not sure how the new code works when intermediate replies are deleted, but here's a scenario raised by the example you gave above:

```
annotation
|
` reply 1
  |
   ` reply 2
```

What happens if reply 1 is deleted? The JWZ algorithm allows us to infer the presence of the missing message and display that information to the user. While we have explicit threaded conversations I think it's important that we don't change to something that essentially lies about what happened. For example, I think producing the rendered output

```
annotation
|
` reply 2
```

would be problematic.

robertknight: In the current implementation in this branch, you would see the output you describe at the end of that comment. It should be fairly straightforward to infer the existence of a missing annotation and create the necessary entry in the hierarchy.

PR #3178, opened 2016-04-05
Add tests for authentication policy

nickstenning:
I noticed that we didn't have tests for this rather important part of the application, and I had a spare moment, so I wrote some.

In order to make `AuthenticationPolicy` easier to test, I also made a small refactoring that unifies logic previously split between the authentication policy and `h.auth.util.bearer_token` into a single function: `h.auth.tokens.authenticated_userid`.


robertknight:
LGTM

finitud:
👍

PR #3179, opened 2016-04-06
Fix 'Location' sort order for annotations

### LINKED ISSUE #3106: Annotation not in document order
### Steps to reproduce
1. Visit https://gamesandlearning.wordpress.com/2016/03/16/playful-annotation-in-the-open-part-2/
2. Verify that sidebar sort is Location

### Expected behaviour
https://hypothes.is/a/AVOEBRmhH9ZO4OKSlQ74 appears second in the sidebar.

### Actual behaviour
It should appear about 2/3 of the way down, to match the target's location in the document.

Chrome extension 0.12.0, Windows 10.

robertknight:
1398661 extracted the code for getting a location from an annotation
into a separate module. However that function expects an Annotation but
the sort predicate function was being passed a Thread.

Consequently the predicate function was returning
Number.POSITIVE_INFINITY for every thread and so sort ordering fell back
to the default 'tie-breaker' comparator - which is the original index of
the element in the list being sorted.

Fixes #3106

### DIFF
--- a/h/static/scripts/app-controller.js
+++ b/h/static/scripts/app-controller.js
@@ -89,7 +89,9 @@ module.exports = function AppController(
         predicateFn = ['-!!message', 'message.updated'];
         break;
       case 'Location':
-        predicateFn = annotationMetadata.location;
+        predicateFn = function (thread) {
+          return annotationMetadata.location(thread.message);
+        };
         break;
     }
     $scope.sort = {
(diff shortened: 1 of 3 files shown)


robertknight:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;wMW42nVu&#x2F;314-bug-annotations-not-in-sort-order"> [BUG] Annotations not in sort order</a>

nickstenning:
LGTM.

PR #3180, opened 2016-04-06
Scroll to annotation when clicking at bottom of collapsed quote

robertknight:
Collapsed quotes can be expanded either explicitly by clicking the
'More' link on the quote or 'implicitly' by clicking anywhere at the
bottom of the quote. In that case, the clicked annotation should
be scrolled into view.

This commit separates the behavior when clicking on the toggle link vs
clicking at the bottom of the card.

In the template, the bottom area has been moved before the inline
controls so that the inline controls are above the bottom area and
receive click events first.

### DIFF
--- a/h/static/scripts/directive/excerpt.js
+++ b/h/static/scripts/directive/excerpt.js
@@ -22,13 +22,23 @@ function ExcerptController() {
   };
 
   this.toggle = function (event) {
+    // When the user clicks a link explicitly to toggle the collapsed state,
+    // the event is not propagated.
     event.stopPropagation();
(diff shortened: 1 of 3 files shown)


robertknight:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;b15FiBmg&#x2F;296-clicking-on-collapsed-quote-doesn-t-scroll-to-annotation-in-page"> Clicking on collapsed quote doesn&#39;t scroll to annotation in page</a>

nickstenning:
LGTM.

PR #3181, opened 2016-04-06
A simple JSON-LD renderer for individual annotations

nickstenning:
This is a simplistic but functional first pass at providing a W3C Web Annotation compatible view onto our annotation data.

It adds a second route,

    /api/annotations/<id>.jsonld

which renders the relevant annotation in a JSON-LD format which should hopefully be compatible with the format defined in https://www.w3.org/TR/annotation-model/.

Huge thanks go to @chdorner who did a lot of the work that made this an easy change.

### DIFF
--- a/h/api/presenters.py
+++ b/h/api/presenters.py
@@ -15,6 +15,16 @@ class AnnotationBasePresenter(object):
         self.request = request
 
     @property
+    def created(self):
+        if self.annotation.created:
+            return utc_iso8601(self.annotation.created)
+
+    @property
+    def updated(self):
+        if self.annotation.updated:
+            return utc_iso8601(self.annotation.updated)
+
+    @property
     def links(self):
         """A dictionary of named hypermedia links for this annotation."""
         # Named link generators are registered elsewhere in the code. See
@@ -27,6 +37,28 @@ class AnnotationBasePresenter(object):
             if link is not None:
                 out[name] = link
         return out
+
+    @property
+    def text(self):
+        if self.annotation.text:
+            return self.annotation.text
+        else:
+            return ''
+
+    @property
+    def tags(self):
+        if self.annotation.tags:
+            return self.annotation.tags
+        else:
+            return []
+
+    @property
+    def target(self):
+        target = {'source': self.annotation.target_uri}
+        if self.annotation.target_selectors:
+            target['selector'] = self.annotation.target_selectors
+
+        return [target]
 
 
 class AnnotationJSONPresenter(AnnotationBasePresenter):
@@ -57,30 +89,6 @@ class AnnotationJSONPresenter(AnnotationBasePresenter):
         return annotation
 
     @property
-    def created(self):
-        if self.annotation.created:
-            return utc_iso8601(self.annotation.created)
-
-    @property
-    def updated(self):
-        if self.annotation.updated:
-            return utc_iso8601(self.annotation.updated)
-
-    @property
-    def text(self):
-        if self.annotation.text:
-            return self.annotation.text
-        else:
-            return ''
-
-    @property
-    def tags(self):
-        if self.annotation.tags:
-            return self.annotation.tags
-        else:
-            return []
-
-    @property
     def permissions(self):
         read = self.annotation.userid
         if self.annotation.shared:
@@ -90,14 +98,6 @@ class AnnotationJSONPresenter(AnnotationBasePresenter):
                 'admin': [self.annotation.userid],
                 'update': [self.annotation.userid],
                 'delete': [self.annotation.userid]}
-
-    @property
-    def target(self):
-        target = {'source': self.annotation.target_uri}
-        if self.annotation.target_selectors:
-            target['selector'] = self.annotation.target_selectors
-
-        return [target]
 
 
 class DocumentJSONPresenter(object):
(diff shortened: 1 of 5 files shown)


nickstenning:
It's worth noting that I haven't updated the API documentation yet, as I'd like to get this out in the wild so I can solicit a few people's input on whether the format is correct before we document it.

judell:
Excellent, thanks!

h/api/test/presenters_test.py
```diff
@@ -29,6 +31,40 @@ def test_constructor_args(self):
         assert presenter.request == request
         assert presenter.annotation == annotation
 
+    def test_created_none_if_missing(self):
```
seanh: I think `test_created_returns_None_if_missing()` would be clearer (more explicit) and consistent with some of the other test names here (e.g. it's `test_created_uses_iso_format()` not `test_created_iso_format()`), although the test naming style in this file is a mish mash.

seanh:
LGTM, very nice and simple

h/api/test/presenters_test.py
nickstenning: Fixed!

h/api/presenters.py
```diff
@@ -28,6 +38,28 @@ def links(self):
...
+        else:
+            return []
+
+    @property
+    def target(self):
+        target = {'source': self.annotation.target_uri}
+        if self.annotation.target_selectors:
+            target['selector'] = self.annotation.target_selectors
+
+        return [target]
```
robertknight: The examples in https://www.w3.org/TR/annotation-model/ use a single object rather than an array for the case where an annotation has a single target. Should we do the same?

h/api/presenters.py
```diff
@@ -28,6 +38,28 @@ def links(self):
...
+        if self.annotation.tags:
+            return self.annotation.tags
+        else:
+            return []
+
+    @property
+    def target(self):
+        target = {'source': self.annotation.target_uri}
+        if self.annotation.target_selectors:
+            target['selector'] = self.annotation.target_selectors
```
robertknight: I'll note here that the format of some of the selectors, eg. the RangeSelector in the spec ( https://hyp.is/AVPw1LheH9ZO4OKSlnLY/www.w3.org/TR/annotation-model/ ) differs from the type of the selector with the same name in our existing annotations, so they would need to be transformed as well to be fully compliant.

h/api/test/presenters_test.py
```diff
@@ -78,6 +114,52 @@ def test_link_generators_called_with_request_and_annotation(self):
 
         dummy_link_generator.assert_called_once_with(request, annotation)
 
+    def test_text(self):
+        request = DummyRequest()
+        ann = mock.Mock(text='It is magical!')
+        presenter = AnnotationBasePresenter(request, ann)
+
+        assert 'It is magical!' == presenter.text
```
robertknight: The earlier asserts in this diff use `assert actual <predicate> expected` but here there is a switch to `assert expected <predicate> actual`. I'd prefer it was consistent if possible.

h/api/test/presenters_test.py
```diff
@@ -78,6 +114,52 @@ def test_link_generators_called_with_request_and_annotation(self):
...
+        request = DummyRequest()
+        ann = mock.Mock(tags=None)
+        presenter = AnnotationBasePresenter(request, ann)
+
+        assert [] == presenter.tags
+
+    def test_target(self):
+        request = DummyRequest()
+        ann = mock.Mock(target_uri='http://example.com',
+                        target_selectors={'PositionSelector': {'start': 0, 'end': 12}})
```
robertknight: This should be `TextPositionSelector` according to the spec

h/api/test/presenters_test.py
robertknight: Oh I see, this was existing code moved from `AnnotationJSONPresenter`.

h/api/test/presenters_test.py
robertknight: (I'll note that this is also just code moved from elsewhere)

robertknight:
I've had a look through this. Its a good start. A few notes from reading through the spec:

1. We don't currently set the `motivation` property. For replies, perhaps that should be set as `replying`
2. The `RangeSelector` format in the spec is different than our current format. What we have in H looks like the properties of a DOM `Range` object, whereas in the spec a `RangeSelector` is a combination of two other types of selector from the spec.
3. The `generator` property is not set here, but that might be a useful thing to add, eg. for debugging purposes
4. For page notes, the target should probably be rendered using the `target: {id: <URL>}` format rather than `target: {source: <URL>, ...}`

BigBlueHat:
@robertknight as the Web Annotation WG is pushing toward the next phase for the documents--much of which is driven by testing--it'd be great to see the issues you listed here addressed.

I'd be happy to help clarify anything or sort issues if you need help. @azaroth42 might also be up for some assisting. 😃 

Cheers!
🎩

h/api/presenters.py
azaroth42: Single object instead of a singleton array would be good, but either is conformant.

azaroth42:
:+1: Would love to see H. up to speed in the very near future to contribute to getting the specifications past the final hurdles.

PR #3183, opened 2016-04-07
Fix Dash compatibility issue in release script

robertknight:
In Dash, `local var=$(expr)` fails if the output of `$(expr)` contains
whitespace. The following outputs 'one two' in bash but 'one' in Dash:

    #!/bin/sh

    func () {
        local x=$(echo one two)
        echo $x
    }

    func

Quoting the subshell capture expression fixes the problem.
I didn't modify the `local version=` line because the version should never contain whitespace.


nickstenning:
LGTM.

PR #3184, opened 2016-04-07
Add a new Elasticsearch index

seanh:


seanh:
Next steps will be:

1. Worker process to read annotation create/update/delete messages off nsq and fill the new search index
2. Get reading and searching to work from the new search index instead of the old one, when postgres is on

_Updating_ annotations in Postgres still needs to be implemented as well (creating is merged and there's a PR for deleting).

h/api/search/__init__.py
```diff
@@ -40,11 +68,23 @@ def includeme(config):
...
+    # This should be used to write to the legacy search index.
+    config.add_request_method(
+        lambda r: _legacy_get_client(r.registry.settings),
+        name='legacy_es',
+        reify=True)
+
+    # request.es is a client for either the new or the legacy Elasticsearch
+    # index, depending on the 'postgres' feature flaf.
+    # This should always be used to read from the search index.
+    config.add_request_method(lambda r: _get_client_or_legacy_client(r),
```
nickstenning: `lambda r: _get_client_or_legacy_client(r)` == `_get_client_or_legacy_client`

seanh: Yeah but that doesn't use a lambda :)

seanh: Fixed

seanh:
Notes:

* We shouldn't need to make any changes to the staging or production environments. Deploying this code should simply create a new "hypothesis" index in the existing Elasticsearch host (and then start using that index, when the "postgres" flag is turned on)

* `hypothesis annotool` is broken on master

* `hypothesis reindex`, `python scripts/migrate.py` and the nipsa worker always use `request.legacy_es`, don't support working with the new index.

* It turns out that creating, updating and deleting annotations in `storage` just uses the "annotator" search index anyway, even if the postgres feature flag is on. They do not use `request.es` or `request.legacy_es`. They use the annotator subclasses in `h.api.models.elastic`, and annotator's `es` is configured using `'legacy.es.index'` in `h/api/db/__init__.py`.

* Searching for annotations on the other hand will now use the new index (and therefore get no results) when the feature flag is on

seanh:
This is ready for review now

chdorner:
LGTM.

PR #3186, opened 2016-04-07
Ignore known and expected errors during sidebar injection

robertknight:
In addition to the known issues that can prevent sidebar injection which
are represented by custom Error classes in the extension, there are also
a set of expected errors which can happen, such as the tab being closed
whilst the sidebar is being injected.

Avoid logging these and reporting them to Sentry.


nickstenning:
LGTM.

PR #3187, opened 2016-04-07
Replace node page router

chdorner:
We've been running into a few bugs on our end because of unwanted features coming from the [page client side router](https://www.npmjs.com/package/page). Since we only use a tiny subset of its features, we should run our own implementation that fits our needs.

We don't need all the client-side routing part, we only need to be able
to selectively run certain JavaScript code depending on which site the
user is at the moment.
This implementation is not as advanced as the one from the `page`
package, or what is possible with the `url-pattern` package. But we
don't need any fancy URL matching at the moment, once we do we can
revisit and maybe introduce something like `url-pattern` instead of just
a crude `document.location.pathname === path`.

### DIFF
--- a/h/static/scripts/test/page-test.js
+++ b/h/static/scripts/test/page-test.js
@@ -1,0 +1,34 @@
+'use strict';
+
+var page = require('../page');
+
+// helper to dispatch a native event to an element
+function sendEvent(element, eventType) {
+  // createEvent() used instead of Event constructor
+  // for PhantomJS compatibility
+  var event = document.createEvent('Event');
+  event.initEvent(eventType, true /* bubbles */, true /* cancelable */);
+  element.dispatchEvent(event);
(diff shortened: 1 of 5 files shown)


h/static/scripts/test/page-test.js
```diff
@@ -0,0 +1,40 @@
...
+  element.dispatchEvent(event);
+
+  return event;
+}
+
+describe('page', function () {
+  it('it adds the callback when the url path matches', function () {
+    var spy = sinon.spy();
+
+    page(document.location.pathname, function () {
```
robertknight: You can just pass the `spy` as the callback argument to `page` here.

chdorner: Fixed. Thanks

robertknight:
LGTM

PR #3188, opened 2016-04-08
Re-implement collapsible excerpt overflow state checking

robertknight:
This PR re-implements the way annotation bodies and quotes check whether they are collapsed and need to display the controls to expand/collapse, in order to fix performance issues with the `<excerpt>` and a problem reported via Sentry where the excerpt could get into a loop between showing the controls and not showing them.

----

The collapsible excerpts used for annotation quotes and bodies need to compute the height of their contents in order to determine:

 1. Whether to show controls for expanding and collapsing the excerpt
 2. The height of the excerpt when collapsed. If the height exceeds the maximum collapsed height by only a small amount, such as a 3-line quote, then the full content is shown and the excerpt is allowed to be taller than the collapsed height.

Previously, the height of the excerpt was calculated 'on-demand' whenever Angular needed to compute any properties of the view that depended on it, by calling the `overflowing()` function in the excerpt, which in turn used `Element.scrollHeight` to measure the height of the content.

This was problematic however because:
 1. `Element.scrollHeight` forces the browser to synchronously lay out the content of the element, which is expensive
 2. The excerpt could trigger an infinite cycle, if its content and the window height were _just right_ [1]

This PR fixes the cycle and performance problems by only computing the overflow state in response to changes in the excerpt's content (eg. annotation quote or body text) and certain events (window resize, media load).

~~Additionally, it defers calculation of the overflow state until the excerpt is actually in the viewport.~~ (_Update_: I removed this because determining whether the excerpt is in the viewport was causing exactly the same layout flush that I was originally trying to avoid)

----

[1] AIUI, the sequence of events which triggered the cycle is:

1. Excerpt overflow state is unknown, so the `max-height` property is not set. If the contents of the annotation list are tall enough, the window shows a scroll bar
2. The overflow is state is calculated, and the excerpt is found to be overflowing, so its `max-height` property is set, reducing the height of the excerpt.
3. If the annotation list is the right height, (2) may cause the content to now become shorter than the window height and the scrollbar will disappear
4. Since the vertical scrollbar has disappeared, this gives the excerpt's content more horizontal room
5. When the excerpt's overflow state is recalculated on the next digest cycle, it is no longer overflowing, because the extra horizontal space means that the content requires less height.
6. If the excerpt's height is now between the collapsed height and the collapsed height + threshold, it will now expand vertically
7. As a result of (6), this may cause the window to show a scrollbar and we go back to step 2 above.

### DIFF
--- a/h/static/scripts/directive/excerpt.js
+++ b/h/static/scripts/directive/excerpt.js
@@ -9,16 +9,16 @@ function ExcerptController() {
     this.animate = true;
   }
 
-  this.enabled = this.enabled || function () {
-    return true;
-  };
+  if (this.enabled === undefined) {
+    this.enabled = true;
+  }
 
   this.isExpandable = function () {
-    return this.overflowing() && this.collapse;
+    return this.overflowing && this.collapse;
   };
 
   this.isCollapsible = function () {
-    return this.overflowing() && !this.collapse;
+    return this.overflowing && !this.collapse;
   };
 
   this.toggle = function (event) {
@@ -37,7 +37,7 @@ function ExcerptController() {
   };
 
   this.showInlineControls = function () {
-    return this.overflowing() && this.inlineControls;
+    return this.overflowing && this.inlineControls;
   };
 
   this.bottomShadowStyles = function () {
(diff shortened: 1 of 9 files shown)


robertknight:
Some remaining tasks:

- [ ] - Tests for `element-visibility-watcher.js`
- [ ] - Build a Chrome extension against the prod service and do some more thorough testing of this

h/static/scripts/directive/excerpt.js
```diff
@@ -69,63 +69,126 @@ function excerpt() {
...
+        }
+        if (overflowing === ctrl.overflowing) {
+          return;
+        }
+
+        ctrl.overflowing = overflowing;
+        if (ctrl.onCollapsibleChanged) {
+         ctrl.onCollapsibleChanged({collapsible: ctrl.overflowing});
+        }
+        //scope.$digest();
```
nickstenning: Let's get rid of this cruft.

h/static/scripts/directive/test/excerpt-test.js
```diff
@@ -2,10 +2,24 @@
...
+ *
+ * This happens asynchronously after an <excerpt> is created in order to wait
+ * for Angular directives used by the <excerpt>'s content to fully resolve.
+ *
+ * @return {Promise}
+ */
+function waitForLayout(element) {
+  element.scope.$digest();
+  return new Promise(function (resolve) {
+    window.requestAnimationFrame(resolve);
```
nickstenning: In some places we do `require('raf')` for this. In other places we use `$$rAF` from angular internals. What's the status of the polyfill? Can we (perhaps in another PR) make this consistent across the client application?

h/static/scripts/directive/test/excerpt-test.js
```diff
@@ -19,17 +33,29 @@ describe('excerpt directive', function () {
...
-    attrs = assign(defaultAttrs, attrs);
+    attrs = Object.assign(defaultAttrs, attrs);
     return util.createDirective(document, 'excerpt', attrs, {}, content);
   }
 
   function height(el) {
     return el.querySelector('.excerpt').offsetHeight;
   }
 
+  var defaultRAF = window.requestAnimationFrame;
```
nickstenning: Is there a reason you wouldn't use `sinon.stub` for this?

h/static/scripts/directive/test/util.js
```diff
@@ -96,8 +96,8 @@ function createDirective(document, name, attrs, initialScope, initialHtml, opts)
   opts = opts || {};
   opts.parentElement = opts.parentElement || document.body;
 
-  // create a template consisting of a single element, the directive
-  // we want to create and compile it
+  // Create a template consisting of a single element, the directive
```
nickstenning: Feel free to pull obviously-safe cleanups like this into their own commits and push them directly to master to avoid noise in pull requests.

h/static/scripts/directive/excerpt.js
```diff
@@ -69,63 +69,126 @@ function excerpt() {
     controllerAs: 'vm',
     link: function (scope, elem, attrs, ctrl) {
```
nickstenning: It seems to me that this link function is getting a bit unwieldy. Could you encapsulate the functionality of the overflow checking in a component that could be independently unit tested, rather than simply testing the functional behaviour of a fairly large and inscrutable blob of code?

nickstenning:
Would suggest pulling 1ceea677762620a9ad833ec9f188d3f10ff9404e into its own PR.

nickstenning:
I've made a bunch of comments on the code, but I'm also unable to verify that this actually works as intended. Creating annotations with long bodies doesn't seem to reliably result in them being collapsed as expected. Reloading the page also doesn't necessarily result in the annotation being collapsed. Even resizing the page or sidebar doesn't seem to reliably trigger a collapse.

Here's a couple of examples I can reproduce reliably:

---

1. Create an annotation with several paragraphs of text in the body. My example has about 500 words in 6 paragraphs.
2. Hit "Post to public".

Result: annotation is truncated but the drop shadow and "More" link aren't shown.

---

1. Create an annotation with several paragraphs of text in the body. My example has about 500 words in 6 paragraphs.
2. Hit "Post to public".
3. Reload the page.
4. Hit "Edit".
5. Without making any changes, hit "Post to public".

Result: the annotation appears correctly truncated on initial page load, but after step 5 isn't truncated at all.

---

1. Make a reply to an annotation with several paragraphs of text in the body. My example has about 500 words in 6 paragraphs.
2. Hit "Post to public".
3. Reload the page.
4. Expand the reply thread.

Result: the reply isn't truncated at all.

h/static/scripts/directive/test/excerpt-test.js
robertknight: `window.requestAnimationFrame` is now available in all browsers that we support (IE >= 10, Safari >= 6, Chrome/Firefox for a very long time), so we don't need to polyfill it any more. Having said that, having an adapter makes for cleaner tests. I'll have a think on this.

robertknight:
Thanks for testing @nickstenning . I've revised and simplified the PR. _Mea culpa_, I was always testing by reloading H on a page or the stream and checking whether excerpts were collapsed and hadn't checked switching between editing/viewing modes properly.

The issues you reported in that last comment were because the `<excerpt>` was grabbing a reference to the content element once when initialized and then always testing the height of that element thereafter. When switching between the enabled and disabled states however, the content element changes, so the excerpt ended up testing the height of an element which was no longer in the DOM. The fix is simply to get a fresh reference to the content element whenever the overflow state is recalculated.

Another issue was the typing of the `enabled` property. A function was used in some places and a boolean in others. Now it is an optional boolean which is assigned a default value in the same way as the `animated` and `collapsed`properties.

h/static/scripts/directive/excerpt.js
robertknight: Had a go at this, looks like the answer is yes. Will revise accordingly.

robertknight:
I have moved the logic for checking for overflow and generating the appropriate CSS content styling has been moved into the `ExcerptOverflowMonitor` helper class.

This was quite an interesting exercise in the merits of unit (contract + collaboration) vs. functional testing. On the one hand, the tests are now easier to debug if they fail and it was easier to write more comprehensive tests for the overflow checking. On the other hand, they are now also more tightly coupled to the implementation - if for some reason it was ever necessary to change that substantially (again) in future, more work would be involved.

nickstenning:
We're closing in on this one :)

I can still reproduce my last failing case from the original comment.

nickstenning:
Also, if I create an annotation with just enough text such that I can get it to trigger the truncation when the sidebar is narrow, but resizing the sidebar big enough makes it go away, I see some pretty odd behaviour when it transitions between the two. Specifically:

- The drop-shadow on the annotation seems to appear substantially before the More/Less link.
- Sometimes when resizing down, the drop-shadow will appear, but the More/Less link never appears until I trigger another digest cycle by mousing over the annotation card.

robertknight:
Both of the above issues should now be fixed:

1. The `<excerpt>` now updates when a parent element is shown or hidden via `ng-show` or `ng-hide`.
2. The More/Less link state is now correctly updated as soon as the excerpt's collapsibility changes.

The issue for (2) was that when the overflow state changed, a digest was triggered for the excerpt alone. The usual Angular solution to notify it about changes outside the framework is via `$scope.$apply`. That unfortunately triggers a full digest cycle which is expensive if a bunch of excerpts all update their overflow state at the same time, and unnecessary in the common case where an excerpt is not overflowing at all. Here I've opted to explicitly call `$scope.$digest` in the affected `<annotation>` when it becomes collapsible.

nickstenning:
Ok.

- replies: fixed
- more/less button and drop shadow out of sync: fixed
- annotation card doesn't resize until you mouse over it: **still broken**

nickstenning:
This last problem exists on master too, so I'm going to merge this.

PR #3189, opened 2016-04-08
Replace homebrew background workers with Celery

nickstenning:
Until now, we've used `h.worker`, a home-grown worker runner which reads tasks from NSQ. It has served its purpose, but is increasingly inflexible, as it currently assumes a process-per-job-type worker model which makes deploying new job types tricky, and consumes more resources as we add more job types, even if those jobs are infrequently used.

Rather than continuing to develop `h.worker`, it makes sense to switch to the well-known and actively developed Celery project to run background tasks. This allows us to configure tasks in a straightforward fashion (basically just functions) and dispatch to them easily from anywhere in the code.

In this pull request, Celery is configured to use Redis as a message broker. We already use Redis for session storage in production, and this PR configures Celery to use a different Redis "database" so that keys do not clash with those from `pyramid_redis_sessions`.

It's worth noting that Redis is *not* a message queue (it does not provide reliable persistence) but at the moment we only use background jobs for two things:

1. Sending emails.
2. Processing "NIPSA" jobs where we have to find all of a user's annotations and flag/unflag them.

For these tasks, it seemed (to me) to be reasonable to reuse Redis as a message broker rather than requiring that we set up RabbitMQ (the normal message broker used with Celery) in production. I think we should have a conversation about whether this is the right choice at this stage, because we have a number of options:

1. Mandate (or at least document) the use of RabbitMQ in dev and prod. (Maximising dev/prod parity.)
2. Use redis in dev (for now) and deploy RabbitMQ in production. (Maximising ease of setup in dev, but ensuring persistent messages in prod.)
3. Use redis in both (for now). (Simplest deployment.)

---

Here's a quick overview in a little more detail of the changes in this branch:

- `hypothesis-worker` is gone and replaced with `hypothesis-celery` which is a simple wrapper around the default Celery CLI.
- `h.mailer` and `h.nipsa.worker` are rewritten to define Celery tasks which can be dispatched asynchronously from anywhere in the application.
- `h.notification.worker` has been removed: responsibility for generating relevant notification emails has moved into the web worker process, and those emails are sent asynchronously by the `h.mailer` Celery task.
- When a Celery worker boots, a fake request is created and attached to the Celery application, much as `h.worker` did.
- Errors handling celery tasks are reported to Sentry.
- Feature flag state is flushed before handling each task.
- Task success results in the request transaction being committed, and task failure results in the request transaction being aborted.

---

Although the diff says this branch adds code, it's worth noting that most of the additions (in relative terms) are in test code:

```
# Non-test changes
$ git diff master --name-only | egrep -v 'test|docs' | xargs git diff master --stat --
...
10 files changed, 155 insertions(+), 180 deletions(-)

# Test and docs changes
$ git diff master --name-only | egrep 'test|docs' | xargs git diff master --stat --
...
8 files changed, 359 insertions(+), 181 deletions(-)
```

---

Questions and issues to resolve before merge:

- [x] Should we enable `CELERY_ACKS_LATE` for consistency with default NSQ behaviour. Probably.
- [x] Are we going to go with option 1, 2, or 3 w.r.t. Redis and RabbitMQ?


nickstenning:
My proposal is that we deploy against RabbitMQ in production, and suggest people use Redis in development, until such time as we remove our dependency on NSQ, and then we switch to using Rabbit everywhere.

For what it's worth, getting Rabbit running in development is super easy:

```
docker run -d --hostname rabbit --name rabbit -p 5672:5672 -p 15672:15672 rabbitmq:3-management
```

And then run celery with

```
BROKER_URL=amqp://guest:guest@docker:5672//
```

h/nipsa/worker.py
```diff
@@ -31,40 +34,26 @@ def remove_nipsa_action(index, annotation):
...
-
-    annotations = helpers.scan(client=client, query=query)
-    actions = [action_func(index, a) for a in annotations]
-    helpers.bulk(client=client, actions=actions)
-
-
-def worker(request):
-    """Worker function for NIPSA'ing and un-NIPSA'ing users.
+def bulk_update_annotations(client, query, action):
+    """Update annotations for a user with the given NIPSA state."""
```
chdorner: Would be nice to have a bit more documentation here, especially in regards to the `action` parameter taking a function which should return an ElasticSearch action. I can see this becoming confusing after not looking at this code for a few months.

docs/hacking/install.rst
```diff
@@ -122,23 +124,25 @@ system that Docker can be installed on:
...
+   the `official PostgreSQL image <https://hub.docker.com/_/postgres/>`_, the
+   `official Redis image <https://hub.docker.com/_/redis/>`_,
    and our custom
    `Elasticsearch with ICU image <https://hub.docker.com/r/nickstenning/elasticsearch-icu/>`_:
 
    .. code-block:: bash
 
       docker run -d --name nsqd -p 4150:4150 -p 4151:4151 nsqio/nsq /nsqd
       docker run -d --name postgres -p 5432:5432 postgres
+      docker run -d --name redis -p 6379:6379 redis
```
chdorner: We could also encourage people to use the `alpine` tag of the official redis images.

chdorner:
This is looking very nice!

In regards to the open questions, here are my two cents:

* I would like to see us not using Redis as a message queue in production/staging if we can avoid it (for the reasons you mentioned above).
* We were briefly talking about the problems of RabbitMQ, and @nickstenning said that most issues people have with running RabbitMQ is due to not setting it up correctly. I would like to hear a bit more about these issues and how to correctly setting it up to not run into these issues.
* In the interest of not needing to run a lot of containers locally while developing, I'm fine with using redis until we stop using NSQ, at that point we should also start encouraging people to start using RabbitMQ locally.

In regards to acks_late, if that's the behavior that we have from NSQ, then we should probably enable it, at least for the time being to keep the moving parts to a minimum. We can always revisit it later on.

Some other questions I have after skimming through the Celery docs:

* Celery doesn't support priorities at the moment (because RabbitMQ doesn't implement them yet). Could that lead to problems for us when we generate a lot of work for one Celery worker type that delays the execution of other work by an unacceptable amount of time? For example the new Elasticsearch indexing task could take a very long time when we decide to re-index all of our Postgres annotations, does that mean that (welcome) emails would not be sent out for that period?

docs/hacking/install.rst
nickstenning: My feeling here is that soon enough the `alpine` tag will become the official `latest`. Let's wait for the upstream maintainers to make that call.

nickstenning:
> - I would like to see us not using Redis as a message queue in production/staging if we can avoid it (for the reasons you mentioned above).

Agreed. Let's deploy Rabbit. We could also deploy against Redis until such time as we have Rabbit up and running (to avoid having to keep this branch open for longer). What do you reckon?

> - We were briefly talking about the problems of RabbitMQ, and @nickstenning said that most issues people have with running RabbitMQ is due to not setting it up correctly. I would like to hear a bit more about these issues and how to correctly setting it up to not run into these issues.

Yep, let's chat about this in Slack. IME it mainly comes down to a) setting memory and disk limits correctly and b) understanding what happens when consumers can't keep up with producers -- i.e. configuring an appropriate response to backpressure. We're not going to be needing a clustered RabbitMQ for a while, so that eliminates a whole host of potential complexity.

> Celery doesn't support priorities at the moment (because RabbitMQ doesn't implement them yet). 

It depends what you mean by 'priorities'. The standard way of implementing priorities in Celery is to have multiple named queues, and dedicated workers for each. For example, we'd probably have an "indexer" queue and dedicated workers to process search index update tasks, and then use the default "celery" queue for job workers. You can read more about that [in the Celery routing documentation](http://docs.celeryproject.org/en/latest/userguide/routing.html).

Oh, and FWIW, RabbitMQ [does actually support priority queues since v3.5.0](https://www.rabbitmq.com/priority.html) but AFAIK there's no Celery support for that yet.

So, I'm going to enable CELERY_ACKS_LATE in this PR, just so we're not changing semantics of job delivery, and then we can work out how we want to get this into production.

seanh:
I've looked through this and it all looks great to me. It's gonna be nice to have one worker process to run in dev instead of several.

I don't know a whole lot about Celery, but will get the docs and have a read when I get a chance.

Agree that we should use a persistent queue in production and plan to move towards using the same queue in dev. I'll leave the particular choice or worker up to you two as it seems to mostly come down to server maintenance characteristics.

I'm happy for this to be merged once @chdorner is happy.

PR #3190, opened 2016-04-08
Delete annotations from PostgreSQL

seanh:
Delete annotations from both the legacy Elasticsearch "database" and the
new PostgreSQL database when the 'postgres' feature flag is on.

### DIFF
--- a/h/api/storage.py
+++ b/h/api/storage.py
@@ -189,8 +189,12 @@ def delete_annotation(request, id):
     :param id: the annotation id
     :type id: str
     """
-    annotation = models.elastic.Annotation.fetch(id)
-    annotation.delete()
+    if _postgres_enabled(request):
+        annotation = fetch_annotation(request, id, _postgres=True)
+        request.db.delete(annotation)
+
+    legacy_annotation = fetch_annotation(request, id, _postgres=False)
+    legacy_annotation.delete()
 
 
 def expand_uri(request, uri):
--- a/h/api/test/storage_test.py
+++ b/h/api/test/storage_test.py
@@ -538,12 +538,83 @@ class TestCreateAnnotation(object):
             }
         }
 
-    @pytest.fixture
-    def fetch_annotation(self, request):
-        patcher = patch('h.api.storage.fetch_annotation', autospec=True)
-        fetch_annotation = patcher.start()
-        request.addfinalizer(patcher.stop)
-        return fetch_annotation
+
+@pytest.mark.usefixtures('fetch_annotation')
+class TestDeleteAnnotationLegacy(object):
+
+    """Tests for delete_annotation() when the postgres feature flag is off."""
+
+    def test_it_only_fetches_the_legacy_annotation(self, fetch_annotation):
+        request = self.mock_request()
+
+        storage.delete_annotation(request, "test_id")
+
+        fetch_annotation.assert_called_once_with(
+            request, "test_id", _postgres=False)
+
+    def test_it_deletes_the_legacy_annotation(self, fetch_annotation):
+        storage.delete_annotation(self.mock_request(), "test_id")
+
+        fetch_annotation.return_value.delete.assert_called_once_with()
+
+    def mock_request(self):
+        request = DummyRequest()
+        request.feature = mock.Mock(return_value=False)
+        return request
+
+
+@pytest.mark.usefixtures('fetch_annotation')
+class TestDeleteAnnotation(object):
+
+    def test_it_fetches_the_annotation(self, fetch_annotation):
+        request = self.mock_request()
+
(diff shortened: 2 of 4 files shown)


h/api/storage.py
```diff
@@ -189,8 +189,12 @@ def delete_annotation(request, id):
     :param id: the annotation id
     :type id: str
     """
-    annotation = models.elastic.Annotation.fetch(id)
-    annotation.delete()
+    if _postgres_enabled(request):
+        annotation = fetch_annotation(request, id, _postgres=True)
+        request.db.delete(annotation)
+
+    legacy_annotation = fetch_annotation(request, id, _postgres=False)
```
seanh: Note that I changed the legacy behaviour to call `fetch_annotation()` instead of calling `models.elastic.Annotation.fetch(id)`. I don't think this makes any difference, it's just more consistent and less duplication. Passing `_postgres=False` _is_ necessary here because the postgres feature flag might be true at this point but we want to fetch from Es not Pg.

h/api/storage.py
```diff
@@ -189,8 +189,12 @@ def delete_annotation(request, id):
     :param id: the annotation id
     :type id: str
     """
-    annotation = models.elastic.Annotation.fetch(id)
-    annotation.delete()
+    if _postgres_enabled(request):
+        annotation = fetch_annotation(request, id, _postgres=True)
```
seanh: Passing `_postgres=True` isn't strictly necessary here as the postgres feature flat will always be on when we hit this line so `fetch_annotation()` would fetch from Pg anyway, but no harm in being explicit.

h/api/test/storage_test.py
```diff
@@ -538,12 +538,83 @@ def annotation_data(self):
...
+class TestDeleteAnnotationLegacy(object):
+
+    """Tests for delete_annotation() when the postgres feature flag is off."""
+
+    def test_it_only_fetches_the_legacy_annotation(self, fetch_annotation):
+        request = self.mock_request()
+
+        storage.delete_annotation(request, "test_id")
+
+        fetch_annotation.assert_called_once_with(
```
seanh: `assert_called_once_with()` not only ensures that it fetched the annotation from Es but also that it did not fetch it from Pg.

h/api/test/storage_test.py
```diff
@@ -538,12 +538,83 @@ def annotation_data(self):
...
+class TestDeleteAnnotation(object):
+
+    def test_it_fetches_the_annotation(self, fetch_annotation):
+        request = self.mock_request()
+
+        storage.delete_annotation(request, "test_id")
+
+        assert fetch_annotation.call_args_list[0] == mock.call(request,
+                                                               "test_id",
+                                                               _postgres=True)
```
seanh: Once the Es code is gone this can become an `assert_called_once_with()`.

h/api/test/storage_test.py
```diff
@@ -538,12 +538,83 @@ def annotation_data(self):
...
+
+    def test_it_deletes_the_annotation(self, fetch_annotation):
+        request = self.mock_request()
+        first_return_value = mock.Mock()
+        second_return_value = mock.Mock()
+        fetch_annotation.side_effect = [
+            first_return_value,
+            second_return_value,
+        ]
```
seanh: We need to make sure it deletes the Postgres annotation not the Es one so it's necessary to distinguish between `fetch_annotation()`'s first and second return values.

h/api/test/storage_test.py
```diff
@@ -538,12 +538,83 @@ def annotation_data(self):
...
+
+        request.db.delete.assert_called_once_with(first_return_value)
+
+    def test_it_fetches_the_legacy_annotation(self, fetch_annotation):
+        request = self.mock_request()
+
+        storage.delete_annotation(request, "test_id")
+
+
+        assert fetch_annotation.call_args == mock.call(request,
```
seanh: `call_args` is always the last call.

h/api/views.py
```diff
@@ -228,7 +228,10 @@ def delete(annotation, request):
     # process the delete event. For example, the streamer needs to know the
     # target URLs of the deleted annotation in order to know which clients to
     # forward the delete event to.
-    _publish_annotation_event(request, annotation, 'delete')
+    _publish_annotation_event(
+        request,
+        AnnotationJSONPresenter(request, annotation).asdict(),
```
seanh: `annotation` is sometimes a Pg annotation not an Es one now, and Pg annotations aren't JSON serializable, so present the annotation and call the presenter's `asdict()`, this works with both Pg and Es annotations.

seanh:
This is ready for review.

chdorner:
LGTM

PR #3191, opened 2016-04-08
Clean up patch boilerplate

nickstenning:
So... Sean made a comment in Slack the other day about [monkeypatch](https://pytest.org/latest/monkeypatch.html), which caused me to look into our tests...  and *boy* had things gotten out of hand.

The same boilerplate has ended up replicated in many, MANY, places in our test code.

This PR extracts the common usecase into a helper called `patch`, which usually reduces the amount of boilerplate needed by a factor of 5 or so.

There are probably more places where we need to examine why we're patching SO MANY THINGS in order to write tests (because that's usually a sign of poor factoring), but at least this stops the madness.

### DIFF
--- a/h/accounts/test/__init___test.py
+++ b/h/accounts/test/__init___test.py
@@ -152,15 +152,10 @@ def test_authenticated_user_calls_get_user(get_user):
 
 
 @pytest.fixture
-def util(request):
-    patcher = mock.patch('h.accounts.util')
-    module = patcher.start()
-    request.addfinalizer(patcher.stop)
-    return module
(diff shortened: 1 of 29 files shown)


chdorner:
LGTM. There are merge conflicts happening, will merge once they're fixed.

nickstenning:
Fixed!

seanh:
> There are probably more places where we need to examine why we're patching SO MANY THINGS in 
> order to write tests (because that's usually a sign of poor factoring)

Yep, this has been worrying me as well

PR #3194, opened 2016-04-09
Remove routes_mapper fixture and use configurator routes

nickstenning:
Some of our tests, mostly views tests, need the pyramid application to be configured with a routes mapper so that `request.route_path` and `request.route_url` work as expected.

We had a clunky fixture, `routes_mapper`, which provided a dummy route mapper, but it's actually cleaner and easier to use the proper Pyramid routes mapper, which is set up by `pyramid.testing.setUp`. All that is required is that tests explicitly defined the routes they expect to exist using `config.add_route`.

This commit removes `routes_mapper` and also cleans up the fixture definitions so that we're not calling `pyramid.testing.setUp` for every test unconditionally. In my testing this knocks 5-6 seconds (25-30%) off test suite runtime.

### DIFF
--- a/h/accounts/test/views_test.py
+++ b/h/accounts/test/views_test.py
@@ -34,9 +34,10 @@ class FakeSubscription(object):
 
 class FakeUser(object):
     def __init__(self, **kwargs):
-        self.activation = None
-        for k in kwargs:
-            setattr(self, k, kwargs[k])
+        defaults = {'activation': None, 'email': None, 'password': None}
+        defaults.update(kwargs)
+        for k, v in defaults.items():
+            setattr(self, k, v)
 
 
 class FakeSerializer(object):
@@ -79,7 +80,7 @@ def mock_flash_function():
                                 return_value=None)
 
 
-@pytest.mark.usefixtures('routes_mapper')
+@pytest.mark.usefixtures('routes')
 class TestAuthController(object):
 
     def test_post_redirects_when_logged_in(self, authn_policy):
@@ -207,9 +208,14 @@ class TestAuthController(object):
 
         assert result.headers['x-erase-fingerprints'] == 'on the hob'
 
+    @pytest.fixture
+    def routes(self, config):
+        config.add_route('forgot_password', '/forgot')
+        config.add_route('index', '/index')
+        config.add_route('stream', '/stream')
 
-@pytest.mark.usefixtures('routes_mapper',
-                         'session')
+
+@pytest.mark.usefixtures('session')
 class TestAjaxAuthController(object):
 
     def test_login_returns_status_okay_when_validation_succeeds(self):
@@ -303,7 +309,7 @@ class TestAjaxAuthController(object):
 @pytest.mark.usefixtures('activation_model',
                          'authn_policy',
                          'mailer',
-                         'routes_mapper')
+                         'routes')
 class TestForgotPasswordController(object):
 
     def test_post_returns_form_when_validation_fails(self):
@@ -398,8 +404,14 @@ class TestForgotPasswordController(object):
         with pytest.raises(httpexceptions.HTTPFound):
             views.ForgotPasswordController(request).get()
 
+    @pytest.fixture
+    def routes(self, config):
+        config.add_route('index', '/index')
+        config.add_route('reset_password', '/reset')
+        config.add_route('reset_password_with_code', '/reset-with-code')
 
-@pytest.mark.usefixtures('routes_mapper')
+
+@pytest.mark.usefixtures('routes')
 class TestResetPasswordController(object):
 
     def test_post_returns_form_when_validation_fails(self):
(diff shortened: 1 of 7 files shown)


h/accounts/test/views_test.py
```diff
@@ -34,9 +34,10 @@ def __init__(self, type_, active):
 
 class FakeUser(object):
     def __init__(self, **kwargs):
-        self.activation = None
-        for k in kwargs:
-            setattr(self, k, kwargs[k])
+        defaults = {'activation': None, 'email': None, 'password': None}
```
nickstenning: These changes are to simplify the changed `ProfileController` tests later on in this file.

h/accounts/test/views_test.py
```diff
@@ -775,111 +801,102 @@ def test_get_when_logged_in_already_logged_in_to_different_account(self):
             "You're already signed in to a different account")
 
 
-@pytest.mark.usefixtures('routes_mapper')
 class TestProfileController(object):
 
-    def test_post_400s_with_no_formid(self):
-        user = FakeUser()
-        request = DummyRequest(post={}, authenticated_user=user)
+    @pytest.fixture(autouse=True)
```
nickstenning: We have to call `config.begin(request)` for these tests, unfortunately, so that `pyramid.threadlocal.get_current_request` returns the correct request. This is because `deform_jinja2` calls `get_current_request`...

h/accounts/views.py
```diff
@@ -141,6 +141,11 @@ def _logout(self):
                renderer='json')
 class AjaxAuthController(AuthController):
 
+    def __init__(self, request):
```
nickstenning: This is added so that `AjaxAuthController` doesn't do a bunch of setup (from the `AuthController` constructor) that is not necessary. It means we don't need to register routes for `TestAjaxAuthController`.

h/test/tweens_test.py
```diff
@@ -48,6 +48,7 @@ def test_tween_csp_uri():
     request = DummyRequest()
     request.registry.settings.update({
         'csp.enabled': True,
+        'csp.report_only': False,
```
nickstenning: The values in `request.registry.settings` are set only once per session, so these tests were order dependent. Setting `csp.report_only` explicitly removes the order-dependence.

nickstenning:
Maybe @seanh you could take a look at this tomorrow?

seanh:
Happy to

h/accounts/test/views_test.py
```diff
@@ -79,9 +80,14 @@ def mock_flash_function():
...
 
 
-@pytest.mark.usefixtures('routes_mapper')
 class TestAuthController(object):
 
+    @pytest.fixture(autouse=True)
+    def routes(self, config):
+        config.add_route('forgot_password', '/forgot')
+        config.add_route('index', '/index')
+        config.add_route('stream', '/stream')
```
seanh: Any reason you used a class-level autouse fixture, rather than usefixtures?

seanh: P.S. When I've been defining fixtures and helper methods in classes I've been putting them at the bottom of the class (which seemed to match the style I'd seen elsewhere in our code of putting fixtures and helpers at the bottom of modules.

seanh:
Not sure if I like names like `req` instead of `request`. I like the idea of it being a fixture though, I've tended to define a request helper method in test classes and then had tests to `request = self.request()`, but using a fixture is nicer.

seanh:
I suspect there are some places in the test where we're mocking Pyramid URL-generating methods on the request (`route_url()` etc) where we could be using a fixture like theses ones instead. I can look into that if you want..

h/accounts/test/views_test.py
seanh: The reason I ask about autouse fixtures is that you _can_ do this with class-level fixtures:

```python
@pytest.mark.usefixtures('my_fixture')
class TestExample:

    @pytest.fixture
    def my_fixture(self):
        ...

    ...
```

In my mind that has the advantage that `usefixtures` can also be used to activate a fixture for all methods of a class when that fixture is defined outside of that class. For example, any test module that has a fixture that more than one of its classes can share, can define that fixture at module level.

Always using `usefixtures` to activate a fixture for all methods of a class, wherever the fixture is defined, seems more consistent than sometimes using `autouse` and sometimes `usefixtures`.

(Also I've already done it in the above style in some existing test classes.)

seanh:
LGTM, I'd like a response to the autouse vs usefixtures thing and I think moving fixtures and helpers to the bottom of classes would be good (for consistency if nothing else). Otherwise happy to merge

h/accounts/test/views_test.py
nickstenning: > Any reason you used a class-level autouse fixture, rather than usefixtures?

No particular reason. It just seems to me to be a more obvious way to use fixtures than having to declare the fixture in one place and then declare that it is used later.

Happy to switch to using `usefixtures` everywhere.

nickstenning:
> Not sure if I like names like req instead of request.

We don't have an option. `request` is a defined fixture name, and refers to the "request" for fixtures made by a test function. We can't declare a fixture called `request` without half of our other fixtures breaking.

nickstenning:
I've replaced the use of `autouse` in most of the cases. In `h/test/admin_test.py` I've left it in because it's clearer and simpler to have one autoused module-level fixture.

PR #3197, opened 2016-04-10
Add copy to clipboard functionality.

sheetaluk:
Add basic copy to clipboard functionality so user can use a button
click to copy a direct link to his clipboard, when supported by the
browser.

* Refactor annotation share dialog into a component.
* Add tests for annotation share dialog.


h/static/scripts/app.js
```diff
@@ -173,6 +173,7 @@ module.exports = angular.module('h', [
   .directive('signinControl', require('./directive/signin-control'))
   .directive('sortDropdown', require('./directive/sort-dropdown'))
   .directive('topBar', require('./directive/top-bar'))
+  .directive('annotationShareDialog', require('./directive/annotation-share-dialog'))
```
robertknight: This list should be sorted alphabetically

h/static/scripts/directive/annotation-share-dialog.js
```diff
@@ -0,0 +1,42 @@
...
+    bindToController: true,
+    controllerAs: 'vm',
+    restrict: 'E',
+    template: require('../../../templates/client/annotation_share_dialog.html'),
+    controller: ['$timeout', '$scope', function ($timeout, $scope) {
+      this.copyToClipboard = function (event) {
+        var $container = angular.element(event.currentTarget).parent();
+        var shareLinkInput = $container.find('input')[0];
+
+        $scope.showCopyToClipboardMessage = true;
```
robertknight: Rather than maintain two separate pieces of state here (the message and whether to show a message), you could just make `copyToClipboardMessage` null when there is no message.

h/static/scripts/directive/annotation.js
```diff
@@ -655,9 +655,6 @@ function AnnotationController(
   vm.share = function(event) {
     var $container = angular.element(event.currentTarget).parent();
     $container.addClass('open');
-    var shareLinkInput = $container.find('input')[0];
```
robertknight: This logic ensured that the link was selected, ready to copy via pressing Ctrl/Cmd+C as soon as the link appeared. Was it an explicit decision to drop that behavior?

h/static/scripts/directive/test/annotation-share-dialog-test.js
```diff
@@ -0,0 +1,101 @@
+'use strict';
+
+var angular = require('angular');
```
robertknight: The format of imports as the top of JS files should be:

```js
'use strict';

<external imports>

<local imports>

<declarations>
```

With a blank line between each section.

h/static/scripts/directive/test/annotation-share-dialog-test.js
```diff
@@ -0,0 +1,101 @@
...
+        }
+      );
+    });
+
+    afterEach(function () {
+      stub.restore();
+    });
+
+    it('sets success class on successful copy to clipboard', function () {
+      var copyBtn = element.find('.annotation-share-dialog-link__btn');
```
robertknight: I suggest extracting these two lines into a helper function (eg. `getCopyBtn()`) which you can re-use in each test.

h/static/styles/annotation-share-dialog.scss
```diff
@@ -0,0 +1,130 @@
+.share-dialog-wrapper.open .annotation-share-dialog {
+  display: flex;
+  flex-direction: column;
+}
+
+.annotation-share-dialog {
+  display: none;
```
robertknight: Something to think about here is how we ensure visual consistency between this and any other dialogs that we might have in future. For example, should the styling of the dialog outline here match the styling of the groups dropdown and sign-in dropdown?

h/static/styles/annotation-share-dialog.scss
```diff
@@ -0,0 +1,130 @@
...
+  z-index: 1;
+  background: $white;
+  border: 1px solid $gray-lighter;
+  border-radius: 2px;
+  width: 200px;
+  font-size: $body1-font-size;
+  cursor: default;
+  box-shadow: 0px 0px 4px 0px rgba(0,0,0,0.15);
+
+  &:after, &:before {
```
robertknight: I suggest adding a note here to explain what the `:after` and `:before` pseudo-selectors are for.

h/static/styles/annotation-share-dialog.scss
```diff
@@ -0,0 +1,130 @@
...
+    border-width: 6px;
+    margin-right: -6px;
+  }
+}
+
+.annotation-share-dialog-msg {
+  color: $gray-light;
+  margin: -5px 10px 10px 10px;
+  line-height: 15px;
+  font-size: 11px;
```
robertknight: The font size and line height should be defined in the typography palette in `variables.scss` if possible

h/static/styles/annotation-share-dialog.scss
```diff
@@ -0,0 +1,130 @@
...
+  font-size: 20px;
+  cursor: pointer;
+}
+
+.annotation-share-dialog-target__icon:hover {
+  color: $dg-1;
+}
+
+.annotation-share-dialog-link__feedback {
+  font-size: 11px;
```
robertknight: This should also use the typography and color palettes if possible.

h/static/styles/app.scss
```diff
@@ -28,6 +28,7 @@ $base-line-height: 20px;
 @import './thread';
 @import './top-bar';
 @import './loggedout-message';
+@import './annotation-share-dialog';
```
robertknight: These should be alphabetically sorted.

robertknight:
I see this error in the dev tools when opening the share dialog:

```
angular.js:13236 TypeError: shareLinkInput.select is not a function
    at AnnotationController.vm.share (annotation.js:661)
    at angular.js:14652
    at expensiveCheckFn (angular.js:15076)
    at callback (angular.js:24546)
    at Scope.$eval (angular.js:16820)
    at Scope.$apply (angular.js:16920)
    at HTMLButtonElement.<anonymous> (angular.js:24551)
    at defaultHandlerWrapper (angular.js:3398)
    at HTMLButtonElement.eventHandler (angular.js:3386)
```

h/static/scripts/directive/annotation.js
```diff
@@ -655,8 +655,8 @@ function AnnotationController(
...
-    shareLinkInput.focus();
+    var shareLinkInput = $container.find('.annotation-share-dialog-link__text');
```
robertknight: The logic for selecting the input field should live in the share dialog, rather than in the parent annotation card.

robertknight: CSS classes that exist for use by code rather than just being presentational should have a `js-` prefix or `is-` prefix for classes to indicate a certain state. eg. `js-input` so it is obvious that there is code relying on them. So you would have eg. `<input class="some-dialog__label js-input">` I'm afraid that's a part of the not-yet-written-down style guide for CSS.

h/static/styles/annotation-share-dialog.scss
robertknight: What was the status on this? I think the 11px font size was from the palette wasn't it?

h/static/scripts/directive/annotation-share-dialog.js
```diff
@@ -0,0 +1,40 @@
...
+
+var angular = require('angular');
+
+module.exports = function () {
+  return {
+    bindToController: true,
+    controllerAs: 'vm',
+    restrict: 'E',
+    template: require('../../../templates/client/annotation_share_dialog.html'),
+    controller: ['$timeout', '$scope', function ($timeout, $scope) {
```
robertknight: It should be possible to use an `// @ngInject` comment here rather than using the array syntax, eg.

```js
// @ngInject
controller: function ($scope) {
...
}
```

The `$timeout` variable is no longer used.

h/static/scripts/directive/annotation-share-dialog.js
```diff
@@ -0,0 +1,40 @@
...
+        var shareLinkInput = $container.find('input')[0];
+
+        try {
+          shareLinkInput.select();
+          document.execCommand('copy');
+
+          $scope.copyToClipboardMessage = 'Link copied to clipboard!';
+          $scope.copyToClipboardMessageClass = 'annotation-share-dialog-link--success';
+        } catch (ex) {
+          $scope.copyToClipboardMessage = 'Please use Ctrl+C to copy link.';
```
robertknight: Unfortunately this message does not fit in the available space, at least in my testing under Chrome. We also need to take account of the fact that it doesn't make sense on mobile devices.

I can't immediately think of a good short message which is applicable to a reasonable range of desktop and mobile platforms though. "Select and copy to share"?

h/static/styles/annotation-share-dialog.scss
```diff
@@ -0,0 +1,132 @@
...
+  border: 1px solid #d3d3d3;
+  border-radius: 2px;
+}
+
+.annotation-share-dialog-link--success {
+  padding: 0px 11px 0px 11px;
+}
+
+.annotation-share-dialog-link--fail {
+  padding: 0px 24px 0px 24px;
```
robertknight: Why is the padding around the label different for the two different states?

h/static/styles/annotation.scss
```diff
@@ -257,6 +175,26 @@ $annotation-card-left-padding: 10px;
...
+}
+
+.share-dialog--actions {
+  padding: 3px;
+  width: inherit;
+}
+
+.share-dialog__link {
+  padding: 0;
+  width: 220px;
```
robertknight: I generally prefer to use `min-width` and `min-height` to set minimum dimensions for a component with a bunch of sub-elements. This avoids the risk of cutting off content if the component's natural size is larger than the minimum for any reason - eg. a label being longer than anticipated.

h/static/scripts/directive/annotation-share-dialog.js
```diff
@@ -0,0 +1,40 @@
...
+    controller: ['$timeout', '$scope', function ($timeout, $scope) {
+      this.copyToClipboard = function (event) {
+        var $container = angular.element(event.currentTarget).parent();
+        var shareLinkInput = $container.find('input')[0];
+
+        try {
+          shareLinkInput.select();
+          document.execCommand('copy');
+
+          $scope.copyToClipboardMessage = 'Link copied to clipboard!';
```
robertknight: Rather than setting these properties on the scope directly, it would be slightly more idiomatic to set them on the controller, then access them in the scope via `vm.copyToClipboardMessage` for example.

h/static/styles/annotation-share-dialog.scss
robertknight: What was the status on the font size here? I'm thinking back to Conor's comments about having everything defined in one place so we can easily adjust if necessary in future.

h/static/scripts/directive/annotation-share-dialog.js
```diff
@@ -0,0 +1,71 @@
...
+              document.addEventListener('click', hideListener);
+            },
+            0
+          );
+        }
+      }.bind(this));
+
+      this.copyToClipboard = function (event) {
+        var $container = angular.element(event.currentTarget).parent();
+        var shareLinkInput = $container.find('input')[0];
```
robertknight: This duplicates the `shareLinkInput` variable above. These two lines can just be removed.

h/static/scripts/directive/test/annotation-share-dialog-test.js
```diff
@@ -0,0 +1,129 @@
...
+      var clock = sinon.useFakeTimers();
+      var expectedMessage = 'Link copied to clipboard!';
+
+      getCopyBtn().click();
+
+      clock.tick(1999);
+      clock.restore();
+
+      var actualMessage = element.find('.annotation-share-dialog-link__feedback').text();
+      assert.isBelow(actualMessage.indexOf(expectedMessage), 0);
```
robertknight: Won't block merge on this but in future I think it would be clearer to write tests for the presence/absence of a substring by comparing for equality/inequality to -1 rather than above/below zero.

h/static/scripts/directive/test/annotation-share-dialog-test.js
```diff
@@ -0,0 +1,129 @@
...
+    it('sets success class on successful copy to clipboard', function () {
+      var expectedMessage = 'Link copied to clipboard!';
+
+      getCopyBtn().click();
+
+      var actualMessage = element.find('.annotation-share-dialog-link__feedback').text();
+      assert.isAbove(actualMessage.indexOf(expectedMessage), -1);
+    });
+
+    it('unsets success class after one second on successful copy to clipboard', function () {
```
robertknight: The `it()` descriptions here are for an older version of the code. I'd suggest "displays message after successful copy" and "displays message after failed copy" as the labels for this test and the one below.

h/static/scripts/directive/test/annotation-share-dialog-test.js
robertknight: Actually, I forgot - there is a better option which gives a nicer error message if the test fails: [assert.include](http://chaijs.com/api/assert/#method_include)

robertknight:
LGTM. In testing I was (slightly) surprised to see that this does work in Internet Explorer and current versions of Microsoft Edge as well as Firefox and Chrome.

PR #3198, opened 2016-04-11
Add the skeleton of a functional test suite

nickstenning:
This commit adds the basic components of a functional test suite, which tests the entire WSGI application as one unit.

These tests currently do not run on Travis. It's likely we won't want them to run on Travis long-term, as it's likely the functional test suite will end up rather slow. For now, they can be run manually with

    tox -e functional

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -14,6 +14,7 @@ graft h/migrations
 graft h/static
 graft h/templates
 graft scripts
+prune tests
 graft vendor
 global-exclude __pycache__
 global-exclude *.py[co]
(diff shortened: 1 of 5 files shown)


nickstenning:
Gah, this is a nightmare. This is failing because the Travis-provided Elasticsearch doesn't have the ICU plugin installed. The simplest way to install the plugin involves switching away from containerised builds, which I'd really like to avoid. (VM build are often much slower to start.)

nickstenning:
@chdorner as discussed. Functional suite doesn't run on Travis, but I figure it's better we get started on it.

PR #3200, opened 2016-04-11
Displaying local file names in the Groups document list

chdorner:
Changes how local files are being displayed to:
2016-04-11-183100_481x438_scrot

1. Local file without a title
1. Link with no title
1. Local file with a title
1. ...

### DIFF
--- a/h/presenters.py
+++ b/h/presenters.py
@@ -9,7 +9,7 @@ import jinja2
 from h._compat import text_type
 
 
-def _format_document_link(href, title, link_text, hostname):
+def _format_document_link(href, title, link_text, host_or_filename):
     """Return a document link for the given components.
 
     Helper function for the .document_link property below.
@@ -19,33 +19,35 @@ def _format_document_link(href, title, link_text, hostname):
         double-escaped.
 
     """
-    if hostname and hostname in link_text:
-        hostname = ""
+    if href and host_or_filename and host_or_filename in link_text:
+        host_or_filename = ""
+    elif not href and title == host_or_filename:
+        title = ""
 
-    def truncate(content, length=60):
+    def truncate(content, length=55):
         """Truncate the given string to at most length chars."""
         if len(content) <= length:
             return content
         else:
             return content[:length] + jinja2.Markup("&hellip;")
 
-    hostname = truncate(hostname)
+    host_or_filename = truncate(host_or_filename)
     link_text = truncate(link_text)
 
-    if href and hostname:
-        link = '<a href="{href}" title="{title}">{link_text}</a><br>({hostname})'
-    elif hostname:
-        link = '<a title="{title}">{link_text}</a><br>({hostname})'
+    if href and host_or_filename:
+        link = '<a href="{href}" title="{title}">{link_text}</a><br>{host_or_filename}'
     elif href:
         link = '<a href="{href}" title="{title}">{link_text}</a>'
     else:
-        link = '<a title="{title}">{link_text}</a>'
+        link = '<em>Local file:</em> {title}'
+        if host_or_filename:
+            link += '<br>{host_or_filename}'
 
     link = link.format(
         href=jinja2.escape(href),
         title=jinja2.escape(title),
         link_text=jinja2.escape(link_text),
-        hostname=jinja2.escape(hostname))
+        host_or_filename=jinja2.escape(host_or_filename))
 
     return jinja2.Markup(link)
 
@@ -102,7 +104,7 @@ class AnnotationHTMLPresenter(object):
 
         """
         if self.filename:
-            return jinja2.escape(self.filename)
+            return jinja2.escape(urllib2.unquote(self.filename))
         else:
             hostname = urlparse.urlparse(self.uri).hostname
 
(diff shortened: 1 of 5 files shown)


chdorner:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;a36GEutS&#x2F;234-displaying-local-file-names-in-the-groups-document-list"> Displaying local file names in the Groups document list</a>

h/presenters.py
```diff
@@ -33,13 +35,13 @@ def truncate(content, length=60):
...
-    elif hostname:
-        link = '<a title="{title}">{link_text}</a><br>({hostname})'
+        link = '<a href="{href}" title="{title}">{link_text}</a><br>{hostname}'
     elif href:
         link = '<a href="{href}" title="{title}">{link_text}</a>'
     else:
-        link = '<a title="{title}">{link_text}</a>'
+        link = '<em>Local file:</em> {title}'
+        if hostname:
+            link += '<br>{hostname}'
```
chdorner: This is showing the `Local file` prefix for all documents where the link is not http or https, which is technically wrong, but leaving the old `a` tag without an `href` and branching off again didn't feel like the right way either. Any other thoughts on this?

robertknight: I think this is reasonable behavior, in the absence of actual data from our database telling us otherwise.

h/static/styles/group-form.scss
```diff
@@ -105,6 +105,11 @@
 
     &__list li {
       margin-bottom: 1.5em;
+
+      em {
```
robertknight: For consistency with the other style names, I think this would be better with a span with a suitable class rather than `<em>`, eg. `.group-documents-list__doctype` (in the `.scss` file, that means moving the contents of the `em {}` block here to a `&__doctype {}` block)

robertknight: Edit: Ignore the above, I see we're using just `<em>` to avoid needing to know where the link will be used when it is generated.

h/presenters.py
```diff
@@ -19,10 +19,12 @@ def _format_document_link(href, title, link_text, hostname):
         double-escaped.
 
     """
-    if hostname and hostname in link_text:
+    if href and hostname and hostname in link_text:
```
robertknight: It wasn't initially clear to me reading this that `hostname` could instead be the filename of the document. Worth adding a comment in the function's docstring?

h/test/presenters_test.py
```diff
@@ -399,9 +407,64 @@ def test_document_link_hostname_same_as_link_text(hostname_or_filename, href,
...
+def test_document_link_hostname_same_as_link_text_ignores_when_href_missing(hostname_or_filename,
+                                                                            href,
+                                                                            title,
+                                                                            link_text):
+    """
+    It should keep the hostname the same when href is missing, even if it
+    is the same as the link text.
+    """
+    href.return_value = None
+    title.return_value = "www.example.com"
```
robertknight: It would be clearer to use a representative title, ie. a filename (eg. "example.pdf") rather than a domain name in this test, since 'hostname' is in fact a filename in this context.

h/presenters.py
chdorner: Here is some data of how many annotations we have where the `url` is not one of `http://`, `https://`, `file://`

```
 1.80% - urn:x-pdf:
 0.07% - blob:http://
 0.04% - zotero://
 0.02% - ftp://
 0.02% - [object ...]
 0.01% - chrome-extension://
<0.01% - blob:(some id)
<0.01% - urn:x_pdf:
```

Considering that most of these uris can be considered local files, I think we can leave it like this.

h/presenters.py
chdorner: Yes, I didn't want to go off and refactor this whole presenter, but I've renamed it to `host_or_filename`.

h/test/presenters_test.py
chdorner: Fixed.

nickstenning:
LGTY, @robertknight?

robertknight:
In testing this locally with a file I had handy, I see the output rendered as:

```
Local file: [u'Employmentpolicy.pdf']
Employmentpolicy.pdf
```

nickstenning:
@robertknight diagnostic information would help here. Are you using the Chrome extension? Bookmarklet? In which browser?

I can't replicate this in Chrome v51.

nickstenning:
I can't replicate this in Firefox 43 with the bookmarklet, either.

robertknight:
I'm testing this in Chrome v50, using the Chrome extension to create the PDF. Christof was unable to reproduce with the same PDF locally. One possible difference is that I have the Postgres flags turned off locally.

nickstenning:
Postgres flags shouldn't affect this at all. I'm using the same PDF as you. Weird.

robertknight:
So to summarize a lengthy Slack conversation, the issue can be reproduced by:

1. Creating an annotation on a document which has not previously been annotated
2. Editing the annotation

When the document is initially created, the title is stored as a string in ElasticSearch. When sending the response to the client, AnnotationJSONPresenter renders an array. When submitting an update, the title field is now sent _back_ to the server as an array, which is then stored in ES. When retrieving only the title of the document from ES, the content was not converted to an array.

This is one of many things that nicely illustrate part of what the ES -> Postgres migration is for.

Issue #2970, opened 2016-02-17
"$sanitize:badparse" error from invalid HTML generated by showdown


robertknight:
**Steps to reproduce**

1. Create a new annotation and paste in the following text: `<one two.>`
2. Click 'Post'

**Expected**

I think we should probably disallow custom HTML tags in markdown generated by the user and just format them as plain text.

**Actual**

1. Nothing appears in the text field
2. An error appears in the console

```
Error: [$sanitize:badparse] The sanitizer was unable to parse the following block of html: <one two.></p>
http://errors.angularjs.org/1.4.7/$sanitize/badparse?p0=%3Cone%20two.%3E%3C%2Fp%3E
minErr/<@https://hypothes.is/assets/scripts/vendor/angular.min.js?310b5ef0:1:775
htmlParser@https://hypothes.is/assets/scripts/vendor/angular-sanitize.min.js?ea3db86c:1:5648
$SanitizeProvider/this.$get</<@https://hypothes.is/assets/scripts/vendor/angular-sanitize.min.js?ea3db86c:1:213
[22]</module.exports</<.link/renderMathAndMarkdown/parts<@https://hypothes.is/assets/scripts/app.min.js?910709f6:2:24260
[22]</module.exports</<.link/renderMathAndMarkdown@https://hypothes.is/assets/scripts/app.min.js?910709f6:2:23859
[22]</module.exports</<.link/ctrl.$render@https://hypothes.is/assets/scripts/app.min.js?910709f6:2:25409
ngModelWatch@https://hypothes.is/assets/scripts/vendor/angular.min.js?310b5ef0:8:28070
...
```

robertknight:
The error reported here no longer happens in Angular 1.5. The issue is now a more minor one. If the entered text is "foo <one" with no closing tag present then nothing after the '<' is rendered. GitHub on the other hand is more tolerant about malformed markup.

At this point I don't think it is worth spending a lot of time on, so I propose to continue doing some refactoring to extract the rendering code into its own testable unit, add a test case that checks no errors occur when parsing input like this and then move on to something else.

nickstenning:
Seeing as the reported error is now fixed... closing.

PR #3201, opened 2016-04-11
Markdown editor input refactor and additional tests

robertknight:
This is some cleanup of the markdown editor and additional tests:

- Change the `<markdown>` component to use the same data-in/events-out pattern as other recently written components by replacing the `ng-model` input with a `text` input property and `on-edit-text` output. This simplifies the implementation as well, which was effectively bypassing `ng-model`'s two-way binding in any case.
- Add tests for the previewing state
- Add tests for what is rendered when the annotation has no body
- Remove the unused 'required' input

### DIFF
--- a/h/static/scripts/directive/annotation.js
+++ b/h/static/scripts/directive/annotation.js
@@ -713,6 +713,10 @@ function AnnotationController(
     $scope.$digest();
   };
 
+  vm.setText = function (text) {
+    vm.form.text = text;
+  };
+
   init();
 }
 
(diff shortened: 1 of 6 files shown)


nickstenning:
Had a quick look but not entirely sure about the rebase conflict here so I'll leave this one to you.

robertknight:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;7KGhVjgr&#x2F;275-fix-sanitize-badparse-error-caused-by-malformed-html-generated-by-markdown-formatting-2970"> Fix $sanitize.badparse error caused by malformed HTML generated by markdown formatting (#2970)</a>

nickstenning:
LGTM.

PR #3202, opened 2016-04-12
Remove monkeypatch

seanh:
@nickstenning

### DIFF
--- a/h/api/search/test/client_test.py
+++ b/h/api/search/test/client_test.py
@@ -50,16 +50,13 @@ class TestIndexAnnotation:
     @pytest.fixture
     def presenters(self, patch):
         presenters = patch('h.api.search.client.presenters')
-        presenters.AnnotationJSONPresenter.return_value = mock.Mock(
-            asdict=mock.Mock(
-                return_value={
-                    'id': 'test_annotation_id',
-                    'target': [
-                        {
-                            'source': 'http://example.com/example',
-                        },
-                    ],
-                }
-            )
-        )
+        presenter = presenters.AnnotationJSONPresenter.return_value
+        presenter.asdict.return_value = {
+            'id': 'test_annotation_id',
+            'target': [
+                {
+                    'source': 'http://example.com/example',
+                },
+            ],
+        }
         return presenters
--- a/h/api/search/test/client_test.py
+++ b/h/api/search/test/client_test.py
@@ -10,16 +10,16 @@ from h.api.search import client
 @pytest.mark.usefixtures('elasticsearch', 'presenters')
 class TestIndexAnnotation:
 
-    def test_it_presents_the_annotation(self):
+    def test_it_presents_the_annotation(self, presenters):
         request = mock.Mock()
         annotation = mock.Mock()
 
         self.client().index_annotation(request, annotation)
 
-        client.presenters.AnnotationJSONPresenter.assert_called_once_with(
+        presenters.AnnotationJSONPresenter.assert_called_once_with(
             request, annotation)
 
-    def test_it_indexes_the_annotation(self):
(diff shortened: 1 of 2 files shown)


h/api/search/test/client_test.py
```diff
@@ -10,56 +10,47 @@
...
-        client_.index_annotation(mock.Mock(), annotation)
+        client_obj.index_annotation(mock.Mock(), annotation)
 
-        assert client_.conn.index.call_args[1]['body']['target'][0]['scope'] == [
-            annotation.target_uri_normalized]
+        assert client_obj.conn.index.call_args[1]['body']['target'][0]['scope'] == [annotation.target_uri_normalized]
 
-    def client(self):
+    @pytest.fixture
+    def client_obj(self):
```
nickstenning: It would seem nicer to me to call this `client` and if necessary do the import inside the function.

h/api/search/test/client_test.py
```diff
@@ -10,56 +10,47 @@
...
     @pytest.fixture
-    def presenters(self, monkeypatch):
-        monkeypatch.setattr(
-            'h.api.search.client.presenters',
-            mock.Mock(spec=client.presenters),
-        )
-        client.presenters.AnnotationJSONPresenter.return_value = mock.Mock(
+    def presenters(self, patch):
+        presenters = patch('h.api.search.client.presenters')
+        presenters.AnnotationJSONPresenter.return_value = mock.Mock(
```
nickstenning: You can also just chain these.

```python
presenter = presenters.AnnotationJSONPresenter.return_value
presenter.asdict.return_value = {...}
```

seanh: Sure, the existing way does fit on one line though, and avoids an additional variable

seanh: Oh I see what you mean, yeah that's nicer

h/api/search/test/client_test.py
seanh: Although if we every add any new top-level functions or classes to `h.api.search.client` and want to test them, we won't be able to...

nickstenning: Why not?

seanh: If we add a top-level `foobar()` function to the client module:

```python
def foobar():
    ...
```

Then normally we'd want tests to call it like:

```
from h.api.search import client

def my_test():
    client.foobar()
```

But if `from h.api.search import client` isn't imported at the top of the module we won't be able to call that function, at least not in that way.

nickstenning: But what's stopping you from importing at the top of the test module?

I'm merely suggesting that you separate the fixture definition from the module namespace. What you're proposing is entirely compatible with a fixture defined in a single class which looks like:

```python
@pytest.fixture
def client():
    from h.api.search.client import Client
    return Client(...)
```

And the `foobar` function would undoubtedly not be tested in the same class as the `Client` class. I'm not seeing where there's a conflict...

seanh: Hmm. But if a test method has a fixture named `client` as an argument, then it can't access the client module itself via the top-level name `client` that was imported to at the top of the test module. It'll trigger pylint's `redefined-outer-name` warning, too. I don't think it's very nice to use the same name twice in the same module, for two different things. I can change it if you really want to...

nickstenning: Things like `client_obj` or `search_lib` are ugly hacks to work around name collisions. Far better in my opinion to simply not have name collisions by, for example, ensuring that `TestClient` has no need to reference the top-level `client` module...

seanh: Changed it

nickstenning:
LGTM.

PR #3203, opened 2016-04-12
Add deleting of annotations from new search index

seanh:


nickstenning:
This LGTM, but I have two thoughts.

1. Rather than having separate test classes for `index_annotation` and `delete_annotation`, it would be more conventional, and probably somewhat more convenient, to have one test class for the `Client` class, `TestClient`. It would have saved you from having to do all the fixture juggling in this PR.

2. Given that a) we are only indexing annotations at the moment, and b) we're not deleting an annotation, we're merely removing it from the index, I think naming these two functions `index` and `remove` would be simpler and clearer.

nickstenning:
I guess on reflection `delete` is fine, but `index` and `delete` still seem simpler than `index_ann...`/`delete_ann...`.

seanh:
1. I think I prefer to have one test class per method. That's the pattern I've been using elsewhere, but also I think it makes use of per-class fixtures better, you put a `usefixtures` on the class with all the fixtures that every test of that method uses, tests for different methods even within the same class often use different fixtures. Granted that's not really taken advantage of in this case, but I think it is in more complex cases, and this is consistent with their pattern. It also avoids having to name every test method `    def test_index_annotation_*(self, ...)`. I'm not sure what the advantage of combining them into one class would be, in terms of the final code?

2. index and delete are Elasticsearch's terms.

3. The reason it's `index_annotation` and `delete_annotation` is that I imagine this `Client` class might be indexing and deleting other types of document in the future, though it isn't currently.

nickstenning:
One test class per tested unit makes more sense to me. A class is by definition an object with shared state, and so there will be certain behaviours of a class that simply cannot be tested with "one test class per method."

>  you put a usefixtures on the class with all the fixtures that every test of that method uses, tests for different methods even within the same class often use different fixtures.

Yes, and you can also put `usefixtures` on individual test methods. I don't see the argument here.

nickstenning:
> I imagine this Client class might be indexing and deleting other types of document in the future, though it isn't currently.

YAGNI.

Make design decisions for the requirements we have now, not the requirements we might have in the future.

seanh:
Removing `_annotation()` I can definitely go for

seanh:
I'm not sure I understand what the behaviours of a class that can't be tested with one test class per method are? One of the things I liked about one test class per method was that I had one `usefixtures` on the test class instead of an `@foomethod_fixtures` on every test method, I felt it reduced boilerplate.

nickstenning:
Behaviours that require interaction between multiple methods? A class is not a collection of functions. It's a collection of functions *and* associated state. So some behaviours of a class require calls to multiple functions to verify. Here is [just one simple example](https://github.com/hypothesis/h/pull/3204/files#diff-c1d09f6442ff2066236ca60a5f21a922R104).

As for your concern about `@foomethod_fixtures` on every test method -- let's worry about that when it happens. In practice I think you'll more often be using one base set of fixtures for the class, and occasionally having to pass in more to specific test functions.

seanh:
I'm not quite convinced yet about one test class per class. It seems better to me to split tests up into smaller, easier to understand classes, and test one method at a time so that each test has only one reason to fail. In some instances we may want to additionally add some multi method tests, which could be added in separate test classes, like `TestEncodeDecodeSanity`.

Anyway, this pr has been rewritten as functions named `index()` and `delete()`

nickstenning:
LGTM. I think I probably would have called the `es` parameter `client`, but that's enough fannying about with this PR :)

PR #3204, opened 2016-04-12
Convert h.features to package

nickstenning:
In preparation for some changes to the feature flags module, convert `h.features` into a package.

Moves the various parts of the feature flags system into their own modules. The `Feature` model moves into its own `models` module, and the `Client` class into a `client` module. The set of current valid flags is part of the model, so moves there.

I've made a couple of small changes that allow the details of the `FEATURES` and `FEATURES_PENDING_REMOVAL` dictionaries to be hidden behind the interface exposed by the model class, including substantial simplifications to the `Client` class that results from reusing the
existing `Feature.all()` method rather than reimplementing its functionality within the client.

The `remove_old_flags` function also moves into the model as a classmethod, again so that everything that needs to know about the `FEATURES{,_PENDING_REMOVAL}` dictionaries is in one place.

The tests for `Client` have been simplified, and tests added for model methods that were missing them.

---

I've also added a commit that fixes a minor bug in the behaviour of the client's caching -- it would previously fail to cache data if there were no feature flags currently active in the application.


h/features/client.py
```diff
@@ -0,0 +1,81 @@
...
+        """
+        Determine if the named feature is enabled for the current request.
+
+        If the feature has no override in the database, it will default to
+        False. Features must be documented, and an UnknownFeatureError will be
+        thrown if an undocumented feature is interrogated.
+
+        When the internal cache is empty, it will automatically load the
+        feature flags from the database first.
+        """
```
seanh: Is this docstring just a duplicate of the class one? On a class like this, it might be okay to just docstring the class and not the method. Then `enabled()` could maybe just be moved into `__call__()` as well

seanh:
Curious about the use of `fetcher` instead of just calling `models.Feature.all()` directly? Tests can then pass in a mock fetcher, but tests could easily patch `models.Feature.all()` as well..

seanh:
lgtm

PR #3206, opened 2016-04-13
Add Postgres UpdateAnnotationSchema

seanh:
* Move a bunch of validation code out of `CreateAnnotationSchema` into `AnnotationSchema`, so that `UpdateAnnotationSchema` can share it. Move the tests too.

* Add new `UpdateAnnotationSchema` intended for use when updating annotations in Postgres (view and storage code for updating annotations in Postgres not written yet)

* The legacy Elasticsearch `UpdateAnnotationSchema` becomes `LegacyUpdateAnnotationSchema`

* The view calls `LegacyUpdateAnnotationSchema`, the new `UpdateAnnotationSchema` isn't ready for use until the storage code for updating annotations in Postgres is written

* The new `UpdateAnnotationSchema` doesn't do the "user must be admin of annotation to change its permissions" test that the legacy one does. This test was always passed, I believe. But we can't implement it in terms of the new Postgres annotation model anyway.

### DIFF
--- a/h/api/schemas.py
+++ b/h/api/schemas.py
@@ -307,6 +307,26 @@ class LegacyCreateAnnotationSchema(object):
 
 class UpdateAnnotationSchema(object):
 
+    """Validate the POSTed data of an update annotation request."""
+
+    def __init__(self, request, annotation):
+        self.request = request
+        self.annotation = annotation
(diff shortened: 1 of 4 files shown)


nickstenning:
LGTM.

Issue #2935, opened 2016-02-10
Angular error: [$sanitize:badparse] The sanitizer was unable to parse the following.


nickstenning:
### Steps to reproduce

1. Go to http://www.nytimes.com/2015/01/17/science/earth/2014-was-hottest-year-on-record-surpassing-2010.html and activate Hypothesis.

### Result

The console contains this error:

```
angular.min.js?d1322f9d:5 Error: [$sanitize:badparse] The sanitizer was unable to parse the following block of html: <img</a> src="http://ipccar5summary.org/wp-content/uploads/2014/10/WGI_FSPM1b_Observed-change-in-surface-temperature-1901%E2%80%932012.png" alt="Image Description" /></p>
http://errors.angularjs.org/1.4.7/$sanitize/badparse?p0=%3Cimg%3C%2Fa%3E%20…%2580%25932012.png%22%20alt%3D%22Image%20Description%22%20%2F%3E%3C%2Fp%3E
    at angular.min.js?d1322f9d:1
    at htmlParser (angular-sanitize.min.js?ea3db86c:1)
    at angular-sanitize.min.js?ea3db86c:1
    at app.min.js?63e82ed1:2
    at renderMathAndMarkdown (app.min.js?63e82ed1:2)
    at ctrl.$render (app.min.js?63e82ed1:2)
    at Object.ngModelWatch (angular.min.js?d1322f9d:8)
    at Scope.$digest (angular.min.js?d1322f9d:6)
    at app.min.js?63e82ed1:3
    at render (app.min.js?63e82ed1:3)
```

robertknight:
See also #2970

nickstenning:
According to @robertknight this was fixed by upgrading to Angular 1.5. This bug no longer repros on the page linked.

PR #3208, opened 2016-04-14
Markdown rendering refactor and math sanitization fixes

robertknight:
This PR separates out the markdown and math rendering code into a separate module from the markdown editor and adds tests. This includes tests for the issue in (#2935, #2970) which were inadvertently fixed by the Angular 1.5.0 upgrade.

In the process and following @seanh 's comments in the initial review, I found several bugs in the way that math was rendered and sanitized, which necessitated a restructuring of the rendering logic to fix properly.

1. The previous implementation allowed inline math blocks to start or end with `\(` or `\)`, rather than requiring them to start only with `\(` and end only with `\)`. This is fixed in this PR
2. The result of inline math parsing of a paragraph was fed through markdown rendering, so if the rendered math contained text which is also valid markdown, that would be rendered as markdown.
3. It turns out that ngSanitize can mangle the relatively complex HTML generated by KaTeX. Previously the non-HTML _input_ (!) to KaTeX was sanitized and the result was sanitized for inline math but not for block math. Since KaTeX escapes all text input, passing the inputs or outputs through ngSanitize is unnecessary. This is fixed as follows:
 1. The input markdown is first processed to extract math blocks and replace them with placeholders
 1. The resulting markdown is then rendered and sanitized
 1. The extracted math blocks are then rendered with KaTeX and the placeholders (`{math:<ID>}` tags) are replaced with the rendered HTML.


robertknight:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;7KGhVjgr&#x2F;275-fix-sanitize-badparse-error-caused-by-malformed-html-generated-by-markdown-formatting-2970"> Fix $sanitize.badparse error caused by malformed HTML generated by markdown formatting (#2970)</a>

seanh:
Needs a rebase now

seanh:
I'm seeing some differences in the math rendering between this branch and master. Annotation body:

```
sdsa  **foo**

$$\lim_{x \to \infty} \exp(-x) = 0$$
```

On master:

screenshot from 2016-04-18 14-18-32

On this branch:

screenshot from 2016-04-18 14-19-37

Also noticed that multiline math seems to be rendered as math on this branch, but is not on master:

```
sdsa  **foo**

$$
\lim_{x \to \infty} \exp(-x) = 0
$$
```

robertknight:
> I'm seeing some differences in the math rendering between this branch and master. Annotation body:

It looks like this is down to running `$sanitize` on the whole HTML string that is generated. Previously `$sanitize` was not being run on the HTML generated by KaTeX.

> Also noticed that multiline math seems to be rendered as math on this branch, but is not on master:

I think I'll consider that a bug-fix.

robertknight:
OK. It took me a while to dig around and figure out what the correct approach to handling HTML sanitization and KaTeX is. It turns out that KaTeX escapes all text input and does not allow insertion of HTML tags, so we don't need to sanitize its inputs or outputs. Aside from being unnecessary work, sanitizing the output with ngSanitize can mangle KaTeX' relatively complex HTML. I've restructured the way rendering works to feed only the markdown blocks through ngSanitize and added tests.

h/static/scripts/render-markdown.js
```diff
@@ -0,0 +1,130 @@
...
+function targetBlank() {
+  function filter(text) {
+    return text.replace(/<a href=/g, '<a target="_blank" href=');
+  }
+  return [{type: 'output', filter: filter}];
+}
+
+var converter;
+
+function renderMarkdown(markdown) {
```
seanh: It's fairly obvious but I think a docstring on this wouldn't hurt, to clarify the input and output. AIUI the input is markdown source text (not including KaTeX source text or HTML from KaTeX - those bits have already been replaced by placeholders) and the output is HTML.

h/static/scripts/render-markdown.js
```diff
@@ -0,0 +1,130 @@
...
+    pos = mathStart + replacement.length;
+  }
+
+  return {
+    mathBlocks: mathBlocks,
+    content: replacedContent,
+  };
+}
+
+function insertMath(html, mathBlocks) {
```
seanh: Also could do with a docstring

seanh:
The increase in clarity of the code here is huge

seanh:
On the tests, I guess I feel it'd be nice to have unit tests for the individual functions in `render-markdown.js`. `extractMath()` and `insertMath()` at least seem complex enough to warrant their own unit tests, the other two are pretty simple, but it looks like you've gone for an approach of just testing everything indirectly via the top-level function `renderMathAndMarkdown()` (which is slightly confusingly called `renderMarkdown()` in the tests and to other code, although the module contains a different function called `renderMarkdown()`)

PR #3210, opened 2016-04-14
 Fix error in session.js when an /app request fails 

robertknight:
When a request to the session endpoint fails with a network
error, avoid trying to parse the response in process().

This is the most frequently reported Sentry event - https://app.getsentry.com/hypothesis/prod-client/issues/116030860/

### DIFF
--- a/h/static/scripts/app.js
+++ b/h/static/scripts/app.js
@@ -28,30 +28,33 @@ if (settings.raven) {
 var mail = require('./vendor/jwz');
 var streamer = require('./streamer');
 
-var resolve =
-  // Ensure that we have available a) the current authenticated userid, and b)
-  // the list of user groups.
-  {
-    sessionState: ['session', function (session) { return session.load(); }],
-    store: ['store', function (store) { return store.$promise; }],
-    streamer: streamer.connect,
-    threading: [
-      'annotationMapper', 'drafts', 'threading',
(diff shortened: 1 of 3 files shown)


h/static/scripts/app.js
```diff
@@ -28,30 +28,33 @@ if (settings.raven) {
...
-        threading.root = mail.messageContainer();
-
-        // Reload all new, unsaved annotations
-        threading.thread(drafts.unsaved());
-
-        return threading;
-      }
-    ]
-  };
+// Fetch external state that the app needs before it can run. This includes the
```
robertknight: This is just a cleanup of this function, no code changes other than using `@ngInject`. See the first commit in the PR.

h/static/scripts/session.js
```diff
@@ -154,8 +154,11 @@ function session($http, $resource, $rootScope, flash, raven, settings) {
     return model;
   };
 
-  function process(data) {
-    // Parse as json
+  function process(data, headersGetter, status) {
+    if (status < 200) {
```
nickstenning: What about if `status == 504`?

robertknight: Good point! - Presumably any status >= 500 may also not be valid JSON?

chdorner:
LGTM.

PR #3211, opened 2016-04-14
Retry requests to /api endpoint

robertknight:
Handle requests to the /api endpoint that fail with network errors in
the same way as failed requests to the /app endpoint by retrying the
request with expontential backoff until it succeeds.

Otherwise, if the request fails then `store.(Annotation|Search)Resource`
are not populated and code that tries to make API requests fails.

I believe this is the cause of errors relating to AnnotationResource in Sentry. See https://app.getsentry.com/hypothesis/prod-client/?query=is%3Aunresolved%20AnnotationResource

### DIFF
--- a/h/static/scripts/store.js
+++ b/h/static/scripts/store.js
@@ -2,6 +2,8 @@
 
 var angular = require('angular');
 
+var retryUtil = require('./retry-util');
+
 function prependTransform(defaults, transform) {
   // We can't guarantee that the default transformation is an array
   var result = angular.isArray(defaults) ? defaults.slice(0) : [defaults];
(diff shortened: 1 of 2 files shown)


h/static/scripts/store.js
```diff
@@ -99,29 +105,28 @@ function store($http, $resource, settings) {
   };
 
   // We call the API root and it gives back the actions it provides.
-  instance.$resolved = false;
```
seanh: I guess `$resolved` was no longer needed?

seanh:
Tested by making the API endpoint randomly fail, works great.

You can't actually annotate until (if and when) it makes a successful APO request, so it might be good to show somehow that it's "loading" until then

PR #3213, opened 2016-04-14
Dispatch markdown editor input changes from $apply() block

### LINKED ISSUE #3212: Post annotation button not activating
Someone been messing with AnnotationController? :)

untitled screencast 12

(When you create a new annotation and type something in the textarea that post button does not become active, clicking on it does nothing. In the gif you can see it activate when I click elsewhere in the window to turn off screencastify)

robertknight:
The input event handler's callback was not invoked inside an $apply()
block, so the $digest function was never called to propagate the changes
in vm.setEditText() in annotation.js. This happened to work previously
because ngModel's $setViewValue() was called to apply input changes,
which internally uses $scope.$apply().

This commit wraps the input events from the markdown editor with $apply
but debounces them. Since $apply() triggers every single watcher in the
application, running it on every keystroke causes noticeable lag
otherwise.

Fixes #3212

### DIFF
--- a/h/static/scripts/directive/test/markdown-test.js
+++ b/h/static/scripts/directive/test/markdown-test.js
@@ -44,6 +44,12 @@ describe('markdown', function () {
             return 'math:' + input.replace(/$$/g, '');
           },
         },
+        'lodash.debounce': function (fn) {
+          // Make input change debouncing synchronous in tests
+          return function () {
+            fn();
+          };
+        },
         '../markdown-commands': {
           convertSelectionToLink: mockFormattingCommand,
           toggleBlockStyle: mockFormattingCommand,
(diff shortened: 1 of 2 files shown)


h/static/scripts/directive/test/markdown-test.js
```diff
@@ -44,6 +44,12 @@ describe('markdown', function () {
             return 'math:' + input.replace(/$$/g, '');
           },
         },
+        'lodash.debounce': function (fn) {
+          // Make input change debouncing synchronous in tests
+          return function () {
```
nickstenning: `return fn;`?

nickstenning:
LGTM. Not going to wait for the rest of the tests.

h/static/scripts/directive/test/markdown-test.js
robertknight: Yes, indeed. :blush:

PR #3214, opened 2016-04-14
Remove empty annotations when a new annotation is created

robertknight:
When a new annotation is created, remove any empty annotation cards that
currently exist. This fixes an issue where it was easy to accidentally
create a large number of blank annotations or empty replies.

Any annotations which are not new (ie. have an ID assigned) or have tags
or text entered by the user are kept.

This logic applies to both annotations and replies.

### DIFF
--- a/h/static/scripts/directive/annotation.js
+++ b/h/static/scripts/directive/annotation.js
@@ -1,6 +1,8 @@
 /* jshint node: true */
 'use strict';
 
+var angular = require('angular');
+
 var annotationMetadata = require('../annotation-metadata');
 var dateUtil = require('../date-util');
 var documentDomain = require('../filter/document-domain');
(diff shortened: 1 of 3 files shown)


robertknight:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;ipHs24Hd&#x2F;295-remove-empty-annotation-editor-when-user-creates-a-new-annotation"> Remove empty annotation editor when user creates a new annotation</a>

h/static/scripts/directive/annotation.js
```diff
@@ -306,6 +312,12 @@ function AnnotationController(
     }
   }
 
+  function deleteIfNewAndEmpty() {
+    if (!domainModel.id && !vm.form.text && vm.form.tags.length === 0) {
```
seanh: Should this use `isNew()`?

h/static/scripts/directive/annotation.js
```diff
@@ -360,7 +372,7 @@ function AnnotationController(
       // Highlights are always private.
       domainModel.permissions = permissions.private();
       domainModel.$create().then(function() {
-        $rootScope.$emit('annotationCreated', domainModel);
+        $rootScope.$emit(events.ANNOTATION_CREATED, domainModel);
```
seanh: Is it necessary to close an existing empty unsaved annotation on creating a new highlight? Creating a highlight won't result in two open annotation editors

seanh:
Otherwise looks good

h/static/scripts/directive/annotation.js
robertknight: That would probably be slightly clearer.

h/static/scripts/directive/annotation.js
robertknight: Not strictly necessary but I don't think it is a problem either, and one less code path to test.

PR #3215, opened 2016-04-14
Scroll share dialog into view when created

robertknight:
This PR converts the `<share-dialog>` component from CoffeeScript to JS and updates the style for consistency with other components (first commit), and adds logic to scroll the dialog into view when it appears (second commit).

Fixes #3043

PR #3216, opened 2016-04-14
Show product version and environment info.

sheetaluk:
Show a dismissable card to the the top of the sidebar containing
information about the version of the tool they have installed as
well as environment information so that we have more information
while troubleshooting errors.

### DIFF
--- a/h/static/scripts/directive/about-this-version-dialog.js
+++ b/h/static/scripts/directive/about-this-version-dialog.js
@@ -1,0 +1,39 @@
+'use strict';
+
+/**
+ * @ngdoc directive
+ * @name aboutThisVersionDialog
+ * @description Displays product version and environment info
+ */
+// @ngInject
+module.exports = function () {
+  return {
+    bindToController: true,
+    controllerAs: 'vm',
+    // @ngInject
+    controller: function ($scope, $window, settings, crossframe) {
+      this.userAgent = $window.navigator.userAgent;
+      this.version = settings.release;
+      this.dateTime = new Date();
+
+      $scope.$watchCollection(
+        function () {
+          return crossframe.frames;
+        },
+        function (frames) {
+          if (frames.length === 0) {
+            return;
+          }
+          this.url = frames[0].uri;
+          this.documentFingerprint = frames[0].documentFingerprint;
+        }.bind(this)
+      );
+    },
+    restrict: 'E',
+    template: require('../../../templates/client/about_this_version_dialog.html'),
+    scope: {
+      auth: '<',
+      onClose: '&',
+    }
+  };
+};
--- a/h/templates/app.html.jinja2
+++ b/h/templates/app.html.jinja2
@@ -29,6 +29,7 @@
       on-logout="logout()"
       account-dialog="accountDialog"
       share-dialog="shareDialog"
+      on-show-about-version-dialog="aboutThisVersionDialog.visible = true"
       is-sidebar="::isSidebar"
       search-controller="search"
       sort-by="sort.name"
@@ -49,6 +50,10 @@
                 on-close="accountDialog.visible=false"></login-form>
       <sidebar-tutorial ng-if="isSidebar"></sidebar-tutorial>
       <share-dialog ng-show="shareDialog.visible"></share-dialog>
+      <about-this-version-dialog ng-if="aboutThisVersionDialog.visible"
+        on-close="aboutThisVersionDialog.visible = false"
+        auth="auth">
+      </about-this-version-dialog>
       <main ng-view=""></main>
     </div>
 
--- a/h/templates/client/about_this_version_dialog.html
+++ b/h/templates/client/about_this_version_dialog.html
@@ -1,0 +1,37 @@
+<div class="sheet">
+  <i class="close h-icon-close"
+    role="button"
+    title="Close"
+    ng-click="vm.onClose()"></i>
+  <div class="about-this-version">
+    <ul class="nav nav-tabs">
+      <li class="active"><a href="">About this version</a></li>
+    </ul>
+    <dl class="tab-content">
+      <dt class="about-this-version__key">Version: </dt>
+      <dd class="about-this-version__val">{{ vm.version }}</dd>
+      <dt class="about-this-version__key">User agent: </dt>
+      <dd class="about-this-version__val">{{ vm.userAgent }}</dd>
+      <dt class="about-this-version__key">URL: </dt>
+      <dd class="about-this-version__val">{{ vm.url }}</dd>
+      <div ng-if="vm.documentFingerprint">
+        <dt class="about-this-version__key">PDF fingerprint: </dt>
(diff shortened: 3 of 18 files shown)


robertknight:
When logged out, there is an issue that the card's content overflow its boundaries.

3216-about-this-version-dialog

h/static/scripts/directive/about-this-version-dialog.js
```diff
@@ -0,0 +1,38 @@
...
+    controller: function ($scope, $window, settings, crossframe) {
+      this.userAgent = $window.navigator.userAgent;
+      this.version = settings.release;
+
+      $scope.$watchCollection(
+        function () {
+          return crossframe.frames;
+        },
+        function (frames) {
+          if (!frames || frames.length <= 0) {
```
robertknight: `frames.length` can never be `< 0` and frames is guaranteed to be an array, so this can be simplified to `if (frames.length === 0)`

h/templates/client/about_this_version_dialog.html
```diff
@@ -0,0 +1,25 @@
+<div class="sheet">
+  <i class="close h-icon-close"
+    role="button"
+    title="Close"
+    ng-click="vm.aboutThisVersionDialog.visible = false"></i>
```
robertknight: As a general principle, new components should not directly manipulate data that is passed in via the scope but rather emit events when the user performs some action, and the parent handles the change.

So in this case, rather than the dialog itself setting the flag which indicates which dialog is visible, it would have an `onClose` event which is triggered when the user clicks the 'Close' button. https://github.com/hypothesis/h/pull/3215/files has a rewrite of the existing `<share-dialog>` to follow this pattern.

h/templates/client/about_this_version_dialog.html
```diff
@@ -0,0 +1,25 @@
...
+      <dt class="about-this-version__key">User Agent: </dt>
+      <dd class="about-this-version__val">{{ vm.userAgent }}</dd>
+      <dt class="about-this-version__key">URL: </dt>
+      <dd class="about-this-version__val">{{ vm.url }}</dd>
+      <div ng-if="vm.documentFingerprint">
+        <dt class="about-this-version__key">PDF fingerprint: </dt>
+        <dd class="about-this-version__val">{{ vm.documentFingerprint }}</dd>
+      </div>
+      <dt class="about-this-version__key">User ID: </dt>
+      <dd class="about-this-version__val">{{ vm.auth.userid.split(":")[1] }}</dd>
```
robertknight: There is a `parseAccountID` function in `scripts/filter/persona.js` which you can use for this.

h/templates/app.html.jinja2
```diff
@@ -49,6 +50,10 @@
                 on-close="accountDialog.visible=false"></login-form>
       <sidebar-tutorial ng-if="isSidebar"></sidebar-tutorial>
       <share-dialog ng-show="shareDialog.visible"></share-dialog>
+      <about-this-version-dialog ng-show="aboutThisVersionDialog.visible"
```
robertknight: This should use `ng-if` rather than `ng-show`. The difference is that `ng-show` hides or shows the dialog using CSS depending on whether it is visible. `ng-if` doesn't put the element in the page at all if it is not visible. The benefit of removing the component from the page entirely is that Angular does not have to check for changes to the values of expressions in the `<about-this-version-dialog>` template whenever user input or a timeout or an HTTP request occurs (which might otherwise change the values of what is displayed in the dialog).

robertknight: The `<share-dialog>` above uses `ng-show` but this is fixed in https://github.com/hypothesis/h/pull/3215

h/templates/client/about_this_version_dialog.html
```diff
@@ -0,0 +1,32 @@
+<div class="sheet">
+  <i class="close h-icon-close"
+    role="button"
+    title="Close"
+    ng-click="vm.aboutThisVersionDialog.visible = false"></i>
+  <div class="about-this-version">
+    <ul class="nav nav-tabs">
+      <li class="active"><a href="">About this verison</a></li>
```
lenazun: Tiny typo:  verison -> version

h/templates/client/about_this_version_dialog.html
```diff
@@ -0,0 +1,34 @@
...
+        <dd class="about-this-version__val">{{ vm.documentFingerprint }}</dd>
+      </div>
+      <div ng-if="vm.auth.userid">
+        <dt class="about-this-version__key">Username: </dt>
+        <dd class="about-this-version__val">{{ vm.auth.username }}</dd>
+      </div>
+      <dt class="about-this-version__key">Date: </dt>
+      <dd class="about-this-version__val">{{ vm.dateTime | date:'dd MMM yyyy HH:mm:ss Z' }}</dd>
+      <div class="about-this-version-feedback">
+        <a class="about-this-version-feedback__link" href="mailto:support@hypothes.is?subject=Hypothesis%20feedback&amp;body=Version:%20{{ vm.version }}%0D%0AUser%20Agent:%20{{vm.userAgent}}%0D%0AURL:%20{{ vm.url }}%0D%0APDF%20fingerprint:%20{{ vm.documentFingerprint ? vm.documentFingerprint : '-' }}%0D%0AUsername:%20{{ vm.auth.username ? vm.auth.username : '-' }}%0D%0ADate:%20{{ vm.dateTime | date:'dd MMM yyyy HH:mm:ss Z' }} ">
```
robertknight: The support link is difficult to read in the template. I would suggest generating it via a helper function in the controller.

h/static/scripts/directive/about-this-version-dialog.js
```diff
@@ -0,0 +1,39 @@
...
+// @ngInject
+module.exports = function () {
+  return {
+    bindToController: true,
+    controllerAs: 'vm',
+    // @ngInject
+    controller: function ($scope, $window, settings, crossframe) {
+      this.userAgent = $window.navigator.userAgent;
+      this.version = settings.release;
+      this.dateTime = new Date();
```
robertknight: I don't feel that strongly about it but I didn't see the date mentioned on the card and I'm not sure it will be that useful given that we have the version number.

dwhly: @robertknight I think it's actually quite useful to have the date.  I added a note on the trello card advocating for it last week.  https://trello.com/c/LGxxgOwd/298-add-product-version-and-environment-info-to-the-tool

TL:DR, there are all kinds of reasons why knowing exactly when someone reported an issue with our system might be important. For instance, their issue may have been the result of an operational situation rather than a code issue.  Without the date we don't know exactly when the issue was, and whether perhaps it might not even be a current problem.  It seems a simple thing to include.

robertknight:
I found an issue in the Chrome extension where clicking the 'Send feedback' link did not result in my mail client being shown when viewing a PDF. It did work however when viewing a normal web page.

robertknight:
I'm going to merge this now. The issue with the 'Send feedback' link in the PDF viewer can be looked into separately.

PR #3217, opened 2016-04-15
Realtime messaging with RabbitMQ

chdorner:
This PR aims to switch our message processing from NSQ to RabbitMQ.

The new `h.realtime` module sets up a RabbitMQ exchange called
`realtime`, which is a direct exchange that is not persisted
(non-durable and delivery mode is transient). Alongside convenience classes for publishing (`h.realtime.Publisher`) and consuming (`h.realtime.Consumer`) real-time messages.

Each consumer creates its own queue with a unique name, which ensures that every message gets routed to all consumers listening for the configured routing key.

The `h.realtime.Consumer` is built with the `kombu.mixins.ConsumerMixin`
class, and after initialization is started by calling the `run()`
method. It should, provided no connection errors are happening, run
forever until stopped.

The `h.realtime.Publisher` is meant to be mounted on the Pyramid request
(accessible as `request.realtime`) and provides two convenience methods
for publishing annotation and user messages to RabbitMQ and its
subscribers.

This PR also adds RabbitMQ as a new required dependency for running h locally. I will switch over the default broker for Celery to RabbitMQ in a later PR.


h/config.py
```diff
@@ -60,6 +62,10 @@
     EnvSetting('nsq.reader.addresses', 'NSQ_READER_ADDRS'),
     EnvSetting('nsq.writer.address', 'NSQ_WRITER_ADDR'),
     EnvSetting('origins', 'ALLOWED_ORIGINS'),
+    EnvSetting('rabbitmq.address', 'RABBITMQ_ADDRESS'),
+    EnvSetting('rabbitmq.user', 'RABBITMQ_USER'),
+    EnvSetting('rabbitmq.password', 'RABBITMQ_PASSWORD'),
+    EnvSetting('rabbitmq.vhost', 'RABBITMQ_VHOST'),
```
chdorner: I'm not a big fan of this part, but I changed it to make sure that linking a docker container works with.

Otherwise we could just reuse the `BROKER_URL` environment variable for the `h.realtime` module.

Any thoughts? Can we decide not to care about supporting linking docker containers and it magically working? (Similar to what we already do with postgres).

chdorner:
This is ready for an initial review now, since it is quite a big change it would be good to get both @nickstenning's and @seanh's :eyes: on this.

chdorner:
Switching it back to WIP, found an issue locally.

chdorner:
Bug fixed. I've set up the RabbitMQ exchange with the type `fanout` which doesn't take the routing key into account and just routes the message to each queue which is bound to that exchange. Which resulted in both topic handlers being called for any message sent to the realtime exchange.

We are now using a `direct` exchange, which routes the messages to each queue that is bound to the exchange and listens to the correct routing key. We already ensure that the queues are not shared between Python processes (by giving them a unique name), so each message now gets sent to the correct consumers.

chdorner:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;h4a1B6Dh&#x2F;323-switch-real-time-messaging-to-rabbitmq"> Switch real-time messaging to RabbitMQ</a>

h/realtime.py
```diff
@@ -0,0 +1,122 @@
+import base64
```
nickstenning: Needs a `# -*- coding: utf-8 -*-` header.

h/streamer/messages.py
```diff
@@ -1,69 +1,50 @@
-# -*- coding: utf-8 -*-
```
nickstenning: This should stay.

h/realtime.py
chdorner: Fixed.

h/streamer/messages.py
chdorner: And fixed.

h/config.py
chdorner: Might also be worth considering not reusing `BROKER_URL` (and possibly renaming it to `CELERY_BROKER`) since Celery does work with other message brokers than RabbitMQ, while our new real-time module only works with an amqp compatible message broker.

chdorner: As discussed in Slack: Supporting docker links auto-magically is increasingly becoming a problem. It is not very easy to support docker links with services that are not set up with default configuration (like a username/password). So we decided to stop supporting docker links (in a new pull request that will come tomorrow).
In the mean time, there's no need to add support for docker links for RabbitMQ, so it's back to the `BROKER_URL` environment variable (with the addition of a `CELERY_BROKER_URL` that gets prioritized over `BROKER_URL` for Celery)

h/test/realtime_test.py
```diff
@@ -0,0 +1,151 @@
...
+    def get_exchange(self, patch):
+        return patch('h.realtime.get_exchange')
+
+    @pytest.fixture
+    def producer_pool(self, patch):
+        return patch('h.realtime.producer_pool')
+
+    @pytest.fixture
+    def req(self):
+        request = mock.MagicMock(registry=mock.MagicMock(settings={}))
```
nickstenning: In circumstances like this it's probably better and easier to use a `pyramid.testing.DummyRequest` object.

h/test/realtime_test.py
```diff
@@ -0,0 +1,151 @@
...
+        exchange = realtime.get_exchange()
+        assert exchange.durable is False
+
+    def test_delivery_mode(self):
+        """Test that delivery mode is 1 (transient)"""
+        exchange = realtime.get_exchange()
+        assert exchange.delivery_mode == 1
+
+    @pytest.fixture
+    def Exchange(self, patch):
```
nickstenning: You're not using this fixture in 3/4 of these tests. I guess that means it's not even really needed?

setup.py
```diff
@@ -41,11 +41,11 @@
     'deform-jinja2>=0.5,<0.6',
     'elasticsearch>=1.1.0,<2.0.0',
     'gevent>=1.0.2,<1.1.0',
-    'gnsq>=0.3.0,<0.4.0',
     'gunicorn>=19.2,<20',
     'itsdangerous>=0.24',
     'jsonpointer==1.0',
     'jsonschema>=2.5.1,<2.6',
+    'kombu>=3.0.35',
```
nickstenning: As with the other version specifications in this file, it's probably wise to add an upper bound on the assumption that the project is semantically versioned, so `kombu>=3.0.35,<3.1`.

h/test/realtime_test.py
```diff
@@ -0,0 +1,151 @@
...
+        exchange = realtime.get_exchange()
+
+        publisher = realtime.Publisher(req)
+        publisher.publish_user(payload)
+
+        producer.publish.assert_called_once_with(
+           payload, exchange=exchange, declare=[exchange], routing_key='user')
+
+    @pytest.fixture
+    def get_connection(self, patch):
```
nickstenning: This fixture doesn't seem to be used. Am I missing something?

h/test/realtime_test.py
```diff
@@ -0,0 +1,151 @@
...
+        return patch('h.realtime.get_connection')
+
+    @pytest.fixture
+    def get_exchange(self, patch):
```
nickstenning: This fixture doesn't seem to be used. Am I missing something?

nickstenning:
I've taken the liberty of addressing some of the little issues with tests so we can get this merged.

nickstenning:
OK. I'm going to go ahead and merge this.

PR #3218, opened 2016-04-15
Update front-end testing dependencies to current versions and switch to mocha reporter

robertknight:
Two commits here, the first updates the front-end testing dependencies (Karma, mocha, chai, sinon etc.) to current npm versions. The exception is that PhantomJS remains at 1.9.x because the upgrade to PhantomJS 2.x resulted in bunch of test failures and will require more work to finish off.

The second switches from the default 'dots' reporter to 'karma-mocha-reporter', which produces output that is easier to read and more helpful when test assertions fail. See https://www.npmjs.com/package/karma-mocha-reporter#showdiff


h/browser/chrome/karma.config.js
```diff
@@ -50,14 +49,24 @@ module.exports = function(config) {
       configure: function(bundle) {
         bundle
           .plugin('proxyquire-universal')
-          .require('phantom-ownpropertynames/implement', {entry: true});
+          // fix for Proxyquire in PhantomJS 1.x.
+          // See https://github.com/bitwit/proxyquireify-phantom-menace
+          .require('../../../node_modules/phantom-ownpropertynames/implement',
```
nickstenning: Is there a reason this has changed to a path rather than a package name?

robertknight: Something changed in how the `require()` is resolved here with the dependency upgrade, with the result that the path is always resolved relative to the location of the `karma.config.js` file rather than using the normal Node require resolution algorithm. I haven't dug into the upgraded modules to find out exactly what caused the issue.

However, reading this again it occurred to me that using `require.resolve('phantom-ownpropertynames/implement')` would be a neater that hard-coding the relative location of the `node_modules` folder.

h/static/scripts/karma.config.js
```diff
@@ -61,14 +58,22 @@ module.exports = function(config) {
           .plugin('proxyquire-universal')
           // fix for Proxyquire in PhantomJS 1.x.
           // See https://github.com/bitwit/proxyquireify-phantom-menace
-          .require('phantom-ownpropertynames/implement', {entry: true});
+          .require('../../../node_modules/phantom-ownpropertynames/implement',
```
robertknight: NB: See above comments about the path in the other Karma config file.

nickstenning:
LGTM.

PR #3219, opened 2016-04-15
Dev install docs corrections

seanh:
I feel like it wasn't [that long ago](https://github.com/hypothesis/h/pull/3046) when I last tested these docs on a fresh Ubuntu VM and fixed them, but they've already gotten broken in several new ways :(

Anyway, this fixes the docs so that they work on a fresh Ubuntu 14.04 again (I didn't test OS X, suspect it's slightly broken too), corrects some inconsistencies, and removes some duplication, fixes Sphinx warnings and syntax mistakes etc.

It's not perfect. For one thing I think the [list of requirements](http://h.readthedocs.org/en/latest/hacking/install.html#requirements) at the top (Python and Node) is bizarre. I added git to the list. But the _Installing the system dependencies_ sections that immediately follow tell you to apt-get or brew install what looks like a completely different list of requirements. I'm not sure that the list at the top of the file is any use to anyone and would vote for removing it.

Maybe we could also cut out some of the tips and things, it looks quite cluttered I think and hinders clarity.

Otherwise I think these docs and the install process are quite good compared to many that I've been through.

### DIFF
--- a/docs/hacking/install.rst
+++ b/docs/hacking/install.rst
@@ -117,15 +117,15 @@ installed on:
 
    .. code-block:: bash
 
-      docker run -d --name nsqd -p 4150:4150 -p 4151:4151 nsqio/nsq /nsqd
-      docker run -d --name postgres -p 5432:5432 postgres
-      docker run -d --name redis -p 6379:6379 redis
-      docker run -d --name elasticsearch -p 9200:9200 -p 9300:9300 nickstenning/elasticsearch-icu
+      sudo docker run -d --name nsqd -p 4150:4150 -p 4151:4151 nsqio/nsq /nsqd
+      sudo docker run -d --name postgres -p 5432:5432 postgres
+      sudo docker run -d --name redis -p 6379:6379 redis
+      sudo docker run -d --name elasticsearch -p 9200:9200 -p 9300:9300 nickstenning/elasticsearch-icu
 
    You'll now have four Docker containers named ``nsqd``, ``postgres``,
    ``redis`` and ``elasticsearch`` running and exposing the nsqd service on
    ports 4150 and 4151, Elasticsearch on 9200 and 9300, Redis on 6379, and
-   PostgreSQL on 5432. You should be able to see them by running ``docker ps``.
+   PostgreSQL on 5432. You should be able to see them by running ``sudo docker ps``.
    You should also be able to visit your Elasticsearch service by opening
    http://127.0.0.1:9200/ in a browser, and connect to your PostgreSQL by
    running ``psql postgresql://postgres@localhost/postgres`` (if you have psql
@@ -139,14 +139,14 @@ installed on:
 
       .. code-block:: bash
 
-         docker start postgres elasticsearch nsqd redis
+         sudo docker start postgres elasticsearch nsqd redis
 
 3. Create the `htest` database in the ``postgres`` container. This is needed
    to run the h tests:
 
    .. code-block:: bash
 
-      docker run -it --link postgres:postgres --rm postgres sh -c 'exec psql -h "$POSTGRES_PORT_5432_TCP_ADDR" -p "$POSTGRES_PORT_5432_TCP_PORT" -U postgres -c "CREATE DATABASE htest;"'
+      sudo docker run -it --link postgres:postgres --rm postgres sh -c 'exec psql -h "$POSTGRES_PORT_5432_TCP_ADDR" -p "$POSTGRES_PORT_5432_TCP_PORT" -U postgres -c "CREATE DATABASE htest;"'
 
 
 .. tip::
@@ -157,7 +157,7 @@ installed on:
 
    .. code-block:: bash
 
-      docker run -it --link postgres:postgres --rm postgres sh -c 'exec psql -h "$POSTGRES_PORT_5432_TCP_ADDR" -p "$POSTGRES_PORT_5432_TCP_PORT" -U postgres'
+      sudo docker run -it --link postgres:postgres --rm postgres sh -c 'exec psql -h "$POSTGRES_PORT_5432_TCP_ADDR" -p "$POSTGRES_PORT_5432_TCP_PORT" -U postgres'
 
    This runs psql in a fourth Docker container (from the same official
    PostgreSQL image, which also contains psql) and links it to your named
@@ -172,7 +172,7 @@ installed on:
 
    .. code-block:: bash
 
-      docker logs nsqd
+      sudo docker logs nsqd
 
    For more on how to use Docker see the `Docker website`_.
 
--- a/docs/hacking/install.rst
+++ b/docs/hacking/install.rst
@@ -3,7 +3,7 @@ Installing h in a development environment
 
 This document contains instructions for setting up a development environment
 for h. If you are looking for instructions for deploying h in a production
-environment, please consult the :doc:`/INSTALL` instead.
+environment, please consult the :doc:`/INSTALL` document instead.
 
 
 Requirements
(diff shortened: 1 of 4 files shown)


docs/hacking/install.rst
```diff
@@ -114,10 +96,13 @@ Install the following packages:
 Installing the services
 -----------------------
 
-h requires Elasticsearch_ 1.0+ with the `Elasticsearch ICU Analysis`_ plugin,
-`NSQ`_ 0.3+ and `PostgreSQL`_ 9.4+. You can install these services however you
-want, but the easiest way is by using Docker. This should work on any operating
-system that Docker can be installed on:
+h requires trhe following external services:
```
chdorner: `:%s/trhe/the/`

docs/hacking/install.rst
```diff
@@ -186,64 +171,139 @@ system that Docker can be installed on:
...
+When ``gulp watch`` is running, you can visit http://localhost:3000
 to see a page with an embedded Hypothesis client which will automatically reload
 when styles, templates or JavaScript source files are changed.
 
 .. _Gulp: http://gulpjs.com/
 
-If you are running background tasks, you must have a Celery worker running. To
-run a Celery worker in development, run:
+
+Running the WebSocket for realtime features
```
chdorner: `:%s/realtime/real-time/`

docs/hacking/install.rst
```diff
@@ -186,64 +171,139 @@ system that Docker can be installed on:
...
+-------------------------------------------
+
+To test h's realtime features (for example, seeing new or updated annotations
```
chdorner: `:%s/realtime/real-time/`

chdorner:
This looks good to me. I'll leave it to @nickstenning to merge since I know you too were talking on Slack about the installation docs.

When this or #3217 gets merged, the other pull request needs changing again.

seanh:
Fixed the two that Christof pointed out

docs/hacking/install.rst
```diff
@@ -131,15 +116,15 @@ system that Docker can be installed on:
 
    .. code-block:: bash
 
-      docker run -d --name nsqd -p 4150:4150 -p 4151:4151 nsqio/nsq /nsqd
-      docker run -d --name postgres -p 5432:5432 postgres
-      docker run -d --name redis -p 6379:6379 redis
-      docker run -d --name elasticsearch -p 9200:9200 -p 9300:9300 nickstenning/elasticsearch-icu
+      sudo docker run -d --name nsqd -p 4150:4150 -p 4151:4151 nsqio/nsq /nsqd
```
nickstenning: I'm not sure about adding `sudo` here. It's entirely dependent on the configuration of docker whether or not using the docker CLI requires you to be root. Following the default instructions for setting up docker on OS X, for example, you don't need to be root to run `docker`.

This really boils down to so much of the struggle with writing documentation -- do we have a responsibility to document other people's systems (how to install and configure postgres, for example). In this case, I'd vote explain what you need to do to docker, and leave it up to people to work out how docker works on their system.

docs/hacking/install.rst
```diff
@@ -186,64 +171,139 @@ system that Docker can be installed on:
...
 
    For more on how to use Docker see the `Docker website`_.
 
 
 .. _Docker website: https://www.docker.com/
 
 
-Running h
----------
+Installing the gulp command
```
nickstenning: I'm not sure I understand why this is here. It has no context and more importantly it's not necessary. You don't need gulp installed globally to run `make dev`.

docs/hacking/install.rst
```diff
@@ -186,64 +171,139 @@ system that Docker can be installed on:
...
+
+Change into the ``h`` directory from the remainder of the installation
+process:
+
+.. code-block:: bash
+
+   cd h
+
+
+Creating a Python virtual environment
```
nickstenning: This is another case where I really can't endorse us attempting to teach people how the entire Python ecosystem works. If you want to link to http://docs.python-guide.org/en/latest/dev/virtualenvs/ or something, that would be great. But I *really* don't want us to have to maintain a set of documentation on how to use virtualenv.

docs/hacking/install.rst
```diff
@@ -186,64 +171,139 @@ system that Docker can be installed on:
...
+
+   gunicorn --reload --paste conf/development-websocket.ini
+
+
+Running Celery for background tasks
+-----------------------------------
+
+To test h's background tasks (for example, sending emails) you need to run a
+Celery worker process.
+Open a new terminal, ``cd`` into your ``h`` directory,
```
nickstenning: Really? Do we need to tell them to turn their computer on first, too?

docs/hacking/install.rst
seanh: Ah, on linux these commands will fail without sudo unless you've taken additional steps, so I [added sudo to make it consistent with other commands in these docs that require sudo](https://github.com/hypothesis/h/pull/3219/commits/8624a4bbc11f49e7c5dc5a82e55c1b8df703b06f). If you don't then the commands fail with an error message that doesn't really suggest that it's a sudo / permissions thing. This is an OS X vs Linux difference then. We could remove the sudos, and add half a sentence somewhere to say "install docker and if necessary (e.g. on linux) configure it to run without sudo"?

docs/hacking/install.rst
seanh: Gulp is used in multiple places further down the page, so moving it here removed duplication and also I think it's clearer to have separate steps like "install gulp" in separate sections. https://github.com/hypothesis/h/pull/3219/commits/c55ba7ba0ba219490b8f75c58afc262171f4b8d4

docs/hacking/install.rst
seanh: I do link to the hitchhiker's guide on venvs (actually the link was already there and I left it). Creating a venv [is necessary for it to work](https://github.com/hypothesis/h/blob/master/Makefile#L8). I don't think these docs'll be a huge maintenance burden, it's one line plus a tip box that links to an external guide

docs/hacking/install.rst
nickstenning: I think adding a sentence that says "you may need to run `docker` commands as root" would be enough.

docs/hacking/install.rst
seanh: What I have gone for here (and in general on this page) is to keep the docs as short and concise as possible (one sentence in this case), linking to external resources for further reading where relevant, while also making it possible to follow the docs without any prior knowledge.

Of course "without prior knowledge" is open to interpretation so I don't explain what a computer is or anything.

But this sentence is making it clear that hypothesis-celery is a separate command that needs to be run in a separate terminal, from the h dir, and with the venv activated. Again I think the burden to maintaining this is low and it should be helpful to people who aren't Python devs

seanh: I guess the question here is who is the audience for these docs, and what level of technical knowledge do they have? I set the bar lower than "Python development experience", someone who knows how to use a computer but has never done Python dev (so doesn't know what a virtualenv is, how to activate one, or that you need to re-activate it in each shell) should be able to follow these docs. So I tried to give enough instructions for that person, and then keep it as minimal as possible. So for example not explaining what a "terminal" is or how to open it.

nickstenning:
Okay, this will definitely now need a bit of a tweak to account for the changes to the docs in #3217.

nickstenning:
But I guess that can come later...

nickstenning:
I'm still not super keen for us to document how to use virtualenv in our developer documentation, but this is obviously an improvement to our dev docs overall, so I'll merge it now.

PR #3222, opened 2016-04-15
Convert admin to package

nickstenning:
Another module that had grown pretty darn unwieldy. I've split this into a package with separate views modules for each section in the admin. Essentially no code changes.

I hope that over time, the core business logic of some of these views can move into their modules in `h.admin` such as `h.admin.users`, `h.admin.groups`, etc., leaving the views trivial.


chdorner:
This is looking much better now!

PR #3224, opened 2016-04-18
Add a helper function for writing parameterized JS tests

robertknight:
Mocha lacks built-in support [1] for writing parameterized tests and the
suggested solution [2] involves a bunch of boilerplate which has IMO
resulted in different styles of parameterized tests in our codebase and
not having parameterized tests when they would be useful to attain more
complete coverage.

This adds a helper inspired by [3] for writing parameterized tests and
switches several existing places in our code to use it.

[1] https://github.com/mochajs/mocha/issues/1454
[2] https://mochajs.org/#dynamically-generating-tests
[3] https://github.com/lawrencec/Unroll

### DIFF
--- a/h/browser/chrome/test/uri-info-test.js
+++ b/h/browser/chrome/test/uri-info-test.js
@@ -1,9 +1,10 @@
 'use strict';
 
+var toResult = require('../../../static/scripts/test/promise-util').toResult;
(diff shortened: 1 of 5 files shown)


chdorner:
Nice and simple solution! LGTM.

PR #3226, opened 2016-04-19
CSP Violation: old homepage

chdorner:
This fixes various CSP violations on the old homepage:

* We had inline styles for the press release banner
* We had inline Javascript for toggling the Chrome extension / Bookmarklet buttons
* `onclick` handler of the Chrome extension install button
* `onclick` handler of the Bookmarklet install button

Since I'm pretty much just moving existing (untested) code around, and because it is the old homepage which will hopefully not be around for much longer, I took the liberty to not write tests for the Javascript part.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -162,6 +162,7 @@ var styleFiles = [
   './h/static/styles/front-page/main.css',
   './h/static/styles/help-page.scss',
   './h/static/styles/site.scss',
+  './h/static/styles/old-home.scss',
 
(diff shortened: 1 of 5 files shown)


chdorner:
The `'chrome' is not defined` Hound warning still exists, but I think it can be safely ignored?

nickstenning:
I reckon this should probably move into a `DOMContentLoaded` handler, but it seems to work for now, so LGTM.

PR #3228, opened 2016-04-19
CSP violation: client config on /docs/help and /welcome

chdorner:
This fixes the CSP violation we had due to the inline script that contains the client configuration. There is a new way to configure the client which is a script tag with type `application/json` and the class `js-hypothesis-config` which does not trigger a CSP error.

### DIFF
--- a/h/templates/help.html.jinja2
+++ b/h/templates/help.html.jinja2
@@ -3,9 +3,7 @@
 {% block meta %}
   <link href='//fonts.googleapis.com/css?family=Lato:400,300' rel='stylesheet' type='text/css'>
   {{ super() }}
-  <script>
-    function hypothesisConfig() { return {firstRun: true}; }
-  </script>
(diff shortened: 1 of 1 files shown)


h/templates/help.html.jinja2
```diff
@@ -3,9 +3,7 @@
 {% block meta %}
   <link href='//fonts.googleapis.com/css?family=Lato:400,300' rel='stylesheet' type='text/css'>
   {{ super() }}
-  <script>
-    function hypothesisConfig() { return {firstRun: true}; }
-  </script>
+  <script type="application/json" class="js-hypothesis-config">{"firstRun":"true"}</script>`
```
robertknight: The `true` here should be a boolean value rather than a string. I expect it works because the code just does `if (config.firstRun)`, but still better to use the right data type.

chdorner: :+1: Fixed.

nickstenning:
:+1:

PR #3229, opened 2016-04-19
Do not authenticate users who do not exist

seanh:
https://app.getsentry.com/hypothesis/prod/issues/123706115/activity/

### DIFF
--- a/h/auth/policy.py
+++ b/h/auth/policy.py
@@ -4,6 +4,7 @@ from pyramid import authentication
 from pyramid import interfaces
 from zope import interface
 
+from h import accounts
 from h.auth import tokens
 from h.auth import util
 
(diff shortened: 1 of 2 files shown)


nickstenning:
I'm not sure about this solution. This is adding a round-trip to the database for every authenticated request, which I'd like to avoid.

I think the "right" way to do this would probably be to drop the relevant session object from redis when delete a user. While worth investigating, I think that's probably not possible at the moment given the way that the session store is keyed, so perhaps a better alternative (given the rarity of the problem this is supposed to fix) is simply to detect the issue in the callable that implements the `request.authenticated_user` property, and invalidate the session and raise a redirect at that point.

seanh:
Closed in favour of https://github.com/hypothesis/h/pull/3231

PR #3231, opened 2016-04-19
Log the user out if they don't exist

### LINKED ISSUE #3227: Deleted account group read crash
1. Join (don't create) a group
2. Have a different (admin) user delete your account, from another browser
3. You're still authenticated in your browser even though your account has been deleted. Try to load the group's page -> crash.

The problem is that in the case of an authenticated session fro a user account `request.authenticated_userid` is set but `request.authenticated_user` is `None`, and the code doesn't expect that.

```
AttributeError: 'NoneType' object has no attribute 'groups'
  File "pyramid/tweens.py", line 20, in excview_tween
    response = handler(request)
  File "pyramid_tm/__init__.py", line 101, in tm_tween
    reraise(*exc_info)
  File "pyramid_tm/__init__.py", line 83, in tm_tween
    response = handler(request)
  File "h/tweens.py", line 24, in conditional_http_tween
    response = handler(request)
  File "pyramid/router.py", line 145, in handle_request
    view_name
  File "pyramid/view.py", line 541, in _call_view
  File "pyramid/config/views.py", line 602, in __call__
    return view(context, request)
  File "pyramid/config/views.py", line 328, in attr_view
    return view(context, request)
  File "pyramid/config/views.py", line 304, in predicate_wrapper
    return view(context, request)
  File "pyramid/config/views.py", line 385, in viewresult_to_response
    result = view(context, request)
  File "pyramid/config/views.py", line 507, in _requestonly_view
    response = view(request)
  File "h/groups/views.py", line 172, in read
    if group in request.authenticated_user.groups:
```

seanh:
Problem:

It's possible for request.authenticated_userid to be set because the user has
a valid browser session but request.authenticated_user to be None at the same
time because the user doesn't exist in the database.

This happens when we delete a user account because we don't (and can't
easily) invalidate the user's session at delete time.

This leaves the application in an invalid state (contradiction between
request.authenticated_userid and request.authenticated_user, and
request.authenticated_userid referring to a user that no longer exists) and
parts of the code that don't expect this will crash.

Solution:

Ideally we would invalidate the user's sessions when we delete them but
this is difficult to do. So instead:

If a piece of code tries to access request.authenticated_user and the
user isn't found in the db then invalidate the user's session and
redirect the browser.

The user will be logged out.

We redirect to the same URL that was requested so that for example if you were
loading the group read page you'll get the logged-out version of the group read
page and not some other page.

We decided to do this when request.authenticated_user is requested because it
avoids doing an additional database lookup (to check that
request.authenticated_user would match what request.authenticated_userid says)
for requests that would not have needed request.authenticated_user.

This avoids an additional db lookup on every request. On the other hand
deleted users can still remain authenticated and do things as long as they
don't make a request that tries to use request.authenticated_user.

Add tests for the new solution and also clean up a few things in the
existing authenticated_user() tests:

- Introduce @authenticated_user_fixtures, apply the same fixtures to all
  authenticated_user() tests without repetition
- Make get_user a fixture to work with @authenticated_user_fixtures
- Use DummyRequest and authn_policy, more realistic test mocks
- Add test for the return value, was untested
- Small improvements to test docstrings

Fixes #3227

### DIFF
--- a/h/accounts/__init__.py
+++ b/h/accounts/__init__.py
@@ -1,5 +1,6 @@
 # -*- coding: utf-8 -*-
 from itsdangerous import URLSafeTimedSerializer
+from pyramid import httpexceptions
 
 from h.security import derive_key
 
@@ -79,8 +80,18 @@ def authenticated_user(request):
     :rtype: h.accounts.models.User or None
 
     """
-    return get_user(request.authenticated_userid, request)
+    user = get_user(request.authenticated_userid, request)
 
+    # If the authenticated user doesn't exist in the db then log them out.
+    # This happens when we delete a user account but the user still has a
+    # valid session for that account (because we don't invalidate sessions
+    # when we delete user accounts).
+    if request.authenticated_userid and not user:
+        request.session.invalidate()
+        raise httpexceptions.HTTPFound(
+            location=request.current_route_url())
+
+    return user
 
 def includeme(config):
     """A local identity provider."""
--- a/h/accounts/test/__init___test.py
+++ b/h/accounts/test/__init___test.py
@@ -1,6 +1,8 @@
 # -*- coding: utf-8 -*-
 import mock
 import pytest
+from pyramid import httpexceptions
+from pyramid import testing
 
 from h import accounts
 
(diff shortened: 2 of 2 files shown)


nickstenning:
LGTM.

PR #3232, opened 2016-04-20
Preserve page query string when rendering feed URLs

### LINKED ISSUE #3003: RSS Feeds for tag view is not correct
Here are tagged annotations for a class project just getting started:
https://hypothes.is/stream?q=tag:inf115

The RSS feed in the source is the general wide hypothesis stream

`<link rel="alternate" type="application/atom+xml" href="https://hypothes.is/stream.atom"/>`
`<link rel="alternate" type="application/rss+xml" href="https://hypothes.is/stream.rss"/>`

but the feeds I found for this tag by guessing are

`https://hypothes.is/stream.rss?tag=inf115`
`https://hypothes.is/stream.atom?tag=inf115`

So feeds are not discoverable (or identified).

nickstenning:
The Atom and RSS feed URLs were always to the unfiltered stream. This makes sure that the query string used for the request to the stream page is also used in the URL to the feeds.

Fixes #3003.

### DIFF
--- a/h/views/main.py
+++ b/h/views/main.py
@@ -55,8 +55,8 @@ def robots(context, request):
 
 @view_config(route_name='stream')
 def stream(context, request):
-    atom = request.route_url('stream_atom')
-    rss = request.route_url('stream_rss')
+    atom = request.route_url('stream_atom', _query=request.query_string)
+    rss = request.route_url('stream_rss', _query=request.query_string)
     return render_app(request, {
         'link_tags': [
             {'rel': 'alternate', 'href': atom, 'type': 'application/atom+xml'},


chdorner:
This doesn't seem to work when I'm testing it out locally.
When I search the stream for my annotations I search for `acct:chdorner@localhost`, which goes to `/stream?q=acct:chdorner@localhost`. But this only renders the app which then does a search query through the API with `/api/search?any=acct%3Achdorner%40localhost&...`.

The feeds (atom/rss) are searching [directly](https://github.com/hypothesis/h/blob/master/h/feeds/views.py#L14) though. Which means that to make it work we need to pretty much rename the `q` param to `any`: `/stream.atom?any=acct:chdorner@localhost`

nickstenning:
You're absolutely right. This was stupid on my part. I assumed that that `/stream` and `/stream.{atom,rss}` used the same query param format. (Silly me... 🙄)

PR #3233, opened 2016-04-20
Fix Hypothesis on local HTML documents

### LINKED ISSUE #2005: No check for limitation on embedding to local HTML
When I open a local HTML file in FF, and try to deploy H using the bookmarklet, I get a pop-up with an error message explaining that 

> Sorry, Hypothesis doesn't work on this type of file. Only PDF's can be annotated locally.

That's all well and good, but when I try to do the same thing with an embed (that is, open a local HTML file that has H embedded), I get no explanation, only a crash on the console:

> SyntaxError: An invalid or illegal string was specified

Since we are bumping into the same limitation, ideally, we should use the same check, and the same warning message.

nickstenning:
The only thing stopping Hypothesis from working on local (i.e. file:// URL) HTML documents is the failure of the discovery module due to quirks in the `postMessage` security policy.

This commit works around that issue so that embedding or injecting Hypothesis on local HTML documents works as expected.

Fixes #2005.

### DIFF
--- a/h/static/scripts/discovery.coffee
+++ b/h/static/scripts/discovery.coffee
@@ -86,6 +86,15 @@ module.exports = class Discovery
   _onMessage: (event) =>
     {source, origin, data} = event
 
+    # If `origin` is 'null' the source frame is a file URL or loaded over some
+    # other scheme for which the `origin` is undefined. In this case, the only
+    # way to ensure the message arrives is to use the wildcard origin. See:
+    #
+    #   https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage
+    #
+    if origin is 'null'
+      origin = '*'
+
     # Check if the message is at all related to our discovery mechanism
     match = data.match? /^__cross_frame_dhcp_(discovery|offer|request|ack)(?::(\d+))?$/
     return unless match
--- a/h/static/scripts/test/discovery-test.coffee
+++ b/h/static/scripts/test/discovery-test.coffee
@@ -74,6 +74,18 @@ describe 'Discovery', ->
       matcher = sinon.match(/__cross_frame_dhcp_ack:\d+/)
       assert.calledWith(fakeTopWindow.postMessage, matcher, 'top')
 
+    it 'sends an "ack" to the wildcard origin if a request comes from a frame with null origin', ->
+      fakeFrameWindow.addEventListener.yields({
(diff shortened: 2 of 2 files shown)


chdorner:
I'm still getting the error saying that the extension doesn't work with local HTML documents (I did `gulp build && hypothesis-buildext ...` which I believe is all that is necessary to get a new development extension).

judell:
> Fixes #2005.

Wow. Sweet!

nickstenning:
Yep, @chdorner. This doesn't update the extension to work on local HTML documents, but it does lift the restriction that made that warning necessary. If you want to test this, you can add a `<script src="http://localhost:5000/embed.js" async defer></script>` or similar to a local HTML document and check that it works as expected.

h/static/scripts/discovery.coffee
```diff
@@ -86,6 +86,15 @@ module.exports = class Discovery
   _onMessage: (event) =>
     {source, origin, data} = event
 
+    # If `origin` is 'null' the source frame is a file URL or loaded over some
+    # other scheme for which the `origin` is undefined. In this case, the only
+    # way to ensure the message arrives is to use the wildcard origin. See:
+    #
+    #   https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage
```
robertknight: I couldn't see how to derive from [the spec](https://hyp.is/AVQ4OV7DH9ZO4OKSl1Cg/www.w3.org/TR/2008/WD-html5-20080610/comms.html) that "null" would definitely be used as the serialization of the origin in the case where the source frame is a file URL, or whether this value might be UA-defined.

I'm currently doing some testing in a few different browsers I have available.

robertknight: Firefox and Chrome exhibit the same behavior at looking at the implementation, it looks like they are referencing this spec: https://hyp.is/AVQ4XdxnH9ZO4OKSl1Db/tools.ietf.org/html/rfc6454

robertknight:
LGTM.

robertknight:
@nickstenning So what is left to do to enable this, aside from removing the checks in the Chrome extension that currently prevent injection into local HTML pages? There was an open task to avoid including `file://` URLs in annotations. Once we've done that, it is possible that we won't have any URIs at all for HTML documents (if they include no meta tags etc.).

PR #3234, opened 2016-04-20
Fix an off-by-one error in PDF anchoring code

### LINKED ISSUE #2418: Annotation editor fails to appear for certain PDF selections
### Steps to reproduce:

1. Start Hypothesis on http://courses.csail.mit.edu/18.337/2015/docs/50YearsDataScience.pdf.
1. Go to page 8.
1. Select "Lost in the hoopla about such skills"
1. Adder appears, click New Note

### Expected behaviour

New, empty annotation appears for editing in the sidebar.

### Actual behaviour

Nothing appears! Note that if you select "ost in the hoopla about such skills" it seems to work just fine.

nickstenning:
The PDF anchoring code systematically failed to anchor annotations made at the very start of a PDF page, due to an off-by-one error in determining the page associated with a given text offset.

Fixes #2418.

### DIFF
--- a/h/static/scripts/annotator/anchoring/pdf.coffee
+++ b/h/static/scripts/annotator/anchoring/pdf.coffee
@@ -62,8 +62,34 @@ findPage = (offset) ->
   index = 0
   total = 0
 
+  # We call `count` once for each page, in order. The passed offset is found on
+  # the first page where the cumulative length of the text content exceeds the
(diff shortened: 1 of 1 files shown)


chdorner:
LGTM. Tested it out locally and I can't reproduce the problem as described in #2418 anymore.

PR #3235, opened 2016-04-20
Remove sudo from docker commands in docs

seanh:
An awkward thing about our dev install docs is:

- On Linux you need to use sudo with all docker commands or you'll get
  an unhelpful error message.
- Unless you took additional steps during docker install to enable
  docker without sudo, in which case sudo is unnecessary but does no
  harm
- On OS X sudo docker is always unnecessary, but does no harm

Remove sudo from all docker commands, the docs now assume that linux
users took the extra steps in the docker install docs to enable no sudo,
and add the error message that they would get if they didn't to a new
troubleshooting section.

Troubleshooting sections at the bottoms of pages can be handy in docs:

- Keep lots of "if this happens then do this" details from cluttering
  the main flow of the docs

- While letting you cover as many potential problems as you like. It scales, this list at the bottom of the page can grow as long as it wants without harming the structure and flow of the rest of the docs.

- Put lots of problems in one place, each with their own subtitle, easy
  to find, easy to scan or search for

- Clear and simple docs structure. If you want to document a problem and
  solution, now you know where.

- Works well when the same problem can occur at multiple places in the
  instructions, we only need to document it in one place

- If some problem and solution is no longer relevant it can be deleted,
  but in the likely event that it does not get deleted it's not really
  doing any harm sitting at the bottom of the page where no one will
  look at it unless they get that error

### DIFF
--- a/docs/hacking/install.rst
+++ b/docs/hacking/install.rst
@@ -116,15 +116,15 @@ installed on:
 
    .. code-block:: bash
 
-      sudo docker run -d --name postgres -p 5432:5432 postgres
-      sudo docker run -d --name elasticsearch -p 9200:9200 -p 9300:9300 nickstenning/elasticsearch-icu
-      sudo docker run -d --name rabbitmq -p 5672:5672 -p 15672:15672 --hostname rabbit rabbitmq:3-management
-      sudo docker run -d --name redis -p 6379:6379 redis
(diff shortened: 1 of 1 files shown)


seanh:
@nickstenning

nickstenning:
:+1:

PR #3238, opened 2016-04-20
[Update annotations in Pg 1/3]: Fix reply template for postgres

seanh:


seanh:
I've tested manually:

* You don't get notified for newly created annotations (as opposed to replies)
* You don't get notified on annotation edits
* You don't get notified when replying to your own annotation, or to your own reply
* You don't get notified when someone makes a private reply to your annotation or reply
* If you make an annotation in a group, then leave the group, then someone replies to it, you don't get notified
* You _do_ get notified when someone makes a public or group reply to your annotation or reply

All with and without postgres on. No crashes for any of the above. Notification emails look right.

seanh:
@chdorner will probably want to have a look at some of the model code changes here

h/api/models/annotation.py
```diff
@@ -103,6 +103,21 @@ def document(self):
         if self.documents:
             return self.documents[0]
 
+    @property
+    def parent_id(self):
+        """
+        Return the ID of the annotation that this annotation is a reply to.
+
+        Returns the ID of the top-level annotation at the top of the thread,
```
nickstenning: I don't think this docstring is accurate. The function returns `references[-1]`, which is the immediate ancestor, not the thread root.

h/api/models/annotation.py
```diff
@@ -103,6 +103,21 @@ def document(self):
...
+        regardless of whether this reply is a direct reply to that annotation
+        or a reply to a reply.
+
+        Returns None of this annotation is not a reply.
```
nickstenning: s/of/if/

h/api/models/document.py
```diff
@@ -37,7 +37,7 @@ def title(self):
         titles = [m.value for m in self.meta if m.type == 'title']
 
         if titles:
-            return titles[0]
+            return titles[0][0]
```
nickstenning: Something has clearly gone wrong here. The value of a single meta instance should not be an array, I don't think.

h/notification/reply_template.py
```diff
@@ -106,17 +98,24 @@ def generate_notifications(request, annotation, action):
...
-    # of the reply.
-    child_read_permissions = annotation.get('permissions', {}).get('read', [])
-    parent_principals = auth.effective_principals(parent['user'], request)
-    read_principals = translate_annotation_principals(child_read_permissions)
-    if not set(parent_principals).intersection(read_principals):
+    # Don't send reply notifications to the author of the parent annotation if
+    # the reply is in a group that the author of the parent annotation isn't
+    # (the author of the parent annotation won't be able to read the reply).
+    parent_principals = auth.effective_principals(parent.userid, request)
+    if annotation.groupid == '__world__':
```
nickstenning: This seems a bit fragile. Might be better to fetch the current authorization policy and use `policy.permits(annotation, parent_principals, 'read')`. That would probably eliminate some code higher up here, too.

h/api/models/document.py
nickstenning: Had a look at the `DocumentMeta` model and I'm clearly wrong, but this seems messy, as well as prone to failure. What if a meta row exists but the value is an empty list

I assume what's supposed to be happening is that there is only supposed be one meta row for each document with type 'title' -- indeed I imagine that's enforced by a database constraint -- but as with all meta objects, we can have multiple values (`<title>`, DC metadata title, etc.).

It just needs to be a bit clearer that one of these arrays is to deal with the case that we have no title information, and the other is to deal with the fact that we can have multiple values for the title.

h/notification/test/reply_template_test.py
```diff
@@ -227,6 +232,43 @@ def test_unsubscribe_url_generation():
 
     request.route_url.assert_called_with('unsubscribe', token='TOKEN')
 
+
+class TestFormatTimestamp:
+
+    @pytest.mark.parametrize('in_date,out_str', [
+        (
+            datetime.datetime(2016, 04, 14, 16, 45, 36, 529730),
```
nickstenning: `04` is a number in octal format. Probably not what you wanted here.

h/api/models/elastic.py
```diff
@@ -117,7 +117,7 @@ def shared(self):
         elif len(read_perms) > 1 and self.groupid == '__world__' and gperm in read_perms:
             read_perms = [gperm]
 
-        if (read_perms != [gperm] and read_perms != [self.userid]):
+        if (gperm not in read_perms and read_perms != [self.userid]):
```
nickstenning: What's the motivation for this change? This entire function is intended to be very conservative in its interpretation of the permissions field, precisely so we can be sure we're importing correct data into PostgreSQL.

If there are annotations where `read_perms` hasn't been successfully mapped to a single-element list by one of the clauses immediately above this line, that's a problem, and we shouldn't fix it by relaxing expectations here.

h/notification/reply_template.py
```diff
@@ -26,55 +26,47 @@
...
-def format_timestamp(timestamp):
-    # Currently we cut the UTC format because time.strptime has problems
-    # parsing it, and of course it'd only correct the backend's timezone
-    # which is not meaningful for international users. This trims the
-    # timezone in the format +00:00.
-    timestamp = re.sub(r'\+\d\d:\d\d$', '', timestamp)
-    timestamp_format = '%Y-%m-%dT%H:%M:%S.%f'
-    parsed = datetime.strptime(timestamp, timestamp_format)
-
+def format_timestamp(timestamp, now=datetime.now):
```
nickstenning: I know you're just preserving behaviour, but this should definitely be `utcnow`.

h/api/models/annotation.py
seanh: Huh. I'll check this

h/api/models/annotation.py
seanh: Done

h/api/models/document.py
seanh: I'm not sure what the model should be here:

1. _Should_ the DocumentMeta model have an array for the value column as it does? Does it have to? If not we can change the model to just have a string

2. If it does have to be an array, is there always at least one item in the array? Or we have to watch our for empty lists? Does it ever make sense  to have a DocumentMeta like this:

   ```
   type: "title",
   value: [],
   ...
   ```

We should probably add a `title` or maybe `first_title` property to the `DocumentMeta` class to return either the first title or an empty string

h/notification/test/reply_template_test.py
seanh: Done

h/api/models/elastic.py
seanh: I didn't realise that the function was supposed to be as conservative as possible in its interpretation.

The reason for this change is that the existing code will match:

```
["group:wibble"]
```

but will _not_ match:

```
["group:wibble", "acct:seanh@hypothes.is"]
```

as being shared, which of course isn't a _correct_ interpretation of that ACL. And the `reply_template` tests do use `["group:wibble", "acct:seanh@hypothes.is"]`-type read ACLs in some global shared fixtures and expect those annotations to be treated as shared or the tests fail.

The alternative would be to fix those tests, either to just use `["group:wibble"]` or to patch the model code.

h/notification/reply_template.py
seanh: Minimising changes to legacy code, but I don't see how this could break anything... Done

h/api/models/annotation.py
seanh: Done

h/notification/reply_template.py
seanh: Stupid question but I don't see how to fetch the current authz policy so that I can call permits()?

h/api/models/document.py
chdorner: The reason why it is an array is due to the fact that there _can_ be multiple values depending on the type of value we store. The annotator [uses arrays](https://github.com/openannotation/annotator/blob/v1.2.x/src/plugin/document.coffee#L68-L81).
Then again, things are not as consistent, the title for example is [never an array](https://github.com/openannotation/annotator/blob/v1.2.x/src/plugin/document.coffee#L83-L97).

chdorner: And, I would rather stay on the safe side and not risk an `IndexError` here.

The only reason this method exists is to have a consistent API with the `elastic.Document` if I remember correctly. And I [recently](https://github.com/hypothesis/h/commit/6ea11a67f4a12f399eb82ddf5e3c72670a64cfa0) changed this very method in `elastic.Document` but forgot to change the PG one.

It should behave the same way, so maybe it should look something along the lines of:
```python
@property
def title(self):
    titles = [m.value for m in self.meta if m.type == 'title']
    try:
        return titles[0][0]
    except IndexError:
        return None
```

h/api/models/elastic.py
chdorner: It's worth noting that all the code in `h.api.models.elastic` is being used in the postgres migration script. So any change would need to be tested by running the migration again.

nickstenning: > The alternative would be to fix those tests, either to just use ["group:wibble"] or to patch the model code.

If the existing tests don't match real data that exists in Elasticsearch, then yes, let's fix the tests.

h/notification/reply_template.py
seanh: Here's a go at a solution to this: https://github.com/hypothesis/h/pull/3238/commits/cf262939b0fc2976380346aa2088a3343637b01f Fixing the tests / writing tests for it now

h/api/models/elastic.py
seanh: I think the failing tests got fixed by some other test changes I made, so I've rebased this commit out of the pr

h/api/models/document.py
seanh: Done as Christof suggests

seanh:
All done. I haven't re-tested manually to make sure it actually works, will do that once you guys are happy with it

h/auth/util.py
```diff
@@ -54,6 +55,29 @@ def group_principals(user):
...
+                          'acct:philip@hypothes.is',
+                          'read'):
+            print 'philip can read this annotation'
+        else:
+            print "philip isn't allowed to read this annotation"
+
+    """
+    policy = request.registry.queryUtility(authorization.IAuthorizationPolicy)
+    users_principals = effective_principals(userid, request)
+    return bool(policy.permits(context, users_principals, permission))
```
chdorner: The implementation is pretty straightforward, but considering it is auth related I would sleep better at night if we had some tests for this function.

chdorner:
This is looking good to me now, would just like to see some tests for the `has_permission` function.

h/auth/util.py
seanh: I'll wait to see if this is what @nickstenning was asking for when he asked for this code to be simplified (full commit for this bit is <https://github.com/hypothesis/h/pull/3238/commits/b8c4d1061a1b6cacb8c3deee7c4cd47e2411a557>), if it's good then I'll add tests for it.

I think mocking the authorization policy might makes tests for this bit of code kind of pointless - "It retrieves the (mock) authorization policy; It calls the (mock) effective principals; It calls permits; It returns True or False if the mock permits() returns truthy or falsey" - that seems like just describing the code line by line not testing it. So maybe I'll test against Pyramid's real `ACLAuthorizationPolicy`.

nickstenning: Agreed that this should be tested. You don't need to mock everything. As you suggested, just set up a basic check that with an `ACLAuthorizationPolicy`, and an appropriate context object, this returns the right values on the basis of stubbed effective principals and the permission name.

h/auth/util.py
```diff
@@ -1,5 +1,6 @@
 # -*- coding: utf-8 -*-
 
+from pyramid import authorization
```
nickstenning: `IAuthorizationPolicy` is supposed to come from `pyramid.interfaces`, not `pyramid.authorization`.

nickstenning:
Looking pretty close to ready.

nickstenning:
I'm going to go ahead and merge this. I have some tests for the `has_permission` that I can add, and a branch which builds on this to purge some of the really terrible code from the `h.notification` package.

Issue #3236, opened 2016-04-20
annotations missing at https://blog.jonudell.net/2016/04/02/owning-and-sharing-your-words/


judell:
### Repro

Visit https://blog.jonudell.net/2016/04/02/owning-and-sharing-your-words/

### Expected behaviour
These annotations:

AVPdOBrbH9ZO4OKSlh7j
AVPdWm4DH9ZO4OKSlh9Z
AVPhMs-TH9ZO4OKSljDi
AVPhNXY3H9ZO4OKSljD0
AVPhOaUfH9ZO4OKSljEk
AVPg-X33H9ZO4OKSli8a
AVPg89q2H9ZO4OKSli71
AVPg7kYVH9ZO4OKSli7i
AVPhBeBJH9ZO4OKSli9O
AVPhG7bEH9ZO4OKSli-x
AVPhIapZH9ZO4OKSli_s
AVPhMIf9H9ZO4OKSljDV
AVPhKo54H9ZO4OKSljC2
AVPhB8GtH9ZO4OKSli9Q
AVPhHOFuH9ZO4OKSli-1
AVPhMcUDH9ZO4OKSljDd
AVPhNq6KH9ZO4OKSljEC
AVPhoScOH9ZO4OKSljRE
AVPhk8_nH9ZO4OKSljPd
AVPg5ZNjH9ZO4OKSli7B
AVPhImEmH9ZO4OKSljAH
AVPhKUroH9ZO4OKSljCr
AVPg682nH9ZO4OKSli7b
AVPg9jPTH9ZO4OKSli7_
AVPg7HbRH9ZO4OKSli7d
AVPhC4PZH9ZO4OKSli9h
AVPg-kpXH9ZO4OKSli8g
AVPhJuUzH9ZO4OKSljCd
AVPhEuK-H9ZO4OKSli9y
AVPhLq9RH9ZO4OKSljDO
AVPhM9RHH9ZO4OKSljDn
AVPhNeQRH9ZO4OKSljD8
AVPhPMfgH9ZO4OKSljE3
AVPhOOntH9ZO4OKSljEc
AVPivBCZH9ZO4OKSljtH
AVPuRnLVH9ZO4OKSlm04
AVPnhEDBH9ZO4OKSlk4f
AVPg6Ap5H9ZO4OKSli7J
AVPg712-H9ZO4OKSli7m
AVPhGWgAH9ZO4OKSli-U
AVPxk6DgH9ZO4OKSlnZq
[...]

PR #3241, opened 2016-04-22
Return all canonical uris when expanding a canonical uri

chdorner:
_tldr;_ This would let us kill two birds with one stone, it allows us to surface old and new annotations 1) when the canonical uri changes, 2) a proposed fix for #3236 

_Long version:_
Before we always only returned the passed-in uri when it ended up being
a canonical uri for the stored document. This works well since all
annotations will have the canonical uri as the `target[source]`.

The only issue is that when a website changes its canonical uri, then we
can't surface the annotations that were made previously because they
won't have the same canonical uri.

This commit fixes that behavior in returning all canonical uris when we
detect that the passed-in uri is a canonical uri.

### DIFF
--- a/h/api/storage.py
+++ b/h/api/storage.py
@@ -222,13 +222,16 @@ def expand_uri(request, uri):
     if doc is None:
         return [uri]
 
-    # We check if the match was a "canonical" link. If so, all annotations
-    # created on that page are guaranteed to have that as their target.source
-    # field, so we don't need to expand to other URIs and risk false positives.
+    # We check if the match was a "canonical" link. If so, we return a list of
+    # all canonical uris we have stored in our database. This will surface all
+    # annotations since they are guaranteed to have the canonical uri as their
+    # target.source field.
+    # Since we return all canonical uris we have stored we also make sure that
+    # this behaviour still works when the website changes its canonical uri.
     docuris = doc.document_uris
     for docuri in docuris:
         if docuri.uri == uri and docuri.type == 'rel-canonical':
-            return [uri]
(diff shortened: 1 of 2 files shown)


nickstenning:
Seems reasonable, but it would still require us to be told about the association between old and new canonical URIs. In this case I guess you've fixed that manually...

judell:
"it would still require us to be told about the association between old and new canonical URIs. In this case I guess you've fixed that manually..."

@chdorner can you unpack that bit for me? In the WP case we currently have:

<link rel="canonical" href="https://blog.jonudell.net/2016/04/02/owning-and-sharing-your-words/" />

I presume this was formerly the http flavor. Since we had that one, then added the https one, didn't we already have both in hand? In other words I'm wondering if this change obviated the need for the manual fixes you made.

chdorner:
Both parts were necessary to fix the WP case.

Quick definition of words for the next paragraph:
* _old_ annotation means an annotation created before the site switched its canonical URL
* _new_ annotation means an annotation created after the site switched its canonical URL

The code fix here is so that we can surface both the _old_ and _new_ annotations on a website that changed its canonical URL. But this only works when we know that two documents are the same, which was the case for this WP.com because both old and new annotations shared at least one document link (`rel-shortlink`).

Me manually adding the https version of http WP.com documents made sure that we can surface all annotations on documents that did not get a new annotation after the https switch.

nickstenning:
So, yeah. This is not the right thing to do. Sorry, I didn't think this through well enough when I reviewed it. See #3249, which is caused by this change.

PR #3242, opened 2016-04-22
Sync PG annotations to ES

chdorner:
This adds a new celery worker for syncing annotations created in Postgres to the new Elasticsearch index.

This PR depends on the merge of #3238, at which point the first commit needs to be removed (I just made sure that it doesn't raise an error, it isn't really fixing all the issues).
_Update:_ #3238 is now merged, so this is ready to get merged as well.

This is now waiting for #3250 to be merged.
_Update:_ #3250 is now merged.


h/annotation_sync.py
```diff
@@ -0,0 +1,31 @@
+# -*- coding: utf-8 -*-
```
nickstenning: `annotation_sync` seems like a bit of an odd name to me. Perhaps `h.indexer` or even `h.search`?

h/annotation_sync.py
```diff
@@ -0,0 +1,31 @@
...
+
+from h.api import presenters
+from h.api.search.index import index
+from h.api.search.index import delete
+
+__all__ = ('index_annotation',)
+
+
+@celery.task
+def index_annotation(action, annotation):
```
nickstenning: It might be nicer to have two different tasks for indexing and deleting, rather than relying on `action`. It'll make stuff easier to test, too.

h/api/search/index.py
```diff
@@ -23,25 +23,19 @@ def index(es, annotation, request):
...
         index=es.index,
         doc_type=es.t.annotation,
         body=annotation_dict,
-        id=annotation_dict["id"],
+        id=annotation_dict['id'],
     )
 
 
-def delete(es, annotation):
+def delete(es, annotation_dict):
```
nickstenning: This could just take an id -- would save the cost of serializing and sending the whole thing through Rabbit.

h/api/presenters.py
```diff
@@ -54,7 +54,8 @@ def tags(self):
 
     @property
     def target(self):
-        target = {'source': self.annotation.target_uri}
+        target = {'source': self.annotation.target_uri,
+                  'scope': self.annotation.target_uri_normalized}
```
nickstenning: Ahhh, here is where this gets messy. There's no need for this `scope` rubbish to be exposed to clients, and so I'd really rather we didn't start adding stuff to the presenter layer just for the sake of the search index.

Perhaps it would be better to move the responsibility for adding the scope field into `h.api.search.index.index`?

nickstenning: To be clear, I hope that eventually we don't use the presenters at all for search indexing -- the requirements are fundamentally different. It's a nice shortcut at the moment because it preserves (mostly) the format our search index is already in, but I'd like us to avoid having to unpick two different sets of requirements from the presenters in the future. So if there's stuff that only needs to be in the search index, it shouldn't be in the presenters.

h/api/search/index.py
chdorner: Actually, this should keep the annotation, and the celery task should load the annotation based on the id. But yes, you're right.

h/annotation_sync.py
chdorner: You don't seem to like file/module/package names with multiple words :stuck_out_tongue: 
I like `h.indexer` though, renamed it to that :+1:

h/annotation_sync.py
chdorner: Fixed.

h/api/search/index.py
chdorner: Fixed.

h/api/presenters.py
chdorner: The celery tasks now get an id, which means we load the annotation from the database, so we have access to the `target_uri_normalized` property again, so all of this is not needed anymore.
Removed these 2-3 commits.

h/indexer.py
```diff
@@ -0,0 +1,38 @@
...
+
+__all__ = (
+    'add_annotation',
+    'delete_annotation',
+)
+
+
+@celery.task
+def add_annotation(id_):
+    annotation = storage.fetch_annotation(celery.request, id_)
```
nickstenning: Ah, ok. This is a good idea, but you're going to need to be a bit careful if you do this. Because the view function is called inside a transaction, that means that you're going to be calling `add_annotation.delay(id)` while the transaction is still open.

While not likely in practice, that means that in theory that message could race to rabbit, get dispatched to a worker, and then do a `SELECT` against the database before the transaction is committed. Which would be Bad (:tm:).

So, while this is absolutely fine, you'll need to change either the subscriber to trigger the job in a [response callback](http://docs.pylonsproject.org/projects/pyramid/en/latest/narr/hooks.html#using-response-callbacks). 

In fact, we should probably trigger all `AnnotationEvents` inside response callbacks (being sure not to do so if the response is an error), so perhaps it would be better to do it once in `h.api.views`. This would ensure that:

- `AnnotationEvents` can't modify annotations, as they don't have access to the session
- they only get fired once the data is successfully written to the database
- errors in `AnnotationEvent` subscribers don't cause the transaction to abort (which, given [we've already got code to ensure this in one case](https://github.com/hypothesis/h/blob/1b184e98549010a86646a2dea9ba66b8a50ae969/h/subscribers.py#L53-L63), seems like the right thing to do)

nickstenning: And yes, [response callbacks are called after the transaction is committed](http://docs.pylonsproject.org/projects/pyramid/en/latest/narr/router.html) (`pyramid_tm` is implemented as a tween).

chdorner: Will work on this in a separate PR, and marking this PR as WIP until the other one is merged.

chdorner:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;Td18UaqW&#x2F;304-postgres-offline-search-index-worker"> Postgres: offline search index worker.</a>

seanh:
I've had a look over this as well, all looks good to me

h/api/presenters.py
seanh: This issue <https://github.com/hypothesis/h/issues/3261> that I ran into elsewhere was also to do with unwanted stuff getting into the search index because we use the presenters for it. I was thinking we needed a separate presenter for the search index though, not that we didn't want to use a presenter for the search index at all?

h/indexer.py
chdorner: #3250 is now merged. I've rebased this branch and fixed a small issue due to the `AnnotationEvent` not containing the `h.api.model.Annotation` object anymore but the serialized version of an annotation. This is now ready for review again.

h/api/search/index.py
```diff
@@ -51,9 +51,9 @@ def delete(es, annotation_id):
     :param es: the Elasticsearch client object to use
     :type es: h.api.search.Client
 
-    :param annotation: the annotation whose corresponding document to delete
-        from the search index
-    :type annotation: h.api.models.Annotation
+    :param annotation_id: the annotation id whose corresponding document to
+        delete from the search index
+    :type annotation: str
```
nickstenning: s/annotation/annotation_id/

h/api/search/test/index_test.py
```diff
@@ -58,7 +58,9 @@ def presenters(self, patch):
 class TestDeleteAnnotation:
 
     def test_it_deletes_the_annotation(self, es):
-        index.delete(es, 'test_annotation_id')
+        id_ = 'test_annotation_id'
```
nickstenning: Might be worth removing this no-op change from the diff.

h/indexer.py
```diff
@@ -0,0 +1,38 @@
...
+@celery.task
+def delete_annotation(id_):
+    delete(celery.request.es, id_)
+
+
+def subscribe_annotation_event(event):
+    if not event.request.feature('postgres'):
+        return
+
+    if event.action == 'create':
```
nickstenning: This should also happen on update, no?

nickstenning:
Ok, all good except we should be reindexing on update, too. I know the postgres code for update isn't done yet, but I think we should have the indexer responding to those events already otherwise we're going to forget it.

chdorner:
Just addressed the last comments and also defined the `celery` and `indexer` queues in `h.celery`.

nickstenning:
LGTM.

PR #3243, opened 2016-04-22
[Update annotations in Pg 2/3]: Do all validation in schemas

seanh:
When creating new annotations a couple of validation rules were
implemented in h.api.storage.create_annotation() instead of in
h.api.schemas with the rest, because we didn't want h.api.schemas to
access the db:

1. Replies must have the same group as their parent.

   Actually this is transformation not validation - it just overwrites
   the group of a reply with that of its parent.

   This needed to be implemented in storage because it requires access
   to the db to look up the parent annotation.

2. The parent ID that a reply is in reply to must be an ID of an
   annotation that actually exists in the database.
   Requires db.

3. The user must have permission to create annotations in the group
   they're trying to create one in.

   This doesn't require db access, just effective_principals, but it has
   to happen _after_ 1 above because 1 might change the group so it had
   to be in storage for that reason.

There's a number of reasons why it's desirable to do _all_ validation in
schemas and none in storage:

a. Have all the validation code implemented in one place, and one call
   to validate() does all validation and transformation at once.

b. No validation or transformation going on in storage, keeps it simple.

c. The 3 bits of validation and transformation code above are actually needed
   by both storage.create_annotation() and the (to be implemented)
   PostgreSQL version of storage.update_annotation(). This means we'd
   have to either.

   i.   Duplicate the code and tests in create_annotation() and
        update_annotation(). But then we have duplication.
   ii.  Add a private helper function for create_annotation() and
        update_annotation() to call and unit test that function. But we
        don't like unit testing private functions.
  iii.  Make the private helper function public. But it shouldn't be
        public as it's not intended to be part of storage's public API.
   iv.  Make a new module to contain this validation code for storage to
        call and unit test that module.

   We _could_ do iv. but schemas.py already contains a place to put
   validation code that's shared between create and update
   (both CreateAnnotationSchema and UpdateAnnotationSchema inherit from
   AnnotationSchema) and putting all the validation code into schemas
   has the advantages a and b as well.

So move this validation code into schemas, which is going to make
implementing the PostgreSQL version of storage.update_annotation()
with no duplicate code or tests simpler.

Schemas is still a pure module: it takes dicts as inputs, gives dicts as
outputs, and doesn't create, update or delete anything in the db.

But it does import and read the db (via storage).

If we really don't want schemas to import storage at all then I'd
suggest [using jsonschema format functions](https://github.com/hypothesis/h/pull/3225) which would allow
schemas to declare that it depends on these validation functions, while
the validation functions themselves (which need access to storage) are
actually implemented in a different module, and then h.api.views injects
the validation functions from the separate module into schemas.
All validation declared in one place in schemas and carried out by one
call to validate(), no validation in storage, and no storage in schemas.

But for now this way is simple and direct.

### DIFF
--- a/h/api/schemas.py
+++ b/h/api/schemas.py
@@ -8,6 +8,7 @@ from pyramid import i18n
 from pyramid import security
 
 from h.api import parse_document_claims
+from h.api import storage
 
 _ = i18n.TranslationStringFactory(__package__)
 
@@ -199,11 +200,30 @@ class AnnotationSchema(JSONSchema):
         new_appstruct['groupid'] = appstruct.pop('group', u'__world__')
         new_appstruct['references'] = appstruct.pop('references', [])
 
-        # Replies always get the same groupid as their parent. The parent's
-        # groupid is added to the reply annotation later by the storage code.
-        # Here we just delete any group sent by the client from replies.
-        if new_appstruct['references'] and 'groupid' in new_appstruct:
-            del new_appstruct['groupid']
+        # Replies must have the same group as their parent.
+        if new_appstruct['references']:
+            top_level_annotation_id = new_appstruct['references'][0]
+            top_level_annotation = storage.fetch_annotation(
+                self.request,
+                top_level_annotation_id,
+                _postgres=True)
+            if top_level_annotation:
+                new_appstruct['groupid'] = top_level_annotation.groupid
+            else:
+                raise ValidationError(
+                    'references.0: ' +
+                    _('Annotation {annotation_id} does not exist').format(
+                        annotation_id=top_level_annotation_id)
+                )
+
+        # The user must have permission to create an annotation in the group
+        # they've asked to create one in.
+        if new_appstruct['groupid'] != '__world__':
+            group_principal = 'group:{}'.format(new_appstruct['groupid'])
+            if group_principal not in self.request.effective_principals:
+                raise ValidationError(
+                    'group: ' + _('You may not create annotations in groups '
+                                  'you are not a member of!'))
 
         new_appstruct['extra'] = appstruct
 
--- a/h/api/storage.py
+++ b/h/api/storage.py
@@ -11,7 +11,6 @@ from functools import partial
 
 from pyramid import i18n
 
-from h.api import schemas
 from h.api import transform
 from h.api import models
 from h.api.events import AnnotationBeforeSaveEvent
@@ -81,31 +80,6 @@ def create_annotation(request, data):
     document_uri_dicts = data['document']['document_uri_dicts']
     document_meta_dicts = data['document']['document_meta_dicts']
     del data['document']
-
-    # Replies must have the same group as their parent.
-    if data['references']:
-        top_level_annotation_id = data['references'][0]
-        top_level_annotation = fetch_annotation(request,
-                                                top_level_annotation_id,
-                                                _postgres=True)
-        if top_level_annotation:
-            data['groupid'] = top_level_annotation.groupid
-        else:
-            raise schemas.ValidationError(
-                'references.0: ' +
-                _('Annotation {annotation_id} does not exist').format(
-                    annotation_id=top_level_annotation_id)
-            )
-
-    # The user must have permission to create an annotation in the group
-    # they've asked to create one in.
-    if data['groupid'] != '__world__':
-        group_principal = 'group:{}'.format(data['groupid'])
-        if group_principal not in request.effective_principals:
-            raise schemas.ValidationError('group: ' +
-                                          _('You may not create annotations '
-                                            'in groups you are not a member '
-                                            'of!'))
 
     annotation = models.Annotation(**data)
     request.db.add(annotation)
--- a/h/api/storage.py
+++ b/h/api/storage.py
@@ -14,6 +14,7 @@ from pyramid import i18n
 from h.api import transform
 from h.api import models
 from h.api.events import AnnotationBeforeSaveEvent
+from h.api.db import types
 
 
 _ = i18n.TranslationStringFactory(__package__)
@@ -51,7 +52,10 @@ def fetch_annotation(request, id, _postgres=None):
         _postgres = _postgres_enabled(request)
 
     if _postgres:
-        return request.db.query(models.Annotation).get(id)
+        try:
+            return request.db.query(models.Annotation).get(id)
+        except types.InvalidUUID:
+            return None
 
     return models.elastic.Annotation.fetch(id)
 
--- a/h/api/test/schemas_test.py
+++ b/h/api/test/schemas_test.py
@@ -39,6 +39,7 @@ class TestJSONSchema(object):
         assert message.startswith("123 is not of type 'string'")
 
 
+@pytest.mark.usefixtures('authn_policy', 'storage')
 class TestAnnotationSchema(object):
 
     def test_it_does_not_raise_for_minimal_valid_data(self):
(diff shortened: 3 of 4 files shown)


seanh:
By th

nickstenning:
So, I see your motivation here, but I really think we should endeavour to keep `h.api.schemas` about _validating the data structure sent by the client_. That is, getting the data sent by the client into a form where we can unambiguously interpret it, _not_ about doing all pre-commit validation.

I would wager that we're _never_ going to do all validation in the `schemas` module, or in any other one place, because data validation means different things at different layers of the application, and there are different kinds of validation to be done depending on where data is coming from. There are a couple of design principles that I think would encourage us not to extend the validation done in this package as far as checking for domain model consistency:

1. ease of testing -- it's a lot easier to test functions that map data structures to other data structure than it is to test functions that read data from the database
2. single responsibility principle -- SRP ("a class should only have one reason to change") would advise that schemas should not know about *both* the domain model *and* the incoming client wire format

In both cases, these apply more to `schemas` than to `storage`, because the former will (I hope) stick around, whereas the latter is, at least in its current incarnation, mostly a big pile of compromises to allow us to do this data migration safely.

I'd argue that, right now, these checks *do* live in `storage`, and they should be tested as part of the test suites for `create_annotation` and `update_annotation`.

As to how `schemas` should behave, I think that being as accepting as possible while simplifying the necessary code upstream is the way to go. Here's one way of thinking about it:

1. There is only one scenario in which the user gets to determine the group: they are creating a new annotation. On update, the group is predetermined and cannot be changed. On reply creation, the same is true.
2. So, in `schemas`, we simply remove the `group` from the incoming data unless it's a create and the data is not a reply.
3. In `storage`, if `group` is not in the data and we're dealing either with a reply creation, we need to fetch the right group from the thread root.
4. In `storage`, validate that the user is allowed to write to the selected group, and raise if not.

If, in future, we wanted to go as far as validating that the user didn't send a `group` field on update, or when creating a reply, that would be fine with me. But at the moment that's going to complicate things, and would necessitate changes to our client that I'd rather we didn't get into. So simply dropping the fields (just as we do for [a bunch of others](https://github.com/hypothesis/h/blob/455bb7ef4859e06004686f9d56c7b784f76c86c0/h/api/schemas.py#L15)) is absolutely fine, I think.

Doing it this way may add a little bit of duplication to `storage`, but that's absolutely fine. The module is likely to change significantly when we remove the Elasticsearch storage functionality, and we can reassess how we want this to happen then, perhaps moving some of the above validation into SQLAlchemy validators.

It will also allow us to be a little bit more intelligent about the errors we return to users. Rather than returning a 400 from a generic `ValidationError` when someone tries to write to a group they have no permissions on, we can add a new exception view to `h.api.views` that catches `h.api.storage.GroupPermissionError` or whatever you want to call it.

seanh:
Alright, I don't want to get too distracted at this point by the wider discussion about whether _all_ validation (that we have now and that we might add in the future) can or should all be implemented in `schemas`. I don't think we need to make a decision on that now. But all th

The problem I have is that there are two validation rules currently (on master) implemented in `h.api.storage.create_annotation()` and they now need to be shared, one way or another, between `storage.create_annotation()` and `update_annotation()`. It sounds like you really don't want `h.api.schemas` to be importing `storage` at all (even just to read the data). In that case there are two solutions that come to mind:

This does have the advantage that, in a sense, all validation that we currently do is being done in `h.api.schemas`. Some of the rules are actually implemented and unit tested in `storage` but they're declared in `schemas` as part of the annotation schema and they get called when `views` calls the `validate()` method from schemas. I do think having them in "one place" in this way is a +1, but that doesn't mean we have to draw a line right now that says that any more validation that we add in the future has to be done in that one place as well. The main reason for it is that it's a nice way to solve the immediate problem of how to share these validation rules between create and update without duplicating any code or tests.

seanh:
If we want schemas to have a single responsibility and not know about the domain model, doesn't that mean that all the "transforming the data from the client into a format suitable for storing in the database" code ought to go in `storage`, not in `schemas`? Schemas does implicitly know a lot about the domain model because as well as validating it also transforms the data into the format necessary for storing in the domain model. The storage code assumes the data has been transformed into a valid form by schemas and just passes it straight into the model without further transformation (apart from the three particular validation/transformation rules that I'm refactoring in this pull request). So if schemas didn't transform the data into the same format as it's stored in the model, storage would then crash.

Anyway, regarding the three validation rules that this PR is about:

I could just go ahead and copy-paste the validation code from `storage.create_annotation()` into `update_annotation()` but note that it's actually three separate validation rules (parent ID of reply must exist, reply must have same group as parent, and user must have permissions for group) plus the tests for those three, and this list of shared validation rules seems likely to grow.

How about this? Add functions named something like `reply_must_have_same_group_as_parent()`, `parent_id_of_reply_must_exist()` and `user_must_have_permission_for_group()` to the public API of `storage` and unit test them in the storage tests. These could either raise or return `False` if validation fails. Then reference these new validation functions by name in `schemas.AnnotationSchema` as jsonschema format functions, and have `views` inject the actual implementations of them into the schema.

So `schemas` wouldn't be importing storage at all, and these validation functions would be implemented and unit tested in `storage` not in `schemas`. At the same time it keeps all the validation in one place in that all the validation rules are declared in `schemas` as part of the annotation schema and they all get called when `views` calls the `validate()` method from schemas, and no more validation is done in `storage.create_annotation()` or `update_annotation()`. I do think having them in "one place" in this way is a +1, but that doesn't mean we have to draw a line right now that says that any more validation that we add in the future has to be done in that one place as well. The main reason for it is that it's a nice way to solve the immediate problem of how to share these validation rules between create and update without duplicating any code or tests.

chdorner:
_tldr is at the end_

It looks like you two seem to have different views on how validation should happen. Let me try and respond to some of the points above as an impartial third party (I am Swiss after all).

Also bear in mind that I don't fully understand what should and should not go into a `schemas` module, most of what I believe to be true/false is based on reading the discussions that happened around the new code for our move to Postgres.

To set the current situation based on the current code in `schemas` and `storage` (and to make sure I understand the problem):

1. We do validate and transform the clients input in `schemas` with `CreateAnnotationSchema` and `UpdateAnnotationSchema` (the update does the same as create, except that it additionally checks that the group did not change).
1. We do mostly transform but as a side effect also validate the data that has been transformed by the `schemas` level in `storage`
  1. When the annotation is a reply we set the same group as its toplevel parent, and as a side-effect validate that the toplevel parent still exists ([code](https://github.com/hypothesis/h/blob/316d1cf6a6d2f113a1842694ad3294821d5d0ea1/h/api/storage.py#L85-L98)).
  1. We validate that the user has permission to create the annotation in the specified group ([code](https://github.com/hypothesis/h/blob/316d1cf6a6d2f113a1842694ad3294821d5d0ea1/h/api/storage.py#L100-L108)).

__To address Sean's reasons he listed in the initial PR description as the motivation to move all validation into one place (`schemas`):__

> a. Have all the validation code implemented in one place, and one call to validate() does all validation and transformation at once.

In general this sounds very sensible to me, although we have to consider the implications this has (which I will address further down).

> b. No validation or transformation going on in storage, keeps it simple.

Same as with the previous point, this sounds very sensible to me. This might have less implications, since we don't _have_ to that code into `schemas` but we could move it into a different module and it can be called from `storage`.

> c. The 3 bits of validation and transformation code above are actually needed by both storage.create_annotation() and the (to be implemented) PostgreSQL version of storage.update_annotation(). This means we'd have to either.

This is only partially true as Nick described. The transformation of setting the group of the toplevel parent arguably does not have to run for `storage.update_annotation()` as we don't allow changing the group of an annotation. We still need to make sure that the toplevel parent annotation exists, and that the user has permission to write to that group.

__To address Sean's possible solutions from the initial PR description:__

> i. Duplicate the code and tests in create_annotation() and update_annotation(). But then we have duplication.

This is clearly not the nicest solution, as Sean already mentioned.

> ii. Add a private helper function for create_annotation() and update_annotation() to call and unit test that function. But we don't like unit testing private functions.

I personally think that duplicating test code is not as bad as duplicating implementation code. Although I always prefer not to duplicate any code.

The test duplication could be avoided by parametrizing the tests to run both `create_annotation()` and `update_annotation()` with a given set of initial state, input, and expected output. This might not be the nicest Python code ever written (which is a potential downside), but I'm pretty sure it could be done.

> iii. Make the private helper function public. But it shouldn't be public as it's not intended to be part of storage's public API.

I agree, although see my response to the next solution (_iv_).

> iv. Make a new module to contain this validation code for storage to call and unit test that module.

I think this solution is my second favourite, depending on how bad we consider test duplication / parametrizing of the tests as described in my response to solution _ii_ to be.


__To adress Nick's points and Sean's answer:__

> I would wager that we're never going to do all validation in the `schemas` module, or in any other one place, because data validation means different things at different layers of the application, and there are different kinds of validation to be done depending on where data is coming from.

I tend to agree with this in this particular situation, but I'm not sure if we can make this a rule without considering the situation. Yes, we do validate and transform data in two different parts of the code, but they are arguably at different levels.

To me `schemas` makes sure that what the client sent us is structurally sound (it has all the required fields, the fields have the correct types, things like that), and then `storage` (or a possible new module) makes sure the content of those fields is correct by taking the current master state (our database) and the ACL system into account.

> Doing it this way may add a little bit of duplication to `storage`, but that's absolutely fine.  The module is likely to change significantly when we remove the Elasticsearch storage functionality, and we can reassess how we want this to happen then, perhaps moving some of the above validation into SQLAlchemy validators.

This is another point I think we shouldn't underestimate. The `storage` module is quite a mess at the moment. Which is completely fine and to be expected with having to support multiple storage engines at the same time. But its current state might make it harder to reason about how far we're willing to make compromises (validating and transforming at different levels vs. one place).

> Rather than returning a 400 from a generic ValidationError when someone tries to write to a group they have no permissions on, we can add a new exception view to h.api.views that catches h.api.storage.GroupPermissionError or whatever you want to call it.

This sounds like a big plus to me. Considering what I wrote above about how I see the responsibilities of the `schemas` and `storage` modules:

1. Structural validation failed (`schemas`): a 400 error, the client is at fault for sending us weird data
2. Content validation failed (`storage`): could be any error code, 401/403 for missing permissions, missing parent annotation maybe a 409

> If we want schemas to have a single responsibility and not know about the domain model, doesn't that mean that all the "transforming the data from the client into a format suitable for storing in the database" code ought to go in storage, not in schemas?

I don't think that this is what it means. We made the decision that schemas emits a format that is easily usable by storage, but this doesn't have to mean that this transformation has to go into `storage`. That being said, it might not be ideal to assume that whatever `schemas` produces will be blindly accepted by `storage`, but I think that is a valid tradeoff at the moment but should be revisited later.

__My personal conclusion:__

For now, I would move these 2-3 validations/transformations into private functions in the `storage` module. I would then try and make the tests as nice as possible (parametrizing them, if not possible then just duplicating them).

But, once we switched over to Postgres and removed ElasticSearch support for our main data store, I would revisit the validation/transformation and judge it without all the code to support two data stores.

I also think that the `jsonschema` approach sounds interesting, but as with the desire to combine different levels of validations, I would prefer to decide about adding another dependency once we removed ElasticSearch support.

nickstenning:
Perhaps I wasn't clear enough, but you keep mentioning duplication as a thing you're trying to avoid. One of the points I was trying to make was that there *isn't* any duplication.

You characterised this as addressing three validation rules:

1. Replies must have the same group as their parent.

2. The parent ID that a reply is in reply to must be an ID of an
annotation that actually exists in the database.

3. The user must have permission to create annotations in the group
they're trying to create one in.

*None* of these need to happen on update. If you simply drop the relevant fields (`group` and `references`) from the data passed by the client on update, nothing about the current implementation will change, and there will be no need to do these checks in `create_annotation`.

seanh:
Alright, honestly I still prefer the jsonschema formats approach, but in the interest of getting this merged I've started rewriting <https://github.com/hypothesis/h/pull/3245> and its tests to work on top of master instead of on top of this branch, and to work in the way that Nick suggested.

Also fixing the semantics of the update `PUT` requests to be more like (a broken version of) `PATCH` than `PUT`, to be compatible with the legacy behaviour.

PR #3244, opened 2016-04-22
Add conversation thread construction module

robertknight:
This is the first part of the new threading UI implementation
whose design is outlined at https://github.com/hypothesis/h/pull/3176/

build-thread implements the logic for constructing an immutable view
model for a list of conversations (a 'Thread') from a list of
annotations and replies plus an object describing the current UI state.

This view model is then rendered by an `<annotation-thread>` component (to follow in a separate PR)

Once the other parts of the new implementation are ready, this will
replace 'threading', 'thread-filter' and the JWZ script.


robertknight:
There is some discussion happening [here](https://github.com/hypothesis/h/pull/3176#discussion_r60830737) about how to handle annotations that don't exist but whose previous existence is implied by the reference field of a reply.

Keeping the behavior the same as on master will require that `buildThread()` create Thread objects with either no annotation or a placeholder annotation in that case.

nickstenning:
Yep, I'd just like to reiterate that while I'm not convinced that a fully-threaded conversation model is best for us long term, that _is_ the model we have right now. And _while_ we have that model, we should endeavour to make sure we reflect accurately what happened when, which means _not_ silently dropping messages from the thread, as in the following example (copied from the linked discussion).

Consider the following conversation:

```
annotation
|
` reply 1
  |
   ` reply 2
```

What happens if reply 1 is deleted? The JWZ algorithm allows us to infer the presence of the missing message and display that information to the user, and so we currently render something like: 

```
annotation
|
` [message unavailable]
  |
   ` reply 2
```

While we have explicit threaded conversations I think it's important that we don't change to something that essentially lies about what happened. For example, I think producing the rendered output

```
annotation
|
` reply 2
```

would be problematic.

robertknight:
> And while we have that model, we should endeavour to make sure we reflect accurately
> what happened when, which means not silently dropping messages from the thread,
> as in the following example (copied from the linked discussion).

This branch now implements the same behavior as master, where placeholders are created for missing annotations. See [these tests](https://github.com/hypothesis/h/pull/3244/files#diff-391821aa1d527c759ec0da60569dbe6dR118).

robertknight:

finitud:
👍

nickstenning:
LGTM.

PR #3245, opened 2016-04-22
[Update annotations in Pg 3/3]: Storage and view code for updating annotations in Postgres

seanh:
This is branched off <https://github.com/hypothesis/h/pull/3243> so the diff is wrong. You can't really use this without the (upcoming) view code for making annotations in Pg. But the tests pass, and it still works with postgres off.

### DIFF
--- a/h/api/schemas.py
+++ b/h/api/schemas.py
@@ -8,6 +8,7 @@ from pyramid import i18n
 from pyramid import security
 
 from h.api import parse_document_claims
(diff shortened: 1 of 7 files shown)


seanh:
This is done but I'm leaving it marked as wip 'cos it needs to be tested manually and given a look over.

seanh:
Closing in favor of https://github.com/hypothesis/h/pull/3256

PR #3246, opened 2016-04-23
Move email generation responsibilities out of `h.notification`

nickstenning:
The preface: `h.notification` is some of the hairiest code in our codebase. We've all known it needs fixing for a while. I had a 10-hour flight, and [Sean had already done some of the hard work](https://github.com/hypothesis/h/pull/3238).

This is one of a couple of pull requests I have lined up to improve the state of this package. In rough order, here's how I think it's going to go down:

- move email rendering/sending responsibilities out of `h.notification` and into `h.emails`
- make `h.notification` less of a nightmare in general. A big part of this will be improving the test suite.
- replace the `Subscriptions` model with one that does the opposite (records opt-outs)
- move other email rendering responsibilities into `h.emails`
- move email templates into MailChimp?

This PR only does the first of this list, and here's an extract from the relevant commit message:

---

Rather than the `generate_notifications` function yielding rendered email data, we now have a business-logic-only function which checks to see if a given annotation action merits a notification, and if so yields a `Notification` object, which contains the data necessary for rendering the notification, namely:

- the reply annotation,
- the user who made the reply annotation,
- the annotation being replied to,
- the user who made the annotation being replied to,
- the document object for the annotated page

The responsibility for actually generating the email has moved into a new `h.emails` package, where `h.emails.reply_notification.generate` now accepts the notification object as a parameter and returns the data needed to construct an email.

As a result of the simplification of `h.notification.reply_template` in this commit, the following modules have also been removed:

- `h.notification.gateway`
- `h.notification.notifier`
- `h.notification.types`

---

It's worth noting that the tests for the new `h.email.reply_notification` module are a little bit involved, because I wanted to test against both the old and new `Annotation` models. While relying on `pytest` fixtures as heavily as these tests do is probably not a great idea in the long run, it does make this much easier to write.

The basic concept is that there's a fixture called `storage_driver` which takes one of two values: `'postgres'` or `'elastic'`. This bubbles up through other fixtures so that in the end, each of the tests in `test_reply_notification.py` are run twice, once with the old annotation model, once with the new.


finitud:
LGTM

PR #3247, opened 2016-04-25
Change display of datetimes on annotations

seanh:


seanh:
@robertknight This is WIP but you can see where I'm going with it

h/static/scripts/time.js
```diff
@@ -6,43 +6,79 @@ var day = hour * 24;
...
-  [2 * minute, 'a minute ago',   minute],
-  [hour,       '{} minutes ago', minute],
-  [2 * hour,   'an hour ago',    hour],
-  [day,        '{} hours ago',   hour],
-  [2 * day,    'a day ago',      day],
-  [month,      '{} days ago',    day],
-  [year,       '{} months ago',  month],
-  [2 * year,   'one year ago',   year],
-  [Infinity,   '{} years ago',   year]
+  [lessThanThirtySecondsAgo,    function () {return 'Just now';},      1],
```
robertknight: The last item in each tuple is now redundant. I think this would be clearer to read if written as an array of formatters:

```js
var formatters = [
  {test: testFunc, format: formatFunc},
  {test: testFunc2, format: formatFunc2},
  ...
];
```

robertknight:
@seanh - Had a quick look over this. The general approach looks fine to me and as you say, makes it easier to change in future.

h/static/scripts/time.js
```diff
@@ -6,43 +6,79 @@ var day = hour * 24;
...
+function nMin(date, now) {
+  return '{} min'.replace('{}', Math.floor(delta(date, now) / minute));
+}
+
+function nHr(date, now) {
+  return '{} hr'.replace('{}', Math.floor(delta(date, now) / hour));
+}
+
+function dayAndMonth(date) {
+  return date.toLocaleDateString(undefined, {day: '2-digit', month: 'short'});
```
robertknight: So we don't forget - it is critical to cache the date formatter using `Intl.DateTimeFormat` (see date-util.js) as it turns out that calling `toLocaleDateString()` on both Chrome and Firefox appears to create a new formatter for each call which is very expensive.

nickstenning:
What needs doing here?

seanh:
Tests, caching, I'm working on it..

h/static/scripts/time.js
seanh: [Done](https://github.com/hypothesis/h/pull/3247/commits/24786940cfb853d2843862a01046f03291c78a7f)

h/static/scripts/time.js
```diff
@@ -6,43 +6,115 @@ var day = hour * 24;
...
+function nMin(date, now) {
+  return '{} min'.replace('{}', Math.floor(delta(date, now) / minute));
+}
+
+function nHr(date, now) {
+  return '{} hr'.replace('{}', Math.floor(delta(date, now) / hour));
+}
+
+// Cached DateTimeFormat instances,
+// because instantiating a DateTimeFormat is expensive.
```
robertknight: Clarification here - instantiating a DateTimeFormat was very expensive in the browsers I tested and `Date.toLocaleDateTime()` which is the obvious thing to use in this context appears to do the same thing behind the scenes.

Something I have seen occasionally in the Angular code which I quite like is a link to a benchmark which can easily be run. JSPerf used to be popular for this but the site is currently being rebuilt. An alternative would be a link to a simple Gist which we can easily run in future.

h/static/scripts/time.js
```diff
@@ -6,43 +6,115 @@ var day = hour * 24;
 var month = day * 30;
 var year = day * 365;
 
+function lessThanThirtySecondsAgo(date, now) {
+  return ((now - date) < 30000);
```
robertknight: For consistency with the functions below I suggest writing these two constants as `N * 1000`

h/static/scripts/test/time-test.js
```diff
@@ -50,6 +51,34 @@ describe('time', function () {
...
   describe('.toFuzzyString', function () {
+
+    var originalIntl;
+
+    before(function () {
+
+      // Intl isn't implemented in PhantomJS so we have to mock it in these
+      // tests. The mock version returns the results we'd expect from the real
+      // one for the test values that we use.
+      originalIntl = window.Intl;
```
robertknight: Rather than stubbing `window.Intl` here you could make it an optional third argument to the `format()` function which defaults to `window.Intl` if undefined. eg:

```js
function format(date, options, Intl) {
  if (typeof Intl === 'undefined') {
    Intl = window.Intl;
  }
  if (Intl && Intl.DateTimeFormat) {
   ...
  } else {
   ...
  }
}
```

Then you can test the two cases via `format(date, options, FakeIntl)` and `format(date, options, null)`

seanh: The tests don't actually call `format()` though, they call `toFuzzyString()` which (sometimes) gets `dayAndMonth()` or `dayAndMonthAndYear()` (via `BREAKPOINTS` and `getBreakpoint()`) and those two both call `format()`. We'd have to add the optional `Intl` argument to `toFuzzyString()`, `dayAndMonth()`, `dayAndMonthAndYear()` and `format()` and pass it down. Or did you mean that we should add some direct unit tests for `format()`?

h/static/scripts/time.js
seanh: Agree on renaming it and changing them to objects.

But are the last items redundant? They're used by `nextFuzzyUpdate()` to decide when to re-format the timestamp.

h/static/scripts/time.js
seanh: Done

h/static/scripts/time.js
seanh: Done

h/static/scripts/time.js
seanh: I've changed these to objects

h/static/scripts/test/time-test.js
robertknight: @seanh - In general, there are a few approaches you can take for passing dependencies down a chain of functions that I know of:

1. Just pass the dependency down explicitly as an argument to each function
2. Do the same as (1) but bundle the dependencies up into a context object (ie. a record/object which maps keys to dependencies). Pyramid's `request` object or Angular's `$scope` are effectively instances of this.
3. Instead of passing dependency 'A' to 'B' which calls 'C' that in turn uses 'A', you instead create a version of 'C' that has 'A' already resolved and pass that in to 'B'. ie. You would make the `format()` function itself an argument to `toFuzzyString()`. This is what you're effectively doing with Angular's dependency injection.

robertknight: In this case, the chain of functions isn't very long so either (1) or (3) would work. You could implement (3) by making `format()` a separately unit-testable function and an argument to `toFuzzyString()`

seanh: I don't think either 2 or 3 adds much in this case so would go for 1.

But what result are we hoping for from the tests here? If we add an optional `Intl` argument there'll be four possible combinations of the two `Intl` values:

1. The test pass in a mock `Intl` and `window.Intl` is defined (e.g. we're running the tests in Chrome) - it should use th

seanh: I don't think either 2 or 3 adds much in this case so would go for 1.

But what result are we hoping for from the tests here? If we add an optional `Intl` argument there'll be four possible combinations of the two `Intl` values:

1. The test pass in a mock `Intl` and `window.Intl` is defined (e.g. we're running the tests in Chrome) - it should use the mock

2. The test pass in a mock `Intl` and `window.Intl` is undefined (e.g. we're running the tests in PhantomJS) - it should use the mock

3. The test do not pass in a mock `Intl` and `window.Intl` is defined (e.g. we're running the tests in Chrome) - it will use the real `window.Intl`

4. The test do not pass in a mock `Intl` and `window.Intl` is undefined (e.g. we're running the tests in PhantomJS) - it will fall back on `date.toDateString()`.

3 and 4 are a difference between PhantomJS and Chrome. If I have a "it falls back on date.toDateString() when Intl is undefined" test, that test will pass in PhantomJS but fail in Chrome. Unless I replace the real `window.Intl` with `undefined`, in which case the test will pass in both PhantomJS and Chrome.

seanh: Oh wait, I just saw that you want the tests to actually pass in `null` ... I guess that works!

seanh: [Done](https://github.com/hypothesis/h/pull/3247/commits/6f060334ba91db64864bdb618ecda045a50a81b1)

seanh:
Gonna test this manually now...

seanh:
Looks like it works:

screenshot from 2016-05-09 17-32-43

screenshot from 2016-05-09 17-33-19

screenshot from 2016-05-09 17-33-34

screenshot from 2016-05-09 17-34-37

screenshot from 2016-05-09 17-41-59

screenshot from 2016-05-09 17-42-29

screenshot from 2016-05-09 17-47-01

screenshot from 2016-05-09 17-46-11

seanh:
Argh, just noticed that it needs to be "3hr**s**" etc when > 1

seanh:
Ok, fixed

seanh:
Note that the "n secs", "n min(s)" and "n hr(s)" strings are not internationalized (this is according to the Trello card I believe).

The  "DD Mon" and "DD Mon YYYY" strings should be internationalized although we don't have a test for this.

When `Intl.DateTimeFormat()` is not present then the "DD Mon YYYY" string is not internationalized (it uses `toDateString()` not `toLocaleDateString()`).

h/static/scripts/test/time-test.js
```diff
@@ -50,18 +51,35 @@ describe('time', function () {
   });
 
   describe('.toFuzzyString', function () {
+
+    function mockIntl() {
```
robertknight: Not going to stop me merging but we generally refer to faked dependencies as fakes rather than mocks in the front-end code. Strictly speaking mocks have pre-programmed usage expectations and fakes do not, but the main reason is just consistency.

robertknight:
LGTM. Looks much nicer than the more verbose formatting we had previously IMO.

PR #3248, opened 2016-04-25
Sheetaluk/297 show annotation adder on text selected before launching the tool

sheetaluk:


h/static/scripts/annotator/guest.coffee
```diff
@@ -375,6 +381,41 @@ module.exports = class Guest extends Annotator
       .removeClass('h-icon-annotate')
       .addClass('h-icon-note');
 
+  getAdderPosition: (selection) ->
```
robertknight: I would definitely suggest pulling this function out into its own module for easier testing, since this is the kind of task where a change can easily fix one case and break another. In doing so, any temporary elements that are created to get the selection location should be removed before the function returns.

robertknight:
After some collaborative experimentation, this has been superseded by #3257

PR #3250, opened 2016-04-26
Publish events in response callback

chdorner:
We now publish pyramid `AnnotationEvent` after the request's transaction has been successfully committed. This is necessary to prevent a race condition with the background indexer (#3242) where the celery task could be running before the transaction that creates the Postgres annotation is committed.
This is desirable on a more general level anyway, since many subscribers are kicking off some kind of task (sending an email, notifying websocket clients, etc.) which we don't want to run when the transaction could fail. But this PR only includes the implementation for moving the publishing of the `h.api.events.AnnotationEvent` into a pyramid [response callback](http://docs.pylonsproject.org/projects/pyramid/en/latest/narr/hooks.html#using-response-callbacks) ([here](http://docs.pylonsproject.org/projects/pyramid/en/latest/narr/router.html) a diagram that illustrates when different components are being called during a request).

It is now possible (and encouraged) to instead of publishing events immediately in the views to enqueue them with `request.after_commit_events.append()`. We have one response callback that publishes all events to the registry in one isolated transaction.
Currently, this means that we can't put mapped SQLAlchemy objects in the events, as those will be detached from the session when accessed in a different transaction. For the moment we just load the annotation from the database again (where necessary), this adds a few Postgres roundtrips which is not ideal, but good enough for now :tm:. We should find more optimized ways of dealing with reloads - mostly moving functionality of subscribers into celery tasks, and / or caching the newly mapped annotations in the event so other subscribers can re-use them.

I have manually tested 1) reply notifications, 2) synchronous ES indexing (which be removed soon anyway), and 3) real-time websockets creating/deleting annotations, and verified that all these still work.

### DIFF
--- a/h/api/events.py
+++ b/h/api/events.py
@@ -4,10 +4,15 @@
 class AnnotationEvent(object):
     """An event representing an action on an annotation."""
 
-    def __init__(self, request, annotation, action):
+    def __init__(self, request, annotation_dict, action):
         self.request = request
-        self.annotation = annotation
+        self.annotation_dict = annotation_dict
         self.action = action
+
+    @property
+    def annotation_id(self):
+        if self.annotation_dict:
+            return self.annotation_dict.get('id')
 
 
 class AnnotationBeforeSaveEvent(object):
--- a/h/api/test/subscribers_test.py
+++ b/h/api/test/subscribers_test.py
@@ -6,7 +6,6 @@ import pytest
 from pyramid import testing
 
 from h.api import events
-from h.api import models
 from h.api import subscribers
 
 
@@ -29,14 +28,23 @@ class TestIndexAnnotationEvent:
 
         assert not delete.called
 
-    def test_it_calls_index_when_action_is_create(self, index, delete):
+    def test_it_fetches_the_annotation_when_action_is_create(self, fetch_annotation):
+        event = self.event('create')
+        event.request.feature.return_value = True
+
+        subscribers.index_annotation_event(event)
+
+        fetch_annotation.assert_called_once_with(event.request,
+                                                 event.annotation_id)
+
+    def test_it_calls_index_when_action_is_create(self, fetch_annotation, index, delete):
         event = self.event('create')
         event.request.feature.return_value = True
 
         subscribers.index_annotation_event(event)
 
         index.assert_called_once_with(
-            event.request.es, event.annotation, event.request)
+            event.request.es, fetch_annotation.return_value, event.request)
         assert not delete.called
 
     def test_it_calls_delete_when_action_is_delete(self, delete, index):
@@ -45,16 +53,20 @@ class TestIndexAnnotationEvent:
 
         subscribers.index_annotation_event(event)
 
-        delete.assert_called_once_with(event.request.es, event.annotation)
+        delete.assert_called_once_with(event.request.es, event.annotation_id)
         assert not index.called
 
     def event(self, action):
         return mock.Mock(
             spec=events.AnnotationEvent(testing.DummyRequest(),
-                                        mock.Mock(spec=models.Annotation()),
+                                        {'id': 'test_annotation_id'},
                                         action),
             action=action,
         )
+
+    @pytest.fixture
+    def fetch_annotation(self, patch):
+        return patch('h.api.storage.fetch_annotation')
 
     @pytest.fixture
     def index(self, patch):
--- a/h/api/views.py
+++ b/h/api/views.py
@@ -254,7 +254,7 @@ def _present_searchdict(request, mapping):
 def _publish_annotation_event(request, annotation, action):
     """Publish an event to the annotations queue for this annotation action"""
     event = AnnotationEvent(request, annotation, action)
-    request.registry.notify(event)
+    request.notify_after_commit(event)
 
 
 def includeme(config):
--- a/h/api/views.py
+++ b/h/api/views.py
@@ -177,11 +177,15 @@ def create(request):
                                                          legacy_appstruct)
 
     if request.feature('postgres'):
-        _publish_annotation_event(request, annotation, 'create')
-        return AnnotationJSONPresenter(request, annotation).asdict()
+        annotation_dict = AnnotationJSONPresenter(request, annotation).asdict()
+        _publish_annotation_event(request, annotation, 'create',
+                                  annotation_dict=annotation_dict)
+        return annotation_dict
 
-    _publish_annotation_event(request, legacy_annotation, 'create')
-    return AnnotationJSONPresenter(request, legacy_annotation).asdict()
+    annotation_dict = AnnotationJSONPresenter(request, legacy_annotation).asdict()
+    _publish_annotation_event(request, legacy_annotation, 'create',
+                              annotation_dict=annotation_dict)
+    return annotation_dict
 
 
 @api_config(route_name='api.annotation', request_method='GET', permission='read')
@@ -209,10 +213,10 @@ def update(annotation, request):
     appstruct = schema.validate(_json_payload(request))
     annotation = storage.update_annotation(request, annotation.id, appstruct)
 
-    _publish_annotation_event(request, annotation, 'update')
-
-    presenter = AnnotationJSONPresenter(request, annotation)
-    return presenter.asdict()
+    annotation_dict = AnnotationJSONPresenter(request, annotation).asdict()
+    _publish_annotation_event(request, annotation, 'update',
+                              annotation_dict=annotation_dict)
+    return annotation_dict
 
 
 @api_config(route_name='api.annotation', request_method='DELETE', permission='delete')
@@ -251,9 +255,12 @@ def _present_searchdict(request, mapping):
     return AnnotationJSONPresenter(request, ann).asdict()
 
 
-def _publish_annotation_event(request, annotation, action):
+def _publish_annotation_event(request, annotation, action, annotation_dict=None):
     """Publish an event to the annotations queue for this annotation action"""
-    event = AnnotationEvent(request, annotation, action)
+    if annotation_dict is None:
+        annotation_dict = AnnotationJSONPresenter(request, annotation).asdict()
+
+    event = AnnotationEvent(request, annotation_dict, action)
     request.notify_after_commit(event)
 
 
--- a/h/app.py
+++ b/h/app.py
@@ -4,6 +4,7 @@
 
 from __future__ import unicode_literals
 
+import collections
 import logging
 
 from pyramid.tweens import EXCVIEW
@@ -44,6 +45,32 @@ def create_app(global_config, **settings):
     config = configure(settings=settings)
     config.include(__name__)
     return config.make_wsgi_app()
+
+
+class EventQueue(object):
+    def __init__(self, request):
+        self.request = request
+        self.queue = collections.deque()
+
+        request.add_response_callback(self.response_callback)
+
+    def __call__(self, event):
+        self.queue.append(event)
+
+    def publish_all(self):
+        while True:
+            try:
+                event = self.queue.popleft()
+            except IndexError:
+                break
+            self.request.registry.notify(event)
+
+    def response_callback(self, request, response):
+        if request.exception is not None:
+            return
+
+        with request.tm:
+            self.publish_all()
 
 
 def includeme(config):
@@ -93,6 +120,9 @@ def includeme(config):
             "style-src": ["'self'", "fonts.googleapis.com"],
         },
     })
+
+    # After commit events
+    config.add_request_method(EventQueue, name='notify_after_commit', reify=True)
 
     # API module
     #
--- a/h/test/app_test.py
+++ b/h/test/app_test.py
@@ -1,0 +1,57 @@
+# -*- coding: utf-8 -*-
+
+import mock
+import pytest
+
+from pyramid.testing import DummyRequest
+
+from h import app
+
+
+class TestEventQueue(object):
+    def test_init_adds_response_callback(self):
+        request = mock.Mock()
+        queue = app.EventQueue(request)
+
+        request.add_response_callback.assert_called_once_with(queue.response_callback)
+
+    def test_call_appends_event_to_queue(self):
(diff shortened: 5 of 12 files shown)


seanh:
Can you make a dict from the annotation (e.g. via `AnnotationJSONPresenter.asdict()`) and put that dict in the event? Then the receiver of the event has all the data (as it was at the time of the event being sent) as a dict not a detached orm object, and they don't need to go to the db

chdorner:
@seanh This would work for some, but I don't think we can assume that all subscribers can work with the dictionary instead of the Annotation object (the [reply_template](https://github.com/hypothesis/h/blob/master/h/notification/reply_template.py) for example)

seanh:
I think reply_template ought to be able to work with a dict, which is different from saying it does, though.

Maybe the first event subscriber that gets the event and needs the annotation can fetch it and save it on the event object. Each subscriber can check for `event.annotation` and if it's not there fetch it and add it.

nickstenning:
We have at least three options, I think:

1. Events receive annotation mapped objects.
2. Events receive some (possibly partial) serialized representation of annotations.
3. Events receive only IDs.

Option 1. is problematic because there are parts of the annotation model that aren't usable when the [instance is in a detached state](http://docs.sqlalchemy.org/en/latest/orm/session_state_management.html). We could in principle make sure the annotation model has no lazy attributes, but this probably creates a load more database traffic, and it's worth working out whether or not we even need the extra data before we go down this route.

Options 2 and 3 don't have any immediate problems, but (as highlighted by Christof) can cause extra work for subscribers and their downstream consumers.

---

In order to work out what we need, let's make sure we know what we currently use these events for:

1. The "realtime" annotation streamer. At the moment [the subscriber for this](https://github.com/hypothesis/h/blob/dfac3a7c8da031b09eb6362e2a898f8f57c1f0f6/h/subscribers.py#L27) presents the annotation to JSON and then pushes it into Rabbit. At the other end the consumer [reconstructs the annotation object](https://github.com/hypothesis/h/blob/dfac3a7c8da031b09eb6362e2a898f8f57c1f0f6/h/streamer/messages.py#L94) but proceeds to use it mostly as a dictionary.

2. Reply notification emails. At the moment [the subscriber for this](https://github.com/hypothesis/h/blob/dfac3a7c8da031b09eb6362e2a898f8f57c1f0f6/h/subscribers.py#L41) uses an annotation model instance to determine whether a reply notification should be sent, and then queues the actual sending of any mail in a Celery task.

That's it.

We're also proposing using `AnnotationEvents` to trigger updates of the search index. That means: on 'create' or 'update', generate some JSON and shove it in the search index. On 'delete', remove a given ID from the index.

---

So, given those requirements, which of options 1, 2, or 3 seem like the most sensible? I think 1 is probably unwise, as it:

- encourages putting logic in the subscriber, which is run on the web thread
- constrains our use of lazy loading in the model to minimise database load

2 is a possibility, and potentially has the advantage that the workers can use the serialised representation of the annotation rather than having to hit the database themselves. However:

- this potentially increases the complexity of every subscriber/worker, as they have to parse a JSON representation rather than using all the helpful functionality exposed by model instances
- it means increased cost on the web thread to serialise the annotation, even if the subscribers are only using the ID
- it means increased transport costs if subscribers send the full serialised representation through the queue

3. seems like the best option to me, although it will require some changes to how our existing use of background workers. It:

- moves the greatest amount of work off the web process by providing incentives to keep subscribers small
- means that worker processes will fetch their own copy of the annotation from the database, and thus they'll get the full power of the sqla model code.
- minimises transport costs in the message queueing system

But this approach would require a couple of modifications to the current code. Specifically:

- The "realtime" code will need to be updated to fetch its own annotation objects, do its own serialisation, and use model instances internally rather than dictionaries.
- The way that annotation deletion happens will probably have to change. Currently annotations are just deleted in the web worker, but the before-deletion serialisation is sent to the "realtime" code so it can work out (from the URLs) which clients need to receive that event. We'll probably need to add a "deleted" flag to the annotation table, which can be used to find deleted annotations, and then we can clean up deleted annotations periodically rather than immediately.
- The work currently being done in the subscriber for reply notifications will probably want to move into a Celery task.

chdorner:
Summary of the discussion we had about this today:

* This is mostly fine for now, an added 1-2 queries is not great and should be fixed in the future. But it's good enough to get it merged
* Real-time messaging and websockets support is broken when deleting annotations, this should be fixed before merging this.

chdorner:
This is ready for review now.

chdorner:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;Td18UaqW&#x2F;304-postgres-offline-search-index-worker"> Postgres: offline search index worker.</a>

chdorner:
Once this has been merged, #3242 needs rebasing, few changes (changing the `h.indexer` subscriber to pass `event.annotation_id` into the celery task), short testing to make sure it still works as expected, and then is ready for review/merge as well.

h/app.py
```diff
@@ -46,6 +47,32 @@ def create_app(global_config, **settings):
     return config.make_wsgi_app()
 
 
+def publish_after_commit_events(request, response):
+    if request.exception is not None:
+        return
+
+    if len(request.after_commit_events) == 0:
```
seanh: I think this can just be `if request.after_commit_events:`, but do you need to return here at all? Won't the `while` loop in `publish_all()` just do nothing if `after_commit_events` is `[]`?

h/app.py
```diff
@@ -94,6 +121,13 @@ def includeme(config):
         },
     })
 
+    # After commit events
+    config.add_request_method(EventQueue, name='after_commit_events', reify=True)
```
seanh: I wonder if we should make the interface for sending an event to be sent after commit look more similar to the normal interface for sending an event? Maybe something like:

```python
request.registry.notify(event)      # Send event immediately.
request.notify_after_commit(event)  # Send event after commit.
```

Just for consistency (maybe it can even be `request.registry.notify_after_commit()`? not sure) but also because I don't think we need the full API of a list object on the request (it looks like code that has access to the request could also do things like delete, insert, reorder the list, not just append to it?)

h/app.py
```diff
@@ -46,6 +47,32 @@ def create_app(global_config, **settings):
...
+        request.after_commit_events.publish_all()
+
+
+class EventQueue(collections.deque):
+    def __init__(self, request):
+        self.request = request
+
+    def publish_all(self):
+        while self.has_events():
+            event = self.popleft()
```
seanh: I like the use of deque and popleft() here to make sure we never send an event twice, smart

h/test/app_test.py
```diff
@@ -0,0 +1,70 @@
...
+        request = self.makerequest(events=[])
+        app.publish_after_commit_events(request, None)
+        assert not request.after_commit_events.publish_all.called
+
+    def test_publishes_events(self):
+        request = self.makerequest(events=[mock.Mock()])
+        app.publish_after_commit_events(request, None)
+        assert request.after_commit_events.publish_all.called
+
+    def makerequest(self, exception=None, events=None):
```
seanh: I've been using the name `mock_request()` for test helper functions that return request objects elsewhere, might be good to keep the naming style consistent

h/test/app_test.py
```diff
@@ -0,0 +1,70 @@
...
+        if events:
+            event_queue.__len__.return_value = len(events)
+
+        request = DummyRequest(tm=mock.MagicMock(),
+                               exception=exception,
+                               after_commit_events=event_queue)
+        return request
+
+class TestEventQueue(object):
+    def test_publish_all_pops_from_left(self, popleft):
```
seanh: I wonder if what we really want to test here is that `self.request.registry.notify()` is called with the events in the right order? And/or that there are no more events on the queue after `publish_all()` has completed successfully? Testing that it calls `popleft()` seems a little like testing an internal implementation details of the method itself to me, and I'm not sure what the purpose of the test it.

h/app.py
seanh: Oh, and the use of `while` to be able to iterate over the list while modifying it :)

h/app.py
```diff
@@ -46,6 +47,32 @@ def create_app(global_config, **settings):
...
+            self.request.registry.notify(event)
+
+    def has_events(self):
+        if len(self) == 0:
+            return False
+        return True
```
seanh: `return len(self) == 0`. But this might even be simple enough that it doesn't even need to be a separate method? Actually you don't even need `len()`, you _could_ do:

```python
while True:
    try:
        event = self.popleft()
    except IndexError:
        break
    self.request.registry.notify(event)
```

which is an "easier to ask for forgiveness than permission" approach.

(I don't think there's anything particularly wrong with the `len()` based approach though, at least in this case when th

h/api/events.py
```diff
@@ -4,9 +4,10 @@
 class AnnotationEvent(object):
     """An event representing an action on an annotation."""
 
-    def __init__(self, request, annotation, action):
+    def __init__(self, request, annotation_id, serialized_annotation, action):
         self.request = request
-        self.annotation = annotation
+        self.annotation_id = annotation_id
+        self.serialized_annotation = serialized_annotation
```
seanh: This is a dict right? Would `annotation_dict` be a better name?

seanh: Is `annotation_id` needed? Or can code just use `serialized_annotation["id"]`? I guess it's not doing any harm though and it's convenient

h/api/test/subscribers_test.py
```diff
@@ -29,14 +29,23 @@ def test_it_does_not_delete_annotations_if_postgres_is_off(self, delete):
...
 
-    def test_it_calls_index_when_action_is_create(self, index, delete):
+    def test_it_fetches_the_annotation_when_action_is_create(self, fetch_annotation):
+        event = self.event('create')
+        event.request.feature.return_value = True
+
+        subscribers.index_annotation_event(event)
+
+        fetch_annotation.assert_called_once_with(event.request,
+                                                         event.annotation_id)
```
seanh: Looks like an whitespace typo here

h/api/views.py
```diff
@@ -177,11 +177,13 @@ def create(request):
                                                          legacy_appstruct)
 
     if request.feature('postgres'):
-        _publish_annotation_event(request, annotation, 'create')
-        return AnnotationJSONPresenter(request, annotation).asdict()
+        presented = AnnotationJSONPresenter(request, annotation).asdict()
+        _publish_annotation_event(request, annotation, 'create', serialized=presented)
+        return presented
```
seanh: I think the variable name `presented`, the param name `serialized` and the method name `asdict()` are all different terms for the same thing. Maybe rename both `presented` and `serialized` to `annotation_dict`?

seanh:
I'm happy with the general approach here, made a bunch of minor comments above

h/app.py
chdorner: I would like to return early here so we don't have to start a new database transaction, most requests won't have any after-commit events, so we would just waste some extra round-trips to the database server.
I've changed it to `if not request.after_commit_events:` though.

h/app.py
chdorner: Sure. I still need to put the list on the request though for storing the events for the life-time of one request. Unless there is another way to handle that?

How about keeping it like that (or renaming to `after_commit_event_queue`) and then adding a new `notify_after_commit` method to the request which basically does `request.after_commit_event_queue.append(...)`. Then we have a nicer API but can still rely on Pyramid handling the storage in memory for each request.

h/test/app_test.py
chdorner: :+1: Fixed.

h/test/app_test.py
chdorner: Yes, so that it processes the events in the right order (FIFO).

h/app.py
chdorner: I split this out into it's own method for easier testing of the `publish_all` method. But seems like this is not needed anymore. I did implement your suggestion.

h/app.py
seanh: Ah, I didn't realise that would trigger a db transaction. Nice! Might be worth adding a comment or test to document that too, otherwise someone might come along and "optimise" it later

h/app.py
seanh: Can't think of anywhere you can put the list off the top of my head besides on the request. I think `request._after_commit_event_queue` (mark it as private) is fine.

h/test/app_test.py
seanh: I think you should test that `self.request.registry.notify()` is called the right number of time with the rights args in the right order

h/app.py
nickstenning: Not fully up-to-speed with the context here, but I'd suggest keeping "private" (i.e. implementation-specific) data structures on the registry, and "public" interfaces (such as for enqueueing an event) on the request object as request methods.

h/api/events.py
chdorner: Changed it to be `annotation_dict` and made the annotation id a property on the event which fetches it from the dict :+1:

h/app.py
chdorner: Did add a test for that :+1:

h/app.py
chdorner: As discussed in Slack, this is now implemented like the `h.features.Client` as a callable class. Adding events is done via `request.notify_after_commit(event)` as Sean suggested, but everything is done inside `h.app.EventQueue` (the internal `collections.deque`, adding the response callback, the response callback itself, etc.)

h/test/app_test.py
chdorner: Implemented that, it's much nicer :+1:

nickstenning:
There's a bunch of tests and view code in here that still references `after_commit_event.append`. You'll want to update that to reference `notify_after_commit`, I'd think.

chdorner:
@nickstenning ah, there was some rebase weirdness going on, had to fix that file multiple times. It is good now though!

h/app.py
seanh: Sounds good

nickstenning:
Ok, looking good to me. Just poking around locally, making sure it works as advertised.

PR #3252, opened 2016-04-26
Fix a couple of slow front-end tests and enable slow test reporting

robertknight:
This fixes a couple of front-end tests running much more slowly, relatively speaking, than they should have been. One of these, the 'Bridge' test, has flaked out on Travis a few times in the past.

I've also enabled slow test reporting in Karma with a threshold set at 500ms (vs. the test timeout of 2000ms) which should help us to spot slow tests before they cause Travis failures.

### DIFF
--- a/h/static/scripts/test/bridge-test.coffee
+++ b/h/static/scripts/test/bridge-test.coffee
@@ -37,7 +37,9 @@ describe 'Bridge', ->
 
     it 'adds the channel to the .links property', ->
       channel = createChannel()
-      assert.include(bridge.links, {channel: channel, window: fakeWindow})
+      assert.isTrue(bridge.links.some((link) ->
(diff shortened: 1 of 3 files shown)


h/static/scripts/test/bridge-test.coffee
```diff
@@ -37,7 +37,9 @@ describe 'Bridge', ->
 
     it 'adds the channel to the .links property', ->
       channel = createChannel()
-      assert.include(bridge.links, {channel: channel, window: fakeWindow})
+      assert.isTrue(bridge.links.some((link) ->
```
robertknight: See the commit message for an explanation. The reason for this change is to avoid a deep equality test on the _values_ of the `channel` and `fakeWindow` objects.

nickstenning:
:pray:

PR #3253, opened 2016-04-26
Convert AnnotationViewerController to JS

robertknight:
This is a straightforward conversion of the existing code, no refactoring.

PR #3254, opened 2016-04-26
New threading UI implementation

robertknight:
**This should be reviewed after #3244 and ~~#3253~~, after which this will need rebasing**

This is a new implementation of annotation UI threading, implementing the design in #3176 . #3244 covers the first part, which is the process of taking a list of annotations from the server and converting them into a sorted list of nested `Thread` objects that mirrors how annotations and replies appear in the sidebar.

This PR implements the second and third parts:
1. Listening for annotations being loaded and unloaded and rebuilding the `Thread` objects, in `root-thread.js` and `annotation-ui.js`.
2. Rendering the Thread objects as a list of nested annotation cards, in `annotation-thread.{js,html}`

Things to focus on in functional testing are:
- Basic display of annotation cards and replies
- Toggling the expanded/collapsed state of replies
- Switching groups
- Changing the sort order
- Changing the search query in the sidebar
- Using 'View N more in conversation' to show hidden annotations in a conversation thread that do not match the search query. Note that this is only shown when a search is active in the sidebar and only _some_ of the annotations in a thread match the query.

For functional testing, a Chrome extension build is available from https://www.dropbox.com/s/3fm3xrk9a96bjw0/new-threading-implementation.crx?dl=0

This PR adds one new dependency, [seamless-immutable](https://github.com/rtfeldman/seamless-immutable) which is used to 'deep-freeze' Thread objects in development builds and catch unintended modifications to them.


robertknight:

seanh:
This "2 replies" link looks like it should collapse/uncollapse just the two second level replies, not the first level reply as well:

untitled screencast - edited 15

dwhly:
> This "2 replies" link looks like it should collapse/uncollapse just the two second level replies, not the first level reply as well

Correct.  The single line collapse is what the triangle is for.

robertknight:
> This "2 replies" link looks like it should collapse/uncollapse just the two second level
> replies, not the first level reply as well:

This makes sense. However the behavior here is the same is on master and as this isn't a completely trivial change to make, I'd like to keep this behavior as-is and revisit separately.

nickstenning:
Is there any way we can break this work into some smaller steps? Reviewing a 2.5KLoC diff is not my idea of a good time 😉 .

robertknight:
I'll have another look over and see what I can do.

robertknight:
Closing this PR in favor of several smaller, more incremental PRs.

PR #3255, opened 2016-04-27
Clean up reply notifications code

nickstenning:
This is a bunch of hopefully behaviour-preserving cleanups to the code that is responsible for generating reply notifications.

- Rename `h.notification.reply_template` to `h.notification.reply`
- Simplify and consolidate `generate_notifications` into `get_notification`, and introduce the `h.notification.reply.Notification` data container.
- Fix a bunch of issues with the unsubscribe view.
- Clean up unneeded code in the `Subscriptions` model.
- Remove an unneeded `LoginEvent` subscriber.


nickstenning:
This is ready for review. I'm not sure why @codecov-io is unhappy, but I don't think it's cause for concern.

@finitud, @seanh, fancy taking a look?

finitud:
LGTM

PR #3256, opened 2016-04-27
View and storage code for updating annotations in Postgres

seanh:
This pull request contains:

* New `views` and `storage` code for updating annotations in Postgres.

* New code to index annotation updates into the new Es index synchronously

* Bug fix: don't crash when fetching an annotation with an invalid ID from postgres (iirc this happens when creating an annotation with an invalid ID in its references)

* Changes to the `schemas` code for updating annotations in Postgres, which had already been merged into master. To be compatible with the legacy behaviour we need to allow partial updates (sort of `PATCH`-like) when only some of the fields of an annotation are sent in an update annotation request. On master if you omit the tags field in an update for example, for the postgres storage it will do the same thing it does on a create: it will replace any existing tags with the default value which is an empty list. Same for the rest of the fields. But the legacy behaviour in this instance is that if no tags are passed in an update (not even an empty list) then _don't change the annotation's tags_.

   Updating the Postgres code to match the legacy code in this regard is actually a change to `schemas` not to `storage`. `storage` just accepts whatever fields it gets from `schemas` and saves them to the model, so if `schemas` sends `'tags': []` then `storage` will overwrite the annotation's existing tags with an empty list. So we change schemas to send no `'tags'` key in the data, if none was sent in the request.

* Change to `schemas` which so that it prevents you from changing the references or group of an annotation by deleting these from the data before it gets sent to `storage`. <del>That is unfortunately currently mixed in with other changes in a commit whose commit message describes only those other changes, I'll split it out tomorrow.</del> (done)

* Change to `schemas` so that `permissions` is no longer a required field for creates and updates. Makes it consistent with the rest of the fields


seanh:
One known issue opened: https://github.com/hypothesis/h/issues/3261

h/api/storage.py
```diff
@@ -132,24 +165,69 @@ def create_annotation(request, data):
...
+    document = data.pop('document', None)
+
+    annotation = models.Annotation.query.get(annotation_id)
+
+    # Just modifying the annotation.extra dict in place doesn't let sqlalchemy
+    # know that it has changed and needs to be saved to the database.
+    # Work around this for now by copying, modifying then assigning.
+    extra = dict(annotation.extra)
+    extra.update(data.pop('extra', {}))
+    annotation.extra = extra
```
seanh: The correct fix for this (which I'd like to propose we do in a separate PR) is to use a dict-like class that also lets sqlalchemy know whenever it's mutated (exactly how you do this I dunno yet but it's a thing).

Christof pointed out that `target_selector` and `tags` also have the same problem. We're getting away with it on those two because we always completely replace them, never modify them in place.

h/api/storage.py
```diff
@@ -114,15 +118,44 @@ def create_annotation(request, data):
...
+        request.db, annotation, document_meta_dicts, document_uri_dicts)
+
+    return annotation
+
+
+def update_document_metadata(session,
+                             annotation,
+                             document_meta_dicts,
+                             document_uri_dicts):
+    """
```
seanh: I've added this function to `storage`'s public API. It is not called by any code outside of `storage`, but maybe it is a potentially useful thing to have in th

seanh:
This is ready to be reviewed now.

h/api/storage.py
```diff
@@ -132,24 +165,69 @@ def create_annotation(request, data):
...
+    annotation.extra = extra
+
+    for key, value in data.items():
+        setattr(annotation, key, value)
+
+    if document:
+        document_uri_dicts = document['document_uri_dicts']
+        document_meta_dicts = document['document_meta_dicts']
+        update_document_metadata(
+            session, annotation, document_meta_dicts, document_uri_dicts)
```
seanh: So what we do with the `"document"` on annotation update is exactly the same as what we do with it on create. Is this right?

seanh:
I could break this up into several consecutive pull requests...

h/api/storage.py
chdorner: Yes, `models.document` would be my preference.

h/api/storage.py
```diff
@@ -180,6 +180,43 @@ def update_document_metadata(session,
...
+
+    annotation = models.Annotation.query.get(annotation_id)
+
+    for key, value in data.items():
+        setattr(annotation, key, value)
+
+    update_document_metadata(
+        session, annotation, document_meta_dicts, document_uri_dicts)
+
+    return annotation
```
chdorner: Just wondering if it would make sense to check that user has access to write to the group, mainly for enforcing ACL with a buggy / badly acting client.

h/api/subscribers.py
```diff
@@ -10,7 +10,7 @@ def index_annotation_event(event):
     if not event.request.feature('postgres'):
         return
 
-    if event.action == 'create':
+    if event.action in ['create', 'update']:
```
chdorner: Depending on which PR gets merged first (see #3242), this has to move into the new `h.indexer`

h/api/storage.py
seanh: It's enforced by `UpdateAnnotationSchema` deleting any `groupid` from the data before sending it to `storage`. You can't update the group of an annotation.

If you created an annotation in a group, then left that group, you would still be able to update that annotation.

h/api/views.py
```diff
@@ -201,20 +202,37 @@ def read_jsonld(annotation, request):
...
 
-    presenter = AnnotationJSONPresenter(request, annotation)
-    return presenter.asdict()
+    # Validate the annotation for, and update the annotation in, PostgreSQL.
+    if request.feature('postgres'):
+        schema = schemas.UpdateAnnotationSchema(request, annotation=annotation)
+        appstruct = schema.validate(copy.deepcopy(json_payload))
+        annotation = storage.update_annotation(request.db,
+                                               annotation.id,
+                                               appstruct)
```
chdorner: You were saying that we support partial updating of annotations. We might have to build something that makes sure that partial updating doesn't fail in case we don't have the annotation in postgres yet. Although we probably will only miss a very small amount of the newest annotations for a short period after switching on the feature and until the last migration run has finished.

seanh: Hmm. I tried this:

1. Create an annotation with postgres on
2. Delete the annotation from postgres with psql.
3. Annotation will still be showing in the client so you can edit it. Try to update the annotation.

This simulates what would happen if we somehow had an annotation in the (new) Es index but not in Pg and someone tried to update it. What happens is that `h.api.resources.AnnotationFactory.__getitem__()` calls `storage.fetch_annotation()` which returns `None` (because it's not in the db) and then we 404. It doesn't get to the view, schema or storage code for updating the annotation.

Do we want to do something about this? Like on an annotation update attempt, if the annotation isn't found in Pg, attempt to fetch it from the legacy Es and create it in Pg?

I think maybe we just want to keep it simple and not handle this case on the update path

h/api/storage.py
seanh: @nickstenning Do you have an opinion on this?

nickstenning: I think for now `models.document` seems the right place. It's mostly calling helper functions that are already there, right?

seanh: Yeah. On reflection it definitely doesn't make sense as part of the `storage` public API, because a different `storage` implementation other than our Pg one might not need any `update_document_metadata()` at all.

I sort of think it belongs in the same place as `storage.create_annotation() and `update_annotation()` because it's "logic for dealing with the model" in the same way as they do. That argues for a private helper function, I guess.

But I'm fine with putting it in `h.api.models.document` and we do already have a bunch of "logic for dealing with the document models" helper functions in there already

h/api/storage.py
chdorner: :+1:
something as the `MutableDict` [here](http://docs.sqlalchemy.org/en/rel_1_0/orm/extensions/mutable.html) should be good, plus a `MutableList` for the `tags` and `references`. Even though we don't mutate some of these data structures but already set the new one on the existing model, I would still like to prevent future bugs by consistently using these special data structures with mutation tracking built-in.
But as you said, that should be done in a separate PR later.

h/api/storage.py
chdorner: I think this is how it currently works and how it is still supposed to work in the future:

* `DocumentMeta`
  * Update existing records (existing is the same `claimant_normalized` and `type` as defined by the unique constraint on the table)
  * Create new records when they don't exist
  * Never delete records that are not present in the current annotation's document data (just leave them as they are)
* `DocumentURI`
  * Ignore (or update) existing records (almost all fields are in the unique constraint, so there isn't much to update)
  * Create non-existing records
  * Never delete any records

h/api/views.py
chdorner: Yes, let's keep it simple and we should keep an eye on this during testing to see if it's even noticeable or not :+1:

h/api/storage.py
seanh: Definitely agree these are bugs waiting to happen. Made an issue for it https://github.com/hypothesis/h/issues/3271

h/api/test/storage_test.py
```diff
@@ -627,6 +623,74 @@ def db(self):
...
+
+    def test_it_updates_the_annotation(self, annotation_data, models):
+        annotation = models.Annotation.query.get.return_value = mock.Mock()
+
+        storage.update_annotation(mock.Mock(),
+                                  'test_annotation_id',
+                                  annotation_data)
+
+        for key, value in annotation_data.items():
+            assert getattr(annotation, key) == value
```
chdorner: This and the test above might benefit from going to the database directly and not using mocks. Otherwise looks good.

h/api/views.py
```diff
@@ -201,20 +202,37 @@ def read_jsonld(annotation, request):
...
-    schema = schemas.LegacyUpdateAnnotationSchema(request,
-                                                  annotation=annotation)
-    appstruct = schema.validate(_json_payload(request))
-    annotation = storage.legacy_update_annotation(request,
-                                                  annotation.id,
-                                                  appstruct)
+    json_payload = _json_payload(request)
 
-    _publish_annotation_event(request, annotation, 'update')
+    legacy_annotation = elastic.Annotation.fetch(annotation.id)
```
chdorner: You could just use `storage.fetch_annotation(request, annotation.id, _postgres=False)` to still go through the storage module but disable the feature flag check.

chdorner: And what's the reason to first fetch the legacy annotation? It doesn't seem to be used in the postgres block and it doesn't raise any exceptions when it can't find the annotation.

h/api/storage.py
seanh: Done

h/api/schemas.py
```diff
@@ -226,9 +223,11 @@ def validate(self, data):
         new_appstruct['text'] = appstruct.pop('text', u'')
         new_appstruct['tags'] = appstruct.pop('tags', [])
 
+        permissions = appstruct.pop('permissions', None)
+
         # Replace the client's complex permissions object with a simple shared
         # boolean.
-        if appstruct.pop('permissions')['read'] == [new_appstruct['userid']]:
+        if permissions and permissions['read'] == [new_appstruct['userid']]:
```
chdorner: I'm not sure if we want to make the assumption that when the permissions field is missing that it means the annotation is public?

h/api/views.py
seanh: Ok, this function was a mess, I've refactored it in a way that should make the code a lot clearer without changing the behaviour. Also stopped it from unnecessarily fetching the legacy annotation when postgres if off (in that case the legacy annotation is the one already passed to the view as argument). And called fetch_annotation() instead of models.elastic when it does need to fetch the legacy annotation, as you suggested.

h/api/schemas.py
seanh: This is an out of date diff but the current behaviour is still to create public annotations if no permissions are given.

We could make it default to private instead?

When hacking around on the command line, since all the fields are optional, it's really easy to create public annotations currently because you don't have to pass any perms. If we changed it to private by default you'd have to pass the pretty annoying permissions object to make a public annotation. 

On the other hand we'll be able to make the API accept a simple shared boolean in the future.

And maybe people hacking around with the API ought to be doing it in private by default!

So I'm happy to change it to private by default.

@nickstenning ?

h/api/test/storage_test.py
seanh: Hmm, I dunno. Maybe. On the other hand this isn't the `models` package, it's a user of the models package, so I feel it ought to mock out the models package and leave it up to the models package tests to make sure the db works. All `storage` does is set the values on the model object, and that's what this tests. I could be convinced..

h/api/storage.py
seanh: Ok, I think that's what it does. And it also calls `merge_documents()` and updates the `.updated` of the document. I believe this is correct then

h/api/models/test/document_test.py
```diff
@@ -434,6 +435,202 @@ def merge_data(self, request):
         return (master, duplicate)
 
 
+class TestUpdateDocumentMetadata(object):
+
+    def test_it_calls_find_or_create_by_uris(self,
```
nickstenning: You don't need to change this now, but in general it's really helpful to me when reading the tests if they're named in such a way that their purpose is clear. Test names that express motivation, not implementation, are very helpful:

```
test_retrieves_document_using_uri
test_finds_existing_documents
```

are probably better names than

```
test_it_calls_some_function
```

if for no other reason than a code refactoring that results in `some_function` being renamed will result in test names that make no sense. More broadly, good test names (or, failing that, good test docstrings) really help when trying to determine if a test failure is because the tests are broken or the code is broken.

h/api/schemas.py
```diff
@@ -156,79 +151,19 @@ class AnnotationSchema(JSONSchema):
                 'type': 'string',
             },
         },
-        'required': [
-            'permissions',
-        ],
     }
 
-    def __init__(self, request):
+    def __init__(self):
```
nickstenning: Seems like this function isn't necessary any more?

h/api/schemas.py
```diff
@@ -156,79 +151,19 @@ class AnnotationSchema(JSONSchema):
...
         super(AnnotationSchema, self).__init__()
-        self.request = request
 
     def validate(self, data):
```
nickstenning: Seems like this function isn't necessary any more?

h/api/schemas.py
```diff
@@ -367,10 +362,72 @@ def validate(self, data):
...
+
+    :param permissions: the permissions dict sent by the client in an
+        annotation create or update request
+    :type permissions: dict
+
+    :param userid: the userid of the user who created the annotation
+    :type userid: unicode
+
+    """
+    return permissions['read'] != [userid]
```
nickstenning: Is there a reason this is expressed this way round rather than `permissions['read'] == ['group:__world__']`? Defaulting to closed in the event of unrecognised data seems a bit safer than defaulting to open.

h/api/storage.py
```diff
@@ -33,15 +34,15 @@ def annotation_from_dict(data):
     return models.elastic.Annotation(data)
 
 
-def fetch_annotation(request, id, _postgres=None):
+def fetch_annotation(request, annotation_id, _postgres=None):
```
nickstenning: Is there a particular reason for this change? What kind of id would you pass into a function called `fetch_annotation` other than an annotation id?

h/api/models/test/document_test.py
seanh: Hmm, yeah this is a good point, this naming style's not great

h/api/schemas.py
seanh: Exactly what we were discussing [here](https://github.com/hypothesis/h/pull/3256#discussion_r61897991), I'll change it to private by default then

h/api/storage.py
```diff
@@ -114,42 +118,60 @@ def create_annotation(request, data):
...
+        document_uri_dicts = document['document_uri_dicts']
+        document_meta_dicts = document['document_meta_dicts']
+        models.update_document_metadata(
+            session, annotation, document_meta_dicts, document_uri_dicts)
 
     return annotation
 
 
-def update_annotation(request, id, data):
+def legacy_update_annotation(request, annotation_id, data):
```
nickstenning: Same question as above. I'm not sure `annotation_id` improves the readability of this at all... to me it's just "stuttering" like `conn_t conn = (conn_t) { ... }` in C.

h/api/schemas.py
seanh: Oh, but it wouldn't necessarily be `['group:__world__']` though, it could be shared within some other group. But I will make it default to private though

h/api/test/schemas_test.py
```diff
@@ -237,18 +237,8 @@ def test_it_raises_if_group_is_not_a_string(self):
 
         assert str(err.value) == "group: False is not of type 'string'"
 
-    def test_it_raises_if_permissions_is_missing(self):
```
nickstenning: Sorry, this a big diff and I can't find it. What's the context here? Is permissions no longer required? If so, how so?

h/api/storage.py
seanh: [It's 'cos pylint doesn't like it](https://github.com/hypothesis/h/pull/3256/commits/831f6467f815c44f5abe56a50fefa5a2f67444db). Overriding a builtin does seem bad. On the other hand I've never actually used the `id` builtin, so I would not be against disabling this pylint warning

h/api/schemas.py
seanh: Nope

h/api/schemas.py
seanh: [Private it is](https://github.com/hypothesis/h/pull/3256#discussion_r61923288)

h/api/test/schemas_test.py
```diff
@@ -42,13 +42,13 @@ def test_it_sets_appropriate_error_message_when_data_invalid(self):
 class TestAnnotationSchema(object):
```
nickstenning: Perhaps this is a silly question, but why/since when do we have 350 lines of tests all of which are basically testing `jsonschema`, and not our own code...?

h/api/storage.py
nickstenning: Ok. Nobody _ever_ uses the `id` function. Let's not follow Pylint's advice if it means writing worse code. Does the `good-names` directive in `.prospector.yaml` not catch this? I'm 👍 on disabling that warning if you want to.

h/api/views.py
```diff
@@ -184,7 +185,9 @@ def create(request):
     return AnnotationJSONPresenter(request, legacy_annotation).asdict()
 
 
-@api_config(route_name='api.annotation', request_method='GET', permission='read')
+@api_config(route_name='api.annotation',
```
nickstenning: It's really helpful when reviewing if semantic and non-semantic changes aren't bundled up in the same diff. Reformatting code to fit in 78 characters while making meaningful changes to the same code makes it much more likely that things are going to be missed, so let's not do this. Not least when we have the line-length rules disabled in our setup.cfg anyway because they're stupid.

h/api/storage.py
seanh: I'm good with this. My only plea is that we disable linter things that we don't follow, otherwise actually useful linter warnings get buried. We've a lot of rule disabling to do. Hoping to find some time to look into it at some point

nickstenning:
Okay, made a bunch of comments but in all honest this is an enormous diff, probably too big to meaningfully review in one piece.

Is there a straightforward way this could be broken up into chunks?

If not, then at this stage I'm happy to put some more time into reviewing it, but in general a 2KLoC diff is probably about an order of magnitude larger than I think we should be aiming for.

h/api/views.py
seanh: Sure, totally agree. This _was_ in a separate pep8 commit at least

seanh:
Ok, let me sort out the comments so far, then I can resubmit it as separate PRs. It should be straightforward to do so I think...

nickstenning:
Thank you so much! 🙏

h/api/test/schemas_test.py
seanh: I'd argue that these are testing that the schema that we've implemented in jsonschema does what we think it does, i.e. testing that we're using jsonschema correctly.

h/api/test/schemas_test.py
seanh: [It's no longer required](https://github.com/hypothesis/h/pull/3256/commits/34b8c2a6a5ae7d36850fc65b0582e11c4a7de2a1) because:

* When creating it defaults to shared (which I'm planning to change to not shared)
* When updating it defaults to no change
* Also to make permissions consistent with the rest of the fields (it was the only required one)

h/api/schemas.py
seanh: [Done](https://github.com/hypothesis/h/pull/3256/commits/5001fc42eb4687044b6741033a48fe91d06796e6)

h/api/schemas.py
seanh: This check is actually the same way that it's done [on master currently](https://github.com/hypothesis/h/blob/master/h/api/schemas.py#L184), I just moved it into a helper function.

To make it safer I could:

1. Return `True` if `len(permissions['read']) == 1 and permissions['read'][0].startswith('group:')`, or

2. Pass `appstruct['group']` into `_shared()` then return `True` if `permissions['read'] == ['group:{id}'.format(id=groupid)`.

Worth it?

h/api/test/schemas_test.py
nickstenning: Hmm. Not so sure. The JSON schema is a data structure. I'm not sure what a test saying "field foo is required" adds over a line in a JSON schema that says `required: True`.

Even if we think the tests add value, they're currently incredibly verbose and coupled to various details of jsonschema, such as error message formats.

Not for this PR, but if anything this should be a big list of failing cases with comments next to them, run through one test with `pytest.mark.parametrize`.

seanh: Yeah they're very repetitive, it does look like a strong candidate for a `parametrize()`, definitely.

I _do_ think the tests add value, though. Because I think it's totally possible to create a json schema that I think does one thing, but actually doesn't do exactly what I think, and these tests make sure that it does. The json schema is a complex and very nested data structure that it's totally easy (and I did) make mistakes with. It's either test everything manually or write a set of unit tests.

Also when making changes to a jsonschema, the corresponding unit test changes in the same commit will document (probably better than the commit message) exactly what the programmer was trying to change.

These tests would also allow us to substitute jsonschema with something else (that had the same interface) and know we hadn't changed behaviour.

In what ways are they too coupled to details of jsonschema? Other than the error message strings, but I don't think there's much to be done about that, I do want to test that when the validation is giving an error it's the error that I think it's giving. Otherwise the validation when called by the test could be failing for some completely different reason (e.g. there is something else invalid about the test data), the actual thing I'm trying to test could be broken, and the test could still be passing.

Also I just want to test that we're giving out the right error messages.

I don't think there's any other way to test "jsonschema validation failed for the reason I expected it to fail" other than asserting against the error message strings?

h/api/models/test/document_test.py
seanh: I've renamed most of these

h/api/schemas.py
seanh: [Done](https://github.com/hypothesis/h/pull/3256/commits/238d7989475ffbeb3233b090f72172dcd23a12b6)

h/api/schemas.py
seanh: Removed

h/api/schemas.py
seanh: Removed

h/api/storage.py
seanh: Done

h/api/storage.py
seanh: [Done](https://github.com/hypothesis/h/pull/3256/commits/9645f6351ab37ebccd883e8deca48a0992878c6a) by using `id_` instead of `annotation_id`

seanh:
I think I've responded to all the comments here, and I just rebased this on master, lots of conflicts. Next I'll split it up into multiple PRs

seanh:
This is now being split up into multiple consecutive PRs, the first of which is https://github.com/hypothesis/h/pull/3279

PR #3257, opened 2016-04-28
Show adder if user makes a selection and then launches the plugin.

robertknight:
If a user selects some text, and then realises that Hypothesis isn't
active, their first reaction will probably be to activate Hypothesis.
In this scenario, we should pop up the adder near the selection on launch.

 * If there is a selection when the 'ready' event fires, show the adder
   at the focus point of the selection.

 * For consistency, show the adder at the focus point of the selection
   after a selection whilst Hypothesis is active.

 * Add 'range-util' module which provides utility functions to get
   the coordinates at the end of the selection.

   'range-util' uses the DOM Range, Selection and NodeIterator APIs directly
   rather than Annotator's Range functions for several reasons:

   1) Given that H now requires IE >= 10, all of our target browsers
      have all the APIs that we need.

   2) Using the DOM APIs directly proved easier to debug and test.

### DIFF
--- a/h/static/scripts/annotator/range-util.js
+++ b/h/static/scripts/annotator/range-util.js
@@ -116,12 +116,13 @@ function getTextBoundingBoxes(range) {
 }
 
 /**
- * Returns the rectangle for the line of text containing the focus point
- * of a Selection.
+ * Returns the rectangle, in document coordinates, for the line of text
+ * containing the focus point of a Selection.
+ *
+ * Returns null if the selection is empty.
  *
  * @param {Selection} selection
- * @return {Object?} A rect containing the coordinates in the document of the
- *         line of text containing the focus point of the selection.
+ * @return {Rect?}
  */
 function selectionFocusRect(selection) {
   if (selection.isCollapsed) {
--- a/h/static/scripts/annotator/range-util.js
+++ b/h/static/scripts/annotator/range-util.js
@@ -1,0 +1,143 @@
+'use strict';
+
+function translate(rect, x, y) {
+  return {
+    left: rect.left + x,
+    top: rect.top + y,
+    width: rect.width,
+    height: rect.height,
+  };
+}
+
+function mapViewportRectToDocument(window, rect) {
+  // `pageXOffset` and `pageYOffset` are used rather than `scrollX`
+  // and `scrollY` for IE 10/11 compatibility.
+  return translate(rect, window.pageXOffset, window.pageYOffset);
+}
+
+/**
+ * Returns true if the start point of a selection occurs after the end point,
+ * in document order.
+ */
+function isSelectionBackwards(selection) {
+  if (selection.focusNode === selection.anchorNode) {
+    return selection.focusOffset < selection.anchorOffset;
+  }
+
+  var range = selection.getRangeAt(0);
+  return range.startContainer === selection.focusNode;
+}
+
+/**
+ * Returns true if `node` is between the `startContainer` and `endContainer` of
+ * the given `range`, inclusive.
+ *
+ * @param {Range} range
+ * @param {Node} node
+ */
+function isNodeInRange(range, node) {
+  if (node === range.startContainer || node === range.endContainer) {
+    return true;
+  }
+
+  /* jshint -W016 */
+  var isAfterStart = range.startContainer.compareDocumentPosition(node) &
+   Node.DOCUMENT_POSITION_FOLLOWING;
+  var isBeforeEnd = range.endContainer.compareDocumentPosition(node) &
+   Node.DOCUMENT_POSITION_PRECEDING;
+
+  return isAfterStart && isBeforeEnd;
+}
+
+/**
+ * Iterate over all Node(s) in `range` in document order and invoke `callback`
+ * for each of them.
+ *
+ * @param {Range} range
+ * @param {Function} callback
+ */
+function forEachNodeInRange(range, callback) {
+  var root = range.commonAncestorContainer;
+
+  // The `whatToShow`, `filter` and `expandEntityReferences` arguments are
+  // mandatory in IE although optional according to the spec.
+  var nodeIter = root.ownerDocument.createNodeIterator(root,
+    NodeFilter.SHOW_ALL, null /* filter */, false /* expandEntityReferences */);
+
+  /* jshint -W084 */
+  var currentNode;
+  while (currentNode = nodeIter.nextNode()) {
+    if (isNodeInRange(range, currentNode)) {
+      callback(currentNode);
+    }
+  }
+}
+
+/**
+ * Returns the bounding rectangles of non-whitespace text nodes in `range`.
+ *
+ * @param {Range} range
+ * @return {Array<Rect>} Array of bounding rects in document coordinates.
+ */
+function getTextBoundingBoxes(range) {
+  var whitespaceOnly = /^\s*$/;
+  var textNodes = [];
+  forEachNodeInRange(range, function (node) {
+    if (node.nodeType === Node.TEXT_NODE &&
+        !node.textContent.match(whitespaceOnly)) {
+      textNodes.push(node);
+    }
+  });
+
+  var rects = [];
+  textNodes.forEach(function (node) {
+    var nodeRange = node.ownerDocument.createRange();
+    nodeRange.selectNodeContents(node);
+    if (node === range.startContainer) {
+      nodeRange.setStart(node, range.startOffset);
+    }
+    if (node === range.endContainer) {
+      nodeRange.setEnd(node, range.endOffset);
+    }
+    if (nodeRange.collapsed) {
+      // If the range ends at the start of this text node or starts at the end
+      // of this node then do not include it.
+      return;
+    }
+
+    // Measure the range and translate from viewport to document coordinates
+    var viewportRects = Array.from(nodeRange.getClientRects());
+    nodeRange.detach();
+    rects = rects.concat(viewportRects.map(function (rect) {
+      return mapViewportRectToDocument(node.ownerDocument.defaultView, rect);
+    }));
+  });
+  return rects;
+}
+
+/**
+ * Returns the coordinates of the end (ie. at the focus point) of the text in a
+ * selection.
+ *
+ * @param {Selection} selection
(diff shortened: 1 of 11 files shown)


robertknight:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;1NElt1Oc&#x2F;297-show-annotation-adder-on-text-selected-before-launching-the-tool"> Show annotation adder on text selected before launching the tool</a>

seanh:
When selecting text with h already activated, I strongly prefer the positioning of the adder on this branch than on master.

On master, it's fairly easy to make it do weird things:

untitled screencast - edited 1

That's an extreme case but it can and I think it does happen in usage, and I think it's quite confusing - 
annotate / highlight what? The adder is not pointing at and is nowhere near to the selected text. The less skilled you are with your mouse or  touchpad the more likely this is to happen.

Here's a less extreme problem:

untitled screencast - edited 2

Adder is pointing right at some text, but _not_ the text that it's going to annotate! And is obscuring the text that it _is_ going to annotate. If the selection was smaller it might be completely obscured which would be even more confusing, no way at all to see that it isn't going to annotate "There are a few types of an...", but it's not.

Those cases and probably more that you could find are just completely wrong in my opinion.

But even when the adder does happen to appear in a reasonable position relative to the selected text, it does bug me that it's not exactly the same relative position each time, it's a little off each time in this way or that, it's inconsistent, just feels slightly broken to me.

The adder literally has an arrow on it, that arrow should be pointing at the selection, not at nothing or something else.

So I much prefer the behaviour on this branch. So much more consistent.

I would have put the adder's arrow at the top-middle of the selection though, like on iOS, not on the top-right. Clearer to point at the middle of a thing you're trying to indicate than at the edge. Also it just makes things line up nicer I think.

There is a problem on this branch when the selection is near to the edge of the window:

untitled screencast - edited 2

Try selecting text on iOS and looking at the popup - it has a nice solution to this. I imagine it'd be a little work to implement though.

seanh:
Couldn't resist one more example of the adder behaviour on master. Which piece of text do you think a new user will think it's going to annotate?

screenshot from 2016-04-28 20-26-35

seanh:
Screenshots of the other, smaller problem that I think the adder positioning has on master. When it doesn't go completely wrong as in the examples above, it just feels kind of messy. Same exact selection three times, adder position is slightly different each time:

screenshot from 2016-04-28 20-48-09
screenshot from 2016-04-28 20-47-38
screenshot from 2016-04-28 20-47-08

The same selection three times on this branch (much nicer):

screenshot from 2016-04-28 20-55-59

screenshot from 2016-04-28 20-55-59

screenshot from 2016-04-28 20-55-59

dwhly:
Agree with the notion that the adder shouldn't ever hang off the edge of the screen.  Particularly common on mobile I'd imagine.

seanh:
Another issue on this branch is when the selection is near the top of the window:

untitled screencast - edited 4

In that case I could scroll, but if there's no more room to scroll you just can't get it at all:

screenshot from 2016-04-29 09-08-47

My suggestion would be to detect this situation and make the adder appear below the selection instead of above it.

seanh:
This branch _does_ seem to handle well when I start a selection and then keep extending the selection downwards so that the top of the selection scrolls off the top of the window (the adder just appears at the bottom of the selection, so it's not a problem that the top of the selection is out of view):

untitled screencast 13

Similarly it does fine if I do the same thing but starting the selection at the bottom and extending upwards - the adder appears at the _top_ of the selection so it doesn't matter that the bottom is out of view (we are seeing a repeat of the selection-is-too-close-to-top-of-window problem reported earlier though):

untitled screencast 14

seanh:
There _is_ a problem if I make a selection whose top and bottom are both out of view e.g. by triple-clicking on a large paragraph:

untitled screencast 15

You see the adder appear briefly because Chrome first detects a double click and selects just one word,
then I click the third time and it extends the selection to the whole paragraph. Adder appears off screen. 

What iOS does in this situation: normally it prefers to put the adder _above_ the selection, if the top of the selection is in view and there's space above it for the adder to be in view. Secondarily it prefers to put the adder _below_ the selection, if the bottom of the selection is in view and there's space below it for the adder to be in view. If neither the top or bottom is in view, or they are but there isn't space in view above or below for the adder, then it just puts the adder bang in the middle of the selection (both horizontally and vertically centered, where the bounding box used for centering is the bounding box of _the part of the selection that is in view_). This obscures part of the selection with the adder, that's why either above or below is preffered if possible.

seanh:
You could implement this as three separate adder positioning functions (or classes or whatever): above, below and middle. Each one either returns a position for the adder or returns null if it would position the adder off screen. You just try each in turn until one succeeds. The final one always succeeds.

seanh:
So overall I think the aims for adder positioning should be, in order of importance:

1. The adder itself must always be completely in view

2. The arrow on the adder should always be "touching" the selection (either the adder is above the selection with its arrow pointing down, or below it with arrow up, or adder is over selection with arrow down). It doesn't literally have to be touching if we decide we _want_ a few pixels between selection and arrow that's fine. But you know what I mean - not pointing at empty space or at other text.

3. The adder should not obscure any part of the selection if at all possible (but if the adder can't be positioned either above or below the selection without violating 1, then you have to violate 3)

seanh:
.... aaaand just wanted to confirm that while the problems with the adder appearing off the right or top sides of the screen are pretty bad, they also happen with the adder behaviour on master. You could argue for merging this branch then fixing that as a separate issue I think.

The problem where you triple-click to select a large paragraph, resulting in a selection whose top and bottom are both out of view, does not happen on master of course (it just shows the adder where you clicked). I think that's probably a fairly obscure case, but one that should be fixed one day.

seanh:
Re-testing the new version of this...

seanh:
I _love_ how you positioned the adder near to the end of the selection (but with the arrow still clearly pointing at the selection) rather than in the middle as I suggested, so that it's closer to the mouse. That is really nice.

I was a little confused by this apparently unnecessary moving of the selection from top to bottom depending on the direction in which I made the selection:

untitled screencast - edited 6

However, I think it's to do with wanting the adder to appear near where the mouse is? As can be seen in multiline selections:

untitled screencast - edited 7

The bigger the selection, the more of a difference this touch makes to how close the adder likely is to the mouse pointer.

seanh:
This principle of putting the adder at the end of the selection so that it's nearer to the mouse can result in the adder obscuring part of the selection, if the selection is near the top or bottom of the window.

Here's a selection near the bottom of the window, so the adder can't appear below the selection:

untitled screencast - edited 8

And the reverse where the selection's at the top of the window:

untitled screencast - edited 10

Personally I would choose _don't obscure any of the selection_ over _put the adder closer to the mouse_ but I think it's borderline.

seanh:
It looks as if you have taken some care to prevent the adder from appearing out of view off the right edge of the window? Here's a selection near the right edge:

untitled screencast - edited 11

The adder is partly out of view, just a tiny bit, but enough in view that I can click either button. Completely in view plus a few pixels space before the edge would be nicer imo.

Funnily on the very left hand edge of the window the adder does seem to end up completely in view. But this gif shows a different problem which is that at the edges of the screen the arrow can end up not pointing at the selection:

untitled screencast - edited 13

What iOS does in this case is actually change the position of the arrow on the adder, leaving the main adder body in the same position but sliding the arrow along the adder until it is pointing at the selection.

seanh:
In this instance of selecting a large paragraph whose end is off the bottom of the screen by triple clicking, it would have been better for the adder to appear at the top of the selection:

untitled screencast 16

In the case where the _top_ of the large selection is off screen but the bottom is on screen, it does work well:

untitled screencast 17

seanh:
Similar problem when both the top and bottom of the selection are off screen, in this case you can't avoid obscuring some of the selection, but I think it's bad that the arrow is pointing at the bottom of the window, it would be better just to put the adder in the middle of the selection as iOS does:

untitled screencast 18

robertknight:
> However, I think it's to do with wanting the adder to appear near where the mouse is?
> As can be seen in multiline selections:

Yes. The idea is that the logic is basic but should give reasonable results and feel predictable.

> Completely in view plus a few pixels space before the edge would be nicer imo.

Yes. The bigger problem right now is that the sidebar is prone to popping open and covering up everything when you make a selection near the right edge of the window. I think that's something to come back to separately.

seanh:
You should reposition the adder when the selection changes after it has first been made:

untitled screencast 19

(Here I'm changing the selection with the keyboard, you can also do this on mobile)

seanh:
The adder should reposition itself after scroll events, or it can end up off screen while the selection is still on screen. This will confuse some people sometimes, trust me:

untitled screencast 19

If the_selection_ is scrolled off screen then the adder should allow itself to be scrolled off screen as well though. But watch out for the selection then being scrolled back on screen but the scrolling stopping while the adder's still off screen! Adder always needs to be above selection if selection is too near bottom of screen, always below selection it selection is too near top of screen, and this positioning always needs to be updated after selection change and after scroll events.

dwhly:
Excellent example of how seemingly simple things can be quite hard to do properly.

seanh:
This is already _so_ much better than the adder positioning on master. There are many details that aren't perfect yet (and I'm sure still more that we haven't found yet) but I'd be **very** happy to see the behaviour as it is on this branch now get into production.

h/static/scripts/annotator/range-util.js
```diff
@@ -0,0 +1,146 @@
...
+    rects = rects.concat(viewportRects.map(function (rect) {
+      return mapViewportRectToDocument(node.ownerDocument.defaultView, rect);
+    }));
+  });
+  return rects;
+}
+
+/**
+ * Returns the rectangle, in document coordinates, for the line of text
+ * containing the focus point of a Selection.
```
seanh: I'm wondering about a better name for this. I'm not entirely sure what focus rect / point means. I think what this function does is return the bounding box of the "last" word in the selection, where the definition of "last" depends on (our guess at) which direction the selection is made in? In that case `lastBoundingBox(selection)` might be a better name?

Also the function name says focus rect but the docstring says focus point, seems like the docstring is wrong.

PR #3260, opened 2016-04-28
Facelift for annotation cards.

sheetaluk:
Please don't merge, even if it is perfect :)

Things to clarify:
1. Should the style for "...More" and "Less" for annotation quotes be updated?
2. Should the style for tags change too?
3.What should the colour for "Only me" and the group name in the annotation_header__share-info be?
4. Should we restyle the annotator license that appears below the editor?
5. Also, what colour should all the links turn on hover?
6. Should the whole annotation card be clickable or should we make only the links(and clickable elements) clickable?

Trello: https://trello.com/c/aXCXxzx2/327-update-the-visual-design-of-the-annotation-cards

### DIFF
--- a/h/static/styles/annotation.scss
+++ b/h/static/styles/annotation.scss
@@ -2,18 +2,52 @@
 
 //ANNOTATION CARD////////////////////////////////
 
+@mixin font-small {
+  font-size: $small-font-size;
+  line-height: $small-line-height;
+}
+
+@mixin font-normal {
+  font-size: $normal-font-size;
+  line-height: $normal-line-height;
+}
+
 // the padding at the left-edge of the annotation card
 // and its container
 $annotation-card-left-padding: 10px;
 
 .annotation-card {
-  background-color: white;
+  box-shadow: 0px 1px 1px 1px rgba(0, 0, 0, 0.10);
   border-radius: 2px;
-  border: 1px solid $color-mercury;
-  padding-left: 12px;
-  padding-right: 12px;
-  padding-top: 12px;
-  padding-bottom: 15px;
+  padding: $layout-h-margin;
+  background: $white;
+}
+
+.annotation-card:hover {
+  box-shadow: 0px 2px 3px 0px rgba(0, 0, 0, 0.15);
+  border-radius: 2px;
+
+  .annotation-header__user {
+    color: $grey-7;
+  }
+
+  .annotation-quote {
+    border-left: $highlight 3px solid;
+    color: $grey-5;
+  }
+
+  .annotation-replies__link {
+    color: $grey-6;
+    text-decoration: underline;
+  }
+
+  .annotation-action-btn {
+    color: $grey-6;
+  }
+
+  .annotation-header__timestamp {
+    color: $grey-5;
+  }
 }
 
 .annotation {
@@ -24,13 +58,8 @@ $annotation-card-left-padding: 10px;
 }
 
 .annotation-link {
-  font-size: $body1-font-size;
-  color: $color-gray;
-  &:focus { outline: 0; }
-
-  .annotation:hover & {
-    color: $link-color;
-  }
+  @include font-small;
+  color: $grey-4;
 }
 
 .annotation-quote-list,
@@ -43,11 +72,11 @@ $annotation-card-left-padding: 10px;
   display: flex;
   flex-direction: row;
   align-items: flex-start;
-  margin-bottom: 5px;
 }
 
 .annotation-header__share-info {
-  color: $color-gray;
+  color: $grey-6;
+  @include font-normal;
 }
 
 .annotation-header__group {
@@ -60,10 +89,18 @@ $annotation-card-left-padding: 10px;
   margin-left: 5px;
 }
 
+.annotation-text {
+  @include font-normal;
+  color: $grey-6;
+}
+
 // the footer at the bottom of an annotation displaying
 // the annotation actions and reply counts
 .annotation-footer {
-  margin-bottom: 0;
+  @include font-normal;
+  color: $grey-5;
+  margin-bottom: 5px;
+  margin-top: 20px;
 }
 
 .u-flex-spacer {
@@ -71,8 +108,8 @@ $annotation-card-left-padding: 10px;
 }
 
 .annotation-quote-list {
-  margin-top: 14px;
-  margin-bottom: 14px;
+  margin-top: $layout-h-margin;
+  margin-bottom: $layout-h-margin;
 }
 
 .annotation-media-embed {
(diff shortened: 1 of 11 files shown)


robertknight:
> Should the style for "...More" and "Less" for annotation quotes be updated?

I believe this was already designed with the new design system in mind, so that won't be necessary, but we'll need to double-check this.

> Should the style for tags change too?

I'd suggest at the least the font style and size for tags should match the annotation body font size.

> What should the colour for "Only me" and the group name in the annotation_header__share-info be?

Conor can make an a final call on that but I'd suggest making it the same color as the icons at the bottom of the card.

> Should we restyle the annotator license that appears below the editor?

I think that probably will want tweaking but it can be done as a separate PR.

> Also, what colour should all the links turn on hover?

Which links do you mean specifically? Are you referring to links in the body of annotations which are still red at the moment?

All of the recently added clickable elements turn from lighter to darker grey on hover so I'd guess that we'll do the same for all links. I did have an accessibility concern that some of the un-hovered greys might be too light but that is something we can adjust.

> Should the whole annotation card be clickable or should we make only
> the links(and clickable elements) clickable?

The current behavior is that clicking anywhere on the annotation card should scroll the view to that card. I suggest we keep that as-is for the moment.

h/static/styles/annotation.scss
```diff
@@ -140,18 +176,18 @@ $annotation-card-left-padding: 10px;
...
-
   @include icons {
-    font-size: 13px;
+    font-size: $normal-font-size;
     vertical-align: -2px;
     margin-right: 1px;
   }
 }
 
+.annotation-license__link{
```
robertknight: Nit. Should have a space before the '{'

h/static/styles/variables.scss
```diff
@@ -42,6 +42,21 @@ $success-color: #1cbd41 !default;
 $mask-start-color: rgba($white, 0) !default;
 $mask-end-color: $white !default;
 
+
+// Variables for the new color palette
+// ------------------------------------
+$grey-1: #F2F2F2;
```
robertknight: There is another set of 'Variables for the new color palette' at the bottom of this file, with names like `lg-1`, `dg-2`. Can you make sure that we only have one set (eg. remove the others and replace with references to these ones)

h/templates/client/annotation.html
```diff
@@ -6,46 +6,36 @@
...
     <!-- User -->
     <span ng-if="vm.user()">
-      <span>
-        <a class="annotation-user"
-           target="_blank"
-           ng-href="{{::vm.serviceUrl}}u/{{vm.user()}}"
-           >{{vm.username()}}</a>
-      </span>
-
-      <span class="annotation-collapsed-replies">
```
robertknight: If you have a stack of nested replies, there is no longer any indication that there are replies to a collapsed reply. Is that an intentional design decision?

eg. If you have:

```
Annotation
|-- Reply A
      |--- Reply B
```

And you collapse 'Reply A', that used to show as:
```
Annotation
|-- Reply A (1 reply)
```
I don't particularly mind the change, just need to be sure that it is intentional.

robertknight:
I had some small comments on the CSS but this is a good start. We'll need to give Conor a chance to try this out and approve the changes before merging.

Some things I noticed comparing the current tip of this branch to the designs at https://trello-attachments.s3.amazonaws.com/571654d75b466478e53c0544/3500x600/68fc0bb587e3a1e79ca37230e8e7d279/Annotation_Card_-__specs.png 

- The annotation card should have a different shadow in its hovered and non-hovered states
- The color of the annotation body text should be different in hovered and non-hovered states, but the shade used in the hover state isn't labeled in the design
- The label of the 'Replies' link at the bottom of the card changed to 'Show replies (N)', and presumably 'Hide replies' if the replies are already visible. I presume this will be part of the "Reply threading" design card?
- The font used for annotation quotes changed from serif to sans-serif in the design

h/templates/client/annotation.html
sheetaluk: Line 136 has that same html. But this one, right here, I'm having a hard time figuring out where its being used.

h/static/styles/annotation.scss
sheetaluk: Done.

h/static/styles/variables.scss
sheetaluk: I'll do that as part of a clean up right after I restyle the annotation card and the replies.

h/static/styles/annotation.scss
```diff
@@ -2,18 +2,52 @@
...
+  line-height: $normal-line-height;
+}
+
 // the padding at the left-edge of the annotation card
 // and its container
 $annotation-card-left-padding: 10px;
 
 .annotation-card {
-  background-color: white;
+  box-shadow: 0px 1px 1px 1px rgba(0, 0, 0, 0.10);
```
robertknight: This does not quite match the spec. The box-shadow in the non-hovered state is `0px 1px 1px 0px` (see https://trello-attachments.s3.amazonaws.com/571654d75b466478e53c0544/3500x600/68fc0bb587e3a1e79ca37230e8e7d279/Annotation_Card_-__specs.png ). Sounds very subtle but makes a difference to how much emphasis hovered cards have

robertknight:
To clarify the comments above about nested replies: If there is a thread structured like this:

3260-nested-replies

On master when collapsed it looks like this:

3260 - nested replies collapsed

On this branch the reply count next to the reply is missing:

3260 - nested new-look replies collapsed

robertknight:
A comment I have on the design rather than the implementation is that I think the level of higher level of contrast on usernames is now a little distracting from the contents of the card and seems out of place with the greys that are used in the top bar. Thoughts @conordelahunty ?

conordelahunty:
>1. Should the style for "...More" and "Less" for annotation quotes be updated?
No, the current styling should work, but as Robert pointed out in a screen shot there seems to be a little visual bug where some of the third line is showing

>2. Should the style for tags change too?
No, we will keep this the same for now (it may get updated in the future when we work out how tags will look on Activity pages)

>3.What should the colour for "Only me" and the group name in the annotation_header__share-info be?
grey-5 (sorry, missed that on on the spec design)

>4. Should we restyle the annotator license that appears below the editor?
Leave it for now, we can treat that separately.

>5. Also, what colour should all the links turn on hover?
How do we handle this right now? Is it a specific # value or is it a percentage?

>6. Should the whole annotation card be clickable or should we make only the links(and clickable elements) clickable?
Let's keep the current behaviour for now.

Some other things:
- The font weight should be 400 (normal), not 300 as it currently is
- The spacing between sections (between the highlight and the annotation body, between the annotation body and the action buttons, etc.) is incorrect. Can you take another look. Some appear to be double the designed values.

conordelahunty:
> The color of the annotation body text should be different in hovered and non-hovered states, but the shade used in the hover state isn't labeled in the design
SUK: @conordelahunty please help.

grey-7

conordelahunty:
> A comment I have on the design rather than the implementation is that I think the level of higher level of contrast on usernames is now a little distracting from the contents of the card and seems out of place with the greys that are used in the top bar. Thoughts @conordelahunty ?

@robertknight I hear. Let's see if we can get the design as request done first and then see how it feels. I think fixing the font weight and the spacing of the annotation body might make an improvement too.

sheetaluk:
There changes have been pulled into: https://github.com/hypothesis/h/pull/3274/commits.

PR #3262, opened 2016-04-28
Make nipsa work with Postgres feature flag

chdorner:
This is mostly pretty straightforward:

* Emit `AnnotationBeforeSaveEvent` in `h.api.search.index` before sending the presented dictionary to Elasticsearch.
* Make sure the nipsa celery task adds/removes the `"nipsa": true` field to all indexed documents of a user
* Emit `AnnotationBeforeSaveEvent` in `h.subscribers.publish_annotation_event` before sending the real-time message so that `h.streamer` can filter out these annotations

### DIFF
--- a/h/subscribers.py
+++ b/h/subscribers.py
@@ -7,7 +7,7 @@ from h import emails
 from h import mailer
 from h.api import presenters
 from h.api import storage
-from h.api.events import AnnotationBeforeSaveEvent
+from h.api.events import AnnotationTransformEvent
 from h.notification import reply
 
 
@@ -31,8 +31,8 @@ def add_renderer_globals(event):
 def publish_annotation_event(event):
     """Publish an annotation event to the message queue."""
     annotation_dict = deepcopy(event.annotation_dict)
-    before_save_event = AnnotationBeforeSaveEvent(event.request,
-                                                  annotation_dict)
+    before_save_event = AnnotationTransformEvent(event.request,
+                                                 annotation_dict)
     event.request.registry.notify(before_save_event)
 
     data = {
--- a/h/subscribers.py
+++ b/h/subscribers.py
@@ -1,10 +1,13 @@
 # -*- coding: utf-8 -*-
 
+from copy import deepcopy
+
 from h import __version__
 from h import emails
 from h import mailer
 from h.api import presenters
 from h.api import storage
+from h.api.events import AnnotationBeforeSaveEvent
 from h.notification import reply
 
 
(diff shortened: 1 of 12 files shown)


h/subscribers.py
```diff
@@ -30,6 +31,10 @@ def publish_annotation_event(event):
     annotation_dict = presenters.AnnotationJSONPresenter(
         event.request, event.annotation).asdict()
 
+    before_save_event = AnnotationBeforeSaveEvent(event.request,
+                                                  annotation_dict)
+    event.request.registry.notify(before_save_event)
```
chdorner: I'm not super happy with the naming here since we are not _saving_ the annotation, but still need to allow `h.nipsa` to hook in and add the `nipsa` field when necessary.

chdorner:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;Ikz0hk7Z&#x2F;340-postgres-make-sure-nipsa-still-works"> Postgres: Make sure nipsa still works</a>

seanh:
Don't we want to just send one `AnnotationBeforeSaveEvent` for each annotation that is created or updated? This seems to send three with the "postgres" feature flag on. The legacy code in `storage.py` sends one, then the `publish_annotation_event()` sends one, then `search.index.index()` sends one.

I might be misunderstanding something but I'd think we want one event to be sent by `storage` before saving a new or updated event to the db?

chdorner:
@seanh you make a good point about how we want to have nipsa work with postgres. It is now implemented in a way that makes postgres unaware of any nipsa features but hooks into the synchronization of annotations from postgres to elasticsearch. I also mentioned above that I'm not super happy with the naming of the event, but wasn't sure if we want to rename / repurpose the event for a different use case as it is part of `h.api`'s public API.

If we decide to store the nipsa flag in Postgres as well then we would need to put it into the `extra` field and then build another hook into the rendering of annotations to remove the nipsa field from the JSON before it is being sent back to the user.

> This seems to send three with the "postgres" feature flag on. The legacy code in storage.py sends one, then the publish_annotation_event() sends one, then search.index.index() sends one.

Nipsa is currently the only code that listens to the `AnnotationBeforeSaveEvent`event, and it is emitted in:

* `h.api.storage._prepare` which is being called in `h.api.storage.update_annotation` (which will soon be `legacy_update_annotation`) and `h.api.storage.legacy_annotation`.
* `h.api.search.index` - new in this branch for getting the `nipsa` field into the new ES index.

After we remove the legacy storage code it will only be called in `h.api.search.index`. Which might justify a repurpose / rename of this event.

finitud:
Looking good. Is the idea that `nipsa` will keep being implemented via the ES index, or will we eventually have a `nipsa`-aware postgres?

chdorner:
@finitud I am not sure about that yet. We probably need to build something into the Postgres storage (`extra` field?) that works similarly, as soon as we start loading annotations from Postgres instead of ElasticSearch, but that isn't on the roadmap yet I believe.

finitud:
Makes sense. LGTM 👍

nickstenning:
Seems to work 👍

PR #3263, opened 2016-04-29
Build Firefox WebExtensions during Travis builds

robertknight:
_There is no hurry on this as it isn't part of our current focus but I've been maintaining this locally because its much nicer for regular use of H in Firefox than the bookmarklet. It is a fairly small set of changes and having regular builds will make it easy for me to communicate with Firefox developers to get any issues that we find fixed._

This adds a `hypothesis-buildext firefox` command which will build a Firefox WebExtension which is compatible with the current developer edition of Firefox and builds stage and production extensions on each Travis build in the same way that Chrome stage/prod extensions are built.

**What works?**
- Installation in Firefox Developer Edition builds
- The badge count
- Activating and deactivating the extension on a page
- If the URL contains a '#annotations' fragment, the extension will auto-inject

**What is missing?**
- The extension supports '#annotations' direct link fragments but communication between Bouncer and the Firefox extension in order to make Bouncer aware that an extension is present is not yet implemented.
- The resulting extension is not currently signed using the addons.mozilla.org (AMO) APIs so it can only be used with Developer Edition builds
- The extension will not check for updates
- Activating the extension on PDFs is not working yet

**Summary of changes in this PR**
 * Modify the `hypothesis-buildext` command to accept `firefox` as a browser argument. This runs the same code as the Chrome build except for the arguments passed when rendering the JSON manifest.
 * Add a handful of small workarounds for incompatibilities between Chrome and Firefox
 * Skip some of the installation and update-checking logic in the extension under Firefox, since these APIs are not yet implemented.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,8 +32,8 @@ matrix:
       addons:
         artifacts:
           paths:
-            # Upload all built zipfiles (browser extensions)
-            - $(ls build/*.zip | tr "\n" ":")
+            # Upload all built extension packages
+            - $(ls build/*.zip build/*.xpi | tr "\n" ":")
       before_install:
         # Unfortunately it seems to be impossible to call `nvm install` from an
         # external script on Travis, so we have to inline this one.
--- a/.jshintrc
+++ b/.jshintrc
@@ -12,7 +12,6 @@
   "esnext": true,
   "globals": {
     "chrome": false,
-    "h": false,
     "Promise": false,
     "chai": false,
     "sinon": false,
@@ -31,6 +30,7 @@
     "assert",
     "before",
     "beforeEach",
+    "chrome",
     "context",
     "describe",
     "it",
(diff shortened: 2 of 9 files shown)


.travis.yml
```diff
@@ -33,7 +33,7 @@ matrix:
         artifacts:
           paths:
             # Upload all built zipfiles (browser extensions)
```
nickstenning: Please update this comment :)

nickstenning:
This looks great, and I'm happy to merge it.

At some point perhaps it would be good to clear up the nomenclature -- this all lives in a folder called `h/browser/chrome`. Is that the right name for it, do you think?

robertknight:
> At some point perhaps it would be good to clear up the nomenclature --
> this all lives in a folder called h/browser/chrome. Is that the right name for it, do you think?

Agreed, this needs to change. I'm pretty sure that the vast majority of the code will be shared between all browser extensions. Say `h/browser-extension`? `h/extension` for the moment. When we re-organize the codebase in future to separate out the client and the website then I imagine there will be a root folder for the client and the extension in a folder underneath that.

I'd suggest that I move the files as part of a follow up PR to avoid mixing that with the code changes in this PR.

PR #3272, opened 2016-05-03
Remove compass-mixins dependency

robertknight:
The immediate problem this dependency causes is that it it fails to
build with the current version of node-sass (see
https://github.com/Igosuki/compass-mixins/issues/84) which in turn means
that H cannot be built with Node 6.0. Removing this dependency allows
the node-sass dependency to be unpinned so that the current version
can be used.

The primary functionality that compass-mixins was used for was vendor
prefixing, which is now taken care of by autoprefixer, plus CSS fixes
for browsers which we no longer support (IE < 10).

For the remaining handful of mixins for resets and color mixing which we
are still using, this commit just extracts the few lines of code for
each from the compass documentation and includes it directly in our SCSS
files.


nickstenning:
Seems ok here!

PR #3273, opened 2016-05-03
Add a utility script to assist with writing the changelog

robertknight:
Given an optional tag, it generates a formatted list of summaries
of changes merged to master since that tag was created.

Change descriptions are taken from pull request titles.

The script requires you to be a cool kid whose parents can afford Python 3.5. This is purely so I could play around with [mypy](http://mypy-lang.org/) and use `subprocess.run`, it isn't super-necessary. I'm happy to rewrite as a Python 2.x script if there are objections.

### DIFF
--- a/scripts/list-changes-since.py
+++ b/scripts/list-changes-since.py
@@ -1,0 +1,117 @@
+#!/usr/bin/env python
+
+import argparse
+import re
+import requests
+import subprocess
+import textwrap
+from subprocess import PIPE
+
+GITHUB_API_URL = 'https://api.github.com'
+
+
+class CompletedProcess(object):
+    def __init__(self, stdout, stderr):
+        self.stdout = stdout
+        self.stderr = stderr
+
+
+# Replace with subprocess.run() when we move to Python 3
+def run(args, **kw):
+    proc = subprocess.Popen(args, **kw)
+    (stdout, stderr) = proc.communicate()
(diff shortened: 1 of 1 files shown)


scripts/list-changes-since.py
```diff
@@ -0,0 +1,96 @@
+#!/usr/bin/env python3
+
+import argparse
+import re
+import requests
```
finitud: This requires an explicit `pip3 install requests`.

finitud:
Nice one! I'm not sure it's a great idea to be mixing python versions within one project though.

nickstenning:
I applaud your enthusiasm for shiny new features of Python 3, but yes -- while `h` is Python 2, its tooling should be Python 2 as well.

Also, I'd really like it if the script printed a warn to the effect of

> Please edit the output before including in CHANGES. Only include entries which have a reasonable chance of being interesting to a downstream consumer of this package, and use language which does not assume detailed knowledge of package internals where possible.

robertknight:
> I applaud your enthusiasm for shiny new features of Python 3,
> but yes -- while h is Python 2, its tooling should be Python 2 as well.
>
> Also, I'd really like it if the script printed a warn to the effect of

Will do on both counts.

robertknight:
I have updated the script to use Python 2.x and print an advisory message as per your suggestion Nick.

nickstenning:
Nice work.

PR #3274, opened 2016-05-03
Sheetaluk/336 reply threading

sheetaluk:
Please don't merge.

* I will add the css to change background colour of a conversation on hover after Rob's reply threading refactor is merged.
* What should show/hide replies in case of 1 reply be: show/hide reply (1), show/hide reply or show/hide replies (1)

### DIFF
--- a/h/static/styles/annotation.scss
+++ b/h/static/styles/annotation.scss
@@ -7,29 +7,56 @@
 $annotation-card-left-padding: 10px;
 
 .annotation-card {
-  background-color: white;
+  box-shadow: 0px 1px 1px 0px rgba(0, 0, 0, 0.10);
   border-radius: 2px;
-  border: 1px solid $color-mercury;
-  padding-left: 12px;
-  padding-right: 12px;
-  padding-top: 12px;
-  padding-bottom: 15px;
+  padding: $layout-h-margin;
+  background: $white;
+}
+
+.annotation-card:hover {
+  box-shadow: 0px 2px 3px 0px rgba(0, 0, 0, 0.15);
+
+  .annotation-header__user {
+    color: $grey-7;
+  }
+
+  .annotation-quote {
+    border-left: $highlight 3px solid;
+    color: $grey-5;
+  }
+}
+
+// When hovering a top-level annotation, show the footer in a hovered state.
+// When hovering a reply (at any level), show the reply's own footer in
+// a hovered state and also the footer of the top-level annotation.
+.annotation-card:hover > .annotation,
+.annotation:hover {
+  .annotation-replies__link,
+  .annotation-replies__count,
+  .annotation-action-btn {
+    color: $grey-6;
+  }
+
+  .annotation-header__timestamp {
+    color: $grey-5;
+  }
 }
 
 .annotation {
   display: block;
   font-family: $sans-font-family;
-  font-weight: 300;
+  font-weight: 400;
   position: relative;
 }
 
 .annotation-link {
-  font-size: $body1-font-size;
-  color: $color-gray;
-  &:focus { outline: 0; }
+  @include font-small;
+  color: $grey-4;
+}
 
-  .annotation:hover & {
-    color: $link-color;
+.annotation-replies:hover {
+  .annotation-replies__link {
+    text-decoration: underline;
   }
 }
 
@@ -42,12 +69,15 @@ $annotation-card-left-padding: 10px;
 .annotation-header {
   display: flex;
   flex-direction: row;
-  align-items: flex-start;
-  margin-bottom: 5px;
+  align-items: baseline;
+  // Margin between top of x-height of username and
+  // top of the annotation card should be ~15px
+  margin-top: -$layout-h-margin + 10px;
 }
 
 .annotation-header__share-info {
-  color: $color-gray;
+  color: $grey-5;
+  @include font-normal;
 }
 
 .annotation-header__group {
@@ -60,10 +90,26 @@ $annotation-card-left-padding: 10px;
   margin-left: 5px;
 }
 
+.annotation-body {
+  @include font-normal;
+  color: $grey-6;
+  // Margin between top of ascent of annotation body and
+  // bottom of ascent of annotation-quote should be ~15px.
+  margin-top: $layout-h-margin - 5px;
+  margin-bottom: $layout-h-margin;
+  margin-right: 0px;
+  margin-left: 0px;
+}
+
 // the footer at the bottom of an annotation displaying
 // the annotation actions and reply counts
 .annotation-footer {
-  margin-bottom: 0;
+  @include font-normal;
+  color: $grey-5;
+  // Margin between bottom of ascent of annotation footer labels
+  // and bottom of annotation card should be ~20px.
+  margin-bottom: $layout-h-margin - 13px;
+  margin-top: $layout-h-margin;
 }
 
 .u-flex-spacer {
@@ -71,8 +117,12 @@ $annotation-card-left-padding: 10px;
 }
 
 .annotation-quote-list {
-  margin-top: 14px;
-  margin-bottom: 14px;
+  // Margin between top of ascent of annotation quote and
+  // bottom of ascent of username should be ~15px
+  margin-top: $layout-h-margin - 5px;
+  // Margin between bottom of ascent of annotation quote and
+  // top of ascent of annotation-body should be ~15px
+  margin-bottom: $layout-h-margin - 3px;
 }
 
 .annotation-media-embed {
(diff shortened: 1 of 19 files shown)


robertknight:
Couple of nits looking at this branch: The baseline of the username and timestamp should be vertically aligned,as should the baseline of the "(Show|hide) replies" label and the action labels on the right.

336-reply-text-vert-positioning

robertknight:
> What should show/hide replies in case of 1 reply be: show/hide reply (1),
> show/hide reply or show/hide replies (1)

For the 'Show all annotations' button that is shown when an annotation is focused, we use 'Show all N annotations' when there are > 1 annotations and 'Show all annotations' if there is one or zero.

IMO it would make sense to display 'Show replies (1)', to make it clear that there _can_ be multiple replies and that this will show all of them if there are.

h/static/styles/thread.scss
```diff
@@ -48,15 +53,37 @@ $thread-padding: $annotation-card-left-padding;
...
+
+    &:hover {
+      & > .threadexp > span {
+        color: $grey-7;
+      }
+    }
+  }
+
+  .thread-message:hover {
+    & > div >.annotation-footer .annotation-action-btn,
```
robertknight: Can we reduce the specificity of the selectors? The more specific selectors get the harder it is to reason about what gets applied in different situations. Could this be simplified to:

```scss
.thread-message:hover {
  .annotation-action-btn, .annotation-replies__link, .annotation-replies__count {
    color: $grey-6;
  }
}
```

?

h/templates/client/annotation.html
```diff
@@ -55,34 +51,34 @@
...
+          | filter : {'type': 'TextQuoteSelector'}
+          track by $index"></blockquote>
     </excerpt>
   </section>
 
   <!-- / Excerpts -- >
 
   <!-- Body -->
-  <section name="text" class="annotation-body">
+  <section name="text" class="annotation-body annotation-text">
```
robertknight: What's the reason for introducing a second class here, could we just add the rules to the existing `.annotation-body` class?

robertknight:
Couple of minor quibbles but otherwise this is looking good. Once you've had a chance to look at those and if Conor is happy, I'm happy to merge this.

One issue I do see in this branch is that because the label of the reply count is now wider but we haven't yet removed the visible labels from the action buttons, you only have to reduce the sidebar width a small amount before the actions and replies get wrapped onto multiple lines.

3274 - multi-line actions

I think the solution for that though is just to get on and implement https://trello.com/c/w7tEeQiL straight after this.

h/static/styles/variables.scss
```diff
@@ -42,6 +42,21 @@ $success-color: #1cbd41 !default;
 $mask-start-color: rgba($white, 0) !default;
 $mask-end-color: $white !default;
 
+
+// Variables for the new color palette
```
robertknight: There are still a bunch of constants at the bottom of this file which are also labeled as the "new color palette" - `$dg1 .. $lg4, $primary, $highlight`. Can you remove all of those which are now unused.

h/static/styles/annotation.scss
```diff
@@ -7,29 +7,57 @@
...
-  padding-right: 12px;
-  padding-top: 12px;
-  padding-bottom: 15px;
+  padding: $layout-h-margin;
+  background: $white;
+}
+
+.annotation-card:hover {
+  box-shadow: 0px 2px 3px 0px rgba(0, 0, 0, 0.15);
+  border-radius: 2px;
```
robertknight: The border radius does not change on hover so we don't need to specify it again here.

h/static/styles/annotation.scss
```diff
@@ -42,12 +70,14 @@ $annotation-card-left-padding: 10px;
 .annotation-header {
   display: flex;
   flex-direction: row;
-  align-items: flex-start;
-  margin-bottom: 5px;
+  align-items: baseline;
+  // Align the baseline of the text with the bottom of a containing block element.
+  margin-top: -$layout-h-margin + 10px;
```
robertknight: This doesn't do what the comment above says - it is setting the top margin not the bottom margin. What might be better here is for the comment to explain the desired spacing from a visual point of view. eg:

`// Margin between x-height of username and top of annotation card should be ~Npx`

That way if we think of a better way of achieving this in future it'll be clear what we are aiming for.

h/static/styles/annotation.scss
```diff
@@ -60,39 +90,61 @@ $annotation-card-left-padding: 10px;
   margin-left: 5px;
 }
 
+.annotation-body {
+  @include font-normal;
+  color: $grey-6;
+  // Align the baseline of the text with the bottom of a containing block element.
```
robertknight: Again here I think it would be better for this comment to state the spacing that we are trying to achieve.

h/static/styles/annotation.scss
```diff
@@ -60,39 +90,61 @@ $annotation-card-left-padding: 10px;
...
+  margin-right: 0px;
+}
+
 // the footer at the bottom of an annotation displaying
 // the annotation actions and reply counts
 .annotation-footer {
-  margin-bottom: 0;
+  @include font-normal;
+  color: $grey-5;
+  // Align the baseline of the text with the bottom of a containing block element.
```
robertknight: Same for this comment

h/static/styles/annotation.scss
```diff
@@ -60,39 +90,61 @@ $annotation-card-left-padding: 10px;
...
 }
 
 .u-flex-spacer {
   flex-grow: 1;
 }
 
 .annotation-quote-list {
-  margin-top: 14px;
-  margin-bottom: 14px;
+  // Align the baseline of the text with the bottom of a containing block element.
```
robertknight: Same for this comment - I think it would be better for this comment to state the spacing that we are trying to achieve.

h/static/styles/excerpt.scss
```diff
@@ -5,6 +5,10 @@
   .excerpt {
     transition: max-height $expand-duration ease-in;
     overflow: hidden;
+
+    p {
```
robertknight: This styling is related to the excerpt's content rather than the excerpt, so it should live alongside the classes that are used to style the annotation content. In this case the `<markdown>` component unfortunately uses the generic `.styled-text` class which is in turn used in a bunch of other places.

One way to solve this would be to make `.styled-text` a mixin so that it can be re-used in different classes, create a `.styled-text` class which just imports that mixin so that other places which use that class today are unaffected and then replace the styling for the markdown body with a different class which imports that mixin and overrides the styling of paragraphs specifically. eg. https://gist.github.com/robertknight/a7ba647b95f6c30f44be9c3aac3810f8

The overriding goal here is to make it as easy as possible to find all the styling which affects a specific element.

h/static/styles/tags-input.scss
```diff
@@ -78,12 +78,13 @@ tags-input {
 
   .tag-list {
     @include pie-clearfix;
-    margin-top: -.4545em;
+    // Align the baseline of the text with the bottom of a containing block element.
```
robertknight: This comment should be updated in the same way as the others.

h/static/styles/thread.scss
```diff
@@ -38,8 +32,27 @@ $thread-padding: $annotation-card-left-padding;
...
 
-.thread-replies .thread:first-child {
-  margin-top: 0.5em;
+.thread-replies {
+  .annotation-action-btn {
+    color: $grey-4;
+  }
+
+  .annotation-footer {
+    // Align the baseline of the text with the bottom of a containing block element.
```
robertknight: These comments also need adjusting

h/static/styles/thread.scss
```diff
@@ -48,15 +61,22 @@ $thread-padding: $annotation-card-left-padding;
...
   // nested threads for annotation replies
   .thread {
-    border-left: 1px dotted $gray-light;
+    border-left: 1px dotted $grey-3;
     padding: 0;
     padding-left: $thread-padding;
-    margin-top: 10px;
+
+    &:hover {
+      & > .threadexp > span {
```
robertknight: I forget, what is this trying to achieve?

sheetaluk: This is the collapse/expand arrow on the left of the username in a reply.

h/static/styles/variables.scss
sheetaluk: There is another card for restyling the About this version dialog (which uses some of these variables). I'll take off all these variables as part of that card.

h/static/styles/thread.scss
robertknight: I'm going to merge this branch now but I suggest that we collaborate on getting the reply hover states implemented afterwards and clear this up afterwards.

As a general note, it is usually always preferable to try to simplify the CSS selectors down to just one or two classes if you can. If not, a comment explaining the desired visual effect is very helpful for future reference.

robertknight:
Looking good. This is a big step towards a design overhaul of H.

PR #3276, opened 2016-05-03
Use custom element names for adder toolbar

### LINKED ISSUE #3207: adder broken on example.com
### Steps to reproduce
1. activate H on example.com
2. make a selection

### Expected behaviour
Normal adder

### Actual behaviour

image

The reason:

    div {
        width: 600px;
        margin: 5em auto;
        padding: 50px;
        background-color: #fff;
        border-radius: 1em;
    }

Maybe nothing to be done about that?

robertknight:
Using custom element names for the adder's container is a simple trick
to reduce the likelihood of the adder's styling being affected by CSS
selectors in the containing page, such as example.com's `div` selector.

FWIW Genius uses the same ~~hack~~ technique.

Other techniques we could consider in addition are same-origin iframes and Shadow DOM (when available).

Fixes #3207

### DIFF
--- a/h/static/scripts/annotator/adder.html
+++ b/h/static/scripts/annotator/adder.html
@@ -1,10 +1,10 @@
-<div class="annotator-adder">
-  <div class="annotator-adder-actions">
-    <button class="annotator-adder-actions__button h-icon-annotate" data-action="comment">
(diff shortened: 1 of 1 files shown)


nickstenning:
LGTM.

PR #3277, opened 2016-05-04
Remove unused IncludeRawExtension

nickstenning:
This was previously used to include angular templates in `app.html` without parsing them as Jinja2 templates. These templates are now included in the script bundles directly using the `stringify` browserify transform.


robertknight:
LGTM

PR #3279, opened 2016-05-05
Storage code for updating annotations in postgres

seanh:
@nickstenning As requested, the first smaller PR split out of <https://github.com/hypothesis/h/pull/3256>. This one contains the storage code only for updating annotations in Postgres, as well as some code cleanup (PEP8 and lint) and some changes that were all ready requested and made on <https://github.com/hypothesis/h/pull/3256>.

This is still probably big enough that it's worth looking at the individual commits rather than the final diff.

Several more pull requests to follow this one.


h/api/test/storage_test.py
```diff
@@ -218,17 +218,79 @@ def mock_request(self):
...
+
+    def test_it_calls_prepare(self, models, partial, transform):
+        storage.legacy_update_annotation(DummyRequest(),
+                                         'test_annotation_id',
+                                         {})
+
+        transform.prepare.assert_called_once_with(
+            models.elastic.Annotation.fetch.return_value, partial.return_value)
+
+    def test_it_inits_AnnotationBeforeSaveEvent(self,
```
chdorner: s/AnnotationBeforeSaveEvent/AnnotationTransformEvent/

h/api/storage.py
```diff
@@ -180,6 +180,46 @@ def update_document_metadata(session,
...
+
+    :returns: the updated annotation
+    :rtype: h.api.models.Annotation
+
+    """
+    # Remove any 'document' field first so that we don't try to save it on the
+    # annotation object.
+    document = data.pop('document', None)
+
+    annotation = models.Annotation.query.get(annotation_id)
```
chdorner: This should go through the passed-in session: `session.query(models.Annotation).get(annotation_id)`

h/api/models/test/document_test.py
```diff
@@ -437,10 +437,10 @@ def merge_data(self, request):
 
 class TestUpdateDocumentMetadata(object):
 
-    def test_it_calls_find_or_create_by_uris(self,
-                                             annotation,
-                                             Document,
-                                             session):
+    def test_it_uses_the_target_uri_to_get_the_document(self,
```
chdorner: :+1: I like these names better then "it calls ..."

chdorner:
This looks good to me, other then a few nitpicky comments (some of them have already been addressed in later commits).

h/api/test/storage_test.py
seanh: Fixed

h/api/storage.py
seanh: Done

nickstenning:
Failing tests need a look.

h/api/models/test/document_test.py
```diff
@@ -434,6 +435,203 @@ def merge_data(self, request):
...
+        )
+
+    def test_if_there_are_multiple_documents_it_merges_them_into_one(
+            self,
+            annotation,
+            Document,
+            merge_documents,
+            session):
+        """If it finds more than one document it calls merge_documents()."""
+        Document.find_or_create_by_uris.return_value = mock.Mock(
```
nickstenning: Not going to block on this at this point, but I really think this would have been a prime candidate for just creating some data in the database and checking the end state was correct, rather than mocking out classmethods on `Document`.

h/api/models/test/document_test.py
```diff
@@ -434,6 +435,203 @@ def merge_data(self, request):
...
+
+    def test_it_calls_first(self, annotation, session, Document):
+        """If it finds only one document it calls first()."""
+        Document.find_or_create_by_uris.return_value = mock.Mock(
+            count=mock.Mock(return_value=1))
+
+        document.update_document_metadata(session, annotation, [], [])
+
+        Document.find_or_create_by_uris.return_value\
+            .first.assert_called_once_with()
```
nickstenning: For example, this test is brittle to the underlying implementation. Calling `queryobj.one()` or `queryobj.one_or_none()` would both be perfectly sensible ways of implementing `update_document_metadata`, and it seems a bit off that these would break the tests while the function could be doing its job perfectly.

Again, I don't think we need to change this, but I'm sharing it to give you some idea of the problems with overly literal tests of code that interacts with something like SQLAlchemy -- a library in which there are at least 3 ways to accomplish any given task.

h/api/storage.py
```diff
@@ -15,6 +15,7 @@
 from h.api import transform
 from h.api import models
 from h.api.events import AnnotationTransformEvent
+from h.api.db import types
```
nickstenning: This should be on the line above.

h/api/storage.py
```diff
@@ -114,42 +118,52 @@ def create_annotation(request, data):
...
+
+    """
+    # Remove any 'document' field first so that we don't try to save it on the
+    # annotation object.
+    document = data.pop('document', None)
+
+    annotation = session.query(models.Annotation).get(id_)
+
+    for key, value in data.items():
+        setattr(annotation, key, value)
```
nickstenning: One thing to be aware of here is that we need to be *really* sure that the client doesn't have the ability to manipulate the keys in the `data` dictionary beyond what we assume to be possible.

In this PR, nothing calls this function outside of tests, but let's just make sure we're aware of the issues with this kind of model assignment code, and not fall into [old traps](http://guides.rubyonrails.org/v3.2.9/security.html#mass-assignment).

seanh:
Tests should be fixed now

h/api/storage.py
seanh: This is done in the schemas module (in the other pull request)

h/api/storage.py
seanh: Done

h/api/models/test/document_test.py
seanh: <https://github.com/hypothesis/h/pull/3279/commits/a6572ace75149af48d93165d036d7da7b45ba1e9> is another good example of the same problem.

The reason I didn't use the real db though, was because I was thinking that `h.api.models` is our package that uses and wraps the db and sqlalchemy, and `storage` is a module that uses `models`. When one of our modules is using another one of our own module, we should mock out the other module in the tests.

There is definitely a problem here but I wonder if the problem is that the API of `models` isn't very good and that `sqlalchemy` is leaking out of models and being used directly in `storage`.

I wonder if we could improve the API of `models` and its encapsulation of sqlalchemy, so that `storage` tests could mock models without being so brittle?

seanh: Oh, except that `update_document_metadata()` isn't in `storage` anymore because it moved into `models`!

So in that case yes, probably it should not mock the db (the reason it does is that this function these tests were in `storage` and then they moved into `models` and I didn't rewrite the tests)

chdorner:
LGTM.

PR #3280, opened 2016-05-05
Schemas code for updating annotations in postgres

seanh:
A few changes to `h.api.schemas` related to updating annotations in Postgres:

- Do partial updates in the same way the legacy Es schemas/storage do https://github.com/hypothesis/h/pull/3280/commits/0d58f63a3f58c6d5364859acc5d3ab3bd2de4375
- Don't allow changing group or references in update, handle this in schemas so storage doesn't worry about it https://github.com/hypothesis/h/pull/3280/commits/caa4d5e4bb41a67b07d662420890f5082fb7c508
- Make permissions optional https://github.com/hypothesis/h/pull/3280/commits/127263a2c97ab7f1d2d91fa8bf517a685f55f2a8
- As well as some pep8 and lint, some bug fixes, and some changes previously requested and made on https://github.com/hypothesis/h/pull/3256

So this is a bit of a grab-bag. I could try to split this up into yet smaller PRs but hopefully this is small enough if reviewed commit by commit.

### DIFF
--- a/h/api/schemas.py
+++ b/h/api/schemas.py
@@ -161,68 +161,6 @@ class AnnotationSchema(JSONSchema):
         ],
     }
 
-    def __init__(self, request):
-        super(AnnotationSchema, self).__init__()
-        self.request = request
-
-    def validate(self, data):
-        appstruct = super(AnnotationSchema, self).validate(data)
-
-        new_appstruct = {}
-
-        # Some fields are not to be set by the user, ignore them.
-        for field in PROTECTED_FIELDS:
-            appstruct.pop(field, None)
-
-        new_appstruct['userid'] = self.request.authenticated_userid
-        new_appstruct['target_uri'] = appstruct.pop('uri', u'')
-        new_appstruct['text'] = appstruct.pop('text', u'')
-        new_appstruct['tags'] = appstruct.pop('tags', [])
-
-        # Replace the client's complex permissions object with a simple shared
-        # boolean.
-        if appstruct.pop('permissions')['read'] == [new_appstruct['userid']]:
-            new_appstruct['shared'] = False
-        else:
-            new_appstruct['shared'] = True
-
-        # The 'target' dict that the client sends is replaced with a single
-        # annotation.target_selectors whose value is the first selector in
-        # the client'ss target.selectors list.
-        # Anything else in the target dict, and any selectors after the first,
-        # are discarded.
-        target = appstruct.pop('target', [])
-        if target:  # Replies and page notes don't have 'target'.
-            target = target[0]  # Multiple targets are ignored.
-            new_appstruct['target_selectors'] = target['selector']
-
-        new_appstruct['groupid'] = appstruct.pop('group', u'__world__')
-        new_appstruct['references'] = appstruct.pop('references', [])
-
-        # Replies always get the same groupid as their parent. The parent's
-        # groupid is added to the reply annotation later by the storage code.
-        # Here we just delete any group sent by the client from replies.
-        if new_appstruct['references'] and 'groupid' in new_appstruct:
-            del new_appstruct['groupid']
-
-        new_appstruct['extra'] = appstruct
-
-        # Transform the "document" dict that the client posts into a convenient
-        # format for creating DocumentURI and DocumentMeta objects later.
-        document_data = appstruct.pop('document', {})
-        document_uri_dicts = parse_document_claims.document_uris_from_data(
-            copy.deepcopy(document_data),
-            claimant=new_appstruct['target_uri'])
-        document_meta_dicts = parse_document_claims.document_metas_from_data(
-            copy.deepcopy(document_data),
-            claimant=new_appstruct['target_uri'])
-        new_appstruct['document'] = {
-            'document_uri_dicts': document_uri_dicts,
-            'document_meta_dicts': document_meta_dicts
-        }
-
-        return new_appstruct
-
 
 class LegacyAnnotationSchema(JSONSchema):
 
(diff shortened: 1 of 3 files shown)


h/api/schemas.py
```diff
@@ -264,17 +192,50 @@ class CreateAnnotationSchema(object):
...
     def validate(self, data):
-        return self.structure.validate(data)
+        appstruct = self.structure.validate(data)
+
+        new_appstruct = {}
+
+        _remove_protected_fields(appstruct)
+
+        new_appstruct['userid'] = self.request.authenticated_userid
+        new_appstruct['target_uri'] = appstruct.pop('uri', u'')
```
chdorner: Does this default value mean that we would write an empty string as the `target_uri` when the client doesn't provide a `uri` field?

seanh: Yes, the default is to create an empty string target URI:

```
~ http POST 'http://localhost:5000/api/annotations' Authorization:'Bearer 6879-e967d22983d7f00b91fadf1a62c2e52c' text=howdy
HTTP/1.1 200 OK
Access-Control-Allow-Origin: *
Connection: close
Content-Length: 580
Content-Type: application/json; charset=UTF-8
Date: Fri, 06 May 2016 09:27:19 GMT
Server: gunicorn/19.4.5

{
    "created": "2016-05-06T09:27:19.328425+00:00",
    "document": {
        "link": [
            {
                "href": ""
            }
        ]
    },
    "group": "__world__",
    "id": "utThDhNsEeaCnXeXOQHVlw",
    "links": {
        "html": "http://localhost:5000/a/utThDhNsEeaCnXeXOQHVlw",
        "json": "http://localhost:5000/api/annotations/utThDhNsEeaCnXeXOQHVlw"
    },
    "permissions": {
        "admin": [
            "acct:fred@localhost"
        ],
        "delete": [
            "acct:fred@localhost"
        ],
        "read": [
            "acct:fred@localhost"
        ],
        "update": [
            "acct:fred@localhost"
        ]
    },
    "tags": [],
    "target": [
        {
            "source": ""
        }
    ],
    "text": "howdy",
    "updated": "2016-05-06T09:27:19.328432+00:00",
    "uri": "",
    "user": "acct:fred@localhost"
}

~ 
```

And in the db:

```
postgres=# SELECT target_uri FROM annotation ORDER BY created DESC LIMIT 1;
 target_uri
------------

(1 row)

postgres=#
```

chdorner: Hmm, we might want to not allow this specific case, because now we can introduce some sensible validations. For example, we [don't migrate](https://github.com/hypothesis/h/blob/master/scripts/migrate.py#L212-L213) any annotations with a missing or empty `uri` field

h/api/schemas.py
```diff
@@ -264,17 +192,50 @@ class CreateAnnotationSchema(object):
...
+        new_appstruct['text'] = appstruct.pop('text', u'')
```
chdorner: We might want to consider using `None` here instead of storing a bunch of empty strings in the database.

h/api/schemas.py
chdorner: On the other hand, these default values haven't been introduced with this PR, so I'm also very happy to have the discussion and then depending on what we decide do the implementation in a new PR.

seanh: Empty document link href is created I guess from the empty uri too

seanh: So what should we do in this case? Is it valid to have an annotation with no URI? Or should we reject this with a validation error? Rather than trying to validate that it's an actual valid URI, we could for now at least check that it's a non-empty string but silly things like "a" would still be allowed

seanh: (This is probably for a separate PR)

chdorner: Yes, not missing and not empty is probably good enough for now.

h/api/schemas.py
```diff
@@ -211,7 +211,7 @@ def validate(self, data):
             new_appstruct['shared'] = _shared(appstruct.pop('permissions'),
                                               new_appstruct['userid'])
         else:
-            new_appstruct['shared'] = True
+            new_appstruct['shared'] = False
```
chdorner: Might be nicer to move this into the `_shared` helper function:

```python
def _shared(permissions, userid):
    if permissions is None:
      return False
    
    # ...
```

chdorner: Hmm, maybe not. This only seems to be used in the `CreateAnnotationSchema` and not when updating.

h/api/schemas.py
seanh: If postgres off, the legacy storage will let me create an annotation with no uri _but_ it puts `null` rather than `''` as the default :(

```json
~ http POST 'http://localhost:5000/api/annotations' Authorization:'Bearer 6879-e967d22983d7f00b91fadf1a62c2e52c' text=hi
HTTP/1.1 200 OK
Access-Control-Allow-Origin: *
Connection: close
Content-Length: 564
Content-Type: application/json; charset=UTF-8
Date: Fri, 06 May 2016 10:20:51 GMT
Server: gunicorn/19.4.5

{
    "created": "2016-05-06T10:20:51.159868+00:00",
    "document": {},
    "group": "__world__",
    "id": "AVSFlYxYsSgLSOURFTGL",
    "links": {
        "html": "http://localhost:5000/a/AVSFlYxYsSgLSOURFTGL",
        "json": "http://localhost:5000/api/annotations/AVSFlYxYsSgLSOURFTGL"
    },
    "nipsa": true,
    "permissions": {
        "admin": [
            "acct:fred@localhost"
        ],
        "delete": [
            "acct:fred@localhost"
        ],
        "read": [
            "group:__world__"
        ],
        "update": [
            "acct:fred@localhost"
        ]
    },
    "tags": [],
    "target": [
        {
            "source": null
        }
    ],
    "text": "hi",
    "updated": "2016-05-06T10:20:51.159883+00:00",
    "uri": null,
    "user": "acct:fred@localhost"
}

~ 
```

seanh: Also it does not insert the empty document.link.href

seanh: Christof and I discussed this in Slack, I think we're both up for a separate follow-up PR that makes a non-empty URI required

h/api/schemas.py
seanh: Yeah, when updating the default behaviour is to not changed the shared value from whatever is currently in the DB.

h/api/schemas.py
seanh: You mean just for "text"? We can certainly do that

chdorner: Yes. Was just going through the default values we set here and the default values/(not-)NULL constraints on the database. We do allow `NULL` on the `text` column, so there's no need to store an empty string. At some later point we might want to assume that an empty string is considered `NULL` as well, which would make querying easier. But that might have other implications and is a whole different discussion :)

seanh: `AnnotationBasePresenter` will tidy this up on the way out (converting `None` to `''`), but anyone dealing with an `Annotation` model object directly will have to be aware that `annotation.text` might be `None` and not a string.

chdorner: Ah, let's leave it as it is then, it's probably not a good idea to change this now. I wasn't thinking it through properly.

h/api/schemas.py
seanh: Opened an issue for this https://github.com/hypothesis/h/issues/3291

h/api/schemas.py
```diff
@@ -367,10 +352,73 @@ def validate(self, data):
...
+    :param permissions: the permissions dict sent by the client in an
+        annotation create or update request
+    :type permissions: dict
+
+    :param groupid: the groupid of the annotation that the permissions dict
+        applies to
+    :type groupid: unicode
+
+    """
+    return permissions['read'] == ['group:{id}'.format(id=groupid)]
```
nickstenning: I think this means I can crash the server by POSTing a permissions object without a `read` key. Is that right?

nickstenning:
Other than what I think is a possible crashing bug, this LGTM.

nickstenning:
Aha! And it's not a crashing bug because the JSONSchema requires that a permissions object has a `read` field. Nice. Merging.

h/api/schemas.py
seanh: I don't think so, read is required: https://github.com/hypothesis/h/blob/master/h/api/schemas.py#L117

PR #3281, opened 2016-05-05
Views code for updating annotations in postgres

seanh:
This needs to be rebased and tested once <https://github.com/hypothesis/h/pull/3279> and <https://github.com/hypothesis/h/pull/3280> are merged

### DIFF
--- a/h/api/views.py
+++ b/h/api/views.py
@@ -235,7 +235,8 @@ def _update_postgres(annotation, request):
 
     annotation_dict = AnnotationJSONPresenter(request, annotation).asdict()
 
-    _publish_annotation_event(request, annotation_dict, 'update')
+    _publish_annotation_event(
+        request, annotation, 'update', annotation_dict=annotation_dict)
 
     return annotation_dict
 
@@ -251,7 +252,8 @@ def _update_elastic(annotation, request, notify):
 
     if notify:
         annotation_dict = AnnotationJSONPresenter(request, annotation).asdict()
-        _publish_annotation_event(request, annotation_dict, 'update')
+        _publish_annotation_event(
+            request, annotation, 'update', annotation_dict=annotation_dict)
         return annotation_dict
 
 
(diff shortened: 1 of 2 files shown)


seanh:
Note to self: this should try to save updates in Pg first then legacy Es, not the other way round (so that if saving to Pg fails, then nothing is saved to Es)

seanh:
This needs testing manually and I want to give it a careful look over, but I think it's probably mostly right

seanh:
Ok, this can be reviewed now, I'm finished working on it

h/api/views.py
```diff
@@ -208,17 +208,46 @@ def read_jsonld(annotation, request):
...
+        _update_elastic(legacy_annotation, request, notify=False)
+        return annotation_dict
+    return _update_elastic(annotation, request, notify=True)
+
+
+def _update_postgres(annotation, request):
+    schema = schemas.UpdateAnnotationSchema(request,
+                                            annotation.target_uri,
+                                            annotation.groupid)
+    appstruct = schema.validate(copy.deepcopy(_json_payload(request)))
```
nickstenning: What's the logic behind calling `copy.deepcopy` here?

1. `.validate()` should not (and AFAICT does not) modify its arguments
2. `_json_payload` [returns a new object every time it is called](https://github.com/Pylons/pyramid/blob/a8cad050d8d0e5095a3bf487132f8d6e315779ec/pyramid/request.py#L236).

nickstenning:
LGTM, other than the `copy.deepcopy` stuff, which looks unnecessary to me.

I do have one comment, which I feel I've made a couple of times before. Mixing functional and non-functional changes in the same pull request makes things unnecessarily hard to review. There would have been no harm (AFAICT) in pulling out stuff like PEP8ing some code and turning `mock_request` into a fixture into separate PRs, and it would greatly improve the readability of the diff for the stuff that actually matters. Reviewing commit by commit isn't a good "solution" here, because one ends up making comments on issues that are already fixed in subsequent commits.

seanh:
Done

nickstenning:
LGTM.

PR #3282, opened 2016-05-06
New threading 1/N - Begin to centralize UI state in annotationUI

robertknight:
This adds the initial infrastructure using [Redux](http://redux.js.org) for centralizing the UI
state as an immutable object accessible via `annotationUI.getState()`
which is updated as a result of actions from the UI, network etc.

For background on why we want to do this, see the design overview at
https://github.com/hypothesis/h/pull/3176 

Additionally this commit removes a couple of tests that checked for
non-mutation of the selected/focused annotation maps and uses
seamless-immutable instead which provides a better guarantee of this,
but only in debug builds.


h/static/scripts/annotation-ui.js
```diff
@@ -1,8 +1,11 @@
 'use strict';
 
+var Immutable = require('seamless-immutable');
+var redux = require('redux');
+
 function value(selection) {
   if (Object.keys(selection).length) {
-    return Object.freeze(selection);
+    return Immutable(selection);
```
robertknight: I'll look into silencing this

robertknight:

h/static/scripts/annotation-ui.js
```diff
@@ -1,8 +1,11 @@
 'use strict';
 
+var immutable = require('seamless-immutable');
+var redux = require('redux');
+
 function value(selection) {
```
nickstenning: This seems like an oddly named function to me. Perhaps calling this `freeze` would make its purpose clearer when used later in this file?

h/static/scripts/annotation-ui.js
```diff
@@ -16,59 +19,106 @@ function initialSelection(settings) {
...
+  });
+}
+
+var types = {
+  SELECT_ANNOTATIONS: 'SELECT_ANNOTATIONS',
+  FOCUS_ANNOTATIONS: 'FOCUS_ANNOTATIONS',
+  SET_HIGHLIGHTS_VISIBLE: 'SET_HIGHLIGHTS_VISIBLE',
+};
+
+function reducer(state, action) {
```
nickstenning: If you're not already familiar with redux, the purpose of this function is completely opaque.

A short paragraph at the top of the file explaining the purpose of the module, and the basic approach to its design (including explaining how redux provides "encapsulated state") would go a long way to making this code more transparent.

h/static/scripts/annotation-ui.js
```diff
@@ -86,18 +136,12 @@ module.exports = function (settings) {
...
     },
 
-    /**
-     * @ngdoc method
-     * @name annotationUI.xorSelectedAnnotations()
-     * @returns nothing
-     * @description takes an array of annotations and adds them to the
-     * selectedAnnotationMap if not present otherwise removes them.
-     */
+    /** Toggle whether annotations are selected or not. */
```
nickstenning: I think just one more line in this comment explaining that for each annotation in `annotations`, its selected state will be toggled, would be helpful.

In addition, although I appreciate you're trying to keep the size of the PRs down, `xorSelectedAnnotations` is a rather unfortunate name. It looks like replacing `xor` with `toggle` would be relatively straightforward and would IMO be a great improvement.

h/static/scripts/test/annotation-ui-controller-test.js
```diff
@@ -22,11 +24,15 @@ describe('AnnotationUIController', function () {
     $scope = $rootScope.$new();
     $scope.search = {};
 
-    annotationUI = {
-      tool: 'comment',
+    var store = createFakeStore({
```
nickstenning: Soooo... I'm looking at `annotation-ui.js`, and I'm having difficulty seeing why you couldn't use a real `annotationUI` object? What am I missing?

nickstenning:
Ok, this is looking great. My only real question is about the tests -- a number of tests here are using a `fakeAnnotationUI`, for which you've written this `create-fake-store` thing, but looking at `annotation-ui` I don't immediately see any reason to use a fake here. You can prepare an `annotation-ui` object as a stub quite easily now, it seems. I'd suggest that we do that, unless I'm missing something?

h/static/scripts/annotation-ui.js
robertknight: I was trying to keep the name consistent with the import name but I agree that `freeze()` would be a better name given how we're using it. The docs for 'seamless-immutable' use `Immutable` as the import name (as if it were a constructor) but it isn't really.

nickstenning: Just in case there was confusion here, I was talking about the function on line 6, currently called `value`.

robertknight:
> I don't immediately see any reason to use a fake here.
> You can prepare an annotation-ui object as a stub quite easily now, it seems.
> I'd suggest that we do that, unless I'm missing something?

I think it depends how purist you want to be about avoiding exercising logic from other units in unit tests.

Typically in a Redux app there are a few pieces which are unit tested as follows:

1. State update function (`(currentState, action) -> newState`). Tests consist of checking outputs for a given set of inputs.
2. Functions that create actions (`(parameters) -> action (+ side effects)`). Tests consist of checking outputs for a given input plus any side effects. These functions are often trivial so separate unit testing might not be worthwhile.
3. Store consumers (eg. `(state) -> DOM Tree + event handlers`). Tests consist of checking that output is correct for a given store state and that event handlers dispatch expected actions.

`createFakeStore()` provides a way to test (3) without (2) and (1). The tests just set the store to a given state and check that the consumers handle it as expected.

Right now, the implementations of (1) and (2) in annotationUI are pretty simple, so I agree there isn't much harm in using the real `annotationUI` service. If the state update / action creation logic starts to get more complex in future then I think you would want to avoid exercising that logic when testing other units.

nickstenning:
LGTM.

nickstenning:
Going to hold off on merging until we've merged #3282 and have some clarity about deployment to production.

PR #3284, opened 2016-05-06
New threading 3/N - Render annotation threads using new threading infrastructure

robertknight:
This replaces the rendering of conversation threads using
the new threading implementation.

 * Add `<annotation-thread>` component to render Thread objects
   generated by rootThread.

 * Expose the Thread generated by the `rootThread` service on
   the scope so that `<annotation-thread>` can render it.

 * Remove logic and tests in `<annotation>` for updating reply
   counts and thread states as these are no longer needed.


robertknight:

nickstenning:
Rebased.

nickstenning:
Most basic stuff seems to be working ok, but it seems to get very confused when using the search filtering:

For example, master:
 
screen shot 2016-05-13 at 12 09 58

This branch:

screen shot 2016-05-13 at 12 10 43

robertknight:
The behavior in master is not ideal but this branch should now match that.

The logic is that the "View N more in conversation" label is shown either if an annotation itself matches the filter or any of its replies (immediate or not) do.

nickstenning:
Okay, much improved! I'm afraid I've found another difference between this branch and master, and I think this is a regression we probably don't want to merge, too.

STR:

1. Create an annotation, and a reply to that annotation containing some keyword: say, "squiggle."
2. Reload the page.
3. Open the replies to the annotation.
4. Close the replies to the annotation.
5. Search in the sidebar for "squiggle".

In master, the "squiggle" reply is expanded and shown in full. On this branch you get this:

screen shot 2016-05-13 at 14 37 43

Steps 3/4 are important -- presumably because they override the "intended" expanded state of a thread to explicitly open or closed from a default state.

robertknight:
> Steps 3/4 are important -- presumably because they override the "intended" expanded state of a thread to explicitly open or closed from a default state.

That seems a likely explanation. The UI state kept by `annotationUI` has an `expanded` map which keeps track of threads that you have explicitly expanded or collapsed. If you haven't explicitly expanded or collapsed a thread then the state depends on whether it is an annotation (replies are collapsed by default) or reply (replies are expanded by default).

robertknight:
> Okay, much improved! I'm afraid I've found another difference between this branch and master, 
> and I think this is a regression we probably don't want to merge, too.

This is now resolved. The behavior now is that the expanded/collapsed states of threads are reset when the filter query is changed (see the `SET_FILTER_QUERY` action logic in `annotation-ui.js`)

This is slightly different than master, which retained explicitly set expanded/collapsed state but exhibited some inconsistencies - eg. clicking the expand/collapse toggle for annotations with replies matching the filter immediately after performing a search was a no-op in some cases.

nickstenning:
Okay, we're getting there...

Now I've noticed that the search result counts don't seem to be consistent with master. Specifically, the search count on master counts the number of matches (including replies), whereas the count on this branch seems to be counting the number of cards which have matches.

robertknight:
> Now I've noticed that the search result counts don't seem to be consistent with master.
> Specifically, the search count on master counts the number of matches (including replies),
> whereas the count on this branch seems to be counting the number of cards which have matches.

The behavior should now be consistent with master. The 'N search results' label displays the total number of visible annotations and replies, including any that have been explicitly shown by clicking 'View N more in conversation'

nickstenning:
:+1:

PR #3285, opened 2016-05-06
New threading 4/N - Virtualize the thread list

robertknight:
Make the sidebar handle "large" (ie. 50+) numbers of annotation threads
much better by virtualizing the thread list. Virtualizing means that we
only instantiate `<annotation-thread>` components for threads which are
either in or 'near' (above/below) the viewport. For the remaining
offscreen threads space is just reserved by a couple of `<li>` elements above
and below the visible threads.

This cuts out the cost of compiling and linking annotation thread directives for offscreen threads when loading annotations and also eliminates all of the watchers that those directives contribute to the `$digest` cycle.

- [x] Fix issue where unsaved text/tags are lost when annotation is scrolled out of view
- [x] Fix slight shift in position of annotation cards when an annotation above the viewport is destroyed and replaced with space in dummy `<li>` element.

### DIFF
--- a/h/static/scripts/test/stream-controller-test.coffee
+++ b/h/static/scripts/test/stream-controller-test.coffee
@@ -1,4 +1,11 @@
+EventEmitter = require('tiny-emitter')
+inherits = require('inherits')
+
 {module, inject} = angular.mock
+
+class FakeRootThread extends EventEmitter
+  constructor: () ->
+    this.thread = sinon.stub()
 
 describe 'StreamController', ->
   $controller = null
@@ -70,9 +77,7 @@ describe 'StreamController', ->
       getFilter: sandbox.stub()
     }
 
-    fakeRootThread = {
-      thread: sandbox.stub()
-    }
+    fakeRootThread = new FakeRootThread()
 
     $provide.value 'annotationMapper', fakeAnnotationMapper
     $provide.value 'annotationUI', fakeAnnotationUI
--- a/h/static/scripts/virtual-thread-list.js
+++ b/h/static/scripts/virtual-thread-list.js
@@ -82,6 +82,28 @@ VirtualThreadList.prototype.setThreadHeight = function (id, height) {
   this._heights[id] = height;
 };
 
+VirtualThreadList.prototype._height = function (id) {
+  // Default guess of the height required for a threads that have not been
+  // measured
+  var DEFAULT_HEIGHT = 200;
+  return this._heights[id] || DEFAULT_HEIGHT;
+};
+
+/** Return the vertical offset of an annotation card from the top of the list. */
+VirtualThreadList.prototype.yOffsetOf = function (id) {
+  var self = this;
+  var allThreads = this._rootThread.children;
+  var matchIndex = allThreads.findIndex(function (thread) {
+    return thread.id === id;
+  });
+  if (matchIndex === -1) {
+    return 0;
+  }
+  return allThreads.slice(0, matchIndex).reduce(function (offset, thread) {
+    return offset + self._height(thread.id);
+  }, 0);
+};
+
 /**
  * Recalculates the set of visible threads and estimates of the amount of space
  * required for offscreen threads above and below the viewport.
@@ -94,9 +116,6 @@ VirtualThreadList.prototype._updateVisibleThreads = function () {
   var MARGIN_ABOVE = 800;
   // Same as MARGIN_ABOVE but for the space below the viewport
   var MARGIN_BELOW = 800;
-  // Default guess of the height required for a threads that have not been
-  // measured
-  var DEFAULT_HEIGHT = 200;
 
   // Estimated height in pixels of annotation cards which are below the
   // viewport and not actually created. This is used to create an empty spacer
@@ -116,7 +135,7 @@ VirtualThreadList.prototype._updateVisibleThreads = function () {
 
   for (var i = 0; i < allThreads.length; i++) {
     thread = allThreads[i];
-    var threadHeight = this._heights[thread.id] || DEFAULT_HEIGHT;
+    var threadHeight = this._height(thread.id);
 
     if (usedHeight + threadHeight < this.window.pageYOffset - MARGIN_ABOVE) {
       // Thread is above viewport
--- a/h/static/scripts/virtual-thread-list.js
+++ b/h/static/scripts/virtual-thread-list.js
@@ -1,0 +1,134 @@
+'use strict';
+
+var EventEmitter = require('tiny-emitter');
+var debounce = require('lodash.debounce');
+var inherits = require('inherits');
+
+/**
+ * VirtualThreadList is a helper for virtualizing the annotation thread list.
+ *
+ * 'Virtualizing' the thread list hugely optimizes updates for the UI by only
+ * creating annotation cards for annotations which are either in or near the
+ * viewport.
+ *
+ * This technique is used in all native UI frameworks but is especially
+ * important as long as Angular is used for the view layer of the application
+ * because every active watcher/template expression contributes overhead to the
+ * $digest cycle and thus towards a feeling of lagginess in the UI.
+ *
+ * @param {Window} container - The Window displaying the list of annotation threads.
+ * @param {Thread} rootThread - The initial Thread object for the top-level
+ *        threads.
+ */
+function VirtualThreadList($scope, window_, rootThread) {
+  var self = this;
+
+  this._rootThread = rootThread;
+
+  // Cache of thread ID -> last-seen height
+  this._heights = {};
+
+  this.window = window_;
+
+  var debouncedUpdate = debounce(function () {
+    self._updateVisibleThreads();
+    $scope.$digest();
+  }, 20);
+  this.window.addEventListener('scroll', debouncedUpdate);
+  this.window.addEventListener('resize', debouncedUpdate);
+
+  this._detach = function () {
+    this.window.removeEventListener('scroll', debouncedUpdate);
+    this.window.removeEventListener('resize', debouncedUpdate);
+  };
+}
+inherits(VirtualThreadList, EventEmitter);
+
+/**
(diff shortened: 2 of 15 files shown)


robertknight:

sheetaluk:
I found a little bug where when the page is refreshed, clicking on the collapse/expand button for threaded replies the first time doesnt do anything. The second click onwards everything seems to work fine.

robertknight:
> I found a little bug where when the page is refreshed, clicking on the collapse/expand button
> for threaded replies the first time doesnt do anything.
> The second click onwards everything seems to work fine.

Thanks! I've fixed this.

robertknight:
Just spotted a fairly major bug with this branch. When annotations are scrolled far enough out of view, any component-local state is lost, this currently includes unsaved changes to annotation text and tags.

This state either needs to be put into the main app store or some ephemeral store that is external to the components.

robertknight:
> Just spotted a fairly major bug with this branch. When annotations are scrolled far enough out of view, any component-local state is lost, this currently includes unsaved changes to annotation text and tags.

Fixed this by using the existing 'drafts' system that we already have for persisting changes when switching groups or user account.

sheetaluk:
Looks like the side bar is showing the domain on which the annotation was made:
https://www.dropbox.com/s/emrznyqfi59i26j/Screenshot%202016-05-12%2006.26.23.png?dl=0

This should only appear in the stream.

robertknight:
> Looks like the side bar is showing the domain on which the annotation was made:
> https://www.dropbox.com/s/emrznyqfi59i26j/Screenshot%202016-05-12%2006.26.23.png?dl=0

Fixed

sheetaluk:
I see "View n more in conversation" under show/hide replies, for annotation-cards with replies, when I click on one of the .annotator-bucket-indicator elements. It doesnt look like master has that.
https://www.dropbox.com/s/aseaktpsz5rnkzk/Screenshot%202016-05-12%2007.38.58.png?dl=0

robertknight:
> I see "View n more in conversation" under show/hide replies, for annotation-cards with replies,
> when I click on one of the .annotator-bucket-indicator elements.

I'm sure this UI is going to change in the not-too-distant future, but fixed. Behavior should now be consistent with master.

robertknight:
> Fix slight shift in position of annotation cards when an annotation above the viewport is destroyed and replaced with space in dummy <li> element.

Ah, this is happening at least in part because the height of annotation cards as computed by the `getThreadHeight()` function in `widget-controller.js` does not include the bottom margin for each card created by the `.stream-list > *` selector.

nickstenning:
- [ ] Creating an annotation that falls outside the currently displayed annotation list doesn't currently autoscroll to the editor as it should.

h/static/scripts/virtual-thread-list.js
```diff
@@ -0,0 +1,142 @@
...
+ * that are expensive to create and consume a lot of memory. For Angular
+ * applications this also helps significantly with UI responsiveness by limiting
+ * the number of watchers (functions created by template expressions or
+ * '$scope.$watch' calls) that have to be run on every '$scope.$digest()' cycle.
+ *
+ * @param {Window} container - The Window displaying the list of annotation threads.
+ * @param {Thread} rootThread - The initial Thread object for the top-level
+ *        threads.
+ */
+function VirtualThreadList($scope, window_, rootThread) {
```
nickstenning: Should/could this not use `$window`?

h/static/scripts/virtual-thread-list.js
```diff
@@ -0,0 +1,142 @@
...
+/**
+ * Recalculates the set of visible threads and estimates of the amount of space
+ * required for offscreen threads above and below the viewport.
+ *
+ * Emits a `changed` event with the recalculated set of visible threads.
+ */
+VirtualThreadList.prototype._updateVisibleThreads = function () {
+  // Space above the viewport in pixels which should be considered 'on-screen'
+  // when calculating the set of visible threads
+  var MARGIN_ABOVE = 800;
```
nickstenning: I'd suggest moving these definitions to the top level of the file, so it's (even) clearer they're "constants".

h/static/scripts/widget-controller.js
```diff
@@ -32,8 +33,49 @@ function groupIDFromSelection(selection, results) {
...
+    // values of the form 'Npx', from which we extract 'N'.
+    var marginHeight = parseFloat(style.marginTop) +
+                       parseFloat(style.marginBottom);
+
+    return elementHeight + marginHeight;
+  }
+
+  var visibleThreads = new VirtualThreadList($scope, window, rootThread.thread());
+  visibleThreads.on('changed', function (state) {
+    $scope.virtualThreadList = {
```
nickstenning: This could be simpler if `VirtualThreadList` returned CSS height values `"300px"` rather than integers:

```
$scope.virtualThreadList = state
```

h/static/scripts/widget-controller.js
```diff
@@ -32,8 +33,49 @@ function groupIDFromSelection(selection, results) {
...
+      visibleThreads: state.visibleThreads,
+      offscreenUpperHeight: state.offscreenUpperHeight + 'px',
+      offscreenLowerHeight: state.offscreenLowerHeight + 'px',
+    };
+
+    scopeTimeout($scope, function () {
```
nickstenning: It's not clear to me what this code is doing...?

nickstenning: Ah, I see, it's the bit responsible for updating the cache... a one-line comment explaining wouldn't go amiss.

h/static/scripts/virtual-thread-list.js
```diff
@@ -0,0 +1,142 @@
...
+
+/**
+ * Sets the actual height for a thread.
+ *
+ * When calculating the amount of space required for offscreen threads,
+ * the actual or 'last-seen' height is used if known. Otherwise an estimate
+ * is used.
+ */
+VirtualThreadList.prototype.setThreadHeight = function (id, height) {
+  if (this._heights[id] === height) {
```
nickstenning: Is there any reason do do this?

h/static/scripts/widget-controller.js
```diff
@@ -32,8 +33,49 @@ function groupIDFromSelection(selection, results) {
...
+      state.visibleThreads.forEach(function (thread) {
+        visibleThreads.setThreadHeight(thread.id, getThreadHeight(thread.id));
```
nickstenning: `getThreadHeight` can return `undefined`, but `setThreadHeight` doesn't have a guard against storing `undefined` AFAICT.

h/static/scripts/virtual-thread-list.js
robertknight: I prefer to have constants scoped to just the function if they are only used in one place, plus we have a clear convention that ALL_CAPS vars are constants, which could be enforced by a tool.

h/static/scripts/virtual-thread-list.js
robertknight: It is pointless at the moment. I think originally I was going to trigger a recalculation of which threads are visible here but that turned out not to be necessary.

h/static/scripts/widget-controller.js
robertknight: The `height` argument to `setThreadHeight()` is actually optional, in that case we revert to using the guesstimate of the thread height. I have updated the `setThreadHeight()` documentation to clarify this.

h/static/scripts/virtual-thread-list.js
robertknight: `VirtualThreadList` is instantiated directly (both in `widget-controller.js` and in the tests) rather than via Angular's injector so we don't need to use the `$window` service here.

h/static/scripts/widget-controller.js
robertknight: I considered this but I thought it slightly nicer to decouple the calculation of the space that needs to be reserved from the mechanism used to actually reserve that space.

robertknight:
I'm seeing an issue on this branch where annotation cards are not appearing under IE 10. They are appearing under Edge.

- [ ] Fix annotation cards not appearing under IE 10

nickstenning:
I've unchecked [this issue](https://github.com/hypothesis/h/pull/3285#issuecomment-220965175) as it doesn't really appear to me to be fixed.

robertknight:
After some additional thought and testing I've reworked the way scrolling the sidebar to a given annotation works.

Scrolling to the Nth annotation is tricky given that the heights for some of the N-1 previous annotations may change from an estimate to an actual measured height as a side-effect of updating the view after scrolling.

0c27edf38e49f04ab1 describes a few approaches I could think of and the one I opted for in the end. The implemented approach is simple but has a downside that the view will sometimes scroll to an estimated position and then quickly adjust and scroll again if the estimate was wrong.

nickstenning:
Ok, I think this is a huge improvement and definitely good enough to merge now. There are a couple of aspects of editor state that we're not currently correctly restoring, which is confusing for a user:

- cursor position within a field
- cursor position between fields (i.e. am I editing the content or the tags?)
- edit/preview state (i.e. am I editing or previewing?)
- uncompleted tags (e.g. start typing a tag, scroll down to remind yourself what the tag is, scroll up, ... oh)
- textarea size (we set `resize: vertical;` on the textarea)

In future we probably want to polish these sharp corners.

nickstenning:
Also: share popup visibility isn't persisted, but that's less surprising, I think.

h/static/scripts/test/stream-controller-test.coffee
```diff
@@ -1,5 +1,12 @@
+EventEmitter = require('tiny-emitter')
+inherits = require('inherits')
```
nickstenning: Post-merge review: unused import.

h/static/scripts/widget-controller.js
nickstenning: Would be nice to see this comment...

Issue #3286, opened 2016-05-06
Add a test that annotations aren't created in Es if they failed to be created in Pg


seanh:
The view code for creating annotations in Postgres has already been merged. There should be tests that if the Postgres validation or storage fails, then no annotation is created in the legacy Elasticsearch storage either. And vice-versa.

PR #3288, opened 2016-05-06
Add some create annotation view tests

seanh:
Fixes https://github.com/hypothesis/h/issues/3286

### DIFF
--- a/h/api/test/views_test.py
+++ b/h/api/test/views_test.py
@@ -6,6 +6,7 @@ import pytest
 from pyramid import testing
 
 from h.api import views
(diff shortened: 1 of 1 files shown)


seanh:
I've tested manually that if the legacy Es validation or storage raises, nothing is committed to Pg (the view function does not catch the raised ValidationError, the db transaction is aborted) but I haven't written a test for this other than the "it raises if legacy validation raises" (the raise being what causes the db transaction to be aborted). I don't think it's very unit testable beyond that.

nickstenning:
LGTM.

PR #3290, opened 2016-05-06
Move EventQueue into h.api and sandbox event notifies

chdorner:
Currently we only use the `request.notify_after_commit()` in `h.api`, although I can see this being used in other circumstances as well, but to not add other h dependencies to `h.api` I'm moving the class into its own `h.api.eventqueue` module.

The second commit makes sure that each notifying each event will be sandboxed, we catch the exception and log it either to Sentry or the exception logger. It is worth noting that if one event has multiple subscribers, these will not be sandboxes, only different events.

### DIFF
--- a/h/api/__init__.py
+++ b/h/api/__init__.py
@@ -3,6 +3,7 @@
 
 def includeme(config):
     config.include('h.api.db')
+    config.include('h.api.eventqueue')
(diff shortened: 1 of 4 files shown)


nickstenning:
LGTM.

Issue #2829, opened 2016-01-02
Show / Hide Highlights endless toggle


fchasen:
When I create a sidebar in the host page and a guest embed in a iframe, the show / hide button toggles on and off very quickly in the sidebar. Inside the iframe the body, the class `annotator-highlight-always-on` is toggled on a off.

Example here: http://fchasen.com/hypothesis/index.html

[hypothesis-test.zip](https://github.com/hypothesis/h/files/76396/hypothesis-test.zip)

nickstenning:
Thanks a lot for filing this issue! We'll triage and take a look at it as soon as possible!

nickstenning:
This bug appears to be fixed in recent versions of Hypothesis. We appear to have introduced new bugs, like the fact that the toggle highlights buttons doesn't seem to correctly apply to subframes, but this one at least can close.

PR #3295, opened 2016-05-09
Call toggleHighlightClass on crossframe message

fchasen:
Calling `setVisibleHighlights` cause an infinite loop, as it would send the crossframe message again.

Not sure if `this.publish 'setVisibleHighlights', shouldShowHighlights` needs to be called in the guest when `toggleHighlightClass` is called.

Fixes: https://github.com/hypothesis/h/issues/2829

### DIFF
--- a/h/static/scripts/annotator/guest.coffee
+++ b/h/static/scripts/annotator/guest.coffee
@@ -46,7 +46,6 @@ module.exports = class Guest extends Annotator
 
   constructor: (element, options) ->
     super
-
     this.adderCtrl = new adder.Adder(@adder[0])
     this.anchors = []
 
@@ -118,7 +117,7 @@ module.exports = class Guest extends Annotator
       .catch((reason) -> cb(reason))
 
     crossframe.on 'setVisibleHighlights', (state) =>
-      this.setVisibleHighlights(state)
+      this.toggleHighlightClass(state)
 
   _setupWrapper: ->
     @wrapper = @element


nickstenning:
Hey @fchasen :)

Thanks for this. It's been quite a while since I went digging in this part of the code. Here's the full source of `Guest#setVisibleHighlights`, for easy reference:

```coffee
   # Pass true to show the highlights in the frame or false to disable.
   setVisibleHighlights: (shouldShowHighlights) ->
     @crossframe?.call('setVisibleHighlights', shouldShowHighlights)
     this.toggleHighlightClass(shouldShowHighlights)
     this.publish 'setVisibleHighlights', shouldShowHighlights
```

So I guess the question is who subscribes to `setVisibleHighlights`? ...

```
$ ag --coffee setVisibleHighlights
h/static/scripts/annotator/guest.coffee
120:    crossframe.on 'setVisibleHighlights', (state) =>
121:      this.setVisibleHighlights(state)
405:  setVisibleHighlights: (shouldShowHighlights) ->
406:    @crossframe?.call('setVisibleHighlights', shouldShowHighlights)
408:    this.publish 'setVisibleHighlights', shouldShowHighlights
430:        this.setVisibleHighlights true

h/static/scripts/annotator/host.coffee
37:      this.setVisibleHighlights(options.showHighlights)

h/static/scripts/annotator/plugin/toolbar.coffee
19:    'setVisibleHighlights': 'onSetVisibleHighlights'
52:          @annotator.setVisibleHighlights state
```

Hmm, that line in `toolbar.coffee` looks suspicious.

```coffee
module.exports = class Toolbar extends Annotator.Plugin
  HIDE_CLASS = 'annotator-hide'

  events:
    'setVisibleHighlights': 'onSetVisibleHighlights'
...
```

So it looks like it subscribes to that event, and that's what's responsible for toggling the state of the little "eye" icon.

Now in practice, this doesn't seem to cause any problems, because the only way of toggling highlights currently is by clicking that button, which calls `annotator.setVisibleHighlights` directly, so the event is always fired. But if something were to fire the `setVisibleHighlights` from elsewhere, we'd presumably want the event to be fired too so that the state of the icon is updated.

robertknight:
> Calling setVisibleHighlights cause an infinite loop, as it would send the crossframe message again.

Looking at `annotation-ui-sync.js` this loop should stop after one iteration because the sidebar only sends the crossframe message if the received setVisibleHighlights state is different from its current state:

```js
// in annotation-ui-sync.js
    setVisibleHighlights: function (state) {
      if (typeof state !== 'boolean') {
        state = true;
      }
      if (annotationUI.visibleHighlights !== state) {
        annotationUI.visibleHighlights = state;
        bridge.call('setVisibleHighlights', state);
      }
    }
```

So the sequence of events when you click on the button will be:

```
1. Sidebar Iframe <== ("setVisibleHighlights", true) <== Bucket bar (the toolbar containing the toggle highlights button)
2. Sidebar Iframe ==> ("setVisibleHighlights", true) ==> Guests
3. Sidebar Iframe <== ("setVisibleHighlights", true) <== Guests
```

So messages (2) and (3) are unnecessary when there is only a single guest connected but it doesn't explain the behavior you saw in EPUBjs when the sidebar toggle was flickering, which implies that `setVisibleHighlights` is being called with alternating true/false values.

Can you test whether this is what is actually happening in your case and if so, find out why the value is being toggled?

fchasen:
Great, I'll take a look at what `state` log as in that function.

There is a simple test case in this issue that may help debug: https://github.com/hypothesis/h/issues/2829

robertknight:
Hi @fchasen - It looks like this issue goes deeper than the surface-level fix here. See #3433 . I'm going to close this pull request for the moment but the test case in #2829 is helpful.

Issue #3293, opened 2016-05-08
Content-based detection of PDFs fails in Chrome 52 (Material Design PDF viewer)


robertknight:
### Steps to reproduce
1. Install Chrome 52
2. Visit http://resource-cms.springer.com/springer-cms/rest/v1/content/7368240/data/v1/Concatenated+PDF+12S1
3. Activate Hypothesis extension

### Expected behaviour
PDF should be opened in the bundled PDF.js viewer and Hypothesis sidebar should appear

### Actual behaviour
Hypothesis icon shows as active but PDF.js viewer does not load

### Browser/system information
Chrome 52 with material design-based PDF viewer. Note the references to the Polymer library in the PDF viewer's HTML content.

### Additional details
The Hypothesis plugin detects PDFs by looking for an `<embed>` tag using this selector: `embed[type="application/pdf"][name="plugin"]`. This does not match the tag in Chrome 52:

```html
<embed id="plugin" type="application/x-google-chrome-pdf" src="http://resource-cms.springer.com/springer-cms/rest/v1/content/7368240/data/v1/Concatenated+PDF+12S1" stream-url="blob:chrome-extension://mhjfbmdgcfjbbpaeojofohoefgiehjai/F29F38E5-18FB-4754-A8A0-9C50C18F83BE" headers="Accept-Ranges: bytes
Age: 17623
Cache-Control: max-age=31536000, public
Connection: keep-alive
Content-Encoding: gzip
Content-Language: en-US
Content-Length: 630703
Content-Type: application/pdf;charset=UTF-8
Date: Sun, 08 May 2016 16:09:00 GMT
ETag: 4a2b8844e6d340f081cfd87c68919f95
Vary: Accept-Encoding
Via: 1.1 varnish-v4
X-Cacheable: YES
X-Varnish: 146974730 145348381
" background-color="0xFF525659" top-toolbar-height="56" full-frame="">
```

dwhly:
How difficult is this to fix?  Important issue.

robertknight:
I'm mistaken. I'm not sure what I did yesterday that caused activating H on a PDF viewer to fail but was working when I tested again this morning and the description is mistaken.

The confusion was caused by the Chrome PDF viewer consisting of a native PDF viewer embedded in an HTML document providing the PDF viewer controls which in turn is instantiated by an `<embed>` tag in the top-level document. ie. Like this:

```html
<html>
<body>
  <embed name="plugin" type="application/pdf" ...>
[...]

PR #3296, opened 2016-05-09
Support PDFs in Firefox WebExtension and add clarifying comments about Chrome PDF detection

robertknight:
This adds support for detecting the PDF.js-based viewer in the Firefox
extension and loading the Hypothesis PDF viewer instead, as we do in the Chrome extension.

After getting thoroughly confused investigating an issue with the Chrome PDF viewer, I added some comments about how the Chrome viewer works as a reference for future. See https://github.com/hypothesis/h/issues/3293 for context.

### DIFF
--- a/h/browser/chrome/lib/detect-content-type.js
+++ b/h/browser/chrome/lib/detect-content-type.js
@@ -1,3 +1,5 @@
+'use strict';
+
 /**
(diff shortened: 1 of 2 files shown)


nickstenning:
LGTM.

PR #3298, opened 2016-05-10
Display hovered conversations with a grey background

robertknight:
Implement the changes to the background hover style from
https://trello.com/c/aXCXxzx2 and specifically this mock: https://trello-attachments.s3.amazonaws.com/571654d75b466478e53c0544/1373x901/1d1bfed40f039ddd84f41012a64c77e1/Annotation_card_refresh.png

### DIFF
--- a/h/static/styles/markdown.scss
+++ b/h/static/styles/markdown.scss
@@ -9,6 +9,7 @@
 }
 
 .markdown-tools {
+  background-color: $white;
   border-top: .1em solid #D3D3D3;
   border-left: .1em solid #D3D3D3;
   border-right: .1em solid #D3D3D3;
--- a/h/static/styles/thread.scss
+++ b/h/static/styles/thread.scss
@@ -5,7 +5,7 @@ $thread-padding: $annotation-card-left-padding;
     margin-bottom: .72em;
   }
 
-  & > li.thread > .threadexp {
+  & > .thread > .thread-collapse-toggle {
     display: none;
   }
 }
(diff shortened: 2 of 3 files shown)


nickstenning:
Looking pretty good, but this is a big change. Do we perhaps want to feature flag this?

Also, couple of niggles.

#### 1. Hovering over a collapsed reply shows "blocky outcrop"

I'm guessing the rectangular block under the collapse indicator shouldn't be there?

screen shot 2016-05-10 at 15 44 56

#### 2. Hovering over one reply highlights all reply threads

The mockup shows each reply and its children being independently hoverable, but that's not how this PR works:

screen shot 2016-05-10 at 15 47 04

----

Again, in the interest of not having to merge this all at once and get it perfect first time, I'd be in favour of adding a feature flag which can be used to add a CSS selector, which in turn can be used to test this design before turning it on for everyone.

nickstenning:
Hahaha! I didn't realise the overall design changes landed on master already, so I checked this out and I though this branch was responsible for everything.

robertknight:
Conor has just given me some feedback on this via Slack:

```
Colour is incorrect:
Grey-2 is wrong, it should be grey-1Threading:
The hover is working not as designed. If there are two different “threads” of conversation in the replies both have the background colour on hover, whereas only one of the “threads” should be highlighted on hover.Spacing:
The top and bottom of the hover state could be increased (if possible and easy).
```

robertknight:
> Do we perhaps want to feature flag this?

That was the original plan for all of the design changes but it would have been pretty fiddly in the end as we had a lot of little changes spread across different parts of the UI, including places where we can't currently access feature flags, so we went with splitting the whole design up into smaller separately-landable pieces instead.

robertknight:
I'm going to rework this on top of #3284

robertknight:
I'm in the process of reworking this branch since #3284 was merged. The styling for threading needs a cleanup so that will happen first.

PR #3299, opened 2016-05-10
Add custom tooltips for annotation card footer actions

robertknight:
* Add an 'h-tooltip' attribute directive which displays a tooltip above
   the associated element as soon as it is hovered and removes the
   tooltip on mouseout or when the associated element is destroyed.

 * Use the custom tooltip for buttons in the annotation card footer
   and remove the labels.


robertknight:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;w7tEeQiL&#x2F;341-custom-tooltips-for-buttons"> Custom tooltips for buttons</a>

nickstenning:
LGTM. And ❤️ 💚 💙  for ARIA labelling!

PR #3300, opened 2016-05-10
Remove extra margin set by first and last p in an annotation body.

sheetaluk:
We want to retain margins between paragraphs in the annotation-body, which includes threaded-replies.
So remove top margin and the bottom margin for just the first child and the last child. of an annotation-body.


robertknight:
LGTM

Issue #3291, opened 2016-05-06
Make a non-empty uri required when creating new annotations


seanh:
See <https://github.com/hypothesis/h/pull/3280#discussion_r62304061>

seanh:
Oops, looks like GitHub's issue auto closing had a false positive here

PR #3304, opened 2016-05-10
Do not test annotation schema

seanh:
I think this is going to make it easier to fix <https://github.com/hypothesis/h/issues/3291>, but also it just makes the test code a lot simpler and reduces test code duplication:

* Don't test `AnnotationSchema` directly, instead only test it indirectly via `CreateAnnotationSchema` and `UpdateAnnotationSchema`.

   `AnnotationSchema` isn't part of the public API of the schemas module, it's a base class for `CreateAnnotationSchema` and `UpdateAnnotationSchema`.

* Move the tests that were testing `AnnotationSchema` directly into a new
`TestCreateUpdateAnnotationSchema` test class which is parametrized to run
all of its tests against _both_ `CreateAnnotationSchema` and
`UpdateAnnotationSchema`. This allows us to test both of the child classes
rather than just testing the base class, and also allows us to avoid
duplicating all the test code.

* `CreateAnnotationSchema` and `UpdateAnnotationSchema` each still have their
own unit tests for the behaviors that apply to only one class or the
other, but those unit tests no longer mock `AnnotationSchema`, instead
calling the real `AnnotationSchema`. This makes these tests simpler as
they can just pass input data into `validate()` instead of having to mock
`AnnotationSchema.return_value.validate.return_value`.

   This will allow us to test future code changes where, for example,
`CreateAnnotationSchema` inherits from `AnnotationSchema` and modifies
`AnnotationSchema.schema` in a way that changes the behaviour of
`AnnotationSchema.validate()`. Since the `CreateAnnotationSchema` tests no
longer mock `AnnotationSchema`, they can test behavior changes to the real
`AnnotationSchema.validate()`.

* Remove a lot of test code duplication by moving duplicate tests from `TestCreateAnnotationSchema` and `TestUpdateAnnotationSchema` into `TestCreateUpdateAnnotationSchema`.

* I also spotted and removed some no longer necessary input data fixtures, and a couple of no longer necessary tests

### DIFF
--- a/h/api/test/schemas_test.py
+++ b/h/api/test/schemas_test.py
@@ -39,19 +39,34 @@ class TestJSONSchema(object):
         assert message.startswith("123 is not of type 'string'")
 
 
-class TestAnnotationSchema(object):
+def create_annotation_schema_validate(data):
+    schema = schemas.CreateAnnotationSchema(testing.DummyRequest())
+    return schema.validate(data)
 
-    def test_it_does_not_raise_for_minimal_valid_data(self):
-        schema = schemas.AnnotationSchema()
 
-        # Use only the required fields.
-        schema.validate(self.valid_input_data())
+def update_annotation_schema_validate(data,
+                                      existing_target_uri='',
+                                      groupid=''):
+    schema = schemas.UpdateAnnotationSchema(testing.DummyRequest(),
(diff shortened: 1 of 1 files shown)


nickstenning:
LGTM.

PR #3305, opened 2016-05-10
Update visual design of about dialog.

sheetaluk:
* Also remove duplicated variables in css.

### DIFF
--- a/h/static/styles/about-this-version-dialog.scss
+++ b/h/static/styles/about-this-version-dialog.scss
@@ -1,23 +1,49 @@
-.about-this-version__key {
+.about-this-version {
+  @include font-normal;
+  background: $grey-3;
+  margin-bottom: .72em;
+  padding: $layout-h-margin;
+  border-radius: 2px;
+}
+
+.about-this-version-title {
+  color: $grey-6;
+  font-weight: bold;
+  // Margin between top of the dialog and
+  // top of x-height of title should be ~15px.
+  margin-top: -5px;
+}
+
+.about-this-version-content {
+  // Margin between bottom of ascent of title and
+  // top of x-height of content should be 20px.
+  margin-top: 11px;
+}
+
+.about-this-version-content__key {
   width: 100px;
   float: left;
-  color: $gray-light;
+  color: $grey-4;
 }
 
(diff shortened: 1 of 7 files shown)


h/static/styles/about-this-version-dialog.scss
```diff
@@ -1,23 +1,49 @@
+.about-this-version-dialog {
```
nickstenning: By my understanding of the BEM pattern for CSS naming, shouldn't this top one be called `.about-this-version`, and then the various components be things like `.about-this-version__title`, `.about-this-version__content`?

Perhaps I have the wrong end of the stick. It wouldn't be the first time. CSS not my forte.

h/templates/client/about_this_version_dialog.html
```diff
@@ -1,37 +1,35 @@
...
-      <dd class="about-this-version__val">{{ vm.dateTime | date:'dd MMM yyyy HH:mm:ss Z' }}</dd>
-      <feedback-link
-        version="vm.version"
-        user-agent="vm.userAgent"
-        url="vm.url"
-        document-fingerprint="vm.documentFingerprint"
-        auth="vm.auth"
-        date-time="vm.dateTime">
-      </feedback-link>
+  <div class="about-this-version-title">
```
nickstenning: If this is a title, could we use an `<hX>` tag for it? Semantic markup is nice.

h/static/styles/about-this-version-dialog.scss
robertknight: Guidelines

The short answer is yes, though we're inconsistent about this elsewhere in the codebase at the moment.

The original BEM guideline was that a 'block' (the first part of the name before the `__`) is "something that is self contained" whereas an 'element' is "something that only makes sense within the context of its containing block". In that context it is sometimes not completely obvious what should be a 'block' vs. an 'element'

For our purposes, the important thing is to make it easy to find the styling that relates to a given component and be able to make changes without having unexpected effects elsewhere. In that context, having the name of the element in the markup, the block name in the styles and the name of the relevant sources files all match is helpful.

nickstenning:
LGTM.

PR #3302, opened 2016-05-10
Allow the creation of explicit cohorts for feature flags

finitud:
This PR is meant to keep track of the work done on the card:
https://trello.com/c/qn2cteEe/332-allow-the-creation-of-explicit-cohorts-for-feature-flags

The rough plan for this piece of work is as follows:
- [x] Add the `FeatureCohort` model in `h.features` (see PR #3309)
- [x] Database migration to create `featurecohort` and `featurecohort_user` tables (see PR #3317)
- [ ] Admin panel to create/delete cohorts (see PR #3344, #3330)
- [ ] Admin panel to edit cohorts (add/remove users)
- [ ] Link cohort list items to each respective edit cohort page.
- [ ] Add logic to `h.features.client` to handle feature cohorts (validation in `Client._state` method).

PR #3309, opened 2016-05-10
Feature flags cohorts: Add Cohort model

finitud:
This is the first of the pull requests referred to in #3302. It defines the `Cohort` model on `h.features.models`.

### DIFF
--- a/h/features/models.py
+++ b/h/features/models.py
@@ -116,9 +116,9 @@ class FeatureCohort(Base, mixins.Timestamps):
     name = sa.Column(sa.UnicodeText(), nullable=False, index=True)
 
     # Cohort membership
-    members = sa.orm.relationship(
-        'User', secondary='featurecohort_user', backref=sa.orm.backref(
-            'cohorts', order_by='FeatureCohort.name'))
+    members = sa.orm.relationship('User',
+                                  secondary='featurecohort_user',
+                                  backref='cohorts')
 
     def __init__(self, name):
         self.name = name
--- a/h/features/models.py
+++ b/h/features/models.py
@@ -3,10 +3,8 @@
 from __future__ import unicode_literals
 
 import logging
-import slugify
 
 import sqlalchemy as sa
-from sqlalchemy.orm import exc
 
 from h.db import Base
 from h.db import mixins
@@ -18,9 +16,6 @@ FEATURES = {
     'new_homepage': "Show the new homepage design?",
     'postgres': 'Read/write annotation and document data from/to postgres'
 }
-
-COHORT_NAME_MIN_LENGTH = 4
-COHORT_NAME_MAX_LENGTH = 25
 
 
 # Once a feature has been fully deployed, we remove the flag from the codebase.
@@ -125,34 +120,8 @@ class Cohort(Base, mixins.Timestamps):
         'User', secondary='user_cohort', backref=sa.orm.backref(
             'cohorts', order_by='Cohort.name'))
 
-    def __init__(self, name, creator):
+    def __init__(self, name):
         self.name = name
-        self.creator = creator
-        self.members.append(creator)
-
-    @sa.orm.validates('name')
-    def validate_name(self, key, name):
-        if not COHORT_NAME_MIN_LENGTH <= len(name) <= COHORT_NAME_MAX_LENGTH:
-            raise ValueError('name must be between {min} and {max} characters '
-                             'long'.format(min=COHORT_NAME_MIN_LENGTH,
-                                           max=COHORT_NAME_MAX_LENGTH))
-        return name
-
-    def __repr__(self):
-        return '<Cohort: %s>' % self.slug
-
-    @property
-    def slug(self):
-        """A version of this group's name suitable for use in a URL."""
-        return slugify.slugify(self.name)
-
-    @classmethod
-    def get_by_id(cls, id_):
-        """Return the cohort with the given id, or None."""
-        try:
-            return cls.query.filter(cls.id == id_).one()
-        except exc.NoResultFound:
-            return None
 
 
 USER_COHORT_TABLE = sa.Table(
--- a/h/features/models.py
+++ b/h/features/models.py
@@ -109,29 +109,30 @@ class Feature(Base):
         return '<Feature {f.name} everyone={f.everyone}>'.format(f=self)
 
 
-class Cohort(Base, mixins.Timestamps):
-    __tablename__ = 'cohort'
+class FeatureCohort(Base, mixins.Timestamps):
+    __tablename__ = 'featurecohort'
 
     id = sa.Column(sa.Integer, autoincrement=True, primary_key=True)
     name = sa.Column(sa.UnicodeText(), nullable=False, index=True)
 
     # Cohort membership
     members = sa.orm.relationship(
-        'User', secondary='user_cohort', backref=sa.orm.backref(
-            'cohorts', order_by='Cohort.name'))
+        'User', secondary='featurecohort_user', backref=sa.orm.backref(
+            'cohorts', order_by='FeatureCohort.name'))
 
     def __init__(self, name):
         self.name = name
 
 
 USER_COHORT_TABLE = sa.Table(
-    'user_cohort', Base.metadata,
+    'featurecohort_user', Base.metadata,
+    sa.Column('id', sa.Integer, nullable=False),
+    sa.Column('cohort_id',
+              sa.Integer,
+              sa.ForeignKey('featurecohort.id'),
+              nullable=False),
     sa.Column('user_id',
               sa.Integer,
               sa.ForeignKey('user.id'),
               nullable=False),
-    sa.Column('cohort_id',
-              sa.Integer,
-              sa.ForeignKey('cohort.id'),
-              nullable=False)
 )
--- a/h/features/models.py
+++ b/h/features/models.py
@@ -17,7 +17,6 @@ FEATURES = {
     'postgres': 'Read/write annotation and document data from/to postgres'
 }
 
-
 # Once a feature has been fully deployed, we remove the flag from the codebase.
 # We can't do this in one step, because removing it entirely will cause stage
 # to remove the flag data from the database on boot, which will in turn disable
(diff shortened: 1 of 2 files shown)


h/features/models.py
```diff
@@ -105,3 +112,57 @@ def remove_old_flags(cls):
...
+    def __repr__(self):
+        return '<Cohort: %s>' % self.slug
+
+    @property
+    def slug(self):
+        """A version of this group's name suitable for use in a URL."""
+        return slugify.slugify(self.name)
+
+    @classmethod
+    def get_by_id(cls, id_):
```
chdorner: You can just use SQLAlchemy's [get](http://docs.sqlalchemy.org/en/latest/orm/query.html#sqlalchemy.orm.query.Query.get) method which returns `None` when it can't be found.

And we also recently tried to pass in the database session to class methods on models to correctly scope these queries in the current transaction like [here](https://github.com/hypothesis/h/blob/master/h/api/models/document.py#L43-L55), which in this case would look like: `session.query(cls).get(id_)`

h/features/test/models_test.py
```diff
@@ -3,8 +3,11 @@
 import mock
 import pytest
 
+from h.groups import models
```
chdorner: This doesn't seem to be used anywhere in these tests?

h/features/models.py
```diff
@@ -105,3 +112,57 @@ def remove_old_flags(cls):
 
     def __repr__(self):
         return '<Feature {f.name} everyone={f.everyone}>'.format(f=self)
+
+
+class Cohort(Base, mixins.Timestamps):
```
nickstenning: So, thinking about this some more, I think this should be called `FeatureCohort`, the table should be called `featurecohort`, and the join table should be `featurecohort_user`.

The reason for this is that it's good for "ownership" of parts of the database to be clear. By ensuring all these tables start with `feature` it's clear they're owned by the `h.features` package.

It's worth saying that in general I'd shy away from names like `FeatureCohort` in Python because it "stutters" (e.g. `features.Cohort` is much nicer than `features.FeatureCohort`) but I think we maybe have to make an exception for models, which should match the table name. Or maybe not, maybe `Cohort` with a table name of `featurecohort` is fine? Thoughts? @chdorner? @seanh?

h/features/models.py
```diff
@@ -105,3 +112,57 @@ def remove_old_flags(cls):
...
+    name = sa.Column(sa.UnicodeText(), nullable=False, index=True)
+
+    # Cohort membership
+    members = sa.orm.relationship(
+        'User', secondary='user_cohort', backref=sa.orm.backref(
+            'cohorts', order_by='Cohort.name'))
+
+    def __init__(self, name, creator):
+        self.name = name
+        self.creator = creator
```
nickstenning: Cohorts don't need a creator. And indeed don't have one...

h/features/models.py
```diff
@@ -105,3 +112,57 @@ def remove_old_flags(cls):
...
+        self.members.append(creator)
+
+    @sa.orm.validates('name')
```
nickstenning: None of this is necessary IMO.

h/features/models.py
```diff
@@ -105,3 +112,57 @@ def remove_old_flags(cls):
...
+        if not COHORT_NAME_MIN_LENGTH <= len(name) <= COHORT_NAME_MAX_LENGTH:
+            raise ValueError('name must be between {min} and {max} characters '
+                             'long'.format(min=COHORT_NAME_MIN_LENGTH,
+                                           max=COHORT_NAME_MAX_LENGTH))
+        return name
+
+    def __repr__(self):
+        return '<Cohort: %s>' % self.slug
+
+    @property
```
nickstenning: Also not necessary.

h/features/models.py
nickstenning: :+1: I'd just remove this entirely.

h/features/models.py
```diff
@@ -105,3 +112,57 @@ def remove_old_flags(cls):
...
+    @classmethod
+    def get_by_id(cls, id_):
+        """Return the cohort with the given id, or None."""
+        try:
+            return cls.query.filter(cls.id == id_).one()
+        except exc.NoResultFound:
+            return None
+
+
+USER_COHORT_TABLE = sa.Table(
```
nickstenning: So in addition to the suggestion for naming this the other way around (see above), we need to not make the same mistake we made with the `user_group` table, namely:

- we should have a uniqueness constraint on `(cohort_id, user_id)`
- the table should have a PK column

The former because >1 rows for any user+cohort pair makes removal harder than it should be. The latter because fixing problems like that becomes harder when we don't have a PK to sort by.

h/features/models.py
chdorner: I don't mind the model class name not matching the table name, as long as it's easy to figure out which ones belong together. The one downside I can think off is that it makes it harder to rename the package as running a migration to rename a table in production without downtime is not very easy.

finitud: Just wondering whether `cohort` would be expressive enough as a table name, since we're not planning to overload the term for now? But otherwise I'd slightly prefer going for `featurecohort` to keep the model and the table names lined up.

h/features/models.py
finitud: Do you mean the name validation?

nickstenning: Yep.

h/features/models.py
finitud: Is this `USER_COHORT_TABLE` variable here necessary, now that I look at it again?

h/features/models.py
nickstenning: Okay, happy to keep this as `cohort` (and thus `Cohort`) for now, but please let's have `cohort_user`, not `user_cohort`.

h/features/models.py
nickstenning: It's not strictly necessary, but it's good practice to be able to reference the join table from outside this module, so I'd keep it.

finitud: 👍

h/features/models.py
```diff
@@ -105,3 +107,32 @@ def remove_old_flags(cls):
...
+    # Cohort membership
+    members = sa.orm.relationship(
+        'User', secondary='featurecohort_user', backref=sa.orm.backref(
+            'cohorts', order_by='FeatureCohort.name'))
+
+    def __init__(self, name):
+        self.name = name
+
+
+USER_COHORT_TABLE = sa.Table(
```
nickstenning: This should probably be called FEATURECOHORT_USER_TABLE for consistency now.

h/features/models.py
```diff
@@ -105,3 +107,32 @@ def remove_old_flags(cls):
...
+    'featurecohort_user', Base.metadata,
+    sa.Column('id', sa.Integer, nullable=False),
```
nickstenning: This probably needs to be explicitly configured as a primary key: `autoincrement=True, primary_key=True`

h/features/test/models_test.py
```diff
@@ -71,3 +73,35 @@ def features_pending_removal_override(self, request):
...
+        cohort = FeatureCohort(name=name)
+        db.Session.add(cohort)
+        db.Session.flush()
+
+        assert cohort.id
+        assert cohort.name == name
+        assert cohort.created
+        assert cohort.updated
+
+    def test_get_by_id_when_id_does_exist(self):
```
nickstenning: Honestly, this test and the next one are just testing sqlalchemy. We don't need to test that sqlalchemy works.

h/features/models.py
```diff
@@ -105,3 +107,32 @@ def remove_old_flags(cls):
...
+    sa.Column('cohort_id',
+              sa.Integer,
+              sa.ForeignKey('featurecohort.id'),
+              nullable=False),
+    sa.Column('user_id',
+              sa.Integer,
+              sa.ForeignKey('user.id'),
+              nullable=False),
```
nickstenning: Add a `sa.UniqueConstraint` here that guarantees only one row can exist for a given (user, cohort) pair.

h/features/test/models_test.py
```diff
@@ -71,3 +73,35 @@ def features_pending_removal_override(self, request):
         }, clear=True)
         patcher.start()
         request.addfinalizer(patcher.stop)
+
+
+class TestCohort(object):
```
nickstenning: If it's called FeatureCohort this should be called TestFeatureCohort.

h/features/test/models_test.py
```diff
@@ -3,8 +3,10 @@
 import mock
 import pytest
 
+from h.test import factories
```
nickstenning: Unused import.

h/features/models.py
```diff
@@ -105,3 +107,32 @@ def remove_old_flags(cls):
...
+class FeatureCohort(Base, mixins.Timestamps):
+    __tablename__ = 'featurecohort'
+
+    id = sa.Column(sa.Integer, autoincrement=True, primary_key=True)
+    name = sa.Column(sa.UnicodeText(), nullable=False, index=True)
+
+    # Cohort membership
+    members = sa.orm.relationship(
+        'User', secondary='featurecohort_user', backref=sa.orm.backref(
+            'cohorts', order_by='FeatureCohort.name'))
```
nickstenning: The `order_by` probably isn't necessary so this can just become `backref='cohorts'`

nickstenning:
I'm a bit confused, because I thought [we'd agreed to leave the names mostly as they were](https://github.com/hypothesis/h/pull/3309#discussion_r62828760), but I'm happy with this too, so don't change it back.

Couple more comments -- some of which are repeated from last time.

h/features/models.py
```diff
@@ -16,6 +17,7 @@
     'postgres': 'Read/write annotation and document data from/to postgres'
 }
 
+
```
nickstenning: Could we get rid of this?

h/features/test/models_test.py
finitud: Makes sense. This leaves no tests left for this model. Should I add any particular ones? Test uniqueness of (user, cohort) pairs?

nickstenning: Don't worry about it. It's a super simple model!

h/features/models.py
finitud: I had this one and the next one fixed in the migration, will open a PR for it now.

nickstenning: Sure, but you can just change the keyword arguments to `sa.Column` here, and then we can get this merged.

nickstenning: The model should provide a complete description of the table, and the migration describes how to get from where we are now to where we want to be. There will necessarily be duplication between them.

nickstenning:
LGTM.

PR #3310, opened 2016-05-11
Fix EBADF crash with ws4py and gevent 1.1

chdorner:
We now override ws4py's run_application due to the websocket server
crashing with EBADF. A change in gevent (1.1) causes all sockets to be
closed when a WSGI handler returns. ws4py starts a new greenlet for
each new websocket connection used to return from the WSGI handler.
The fix is taken from [here](https://github.com/Lawouach/WebSocket-for-Python/pull/180) and waits for the greenlet to finish before
returning from the WSGI handler.

More information at:
- https://github.com/Lawouach/WebSocket-for-Python/issues/170
- https://github.com/gevent/gevent/issues/633

### DIFF
--- a/h/websocket.py
+++ b/h/websocket.py
@@ -7,6 +7,7 @@ and a stripped-down Pyramid application which exposes a single endpoint for
 serving the "streamer" over the websocket.
 """
 
+import gevent
 from gunicorn.workers.ggevent import GeventPyWSGIWorker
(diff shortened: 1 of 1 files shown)


nickstenning:
Seems to fix any issues for me. :party:

seanh:
Works for me too (once I finally got the websocket configured properly)

PR #3313, opened 2016-05-11
Anchor annotations sequentially

### LINKED ISSUE #3278: Inner ranges not highlighted when anchoring nested annotations using quotes or text positions
## Update

_When anchoring two or more annotations using quote or position selectors, such that one annotation is partially or entirely contained within another, anchoring fails to wrap the inner annotation in a highlight span. As a result, clicking the inner text does not display the corresponding annotation in the sidebar. See https://github.com/hypothesis/h/issues/3278#issuecomment-217466106 for a minimal test case._

## Original Description

At http://jonudell.net/h/sitc-sulfate.html, consider two paragraphs. Para 1 begins "Under the anaerobic conditions..." and para 2 (immediately following) begins "Sulfate in sewage..."

Para 1 contains this annotation: http://jonudell.net/h/sitc-sulfate.html#annotations:AVPS1ONYH9ZO4OKSlgOL. That direct link does the expected thing (i.e. the card is selected in the sidebar, the doc scrolls to the highlight).

image

But there's a nested annotation that we cannot see (i.e., that does not appear with a darker highlight) and that does not sync with the sidebar when following the direct link:

image

http://jonudell.net/h/sitc-sulfate.html#annotations:AVPS1PxNH9ZO4OKSlgO3

```
                {
                    "type": "TextQuoteSelector",
                    "prefix": "fide in gravity sewer sections, ",
                    "exact": "sewage pumping stations",
                    "suffix": ", and inlet structures of wastew"
                }
```

Contrast this with para 2 where the sentence beginning "Coagulation plays an important role" has a nested annotation on "disinfection by-products".

A direct link to the outer annotation: 

http://jonudell.net/h/sitc-sulfate.html#annotations:AVPS1Nu2H9ZO4OKSlgN9

image

To the inner annotation:

http://jonudell.net/h/sitc-sulfate.html#annotations:AVPS1NUHH9ZO4OKSlgNx

image

Why the difference between para 1 and para 2?

robertknight:
When anchoring annotations which overlap, inserting the highlight
for an annotation could fail if the DOM was mutated in-between the
annotation's location being resolved to a DOM Range and that DOM Range
being highlighted.

Given two annotations, #1 and #2 where #2's range is contained within #1
(eg. #1 annotates a sentence and #2 annotates a phrase within that
sentence), the following could happen:

 1. Guest#anchor() (the function that anchors annotations) called with annotation #1
 2. Guest#anchor() called with annotation #2
 3. Annotation #1 resolved to range R1
 4. Annotation #2 resolved to range R2
 5. Range R1 highlighted. This wraps the first annotation in a `<span class="annotator-hl">` tag
    and makes R2 invalid in the process.
 6. Range R2 highlighting attempted. This fails because R2 is now invalid. Consequently annotation #2 appears in the sidebar but is not highlighted in the document.

This commit resolves the issue by anchoring annotations sequentially, so
the sequence would now be:

 1. Guest#anchor() called with annotation #1
 2. Annotation #1 resolved to range R1
 3. Range R1 highlighted
 4. Guest #anchor() called with annotation #2
 5. Annotation #2 resolved to range R2
 6. Range R2 highlighted

It might be possible to optimize the process by only anchoring
annotations which overlap sequentially. This commit takes the simplest
approach and just anchors all annotations sequentially.

Fixes #3278

### DIFF
--- a/h/static/scripts/annotator/guest.coffee
+++ b/h/static/scripts/annotator/guest.coffee
@@ -6,6 +6,7 @@ scrollIntoView = require('scroll-into-view')
 Annotator = require('annotator')
 $ = Annotator.$
 
+TaskQueue = require('./task-queue')
 adder = require('./adder')
 highlighter = require('./highlighter')
 rangeUtil = require('./range-util')
@@ -49,6 +50,9 @@ module.exports = class Guest extends Annotator
 
     this.adderCtrl = new adder.Adder(@adder[0])
     this.anchors = []
+
+    # Queue of annotation anchoring tasks
+    this._anchoringQueue = new TaskQueue()
 
     cfOptions =
       on: (event, handler) =>
@@ -98,8 +102,14 @@ module.exports = class Guest extends Annotator
       this.detach(annotation)
 
     this.subscribe 'annotationsLoaded', (annotations) =>
-      for annotation in annotations
-        this.anchor(annotation)
+      self = this
+      annotations.forEach((annotation) ->
+        # Annotations are anchored sequentially so that the DOM is not mutated
+        # in-between an annotation's location in the document being resolved to
+        # a DOM Range and the Range being highlighted by wrapping it in an
+        # 'annotator-hl' span.
+        self._anchoringQueue.push(-> self.anchor(annotation))
+      )
 
   _connectAnnotationUISync: (crossframe) ->
     crossframe.on 'focusAnnotations', (tags=[]) =>
--- a/h/static/scripts/annotator/task-queue.js
+++ b/h/static/scripts/annotator/task-queue.js
@@ -1,0 +1,43 @@
+'use strict';
+
+/**
+ * A queue which executes a series of Promise-yielding tasks sequentially.
+ *
+ * Call push() to enqueue a new task. All tasks are executed in serial.
+ *
+ * If we end up with more sophisticated needs, we might want to look at
+ * a package such as bottleneck.
+ */
+function TaskQueue() {
+  this._queue = [];
+  this._activeTask = null;
+}
+
+TaskQueue.prototype._runNext = function () {
+  var self = this;
+  if (this._activeTask || !this._queue.length) {
+    return;
+  }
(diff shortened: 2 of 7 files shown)


nickstenning:
Hmm. This seems to completely break highlighting for me. I made a bunch of highlights on the test page (/docs/help) some overlapping, some not. When I reload the page only one highlight is rendered. Nothing in the console.

robertknight:
I couldn't reproduce this just now but this branch _did_ have that exact issue at some point late yesterday before I fixed it. Can you check that you've got the latest tip of the branch checked out?

nickstenning:
You're quite right -- I was looking at 3545ac6.

Drawing highlights now seems pretty slow, though. Is that a tradeoff we're prepared to accept?

robertknight:
> Drawing highlights now seems pretty slow, though. Is that a tradeoff we're prepared to accept?

I was worried that might happen. We only need to anchor annotations that overlap sequentially. I think we should hold off merging this until that's been explored.

@judell  suggested an alternative approach to the SitC folks which was to sort annotations by quote length, which has the effect that for any range `C` contained by another set of ranges `S`, `C` will always be highlighted before ranges in `S`.

The logic might be something like:
1. Locate all annotations in document concurrently
2. Sort resulting ranges such that ranges are highlighted before any ranges that contain them
3. Attempt to highlight each range in the document. If the location has become invalidated by the time we come to highlight the range, then re-queue the annotation for another anchoring attempt.

robertknight:
Just had a discussion with @judell . The Science in the Classroom project has a solution that solves their problem so this is less urgent. I think it would make sense to merge the first commit in this branch - the anchoring integration tests and then revisit the fix separately.

robertknight:
I'm in the process of extracting out the tests from this into a separate `quote-anchor-integration-test` branch. I created a PR at https://github.com/hypothesis/h/pull/3351 . This PR URL isn't accessible at present. GitHub appears to be having issues.

PR #3314, opened 2016-05-11
Make uri required when creating new annotations

seanh:
The problem: we want to make a non-empty "uri" a required field when
creating new annotations, but _not_ when updating annotations.

I could easily do this by adding Python code to
CreateAnnotationSchema.validate(), but I would rather do things like
this that _can_ be done in JSONSchema in JSONSchema.

If I make "uri" required in the AnnotationSchema.schema class property,
because this is a class property and both CreateAnnotationSchema and
UpdateAnnotationSchema use an AnnotationSchema instance "uri" will now
be required both when creating and when updating annotations.

In general the JSONSchema validation that's done when creating and
updating cannot differ. Creating and updating validation can only differ
in the CreateAnnotationSchema.validate() and
UpdateAnnotationSchema.validate() methods in the code that runs after
AnnotationSchema.validate() is called.

The proposed solution:

1. Make AnnotationSchema.schema an instance attribute not a class
   attribute.

2. Make CreateAnnotationSchema add "uri" to "required" in the schema of
   its AnnotationSchema instance. This does not affect UpdateAnnotationSchema
   which has its own AnnotationSchema instance.

3. Because it uses the schema attribute, JSONSchema.validator now needs to be
   instantiated _after_ CreateAnnotationSchema has modified the schema.
   So move this from an instance attribute created in
   JSONSchema.__init__() (which happens before CreateAnnotationSchema
   modifies the schema) to an @property method which is not called until
   the validator is actually needed, when validate() is called.

   This means that an AnnotationSchema instance now creates a new
   validator each time its validate() is called, instead of creating
   just one and reusing it, but I think this is ok.

4. Make JSONSchema.schema an instance attribute as well, just to be
   consistent with AnnotationSchema.schema, there seems no point in
   mixing class and instance attributes, although I think we might get
   away with it.

### DIFF
--- a/h/api/schemas.py
+++ b/h/api/schemas.py
@@ -21,14 +21,16 @@ class JSONSchema(object):
     """
     Validate data according to a Draft 4 JSON Schema.
 
-    Inherit from this class and override the `schema` class property with a
-    valid JSON schema.
+    Inherit from this class and override the `schema` property with a valid
+    JSON schema.
     """
 
-    schema = {}
-
     def __init__(self):
-        self.validator = jsonschema.Draft4Validator(self.schema)
+        self.schema = {}
+
+    @property
+    def validator(self):
+        return jsonschema.Draft4Validator(self.schema)
 
     def validate(self, data):
         """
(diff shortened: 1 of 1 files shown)


seanh:
I'm not sure we really need `AnnotationSchema` or `JSONSchema` classes at all.

`JSONSchema`'s `validate()` method could become a `_validate()` helper function with `schema` as an additional argument.

Then we could just have a `_base_schema()` helper function, each of `CreateAnnotationSchema` and `UpdateAnnotationSchema` could just call `_base_schema()`, modify the schema as necessary, then pass it to `_validate(data_dict, schema_dict)`.

`CreateAnnotationSchema` and `UpdateAnnotationSchema` could probably also just be functions...

seanh:
However the change in this PR, changing `schema` from a class attribute into an instance attribute and changing `validator` from an instance attribute into a property, is a minimal change that allows the json schema to differ between create and update.

The reason I made `validator` an `@property` instead of just a local variable in `validate()`, by the way, is so that subclasses can still override it if they wanted to (no one is doing so but it was possible before so maybe it should remain possible now)

nickstenning:
Honestly, I think this approach is substantially more surprising and prone to error than just adding the two lines of python code to `CreateAnnotationSchema.validate()` for now.

PR #3316, opened 2016-05-12
Add Jenkinsfile to configure build pipeline

nickstenning:
This adds a Jenkinsfile, which is a Groovy script interpreted by the Jenkins Pipeline plugin in order to schedule our build.

Adding it to the source code repository means we can have Travis-like branch builds which we can configure to behave differently to master branch builds.

Here, we configure three build steps:

2. Build: builds the Docker image and tags it with the git revision
2. Test: runs the backend tests -- frontend tests will come in due course
3. Push: if we're on the master branch, we push the image (tagged as "auto") up to the Docker Hub


nickstenning:
Note also that because this branch contains a `Jenkinsfile`, Jenkins built it and has updated the commit status here.

robertknight:
LGTM

PR #3317, opened 2016-05-12
Add featurecohort table migration

finitud:
This is part of the work described in #3302.

### DIFF
--- a/h/migrations/versions/3bcd62dd7260_add_featurecohort_table.py
+++ b/h/migrations/versions/3bcd62dd7260_add_featurecohort_table.py
@@ -1,0 +1,38 @@
+"""add cohort table
+
+Revision ID: 3bcd62dd7260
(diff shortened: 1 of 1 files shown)


h/migrations/versions/3bcd62dd7260_add_featurecohort_table.py
```diff
@@ -0,0 +1,38 @@
...
+down_revision = 'dfa82518915a'
+
+from alembic import op
+import sqlalchemy as sa
+
+
+def upgrade():
+    op.create_table('featurecohort',
+        sa.Column('id', sa.Integer(), nullable=False),
+        sa.Column('name', sa.Unicode(length=100), nullable=False),
```
nickstenning: This isn't consistent with the model. The model defines the field as `UnicodeText`, not `Unicode`, and it doesn't specify a length.

PR #3318, opened 2016-05-12
Fix "Set changed size ..." error in websocket server

nickstenning:
This commit imports and modifies a few more pieces from the buggy ws4py WSGI
integration. The problem here specifically is in the GEventWebSocketPool's
"clear" method, which iterates over a set of greenlets that isn't guaranteed to
remain of constant size during iteration (clients may join, or clients may
close their connections).

So we replace a

    for greenlet in self:

with a

    for greenlet in list(self):

and we're done.

But by that point we've modified basically everything in
`ws4py.server.geventserver` and it doesn't make sense any more to keep
monkeypatching, so I've copied and pasted the half-dozen remaining lines of code into our codebase.


chdorner:
As we already talked about, would be nice to have some tests for all this ws4py code that we're patching/copying. But that can happen separately.

LGTM.

PR #3319, opened 2016-05-12
Add a `hypothesis migrate` command

nickstenning:
This command wraps Alembic, so that you don't need to remember to
specify the config file for every alembic command. Instead of running

    alembic -c conf/alembic.ini upgrade head

you can now run

    hypothesis migrate upgrade head


chdorner:
Lovely! I always forget that alembic has it's own config file and not `conf/{development-}app.ini`. Now I can safely forget about it.
LGTM.

PR #3320, opened 2016-05-12
Prune old and crufty migrations

nickstenning:
There's no point keeping these around for ever. This commit removes:

- all but the latest 10 migrations
- a standalone annotation migration script from before the ark
- a deeply unhelpful README


robertknight:
LGTM

PR #3323, opened 2016-05-12
Postgres to ElasticSearch reindexer

chdorner:
This PR adds:

* A class for batch indexing all (or a big set of) annotations: `h.api.search.index.BatchIndexing`
* A class for batch removing deleted annotations from the search index: `h.api.search.index.BatchDeleting`
* A celery task for an easy way to run both at the same time in a worker: `hypothesis-celery call h.indexer.reindex`

I've added more detailed comments in the code, but all SQL queries, ElasticSearch queries and operations have been optimized as far as possible for causing the least amount of load on the respective Postgres/ElasticSearch servers, and for keeping the memory footprint of the Python process running these as low as possible.

I've tested it out locally and running both the indexing and deleting takes between 5 and 5:30 minutes on my local machine with approximately 350,000 annotations.

### DIFF
--- a/h/api/models/annotation.py
+++ b/h/api/models/annotation.py
@@ -79,11 +79,10 @@ class Annotation(Base, mixins.Timestamps):
     #: Any additional serialisable data provided by the client.
     extra = sa.Column(pg.JSONB, nullable=True)
 
-    document_uris = sa.orm.relationship('DocumentURI',
-                                        viewonly=True,
-                                        foreign_keys='DocumentURI.uri_normalized',
-                                        primaryjoin='DocumentURI.uri_normalized == Annotation.target_uri_normalized')
-    documents = association_proxy('document_uris', 'document')
+    document = sa.orm.relationship('Document',
+                                   secondary='join(DocumentURI, Document, DocumentURI.document_id == Document.id)',
+                                   primaryjoin='Annotation.target_uri_normalized == DocumentURI.uri_normalized',
+                                   uselist=False)
 
     @hybrid_property
     def target_uri(self):
@@ -97,11 +96,6 @@ class Annotation(Base, mixins.Timestamps):
     @hybrid_property
     def target_uri_normalized(self):
         return self._target_uri_normalized
-
-    @hybrid_property
-    def document(self):
-        if self.documents:
-            return self.documents[0]
 
     @property
     def parent_id(self):
--- a/h/api/models/annotation.py
+++ b/h/api/models/annotation.py
@@ -28,6 +28,7 @@ class Annotation(Base, mixins.Timestamps):
         #   http://www.postgresql.org/docs/9.5/static/gin-intro.html
         #
         sa.Index('ix__annotation_tags', 'tags', postgresql_using='gin'),
+        sa.Index('ix__annotation_updated', 'updated'),
     )
 
     #: Annotation ID: these are stored as UUIDs in the database, and mapped
--- a/h/api/models/document.py
+++ b/h/api/models/document.py
@@ -26,8 +26,12 @@ class Document(Base, mixins.Timestamps):
     #        dependency on the annotator-store is removed, as it clashes with
     #        making the Postgres and Elasticsearch interface of a Document
     #        object behave the same way.
-    document_uris = sa.orm.relationship('DocumentURI', backref='document')
-    meta = sa.orm.relationship('DocumentMeta', backref='document')
+    document_uris = sa.orm.relationship('DocumentURI',
+                                        backref='document',
+                                        order_by='DocumentURI.updated.desc()')
+    meta = sa.orm.relationship('DocumentMeta',
+                               backref='document',
+                               order_by='DocumentMeta.updated.desc()')
 
     def __repr__(self):
         return '<Document %s>' % self.id
--- a/h/api/models/document.py
+++ b/h/api/models/document.py
@@ -93,6 +93,7 @@ class DocumentURI(Base, mixins.Timestamps):
                             'uri_normalized',
                             'type',
                             'content_type'),
+        sa.Index('ix__document_uri_updated', 'updated'),
     )
 
     id = sa.Column(sa.Integer, autoincrement=True, primary_key=True)
@@ -153,6 +154,7 @@ class DocumentMeta(Base, mixins.Timestamps):
     __tablename__ = 'document_meta'
     __table_args__ = (
         sa.UniqueConstraint('claimant_normalized', 'type'),
+        sa.Index('ix__document_meta_updated', 'updated'),
     )
 
     id = sa.Column(sa.Integer, autoincrement=True, primary_key=True)
--- a/h/api/models/document.py
+++ b/h/api/models/document.py
@@ -93,6 +93,7 @@ class DocumentURI(Base, mixins.Timestamps):
                             'uri_normalized',
                             'type',
                             'content_type'),
+        sa.Index('ix__document_uri_document_id', 'document_id'),
         sa.Index('ix__document_uri_updated', 'updated'),
     )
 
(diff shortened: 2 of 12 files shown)


h/api/models/annotation.py
```diff
@@ -79,11 +80,10 @@ class Annotation(Base, mixins.Timestamps):
...
 
-    document_uris = sa.orm.relationship('DocumentURI',
-                                        viewonly=True,
-                                        foreign_keys='DocumentURI.uri_normalized',
-                                        primaryjoin='DocumentURI.uri_normalized == Annotation.target_uri_normalized')
-    documents = association_proxy('document_uris', 'document')
+    document = sa.orm.relationship('Document',
+                                   secondary='join(DocumentURI, Document, DocumentURI.document_id == Document.id)',
+                                   primaryjoin='Annotation.target_uri_normalized == DocumentURI.uri_normalized',
+                                   uselist=False)
```
nickstenning: Much nicer 👍

h/api/models/document.py
```diff
@@ -26,8 +26,12 @@ class Document(Base, mixins.Timestamps):
     #        dependency on the annotator-store is removed, as it clashes with
     #        making the Postgres and Elasticsearch interface of a Document
     #        object behave the same way.
-    document_uris = sa.orm.relationship('DocumentURI', backref='document')
-    meta = sa.orm.relationship('DocumentMeta', backref='document')
+    document_uris = sa.orm.relationship('DocumentURI',
+                                        backref='document',
+                                        order_by='DocumentURI.updated.desc()')
```
nickstenning: If we're sorting on these do we want an index on them, too?

h/api/models/document.py
```diff
@@ -26,8 +26,12 @@ class Document(Base, mixins.Timestamps):
...
+    meta = sa.orm.relationship('DocumentMeta',
+                               backref='document',
+                               order_by='DocumentMeta.updated.desc()')
```
nickstenning: Ditto.

h/api/search/index.py
```diff
@@ -69,3 +73,183 @@ def delete(es, annotation_id):
...
+                                             chunk_size=100,
+                                             raise_on_error=False,
+                                             expand_action_callback=self._present)
+        errored = set()
+        for ok, item in indexing:
+            if not ok:
+                errored.add(item['index']['_id'])
+        return errored
+
+    def _present(self, annotation):
```
nickstenning: I'd suggest calling this "prepare" or something a bit more generic -- I'm worried that we're falling slowly into the trap of assuming that we'll always use a presenter to turn an annotation object into a thing for the search index, and that definitely doesn't have to be true!

h/api/search/index.py
```diff
@@ -69,3 +73,183 @@ def delete(es, annotation_id):
...
+        action = {'index': {'_index': self.es_client.index,
+                            '_type': self.es_client.t.annotation,
+                            '_id': annotation.id}}
+        data = presenters.AnnotationJSONPresenter(self.request, annotation).asdict()
```
nickstenning: For similar reasons, perhaps add a FIXME/TODO here to note that the `AnnotationJSONPresenter` is by no means the right way to prepare a document for indexing from an annotation, but it serves in the interim to replicate the structure of the old index.

h/api/search/index.py
```diff
@@ -69,3 +73,183 @@ def delete(es, annotation_id):
...
+        # This is using a windowed query for loading all annotations in batches.
+        # It is the most performant way of loading a big set of records from
+        # the database while still supporting eagerloading of associated
+        # document data.
+
+        updated = self.session.query(models.Annotation.updated). \
+                execution_options(stream_results=True). \
+                order_by(models.Annotation.updated.asc()).all()
+
+        windows = [updated[x:x+chunksize] for x in range(0, len(updated), chunksize)]
```
nickstenning: It doesn't look to me like there's a reason you'd need to store all of the intermediate values here, is there? Couldn't you store something like:

```python
windows = [(updated[x], updated[x+chunksize-1]) for x in range(0, len(updated), chunksize)]
```

If you were feeling really nice, I'd probably do something like:

```python
class Window(namedtuple('Window', ['start', 'end'])):
    pass
```

so you could then do

```python
windows = [Window(start=updated[x], end=updated[x+chunksize-1]) 
           for x in range(0, len(updated), chunksize)]
...

for window in windows:
    batch = basequery.filter(models.Annotation.updated.between(window.start, window.end)).all()
    ...
```

nickstenning: Probably also want to use `xrange` here given the size of `len(updated)` -- you'll need to add something to `_compat` for this.

h/api/search/index.py
```diff
@@ -69,3 +73,183 @@ def delete(es, annotation_id):
...
+            filter(models.Annotation.id.in_(annotation_ids)).all()
+
+        for a in annotations:
+            yield a
+
+    def _eager_loaded_query(self):
+        return self.session.query(models.Annotation).options(
+            subqueryload(models.Annotation.document).subqueryload(models.Document.document_uris),
+            subqueryload(models.Annotation.document).subqueryload(models.Document.meta)
+        )
```
nickstenning: SQLAlchemy ^*&@ yeah! 🤘

h/api/search/index.py
```diff
@@ -69,3 +73,183 @@ def delete(es, annotation_id):
...
+        return errored
+
+    def _present(self, id_):
+        action = {'delete': {'_index': self.es_client.index,
+                             '_type': self.es_client.t.annotation,
+                             '_id': id_}}
+        return (action, None)
+
+    def _pg_ids(self):
+        # This is using a Postgres cursor for better query performance.
```
nickstenning: If this works for now, that's fine, but in general we can't keep building a set of all the ids in the database. Presumably we could return these in batches and do a `difference_update` for each batch?

h/api/search/test/index_test.py
```diff
@@ -101,6 +112,245 @@ def log(self, patch):
...
+
+    @pytest.fixture
+    def index(self, patch):
+        return patch('h.api.search.index.BatchIndexer.index')
+
+    @pytest.fixture
+    def streaming_bulk(self, patch):
+        return patch('h.api.search.index.es_helpers.streaming_bulk')
+
+    @pytest.fixture
```
nickstenning: Doesn't look like you're using it as a fixture any more so this should probably go.

h/indexer.py
```diff
@@ -1,17 +1,26 @@
...
 
 __all__ = (
     'add_annotation',
     'delete_annotation',
+    'reindex',
 )
 
 
+log = logging.getLogger(__name__)
+log.setLevel(logging.DEBUG)
```
nickstenning: This should probably go?

nickstenning:
Looking pretty good to me except for the few issues I've raised.

h/api/models/document.py
chdorner: Added an index to both DocumentMeta/URI `updated` column.

When checking the query with `EXPLAIN ANALYZE ...` I saw that it isn't using an index for filtering the `document_id` column. Turns out Postgres is [not automatically adding indices for foreign keys](https://hyp.is/AVSqPuhUFtPwhMO7Dnnm/www.postgresql.org/docs/9.4/static/ddl-constraints.html). There is a migration for that as well now.

h/api/models/document.py
chdorner: Ditto from above.

h/api/search/index.py
chdorner: Good point. Fixed.

h/api/search/index.py
chdorner: Done.

h/api/search/index.py
chdorner: Done, done, and done.

h/api/search/test/index_test.py
chdorner: Done

h/indexer.py
chdorner: It's gone.

h/api/search/index.py
chdorner: It is now returning a SQLAlchemy query and we use the `batch_iter` function in `deleted_annotation_ids`, very similar to the way we load ElasticSearch annotation ids. :+1:

chdorner:
I've addressed all comments. It is now even more query / memory optimized!

nickstenning:
LGTM.

PR #3326, opened 2016-05-12
Show appropriate messages if an annotation is unavailable to a user.

sheetaluk:
If a selected annotation is unavailable to a user, prompt user to sign in
if he/she isn't signed in.
If a selected annotation is unavailable to a user and he/she is already
signed in, display a message.

Trello:
https://trello.com/c/scNGUMBk/330-improve-copy-to-clarify-direct-linking-functionality

### DIFF
--- a/h/static/styles/thread.scss
+++ b/h/static/styles/thread.scss
@@ -27,7 +27,6 @@ $thread-padding: $annotation-card-left-padding;
   align-items: center;
 
   &__label {
-    max-width: 160px;
     text-align: center;
   }
 
--- a/h/templates/client/viewer.html
+++ b/h/templates/client/viewer.html
@@ -27,7 +27,16 @@
       ng-if="selectedAnnotationUnavailable()">
     <div class="annotation-unavailable-message__icon"></div>
     <p class="annotation-unavailable-message__label">
-      You do not have permission to see this annotation
+      <span ng-if="auth.status === 'signed-out'">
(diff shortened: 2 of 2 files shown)


robertknight:
The functionality works fine but the text wrapping looks odd:

3326-message-wrapping

I'd suggest that either each sentence should appear on its own line or the whole message should appear on one line.

h/static/styles/thread.scss
```diff
@@ -20,7 +20,7 @@ $thread-padding: $annotation-card-left-padding;
   align-items: center;
 
   &__label {
-    max-width: 160px;
+    max-width: 180px;
```
robertknight: You can't tell from looking at this what the intended visual effect is, that a line break should appear after the end of the sentence. Any reason not to just use `<br>` tags in the template?

finitud:
May I suggest to use singular "they" instead of "he/she" for better inclusiveness?

robertknight:
> May I suggest to use singular "they" instead of "he/she" for better inclusiveness?

I presume you're referring to the commit message @finitud ?

sheetaluk:
1. Changed the commit message.
2. Removed styling for width and added <br>

Here is how it looks:
https://www.dropbox.com/s/dgg1hl61ku3hcrr/Screenshot%202016-05-25%2011.54.18.png?dl=0

nickstenning:
LGTM.

PR #3328, opened 2016-05-13
Fix database/index autoconfiguration in dev environment

nickstenning:
Database/search index autocreation was broken for `hypothesis devserver` due to the MODEL_CREATE_ALL environment variable overrides in `h.cli.bootstrap`. This commit fixes this problem by:

- adding an environment variable, SEARCH_AUTOCONFIG, to correspond to the h.search.autoconfig setting
- setting SEARCH_AUTOCONFIG and MODEL_CREATE_ALL explicitly in the `initdb` command and for the web process run by `devserver`
- removing the affected settings from development-app.ini to prevent unintended use of the autocreation feature

### DIFF
--- a/h/cli/__init__.py
+++ b/h/cli/__init__.py
@@ -24,7 +24,7 @@ SUBCOMMANDS = (
 )
 
 
-def bootstrap(app_url, dev=False, create_db=False):
+def bootstrap(app_url, dev=False):
     """
     Bootstrap the application from the given arguments.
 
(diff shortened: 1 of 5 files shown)


h/cli/__init__.py
```diff
@@ -37,13 +37,6 @@ def bootstrap(app_url, dev=False, create_db=False):
     # up an entire application to build the extensions.
     os.environ['H_SCRIPT'] = 'true'
 
-    # Override other important environment variables
-    os.environ['MODEL_CREATE_ALL'] = 'True' if create_db else 'False'
-    os.environ['MODEL_DROP_ALL'] = 'False'
-
-    if dev:
-        os.environ['SECRET_KEY'] = 'notsecret'
```
nickstenning: This is no longer needed as all these commands now bootstrap against `development-app.ini` when appropriate, which already defines a dummy secret key for development.

chdorner:
LGTM.

PR #3329, opened 2016-05-13
Remove H_ prefix from websocket URL setting

nickstenning:
This is well-intentioned, but inconsistent with all other settings.


robertknight:
LGTM

chdorner:
And here's the corresponding change in the playbook repo: https://github.com/hypothesis/playbook/commit/6b33692ac4a8fb171cb8af1ade680748117fc68e

PR #3330, opened 2016-05-13
Feature flags cohorts: Add Feature Cohorts admin panel

finitud:
This pull request adds an admin panel where an administrator can see a list of existing feature cohorts and create new ones.

See #3302 for further description.


finitud:
From Slack: 
> let's make the "cohorts" admin pages part of the features admin.
> so `/admin/features/cohorts` and probably use some subnav within those pages rather than a new top-level admin page"

h/admin/views/cohorts.py
```diff
@@ -0,0 +1,35 @@
...
+from h import paginator
+
+
+@view_config(route_name='admin_cohorts',
+             request_method='GET',
+             renderer='h:templates/admin/cohorts.html.jinja2',
+             permission='admin_cohorts')
+@paginator.paginate
+def cohorts_index(context, request):
+    return models.FeatureCohort.query.order_by(models.FeatureCohort.created.desc())
```
chdorner: This might not be too important for the admin views, but in general it's a good idea to scope all model queries through the same transaction as the request is in with:
```python
return request.db.query(models.FeatureCohort).order_by(models.FeatureCohort.created.desc())
```

h/admin/views/cohorts.py
```diff
@@ -0,0 +1,35 @@
...
+             request_method='POST',
+             request_param='add',
+             renderer='h:templates/admin/cohorts.html.jinja2',
+             permission='admin_cohorts')
+def cohorts_add(request):
+    """Create a new feature cohort."""
+    cohort_name = request.params['add']
+    cohort = models.FeatureCohort(name=cohort_name)
+    request.db.add(cohort)
+    request.db.flush()
```
chdorner: There's no need to flush manually, SQLAlchemy does that automatically when the transaction gets committed.

h/admin/views/__init__.py
```diff
@@ -6,6 +6,7 @@ def includeme(config):
     config.include('.admins')
     config.include('.badge')
     config.include('.features')
+    config.include('.cohorts')
```
nickstenning: Personally I'd be inclined to have these views in the `features.py` module, as they're really part of the same logical chunk of functionality, but this isn't a killer either way.

h/admin/views/cohorts.py
```diff
@@ -0,0 +1,35 @@
...
+from pyramid import httpexceptions as exc
+
+from h import models
+from h import paginator
+
+
+@view_config(route_name='admin_cohorts',
+             request_method='GET',
+             renderer='h:templates/admin/cohorts.html.jinja2',
+             permission='admin_cohorts')
```
nickstenning: The permission here should probably just be `admin_feature`, not `admin_cohorts` -- you can either admin features or you can't -- there's no need to be more fine-grained than that.

h/admin/views/cohorts.py
nickstenning: It's probably also more intuitive to order feature cohorts by their name rather than their created date.

h/admin/views/cohorts.py
nickstenning: :+1:

h/templates/admin/cohorts.html.jinja2
```diff
@@ -0,0 +1,49 @@
...
+
+{% set page_id = 'cohorts' %}
+{% set page_title = 'Feature Cohorts' %}
+
+{% block content %}
+  <p>
+    On this page you can see a list of all the feature cohorts.
+  </p>
+  <p>
+    Back to <a href="{{ request.route_url('admin_features') }}">Feature flags</a>
```
nickstenning: It would be nice if we could extend our use of the bootstrap navigation bar rather than adding some ad-hoc links to the page content like this. See https://getbootstrap.com/components/#navbar for some examples -- and perhaps @robertknight for some help if you're not sure what to do.

h/admin/views/features.py
```diff
@@ -33,5 +34,30 @@ def features_save(request):
...
+
+
+@view_config(route_name='admin_cohorts',
+             request_method='POST',
+             request_param='add',
+             renderer='h:templates/admin/cohorts.html.jinja2',
+             permission='admin_features')
+def cohorts_add(request):
+    """Create a new feature cohort."""
+    cohort_name = request.params['add']
```
chdorner: Might be good to not allow creating cohorts with an empty name. I'll merge this anyway as it's not the end of the world :)

chdorner:
LGTM.

PR #3332, opened 2016-05-16
Use distinct database session for h.api tests

nickstenning:
The h.api tests formerly used the default scoped session in h.db when testing. This commit gives these tests their own independently managed database session, thus allowing them to be run independently of the rest of the h test suite.


chdorner:
LGTM

PR #3333, opened 2016-05-16
Use config.testing_securitypolicy, not authn_policy fixture

nickstenning:
Changes the tests that need an authentication policy to use the Pyramid testing_securitypolicy helper, rather than our own homebrew authn_policy fixture.

This simplifies extracting the `h.api` tests to run on their own.

### DIFF
--- a/h/api/schemas.py
+++ b/h/api/schemas.py
@@ -256,7 +256,7 @@ class LegacyCreateAnnotationSchema(object):
         # group they've asked to create one in.
         if 'group' in appstruct:
             if appstruct['group'] == '__world__':
-                group_principal = security.Everyone
+                group_principal = security.Authenticated
             else:
                 group_principal = 'group:{}'.format(appstruct['group'])
             if group_principal not in self.request.effective_principals:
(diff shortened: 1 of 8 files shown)


h/api/schemas.py
```diff
@@ -256,7 +256,7 @@ def validate(self, data):
         # group they've asked to create one in.
         if 'group' in appstruct:
             if appstruct['group'] == '__world__':
-                group_principal = security.Everyone
+                group_principal = security.Authenticated
```
nickstenning: Users need to be in `security.Authenticated` to create an annotation in `__world__`, not just in `security.Everyone`, which includes unauthenticated users. This change is necessary so that [this test](https://github.com/hypothesis/h/pull/3333/files#diff-bb0455f689fc16e7f1d9fc5556c04bc5R436) continues to pass.

chdorner: I remember I raised this in a PR. This shouldn't have any side-effects since unauthenticated users couldn't even hit this code path. :+1:

chdorner:
LGTM

PR #3335, opened 2016-05-17
Enable Hound Python checking

chdorner:


nickstenning:
I ran into this before. Looks like Hound can't cope with the config file not being a .ini... ARGH.

nickstenning:
https://github.com/houndci/python/issues/5#issuecomment-221269926

nickstenning:
Going to suggest we close this for now.

PR #3340, opened 2016-05-19
Drive the limit/offset tests with data tables

nickstenning:
This reduces a bunch of repetition in the limit/offset parameter tests for the query builder and makes it trivial to add new cases.


chdorner:
LGTM

PR #3341, opened 2016-05-19
Clamp allowable values for "limit" parameter

nickstenning:
We don't want users to be able to specify arbitrarily high values for the "limit" parameter to the search API.

This commit ensures that the value is clamped to 200.

_**N.B.** This PR depends on #3340 and will need rebasing once that is merged._


chdorner:
LGTM. (Finally!)

PR #3342, opened 2016-05-19
Add options page to extension, allow disabling badge

nickstenning:
This is an experimental commit to add an options pane to the Chrome
extension, with the first available option being the ability to disable
the annotation count badge.

The settings are persisted in a dictionary using the `chrome.storage`
API, and used to disable the code responsible for updating the badge
state when a tab changes.

### DIFF
--- a/h/browser/chrome/lib/options.html
+++ b/h/browser/chrome/lib/options.html
@@ -1,0 +1,26 @@
+<!DOCTYPE html>
+<html>
+<head>
+  <title>Hypothesis Extension Settings</title>
+  <style>
+    .hint {
+      font-size: 90%;
+    }
+  </style>
+</head>
+
+<body>
+  <main>
+    <input type="checkbox" id="badge">
+    <label for="badge">Enable extension badge</label>
+    <p class="hint">
+        By default, the extension button displays the number of annotations found
+        on the pages you visit. Disable this if you find the badge distracting or
+        if you do not wish the extension to make requests to the Hypothesis
+        service for this data.
+    </p>
+
+    <script src="options.js"></script>
+  </main>
+</body>
+</html>
--- a/h/browser/chrome/lib/hypothesis-chrome-extension.js
+++ b/h/browser/chrome/lib/hypothesis-chrome-extension.js
@@ -41,6 +41,7 @@ var TAB_STATUS_COMPLETE = 'complete';
  */
 function HypothesisChromeExtension(dependencies) {
   var chromeTabs = dependencies.chromeTabs;
+  var chromeStorage = dependencies.chromeStorage;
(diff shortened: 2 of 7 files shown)


nickstenning:
@robertknight I'd be interested to hear what you make of this. It seems pretty straightforward to add an options page to the Chrome extension.

I haven't updated the `HypothesisChromeExtension` tests, but will do so if we think this is roughly the right approach.

robertknight:
I'm definitely in favor of adding this preference. Some UI feedback:

1. The settings UI should follow the conventions of Chrome's settings UI (at chrome://settings/ ), specifically the setting should auto-apply when the checkbox is toggled, rather than requiring the user to click 'Save changes'. The option should also be phrased without a question mark at the end.
2. When the option is turned on or off, it should ideally refresh the state of the badge for open tabs immediately. This would provide immediate feedback that the setting change has taken effect.

h/browser/chrome/lib/options.html
```diff
@@ -0,0 +1,26 @@
+<!DOCTYPE html>
+<html>
+<head>
+  <title>Hypothesis Extension</title>
```
robertknight: Small nit: Even though this is not visible to the user, I'd suggest setting this to "Hypothesis Extension Settings" so that it becomes more obvious which page is being referred to if this shows up in the Task Manager etc.

h/browser/chrome/lib/options.html
```diff
@@ -0,0 +1,26 @@
...
+      font-size: 90%;
+    }
+  </style>
+</head>
+
+<body>
+  <input type="checkbox" id="badge">
+  <label for="badge">Enable extension badge?</label>
+  <p class="hint">
+      By default, the extension button has a small badge that displays the
```
robertknight: This is a good description but I think it would be preferable to shorten it to a single sentence if possible and then link to https://hypothes.is/blog/hypothes-is-chrome-extension-permissions/ for more details.

nickstenning: I'm not really sure I understand. This isn't about permissions, it's about the badge. Linking to the permissions blog post is just going to be confusing.

nickstenning:
> The settings UI should follow the conventions of Chrome's settings UI (at chrome://settings/ ), specifically the setting should auto-apply when the checkbox is toggled, rather than requiring the user to click 'Save changes'. The option should also be phrased without a question mark at the end.

I've done this.

> When the option is turned on or off, it should ideally refresh the state of the badge for open tabs immediately. This would provide immediate feedback that the setting change has taken effect.

Let's leave this for a later version. I think it's going to be a bit more involved to make that happen than I can justify right now.

robertknight:
LGTM. I think the revised wording comes across well.

PR #3343, opened 2016-05-20
Fix up h.api.search.index_test database session usage

nickstenning:
Since 6a365ce, h.api tests should be using their own isolated database session. This was simply overlooked because the work was going on at the same time as the test rejiggery.


chdorner:
:tada:

PR #3344, opened 2016-05-20
Add the FeatureCohort model to h.models

finitud:
Following my own doctrine, I've split #3330 in two atomic pull requests. This one just makes the `FeatureCohort` model accessible to the rest of the application, and should be merged in before #3330.


nickstenning:
LGTM.

PR #3345, opened 2016-05-21
Allow authenticating to the WebSocket with an API token

nickstenning:
In order to programatically authenticate to the WebSocket endpoint at the moment one has to do ridiculous things like POST to the /login form and extract a cookie.

This commit makes it possible to authenticate to the WebSocket endpoint with an API token.

### DIFF
--- a/h/auth/policy.py
+++ b/h/auth/policy.py
@@ -15,12 +15,12 @@ class AuthenticationPolicy(object):
         self.session_policy = authentication.SessionAuthenticationPolicy()
 
     def authenticated_userid(self, request):
-        if _is_api_request(request):
+        if _is_token_request(request):
             return tokens.authenticated_userid(request)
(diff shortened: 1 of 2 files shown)


chdorner:
LGTM

tgbugs:
Did this go live? I'm trying to auth using the same header that works for 
`{'Authorization': 'Bearer ' + api_token}` and 
`{'Authorization': 'Bearer ' + api_token, 'Content-Type': 'application/json;charset=utf-8'}`
neither seems to produce the same behaviour as using the cookie auth.

dwhly:
https://github.com/hypothesis/h/releases/tag/v0.23.1

tgbugs:
Got it, thanks!

nickstenning:
Yep, sorry about that. Serves me right for doing things in a rush. I have a train journey tomorrow and will look at fixing this properly.

PR #3346, opened 2016-05-21
Add enter/exit animation when adder is shown

robertknight:
Add an entry animation for the adder which makes it 'pop up' or 'pop
down', depending on which direction the arrow is pointing.

Additionally, animate the color transitions when hovering over the
toolbar buttons to soften the changes.

### DIFF
--- a/h/static/scripts/annotator/adder.js
+++ b/h/static/scripts/annotator/adder.js
@@ -39,6 +39,7 @@ function Adder(element) {
   this.element = element;
 
   var view = element.ownerDocument.defaultView;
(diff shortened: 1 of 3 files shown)


nickstenning:
I'd like @conordelahunty to sign off on this one before merging.

conordelahunty:
@robertknight Can you send me a build

conordelahunty:
Have tweeked the speed a little with Robert and am happy to ship that change.

nickstenning:
:+1:

PR #3347, opened 2016-05-21
Show adder when the selection changes

robertknight:
This PR makes the adder show up in the correct position when making or adjusting a selection on touch devices.

The problem with the previous implementation is that it relied on 'mouseup' or 'touchend' events being emitted when the user made a selection. On iOS and Android under all browsers I tested, neither of these events are emitted when changing the selection with the OS selection changes. The 'touchend' event is sometimes emitted when the initial word or paragraph is selected via a long-press.

This PR fixes the problem by listening for `selectionchange` events in addition to mouse events, with the following behavior:

* When selectionchange events occur, we buffer the changes and wait for a pause of 100ms before checking for a selection and showing or hiding the adder. This avoids the adder moving around continually whilst moving the OS drag handles
* When a selectionchange event occurs whilst the user is dragging the mouse, we wait until the user finishes dragging the mouse before showing the adder. Again this avoids popping up the adder mid-selection

The implementation uses [Observables](https://github.com/zenparsing/es-observable) (a proposed API for ES.next) via the minimal [zen-observable](https://github.com/zenparsing/zen-observable) package. This provides a nice way to work with streams of UI events (or some other push-based source) in order to implement things like buffering, delays, merging events from different sources etc.


h/static/scripts/annotator/selections.js
```diff
@@ -0,0 +1,129 @@
...
+
+var Observable = require('zen-observable');
+
+/**
+ * Returns an observable of events emitted by `src`.
+ *
+ * @param {EventTarget} src - The event source.
+ * @param {Array<string>} eventNames - List of events to subscribe to
+ */
+function listen(src, eventNames) {
```
robertknight: Note: The `listen()`, `buffer()` and `merge()` functions in this file are generic utilities for creating and working with observables. It would probably make sense to extract them into a separate file.

robertknight:
@sheetaluk - Could you take a look at this? It should work on both iDevices and Android.

robertknight:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;7CQdhTOg&#x2F;354-make-adder-adjust-position-when-selection-changes-on-touch-devices"> Make adder adjust position when selection changes on touch devices</a>

chdorner:
LGTM. Selecting text on mobile devices is gradually getting better :tada:

judell:
Things are much improved on Windows Phone as well. Not many will care, and I myself am likely not long for this platform, but it's a nice validation of this work.

swapnildeshmukh15:
Hello guys, 

  I would like to know if this annotations will work with URL that we call in webview on Android or iOS?

PR #3349, opened 2016-05-23
Add reindex CLI command

chdorner:
* Move `h.indexer.reindex` code into `h.api.search.index` (so it can be called from multiple places)
* Rename celery task to `h.indexer.reindex_annotations`
* Rename old reindex task to `legacy_reindex`
* Add new reindex task as `reindex`

### DIFF
--- a/h/api/search/__init__.py
+++ b/h/api/search/__init__.py
@@ -76,6 +76,14 @@ def includeme(config):
         name='legacy_es',
         reify=True)
 
+    # request.new_es is always a client for the legacy Elasticsearch index,
+    # regardless of whether the 'postgres' feature flag is on.
+    # This should be used to write to the new search index.
+    config.add_request_method(
+        lambda r: _get_client(r.registry.settings),
+        name='new_es',
+        reify=True)
+
     # request.es is a client for either the new or the legacy Elasticsearch
     # index, depending on the 'postgres' feature flaf.
     # This should always be used to read from the search index.
--- a/h/api/search/__init__.py
+++ b/h/api/search/__init__.py
@@ -71,6 +71,7 @@ def includeme(config):
     # request.legacy_es is always a client for the legacy Elasticsearch index,
     # regardless of whether the 'postgres' feature flag is on.
     # This should be used to write to the legacy search index.
+    # TODO: Remove when postgres migration is done
     config.add_request_method(
         lambda r: _legacy_get_client(r.registry.settings),
         name='legacy_es',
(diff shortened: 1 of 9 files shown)


h/indexer.py
```diff
@@ -32,15 +31,11 @@ def delete_annotation(id_):
 
 
 @celery.task
-def reindex():
+def reindex_annotations():
     if not celery.request.feature('postgres'):
```
nickstenning: Didn't you also need to remove this? Or are you saving that for another PR?

tests/h/api/search/index_test.py
```diff
@@ -111,6 +111,31 @@ def log(self, patch):
         return patch('h.api.search.index.log')
 
 
+@pytest.mark.usefixtures('BatchIndexer', 'BatchDeleter')
+class TestReindex(object):
+    def test_it_indexes_all_annotations(self, BatchIndexer):
+        session, es, request = mock.Mock(), mock.Mock(), mock.Mock()
```
nickstenning: For situations like this where you're just passing named things around, it's probably nicer to use `mock.sentinel`:

```python
s = mock.sentinel
index.reindex(s.session, s.es, s.request)
BatchIndexer.assert_called_once_with(s.session, s.es, s.request)
```

tests/h/indexer_test.py
```diff
@@ -135,6 +123,7 @@ def delete_annotation(self, patch):
 @pytest.fixture
 def celery(patch):
     cel = patch('h.indexer.celery')
-    cel.request = DummyRequest(db=mock.Mock(), es=mock.Mock(), feature=mock.Mock())
+    cel.request = DummyRequest(
+        db=mock.Mock(), es=mock.Mock(), new_es=mock.Mock(), feature=mock.Mock())
```
nickstenning: Does it need `es` any more?

h/api/search/__init__.py
```diff
@@ -76,6 +76,14 @@ def includeme(config):
         name='legacy_es',
         reify=True)
 
+    # request.new_es is always a client for the legacy Elasticsearch index,
+    # regardless of whether the 'postgres' feature flag is on.
+    # This should be used to write to the new search index.
```
nickstenning: Perhaps just add a `TODO` to remove this once the migration is done?

tests/h/api/search/index_test.py
chdorner: Yeah, I need to start using `mock.sentinel` more. Didn't do the assignment to a shorter variable though.

tests/h/indexer_test.py
chdorner: Yes, the other tasks still use `request.es` because they never run when the feature is off (i.e. [here](https://github.com/hypothesis/h/blob/reindex-cli-command/tests/h/indexer_test.py#L28))

h/api/search/__init__.py
chdorner: :+1:

h/indexer.py
chdorner: I still want to keep the celery task for reindex to check for the feature. The CLI command on the other hand does run regardless of the flag.

nickstenning:
LGTM.

PR #3350, opened 2016-05-23
Add target[scope] when batch indexing an annotation

chdorner:
We still use this field when searching, but the batch indexer did not set it
the same way we do it for indexing individual annotations. Now we do :tada:


nickstenning:
LGTM.

PR #3352, opened 2016-05-23
Update Python code style docs

seanh:
Update our Python code style documentation to match our actual practice:

- Removed "Most important, though, is to follow the style of the code
  you are modifying, if your edits are not new files." The advice to
  follow existing style is already in both PEP 8 and the Google Python
  Style Guide so there's no need to repeat it. Also, I'd argue that
  whether or not following existing style is the _most_ important thing
  is a context-dependent decision.

- Removed "Please stick to strict, 80-column line limits except for small
  exceptions that would still be readable if they were truncated."
  Again this is already stated in PEP 8 and the Google guide (and it's
  79 chars not 80).

- Removed "Eliminate trailing whitespace wherever possible", this is in
  PEP 8.

- Added recommendation to use Flake8 first, Prospector second
  (previously it said Prospector only).

- Removed mention of `make lint`.
  I'm not sure whether `make lint` is actually a useful command to have
  since it runs Prospector only and not Flake8 (or any of our JavaScript
  or other lint tools) and runs it on the whole codebase at once.

- Moved the Google Python Style Guide to a new "Additional reading"
  section and clarified that it isn't mandatory.

### DIFF
--- a/docs/hacking/code-style.rst
+++ b/docs/hacking/code-style.rst
@@ -1,42 +1,46 @@
 Code style
 ##########
 
-Notes on code style follow for the different languages used in the
-project. Most important, though, is to follow the style of the code
-you are modifying, if your edits are not new files.
-
-Please stick to strict, 80-column line limits except for small
-exceptions that would still be readable if they were truncated.
-
-Eliminate trailing whitespace wherever possible.
-
-Linting
--------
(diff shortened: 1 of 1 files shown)


nickstenning:
LGTM.

PR #3353, opened 2016-05-23
Remove `make lint`

seanh:
Does anyone actually use this? It runs Prospector only, even though our
primary Python linter is now Flake8, and it runs it on the entire
codebase at once (and doesn't accept args to run on individual files).
It doesn't run any of the JavaScript linters that we use either.

It seems simpler just to let people run the linters on the code they're
currently editing using things like text editor plugins,
GitHub-integrated CI tools on pull requests, etc.


nickstenning:
LGTM.

PR #3354, opened 2016-05-23
Add docs on docstrings

seanh:
This is branched off https://github.com/hypothesis/h/pull/3352.

For the sake of starting a discussion It says that _all_ public stuff should have docstrings (which is what PEP 8 and PEP 257 say), but do we want to change that to something else?

Apart from that it adds a link to PEP 257 and adds some links on how to use Sphinx stuff to help with docstrings.


nickstenning:
LGTM.

PR #3355, opened 2016-05-24
Selector is not a required property for page notes

nickstenning:
We need to accept annotation updates where the target has no selectors,
because this is a valid "page note".


chdorner:
LGTM.

PR #3357, opened 2016-05-24
Remove legacy ES code

chdorner:
This PR removes most legacy code. It makes h stop writing / reading / updating from/to the old index, and removes a few other legacy ES related code paths.

More importantly, this PR does not remove the `h.api.models.elastic` module (and thus the `annotator` dependency) due to `h.api.storage.annotation_from_dict` which is still being used when presenting search results, in the streamer, and in the groups and feed views.
We need to write some code to be able to remove these last code paths. The solution for what to do here still needs to be discussed (do we want to go all the way and load annotations from PG after search results come back from ES?).

But nevertheless, this PR should be good to be reviewed. Merging + deploying should happen as soon as we're happy that we don't want to switch back and turn the `postgres` feature flag off.

I've structured the commits to be around removing legacy code from individual features. If the PR is too big please let me know. Splitting it up into multiple PR's is very easy (although I'd like to keep it as one deployment).


chdorner:
I'll nominate both @seanh and @nickstenning to review and test this out locally.

chdorner:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;XW03lN5D&#x2F;349-postgres-migration-cleanup"> Postgres migration cleanup</a>

h/api/search/__init__.py
```diff
@@ -77,21 +77,13 @@ def includeme(config):
...
-    # request.new_es is always a client for the legacy Elasticsearch index,
-    # regardless of whether the 'postgres' feature flag is on.
-    # This should be used to write to the new search index.
-    # TODO: Remove when postgres migration is done
-    config.add_request_method(
-        lambda r: _get_client(r.registry.settings),
-        name='new_es',
-        reify=True)
-
     # request.es is a client for either the new or the legacy Elasticsearch
```
nickstenning: This comment needs updating and `_get_client_or_legacy_client` can be removed.

h/api/storage.py
```diff
@@ -209,12 +160,8 @@ def delete_annotation(request, id_):
...
     :type id_: str
     """
-    if _postgres_enabled(request):
-        annotation = fetch_annotation(request, id_, _postgres=True)
-        request.db.delete(annotation)
-
-    legacy_annotation = fetch_annotation(request, id_, _postgres=False)
-    legacy_annotation.delete()
+    annotation = fetch_annotation(request, id_)
+    request.db.delete(annotation)
```
nickstenning: No need to change this now, but this should now really be doing something like `request.db.query(Annotation).filter_by(id=id_).delete()` rather than fetching the thing first.

h/api/storage.py
```diff
@@ -252,18 +195,3 @@ def expand_uri(request, uri):
             return [uri]
 
     return [docuri.uri for docuri in docuris]
-
-
-def _prepare(request, annotation):
-    """Prepare the given annotation for storage."""
-    fetcher = partial(fetch_annotation, request, _postgres=False)
-    transform.prepare(annotation, fetcher)
```
nickstenning: I'm not 100% sure but I think this means the entirety of `h.api.transform` can go now, too.

tests/h/api/storage_test.py
```diff
@@ -426,11 +195,7 @@ def test_it_does_not_crash_if_no_text_or_tags(self):
         storage.create_annotation(self.mock_request(), data)
 
     def mock_request(self):
-        request = DummyRequest(
-            feature=mock.Mock(
-                side_effect=lambda flag: flag == "postgres_write"),
```
nickstenning: ¯\_(ツ)_/¯

nickstenning:
Diff is looking pretty good. I think you can remove `h.api.transform` and tests as well now.

h/api/storage.py
chdorner: Created a card for it: https://trello.com/c/vW2ihM1m/353-improve-pg-annotation-delete-code

h/api/storage.py
chdorner: Yeah, it wasn't being used anywhere. It is gone now.

h/api/search/__init__.py
chdorner: I've removed this comment and moved the code to where the old comment was. So once we can get rid of all legacy code we just need to remove the block where it adds the `legacy_es` method.

chdorner:
I've addressed the comments. Will keep it labeled as WIP until we release 0.24.0, this should then go into the 0.25.0 release.

tests/h/api/storage_test.py
```diff
@@ -646,25 +386,6 @@ def test_it_deletes_the_annotation(self, fetch_annotation, mock_request):
...
-            second_return_value,
-        ]
-
-        storage.delete_annotation(mock_request, "test_id")
-
-        second_return_value.delete.assert_called_once_with()
-
     @pytest.fixture
     def mock_request(self, session):
         request = DummyRequest()
```
seanh: It's probably not necessary for this to be a fixture anymore, though maybe kind of tedious to change

tests/h/api/views_test.py
```diff
@@ -351,35 +214,9 @@ def test_it_returns_presented_annotation(self,
...
-            views.create(mock_request)
-        except ValidationError:
-            pass
-
-        assert not storage.legacy_create_annotation.called
-
     @pytest.fixture
     def mock_request(self):
-        return mock.Mock(feature=mock.Mock(return_value=True))
+        return mock.Mock()
```
seanh: Tests can probably use `testing.DummyRequest()` instead of this fixture now

h/api/schemas.py
```diff
@@ -233,37 +200,6 @@ def validate(self, data):
...
-        _remove_protected_fields(appstruct)
-
-        # Set the annotation user field to the request user.
-        appstruct['user'] = self.request.authenticated_userid
-
-        # Assert that the user has permission to create an annotation in the
-        # group they've asked to create one in.
-        if 'group' in appstruct:
-            if appstruct['group'] == '__world__':
-                group_principal = security.Authenticated
```
seanh: I think `security` is now an unused import

h/api/storage.py
```diff
@@ -45,28 +41,13 @@ def fetch_annotation(request, id_, _postgres=None):
...
-    return models.elastic.Annotation.fetch(id_)
-
 
-def legacy_create_annotation(request, data):
-    annotation = models.elastic.Annotation(data)
-    # FIXME: this should happen when indexing, not storing.
-    _prepare(request, annotation)
-    annotation.save()
-    return annotation
+    try:
```
seanh: There shouldn't be a blank line here (pep257)

h/api/views.py
```diff
@@ -209,17 +188,7 @@ def read_jsonld(annotation, request):
...
-        legacy_annotation = storage.fetch_annotation(request,
-                                                     annotation.id,
-                                                     _postgres=False)
-        _update_elastic(legacy_annotation, request, notify=False)
-        return annotation_dict
-    return _update_elastic(annotation, request, notify=True)
 
-
-def _update_postgres(annotation, request):
     schema = schemas.UpdateAnnotationSchema(request,
```
seanh: Shouldn't be a blank line before this one

seanh:
Gave this some testing, LGTM

nickstenning:
LGTM.

PR #3358, opened 2016-05-25
Fix strict-mode violations in streamer.js

robertknight:
Fix several missing 'var' declarations and JSHint lint violations in streamer.js


nickstenning:
LGTM.

PR #3360, opened 2016-05-25
Improve logic for scrolling sidebar to annotation with a given ID

robertknight:
The previous logic contained several errors and omissions which could
result in the card not appearing at the top of the viewport after
scrolling.

 1. If the set of visible threads changed multiple times before the timeout
    that triggered re-measuring of visible thread heights changed, then
    a measured height could be replaced with 0.

    Add checks in setThreadHeight() to catch this.

 2. Account for the target scroll offset changing as a result of
    the document height increasing after the actual heights of
    cards become known and consequently the maximum possible
    scroll offset changes.

 3. Correct calculation of the position that we need to scroll the
    window to in order to have a card positioned at the top
    of the window. For the first card in the list, the position
    should be 0 (as returned by visibleThreads.yOffsetOf(...)),
    and for the Nth card it should be the total height of the N-1
    previous cards.

### DIFF
--- a/h/static/scripts/widget-controller.js
+++ b/h/static/scripts/widget-controller.js
@@ -36,6 +36,11 @@ module.exports = function WidgetController(
   drafts, groups, rootThread, settings, streamer, streamFilter, store,
   VirtualThreadList
 ) {
+
+  /**
+   * Returns the height of the thread for an annotation if it exists in the view
+   * or undefined otherwise.
+   */
   function getThreadHeight(id) {
     var threadElement = document.getElementById(id);
     if (!threadElement) {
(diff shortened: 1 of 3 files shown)


h/static/scripts/widget-controller.js
```diff
@@ -66,7 +75,11 @@ module.exports = function WidgetController(
 
     scopeTimeout($scope, function () {
       state.visibleThreads.forEach(function (thread) {
-        visibleThreads.setThreadHeight(thread.id, getThreadHeight(thread.id));
+        var height = getThreadHeight(thread.id);
```
robertknight: @nickstenning You get to say I-told-you-so about this ;)

h/static/scripts/widget-controller.js
```diff
@@ -334,29 +347,33 @@ module.exports = function WidgetController(
...
 
   /**
-   * Return the offset between the top of the window and the top of the
-   * first annotation card.
+   * Return the vertical scroll offset for the document in order to position the
+   * annotation thread with a given `id` or $$tag at the top-left corner
+   * of the view.
    */
-  function cardListYOffset() {
-    var cardListTopEl = document.querySelector('.js-thread-list-top');
```
nickstenning: This class remains on the spacer in the template but is now not referenced anywhere.

nickstenning:
LGTM.

PR #3361, opened 2016-05-25
Disable mccabe, pep8 and pyflakes in prospector

seanh:
We now use flake8 first and then prospector as a second pass linter.
Flake8 already runs mccabe, pep8 and pyflakes so no need for prospector
to also run them.


nickstenning:
LGTM.

PR #3363, opened 2016-05-25
Add py3kwarn to dev dependencies

seanh:


nickstenning:
I'm happy to add this, but is there a reason?

nickstenning:
Whatever. Can't hurt.

seanh:
I just noticed that I had `py3kwarn` in my Syntastic config (and it _has_ been telling me interesting things about Python 2-3) but after deleting and recreating my h virtualenv following our docs Syntastic was (silently) not running py3kwarn anymore 'cos it wasn't installed. So I just figured it can't hurt to have it.

PR #3365, opened 2016-05-25
Re-enable pep257's "missing docstring" messages

seanh:
We recently added "All public modules, functions, classes, and methods
should normally have docstrings" to our code style docs, so it seems
like we should have these messages enabled.

This disabling of pep257 missing docstring warnings wasn't actually
working anyway, since pylint's missing-docstring message wasn't
disabled, and since we run both pylint and pep257 at the same time (via
prospector) pylint would still complain about all the same missing
docstrings!

So this commit also disables pylint's missing docstrings warnings. We
run both pylint and pep257 at the same time via prospector, so there's
no point in having them both complain about missing docstrings.

### DIFF
--- a/.prospector.yaml
+++ b/.prospector.yaml
@@ -19,6 +19,8 @@ pylint:
   disable:
     - line-too-long  # PEP8 checks this and doesn't complain about
                      # unavoidable long lines (such as URLs).
+    - missing-docstring  # The pep257 tool reports missing docstrings to us
+                         # so we don't need pylint to do so.
     - too-few-public-methods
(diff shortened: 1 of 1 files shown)


seanh:
Although actually I don't think there's any case-by-case way to disable messages from `pep257`. I wonder if we should disable all the `pep257` missing docstring messages, but enable the pylint ones, because the pylint ones can be silenced with `# pylint` comments?

nickstenning:
If we need to invert the sense of this later, that's fine. For now, LGTM.

PR #3370, opened 2016-05-26
Add padding, fix hover state colours, tweak shadow

conordelahunty:
Making this change to fix visual alignment of buttons and increase the overall visibility of the adder.


robertknight:
LGTM

PR #3371, opened 2016-05-26
devserver: Extend ALLOWED_ORIGINS instead of overriding it

chdorner:
Which allows for a local environment variable that includes the chrome
extension URL to have websockets work out of the box.


nickstenning:
LGTM.

PR #3372, opened 2016-05-26
Remove docs about redirecting docs pages

seanh:
The Read the Docs UI for maintaining redirects is terrible.
You can't see which redirect is which when you want to remove redirects.
And there is apparently no way to tell it to redirect for only a
specific named version of the docs such as "latest", so either you break
links for old (but valid) docs.


chdorner:
LTGM.

PR #3374, opened 2016-05-26
Send up to 200 replies per search request

### LINKED ISSUE #3251: Replies showing "Message not available" in the sidebar
### Steps to reproduce
1. Open this direct link: https://hyp.is/AVRP4RFaH9ZO4OKSl57O/www.digitalpedagogylab.com/hybridped/teaching-in-our-right-minds/
2. Open the same as a standalone link: https://hypothes.is/a/AVRP4RFaH9ZO4OKSl57O
3. Expose the replies in each.

### Expected behaviour
The threads in both should be the same.

### Actual behaviour
Inside the sidebar, many of the replies are collapsed with "Message not available".

### Additional details
If you watch the annotation load in the sidebar via the direct link you'll see the reply count pause for about 15 seconds at "2" and then it goes to "7".  It's kind of like it's waiting for some of those missing replies and then gives up.

seanh:
Increase the upper limit on the number of reply annotations sent in a
single search request response with _separate_replies is True from 100
to 200.

This reduces the number of "Message not available" annotation cards seen
in the browser in place of reply annotations, on pages that have a lot
of replies.

Fixes #3251

### DIFF
--- a/h/api/search/core.py
+++ b/h/api/search/core.py
@@ -57,7 +57,7 @@ def search(request, params, private=True, separate_replies=False):
             [h['_id'] for h in results['hits']['hits']]))
         reply_results = es.conn.search(index=es.index,
                                        doc_type=es.t.annotation,
-                                       body=builder.build({'limit': 100}))
+                                       body=builder.build({'limit': 200}))
 
         if len(reply_results['hits']['hits']) < reply_results['hits']['total']:
             log.warn("The number of reply annotations exceeded the page size "


seanh:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;hgNerzsJ&#x2F;343-bug-message-not-available-in-threads"> [BUG] Message not available in threads</a>

chdorner:
:+1:

PR #3375, opened 2016-05-27
Add edit view for feature cohorts

finitud:
This allows an admin to add/remove users to a feature cohort.

### DIFF
--- a/h/admin/__init__.py
+++ b/h/admin/__init__.py
@@ -9,7 +9,7 @@ def includeme(config):
     config.add_route('admin_badge', '/badge')
     config.add_route('admin_features', '/features')
     config.add_route('admin_cohorts', '/features/cohorts')
-    config.add_route('admin_edit_cohort', '/features/cohorts/{id}')
+    config.add_route('admin_cohorts_edit', '/features/cohorts/{id}')
     config.add_route('admin_groups', '/groups')
     config.add_route('admin_groups_csv', '/groups.csv')
     config.add_route('admin_nipsa', '/nipsa')
--- a/h/admin/__init__.py
+++ b/h/admin/__init__.py
@@ -9,6 +9,7 @@ def includeme(config):
     config.add_route('admin_badge', '/badge')
     config.add_route('admin_features', '/features')
     config.add_route('admin_cohorts', '/features/cohorts')
+    config.add_route('admin_edit_cohort', '/features/cohorts/{id}')
     config.add_route('admin_groups', '/groups')
     config.add_route('admin_groups_csv', '/groups.csv')
     config.add_route('admin_nipsa', '/nipsa')
--- a/h/admin/views/features.py
+++ b/h/admin/views/features.py
@@ -101,7 +101,7 @@ def cohorts_edit_remove(request):
     cohort_id = request.matchdict['id']
 
     cohort = request.db.query(models.FeatureCohort).get(cohort_id)
-    member = request.db.query(models.User).filter_by(uid=member_name).first()
+    member = request.db.query(models.User).filter_by(username=member_name).first()
     try:
         cohort.members.remove(member)
     except ValueError:
--- a/h/admin/views/features.py
+++ b/h/admin/views/features.py
@@ -102,7 +102,12 @@ def cohorts_edit_remove(request):
 
     cohort = request.db.query(models.FeatureCohort).get(cohort_id)
     member = request.db.query(models.User).filter_by(uid=member_name).first()
-    cohort.members.remove(member)
+    try:
+        cohort.members.remove(member)
+    except ValueError:
+        request.session.flash(
+            _("User {member_name} doesn't exist.".format(member_name=member_name)),
+            "error")
 
     url = request.route_url('admin_cohorts_edit', id=cohort_id)
     return httpexceptions.HTTPSeeOther(url)
--- a/h/admin/views/features.py
+++ b/h/admin/views/features.py
@@ -59,7 +59,7 @@ def cohorts_add(request):
     return httpexceptions.HTTPSeeOther(url)
 
 
-@view_config(route_name='admin_edit_cohort',
+@view_config(route_name='admin_cohorts_edit',
              request_method='GET',
              renderer='h:templates/admin/edit_cohort.html.jinja2',
              permission='admin_features')
@@ -69,7 +69,7 @@ def cohorts_edit(context, request):
     return {'cohort': cohort, 'members': cohort.members}
 
 
-@view_config(route_name='admin_edit_cohort',
+@view_config(route_name='admin_cohorts_edit',
              request_method='POST',
              request_param='add',
              renderer='h:templates/admin/edit_cohort.html.jinja2',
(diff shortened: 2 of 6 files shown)


h/admin/__init__.py
```diff
@@ -9,6 +9,7 @@ def includeme(config):
     config.add_route('admin_badge', '/badge')
     config.add_route('admin_features', '/features')
     config.add_route('admin_cohorts', '/features/cohorts')
+    config.add_route('admin_edit_cohort', '/features/cohorts/{id}')
```
nickstenning: For consistency with the other views here this should probably be called `admin_cohorts_edit`.

h/admin/views/features.py
```diff
@@ -59,5 +59,54 @@ def cohorts_add(request):
...
+             request_param='remove',
+             renderer='h:templates/admin/edit_cohort.html.jinja2',
+             permission='admin_features')
+def cohorts_edit_remove(request):
+    member_name = request.params['remove']
+    cohort_id = request.matchdict['id']
+
+    cohort = request.db.query(models.FeatureCohort).get(cohort_id)
+    member = request.db.query(models.User).filter_by(uid=member_name).first()
+    cohort.members.remove(member)
```
nickstenning: While you actually have to construct such a request by hand at the moment, this view still shouldn't crash if the username passed is not a member of the cohort, as it will currently.

h/admin/views/features.py
```diff
@@ -59,5 +59,54 @@ def cohorts_add(request):
...
+             request_method='POST',
+             request_param='remove',
+             renderer='h:templates/admin/edit_cohort.html.jinja2',
+             permission='admin_features')
+def cohorts_edit_remove(request):
+    member_name = request.params['remove']
+    cohort_id = request.matchdict['id']
+
+    cohort = request.db.query(models.FeatureCohort).get(cohort_id)
+    member = request.db.query(models.User).filter_by(uid=member_name).first()
```
nickstenning: Any particular reason you use `get_by_username` above and `filter_by(uid=...)` here? It should probably be consistent.

h/templates/admin/edit_cohort.html.jinja2
```diff
@@ -0,0 +1,59 @@
...
+      <h3 class="panel-title">Members</h3>
+    </div>
+    <div class="panel-body">
+      {% if members %}
+        <form method="POST" action="{{ request.route_url('admin_edit_cohort', id=cohort.id) }}">
+          <input type="hidden" name="cohort" value="{{ cohort.id }}">
+          <ul>
+            {% for user in members %}
+              <li>
+                {{ user.uid }}
```
nickstenning: This should probably be a username. `uid` is a column we use to ensure that we can't create usernames that differ only by case or dots, and shouldn't be exposed in the web interface.

tests/h/admin/views/features_test.py
```diff
@@ -75,21 +77,58 @@ def test_cohorts_index_when_no_cohorts():
...
     result = views.cohorts_index({}, req)
     assert len(result["results"]) == 1
-    assert result["results"][0].name == "cohort"
+
+    cohort = result["results"][0]
+    assert cohort.name == "cohort"
+    assert len(cohort.members) == 0
+
+
+def create_cohort(req):
```
nickstenning: I wouldn't recommend using other view functions to prepare the database. Just create a `FeatureCohort` directly and add it to the database.

Doing it this way leaves these tests (for the add and remove user from cohort views) liable to break if a different view (create a cohort) changes, which I'd say would be a surprising for somebody working on the latter view.

tests/h/admin/views/features_test.py
```diff
@@ -75,21 +77,58 @@ def test_cohorts_index_when_no_cohorts():
...
+    req.params['add'] = 'cohort'
+    views.cohorts_add(req)
+    return views.cohorts_index({}, req)['results'][0]
+
+
+def add_user_to_cohort(req, user, cohort):
```
nickstenning: Ditto. You could achieve the same database state in about as many lines by just creating a `FeatureCohort` object yourself and adding it to the session.

tests/h/admin/views/features_test.py
```diff
@@ -75,21 +77,58 @@ def test_cohorts_index_when_no_cohorts():
...
+    req.matchdict['id'] = cohort.id
+    req.params['add'] = user.username
+    views.cohorts_edit_add(req)
+
+
+def test_cohort_edit_add_and_remove_user():
```
nickstenning: Tests should test one thing at a time, so I'd suggest testing the `cohorts_edit`, `cohorts_edit_add`, and `cohorts_edit_remove` view functions in (at least) three different tests.

nickstenning: And in case it's not clear _why_ tests "should test one thing at a time," it's so it's easy to identify what exactly is broken when a test fails. In this case, there are (at least) three different functions (not counting the use of `create_cohort` and by extension the `cohorts_add` and `cohorts_index` views) which could have broken, and I'd then have to drill into this test before I even understood what was broken.

h/admin/views/features.py
```diff
@@ -59,5 +59,54 @@ def cohorts_add(request):
...
+    if member is None:
+        request.session.flash(
+            _("User {member_name} doesn't exist.".format(member_name=member_name)),
+            "error")
+    else:
+        cohort = request.db.query(models.FeatureCohort).get(cohort_id)
+        cohort.members.append(member)
+
+    url = request.route_url('admin_edit_cohort', id=cohort_id)
+    return httpexceptions.HTTPSeeOther(url)
```
nickstenning: It's probably best to always `raise` HTTPExceptions. There are some you can return (like 302) but there are others which you _can_ return but probably shouldn't (like 404, which will render a 404 but will ignore your custom 404 templates if you `return` rather than `raise` it). For the sake of consistency, and to minimise the amount of thinking you have to do when reading a function like this, I'd suggest always `raise` if the response is an instance of `HTTPException`.

h/admin/__init__.py
nickstenning: That would also make it consistent with the name of the view functions which implement it, which is always nice.

h/admin/views/features.py
finitud: I forgot to wrap the one above in the `request.db` session, for what I can tell I can't `get_by_username` if I use `request.db.query`. Changing the one above probably makes more sense?

nickstenning: Don't use uid. Just use `get_by_username` for now. Yes, it needs to change (see #3423), but `uid` is an internal helper to ensure username uniqueness, not a replacement for username.

chdorner: #3423 is now merged. So `get_by_username` should now accept a session as well (after a rebase).

nickstenning: Not just yet ;)

h/admin/views/features.py
nickstenning: Ignore this. It's certainly true that ideally we'd always raise HTTPExceptions but it doesn't work, as I just reminded myself, because it results in the request transaction being aborted.

nickstenning:
LGTM.

PR #3376, opened 2016-05-27
Implement new design for hovered conversation threads

robertknight:
Implement the changes to the background hover style and spacing. This initially started from
https://trello.com/c/aXCXxzx2 but some changes were made based on design review during the implementation:

- The grey background for cards on hover has intentionally not been implemented
- The dashed line alongside cards now aligns to the top of the 'Only me' indicator if present on a reply.
- The 'Show/hide replies' toggle is no longer shown for replies

Implementation notes:

 * Use `--reply` modifier classes on <annotation> and
   <annotation-thread> elements to style replies differently from
   top-level annotations. This avoids mixing style classes from
   different components (annotation-thread and annotation) in the same
   file which was required when using descendant selectors.

 * Rename `thread` CSS classes to match the name of the component
   that they are used in (annotation-thread)

 * Move the `annotation-unavailable-message` class to `app.scss` alongside
   the classes for other elements of the root template.

- [x] Show expand/collapse toggle in dark grey only when annotation (not replies) are hovered
- [x] Design review with Conor
- [x] Decide whether to keep or remove the 'Hide replies' option for nested replies


h/static/styles/annotation-thread.scss
```diff
@@ -0,0 +1,83 @@
...
+  height: 100%;
+
+  &.is-open {
+    // When the thread is expanded, the top of the dashed line is should be
+    // aligned with the reply's body.
+    height: 30px;
+  }
+}
+
+.annotation-thread__load-more {
```
robertknight: - [ ] - Remove this. I don't think it is needed any more since we're using flexbox for layout

robertknight:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;7nr9z7YA&#x2F;336-reply-threading"> Reply threading</a>

robertknight:
This is now ready pending a final sign-off from @conordelahunty of the design changes.

h/static/styles/annotation-thread.scss
```diff
@@ -0,0 +1,72 @@
+.annotation-thread {
+  display: flex;
+  flex-direction: row;
+}
+
+// Direct or nested reply to an annotation
+.annotation-thread--reply {
+  // The negative margin is used so that the hovered background of reply threads
```
robertknight: Note: This needs to be removed and simplified as we decided not to implement the hovered background from the specs. The comment is also out of date here.

robertknight:
@sheetaluk - This is ready for review now.

dwhly:
Is there a current .crx?

robertknight:
@dwhly - There will be shortly. Currently investigating some differences in how Chrome and Firefox render dashed lines.

robertknight:
@dwhly - Latest build is on the Trello card at https://trello.com/c/7nr9z7YA/336-reply-threading

PR #3377, opened 2016-05-27
Add integration test for anchoring

robertknight:
This adds an integration test for anchoring of annotations loaded
into the annotator guest via the sidebar.

This was extracted from https://github.com/hypothesis/h/pull/3313

The test case for nested annotations has been marked as an expected
failure until a better resolution for #3278 has been found.

### DIFF
--- a/h/static/scripts/annotator/test/integration/anchoring-test.js
+++ b/h/static/scripts/annotator/test/integration/anchoring-test.js
@@ -1,0 +1,102 @@
+// Tests that the expected parts of the page are highlighted when annotations
+// with various combinations of selector are anchored.
+
+'use strict';
+
+var Annotator = require('annotator');
+
+var unroll = require('../../../test/util').unroll;
+var Guest = require('../../guest');
+
+function quoteSelector(quote) {
+  return {
+    type: 'TextQuoteSelector',
+    exact: quote,
+  };
+}
+
+/** Generate an annotation that matches a text quote in a page. */
+function annotateQuote(quote) {
+  return {
+    target: [{
+      selector: [quoteSelector(quote)],
+    }],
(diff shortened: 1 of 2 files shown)


nickstenning:
Hmmm. I'm not really sure I see what's added by making this an "integration" test. Looking at the subscriber for the `annotationsLoaded` event it seems like this test would be a lot clearer if we simply tested that the `anchor` method on `Guest` resulted in appropriate highlights being drawn in the page.

nickstenning:
But of course it's perfectly possible I'm missing something, so please correct me if so!

robertknight:
It is an integration test primarily in the sense that it covers all of the anchoring and highlighting logic which spans several modules.

I'm emitting the "annotationsLoaded" event primarily because that is the entry point that the actual application uses.

nickstenning:
Yes, I get that, but if you actually look at the subscriber for that event it's just a loop over `this.anchor` — I really don't think the complexity of the tests here (in particular the fake cross frame stub) is adding enough value to justify itself.

Most of the value would be realisable just by testing `Guest#anchor` directly.

robertknight:
I do think we should have a test that covers the actual sequence of events that happens in the app but I agree that if this test is focused on testing the anchoring, just using `Guest.anchor()` directly is simpler.

I have modified the test accordingly.

nickstenning:
LGTM. Thanks, looks a lot simpler now :)

robertknight:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;V0RfL0fN&#x2F;345-spike-to-investigate-problem-with-nested-annotations"> Spike: to investigate problem with nested annotations</a>

PR #3378, opened 2016-05-27
Load annotation from DB in streamer

chdorner:
We need to get rid of the legacy `h.api.storage.annotation_from_dict`, plus we always knew that sending whole JSON blobs of annotations through the realtime system is not the best idea.
This PR gets rid of both of these by sending only the annotation id over the wire, the streamer side then loads the annotation, serializes it, and then sends the notification to the client.


h/streamer/messages.py
```diff
@@ -91,33 +93,51 @@ def handle_annotation_event(message, socket):
     annotation event, otherwise a dict containing information about the event.
     """
     action = message['action']
-    annotation = storage.annotation_from_dict(message['annotation'])
```
chdorner: One down, three to go :gun:

h/streamer/messages.py
```diff
@@ -91,33 +93,51 @@ def handle_annotation_event(message, socket):
...
+    notification = {
+        'type': 'annotation-notification',
+        'options': {'action': action},
+    }
+    id_ = message['annotation_id']
+
+    # Return early when action is delete
+    if action == 'delete':
+        notification['payload'] = [{'id': id_}]
+        return notification
```
chdorner: We've been talking about the case where annotations get deleted. This means that we can't load them from the database anymore at this moment in time.
The client already supported removing annotations by just having an id (with a small fix). But I wonder if this now means that we can't filter anything, so when a user is on a URL with the sidebar open, they will get every single delete action pushed through the websocket even when it's for a completely different URL.

nickstenning: Is there any way we could fix this by sending the relevant URL/URLs with delete notifications for the time being?

h/streamer/messages.py
```diff
@@ -91,33 +93,51 @@ def handle_annotation_event(message, socket):
...
+    annotation = storage.fetch_annotation(socket.request.db, id_)
+
+    if annotation is None:
         return None
 
-    # We don't send anything until we have received a filter from the client
-    if socket.filter is None:
+    serialized = presenters.AnnotationJSONPresenter(
+        socket.request, annotation).asdict()
+    transform_event = events.AnnotationTransformEvent(
```
nickstenning: These seems weird to me. The purpose of the `TransformEvent` is to allow bits of the application to modify the payload that's indexed in ES, right?

I guess it's here because it used to be in `publish_annotation_event`, but that seems odd. I'm pretty sure the only reason it's actually needed here is for NIPSA, and that we address more sensibly, I think, by doing the following:

```python
from h.nipsa.logic import has_nipsa
if has_nipsa(annotation.userid) and socket.request.authenticated_userid != annotation.userid:
    return None
```

Side note: this is still pretty ugly and I think maybe we should be thinking about rolling the NIPSA table into our user table so that what actually do is more like

- add an `annotations` relation on the `User` model with a `backref`
- then check `annotation.user.is_nipsa_flagged` (or whatever) rather than having to use `h.nipsa.logic.has_nipsa`

but I think that is definitely a job for another day.

chdorner: Yes, that's better. For some reason I though I'm still in `h.api` and shouldn't use the nipsa package directly.

h/streamer/messages.py
chdorner: Hmm, just wanted to implement this and saw that it's going to be a bit more complicated for the `/stream` page. We do add the same filters as the page has (user, group, tag, etc.). It looks like the only viable solution for this is what we talked about not deleting annotations immediately but marking them as deleted and then later on actually removing them from the database. Which is quite a bit of work. Can you think of another solution?

nickstenning: Hmm. Does the filter code really require an annotation object, or is a dictionary good enough? If the latter, we could keep sending the serialized annotation only for deletes as a temporary measure. It's ugly as hell, but...

h/streamer/messages.py
```diff
@@ -91,33 +93,48 @@ def handle_annotation_event(message, socket):
...
+    id_ = message['annotation_id']
+
+    # Return early when action is delete
+    if action == 'delete':
+        notification['payload'] = [{'id': id_}]
+        return notification
+
+    annotation = storage.fetch_annotation(socket.request.db, id_)
+
+    if annotation is None:
```
nickstenning: We should probably ensure this case (receive message for annotation that no longer exists in the database) is covered by the tests.

h/streamer/messages.py
chdorner: Okay, it is now sending the serialized annotation again (only for delete events) and I've had to rewrite parts of the `streamer.messages.handle_annotation_message`.

h/streamer/messages.py
chdorner: I've rewritten parts of this function and decided to actually remove this case. The case where it is not a `delete` event and it fails to load the annotation from the database is an unexpected error and should result in a exception that we will get notified about.

nickstenning: Yeah, unfortunately that's not true. If I issue an update and a delete in quick succession it's perfectly possible that this function won't process the update until after the annotation ha been deleted.

chdorner: That is a very good point. I just put that back and added a test for it.

nickstenning:
LGTM.

PR #3379, opened 2016-05-27
Move canonical annotation list sort state to Redux store

robertknight:
This is a step towards getting us to a place where all important state in the front-end app is managed in the same way - kept in the Redux store and modified only via actions.

The sort state was duplicated as a property on the $scope and also
stored in the Redux store.

Instead just expose functions on the scope which retrieve this state
directly from the store and dispatch an action to the store when the
user selects a different sort mode.

### DIFF
--- a/h/static/scripts/annotation-ui.js
+++ b/h/static/scripts/annotation-ui.js
@@ -52,7 +52,10 @@ function initialState(settings) {
 
     filterQuery: null,
 
-    sortMode: 'Location',
+    // Key by which annotations are currently sorted.
+    sortKey: 'Location',
+    // Keys by which annotations can be sorted.
+    sortKeysAvailable: ['Newest', 'Oldest', 'Location'],
   });
 }
 
@@ -66,7 +69,7 @@ var types = {
   REMOVE_ANNOTATIONS: 'REMOVE_ANNOTATIONS',
   CLEAR_ANNOTATIONS: 'CLEAR_ANNOTATIONS',
   SET_FILTER_QUERY: 'SET_FILTER_QUERY',
-  SORT_BY: 'SORT_BY',
+  SET_SORT_KEY: 'SET_SORT_KEY',
 };
 
 function excludeAnnotations(current, annotations) {
@@ -116,8 +119,8 @@ function reducer(state, action) {
         forceVisible: {},
         expanded: {},
       });
-    case types.SORT_BY:
-      return Object.assign({}, state, {sortMode: action.mode});
+    case types.SET_SORT_KEY:
+      return Object.assign({}, state, {sortKey: action.key});
     default:
       return state;
   }
@@ -301,11 +304,11 @@ module.exports = function (settings) {
       });
     },
 
-    /** Sets the sort mode for the annotation list. */
-    sortBy: function (mode) {
+    /** Sets the sort key for the annotation list. */
+    setSortKey: function (key) {
       store.dispatch({
-        type: types.SORT_BY,
-        mode: mode,
+        type: types.SET_SORT_KEY,
+        key: key,
       });
     },
   };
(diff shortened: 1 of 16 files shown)


h/static/scripts/app-controller.js
```diff
@@ -79,31 +80,6 @@ module.exports = function AppController(
     }
   });
 
-  $scope.$watch('sort.name', function (name) {
```
robertknight: List code here was already dead and replaced by logic in `root-thread.js`

h/static/scripts/test/app-controller-test.js
```diff
@@ -292,24 +293,4 @@ describe('AppController', function () {
       assert.equal(fakeWindow.confirm.callCount, 0);
     });
   });
-
-  describe('sorting', function () {
```
robertknight: This code was already dead and replaced by tests in `root-thread-test.js`

h/static/scripts/annotation-ui.js
```diff
@@ -96,8 +100,23 @@ function annotationsReducer(state, action) {
   }
 }
 
+function sortReducer(state, action) {
+  switch (action.type) {
+    case types.SORT_BY:
+      return Object.assign({}, state, {name: action.mode});
+    case types.SET_SORT_OPTIONS:
+      return Object.assign({}, state, {
+        name: action.mode,
```
nickstenning: It would be nice if we could take this opportunity to unify the naming here. We're calling it "name" in the object, but "mode" in the action. And "name" is a pretty odd choice, IMO. Perhaps refer to it as the sort "key" everywhere, and rename "options" to "availableKeys"?

h/static/scripts/stream-controller.coffee
```diff
@@ -64,9 +60,9 @@ module.exports = class StreamController
       };
     );
 
+    annotationUI.setSortOptions(['Newest', 'Oldest'], 'Newest');
```
nickstenning: You know, I've never noticed that you could sort the stream. It's a pretty useless feature, as it sorts the fetched cards, not the query. Perhaps consult with @conordelahunty on whether we even want to retain this? If not, we could get rid of all the extra code for changing the set of available sort keys.

nickstenning: We have agreement from @conordelahunty that this can go.

nickstenning:
Since you're busy doing more interesting things, @robertknight, I've taken the liberty of updating this broadly in line with my comments above. Namely:

- All the naming is now consistent across the application
- The ability to sort the stream has gone as per discussion with Conor
- The ability to change the list of `sortKeysAvailable` is gone, because it's not needed any more
- The option for the `sortDropdown` directive to appear as anything other than an icon has gone -- let me know if I missed any CSS here

h/static/scripts/annotation-ui.js
```diff
@@ -52,7 +52,10 @@ function initialState(settings) {
 
     filterQuery: null,
 
-    sortMode: 'Location',
+    // Key by which annotations are currently sorted.
+    sortKey: 'Location',
+    // Keys by which annotations can be sorted.
+    sortKeysAvailable: ['Newest', 'Oldest', 'Location'],
```
nickstenning: I've turned this from an object into a couple of top-level keys. Flat structures are much easier to wrap your head around than nested ones. Worth keeping in mind, I think.

robertknight: As the code is right now that makes sense. One reason that it can be useful to use objects to group fields as the state gets more complex is that makes it easier to split up the main `(state, action) => nextState` update function into smaller ones which each take a specific part of the existing state as input and return the new version of that piece as a result.

robertknight:
The revised version of this LGTM.

PR #3381, opened 2016-05-30
Make "/app" endpoints relative to serviceUrl

nickstenning:
As reported by Donny Donovan from EJP, requests to "/app*" related endpoints weren't correctly being constructed relative to the configured "serviceUrl", which makes the client less portable.

This change removes a leading slash from the path supplied to the URL constructor, ensuring that if "serviceUrl" is not at the root of a domain, the "/app*" paths will still be correctly constructed.


robertknight:
LGTM

PR #3383, opened 2016-05-30
Bump SQLAlchemy dependency

robertknight:
SQLAlchemy >= 1.0.9 is required for the Query.one_or_none() method.

Bumping to v1.0.13 as that is the version I've actually tested.


nickstenning:
👍

PR #3384, opened 2016-05-30
Ensure newest annotations get reindexed first

chdorner:
Users are more likely to notice syncing issues with their latest annotations
rather than the old ones.
This basically just adds a `ORDER BY updated desc` at the end of the query and
shuffles the calculation of the time windows around so that the start of a window
is always before the end of the window.


nickstenning:
LGTM.

PR #3385, opened 2016-05-30
Fix client on pages that define a variable called `global`

### LINKED ISSUE #2723: Annotator.Util undefined on pages that define a variable called `global`
**See https://github.com/hypothesis/h/issues/2397#issuecomment-126376613 for a minimal repro case of this issue**

Launching H via the extension or bookmarklet fails on http://www.thebookseller.com/futurebook/publishing-has-entered-new-and-different-era with a 'cannot read `Util` of undefined` error or variation on that.

```js
        var Annotator = require("annotator");
        var g = Annotator.Util.getGlobal(); // 'Annotator' is undefined here
```

I'm sure I investigated something similar before that was related to a site defining its own version of 'global' or something similar.

robertknight:
Modules which assume the existence of a global variable called "global"
(which exists in Node) are wrapped by Browserify during the build which
normally results in "global" being aliased to "self" or "window" in a
browser environment.

If code on the page into which H is loaded defines a global variable
called "global" however, that gets used instead and this can break such
modules. In the case of https://www.civilsprep.com, "global" is a
reference to a DOM node for example.

This commit fixes the issue by only aliasing "global" (as seen by the
module using it) to either "window" or "self" and not to any existing
variable called "global".

Fixes #2723

### DIFF
--- a/scripts/gulp/create-bundle.js
+++ b/scripts/gulp/create-bundle.js
@@ -86,7 +86,18 @@ module.exports = function createBundle(config, buildOpts) {
       'querystring',
     ],
     insertGlobalVars: {
+      // The Browserify polyfill for the `Buffer` global is large and
+      // unnecessary, but can get pulled into the bundle by modules that can
+      // optionally use it if present.
       Buffer: undefined,
+      // Override the default stub for the `global` var which defaults to
+      // the `global`, `self` and `window` globals in that order.
+      //
+      // This can break on web pages which provide their own definition of
+      // `global`. See https://github.com/hypothesis/h/issues/2723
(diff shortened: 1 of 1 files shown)


robertknight:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;c4iARNt9&#x2F;333-bug-annotator-util-undefined-on-pages-that-define-a-variable-called-global"> [BUG] Annotator.Util undefined on pages that define a variable called `global`</a>

chdorner:
LGTM. Tested it on the page and injecting the client works as expected now :+1:

PR #3388, opened 2016-05-30
Get groups documents from postgres

seanh:


seanh:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;pTuDjUPh&#x2F;351-postgres-cleanup-groups-page"> Postgres cleanup: Groups page</a>

seanh:
This changes the group read page to get the list of the group's annotated documents from Postgres instead of from Elasticsearch.

It eager loads the `Document` and `DocumentMeta` objects so it gets them all in one SQL query.

I don't think it's the _most_ efficient query, since I think it's loading the entire `Annotation`, `Document` and `DocumentMeta` when not all of their fields are really necessary. It minimises the number of SQL queries but it doesn't minimise the size of the SQL result. This was the easiest way since I wanted to reuse the existing `AnnotationHTMLPresenter` and its `.document_link` property. It _could_ be optimised further if necessary.

`groups/views_test.py` is changed to use the real db since there's now a rather complex sqlalchemy query in `views.py` (maybe this query should move into `models.py`).

The way it gets the list of the group's annotated documents is a little odd. It follows how we did it before with Elasticsearch. What it actually gets is the 25 most recently updated _annotations_ with unique `.target_uri_normalized` properties, then it uses `AnnotationHTMLPresenter.document_link` on those. I don't think this is quite right, I think the Postgres model now allows for two completely different `DocumentURI`s (with different `.uri_normalized` properties) to point to the same `Document`.

So I think that what we _really_ need to do here is get the group's 25 most recently annotated `Document`s and compute the document links from those. We need a `DocumentHTMLPresenter` not an `AnnotationHTMLPresenter`.

seanh:
Something like this I think:

```
    documents = (
        request.db.query(models.Document)
        .options(joinedload(models.Document.meta))
        .filter(models.Annotation.shared == expression.true())
        .filter(models.Annotation.groupid == group.pubid)
        .filter(models.Annotation.target_uri_normalized
                == models.DocumentURI.uri_normalized)
        .filter(models.DocumentURI.document_id == models.Document.id)
        .order_by(models.Annotation.created)
        [:25]
    )
```

Which unfortunately reimplements the `Annotation.target_uri_normalized == DocumentURI.uri_normalized` -> `DocumentURI.document_id == Document.id` relationship (already implemented in the `Annotation.document` property) to get an annotation's document.

And then `document_link()` will have to be moved from `AnnotationHTMLPresenter` onto a new `DocumentHTMLPresenter`.

h/groups/views.py
```diff
@@ -101,26 +103,22 @@ def _read_group(request, group):
...
-                break
-
-    document_links = [annotation.document_link
-                      for annotation in annotations_by_uri.values()]
+    annotations = (
+        request.db.query(models.Annotation)
+        .options(
+            joinedload(models.Annotation.document)
+            .joinedload(models.Document.meta))
+        .filter(models.Annotation.shared == expression.true())
```
nickstenning:
```python
.filter_by(shared=True, groupid=group.pubid)
```
would probably be a bit clearer here.

seanh: Oh yeah, I did have more than one class in the `request.db.query(models.Annotation)` bit at one point which is why it seemed like I had to use `filter()`. It can be less verbose now

h/groups/views.py
```diff
@@ -101,26 +103,21 @@ def _read_group(request, group):
...
-            annotations_by_uri[normalized_uri] = annotation
-            if len(annotations_by_uri) >= 25:
-                break
-
-    document_links = [annotation.document_link
-                      for annotation in annotations_by_uri.values()]
+    documents = (
+        request.db.query(models.Document)
+        .options(joinedload(models.Document.meta))
+        .filter(models.Annotation.shared == expression.true())
```
nickstenning: Not sure how this can possibly work if you haven't told it how to map a document to an annotation...?

seanh: It does seem to work. There's a couple of filters below this line that achieve that mapping:

```python
.filter(models.Annotation.target_uri_normalized == models.DocumentURI.uri_normalized)
.filter(models.DocumentURI.document_id == models.Document.id)
```

It doesn't seem to be necessary for these filters to be stated _before_ the shared annotations filter (and iirc from the days when I used to actually write SQL queries, that rings true to me).

seanh: For readability those should probably be changed to map a document to an annotation and not vice-versa now (though of course it sends up actually doing the same thing)

h/groups/views.py
```diff
@@ -101,26 +104,27 @@ def _read_group(request, group):
...
+        .filter(
+            models.Document.annotations.any(
+                and_(
+                    models.Annotation.shared == expression.true(),
+                    models.Annotation.groupid == group.pubid
+                )
+            )
+        )
+        .order_by(models.Annotation.created)
+        [:25]
```
nickstenning: `.limit(25)` would be nicer. I don't think `__getitem__` is smart enough to do this on the server.

nickstenning: Hmm. Maybe it is...

nickstenning:
This is looking pretty good. Any particular reason it's still marked WIP?

seanh:
Closed in favor of https://github.com/hypothesis/h/pull/3409

PR #3392, opened 2016-05-31
Load annotations from DB in RSS/Atom feeds

chdorner:
This is WIP and blocked because I need the `AnnotationHTMLPresenter` changes from #3388.

### DIFF
--- a/h/api/storage.py
+++ b/h/api/storage.py
@@ -49,6 +49,17 @@ def fetch_annotation(session, id_):
         return None
 
 
+def fetch_ordered_annotations(session, ids):
+    if not ids:
+        return []
+
+    ordering = {x: i for i, x in enumerate(ids)}
+
+    anns = session.query(models.Annotation).filter(models.Annotation.id.in_(ids)).all()
+    anns = sorted(anns, key=lambda a: ordering.get(a.id))
+    return anns
+
+
 def create_annotation(request, data):
     """
     Create an annotation from passed data.
--- a/h/feeds/util.py
+++ b/h/feeds/util.py
@@ -20,4 +20,4 @@ def tag_uri_for_annotation(annotation, annotation_url):
     domain = urlparse.urlparse(annotation_url(annotation)).hostname
     return u"tag:{domain},{date}:{id_}".format(domain=domain,
                                                date=FEED_TAG_DATE,
-                                               id_=annotation["id"])
+                                               id_=annotation.id)
--- a/h/feeds/atom.py
+++ b/h/feeds/atom.py
@@ -42,9 +42,6 @@ def _feed_entry_from_annotation(
             {"rel": "alternate", "type": "application/json",
              "href": annotation_api_url(annotation.annotation)}
         )
-    # TODO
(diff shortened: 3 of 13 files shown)


h/feeds/util.py
```diff
@@ -1,11 +1,27 @@
...
 
 # See RFC4151 for details of the use and format of the tag date:
 #
 #   https://tools.ietf.org/html/rfc4151#section-2.1
 FEED_TAG_DATE='2015-09'
 
+ZERO = timedelta(0)
+
+
+class UTC(tzinfo):
```
nickstenning: I'd suggest that instead of converting a naive datetime into a timezone-aware datetime, just change the format.

nickstenning: That is, hard code `Z` or `+0000` rather than using `%Z`.

h/api/storage.py
```diff
@@ -49,6 +49,15 @@ def fetch_annotation(session, id_):
         return None
 
 
+def fetch_ordered_annotations(session, ids):
+    if not ids:
+        return []
+
+    anns = session.query(models.Annotation).filter(models.Annotation.id.in_(ids)).all()
+    anns = sorted(anns, key=lambda a: ids.index(a.id))
```
nickstenning: This is going to be crazy slow if the list is too large, because `.index` is an O(N) operation.

Probably way better to built a lookup table first as:

```python
ordering = {x: i for i, x in enumerate(ids)}
```

and then use

```python
anns = sorted(anns, key=ordering.get)
```

nickstenning: FWIW:

```python
In [1]: timeit.timeit('sorted(anns, key=order.index)', setup='import random; anns=range(200); order=range(200); random.shuffle(order)', number=10000)
Out[1]: 4.671551942825317

In [2]: timeit.timeit('orderdict={k: i for i, k in enumerate(order)}; sorted(anns, key=orderdict.get)', setup='import random; anns=range(200); order=range(200); random.shuffle(order)', number=10000)
Out[2]: 0.7863879108428955
```

h/feeds/util.py
chdorner: Sure, if that's good enough then why not. Had to do a bit more work for atom feeds since they just used the `isoformat()` method, but the `UTC` class and converting naive to aware datetimes is gone now.

h/api/storage.py
chdorner: Yeah you're right. Especially since it has quite a dramatic impact even with low numbers like 200.

chdorner:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;kMNdNHWM&#x2F;352-postgres-cleanup-rss-atom-feeds"> Postgres cleanup: RSS&#x2F;Atom feeds</a>

chdorner:
Closing this because it can't go into master yet as it would break the groups page. @seanh and me decided to merge this and #3388 into a separate branch which then gets merged into master in one go.

chdorner:
New PR is here: #3408 3408

PR #3395, opened 2016-06-01
Use the "HTML" link for the annotation timestamp

nickstenning:
The timestamp wasn't originally specified to be switched to the "in-context" link, so this commit makes sure that it always points to the "HTML" representation of the annotation, if that link is available.

### DIFF
--- a/h/static/scripts/directive/test/annotation-test.js
+++ b/h/static/scripts/directive/test/annotation-test.js
@@ -1424,30 +1424,54 @@ describe('annotation', function() {
       });
     });
 
-    describe('annotation metadata', function () {
-      function findLink(directive) {
-        var links = directive.element[0]
-          .querySelectorAll('header .annotation-header__timestamp');
-        return links[links.length-1];
-      }
-
-      it('displays HTML links when in-context links are not available', function () {
-        var annotation = Object.assign({}, fixtures.defaultAnnotation(), {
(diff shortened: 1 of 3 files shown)


nickstenning:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;SmrrbhlV&#x2F;360-bug-timestamp-on-cards-should-link-to-the-permalink-not-the-direct-link"> [Bug] Timestamp on cards should link to the permalink, not the direct link.</a>

h/static/scripts/directive/test/annotation-test.js
```diff
@@ -1424,30 +1424,53 @@ describe('annotation', function() {
...
+
+        assert.equal(controller.linkInContext, annotation.links.html);
       });
 
-      it('displays in-context links when available', function () {
+      it('linkHTML uses the HTML link when available', function () {
         var annotation = Object.assign({}, fixtures.defaultAnnotation(), {
           links: {
             html: 'https://test.hypothes.is/a/deadbeef',
-            incontext: 'https://hpt.is/deadbeef'
```
seanh: Do we want to test that `linkHTML` is always the HTML link, even when `incontext` is available?

nickstenning: Good idea.

PR #3400, opened 2016-06-01
Remove additional margin at foot of annotation card

### LINKED ISSUE #3387: Top and bottom margins not the same on cards
### Steps to reproduce
1. Open a card (on master)

### Expected behaviour
Margins should be the same (I believe).

### Actual behaviour
Lower one is larger.

screenshot_2016-05-30_09_44_21

nickstenning:
There should be 15px, not 20px of margin at the bottom of the annotation card, just as at the top.

Removing this CSS means that the annotation footer has no bottom margin and the 15px is provided for all sizes of the card by padding on .annotation-card.

Fixes #3387.

### DIFF
--- a/h/static/styles/annotation.scss
+++ b/h/static/styles/annotation.scss
@@ -108,9 +108,6 @@ $annotation-card-left-padding: 10px;
 .annotation-footer {
   @include font-normal;
   color: $grey-5;
-  // Margin between bottom of ascent of annotation footer labels
-  // and bottom of annotation card should be ~20px.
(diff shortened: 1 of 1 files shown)


nickstenning:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;4yQ7AHK4&#x2F;366-bug-annotation-card-margin"> [BUG] Annotation card margin</a>

chdorner:
LGTM.

PR #3402, opened 2016-06-02
Fix distracting 404s in karma/phantom test output

nickstenning:
For the longest time our tests have spat out a couple of ugly warnings when running:

    02 06 2016 15:19:20.485:WARN [web-server]: 404: /undefined?
    02 06 2016 15:19:20.488:WARN [web-server]: 404: /undefined?

This has annoyed me for ages so I decided to work it out. Turns out the annotator Host tests were injecting an iframe with a `src` property of `/undefined?`, which resulted in the request from Phantom to the Karma server. This PR just adds a blank file which can be served by Karma for this purpose.

Running these tests also failed in a real browser (Chrome) because we were doing `.indexOf` on an object which was undefined. d664dff fixes this issue.

### DIFF
--- a/h/static/scripts/annotator/host.coffee
+++ b/h/static/scripts/annotator/host.coffee
@@ -6,10 +6,14 @@ Guest = require('./guest')
 
 module.exports = class Host extends Guest
   constructor: (element, options) ->
+    # Make a copy of all options except `options.app`, the app base URL.
     appOpts = queryString.stringify(
       Object.assign({}, options, {app: undefined})
     )
-    options.app += (if '?' in options.app then '&' else '?') + appOpts
+    if options.app and '?' in options.app
+      options.app += '&' + appOpts
+    else
+      options.app += '?' + appOpts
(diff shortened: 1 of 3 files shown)


h/static/scripts/annotator/host.coffee
```diff
@@ -6,10 +6,14 @@ Guest = require('./guest')
...
   constructor: (element, options) ->
+    # Make a copy of all options except `options.app`, the app base URL.
     appOpts = queryString.stringify(
       Object.assign({}, options, {app: undefined})
     )
-    options.app += (if '?' in options.app then '&' else '?') + appOpts
+    if options.app and '?' in options.app
+      options.app += '&' + appOpts
+    else
+      options.app += '?' + appOpts
```
robertknight: We should just throw an error here if options.app is undefined, since the result will end up being `undefined?...` otherwise.

robertknight:
LGTM

PR #3403, opened 2016-06-02
Pg cleanup feeds fix annotation html presenter

seanh:


seanh:
There's a lot of complexity in `AnnotationHTMLPresenter` that's to do with the variability of the legacy Elastic annotations, I guess all that can be simplified now..

seanh:
@chdorner I think there's probably more that can be done to clean up the feeds code and their tests now that we don't have legacy annotations, but for now I think this should be enough to get your PR unstuck.

h/presenters.py
```diff
@@ -128,14 +125,13 @@ def title(self):
 
         """
         document_ = self.annotation.document
```
chdorner: `annotation.document` can be `None` with the new PG setup. We should probably add a check for this.

h/presenters.py
```diff
@@ -128,14 +125,13 @@ def title(self):
...
-        if document_:
-            title = document_.title
-            if title:
-                # Convert non-string titles into strings.
-                # We're assuming that title cannot be a byte string.
-                title = text_type(title)
+        title = document_.title
```
chdorner: `annotation.document` can be `None` with the new PG setup. We should probably add a check for this.

chdorner:
Looking good. Two things that might cause problems in the future that we should probably discuss.

h/api/models/annotation.py
```diff
@@ -86,6 +86,17 @@ class Annotation(Base, mixins.Timestamps):
                                    viewonly=True,
                                    uselist=False)
 
+    @property
+    def target_links(self):
+        """A list of the URLs to this annotation's targets."""
+        links = []
+        for target_selector in self.target_selectors:
+            source = target_selector.get("source")
```
seanh: Ah, it looks like the `target` object submitted by the client is a list of dicts with `selector` and `source` fields. These used to all be stored in legacy Es, and `h.api.models.elastic.Annotation.target_links` returns a list of all these `source` fields. But _now_ we discard all targets other than the first, and we discard all fields of the first target other than `selector`. So `source` is discarded and not stored in Postgres, even though the client still sends it.

I guess we should just delete this `target_links` property and remove the code from feeds that uses it

seanh: Ok: https://github.com/hypothesis/h/pull/3403/commits/fa013ff4f7e8da8683ad55a1c6491b635b37774a

seanh: Rebased this into the history now

h/presenters.py
seanh: Hmm, are annotations with no document something we want to support?

But for now I've rebased out the commit that removed this `if document_` test

seanh:
Both done @chdorner

chdorner:
LGTM. Thanks for this!

PR #3404, opened 2016-06-02
Use PDF fingerprint as primary search URI for PDFs instead of url.

sheetaluk:
https://trello.com/c/DVUemKwi/329-use-pdf-fingerprint-as-primary-search-uri-for-pdfs

### DIFF
--- a/h/static/scripts/annotator/plugin/pdf-metadata.js
+++ b/h/static/scripts/annotator/plugin/pdf-metadata.js
@@ -1,0 +1,68 @@
+'use strict';
+
+/**
+ * This PDFMetadata service extracts metadata about a loading/loaded PDF
+ * document from a PDF.js PDFViewerApplication object.
+ *
+ * This hides from users of this service the need to wait until the PDF document
+ * is loaded before extracting the relevant metadata.
+ */
+function PDFMetadata(app) {
+  this._loaded = new Promise(function (resolve) {
+    var finish = function () {
+      window.removeEventListener('documentload', finish);
+      resolve(app);
+    };
+
+    if (app.documentFingerprint) {
+      resolve(app);
+    } else {
+      window.addEventListener('documentload', finish);
+    }
+  });
+}
+
+/**
+ * Returns a promise of the URI of the loaded PDF.
+ */
+PDFMetadata.prototype.getUri = function () {
+  return this._loaded.then(function (app) {
+    return app.url;
+  });
+};
+
+/**
+ * Returns a promise of a metadata object, containing:
+ *
+ * title(string) - The document title
+ * link(array) - An array of link objects representing URIs for the document
+ * documentFingerprint(string) - The document fingerprint
+ */
+PDFMetadata.prototype.getMetadata = function () {
+  return this._loaded.then(function (app) {
+    var title = document.title;
+
+    if (app.metadata && app.metadata.has('dc:title') && app.metadata.get('dc:title') !== 'Untitled') {
+      title = app.metadata.get('dc:title');
+    } else if (app.documentInfo && app.documentInfo.Title) {
+      title = app.documentInfo.Title;
+    }
+
+    var link = [
+      {href: fingerprintToURN(app.documentFingerprint)},
+      {href: app.url}
+    ];
+
+    return {
+      title: title,
+      link: link,
+      documentFingerprint: app.documentFingerprint,
+    };
+  });
+};
+
+function fingerprintToURN(fingerprint) {
+  return 'urn:x-pdf:' + String(fingerprint);
+}
+
+module.exports = PDFMetadata;
--- a/h/static/scripts/annotator/plugin/pdf-metadata.js
+++ b/h/static/scripts/annotator/plugin/pdf-metadata.js
@@ -27,7 +27,7 @@ function PDFMetadata(app) {
  */
 PDFMetadata.prototype.getUri = function () {
   return this._loaded.then(function (app) {
-    return app.url;
+    return fingerprintToURN(app.documentFingerprint);
   });
 };
 
--- a/h/static/scripts/annotator/plugin/test/pdf-metadata-test.js
+++ b/h/static/scripts/annotator/plugin/test/pdf-metadata-test.js
@@ -1,0 +1,82 @@
+'use strict';
+
+var PDFMetadata = require('../pdf-metadata');
+
+describe('pdf-metadata', function () {
+  describe('get pdf metadata when documentload event fires', function () {
+    it('should get pdf urn', function () {
+      var fakeApp = {};
+      var pdfMetadata = new PDFMetadata(fakeApp);
+
+      var event = document.createEvent('Event');
+      event.initEvent('documentload', false, false);
+      fakeApp.url = 'http://example.com/foo.pdf';
+      window.dispatchEvent(event);
+
+      return pdfMetadata.getUri().then(function (uri) {
+        assert.equal('http://example.com/foo.pdf', uri);
+      });
+    });
+  });
+
+  describe('get pdf metadata when documentFingerprint is set', function () {
+    var pdfMetadata;
+    var fakePDFViewerApplication = {
+      documentFingerprint: 'fakeFingerprint',
+      documentInfo: {
+        Title: 'fakeTitle',
+      },
+      metadata: {
+        metadata: {
+          'dc:title': 'fakeTitle',
+        }
+      },
+      url: 'fakeUrl',
+    };
+
+    beforeEach(function () {
+      pdfMetadata = new PDFMetadata(fakePDFViewerApplication);
+    });
+
+    describe('#getUri', function () {
+      it('returns the PDF URL as its URI', function () {
+        return pdfMetadata.getUri().then(function (uri) {
+          assert.equal('fakeUrl', uri);
+        });
+      });
+    });
+
+    describe('#getMetadata', function () {
+      it('should get a metadataobject with dc:title', function () {
+        var expectedMetadata = {
+          title: 'dcTitle',
+          link: [{href: 'urn:x-pdf:' + fakePDFViewerApplication.documentFingerprint},
+            {href: fakePDFViewerApplication.url}],
+          documentFingerprint: fakePDFViewerApplication.documentFingerprint,
+        };
+
+        fakePDFViewerApplication.metadata.has = sinon.stub().returns(true);
+        fakePDFViewerApplication.metadata.get = sinon.stub().returns('dcTitle');
+
+        return pdfMetadata.getMetadata().then(function (actualMetadata) {
+          assert.deepEqual(expectedMetadata, actualMetadata);
+        });
+      });
+
+      it('should get a metadataobject with documentInfo.Title', function () {
+        var expectedMetadata = {
+          title: fakePDFViewerApplication.documentInfo.Title,
+          link: [{href: 'urn:x-pdf:' + fakePDFViewerApplication.documentFingerprint},
+            {href: fakePDFViewerApplication.url}],
+          documentFingerprint: fakePDFViewerApplication.documentFingerprint,
+        };
+
+        fakePDFViewerApplication.metadata.has = sinon.stub().returns(false);
+
+        return pdfMetadata.getMetadata().then(function (actualMetadata) {
+          assert.deepEqual(expectedMetadata, actualMetadata);
+        });
+      });
+    });
+  });
+});
--- a/h/static/scripts/annotator/plugin/test/pdf-metadata-test.js
+++ b/h/static/scripts/annotator/plugin/test/pdf-metadata-test.js
@@ -3,23 +3,29 @@
 var PDFMetadata = require('../pdf-metadata');
 
 describe('pdf-metadata', function () {
-  describe('get pdf metadata when documentload event fires', function () {
-    it('should get pdf urn', function () {
-      var fakeApp = {};
-      var pdfMetadata = new PDFMetadata(fakeApp);
+  it('waits for the PDF to load before returning metadata', function () {
+    var fakeApp = {};
+    var pdfMetadata = new PDFMetadata(fakeApp);
 
-      var event = document.createEvent('Event');
-      event.initEvent('documentload', false, false);
-      fakeApp.documentFingerprint = 'fakeFingerprint';
-      window.dispatchEvent(event);
+    var event = document.createEvent('Event');
+    event.initEvent('documentload', false, false);
+    fakeApp.documentFingerprint = 'fakeFingerprint';
+    window.dispatchEvent(event);
 
-      return pdfMetadata.getUri().then(function (uri) {
-        assert.equal('urn:x-pdf:fakeFingerprint', uri);
-      });
+    return pdfMetadata.getUri().then(function (uri) {
+      assert.equal(uri, 'urn:x-pdf:fakeFingerprint');
     });
   });
 
-  describe('get pdf metadata when documentFingerprint is set', function () {
+  it('does not wait for the PDF to load if it has already loaded', function () {
+    var fakePDFViewerApplication = {documentFingerprint: 'fakeFingerprint'};
+    var pdfMetadata = new PDFMetadata(fakePDFViewerApplication);
+    return pdfMetadata.getUri().then(function (uri) {
+      assert.equal(uri, 'urn:x-pdf:fakeFingerprint');
+    });
+  });
+
+  describe('metadata sources', function () {
     var pdfMetadata;
     var fakePDFViewerApplication = {
       documentFingerprint: 'fakeFingerprint',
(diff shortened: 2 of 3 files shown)


seanh:
I think this has broken the `uri` field that the client sends to the API when searching for the annotations of the document

seanh:
I'm seeing `uri=urn%3Ax-pdf%3Aundefined` in the `http://localhost:5000/api/search` request when launching the sidebar

seanh:
I wonder if `PDFViewerApplication.documentFingerprint` is not defined yet at the time of the `documentload` event, and `PDF.uri` now needs to wait for some other, later event

sheetaluk:
@seanh I can't seem to be able to reproduce it. 
https://www.dropbox.com/s/xf0kea9c2sl30je/Screenshot%202016-06-02%2014.56.17.png?dl=0

Here are the 2 scenarios I tried:
I refreshed the page while H was active.
I opened a pdf and then enabled the H plugin.

robertknight:
Sheetal and me were discussing this on Slack yesterday. One of the questions we had was what the fingerprint was, where it comes from and what the risks of this change were.

For future reference, the fingerprint returned by PDF.js is a hex string which may come from one of two sources:

1. It may be the "Permanent File Identifier" which is embedded in the PDF by the software that creates it, when the PDF was created. See this part of the PDF specification: https://hyp.is/ehLvjij4EeabInuHElZdQQ/www.adobe.com/content/dam/Adobe/en/devnet/acrobat/pdfs/PDF32000_2008.pdf
2. If no such information exists in the PDF document, PDF.js will generate one using an undocumented and non-standardized algorithm which could change in future versions of PDF.js. The [current implementation](https://github.com/mozilla/pdf.js/commit/aa53319c875205ffe7cd8b41a0280e19445ca5fb) computes the MD5 hash of the first 1024 bytes of the file.

So there are some potential risks to consider:

1. If the algorithm for computing the digest changes in a future version of PDF.js, that would alter the fingerprint and all annotations would fail to anchor.
2. If there is PDF generating software out there which generates an ID but does so poorly, that could result in many different PDFs having the same ID.
3. If the PDF at a particular URL is updated in a way that doesn't preserve the ID, then currently (on master) annotations could still anchor as the annotations are associated with a PDF by URL rather than by content. If we switch to only using a fingerprint as the search query, the existing annotations will then fail to anchor.

nickstenning:
@robertknight Yep, that's all correct. In the long run we'll address issues like this by sending all known URIs for a document when retrieving annotations. In the meantime, this fixes a set of common failure modes where people share and annotate a PDF and are surprised that annotations don't load.

seanh:
Weird, I'm still seeing the above problem, `PDFViewerApplication.documentFingerprint` is `undefined` the first time it's accessed when the sidebar is sending the search API request. If you later create and save an annotation the exact same line of code that this PR modifies is called again and this time `PDFViewerApplication.documentFingerprint` is defined.

I seem to be able to be able to reproduce this reliably every time, so if it's a race condition it seems to almost always come out this way on my machine.

seanh:
This does look like a race though, if I stop it in the debugger at the point when `PDFViewerApplication.documentFingerprint` is first accessed and is undefined it looks as if the PDF viewer app has started but the PDF itself hasn't been rendered yet. I think there is some "the actual PDF file has finished rendering" event that we need to be waiting on instead of `documentload`.

seanh:
Ok, actually I think `documentload` might be the right event to wait for. The problem appears to be that, on my system, on [this line](https://github.com/hypothesis/h/blob/a1d8964967188171f0408bdf323750abe837513e/h/static/scripts/annotator/plugin/pdf.coffee#L16) `PDFViewerApplication.loading` is always `undefined` so we never actually wait for `documentload`. `@documentLoaded` is always simple initialised as `Promise.resolve()`.

This change makes it actually wait for `documentload` and sent the correct URI in the search, for me:

```diff
diff --git a/h/static/scripts/annotator/plugin/pdf.coffee b/h/static/scripts/annotator/plugin/pdf.coffee
index 547f51c..9fd6fcd 100644
--- a/h/static/scripts/annotator/plugin/pdf.coffee
+++ b/h/static/scripts/annotator/plugin/pdf.coffee
@@ -13,14 +13,11 @@ module.exports = class PDF extends Annotator.Plugin
     @pdfViewer = PDFViewerApplication.pdfViewer
     @pdfViewer.viewer.classList.add('has-transparent-text-layer')
 
-    if PDFViewerApplication.loading
-      @documentLoaded = new Promise (resolve) ->
-        finish = (evt) ->
-          window.removeEventListener('documentload', finish)
-          resolve()
-        window.addEventListener('documentload', finish)
-    else
-      @documentLoaded = Promise.resolve()
+    @documentLoaded = new Promise (resolve) ->
+      finish = (evt) ->
+        window.removeEventListener('documentload', finish)
+        resolve()
+      window.addEventListener('documentload', finish)
 
     @observer = new MutationObserver((mutations) => this._update())
     @observer.observe(@pdfViewer.viewer, {
```

I'm not sure what the point of this `if PDFViewerApplication.loading` that I've deleted was. There seems to be no problem with just waiting for the `documentload` event here.

But this is guess work, I don't know exactly `PDFViewerApplication.loading` is supposed to be or why we were switching on it here.

It never seems to happen on my system but is it possible that `documentload` could fire _before_ the code above is executed, and then it won't be fired again. so `finish()` would never get called? Or would `finish()` just be called immediately in that case?

seanh:
[The commit that introduced the `if PDFViewerApplication.loading`](d44df9045d5df25ad50d3bff74acc48945ca2cf1)

seanh:
I think [PDFViewerApplication.loading was removed from PDF.js](http://lists.alioth.debian.org/pipermail/pkg-javascript-commits/2015-August/017167.html)

chdorner:
fwiw: I'm getting this as the search url: `urn:x-pdf:9597c618bc90afa4a078ca72b2dd061c` for [this PDF](http://www.adobe.com/content/dam/Adobe/en/devnet/acrobat/pdfs/PDF32000_2008.pdf) (which is linked above) on Chromium 51.0.2704.63 (and Linux).

seanh:
I suspect it's a race condition that depends on how fast pdf.js loads the beginning of the pdf

dwhly-proj:
Riffing on @nickstenning remark: 
Should we search the fingerprint as the primary or the exclusive identifier?  In other words will we also search for annotations associated with the URL of the PDF?  (I wonder if PDFs that are watermarked by publishers, for instance journals that put the subscribers name in the margin of the document or use encoded pixels, will have different PDF fingerprints for a PDF at the same URL).  

Related: For existing PDF annotations, going back to the release of the feature, were we storing the PDF fingerprint in a way that is now available to us within our annotation db?

Related to @robertknight's observation about the PDFjs generated fingerprint is: I wonder the percentage of PDFs that we encounter on average have a fingerprint stored at birth.

sheetaluk:
Can reproduce the issue with the file attached:
image
[The Standardistas [2013] A Pocket Guide to The Craft of Words.pdf](https://github.com/hypothesis/h/files/297695/The.Standardistas.2013.A.Pocket.Guide.to.The.Craft.of.Words.pdf)

sheetaluk:
@seanh might be right. Looks like the line `PDFViewerApplication.loading` in h/h/static/scripts/annotator/plugin/pdf.coffee was introduced in May 2015. `pdf.js` was updated in August 2015, and removes `PDFViewerApplication.loading`.

Tested out the change on the attached two pdfs locally and remotely and confirmed it works.
[The Standardistas [2013] A Pocket Guide to The Craft of Words.pdf](https://github.com/hypothesis/h/files/297759/The.Standardistas.2013.A.Pocket.Guide.to.The.Craft.of.Words.pdf)
[poof.pdf](https://github.com/hypothesis/h/files/297760/poof.pdf)

chdorner:
Tested it again locally with the two attached PDFs.

Without 94e5da7:
* _poof.pdf_: worked
* _Standardistas..._: got `undefined` for the fingerprint

With 94e5da7:
* _poof.pdf_: worked
* _Standardistas..._: worked

:tada:

nickstenning:
Is there a reason this is still marked as WIP?

robertknight:
@nickstenning - Yes, some refactoring is currently being done to move metadata extraction out of guest.coffee to make it easier to test/alter.

@sheetaluk - I can see a potential issue with 94e5da76d61b as well. If the PDF loads before the Hypothesis code executes, then the `documentLoaded` promise will never resolve. We either need to guarantee that the H code executes first, or we'll need to find the replacement for `PDFViewerApplication.loading`

nickstenning:
A quick glance at PDF.js's [`viewer.js`] shows that `PDFViewerApplication.documentFingerprint` [is always set before](https://github.com/mbbaig/pdf.js/blob/d65ef81af500aa4ecd899b37efbc4139f8bd94a7/web/viewer.js#L819) the [`documentload` events are fired](https://github.com/mbbaig/pdf.js/blob/d65ef81af500aa4ecd899b37efbc4139f8bd94a7/web/viewer.js#L848).

So, to avoid any chance of a race condition, I'd suggest you do the following:

1. First, add a `documentload` listener.
2. Then, check if `documentFingerprint` is set. If so, you can resolve the promise.
3. Otherwise, resolve the promise when `documentload` triggers.

h/static/scripts/annotator/plugin/pdf-metadata.js
```diff
@@ -0,0 +1,73 @@
...
+    var metadata;
+    var urn;
+    var title;
+    var documentFingerprint;
+    var link = [];
+
+    info = this._PDFViewerApplication.documentInfo;
+    metadata = this._PDFViewerApplication.metadata;
+
+    if ((typeof metadata !== 'undefined' && metadata !== null ? metadata.has('dc:title') : void 0) &&
```
robertknight: I know this was ported from the original code but the convention we're using for more recently written code is just to use truthiness where possible to test for an object being null, so `if (metadata && metadata.has('dc:title')`. The rationale is this is more succinct and easier to read.

h/static/scripts/annotator/plugin/test/pdf-metadata-test.js
```diff
@@ -0,0 +1,73 @@
...
+    pdfMetadata = new PDFMetadataService(fakePDFViewerApplication);
+    pdfMetadata._documentLoaded = Promise.resolve();
+    sandbox = sinon.sandbox.create();
+  });
+
+  afterEach(function () {
+    sandbox.restore();
+  });
+
+  describe('.getUrn', function () {
```
robertknight: Nit. Should use `#name` (note leading '#') to refer to instance fields/methods in describe/it tests, following JSDoc conventions. Using a period denotes a class property.

h/static/scripts/annotator/plugin/test/pdf-metadata-test.js
```diff
@@ -0,0 +1,73 @@
+'use strict';
+
+var PDFMetadataService = require('../pdf-metadata');
```
robertknight: The name here should match the class name if possible, so just `PDFMetadata` rather than `PDFMetadataService`.

h/static/scripts/annotator/plugin/test/pdf-metadata-test.js
```diff
@@ -0,0 +1,73 @@
...
+      'Title': 'fakeTitle',
+    },
+    'metadata': {
+      'metadata': {
+        'dc:title': 'fakeTitle',
+      }
+    },
+    'url': 'fakeUrl',
+  };
+  var sandbox;
```
robertknight: Sandboxes are generally only needed if you are modifying the global environment in some way, eg. by using fake clocks or modifying global properties and later need an easy way to undo those changes after a test runs. If you just need to create a stub in order to test that a particular method was called you can use `sinon.stub()` rather than `sandbox.stub()`.

h/static/scripts/annotator/plugin/pdf-metadata.js
```diff
@@ -0,0 +1,73 @@
...
+    } else {
+      title = document.title;
+    }
+
+    urn = 'urn:x-pdf:' + this._PDFViewerApplication.documentFingerprint;
+    link = [{href: urn}, {href: this._PDFViewerApplication.url}];
+    documentFingerprint = this._PDFViewerApplication.documentFingerprint;
+
+    return {
+      'title': title,
```
robertknight: Nit: Keys in object literals should only be quoted if they contain special characters that are not legal in field names.

h/static/scripts/annotator/plugin/pdf-metadata.js
```diff
@@ -0,0 +1,73 @@
...
+ * #3. link - an array of two objects, the urn and the url strings of the pdf.
+ * #4. documentFingerprint - the string representing the pdf fingerprint.
+ */
+function PDFMetadata(PDFViewerApplication) {
+  this._PDFViewerApplication = PDFViewerApplication;
+  this._documentLoaded = new Promise(function (resolve) {
+    var finish;
+    finish = function (evt) {
+      window.removeEventListener('documentload', finish);
+      return resolve();
```
robertknight: You don't need to return the result of `resolve()` here, just call it.

h/static/scripts/annotator/plugin/test/pdf-metadata-test.js
```diff
@@ -0,0 +1,73 @@
...
+
+  beforeEach(function () {
+    pdfMetadata = new PDFMetadataService(fakePDFViewerApplication);
+    pdfMetadata._documentLoaded = Promise.resolve();
```
robertknight: Setting `_documentLoaded` directly means that we're not exercising the code that listens for the relevant event. 

You can simulate events with:

```
var event = document.createEvent('Event')
event.initEvent('documentload', false /* bubbles */, false /* cancelable */);
window.dispatchEvent(event);
```
(In an actual browser the first two lines can be replaced with just `var event = new Event(<event type>)` but that is not supported in PhantomJS)

h/static/scripts/annotator/plugin/pdf-metadata.js
```diff
@@ -0,0 +1,73 @@
...
+ * This PDFMetadata service provides implementations for extracting fields
+ * of a pdf document from the input PDFViewerApplication object.
+ *
+ * The fields extracted are:
+ * #1. urn - a string which uniquely identifies a pdf document.
+ * #2. title - a string representing the title of the pdf document.
+ * #3. link - an array of two objects, the urn and the url strings of the pdf.
+ * #4. documentFingerprint - the string representing the pdf fingerprint.
+ */
+function PDFMetadata(PDFViewerApplication) {
```
nickstenning: I'd suggest that you don't overload the global `PDFViewerApplication` with the name of the parameter. Call it `application` or even `app`.

h/static/scripts/annotator/plugin/test/pdf-metadata-test.js
nickstenning: TBH using a sandbox is good practice if you have state that lives outside a test method, as it makes it much harder to forget to tear it down.

nickstenning:
I'll just note that as written this still exhibits a potential race condition [as described by Robert](https://github.com/hypothesis/h/pull/3404#issuecomment-224295150). I'd still suggest a strategy like the one [I suggested](https://github.com/hypothesis/h/pull/3404#issuecomment-224302384).

h/static/scripts/annotator/plugin/pdf-metadata.js
```diff
@@ -0,0 +1,78 @@
...
+    finish = function () {
+      window.removeEventListener('documentload', finish);
+      resolve();
+    };
+
+    if (this._pdfViewerApplication.documentFingerprint) {
+      resolve();
+    }
+
+    window.addEventListener('documentload', finish);
```
nickstenning: This is fine, in the sense that resolving a promise multiple times should be safe, but it might be worth moving this into an `else` clause just to make the intent clearer.

h/static/scripts/annotator/plugin/pdf-metadata.js
```diff
@@ -0,0 +1,78 @@
...
+  }.bind(this));
```
nickstenning: Using `Function.bind` can get a bit messy if you have nested scopes. It's more common in our code base to use the closure pattern:

```javascript
var self = this;

function () {
    self.foo();
}
```

h/static/scripts/annotator/plugin/pdf-metadata.js
```diff
@@ -0,0 +1,78 @@
...
+};
+
+/**
+ * Returns a promise of an object with the fields
+ * title(string), link(array) and documentFingerprint(string)
+ * after the pdf loads succesfully and the 'documentload' event fires.
+ */
+PDFMetadata.prototype.getMetadata = function () {
+  return this._documentLoaded.then(function () {
+    var info;
```
robertknight: Where possible it is preferable to put the declaration and initialization on one line, eg. `var info = this._pdfViewerApplication.documentInfo` instead of `var info; info = this. ...`. There are two reasons for this: Fewer lines of code, and when we convert the codebase to ES6, we can use `const varName = <initializer>` to help guard against accidental re-assignments.

nickstenning:
I've made a couple of changes here for readability. I hope that's ok.

1. Rather than using `Function.prototype.bind` or `var self = this`, I've modified it so that the `PDFViewerApplication` is simply the resolved value of the promise.
2. I've split the PR into two commits. The first is just a refactoring which introduces the `PDFMetadata` service, while the second implements the functional change -- switching to fingerprint URNs rather then document URLs.

nickstenning:
Ok, shall we get this merged, @robertknight? LGTM1.

PR #3405, opened 2016-06-02
Remove path for local pdf files.

sheetaluk:
Currently, when annotating a local PDF, the `file://` URL will be submitted to the server. While submitting a document title or filename (basename) is appropriate, the whole file path is not.

https://trello.com/c/K8t2kl9m/331-don-t-save-file-urls-in-document-metadata


seanh:
This looks good to me. There are two more instances of the path remaining on this branch, but those will both be gone when https://github.com/hypothesis/h/pull/3404 is merged, so I suggest getting https://github.com/hypothesis/h/pull/3404 merged first then quickly re-testing and merging this.

nickstenning:
We already have the filename submitted as `document.title` for documents that don't have a title otherwise set, and entries in `links` should be URIs, so I think you can just omit it rather than modifying it.

PR #3407, opened 2016-06-03
Use new TokenAuthenticationPolicy and remove cruft

nickstenning:
This commit updates our primary AuthenticationPolicy so it delegates to one or other of a SessionAuthenticationPolicy or the new TokenAuthenticationPolicy depending on the request path. Both "upstream" authentication policies use the same "groupfinder" callback to provide a full list of effective principals for the authenticated user.

This allows to perform the following cleanups:

- h.auth.util.effective_principals turns into h.auth.util.groupfinder, and thus can be dramatically simplified: it no longer reimplements the default logic of Pyramid w.r.t. system principals.
- h.auth.util.group_principals is folded into groupfinder.
- h.auth.util.has_permission is no longer needed, with one minor change to the reply notification code.

This paves the way to further simplification and generalisation of the AuthenticationPolicy, so that we can make changing authentication for different routes in the application easily (in particular, I'm lining this up for the websocket endpoint.)

_**N.B.** This is based on #3406, and will want rebasing when that merges._


h/notification/reply.py
```diff
@@ -89,8 +89,8 @@ def get_notification(request, annotation, action):
         return
 
     # Don't send reply notifications to the author of the parent annotation if
-    # the author doesn't have permission to read the reply.
-    if not auth.has_permission(request, reply, parent.userid, 'read'):
+    # the reply was private.
+    if not reply.shared:
```
nickstenning: It's worth noting that this simplification is valid because of [this code](https://github.com/hypothesis/h/blob/3a2f303177d6a4883bb73d3c3fe23ac027336713/h/api/storage.py#L69-L81), which ensures that replies are always created in the same group as their parent.

seanh:
Rebased

tests/h/auth/util_test.py
```diff
@@ -1,173 +1,50 @@
...
+@pytest.mark.parametrize('user,principals', (
+    # User isn't found in the database: they're not authenticated at all
+    (None, None),
+    # User found but not staff, admin, or a member of any groups: no additional principals
+    (FakeUser(admin=False, staff=False, groups=[]),
+     []),
+    # User is admin: role.Admin should be in principals
+    (FakeUser(admin=True, staff=False, groups=[]),
+     [role.Admin]),
+    # User is staff: role.Staff§ should be in principals
```
seanh: §?

nickstenning: Ooops!

nickstenning: Fixed in 2388d0e.

PR #3408, opened 2016-06-03
Load annotations from DB in RSS/Atom feeds

chdorner:
_#3388 and this have depend on changes in the presenters, we should review these PRs and merge them into a new "pg-cleanup" branch, and then merge that one into master_

This is the second part of removing the usage of `storage.annotation_from_dict`. We now also load the annotations from the database after we search in the index for the Atom/RSS feeds.
Over time all searching should then load the data from the database, which will let us only keep searchable data in the index instead of the whole annotation object.
Sean also did some great changes to the `AnnotationHTMLPresenter` making it work with the annotation DB model.

### DIFF
--- a/h/feeds/rss.py
+++ b/h/feeds/rss.py
@@ -20,8 +20,7 @@ def _pubDate_string_from_annotation(annotation):
     element of an RSS feed.
 
     """
-    if annotation.created:
-        return annotation.created.strftime('%a, %d %b %Y %H:%M:%S +0000')
+    return annotation.created.strftime('%a, %d %b %Y %H:%M:%S +0000')
 
 
 def _feed_item_from_annotation(annotation, annotation_url):
@@ -71,7 +70,7 @@ def feed_from_annotations(annotations, annotation_url, rss_url, html_url,
             for annotation in annotations]
     }
 
-    if annotations and annotations[0].updated:
+    if annotations:
         feed['pubDate'] = annotations[0].updated.strftime('%a, %d %b %Y %H:%M:%S UTC')
 
     return feed
--- a/h/feeds/rss.py
+++ b/h/feeds/rss.py
@@ -1,6 +1,5 @@
 """Functions for generating RSS feeds."""
 from pyramid import i18n
-from dateutil import parser
 
 from h import presenters
 from h import util
@@ -21,8 +20,8 @@ def _pubDate_string_from_annotation(annotation):
     element of an RSS feed.
 
     """
-    return parser.parse(annotation['created']).strftime(
-        '%a, %d %b %Y %H:%M:%S %z')
+    if annotation.created:
+        return annotation.created.strftime('%a, %d %b %Y %H:%M:%S +0000')
 
 
 def _feed_item_from_annotation(annotation, annotation_url):
@@ -35,16 +34,15 @@ def _feed_item_from_annotation(annotation, annotation_url):
 
     """
     try:
-        name = util.user.split_user(annotation["user"])["username"]
+        name = util.user.split_user(annotation.userid)["username"]
     except ValueError:
-        name = annotation["user"]
+        name = annotation.userid
     return {
         "author": {"name": name},
         "title": annotation.title,
         "description": annotation.description,
         "pubDate": _pubDate_string_from_annotation(annotation),
-        "guid": h.feeds.util.tag_uri_for_annotation(
-            annotation, annotation_url),
+        "guid": h.feeds.util.tag_uri_for_annotation(annotation, annotation_url),
         "link": annotation_url(annotation)
     }
 
@@ -73,8 +71,7 @@ def feed_from_annotations(annotations, annotation_url, rss_url, html_url,
             for annotation in annotations]
     }
 
-    if annotations:
-        feed['pubDate'] = parser.parse(annotations[0]['updated']).strftime(
-            '%a, %d %b %Y %H:%M:%S %Z')
+    if annotations and annotations[0].updated:
+        feed['pubDate'] = annotations[0].updated.strftime('%a, %d %b %Y %H:%M:%S UTC')
 
     return feed
--- a/h/presenters.py
+++ b/h/presenters.py
@@ -62,9 +62,6 @@ class AnnotationHTMLPresenter(object):
     def __getattr__(self, attr):
         return getattr(self.annotation, attr)
 
-    def __getitem__(self, key):
-        return self.annotation[key]
-
     @property
     def uri(self):
         return jinja2.escape(self.annotation.target_uri)
--- a/h/presenters.py
+++ b/h/presenters.py
@@ -67,7 +67,7 @@ class AnnotationHTMLPresenter(object):
 
     @property
     def uri(self):
-        return jinja2.escape(self.annotation.uri)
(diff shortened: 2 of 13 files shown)


seanh:
This all looks fine to me but some of the commits are mine so I think @nickstenning or @finitud should review. If we can get this merged into `pg-cleanup` then I can re-do https://github.com/hypothesis/h/pull/3388 to be against `pg-cleanup` (and then once #3388 is merged into `pg-cleanup` then we can merge `pg-cleanup` into master)

h/feeds/rss.py
```diff
@@ -21,8 +20,8 @@ def _pubDate_string_from_annotation(annotation):
     element of an RSS feed.
 
     """
-    return parser.parse(annotation['created']).strftime(
-        '%a, %d %b %Y %H:%M:%S %z')
+    if annotation.created:
```
nickstenning: Is it possible for `annotation.created` to be falsy?

h/feeds/rss.py
```diff
@@ -73,8 +71,7 @@ def feed_from_annotations(annotations, annotation_url, rss_url, html_url,
             for annotation in annotations]
     }
 
-    if annotations:
-        feed['pubDate'] = parser.parse(annotations[0]['updated']).strftime(
-            '%a, %d %b %Y %H:%M:%S %Z')
+    if annotations and annotations[0].updated:
```
nickstenning: Ditto. If `annotations` is truthy, then so is `annotations[0]` (unless it can somehow be something other than an annotation), and `.updated` is not nullable. Overly defensive or am I missing something?

h/presenters.py
```diff
@@ -62,12 +62,9 @@ def __init__(self, annotation):
     def __getattr__(self, attr):
         return getattr(self.annotation, attr)
 
-    def __getitem__(self, key):
-        return self.annotation[key]
-
```
nickstenning: :+1: it would be nice to get rid of the `__getattr__` as well if possible.

h/feeds/rss.py
chdorner: This is mostly for easier testing, so we don't have to always have an object that returns a datetime for `updated` and `created`. I can remove it though if you feel strongly about it.

h/presenters.py
```diff
@@ -244,20 +241,13 @@ def description(self):
 
         """
         def get_selection():
-            targets = self.annotation.get("target")
-            if not isinstance(targets, list):
-                return
+            targets = self.annotation.target_selectors
```
nickstenning: I don't think this is right. Annotations in our current database schema have a single target, and `target_selectors` is an array of selectors. So this code now will always return `None` because `target` is a selector and `target.get("selector")` will always be `None`.

tests/factories.py
```diff
@@ -91,10 +91,8 @@ def uri(stub):
         return "http://example.com/document_{n}".format(n=stub.random_number)
 
     @factory.LazyAttribute
-    def target(stub):
+    def target_selectors(stub):
         return [{
```
nickstenning: Ditto here. This is now returning an object which doesn't look like anything that will ever be returned from `target_selectors` on a real `Annotation` object.

tests/factories.py
```diff
@@ -91,10 +91,8 @@ def uri(stub):
         return "http://example.com/document_{n}".format(n=stub.random_number)
 
     @factory.LazyAttribute
-    def target(stub):
+    def target_selectors(stub):
```
nickstenning: Ditto here. This is now returning a data structure that isn't compatible with the one returned by `target_selectors` on a real `Annotation` object.

tests/h/feeds/atom_test.py
```diff
@@ -1,13 +1,26 @@
...
+from h.api import models
 from h.feeds import atom
 
 from ... import factories
 
 
+def _annotation(**kwargs):
+    args = {
+        'userid': 'acct:janebloggs@hypothes.is',
+        'created': datetime.now(),
```
nickstenning: To avoid confusion about what we expect datetimes in our system to be, I'd suggest _always_ using `utcnow()`, even in cases like this where it doesn't matter.

tests/h/feeds/rss_test.py
```diff
@@ -17,24 +18,31 @@ def _annotation_url():
     return mock.Mock(return_value='https://hypothes.is/a/id')
 
 
+def _annotation(**kwargs):
+    args = dict(
```
nickstenning: This seems weirdly inconsistent. Let's have `{'userid': ...}` here not `dict(userid=...)`.

tests/h/presenters_test.py
```diff
@@ -560,10 +519,10 @@ def test_document_link_no_href_and_no_hostname(hostname_or_filename, href,
 
 def test_description():
     annotation = _annotation(
-        annotation={
-            "target": [{'selector': [{'exact': 'selected text'}]}],
-            "text": "entered text"
-        }
+        annotation=mock.Mock(
+            target_selectors=[{'selector': [{'exact': 'selected text'}]}],
```
nickstenning: One more here. This should just be `[{'exact': 'selected text'}]`.

h/feeds/rss.py
seanh: I don't think so, it's missing in some tests but I think the tests are wrong

h/feeds/rss.py
seanh: This is what `factories.py` is for, at some point we should refactor these tests to use that

h/feeds/rss.py
seanh: Done

h/feeds/rss.py
seanh: Done

h/presenters.py
seanh: `__getattr__()` is a convenience so that if you've wrapped an `Annotation` in an `AnnotationHTMLPresenter` then you can still access the `Annotation`s properties, not just the `AnnotationHTMLPresenter`'s own properties, on the `AnnotationHTMLPresenter`. You don't have to keep references to both the `Annotation` and the `AnnotationHTMLPresenter` and remember which properties to get from which.

I'm fine with it being removed if you want but I'd vote to do it as a separate PR as I think it's separate from getting the feeds code to use Postgres instead of Elastic.

h/presenters.py
seanh: Yep

seanh: Done

tests/factories.py
seanh: Done

tests/h/feeds/atom_test.py
seanh: Done

tests/h/feeds/rss_test.py
seanh: Done

tests/h/presenters_test.py
seanh: Done

seanh:
@nickstenning All done

nickstenning:
Great. It would be nice if the history for this could be cleaned up into something slightly less messy.

seanh:
Rebased

nickstenning:
LGTM.

PR #3409, opened 2016-06-03
Load documents from groups page from Postgres instead of from Elasticsearch

seanh:
This depends on https://github.com/hypothesis/h/pull/3408 and https://github.com/hypothesis/h/pull/3410

### DIFF
--- a/h/groups/models.py
+++ b/h/groups/models.py
@@ -4,6 +4,7 @@ import sqlalchemy as sa
 from sqlalchemy.orm import exc
 import slugify
 
+from h.api import models
 from h.db import Base
 from h.db import mixins
 from h import pubid
@@ -57,6 +58,28 @@ class Group(Base, mixins.Timestamps):
 
     def __repr__(self):
         return '<Group: %s>' % self.slug
+
+    def documents(self, limit=25):
+        """
+        Return this group's most recently annotated documents.
+
+        Only returns documents that have shared annotations in this group,
+        not documents that only have private annotations in the group.
+
+        """
+        documents = []
+        annotations = (
+            sa.orm.object_session(self).query(models.Annotation)
+            .filter_by(groupid=self.pubid, shared=True)
+            .order_by(models.Annotation.updated.desc())
+            .limit(1000))
+        for annotation in annotations:
+            if annotation.document not in documents:
+                documents.append(annotation.document)
+                if len(documents) >= limit:
+                    break
+
+        return documents
 
     @classmethod
     def get_by_pubid(cls, pubid):
--- a/h/groups/views.py
+++ b/h/groups/views.py
@@ -1,19 +1,13 @@
 # -*- coding: utf-8 -*-
 
-import collections
-
 import deform
 from pyramid import httpexceptions as exc
 from pyramid.view import view_config
 from pyramid import renderers
 
-
 from h import i18n
 from h import models
 from h import presenters
-from h.api import search
-from h.api import storage
-from h.api import uri
 from h.groups import schemas
 
 import h.session
@@ -99,38 +93,17 @@ def _read_group(request, group):
     group visits the group's URL.
 
     """
-    url = request.route_url('group_read', pubid=group.pubid, slug=group.slug)
-
-    result = search.search(request,
-                           private=False,
-                           params={"group": group.pubid, "limit": 1000})
-    annotations = [presenters.AnnotationHTMLPresenter(
-                       storage.annotation_from_dict(a))
-                   for a in result['rows']]
-
-    # Group the annotations by URI.
-    # Create a dict mapping the (normalized) URIs of the annotated documents
-    # to the most recent annotation of each document.
-    annotations_by_uri = collections.OrderedDict()
-    for annotation in annotations:
-        normalized_uri = uri.normalize(annotation.uri)
-        if normalized_uri not in annotations_by_uri:
-            annotations_by_uri[normalized_uri] = annotation
-            if len(annotations_by_uri) >= 25:
-                break
-
-    document_links = [annotation.document_link
-                      for annotation in annotations_by_uri.values()]
-
-    template_data = {
-        'group': group,
-        'group_url': url,
-        'document_links': document_links
-    }
-
     return renderers.render_to_response(
         renderer_name='h:templates/groups/share.html.jinja2',
-        value=template_data, request=request)
+        request=request,
+        value={
+            'group': group,
+            'group_url': request.route_url('group_read',
+                                           pubid=group.pubid,
+                                           slug=group.slug),
+            'document_links': [presenters.DocumentHTMLPresenter(document).link
+                               for document in group.documents()]
+        })
 
 
 def _join(request, group):
--- a/tests/h/groups/views_test.py
+++ b/tests/h/groups/views_test.py
@@ -1,8 +1,11 @@
 # -*- coding: utf-8 -*-
+import itertools
+
 import deform
 import mock
 import pytest
 from pyramid import httpexceptions
+from pyramid import testing
 
 from h.groups import views
 
@@ -161,8 +164,10 @@ def test_create_publishes_join_event(Group, session_model):
 
 
 # The fixtures required to mock all of read()'s dependencies.
-read_fixtures = pytest.mark.usefixtures(
-    'search', 'Group', 'renderers', 'uri', 'presenters')
+read_fixtures = pytest.mark.usefixtures('Group',
+                                        'renderers',
+                                        'routes',
+                                        'presenters')
 
 
 @read_fixtures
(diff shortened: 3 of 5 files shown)


seanh:
Things I still want to do here:

* Re-check that the SQL query is optimal and it isn't sending more queries than necessary

seanh:
The SQL queries that the groups page does on this branch - this looks good to me, it is not doing one query for each of the group's documents or anything:

Startup:

```sql
SELECT CAST('test plain returns' AS VARCHAR(60)) AS anon_1
SELECT CAST('test unicode returns' AS VARCHAR(60)) AS anon_1
DELETE FROM feature WHERE feature.name NOT IN (%(name_1)s, %(name_2)s, %(name_3)s, %(name_4)s, %(name_5)s)
```

The groups page gets the group itself:

```sql
SELECT "group".created AS group_created, "group".updated AS group_updated, "group".id AS group_id, "group".pubid AS group_pubid, "group".name AS group_name, "group".creator_id AS group_creator_id 
FROM "group" 
WHERE "group".pubid = %(pubid_1)s 
 LIMIT %(param_1)s
```

Something gets the user (authorized user? group creator?):

```sql
SELECT "user".username AS user_username, "user".password AS user_password, "user".id AS user_id, "user".uid AS user_uid, "user".admin AS user_admin, "user".staff AS user_staff, "user".sidebar_tutorial_dismissed AS user_sidebar_tutorial_dismissed, "user".email AS user_email, "user".last_login_date AS user_last_login_date, "user".registered_date AS user_registered_date, "user".activation_id AS user_activation_id, "user".salt AS user_salt, "user".password_updated AS user_password_updated 
FROM "user" 
WHERE "user".uid = %(uid_1)s 
 LIMIT %(param_1)s
```

Checking whether the user is a member of the group:

```sql
SELECT "group".created AS group_created, "group".updated AS group_updated, "group".id AS group_id, "group".pubid AS group_pubid, "group".name AS group_name, "group".creator_id AS group_creator_id 
FROM "group", user_group 
WHERE %(param_1)s = user_group.user_id AND "group".id = user_group.group_id ORDER BY "group".name
```

This is the big new query from this PR, the groups page gets the group's documents:

```sql
SELECT anon_1.document_created AS anon_1_document_created, anon_1.document_updated AS anon_1_document_updated, anon_1.document_id AS anon_1_document_id, anon_1.annotation_created AS anon_1_annotation_created, document_uri_1.claimant AS document_uri_1_claimant, document_uri_1.claimant_normalized AS document_uri_1_claimant_normalized, document_uri_1.uri AS document_uri_1_uri, document_uri_1.uri_normalized AS document_uri_1_uri_normalized, document_uri_1.created AS document_uri_1_created, document_uri_1.updated AS document_uri_1_updated, document_uri_1.id AS document_uri_1_id, document_uri_1.type AS document_uri_1_type, document_uri_1.content_type AS document_uri_1_content_type, document_uri_1.document_id AS document_uri_1_document_id, document_meta_1.claimant AS document_meta_1_claimant, document_meta_1.claimant_normalized AS document_meta_1_claimant_normalized, document_meta_1.created AS document_meta_1_created, document_meta_1.updated AS document_meta_1_updated, document_meta_1.id AS document_meta_1_id, document_meta_1.type AS document_meta_1_type, document_meta_1.value AS document_meta_1_value, document_meta_1.document_id AS document_meta_1_document_id 
FROM (SELECT document.created AS document_created, document.updated AS document_updated, document.id AS document_id, annotation.created AS annotation_created 
FROM document, annotation 
WHERE EXISTS (SELECT 1 
FROM document_uri, annotation 
WHERE document.id = document_uri.document_id AND annotation.target_uri_normalized = document_uri.uri_normalized AND annotation.shared = true AND annotation.groupid = %(groupid_1)s) ORDER BY annotation.created 
 LIMIT %(param_1)s) AS anon_1 LEFT OUTER JOIN document_uri AS document_uri_1 ON anon_1.document_id = document_uri_1.document_id LEFT OUTER JOIN document_meta AS document_meta_1 ON anon_1.document_id = document_meta_1.document_id ORDER BY anon_1.annotation_created, document_uri_1.updated DESC, document_meta_1.updated DESC
```

Something is checking whether the user is a member of the group again:

```sql
2016-06-06 11:41:18,854 [29798] [sqlalchemy.engine.base.Engine:INFO] SELECT "user".username AS user_username, "user".password AS user_password, "user".id AS user_id, "user".uid AS user_uid, "user".admin AS user_admin, "user".staff AS user_staff, "user".sidebar_tutorial_dismissed AS user_sidebar_tutorial_dismissed, "user".email AS user_email, "user".last_login_date AS user_last_login_date, "user".registered_date AS user_registered_date, "user".activation_id AS user_activation_id, "user".salt AS user_salt, "user".password_updated AS user_password_updated 
FROM "user", user_group 
WHERE %(param_1)s = user_group.group_id AND "user".id = user_group.user_id
```

seanh:
This can be reviewed now

h/groups/views.py
```diff
@@ -101,26 +104,28 @@ def _read_group(request, group):
...
+    # a. is shared, and
+    # b. is in this group.
+    documents = (
+        request.db.query(models.Document)
+        .options(joinedload(models.Document.meta))
+        .options(joinedload(models.Document.document_uris))
+        .filter(
+            models.Document.annotations.any(
+                and_(
+                    models.Annotation.shared == expression.true(),
```
chdorner: Is there any advantage of using `sqlalchemy.sql.expression.true()` instead of just `True`?

h/groups/views.py
```diff
@@ -101,26 +104,28 @@ def _read_group(request, group):
...
+                    models.Annotation.groupid == group.pubid
+                )
+            )
+        )
+        .order_by(models.Annotation.created)
+        [:25]
```
chdorner: As @nickstenning mentioned in [the old PR](https://github.com/hypothesis/h/pull/3388#discussion_r65523132), `.limit(25)` might be easier to read.

tests/h/groups/views_test.py
```diff
@@ -544,6 +562,61 @@ def test_leave_publishes_leave_event(Group, session_model):
...
+
+    """
+    annotation = models.Annotation(
+        userid=u'fred', groupid=groupid, shared=shared,
+        target_uri=document.document_uris[0].uri)
+    db.Session.add(annotation)
+    return annotation
+
+
+def document(db, uri):
```
chdorner: Nit: Might be nice to have the db session similar to the `annotation` helper function, or the other way around.

h/groups/views.py
```diff
@@ -101,26 +104,28 @@ def _read_group(request, group):
...
-            annotations_by_uri[normalized_uri] = annotation
-            if len(annotations_by_uri) >= 25:
-                break
-
-    document_links = [annotation.document_link
-                      for annotation in annotations_by_uri.values()]
+    # Find documents that have at least one annotation that:
+    # a. is shared, and
+    # b. is in this group.
+    documents = (
```
chdorner: I've tested this locally with a group that has >10k annotations and this query unfortunately takes about 2 minutes on my local machine.
I sat together with @nickstenning and we came up with this:
```
request.db.query(models.Document) \
        .options(joinedload(models.Document.meta)) \
        .options(joinedload(models.Document.document_uris)) \
        .join(models.Document.annotations) \
        .filter(models.Annotation.groupid == group.pubid, models.Annotation.shared == True) \
        .order_by(models.Annotation.updated.desc()) \
        .limit(25).all()
```

This query takes somewhere in between 3-6ms to execute (on the same group). The main difference is that it is using JOIN queries instead lots of subqueries.

The above query orders by `updated` which is not really what we want, but we don't have an index on `created` at the moment. So we should add an index on that column and then order by `created`.

h/groups/views.py
seanh: `== True` produces linter warnings from both Flake8 and PyLint. You're supposed to do `is True`. But `is True` doesn't work in this case (it never finds any annotations). Of course you could argue that the linters are wrong in this case. But the choice was between 1. Ignoring the linter warnings/creating linter noise, 2. Putting in linter comments to silence this line, or 3. Using `true()`.

h/groups/views.py
seanh: I'd argue for `[:25]` as more readable, being the standard way to limit a sequence in Python and sqlalchemy (the examples in the sqla tutorial docs use slicing not `limit()`, though of course `limit()` is documented when you get to the API docs). I'm happy to change this if you prefer `limit()` though.

h/groups/views.py
chdorner: Ah, right. Didn't think of that. Using `expression.true()` sounds good then

h/groups/views.py
chdorner: Considering we are creating a SQL query here I have a slight preference for using `limit()`, but not feeling strongly about it (we should just make sure that we use the same style for new code).

tests/h/groups/views_test.py
seanh: Do you think we should have a `db_session` fixture like we do for the API tests?

I think we'd end up with `db_session` as a fixture and `annotation()` as a helper function though, rather than both as fixtures or both as helper functions.

`annotation()` is a function because it's often called more than once by a single test, passing different values to it each time, sometimes those values are documents produced by earlier calls to `document()` in the same test function ... I think it's just easier to do this as a helper function. But for something like a `db_session` that can easily be done as a fixture I'd do so.

h/groups/views.py
seanh: Ah, for some reason I didn't think to test the query with groups with large numbers of annotations. Funnily enough I was trying to get this query to work with joins instead of subqueries initially precisely because it

SQL query that the original Python/SQLAlchemy code produced:

```sql
SELECT ... FROM
  SELECT ... FROM document, annotation WHERE EXISTS
    SELECT 1 FROM document_uri, annotation
      WHERE document.id = document_uri.document_id
      AND annotation.target_uri_normalized = document_uri.uri_normalized
      AND annotation.shared = true
      AND annotation.groupid = %(groupid_1)s
  ORDER BY annotation.created
  LIMIT %(param_1)s
AS anon_1 LEFT OUTER JOIN document_uri AS document_uri_1 ON anon_1.document_id = document_uri_1.document_id
LEFT OUTER JOIN document_meta AS document_meta_1 ON anon_1.document_id = document_meta_1.document_id
ORDER BY anon_1.annotation_created, document_uri_1.updated DESC, document_meta_1.updated DESC
```

Query that the above alternative with `join()` instead of `any()` produces:

```sql
SELECT ... FROM
  SELECT ... FROM document JOIN (document_uri AS document_uri_2
                                 JOIN document AS document_1 ON document_uri_2.document_id = document_1.id)
                           ON document_1.id = document_uri_2.document_id
                           JOIN annotation ON annotation.target_uri_normalized = document_uri_2.uri_normalized 
  WHERE annotation.groupid = %(groupid_1)s
  AND annotation.shared = true
  ORDER BY annotation.updated DESC 
  LIMIT %(param_1)s)
  AS anon_1 LEFT OUTER JOIN document_uri AS document_uri_1 ON anon_1.document_id = document_uri_1.document_id
  LEFT OUTER JOIN document_meta AS document_meta_1 ON anon_1.document_id = document_meta_1.document_id
  ORDER BY anon_1.annotation_updated DESC, document_uri_1.updated DESC, document_meta_1.updated DESC
```

So as you say the second subquery is replaced by a (double) join.

I was actually trying to write the query this way in the first place but I couldn't figure out how to use the `annotations` relationship in a join, thanks!

h/groups/models.py
```diff
@@ -58,6 +59,24 @@ def slug(self):
     def __repr__(self):
         return '<Group: %s>' % self.slug
 
+    def documents(self, db, limit=25):
```
nickstenning: For an instance method, use `sqlalchemy.orm.object_session` to retrieve the session, don't pass it in.

nickstenning:
Marking as WIP because I think we're agreed that this isn't working as-is.

h/groups/models.py
seanh: Done

h/groups/views.py
seanh: Ok the above query was incorrect, I've changed it to something completely different. Unfortunately I don't think there's any correct and efficient way to get a group's 25 most recently annotated documents currently, but the way that I've changed it to is more or less as good ad the previous Elasticsearch solution (it might be slower than Es on groups with lots of annotations).

h/groups/views.py
seanh: Done

h/groups/models.py
```diff
@@ -58,6 +59,34 @@ def slug(self):
...
+            .options(sa.orm.joinedload('document')
+                     .joinedload('document_uris'))
+            .options(sa.orm.joinedload('document')
+                     .joinedload('meta'))
+            .filter_by(groupid=self.pubid, shared=True)
+            .order_by(models.Annotation.updated.desc())
+            .limit(1000)
+            .all())
+        for annotation in annotations:
+            if annotation.document not in documents:
```
nickstenning: This smells like a `set` to me.

h/groups/models.py
```diff
@@ -58,6 +59,34 @@ def slug(self):
...
+            sa.orm.object_session(self).query(models.Annotation)
+            .options(sa.orm.load_only('id'))
+            .options(sa.orm.joinedload('document')
+                     .joinedload('document_uris'))
+            .options(sa.orm.joinedload('document')
+                     .joinedload('meta'))
+            .filter_by(groupid=self.pubid, shared=True)
+            .order_by(models.Annotation.updated.desc())
+            .limit(1000)
+            .all())
```
nickstenning: A query object is iterable on its own: there's no need to call `.all()` if you're about to iterate over it.

h/groups/models.py
```diff
@@ -58,6 +59,34 @@ def slug(self):
...
+        for annotation in annotations:
+            if annotation.document not in documents:
+                documents.append(annotation.document)
+                if len(documents) >= 25:
```
nickstenning: You've got a keyword argument you're not using here ;)

h/groups/models.py
seanh: Oh, cool

h/groups/models.py
seanh: It does, but a set would lose the order

h/groups/models.py
seanh: D'oh

h/groups/models.py
nickstenning: Hmm, ok. There are ways around that (use a set of ids, say, and a list) but it's only 25 by default so I guess it'll be fine.

h/groups/models.py
seanh: Fixed

h/groups/models.py
seanh: Done

seanh:
@chdorner I've reproduced the 11s+ load times on the page of a group with lots of annotations. I've removed the eager loading and the attempt to reduce the amount of data returned by returning only annotation IDs, this reduces the page load time to much less than 1s, close to 0.5s, for me (on my laptop). It's still not very fast but it's much better

chdorner:
LGTM.

PR #3410, opened 2016-06-03
Add DocumentHTMLPresenter class

seanh:
This isn't used yet but it will be used by the new groups page code in <https://github.com/hypothesis/h/pull/3409>

### DIFF
--- a/tests/h/presenters_test.py
+++ b/tests/h/presenters_test.py
@@ -1,3 +1,5 @@
+import datetime
+
 import pytest
 import mock
 import jinja2
@@ -44,7 +46,8 @@ class TestAnnotationHTMLPresenter(object):
     def test_created_day_string_from_annotation(self):
         annotation = self._annotation(
             annotation=mock.Mock(
-                created="2015-09-04T17:37:49.517852+00:00"))
+                created=datetime.datetime(2015, 9, 4, 17, 37, 49, 517852))
+            )
         assert annotation.created_day_string == '2015-09-04'
 
 
--- a/tests/h/presenters_test.py
+++ b/tests/h/presenters_test.py
@@ -5,573 +5,618 @@ import jinja2
 from h import presenters
 
 
-def _annotation(annotation=None, **kwargs):
-    """Return an AnnotationHTMLPresenter for the given annotation.
+class TestAnnotationHTMLPresenter(object):
 
-    If no annotation is given a mock will be used, and any keyword arguments
-    will be forwarded to the mock.Mock() constructor.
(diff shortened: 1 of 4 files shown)


seanh:
The diff on this is pretty misleading and GitHub is showing the two commits in the wrong order. What this PR does:

1. Move `AnnotationHTMLPresenter`'s tests into a new `TestAnnotationHTMLPresenter`, without changing the tests.

2. Add new `DocumentHTMLPresenter` class and move several properties from `AnnotationHTMLPresenter` into it.

3. Also move the relevant tests and fixtures from `TestAnnotationHTMLPresenter` into `TestDocumentHTMLPresenter`

4. Leaves behind methods on `AnnotationHTMLPresenter` in place of the ones that got moved into `DocumentHTMLPresenter`, like this:

   ```python
    @property
    def filename(self):
        """Return the filename of this annotation's document."""
        return self.document.filename
   ```

   (`self.document = DocumentHTMLPresenter(self.annotation.document)`)

   Also moved these all to the bottom of the class.

5. Fixed a couple of failing tests in other modules because `AnnotationHTMLPresenter` now requires the `Annotation` to have a `Document`.

I should try getting rid of the backwards-compat methods in 4...

seanh:
> I should try getting rid of the backwards-compat methods in 4...

Hmm, on the other hand it's actually pretty convenient both for the feeds code and its tests if the `DocumentHTMLPresenter` properties are accessible on the `AnnotationHTMLPresenter` object..

tests/h/presenters_test.py
```diff
@@ -5,573 +5,335 @@
...
-    filename.return_value = ""  # This is not a file:// URI.
+        assert annotation.description == (
+            "&lt;blockquote&gt;selected text&lt;/blockquote&gt;entered text")
 
-    annotation = _annotation(annotation)
-    assert isinstance(annotation.title, unicode)
+    def test_created_day_string_from_annotation(self):
+        annotation = self._annotation(
+            annotation=mock.Mock(
+                created="2015-09-04T17:37:49.517852+00:00"))
```
chdorner: This should be `datetime.datetime.datetime(2015, 9, 4, 17, 37, 49, 517852)` now.

chdorner:
Just one tiny oversight in the tests now that we work with `datetime.datetime` objects instead of ISO8601 strings. Otherwise looks great.

tests/h/presenters_test.py
seanh: Done. Thanks @chdorner !

chdorner:
LGTM.

Issue #3261, opened 2016-04-28
"links" field appearing in new Es index


seanh:
`AnnotationJSONPresenter` adds a (generated, not stored in db) "links" fields to the annotations we return from the API. <https://github.com/hypothesis/h/pull/3256/commits/fff6df028f1c524559569da120ead615ba385a64> makes sure this field doesn't get into the Pg database. But the "links" field is still appearing in the new Es index (but not in the legacy index). The reason is that `AnnotationJSONPresenter` is used to dictize the annotations for going into the new Es index.

Design problem?

Using `AnnotationJSONPresenter` for both means a 1:1 mapping between what annotations look like in Es and what our API returns.

Seems we probably want two different presenter classes?

seanh:
@nickstenning @chdorner What do we want to do about this? How about the suggestion to add a new `AnnotationSearchIndexPresenter` for presenting annotations? Or instead of a presenter class just add a new `h.api.search.transform.transform()` function for converting an annotation to a dict? (Will need to add the `transform` module as well)

seanh:
The problem is this line: https://github.com/hypothesis/h/blob/master/h/api/search/index.py#L31

seanh:
Actually it looks like there's only one `index()` function where this annotation -> dict for search indexing transformation needs to be done, so it can just be done inline in that function, or as a private helper function of that function, doesn't need a `transform` module

chdorner:
@seanh I'm just working on the second place where that is needed: https://github.com/hypothesis/h/blob/pg-es-reindex/h/indexer.py#L93

I generally like the idea of a specific search presenter, the only problem I see at the moment is that we still return the JSON returned by ElasticSearch back to the client, so if we decide to create a search presenter it will have to stay in sync with the `AnnotationJSONPresenter`.

seanh:
Right. Until the postgres migration is finished and the legacy Es storage is gone I think the annotations in the new Es search index should be exactly the same as the annotations in the legacy Es database. Then once the migration is finished and the legacy Es storage is gone we'll no longer be returning data from Es to the user (not even indirectly by passing it through `AnnotationJSONPresenter` first or anything), we'll be getting the Postgres annotation model object(s) and presenting them with `AnnotationJSONPresenter` and returning them. So at that point a search presenter could diverge from `AnnotationJSONPresenter` as much as we wanted. But for now it would be the same except without the "links" field.

[...]

PR #3413, opened 2016-06-06
Remove the "links" field from the search index

seanh:
Remove the "links" field of annotations from the search index.

This field did not exist in the search index back when we used the
legacy Elasticsearch storage, it only existed in the annotations
returned by the API, but it got accidentally added into the search index when
we moved to the new Postgres storage becayse the new search indexer uses
AnnotationJSONPresenter which is the presenter used for returning
annotations from the API.

Fundamentally we don't want the annotations returned by the API to have
to look exactly the same as the annotations in the search index - these
are two different things, and may evolve in different directions
according to different needs.

So add a new AnnotationSearchIndexPresenter and have the search indexer
use that instead of AnnotationJSONPresenter. Now the annotation formats
used by the API and search index are separated in the code and can be
modified independently.

Note that both AnnotationJSONPresenter and
AnnotationSearchIndexPresenter still use DocumentJSONPresenter for the
document sub-object of annotations - it has not been necessary to
separate out a DocumentSearchIndexPresenter yet.

Fixes https://github.com/hypothesis/h/issues/3261

### DIFF
--- a/h/api/presenters.py
+++ b/h/api/presenters.py
@@ -62,6 +62,9 @@ class AnnotationBasePresenter(object):
 
 
 class AnnotationJSONPresenter(AnnotationBasePresenter):
+
+    """Present an annotation in the JSON format returned by API requests."""
+
     def asdict(self):
         docpresenter = DocumentJSONPresenter(self.annotation.document)
 
@@ -90,14 +93,43 @@ class AnnotationJSONPresenter(AnnotationBasePresenter):
 
     @property
     def permissions(self):
-        read = self.annotation.userid
-        if self.annotation.shared:
-            read = 'group:{}'.format(self.annotation.groupid)
+        return _permissions(self.annotation)
 
-        return {'read': [read],
-                'admin': [self.annotation.userid],
-                'update': [self.annotation.userid],
-                'delete': [self.annotation.userid]}
+
+class AnnotationSearchIndexPresenter(AnnotationBasePresenter):
+
+    """Present an annotation in the JSON format used in the search index."""
+
+    def asdict(self):
+        docpresenter = DocumentJSONPresenter(self.annotation.document)
+
+        base = {
+            'id': self.annotation.id,
+            'created': self.created,
+            'updated': self.updated,
+            'user': self.annotation.userid,
+            'uri': self.annotation.target_uri,
+            'text': self.text,
+            'tags': self.tags,
+            'group': self.annotation.groupid,
+            'permissions': self.permissions,
+            'target': self.target,
+            'document': docpresenter.asdict(),
+        }
+
+        base['target'][0]['scope'] = [self.annotation.target_uri_normalized]
+
+        if self.annotation.references:
+            base['references'] = self.annotation.references
+
+        annotation = copy.copy(self.annotation.extra) or {}
+        annotation.update(base)
+
+        return annotation
+
+    @property
+    def permissions(self):
+        return _permissions(self.annotation)
 
 
 class AnnotationJSONLDPresenter(AnnotationBasePresenter):
@@ -238,6 +270,24 @@ def deep_merge_dict(a, b):
 
 def _json_link(request, annotation):
     return request.route_url('api.annotation', id=annotation.id)
+
+
+def _permissions(annotation):
+    """
+    Return a permissions dict for the given annotation.
+
+    Converts our simple internal annotation storage format into the legacy
+    complex permissions dict format that is still used in some places.
+
+    """
+    read = annotation.userid
+    if annotation.shared:
+        read = 'group:{}'.format(annotation.groupid)
+
+    return {'read': [read],
+            'admin': [annotation.userid],
+            'update': [annotation.userid],
+            'delete': [annotation.userid]}
 
 
 def includeme(config):
--- a/h/api/search/index.py
+++ b/h/api/search/index.py
@@ -38,14 +38,8 @@ def index(es, annotation, request):
     :type annotation: h.api.models.Annotation
 
     """
-    # FIXME: this should not use the same presenter as we use to render
-    # annotations for clients. It is useful in the mean time until we get rid of
-    # the legacy ElasticSearch annotation storage.
-    annotation_dict = presenters.AnnotationJSONPresenter(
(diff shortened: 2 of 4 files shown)


h/api/presenters.py
```diff
@@ -90,14 +93,41 @@ def asdict(self):
 
     @property
     def permissions(self):
-        read = self.annotation.userid
-        if self.annotation.shared:
-            read = 'group:{}'.format(self.annotation.groupid)
+        return _permissions(self.annotation)
+
+
+class AnnotationSearchIndexPresenter(AnnotationBasePresenter):
```
chdorner: Some basic testing would be nice here, so we're not going to accidentally break search indexing.

h/api/search/index.py
```diff
@@ -41,7 +41,7 @@ def index(es, annotation, request):
     # FIXME: this should not use the same presenter as we use to render
     # annotations for clients. It is useful in the mean time until we get rid of
     # the legacy ElasticSearch annotation storage.
```
chdorner: This comment can go now as well.

chdorner:
Looks good except for a few small things here and there. I just checked how the rendering works in the API and we're still using `annotation_from_dict` (which was not on my list, I will create a new card for that), we just have to remember to still properly present it when rendering.

h/api/search/index.py
```diff
@@ -41,7 +41,7 @@ def index(es, annotation, request):
...
-    annotation_dict = presenters.AnnotationJSONPresenter(
+    annotation_dict = presenters.AnnotationSearchIndexPresenter(
         request, annotation).asdict()
 
     annotation_dict['target'][0]['scope'] = [
```
nickstenning: This should probably move into the search index presenter, then?

h/api/presenters.py
```diff
@@ -90,14 +93,41 @@ def asdict(self):
...
+            'created': self.created,
+            'updated': self.updated,
+            'user': self.annotation.userid,
+            'uri': self.annotation.target_uri,
+            'text': self.text,
+            'tags': self.tags,
+            'group': self.annotation.groupid,
+            'permissions': self.permissions,
+            'target': self.target,
+            'document': docpresenter.asdict(),
```
nickstenning: On the basis that we don't index this, I think we can probably remove this now.

h/api/presenters.py
```diff
@@ -90,14 +93,41 @@ def asdict(self):
...
+        }
+
+        if self.annotation.references:
+            base['references'] = self.annotation.references
 
-        return {'read': [read],
-                'admin': [self.annotation.userid],
-                'update': [self.annotation.userid],
-                'delete': [self.annotation.userid]}
+        annotation = copy.copy(self.annotation.extra) or {}
```
nickstenning: We should probably also not be indexing extras, because doing so can result in nasty mapping clashes.

h/api/search/index.py
nickstenning: If you do, you can move it in as `uri_normalized` rather than scope and make the appropriate changes to the search query builder.

nickstenning:
I've made a couple of comments, not all of which are suggesting changes to this PR, but changes which are made possible/desirable as a result of doing this work. I think for me all I'd like to see is a few tests for this and then we can merge and follow up on the other stuff I've mentioned.

h/api/search/index.py
```diff
@@ -147,7 +147,7 @@ def _prepare(self, annotation):
         # FIXME: this should not use the same presenter as we use to render
         # annotations for clients. It is useful in the mean time until we get rid of
         # the legacy ElasticSearch annotation storage.
-        data = presenters.AnnotationJSONPresenter(self.request, annotation).asdict()
+        data = presenters.AnnotationSearchIndexPresenter(self.request, annotation).asdict()
         data['target'][0]['scope'] = [annotation.target_uri_normalized]
```
nickstenning: Ditto above. The reason this is separate is because we didn't want to add this to `AnnotationJSONPresenter`. There's no reason not to add it to `AnnotationSearchIndexPresenter`.

h/api/presenters.py
chdorner: Not yet I believe, because the search endpoint still uses `annotation_from_dict` and then presents the response from ES. As soon as that endpoint goes to PG as well after searching, then we can remove this.

h/api/presenters.py
chdorner: Same as above, we still need to keep this for a bit.

h/api/presenters.py
nickstenning: Ugh, right. Sorry.

h/api/presenters.py
nickstenning: :+1:

h/api/search/index.py
seanh: I'd like to do the rename in a separate pr

h/api/presenters.py
seanh: Tests added

h/api/search/index.py
seanh: Done

h/api/search/index.py
seanh: Done

seanh:
If this can be merged it'll unblock my work on the new `Annotation` test factory

tests/h/api/presenters_test.py
```diff
@@ -446,3 +562,16 @@ def tzname(self, dt):
...
+def clear_registry():
+    # If you call add_annotation_link_generator() as some tests in this module
+    # do it'll add stuff to the request registry that you pass to it.
+    # Then when a later test runs and creates a _new_ DummyRequest instance,
+    # that request's registry will still contain the stuff added by the
+    # previous test!
+    # DummyRequest.registry is shared across requests.
+    # So rather than requiring each test that adds to the registry to also
+    # clear it, this autouse fixture just clears the registry after every test.
+    DummyRequest().registry.clear()
```
nickstenning: Hmm. I think the cleanup here should probably be on the relevant tests, or global. But ¯\_(ツ)_/¯... works for now.

nickstenning:
LGTM.

PR #3415, opened 2016-06-06
Add non-nullable and default values to annotation.extra

### LINKED ISSUE #3356: Set Annotation.extra column `NOT NULL` and default to `{}`
The workaround for editing mutable columns means that we cannot allow `NULL` values in the `annotation.extras` column.

New annotations created after the switch to postgres already set the default value to `{}`, migrated annotations still had the value set to `NULL`. I fixed this manually by setting the value for all `NULL` records to `{}`.

We should set a `NOT NULL` constraint on this column with a default value of `{}`.

seanh:
Fixes #3356

### DIFF
--- a/h/api/models/annotation.py
+++ b/h/api/models/annotation.py
@@ -78,7 +78,7 @@ class Annotation(Base, mixins.Timestamps):
                            server_default=sa.text('ARRAY[]::uuid[]'))
 
     #: Any additional serialisable data provided by the client.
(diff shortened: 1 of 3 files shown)


chdorner:
LGTM. It's not an issue for us right now, but it's worth noting that the order of these things is quite important (and could lead to issues). Generally it should be:

1. Migration to set server-side default value
1. Run a query to update the `NULL` values to `'{}'` (our views code already did this part)
1. Migration to set the `NOT NULL` constraint

PR #3416, opened 2016-06-06
Add move_uri cli command

chdorner:
Which allows for easier moving of annotations and document equivalence data
from one URL to another.

This will **replace** the annotation's ``target_uri`` and all the
document uri's ``claimant``, plus the matching ``uri`` for
self-claim and canonical uris.

I've tried and make the command as safe as possible to use, this is how it looks like:
```shell
$ hypothesis --dev move_uri \
  --old http://www.hybridpedagogy.com/announcements/when-words-go-global-a-digped-chat/ \
  --new http://www.digitalpedagogylab.com/hybridped/when-words-go-global-a-digped-chat/

Changing all annotations and document data matching:
"http://www.hybridpedagogy.com/announcements/when-words-go-global-a-digped-chat/"
to:
"http://www.digitalpedagogylab.com/hybridped/when-words-go-global-a-digped-chat/"
This will affect 1 annotations, 3 document uri claimants, and 2 document uri self-claims or canonical uris.
Are you sure? [y/N]: 
```

### DIFF
--- a/h/cli/commands/move_uri.py
+++ b/h/cli/commands/move_uri.py
@@ -1,0 +1,80 @@
+# -*- coding: utf-8 -*-
+
+import click
+
+from h import models
+from h.api import uri
+from h.api.search.index import BatchIndexer
+
+
+@click.command('move-uri')
+@click.option('--old', required=True,
+              help='Old URI with annotations and documents.')
+@click.option('--new', required=True, confirmation_prompt=True,
+              help='New URI for matching annotations and documents.')
+@click.pass_context
+def move_uri(ctx, old, new):
+    """
+    Move annotations and document equivalence data from one URL to another.
+
+    This will **replace** the annotation's ``target_uri`` and all the
+    document uri's ``claimant``, plus the matching ``uri`` for self-claim and
+    canonical uris.
+    """
+
+    request = ctx.obj['bootstrap']()
+
+    annotations = _fetch_annotations(request.db, old)
+    docuris_claimant = _fetch_document_uri_claimants(request.db, old)
+    docuris_uri = _fetch_document_uri_canonical_self_claim(request.db, old)
+
+    prompt = ('Changing all annotations and document data matching:\n' +
+              '"{old}"\nto:\n"{new}"\n' +
+              'This will affect {ann_count} annotations, {doc_claimant} ' +
+              'document uri claimants, and {doc_uri} document uri self-claims ' +
+              'or canonical uris.\n' +
+              'Are you sure? [y/N]').format(old=old, new=new,
+                                            ann_count=len(annotations),
+                                            doc_claimant=len(docuris_claimant),
+                                            doc_uri=len(docuris_uri))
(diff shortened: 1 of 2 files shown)


chdorner:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;aR9F1uC9&#x2F;367-move-digitalpedagogy-annotations-to-new-domain"> Move digitalpedagogy annotations to new domain</a>

seanh:
Code looks fine.

I wonder if we can do something safer with URI normalization here though? Instead of changing target_uri, claimant, etc leave them as they originally where but store something that tells the URI normalization code that `old.com/*` should be transformed to `new.com/*` as part of normalization, and then re-run the normalization to update the normalized URIs. Have a db table of these "additional transformations" to apply during normalization and a CRUD API for that table. If you accidentally added a wrong rule you could just delete it and re-run the normalization. This probably requires a lot more thought though..

nickstenning:
In general @seanh that's the approach we'd like to take to "fixing" document equivalence data, and is _precisely_ the reason why the new document equivalence data is stored as it is (i.e. with an explicit "claimant" URI) -- so we can add rules like "ignore `rel=canonical` tags from `example.com`".

But that'll be a chunk of work, and it's complicated even more by the rules surrounding searching for annotations by a URI we know to be canonical (we skip the step where we expand the search to all equivalent URIs in this case).

For now this addresses a specific need that we don't really have any other good approaches for.

h/cli/commands/move_uri.py
```diff
@@ -0,0 +1,80 @@
+# -*- coding: utf-8 -*-
+
+import click
+
+from h import models
+from h.api import uri
+from h.api.search.index import BatchIndexer
+
+
+@click.command()
```
nickstenning: I have a fanatical dislike for underscores on the command line, sorry. Can we make this `@click.command('move-uri')`?

h/cli/commands/move_uri.py
```diff
@@ -0,0 +1,80 @@
...
+
+def _fetch_document_uri_claimants(session, uri_):
+    return session.query(models.DocumentURI).filter(
+        models.DocumentURI.claimant_normalized == uri.normalize(uri_)).all()
+
+
+def _fetch_document_uri_canonical_self_claim(session, uri_):
+    return session.query(models.DocumentURI).filter(
+        models.DocumentURI.uri_normalized == uri.normalize(uri_),
+        models.DocumentURI.type.in_([u'self-claim', u'rel-canonical'])).all()
```
nickstenning: I'm not sure I understand the motivation for limiting ourselves to these claim types here?

chdorner: This is to change as little as possible while still have the document equivalence work (with our special case for canonical uris). The next annotation will then automatically create new document uris again.

h/cli/commands/move_uri.py
chdorner: Sure, done.

nickstenning:
LGTM.

PR #3417, opened 2016-06-06
Persist in-place mutations of annotation tags and extras

seanh:
`target_selectors` is also a mutable object mutations of which are not being persisted to the database but that one is a more difficult case. I think these fixes for tags and extras can be merged separately first.

### DIFF
--- a/h/api/db/types.py
+++ b/h/api/db/types.py
@@ -9,6 +9,7 @@ import uuid
 from sqlalchemy import types
 from sqlalchemy.dialects import postgresql
 from sqlalchemy.exc import DontWrapMixin
+from sqlalchemy.ext.mutable import Mutable
 
 
 # A magic byte (expressed as two hexadecimal nibbles) which we use to expand a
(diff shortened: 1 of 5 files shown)


tests/h/api/models/annotation_test.py
```diff
@@ -83,6 +83,94 @@ def test_acl_group_shared():
...
+
+    Setting an Annotation.extra value in-place:
+
+        my_annotation.extra['foo'] = 'bar'
+
+    should be persisted to the database.
+
+    """
+    annotation = Annotation(userid='fred')
+    annotation.extra = {}  # FIXME: There should be a default value for this.
```
seanh: https://github.com/hypothesis/h/pull/3415 should allow this line to be removed

nickstenning: Now merged.

tests/h/api/models/annotation_test.py
```diff
@@ -83,6 +83,94 @@ def test_acl_group_shared():
...
+
+    # Add the annotation to the db so that annotation.id gets generated.
```
nickstenning: Possibly leaning a little towards the `# Add 1 to i` style of commenting here ;)

tests/h/api/models/annotation_test.py
```diff
@@ -83,6 +83,94 @@ def test_acl_group_shared():
...
+    db_session.add(annotation)
+
+    # We need to flush the db here so that the initial value of {} for
+    # annotation.tags gets persisted, otherwise this test would never fail.
+    db_session.flush()
+
+    # Make an in-place change to annotation.extra.
```
nickstenning: Ditto.

h/api/db/types.py
```diff
@@ -65,6 +66,79 @@ def process_result_value(self, value, dialect):
...
+    @classmethod
+    def coerce(cls, key, value):
+        if not isinstance(value, MutableList):
+            if isinstance(value, list):
+                return MutableList(value)
+            return Mutable.coerce(key, value)
+        else:
+            return value
+
+    def __setitem__(self, idx, value):
```
nickstenning: It wouldn't be too hard, I don't think, to write a fairly concise set of tests for this class. Something like the following, perhaps:

```python
@mock.patch.object(MutableList, 'changed')
@pytest.mark.parametrize('start,end,retval,op,args', (
    (['foo'], ['bar'], None, '__setitem__', (0, 'bar')),
    ([1,2,3], [1,4,5], None, '__setslice__', (1, 3, [4, 5])),
    ...
))
def test_mutable_list_operation(start, end, retval, op, args, changed):
    val = MutableList(start)
    result = getattr(val, op)(*args)
    
    assert result == retval
    assert val == end
    changed.assert_called_once_with()
```

h/api/db/types.py
```diff
@@ -65,6 +66,79 @@ def process_result_value(self, value, dialect):
         return _get_urlsafe_from_hex(hexstring)
 
 
+class MutableList(Mutable, list):
```
nickstenning: Perhaps make a note here that we can remove this entirely when SQLAlchemy 1.1 is released in favour of `sqlalchemy.ext.mutable.MutableList`.

h/api/db/types.py
seanh: Done

h/api/db/types.py
seanh: Done

tests/h/api/models/annotation_test.py
seanh: Deleted

tests/h/api/models/annotation_test.py
seanh: Deleted

tests/h/api/models/annotation_test.py
seanh: Line deleted

tests/h/api/db/types_test.py
```diff
@@ -109,3 +110,51 @@ def test_annotation_selector_deserialize_missing_text_quote_selector():
...
+        lambda l: l.__setslice__(1, 3, ['a', 'b']),
+        lambda l: l.__delitem__(0),
+        lambda l: l.__delslice__(1, 3),
+        lambda l: l.append('value'),
+        lambda l: l.insert(0, 'value'),
+        lambda l: l.extend(['value']),
+        lambda l: l.pop(),
+        lambda l: l.remove(1),
+        lambda l: l.sort(),
+        lambda l: l.reverse(),
```
nickstenning: Nice! Much nicer than what I proposed :)

nickstenning:
LGTM.

PR #3420, opened 2016-06-07
List pull requests that were merged without a merge commit in list-changes-since.py

robertknight:
The changelog helper script relied on parsing merge commit messages in order to
get the numbers of pull requests merged since a given tag.

This did not include pull requests that were merged via a fast-forward merge.

Fix this by using the GitHub API to enumerate pull requests in descending
last-update order and finding those merged since the tag was created.

There is a caveat that this relies on the system time being reasonably accurate
on the system where the tag was created.

### DIFF
--- a/scripts/list-changes-since.py
+++ b/scripts/list-changes-since.py
@@ -1,6 +1,7 @@
 #!/usr/bin/env python
 
 import argparse
+import dateutil.parser
 import re
 import requests
 import subprocess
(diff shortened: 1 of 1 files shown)


scripts/list-changes-since.py
```diff
@@ -35,28 +36,70 @@ def get_last_tag():
...
+    while url:
+        res = requests.get(url, headers=headers, params=params)
+        if res.status_code != 200:
+            raise Exception('GitHub request failed:', res.text)
+
+        page = res.json()
+        if isinstance(page, list):
+            for item in page:
+                yield item
+            url = res.links.get('next') and res.links.get('next')['url']
```
nickstenning: FWIW this kind of thing isn't really idiomatic in Python, IMO. Probably something more like:

```python
try:
    url = res.links['next']['url']
except KeyError:
    url = None
```

nickstenning:
Well, it seems to work.

But in the interests of always trying to simplify things, perhaps this script would be a lot shorter if it just looked through the git history for commits where the commit message first line matches one of these patterns?

    ^Merge pull request #\d+
    \(#\d+\)$

No?

robertknight:
> But in the interests of always trying to simplify things, perhaps this script would be a lot shorter
> if it just looked through the git history for commits where the commit message first line matches
> one of these patterns?

That seems reasonable. Just a matter of finding out whether parsing the commit log would miss any issues that the current implementation finds, or vice-versa.

PR #3422, opened 2016-06-08
Add Hypothesis-driven property-based tests for query builder

nickstenning:
This has the potential to cause much confusion. I'm sorry.

But I think the payoff is worth it.

Hypothesis is the thing we're building.

But Hypothesis is also an extremely capable property-based testing library for Python: https://github.com/HypothesisWorks/hypothesis-python

Property-based testing is an approach to testing code in which you write tests which assert the invariant properties of the system under test. For example, the tests in this commit check the following invariants of the code responsible for parsing the "limit" parameter to the search query builder:

- Any input of any description (assuming it's a string) will only ever result in an output in the range 0...LIMIT_MAX.
- Any input which is a string that is an integer in the range 0...LIMIT_MAX (e.g. "24") should result in an output which is the corresponding integer.

This is a very simple example of a property-based test, but even the few lines of code below result in *much* more comprehensive testing of the component in question than the small list of examples which it replaces. This is because Hypothesis typically generates dozens to hundreds of examples with which it tries to invalidate the invariant you've set up.

For more about Hypothesis and property-based testing, see [1].

For more about how to generate the right data for a Hypothesis-based test, see [2].

[1]: http://hypothesis.works/
[2]: http://hypothesis.works/articles/generating-the-right-data/

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,7 @@ node_modules/
 # py.test state cache
 .cache/
 .coverage
+.hypothesis/
 .tox/
 
 .eggs/
--- a/tests/h/api/search/query_test.py
+++ b/tests/h/api/search/query_test.py
@@ -1,7 +1,9 @@
 # -*- coding: utf-8 -*-
 
-import pytest
 import mock
+import pytest
+from hypothesis import strategies as st
+from hypothesis import given
 from webob import multidict
 
 from h.api.search import query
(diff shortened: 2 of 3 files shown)


chdorner:
LGTM. Might be good to have @seanh and/or @finitud look at this as well.

nickstenning:
Let's get this merged, please. @seanh, @finitud?

seanh:
Reading about Hypothesis now

finitud:
Reading it now as well

seanh:
LGTM, I'll read up a little more on hypothesis but no need for this PR to wait on that. I'm not sure yet how it makes the test results reproducible/predictable across different developer's machines and Travis (unless we just manage cases that come up using `@example()`?), though I see you've marked all the tests as fuzz.

nickstenning:
>  I'm not sure yet how it makes the test results reproducible/predictable across different developer's machines and Travis

It doesn't. But it will print enough information on a failure to be able to reproduce it.

Across multiple runs on the same machine it will save state (including failing examples it has found) in the .hypothesis folder.

finitud:
LGTM 👍

nickstenning:
If all three of you have said LGTM, perhaps you could push the green button? 😉

PR #3423, opened 2016-06-08
First steps towards removing the query property

nickstenning:
The use of a "magic" query property makes testing difficult and causes confusion about which session is the correct session to use when writing new code.

This PR is part of an effort to standardise all ORM classmethods so they take a database session as their first parameter and do not rely on a thread-local query property.

There is a bunch more work to do here, some of which requires unpicking of other code, but this is a start.


chdorner:
Looking good.

PR #3424, opened 2016-06-08
Remove the h.accounts.make_admin helper

nickstenning:
As part of the work started in #3423, I'm working up to making the `User.get_by_username` classmethod take an explicit database session rather than using the query property.

There are a couple of places that use `User.get_by_username` that don't currently have access to the request object, including these helpers (`make_admin` and `make_staff`) in `h.accounts`.

So, my options are:

1. Pass the request object into these helpers.
2. Remove the layer of indirection that the helpers represent.

I've opted here for the latter, because they don't really seem to be doing anything particularly useful. There are only two places that `make_admin` is used, namely the CLI and the admin panel. In the CLI, the NoSuchUserError is not caught and thus bubbles up to the user as a traceback, which isn't very nice. In the admin view, the code needed to handle the NoSuchUserError is almost exactly as verbose as the code needed without this helper.

Moreover, in the admin views, the helper can only be used to make a user an admin, not to remove their admin status, so the companion view (`admins_remove`) already knows about the `User` model.

### DIFF
--- a/h/accounts/__init__.py
+++ b/h/accounts/__init__.py
@@ -34,15 +34,6 @@ class JSONError(Error):
     pass
 
 
-def make_admin(username):
-    """Make the given user an admin."""
-    user = models.User.get_by_username(username)
-    if user:
-        user.admin = True
-    else:
-        raise NoSuchUserError
-
-
 def make_staff(username):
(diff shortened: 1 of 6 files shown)


chdorner:
LGTM.

PR #3425, opened 2016-06-08
Clean up admin "admins" view and associated tests

nickstenning:
This commit is mainly about cleaning up the tests for the views responsible for managing which users are admins so that they are consistent with the test changes made in #3424.

It also makes the `admins_add` and `admins_remove` views consistent in their behaviour (they now both do redirect-after-post rather than just one of them), and fixes a bug that allowed the `admins_remove` view to crash if the username provided to it was invalid.

### DIFF
--- a/tests/h/admin/views/admins_test.py
+++ b/tests/h/admin/views/admins_test.py
@@ -1,5 +1,7 @@
 # -*- coding: utf-8 -*-
 
+from __future__ import unicode_literals
+
 from mock import Mock
 from pyramid import httpexceptions
 from pyramid.testing import DummyRequest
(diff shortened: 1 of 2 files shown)


tests/h/admin/views/admins_test.py
```diff
@@ -8,186 +10,128 @@
...
 
-@admins_remove_fixtures
-def test_admins_remove_returns_redirect_when_too_few_admins(User):
-    User.admins.return_value = [Mock(username="fred")]
-    request = DummyRequest(params={"remove": "fred"})
-
-    response = views.admins_remove(request)
-
-    assert isinstance(response, httpexceptions.HTTPSeeOther)
+    admins = ['agnos', 'bojan', 'cristof']
```
chdorner: Never seen this version of my name before, have :100: points :smile:

chdorner:
LGTM. Nice that we now redirect to the index page.

tests/h/admin/views/admins_test.py
nickstenning: Deliberately not you 😉

PR #3426, opened 2016-06-08
Remove the h.accounts.make_staff helper

nickstenning:
This is exactly the same set of changes as made in #3424 and #3425, only for
the `make_staff` helper and views rather than for the `make_admin` helper and
views.


chdorner:
LGTM.

PR #3429, opened 2016-06-09
Do not use query property in User.get_by_username

nickstenning:
The use of a "magic" query property makes testing difficult and causes confusion about which session is the correct session to use when writing new code.

This commit is part of an effort to standardise all ORM classmethods so they take a database session as their first parameter and do not rely on a thread-local query property.

### DIFF
--- a/tests/h/admin/views/admins_test.py
+++ b/tests/h/admin/views/admins_test.py
@@ -2,7 +2,7 @@
 
 from __future__ import unicode_literals
 
-from mock import Mock
+import mock
 from pyramid import httpexceptions
 from pyramid.testing import DummyRequest
 import pytest
@@ -12,18 +12,14 @@ from h.admin.views import admins as views
 
 @pytest.mark.usefixtures('routes')
 class TestAdminsIndex(object):
-    def test_when_no_admins(self):
-        request = DummyRequest()
-
-        result = views.admins_index(request)
+    def test_when_no_admins(self, req):
+        result = views.admins_index(req)
 
         assert result["admin_users"] == []
 
     @pytest.mark.usefixtures('users')
-    def test_context_contains_admin_usernames(self):
-        request = DummyRequest()
-
-        result = views.admins_index(request)
+    def test_context_contains_admin_usernames(self, req):
+        result = views.admins_index(req)
 
         assert set(result["admin_users"]) == set(["agnos", "bojan", "cristof"])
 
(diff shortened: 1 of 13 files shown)


tests/h/admin/views/admins_test.py
```diff
@@ -12,102 +12,111 @@
...
-        result = views.admins_remove(request)
+        result = views.admins_remove(req)
 
         assert isinstance(result, httpexceptions.HTTPSeeOther)
         assert result.location == '/adm/admins'
 
 
 @pytest.fixture
+def req(db_session):
+    return DummyRequest(db=db_session)
```
seanh: I've used `mock_request` as the name for mock request fixtures elsewhere, I think it's more readable as overly terse names like `req` can be unclear (especially as you get more and more of them in the code). `mock_request` would be consistent with existing fixtures elsewhere. I know that `request` isn't a possible name for a custom fixture.

seanh:
Otherwise looks great

tests/h/admin/views/admins_test.py
nickstenning: It wouldn't be consistent with tests in almost exactly the same style in `admins_test.py` and `staff_test.py`, and it's not a mock (it's a stub) so I'm not sure I want to change it to `mock_request`...

...TBH I'm also not sure what's wrong with short names in this context. Is there anything else `req` could plausibly mean here?

nickstenning: Oops, sorry -- this *is* `admins_test.py` -- that'll teach me to read PRs on my phone.

Either way, my point stands. It's not a mock, and I fail to see how `req` could be ambiguous in the context of a views test.

seanh: Would `stub_request` or `dummy_request` also be options? It's not that `req` is ambiguous it's just that I find it overly terse, I think it puts extra mental work on the reader to have to decode and then remember that req means request, ctx means context, etc etc.

nickstenning:
Seeing as you've merged #3432 which used the same pattern in the tests, it now seems like a bit of a silly exercise to go and rename everything from `req`, not least because:

```
~/src/h $ ag '\breq\b' tests/ | wc -l
     125
~/src/h $ ag '\bstub_request\b' tests/ | wc -l
       0
~/src/h $ ag '\bdummy_request\b' tests/ | wc -l
      11
```

PR #3430, opened 2016-06-09
Remove User.{admins,staff_members} classmethods

nickstenning:
These helpers used the class query property, which we're trying to remove. In addition, it's not really clear that they provided a useful abstraction, given that `staff_members` was used in only one place, and `admins` was used in two places, one of which resulted in a less efficient use of the database as a result of the indirection.

This commit removes both of them and simply uses SQLAlchemy in the view functions to achieve the same effect without relying on the "magic" query property.


chdorner:
This is looking good to me, has to wait for #3429 to be merged, so I'll leave it open.

PR #3434, opened 2016-06-09
Append URI from document to PDF incontext links

chdorner:
We now use the PDF fingerprint as the target_uri of the annotation. In
order to still be able to append a potential HTTP URL to the incontext
link, we are trying to load an HTTP URL from the document uris.

The same work still needs to be done in Bouncer.


nickstenning:
LGTM.

PR #3436, opened 2016-06-09
Make thread building more idiomatic by using memoization

robertknight:
In Redux, the idiomatic way to transform the normalized state of the application into a form that maps closer to what the UI looks like is to do:
```
  derivedData = transform(select(store.getState()))
```
Where the `select` function extracts the relevant fields from the state and the `transform` function then computes the derived data. In our case, the input state is the list of annotations, current selection/sort etc. and the derived data is a tree-structure of annotation threads and replies that are visible according to the current UI settings.

Because the store state is immutable, both `select` and `transform` can be trivially memoized to avoid unnecessary recalculations and view updates. In other words, if `select` is called with the same object twice it can just return the previously calculated value on the second call and likewise for `transform`.

Modifying the `RootThread` class to use this pattern simplifies it by avoiding inheritance from EventEmitter. In future this will make it easier to avoid rebuilding the thread if parts of the state that do not affect the visible threads changes.

### DIFF
--- a/h/static/scripts/annotation-viewer-controller.js
+++ b/h/static/scripts/annotation-viewer-controller.js
@@ -17,17 +17,17 @@ function AnnotationViewerController (
     $location.path('/stream').search('q', query);
   };
 
-  rootThread.on('changed', function (thread) {
+  function thread() {
+    return rootThread.thread(annotationUI.getState());
+  }
+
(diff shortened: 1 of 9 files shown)


nickstenning:
LGTM.

PR #3437, opened 2016-06-09
Allow admins to toggle specific features for a features cohort

finitud:


h/admin/views/features.py
```diff
@@ -52,8 +52,11 @@ def cohorts_index(context, request):
 def cohorts_add(request):
     """Create a new feature cohort."""
     cohort_name = request.params['add']
-    cohort = models.FeatureCohort(name=cohort_name)
-    request.db.add(cohort)
+    if cohort_name not in models.FeatureCohort.RESERVED_FEATURECOHORT_NAMES:
```
nickstenning: Couple of different things to say to this:

1. Don't worry about it -- it's an admin interface. If people create a cohort with a silly name we can address that by communication rather than code.
2. If you did want to do this it would probably be better to implement it further back (i.e. on the model, using a [validator](http://docs.sqlalchemy.org/en/latest/orm/mapped_attributes.html#simple-validators)) rather than in the view. (Although the view would need to know how to respond to a validation error.)
3. Namespaces are great! If you're storing a set of reserved names on the `FeatureCohort` class, then it's clear they refer to reserved names for feature cohorts, so you can just call it `RESERVED_NAMES` rather than `RESERVED_FEATURECOHORT_NAMES`.

In summary, though, I don't think this is necessarily needed s

h/features/models.py
```diff
@@ -35,6 +35,23 @@
 #
 FEATURES_PENDING_REMOVAL = {}
 
+FEATURE_FEATURECOHORT_TABLE = sa.Table(
+    'feature_featurecohort', Base.metadata,
```
nickstenning: For the sake of consistency, let's have both join tables in the same part of the file as one another, and named the same way. That is: `featurecohort_feature` (to go with `featurecohort_user`) rather than `feature_featurecohort`).

h/features/models.py
```diff
@@ -35,6 +35,23 @@
...
+    sa.Column('id',
+              sa.Integer(),
+              nullable=False),
+    sa.Column('feature_id',
+              sa.Integer(),
+              nullable=False),
+    sa.Column('cohort_id',
+              sa.Integer(),
+              nullable=False),
+    sa.PrimaryKeyConstraint('id'),
```
nickstenning: Perhaps use the `primary_key` and `autoincrement` keyword arguments as you've done for the user join table.

h/features/models.py
```diff
@@ -35,6 +35,23 @@
...
+    sa.ForeignKeyConstraint(['cohort_id'], ['featurecohort.id']),
```
nickstenning: Again, let's be consistent in how we do these things and use the same pattern as you used for the user join table below.

h/features/models.py
```diff
@@ -63,6 +80,10 @@ class Feature(Base):
                       default=False,
                       server_default=sa.sql.expression.false())
 
+    featurecohorts = sa.orm.relationship('FeatureCohort',
```
nickstenning: This can probably just be `cohorts` -- because a `cohorts` attribute on a `feature` is self-evidently in reference to feature cohorts as opposed to any other kind of cohort.

h/features/models.py
```diff
@@ -63,6 +80,10 @@ class Feature(Base):
...
+                                         secondary=FEATURE_FEATURECOHORT_TABLE,
+                                         backref='cohorts')
```
nickstenning: The `backref` property determines the name of the attribute on the *other side* of the relationship. So this would put a `cohorts` attribute on instances of the `FeatureCohort` class, which I don't imagine is what you intended.

h/features/models.py
nickstenning: That said, I think it would probably make more sense to define this relationship on the `FeatureCohort` class rather than here. Think of it this way: features don't need feature cohorts to work, but the reverse is not true. It's nicer to be able to remove the functionality just by deleting the `FeatureCohort` class than if you have to go and remove relationships from other objects.

h/features/models.py
finitud: Makes sense.

h/admin/views/features.py
```diff
@@ -29,6 +32,15 @@ def features_save(request):
                 setattr(feat, attr, True)
             else:
                 setattr(feat, attr, False)
+        for cohort in request.db.query(models.FeatureCohort).all():
+            val = request.POST.get('{0}[{1}]'.format(feat.name, cohort.name))
```
nickstenning: Ok, I see now why you added stuff to precent cohorts called 'everyone', 'admin' and 'staff'. I'd suggest naming the form fields in such a way that this can't cause a problem, such as `<featurename>[cohort][<cohortname>]`.

h/admin/views/features.py
```diff
@@ -29,6 +32,15 @@ def features_save(request):
                 setattr(feat, attr, True)
             else:
                 setattr(feat, attr, False)
+        for cohort in request.db.query(models.FeatureCohort).all():
```
nickstenning: This function is now getting to the point where it needs splitting up into some helpers, I think. Leave it for now, but in future maybe consider factoring stuff out when it gets this busy.

h/features/client.py
```diff
@@ -79,4 +79,7 @@ def _state(self, feature):
         # member.
         if feature.staff and role.Staff in self.request.effective_principals:
             return True
+        for cohort in feature.cohorts:
```
nickstenning: This would probably be clearer (and would definitely be more efficient) if written as:

```python
if set(feature.cohorts) & set(self.request.authenticated_user.cohorts):
    return True
```

tests/h/features/client_test.py
```diff
@@ -125,11 +133,13 @@ def fetcher(self):
...
             Feature(name='on-for-staff', staff=True),
             Feature(name='on-for-admins', admins=True),
+            Feature(name='on-for-cohort', cohorts=[FeatureCohort(name='cohort')])
         ])
 
     @pytest.fixture
     def req(self):
-        return DummyRequest(db=mock.sentinel.db_session)
+        return DummyRequest(db=mock.sentinel.db_session,
+                            authenticated_user='acct:foo@example.com')
```
nickstenning: This is a bit weird, because `authenticated_user` isn't a string in the real application, it's a user object. I suspect you'll discover that this will need to change anyway if you address the comments above.

tests/h/features/client_test.py
```diff
@@ -71,6 +71,15 @@ def test_call_false_if_everyone_false(self, client):
     def test_call_true_if_everyone_true(self, client):
         assert client('on-for-everyone') is True
 
+    def test_call_false_if_cohort_disabled(self, client):
+        assert client('on-for-cohort') is False
+
+    def test_call_true_if_cohort_enabled(self, patch, client, user, cohort):
+        cohort_model = patch('h.features.models.FeatureCohort')
+        cohort_model.members = [user]
```
nickstenning: Is this even needed now?

finitud: It isn't. Removing.

nickstenning:
LGTM.

PR #3438, opened 2016-06-09
Simplify annotation key event handling

robertknight:
Rather than imperatively adding a 'keydown' event handler in the
<annotation> link function, use `ng-keydown` to set up an event handler
instead.

With this we no longer need a link function.


nickstenning:
LGTM.

PR #3441, opened 2016-06-09
Remove path for local pdf files.

sheetaluk:
Currently, when annotating a local PDF, the `file://` URL will be submitted to the server.
It is not appropriate to submit local file paths to the server.
Since entries in document.link should be URIs, omit the URL in the case of local files.

https://trello.com/c/K8t2kl9m/331-don-t-save-file-urls-in-document-metadata
53725fd

Also, closed the PR below in favour of this one
https://github.com/hypothesis/h/pull/3405


robertknight:
LGTM. One extra thing that would be useful to have is a test that HTTP(S) URLs _are_ saved with the document.

PR #3443, opened 2016-06-10
Update to Angular 1.5.6

robertknight:
Alongside the usual mix of bug fixes and performance improvements, this brings support for lifecycle hooks which get called when components are mounted, when their inputs are changed and when they are destroyed, as a cleaner alternative to `$scope.$watch` hooks.

See https://github.com/angular/angular.js/commit/9cd9956d


nickstenning:
Did you mean to use "^1.5.6"? Are we confident that Angular 1.9 isn't going to break our code?

robertknight:
> Did you mean to use "^1.5.6"? Are we confident that Angular 1.9 isn't going to break our code?

Recent Angular upgrades have been very smooth* but fair point, I'm not completely confident that Angular 1.6 couldn't cause any breakage. I've changed the version matching to match any 1.5.x release instead which I think is sufficiently safe.

nickstenning:
Seems ok to me. At some point I think you might want to consider shrinkwrap... we're very likely going to do something similar for the Python application.

PR #3444, opened 2016-06-10
Extract timestamp out into its own component

robertknight:
This makes it easier to test the refresh logic directly and decouples it
from the rest of the annotation display.

### DIFF
--- a/h/static/scripts/directive/test/util.js
+++ b/h/static/scripts/directive/test/util.js
@@ -151,6 +151,12 @@ function createDirective(document, name, attrs, initialScope, initialHtml, opts)
     element.scope = childScope;
     childScope.$digest();
     element.ctrl = element.controller(name);
+
+    if (!element.ctrl) {
+      throw new Error('Failed to create "' + name + '" directive in test.' +
+        'Did you forget to register it with angular.module(...).directive() ?');
+    }
+
     return element;
   };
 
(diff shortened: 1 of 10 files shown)


h/static/scripts/directive/test/util.js
```diff
@@ -151,6 +151,12 @@ function createDirective(document, name, attrs, initialScope, initialHtml, opts)
     element.scope = childScope;
     childScope.$digest();
     element.ctrl = element.controller(name);
+
+    if (!element.ctrl) {
```
robertknight: I added this because it bit me a few times. This is also why no-op controllers were added to 3 components.

h/static/scripts/directive/timestamp.js
```diff
@@ -0,0 +1,67 @@
+'use strict';
+
+var dateUtil = require('../date-util');
+
+// @ngInject
+function TimestampController($scope, time) {
+  var vm = this;
+
+  /** A fuzzy, relative (eg. '6 days ago') format of the annotation's
```
nickstenning: Perhaps update the comments to not reference annotations specifically, now that this is a general purpose directive?

nickstenning:
LGTM other than a small comment about docs.

nickstenning:
👍

PR #3446, opened 2016-06-10
Merge documents after moving annotations

chdorner:
Try and merge documents after replacing uris.
The page under the new URI might already have gotten new annotations.
Which means that in this case we have to merge these two documents
together.


nickstenning:
LGTM.

PR #3447, opened 2016-06-10
Pg cleanup

seanh:


seanh:
Looks good to me, I've briefly tested on prod data and the rss, atom and groups pages seem to be working fine

chdorner:
Just gave the feeds and the groups pages a quick test run and it's looks fine.

PR #3448, opened 2016-06-10
Don't return None in group.documents()

seanh:
group.documents() shouldn't return None in the list of the group's documents
when the group has an annotation that has no document.

Some annotations have no document and annotation.document will be None.
In this case nothing from this annotation should be added to the list of
documents, it should not return None in the list of documents that it
returns.


seanh:
Oh, I accidentally committed those profile files again

seanh:
Fixed

nickstenning:
LGTM.

PR #3449, opened 2016-06-10
Remove __getattr__() from AnnotationHTMLPresenter

seanh:
This was convenient but can also cause some tricky behaviour when
@property methods of AnnotationHTMLPresenter or DocumentHTMLPresneter
fail with things like AttributeError and then it falls back on
__getattr__() and tries to get the attribute from Annotation.

Simple and explicit is better, less confusing and unexpected behaviour.

For convenience explicitly forward some Annotation properties that the
h.feeds uses and that were previously relying on the __getattr__().


nickstenning:
LGTM.

PR #3451, opened 2016-06-10
Don't crash when annotation has no document

seanh:
AnnotationHTMLPresenter hadn't been written to handle the case when an
annotation has no document, and it crashes.

It's not clear what AnnotationHTMLPresenter properties such as
.document_link etc should return in this case. But for now just make
them return empty strings - at least it's better than crashing.


nickstenning:
LGTM.

PR #3453, opened 2016-06-11
Fix adder not being shown if there is a selection when the plugin is activated

### LINKED ISSUE #3452: Adder no longer automatically shows when text is selected on extension activate.
We just shipped an enhancement that would allow you to select some text before the extension was on, and preserve the selection through activation, surfacing the Adder for you.

That appears to have regressed.

### Steps to reproduce
1. Select some text
2. Click on the extension

### Expected behaviour
Adder should deploy

### Actual behaviour
It doesn't

Version:
0.27.0
User agent:
Mozilla/5.0 (Macintosh; Intel Mac OS X 10_10_5) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/51.0.2704.84 Safari/537.36
URL:
http://www.huffingtonpost.com/entry/david-de-gea-gang-rape_us_575af789e4b00f97fba82bea
Username:
dwhly
Date:
10 Jun 2016 14:25:09 -0700

robertknight:
`selections()` tried to listen for a 'ready' event on the DOM Document
object. The 'ready' event however is jQuery-specific however so that
never gets triggered.

Instead change the `selections()` function to perform an initial check
for a selection on the next tick and emit a null|DOMRange.

Fixes #3452

### DIFF
--- a/h/static/scripts/annotator/selections.js
+++ b/h/static/scripts/annotator/selections.js
@@ -2,11 +2,24 @@
 
 var observable = require('../util/observable');
 
+/** Returns the selected `DOMRange` in `document`. */
+function selectedRange(document) {
+  var selection = document.getSelection();
+  if (!selection.rangeCount || selection.getRangeAt(0).collapsed) {
+    return null;
+  } else {
+    return selection.getRangeAt(0);
+  }
+}
+
 /**
  * Returns an Observable stream of text selections in the current document.
  *
  * New values are emitted when the user finishes making a selection
  * (represented by a `DOMRange`) or clears a selection (represented by `null`).
+ *
+ * A value will be emitted with the selected range at the time of subscription
+ * on the next tick.
  *
  * @return Observable<DOMRange|null>
  */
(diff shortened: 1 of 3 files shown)


nickstenning:
Any reason you wouldn't just swap out `ready` for `DOMContentLoaded`?

h/static/scripts/annotator/selections.js
```diff
@@ -1,13 +1,28 @@
 'use strict';
 
+var Observable = require('zen-observable');
```
nickstenning: Requiring both `Observable` and `observable` in the same file is pretty ugly. It looks like `../util/observable` already exposes the constructor so you could just use `observable.Observable.of` if necessary...?

robertknight:
> Any reason you wouldn't just swap out ready for DOMContentLoaded?

In this case the code is usually being run after `DOMContentLoaded` has been emitted. I believe the reason that this was working previously is that jQuery's ready is fired when the library loads if the DOM is already initialized.

nickstenning:
Right, and actually the whole thing is a bit weird because Annotator is initialized in a domready handler: https://github.com/hypothesis/h/blob/master/h/static/scripts/annotator/main.js#L37

nickstenning:
(So the initial check seems like the right approach.)

nickstenning:
LGTM.

PR #3454, opened 2016-06-13
Initialize `tags` and `text` fields of all annotations loaded into the app

robertknight:
Several places in the `<annotation>` component assumed that the `tags`
field was an array, as did `<tag-editor>`. The `tags` and `text` fields
are always set by the server for annotations received from the API but
are not set by Annotator when creating new annotations in the page.

Initializing these fields in the reducer, which is the only place that
the loaded set of annotations can be modified, means that the rest of
the app can rely on them having the expected type.

This fixes an error when saving new annotations.

### DIFF
--- a/h/static/scripts/annotation-ui.js
+++ b/h/static/scripts/annotation-ui.js
@@ -84,11 +84,31 @@ function excludeAnnotations(current, annotations) {
   });
 }
 
+/**
+ * Initialize all required fields of `annot` to the expected types.
+ *
+ * Initializing these fields when annotations are added allows the rest of the
+ * app to assume their existence.
+ */
+function addRequiredFields(annot) {
+  // Modify `annot` rather than returning a copy because `annot` is currently
+  // a `store.AnnotationResource` instance rather than a plain object and has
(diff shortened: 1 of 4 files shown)


nickstenning:
Why not fix the places that assume `tags` has to be set? It seems to me like it would be less code and less brittle to changes in the API?

robertknight:
> Why not fix the places that assume tags has to be set?

I think it will lead to less code overall (fewer checks for null-ness of the `tags` field) but more importantly fewer mistakes from code that fails to account for the possibility that `tags` is null. While investigating the initial error I saw in the console I found a couple of other places that incorrectly made the same assumption.

nickstenning:
I'm honestly not sure. Don't forget, we're trying to build something that in the end will ingest annotation data from all kinds of sources, and I don't know that building in assumptions like "all annotations have tags fields" is a great idea.

robertknight:
Had some discussion on Slack about the last comment. We decided to initialize the tags and text fields but in the same place where we initialize other required fields. I'll close this and create a separate PR

PR #3457, opened 2016-06-13
Do not use query property in Feature.{all,remove_old_flags}

nickstenning:
The use of a "magic" query property makes testing difficult and causes
confusion about which session is the correct session to use when writing
new code.

This commit is part of an effort to standardise all ORM classmethods so
they take a database session as their first parameter and do not rely on a
thread-local query property.


chdorner:
LGTM.

PR #3458, opened 2016-06-13
Add a NIPSA service to replace h.nipsa.logic

nickstenning:
One approach to limiting coupling between views and the models they control is to introduce a service layer which is responsible for encapsulating business logic and dependencies of that logic on other parts of the application.

This commit uses the `pyramid_services` library, which is a small library formalising a service layer pattern, in which services are registered with the application registry by name or by interface (I've chosen to go with name here, for the sake of simplicity). These services can then be retrieved by anything which can call `request.find_service(...)`.

Crucially, one can register service factories which in turn have access to the request object and thus can extract other dependencies (such as the database session) without calling code having to pass them in. This is the key difference between this approach and the more direct "logic" module approach.

### DIFF
--- a/h/nipsa/services.py
+++ b/h/nipsa/services.py
@@ -1,0 +1,63 @@
+# -*- coding: utf-8 -*-
+
+from h.nipsa import models
+from h.nipsa import worker
+
+
+class NipsaService(object):
+
+    """
+    A service which provides access to the state of "not-in-public-site-areas"
+    (NIPSA) flags on userids.
+    """
+
+    def __init__(self, session):
+        self.session = session
+
+    @property
+    def flagged_userids(self):
+        """
+        A list of all the NIPSA'd user IDs.
+
+        :rtype: list of unicode strings
+        """
+        return [u.userid for u in self.session.query(models.NipsaUser)]
+
+    def is_flagged(self, userid):
+        """Return whether the given userid is flagged as "NIPSA"."""
+        return self._user_query(userid).one_or_none() is not None
+
+    def flag(self, userid):
+        """
+        Add a NIPSA flag for a userid.
+
+        Add the given user's ID to the list of NIPSA'd user IDs. If the user
+        is already NIPSA'd then nothing will happen (but an "add_nipsa"
+        message for the user will still be published to the queue).
+        """
+        nipsa_user = self._user_query(userid).one_or_none()
+        if nipsa_user is None:
+            nipsa_user = models.NipsaUser(userid)
+            self.session.add(nipsa_user)
+
+        worker.add_nipsa.delay(userid)
+
+    def unflag(self, userid):
+        """
+        Remove the NIPSA flag for a userid.
+
+        If the user isn't NIPSA'd then nothing will happen (but a
+        "remove_nipsa" message for the user will still be published to the
+        queue).
+        """
+        self._user_query(userid).delete()
+
+        worker.remove_nipsa.delay(userid)
+
+    def _user_query(self, userid):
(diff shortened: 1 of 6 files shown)


nickstenning:
Note that the service is entirely unused in this commit: I've just split this into it's own PR so it's easier to review. If you want to see how this simplifies things later on, see [this commit](https://github.com/hypothesis/h/pull/3459/commits/ac6a622888c99fc45437d2382c780fbc11bbd97a) from #3459.

chdorner:
Looks like a nice approach. Will let @seanh take a look as well.

h/nipsa/services.py
```diff
@@ -0,0 +1,63 @@
...
+        Add a NIPSA flag for a userid.
+
+        Add the given user's ID to the list of NIPSA'd user IDs. If the user
+        is already NIPSA'd then nothing will happen (but an "add_nipsa"
+        message for the user will still be published to the queue).
+        """
+        nipsa_user = self._user_query(userid).one_or_none()
+        if nipsa_user is None:
+            nipsa_user = models.NipsaUser(userid)
+            self.session.add(nipsa_user)
```
seanh: Not important here, and presumably not this PR's fault, but this could crash if two requests tried to nipsa the same user at the same time couldn't it?

tests/h/nipsa/services_test.py
```diff
@@ -0,0 +1,95 @@
...
+
+from h.nipsa.models import NipsaUser
+from h.nipsa.services import NipsaService
+from h.nipsa.services import nipsa_factory
+
+
+@pytest.mark.usefixtures('nipsa_users', 'worker')
+class TestNipsaService(object):
+    def test_flagged_userids_returns_list_of_userids(self, db_session):
+        svc = NipsaService(db_session)
```
seanh: Could be an `@fixture` in the `TestNipsaService` class

tests/h/nipsa/services_test.py
```diff
@@ -0,0 +1,95 @@
...
+    request = DummyRequest(db=mock.sentinel.db_session)
+
+    svc = nipsa_factory(None, request)
+
+    assert isinstance(svc, NipsaService)
+    assert svc.session == mock.sentinel.db_session
+
+
+@pytest.fixture
+def nipsa_users(db_session):
```
seanh: Not important, but this could be in the `TestNipsaService` class which is the only place it's used

seanh:
Nice. I think maybe just replacing the repeated `svc = NipsaService(db_session)` in the tests with a fixture would be worthwhile? Otherwise looks great

h/nipsa/services.py
nickstenning: Sure, but I don't think that's very likely -- certainly not as likely as two people trying to annotate the same page at the same time. There's a big difference between the two situations, too: in the annotation metadata situation the result was that the entire transaction failed and the annotation wasn't created. In this situation the request will error, but the database will have ended up in the correct state regardless.

tests/h/nipsa/services_test.py
nickstenning: Hmm. Maybe. Not sure about making the system-under-test a fixture unless constructing is particularly burdensome.

tests/h/nipsa/services_test.py
nickstenning: Does it matter where it's defined? Less indentation == good?

seanh: In this case I don't suppose it does matter. In larger test files with lots of test classes I've found it useful to put fixtures only used by one class in that class, just for the sake of namespacing and of keeping related code together

PR #3459, opened 2016-06-13
Reimplement NIPSA operations using NipsaService

nickstenning:
This commit removes h.nipsa.logic and replaces code that used it with code written to use the new NipsaService. This has the advantage that calling code now doesn't need to concern itself with the dependency the NipsaService has on the database session.

This results in a simplification of the NipsaUser model, and a removal of any reliance on the h.db.Base query property.


seanh:
Rebased

PR #3460, opened 2016-06-13
Initialize `tags` and `text` fields to expected types for new annotations

robertknight:
Several places in the `<annotation>` component assumed that the `tags`
field was an array, as did `<tag-editor>`. The `tags` and `text` fields
are always set by the server for annotations received from the API but
are not set by Annotator when creating new annotations in the page.

It would be better to initialize all of the required fields for
annotations when they are loaded into the app rather than in the
`<annotation>` component. For now however keep the initialization of all
required fields in the same place.

This fixes an error when saving new annotations.


nickstenning:
LGTM.

PR #3461, opened 2016-06-13
Load annotations from DB for /api/search

chdorner:
Similar to what we're doing with the RSS/Atom feeds. After searching in Elasticsearch, we'll load the annotations from the database.
This gets rid of the last `annotation_from_dict` usage in the code (except for the annotation test factory).

This also adds the option for `storage.fetch_ordered_annotations` to eager load the document data.

### DIFF
--- a/h/api/storage.py
+++ b/h/api/storage.py
@@ -8,6 +8,7 @@ assumed to be validated.
 """
 
 from pyramid import i18n
+from sqlalchemy.orm import subqueryload
 
 from h.api import schemas
 from h.api import models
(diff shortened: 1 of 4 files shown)


h/api/storage.py
```diff
@@ -49,13 +50,21 @@ def fetch_annotation(session, id_):
...
-    anns = session.query(models.Annotation).filter(models.Annotation.id.in_(ids)).all()
+    query = session.query(models.Annotation).filter(models.Annotation.id.in_(ids))
+
+    if load_documents:
+        query = query.options(
+            subqueryload(models.Annotation.document).subqueryload(models.Document.document_uris),
+            subqueryload(models.Annotation.document).subqueryload(models.Document.meta)
+        )
+
+    anns = query.all()
```
nickstenning: This isn't necessary -- just do `anns = sorted(query, key=...)`.

nickstenning:
Other than one tiny niggle, this looks great.

nickstenning:
LGTM.

PR #3462, opened 2016-06-13
Optimize feed annotations loading

chdorner:
This is based on #3461 and will need rebasing once that one is merged.
It enables the eager loading of document data for the RSS/Atom feeds.


nickstenning:
LGTM.

PR #3463, opened 2016-06-13
Fix cursor style for annotation cards

robertknight:
Set the annotation cursor to a pointer for annotation cards to indicate
that it is clickable (it scrolls the page to that annotation) except for
the body which has a text cursor to indicate that the text is
selectable.


nickstenning:
LGTM.

PR #3464, opened 2016-06-13
Unify styling of annotation quotes and quotes in annotation bodies

### LINKED ISSUE #3435: Selection styling and quoted text styling no longer match.
### Steps to reproduce
1. Quote some text in the body of an annotation
2. Compare that styling to the selection of that annotation.

### Expected behaviour
They should be the same

### Actual behaviour
They're not. One is serif the other is sans.

Chrome extension

### Additional details
screenshot 2016-06-09 06 36 23

@conordelahunty ^

robertknight:
Make the styling of quoted snippets of the page consistent with the
styling of block quotes in annotation bodies.

The styling for `<del>` and `<ins>` elements in quotes has been removed
because it is not currently used.

Fixes #3435

### DIFF
--- a/h/static/styles/annotation.scss
+++ b/h/static/styles/annotation.scss
@@ -171,18 +171,6 @@ $annotation-card-left-padding: 10px;
 
 .annotation-quote {
   @include quote;
-  @include font-normal;
-  color: $grey-4;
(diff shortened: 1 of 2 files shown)


nickstenning:
LGTM.

PR #3466, opened 2016-06-13
Fix concurrency crash when creating document metadata

### LINKED ISSUE #3465: Crash when two requests simultaneously create document metadata

1. Browse to a page that:

3. Annotate the page

seanh:
Fixes #3465.

Creating a new annotation can crash when two concurrent requests are
trying to create new annotations of the same document:

My theory is that this can happen due to a race condition / concurrency
problem when two people simultaneously try to annotate a page that has
not been annotated before:

1. Browse to a page that:

   a. Will generate more than one `DocumentMeta` (e.g. has a title and
   favicon)

   b. Has not been annotated yet (or more accurately does not currently
   have any `DocumentMeta`s in the db)

2. Put a `set_trace()` before
   [this line](https://github.com/hypothesis/h/blob/46948765cc7e49b18b7d8030ad29b9f137e86de0/h/api/models/document.py#L413)

3. Annotate the page

4. The first time it hits your `set_trace()` hit `c` for continue,
   allowing it to create the first of two `DocumentMeta`s. I suppose the
   ordering here is undefined but for me it was always the title
   `DocumentMeta` that got created first. This creates the
   `DocumentMeta` in the `sqlalchemy` session but doesn't flush or
   commit it yet. The code should now be stopped on your `set_trace()`
   for the second (and last) time.

5. Go over to a `psql` shell for your db and insert a `DocumentMeta`
   with the same `claimant_normalized` and `type` as the one that was
   created in the previous step: `INSERT INTO document_meta
   (document_id, claimant, claimant_normalized, type, value) VALUES
   (1128, 'http://localhost:4545/', 'http://localhost:4545', 'title',
   '{"foo"}');`

6. Go back and hit `c` to continue processing the request. The query for
   existing `DocumentMeta`s on
   [this line](https://github.com/hypothesis/h/blob/46948765cc7e49b18b7d8030ad29b9f137e86de0/h/api/models/document.py#L318)
   will cause the `DocumentMeta` that was created in step 4 to be
   flushed, which will violate a database constraint because of the
   `DocumentMeta` that you added in step 5, and you will see the
   traceback.

The problem is that `create_or_update_document_meta()` has the logic:
"If there's already a DocumentMeta then return it, else create a new one" and
this cannot work because a concurrent request may create a new
DocumentMeta in-between the "if" and the "create a new one" and this
will then cause the "create a new one" to crash with an IntegrityError
from the database.

It is not possible to simply catch this IntegrityError and return the
DocumentMeta that was inserted by the concurrent request instead,
because once the IntegrityError has happened it's necessary to rollback
the transaction and this will lose any other objects that've been added
to the session outside of `create_or_update_document_meta()` but not
committed yet (e.g. `DocumentURI`s, the new `Annotation`).

So what we have to do is create a nested transaction for the "get the
existing DocumentMeta or create a new one" logic so that if "create a
new one" fails because of a concurrent request we can simply rollback
this nested transaction and return the `DocumentMeta` created by the
concurrent request instead without losing any other work already done on
the current request's session.

### DIFF
--- a/h/api/models/document.py
+++ b/h/api/models/document.py
@@ -266,6 +266,51 @@ def create_or_update_document_uri(session,
     docuri.updated = updated
 
 
+def get_or_create_document_meta(session,
+                                claimant,
+                                type,
+                                value,
+                                document,
+                                created,
+                                updated):
+    """
+    Return the DocumentMeta with the given claimant_normalized and type.
+
+    If no such DocumentMeta exists in the database then create a new one using
+    the given arguments and return it.
+
+    """
+    def get_existing_document_meta():
+        """Return a DocumentMeta with claimant_normalized and type or None."""
+        return session.query(DocumentMeta).filter(
+            DocumentMeta.claimant_normalized == uri_normalize(claimant),
+            DocumentMeta.type == type).one_or_none()
+
+    document_meta = get_existing_document_meta()
+
+    if document_meta:
+        return document_meta
+
+    # There wasn't an existing DocumentMeta in the db, so create a new one.
+    try:
+        with session.begin_nested():
+            document_meta = DocumentMeta(claimant=claimant,
+                                         type=type,
+                                         value=value,
+                                         document=document,
+                                         created=created,
+                                         updated=updated,
+                                         )
+            session.add(document_meta)
+    except sa.exc.IntegrityError:
+        # It looks like a concurrent request added an equivalent DocumentMeta
+        # before we could save ours.
+        document_meta = get_existing_document_meta()
+        assert document_meta
+
+    return document_meta
+
+
 def create_or_update_document_meta(session,
                                    claimant,
                                    type,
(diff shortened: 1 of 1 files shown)


seanh:
It's still possible for a concurrent request to make this crash if:

1.After `get_or_create_document_meta()` does a `SELECT` to find the existing meta, but before it commits the transaction in which is `INSERT`s a new meta, a concurrent request `INSERT`s an equivalent meta.  The code will go to the `except sa.exc.IntegrityError` block where it `SELECT`s and returns the meta created by the concurrent request.

2. Before `get_or_create_document_meta()` does the `SELECT` to find the meta created by the concurrent request another concurrent request deletes that meta.

[This assert](https://github.com/hypothesis/h/pull/3466/commits/89eb550ff78db9b86c4ecd24c3d2c997a8516a13#diff-90afc90779f2946eb7d36018acbf9f55R301) can fail (or if the `assert` wasn't there then you'd get an `AttributeError` somewhere later on.

This is obscure enough that I think we can ignore it? Since it requires two concurrent requests to both insert and then delete an equivalent meta at just the right times.

Other than that, I can't find any way for a concurrent request to make the code crash by just `INSERT`ing at devious times,

seanh:
This is still WIP because it needs tests and also `get_or_create_document_meta()` should maybe be private

seanh:
Closed in favor of https://github.com/hypothesis/h/pull/3478

PR #3467, opened 2016-06-13
Add self-documenting 'help' target to Makefile

finitud:
This is just a small "niceness" taken from:

http://marmelab.com/blog/2016/02/29/auto-documented-makefile.html

I've opted for leaving the default make rule as it is (run the tests), but if people like this help option it can be made the default rule.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,6 +17,7 @@ default: test
 build/manifest.json: node_modules/.uptodate
 	$(NPM_BIN)/gulp build
 
+## Clean up runtime artifacts (needed after a version update)
 .PHONY: clean
 clean:
 	find . -type f -name "*.py[co]" -delete
@@ -24,6 +25,7 @@ clean:
 	rm -f node_modules/.uptodate h.egg-info/.uptodate
 	rm -rf build dist
 
+## Run the development H server locally
 .PHONY: dev
 dev: build/manifest.json h.egg-info/.uptodate
 	@hypothesis devserver
@@ -37,10 +39,12 @@ dist/h-$(BUILD_ID).tar.gz:
 dist/h-$(BUILD_ID): dist/h-$(BUILD_ID).tar.gz
 	tar -C dist -zxf $<
 
+## Build hypothesis/hypothesis docker image
 .PHONY: docker
 docker: dist/h-$(BUILD_ID)
 	docker build -t hypothesis/hypothesis:$(DOCKER_TAG) $<
 
+## Run test suite
 .PHONY: test
 test: node_modules/.uptodate
 	@pip install -q tox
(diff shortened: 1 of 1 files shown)


nickstenning:
This is a neat trick, but I'm a bit dubious about the value. `make help` isn't a standard way of finding out what's in a Makefile, so the only way I'd actually discover that this command exists is by... reading the Makefile.

Happy to go by what others think but for me this swings closer to "inscrutable magic" than "helpful documentation."

nickstenning:
(FWIW I'm all in favour of adding some better comments to the Makefile. Just not sure about the `help` target trickery.)

finitud:
It's not standard, but it's not unheard of... probably works best if it's the default target, and with a bigger `Makefile` (ours is quite simple as far as `Makefile`s go). Happy to drop this one though if it doesn't really help.

chdorner:
I agree that `make help` isn't really common knowledge yet. But if we can agree on the comment going at the end of the task definition line, then it doesn't hurt to keep the `make help` task imo.

nickstenning:
I'm really not a fan of changing the default target. There is a long history of `make` doing something useful, not printing help (thing `./configure && make && make install`).

Perhaps the best solution here is just to have the `make help` target but just write it explicitly:

```
.PHONY: help
help:
        @echo help        Display this help message
        @echo clean       Remove build artifacts
```

Etc. Thoughts?

chdorner:
I do like having the documentation right where the task definitions are. [Here's](https://gist.github.com/klmr/575726c7e05d8780505a) a version that supports multi-line help comments (although the implementation of the help task is pretty insane).

finitud:
I agree that the Unix tradition is to have `make` do the build (which in our case is running the tests). I'm split between the two options of having documentation on each target and have a magic `help` rule, or having an explicit `help` rule as @nickstenning suggests.

nickstenning:
>  Here's a version that supports multi-line help comments

We will add that make task to our Makefile over my dead body 😂...

nickstenning:
There's a difference between comments, which explain how and why a task does something a particular way, and the help task, which is intended to provide a short summary of available tasks. Why not have both?

If we can avoid having something as fragile and "magical" as that help task in our Makefile, that would be good, IMO.

chdorner:
> There's a difference between comments, which explain how and why a task does something a particular way, and the help task, which is intended to provide a short summary of available tasks. Why not have both?

This sounds like a good middle ground to me :+1:

finitud:
Sounds good to me as well, will do this later today.

finitud:
I have rewritten this one so that targets keep the comments (but in a separate line for better readability), and the `help` target is now explicit.

nickstenning:
Not sure it's "self-documenting" any more, but LGTM.

PR #3469, opened 2016-06-13
Fixes #3264 with whitespace: no-wrap and flex-shrink: 0

### LINKED ISSUE #3264: CSS issue with sidebar search input
Opening and closing the sidebar search _while signed out_ causes the group display to break:

untitled screencast - edited 5

(P.S. Love the first annotation that every new user of h sees there)

bZichett:


h/static/styles/top-bar.scss
```diff
@@ -31,10 +31,15 @@
   padding-right: $h-padding;
 
   height: 100%;
+
+  .signin-control{
```
robertknight: We try to avoid nested selectors, so this declaration should go at the top-level of the file if possible.

robertknight:
Thank-you for this! I'll take a closer look later but the change looks sensible to me.

bZichett:
Yep good policy.  I missed the dedicated signin-control.scss file last commit anyway, and that made much more sense anyway, so the small class is placed there now.

Also, not sure if it matters too much ... usually I'd personally try to avoid applying css to an angular generated tag name ... but the class wouldn't have had be created if it was just applied to the "signin-control" directive tag.

robertknight:
> Also, not sure if it matters too much ... usually I'd personally try to avoid applying
> css to an angular generated tag name

Agreed. I usually treat the custom HTML element as a hidden implementation detail and instead style the top-level span/div element inside it.

robertknight:
LGTM. Thank-you very much @bZichett !

PR #3474, opened 2016-06-15
Fix standalone annotation reply pages

robertknight:
This is a proposed fix for standalone pages for replies. Rather than just showing the reply itself, it fetches and displays the whole conversation thread. I think this provides better context for the user:

reply-standalone-page

Something that is missing from the above but I would like to include in a follow-up PR is to highlight the reply that was referenced in the URL. This is something we can then also use when direct-linking to replies.

The implementation uses the existing API. The app first fetches the annotation specified in the URL, if it is a top-level annotation it fetches the replies, otherwise it first fetches the top-level parent and then fetches the replies.

A case I'm not yet handling here is when the top-level annotation no longer exists but the reply still does. In that case the current behavior will be to display the "Message not available". What I intend to do in a separate PR is show "Message not available" but still show the replies that do exist. Likewise if a reply mid-way down the thread is missing, the missing replies will be replaced by "Message not available" but any replies to the missing reply should still be shown.

### DIFF
--- a/h/static/scripts/annotation-viewer-controller.js
+++ b/h/static/scripts/annotation-viewer-controller.js
@@ -2,6 +2,28 @@
 
 var angular = require('angular');
 
+/**
+ * Fetch all annotations in the same thread as `id`.
+ *
+ * @return Promise<Array<Annotation>>
+ */
+function fetchThread(store, id) {
+  var annot;
+  return store.AnnotationResource.get({id: id}).$promise.then(function (annot) {
+    if (annot.references && annot.references.length) {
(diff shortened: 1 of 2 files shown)


nickstenning:
LGTM.

robertknight:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;vOUw3BbE&#x2F;372-bug-message-not-available-when-viewing-standalone-annotation-page-for-reply"> [BUG] &quot;Message not available&quot; when viewing standalone annotation page for reply</a>

PR #3476, opened 2016-06-15
Prevent features cohort client from crashing when unauthenticated

finitud:


chdorner:
LGTM. Thanks for the fast fix!

PR #3478, opened 2016-06-15
Don't crash when creating DocumentMetas concurrently

### LINKED ISSUE #3465: Crash when two requests simultaneously create document metadata

seanh:
Fixes #3465.

Creating a new annotation can crash when two concurrent requests are
trying to create new annotations of the same document:

My theory is that this can happen due to a concurrency problem when two
people simultaneously try to annotate a page that has not been annotated
before:

1. Browse to a page that:

   a. Will generate more than one `DocumentMeta` (e.g. has a title and
   favicon)

   b. Has not been annotated yet (or more accurately does not currently
   have any `DocumentMeta`s in the db)

2. Put a `set_trace()` before
   [this line](https://github.com/hypothesis/h/blob/46948765cc7e49b18b7d8030ad29b9f137e86de0/h/api/models/document.py#L413)

3. Annotate the page

4. The first time it hits your `set_trace()` hit `c` for continue,
   allowing it to create the first of two `DocumentMeta`s. I suppose the
   ordering here is undefined but for me it was always the title
   `DocumentMeta` that got created first. This creates the
   `DocumentMeta` in the `sqlalchemy` session but doesn't flush or
   commit it yet. The code should now be stopped on your `set_trace()`
   for the second (and last) time.

5. Go over to a `psql` shell for your db and insert a `DocumentMeta`
   with the same `claimant_normalized` and `type` as the one that was
   created in the previous step: `INSERT INTO document_meta
   (document_id, claimant, claimant_normalized, type, value) VALUES
   (1128, 'http://localhost:4545/', 'http://localhost:4545', 'title',
   '{"foo"}');`

6. Go back and hit `c` to continue processing the request. The query for
   existing `DocumentMeta`s on
   [this line](https://github.com/hypothesis/h/blob/46948765cc7e49b18b7d8030ad29b9f137e86de0/h/api/models/document.py#L318)
   will cause the `DocumentMeta` that was created in step 4 to be
   flushed, which will violate a database constraint because of the
   `DocumentMeta` that you added in step 5, and you will see the
   traceback.

The problem is that `create_or_update_document_meta()` has the logic:
"If there's already a DocumentMeta then return it, else create a new one" and
this cannot work because a concurrent request may create a new
DocumentMeta in-between the "if" and the "create a new one" and this
will then cause the "create a new one" to crash with an IntegrityError
from the database.

It is not possible to simply catch this IntegrityError and return the
DocumentMeta that was inserted by the concurrent request instead,
because once the IntegrityError has happened it's necessary to rollback
the transaction and this will lose any other objects that've been added
to the session outside of `create_or_update_document_meta()` but not
committed yet (e.g. `DocumentURI`s, the new `Annotation`).

So instead we raise a TransientError subclass which causes pyramid_tm to
abort the transaction and try replaying the request.

See: http://docs.pylonsproject.org/projects/pyramid_tm/en/latest/#retrying

### DIFF
--- a/h/api/models/__init__.py
+++ b/h/api/models/__init__.py
@@ -27,4 +27,5 @@ __all__ = (
     'DocumentURI',
     'elastic',
     'merge_documents',
+    'update_document_metadata',
 )
--- a/h/api/models/__init__.py
+++ b/h/api/models/__init__.py
@@ -13,8 +13,6 @@ functions in `h.api.storage`.
 
 from h.api.models import elastic
 from h.api.models.annotation import Annotation
-from h.api.models.document import create_or_update_document_meta
-from h.api.models.document import create_or_update_document_uri
 from h.api.models.document import Document
 from h.api.models.document import DocumentMeta
 from h.api.models.document import DocumentURI
@@ -24,8 +22,6 @@ from h.api.models.document import update_document_metadata
 
 __all__ = (
     'Annotation',
-    'create_or_update_document_meta',
-    'create_or_update_document_uri',
     'Document',
     'DocumentMeta',
     'DocumentURI',
--- a/h/api/models/document.py
+++ b/h/api/models/document.py
@@ -8,6 +8,7 @@ import logging
 import sqlalchemy as sa
 from sqlalchemy.dialects import postgresql as pg
 from sqlalchemy.ext.hybrid import hybrid_property
+import transaction
 
 from h.api.db import Base
 from h.api.db import mixins
@@ -15,6 +16,17 @@ from h.api.uri import normalize as uri_normalize
 
 
 log = logging.getLogger(__name__)
+
+
+class ConcurrentDocumentMetaCreateError(transaction.interfaces.TransientError):
+    """
+    Raised when concurrent requests try to create the same DocumentMeta.
+
+    Because this is a TransientError subclass, raising it will cause pyramid_tm
+    to abort the current transaction and try replaying the whole request.
+    """
+
+    pass
 
 
 class Document(Base, mixins.Timestamps):
@@ -312,6 +324,10 @@ def create_or_update_document_meta(session,
         attribute to
     :type updated: datetime.datetime
 
+    :raises :py:exc:`sqlalchemy.exc.IntegrityError`: if a concurrent request
+        commits an equivalent DocumentMeta just before this request attempts
+        to do the same
+
     """
     existing_dm = session.query(DocumentMeta).filter(
         DocumentMeta.claimant_normalized == uri_normalize(claimant),
@@ -326,6 +342,11 @@ def create_or_update_document_meta(session,
                     created=created,
                     updated=updated,
                     ))
+        # If a concurrent request has just committed a DocumentMeta that
+        # conflicts with the one we just added, then flush() will raise
+        # IntegrityError. Best to flush now and have it raised here, rather
+        # than at some random point later in the code.
+        session.flush()
     else:
         existing_dm.value = value
         existing_dm.updated = updated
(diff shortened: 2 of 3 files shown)


nickstenning:
This looks fine, but it's not a problem restricted to `DocumentMeta`. This should probably wrap the whole of the inside of `update_document_metadata`.

h/api/models/__init__.py
```diff
@@ -13,6 +13,7 @@
 
 from h.api.models import elastic
 from h.api.models.annotation import Annotation
+from h.api.models.document import ConcurrentDocumentMetaCreateError
```
nickstenning: This doesn't need to be "public". 

For that matter, neither do the "create_or_update_document_*" functions...

seanh:
> This looks fine, but it's not a problem restricted to DocumentMeta. This should probably wrap the whole of the inside of update_document_metadata.

If I remove `create_or_update_document_meta()` and `create_or_update_document_uri()` (which I'm guessing also needs the same bug fix as this) from the public interface then yes. But those functions are complicated enough that they have their own unit tests, and we don't unit test private functions. Would I move all those unit tests to be testing the private functions indirectly through `update_document_metadata()`? Or move `create_or_update_document_meta()` and `create_or_update_document_uri()` into their own module?

I guess I'd vote for testing the private functions indirectly

h/api/models/__init__.py
seanh: Right, as long as `create_or_update_document_meta()` is public then this does need to be public. If `create_or_update_document_meta()` becomes private then yes

nickstenning:
@seanh: I was talking about exposing them in the `__init__` module. They can remain public in the `document` module, that's fine.

seanh:
Actually, `update_document_metadata()` is gonna raise `ConcurrentDocumentMetaCreateError` now, and `update_document_metadata()` is part of the `h.api.models` public API, so doesn't `ConcurrentDocumentMetaCreateError` need to be part of the public API as well?

nickstenning:
Given that nothing in our code is supposed to catch it, no.

seanh:
Ok, I've moved the `try ... except sa.exc.IntegrityError:` into `update_document_metadata()` but it doesn't catch errors from `create_or_update_document_uri()` yet, it probably should but I just want to quickly test what happens with that function and concurrent creations first..

seanh:
I think we could merge this as is? And find and fix other concurrency errors in this module separately

seanh:
I think there's a little bit more to concurrency issues and `DocumentURI` than it might first seem, so I'm proposing merging this fix for a `DocumentMeta` crash on its own even though it blatantly ignores a similar nearby `DocumentURI` crash. I'm opening GitHub issues for the `DocumentURI` stuff..

nickstenning:
Well, possibly, yes, but I don't understand what's complicated here? There is a broad class of race-condition introduced by the following kind of code:

```python
if something_exists_in_the_database():
    use_that_thing()
else:
    create_a_new_thing_in_the_database()
```

All that's required to fix the entire shebang is what I suggested earlier:

> This should probably wrap the whole of the inside of update_document_metadata.

To be even more explicit about it, rename `update_document_metadata` to `_update_document_metadata` and then rewrite `update_document_metadata` as follows:

```python
def update_document_metadata(session, ...):
    try:
        _update_document_metadata(session, ...)
        session.flush()
    except sa.exc.IntegrityError:
        raise ConcurrentDocumentUpdateError()
```

That's all there is to it. We can merge this if you want, but you're only going to have to rewrite the tests for it anyway...

nickstenning:
Closing in favour of #3493 as discussed on Slack.

PR #3484, opened 2016-06-15
Add some timers to the streamer code

nickstenning:
So we have some idea how long message processing is taking.


chdorner:
LGTM.

PR #3485, opened 2016-06-15
Repurposed About panel for Help

lenazun:
This PR consolidates different Help-related options in the signed-in menu to just one.

The Menu now only says Help

menu

And what gets displayed is the About this Version card, and I've repurposed the "send feedback" link to send an email to support (including all the details on the card).  

help

I have questions:
@conordelahunty : is this visually what you had in mind? Is the copy OK or should we write anything else? Is it OK to omit the feedback link at the bottom? (It was going to the same place as support)

@robertknight, @sheetaluk: I've repurposed feedback_link.html, some styles and names that refer to "About this version".  Should I rename everything to "support" or "help", or is it OK to leave it as it is?


lenazun:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;RwDV4WYx&#x2F;377-simplify-menu-for-support-in-our-app"> Simplify menu for support in our app</a>

nickstenning:
This looks great! Should we have a short sentence linking people to `/docs/help`, do you think?

And yes, we should probably rename everything like `about-this-version`, `about_this_version`, `aboutThisVersion` to `help`.

conordelahunty:
Looking good Lena, this is basically what I had in mind.

> Is it OK to omit the feedback link at the bottom?

Yep.

I wonder if we could add a little more copy. Something like "<Send us a message> if you have any questions or want to give us feedback. You can also visit our <help page>."

robertknight:
> I've repurposed feedback_link.html, some styles and names that refer to "About this version".
> Should I rename everything to "support" or "help", or is it OK to leave it as it is? 

Component names should generally be nouns of 2-3 words, usually of the form `$role-$type-of-element` (eg. `feedback-link`, `share-dialog`, `search-form`), preferably matching what the piece of the UI is called in Trello cards, design discussion etc. I'd suggest `help-panel` or `help-dialog`. At the moment it is a pane or panel rather than a dialog. I can imagine in future we might want to make it a modal on smaller devices though.

lenazun:
Thank you all for your comments! I renamed all the things, and made the suggested changes to the copy (including a link to the docs/help)

added_copy

dwhly:
Only question for you all is whether modifying the menu text from "Help" to "Help / Feedback" makes sense?

Otherwise looks great!

h/static/scripts/directive/signin-control.js
```diff
@@ -17,7 +17,7 @@ module.exports = function () {
       /**
        * Called when the user clicks on the "About this version" text.
```
robertknight: The reference to "About this version" needs updating

robertknight:
The code changes look code but I'm afraid it wasn't at all obvious to me when I tried this just now that "Send us a message" and "help documents" were clickable links because they have only a tiny contrast change from the surrounding text. This might be as simple as underlining the links. Thoughts @conordelahunty ?

robertknight:
This what it looks like with underlined links:

underlined-links

h/templates/client/help_panel.html
```diff
@@ -0,0 +1,43 @@
...
+    <help-link
+      version="vm.version"
+      user-agent="vm.userAgent"
+      url="vm.url"
+      document-fingerprint="vm.documentFingerprint"
+      auth="vm.auth"
+      date-time="vm.dateTime">
+    </help-link>  
+    if you have any questions or want to give us feedback. 
+    You can also visit our <a class="help-panel-content__link" href="/docs/help" target="_blank"> help documents.</a>
```
robertknight: The `/docs/help` link will not work in the Chrome extension where the page is served from `chrome-extension://<ID>/viewer`. You need to prepend `{{vm.serviceUrl}}` where `vm.serviceUrl` is the URL of the Hypothesis service that the app is talking to.

conordelahunty:
@robertknight Yes to underlines

robertknight:
OK, I've made the change to underline links and also changed `/docs/help` to use `serviceUrl`. I also pruned a little bit of extraneous spacing around the 'Send us a message' link. Hope that's OK @lenazun

lenazun:
Thank you for the changes @robertknight !

PR #3486, opened 2016-06-15
Record approximate time-in-queue info for "realtime"

nickstenning:
While using wall clocks to calculate time differences is always going to
be something of a crapshoot, it would be good to know roughly how long
messages are spending in RabbitMQ.

This commit adds a "timestamp" header to every message published by
h.realtime.Publisher, which h.realtime.Consumer can then read and use to
compute the approximate time-in-queue.

### DIFF
--- a/h/realtime.py
+++ b/h/realtime.py
@@ -3,6 +3,7 @@
 import base64
 import random
 import struct
+from datetime import datetime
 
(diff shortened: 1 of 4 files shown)


chdorner:
LGTM.

PR #3487, opened 2016-06-16
Refactor search input

robertknight:
This is a refactoring of the way the search in the top bar works to make it more consistent with other recently updated pieces of the UI, specifically:

 * First commit: Convert `<simple-search>` to a modern component style which takes the current search query as input and calls a callback when it changes, rather than modifying a two-way binding directly and then `$scope.$watch()`'ing that elsewhere. This allows us to store the current filter query in just one place in the app (`annotationUI.getState().filterQuery`)
 * Second commit: Rename `<simple-search>` to `<search-input>` and convert from CoffeeScript to JS, with a new set of tests that use the `createDirective()` helper to simplify things and test the behavior more "from the outside" with less reliance on implementation details.


robertknight:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;emzDGxsD&#x2F;375-auto-expand-conversation-thread-for-direct-linked-annotations"> Auto-expand conversation thread for direct-linked annotations</a>

nickstenning:
Branch needs a rebase.

PR #3488, opened 2016-06-16
Add Makefile documentation for dist target and hint for builds

finitud:
Decided to add some hints to the `Makefile` because I keep forgetting to `make dist` before `make extensions`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -30,6 +30,7 @@ clean:
 dev: build/manifest.json h.egg-info/.uptodate
 	@hypothesis devserver
 
+## Build the python H distribution
 .PHONY: dist
 dist: dist/h-$(BUILD_ID).tar.gz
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -114,8 +115,13 @@ node_modules/.uptodate: package.json
...
+	@echo
 	@echo " clean      Clean up runtime artifacts (needed after a version update)"
 	@echo " dev        Run the development H server locally"
+	@echo " dist       Build the python H distribution"
 	@echo " docker     Build hypothesis/hypothesis docker image"
 	@echo " extensions Build the browser extensions"
 	@echo " test       Run the test suite (default)"
+	@echo
+	@echo "To run the test suite and build the whole app, you can do:"
+	@echo " $ make clean && make && make dist && make extensions"
```
nickstenning: AKA `make clean test dist extensions`

robertknight:
The `make extensions` target is only really used by Travis CI. In development the normal workflow is to run `make dev` which will auto-rebuild client assets when they change and then run `hypothesis-buildext <browser> --service <service>` which is then instant.

This is completely non-obvious however. Perhaps it would make sense to add a `dev-extension` target in the Makefile which documents this workflow and some comments about the use of `make extensions` in the Makefile.

finitud:
This is a great idea, I will add some comments on this in the `Makefile`.

Makefile
nickstenning: Did you see this comment? I'm happy to merge this but was waiting for you to update this.

finitud: Yes – I'm going to rewrite this a bit though, based on @robertknight's comments as well. Will mark the PR as WIP for now.

nickstenning:
Can we get this updated, please? Stale PRs are :-1: and we should either be updating/merging or closing.

finitud:
Closing for now.

PR #3490, opened 2016-06-16
Rewrite factories.Annotation to use Postgres

seanh:
Rewrite factories.Annotation to generate the new h.api.models.Annotation
objects that use the new PostgreSQL annotation storage, instead of
generating legacy Elasticsearch storage objects.

In the process also add Document, DocumentURI and DocumentMeta factories that the
Annotation factory uses, these can also be used independently.

Demo:

```python
>>> a = factories.Annotation()
>>> a.tags  # Each annotation gets a random number of randomly generated tags.
[]
>>> factories.Annotation().tags
[u'quo', u'eos']
>>> factories.Annotation().tags
[u'enim', u'suscipit', u'mollitia', u'ab']
>>> a.target_uri  # Randomly generated fake URI.
u'http://www.weimann.com/faq.jsp'
>>> a.text  # Randomly generated text.
u'Commodi magnam pariatur nulla itaque repellat reprehenderit odit. Ipsum molestiae adipisci beatae totam accusantium. Incidunt beatae iste ab blanditiis pariatur voluptatibus beatae. Harum eligendi nulla nisi maxime maxime consequatur quod.'
>>> a.userid  # Randomly generated fake user ID.
u'wpfeffer'
>>> a.target_selectors  # This is hard-coded.
[{u'type': u'RangeSelector', u'startContainer': u'/div[1]/article[1]/section[1]/div[1]/div[2]/div[1]', u'endContainer': u'/div[1]/article[1]/section[1]/div[1]/div[2]/div[1]', u'startOffset': 0, u'endOffset': 76}, {u'start': 286, u'end': 362, u'type': u'TextPositionSelector'}, {u'type': u'TextQuoteSelector', u'prefix': u' browser extension.\n            ', u'exact': u'If you wish to install Hypothesis on your own site then head over to GitHub.', u'suffix': u'\n          \n        \n      \n    '}]
>>> a.document  # Either a new Document will be created and added to the db,
                # or an existing one will be used
                # (it actually just calls update_document_metadata()).
<Document 42>
>>> a.document.title  # This comes from the generated DocumentMetas.
u'revolutionize dynamic e-tailers'
>>> a.document.document_uris  # A random number of randomly generated
                              # DocumentURIs are added to the db.
[<DocumentURI 52>, <DocumentURI 53>, <DocumentURI 54>]
>>> # They all match the target_uri of the annotation:
>>> [(u.claimant, u.claimant_normalized, u.uri, u.uri_normalized) for u in a.document.document_uris]
[(u'http://www.weimann.com/faq.jsp', u'http://www.weimann.com/faq.jsp',
  u'http://www.weimann.com/faq.jsp', u'http://www.weimann.com/faq.jsp'),
 (u'http://www.weimann.com/faq.jsp', u'http://www.weimann.com/faq.jsp',
  u'http://www.weimann.com/faq.jsp', u'http://www.weimann.com/faq.jsp'),
 (u'http://www.weimann.com/faq.jsp', u'http://www.weimann.com/faq.jsp',
  u'http://www.weimann.com/faq.jsp', u'http://www.weimann.com/faq.jsp')]
>>> # Types and content types come from looping sequences of hardcoded values.
>>> # FIXME: I'm not sure why there's a None in here.
>>> [(u.type, u.content_type) for u in a.document.document_uris]
[(u'self-claim', None), (u'rel-alternate', u'text/html'),
 (u'rel-canonical', u'application/pdf')]
>>> a.document.meta # A random number of randomly generated DocumentMetas are
                    # added to the db.
[<DocumentMeta 124>]
>>> a.document.meta[0].claimant  # It matches the annotation's target_uri.
u'http://www.weimann.com/faq.jsp'
>>> # Type and value loop through lists of hardcoded values.
>>> a.document.meta[0].type
u'title'
>>> a.document.meta[0].value
[u'revolutionize dynamic e-tailers']
```

Additionally:

* `factories.Annotation.build()` will make an `Annotation()` only (no `Document`s, `DocumentURI`s and `DocumentMeta`s) and will not add it to the db session

* You can also call `factories.Document()`, `factories.DocumentURI()` and `factories.DocumentMeta()` directly. If all you do is call `factories.DocumentURI()` and `factories.DocumentMeta()` and you didn't add any `DocumentURI`s or `DocumentMeta`s into the db in any other way, it should be safe. If there's already stuff in the db it may throw `IntegrityError`. By default it just creates a new `Document` for each `DocumentURI` and `DocumentMeta` and doesn't do any merging of documents.

* You can always call `factories.DocumentURI.build()` and `factories.DocumentMeta.build()` which is safe because it doesn't add them to the db session, then you can create dicts from these unattached `DocumentURI`s and `DocumentMeta`s and pass them into `h.api.models.update_document_metadata()` which will do the right thing (this is actually what `factories.Annotation()` does).

### DIFF
--- a/tests/factories.py
+++ b/tests/factories.py
@@ -2,135 +2,167 @@
 """Factory classes for easily generating test objects."""
 
 from __future__ import absolute_import
+from __future__ import unicode_literals
 
 import random
-import datetime
 
 import factory
+import faker
 
-from h.api import storage
+from h import db
+from h import models
 from h.accounts import models as accounts_models
+from h.api import models as api_models
 
 
-class Annotation(factory.Factory):
-
-    """A factory class that generates test annotation dicts.
-
-    Usage:
-
-        annotation_1 = Annotation()
-        annotation_2 = Annotation()
-
-    A unique "id" and "text", "created" and "updated" strings based on the
-    current time, etc. are generated for each annotation dict returned.
-
-    You can override the value of any attribute by passing in named arguments,
-    for example:
-
-        annotation = Annotation(text="My annotation text")
-
-    There are some special params that don't appear in the annotation dicts
-    themselves but influence the values that do appear in the dicts.
-
-    For example:
-
-        annotation = Annotation(username="my_user")
-
-    will create an annotation with "user": "acct:my_user@127.0.0.1".
-
-        annotation = Annotation(num_tags=3)
-
-    will create an annotation with 3 (randomly generated) tags.
-    Without this argument the number of tags is chosen at random (and may be
-    zero).
-
-    To choose the actual values of the tags do:
-
-        annotation = Annotation(tags=["foo", "bar"])
-
-    A random number is generated for each annotation and used in the URI
-    ("http://example.com/document_3"), document title ("Example Document 3"),
-    etc. To choose this number do:
-
-        annotation = Annotation(random_number=7)
+FAKER = faker.Factory.create()
 
 
-    """
+class Document(factory.alchemy.SQLAlchemyModelFactory):
 
-    class Meta:
-        model = storage.annotation_from_dict({}).__class__
-        exclude = ["username", "random_number", "num_tags", "exact_text",
-                   "document_title"]
+    class Meta:  # pylint: disable=no-init, old-style-class
+        model = models.Document
+        sqlalchemy_session = db.Session
 
-    username = "seanh"
-    random_number = factory.LazyAttribute(lambda n: random.randint(1, 10))
-    num_tags = None
-    exact_text = None
-    document_title = None
 
-    id = factory.Sequence(lambda n: "test_id_{n}".format(n=n + 1))
-    text = factory.Sequence(lambda n: "Test annotation {n}".format(n=n + 1))
-    user = factory.LazyAttribute(
-        lambda n: u"acct:{username}@127.0.0.1".format(username=n.username))
+class DocumentMeta(factory.alchemy.SQLAlchemyModelFactory):
 
-    @factory.LazyAttribute
-    def created(stub):
-        # pylint: disable=no-self-use
-        return datetime.datetime.now().isoformat()
+    class Meta:  # pylint: disable=no-init, old-style-class
+        model = models.DocumentMeta
+        sqlalchemy_session = db.Session
 
-    updated = factory.SelfAttribute('created')
+    # Trying to add two DocumentMetas with the same claimant and type to the
+    # db will crash. We use a sequence instead of something like FAKER.url()
+    # for claimant here so that never happens (unless you pass in your own
+    # claimant).
+    claimant = factory.Sequence(
+        lambda n: 'http://example.com/document_' + str(n) + '/')
 
-    @factory.LazyAttribute
-    def tags(stub):
-        num = (stub.num_tags if stub.num_tags is not None else
-               random.randint(1, 11))
-        return ["tag_{n}".format(n=n) for n in range(1, num + 1)]
+    type = factory.Iterator([
+        'title', 'twitter.url.main_url', 'twitter.title', 'favicon'])
+    document = factory.SubFactory(Document)
 
-    @factory.LazyAttribute
-    def uri(stub):
-        return "http://example.com/document_{n}".format(n=stub.random_number)
+    @factory.lazy_attribute
+    def value(self):
+        if self.type == 'twitter.url.main_url':
+            return [FAKER.url()]
+        elif self.type == 'favicon':
+            return [FAKER.image_url()]
+        else:
+            return [FAKER.bs()]
 
-    @factory.LazyAttribute
-    def target_selectors(stub):
-        return [{"type": "RangeSelector",
-                 "startContainer": "/div[1]/article[1]/",
-                 "endContainer": "/div[1]/article[1]/section[1]",
-                 "startOffset": 0,
-                 "endOffset": 43},
-                {"start": 211,
-                 "end": 254,
-                 "type": "TextPositionSelector"},
-                {"type": "TextQuoteSelector",
-                 "prefix": "text quote prefix",
-                 "exact": (stub.exact_text if stub.exact_text is not None
-                           else "The exact text that was selected"),
-                 "suffix": "text quote suffix"},
-                {"type": "FragmentSelector",
-                 "value": ""}
-            ]
 
-    @factory.LazyAttribute
-    def permissions(stub):
-        return {
-            "admin": [stub.user],
-            "read": ["group:__world__"],
-            "update": [stub.user],
-            "delete": [stub.user]
-        }
+class DocumentURI(factory.alchemy.SQLAlchemyModelFactory):
 
-    @factory.LazyAttribute
-    def document(stub):
-        return {
-            "eprints": {},
-            "title": (stub.document_title or
-                      "Example Document {n}".format(n=stub.random_number)),
-            "twitter": {},
-            "dc": {},
-            "prism": {},
-            "highwire": {},
-            "facebook": {},
-            "link": [{"href": stub.uri}]
-        }
+    class Meta:  # pylint: disable=no-init, old-style-class
+        model = models.DocumentURI
+        sqlalchemy_session = db.Session
+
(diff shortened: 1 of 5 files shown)


tests/factories.py
```diff
@@ -2,135 +2,159 @@
...
+            This doesn't add anything to the database session yet.
+            """
+            document_uri = DocumentURI.build(claimant=self.target_uri,
+                                             uri=self.target_uri)
+            return dict(
+                claimant=document_uri.claimant,
+                uri=document_uri.uri,
+                type=document_uri.type,
+                content_type=document_uri.content_type,
+            )
```
seanh: Here I want to be able to use all of factory_boy's goodness to randomly generate the values for the dict, and I think the only way to do that is to have a factory class to call. I could have written a `DocumentURIDict` factory class, but a `DocumentURI` factory seems more useful. So here I'm actually making an `h.api.models.DocumentURI` object, but not adding it to the db session, then returning a dict of some of its attributes.

It _might_ be possible to do something slightly nicer here, like have a `DocumentURI` factory class that _inherits_ from a `DocumentURIDict` factory without duplication.

tests/factories.py
```diff
@@ -2,135 +2,159 @@
...
+                value=document_meta.value,
+            )
+
+        document_meta_dicts = [document_meta_dict()
+                               for _ in range(random.randint(1, 3))]
+
+        # Make sure that there's always at least one DocumentMeta with
+        # type='title', so that we never get annotation.document.title is None:
+        if 'title' not in [m['type'] for m in document_meta_dicts]:
+            document_meta_dicts.append(document_meta_dict(type='title'))
```
seanh: There are some existing tests that expect `annotation.document.title` to be a string, and without this they would intermittently fail.

tests/h/factories_test.py
```diff
@@ -15,78 +15,30 @@ def test_id(self):
...
-        """Annotations should have an updated date from the current time."""
-        before = datetime.datetime.now()
-
-        annotation = factories.Annotation()
-
-        after = datetime.datetime.now()
-        updated = datetime.datetime.strptime(
-            annotation["updated"], "%Y-%m-%dT%H:%M:%S.%f")
-
-        assert before < updated < after
```
seanh: There's no longer anything about created and updated times in any of the factory classes. These columns have defaults in the db, so it's a feature of the ORM classes they you don't have to pass them in and values will juist be generated, the factory classes just use that.

tests/h/factories_test.py
```diff
@@ -15,78 +15,30 @@ def test_id(self):
...
-
-    def test_tags(self):
-        """It should be possible to choose the number of tags with num_tags."""
-        # If num_tags isn't passed the factory chooses a random number of tags.
-        # Here we choose a num_tags higher than the upper range of this random
-        # choice, so there's no chance of random false positive test passes.
-        annotation = factories.Annotation(num_tags=20)
-        assert len(annotation["tags"]) == 20
-        assert "num_tags" not in annotation
```
seanh: This ability to specify the number of tags was removed. You can do `factories.Annotation(tags=['foo', 'bar'])` though.

tests/h/factories_test.py
```diff
@@ -15,78 +15,30 @@ def test_id(self):
...
 
     def test_custom_tags(self):
-        assert factories.Annotation(tags=["foo", "bar", "gar"])["tags"] == [
+        assert factories.Annotation(tags=["foo", "bar", "gar"]).tags == [
             "foo", "bar", "gar"]
 
-    def test_uri(self):
-        annotation = factories.Annotation(random_number=3)
-        assert annotation["uri"] == "http://example.com/document_3"
-        assert "random_number" not in annotation
```
seanh: This random number thing was removed.

tests/h/factories_test.py
```diff
@@ -15,78 +15,30 @@ def test_id(self):
...
-
-    def test_permissions(self):
-        annotation = factories.Annotation(username="test_user")
-        assert "test_user" in annotation["permissions"]["admin"][0]
-        assert "test_user" in annotation["permissions"]["update"][0]
-        assert "test_user" in annotation["permissions"]["delete"][0]
```
seanh: There's no permissions anymore, and `shared` defaulting to `False` is a feature of the ORM class not the factory.

tests/factories.py
```diff
@@ -2,135 +2,167 @@
 """Factory classes for easily generating test objects."""
 
 from __future__ import absolute_import
+from __future__ import unicode_literals
 
 import random
-import datetime
 
 import factory
+import faker
```
nickstenning: Presumably you need to add this to the `tox.ini` deps now?

tests/factories.py
```diff
@@ -2,135 +2,167 @@
...
-            "title": (stub.document_title or
-                      "Example Document {n}".format(n=stub.random_number)),
-            "twitter": {},
-            "dc": {},
-            "prism": {},
-            "highwire": {},
-            "facebook": {},
-            "link": [{"href": stub.uri}]
-        }
+from h.api import models as api_models
```
nickstenning: Importing `models` for `update_document_metadata` is confusing to me. I think `from h.api.models.document import update_document_metadata` would be clearer.

tests/factories.py
```diff
@@ -2,135 +2,167 @@
...
 import random
-import datetime
 
 import factory
+import faker
 
-from h.api import storage
+from h import db
+from h import models
 from h.accounts import models as accounts_models
```
nickstenning: Unused import?

tests/factories.py
seanh: I don't think so, it's already a dependency of factory_boy

nickstenning: Ah, I see, it's a dependency of factory-boy's.

tests/h/factories_test.py
```diff
@@ -3,90 +3,111 @@
...
-    """Unit tests for the Annotation factory class."""
+    def test_with_default_values(self):
+        document = factories.Document()
 
-    def test_id(self):
-        """Each test annotation should be created with a unique ID."""
-        annotation_1 = factories.Annotation()
-        annotation_2 = factories.Annotation()
+        db.Session.flush()
+        assert document.id
```
nickstenning: Isn't this just testing the document model?

nickstenning:
This is looking great!

I have only a couple of small niggles, and one broader one, namely: is there actually value in testing test factories? Aren't they implicitly tested by the tests that use them, and if they break then the tests that use them will break?

I'd be in favour of just removing `factories_test.py` altogether, if for no other reason than it will give us a more realistic picture from code coverage of which bits of the factory code we're actually using in tests.

tests/factories.py
nickstenning: ECONCURRENTUPDATE

tests/factories.py
seanh: Done

tests/factories.py
seanh: `accounts_models`? It's used by the `User` class at the bottom of the file, which this pr hasn't touched

nickstenning: Oops. Cmd-F fail.

tests/h/factories_test.py
seanh: This is just a super-basic test that the `Document` factory works

seanh:
I considered just deleting `factories_test.py` but in the end I replaced what was there with a much simpler set of tests that are really just intended to break if the factories are completely broken, hopefully helping to make it clear that it's the factories that are broken (and not all the other tests that would be failing).

There's not much too the various factories that isn't just declarative and getting factory_boy to do things for us, so not much point in us testing the behaviour in detail.

Some parts such as `Annotation.make_metadata()` probably cross this line but would probably require some pretty awkward mocking to test because they involve so much randomness.

tests/factories.py
seanh: I would quite like to send a follow-up pr to sort out the `User` factory though. It's fine, but it could use faker to get nicer fake usernames and passwords etc

seanh: Oh, and it _should_ add the user to the db session unless you do `.build()`, it doesn't

nickstenning:
Ok. I'm still not sure that `factories_test.py` is a good idea, but we'll keep it for now.

PR #3492, opened 2016-06-16
Add `pyramid_request` fixture

nickstenning:
Our test code contains a huge variety of different approaches to generating the ubiquitous Pyramid request object. This heterogeneity makes tests harder to read, and easier to break (as when, for example, a view function acquires a dependency on the database session, the stub request had to change).

This commit introduces a `pyramid_request` fixture, which should be in 99% of cases the way tests generate a stub request. This request fixture is preconfigured with some commonly-needed properties -- in particular the database session.

In addition, this PR rearranges some of the other request fixtures in order to improve test isolation -- the settings object is no longer shared between tests, for example.


chdorner:
LGTM.

PR #3493, opened 2016-06-17
Fix race condition bugs in document update code

### LINKED ISSUE #3481: Concurrency issues with creating DocumentURIs
To really pin this down I think that <https://github.com/hypothesis/h/issues/3480> should be fixed first so that concurrency errors become `IntegrityError`s and crash, rather than simply creating duplicate rows in the `document_uri` table.

`h.api.models.document` has two places that contain the logic "if there is not already a matching DocumentURI then create one" (just search the file for lines that init a `DocumentURI()`). If a concurrent request comes along and commits a matching DocumentURI after the _if_ but before the _create_ then (once <https://github.com/hypothesis/h/issues/3480> is fixed) I think it'll crash with an `IntegrityError`).

nickstenning:
In a number of places, the code for updating document metadata performs a series of actions that map to the following pattern:

    1   in_database = check_for_existing_data(*keys)
    2   if in_database:
    3       update_existing_data(*keys, data)
    4   else:
    5       create_data(*keys, data)

This type of code suffers from a race condition problem. Imagine that two requests carrying the same data arrive within moments of one another. For example, if two people annotate the same page at almost exactly the same time.

The most troublesome scenario is the one in which the conditional on line 1 evaluates to false.

In this situation, two application servers (which may be on physically separated machines) execute the same code, and race each other to commit their database transactions. If the first has not committed its transaction by the time line 1 is executed, then the second application server will also follow the code path in which it tries to create new records. When it comes to commit *its* transaction (after the other application server) the database integrity constraints will throw an error.

For updating document metadata, this is not a fatal condition. We can simply throw an error which is an instance of `transaction.interfaces.TransientError`, which will result in the entire request being replayed[1]. Hopefully, the second time around the code will follow the first branch of the conditional and all will be well.

[1]: http://docs.pylonsproject.org/projects/pyramid_tm/en/latest/#retrying

Fixes #3481.

### DIFF
--- a/h/api/models/document.py
+++ b/h/api/models/document.py
@@ -6,6 +6,7 @@ from datetime import datetime
 import logging
 
 import sqlalchemy as sa
+import transaction
 from sqlalchemy.dialects import postgresql as pg
 from sqlalchemy.ext.hybrid import hybrid_property
 
@@ -15,6 +16,10 @@ from h.api.uri import normalize as uri_normalize
 
 
 log = logging.getLogger(__name__)
+
+
+class ConcurrentUpdateError(transaction.interfaces.TransientError):
+    """Raised when concurrent updates to document data conflict."""
 
 
 class Document(Base, mixins.Timestamps):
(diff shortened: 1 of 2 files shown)


seanh:
LGTM

PR #3494, opened 2016-06-17
Fix annotations made on PDFs in non-Safari browsers not appearing in Safari

### LINKED ISSUE #3471: Annotations don't load on PDFs thru via on Safari
### Steps to reproduce
1. Open Safari Browser
2. Browse to https://via.hypothes.is/http://jeremydean.org/blog/wp-content/uploads/2016/06/Long_Live_the_Web.pdf

### Expected behaviour
Annotations should load

### Actual behaviour
They don't

Recent Safari

### Additional details
No console output other than PDF fingerprint.

robertknight:
In Safari a trailing '#' is appended to URIs extracted from web
documents and PDFs. For URLs this makes no difference as the Hypothesis
search API ignores the fragment identifier. However a URN with a
trailing '#' is considered different from a URN without a trailing '#'.
Since the URI for PDFs is now a URN of the form 'urn:x-pdf:...', when
viewing a PDF in Safari annotations made in other browsers were not
fetched and vice-versa.

The trailing '#' is caused by a bug in the `URL.hash` setter in Safari.

According to section 6.3 of the URL spec [1], setting URL.hash to an empty
string should _remove_ the fragment identifier. In Safari however, it
sets the fragment identifier to an empty string. The result is that when
the URL is serialized, the output is eg. 'urn:x-pdf:aabb#' instead of
'urn:x-pdf:aabb'.

Upstream bug: https://bugs.webkit.org/show_bug.cgi?id=158869
Fixes #3471

[1] https://url.spec.whatwg.org/#urlutils-members

### DIFF
--- a/h/static/scripts/annotator/guest.coffee
+++ b/h/static/scripts/annotator/guest.coffee
@@ -19,6 +19,17 @@ animationPromise = (fn) ->
       catch error
         reject(error)
 
+# Normalize the URI for an annotation. This makes it absolute and strips
+# the fragment identifier.
+normalizeURI = (uri, baseURI) ->
+  # Convert to absolute URL
+  url = new URL(uri, baseURI)
+  # Remove the fragment identifier.
+  # This is done on the serialized URL rather than modifying `url.hash` due
+  # to a bug in Safari.
+  # See https://github.com/hypothesis/h/issues/3471#issuecomment-226713750
+  return url.toString().replace(/#.*/, '');
+
 module.exports = class Guest extends Annotator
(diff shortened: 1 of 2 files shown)


robertknight:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;KEU6XOHD&#x2F;380-pdf-annotations-created-in-non-safari-browsers-not-appearing-in-safari"> PDF annotations created in non-Safari browsers not appearing in Safari</a>

nickstenning:
LGTM.

PR #3497, opened 2016-06-17
Tailor API response to client

chdorner:
In order to remove the last bit of legacy code we want to load the annotations from the database after going to Elasticsearch. #3461 introduced this change, but it turned out that for annotations on certain URIs with a lot of document data this was just too slow, it got reverted in 9af5b659106885fe2be533de347afa8d5bc5711f.

We then decided to look at the client code and only render the document data the client actually needs instead of returning everything we have stored. It turned out the client only relies on the `title` property and the `document.link` for getting a non-URN URI when needed, and both of these are only used in the stream.

~~This PR introduces a new property `target.locator` which is either the annotation URI, or when that one is a URN (e.g. PDF fingerprint `urn:x-pdf:...`), then the server tries to load a non URN URI from the document data and returns that one. Which means that whenever we want to link to the source of a document, `target.locator` will always include a linkable URI, unless we don't have one, in which case `target.locator` is missing.~~
In the case of PDFs, the `target.source` (or `uri`) will now always be the best URI we know, if it is a URN then we don't have any other information and shouldn't render a link. These changes were done in #3524.

### DIFF
--- a/h/api/models/document.py
+++ b/h/api/models/document.py
@@ -38,16 +38,19 @@ class Document(Base, mixins.Timestamps):
                                backref='document',
                                order_by='DocumentMeta.updated.desc()')
 
+    meta_titles = sa.orm.relationship('DocumentMeta',
+                                      primaryjoin='and_(Document.id==DocumentMeta.document_id, DocumentMeta.type==u"title")',
+                                      order_by='DocumentMeta.updated.asc()',
+                                      viewonly=True)
+
     def __repr__(self):
         return '<Document %s>' % self.id
 
     @property
     def title(self):
-        titles = [m.value for m in self.meta if m.type == 'title']
-        try:
-            return titles[0][0]
-        except IndexError:
-            return None
+        for meta in self.meta_titles:
+            if meta.value:
+                return meta.value[0]
 
     @classmethod
     def find_by_uris(cls, session, uris):
--- a/h/api/storage.py
+++ b/h/api/storage.py
@@ -8,7 +8,6 @@ assumed to be validated.
 """
 
 from pyramid import i18n
-from sqlalchemy.orm import subqueryload
 
 from h.api import schemas
 from h.api import models
@@ -50,19 +49,37 @@ def fetch_annotation(session, id_):
         return None
 
 
-def fetch_ordered_annotations(session, ids, load_documents=False):
+def fetch_ordered_annotations(session, ids, query_processor=None):
+    """
+    Fetch all annotations with the given ids and order them based on the list
+    of ids.
+
+    The optional `query_processor` parameter allows for passing in a function
+    that can change the query before it is run, especially useful for
+    eager-loading certain data. The function will get the query as an argument
+    and has to return a query object again.
+
+    :param session: the database session
+    :type session: sqlalchemy.orm.session.Session
+
+    :param ids: the list of annotation ids
+    :type ids: list
+
+    :param query_processor: an optional function that takes the query and
+                            returns an updated query
+    :type query_processor: callable
+
+    :returns: the annotation, if found, or None.
+    :rtype: h.api.models.Annotation, NoneType
+    """
     if not ids:
         return []
 
     ordering = {x: i for i, x in enumerate(ids)}
 
     query = session.query(models.Annotation).filter(models.Annotation.id.in_(ids))
-
-    if load_documents:
-        query = query.options(
-            subqueryload(models.Annotation.document).subqueryload(models.Document.document_uris),
-            subqueryload(models.Annotation.document).subqueryload(models.Document.meta)
-        )
+    if query_processor:
+        query = query_processor(query)
 
     anns = sorted(query, key=lambda a: ordering.get(a.id))
     return anns
--- a/h/static/scripts/annotation-metadata.js
+++ b/h/static/scripts/annotation-metadata.js
@@ -13,47 +13,23 @@
  *
  */
 function extractDocumentMetadata(annotation) {
-  var document_;
   var uri = annotation.uri;
   var domain = new URL(uri).hostname;
-  if (annotation.document) {
-    if (uri.indexOf('urn') === 0) {
-      var i;
-      for (i = 0; i < annotation.document.link.length; i++) {
-        var link = annotation.document.link[i];
-        if (link.href.indexOf('urn:') === 0) {
-          continue;
-        }
-        uri = link.href;
-        break;
-      }
-    }
+  var title = domain;
 
-    var documentTitle;
-    if (Array.isArray(annotation.document.title)) {
-      documentTitle = annotation.document.title[0];
-    } else {
-      documentTitle = annotation.document.title;
-    }
-
-    document_ = {
-      uri: uri,
-      domain: domain,
-      title: documentTitle || domain
-    };
-  } else {
-    document_ = {
-      uri: uri,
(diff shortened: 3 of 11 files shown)


chdorner:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;i4d5DZGx&#x2F;371-figure-out-which-document-data-the-client-needs-and-only-render-that"> Figure out which document data the client needs and only render that</a>

h/static/scripts/test/annotation-metadata-test.js
```diff
@@ -80,8 +67,7 @@ describe('annotation-metadata', function () {
       context('when there is no document.title', function() {
         it('returns the domain as the title', function() {
           var model = {
-            document: {},
-            uri: 'http://example.com/',
+            uri: 'http://example.com/'
```
robertknight: Small nit here and in a few other places, the last entry in a multi-line array or object literal should have a  trailing comma:

```js
var foo = [
  entry1,
  entry2,
]
```

This reduces noise in diffs by not changing the content of a line depending on whether it is the last entry or not.

h/static/scripts/annotation-metadata.js
```diff
@@ -15,25 +15,21 @@
...
-      for (i = 0; i < annotation.document.link.length; i++) {
-        var link = annotation.document.link[i];
-        if (link.href.indexOf('urn:') === 0) {
-          continue;
-        }
-        uri = link.href;
-        break;
-      }
+
+  if (Array.isArray(annotation.target) && annotation.target.length > 0) {
```
robertknight: Do we need the `isArray` check here or can we rely on `target` being an array?

h/static/scripts/annotation-metadata.js
```diff
@@ -15,25 +15,21 @@
...
+      uri = locator;
     }
+  }
 
+  var domain = new URL(uri).hostname;
+
+  if (annotation.document) {
     var documentTitle;
-    if (Array.isArray(annotation.document.title)) {
+    if (Array.isArray(annotation.document.title) &&
```
robertknight: Is the type of `annotation.document.title` still `string[] | string` or is it now just `string`?

h/static/scripts/annotation-metadata.js
chdorner: No, you're right. Will fix.

h/static/scripts/annotation-metadata.js
chdorner: It is always an array now, like the other metadata (well, the one that I'm removing in this PR..). I'll change the check to be `annotation.document.title && ...` (since an empty array is falsy in JS, right?)

chdorner:
This should be ready for review now. I've wrote a message to the dev mailing list with a bit more backstory on this change: https://groups.google.com/a/list.hypothes.is/forum/#!topic/dev/0f5n1-1J5yY

h/api/models/document.py
```diff
@@ -38,16 +38,19 @@ class Document(Base, mixins.Timestamps):
                                backref='document',
                                order_by='DocumentMeta.updated.desc()')
 
+    meta_title = sa.orm.relationship('DocumentMeta',
+                                     primaryjoin='and_(Document.id==DocumentMeta.document_id, DocumentMeta.type==u"title")',
+                                     order_by='DocumentMeta.updated.asc()',
+                                     uselist=False,
+                                     viewonly=True)
+
```
chdorner: This produces warnings when there is more than one title document meta record (`SAWarning: Multiple rows returned with uselist=False for eagerly-loaded attribute 'Document.meta_title'`).
It does produce what we want, but I have the feeling that I'm using SQLAlchemy the wrong way to achieve what I want to achieve here. Any ideas?

chdorner:
This is ready for review again.

h/api/models/document.py
nickstenning: Given the join condition, you're always going to have the possibility of there being multiple rows in this queryset. I'd suggest just accepting this and changing it to `meta_titles` with `uselist=True`.

h/api/storage.py
```diff
@@ -50,19 +49,37 @@ def fetch_annotation(session, id_):
...
+    The optional `query_processor` parameter allows for passing in a function
+    that can change the query before it is run, especially useful for
+    eager-loading certain data. The function will get the query as an argument
+    and has to return a query object again.
+
+    :param session: the database session
+    :type session: sqlalchemy.orm.session.Session
+
+    :param ids: the list of annotation ids
+    :type id_: list
```
nickstenning: s/id_/ids/

tests/h/emails/reply_notification_test.py
```diff
@@ -43,13 +43,9 @@ def test_calls_renderers_with_appropriate_context(self,
     def test_falls_back_to_target_uri_for_document_title(self,
                                                          notification,
                                                          pyramid_request,
-                                                         storage_driver,
                                                          html_renderer,
                                                          text_renderer):
-        if storage_driver == 'elastic':
```
nickstenning: `from h.api.models import elastic` is now unused in this file, I think?

nickstenning:
This looks great to me. Yay deleting code. I think @robertknight should be aware of this, but other than the few tiny things I've commented on this can merge.

h/api/storage.py
```diff
@@ -50,19 +49,37 @@ def fetch_annotation(session, id_):
...
+
+    :param query_processor: an optional function that takes the query and
+                            returns an updated query
+    :type query_processor: function
```
nickstenning: You're using query_processor as a callable here (i.e. not specifically as a function), so I'd be inclined to document it as a callable, not a function.

h/static/scripts/annotation-metadata.js
```diff
@@ -13,47 +13,23 @@
...
-    document_.title = document_.title.slice(0, 30) + '…';
+  if (title.length > 30) {
+    title = title.slice(0, 30) + '…';
   }
 
-  return document_;
+  return {
+      uri: uri,
+      domain: domain,
+      title: title,
```
robertknight: Should use two-space indentation here.

robertknight:
Testing this locally I saw this message in the `make dev` output:

```
11:12:29 ws     | /home/robert/.virtualenvs/h/local/lib/python2.7/site-packages/sqlalchemy/orm/strategies.py:624: SAWarning: Multiple rows returned with uselist=False for lazily-loaded attribute 'Document.meta_title' 
11:12:29 ws     |   % self.parent_property)
```

robertknight:
LGTM. I tested annotating different copies of the same PDF and web pages locally via the extension and viewing document metadata on the stream. All worked as expected.

h/api/models/document.py
chdorner: Fixed.

h/api/storage.py
chdorner: Fixed.

tests/h/emails/reply_notification_test.py
chdorner: Fixed.

h/api/storage.py
chdorner: Fixed.

h/static/scripts/annotation-metadata.js
chdorner: Fixed.

robertknight:
LGTM

PR #3499, opened 2016-06-17
Add a service to manage group creation and membership

nickstenning:
In preparation for some substantial reorganisation of the groups views (which in turn is in preparation for changes which we'll be making to groups in the near future) this commit adds a GroupsService which hides some of the details of creating groups and managing their membership.

In this commit the service remains unused.

### DIFF
--- a/h/groups/services.py
+++ b/h/groups/services.py
@@ -1,0 +1,50 @@
+# -*- coding: utf-8 -*-
+
+from h.accounts import get_user
+from h.models import Group
+
+
+class GroupsService(object):
+
+    """A service for manipulating groups and group membership."""
+
+    def __init__(self, session, user_fetcher):
+        self.session = session
+        self.user_fetcher = user_fetcher
+
+    def create(self, name, userid):
+        """
+        Create a new group.
+
+        :param name: the human-readable name of the group
+        :param userid: the userid of the group creator
+
+        :returns: the created group
+        """
+        creator = self.user_fetcher(userid)
+        group = Group(name=name, creator=creator)
+        self.session.add(group)
+        self.session.flush()
+
+        return group
+
+    def member_join(self, group, userid):
+        """Add `userid` to the member list of `group`."""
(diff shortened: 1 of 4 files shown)


h/groups/services.py
```diff
@@ -0,0 +1,50 @@
...
+from h.models import Group
+
+
+class GroupsService(object):
+
+    """A service for manipulating groups and group membership."""
+
+    def __init__(self, session, user_fetcher):
+        self.session = session
+        self.user_fetcher = user_fetcher
```
seanh: Nice

h/groups/services.py
```diff
@@ -0,0 +1,50 @@
...
+        """
+        creator = self.user_fetcher(userid)
+        group = Group(name=name, creator=creator)
+        self.session.add(group)
+        self.session.flush()
+
+        return group
+
+    def member_join(self, group, userid):
+        """Add `userid` to the member list of `group`."""
```
seanh: Single-backticks is actually italics in rST, it's double backticks for code spans. Not sure if that's intentional?

seanh:
LGTM, still loving this services thing. @chdorner @finitud you probably want to have a look at this new "services" pattern if you haven't seen it already?

h/groups/services.py
nickstenning: Single backticks are an [unqualified reference](http://www.sphinx-doc.org/en/stable/markup/inline.html), not italics. I don't know if it's the "right" way to refer to parameters, though...

seanh: Oh yeah, it actually outputs `<cite>...</cite>`.

robertknight:
LGTM. As mentioned [on Slack](https://hypothes-is.slack.com/archives/public/p1466418590001796) I think it would be helpful to have a record somewhere explaining the motivation for introducing the service layer and policies for them - eg. are services allowed to depend on the `request` object (I presume not) or is that only permissible in the factory?

PR #3500, opened 2016-06-17
Rearrange groups views to take advantage of GroupsService

nickstenning:
Now that the business logic associated with administering groups is mostly encapsulated by the GroupsService (see #3499), we can substantially simplify the views code. This is achieved by doing the following:

- Retrieval of the group for join/leave/read views moves into a Pyramid traversal operation, so that each view simply receives the group object directly, rather than having to fetch it itself.

- Rather than each view needing to do its own `if request.authenticated_userid is None:` check, we use the Pyramid `effective_principals` predicate in the view config. This also allows us to remove a conditional from the read view, by matching different view functions dependent on the user's authentication status.

- Relieve the views of the need to do an explicit membership check by adding an ACL to the `Group` model. Only members of the group have the "read" permission, so we can then use `request.has_permission("read")`.

This results in somewhat less mock-heavy test code for views.py, as well as improved test coverage. Hopefully it will also make it easier to add and modify functionality in future, too.

_**N.B.** This PR depends on #3498 and #3499, and will need rebasing after these merge._

### DIFF
--- a/h/app.py
+++ b/h/app.py
@@ -134,7 +134,7 @@ def includeme(config):
     config.include('h.admin', route_prefix='/admin')
     config.include('h.badge')
     config.include('h.feeds')
-    config.include('h.groups')
+    config.include('h.groups', route_prefix='/groups')
     config.include('h.links')
     config.include('h.nipsa')
     config.include('h.notification')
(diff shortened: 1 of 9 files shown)


seanh:
LGTM, worth giving it a good test once #3499 is merged

robertknight:
I've rebased this. Just having a read through.

seanh:
Tested and seems to work well, :+1: from me

robertknight:
LGTM

PR #3503, opened 2016-06-20
Fix a couple of document model tests

seanh:
These tests were mocking classes that the code wasn't really using.

### DIFF
--- a/tests/h/api/models/document_test.py
+++ b/tests/h/api/models/document_test.py
@@ -250,7 +250,7 @@ class TestCreateOrUpdateDocumentURI(object):
         assert document_uri.created > created
         assert document_uri.updated > updated
 
-    def test_it_logs_a_warning_if_document_ids_differ(self, log, DocumentURI):
+    def test_it_logs_a_warning_if_document_ids_differ(self, log):
         """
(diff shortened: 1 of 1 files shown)


tests/h/api/models/document_test.py
```diff
@@ -250,21 +250,23 @@ def test_it_creates_a_new_DocumentURI_if_there_is_no_existing_one(self, db_sessi
...
         If there's an existing DocumentURI and its .document property is
         different to the given document it shoulg log a warning.
 
         """
+        session = mock_db_session()
+
         # existing_document_uri.document won't be equal to the given document.
         existing_document_uri = mock.Mock(document=mock_document())
-        DocumentURI.query.filter.return_value.first.return_value = (
+        session.query.return_value.filter.return_value.first.return_value = (
```
seanh: This is actually unnecessary, since `session` is a mock `session.query().filter().first()` will of course return another mock object, and that mock's `.document` property will be different to the `document=mock_document()` that's passed in to the function under test below.

But I sort of think it's better for making it clear what is being tested to leave this explicitly here than to have the tests implicitly rely on this behaviour of mocks.

nickstenning:
LGTM.

PR #3504, opened 2016-06-20
Separate annotations and notes in sidebar

sheetaluk:


h/static/scripts/annotation-metadata.js
```diff
@@ -70,6 +70,36 @@ function isNew(annotation) {
...
+function isTypeAnnotation(annotation) {
+  if (annotation && !annotation.deleted) {
+    if (annotation.target && annotation.target.length > 0 && annotation.target[0].selector) {
+      return true;
+    };
+  }
+  return false;
+}
+
+function isTypeReply(annotation) {
```
robertknight: There is already an `isReply` function in this file. Any reason not to use that?

h/static/scripts/annotation-metadata.js
```diff
@@ -96,4 +126,7 @@ module.exports = {
   isReply: isReply,
   isNew: isNew,
   location: location,
+  isTypePageNote: isTypePageNote,
```
robertknight: To make the naming more consistent, I suggest we just name this `isPageNote`.

h/static/scripts/root-thread.js
```diff
@@ -63,6 +63,14 @@ function RootThread($rootScope, annotationUI, searchFilter, viewFilter) {
       };
     }
 
+    var threadFilterFn = function (thread) {
+      if(state.selectedTab === 'annotation') {
+        return (metadata.isTypeAnnotation(thread.annotation) || metadata.isTypeReply(thread.annotation));
```
robertknight: `isTypeReply()` will always be false here because only top-level annotations (ie. not replies) are being filtered using `threadFilterFn`.

h/static/scripts/widget-controller.js
```diff
@@ -38,6 +39,27 @@ module.exports = function WidgetController(
 ) {
 
   /**
+   * Returns the number of top level annotations which are of type annotations
+   * and not notes or replies.
+   */
+  function countAnnotations(annotations) {
+    var totalAnnotations = annotations.reduce(function(annotationsCount, annotation) {
```
robertknight: It is usually easier to read a function if local variable names are short. It would be fine to use `total` and `count` rather than `totalAnnotations` and `annotationsCount` here.

h/static/scripts/widget-controller.js
```diff
@@ -38,6 +39,27 @@ module.exports = function WidgetController(
...
+      return metadata.isTypeAnnotation(annotation) ? annotationsCount+1 : annotationsCount;
+    }, 0);
+    return totalAnnotations;
+  };
+
+  /**
+   * Returns the number of top level annotations which are of type notes.
+   */
+  function countNotes(annotations) {
+    var totalNotes = annotations.reduce(function(notesCount, annotation) {
```
robertknight: As above, I'd suggest just calling these variables `total` and `count`.
```

File: h/static/scripts/widget-controller.js [hunk 6]
```diff
@@ -65,15 +87,19 @@ module.exports = function WidgetController(
     return rootThread.thread(annotationUI.getState());
   }
 
+  // By default select the annotation tab
+  annotationUI.selectTab('annotation');
```
robertknight: Rather than dispatching an action to select the default tab on startup, you can just define it as part of the default state of the UI in the `initialState()` function in `annotation-ui.js`.

h/static/scripts/widget-controller.js
```diff
@@ -343,6 +369,11 @@ module.exports = function WidgetController(
     return thread().totalChildren;
   };
 
+  $scope.filterAnnotationsByType = function (type) {
+    $scope.clearSelection();
```
robertknight: Clearing the selection when the current tab changes can be done in the reducer function that handles the state update when the selected tab changes. This will enforce that the selection is cleared when the tab is changed via another means.

h/static/scripts/widget-controller.js
```diff
@@ -343,6 +369,11 @@ module.exports = function WidgetController(
     return thread().totalChildren;
   };
 
+  $scope.filterAnnotationsByType = function (type) {
```
robertknight: Perhaps just call this `selectTab`?

h/static/scripts/annotation-metadata.js
```diff
@@ -70,30 +70,55 @@ function isNew(annotation) {
...
+          return selectors[k].start;
+        }
+      }
+    }
+  }
+  return Number.POSITIVE_INFINITY;
+}
 
 module.exports = {
   extractDocumentMetadata: extractDocumentMetadata,
```
robertknight: Nit: I suggest sorting `module.exports` keys in lexicographical order.

nickstenning:
This is looking pretty good! Here are a couple of things I've noticed so far:

- [ ] The "tabs" seem awfully close to the signin box when you're signed out:

screen shot 2016-06-24 at 13 15 04
screen shot 2016-06-24 at 13 14 46

- [ ] The story card says that if there are no cards in a tab, there should be no number rather than a zero.

- [ ] Looks like this currently breaks standalone pages for page notes.

More generally, I note that this work isn't feature flagged. Last time we worked on a large-ish change to the frontend code I think we talked about how we needed to feature flag these things so they don't get stuck in a branch as we work out the last few niggles.

I don't have a feeling for how hard this is to feature flag at this stage, but perhaps you could look into it?

h/static/scripts/annotation-metadata.js
```diff
@@ -70,30 +70,55 @@ function isNew(annotation) {
   return !annotation.id;
 }
 
+/** Return `true` if the given annotation is a page note, `false` otherwise. */
+function isPageNote(annotation) {
+  if (annotation && !annotation.deleted) {
```
robertknight: Annotations do not have a `deleted` property. See this discussion on Slack: https://hypothes-is.slack.com/archives/public/p1466767221002927

h/static/scripts/annotation-metadata.js
```diff
@@ -70,30 +70,55 @@ function isNew(annotation) {
...
+    }
+    if (annotation.target && (annotation.target.length === 0 || !annotation.target[0].selector)) {
+      return true;
+    }
+  }
+  return false;
+}
+
+/** Return `true` if the given annotation is a top level annotation, `false` otherwise. */
+function isAnnotation(annotation) {
```
robertknight: Couple of comments here:

1. You can simplify the `isPageNote()` function, conceptually, to:

```
isPageNote = annot => !isAnnotation(annot) && !isReply(annot)
```

2. Avoid checks for whether an argument is null/undefined unless there is a clear reason _why_ the parameter would ever be null. Otherwise you have introduced an extra path through your code and until we have better tools, it is difficult to find out after the fact whether such checks are necessary.

h/static/scripts/annotation-ui.js
```diff
@@ -55,6 +58,9 @@ function initialState(settings) {
 
     filterQuery: null,
 
+    // By default select the annotations tab
```
robertknight: This comment is a bit redundant IMO

h/static/scripts/annotation-ui.js
```diff
@@ -125,6 +132,9 @@ function reducer(state, action) {
       return Object.assign({}, state, {expanded: action.expanded});
     case types.HIGHLIGHT_ANNOTATIONS:
       return Object.assign({}, state, {highlighted: action.highlighted});
+    case types.SELECT_TAB:
+      return Object.assign({}, state,
+        {selectedTab: action.tab});
```
robertknight: The `{selectedTab ...}` literal here could go on the line above

h/static/scripts/build-thread.js
```diff
@@ -238,6 +238,11 @@ var defaultOpts = {
    */
   filterFn: undefined,
   /**
+   * Function that returns true if an annotation of a certain
```
robertknight: `threadFilterFn` is a generic predicate function, it doesn't have to be used for filtering annotations by type. So the comment here could be adjusted to something like "A filter function which should return true if a given annotation and its replies should be displayed`"

h/static/scripts/directive/test/selection-tabs-test.js
```diff
@@ -0,0 +1,69 @@
...
+var util = require('./util');
+
+describe('selectionTabs', function () {
+  before(function () {
+    angular.module('app', [])
+      .directive('selectionTabs', require('../selection-tabs'));
+  });
+
+  beforeEach(function () {
+    angular.mock.module('app');
```
robertknight: There is a shortcut you can use here to make injecting fake dependencies easier. `angular.mock.module` takes an optional second argument which is an object mapping dependency names to fakes.

```js
var fakeAnnotationUI = {};
angular.mock.module('app', {
  annotationUI: fakeAnnotationUI,
});
```

h/static/scripts/root-thread.js
```diff
@@ -57,6 +57,17 @@ function RootThread($rootScope, annotationUI, searchFilter, viewFilter) {
       };
     }
 
+    var threadFilterFn;
+    if (!state.filterQuery) {
+      threadFilterFn = function (thread) {
+        if (state.selectedTab === annotationUI.TAB_ANNOTATIONS) {
+          return (metadata.isAnnotation(thread.annotation));
```
robertknight: The braces around the `return`-ed expression here are redundant.

h/static/scripts/test/annotation-metadata-test.js
```diff
@@ -146,4 +146,57 @@ describe('annotation-metadata', function () {
       }), Number.POSITIVE_INFINITY);
     });
   });
+
+  describe('.isPageNote', function () {
+    it ('returns true for an annotation without selectors', function () {
+      assert.isTrue(annotationMetadata.isPageNote({
+        deleted: false,
```
robertknight: As mentioned above, although @nickstenning thinks we might have had a `deleted` property on annotations in the past, we don't currently have one. It might be the case that one is introduced internally on the server in future but the client will never see it, so we don't need to worry about it.

h/static/scripts/test/annotation-metadata-test.js
```diff
@@ -146,4 +146,57 @@ describe('annotation-metadata', function () {
...
+        target: []
+      }));
+    });
+    it ('returns true for an annotation with empty target', function () {
+      assert.isTrue(annotationMetadata.isPageNote({
+        deleted: false,
+        target: [{selector: undefined}]
+      }));
+    });
+    it ('returns false for an annotation without a targets', function () {
```
robertknight: s/without a targets/without a target

h/static/scripts/test/annotation-metadata-test.js
```diff
@@ -146,4 +146,57 @@ describe('annotation-metadata', function () {
...
+      assert.isFalse(annotationMetadata.isPageNote({
+        deleted: false,
+        target: undefined
+      }));
+    });
+    it ('returns false for an annotation which is a reply', function () {
+      assert.isFalse(annotationMetadata.isPageNote({
+        deleted: false,
+        target: [],
+        references: [{}]
```
robertknight: `references` is an array of strings rather than an array of objects.

h/static/scripts/test/annotation-metadata-test.js
```diff
@@ -146,4 +146,57 @@ describe('annotation-metadata', function () {
...
+        deleted: true
+      }));
+    });
+  });
+
+  describe ('.isAnnotation', function () {
+    it ('returns true if an annotation is a top level annotation', function () {
+      assert.isTrue(annotationMetadata.isAnnotation({
+        deleted: false,
+        target: [{selector: {}}]
```
robertknight: `target[n].selector` is an array rather than an object.

h/static/scripts/test/build-thread-test.js
```diff
@@ -262,6 +263,44 @@ describe('build-thread', function () {
         }]);
       });
     });
+
+    context('thread filtering', function () {
```
robertknight: AIUI when you have a set of tests _about_ a particular function/class etc., those are `describe()` blocks. When you have a set of tests that apply _when_ certain a certain condition is met, those are `context()` blocks.

h/static/scripts/test/build-thread-test.js
```diff
@@ -262,6 +263,44 @@ describe('build-thread', function () {
...
+        target: [{selector: {}}],
+      },{
+        id: '2',
+        text: 'note',
+        target: [{selector: undefined}],
+      }];
+
+      it('shows only annotations which are notes', function () {
+        var thread = createThread(fixture, {
+          threadFilterFn: function (thread) {
```
robertknight: From the point of view of `buildThread()`, the threadFilterFn is generic - so actually here you only need a single test which checks that only annotations matching the filter are included. eg. `it('shows only annotations matching the thread filter')`. The filter condition could be anything.

h/static/scripts/test/root-thread-test.js
```diff
@@ -50,6 +50,8 @@ describe('rootThread', function () {
       removeSelectedAnnotation: sinon.stub(),
       addAnnotations: sinon.stub(),
       setCollapsed: sinon.stub(),
+      selectTab: sinon.stub(),
+      TAB_ANNOTATIONS: 'annotation',
```
robertknight: Rather than redefining the constant here I'd suggest importing the _actual_ value. What we should do in future is get ourselves to a state where we can import interfaces, constants etc. in tests without importing the implementation.

h/static/scripts/test/root-thread-test.js
```diff
@@ -179,6 +181,53 @@ describe('rootThread', function () {
...
   });
 
+  describe('when the thread filter query is set', function () {
+    it('generates a thread filter function to filter out annotations', function () {
+      fakeBuildThread.reset();
+      fakeBuildThread.annotation = {target: [{ selector: {} }]};
+
+      fakeAnnotationUI.state = Object.assign({}, fakeAnnotationUI.state,
+        {selectedTab: 'annotation'});
+      fakeAnnotationUI.TAB_ANNOTATIONS = 'annotation';
```
robertknight: Rather than re-defining these constants in each test, just import them once when setting up `fakeAnnotationUI`

h/static/scripts/test/root-thread-test.js
```diff
@@ -179,6 +181,53 @@ describe('rootThread', function () {
     ]);
   });
 
+  describe('when the thread filter query is set', function () {
+    it('generates a thread filter function to filter out annotations', function () {
+      fakeBuildThread.reset();
+      fakeBuildThread.annotation = {target: [{ selector: {} }]};
```
robertknight: `fakeBuildThread` is a fake for the `buildThread()` function. What you are trying to test here is just that the thread filter function passed to `fakeBuildThread` behaves the way that you expect.

To do that you just need something like:

```js
var annot = {target: [{ selector: {} }]};
assert.equal(threadFilterFn(annot), true);
```

h/static/scripts/test/root-thread-test.js
```diff
@@ -179,6 +181,53 @@ describe('rootThread', function () {
     ]);
   });
 
+  describe('when the thread filter query is set', function () {
+    it('generates a thread filter function to filter out annotations', function () {
```
robertknight: The name of this test differs from what the test actually checks. Here this test generates a filter function which only _matches_ (not filters out) annotations. Same with the test below for notes.

h/static/scripts/widget-controller.js
```diff
@@ -4,6 +4,7 @@ var SearchClient = require('./search-client');
 var events = require('./events');
 var memoize = require('./util/memoize');
 var scopeTimeout = require('./util/scope-timeout');
+var metadata = require('./annotation-metadata');
```
robertknight: Imports should be sorted by require path where possible.

h/static/scripts/widget-controller.js
```diff
@@ -72,6 +94,9 @@ module.exports = function WidgetController(
   var visibleThreads = new VirtualThreadList($scope, window, thread());
   annotationUI.subscribe(function () {
     visibleThreads.setRootThread(thread());
+    $scope.totalAnnotations = countAnnotations(annotationUI.getState().annotations);
+    $scope.totalNotes = countNotes(annotationUI.getState().annotations);
+    $scope.tabSelection = annotationUI.getState().selectedTab;
```
robertknight: It confused me what `tabSelection` was when looking through the code above. I would suggest naming it `selectedTab` for consistency.

h/static/scripts/widget-controller.js
```diff
@@ -117,6 +142,25 @@ module.exports = function WidgetController(
     crossframe.call('scrollToAnnotation', annotation.$$tag);
   }
 
+  /** Returns the annotation type - note or annotation of the first annotation
+   *  in `results` whose ID is a key in `selectedAnnotationMap`.
+   */
+  function tabTypeFromSelection(selection, results) {
```
robertknight: I'm trying to follow a convention in new code that either: 1) All paths through the code return no value, in which case it always uses just `return;` or 2) All paths through the code explicitly return a value, in which case either `return <value>` or `return null;`. This makes it easier to spot a mistake where some path fails to return a value but should do.

h/static/scripts/widget-controller.js
```diff
@@ -322,6 +370,8 @@ module.exports = function WidgetController(
   };
 
   $scope.isLoading = isLoading;
+  $scope.tabAnnotations = annotationUI.TAB_ANNOTATIONS;
```
robertknight: Ah, I wondered where `tabAnnotations` was coming from. IMO it would be better to put these constants in a place that they can just be directly imported from the component and tests that use them.

h/templates/client/viewer.html
```diff
@@ -2,6 +2,14 @@
 
 (See gh2642 for rationale for 'ng-show="true"')
  -->
+<selection-tabs ng-if="!search.query() && selectedAnnotationCount() <= 0"
+  tab-selection="tabSelection"
+  total-annotations="totalAnnotations"
+  total-notes="totalNotes"
+  tab-annotations="tabAnnotations"
```
robertknight: As noted above, I think we should remove `tabAnnotations` and `tabNotes` here and just put the constants in a file that can be imported directly from the `selection-tabs` component via a simple import.

conordelahunty:
Looking good. Sheetal and I have discussed some of the spacing and alignment issues as well as some cosmetic changes. 

Also, I second the call for this work to be featured flagged, this can be improved upon greatly if we and the wider team have a few days to test it out and see what feels weak.

nickstenning:
Going to mark this is as WIP. Of course please unmark it when it's ready for review again!

h/static/scripts/test/root-thread-test.js
```diff
@@ -179,6 +185,47 @@ describe('rootThread', function () {
...
+      fakeBuildThread.reset();
+
+      fakeAnnotationUI.state = Object.assign({}, fakeAnnotationUI.state,
+        {selectedTab: 'annotation'});
+
+      rootThread.thread(fakeAnnotationUI.state);
+      var threadFilterFn = fakeBuildThread.args[0][1].threadFilterFn;
+
+      var annotation = {target: [{ selector: {} }]};
+      assert(threadFilterFn({annotation: annotation}));
```
nickstenning: This isn't going to stop me merging, but we should try where possible to use precise assertions -- [Chai has lots of them](http://chaijs.com/api/assert/#method_assert) -- because it improves the error messages we get when things go wrong. This should probably be `assert.isTrue`, because you're expecting a literal boolean.

h/static/scripts/test/root-thread-test.js
```diff
@@ -179,6 +185,47 @@ describe('rootThread', function () {
...
+      fakeBuildThread.reset();
+      fakeBuildThread.annotation = {target: [{}]};
+
+      fakeAnnotationUI.state = Object.assign({}, fakeAnnotationUI.state,
+        {selectedTab: 'note'});
+
+      rootThread.thread(fakeAnnotationUI.state);
+      var threadFilterFn = fakeBuildThread.args[0][1].threadFilterFn;
+
+      assert.equal(threadFilterFn(fakeBuildThread), true);
```
nickstenning: `assert.isTrue` as above.

h/static/scripts/test/root-thread-test.js
```diff
@@ -179,6 +185,47 @@ describe('rootThread', function () {
...
+      fakeBuildThread.reset();
+      fakeBuildThread.annotation = {target: [{}]};
+
+      fakeAnnotationUI.state = Object.assign({}, fakeAnnotationUI.state,
+        {selectedTab: 'annotation'});
+
+      rootThread.thread(fakeAnnotationUI.state);
+      var threadFilterFn = fakeBuildThread.args[0][1].threadFilterFn;
+
+      assert.equal(threadFilterFn(fakeBuildThread), undefined);
```
nickstenning: `assert.isUndefined` as above.

nickstenning:
LGTM. 

Perhaps have a look through the few small comments I left above? And there are a couple of issues to follow up with in more PRs:

> The "tabs" seem awfully close to the signin box when you're signed out

and

 >  **Looks like this currently breaks standalone pages for page notes.**

But because this is feature flagged, we can now merge it. Hooray :)

h/static/scripts/test/root-thread-test.js
robertknight: @sheetaluk - I would strongly 👍 this comment. Debugging failures of `assert.predicate(a, b)` is _much_ easier than `assert(condition)`

PR #3505, opened 2016-06-20
Clean up the welcome page

finitud:
The `/welcome(/)` endpoint now redirects to `/welcome/{slug}`, with a
randomly generated slug, and which renders the exact same page.
This allows for the onboarding experience to start on a page that
is clean of annotations.

See https://trello.com/c/FIiBdyTX/302-clean-up-the-welcome-page

### DIFF
--- a/h/views/__init__.py
+++ b/h/views/__init__.py
@@ -19,7 +19,8 @@ def includeme(config):
 
     # help
     config.add_route('help', '/docs/help')
-    config.add_route('onboarding', '/welcome')
+    config.add_route('onboarding', '/welcome/')
+    config.add_route('custom_onboarding', '/welcome/{welcome_slug}')
 
     # main
     config.add_route('annotation',
--- a/h/views/__init__.py
+++ b/h/views/__init__.py
@@ -20,7 +20,7 @@ def includeme(config):
     # help
     config.add_route('help', '/docs/help')
     config.add_route('onboarding', '/welcome/')
-    config.add_route('custom_onboarding', '/welcome/{welcome_slug}')
+    config.add_route('custom_onboarding', '/welcome/{slug}')
 
     # main
     config.add_route('annotation',
--- a/tests/h/views/help_test.py
+++ b/tests/h/views/help_test.py
@@ -1,0 +1,36 @@
+# -*- coding: utf-8 -*-
+
+from __future__ import unicode_literals
+
+import pytest
+import mock
+from pyramid import httpexceptions
+
+from h.views import help
+
+
+@pytest.mark.usefixtures('routes')
+def test_welcome_page_redirects_to_new_url(pyramid_request):
+    result = help.onboarding_page({}, pyramid_request)
+    assert isinstance(result, httpexceptions.HTTPFound)
+
+
+@pytest.mark.usefixtures('routes')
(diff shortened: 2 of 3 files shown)


h/views/__init__.py
```diff
@@ -19,7 +19,8 @@ def includeme(config):
 
     # help
     config.add_route('help', '/docs/help')
-    config.add_route('onboarding', '/welcome')
+    config.add_route('onboarding', '/welcome/')
+    config.add_route('custom_onboarding', '/welcome/{welcome_slug}')
```
nickstenning: It's a small thing, but let's just call the matchdict item `slug`. It's the `slug` in the welcome page URL so there's no need to reiterate that.

tests/h/views/help_test.py
```diff
@@ -0,0 +1,43 @@
...
+from __future__ import unicode_literals
+
+import pytest
+import mock
+from pyramid import httpexceptions
+
+from h.views import help
+
+
+def test_welcome_page_redirects_to_new_url(pyramid_request, routes):
```
nickstenning: Typically if you don't actually use a fixture within the test, you should use the `pytest.mark.usefixtures` decorator rather than passing it in, e.g.

```python
@pytest.mark.usefixtures('routes')
def test_welcome_page_redirects_to_new_url(pyramid_request):
    ...
```

tests/h/views/help_test.py
```diff
@@ -0,0 +1,43 @@
...
+
+
+def test_custom_welcome_page(pyramid_request, routes):
+    pyramid_request.matched_route.name = "onboarding"
+    result = help.custom_onboarding_page({}, pyramid_request)
+    assert not result['is_help']
+    assert result['is_onboarding']
+
+
+@pytest.fixture
```
nickstenning: I don't think any of this is needed any more now that you've split the view functions up nicely.

tests/h/views/help_test.py
```diff
@@ -0,0 +1,43 @@
...
+
+import pytest
+import mock
+from pyramid import httpexceptions
+
+from h.views import help
+
+
+def test_welcome_page_redirects_to_new_url(pyramid_request, routes):
+    pyramid_request.matched_route.name = "onboarding"
```
nickstenning: This isn't needed.

tests/h/views/help_test.py
```diff
@@ -0,0 +1,43 @@
...
+    result = help.onboarding_page({}, pyramid_request)
+    assert isinstance(result, httpexceptions.HTTPFound)
+
+
+def test_help_page_returns_is_help_true(pyramid_request):
+    pyramid_request.matched_route.name = "help"
```
nickstenning: Ditto.

tests/h/views/help_test.py
```diff
@@ -0,0 +1,43 @@
...
+    result = help.help_page({}, pyramid_request)
+    assert result['is_help']
+
+
+def test_custom_welcome_page(pyramid_request, routes):
+    pyramid_request.matched_route.name = "onboarding"
```
nickstenning: Ditto.

nickstenning:
LGTM. Will merge when tests go green.

PR #3506, opened 2016-06-20
Fix unhelpful username error message when registering a new account

### LINKED ISSUE #3391: Unhelpful error messages for invalid usernames
We don't say what is required of a valid username nor what is wrong with the one given:

screenshot from 2016-05-31 17-54-21

seanh:
Fixes  #3391

### DIFF
--- a/h/accounts/schemas.py
+++ b/h/accounts/schemas.py
@@ -193,7 +193,8 @@ class RegisterSchema(CSRFSchema):
             unblacklisted_username,
         ),
         title=_('Username:'),
-        hint=_('between {min} and {max} characters').format(
+        hint=_('between {min} and {max} characters, containing only letters, '
+               'numbers, periods, and underscores').format(
             min=models.USERNAME_MIN_LENGTH,
             max=models.USERNAME_MAX_LENGTH
         ),
--- a/h/accounts/schemas.py
+++ b/h/accounts/schemas.py
@@ -185,7 +185,10 @@ class RegisterSchema(CSRFSchema):
             colander.Length(
                 min=models.USERNAME_MIN_LENGTH,
                 max=models.USERNAME_MAX_LENGTH),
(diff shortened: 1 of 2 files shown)


h/accounts/schemas.py
```diff
@@ -185,12 +185,16 @@ class RegisterSchema(CSRFSchema):
...
+                '(?i)^[A-Z0-9._]+$',
+                msg=_("Must contain only letters, numbers, periods and "
+                      "underscores")),
             unique_username,
             unblacklisted_username,
         ),
         title=_('Username:'),
-        hint=_('between {min} and {max} characters').format(
+        hint=_('between {min} and {max} characters, containing only letters, '
+               'numbers, periods and underscores').format(
```
seanh: This is shown next to the _Username:_ field on the form. I thought it would be good to tell the user what characters they're allowed to use before they come up with a username, not after.

chdorner: Good point, we should probably require a longer password in the future.

h/accounts/schemas.py
```diff
@@ -185,12 +185,16 @@ class RegisterSchema(CSRFSchema):
             colander.Length(
                 min=models.USERNAME_MIN_LENGTH,
                 max=models.USERNAME_MAX_LENGTH),
-            colander.Regex('(?i)^[A-Z0-9._]+$'),
+            colander.Regex(
+                '(?i)^[A-Z0-9._]+$',
+                msg=_("Must contain only letters, numbers, periods and "
```
chdorner: I'm no native speaker, do we care about the [Oxford comma](https://en.wikipedia.org/wiki/Serial_comma)?

seanh: Hmm, I'm leading towards including it because Wikipedia says most US style guides include it, including Strunk & White which I happen to have here

seanh: Done

chdorner:
LGTM.

PR #3507, opened 2016-06-20
Fix the Help panel on the stream and standalone annotation pages

robertknight:
- Fix incorrect reference to `onShowHelpPanel` property in <top-bar>
   template

 - Do not show URL entry in metadata display if there is no annotation
   guest connected to the sidebar, as is the case in the stream and
   permalink pages.


nickstenning:
LGTM.

PR #3508, opened 2016-06-20
Fix escaped HTML in login error message

### LINKED ISSUE #3501: Login error messages contain escaped HTML
screenshot from 2016-06-20 11-31-00

seanh:
If you try to login to an unactivated account the error message that the
server returns contains HTML:

You haven't activated your account yet. <strong>Please check your email and
open the link to activate your account</strong>.

Render this client-side as HTML rather than escaping the tags.

Fixes #3501.

### DIFF
--- a/h/templates/client/login_form.html
+++ b/h/templates/client/login_form.html
@@ -30,7 +30,8 @@
                 >Please enter your username or email address.</li>
             <li class="form-error"
                 ng-show="login.username.$error.response"
-                >{{login.username.responseErrorMessage}}</li>
+                ng-bind-html="login.username.responseErrorMessage">
+            </li>
           </ul>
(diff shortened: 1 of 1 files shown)


robertknight:
LGTM, although is there any risk of unexpected HTML being returned as the response error message?

seanh:
There are three different error messages that field can return: https://github.com/hypothesis/h/blob/38d0fbe0e7f9251df5aa7dc79eb18defd8221d09/h/accounts/schemas.py#L123 They're all hardcoded error messages that don't contain any variables from user data or anything, so there's no risk. Of course, we don't know what new error messages might be added in the future..

robertknight:
LGTM

PR #3511, opened 2016-06-20
Add script to assist with tagging the Docker image for the H service

robertknight:
I appreciate our deployment workflow might be changing pretty soon after the client/server repo split but for the moment this documents the sequence of steps I've been going through to tag Docker images for new releases.


chdorner:
LGTM.

PR #3512, opened 2016-06-20
Remove the query property!

nickstenning:
Finally, the day is here! Thread-local sessions need be no more! (Although
they're sticking around for the moment).

The use of a "magic" query property makes testing difficult and causes
confusion about which session is the correct session to use when writing new
code.

This commit finally removes the magical query property on the model base class.
There are now only two mechanisms by which code can get hold of a database
session:

1. be passed it explicitly, or be passed an object which holds a reference to a session (such as the request object).
2. construct one yourself

This will make it possible to fully isolate request sessions from one another and to use explicitly request-scoped sessions rather than a thread-local scoped_session registry.


chdorner:
:tada:

PR #3514, opened 2016-06-21
Don't access the database session directly in tests

nickstenning:
All tests should access the database session using the `db_session` fixture. This will make it easier to turn `h.db.Session` into a stateless session factory in future.

This PR doesn't address the test factories or conftest.py. These will be fixed in a subsequent commit.


chdorner:
LGTM.

PR #3515, opened 2016-06-21
Move test factories into tests/common/

nickstenning:
In preparation for putting more generic support modules in tests/common/, this commit moves the test factories module.

I've also removed the factories_test.py file. This is for a couple of reasons:

- we don't test test code (or if we did, would we test test test code?).
- factories are tested in the context where they are used (i.e. in regular tests).
- having factories_test.py in `h/` while `factories.py` lives at the top level or in `common/` doesn't make a lot of sense -- and I'm not inclined to start adding tests to `tests/common/` across the board.
- not explicitly testing `factories.py` means it is more obvious from test coverage if parts of it are not being used and can be removed.


chdorner:
This looks good to me. I'll leave this for @seanh to review and merge, since I remember you two were talking about testing the factories.

seanh:
`tests/common` seems like a good idea.

I'm not sure I understand why `factories_test.py` is tests for tests? `factories.py` is code for generating objects, it's used by tests but it's not itself tests.

I found it convenient while developing the new factories to edit `factories.py`, run `factories_test.py` and see what happened, edit it again. As opposed to having to do the same thing manually, repeatedly in a Python shell if we didn't have any tests to run.

If there's a bug in a factory, I'd rather have that caught close to the source than only have some far away tests for the production code failing, I think that could lead to some quite hard to find bugs as there's non-obvious stuff in the tests factories (especially the dependent generation of document, document_uri and document_metas) that if broken could cause tests that use it to break in non-obvious ways.

Also a bug in a factory may lead to tests that use the factory _passing_ when they should be failing. But the tests for the factory itself should be failing.

Even though the factories that we have right now are very simple and mostly just use features of `factory_boy` and the ORM classes, they may evolve to become more complex and if they do the tests will be there to become more useful for documentation, verification and change management.

[make_metadata()](https://github.com/hypothesis/h/blob/c886c6891fc4a701ccbf43b939e166a61556305d/tests/factories.py#L113) is already fairly non-trivial and worth having a simple test for.

nickstenning:
If part of our code is so complicated to test that we need non-trivial test harnesses, I'd suggest that we should invest more time in simplifying the code under test rather than in writing tests for test harnesses. 

Perhaps "testing tests" is overly specific. "Testing test-related code" is probably more like it.

And as for "factories could be broken in a way that causes tests to pass"... Isn't that true of *any* test code? Isn't that true of the code that tests that the test code is working?

The goals of automated testing are:

1. ensure that the system under test is working as expected
2. provide feedback about the suitably of the design of the system under test

I don't see how testing the factories code fulfils either of these objectives if you consider that the "system under test" is *not* the factories but the underlying model classes.

seanh:
Ok, I don't think there's anything wrong in principle with having tests for test harnesses. I can see that it would be nice if the production code was always simple enough to not require complex test harnesses, but I think _would be nice_ is correct. I think `factories_test.py` is useful because: when making changes to `factories.py` I can write a failing test for the new behaviour that I want, edit the code, run the tests, repeat until it works, instead of having to do it manually in a shell. It's useful to pin down any bugs in `factories.py` close to the code, in its own tests. And it documents the intended behaviour and features of the factories.

The factories _are_ pretty simple at the moment though (even if I think `make_metadata()` is pushing it), so this doesn't seem crucially important.

I'm afraid we might get stuck here so I'm happy to see it deleted if you want?

PR #3516, opened 2016-06-21
Make test for URL-ness of selection stricter

### LINKED ISSUE #3513: Bug: Inserting Links on texts with colons
As soon as a colon is connected directly to a word in a text I want to insert a link into, the editor asks not for the link, but the description.

I created a gif:
colon and link

robertknight:
Increase the strictness of the heuristic for whether the selection is a
URL or not in the "Convert selection to link" editor toolbar command.

Require the selection to contain something that looks like a scheme
followed by an authority, rather than just a scheme.

Fixes #3513

### DIFF
--- a/h/static/scripts/markdown-commands.js
+++ b/h/static/scripts/markdown-commands.js
@@ -101,7 +101,7 @@ function convertSelectionToLink(state, linkType) {
   }
 
   var newState;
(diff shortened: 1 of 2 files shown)


nickstenning:
LGTM.

PR #3517, opened 2016-06-21
Allow multiple 'uri' parameters to the search query

nickstenning:
In order to have a better chance of retrieving all the annotations we know about for a document, it should be possible to query by all known URIs for that document.

This commit makes it possible to supply multiple 'uri' parameters to the API search endpoint. The resulting query will match annotations in which `target.scope` (i.e. the normalised target URI) matches *any* of the URIs in the set expanded from *all* of the query parameter values.

This is an entirely backwards compatible change. In the case where only one `uri` parameter is provided, the query output is identical.

### DIFF
--- a/h/api/search/query.py
+++ b/h/api/search/query.py
@@ -170,13 +170,18 @@ class UriFilter(object):
         self.request = request
 
     def __call__(self, params):
-        uristr = params.pop('uri', None)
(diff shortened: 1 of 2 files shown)


chdorner:
LGTM. It's probably worth to update the documentation as well, right?

PR #3524, opened 2016-06-22
Update PDF fingerprint searching / handling

chdorner:
This PR changes the client to:

* Search by fingerprint and non-file URI when the document is a fingerprint
* Changes the `target.source` of newly created annotations back to a non-file URI (otherwise uses fingerprint)

#3517 made it possible on the backend to search for multiple uris.
#3497 needs the changes of this PR (and has to remove the new `target.locator` property again).

### DIFF
--- a/h/static/scripts/annotator/plugin/pdf-metadata.js
+++ b/h/static/scripts/annotator/plugin/pdf-metadata.js
@@ -27,7 +27,11 @@ function PDFMetadata(app) {
  */
 PDFMetadata.prototype.getUri = function () {
   return this._loaded.then(function (app) {
-    return fingerprintToURN(app.documentFingerprint);
+    var uri = getPDFURL(app);
+    if (!uri) {
+      uri = fingerprintToURN(app.documentFingerprint);
+    }
+    return uri;
   });
 };
 
@@ -52,11 +56,9 @@ PDFMetadata.prototype.getMetadata = function () {
       {href: fingerprintToURN(app.documentFingerprint)}
     ];
 
-    // Local file:// URLs should not be saved in document metadata.
-    // Entries in document.link should be URIs. In the case of
-    // local files, omit the URL.
-    if (app.url.indexOf('file://') !== 0) {
-      link.push({href: app.url});
+    var url = getPDFURL(app);
+    if (url) {
+      link.push({href: url});
     }
 
     return {
@@ -69,6 +71,17 @@ PDFMetadata.prototype.getMetadata = function () {
 
 function fingerprintToURN(fingerprint) {
   return 'urn:x-pdf:' + String(fingerprint);
+}
+
+function getPDFURL(app) {
+  // Local file:// URLs should not be saved in document metadata.
+  // Entries in document.link should be URIs. In the case of
+  // local files, omit the URL.
+  if (app.url.indexOf('file://') !== 0) {
+    return app.url;
+  }
+
+  return null;
 }
 
 module.exports = PDFMetadata;
--- a/h/static/scripts/annotator/plugin/test/pdf-metadata-test.js
+++ b/h/static/scripts/annotator/plugin/test/pdf-metadata-test.js
@@ -9,19 +9,23 @@ describe('pdf-metadata', function () {
 
     var event = document.createEvent('Event');
     event.initEvent('documentload', false, false);
+    fakeApp.url = 'http://fake.com';
     fakeApp.documentFingerprint = 'fakeFingerprint';
     window.dispatchEvent(event);
 
     return pdfMetadata.getUri().then(function (uri) {
-      assert.equal(uri, 'urn:x-pdf:fakeFingerprint');
+      assert.equal(uri, 'http://fake.com');
     });
   });
 
   it('does not wait for the PDF to load if it has already loaded', function () {
-    var fakePDFViewerApplication = {documentFingerprint: 'fakeFingerprint'};
+    var fakePDFViewerApplication = {
+      url: 'http://fake.com',
+      documentFingerprint: 'fakeFingerprint',
+    };
     var pdfMetadata = new PDFMetadata(fakePDFViewerApplication);
     return pdfMetadata.getUri().then(function (uri) {
-      assert.equal(uri, 'urn:x-pdf:fakeFingerprint');
+      assert.equal(uri, 'http://fake.com');
     });
   });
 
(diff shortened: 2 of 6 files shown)


h/static/scripts/cross-frame.coffee
```diff
@@ -46,11 +46,24 @@ module.exports = class CrossFrame
...
+          searchUris = [info.uri]
+
           documentFingerprint = null
           if info.metadata and info.metadata.documentFingerprint
             documentFingerprint = info.metadata.documentFingerprint
+
+            reducer = (uris, link) ->
+              uris.push(link['href'])
+              return uris
+            searchUris = info.metadata.link.reduce(reducer, []);
```
robertknight: If you have an array of values that you want to produce an equal-length array of some derived value, you can use `Array#map` instead. eg. `info.metadata.link.map((link) -> link.href)`

h/static/scripts/annotator/plugin/pdf-metadata.js
```diff
@@ -71,4 +73,13 @@ function fingerprintToURN(fingerprint) {
   return 'urn:x-pdf:' + String(fingerprint);
 }
 
+function getPDFURL(app) {
+    // Local file:// URLs should not be saved in document metadata.
+    // Entries in document.link should be URIs. In the case of
+    // local files, omit the URL.
+    if (app.url.indexOf('file://') !== 0) {
```
robertknight: I prefer that if functions return a value through one path, then they explicitly return a value through all paths rather than just relying on the implicit `return undefined`. That makes it obvious that any null values returned are intentional.

In short, just add `return null` at the end here.

h/static/scripts/widget-controller.js
```diff
@@ -177,7 +177,7 @@ module.exports = function WidgetController(
       // Remove client from list of active search clients
       searchClients.splice(searchClients.indexOf(searchClient), 1);
     });
-    searchClient.get({uri: uri, group: group});
+    searchClient.get({uri: uris, group: group});
```
robertknight: Passing an array here works fine without any changes in `SearchClient`?

h/static/scripts/cross-frame.coffee
chdorner: Much nicer, thanks!

h/static/scripts/annotator/plugin/pdf-metadata.js
chdorner: :+1:

h/static/scripts/widget-controller.js
chdorner: Yes, it looks like Angular (?) converts an array to multiple `uri=` in the query string.

nickstenning:
One detail that I think has been overlooked here is that we probably don't want to search by all URIs on HTML pages -- for the moment. This causes problems when people have weird metadata on the page: see 8100ec4.

chdorner:
@nickstenning as discussed in Slack, [this](https://github.com/hypothesis/h/pull/3524/files#diff-c45cbeb67fa1305bc941ba32b44d869fR54) only runs when the document is a PDF.

h/static/scripts/widget-controller.js
```diff
@@ -219,12 +221,10 @@ module.exports = function WidgetController(
     var group = annotationUI.hasSelectedAnnotations() ?
       null : groups.focused().id;
 
-    for (var i=0; i < urls.length; i++) {
-      _loadAnnotationsFor(urls[i], group);
-    }
+    _loadAnnotationsFor(searchUris, group);
```
robertknight: Since `_loadAnnotationsFor` is now outside of a loop, it gets called here even if `searchUris` is empty. The result when I tested this locally on a PDF is that the client ended up trying to fetch all public annotations. The call to `_loadAnnotationsFor` needs to be moved inside the `if (searchUris.length) {...}` check below.

chdorner: Oh yes, you're right. It did a search first without a `uri` in the params. I moved it down into the same `if` block as the stream filter code.

h/static/scripts/annotator/plugin/test/pdf-metadata-test.js
```diff
@@ -45,7 +49,19 @@ describe('pdf-metadata', function () {
...
 
     describe('#getUri', function () {
-      it('returns the URN-ified document fingerprint as its URI', function () {
+      it('returns the non-file URI', function() {
+        return pdfMetadata.getUri().then(function (uri) {
+          assert.equal(uri, 'fakeUrl');
+        });
+      });
+
+      it('returns the URN-ified document fingerprint as its URI when url is a file url', function () {
```
robertknight: nit. Should be consistent about `URL`, `URI`, `URN` etc. being in caps in test names

h/static/scripts/annotator/plugin/test/pdf-metadata-test.js
```diff
@@ -9,19 +9,23 @@ describe('pdf-metadata', function () {
 
     var event = document.createEvent('Event');
     event.initEvent('documentload', false, false);
+    fakeApp.url = 'fakeUrl';
```
robertknight: I think it makes tests more useful as documentation when the fake data at least looks like the real data, so eg. `http://fake.com` instead of `fakeUrl`.

h/static/scripts/test/cross-frame-test.coffee
```diff
@@ -72,5 +72,21 @@ describe 'CrossFrame', ->
       fakeBridge.onConnect.yields(channel)
       crossframe.connect()
       assert.deepEqual(crossframe.frames, [
-        {channel: channel, uri: uri, documentFingerprint: null}
+        {channel: channel, uri: uri, searchUris: [uri], documentFingerprint: null}
+      ])
+
+    it 'updates the frames array with multiple search uris when PDF', ->
```
robertknight: s/when PDF/when the document is a PDF/

h/static/scripts/test/cross-frame-test.coffee
```diff
@@ -72,5 +72,21 @@ describe 'CrossFrame', ->
...
+      uri = 'http://example.com/test.pdf'
+      fingerprint = 'urn:x-pdf:fingerprint'
+      channel = {call: sandbox.stub().yields(null, {
```
robertknight: `call:` should be on a new line here.

robertknight:
LGTM

PR #3525, opened 2016-06-22
Fix broken document uri unique constraint migration

seanh:


seanh:
Alright, I've tested this on some prod data and it looks right to me.

nickstenning:
I think this is going to be problematic because the work isn't batched, and as a result it's going to lock the `document_uri` table for the entire duration of the migration -- running locally this seems to be at least several minutes so that ain't going to work in production.

seanh:
Weird, I tested this locally on prod data and it ran in a few seconds. I'll look again.

seanh:
Marking this as WIP until I have a chance to investigage how slow it is on prod data, and look into batching it. Didn't have time to get round to it today.

seanh:
Re-tested quickly, this runs in < 15s on prod data on my laptop

nickstenning:
Takes ~23s here. We can't lock the `annotation` and `document*` tables in production for half a minute. It should be straightforward to break this up into multiple transactions which handle some number of rows at a time.

seanh:
@nickstenning Obviously this is WIP but how does this look to you: <https://github.com/hypothesis/h/pull/3525/commits/0c586d2b8556f20b27f754f6b93e6b8142ba7dcc>:

1. Query for 25 document_uri rows.
2. Update the 25 ORM objects
3. Commit
4. Query for 25 document_uri rows...

nickstenning:
That general approach looks good. Two things I noticed while looking at this that you might want to address:

- If you have a query that identifies multiple rows you want to delete, you can just call `.delete` on the query object rather than doing `for thing in query: session.delete(thing)`
- As previously discussed, schema migrations and data migrations should ideally be in different migration files. This allows us to run the data migration repeatedly and idempotently if necessary before running the schema migration.

seanh:
I've batched all the `DELETE`s and `UPDATE`s into transactions of 25 rows at a time and re-tested. Each call to `commit()` takes less than 0.05s on my laptop using production data, the migration completes and adds the `NOT NULL`s to the columns without crashing, and there are still lots of rows left in the table. I didn't pick through the prod data post migration manually to make sure it's right, but I have written tests (see individual commits in this PR) and they pass.

seanh:
> If you have a query that identifies multiple rows you want to delete, you can just call .delete on the 
> query object rather than doing for thing in query: session.delete(thing)

Apparently this doesn't work when `.limit()` has been used.

> As previously discussed, schema migrations and data migrations should ideally be in different 
> migration files. This allows us to run the data migration repeatedly and idempotently if necessary before 
> running the schema migration.

The migration script already on master combines data and schema migrations into one, in this PR I'm just trying to fix the script to 1. Not crash when there are already existing rows in the DB with emptry string types and content_types and 2. Batch the deletes and updates.

Of course it should be pretty simple to separate out the schema changes but maybe a separate PR?

nickstenning:
I've removed the commit/revert you had on this branch, and I've made a couple of small modifications to the migration script. Running it locally all looks good. I'll merge this, then split the migration in two, and then we can get it out to stage.

seanh:
Thanks Nick.

In case I want to write tests for another migration in the future and want to refer to them, the commit that had the tests for this migration was: 12bc84af3f5535865416614deedf0f46fd7edfce

seanh:
Interesting what you did instead of importing the models. I had exactly that problem when I wanted to test. I would have to modify the model code to remove the constraints, then create the db from that, then import data, then run the migration. Otherwise you wouldn't be able to import the pre-migration data because the constraints would already be on the db and it's violate them.

PR #3526, opened 2016-06-22
Fix adder breaking after making a selection containing no text

### LINKED ISSUE #3523: Adder stops working after making a selection that does not include any text
### Steps to reproduce
1. Go to a page with an image on it and activate H
2. Select the image, _being careful not to select only the image and not any whitespace or text around it_. Depending on the page this can be tricky with the mouse.
3. Select a piece of text elsewhere on the page

### Expected behaviour
The adder should appear after step 3.

### Actual behaviour
The adder does not appear and a console error is logged, the same as #3521

Tested with Chrome v52 but the issue should be the same for any browser.

### Additional details
Any other means of making a non-empty selection which does not include any text should also trigger the issue.

robertknight:
Ignore the selection change if the new selection does not contain
any text.

 - Add missing check for `rangeUtil.selectionFocusRect()` returning null,
   which happens if the selection contains no text
 - Add tests for selection change handling
 - Create the `Guest` instance with dependencies stubbed out afresh
   for each test case in guest-test.js

Fixes #3523

### DIFF
--- a/h/static/scripts/annotator/guest.coffee
+++ b/h/static/scripts/annotator/guest.coffee
@@ -350,6 +350,14 @@ module.exports = class Guest extends Annotator
     @crossframe?.call('focusAnnotations', tags)
 
   _onSelection: (range) ->
+    selection = Annotator.Util.getGlobal().getSelection()
+    isBackwards = rangeUtil.isSelectionBackwards(selection)
+    focusRect = rangeUtil.selectionFocusRect(selection)
+    if !focusRect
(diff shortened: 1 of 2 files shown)


robertknight:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;VA5eAGa8&#x2F;383-adder-stops-working-after-rapid-clicking-on-page-in-certain-browsers"> Adder stops working after rapid clicking on page in certain browsers</a>

PR #3527, opened 2016-06-22
PyLint: Don't warn about parens in print("foo")

seanh:


nickstenning:
Looks to me like this warning doesn't fire if files have the correct `from __future__ import print_function` at the top, so in that sense it's probably valid and we shouldn't disable it.

seanh:
Ah right

PR #3529, opened 2016-06-22
Don't show permissions error before annotation

### LINKED ISSUE #3167: "You do not have permission to see this annotation" shown briefly before annotation appears
### Steps to reproduce

1. Run h locally
2. Run bouncer locally
3. Serve a PDF file at localhost (using `python -m SimpleHTTPServer`)
4. Open the localhost PDF file in the browser, activate h, make an annotation, get and follow the direct link to the annotation

I've only been able to reproduce the annotation using a PDF, though I'm not sure the file being a PDF has anything to do with the bug.

### Expected behaviour

The annotation should just be shown, with no "You do not have permission to see this annotation" card shown in its place before it appears.

### Actual behaviour

A "You do not have permission to see this annotation" card appears briefly in the sidebar, then is replaced by the annotation:

Chrome 49.0.2623.75 (64-bit), Ubuntu Gnome 15.10, Hypothesis master branch.

### Additional details

It looks like we don't distinguish between annotations that are unavailable because they haven't been received or threaded yet, and ones that the user doesn't have permission to see: <https://github.com/hypothesis/h/blob/932c06a14fc9444fdb0f7b49fc833990e10b4a63/h/static/scripts/widget-controller.js#L208>.

seanh:
Don't momentarily show a "You do not have permission to see this
annotation" card before showing the annotation when loading a
direct-linked annotation.

Fixes #3167.

When following direct links to annotations on certain documents,
for example large PDF files that take a while to load,
a "You do not have permission to see this annotation" card was being
displayed briefly in place of the annotation card, before quickly being
replaced by the correct annotation card.

Previously the logic of the WidgetController.isLoading() function was to
return true if we are currently waiting for the response to an
annotation search request that we've sent.

The theory is that this was happening:

1. Client loads, but PDF has not yet loaded, so URL is unknown.

2. Since the URL is unknown, the sidebar does not start fetching
   annotations and isLoading() returns false.

3. Since the client a) Has a selection but b) Is not in a loading state
   - it thinks that it failed to fetch the annotation and displays the
   "You do not have permission to see this annotation" message,

4. Once the PDF finishes loading, the URL is updated and that is
   reported to the sidebar which then performs a search query and
   isLoading() starts to return true until the search query response
   arrives then it returns false again.

The fix is to make isLoading() return true if _either_ we've sent an
annotation search request and are waiting on the response _or_ we're
still waiting for the document to finish loading and haven't sent the
search request yet. So instead of returning false, then true, then false
again isLoading() will just keep returning true until both the document
has loaded and the search response has been received, then it will start
returning false.

### DIFF
--- a/h/static/scripts/test/widget-controller-test.js
+++ b/h/static/scripts/test/widget-controller-test.js
@@ -344,6 +344,17 @@ describe('WidgetController', function () {
   });
 
   describe('direct linking messages', function () {
+
+    beforeEach(function () {
+      // The document has finished loading.
+      fakeCrossFrame.frames = [
+        {
+          uri: 'http://www.example.com',
+          searchUris: [],
+        }
+      ];
+    });
+
     it('displays a message if the selection is unavailable', function () {
       annotationUI.selectAnnotations(['missing']);
       $scope.$digest();
@@ -360,6 +371,18 @@ describe('WidgetController', function () {
     it('does not a show a message if there is no selection', function () {
       annotationUI.selectAnnotations([]);
       $scope.$digest();
+      assert.isFalse($scope.selectedAnnotationUnavailable());
+    });
+
+    it("doesn't show a message if the document isn't loaded yet", function () {
+      // No search requests have been sent yet.
+      searchClients = [];
+      // There is a selection but the selected annotation isn't available.
+      annotationUI.selectAnnotations(['missing']);
+      // The document hasn't finished loading.
+      fakeCrossFrame.frames = [];
+      $scope.$digest();
+
       assert.isFalse($scope.selectedAnnotationUnavailable());
     });
 
--- a/h/static/scripts/widget-controller.js
+++ b/h/static/scripts/widget-controller.js
@@ -181,7 +181,17 @@ module.exports = function WidgetController(
   }
 
   function isLoading() {
-    return searchClients.length > 0;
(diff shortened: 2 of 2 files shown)


seanh:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;gXYtlAja&#x2F;361-bug-misleading-message-while-an-accessible-direct-link-annotation-loads"> [Bug] Misleading message while an accessible direct-link annotation loads</a>

robertknight:
This needs rebasing on master following @chdorner 's changes from yesterday.

seanh:
Rebased

seanh:
Not sure why the Jenkins build for this branch keeps failing on pip install

h/static/scripts/test/widget-controller-test.js
```diff
@@ -344,6 +344,17 @@ describe('WidgetController', function () {
   });
 
   describe('direct linking messages', function () {
+
+    beforeEach(function () {
+      // The document has finished loading.
+      fakeCrossFrame.frames = [
```
robertknight: Small nit. The convention is to format this in a more compact way as:

```
fakeCrossFrame.frames = [{
  uri: '...',
  searchUris: [],
}];
```

robertknight:
LGTM

PR #3531, opened 2016-06-22
Fix missing PDF highlights

### LINKED ISSUE #3530: Annotation highlights not working on PDFs
### Steps to reproduce
1.  Go to a cool PDF.  say, http://www.supremecourt.gov/opinions/15pdf/14-1373_83i7.pdf
2.  Annotate a little
3. See annotation highlight disappear before your eyes :(

### Expected behaviour
After annotating, the selected text should remain highlighted. 

### Actual behaviour
After annotating, the selected text is no longer highlighted.  Clicking on a annotation card (sometimes?) focuses on the annotation, but there's no highlight.

Version:
0.29.0
User agent:
Mozilla/5.0 (Macintosh; Intel Mac OS X 10_11_5) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/51.0.2704.103 Safari/537.36
URL:
urn:x-pdf:0906f5d5c8ec3bb0a29606126c6f9c6d
PDF fingerprint:
0906f5d5c8ec3bb0a29606126c6f9c6d
Username:
wikinotes
Date:
22 Jun 2016 10:43:46 -0700

### Additional details

nohighlights

nickstenning:
fda63b4 unfortunately broke highlighting in PDFs, because the PDF anchoring code first finds the text in the document, and then uses a TextPositionSelector and the HTML anchoring code to select a range in
the rendered page for highlighting.

This commit fixes that issue by moving the TextQuoteSelector check up a level into `Guest#anchor`.

I've added two tests for this in `guest-test.coffee` -- one tests that annotations where the target contains no TextQuoteSelector are marked as orphans, and the second tests that the low-level anchoring code is never called in this case.

I have also removed the integration test, on the basis that I don't think integration testing failure cases is a good habit to get into.

Fixes #3530.

### DIFF
--- a/h/static/scripts/annotator/anchoring/html.coffee
+++ b/h/static/scripts/annotator/anchoring/html.coffee
@@ -50,11 +50,6 @@ exports.anchor = (root, selectors, options = {}) ->
       when 'RangeSelector'
         range = selector
 
-  # Assert that there is a quote selector, since otherwise we cannot validate
-  # the rest of the annotation's selectors.
-  if not quote?
-    return Promise.reject(new Error('quote selector not found'))
-
   # Assert the quote matches the stored quote, if applicable
   maybeAssertQuote = (range) ->
     if quote?.exact? and range.toString() != quote.exact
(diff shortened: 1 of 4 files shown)


robertknight:
Changes LGTM. Just going to check out this branch and do some manual testing.

robertknight:
I tested viewing annotations on recently annotated PDFs and HTML documents from the stream, creation of annotations on PDFs and HTML documents and the original page that prompted disabling anchoring for annotations without quote selectors. All worked fine.

PR #3535, opened 2016-06-23
Remove duplicated state in `<annotation>` component

robertknight:
A lot of state was duplicated in the `<annotation>` component, such as the current unsaved changes, the editor state, the document links and metadata and the annotation object being displayed by the annotation card.

The result of all this was several bugs (https://trello.com/c/fotxgB2S, https://trello.com/c/nRaDXSdL) when that state got out of sync.

This PR is a series of changes that:
 * Remove as much state from `<annotation>` as possible at the moment. All we need is the current saved annotation (`vm.annotation`), the current unsaved changes (stored in the `drafts` service) and whether a save is in progress. From that we can compute everything else that is displayed in the view.
 * Gets us closer towards making annotations immutable objects in the sidebar app. The benefit of this is that all the places that derive something from annotation objects (eg. the domain and title associated with an annotation) can avoid recalculation if the annotation instance has not changed.

Combined, these changes allow a lot of code to be removed from `annotation.js` and many of the tests are made redundant or greatly simplified.

This is a large PR and will make a lot more sense reviewed commit-by-commit. I also expect that we'll want to separate them out into several PRs.

### DIFF
--- a/h/static/scripts/annotation-sync.coffee
+++ b/h/static/scripts/annotation-sync.coffee
@@ -53,11 +53,6 @@ module.exports = class AnnotationSync
       this._syncCache(channel)
     @bridge.onConnect(onConnect)
 
-  # Provide a public interface to the annotation cache so that other
-  # sync services can lookup annotations by tag.
-  getAnnotationForTag: (tag) ->
-    @cache[tag] or null
-
   sync: (annotations) ->
     annotations = (this._format a for a in annotations)
     @bridge.call 'sync', annotations, (err, annotations = []) =>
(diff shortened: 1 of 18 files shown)


seanh:
Have you tested:

* Edit an annotation, make some changes, hit cancel instead of save - the original unedited version of the annotation should be restored

* Edit an annotation, make some changes, hit save but have the server return some sort of an error, now hit - the original unedited version of the annotation should be restored

robertknight:
> Edit an annotation, make some changes, hit cancel instead of save -
> the original unedited version of the annotation should be restored

This is achieved now simply by deleting the unsaved draft when the user clicks 'Cancel'

>   Edit an annotation, make some changes, hit save but have the server return some sort of
> an error, now hit - the original unedited version of the annotation should be restored

Actually when an error happens we want to return to edit mode and preserve the unsaved changes until the user explicitly presses Cancel. This is done by setting an `isSaving` flag when the user clicks 'Post' and then unsetting that flag in the event of an error. Unlike the "happy" path, we don't delete the draft in this case.

robertknight:
Discussed the high-level picture of the changes with @seanh yesterday. I'm now going to close this and pull out the individual commits into smaller PRs.

PR #3536, opened 2016-06-23
Use direct links in reply notification emails

seanh:


seanh:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;1u8CHiKr&#x2F;374-direct-link-to-conversation-from-reply-notification-emails"> Direct link to conversation from reply notification emails</a>

h/emails/reply_notification.py
```diff
@@ -33,10 +31,9 @@ def generate(request, notification):
         'document_title': document_title,
         'document_url': notification.parent.target_uri,
         'parent': notification.parent,
-        'parent_url': parent_url,
         'parent_user': notification.parent_user,
         'reply': notification.reply,
-        'reply_url': reply_url,
+        'reply_url': links.incontext_link(request, notification.parent),
```
nickstenning: Shouldn't this be `notification.reply`? Or `parent_url`?

nickstenning:
I think this is missing a bunch of work described in the card. You've changed the template context but not the templates, so now the HTML email is going to contain `<a href="" ...>`.

The story card suggests new copy for the reply notification emails in order to account for the new way we're generating the links. That needs to be integrated here.

seanh:
Just realised this is gonna break it bouncer's off...

nickstenning:
(Sorry, didn't mean to close this.)

seanh:
Also I'm not sure this is correct for replies to replies...

seanh:
I need https://github.com/hypothesis/h/pull/3539 merged before I can finish this

seanh:
untitled screencast 23

seanh:
If bouncer is not available then it uses the reply's individual page instead.

If bouncer _is_ available and the reply is a reply to a reply then it still uses the direct link to the root annotation of the thread, not the intermediate parent annotation.

seanh:
Although it was missing from the Trello card, I left the unsubscribe link in the email.

dwhly:
Yay!

h/emails/reply_notification.py
seanh: It happens to work either way currently (since `incontext_link()` returns a direct link to the thread root anno, any anno in the thread will do as input and give the same output) but I think passing in `notification.reply` is semantically correct so I've changed it to that

h/emails/reply_notification.py
```diff
@@ -20,11 +21,8 @@ def generate(request, notification):
     if not document_title:
         document_title = notification.parent.target_uri
 
-    parent_url = request.route_url('annotation', id=notification.parent.id)
-    reply_url = request.route_url('annotation', id=notification.reply.id)
-
-    reply_user_url = request.route_url('stream.user_query',
-                                       user=notification.reply_user.username)
+    reply_url = (links.incontext_link(request, notification.reply) or
```
nickstenning: IMO it would be clearer and more consistent with, for example, the code immediately above if this was written out explicitly:

```python
reply_url = links.incontext_link(request, notification.reply)
if not reply_url:
    reply_url = request.route_url('annotation', id=notification.reply.id)
```

h/templates/emails/reply_notification.html.jinja2
```diff
@@ -1,23 +1,14 @@
...
-<hr>
-<p>
-  <small>
-    If you&rsquo;d rather not receive these notifications you can
-    <a href="{{ unsubscribe_url }}">unsubscribe now</a>.
-  </small>
-</p>
+<p>{{ reply_user.username }} has
+<a href="{{ reply_url }}">left a reply to your annotation</a>
+on &ldquo;{{ document_title }}&rdquo;:</p>
```
nickstenning: I know it wasn't in the Google Doc, but I think we should probably link the document title to the reply URL as well, especially as `document_title` can fall back to being a URL if we don't have a title.

h/templates/emails/reply_notification.html.jinja2
```diff
@@ -1,23 +1,14 @@
-<p>
-  <a href="{{ reply_user_url }}">{{ reply_user.username }}</a>
-  has just left a reply to your annotation on
-  &ldquo;<a href="{{ document_url }}">{{ document_title }}</a>&rdquo;.
-</p>
-<p>
-  On <a href="{{ parent_url }}">{{ parent.created | human_timestamp }}</a>
```
nickstenning: Looks like the `human_timestamp` Jinja2 extension will be made unused by this change, so let's remove it.

h/templates/emails/reply_notification.txt.jinja2
```diff
@@ -1,15 +1,11 @@
-{{ reply_user.username }} has just left a reply to your annotation on "{{ document_title }}".
-{{ document_url }}
+{{ reply_user.username }} has left a reply to your annotation {{ reply_url }} on "{{ document_title }}":
```
nickstenning: I'm not sure having `{{ reply_url }}` in this line in the plain text version makes much sense.

> seanh has left a reply to your annotation https://hpt.is/pHdz_jxhEeakRBv01oK7Sg/example.com/foo/bar on "Example Page":

This looks odd, I think.

tests/h/emails/reply_notification_test.py
```diff
@@ -162,3 +199,7 @@ def token_serializer(self, pyramid_config):
         serializer.dumps.return_value = 'FAKETOKEN'
         pyramid_config.registry.notification_serializer = serializer
         return serializer
+
+    @pytest.fixture
+    def links(self, patch):
```
nickstenning: Can we please try and keep fixtures alphabetically organised so they're easy to find?

h/templates/emails/reply_notification.html.jinja2
seanh: Wait, I'm confused. `reply_url` is either the in context link to the reply, or the individual page of the reply. When `document_title` falls back to being a URL it'll be the URL of the annotated document (`notification.parent.target_uri`). Presumably we don't want `<a href="{{ in context link to reply }}">{{ URL of document }}</a>`. Did you mean that we should link `document_title` to the URL of the document (`notification.parent.target_uri`)?

nickstenning: The "in-context" link takes you to the annotation "in-context" ... i.e. on the annotated document. Sure, we could link to the target URI, or we could link to the target URI + a fragment that automatically focuses the relevant conversation. I don't see what we lose by doing the latter.

tests/h/emails/reply_notification_test.py
seanh: I've alphabetized the whole list

h/templates/emails/reply_notification.txt.jinja2
seanh: Done

h/templates/emails/reply_notification.html.jinja2
seanh: Done

h/emails/reply_notification.py
seanh: Done.

Why do you say this is clearer though? Is it because the `or` version requires wrapping in parens because it doesn't fit on one line?

In general I would have thought that `foo = thing_1 or thing_2` is quite Pythonic.

seanh:
All done except I'm awaiting feedback on [this comment](https://github.com/hypothesis/h/pull/3536#discussion_r68565209)

h/templates/emails/reply_notification.html.jinja2
seanh: We'll either by linking to the hyp.is URL which may redirect to the target uri plus fragment or to via, or if bouncer is off we'll be linking to the reply's individual page, but okay - done

seanh: These makes me realise that in the case where document_title is a URL we'll be displaying a URL in quotation marks, which might look odd...

h/emails/reply_notification.py
nickstenning: I think in some cases it could be clearer, but here, where `thing_1` and `thing_2` are both function calls with their own arguments and are long enough that they have to split over two lines, it's time to just write it out in full.

nickstenning:
LGTM.

h/emails/reply_notification.py
seanh: Makes sense

PR #3540, opened 2016-06-24
Add thread_root_id property to Postgres annotation model

seanh:
Copy the thread_root_id property from the Elastic to the Postgres
annotation model. Fixes a crash in incontext_link() because it now uses
Postgres not Elastic annotations.


nickstenning:
Fixed in cd1d3950f0280bb0c32ce350ff0138e4e3a42cff already.

Issue #3380, opened 2016-05-29
New annotations not moved to current group when switching groups


robertknight:
### Steps to reproduce
1. Create a new annotation whilst a group _A_ is selected, but don't publish it
2. Switch groups

### Expected behaviour
Annotation should move to new group

### Actual behaviour
Annotation does not move to new group.

### Browser/system information
Firefox 48

### Additional details
- Behavior should be the same whether or not the annotation has any text
- This looks like a regression with the new threading implementation

nickstenning:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;nRaDXSdL&#x2F;606-bug-new-annotations-not-moved-to-current-group-when-switching-groups-3380"><img src="https:&#x2F;&#x2F;github.trello.services&#x2F;images&#x2F;trello-icon.png" width="12" height="12"> [Bug]New annotations not moved to current group when switching groups #3380</a>

PR #3541, opened 2016-06-24
Fix #2965 (1/N) - Fix redundant calls to onEditText() in markdown editor component

### LINKED ISSUE #2965: Annotation search does not find edits made in current session
**Steps to reproduce**

3. Search for _K_

**Expected results**

**Actual results**

robertknight:
_This is the first in a set of PRs to fix #2965 and #3380 which are ultimately caused by certain state for annotations being duplicated and getting out of sync._

This fixes an issue where onEditText() would be called unnecessarily
when the markdown editor lost focus.

Simplify the code by just listening for the 'input' event, which all our
target browsers support.


nickstenning:
:+1:

For anyone else looking at this: http://caniuse.com/#feat=input-event

PR #3542, opened 2016-06-24
Fix #2965 (2/N) - Remove unused alternative delete button from <annotation> component

### LINKED ISSUE #2965: Annotation search does not find edits made in current session

robertknight:
Remove an alternative annotation deletion button which was part of an
older UI for annotation cards and is no longer used.


nickstenning:
LGTM.

PR #3543, opened 2016-06-24
Fix #2965 (3/N) - Match annotations by ID or tag in drafts service and when removing annotations

### LINKED ISSUE #2965: Annotation search does not find edits made in current session

robertknight:
This is the first in a set of PRs which fixes https://github.com/hypothesis/h/issues/2965 and simplifies `annotation.js` by storing and updating annotation objects in a way that fits the standard architecture for apps using Redux: 

1. Annotations become plain, immutable objects which reflect the saved version of the annotation on the server or the initial version of the annotation received from Annotator (for new annotations)
2. The annotation editor derives its state from only:
 - The saved/new annotation
 - The current draft for that annotation (if it is being edited) and
 - A flag that indicates whether the annotation is currently being saved.
3. When an annotation is created or updated on the server, the local instance is replaced with the newly saved version and any local-only properties (eg. whether the annotation was successfully anchored) are copied across from the previous object. The code was already partially doing step (3) but in an incomplete way.

In order to do this, we need to be able to match annotations by ID rather than instance in a few places. Annotations have two IDs - a local tag which is assigned to all annotations when they are sent across the bridge between the `<iframe>` and the page and an ID which is assigned by the server.

Previously we were matching annotations by reference equality or by server-assigned ID. Instead we need to match them by server ID or by local tag. Matching by local tag is needed for new annotations which do not have a server-assigned ID yet.

### DIFF
--- a/h/static/scripts/annotation-ui.js
+++ b/h/static/scripts/annotation-ui.js
@@ -77,15 +77,25 @@ var types = {
   SET_SORT_KEY: 'SET_SORT_KEY',
 };
 
+/**
+ * Return a copy of `current` with all matching annotations in `annotations`
+ * removed.
+ */
 function excludeAnnotations(current, annotations) {
-  var idsAndTags = annotations.reduce(function (map, annot) {
-    var id = annot.id || annot.$$tag;
-    map[id] = true;
-    return map;
-  }, {});
+  var ids = {};
+  var tags = {};
+  annotations.forEach(function (annot) {
+    if (annot.id) {
+      ids[annot.id] = true;
+    }
+    if (annot.$$tag) {
+      tags[annot.$$tag] = true;
+    }
+  });
   return current.filter(function (annot) {
-    var id = annot.id || annot.$$tag;
-    return !idsAndTags[id];
+    var shouldRemove = (annot.id && (annot.id in ids)) ||
+                       (annot.$$tag && (annot.$$tag in tags));
+    return !shouldRemove;
   });
 }
 
--- a/h/static/scripts/test/annotation-ui-test.js
+++ b/h/static/scripts/test/annotation-ui-test.js
@@ -1,10 +1,19 @@
 'use strict';
 
+var immutable = require('seamless-immutable');
+
 var annotationUIFactory = require('../annotation-ui');
-
 var annotationFixtures = require('./annotation-fixtures');
-
 var unroll = require('./util').unroll;
+
+var defaultAnnotation = annotationFixtures.defaultAnnotation;
+
+var fixtures = immutable({
+  pair: [
+    Object.assign(defaultAnnotation(), {id: 1, $$tag: 't1'}),
+    Object.assign(defaultAnnotation(), {id: 2, $$tag: 't2'}),
+  ],
+});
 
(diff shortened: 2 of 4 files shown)


h/static/scripts/annotation-ui.js
```diff
@@ -77,15 +77,24 @@ var types = {
...
+      idsAndTags[annot.id] = true;
+    }
+    if (annot.$$tag) {
+      idsAndTags[annot.$$tag] = true;
+    }
+  });
   return current.filter(function (annot) {
-    var id = annot.id || annot.$$tag;
-    return !idsAndTags[id];
+    var shouldRemove = (annot.id && idsAndTags[annot.id]) ||
```
nickstenning: I know that it's not likely to happen, but there's a potential problem here if angular generates a tag that looks like an annotation ID.

It seems that it would be just as fast and just as clear to have an `ids` set and a `tags` set, rather than one joint `idsAndTags` set, and it would completely avoid this possibility.

nickstenning: On an only partially-related note, it might be easier to see that you're using this object as a set-like thing if the condition you checked was `annot.id in idsAndTags` rather than `idsAndTags[annot.id]`.

h/static/scripts/test/annotation-ui-test.js
```diff
@@ -49,6 +49,21 @@ describe('annotationUI', function () {
       annotationUI.removeAnnotations([annot]);
       assert.deepEqual(annotationUI.getState().annotations, []);
     });
+
+    it('matches annotations to remove by ID', function () {
+      var annot = annotationFixtures.defaultAnnotation();
+      annotationUI.addAnnotations([annot]);
+      annotationUI.removeAnnotations([{id: annot.id}]);
+      assert.deepEqual(annotationUI.getState().annotations, []);
```
nickstenning: I know it's a bit pedantic, but a minimal passing implementation for this test would be one that simply removed all annotations, regardless of id, tag, or anything else. Perhaps it would be wise to add a couple more annotations before performing the remove just so it's a bit clearer that it's doing the right thing?

robertknight:
> I know that it's not likely to happen, but there's a potential problem here if angular
> generates a tag that looks like an annotation ID.

It is our code that generates the tag, so we can make sure this doesn't happen.

h/static/scripts/test/annotation-ui-test.js
robertknight: Good point. Will fix.

nickstenning:
> It is our code that generates the tag value, so we can make sure this doesn't happen.

Even if it is, it's not immediately apparent in the piece of code I was talking about, so I'd still be inclined to keep the `tags` and `ids` sets separate.

robertknight:
> Even if it is, it's not immediately apparent in the piece of code I was talking about,
> so I'd still be inclined to keep the tags and ids sets separate.

Fair enough. I have reworked this and also the subsequent PR which previously relied on the same assumption.

nickstenning:
LGTM.

PR #3546, opened 2016-06-25
Prevent long URLs in quotes and bodies causing overflow

### LINKED ISSUE #3421: Long words in quotes can overflow
### Steps to reproduce
1. Create an annotation which highlights the following quote:

> Lebow, D. G..; Lick, D. W. & Hartman, H. J. (2003/2004). HyLighter and Interactive Annotation: New Technology to Develop Higher-Order Thinking Skills. Inquiry: Critical Thinking Across the Curriculum, XXIII (1 & 2 ) 69-79. https://www.academia.edu/7907843/Lebow_D._G.._Lick_D._W._and_Hartman_H._J._2003_2004_._HyLighter_and_Interactive_Annotation_New_Technology_to_Develop_Higher-Order_Thinking_Skills._Inquiry_Critical_Thinking_Across_the_Curriculum_XXIII_1_and_2_69-79

2. Save annotation

### Expected behaviour
Quote should be wrapped correctly and annotation card width should not be affected

### Actual behaviour
Quote overflows annotation card border

Tested with Firefox v48

### Notes
The `overflow-wrap` CSS property can be used to allow breaking of words if the line does not contain sufficient other breaking points.

The width of the annotation thread's content area needs to be restricted so that it does not overflow the card.

robertknight:
Set a maximum width on the annotation content to prevent it overflowing
the card and set `overflow-wrap` on annotation quotes and bodies so that
long words without any natural breakpoints are broken in the middle if
necessary to prevent overflow.

Fixes #3421

### DIFF
--- a/h/static/styles/annotation-thread.scss
+++ b/h/static/styles/annotation-thread.scss
@@ -39,6 +39,9 @@ li:first-child .annotation-thread--top-reply {
 
 .annotation-thread__content {
   flex-grow: 1;
+
+  // Prevent annotation content from overflowing the container
+  max-width: 100%;
 }
 
 // Darken expand/collapse toggle when an annotation is hovered. This is only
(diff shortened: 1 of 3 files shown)


nickstenning:
LGTM.

PR #3548, opened 2016-06-27
Fix #2965 (4/N) Improve behavior when selecting pre-created annotations

### LINKED ISSUE #2965: Annotation search does not find edits made in current session

### LINKED ISSUE #3547: Selecting a pre-created annotation results in permission error being shown
### Steps to reproduce
1. Activate H in a page, select text and click the 'Annotate' button
2. After the sidebar opens, click on the highlight for the new annotation in the page

### Expected behaviour
The annotation card for the new annotation should remain visible in the sidebar

### Actual behaviour
The annotation card for the new annotation disappears and is replaced by a 'You do not have permission ...' error.

### Additional details
This is happening because the selection mechanism currently does not support selecting annotations which have not yet been created on the server, which I refer to here as "pre-created". We should either fix this mechanism or at least implement some better behavior.

robertknight:
Selecting a pre-created annotation in the page resulted in a
'You do not have permission to see this annotation' being displayed
because the selection mechanism only supports annotations which have a
server-assigned ID.

This commit improves that behavior by ignoring annotations without IDs
when selecting annotations, so clicking on a newly created annotation which
has not been saved will just clear the selection. Changing the selection
mechanism to support annotations without IDs requires changes in quite
a few places so I decided not to do it for the moment.

It also fixes an issue in AnnotationUISync that prevents Annotation
objects being replaced in the UI state after they are updated on the
server. Previously AnnotationUISync would receive the tags of
annotations that had been selected in the page and would then look up
annotations in a cache maintained by AnnotationSync in order to get
Annotation objects that could be passed to `annotationUI` functions
which needed objects with an ID.

This commit avoids the issue by changing the inputs to `annotationUI`
functions to be a list of IDs/tags to (de-)select or focus and then
just passing the tags directly to those functions in AnnotationUISync.

This assumes that annotation IDs and tags never conflict, which is
currently true and something we should be able to easily enforce.

Fixes #3547
Also a step towards fixing #2965 because it fixes the stale cache issue.

### DIFF
--- a/h/static/scripts/annotation-sync.coffee
+++ b/h/static/scripts/annotation-sync.coffee
@@ -53,11 +53,6 @@ module.exports = class AnnotationSync
       this._syncCache(channel)
     @bridge.onConnect(onConnect)
 
-  # Provide a public interface to the annotation cache so that other
-  # sync services can lookup annotations by tag.
-  getAnnotationForTag: (tag) ->
-    @cache[tag] or null
-
   sync: (annotations) ->
     annotations = (this._format a for a in annotations)
     @bridge.call 'sync', annotations, (err, annotations = []) =>
--- a/h/static/scripts/annotation-ui-controller.js
+++ b/h/static/scripts/annotation-ui-controller.js
@@ -21,7 +21,7 @@ function AnnotationUIController($rootScope, $scope, annotationUI) {
   });
 
   $rootScope.$on(events.ANNOTATION_DELETED, function (event, annotation) {
-    annotationUI.removeSelectedAnnotation(annotation);
+    annotationUI.removeSelectedAnnotation(annotation.id);
   });
 }
 
--- a/h/static/scripts/annotation-ui-sync.js
+++ b/h/static/scripts/annotation-ui-sync.js
@@ -9,7 +9,6 @@ var uiConstants = require('./ui-constants');
  * @name AnnotationUISync
  * @param {$window} $window An Angular window service.
  * @param {Bridge} bridge
- * @param {AnnotationSync} annotationSync
  * @param {AnnotationUI} annotationUI An instance of the AnnotatonUI service
  * @description
  * Listens for incoming events over the bridge concerning the annotation
@@ -17,29 +16,36 @@ var uiConstants = require('./ui-constants');
  * that the messages are broadcast out to other frames.
  */
 // @ngInject
-function AnnotationUISync($rootScope, $window, bridge, annotationSync,
-  annotationUI) {
-  // Retrieves annotations from the annotationSync cache.
-  var getAnnotationsByTags = function (tags) {
-    return tags.map(annotationSync.getAnnotationForTag, annotationSync);
-  };
+function AnnotationUISync($rootScope, $window, annotationUI, bridge) {
+  function lookupByTag(tag) {
+    return annotationUI.getState().annotations.find(function (annot) {
+      return annot.$$tag === tag;
+    });
+  }
+
+  function findIDsForTags(tags) {
+    var ids = [];
+    tags.forEach(function (tag) {
+      var annot = lookupByTag(tag);
+      if (annot && annot.id) {
+        ids.push(annot.id);
+      }
+    });
+    return ids;
+  }
 
   var channelListeners = {
     showAnnotations: function (tags) {
-      tags = tags || [];
-      var annotations = getAnnotationsByTags(tags);
-      annotationUI.selectAnnotations(annotations);
+      var ids = findIDsForTags(tags || []);
+      annotationUI.selectAnnotations(ids);
       annotationUI.selectTab(uiConstants.TAB_ANNOTATIONS);
     },
     focusAnnotations: function (tags) {
-      tags = tags || [];
-      var annotations = getAnnotationsByTags(tags);
-      annotationUI.focusAnnotations(annotations);
+      annotationUI.focusAnnotations(tags || []);
     },
     toggleAnnotationSelection: function (tags) {
-      tags = tags || [];
-      var annotations = getAnnotationsByTags(tags);
-      annotationUI.toggleSelectedAnnotations(annotations);
+      var ids = findIDsForTags(tags || []);
+      annotationUI.toggleSelectedAnnotations(ids);
     },
     setVisibleHighlights: function (state) {
       if (typeof state !== 'boolean') {
--- a/h/static/scripts/annotation-ui.js
+++ b/h/static/scripts/annotation-ui.js
@@ -20,6 +20,13 @@ function freeze(selection) {
   }
 }
 
+function toSet(list) {
+  return list.reduce(function (set, key) {
+    set[key] = true;
+    return set;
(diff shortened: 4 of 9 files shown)


nickstenning:
Needs a rebase. Marking as WIP.

seanh:
Tested and it seems to fix the issue for me

h/static/scripts/annotation-sync.coffee
```diff
@@ -53,11 +53,6 @@ module.exports = class AnnotationSync
       this._syncCache(channel)
     @bridge.onConnect(onConnect)
 
-  # Provide a public interface to the annotation cache so that other
-  # sync services can lookup annotations by tag.
-  getAnnotationForTag: (tag) ->
-    @cache[tag] or null
-
   sync: (annotations) ->
```
seanh: This was only used by `annotation-ui-sync` and this pr removes the part of `annotation-ui-sync` that was using it, so this is simply deleted

h/static/scripts/annotation-sync.coffee
```diff
@@ -53,11 +53,6 @@ module.exports = class AnnotationSync
       this._syncCache(channel)
     @bridge.onConnect(onConnect)
 
-  # Provide a public interface to the annotation cache so that other
-  # sync services can lookup annotations by tag.
-  getAnnotationForTag: (tag) ->
-    @cache[tag] or null
-
```
seanh: This was only used by `annotation-ui-sync` and this pr removes the part of `annotation-ui-sync` that was using it, so this is simply deleted

h/static/scripts/annotation-ui-controller.js
```diff
@@ -21,7 +21,7 @@ function AnnotationUIController($rootScope, $scope, annotationUI) {
   });
 
   $rootScope.$on(events.ANNOTATION_DELETED, function (event, annotation) {
-    annotationUI.removeSelectedAnnotation(annotation);
```
seanh: Not your fault, but what a horrible method name. `unselect(annotation)` or `removeFromSelection(annotation)` or something would be clearer.

h/static/scripts/annotation-ui.js
```diff
@@ -290,12 +280,13 @@ module.exports = function (settings) {
     },
 
     /** De-select an annotation. */
-    removeSelectedAnnotation: function (annotation) {
+    removeSelectedAnnotation: function (id) {
```
seanh: Since the function only uses `annotation.id` it can just accept `id` as arg, doesn't need the rest of `annotation`.

h/static/scripts/annotation-ui-controller.js
```diff
@@ -21,7 +21,7 @@ function AnnotationUIController($rootScope, $scope, annotationUI) {
...
+    annotationUI.removeSelectedAnnotation(annotation.id);
```
seanh: This pr changes `removeSelectedAnnotation()` to take an `id` not the whole annotation, so this is just updating the call.

h/static/scripts/annotation-ui-sync.js
```diff
@@ -9,37 +9,43 @@ var uiConstants = require('./ui-constants');
...
+  function findIDsForTags(tags) {
+    var ids = [];
+    tags.forEach(function (tag) {
+      var annot = lookupByTag(tag);
+      if (annot && annot.id) {
+        ids.push(annot.id);
+      }
+    });
+    return ids;
+  }
```
seanh: `selectAnnotations()` (called several times below) now takes annotation ids not whole annotations, hence `getAnnotationsByTags()` is no longer needed, `findIDsForTags()` (which uses `lookupByTag()`) replaces it.

h/static/scripts/annotation-ui-sync.js
```diff
@@ -9,37 +9,43 @@ var uiConstants = require('./ui-constants');
...
-  // Retrieves annotations from the annotationSync cache.
-  var getAnnotationsByTags = function (tags) {
-    return tags.map(annotationSync.getAnnotationForTag, annotationSync);
-  };
+function AnnotationUISync($rootScope, $window, annotationUI, bridge) {
+  function lookupByTag(tag) {
+    return annotationUI.getState().annotations.find(function (annot) {
+      return annot.$$tag === tag;
+    });
+  }
```
seanh: It seems `lookupByTag()` is called only once, by `findIDsForTags()` below

h/static/scripts/annotation-ui-sync.js
```diff
@@ -9,37 +9,43 @@ var uiConstants = require('./ui-constants');
...
-      var annotations = getAnnotationsByTags(tags);
-      annotationUI.focusAnnotations(annotations);
+      annotationUI.focusAnnotations(tags || []);
     },
     toggleAnnotationSelection: function (tags) {
-      tags = tags || [];
-      var annotations = getAnnotationsByTags(tags);
-      annotationUI.toggleSelectedAnnotations(annotations);
+      var ids = findIDsForTags(tags || []);
+      annotationUI.toggleSelectedAnnotations(ids);
```
seanh: These three chunks are all just because the args accepted by `selectAnnotations()`, `focusAnnotations()` and `toggleSelectedAnnotations()` have changed. No behaviour change here.

seanh:
Sorry for taking notes on your PR, my way of trying to make sense of the changeset. Two things:

1. The use of IDs sometimes and tags other times is confusing. I think you said somewhere that changing the code to support selecting of unsaved annotations (by tag, since they don't have ids yet) was too much work for right now, hence I guess why selecting (and deselecting etc) still takes ids not tags as input. But otherwise the move is towards using tags, not ids, to identify annotations throughout the frontend code?

2. Is it possible to break this up into two or more separate commits? One that fixes the 'You do not have permission to see this annotation' problem, and another that fixes the 'Annotation objects being replaced in the UI state' problem?

This is not exactly the clearest part of the codebase and the changeset is a little hard to follow all as one diff.

annotation-sync, annotation-ui-sync, annotation-ui-controller, annotation-ui, the bridge, the channel, the crossframe ... sigh.

nickstenning:

robertknight:
> But otherwise the move is towards using tags, not ids, to identify annotations
> throughout the frontend code?

Once this cleanup is merged and the way tags are allocated to annotations has changed so that we can rely on _all_ annotations having tags, then that might indeed be a sensible thing to do.

One wart is that direct-linking uses the selection mechanism and direct-links refer to IDs rather than tags.

robertknight:
> One that fixes the 'You do not have permission to see this annotation' problem, and
> another that fixes the 'Annotation objects being replaced in the UI state' problem?

There isn't really a separate commit for this because the fix is inherent in the approach. The closest thing is the difference between `getAnnotationsByTags` in the old code which returned _N_ annotations for _N_ tags and `findIDsForTags` in the new code which returns `M <= N` IDs for _N_ tags, where M < N if some tags do not have corresponding IDs. Does that make sense?

PR #3553, opened 2016-06-28
Use external client

nickstenning:
This is a work in progress to show the changes we'd have to make to this repository as a result of splitting the client out into its own repo.

The only complicated bit is updating the Pyramid application's assets subsystem to deal with the fact that we now have two directories which contain built assets. One, as currently, at `build/`, which contains "site" assets, and one, installed from npm, at `node_modules/hypothesis/build/`, which contains the built client.

The rest of the changes are just removing unneeded stuff from the build system, the test harness, and so on.

I've not included the ~30,000 line removal of code commit, but if you want to see it you can look at http://github.com/hypothesis/h/compare/master...use-external-client-code-removed. The tests pass on that branch.

Still to do:

- [x] a couple of minor fixes to the Python tests
- [x] update the `hypothesis-buildext` command to account for new asset location


nickstenning:
The tests are now passing on the version of this branch which has the client code removed.

nickstenning:
One more note. Using this setup, local development of the client in parallel with the web service will have changed a little. You'll need to set up the `h` repo to point to an `npm link`ed version of the client repository, which you can do by running

    npm link

in the `client` repository root, and then

    npm link hypothesis

in the `h` repository root. Once that's done, you can use `gulp watch` as usual in the client repository. Or at least you'll be able to once we've made a few more tweaks to the client repository so it doesn't try and build the extension bundles.

gulpfile.js
```diff
@@ -249,9 +226,9 @@ function changedAssets(prevManifest, newManifest) {
 
 var debouncedLiveReload = debounce(function () {
```
robertknight: The live reload server is used for testing the sidebar, so presumably won't be needed in the main repository any more assuming that the gulp process watching/building client assets is run from the other repo.

nickstenning: :+1: I can add a commit to this branch to remove it.

nickstenning:
Ok. This is now updated to point at the released version of `hypothesis`, v0.31.0 on npm.

Please review and "LGTM" or otherwise, but do *not* merge this branch. When we have consensus on merging, I'll merge the `use-external-client-code-removed` branch, which is the same as this, but with ~30kloc of code removed from the repository and has passing tests.

gulpfile.js
```diff
@@ -296,11 +227,6 @@ gulp.task('watch-manifest', function () {
   }));
 });
 
-gulp.task('start-live-reload-server', function () {
-  var LiveReloadServer = require('./scripts/gulp/live-reload-server');
-  liveReloadServer = new LiveReloadServer(3000, 'http://localhost:5000');
-});
-
 gulp.task('build-app',
```
robertknight: Either in this PR or separately it would make sense to replace mentions of `app` with `site`, since we're no longer building the client here.

robertknight:
One trivial suggestion about changing `app` to `site` for task names in the Gulpfile, otherwise LGTM.

nickstenning:
Merged as 89e6cd0..596e671.

PR #3555, opened 2016-06-29
Trigger Angular digest cycle if necessary after Redux action dispatch

robertknight:
In various places where the program receives external inputs (WebSocket messages, messages from other iframes, async promises being resolved) we need to trigger an Angular digest cycle explicitly in order to sync the UI with the app state after the input is processed.

Since most app state changes, including all those relating to the set of loaded annotations and the current selection, now go through Redux, we can use its [middleware functionality](http://redux.js.org/docs/advanced/Middleware.html) for adding logic around state changes to put this in one place, saving the need to sprinkle `$apply` calls and inject `$rootScope` in places that receive external inputs.

We cannot rely on this in the WebSocket code yet because WS messages can update session state which is not yet stored in Redux, but I have added a note that we should remove the `$apply` calls there once it is.

### DIFF
--- a/h/static/scripts/annotation-ui-sync.js
+++ b/h/static/scripts/annotation-ui-sync.js
@@ -49,20 +49,10 @@ function AnnotationUISync($rootScope, $window, bridge, annotationSync,
     }
   };
 
-  // Because the channel events are all outside of the angular framework we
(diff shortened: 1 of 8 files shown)


nickstenning:

PR #3558, opened 2016-06-29
Normalize http/s

chdorner:
Changes when I run the migration locally based off of a database dump from 2016-06-27:

```
run time: ~1h40m

deleted document meta: 800
types:
  238 facebook.*
  195 title
  118 favicon
  112 twitter.*
   62 highwire.*
   59 dc.*
   10 prism.*
    6 reply_to

deleted document uris: 531
types:
  227 rel-canonical
  105 rel-alternate
   89 self-claim
   53 null
   42 rel-shortlink
    5 dc-doi
    5 empty string
    3 highwire-doi
    2 highwire-pdf

merged documents: 148
```

### DIFF
--- a/h/api/models/document.py
+++ b/h/api/models/document.py
@@ -382,7 +382,7 @@ def create_or_update_document_meta(session,
         raise ConcurrentUpdateError('concurrent document meta updates')
 
 
-def merge_documents(session, documents, updated=datetime.now()):
+def merge_documents(session, documents, updated=None):
     """
     Takes a list of documents and merges them together. It returns the new
     master document.
@@ -390,6 +390,9 @@ def merge_documents(session, documents, updated=datetime.now()):
     The support for setting a specific value for the `updated` should only
     be used during the Postgres migration. It should be removed afterwards.
     """
+    if updated is None:
+        updated = datetime.utcnow()
+
     master = documents[0]
     duplicates = documents[1:]
 
--- a/h/api/models/document.py
+++ b/h/api/models/document.py
@@ -57,6 +57,10 @@ class Document(Base, mixins.Timestamps):
         """Find documents by a list of uris."""
         query_uris = [uri_normalize(u) for u in uris]
 
+        # FIXME: remove after httpx normalization is finished
+        query_uris.extend([uri_normalize(u, httpx_normalization=False)
+                           for u in uris])
+
         matching_claims = (
             session.query(DocumentURI)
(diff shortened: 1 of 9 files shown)


h/api/models/document.py
```diff
@@ -364,14 +382,17 @@ def create_or_update_document_meta(session,
         raise ConcurrentUpdateError('concurrent document meta updates')
 
 
-def merge_documents(session, documents, updated=datetime.now()):
```
nickstenning: Oh.

tests/h/api/uri_test.py
```diff
@@ -30,134 +38,134 @@
     ("http%3A%2F%2Fexample.com%2Ffoo%2F", "http%3A%2F%2Fexample.com%2Ffoo%2F"),
 
     # Should leave already-normalized URLs as they are
```
nickstenning: We should probably change the comment here -- as these aren't "already-normalised" URLs any more.

chdorner: I've actually moved up the subdomain case to the top and then removed these two test cases.

h/api/search/query.py
```diff
@@ -177,8 +177,15 @@ def __call__(self, params):
 
         uris = set()
         for query_uri in query_uris:
-            us = [uri.normalize(u)
-                  for u in storage.expand_uri(self.request.db, query_uri)]
+            expanded = storage.expand_uri(self.request.db, query_uri)
+
+            us = [uri.normalize(u, httpx_normalization=True)
```
nickstenning: Maybe remove the kwarg now that it defaults to `True`?

h/api/models/document.py
```diff
@@ -364,14 +382,17 @@ def create_or_update_document_meta(session,
...
+def merge_documents(session, documents, updated=None):
     """
     Takes a list of documents and merges them together. It returns the new
     master document.
 
     The support for setting a specific value for the `updated` should only
     be used during the Postgres migration. It should be removed afterwards.
     """
+    if updated is None:
+        updated = datetime.now()
```
nickstenning: This should be `utcnow()`.

nickstenning:
This looks good. The only place you're using `uri.normalize` in a query settings without having added the `httpx_normalization=False` fallback is the `move-uri` CLI command, but we just won't use that until the migration has completed successfully.

nickstenning:
LGTM.

PR #3562, opened 2016-06-30
Ensure that we close transient streamer DB sessions

nickstenning:
Accessing `socket.request.db` inside a streamer message handler is problematic, as the new session will grab a connection from the engine connection pool and potentially not return it.

The longer-term fix here is to ensure that all database access is mediated by the session created by `process_work_queue` in `h/streamer/streamer.py`, but this is a temporary hack to ensure that we don't exhaust the connection pool when processing lots of websocket messages.


chdorner:
LGTM.

PR #3563, opened 2016-06-30
Centralise matcher objects in tests/common/matchers.py

nickstenning:
A number of tests (including ones I'm currently working on) use matcher objects like those in this commit. Rather than writing them over and over again for every new test module, it makes sense to centralise them.


seanh:
Great pattern

PR #3564, opened 2016-06-30
Add a links service, to allow generating links to annotations

nickstenning:
Generating links to annotations is currently a responsibility of the `h.api.presenters` module, which uses a passed-in Pyramid request object to do much of the heavy lifting.

This commit adds a links service (currently unused) which will replace this functionality which will allow us to remove the need to pass a request object into the presenters.

This, in turn, will allow us to entirely remove the use of a request object from the streamer code -- thus eliminating the possibility that we accidentally use `request.db` (as we currently do), thus opening unintended connections to the database server.

### DIFF
--- a/h/api/__init__.py
+++ b/h/api/__init__.py
@@ -3,6 +3,7 @@
 
 def includeme(config):
     config.include('h.api.eventqueue')
+    config.include('h.api.links')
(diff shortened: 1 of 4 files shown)


chdorner:
LGTM.

PR #3567, opened 2016-06-30
Remove request dependency from presenter classes

nickstenning:
This PR removes the need to pass a Pyramid request object into a presenter class, which will in turn make it possible to remove all need for a request object from the streamer message handling code.

Instead, the AnnotationJSONPresenter and AnnotationJSONLDPresenter classes accept a "links service" argument which they use to generate links.

### DIFF
--- a/h/websocket.py
+++ b/h/websocket.py
@@ -174,11 +174,27 @@ def create_app(global_config, **settings):
     config.include('h.models')
     config.include('h.db')
 
-    # We have to include search to set up the `request.es` property.
+    # We have to include parts of the `h.api` package in order to provide,
+    # among other things:
+    #
+    #   - the links service
+    #   - the default presenters (and their link registrations)
+    #   - the `request.es` property
+    config.include('h.api.links')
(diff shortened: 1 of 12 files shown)


nickstenning:
I'm sorry this is a larger diff. There is probably a way to make this smaller but I'd quite like to fix production as soon as we can.

h/websocket.py
```diff
@@ -174,12 +174,28 @@ def create_app(global_config, **settings):
...
+    # for annotations.
+    config.include('h.links')
+
     # We have to include nipsa to provide the NIPSA service
     config.include('h.nipsa')
 
+    # And finally we add static routes which can be used for URL generation
+    # within the websocket server.
+    config.add_route('annotation', '/a/{id}', static=True)
+    config.add_route('api.annotation', '/api/annotations/{id}', static=True)
```
chdorner: All these additions seem to be a bit out of place, but considering the production implications, we should maybe have a look at this at some later point.

chdorner:
LGTM.

PR #3568, opened 2016-06-30
Update dev install docs

seanh:


seanh:
This is makred as WIP for a couple of reasons:

1. I'm having an issue that once I run the `npm link` commands then t

seanh:
I think the relationship between site and client is pretty unclear in these docs. When you setup a client dev env, localhost:3000 will embed your dev client and the client will always communicate with your deb web service on localhost:5000.

In contrast when you setup a dev web service, it will embed a built copy of the client from npm unless you take additional steps to link it to your dev client.

And then there's the Chrome extension.

But I think we can maybe leave making this all crystal clear for a later time? After all there is still more splitting up of components and evolving these relations to happen.

seanh:
> I'm having an issue that once I run the npm link commands then the web process starts crashing because icomoon.css is not in the client's manifest.json. Running gulp build build-css watch in the client dir fixes it, but I've left that out of the docs 'cos I'm not sure if this is intentional.

This is fixed by https://github.com/hypothesis/client/pull/6

seanh:
Alright, I've split this up into separate pages for installing the website/API and installing the client. This is ready for review now.

nickstenning:
On the basis that we should probably move the docs for the client into the client repository in due course, I kind of think that we should avoid moving the existing install documentation more than necessary.

Can we keep `docs/developing/install.rst` where it is and simply add `docs/developing/install-client.rst` alongside it for now instead of adding the `installing` directory?

docs/developing/installing/targets.rst
```diff
@@ -0,0 +1,3 @@
+.. _Git repo named h: https://github.com/hypothesis/h/
```
nickstenning: This indirection to me seems overly complicated versus just duplicating the links in each document...

seanh: Admittedly the benefit is slim in this case but I think it might be a good pattern to get into the habit of:

1. De-duplication: if the repo URL changed we'd only have to change it in one place, admittedly it seems unlikely that the repo URL is gonna change.

2. Convenience: it allows you to just write:

        `Git repo named h`_

   each time and have it be linked, instead of having to write

        `Git repo named h <https://github.com/hypothesis/h/>`_

   every time. 

3. Consistency in the source files, even though in this case it doesn't seem likely the URL will change, doing it like this is consistent with the same pattern used for other URLs that we link to multiple times as well and I think it's good to get into the habit / demonstrate best practice.

4. Consistency in the output, since this makes it convenient to write

        `Git repo named h`_

   it increases the chances that whenever we link to the repo it'll always be rendered as
   `<a href="...">Git repo named h</a>` which I think is nice for the reader as they know it's
   the same thing being linked to.

seanh:
The thing is that `docs/developing/installing/index.rst` contains its own `toctree` that includes `website.rst` and `client.rst`, and I think it's nice if the directory structure of the docs dir mirrors the toctree structure of the docs. You have a top-level `index.rst` with a `toctree` that includes all t

. I think I could just move `index.rst` (rename it back to `installing.rst`), `website.rst` and `client.rst` back up into `docs/developing/` and leave all the code (including `toctree`) the same,

seanh:
Maybe I should rename the new `installing` dir to `install` though, then URLs won't break...

seanh:
> Maybe I should rename the new installing dir to install though, then URLs won't break...

^ done

nickstenning:
I still think the `targets.rst` file is overkill. If its contents were clearly things referred to throughout the documentation and which were prone to change, then fine. But I'm not really sure needing to refer to `a git repo named X` is as universal as you seem to be implying, and I'm *really* sure that the GitHub URLs of these repositories are not going to mysteriously change.

nickstenning:
Regardless, this is an improvement. Merging.

PR #3570, opened 2016-07-01
Use streamer session in realtime message handlers

nickstenning:
This PR ensures that all database access inside

1. the realtime message handlers in `h/streamer/messages.py`, and
2. the websocket message handler in `h/streamer/websocket.py`

is done using the streamer-managed session created in `process_work_queue`.

This ensures that:

- the underlying connection is returned to the pool after handling each message...
- ...and thus we don't accidentally overflow the database connection pool when handling lots of messages (or a few messages with lots of clients)
- also, the transaction isolation level set in `streamer.py` is respected

In order to do this, the created session is now passed into the message handlers.

Further, and in order to ensure that this problem doesn't recur, I've removed the `request` property from the `h.websocket.WebSocket` instances that represent client connections to the WebSocket. This means that we can't accidentally create a new database connection by accessing `socket.request.db`. Now the socket instance is informed of its auth status through the WSGI environment. This has the nice side effect of removing our last use of `pyramid.threadlocal` from the codebase

### DIFF
--- a/h/streamer/views.py
+++ b/h/streamer/views.py
@@ -8,7 +8,7 @@ from pyramid.view import view_config
 from ws4py.exc import HandshakeError
 from ws4py.server.wsgiutils import WebSocketWSGIApplication
 
-from h.streamer import websocket
+from h.streamer import streamer, websocket
 
 
 @view_config(route_name='ws')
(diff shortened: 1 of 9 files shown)


h/streamer/views.py
```diff
@@ -21,6 +21,19 @@ def websocket_view(request):
...
         if origin != request.host_url and origin not in allowed:
             return httpexceptions.HTTPForbidden()
+
+    # Provide environment which the WebSocket handler can use...
+    request.environ.update({
+        'h.ws.authenticated_userid': request.authenticated_userid,
+        'h.ws.effective_principals': request.effective_principals,
+        'h.ws.registry': request.registry,
+        'h.ws.streamer_work_queue': streamer.WORK_QUEUE,
+    })
```
chdorner: I'm not sure if this is one of the points that you are a bit unhappy about, but I do like this approach. At this point we're handing the connection off to a non-pyramid part of our application, so we make sure to pass the necessary data on instead of _magically_ getting a request.

nickstenning: No, I like this in general, I just don't like having to pass the entire application registry in. Not sure there's a lot we can do about that right now, though.

chdorner:
LGTM.
I did reproduce the timeout on master and can't reproduce it anymore on this branch :+1:

PR #3574, opened 2016-07-04
Optimize streamer DB access

chdorner:
This was an oversight made in #3378 (specifically in 80c54e5) which had
a pretty serious impact on performance. Now that we load the annotation
from the database based on the id in the realtime message we have to
make sure that we only access the database once per message, and not
once per message per socket. Which would mean that with 1000 connected
clients we would access the database 2000 times (once for loading the
annotation, once for checking if the annotation creator is nipsa'd).

This pull requests refactors the topic handler functions to receive a
list of sockets instead of a socket. This new topic handler then loads
the necessary data from the database and then calls a private function
that returns the payload that we send back to the client.

### DIFF
--- a/h/streamer/messages.py
+++ b/h/streamer/messages.py
@@ -1,7 +1,6 @@
 # -*- coding: utf-8 -*-
 
 from collections import namedtuple
-import json
 import logging
 
 from gevent.queue import Full
@@ -67,8 +66,6 @@ def handle_message(message, session, topic_handlers):
     object. It is assumed that there is a 1:1 request-reply mapping between
     incoming messages and messages to be sent out over the websockets.
     """
-    data = message.payload
-
     try:
         handler = topic_handlers[message.topic]
     except KeyError:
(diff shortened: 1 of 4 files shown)


h/streamer/messages.py
```diff
@@ -79,15 +76,43 @@ def handle_message(message, session, topic_handlers):
...
-def handle_annotation_event(message, socket, session):
+def handle_user_event(message, sockets, _):
+    for socket in sockets:
+        reply = _generate_user_event(message, socket)
+        if reply is None:
+            continue
+        socket.send_json(reply)
+
+
+def _generate_annotation_event(message, socket, annotation, user_nipsad):
```
nickstenning: By way of a general observation rather than feedback on this PR specifically:

I think we could clean this up some more by realising that some of the conditions in this function refer only to the message, while some are about the (message, socket) pair. All of the conditions which are solely about the message could move to the top level, outside the socket loop.

nickstenning:
LGTM.

PR #3575, opened 2016-07-04
Remove pre/during httpx migration support code

chdorner:
Now that we migrated the database we don't need to normalise to both
httpx and non-httpx variations anymore.


chdorner:
We ran the migration on production, so this is good to go now.

nickstenning:
LGTM.

PR #3576, opened 2016-07-05
Allow either "signed-out" to "logged-out" auth status

nickstenning:
While the client transitions from "signed-out" to "logged-out", we simplify the deployment process by simply recognising both/either in app.html.

### DIFF
--- a/h/templates/app.html.jinja2
+++ b/h/templates/app.html.jinja2
@@ -36,7 +36,13 @@
       on-change-sort-key="setSortKey(sortKey)">
     </top-bar>
 
-    <div class="create-account-banner" ng-if="isSidebar && auth.status === 'signed-out'" ng-cloak>
+    {#
+      The client is transitioning from using "signed-out" as an auth status to
+      "logged-out". For the moment we recognise either.
+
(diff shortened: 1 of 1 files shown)


h/templates/app.html.jinja2
```diff
@@ -36,7 +36,13 @@
...
     </top-bar>
 
-    <div class="create-account-banner" ng-if="isSidebar && auth.status === 'signed-out'" ng-cloak>
+    {#
+      The client is transitioning from using "signed-out" as an auth status to
+      "logged-out". For the moment we recognise either.
+
+      TODO: remove the "signed-out" option when the client no longer uses it.
+    #}
+    <div class="create-account-banner" ng-if="isSidebar && (auth.status === 'signed-out' || auth.status === 'logged-out')" ng-cloak>
```
robertknight: There is a reference to "sign in" in the message text here which needs to be updated

robertknight:
LGTM

h/templates/app.html.jinja2
robertknight: @nickstenning Pointed out that this was fixed by a separate PR

PR #3577, opened 2016-07-05
Remove client templates

nickstenning:
These are no longer needed here: they have moved to the `client` repository.


robertknight:
LGTM

PR #3578, opened 2016-07-05
Update copy, comments from "sign in/out" to "log in/out"

nickstenning:
In coordination with hypothesis/client#18, this PR updates copy in the rest of the application to consistently use "log {in,out}" rather than "sign {in,out}".


robertknight:
LGTM

PR #3579, opened 2016-07-05
Remove referer from groups page links

seanh:


seanh:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;Wk6Y59sK&#x2F;388-remove-referer-header-from-link-on-groups-page"> Remove &quot;Referer&quot; header from link on groups page</a>

finitud:
Perhaps it'd be worth adding a test case to check this bit of metadata being set properly?

nickstenning:
Ah! Of course! Nice find.

So, using [Referrer Policy](https://www.w3.org/TR/referrer-policy/#referrer-policies) is a great way to deal with this, but I've just done a bit of research and unfortunately (as with so many things on the web) there are [multiple different versions of the standard in use](http://caniuse.com/#feat=referrer-policy).

The `never` value is from [the old standard](https://wiki.whatwg.org/wiki/Meta_referrer), supported by Edge and Safari. The [new standard](https://www.w3.org/TR/referrer-policy/) has [a backwards compatibility clause](https://www.w3.org/TR/referrer-policy/#determine-policy-for-token), but it also recommends that "Authors are encouraged to avoid the legacy keywords never, default, and always."

I'd suggest we use the `origin` policy instead, which has the same meaning in both old and new standards, and has the advantage of still surfacing "https://hypothes.is" in the analytics of linked sites.

Lastly, it looks like it should be possible to use the `Referrer-Policy` to communicate a policy to the browser. This is probably preferable to tying the fix to the particular templates, and would allow us to implement a "referrer policy" view deriver that wasn't tied to a particular view function.

seanh:
Ah, I used the old version of the spec deliberately for wider browser support (see commit message), but I missed the "origin" option in the old space and the admonition against "never" in the new spec. "origin" is better, I've changed it to that.

I forgot to mention it in the commit message, but the reason that I went for a meta tag instead of an HTTP header was also browser support. I don't think any of Chrome, Firefox, Safari or Edge support the `Referrer-Policy` header. I've added this to the list of rejected solutions in the commit message.

seanh:
I'm not sure how I could write a test for this.

I could test that the group read view returns the correct `meta_attrs` in the dict that it returns. It's not much of a test though.

The only reason that this ends up putting the correct `<meta` tag in is that the code in the `meta_attrs` loop in `base.html.jinja2` does it, and the `meta_attrs` that the view returns matches the format that the base template layout expects.

So it would have to be a "functional" test that actually simulates browsing to the URL and rendering the templates, and then looks for the `<meta` tag in the resulting HTML.

seanh:
Ended up writing a functional test. I had to add a `Group` factory and give the functional tests a copy of the `factories` fixture. Also had to write code to login in functional tests, which turns out to be quite easy, but we may want to put that in a helper function or fixture at some point, once multiple functional tests need to login.

tests/functional/conftest.py
```diff
@@ -5,6 +5,8 @@
 import pytest
 from webtest import TestApp
 
+import tests.common.factories
```
seanh: The other copies of the `factories` fixture do `from ..common import factories`. This won't work in `tests/functional/` because it's not a package, and also aren't relative imports bad? The others also put the import inside the function, which I guess was to avoid a name conflict between the imported module and the fixture function itself? Just importing the module as `tests.common.factories` while verbose avoids the conflict.

nickstenning: You can just make `tests/functional/` a package if you want.

> aren't relative imports bad anyway?

No? The form of "relative import" that can cause all sorts of problems is the _implicit_ relative import (i.e. `import foo` where `foo` is a subpackage of the current package), not _explicit_ relative imports, which are absolutely fine.

Putting the import inside the fixture function is absolutely fine, though. Yes, it helps avoid a name conflict, but it also means we don't actually import the factories module unless somebody uses the factories fixture. In application code this pattern should be used sparingly (IMO) but in test code it's absolutely fine, and can actually lead to a cleaner `conftest.py` (which typically imports all kinds of different modules in order to support various fixtures).

nickstenning:
LGTM.

PR #3581, opened 2016-07-06
Add new user profile columns

chdorner:
These include:

* Display name
* Description
* Location
* Link (column is called `uri` for consistency with other models)
* ORCID

As discussed on Slack, the validation of the `uri` and `orcid` properties should be done in the form schemas.

### DIFF
--- a/h/accounts/models.py
+++ b/h/accounts/models.py
@@ -59,20 +59,36 @@ class User(Base):
 
     id = sa.Column(sa.Integer, autoincrement=True, primary_key=True)
 
-    # Normalised user identifier
+    #: Normalised user identifier
     uid = sa.Column(sa.UnicodeText(), nullable=False, unique=True)
-    # Username as chosen by the user on registration
+    #: Username as chosen by the user on registration
     _username = sa.Column('username',
                           sa.UnicodeText(),
                           nullable=False,
                           unique=True)
 
+    #: The display name which will be used when rendering an annotation.
+    display_name = sa.Column(sa.UnicodeText())
+
+    #: A short user description/bio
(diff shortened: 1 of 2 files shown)


chdorner:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;ffgjL81D&#x2F;393-user-profile-fields"> User profile fields</a>

h/accounts/models.py
```diff
@@ -67,6 +67,16 @@ class User(Base):
                           nullable=False,
                           unique=True)
 
+    display_name = sa.Column('display_name', sa.UnicodeText())
```
nickstenning: No need to duplicate the column name. `sa.Column(sa.UnicodeText())` would be fine.

chdorner: sure, stupid copy + pasta oversight

nickstenning:
It would be nice to have a line of documentation for each of these, c.f. `h/api/models/annotation.py`.

h/accounts/models.py
```diff
@@ -67,6 +67,21 @@ class User(Base):
                           nullable=False,
                           unique=True)
 
+    # The display name which will be used when rendering an annotation.
```
nickstenning: Sorry to be such a pain. The colon after the comment character (`#:`) in the equivalent comments in `annotation.py` are important, because they allow the attribute to be [documented by Sphinx automatically](http://www.sphinx-doc.org/en/stable/ext/autodoc.html#directive-autoattribute). We're not generating documentation for these modules at the moment, but I'd like doc comments like this to be visible when we do.

nickstenning:
LGTM.

PR #3583, opened 2016-07-06
Update PhantomJS

robertknight:
This PR updates PhantomJS and other JS test infrastructure to current versions and removes the `console.log()` ban in tests which was also recently removed from the client repository.

See hypothesis/client#16, hypothesis/client#17, hypothesis/client#23


nickstenning:
LGTM.

PR #3584, opened 2016-07-06
Add an admin form action to rename a user

robertknight:
As a step towards making renaming users (a very common support request) easier, this adds an action to the users admin form to rename users.

In the process, I came across a few other issues that needed to be fixed:

* The `models.User.username` setter only validated the username length, not the character set
* The admin user dashboard could display an out-of-date annotation count if recent changes were still being indexed (TBD: Is this understanding correct?)

----

* [x] Fail more gracefully if new username is already taken by another account
* [x] Fix issue where annotation count resets to zero on initial page load after username is changed
* [x] Fix issue where only username length restrictions are applied to the username, not other validation rules (eg. allowed characters)
* [x] ~~Fix issue where changing a user's name resets the last-updated timestamp of all their annotations to the current time~~ (already fixed with #3586)
* [x] As a separate PR, use the DB rather than ES to lookup a user's annotations in order to delete them, replacing `_all_user_annotations_query`?
* [x] Add tests for renaming a user
* [x] Move `delete_user()` and `rename_user()` functions to service layer? (Just added a new `RenameUserService`

### DIFF
--- a/h/admin/services/user.py
+++ b/h/admin/services/user.py
@@ -28,21 +28,16 @@ class RenameUserService(object):
         self.request = request
         self.session = request.db
 
-    def check(self, user_id, new_username):
-        user = self.session.query(models.User).get(user_id)
-        if user is None:
-            raise UserRenameError('Could not find existing user with id "%d"' % user_id)
-
+    def check(self, new_username):
         existing_user = models.User.get_by_username(self.session, new_username)
         if existing_user:
             raise UserRenameError('Another user already has the username "%s"' % new_username)
 
         return True
 
-    def rename(self, user_id, new_username):
-        self.check(user_id, new_username)
+    def rename(self, user, new_username):
+        self.check(new_username)
 
-        user = self.session.query(models.User).get(user_id)
         old_username = user.username
 
         user.username = new_username
--- a/h/admin/services/user.py
+++ b/h/admin/services/user.py
@@ -1,0 +1,74 @@
+# -*- coding: utf-8 -*-
+
+from __future__ import unicode_literals
+
+from h import models
+from h.api.search.index import BatchIndexer
+from h.util.user import userid_from_username
+
+
+class UserRenameError(Exception):
+    pass
+
+
+class RenameUserService(object):
+    """
+    Renames a user and updates all its annotations and when necessary the NIPSA
+    status.
+
+    ``check`` should be called first
+
+    Validates the new username and updates the User. The permissions of
+    the user's annotations are updated to reflect the new username.
+
+    May raise a ValueError if the new username does not validate or
+    UserRenameError if the new username is already taken by another account.
+    """
+    def __init__(self, request):
+        self.request = request
+        self.session = request.db
+
+    def check(self, old_username, new_username):
+        user = models.User.get_by_username(self.session, old_username)
+        if user is None:
+            raise UserRenameError('Could not find existing user with username "%s"' % old_username)
+
+        existing_user = models.User.get_by_username(self.session, new_username)
+        if existing_user:
+            raise UserRenameError('Another user already has the username "%s"' % new_username)
+
+    def rename(self, old_username, new_username):
+        self.check(old_username, new_username)
+
+        user = models.User.get_by_username(self.session, old_username)
+
+        user.username = new_username
+        ids = self._change_annotations(old_username, new_username)
+        self.request.tm.commit()
+
+        self._reindex_annotations(ids)
+
+    def _change_annotations(self, old_username, new_username):
+        new_userid = userid_from_username(new_username, self.request.auth_domain)
+        annotations = self._fetch_annotations(old_username)
+
+        ids = set()
+        for annotation in annotations:
+            annotation.userid = new_userid
+            ids.add(annotation.id)
+
+        return ids
+
+    def _reindex_annotations(self, ids):
+        indexer = BatchIndexer(self.request.db, self.request.es, self.request)
+        indexer.index(ids)
+
+    def _fetch_annotations(self, username):
+        userid = userid_from_username(username, self.request.auth_domain)
+        return self.session.query(models.Annotation).filter(
+            models.Annotation.userid == userid).yield_per(100)
+
+
+def rename_user_factory(context, request):
+    """Return a RenameUserService instance for the passed context and request."""
+    return RenameUserService(request)
--- a/h/admin/services/user.py
+++ b/h/admin/services/user.py
@@ -63,8 +63,9 @@ class RenameUserService(object):
         return ids
 
     def _reindex_annotations(self, ids):
-        indexer = index.BatchIndexer(self.request.db, self.request.es, self.request)
-        indexer.index(ids)
+        if ids:
+            indexer = index.BatchIndexer(self.request.db, self.request.es, self.request)
+            indexer.index(ids)
 
     def _fetch_annotations(self, username):
         userid = userid_from_username(username, self.request.auth_domain)
--- a/h/admin/services/user.py
+++ b/h/admin/services/user.py
@@ -3,7 +3,7 @@
 from __future__ import unicode_literals
 
 from h import models
-from h.api.search.index import BatchIndexer
+from h.api.search import index
 from h.util.user import userid_from_username
 
 
@@ -28,19 +28,22 @@ class RenameUserService(object):
         self.request = request
         self.session = request.db
 
-    def check(self, old_username, new_username):
-        user = models.User.get_by_username(self.session, old_username)
+    def check(self, user_id, new_username):
+        user = self.session.query(models.User).get(user_id)
         if user is None:
-            raise UserRenameError('Could not find existing user with username "%s"' % old_username)
+            raise UserRenameError('Could not find existing user with id "%d"' % user_id)
 
         existing_user = models.User.get_by_username(self.session, new_username)
         if existing_user:
             raise UserRenameError('Another user already has the username "%s"' % new_username)
 
-    def rename(self, old_username, new_username):
-        self.check(old_username, new_username)
+        return True
 
-        user = models.User.get_by_username(self.session, old_username)
+    def rename(self, user_id, new_username):
+        self.check(user_id, new_username)
+
+        user = self.session.query(models.User).get(user_id)
+        old_username = user.username
 
         user.username = new_username
         ids = self._change_annotations(old_username, new_username)
@@ -60,7 +63,7 @@ class RenameUserService(object):
         return ids
 
     def _reindex_annotations(self, ids):
-        indexer = BatchIndexer(self.request.db, self.request.es, self.request)
+        indexer = index.BatchIndexer(self.request.db, self.request.es, self.request)
         indexer.index(ids)
 
     def _fetch_annotations(self, username):
--- a/h/admin/services/user.py
+++ b/h/admin/services/user.py
@@ -13,20 +13,22 @@ class UserRenameError(Exception):
 
 class RenameUserService(object):
     """
-    Renames a user and updates all its annotations and when necessary the NIPSA
-    status.
+    Renames a user and updates all its annotations.
 
     ``check`` should be called first
 
-    Validates the new username and updates the User. The permissions of
-    the user's annotations are updated to reflect the new username.
+    Validates the new username and updates the User. The user's annotations
+    userid field will be updated. It accepts a reindex function that gets a
+    list of annotation ids, it is then the function's responsibility to reindex
+    these annotations in the search index.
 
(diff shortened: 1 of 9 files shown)


h/admin/views/users.py
```diff
@@ -69,23 +65,59 @@ def users_activate(request):
...
+        # the character set or any other restrictions.
+        user.username = new_username
+
+        # FIXME (2) - We need to trigger re-indexing of any of the user's
+        # annotations here and ensure that annotation permissions are updated
+        # as a result
+
+        # FIXME (3) - Is there already a service layer that this logic can go
+        # into? If not, should we create one and move the user
+        # rename/activate/delete logic there now?
```
chdorner: All of the business logic should probably run in a background worker. We can ensure that the user's username change and the annotations userid change happens at the same time (yay for transactional databases!), and then reindex their annotations (`h.api.search.index.BatchIndexer` can take a list of ids).
As with our current manual solution, there will be a certain time period when searching for the new username will not result in any, or not all annotations of the users, until the search index has been updated.

The problem with putting everything in a background worker is that we can't easily communicate when it is done since it depends on how much is already in the celery queue.

The code itself can still live in a service, but should be called in a celery worker in my opinion, to ensure that we don't run into any timeouts that could leave the index in a bad state (the database should be fine, again, yay for transactions!).

chdorner: Oh, and since Elasticsearch isn't transactional it would be best to change the SQL models, then commit the transaction, and then call the batch indexer. In this case we wouldn't reindex anything when the SQL part fails for whatever reason and minimises the chance of ending up in a bad state.

robertknight:
Following a discussion on Slack, the current plan is to:

1. Perform validation of the new username as part of the request
2. Move the activity of actually renaming the user to a worker so that the search index will not be in an inconsistent state if the whole process takes longer than the request timeout

h/admin/services/user.py
```diff
@@ -0,0 +1,74 @@
...
+from h import models
+from h.api.search.index import BatchIndexer
+from h.util.user import userid_from_username
+
+
+class UserRenameError(Exception):
+    pass
+
+
+class RenameUserService(object):
```
robertknight: Ah, OK. I think I was originally imagining a service with a slightly broader scope that handled all user admin actions. So the way I was originally intending to approach this was 1) Create a `UserAdminService` and move the existing logic for handling user actions here in one PR, 2) Add the rename user functionality in a separate PR.

If you approach this as you have done here, adding a service to rename users first, would you later create a separate service for the different user admin tasks or rename this service and broaden its scope to include other tasks and move the logic for `admin_users_delete` into it?

chdorner: The main reason why I went for a `RenameUserService` rather than a `UserAdminService` was that I felt that the `rename_user` code is complex enough to warrant being split out into a bunch of methods for better readability.

I don't think there's a need right now to move the `delete_user` function somewhere else (given this is work wasn't planned to happen and we are quite busy), but once we are doing that, then we should think about it again and maybe move some stuff around. What do you think?

robertknight: That sounds reasonable to me.

h/admin/services/user.py
```diff
@@ -0,0 +1,74 @@
...
+    def check(self, old_username, new_username):
+        user = models.User.get_by_username(self.session, old_username)
+        if user is None:
+            raise UserRenameError('Could not find existing user with username "%s"' % old_username)
+
+        existing_user = models.User.get_by_username(self.session, new_username)
+        if existing_user:
+            raise UserRenameError('Another user already has the username "%s"' % new_username)
+
+    def rename(self, old_username, new_username):
```
robertknight: Perhaps the first argument here should be the user's integer ID rather than their username.

This is an extreme edge case, but in theory if three tasks to perform a rename were scheduled concurrently (A -> B, A -> C, D->A) this could end up resulting in (A -> B, D -> C) where D -> C is not one of the attempted renames.

h/admin/views/users.py
```diff
@@ -69,23 +63,56 @@ def users_activate(request):
...
+
+        worker.rename_user.delay(old_username, new_username)
+
+        request.session.flash(
+            'The user "%s" will be renamed to "%s" in the backgroud. Refresh this page to see if it\'s already done' %
+            (old_username, new_username), 'success')
+
+        return httpexceptions.HTTPFound(
+            location=request.route_path('admin_users',
+                                        _query=(('username', new_username),)))
```
robertknight: If the user rename operation is asynchronous then presumably the new username might not exist when the client tries to load the redirected page?

chdorner: Yes, which is why I changed the flash message to something like "it is done in the background, reload this page to see changes". Which isn't great, but should do the job for now. It is an admin panel after all ;)

h/admin/services/user.py
chdorner: Done in 91df9b27826af7cf90d72dc557c56f56448bdd9d

h/admin/services/user.py
```diff
@@ -0,0 +1,78 @@
+# -*- coding: utf-8 -*-
```
nickstenning: This seems like a slightly odd place to put this to me. It can stay here for now, but in general I think services should live alongside the models that they are (principally) involved with manipulating, i.e. in `h.accounts.services` in this case.

tests/h/admin/views/users_test.py
```diff
@@ -40,25 +42,17 @@ def test_users_index_looks_up_users_by_email(User, pyramid_request):
...
-                'query': {'match_all': {}}
-            }
-        }
-    }
-    es.conn.count.assert_called_with(index=es.index,
-                                     doc_type=es.t.annotation,
-                                     body=expected_query)
+    pyramid_request.params = {"username": "bob"}
+    result = views.users_index(pyramid_request)
+    print(result)
```
robertknight: Left in by mistake?

h/admin/services/user.py
```diff
@@ -0,0 +1,78 @@
...
+
+    ``check`` should be called first
+
+    Validates the new username and updates the User. The permissions of
+    the user's annotations are updated to reflect the new username.
+
+    May raise a ValueError if the new username does not validate or
+    UserRenameError if the new username is already taken by another account.
+    """
+    def __init__(self, request):
```
nickstenning: It doesn't make a lot of sense to me to pass a request object into a service. Pretty much the whole point of the service factory is to make it possible for the service to *not* require a request object, and pretty much the whole point of the service pattern is to have a layer that is *not* exposed to the internals of Pyramid.

It looks like this is ultimately because you need to construct a `BatchIndexer`. Eventually I don't think `BatchIndexer` should need a request object, but for now perhaps another way of doing this might be as follows:

```python
def rename_user_factory(context, request):
    def _reindex(ids):
        if not ids:
            return
        request.tm.commit()
        indexer = index.BatchIndexer(request.db, request.es, request)
        indexer.index(ids)
    return RenameUserService(session=request.db, auth_domain=request.auth_domain, reindex=_reindex)
```

h/admin/views/users.py
robertknight: This will do for now, but if there is a way we can be smarter about this and provide a better experience that would be great.

As an example I could just throw out there - most rename operations will take less than the 5 second request threshold. Could we wait for the operation to complete up to some time threshold and tell the user it succeeded if it finishes within that time, or show a "your username is being changed" message otherwise?

nickstenning: This isn't about providing a user with the ability to rename their own account. This is an admin interface.

(And to answer your question, no. Any request that takes even 1 second is blocking other requests, so that's not an acceptable way to handle these kinds of changes.)

robertknight:
Testing this locally with a basic user who has just a few annotations I encountered a couple of issues:

1. After renaming a user account and trying to create a new group as that user, I get this error:

```
ProgrammingError: (psycopg2.ProgrammingError) column user_group.id does not exist
LINE 1: ...oup (user_id, group_id) VALUES (2, 110) RETURNING user_group...
                                                             ^
 [SQL: 'INSERT INTO user_group (user_id, group_id) VALUES (%(user_id)s, %(group_id)s) RETURNING user_group.id'] [parameters: {'group_id': 110, 'user_id': 2}]
```

2. After renaming the user account, clients which are logged in as that user get logged out. This could potentially be surprising for users. More importantly though, I don't see explicit logic to do this in the code, so I wondered why it was happening?

h/admin/views/users.py
```diff
@@ -69,23 +63,56 @@ def users_activate(request):
...
+    user = _form_request_user(request)
 
     if user is None:
+        return httpexceptions.HTTPFound(
+            location=request.route_path('admin_users'))
+
+    try:
+        delete_user(request, user)
         request.session.flash(
-            'Cannot find user with username %s' % username, 'error')
```
nickstenning: Did you intend to remove this?

h/admin/views/users.py
```diff
@@ -69,23 +63,56 @@ def users_activate(request):
...
+
+        request.session.flash(
+            'The user "%s" will be renamed to "%s" in the backgroud. Refresh this page to see if it\'s already done' %
+            (old_username, new_username), 'success')
+
+        return httpexceptions.HTTPFound(
+            location=request.route_path('admin_users',
+                                        _query=(('username', new_username),)))
+
+    except (UserRenameError, ValueError) as e:
```
nickstenning: What raises a `ValueError` here?

h/admin/views/users.py
nickstenning: Oh, I see, it's moved into the `_form_request_user` function.

Perhaps these view functions would be cleaner if `_form_request_user` threw a named exception if the user wasn't found, and then you added an exception view to handle that scenario. See `h/accounts/views.py` for examples of exception views.

h/admin/services/user.py
```diff
@@ -0,0 +1,78 @@
...
+
+
+class RenameUserService(object):
+    """
+    Renames a user and updates all its annotations and when necessary the NIPSA
+    status.
+
+    ``check`` should be called first
+
+    Validates the new username and updates the User. The permissions of
```
nickstenning: The permissions? I think it would be better to say `userid`, as that's what's actually being updated.

robertknight:
>  After renaming a user account and trying to create a new group as that user, I get this error:

I think this is a false alarm, I see this with a brand new account that has not been renamed as well.

nickstenning:
> I think this is a false alarm, I see this with a brand new account that has not been renamed as well.

You haven't run the latest migrations.

robertknight:
> You haven't run the latest migrations.

Indeed. Both Sheetal and I got caught out by this at various points recently. Is there an easy way we could check this when `make dev` is run and either run the migration automatically or die with an error?

nickstenning:
In addition to my feedback inline, I have one overall observation about the interface of the `RenameUserService`.

Because the `RenameUserService` functions accept user ids, you're having to fetch the relevant user twice, which isn't ideal, as that then means you have to communicate and handle error conditions twice, too.

The clearest way to write this, IMO, is to leave the responsibility for retrieving the relevant `User` object to the views code, and pass the whole user object in where appropriate. Your views would end up looking like this:

```python
def users_rename(request):
    user = _form_request_user(request)  # throws an UserNotFound exception that is handled by an exception view
    new_username = request.params.get('new_username')

    svc = request.find_service(name='rename_user')
    svc.check_username_available(new_username)  # throws a UsernameNotAvailable exception that is handled by an exception view
    svc.rename_async(user, new_username)  # does not throw (as it simply dispatches a Celery task)

    return httpexceptions.HTTPFound(location=request.route_path('admin_users', ...))
```

That is, all error-path code moves into exception views, and the view doesn't have to do anything other than call service methods. In particular, I'd have the service responsible for dispatching the relevant tasks in the background (c.f. the `NipsaService`), not the view.

h/admin/views/users.py
nickstenning: I worked it out eventually, but see later comments.

h/admin/services/user.py
chdorner: I created it as an admin service specifically to signal that the implementation works but should probably be revisited before we move this out into a `h.accounts.services` and give this ability to users to rename their own account. For all the reasons we already touched on in the comments of this PR (it's at the moment impossible to show any progress to the user, etc.)

chdorner:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;dg5QqFp6&#x2F;397-add-rename-user-functionality-to-admin-dashboard"> Add &#39;Rename user&#39; functionality to admin dashboard</a>

nickstenning:
Still some rough edges, but for a piece of work we weren't even scheduled to do it looks pretty good 😉.

PR #3585, opened 2016-07-06
Escape user input in Deform forms

seanh:
Replace deform_jinja2.jinja2_renderer_factory() with our own Deform
Jinja2 renderer function which uses our existing Jinja2 environment
(which escapes user input because it has auto-escaping enabled).'

deform_jinja2.jinja2_renderer_factory() does not escape user text
rendered in form fields - it doesn't turn on Jinja2's autoescaping, and
there's no way for a caller of this function to turn on autoescaping.

Rather than manually escape every instance of user text in our Deform
Jinja2 templates, replace deform_jinja2.jinja2_renderer_factory() with
our own renderer function that uses our existing Jinja2 environment that
has autoescaping enabled.

### DIFF
--- a/h/form.py
+++ b/h/form.py
@@ -7,6 +7,10 @@ form templates in preference to the defaults. Uses `deform_jinja2` to provide
 the fallback templates in Jinja2 format, which we can then extend and modify as
 necessary.
 """
+import deform
+import jinja2
+import pyramid_jinja2
+from pyramid.path import AssetResolver
 
 
 SEARCH_PATHS = (
@@ -15,16 +19,53 @@ SEARCH_PATHS = (
 )
 
 
-def init():
-    from deform import Form
-    from deform_jinja2 import jinja2_renderer_factory
-    from deform_jinja2.translator import PyramidTranslator
+class Jinja2Renderer(object):
+    """
+    An alternate Deform renderer that uses Jinja2 to render templates.
 
-    renderer = jinja2_renderer_factory(search_paths=SEARCH_PATHS,
-                                       translator=PyramidTranslator())
+    We supply our own Jinja2 renderer because deform_jinja2's one doesn't
+    support autoescaping.
+    """
 
-    Form.set_default_renderer(renderer)
+    def __init__(self, base_env):
+        """
+        Return a new callable Jinja2Renderer object.
+
+        :param base_env: the initial Jinja2 environment that this renderer's
+            Jinja2 environment will be created as an overlay of
+        :type base_env: :py:class:`jinja2.Environment`
+        """
+        self._base_env = base_env
+        self._env = None
+
+    def _get_env(self):
+        """
+        Return this renderer's Jinja2 environment.
+
+        Initialize the environment if it hasn't been initialized already.
+        """
+        if not self._env:
+            resolver = AssetResolver()
+            searchpath = [resolver.resolve(path).abspath()
+                          for path in SEARCH_PATHS]
+            loader = pyramid_jinja2.SmartAssetSpecLoader(searchpath)
+            self._env = self._base_env.overlay(loader=loader)
+        return self._env
+
+    def __call__(self, template_name, **kwargs):
+        """Return the given template rendered with the given keyword args."""
+        if not template_name.endswith('.jinja2'):
+            template_name += '.jinja2'
+
+        template = self._get_env().get_template(template_name)
+
+        return jinja2.Markup(template.render(**kwargs))
 
 
-def includeme(_):
-    init()
+def initialize_deform_renderer(config):
+    deform.Form.set_default_renderer(
+        Jinja2Renderer(config.get_jinja2_environment()))
+
+
+def includeme(config):
+    config.action(None, initialize_deform_renderer, args=[config])
--- a/h/form.py
+++ b/h/form.py
@@ -49,7 +49,7 @@ class Jinja2Renderer(object):
             searchpath = [resolver.resolve(path).abspath()
                           for path in SEARCH_PATHS]
             loader = pyramid_jinja2.SmartAssetSpecLoader(searchpath)
-            self._env = self._base_env.overlay(loader=loader)
+            self._env = self._base_env.overlay(autoescape=True, loader=loader)
         return self._env
 
     def __call__(self, template_name, **kwargs):
--- a/tests/h/conftest.py
+++ b/tests/h/conftest.py
@@ -116,8 +116,8 @@ def db_session(db_engine):
 
 @pytest.fixture(scope='session', autouse=True)
 def deform():
-    """Allow tests that use deform to find our custom templates."""
-    form.init()
+    """Use our custom Deform renderer instead of Deform's default one."""
+    form.initialize_deform_renderer(config=mock.Mock())
 
 
 @pytest.yield_fixture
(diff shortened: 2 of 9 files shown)


h/form.py
```diff
@@ -15,16 +20,26 @@
...
-    from deform_jinja2.translator import PyramidTranslator
+# We supply our own Jinja2 renderer factory because deform_jinja2's one doesn't
+# support autoescaping.
+def jinja2_renderer_factory(config):
+
+    def jinja2_renderer(template_name, **kwargs):
+        searchpath = [pkg_resources.resource_filename(*(path.split(':')))
+                      for path in SEARCH_PATHS]
+
+        environment = config.get_jinja2_environment().overlay(
```
nickstenning: You're still creating a new overlay environment for every single template render call.

On approach to fixing this would be to move this up into the body of the `jinja2_renderer_factory` function and close over it.

seanh: Done. Wish Python 2.7 had `nonlocal`

h/form.py
```diff
@@ -15,16 +20,29 @@
...
-    from deform_jinja2 import jinja2_renderer_factory
-    from deform_jinja2.translator import PyramidTranslator
+# We supply our own Jinja2 renderer factory because deform_jinja2's one doesn't
+# support autoescaping.
+def jinja2_renderer_factory(config):
+    environment = []
+
+    def jinja2_renderer(template_name, **kwargs):
+        if not environment:
+            searchpath = [pkg_resources.resource_filename(*(path.split(':')))
```
nickstenning: We should probably be using `pyramid.path.AssetResolver` to do this rather than this hacky version from `deform_jinja2`.

Something like:

```python
from pyramid.path import AssetResolver

resolver = AssetResolver()

...
    searchpath = [resolver.resolve(p).abspath() for p in SEARCH_PATHS]
```

h/form.py
```diff
@@ -15,16 +20,29 @@
...
+                          for path in SEARCH_PATHS]
+
+            environment.append(config.get_jinja2_environment().overlay(
+                loader=pyramid_jinja2.SmartAssetSpecLoader(
+                    searchpath=searchpath)))
+
+        if '.jinja2' not in template_name:
```
nickstenning: `not template_name.endswith('.jinja2')` would probably be a more accurate test at this point.

h/form.py
```diff
@@ -15,16 +20,29 @@
...
+            template_name += '.jinja2'
+
+        template = environment[0].get_template(template_name)
```
nickstenning: I was a bit confused by why you're storing the environment in a list... but I guess you're working around the fact that the jinja2 environment might not exist when `includeme` is called.

It would be more conventional to use `config.action` much as it's used in `app.py` to ensure that this is set up after the configuration stage of the application boot.

h/form.py
```diff
@@ -15,16 +20,29 @@
 )
 
 
-def init():
-    from deform import Form
-    from deform_jinja2 import jinja2_renderer_factory
-    from deform_jinja2.translator import PyramidTranslator
+# We supply our own Jinja2 renderer factory because deform_jinja2's one doesn't
+# support autoescaping.
+def jinja2_renderer_factory(config):
```
nickstenning: The `jinja2_renderer_factory` in `deform_jinja2` is a class, instances of which are callable. This is a nice pattern for factories which return callables, and avoids some of the ugliness of closures and the stuff you've had to do here with `environment = []`.

I'd suggest adopting the same pattern, and passing in the parent environment and a list of search paths rather than the whole config and relying on a global. This will make it much easier to test the behaviour of the code.

For example:

```python
class RendererFactory(object):
    def __init__(self, parent_env, search_paths=()):
        ...
        self.env = ...

    def __call__(self, template_name, **kwargs):
        template = self.env.get_template(...)
        ...
```

tests/h/conftest.py
```diff
@@ -117,7 +117,7 @@ def restart_savepoint(session, transaction):
 @pytest.fixture(scope='session', autouse=True)
 def deform():
     """Allow tests that use deform to find our custom templates."""
-    form.init()
+    form.includeme(config=mock.Mock())
```
nickstenning: Hmm. I guess this doesn't cause tests to fail, but it certainly seems a little odd, given the docstring.

nickstenning:
This is probably on your list already but we should probably test the renderer factory, and I've made a couple of suggestions which hopefully will make testing it easier.

h/form.py
seanh: Done

h/form.py
seanh: For some reason it was written this way in `deform_jinja2`, but there doesn't seem to be any reason not to change it, done

h/form.py
seanh: Done

h/form.py
seanh: Ok, I've replaced this with a class

tests/h/conftest.py
seanh: I've corrected the docstring, which I think was slightly off. What calling this function actually does is register our custom deform renderer so that it gets used instead of deform's default chameleon template renderer (which crashes if the tests use it), so yes it means that deform can "find our custom templates" but what it really means is that deform will use our renderer not its default one.

I haven't changed the name of this fixture but I think `deform` is an odd name for it, it's not for example patching deform with a mock and returning the mock, a better name for it would probably be `DeformJinja2Renderer`.

It seems to be fine to pass in a mock in place of the configurator. Which will mean that our deform jinja2 renderer ends up with a mock object as its jinja2 env, mock objects as templates, `template.render()` will do nothing and return a mock ... Is it worth looking into passing in a real test configurator instead, so that our templates actually get rendered? I suppose they were actually being rendered before.

seanh: This works:

```diff
diff --git a/tests/h/conftest.py b/tests/h/conftest.py
index 6cbe58e..d41d184 100644
--- a/tests/h/conftest.py
+++ b/tests/h/conftest.py
@@ -11,6 +11,7 @@ import os
 import mock
 import pytest
 
+import jinja2
 import sqlalchemy
 from pyramid import testing
 from pyramid.request import apply_request_extensions
@@ -117,7 +118,15 @@ def db_session(db_engine):
 @pytest.fixture(scope='session', autouse=True)
 def deform():
     """Use our custom Deform renderer instead of Deform's default one."""
-    form.initialize_deform_renderer(config=mock.Mock())
+    config = mock.Mock()  # Mock Pyramid Configurator object.
+
+    # Make a jinja2 environment to stand in for the app's main jinja2 env
+    # in tests that call deform.
+    env = jinja2.Environment()
+    env.globals['_'] = lambda s: s  # Our templates expect an _() function.
+    config.get_jinja2_environment.return_value = env
+
+    form.initialize_deform_renderer(config=config)
 
 
 @pytest.yield_fixture
```

Accounts views tests will now crash if our deform templates contain invalid jinja2 for example, because they actually render the templates.

seanh:
I've added tests.

I've removed all the necessary `| safe`s and manually tested all the forms.

Let me know if you want me to [make the accounts views tests actually render our deform jinja2 templates](https://github.com/hypothesis/h/pull/3585#discussion_r69895368) or if you want to leave that mocked.

h/form.py
```diff
@@ -15,16 +19,53 @@
...
+        if not template_name.endswith('.jinja2'):
+            template_name += '.jinja2'
+
+        template = self._get_env().get_template(template_name)
+
+        return jinja2.Markup(template.render(**kwargs))
+
+
+def initialize_deform_renderer(config):
+    deform.Form.set_default_renderer(
```
nickstenning: It's a small thing, but if you find yourself breaking a line in situations like this, then it's probably a good sign that an extra variable or two would provide some additional clarity:

```python
renderer = Jinja2Renderer(config.get_jinja2_environment())
deform.Form.set_default_renderer(renderer)
```

nickstenning:
LGTM.

PR #3586, opened 2016-07-07
Remove automagical annotation update handling

chdorner:
We had several issues recently where we ran migrations that then
obviously touched the `updated` column, but shouldn't have. So the
solution is to get rid of automatically changing this column and to do
it explicitly in `h.api.storage.update_annotation`.


nickstenning:
LGTM.

PR #3587, opened 2016-07-08
Unify dev and prod session configs

nickstenning:
Prior to this PR, h used a signed cookie-based session in development, while using a session-only cookie and a session persisted in Redis in production.

The development documentation already requires a user to install and configure Redis, so there is no need for this violation of dev-prod parity[1] to continue.

[1]: http://12factor.net/dev-prod-parity


nickstenning:
Turns out I already had this on a branch...

seanh:
lgtm

PR #3589, opened 2016-07-08
Move username charset validation to models layer

robertknight:
This PR is some preparatory refactoring extracted out of https://github.com/hypothesis/h/pull/3584

* Previously the username length was validated in the models layer but the username character set was not. That was only validated as part of the sign-up form validation. In order to support changing the username from other places (eg. the admin panel), add this validation to the models layer.
* Change the username -> userid helper to accept an auth_domain rather than a request so that it can be used from services which do not have a request object. This also saves some mocking in a test.

### DIFF
--- a/h/accounts/models.py
+++ b/h/accounts/models.py
@@ -2,6 +2,7 @@
 import datetime
 import hashlib
 import random
+import re
 import string
 
 import cryptacular.bcrypt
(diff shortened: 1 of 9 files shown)


nickstenning:
LGTM. The `User` model is getting a bit silly, but that's not your fault. I'll submit a small cleanup PR.

nickstenning:
(FWIW I just squash merged this entirely by accident. Sorry.)

PR #3591, opened 2016-07-08
Add unique constraint to user_group table

### LINKED ISSUE #2919: Missing unique constraint on "user_group" join table
We're missing a unique constraint on the "user_group" table, which means that in certain circumstances (probably a race condition somewhere) it's possible for multiple identical join rows to be present in the table. This isn't a problem until someone tries to leave the group, and then SQLAlchemy (because it's awesome) will spot that it's about to drop 2 rows when it was expecting to only drop 1 and bail. For example:

https://app.getsentry.com/hypothesis/prod/issues/107611330/

We should add a unique constraint on the tuple (user_id, group_id) on the join table, and write a migration to sort out the small number of existing duplicated rows.

seanh:
Fixes #2919

1. Remove duplicate rows from user_group table
2. Add (user_id, group_id) unique constraint to user group table, we don't want the same user to be a member of the same group twice
3. Also add a primary key to the user_group table to make it consistent with other relation tables in the db

Quick sanity test on production data for the data migration part:

```bash
$ pg_restore --single-transaction --clean --if-exists --no-owner --no-privileges -d postgresql://postgres@localhost/postgres <PRODUCTION_DB_DUMP.dump>; and alembic -c conf/alembic.ini stamp 6f86796f64e0
2016-07-08 19:04:46 30630 alembic.runtime.migration [INFO] Context impl PostgresqlImpl.
2016-07-08 19:04:46 30630 alembic.runtime.migration [INFO] Will assume transactional DDL.
2016-07-08 19:04:46 30630 alembic.runtime.migration [INFO] Running stamp_revision 9e6b4f70f588 -> 6f86796f64e0

$ psql postgresql://postgres@localhost/postgres -c "SELECT count(*) FROM (SELECT * FROM user_group GROUP BY (user_id, group_id) HAVING count('*') > 1) AS foo;"
 count
-------
    17
(1 row)

$ psql postgresql://postgres@localhost/postgres -c "SELECT count(*) FROM (SELECT DISTINCT (user_id, group_id) FROM user_group) AS foo;"
 count
-------
  8869
(1 row)

$ alembic -c conf/alembic.ini upgrade +1
2016-07-08 19:06:06 31008 alembic.runtime.migration [INFO] Context impl PostgresqlImpl.
2016-07-08 19:06:06 31008 alembic.runtime.migration [INFO] Will assume transactional DDL.
2016-07-08 19:06:06 31008 alembic.runtime.migration [INFO] Running upgrade 6f86796f64e0 -> 9e01b7287da2, Remove duplicates rows from the user_group table.

$ psql postgresql://postgres@localhost/postgres -c "SELECT count(*) FROM (SELECT * FROM user_group GROUP BY (user_id, group_id) HAVING count('*') > 1) AS foo;"
 count
-------
     0
(1 row)

$ psql postgresql://postgres@localhost/postgres -c "SELECT count(*) FROM (SELECT DISTINCT (user_id, group_id) FROM user_group) AS foo;"
 count
-------
  8869
(1 row)

$ 
```

### DIFF
--- a/h/groups/models.py
+++ b/h/groups/models.py
@@ -116,5 +116,6 @@ USER_GROUP_TABLE = sa.Table(
     sa.Column('group_id',
               sa.Integer,
               sa.ForeignKey('group.id'),
-              nullable=False)
+              nullable=False),
+    sa.UniqueConstraint('user_id', 'group_id'),
 )
--- a/h/groups/models.py
+++ b/h/groups/models.py
@@ -109,6 +109,7 @@ class Group(Base, mixins.Timestamps):
 
 USER_GROUP_TABLE = sa.Table(
     'user_group', Base.metadata,
+    sa.Column('id', sa.Integer, autoincrement=True, primary_key=True),
     sa.Column('user_id',
               sa.Integer,
               sa.ForeignKey('user.id'),
--- a/h/migrations/versions/9e01b7287da2_remove_duplicates_from_user_group_table.py
+++ b/h/migrations/versions/9e01b7287da2_remove_duplicates_from_user_group_table.py
@@ -1,0 +1,53 @@
+"""
+Remove duplicate rows from the user_group table.
+
+Revision ID: 9e01b7287da2
+Revises: 6f86796f64e0
+Create Date: 2016-07-08 17:54:57.399139
+"""
+
+from __future__ import unicode_literals
+
+from alembic import op
+import sqlalchemy as sa
+from sqlalchemy.ext.declarative import declarative_base
+from sqlalchemy.orm import sessionmaker
+
+revision = '9e01b7287da2'
+down_revision = '6f86796f64e0'
+
+Session = sessionmaker()
+
+
+user_group = sa.table('user_group',
+                      sa.Column('user_id', sa.Integer),
+                      sa.Column('group_id', sa.Integer))
+
+
+def upgrade():
+    session = Session(bind=op.get_bind())
+
+    # Find all the groups of duplicate user_group rows that have the same
+    # user_id and group_id values.
+    groups = (
+        session.query(user_group)
+        .group_by('user_id', 'group_id')
+        .having(sa.func.count('*') > 1))
+
+    for user_id, group_id in groups:
+        # Delete all the rows from the group of duplicate rows.
+        # This deletes _all_ the rows from the group, we'll have to put back
+        # one row later.
+        session.execute(
+            user_group.delete()
+            .where(user_group.c.user_id == user_id)
+            .where(user_group.c.group_id == group_id)
+        )
+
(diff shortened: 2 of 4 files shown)


h/groups/models.py
```diff
@@ -109,12 +109,14 @@ def created_by(cls, session, user):
 
 USER_GROUP_TABLE = sa.Table(
     'user_group', Base.metadata,
+    sa.Column('id', sa.Integer, primary_key=True),
```
nickstenning: For consistency with the rest of our models this should also specify `autoincrement=True`. Strictly speaking I don't believe it's necessary since sqla 1.1 which introduced `autoincrement='auto'` as the default, but no other models in our code define an integer PK without `autoincrement=True` so it took me a minute to work out whether it was correct.

h/migrations/versions/9e01b7287da2_remove_duplicates_from_user_group_table.py
```diff
@@ -0,0 +1,57 @@
...
+
+from alembic import op
+import sqlalchemy as sa
+from sqlalchemy.ext.declarative import declarative_base
+from sqlalchemy.orm import sessionmaker
+
+revision = '9e01b7287da2'
+down_revision = '6f86796f64e0'
+
+Base = declarative_base()
```
nickstenning: This doesn't appear to be used anywhere.

h/migrations/versions/9e01b7287da2_remove_duplicates_from_user_group_table.py
```diff
@@ -0,0 +1,57 @@
+"""
+Remove duplicates rows from the user_group table.
```
nickstenning: s/duplicates/duplicate/

h/migrations/versions/9e01b7287da2_remove_duplicates_from_user_group_table.py
```diff
@@ -0,0 +1,57 @@
...
+    # Find all the groups of duplicate user_group rows that have the same
+    # user_id and group_id values.
+    groups = (
+        session.query(user_group)
+        .group_by('user_id', 'group_id')
+        .having(sa.func.count('*') > 1))
+
+    for group in groups:
+        user_id = group[0]
+        group_id = group[1]
```
nickstenning: It's a small thing, but it is simpler and clearer in situations like this to unpack the tuple in the `for` statement itself:

```python
    for user_id, group_id in groups:
        ...
```

nickstenning:
Other than the relatively minor stylistic quibbles, this looks good -- I've tried it out and it seems to work well.

seanh:
Changes done

nickstenning:
LGTM.

PR #3593, opened 2016-07-11
Rename account settings URLs

finitud:
As per step#1 in https://trello.com/c/G3kr1awd/392-account-settings-profile-etc-forms-updates

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -13,7 +13,7 @@ Some of the API URLs documented below require a valid API token.
 To use these API URLs you should:
 
 1. Generate yourself an API token on your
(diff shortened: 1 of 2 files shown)


nickstenning:
A quick grep tells me that there's some API documentation which needs updating.

Also, we should update the route names to `account_*` rather than `profile_*`, and rename the relevant templates and layouts.

If you update the API docs I'll merge this, and then perhaps let's do the other changes in another PR.

nickstenning:
@finitud Did you see my comment?

finitud:
@nickstenning yep, taking care of this one now.

nickstenning:
LGTM.

PR #3596, opened 2016-07-12
Activity page skeleton: Refactor search query tests

chdorner:
_This is part one of a four-part series of pull requests that adds a new free-text search parser and a skeleton page for playing with it._

This is refactoring the tests for `h.api.search.query` to use test classes for testing the different filters and matchers. This will help the readability of the other diffs.

_These changes were split out of the PR #3588_

PR #3597, opened 2016-07-12
Activity page skeleton: Add free-text search query parser

chdorner:
_This is part two of a four-part series of pull requests that adds a new free-text search parser and a skeleton page for playing with it. This needs rebasing after #3596 is merged._

This pull request adds a free-text search parser that converts a query like `user:chdorner tag:reading atlantic` into a MultiDict that looks like what the search api expects `{'user': 'chdorner', 'tag': 'reading', 'any': 'atlantic'}`.
It supports different named fields, strings in double or single quotes, and should work with any Unicode characters. For a complete list check out the test examples.

### DIFF
--- a/h/api/search/parser.py
+++ b/h/api/search/parser.py
@@ -1,0 +1,112 @@
+# -*- coding: utf-8 -*-
+
+"""
+The query parser which converts our subset of the Apache Lucene syntax and
+transforms it into a MultiDict structure that h.api.search understands.
+"""
+
+from __future__ import unicode_literals
+
+from collections import namedtuple
+
+import pyparsing as pp
+from webob.multidict import MultiDict
+
+# Enable memoizing of the parsing logic
+pp.ParserElement.enablePackrat()
+
+# Named fields we support when querying (e.g. `user:luke`)
+named_fields = ['user', 'tag', 'group', 'uri']
+
+whitespace = set([
+    "\u0009",  # character tabulation
+    "\u000a",  # line feed
+    "\u000b",  # line tabulation
+    "\u000c",  # form feed
+    "\u000d",  # carriage return
+    "\u0020",  # space
+    "\u0085",  # next line
+    "\u00a0",  # no-break space
+    "\u1680",  # ogham space mark
+    "\u2000",  # en quad
+    "\u2001",  # em quad
+    "\u2002",  # en space
+    "\u2003",  # em space
+    "\u2004",  # three-per-em space
+    "\u2005",  # four-per-em space
+    "\u2006",  # six-per-em space
+    "\u2007",  # figure space
+    "\u2008",  # punctuation space
+    "\u2009",  # thin space
+    "\u200a",  # hair space
+    "\u2028",  # line separator
+    "\u2029",  # paragraph separator
+    "\u202f",  # narrow no-break space
+    "\u205f",  # medium mathematical space
+    "\u3000",  # ideographic space
+])
+
+parser = None
+Match = namedtuple('Match', ['key', 'value'])
+
+
+def parse(q):
+    """Parse a free text, Lucene-like, query string into a MultiDict.
+
+    "user:luke tag:foobar tag:news hello world" is parsed into:
+    ``
+    {
+        "user": "luke",
+        "tag": "foobar",
+        "tag": "news",
+        "any": "hello",
+        "any": "world"
+    }
+    ``
+
+    Supported keys for fields are ``user``, ``group``, ``tag``, ``uri``.
+    Any other search terms will get the key ``any``.
+    """
+    parser = _get_parser()
+    parse_results = parser.parseString(q)
+
+    # The parser returns all matched strings, even the field names, we use a
+    # parse action to turn matches into a key/value pair (Match), but we need
+    # to filter out any other matches that the parser returns.
+    return MultiDict([m for m in parse_results if isinstance(m, Match)])
+
+
+def _get_parser():
+    global parser
+    if parser is None:
+        parser = _make_parser()
+    return parser
+
+
+def _make_parser():
+    word = pp.CharsNotIn(''.join(whitespace))
+    word.skipWhitespace = True
+
+    value = pp.MatchFirst([
+        pp.dblQuotedString.copy().setParseAction(pp.removeQuotes),
+        pp.sglQuotedString.copy().setParseAction(pp.removeQuotes),
+        pp.Empty() + pp.CharsNotIn(''.join(whitespace)),
+    ])
+
+    expressions = []
+
+    for field in named_fields:
+        exp = pp.Suppress(pp.CaselessLiteral(field) + ':') + \
+            value.copy().setParseAction(_decorate_match(field))
+        expressions.append(exp)
+
+    any_ = value.copy().setParseAction(_decorate_match('any'))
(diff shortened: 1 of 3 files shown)


chdorner:

nickstenning:
Rebased.

h/activity/query.py
```diff
@@ -0,0 +1,122 @@
...
+    "\u2029",  # paragraph separator
+    "\u202f",  # narrow no-break space
+    "\u205f",  # medium mathematical space
+    "\u3000",  # ideographic space
+])
+
+parser = None
+
+
+class Match(object):
```
nickstenning: For a simple container like this I'd be inclined to use a `namedtuple`:

```python
from collections import namedtuple

Match = namedtuple('Match', ['key', 'value'])
```

You'll get the `__repr__` and constructor for free.

h/activity/query.py
```diff
@@ -0,0 +1,122 @@
...
+
+    Supported keys for fields are ``user``, ``group``, ``tag``, ``uri``.
+    Any other search terms will get the key ``any``.
+    """
+    parser = _get_parser()
+    parse_results = parser.parseString(q)
+
+    result = MultiDict()
+    for res in parse_results:
+        if not isinstance(res, Match):
```
nickstenning: Might be worth adding a comment to explain what `parse_results` will contain and why only some of the results will be instances of `Match`.

nickstenning: Also, this might be a shorter and equally clear way of writing this code:

```python
return MultiDict([m for m in parse_results if isinstance(m, Match)])
```

chdorner: :+1: to both of these comments.

h/activity/query.py
```diff
@@ -0,0 +1,110 @@
...
+    # Enable memoizing of the parsing logic
+    pp.ParserElement.enablePackrat()
+
+    word = pp.CharsNotIn(' ')
+    word.skipWhitespace = True
+
+    value = pp.MatchFirst([
+        pp.quotedString.copy().setParseAction(pp.removeQuotes),
+        word
+    ])
```
nickstenning: Couple of observations about the preceding five lines:

1. You don't actually seem to use the `whitespace` set at the moment as an argument to `CharsNotIn`, which probably isn't intentional?
2. I can't find any documentation of `skipWhitespace` attr for `CharsNotIn`, so I'd be inclined to use something like `pp.Empty() + pp.CharsNotIn(''.join(whitespace))` instead.
3. This is purely stylistic, but I think it might be clearer to add two `QuotedString` matchers rather than duplicating the `pp.quotedString` singleton. This also removes the need for `pp.removeQuotes`.

All in all, that would result in something like this:

```python
value = pp.MatchFirst([
    pp.QuotedString('"', escchar='\\'),
    pp.QuotedString("'", escchar='\\'),
    pp.Empty() + pp.CharsNotIn(''.join(whitespace)),
])
```

which seems a bit clearer to me?

h/activity/query.py
```diff
@@ -0,0 +1,110 @@
...
+def _get_parser():
+    global parser
+    if parser is None:
+        parser = _make_parser()
+    return parser
+
+
+def _make_parser():
+    # Enable memoizing of the parsing logic
+    pp.ParserElement.enablePackrat()
```
nickstenning: I think this should probably be at the top level of the module. It's a call with global side-effects, so I reckon making it really obvious up at the top of the module would be a good idea.

h/activity/query.py
```diff
@@ -0,0 +1,110 @@
...
+    value = pp.MatchFirst([
+        pp.quotedString.copy().setParseAction(pp.removeQuotes),
+        word
+    ])
+
+    expressions = []
+
+    named_fields = ['user', 'tag', 'group', 'uri']
+    for field in named_fields:
+        exp = (pp.Keyword(field, caseless=True) +
```
nickstenning: I'm reading up on the difference between `Keyword` and `Literal`, and this sounds a bit dodgy:

>  However, unlike the Literal class, the next input character must not be one of the characters in ... [the set of] all the letters and digits plus underbar (“_”) and dollar sign (“$”).

Notably, that set doesn't include non-ASCII characters.

I don't think it matters, because we then immediately require that this keyword is followed by a literal `:`. But in that case we could just use a literal. In addition, because of the parse action approach to decorating the match result, you can suppress the entire keyword match:

```python
exp = pp.Suppress(pp.CaselessLiteral(field) + ':') + value.copy().setParseAction(_decorate_match(field))
```

h/activity/query.py
```diff
@@ -0,0 +1,110 @@
...
+    word.skipWhitespace = True
+
+    value = pp.MatchFirst([
+        pp.quotedString.copy().setParseAction(pp.removeQuotes),
+        word
+    ])
+
+    expressions = []
+
+    named_fields = ['user', 'tag', 'group', 'uri']
```
nickstenning: I'd be inclined to move this to a module constant.

h/activity/query.py
nickstenning: Christof points out that (3) results in different behaviour when parsing quoted strings with odd numbers of characters such as `"hello""` or `"""`. The Elasticsearch query string query parser currently barfs on values with odd numbers of quotes, so I guess we get to treat them however we like.

tests/h/activity/query_test.py
```diff
@@ -0,0 +1,112 @@
...
+char_blacklist = query.whitespace.union(set('\'"'))
+nonwhitespace_chars = st.characters(blacklist_characters=char_blacklist)
+nonwhitespace_text = st.text(alphabet=nonwhitespace_chars, min_size=1)
+
+
+@given(nonwhitespace_text)
+@pytest.mark.fuzz
+def test_parse_with_any_nonwhitespace_text(text):
+    result = query.parse(text)
+    assert result.get('any') == text
```
nickstenning: This doesn't seem to me to be a valid invariant. Although it's unlikely, it's perfectly possible for `user:a` to be generated as an example here, and then the test will fail.

It's probably better to check something like `query.parse('user:' + example)` always results in the correct `user` entry in the results. You could even be fancy and do something like:

```python
@given(kw=st.sampled_from(('user', 'tag', 'group', ...)),
       value=nonwhitespace_text)
def test_parse_with_any_nonwhitespace_text(kw, value):
    result = query.parse(kw + ':' + value)
    assert result.get(kw) == value
```

nickstenning:
This mostly looks great to me! A couple of comments both on the code and the test suite, but nothing major.

I have only one general query (ba-dum TISH) -- which is about where you've put this module. This query parser will in all likelihood be used by more than just activity pages, so I reckon it could (and probably should) be called `h.api.search.parser`.

h/activity/query.py
chdorner: This looks more robust, thanks :+1:

tests/h/activity/query_test.py
chdorner: Went with the fancy version, also reading the fields from the new module-level variable :+1:

chdorner:
Done, and moved it to `h.api.search.parser` (I've had it in the `h.activity` package first because it's only used there now with the intention that it will be moved somewhere else once it will be used in another part of the application).

rimshot

nickstenning:
LGTM.

PR #3598, opened 2016-07-12
Activity page skeleton: Add UserFilter to search

chdorner:
_This is part three of a four-part series of pull requests that adds a new free-text search parser and a skeleton page for playing with it. This needs rebasing after #3597 is merged._

This pull request removes the `user` field from the `AnyMatcher` and adds a new `UserFilter` to our search query builder.
I think this should be a backwards-compatible change, but please look out for issues when reviewing.

### DIFF
--- a/h/api/search/core.py
+++ b/h/api/search/core.py
@@ -78,6 +78,7 @@ def default_querybuilder(request, private=True):
     builder.append_filter(query.AuthFilter(request, private=private))
     builder.append_filter(query.UriFilter(request))
     builder.append_filter(query.GroupFilter())
+    builder.append_filter(query.UserFilter())
     builder.append_matcher(query.AnyMatcher())
     builder.append_matcher(query.TagsMatcher())
     for factory in request.registry.get(FILTERS_KEY, []):
--- a/h/api/search/query.py
+++ b/h/api/search/query.py
@@ -185,6 +185,22 @@ class UriFilter(object):
         return {"terms": {"target.scope": list(uris)}}
 
 
+class UserFilter(object):
+
+    """
+    A filter that selects only annotations where the 'user' parameter matches.
+    """
+
+    def __call__(self, params):
+        if 'user' not in params:
+            return None
+
+        users = [v for k, v in params.items() if k == 'user']
+        del params['user']
+
+        return {'terms': {'user': users}}
+
+
 class AnyMatcher(object):
 
     """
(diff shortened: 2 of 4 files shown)


chdorner:

nickstenning:
Rebased.

nickstenning:
I think we need to be a little careful here, because currently searches like `https://hypothes.is/stream?q=user:chdorner` (that maps to `https://hypothes.is/api/search?user=chdorner` under the hood) work, and this would break that behaviour.

I'd suggest that we inspect the parameter to `user` and apply some transformations to it so that the query is expanded before it's sent to ElasticSearch:

```
chdorner                   →  acct:chdorner@hypothes.is
chdorner@hypothes.is       →  acct:chdorner@hypothes.is
acct:chdorner@hypothes.is  →  acct:chdorner@hypothes.is
```

(Where the domain is filled in with `request.auth_domain` if it's missing, for the time being.)

nickstenning:
As Christof points, out, this will happen automatically due to the analysis rules for the `user` field. In addition, after this PR merges, a search for `user@example.com` will now never match an annotation made by `acct:user@hypothes.is` as it will currently.

nickstenning:
LGTM.

chdorner:
Here's a summary of our Slack discussion about backwards compatibility:

The old version did a [match query](https://www.elastic.co/guide/en/elasticsearch/reference/1.5/query-dsl-match-query.html) on the user field, where the new code does a [terms filter](https://www.elastic.co/guide/en/elasticsearch/reference/1.5/query-dsl-terms-filter.html). The main difference is that the terms filter does _not analyze_ the terms, where the match query _analyzes_ the text before searching.

I did some testing locally, let's assume I have two annotations in the index:
```
id 1 - userid: acct:luke@hypothes.is
id 2 - userid: acct:luke@localhost
```

The results look like this:

Query | Results old code | Results new code
----- | ---------------- | ----------------
`user:luke` | `[1, 2]` | `[1, 2]`
`user:luke@hypothes.is` | `[1]` | `[1]`
`user:luke@localhost` | `[2]` | `[2]`
`user:acct:luke@hypothes.is` | `[1, 2]` | `[1]`
`user:acct:luke@localhost` | `[1, 2]` | `[2]`

This change improves the results, but it is a backwards-incompatible change. _But_ the backwards incompatibility only applies when the database includes users with the same username but different auth domains (and we only have 6 ones like that at the moment, so we can safely ignore these).

PR #3599, opened 2016-07-12
Activity page skeleton: Add skeleton page

chdorner:
_This is part foiur of a four-part series of pull requests that adds a new free-text search parser and a skeleton page for playing with it. This needs rebasing after #3598 is merged._

Finally, this pull request adds a skeleton page with a search form for playing around with the search parser. This is behind a feature flag, so the design shouldn't matter (it looks like the browser default styles).

### DIFF
--- a/h/activity/views.py
+++ b/h/activity/views.py
@@ -1,0 +1,51 @@
+# -*- coding: utf-8 -*-
+
+"""
+Activity pages views.
+"""
+
+from __future__ import unicode_literals
+
+from pyramid import httpexceptions
+from pyramid.view import view_config
+
+from h import models
+from h.api import search as search_lib
+from h.api.search import parser
+from h.api import storage
+
+
+@view_config(route_name='activity.search',
+             request_method='GET',
+             renderer='h:templates/activity/search.html.jinja2')
+def search(request):
+    if not request.feature('activity_pages'):
+        raise httpexceptions.HTTPNotFound()
+
+    results = []
+    total = None
+    if 'q' in request.params:
+        query = parser.parse(request.params['q'])
+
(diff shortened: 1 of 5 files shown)


chdorner:

chdorner:
Rebased.

h/app.py
```diff
@@ -113,6 +113,7 @@ def includeme(config):
     config.include('h.api', route_prefix='/api')
 
     # Core site modules
+    config.include('h.activity')
```
nickstenning: This should go in the block below. Sorry, that's probably not super clear, but the top block is site-wide utilities, infrastructure, plumbing stuff. The bottom block is specific site modules.

h/activity/views.py
```diff
@@ -0,0 +1,51 @@
...
+    if 'q' in request.params:
+        query = parser.parse(request.params['q'])
+
+        out = search_lib.search(request, query)
+        total = out['total']
+
+        anns = storage.fetch_ordered_annotations(request.db,
+                                                 [r['id'] for r in out['rows']])
+        for ann in anns:
+            group = request.db.query(models.Group).filter(models.Group.pubid == ann.groupid).one_or_none()
```
nickstenning: We should probably add an `annotations` relationship (with a `group` backref) to the `Group` to prevent having to do silly stuff like this.

nickstenning:
LGTM.

h/app.py
chdorner: Done on master :+1:

h/activity/views.py
chdorner: Yes, definitely. I think we can do that once we're actually writing proper views code for the activity pages

PR #3600, opened 2016-07-12
New login page design

seanh:
This pull request implements the new design for the `/login` page only, not any other pages. It implements the new design only, not any URL changes (`/login` isn't changing anyway). It adds a feature flag for the new login form design (or it should do before it's merged, the code isn't feature flagged yet at the time of writing)

### DIFF
--- a/h/static/images/icons/hypothesis.svg
+++ b/h/static/images/icons/hypothesis.svg
@@ -1,0 +1,71 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<svg
+   xmlns:sketch="http://www.bohemiancoding.com/sketch/ns"
+   xmlns:dc="http://purl.org/dc/elements/1.1/"
+   xmlns:cc="http://creativecommons.org/ns#"
+   xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#"
+   xmlns:svg="http://www.w3.org/2000/svg"
+   xmlns="http://www.w3.org/2000/svg"
+   xmlns:sodipodi="http://sodipodi.sourceforge.net/DTD/sodipodi-0.dtd"
+   xmlns:inkscape="http://www.inkscape.org/namespaces/inkscape"
+   width="24px"
+   height="28px"
+   viewBox="0 0 24 28"
+   version="1.1"
+   id="svg2"
+   inkscape:version="0.91 r13725"
+   sodipodi:docname="hypothesis.svg">
+  <metadata
+     id="metadata14">
+    <rdf:RDF>
+      <cc:Work
+         rdf:about="">
+        <dc:format>image/svg+xml</dc:format>
+        <dc:type
+           rdf:resource="http://purl.org/dc/dcmitype/StillImage" />
+      </cc:Work>
+    </rdf:RDF>
+  </metadata>
+  <sodipodi:namedview
+     pagecolor="#ffffff"
+     bordercolor="#666666"
+     borderopacity="1"
+     objecttolerance="10"
+     gridtolerance="10"
+     guidetolerance="10"
+     inkscape:pageopacity="0"
+     inkscape:pageshadow="2"
+     inkscape:window-width="1920"
+     inkscape:window-height="1136"
+     id="namedview12"
+     showgrid="false"
+     inkscape:zoom="8.4285714"
+     inkscape:cx="12"
+     inkscape:cy="14"
+     inkscape:window-x="0"
+     inkscape:window-y="27"
+     inkscape:window-maximized="1"
+     inkscape:current-layer="svg2" />
+  <!-- Generator: Sketch 3.5.2 (25235) - http://www.bohemiancoding.com/sketch -->
+  <title
+     id="title4">Rectangle 2 Copy 18</title>
+  <desc
+     id="desc6">Created with Sketch.</desc>
+  <defs
+     id="defs8" />
+  <g
+     id="Page-1"
+     stroke="none"
+     stroke-width="1"
+     fill="none"
+     fill-rule="evenodd"
+     sketch:type="MSPage"
+     style="fill:#3f3f3f;fill-opacity:1">
+    <path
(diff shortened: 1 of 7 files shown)


seanh:
I think all the details on this are correct according to Conor's spec. But the code needs tidying up and it needs putting behind a feature flag.

screenshot from 2016-07-12 18-25-25

untitled screencast - edited 29

seanh:
Oh, and the error messages need work. I think the error message text is the wrong size, and also the layout breaks if the error message is too long and wraps

seanh:
Should give it some accessibility thought too

seanh:
Do we want to disable the browser's own form validation on these pages, replace these validations with ones that come from our own code? I'm not sure if we can style these:

screenshot from 2016-07-13 10-50-48
screenshot from 2016-07-13 10-50-27

seanh:
@robertknight If you're gonna do any reviewing of this now, just note that it's still rough, I still instead to do some finishing off and cleaning up of the code and commit history, but general comments on what you want the approach to be here will still be applicable I think

h/static/styles/site.scss
```diff
@@ -1,5 +1,4 @@
 // minimal bootstrap CSS
-@import './vendor/bootstrap/bootstrap';
```
seanh: This bootstrap import will be removed by https://github.com/hypothesis/h/pull/3601 which should be merged first, so this pr won't have to remove this import

h/static/images/icons/hypothesis.svg
```diff
@@ -0,0 +1,71 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
seanh: If we're gonna do this as components then this image file should probably be in one folder along with the CSS and Jinja2/HTML for this "logged-out navbar" component.

h/static/styles/logged-out-navbar.scss
```diff
@@ -0,0 +1,4 @@
+.logged-out-navbar {
+  margin-left: 30px;
+  margin-top: 15px;
```
seanh: This is basically the amount of white margin space between the top and left edges of the document and any of our page content or elements. The same `30px` is repeated later on `.login-form` so that (if you make the browser window thin enough) the form aligns vertically with the h icon. This makes me think that this `30px` `margin/padding-left` ought to be on `<body>` or some containing element instead of duplicated.

h/static/styles/login-form.scss
```diff
@@ -0,0 +1,148 @@
...
+
+  // Keep a little white space around the form, even on small screens.
+  // This is the same size as the margin-left on the logged-out-navbar.
+  padding-left: 30px;
+  padding-right: 30px;
+}
+
+.login-form__heading {
+  font-size: $font-size-heading2;
+  font-weight: normal;
```
seanh: This should probably be moved to a new `$variable`, it seems that `normal` weight headings are part of our style now

h/static/styles/login-form.scss
```diff
@@ -0,0 +1,148 @@
...
+  // model as a whole remains the same size and does not push down elements
+  // that are positioned inside or below the <input> by a couple of pixels on
+  // focus.
+  margin-bottom: 14px;
+  padding-top: 28px;
+  padding-left: 9px;
+}
+
+.login-form .form-field-error .form-input {
+  border-color: $brand-color;
```
seanh: Not sure about using `$brand-color` for error messages. They can be the same _color_, but there should probably be a separate `$error-color` variable that has the same color value.

seanh:
I haven't really implemented this as components, at least not yet, but here's some thoughts on how components could be done:

Take the logged-out navbar as a handy example of a very simple component:

* Have a single folder, say `h/templates/logged-out-navbar`, that contains:

   - `logged-out-navbar.html.jinja2`
   - `logged-out-navbar.css`
   - `logged-out-navbar.js` (if it had any)

   This might require some tweaking of our assets setup, I don't think it allows templates, CSS and JS in the same folder (has different root folders for each)

* `logged-out-navbar.html.jinja2` will set some variable in the template code that causes `logged-out-navbar.css` and `logged-out-navbar.js` to be automatically included in the page if and only if `logged-out-navbar.html.jinja2` is used (totally doable, though I'm not sure how this will interact with bundling)

* Then all you need to do to re-use the component is `{% include  "h:templates/includes/logged-out-navbar.html.jinja2" %}` in any template

* Some components will require arguments to be passed in to the template when it's included. For example an `<input>` component might require the label text as an argument.

* Some components will position themselves relative to the parent element that the `{% include %}` is, i.e. they'll appear in the page where you used them. Others (like the navbar probably) will only make sense to be used in one place such as at the top of the page, and may absolutely position themselves as such,

seanh:
I've done this by re-using existing from `variables.scss`. Not sure if that's what we want, or if we want a clean start?

seanh:
TODO:

- [ ] https://github.com/hypothesis/h/pull/3601 should be merged first, then https://github.com/hypothesis/h/pull/3600/commits/fa4ebdab637d856a8e5b6bf44f587c1d02208454 can be removed from this PR

- [ ] The new form needs to be put behind a feature flag

- [ ] I think we should disable the browser's built-in validation (see comment with screenshot above)

- [ ] Do we want to do any component-ization of this code here, or can that wait for later pull requests?

- [ ] The CSS classes used by `form.html.jinja2` are not entirely BEM-ish. I've simply made use of the existing CSS classes. We _could_ change that template to be more BEM. Maybe separate PR though?

- [ ] There's an issue with wrapping of long error messages that needs to be fixed.

   Short errors are fine:

   screenshot from 2016-07-13 19-43-42

   But this issue appears when the error gets longer:

   screenshot from 2016-07-13 18-16-54

   The same would of course also happen with the label texts that are inside the boxes, but I think we can assume those will never wrap.

   What is _supposed_ to happen, from the design, is that the wrapped error message pushes down the user-entered text and the bottom of the input box:

   mobile_-_log_in_validation

   I don't think it's possible to implement the above without JavaScript. @robertknight was going to see if he could find a no-JavaScript way. If there isn't a no-JavaScript way then my suggestion would be that we put error messages under the input boxes instead of inside them, and then (in a separate pull request) have some JavaScript progressive enhancement that implements the error messages inside the input.

For separate PRs:

- I think we should do an accessibility review of this new form and see what should be done there. 

- Media queries to change sizes etc depending on screen size.

- Review all the error messages this form can generage.

seanh:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;yt9gXTWA&#x2F;391-logged-out-forms-updates"> Logged-out forms updates</a>

nickstenning:
I've picked this up.

nickstenning:
I'm going to close this for now, as I need to move some CSS and template stuff around in order to make this PR make sense. I'll reopen it once that's done.

PR #3601, opened 2016-07-13
Remove the new homepage feature

seanh:
Remove the new https://hypothes.is/ front page that was behind the
'new_homepage' feature flag, and all of the code that was added to
support it (and that isn't now also being used by something else).


seanh:
@robertknight Is this what you wanted or did I rip out far too much?

robertknight:
> @robertknight Is this what you wanted or did I rip out far too much?

I haven't tested the other site pages with this branch closely yet, but providing everything else still works then yes, this is fine. As @nickstenning said, we can get it back from version control if we need it.

seanh:
I tested the other pages and all seemed fine but please do test yourself as well

nickstenning:
LGTM.

PR #3603, opened 2016-07-14
Simplify client app template

robertknight:
Note that once this is merged, we need to coordinate a release of the H service that includes the changes in hypothesis/client#37

Once hypothesis/client#37 is merged, the H service only needs to include
`<app></app>` in the app.html file it serves and the client will
populate it with the root UI structure.

### DIFF
--- a/h/templates/app.html.jinja2
+++ b/h/templates/app.html.jinja2
@@ -22,7 +22,7 @@
       {% endfor %}
     {% endif %}
   </head>
-  <body ng-csp="">
+  <body>
     <hypothesis-app></hypothesis-app>
 
     <!-- App Configuration !-->
--- a/h/templates/app.html.jinja2
+++ b/h/templates/app.html.jinja2
@@ -22,50 +22,8 @@
       {% endfor %}
     {% endif %}
   </head>
-  <body ng-app="h" ng-strict-di ng-controller="AppController" ng-csp="">
-    <top-bar
-      auth="auth"
-      on-login="login()"
-      on-logout="logout()"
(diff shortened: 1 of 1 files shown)


h/templates/app.html.jinja2
```diff
@@ -22,50 +22,8 @@
...
-        ng-if="shareDialog.visible"
-        on-close="shareDialog.visible = false">
-      </share-dialog>
-      <help-panel ng-if="helpPanel.visible"
-        on-close="helpPanel.visible = false"
-        auth="auth">
-      </help-panel>
-      <main ng-view=""></main>
-    </div>
+  <body ng-csp="">
```
nickstenning: Do we still need `ng-csp` here? Could we wrap the whole of [client/templates/app.html](https://github.com/hypothesis/client/pull/37/files#diff-0a0cc75d40f6228c6858344efd190676) in an `<ng-csp>` tag instead?

robertknight: Now that I re-read the [documentation for ngCsp](https://docs.angularjs.org/api/ng/directive/ngCsp), yes, this should be possible. It adds a constraint that the contents of `app.html` must come before the `<script>` tag for the Angular bundle. I don't foresee that being a problem.

robertknight: Actually, reading the comments in the code (the docs are not that clear), the requirements is that some HTML tag in the document must have an `ng-csp` attribute on it. The init code in `app.js` could add this to the `<body>` tag.

nickstenning:
LGTM.

PR #3605, opened 2016-07-15
Centralise style configuration in base.html layout

nickstenning:
Rather than every single template getting to choose its own style bundle, it makes sense for the base layout to provide a sensible default which can be overridden by templates that extend the layout.

This commit sets things up so that:

1. By default, templates extending `base.html.jinja2` will get the `site_css` bundle included.
2. Legacy templates (those that include CSS from the client) are explicitly marked by setting a `uses_legacy_styles` flag in the template.
3. Templates which are not "legacy" but which wish to override the style bundle can set a `style_bundle` template variable to select a different style bundle.

This commit should have no user-visible side-effects.


robertknight:
LGTM

PR #3606, opened 2016-07-15
Add a "v2" site CSS bundle and make it the default

nickstenning:
This PR adds a `site-v2` CSS bundle, which is currently completely empty, and
makes it the default for all templates which extend from `base.html.jinja2`.

We add opt-outs for those pages (basically just group pages) which don't already explicitly opt-out of the default bundle.

This might seem a counterintuitive approach, as this commit means that the base layout "default bundle" is now not being used by *any* of the templates that extend the layout. But by making the default bundle `site_v2_css` we make it much easier to grep for templates which need updating to use the default bundle, and once fixed we simply *remove overrides* from those templates.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -140,6 +140,7 @@ var styleFiles = [
   './h/static/styles/front-page/main.css',
   './h/static/styles/help-page.scss',
   './h/static/styles/site.scss',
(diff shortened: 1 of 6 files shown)


robertknight:
LGTM

PR #3607, opened 2016-07-15
Remove a couple of unused CSS rules

robertknight:
Remove a couple of CSS rules that were left over from when the
Hypothesis client was part of this repository.


nickstenning:
LGTM.

PR #3609, opened 2016-07-18
CSS cleanups

nickstenning:
A couple of small commits cleaning up our stylesheets:

1. Move `front-page/main.css` to `front-page.css` alongside other "entrypoint" stylesheets.

2. Rename all partials to have leading-underscored names, so it's clear they are partials.


h/static/styles/help-page.scss
```diff
@@ -1,10 +1,10 @@
 $base-font-size: 16px;
 $base-line-height: 22px;
 
-@import './reset';
-@import './common';
-@import './elements';
-@import './mixins/icons';
+@import 'reset';
```
robertknight: I know that SASS does not require `./` to specify relative paths but I preferred to use them as it made the syntax we use for relative imports consistent between the JS and SASS code.

robertknight: If in future we move foundational modules such as colors and typography to a separate module published on npm, then I imagine we would add `node_modules/` to the search path and the imports might look like this:

```scss
@import 'hypothesis-frontend-toolkit/colors'

@import './mixins/something-local'
```

nickstenning: Agreed, but relative imports are horrible in CSS anyway. In #3610 and soon another PR all imports will be absolute and as a result hopefully much clearer.

robertknight: > Agreed, but relative imports are horrible in CSS anyway.

Horrible in SASS you mean? Can you expand on this?

nickstenning: It's more a conversation for beer than a pull request, but here's the thing: there are two different kinds of `@import` in SASS:

1. Those that result in output.
2. Those that do not (i.e. mixin modules).

If you use imports in SASS the same way you're using using imports in other code (i.e. assume they are side-effect free and import them as close as possible to where they're used) then sooner or later you're going to mistake an import of type

robertknight: > Relative imports imply that CSS imports can be used this way, and lull you into a
> false sense of security ("It's ok to import anything from anywhere").

Makes sense. Thanks for the explanation.

PR #3610, opened 2016-07-18
Move all partials into a 'partials' directory

nickstenning:
Moving these files out of the way will make it easier to expand partials into full modules (a directory containing multiple SCSS files) in future, and makes it even easier to see which are the entrypoint CSS files.


robertknight:
> Moving these files out of the way will make it easier to expand partials into
> full modules (a directory containing multiple SCSS files) in future,

Can you give an example of what that might look like?

robertknight:
LGTM

PR #3611, opened 2016-07-18
Lowercase search term in user filter

### LINKED ISSUE #3608: api query for user fails for mixed-case name
This used to work:

https://hypothes.is/api/search?user=SitC

Now it fails, although this works:

https://hypothes.is/api/search?user=sitc

The username really is mixed-case:

```
            "user": "acct:SitC@hypothes.is",
            "document": {
                "title": [
                    "Science in the Classroom | A collection of annotated research papers and accompanying teaching materials"
                ]
            },
            "id": "AVOlEijJH9ZO4OKSlXyR",
            "permissions": {
                "read": [
                    "group:__world__"
                ],
                "admin": [
                    "acct:SitC@hypothes.is"
```

chdorner:
The field analyzer lowercases the value without adding a search analyzer
to lowercase it as well, so we need to do that before we pass the query
to Elasticsearch.

Fixes #3608

### DIFF
--- a/h/api/search/query.py
+++ b/h/api/search/query.py
@@ -195,7 +195,7 @@ class UserFilter(object):
         if 'user' not in params:
             return None
 
-        users = [v for k, v in params.items() if k == 'user']
+        users = [v.lower() for k, v in params.items() if k == 'user']
         del params['user']
 
         return {'terms': {'user': users}}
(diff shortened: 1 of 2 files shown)


nickstenning:
LGTM.

PR #3612, opened 2016-07-18
Simplify CSS imports

nickstenning:
This PR adds a "base partial" which defines variables and includes all mixins.

This makes tracking allowable imports in CSS files much simpler.

This commit removes unused/redundant imports from partial CSS files so that the following now applies:

- entrypoint CSS files *only* import partials (no mixins)
- no partials other than the base partial import anything

In addition, this PR:

- moves a small handful of mixins out of partials and into the `mixins/` directory
- removes some unused variables
- renames '_common.scss' to '_help-page.scss' -- it was previously styles "common" to the web application and the client, but now the help page is the only page with these styles.

_*N.B.* This PR builds on #3609 and #3610 and will need rebasing._


robertknight:
> no partials other than the base partial import anything

This is at least more consistent than what we have now. What I would like to look into options for in future is making dependencies of individual partials explicit.

PR #3613, opened 2016-07-18
Add temporary redirects for renamed user account URLs

finitud:
This is required so that the older versions of the client do not stop
working while the extension gets updated.

### DIFF
--- a/h/tweens.py
+++ b/h/tweens.py
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 
 import collections
+from pyramid import httpexceptions
 
 
 def auth_token(handler, registry):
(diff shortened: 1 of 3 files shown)


nickstenning:
I think there must have been a miscommunication somewhere along the way, because the goal was to serve actual redirects, not to make the forms appear at the old URLs.

h/tweens.py
```diff
@@ -96,3 +97,20 @@ def content_security_policy_tween(request):
         return resp
 
     return content_security_policy_tween
+
+
+def redirect_tween_factory(handler, registry):
+    redirects = {
```
nickstenning: I'd suggest you move this into a module constant, which is then a default argument to the factory function:

```python
REDIRECTS = {...}

def redirect_tween_factory(handler, registry, redirects=REDIRECTS):
    ...
```

Then you can add some tests for this function :)

nickstenning:
LGTM.

PR #3614, opened 2016-07-18
Remove unused SASS variables and mixins

robertknight:
Remove SASS variables and mixins that are unused following separation of the client and web service into separate repositories.

The list of unused variables and mixins was found with [sass-unused](https://github.com/robertknight/sass-unused). I have left in variables from the new color palette which are unused because they will be used very soon by new designs.

### DIFF
--- a/h/static/styles/partials/_base.scss
+++ b/h/static/styles/partials/_base.scss
@@ -20,15 +20,12 @@ $color-gray: #818181;
 $color-silver-chalice: #a6a6a6;
 $color-silver: #bbb;
 $color-alto: #dedede;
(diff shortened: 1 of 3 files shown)


h/static/styles/partials/_base.scss
```diff
@@ -126,3 +114,4 @@ $small-line-height: 12px;
 // Include all mixins, making them available to the stylesheet. N.B. Mixins
 // should *never* output CSS, so this is safe.
 @import 'mixins/all';
+
```
nickstenning: 😦

nickstenning:
LGTM.

PR #3617, opened 2016-07-20
WIP: Search aggregations

chdorner:


h/activity/views.py
```diff
@@ -24,14 +25,20 @@ def search(request):
...
 
-        out = search_lib.search(request, query)
-        total = out['total']
+        search_request = search_lib.Search(request)
+        search_request.append_filter(query.TopLevelAnnotationsFilter())
+        search_request.append_aggregation(query.TagsAggregation())
+        result = search_request.run(search_query)
+        total = result.total
+
+        tags = [t['key'] for t in result.aggregations['tags']['buckets']]
```
chdorner: Instead of this I could also see a solution where the `TagsAggregation` class has a `parse` function which would take the aggregation response and turn it into something more meaningful, in this case just a list of strings.

h/api/search/query.py
```diff
@@ -32,6 +36,7 @@ def build(self, params):
 
         filters = [f(params) for f in self.filters]
         matchers = [m(params) for m in self.matchers]
+        aggregations = dict([(a.key, a()) for a in self.aggregations])
```
nickstenning: So's you know, dict comprehensions are a thing: `{a.key: a() for a in self.aggregations}`.

chdorner: :+1: will do that

seanh:
Wondering whether we need both `class Builder` and `class Search` now, or would just `class Search` do? Looks good though

chdorner:
I've split this up into multiple PRs: #3622, #3623, #3624

PR #3618, opened 2016-07-20
Bucket by time and document

seanh:


seanh:
On first glance the performance of this bucketing code makes quite a hit on the response time.

On my laptop with some production data, on `master` loading `/search` takes 103ms-208ms (very cheap test: just loading the page 10 times in Chrome with caching disabled, looking at the response times in dev tools). On this branch, 210-314ms.

And this is rendering only 20 (!) annotations (the search lib default). If we're bucketing the annotations by document and showing those documents collapsed initially, I think we're gonna want a lot more than 20 annotations per page, in order for that to result in a decent number of documents to display and for it not to be too easy for an annotation storm on one document to flood the results leading to only one document being displayed.

seanh:
An issue with searching for annotations and then rolling them up by document, when the documents are initially collapsed on page load: in order to produce a respectable-looking search result you want a decent-sized list of documents. Say 25 documents per page. A search result page showing 2 documents is going to look silly. But when you search for N annotations, you don't know how many documents that might produce, maybe it'll be 25, maybe it'll be 2.

Pagination and infinite scroll is also going to be strange, if you have to click to go to the next page after just documents, and then the next page has 50 documents on it, and then the page after that has 1, etc.

Basically we want to tell the search engine "give me N documents" not "give me N annotations"

seanh:
> If we're gonna show the results as a list of (collapsed) documents, my guess is we want to tell the search engine "give me N documents" not "give me N annotations". We have to somehow do a search for "the most recent N documents that have an annotation that matches a query, ordered by most recent annotation".

Assuming we find a way to do this query, if your query limit is n documents then you don't know how many annotations that would produce. How many of the most recent annotations that match the query do you have to consume, until you have annotations of 25 different documents? Could be 25 annotations, could be 25000. So the page render time could get unacceptably long.

I think what you'd have to do is query for the documents and num annotations per document only, don't retrieve or render the actual annotations. Then when the user clicks to expand a document, we do a full page reload, render the same search resul

seanh:
**If we're going to be rendering search results as a collapsed list of documents** then I don't think that searching for N annotations and then extracting document buckets from those annotations (as this pull request does) is going to work.

The problem is that you don't know how many unique documents the 20 most recent annotations that matched your search are going to produce. Could be 20, could be 1.

This leads to two issues:

1. Sometimes the search result page will look silly, as it'll contain only one document. Either just one time frame with one document in it, or a few time frames containing the same one document repeated a few times.

2. The number of results per page will change as you paginate through the results: the first page of the search results may have one document on it, then the second page may have 20 documents, then the third page 4 documents, and so on. (Or the same problem but with infinite scrolling.)

We may somehow be able to implement a search for N _documents_ rather than N annotations. We'd need to do a search for _"the most recent N documents that have any annotations that match this query, ordered by most recent annotation"_. But assuming we could implement that search, that wouldn't work either, because you wouldn't know how many of the most recently updated matching _annotations_ you would have to consume in order to find your 25 unique documents. If it's too many annotations, response time will be too slow as it tries to render hundreds or thousands of annotations.

A third approach might be to do the search for N _documents_, but retrieve and render the documents and num. annotations per document only, don't retrieve or render any of the annotations. Then when a document is clicked on to expand it, then you retrieve and render the matching annotations of that document. But this approach also has a similar problem: you don't know how matching annotations of document A you had to consume on your way to finding N unique documents. Expanding document A and retrieving and rendering all of its annotations might again make the page too slow by trying to render too many annotations.

So I think what we'll have to do in the end is:

1. Find a way to search for N _documents_ rather than N annotations; and
2. Retrieve only the most recent M matching annotations of each document, truncating the rest behind a "See more" link

That would give us control of both N and M and thus control of the size of each page of results, keeping the sizes of results pages regular and preventing them from being either too large or too small.

**If the list of documents is always expanded** on page load to show all the annotations under each document, then it's the number of annotations not the number of documents that determinses the size of the page on screen, we'll just need to query for say 25 annotations and we'll get a respectable-looking page showing 25 results (even if all 25 may be indented under just one time frame and document), and as you paginate each page will give you 25 more annotations until you get to the end. The code in this PR would work fine.

nickstenning:
You're absolutely right, of course, but we know that, we've considered the possible problems that raises, and I think we should plough on and see what this solution looks like in practice on a wide range of example queries.

seanh:
This is ready to be reviewed now.

I changed the timeframe bucketing algorithm a little from what I had before, as suggested by Nick, to an algorithm that loops over the annotations (once) and only requests the next timeframe each time it comes to the first annotation that doesn't fit within the current timeframe. This minimizes the amount of work done per annotation.

There are two difficulties with this. One is that you actually need to look at the annotations in order to compute the full list of timeframes required to bucket them, because we have timeframes Jul 2016, Jun 2016 etc that depend on the times of the annotations. It's not just a fixed set of timeframes like "This week", "Last month" , etc.

Second, closely related to the first difficulty is that the value of the next timeframe depends on the value of the next annotation to be bucketed. For example the previous timeframe may have been Jul 2016, but if the very next annotation is from Nov 1968 then the next timeframe should be Nov 1968 and not, for example, Jun 2016. We don't want to generate all the timeframes Jun 2016 - Dec 1968 only to throw them away.

All this means that a standard iterator (`next()`) won't work for producing the list of timeframes. But a pseudo-iterator in which `next(annotation)` takes the next annotation to be bucketed as an argument works great.

There are lots of different kinds of timeframes that we might want to use. "Yesterday, "Last week", "Last month" etc are "static" timeframes in that their values are generated from the current time only, not from the annotations. Timeframes of the form "June 2016" or "2015" etc are "dynamic" timeframes in that to generate their values you need to consider the annotations to be bucketed. "Ages ago" (or say "More than five years ago" etc) might be an "end bucket" timeframe in that it has no cutoff point, any annotation that didn't fit within a previous timeframe would always fit within this one.

I think the current algorithm can handle any combinations of all these kinds of timeframes just by modifying `TimeframeGenerator`, so we should be able to try out whatever we want to.

h/activity/bucketing.py
```diff
@@ -0,0 +1,128 @@
...
+        annotation may also be within the cutoff times of previous timeframes.
+        It's up to the caller to handle this.
+
+        """
+        return annotation.updated >= self.cutoff_time
+
+    def __repr__(self):
+        return '{class_} "{label}" with {n} document buckets'.format(
+            class_=self.__class__, label=self.label,
+            n=len(self.document_buckets))
```
seanh: This is to make reading test failure outputs easier

h/activity/bucketing.py
```diff
@@ -0,0 +1,128 @@
...
+
+
+class TimeframeGenerator(object):
+
+    def __init__(self):
+        self.timeframes = [
+            Timeframe(_("Last 7 days"), utcnow() - datetime.timedelta(days=7)),
```
seanh: We can add as many "static" (i.e. hardcodeable) timeframes as we want to this list. "Just now", "Yesterday", "Last month", etc

h/activity/bucketing.py
```diff
@@ -0,0 +1,128 @@
...
+        """
+        Return the next timeframe to be used for bucketing annotations.
+
+        :param annotation: the next annotation to be bucketed, the returned
+            timeframe is guaranteed to be the correct timeframe for this
+            annotation
+
+        """
+        while self.timeframes:
+            timeframe = self.timeframes.pop(0)
```
seanh: Because we `pop()` here, when `next()` is called again, the next annotation to be bucketed will not be checked against any timeframe that any previous annotation has already been checked against. These checks would have been a waste because they'd never pass.

h/activity/bucketing.py
```diff
@@ -0,0 +1,128 @@
...
+            if timeframe.within_cutoff(annotation):
+                return timeframe
+
+        cutoff_time = datetime.datetime(year=annotation.updated.year,
+                                        month=annotation.updated.month,
+                                        day=annotation.updated.day)
+        timeframe = Timeframe(cutoff_time.strftime('%b %Y'), cutoff_time)
```
seanh: Here we just jump to generating "Jun 2016"-type timeframes (a list that potentially goes on forever, there will always be one of these timeframes for every annotation). We could easily change this code to for example only generate "MON YEAR" timeframes for annotations from this year and then change to "YEAR" timeframes for older ones and then finally return a "More than five years ago" "end bucket" timeframe for really older ones ... Or whatever we might want

h/activity/bucketing.py
```diff
@@ -0,0 +1,128 @@
...
+        return []
+
+    generator = TimeframeGenerator()
+    timeframes = [generator.next(results[0]['annotation'])]
+
+    for result in results:
+        annotation = result['annotation']
+        if not timeframes[-1].within_cutoff(annotation):
+            timeframes.append(generator.next(annotation))
+        timeframes[-1].append(result)
```
seanh: I think this is O(n)

h/templates/activity/search.html.jinja2
```diff
@@ -7,25 +7,31 @@
...
+                  <p>User: {{ result.annotation.userid }}</p>
+                  <p>Group: {{ result.group.name }}</p>
+                  <p>Shared: {{ result.annotation.shared }}</p>
+                  <p>URI: {{ result.annotation.target_uri }}</p>
+                  <p>Text: {{ result.annotation.text }}</p>
+                  <p>Tags: {{ ','.join(result.annotation.tags) }}</p>
+                </li>
+              {% endfor %}
+            </ol>
+          </li>
```
seanh: I wanted nested lists (rather than a table) because the indentation makes it easier to scan the output to see if it's correct

tests/h/activity/test_bucketing.py
```diff
@@ -0,0 +1,147 @@
...
+        self.document_buckets = document_buckets
+
+    def __eq__(self, timeframe):
+        return (self.label == timeframe.label and
+                self.document_buckets == timeframe.document_buckets)
+
+    def __repr__(self):
+        return '{class_} "{label}" with {n} document buckets'.format(
+            class_=self.__class__, label=self.label,
+            n=len(self.document_buckets))
```
seanh: This is also for test fail output

seanh:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;OARFSoMf&#x2F;404-render-annotations-bucketed-by-document-and-time-period-on-activity-page-skeleton"> Render annotations bucketed by document and time period on activity page skeleton</a>

h/activity/bucketing.py
```diff
@@ -0,0 +1,128 @@
...
+
+    """
+    if not results:
+        return []
+
+    generator = TimeframeGenerator()
+    timeframes = [generator.next(results[0]['annotation'])]
+
+    for result in results:
+        annotation = result['annotation']
```
chdorner: I think we should `results` in this function to be just a list of annotations. I just did this dictionary with `annotation` and `group` for easier rendering, but that's skeleton code that should probably not affect this bucketing.

Nick also mentioned recently on my code that we should probably add a relationship in `h.groups.models.Group` to annotations with a `backref`, so the skeleton could then just call `annotation.group` (and we should probably eager-load this relationship).

seanh: Hmm, we can change `views.py` to pass the annotations to `bucket()` _before_ it fetches the group for each annotation and turns them into "results"

chdorner: Yes, we can also just forget about the group by adding the relationship, then the template can get the group out of the annotation object. A document bucket will then just be a list of `h.api.models.Annotation` objects.

seanh: Is it a problem that `Annotation` is in `h.api` but `Group` is in `h.group.models`? iirc we didn't want `h.api` to have any dependencies on the rest of `h`, and I think a `backref` from `h.api.models.Annotation` referencing a relationship in `h.group.models.Group` would be such a dependency, wouldn't it? Not at the actual database level necessarily but at least at the Python model code level.

seanh: Oh no wait, I think it can be done with _only_ changes to `Group` code...

chdorner: Yes, exactly. By adding the relationship to the `Group` code but including a `backref` we can get the relationship with only changing groups code.

seanh: I'm not sure if this relationship is possible, because `Annotation` and `Group` are registered with difference declarative base classes, you'll get something like this:

```
sqlalchemy.exc.InvalidRequestError: When initializing mapper Mapper|Group|group, expression 'Annotation.groupid==Group.pubid' failed to locate a name ("name 'Annotation' is not defined"). If this is a class name, consider adding this relationship() to the <class '__main__.Group'> class after both dependent classes have been defined.
```

We could of course add a _property_ `.group` to `Annotation` as just a Python property, not a sqlalchemy relationship, but then I don't think eager loading would be possible.

seanh: I've pushed a [change that makes the view construct its result dicts after bucketing instead of before](https://github.com/hypothesis/h/pull/3618/commits/f14857adfb37737d17e3c08add6f6fc96b21d1b6), so I think we can deal with this relationship issue in a separate pr if we want to

chdorner:
LGTM. Let's coordinate merging this and my other activity pages PRs since we're both changing `h/activity/views.py` and always have to fix merge conflicts.

seanh:
Sure, let's just get the search aggregations stuff merged first and then rebase this

chdorner:
Rebased and fixed the merge conflicts (just in `h/activity/views.py`)

seanh:
Cheers!

PR #3621, opened 2016-07-22
Remove PhantomJS 1.x-specific polyfills

robertknight:
These are no longer required with PhantomJS 2.x

See https://github.com/hypothesis/client/pull/45


nickstenning:
LGTM.

PR #3622, opened 2016-07-22
Search aggregations (1): Return a SearchResult object when searching

chdorner:
Now that we don't just blindly render what ElasticSearch returns we can
change the return value of `h.api.search.core.search` to be a value
object for easier access and sane default values. This is also done as a
preparation for aggregation support to search.

### DIFF
--- a/h/api/search/core.py
+++ b/h/api/search/core.py
@@ -1,5 +1,6 @@
 # -*- coding: utf-8 -*-
 import logging
+from collections import namedtuple
 
 from h.api.search import query
 
@@ -7,6 +8,11 @@ FILTERS_KEY = 'h.api.search.filters'
 MATCHERS_KEY = 'h.api.search.matchers'
 
 log = logging.getLogger(__name__)
+
+SearchResult = namedtuple('SearchResult', [
+    'total',
+    'annotation_ids',
+    'reply_ids'])
 
 
 def search(request, params, private=True, separate_replies=False):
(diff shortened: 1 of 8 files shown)


chdorner:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;5SPnhoee&#x2F;403-add-aggregation-capability-to-search-package"> Add aggregation capability to search package</a>

h/api/search/core.py
```diff
@@ -41,36 +47,31 @@ def search(request, params, private=True, separate_replies=False):
...
-    results = es.conn.search(index=es.index,
-                             doc_type=es.t.annotation,
-                             body=builder.build(params))
-    total = results['hits']['total']
-    docs = results['hits']['hits']
-    rows = [dict(d['_source'], id=d['_id']) for d in docs]
-    return_value = {"rows": rows, "total": total}
+    response = es.conn.search(index=es.index,
+                              doc_type=es.t.annotation,
+                              body=builder.build(params))
```
nickstenning: This should probably now set `_source=False`.

h/api/search/core.py
```diff
@@ -41,36 +47,31 @@ def search(request, params, private=True, separate_replies=False):
...
         builder = default_querybuilder(request, private=private)
-        builder.append_matcher(query.RepliesMatcher(
-            [h['_id'] for h in results['hits']['hits']]))
-        reply_results = es.conn.search(index=es.index,
-                                       doc_type=es.t.annotation,
-                                       body=builder.build({'limit': 200}))
+        builder.append_matcher(query.RepliesMatcher(annotation_ids))
+        reply_response = es.conn.search(index=es.index,
+                                        doc_type=es.t.annotation,
+                                        body=builder.build({'limit': 200}))
```
nickstenning: Ditto on `_source=False`.

tests/h/api/views_test.py
```diff
@@ -73,8 +73,8 @@ def test_it_searches(self, pyramid_request, search_lib):
                                                   separate_replies=False)
 
     def test_it_loads_annotations_from_database(self, pyramid_request, search_lib, storage):
-        search_lib.search.return_value = {'total': 2,
-                                          'rows': [{'id': 'row-1'}, {'id': 'row-2'}]}
+        search_lib.search.return_value = mock.Mock(total=2,
```
nickstenning: Given that `SearchResult` is just a data structure, it might make more sense to use it here in place of a `Mock` object.

chdorner:
Fixed the review comments and also rebased all subsequent pull requests for easier rebasing once this is merged.

nickstenning:
LGTM. From a little local testing this comes with a (small) bonus performance boost.

PR #3624, opened 2016-07-22
Search aggregations (3): Tags aggregation

chdorner:
_This is based on #3623 and will need rebasing once that one got merged_

Aggregations can be added to search queries the same way we add filters and matchers. The interface of an aggregation is very similar (a callable), but it also needs to expose a `parse_result` method which parses the Elasticsearch result and thus allows us to decouple callers of the `Search` class to know anything about how Elasticsearch renders search results.

There is also a `TagsAggregation` which can be added to search queries, check the activity page skeleton on how it is being used.

PR #3625, opened 2016-07-22
Search aggregations (4): Users aggregation

chdorner:
_This is based on #3624 and will need rebasing once that one got merged_

The aggregation is pretty much the same as the tags aggregation. One caveat is that [analyzed fields have a big impact on how aggregations work](https://www.elastic.co/guide/en/elasticsearch/guide/1.x/aggregations-and-analysis.html). The solution for this is to add a new `user_raw` field to the search index which is not analyzed.

The activity page skeleton will automatically add the users aggregation if the search contains exactly one group.

### DIFF
--- a/h/activity/views.py
+++ b/h/activity/views.py
@@ -25,7 +25,7 @@ def search(request):
     if not request.feature('activity_pages'):
         raise httpexceptions.HTTPNotFound()
 
-    results = []
+    timeframes = []
     total = None
     tags = []
     users = []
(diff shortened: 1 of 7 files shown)


chdorner:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;DpWmSc8w&#x2F;406-add-a-user-aggregation"> Add a &quot;user&quot; aggregation</a>

seanh:
Rebased

PR #3628, opened 2016-07-26
Make HTML enhancements URL-independent

robertknight:
This implements a new system for upgrading elements on the page with
additional functionality which does not require hardcoding the routes
that might use such functionality into the client-side code.

Instead a map of selector to controller classes is created in the root
site bundles and the `upgradeElements` module handles upgrading all
matching elements on the page.

This also gives us a central place to put logic relating to upgrades to
catch issues such as errors upgrading elements or elements which should
be upgraded but for which no matching controller was found.

At some point it will likely make sense to use an existing library to handle
this, but this is a first step and we'll get a better feel for the requirements
as the Activity Pages project evolves.


h/templates/admin/users.html.jinja2
```diff
@@ -70,7 +70,7 @@
 
       <h3>Danger Zone</h3>
 
-      <form id="js-users-delete-form" method="POST" action="{{request.route_path('admin_users_delete')}}" class="form-inline">
+      <form class="js-users-delete-form" method="POST" action="{{request.route_path('admin_users_delete')}}" class="form-inline">
```
nickstenning: This element now has two `class` attributes, which is invalid HTML.

h/static/scripts/upgrade-elements.js
```diff
@@ -0,0 +1,29 @@
...
+ *
+ * @param {Element} root - The root element to search for matching elements
+ * @param {Object} controllers - Object mapping selectors to controller classes.
+ *        For each element matching a given selector, an instance of the
+ *        controller class will be constructed and passed the element in
+ *        order to upgrade it.
+ */
+function upgradeElements(root, controllers) {
+  Object.keys(controllers).forEach(function (selector) {
+    var elements = Array.from(root.querySelectorAll(selector));
```
nickstenning: `Array.from` doesn't appear to be supported at all in IE. Are we sure that's ok?

robertknight: We're polyfilling this and several other ES6 additions to the standard library in `polyfills.js`

nickstenning: :+1:

robertknight: Ah! - But we're not actually loading the polyfill bundle into site pages at the moment. Glad this was caught.

nickstenning:
LGTM.

PR #3630, opened 2016-07-27
Add orphans flag.

sheetaluk:
Add a feature flag for the orphans selection tab.

https://github.com/hypothesis/client/pull/52/files
https://trello.com/c/1NVK2jwv/400-add-a-tab-for-unanchored-annotations


h/features/models.py
```diff
@@ -14,6 +14,7 @@
 FEATURES = {
     'activity_pages': "Show the new activity pages?",
     'selection_tabs': "Show the tabs to select between annotations and notes?",
```
robertknight: I think these are supposed to be in alphabetical order

nickstenning:
LGTM. I fixed the ordering. Merged as afbab38.

PR #3633, opened 2016-07-27
Enable ESLint

robertknight:
To make the H service code consistent with our client code, this adds a lint task which reuses the shared ESLint config from https://github.com/hypothesis/frontend-toolkit/pull/2 and makes fix-ups to the code. _Note that this PR won't be usable until that PR has been merged and the package published to npm_.

I'm doing this now because its easier to get to a clean slate when there isn't much front-end code left in the repository.

Hound does support ESLint but I'm not sure if it supports shared configurations. If not we can get Travis to do this, similar to what was done for the client repo.


seanh:
Looks fine, Jenkins build needs to be fixed

robertknight:
> Looks fine, Jenkins build needs to be fixed

For future references, I added eslint and friends to devDependencies since it is not required for a normal build and therefore should not be included in the Docker image. However, in the gulpfile the module was _always_ being loaded, even if the `lint` task was never being run.

This is now fixed by loading it lazily in the same way as `karma`.

PR #3637, opened 2016-07-27
Expose feature flags to deform templates

robertknight:
Add a mechanism to customize the rendering of form controls by deform
based on properties of the request, such as the active feature flags.

 * Add a `create_form` factory function on the request which returns
   a deform.Form that is configured to use a renderer which exposes the
   request's feature flags as a `feature` callable in the context

 * Replace all direct uses of `deform.Form` with `request.create_form`

 * Remove global override for deform.Form renderer as it is now unused


h/form.py
```diff
@@ -58,14 +65,25 @@ def __call__(self, template_name, **kwargs):
...
+    """
+    Return a callable which creates deform.Form instances for a request.
+
+    The callable accepts the same arguments as deform.Form but the resulting
+    form will be configured to have access to properties of the request in its
+    templates, such as feature flags.
+    """
+    renderer = Jinja2Renderer(config.get_jinja2_environment(), {
+        'feature': request.feature })
+    return partial(deform.Form, renderer=renderer)
```
nickstenning: So this now creates an instance of `Jinja2Renderer` each time you create a form. This isn't necessarily the end of the world, but it's also not necessary.

Perhaps save a single instance of the renderer on the registry and then just deal with the context in the `render` function -- i.e. the `__call__` method.

robertknight:
LGTM

PR #3638, opened 2016-07-27
Add "proxy" auth policy mode

nickstenning:
This change makes it possible to switch the `h` application into a mode where the authenticated user for a request is determined by the value of the X-Forwarded-User request header, rather than retrieved from the session.

This is useful for people who want to run h behind an authenticating reverse proxy which can be responsible for setting the header.

Unsurprisingly, turning this mode on in the wrong context would pose a major security risk, allowing anyone to impersonate anyone else simply by setting an HTTP header. As such, when `PROXY_AUTH` is set, the auth module prints a substantial warning to the logs.

### DIFF
--- a/h/auth/__init__.py
+++ b/h/auth/__init__.py
@@ -2,7 +2,10 @@
 
 """Authentication configuration."""
 
-from pyramid.authentication import SessionAuthenticationPolicy
+import logging
+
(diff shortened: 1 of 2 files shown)


chdorner:
LGTM.

PR #3639, opened 2016-07-29
Add a hook to disable all JS enhancements on the page

robertknight:
Adding `nojs=1` to the URL's query string parameters will disable all JS
enhancements.

This is useful for testing the behavior users will see when JS is slow
to load or otherwise broken across a variety of browsers without needing
to install browser-specific addons or change global browser settings.

We alternatively put this hook in the root site bundle (`site.js`). Putting it in `upgradeElements` has the slight advantage that it will be available in any projects using it.


nickstenning:
Nice.

PR #3640, opened 2016-07-29
Implement new form designs for logged out forms

robertknight:
This PR implements the new form styles that will be rolled out across the site and uses them for the logged-out pages, following the designs at https://trello.com/c/yt9gXTWA/391-logged-out-forms-updates-visual-changes-to-form-style . The new form styles are used by deform whenever the `activity_pages` flag is enabled.

In order to give us a clean slate for Activity Pages, I have created a new `partials-v2/` directory which contains only variables and component styles used by the new pages and the `site-v2.css` bundle now only references styles from that directory. This ensures that no legacy cruft unexpectedly ends up on pages as we migrate them to the new design.

Note that because the new form styles are used whenever the flag is enabled, but only a subset of pages have been migrated to use the new `site-v2` CSS bundle which contains the supporting CSS, other forms will be visually broken (though still usable) when the 'activity_pages' flag is enabled. Nick and I discussed this and I think we can live with it as we'll be migrating all of the pages very shortly.

When the activity_pages flag is disabled, everything should look as before except for a few validation message and form label changes which I didn't think were worth flagging.


seanh:
I think the logo should be horizontally aligned with the form on small screens. Looks like the "Log in" title might be out by a couple pixels (border width?) as well.

screenshot

seanh:
Look like the focused styles aren't working without javascript (you can't see my keyboard cursor in the screenshot but it's in the first text field):

screenshot from 2016-07-29 18-28-35

seanh:
Is it possible to use the "clicking now will let you start typing" mouse cursor whenever the cursor is over the "text area" (it's only showing when the cursor is over the invisible `<input>`)

untitled screencast - edited 32

seanh:
Related to the mouse cursor issue above, with JavaScript disabled if I click inside the "input field" border (but outside of the invisible `<input>`) it does not put the keyboard focus in the `<input>`, you have to actually click inside the (invisible) `<input>`.

It looks as if JavaScript is used to make clicks inside the input's container element, but not inside the input itself, focus the input. If you just make the `<input>` larger using padding, and then use absolute positioning to position the label and error messages inside the `<input>`, then both the mouse cursor style and the click to focus behaviour work correctly without JavaScript (see <https://github.com/hypothesis/h/pull/3600/files>)

seanh:
This is too big a diff to review last thing on a Friday, will take a closer look at the actual code next week (I'm off on Monday though)

robertknight:
> Instead of using a container element for the border, if you just make the <input> itself larger
> using padding, and then use absolute positioning to position the label and error messages inside
> the <input>, then the mouse cursor style will work correctly and the click to focus will work
> without JavaScript

I dismissed this initially because of the issues you encountered before where the validation message could overflow the container, or appear on top of the input if short, or the same could happen to the label. On reflection that's probably not going to be an issue provided we:

A) Ensure that the validation messages are always short and
B) Change the layout for mobile devices so that the validation messages do not appear on top of the input - eg. put them underneath.

I gave this approach another go this morning and you're right - its a simpler way to get the desired cursor and input tab behavior. We are still going to need some logic for the input forms in order to implement the character counter which appears in the input field in one or two other mockups but possibly none then for the basic focus handling.

seanh:
> B) Change the layout for mobile devices so that the validation messages do not appear on top of the input - eg. put them underneath.

That's a great point about mobile devices.

For larger screens, _if_ we want to support long error messages, would it be possible to do a JavaScript progressive enhancement that looks at the height of the error message's element, and if it's bigger than the input's top padding, increases the top padding?

That would leave the visual issue only occurring on non-mobile devices where the user has JavaScript disabled, where we've added a long error message, and the user has triggered it, which seems rare.

We _could_ use the alternative mobile device error-message-below-input style on non-mobile devices as well, only when JavaScript is disabled, and that would completely remove the visual error in all cases. But that definitely seems like polish and detail to me, in reality hardly anyone will ever see the error, literally no one will if we don't have any over-long error messages.

robertknight:
This has now been simplified as per your suggestion @seanh . Consequently, I have removed the controller for input fields entirely for this first cut. As I mentioned above, we are going to need some logic for form inputs to handle extended features such as the character count, but not for the features used by the logged out forms.

h/templates/deform/mapping_item.jinja2
```diff
@@ -1,36 +1,54 @@
...
-<div class="form-group form-field{% if field.error %} form-field-error{% endif %}"
-     {%- if field.description -%}
-     title="{{ _(field.description) }}"
-     {% endif %}
-     id="item-{{ field.oid }}">
-{% endif -%}
+{% if feature('activity_pages') and not field.widget.hidden %}
+  <div class="form-input {% if field.error %}is-error{% endif %}"
+       id="item-{{ field.oid }}">
+    <label class="form-input__label">{{_(field.title)}}</label>
```
robertknight: This is missing the `for` attribute

h/templates/deform/mapping_item.jinja2
```diff
@@ -1,36 +1,54 @@
-{%- if not field.widget.hidden -%}
-<div class="form-group form-field{% if field.error %} form-field-error{% endif %}"
-     {%- if field.description -%}
-     title="{{ _(field.description) }}"
-     {% endif %}
-     id="item-{{ field.oid }}">
-{% endif -%}
+{% if feature('activity_pages') and not field.widget.hidden %}
```
robertknight: In my initial implementation the structure of the HTML was different with and without the feature flag. That's no longer really the case, so I might be able to simplify this to just changing a few CSS classes depending on whether the feature flag is set.

h/static/images/logo.svg
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<svg width="24px" height="28px" viewBox="0 0 24 28" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
+    <!-- Generator: Sketch 3.8.3 (29802) - http://www.bohemiancoding.com/sketch -->
+    <title>Rectangle 2 Copy 17</title>
```
nickstenning: Perhaps run this through svgo?

h/accounts/schemas.py
```diff
@@ -246,11 +246,11 @@ class ResetPasswordSchema(CSRFSchema):
     # call it `user` because when validated, it will return a `User` object.
     user = colander.SchemaNode(
         ResetCode(),
-        title=_('Your reset code:'),
+        title=_('Reset code'),
```
nickstenning: So you've changed the form field titles on the login and reset password forms to be:

- terser
- colon-less

But you haven't done the same for the "register" or "forgot password" forms. I'm guessing this is an oversight...?

h/accounts/views.py
```diff
@@ -315,14 +315,14 @@ def _reset_password(self, user, password):
...
+        tos_link = ('<a class="link" href="/terms-of-service">' +
                     _('Terms of Service') +
                     '</a>')
-        cg_link = ('<a href="/community-guidelines">' +
+        cg_link = ('<a class="link" href="/community-guidelines">' +
                    _('Community Guidelines') +
                    '</a>')
         form_footer = _(
-            'You are agreeing to be bound by our {tos_link} and '
+            'You are agreeing to our {tos_link} and '
```
nickstenning: The length of this footer message causes the "Sign up" button to wrap onto two lines on small screens. We probably want to avoid that.

h/static/images/logo.svg
robertknight: I created a separate PR for this which minifies all SVGs in production builds https://github.com/hypothesis/h/pull/3642

I think it is useful to have the unoptimized version in-tree for diff-ability.

h/accounts/schemas.py
robertknight: Yes. I'm following the text shown in Conor's mocks.

nickstenning: I think you missed my point here, which was that you haven't made equivalent changes for two of the four forms.

robertknight: Sorry, I got the point, I meant that _Yes it was an oversight_ Fixing that now.

h/accounts/views.py
robertknight: Fixed

robertknight: There is a small remaining issue that the underline below links overlaps the line below slightly when the message wraps onto multiple lines. I'll come back to that it is very minor I want to run something by Conor.

nickstenning:
Small thing: "tab" ordering seems a bit off on these forms, in that the primary form button should probably come directly after the last required field. At the moment the ToS/Community Guidelines links (or the Forgot Password? link) are sooner.

nickstenning:
LGTM.

h/accounts/schemas.py
```diff
@@ -191,16 +191,16 @@ class RegisterSchema(CSRFSchema):
             unique_username,
             unblacklisted_username,
         ),
-        title=_('Username:'),
+        title=_('Username'),
         hint=_('between {min} and {max} characters, containing only letters, '
                'numbers, periods, and underscores').format(
```
seanh: When we remove the legacy forms I guess we should remove these hints, they aren't used anymore

nickstenning: Not quite: https://trello.com/c/gGtgaRg9/637-form-field-hint-enhancement

PR #3641, opened 2016-07-29
Enable telling devserver which processes to run

seanh:
Add options to the `hypothesis devserver` command to tell it which of
the processes you would like it to run and which not, instead of it
always running all the processes.

The problem with `hypothesis devserver` is that it doesn't work with pdb
(or any similar Python debugger). Because devserver multiplexes the
output of multiple processes into one terminal, while you were trying to
interact with pdb in say the web process output from the celery process
would be printed and ruin your pdb shell. Also, devserver doesn't know
which of its multiple processes to send the input you're typing to, so
you can't interact with pdb.

Now, if you want to use pdb, you can run `hypothesis devserver --no-web`
and then run the web process separately in another shell (by running the
`gunicorn` command directly) and use pdb as normal.

Or you could debug the celery worker process in the same way but with
`hypothesis devserver --no-worker`.

### DIFF
--- a/h/cli/commands/devserver.py
+++ b/h/cli/commands/devserver.py
@@ -12,7 +12,23 @@ import click
               default=False,
               is_flag=True,
               help='Serve HTTPS rather than plain HTTP.')
-def devserver(https):
+@click.option('--web/--no-web',
+              default=True,
+              help="Whether or not to run the Pyramid app process "
+                   "(default: --web).")
+@click.option('--ws/--no-ws',
+              default=True,
+              help="Whether or not to run the WebSocket process "
+                   "(default: --ws).")
+@click.option('--worker/--no-worker',
+              default=True,
+              help="Whether or not to run the Celery worker process "
+                   "(default: --worker).")
+@click.option('--assets/--no-assets',
+              default=True,
+              help="Whether or not to run the gulp watch process "
+                   "(default: --assets).")
+def devserver(https, web, ws, worker, assets):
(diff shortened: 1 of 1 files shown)


seanh:
I'm sure there's a less repetitive way to implement this using click's composability but this works

seanh:
```
Usage: hypothesis devserver [OPTIONS]

  Run a development server.

  This command will start a development instance of h, consisting of a web
  application, Celery worker, and websocket server. It will also start a
  process which will watch and build the frontend assets.

  By default, the webserver will be accessible at:

      http://localhost:5000

  You can also pass the `--https` flag, which will look for a TLS
  certificate and key in PEM format in the current directory, in files
  called:

      .tlscert.pem     .tlskey.pem

  If you use this flag, the webserver will be accessible at:

      https://localhost:5000

  If you wish this to be the default behaviour, you can set the USE_HTTPS
  environment variable.

Options:
  --https                 Serve HTTPS rather than plain HTTP.
  --web / --no-web        Whether or not to run the Pyramid app process
                          (default: --web).
  --ws / --no-ws          Whether or not to run the WebSocket process
                          (default: --ws).
  --worker / --no-worker  Whether or not to run the Celery worker process
                          (default: --worker).
  --assets / --no-assets  Whether or not to run the gulp watch process
                          (default: --assets).
  --help                  Show this message and exit.
```

nickstenning:
LGTM.

PR #3643, opened 2016-08-01
Replace "firstRun" config option with "openSidebar" and "openLoginForm"

nickstenning:
This commit deprecates the confusingly-named "firstRun" embed script
option with two which actually describe the relevant actions triggered.

I've left the "firstRun" option around for now, for old(er) clients. We
can remove this in due course, as noted in a template comment.


chdorner:
LGTM. Will try and get this out to production today without making a new release.

PR #3646, opened 2016-08-02
Shorten an over-long error message

### LINKED ISSUE #3465: Crash when two requests simultaneously create document metadata

seanh:
This error message was overrunning the border of its container element
in the HTML/CSS.

Fixes #3465.

### DIFF
--- a/h/accounts/schemas.py
+++ b/h/accounts/schemas.py
@@ -169,9 +169,7 @@ class ForgotPasswordSchema(CSRFSchema):
 
         if user is None:
             err = colander.Invalid(node)
-            err['email'] = _('We have no user with the email address '
-                             '"{email}". Correct this address or try '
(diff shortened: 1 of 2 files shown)


seanh:
I think we want to review error messages and maybe someone will decide what consistent voice and style they should all be in etc. For know "Unknown email address" matches the other message that this form generates "Invalid email address" and fits within the box, so I think this is worth merging now.

nickstenning:
As luck would have it we have a card to do just that, already: https://trello.com/c/gY8mnzZk/409-review-error-messages-generated-by-logged-out-forms-and-check-that-they-make-sense-with-the-new-form-designs

nickstenning:
LGTM.

PR #3647, opened 2016-08-02
Left-align h logo with form on new forms

### LINKED ISSUE #3644: Icon, title and form not left-aligned on new login form

seanh:
Fixes #3644

### DIFF
--- a/h/static/styles/partials-v2/_masthead.scss
+++ b/h/static/styles/partials-v2/_masthead.scss
@@ -3,7 +3,7 @@
 
 .masthead {
   padding-bottom: 0;
-  padding-left: 20px;
-  padding-right: 20px;
+  padding-left: 15px;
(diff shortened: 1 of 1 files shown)


seanh:
#3644 also points out that the "Log in" title also appears to be a couple of pixels out of alignment, but I think this might just be the font?

h/static/styles/partials-v2/_masthead.scss
```diff
@@ -3,7 +3,7 @@
 
 .masthead {
   padding-bottom: 0;
-  padding-left: 20px;
-  padding-right: 20px;
+  padding-left: 15px;
+  padding-right: 15px;
```
seanh: By the way, it looks like the `padding-right` isn't actually needed at all?

chdorner:
LGTM.

PR #3651, opened 2016-08-02
URL changes for reset password, forgot password and register.

sheetaluk:
https://trello.com/c/Yiraeqzs/408-logged-out-form-url-changes-and-redirects

Related to:
https://github.com/hypothesis/client/pull/68

Test scenarios:
- In /: create new user
- In acount/settings:
   changed email
   changed password
- In stream:
   logged out and logged in with new password, 
   forgot password and reset it

**N.B.** We will need to coordinate the deployment of this change carefully, as we may need to update our nginx configuration._​

### DIFF
--- a/h/accounts/views.py
+++ b/h/accounts/views.py
@@ -219,7 +219,7 @@ class ForgotPasswordController(object):
         self._send_forgot_password_email(user)
 
         return httpexceptions.HTTPFound(
-            self.request.route_path('reset_password'))
+            self.request.route_path('account_reset'))
 
     def _redirect_if_logged_in(self):
         if self.request.authenticated_userid is not None:
@@ -229,14 +229,14 @@ class ForgotPasswordController(object):
         serializer = self.request.registry.password_reset_serializer
         code = serializer.dumps(user.username)
 
-        link = reset_password_link(self.request, code)
-        message = reset_password_email(user, code, link)
+        link = account_reset_link(self.request, code)
+        message = account_reset_email(user, code, link)
         mailer.send.delay(**message)
 
 
-@view_defaults(route_name='reset_password',
-               renderer='h:templates/accounts/reset_password.html.jinja2')
-class ResetPasswordController(object):
+@view_defaults(route_name='account_reset',
+               renderer='h:templates/accounts/reset.html.jinja2')
+class ResetController(object):
 
     """Controller for handling password reset forms."""
 
@@ -245,7 +245,7 @@ class ResetPasswordController(object):
         self.schema = schemas.ResetPasswordSchema().bind(request=self.request)
         self.form = request.create_form(
             schema=self.schema,
-            action=self.request.route_path('reset_password'),
+            action=self.request.route_path('account_reset'),
             buttons=(_('Save'),))
 
     @view_config(request_method='GET')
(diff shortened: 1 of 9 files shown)


h/accounts/views.py
```diff
@@ -680,9 +680,9 @@ def dismiss_sidebar_tutorial(request):
 def includeme(config):
     config.add_route('login', '/login')
     config.add_route('logout', '/logout')
-    config.add_route('register', '/register')
+    config.add_route('signup', '/signup')
     config.add_route('activate', '/activate/{id}/{code}')
-    config.add_route('forgot_password', '/forgot_password')
+    config.add_route('forgot-password', '/forgot-password')
```
chdorner: It looks like we tend to use underscores for route names (the first argument).

chdorner: The failing tests will also be fixed when you change the route name back to `forgot_password`.

h/accounts/views.py
```diff
@@ -245,7 +245,7 @@ def __init__(self, request):
         self.schema = schemas.ResetPasswordSchema().bind(request=self.request)
         self.form = request.create_form(
             schema=self.schema,
-            action=self.request.route_path('account_reset'),
+            action=self.request.route_path('account/reset'),
```
chdorner: `request.route_path` [takes the route name as the first argument](http://docs.pylonsproject.org/projects/pyramid/en/latest/api/request.html#pyramid.request.Request.route_path), so this should be `account_reset`.

h/accounts/views.py
```diff
@@ -219,7 +219,7 @@ def post(self):
         self._send_forgot_password_email(user)
 
         return httpexceptions.HTTPFound(
-            self.request.route_path('reset_password'))
+            self.request.route_path('account/reset'))
```
chdorner: `request.route_path` [takes the route name as the first argument](http://docs.pylonsproject.org/projects/pyramid/en/latest/api/request.html#pyramid.request.Request.route_path), so this should be `account_reset`.

h/accounts/views.py
```diff
@@ -245,15 +245,15 @@ def __init__(self, request):
         self.schema = schemas.ResetPasswordSchema().bind(request=self.request)
         self.form = request.create_form(
             schema=self.schema,
-            action=self.request.route_path('reset_password'),
+            action=self.request.route_path('account/reset'),
```
chdorner: Same as above, this needs to be `account_reset`.

h/accounts/views.py
```diff
@@ -72,7 +72,7 @@ class AuthController(object):
 
     def __init__(self, request):
         form_footer = '<a class="link" href="{href}">{text}</a>'.format(
-            href=request.route_path('forgot_password'),
+            href=request.route_path('forgot-password'),
```
chdorner: This needs to be `forgot_password` as well.

h/accounts/views.py
```diff
@@ -649,9 +649,9 @@ def reset_password_email(user, reset_code, reset_link):
     }
 
 
-def reset_password_link(request, reset_code):
+def account_reset_link(request, reset_code):
     """Transform an activation code into a password reset link."""
```
chdorner: We might want to change the wording in the docstring as well to say "account reset link".

chdorner:
This is looking great! :+1: 
I will take some time either later today or otherwise tomorrow morning to do some digging in case there are some places we forgot to change in h and the client. Will merge it after that

chdorner:
I'll merge this now. As Nick mentioned on Slack, it would be nice to have the `reset_password` route be redirected as well but that will include some changes to the redirecting tween. We can do that change in a separate PR.

Issue #3631, opened 2016-07-27
Double error message when trying to register account with blacklisted username


seanh:
### Steps to reproduce

1. To to `/register`
2. Enter `twitter` as the username for the new account, and enter a valid email and password
3. Submit the form

### Expected behaviour

A single error message. "An account with this username already exists" is not really correct but will do.

### Actual behaviour

"An account with this username already exists" error message is shown twice.

![screenshot from 2016-07-27 12-31-27](https://cloud.githubusercontent.com/assets/22498/17174040/62cfcd50-53f6-11e6-869e-1a44229a0c02.png)

This does not seem to happen when trying to register an account for a non-blacklisted username where an account with the username actually does already exist - you'll get the error message only once.

nickstenning:
Ok, so the problem here is really that because of the username "pre-registration" that we did before launch, we do actually have some users whose usernames are in the blacklist.

To wit, we have:

- 26 such users
- 25 of whom have "activated" accounts
- 12 of whom have set passwords
- 3 of whom have actually annotated
- totalling 6 annotations

I suggest we review this list and:

- remove users who have aren't activated or have no password set
- remove items from the blacklist that don't need to be there (there are certainly a couple of candidates for that in this list) or where we're prepared to accept the risk of impersonation for a username

Hopefully that way we can get to a state where we have no extant users who are in the blacklist, which would eliminate this problem.

robertknight:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;gY8mnzZk&#x2F;409-review-error-messages-generated-by-logged-out-forms-and-check-that-they-make-sense-with-the-new-form-designs"><img src="https:&#x2F;&#x2F;github.trello.services&#x2F;images&#x2F;trello-icon.png" width="12" height="12"> Review error messages generated by logged-out forms and check that they make sense with the new form designs</a>

lenazun:
@chdorner was this fixed by #3652?

chdorner:
I think #3652 just removed some usernames from the blacklist, we would still have this issue with other usernames? Correct me if I'm wrong @seanh and @nickstenning

PR #3652, opened 2016-08-02
Remove some uids from the blacklist

seanh:


chdorner:
What's the reason for removing these potential usernames from the blacklist?

chdorner:
As discussed in Slack:

> With those usernames that I was removing from the blacklist (`anon`, `atom`, `beta`...) we already have users in the prod db with those usernames, and the accounts are activated and have passwords. It causes this issue https://github.com/hypothesis/h/issues/3631 There are some more user accounts with blacklisted uids in the prod db as well that aren't activated or don't have a password set, I think the plan was to delete those from the db

chdorner:
LGTM.

PR #3653, opened 2016-08-03
Simplify account controller

seanh:
These refactorings are going to make it easier to do <https://trello.com/c/MxOJSA6b/413-return-form-html-fragments-for-ajax-requests>. I want to make use of Pyramid view config predicates to select different view callables when the forms are submitted via Ajax, and this makes the existing code consistent with that approach.

### DIFF
--- a/h/accounts/views.py
+++ b/h/accounts/views.py
@@ -495,33 +495,40 @@ class AccountController(object):
         """Show the user's account."""
         return self._template_data()
 
-    @view_config(request_method='POST')
-    def post(self):
-        """Handle POST payload from account update form."""
-        formid = self.request.POST.get('__formid__')
-        if formid is None or formid not in self.forms:
-            raise httpexceptions.HTTPBadRequest()
+    @view_config(request_method='POST',
+                 request_param='__formid__=email')
+    def post_email_form(self):
+        """Called by Pyramid when the change email form is submitted."""
+        return self.handle_form_submission(
+            self.forms['email'],
+            on_success=self.update_email_address)
 
+    @view_config(request_method='POST',
+                 request_param='__formid__=password')
+    def post_password_form(self):
+        """Called by Pyramid when the change password form is submitted."""
+        return self.handle_form_submission(
+            self.forms['password'],
+            on_success=self.update_password)
+
+    def update_email_address(self, appstruct):
+        self.request.authenticated_user.email = appstruct['email']
+
+    def update_password(self, appstruct):
+        self.request.authenticated_user.password = appstruct['new_password']
+
+    def handle_form_submission(self, form, on_success):
+        """Handle the submission of the given form."""
         try:
-            if formid == 'email':
-                self._handle_email_form()
-            elif formid == 'password':
-                self._handle_password_form()
+            appstruct = form.validate(self.request.POST.items())
         except deform.ValidationFailure:
             return self._template_data()
 
+        on_success(appstruct)
         self.request.session.flash(_("Success. We've saved your changes."),
                                    'success')
         return httpexceptions.HTTPFound(
             location=self.request.route_url('account'))
-
-    def _handle_email_form(self):
-        appstruct = self.forms['email'].validate(self.request.POST.items())
-        self.request.authenticated_user.email = appstruct['email']
-
-    def _handle_password_form(self):
-        appstruct = self.forms['password'].validate(self.request.POST.items())
-        self.request.authenticated_user.password = appstruct['new_password']
 
     def _template_data(self):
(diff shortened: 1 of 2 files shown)


nickstenning:
I've cherry-picked 1202c82e1ff6f8dd590708f34d2267cfb34246a0 onto master, as that's pretty uncontroversial. Going to rebase this to look at the rest.

h/accounts/views.py
```diff
@@ -495,34 +495,49 @@ def get(self):
...
 
+    def _handle_email_form(self):
         try:
-            if formid == 'email':
-                self._handle_email_form()
-            elif formid == 'password':
-                self._handle_password_form()
+            appstruct = self.forms['email'].validate(self.request.POST.items())
+        except deform.ValidationFailure:
+            raise
```
nickstenning: I think I understand why you've done this, but to me this just adds unnecessary noise to understanding what this function does.

One of the nicest features of Pyramid (IMO) is the exception view, which allows you to move error handling out of the main flow of the code, so that understanding what the code is supposed to do in the "happy path" case is as easy as possible. You almost seem to be going down the opposite route here...

In general, I think something like this is what we should be shooting for:

```python
class FormController(object):

    def __init__(self, context, request):
        self.context = context
        self.request = request
        self.form = request.create_form(...)

    @view_config(request_method='GET')
    def get(self):
        return {'form': self.form.render()}

    @view_config(request_method='POST')
    def post(self):
        appstruct = self.form.validate(self.request.POST.items())

        # do stuff with appstruct

        return HTTPFound(location=request.path)

    @view_config(context=deform.ValidationFailure)
    def validation_failure(self):
        return {'form': self.context.render()}
```

The important point here is not that I'm using an exception view per se, but that the error-handling code is a) generic (it will work for any views in this class) and b) distinct from the view code itself, making both the happy path and the error handling code easier to test.

In this particular context (the account controller) it's made a little more complicated by a variety of factors (multiple forms, additional context, and so on), but whether it's an exception view, a decorator, or even just a helper function, we should be trying to make it possible to write correct form handlers across the board, whereas I'm worried that these changes just make it harder to understand what's happening, and are closely tied to this particular controller.

Perhaps an even simpler approach would be a `validate_form` helper function, something like:

```python
class validate_form(request, form):
    try:
        appstruct = form.validate(request.POST.items())
    except deform.ValidationFailure:
        return None, {'form': form.render()}
    else:
        return appstruct, HTTPFound(location=request.path)
```

which you could then use as:

```
@view_config(request_method='POST', request_param='__formid__=email')
def post_email(self):
    appstruct, response = validate_form(self.request, self.forms['email'])
    if appstruct is not None:  # form was valid
        # update email
    else:
        response.update({'email': request.authenticated_user.email})
    return response
```

This relies on `response` having different types when the form is valid vs invalid, but I don't think that's so bad...

seanh: In the case of this particular page it seems to be awkward to use an exception view. The reason is that when a ValidationFailure occurs we need to render two forms:

1. The form that was submitted and failed to validate, with validation errors. If it was the email form that was submitted, for example, then in the view callable `self.forms['email'].render()` will render the email form with validation errors. But in the exception view the same code will render the form _without_ validation errors. Fortunately `self.context.render()` in the exception view will render the submitted form with validation errors, but...

2. We also need to return the other form to the template. This is either `self.forms['email'].render()` or `self.forms['password'].render()` but the exception view doesn't know which. There are ways that the exception view can find out which, or that the main view callable can tell the exception view which, but I think they're all a little awkward.

So I don't think it's possible to _completely_ separate happy path from error handling code, unless this behaviour of deform with exception views can be fixed (I haven't looked into that yet).

So forgetting the exception view for now...

I believe the ValidationFailure error handling as I had it in the _post() method was already generic in that the same code could handle a ValidationFailure from the validation of any form:

```python
try:
    handler()
except deform.ValidationFailure:
    return self._template_data()
```

But there was duplication between the multiple handler methods of the logic to validate the posted data against the form and do the action (change the email or password) if it's valid, this was duplicated in `try ... except ... else ... ` clauses in `_handle_password_form()` and `_handle_email_form()`. I've moved this logic into `_post()` to remove the duplication.

I've also renamed `_post()` to `handle_form_submission()` which I hope is a bit clearer.

This is similar to your `validate_form()` except that instead of returning either an `(appstruct, redirect)` tuple or `(None, template_data)` tuple, it takes an `on_success()` callable as argument. I think this is simpler to use and duplicates less code between the email and password forms.

Of course this does not completely separate happy path from error handling code. Those are slightly mixed together in `handle_form_submission()`. But without the exception view I think this might be the best we can do.

nickstenning:
LGTM.

PR #3655, opened 2016-08-03
Enable submitting accounts forms via xhr

seanh:
Depends on https://github.com/hypothesis/h/pull/3653

### DIFF
--- a/h/accounts/views.py
+++ b/h/accounts/views.py
@@ -58,6 +58,41 @@ def to_xhr_response(request, non_xhr_result, form):
     return form.render()
 
 
+def handle_form_submission(request, form, on_success, on_failure):
+    """
+    Handle the submission of the given form in a standard way.
+
+    :param request: the Pyramid request
+
+    :param form: the form that was submitted
+    :type form:
+
+    :param on_success: A callback function to be called if the form validates
+        successfully. This function should carry out the action that the form
+        submission requests. For example for a change password form, this
+        function would change the user's password.
+    :type on_success: callable
+
+    :param on_failure: A callback function that will be called if form validation
+        fails in order to get the view callable result that should be returned.
+        Note that the result returned by on_failure() will *not* be used if the
+        request is an XHR request.
+    :type on_failure: callable
+
+    """
+    try:
+        appstruct = form.validate(request.POST.items())
+    except deform.ValidationFailure:
+        result = on_failure()
+    else:
+        on_success(appstruct)
+        request.session.flash(_("Success. We've saved your changes."),
+                              'success')
+        result = httpexceptions.HTTPFound(location=request.url)
+
+    return to_xhr_response(request, result, form)
+
+
 @json_view(context=BadCSRFToken)
 def bad_csrf_token(context, request):
     request.response.status_code = 403
@@ -519,38 +554,27 @@ class AccountController(object):
                  request_param='__formid__=email')
     def post_email_form(self):
         """Called by Pyramid when the change email form is submitted."""
-        return self.handle_form_submission(
+        return handle_form_submission(
+            self.request,
             self.forms['email'],
-            on_success=self.update_email_address)
+            on_success=self.update_email_address,
+            on_failure=self._template_data)
 
     @view_config(request_method='POST',
                  request_param='__formid__=password')
     def post_password_form(self):
         """Called by Pyramid when the change password form is submitted."""
-        return self.handle_form_submission(
+        return handle_form_submission(
+            self.request,
             self.forms['password'],
-            on_success=self.update_password)
+            on_success=self.update_password,
+            on_failure=self._template_data)
 
     def update_email_address(self, appstruct):
         self.request.authenticated_user.email = appstruct['email']
 
     def update_password(self, appstruct):
         self.request.authenticated_user.password = appstruct['new_password']
-
-    def handle_form_submission(self, form, on_success):
-        """Handle the submission of the given form."""
-        try:
-            appstruct = form.validate(self.request.POST.items())
-        except deform.ValidationFailure:
-            result = self._template_data()
-        else:
-            on_success(appstruct)
-            self.request.session.flash(_("Success. We've saved your changes."),
-                                       'success')
-            result = httpexceptions.HTTPFound(
-                location=self.request.route_url('account'))
-
-        return to_xhr_response(self.request, result, form)
 
     def _template_data(self):
         """Return the data needed to render accounts.html.jinja2."""
--- a/h/accounts/views.py
+++ b/h/accounts/views.py
@@ -50,6 +50,14 @@ def to_xhr_response(request, non_xhr_result, form):
     should be the result that the view callable would normally return if this
     were not an XHR request.
 
+    :param request: the Pyramid request
+
+    :param non_xhr_result: the view callable result that should be returned if
+        ``request`` is *not* an XHR request
+
+    :param form: the form that was submitted
+    :type form: deform.form.Form
+
     """
     if not request.is_xhr:
         return non_xhr_result
@@ -65,7 +73,7 @@ def handle_form_submission(request, form, on_success, on_failure):
     :param request: the Pyramid request
 
     :param form: the form that was submitted
-    :type form:
+    :type form: deform.form.Form
 
     :param on_success: A callback function to be called if the form validates
         successfully. This function should carry out the action that the form
--- a/h/accounts/views.py
+++ b/h/accounts/views.py
@@ -38,6 +38,26 @@ def ajax_payload(request, data):
     return payload
 
 
+def to_xhr_response(request, non_xhr_result, form):
+    """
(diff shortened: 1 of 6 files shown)


nickstenning:
I've added some comments to #3653. If possible I'd like us to find a generic solution rather than having to double up on the number of views.

Options here seem to include view decorators, or even [view derivers](http://docs.pylonsproject.org/projects/pyramid/en/latest/narr/hooks.html#view-derivers) if we upgrade to Pyramid 1.7.

seanh:
Alright. The code and history needs some cleanup, but as illustration...

The awkward thing here is that the original view callables return different things depending on whether the form submission was valid or not:

1. For an invalid form submission the form HTML page needs to be re-rendered with validation errors. So the view callable returns a dict of the template data needed to render the page.

2. For a valid form submission the view callable returns an `HTTPFound`.

In both cases if the request is an XHR one we want to render a rendering of just the `<form>` element (possibly with different HTTP statuses depending on whether the form submission succeeded or not).

The dict that the view callable returns in case 1 happens to already contain the form that was submitted, so our to-xhr helper could simply extract the form from the dict that the view callable was already returning anyway. This requires the helper to know what the key of the form in the returned dict is, which varies from view to view, but that can be passed an an argument to the decorator (`@to_xhr_result('email_form')`.

But that doesn't work for case 2 where the view callable just returns an `HTTPFound`.

And we don't want the original view callables to have to check whether the request is XHR or not - that would defeat the point of having a reusable helper for this.

So basically I think we need to require the original view callables to return a `(form, result)` tuple where `form` is the deform form that needs to be rendered in the case of an XHR request and `result` is the result (either template data dict or `HTTPFound`) that should be returned if this is not an XHR request.

<https://github.com/hypothesis/h/pull/3655/commits/07c0c75f6f22b440af0f89f5f02ead2674d8616f> implements a version of this that uses a helper function that view callables call. The essence of the pattern is:

<pre>
def to_xhr_result(request, non_xhr_result, form):
    if not request.is_xhr:
        return non_xhr_result

    request.response.text = form.render()
    request.response.content_type = 'text/plain'
    return request.response

@view_config(...)
def my_view_callable(request):
    non_xhr_result = generate_normal_result_in_normal_way(request)
    form = get_the_email_address_form_or_the_password_form()
    <strong>return to_xhr_result(request, non_xhr_result, form)</strong>
</pre>

<https://github.com/hypothesis/h/pull/3655/commits/da9d9e7875ba516945a4d231538a0e099321b8ae> implements the same thing but using a decorator instead of a helper function. In this case the pattern is:

<pre>
def to_xhr_result(view_callable):
    def replacement_view_callable(self):
        form, non_xhr_result = view_callable(self)

        if not self.request.is_xhr:
            return non_xhr_result

        self.request.response.text = form.render()
        self.request.response.content_type = 'text/plain'
        return self.request.response
    return replacement_view_callable

class MyViewClass:

    ...

    <strong>@to_xhr_result</strong>
    @view_config(...)
    def my_view_callable(self):
        non_xhr_result = generate_normal_html_response_in_normal_way(request)
        form = get_the_email_address_form_or_the_password_form()
        <strong>return (form, non_xhr_result)</strong>
</pre>

In both cases I've moved the `form.render()` calls from the view code into the template, this is one simple way to make sure that only the form(s) that the response actually needs get rendered, and no form template is rendered twice for a single response, or rendered but then not used in the response.

Personally I think I prefer the simple helper function solution. It only requires each view callable to be changed in one place, the `return` statement, whereas the decorator requires both adding the `@decorator` statement and changing the `return` statement.

Also the dumb helper function can be called in the same way by view functions that take a `request` argument, view functions that take `request` and `context`, and view methods that take `self`. I think you might need three different decorators (or one smart one) for this. I'd guess that Pyramid 1.7 view derivers would handle that for us, but I still prefer the simple function for the other reasons.

A function is also just simpler and easier to understand than a decorator or view deriver.

Perhaps I've failed to see the benefit of moving to a decorator (and then when we upgrade to Pyramid 1.7, to a view deriver), but my proposal would be to tidy up the helper function version and then remove WIP from this...

seanh:
Alright, this is ready for review.

Now that we have `handle_form_submission()` which is a helper function that any form post view can call to handle a form submission correctly, there's only one place where we need to add XHR support.

I've written `to_xhr_response()` as a separate function anyway, just for the sake of keeping different things separate.

This still lives in `h/accounts/views.py` for now, but I think the next step will be to get the new group form to use it, so that'll involve moving it to a more generic place and if necessary making the code generic, but I'd like to leave that for a later pr. Nonetheless I didn't like how `handle_form_submission()` was a class method but `to_xhr_response()` was a function, so I've gone to the trouble of extracting `handle_form_submission()` into a normal function as well, which required making it a little more generic.

I've made it so that after a successful form submission `handle_form_submission()` always redirects to the same page that the original form submission came from (`httpexceptions.HTTPFound(location=request.url)`). For now this is all we need, but we can always add an optional `redirect_url=None` to `handle_form_submission()` later if we want to.

h/accounts/views.py
```diff
@@ -38,6 +38,70 @@ def ajax_payload(request, data):
     return payload
 
 
+def to_xhr_response(request, non_xhr_result, form):
```
nickstenning: Both this and `handle_form_submission` seem to me to be generic -- not specific to `h.accounts`. Perhaps you could put them in `h.form` and import them here?

h/accounts/views.py
```diff
@@ -38,6 +38,70 @@ def ajax_payload(request, data):
...
+        ``request`` is *not* an XHR request
+
+    :param form: the form that was submitted
+    :type form: deform.form.Form
+
+    """
+    if not request.is_xhr:
+        return non_xhr_result
+
+    request.response.body = form.render().encode()
```
nickstenning: Rather than manually encoding the form content, I think it would probably be better to do something like:

```python
request.override_renderer = 'text'
return form.render()
```

That way details of text encoding, content-type, etc., are left up to Pyramid.

tests/h/accounts/views_test.py
```diff
@@ -994,6 +1143,18 @@ def pyramid_request(self, pyramid_request, fake_db_session):
         return pyramid_request
 
 
+class RedirectMatcher(object):
+    """Matches any HTTPFound redirect to the given URL."""
```
nickstenning: The name doesn't match the behaviour here. Either this should match all redirects (`isinstance(other, httpexceptions.HTTPRedirection`) or it should be called something like `redirect_302_to`.

For consistency with other matcher objects I'd suggest giving it a `lower_underscored` name.

nickstenning:
Tested with curl and seems to work nicely!

h/accounts/views.py
seanh: I think the next step will be to get the new group form to use this, so that'll involve moving it to a more generic place and if necessary making the code generic, but I'd like to leave that for a later pr.

tests/h/accounts/views_test.py
seanh: done

h/accounts/views.py
seanh: done

h/accounts/views.py
nickstenning: Ok. I'm not really sure why they couldn't have moved in this PR, as it seems like a fairly straightforward change... but I'm not going to kick up a fuss about it.

tests/h/accounts/views_test.py
```diff
@@ -59,6 +60,145 @@ def invalid_form(errors=None):
...
+        If ``request`` isn't an XHR request it returns ``non_xhr_result``.
+
+        The calling view callable passes in the result that it would have
+        returned normally if this were not an XHR request as the
+        ``non_xhr_result`` argument. If the given ``request`` is not an XHR
+        request then ``non_xhr_result`` should just be returned unmodified.
+
+        """
+        pyramid_request.is_xhr = False
+        non_xhr_result = mock.Mock()
```
nickstenning: In general, when you're just passing values around, it can often be quicker/clearer to use `mock.sentinel.SOMETHING` rather than constructing a `mock.Mock()` object. It has the advantage that the test will fail if you access an attribute on the object, making it clear that it's no longer being used as an opaque value.

tests/h/accounts/views_test.py
```diff
@@ -59,6 +60,145 @@ def invalid_form(errors=None):
...
+
+    def test_returns_form_if_xhr(self, pyramid_request):
+        """
+        If ``request`` is an XHR request it should return the rendered ``form``.
+
+        It should return ``form`` rendered to a ``<form>`` element HTML snippet.
+
+        """
+        pyramid_request.is_xhr = True
+        form = mock.Mock()
```
nickstenning: Probably good to get into the habit of limiting the interface you're mocking using `spec_set`:

```python
form = mock.Mock(spec_set=['render'])
```

nickstenning:
LGTM.

PR #3656, opened 2016-08-04
Document bucket data

chdorner:
This PR changes the list of annotations in a document bucket to be a `DocumentBucket` object, which in turn contains a list of annotations. And other data like a unique set of tags and users, the URI and the domain of the document. This data is then displayed in the activity pages skeleton.


chdorner:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;LOtREQsy&#x2F;415-backend-support-for-document-component"> Backend support for Document component</a>

tests/h/activity/bucketing_test.py
```diff
@@ -0,0 +1,376 @@
...
+                                            document=document)
+        db_session.add(title_meta)
+        db_session.flush()
+
+        bucket = bucketing.DocumentBucket(document)
+        assert bucket.title == 'The Document Title'
+
+    def test_init_extracts_the_first_http_uri(self, db_session, document):
+        docuri_pdf = factories.DocumentURI(uri='urn:x-pdf:fingerprint',
+                                           updated=datetime.datetime(2016, 05, 02),
```
nickstenning: Careful with numeric literals. `05` is octal in Python 2.x.

h/activity/views.py
```diff
@@ -54,13 +54,11 @@ def eager_load_documents(query):
         timeframes = bucketing.bucket(anns)
 
         for timeframe in timeframes:
-            for document, annotations in timeframe.document_buckets.items():
-                results = []
-                for annotation in annotations:
+            for document, bucket in timeframe.document_buckets.items():
+                for index, annotation in enumerate(bucket.annotations):
```
nickstenning: Yeesh. No need to fix this now, but obviously this can't be how we fetch groups for real.

h/templates/activity/search.html.jinja2
```diff
@@ -36,11 +36,18 @@ Users (first ten):
     <li>
       <p><strong>{{ timeframe.label }}</strong></p>
       <ol>
-        {% for document, results in timeframe.document_buckets.items() %}
+        {% for _, bucket in timeframe.document_buckets.items() %}
```
nickstenning: Tiny thing, but `for bucket in timeframe.document_buckets.values()` would have the same effect and you wouldn't have to discard the documents.

h/activity/views.py
chdorner: Yup, of course. I think as soon as the frontend for this page is underway we should look at the views code, and also add tests (there's only one at the moment checking for the feature flag).

nickstenning:
LGTM.

PR #3658, opened 2016-08-05
Implement static parts of new nav bar

robertknight:
This implements the static parts of the new nav bar for accounts pages and Activity Pages as described at https://trello.com/c/atjT8T9p/395-logged-in-navbar-component

Icons are rendered using inline SVGs using the approach described at https://github.com/blog/2112-delivering-octicons-with-svg and for pretty much the same reasons (especially the ability to style parts of icons on the fly with CSS). On-the-fly icon coloring relies on SVG icons using the `currentColor` value for fills and strokes so that they use the CSS `color` property value from the containing `<svg>` element.

The dynamic parts (the drop-down menus for the 'Groups' link and 'Settings' icon) will follow in subsequent PRs.

The SVG icon helper uses `lxml.etree` instead of `xml.etree` to benefit
from support for namespace prefixes and avoid the problem described at http://stackoverflow.com/questions/8983041

----

- [x] Review the SVG icon helper implementation
- [x] Fix tooltips for SVG icons


h/assets.py
```diff
@@ -132,6 +134,22 @@ def _load_bundles(fp):
 assets_client_view = _add_cors_header(assets_client_view)
 
 
+def svg_icon(request, name, css_class=''):
```
robertknight: There are a few things that should be fixed here:

- This probably shouldn't be a method on the `request` object, since it isn't tied to a request in any way
- If the SVG has a `<title>` tag, that gets used as the tooltip for the icon. We should either strip those out when building the SVGs or remove them here
- Caching of the parsed XML content is missing

robertknight:
I continued working on a new branch based on this in order to implement the dynamic parts of the navbar but it soon became clear that just using the existing `{% include %}` mechanism to include the navbar on different pages was going to get ugly and that it would be better to find a way to create reusable widgets which could be re-used across different pages on the site.

I'm going to close this and open a new PR which uses that approach.

PR #3659, opened 2016-08-08
Fixed DetachedInstanceError in exception view template

seanh:
Exception views are called after the request's transaction has been
aborted due to an exception. If a template rendered by the exception
view then tried to call feature() it would crash with a
DetachedInstanceError as the feature flags client tries to load its
feature flags which involves accessing lazy-loaded properties of
request.authenticated_user which can no longer be loaded because the db
transaction has been aborted.

Forcing the feature flags client to load and cache all of its feature
flags when the template environment is setup ensures that these
lazy-loaded properties have already been loaded by the time any
templates get rendered, so accessing them on a detached instance is fine
because they no longer need to be read from the db.

### DIFF
--- a/h/features/__init__.py
+++ b/h/features/__init__.py
@@ -4,6 +4,8 @@ from __future__ import unicode_literals
 
 import os
 
+from pyramid.events import NewRequest
+
 from h import db
 from h.features import client
 from h.features import models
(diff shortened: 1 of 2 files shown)


nickstenning:
Making the change where you have will make for very surprising debugging if this issue crops up again in the future. 

Requests that have forms to handle will work fine, but requests that don't will suffer from the same problem you describe. 

Why not change the `h.features` module instead so that all users of that code have the same experience?

seanh:
Ok, I've changed this to happen on `NewRequest`, and also made `load()` public and had it call that instead of just calling `all()` and throwing away the result.

I've tested that this still fixes the crash I was seeing (when an exception view tries to render a template that accesses a feature flag).

Both `enabled()` and `all()` check whether `_cache` is `None` and if so, call `load()`. Now that we're calling `load()` on `NewRequest` anyway, the only place that I can see that this behaviour is needed is in `h/celery.py` which calls `clear()` before each task. If we just changed that to `load()` instead, then I think we could get rid of `clear()`, and get rid of the `load()` calls in `enabled()` and `all()`.

nickstenning:
LGTM.

PR #3660, opened 2016-08-08
Update account settings forms styling: Notifications and navbar

finitud:
https://trello.com/c/G3kr1awd/392-account-settings-profile-etc-forms-updates

### DIFF
--- a/h/templates/accounts/notifications.html.jinja2
+++ b/h/templates/accounts/notifications.html.jinja2
@@ -1,13 +1,21 @@
 {% extends "h:templates/layouts/account.html.jinja2" %}
 
+{% if not feature('activity_pages') %}
+{% set uses_legacy_styles = True %}
+{% endif %}
+
 {% set page_route = 'account_notifications' %}
 {% set page_title = 'Notifications' %}
 
 {% block page_content %}
-  <div class="form-vertical">
-    <h2 class="form-heading">
-      <span>{% trans %}Update your notification settings{% endtrans %}</span>
-    </h2>
+  {% if feature('activity_pages') %}
     {{ form }}
-  </div>
+  {% else %}
+    <div class="form-vertical">
+      <h2 class="form-heading">
+        <span>{% trans %}Update your notification settings{% endtrans %}</span>
+      </h2>
+      {{ form }}
+    </div>
+  {% endif %}
 {% endblock page_content %}
--- a/h/templates/layouts/account.html.jinja2
+++ b/h/templates/layouts/account.html.jinja2
@@ -18,23 +18,45 @@
   {% endif %}
   <div class="content paper">
     {% include "h:templates/includes/logo-header.html.jinja2" %}
-    <ul class="nav nav-tabs">
-      {% for route, title in nav_pages %}
-        <li{% if route == page_route %} class="active"{% endif %}>
-            <a href="{{ request.route_url(route) }}">{{ title }}</a>
-        </li>
-      {% endfor %}
-    </ul>
-
-    {% if request.session.peek_flash('success') -%}
-    <div class="form-flash">
-      {% for message in request.session.pop_flash('success') %}
-        <p>{{ message }}</p>
-      {%- endfor %}
-    </div>
-    {%- endif %}
-
-    {% block page_content %}
-    {% endblock page_content %}
+    {% if feature('activity_pages') %}
+      <div class="form">
+        <nav class="tabs">
+          <ul>
+            {% for route, title in nav_pages %}
+              <li class="tabs__item">
+                  <a href="{{ request.route_url(route) }}"
+                     class="tabs__link{% if route == page_route %} is-active{% endif %}">
+                    {{ title }}
+                  </a>
+              </li>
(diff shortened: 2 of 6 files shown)


h/templates/accounts/notifications.html.jinja2
```diff
@@ -1,13 +1,23 @@
...
 
 {% block page_content %}
-  <div class="form-vertical">
-    <h2 class="form-heading">
-      <span>{% trans %}Update your notification settings{% endtrans %}</span>
-    </h2>
-    {{ form }}
-  </div>
+  {% if feature('activity_pages') %}
+    <div class="form content">
```
finitud: Would this be `form--content` or `form__content`?

h/templates/layouts/account.html.jinja2
```diff
@@ -15,14 +15,27 @@
...
-        <li{% if route == page_route %} class="active"{% endif %}>
-            <a href="{{ request.route_url(route) }}">{{ title }}</a>
-        </li>
-      {% endfor %}
-    </ul>
-
+    {% if feature('activity_pages') %}
+      <div class="form content">
+        <h2 class="form-header">
+          <ul class="form-nav">
```
finitud: So `form--content__nav` / `form--content__header` I'm guessing, which answers my previous question. Do we actually need this `h2` element?

h/templates/accounts/notifications.html.jinja2
robertknight: `form` and `content` are two separate blocks here. In older templates, the `content` class was used to denote the main content of the page. We don't actually style this class at all in the new CSS at the moment, so I would suggest that we just leave it out for the moment.

finitud: So just `form` and then `form__nav` or `form__header` below?

h/templates/layouts/account.html.jinja2
robertknight: No. It might sense to make this a `<nav>` element, since this is a list of navigational links.

h/templates/accounts/notifications.html.jinja2
robertknight: Yes, that would be fine. Whether to use say `form__nav` vs. `form-nav` depends on whether the tabs are something that could be re-used on its own or whether it only makes sense in the context of a parent `form`. https://en.bem.info/methodology/faq/#a-block-or-an-element-when-should-i-use-which explains this a bit more.

The idea is to make it clear from the class name whether a class can be used on its own or whether it needs to live inside some other block.

h/templates/layouts/account.html.jinja2
```diff
@@ -15,23 +15,45 @@
...
-      {% for message in request.session.pop_flash('success') %}
-        <p>{{ message }}</p>
-      {%- endfor %}
-    </div>
-    {%- endif %}
-
-    {% block page_content %}
-    {% endblock page_content %}
+    {% if feature('activity_pages') %}
+      <div class="form">
```
nickstenning: So we're using `<div class="form">` in order to get the horizontal alignment right here, but the navigation is clearly not a form.

I'm going to defer to @robertknight on the right way to solve this one, but IMO we definitely shouldn't be using CSS classes which are named "form" for things that aren't forms.

nickstenning:
Visually this looks good, but I'd like to suggest a change to the markup/CSS for the checkbox, because at the moment the tappable area isn't as large as the surrounding grey box, which is IMO misleading.

If you were do to this instead:

```html
<div class="form-checkbox">
    <label for="deformField2-0" class="form-checkbox__label">
      <input class="form-checkbox__input" type="checkbox" name="checkbox" value="reply" id="deformField2-0"> Email me when someone replies to one of my annotations.
    </label>
  </div>
```

then with CSS like this

```css
.form-checkbox {
  ...
  padding: 0
}

.form-checkbox__label {
  ...
  padding: 15px;
  display: inline-block;
  cursor: pointer;
}

.form-checkbox__input {
  ...
  margin-right: 10px;
}
```

You'll end up with an element that is visually identical, but now the entire box is clickable and can be used to toggle the checkbox.

robertknight:
> I'm going to defer to @robertknight on the right way to solve this one, but IMO we definitely
> shouldn't be using CSS classes which are named "form" for things that aren't forms.

That's a fair point, although this is consistent with the structure of the markup for the new signup, login forms. We could perhaps rename this CSS component to `form-container` to make it clear that it doesn't refer to the `<form>` element itself? If this is done, then the templates for the login forms (`{login, signup, forgot_password, reset}.html.jinja2`) will need to be updated.

> You'll end up with an element that is visually identical, but now the entire box is clickable
> and can be used to toggle the checkbox.

👍 Nice trick

nickstenning:
> We could perhaps rename this CSS component to form-container to make it clear that it doesn't refer to the <form> element itself? If this is done, then the templates for the login forms ({login, signup, forgot_password, reset}.html.jinja2) will need to be updated.

Let's do this, but in a followup PR.

nickstenning:
Looking better!

PR #3661, opened 2016-08-08
Show tracebacks when functional tests 500

seanh:
When a functional test fails with a 500 Server Error show the in-app
traceback of the exception that produced the 500, rather than just the
traceback from webtest where it received the 500.

This is useful for debugging functional test errors.


nickstenning:
Alphabetical ordering please! Otherwise :+1: -- already made the same change [over here](https://github.com/hypothesis/h/commit/88c2020acf75e27ca55084f991cea5f7486952b5).

nickstenning:
Fixed the ordering.

PR #3662, opened 2016-08-08
Redirect /reset_password

chdorner:
Redirect based on prefix matches.

This will allow us to crudely do redirects of `/reset_password/{code}`
by matching the `/reset_password` prefix, and then appending the rest at
the end of the new route.


nickstenning:
LGTM.

PR #3663, opened 2016-08-08
Extract memex

nickstenning:
This pull request extracts the `h.api` package into its own package, called `memex`, which can be released independently of the rest of h.

Here's a brief overview of what has happened in this PR:

- The `h.api` package and all references to it have been renamed to `memex`.
- All of the python packaging metadata has been updated to point to `memex` as the installable package, rather than `h`. The `h` application is now not an installable package -- it's just a bunch of Python modules which get built into a Docker image.
- The `memex` package moves into a `src/` directory. The most important reason for this change is that it ensures that our tests run against the _installed_ package, and not the version of the package on disk. For more about the motivation for this layout, see [this article](https://blog.ionelmc.ro/2014/05/25/python-packaging/).
- Updates to the test infrastructure so that `h` tests and `memex` tests run in different tox environments. For now this results in a little duplication between the different `conftest.py` files.
- A new versioning scheme for the `h` application. `h` versions are now of the form (YYYYMMDD+gXXXXXXX[.dirty]) where YYYYMMDD is the date of the most recent commit, and XXXXXXX is the abbreviated commit hash. This relieves us from the need to manually bump the version for the web service.
- `memex` has a manually bumped version, which starts here at `0.39.0+dev`.
- An updated Dockerfile for building the `h` application. This new Dockerfile allows for much faster builds (most of the time).

Due to the changes to imports, some imports will now be in the wrong part of the file. I suggest we leave that as-is in this PR and do a separate pass over the codebase with [isort](https://github.com/timothycrosley/isort) after merge.


setup.cfg
```diff
@@ -1,13 +1,3 @@
-[check-manifest]
-ignore =
-    */test
-    */test/*
-    .*
-    h/browser
-    h/browser/*
-    node_modules
-    tox.ini
-
```
nickstenning: In case it's not clear -- I've removed the `check-manifest` Travis job and config because moving everything into `src/` makes it next to impossible to screw up the `MANIFEST.in` file as it can now just say `graft src` and everything is included.

chdorner:
LGTM. Let's get this on staging and see if it works, looks good locally.

PR #3664, opened 2016-08-09
Remove sourcemap uploader

nickstenning:
This was only ever needed for files built into the Chrome extension, where the source file URLs were not publicly fetchable by the Sentry application.

This does not apply to the site assets and so this isn't needed in this repository.


robertknight:
> This was only ever needed for files built into the Chrome extension,
> where the source file URLs were not publicly fetchable by the Sentry application.

I vaguely recall that Sentry had intermittent problems fetching sourcemaps from external URLs. I'm happy to merge this PR for the moment but if it turns out that Sentry is unreliable at fetching sourcemaps for the site, we might need to revert back to uploading them directly.

PR #3665, opened 2016-08-09
Add migration for Annotation.text_rendered column

chdorner:
This will have to go out to at least staging and then we have to run the migration before we can merge #3666.


chdorner:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;GyZdQ78y&#x2F;414-render-annotation-bodies-as-markdown"> Render annotation bodies as Markdown</a>

PR #3666, opened 2016-08-09
Server-side annotation text markdown rendering

chdorner:
This adds server-side annotation text markdown rendering capabilities.

We use [mistune](https://pypi.python.org/pypi/mistune) for Markdown rendering, and for now make it ignore inline math and math blocks (see `MathMarkdown`, `MathInlineLexer`, `MathBlockLexer`, and `MathRenderer` in `h.api.markdown`).

After converting Markdown to HTML we run the text through [bleach](https://pypi.python.org/pypi/bleach) to sanitize the HTML. We bleach's default list of [allowed tags](https://github.com/mozilla/bleach/blob/v1.4.3/bleach/__init__.py#L23-L36) and [allowed attributes](https://github.com/mozilla/bleach/blob/v1.4.3/bleach/__init__.py#L38-L42), plus add anything that Markdown generates (`h.api.markdown.MARKDOWN_TAGS`, `h.api.markdown.MARKDOWN_ATTRIBUTES`).

For testing purposes, I've created an annotation with the [this text](https://gist.github.com/chdorner/c4ac4d0d8d21cfd00eeb1a5962e51425) which should include all Markdown features plus an oEmbed link (Youtube), and the inline and block math we support.

### DIFF
--- a/src/memex/markdown.py
+++ b/src/memex/markdown.py
@@ -5,6 +5,7 @@ from __future__ import unicode_literals
 import re
 
 import bleach
+from bleach import callbacks as linkify_callbacks
 import mistune
 
 MARKDOWN_TAGS = [
@@ -13,8 +14,18 @@ MARKDOWN_TAGS = [
 ]
 ALLOWED_TAGS = set(bleach.ALLOWED_TAGS + MARKDOWN_TAGS)
 
+
+def _filter_link_attributes(name, value):
+    if name in ['href', 'title']:
+        return True
+
+    if name == 'target' and value == '_blank':
+        return True
+
+    return False
+
 MARKDOWN_ATTRIBUTES = {
-    'a': ['href', 'title'],
+    'a': _filter_link_attributes,
     'img': ['alt', 'src', 'title'],
 }
 ALLOWED_ATTRIBUTES = dict(bleach.ALLOWED_ATTRIBUTES.items() + MARKDOWN_ATTRIBUTES.items())
@@ -70,7 +81,11 @@ def render(text):
 
 
 def sanitize(text):
-    return bleach.clean(text,
+    linkified = bleach.linkify(text, callbacks=[
+        linkify_callbacks.target_blank
+    ])
+
+    return bleach.clean(linkified,
                         tags=ALLOWED_TAGS,
                         attributes=ALLOWED_ATTRIBUTES)
 
--- a/src/memex/markdown.py
+++ b/src/memex/markdown.py
@@ -4,7 +4,20 @@ from __future__ import unicode_literals
 
 import re
 
+import bleach
 import mistune
+
+MARKDOWN_TAGS = [
+    'a', 'blockquote', 'code', 'em', 'hr', 'h1', 'h2', 'h3', 'h4', 'h5', 'h6',
+    'img', 'li', 'ol', 'p', 'pre', 'strong', 'ul',
+]
+ALLOWED_TAGS = set(bleach.ALLOWED_TAGS + MARKDOWN_TAGS)
+
+MARKDOWN_ATTRIBUTES = {
+    'a': ['href', 'title'],
+    'img': ['alt', 'src', 'title'],
+}
+ALLOWED_ATTRIBUTES = dict(bleach.ALLOWED_ATTRIBUTES.items() + MARKDOWN_ATTRIBUTES.items())
 
 # Singleton instance of the Markdown instance
 markdown = None
@@ -52,8 +65,14 @@ class MathRenderer(mistune.Renderer):
 def render(text):
     if text is not None:
         render = _get_markdown()
-        return render(text)
+        return sanitize(render(text))
     return None
+
+
+def sanitize(text):
+    return bleach.clean(text,
+                        tags=ALLOWED_TAGS,
+                        attributes=ALLOWED_ATTRIBUTES)
 
 
 def _get_markdown():
--- a/src/memex/markdown.py
+++ b/src/memex/markdown.py
@@ -8,6 +8,8 @@ import bleach
 from bleach import callbacks as linkify_callbacks
 import mistune
 
+LINK_REL = 'nofollow noopener'
+
 MARKDOWN_TAGS = [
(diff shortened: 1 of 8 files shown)


chdorner:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;GyZdQ78y&#x2F;414-render-annotation-bodies-as-markdown"> Render annotation bodies as Markdown</a>

tests/h/api/markdown_test.py
```diff
@@ -0,0 +1,73 @@
...
+    ])
+    def test_it_allows_markdown_html(self, text):
+        assert markdown.sanitize(text) == text
+
+    @pytest.mark.parametrize("text,expected", [
+        ('<script>evil()</script>', '&lt;script&gt;evil()&lt;/script&gt;'),
+        ('<a href="#" onclick="evil()">foobar</a>', '<a href="#">foobar</a>'),
+        ('<a href="#" onclick=evil()>foobar</a>', '<a href="#">foobar</a>'),
+        ('<a href="javascript:alert(\'evil\')">foobar</a>', '<a>foobar</a>'),
+        ('<img src="/evil.jpg" onclick="evil()">', '<img src="/evil.jpg">'),
```
nickstenning: I've just checked manually that bleach protects against a image src XSS (e.g. ``) but it would probably be worth adding some examples here for that, too.

h/cli/commands/render_annotation_text.py
```diff
@@ -0,0 +1,54 @@
+# -*- coding: utf-8 -*-
```
nickstenning: Perhaps this should be a data migration?

I know why you've added it here (in case we change the rendering algorithm in the future and need to rerun it) but I actually think even then putting it in a data migration has the advantage that it makes it explicit in the repo history that the format has changed and we need to update the database. It's not hard to retrieve a migration from the past and create a new one with the same content. And it means we don't end up with an unbounded number of these one-off commands in the CLI.

robertknight:
I ran into an issue with the migration:

```
(h)  ~/h/r/h > ./bin/hypothesis migrate upgrade head
2016-08-10 15:42:25 2032 alembic.runtime.migration [INFO] Context impl PostgresqlImpl.
2016-08-10 15:42:25 2032 alembic.runtime.migration [INFO] Will assume transactional DDL.
2016-08-10 15:42:25 2032 alembic.runtime.migration [INFO] Running upgrade 6b801ecc60f1 -> 39b1935d9e7b, Add the text_rendered column to the annotation table
2016-08-10 15:42:25 2032 alembic.runtime.migration [INFO] Running upgrade 39b1935d9e7b -> d536d9a342f3, Fill in annotation text_rendered column
...
sqlalchemy.exc.ProgrammingError: (psycopg2.ProgrammingError) column annotation.text_rendered does not exist
LINE 1: ...tion_updated, annotation.text AS annotation_text, annotation...
                                                             ^
 [SQL: 'SELECT annotation.id AS annotation_id, annotation.updated AS annotation_updated, annotation.text AS annotation_text, annotation.text_rendered AS annotation_text_rendered \nFROM annotation \nWHERE annotation.updated BETWEEN %(updated_1)s AND %(updated_2)s ORDER BY annotation.updated ASC'] [parameters: {'updated_1': datetime.datetime(2016, 5, 30, 10, 38, 24, 553896), 'updated_2': datetime.datetime(2016, 8, 10, 14, 31, 25, 955777)}]
```

When I subsequently ran the migrations one by one they worked, but presumably it should be possible for `hypothesis migrate` to run both of these migrations together?

chdorner:
The migration was actually an issue with our alembic configuration, and will be fixed with #3676.

src/memex/markdown.py
```diff
@@ -0,0 +1,84 @@
...
+]
+ALLOWED_TAGS = set(bleach.ALLOWED_TAGS + MARKDOWN_TAGS)
+
+MARKDOWN_ATTRIBUTES = {
+    'a': ['href', 'title'],
+    'img': ['alt', 'src', 'title'],
+}
+ALLOWED_ATTRIBUTES = dict(bleach.ALLOWED_ATTRIBUTES.items() + MARKDOWN_ATTRIBUTES.items())
+
+markdown = None
```
robertknight: Could use a comment to explain what this global var is

tests/memex/markdown_test.py
```diff
@@ -0,0 +1,74 @@
...
+    def test_it_allows_markdown_html(self, text):
+        assert markdown.sanitize(text) == text
+
+    @pytest.mark.parametrize("text,expected", [
+        ('<script>evil()</script>', '&lt;script&gt;evil()&lt;/script&gt;'),
+        ('<a href="#" onclick="evil()">foobar</a>', '<a href="#">foobar</a>'),
+        ('<a href="#" onclick=evil()>foobar</a>', '<a href="#">foobar</a>'),
+        ('<a href="javascript:alert(\'evil\')">foobar</a>', '<a>foobar</a>'),
+        ('<img src="/evil.jpg" onclick="evil()">', '<img src="/evil.jpg">'),
+        ('<img src="javascript:alert(\'evil\')">', '<img>'),
```
robertknight: One additional issue, which I don't think we need to deal with right now but I'll note here, is that we discovered an issue recently with the client where it adds `target="_blank"` to links generated from markdown but not to HTML included directly. This means that if the HTML is served in an iframe, clicking the link ends up navigating just the iframe instead of opening a new tab or navigating the top level window.

src/memex/markdown.py
```diff
@@ -0,0 +1,84 @@
...
+    def parse_block_math(self, m):
+        self.tokens.append({
+            'type': 'block_math',
+            'text': m.group(1)
+        })
+
+
+class MathRenderer(mistune.Renderer):
+    def __init__(self, **kwargs):
+        self.options = kwargs
```
robertknight: Should this call the base class constructor of `mistune.Renderer` here, passing `kwargs`?

robertknight:
A couple of very minor code issues, otherwise LGTM.

nickstenning:
Really nice work on this. LGTM.

PR #3667, opened 2016-08-09
Define a default script bundle in the base layout

robertknight:
Remove the repeated markup for generating `<script>` tags in each page that uses the `site_js` bundle and move it to the base layout instead.

Individual pages or layouts can still override the default bundle and provide their own. Currently only the old homepage does this.

Also add `.jinja2` to `.editorconfig` so that Vim uses correct indentation for new Jinja2 templates.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -7,7 +7,7 @@ indent_style = space
 insert_final_newline = true
 trim_trailing_whitespace = true
 
-[*.{html,js,json,scss,yaml,yml}]
+[*.{html,jinja2,js,json,scss,yaml,yml}]
 indent_size = 2
 
 [*.py]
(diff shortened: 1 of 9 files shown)


nickstenning:
This looks fine. Just checking: you're aware that this will add the `site_js` bundle to a bunch of pages that didn't have it before (the help page, the 404 pages, etc.).

I'll merge it and if you think of any issues that could be caused by including the bundle where it's (currently) not needed, just shout.

nickstenning:
P.S. In future, can I suggest multiple different PRs for multiple independent changes? The `.editorconfig` change is small and unrelated to the main change in this PR, and could easily be merged quickly and without discussion. In addition, reviewing more involved changes is much easier when you're not distracted by other unrelated things.

robertknight:
> Just checking: you're aware that this will add the site_js bundle to a bunch of pages that didn't
> have it before (the help page, the 404 pages, etc.).

Yes, given the way the site bundle works this should be fine. There is no page-specific logic any more, instead the bundle just hunts for elements it can upgrade when it runs. If the page contains nothing to upgrade then the bundle is a no-op.

PR #3668, opened 2016-08-09
Add missing reset_password redirect

chdorner:
#3662 made it possible to redirect based on prefix matches, but was missing the actual redirect for the `/reset_password` path.


nickstenning:
I assumed that was intentional :)

chdorner:
Nope :) I think I wanted to split it up into two commits and then forgot to actually do the second commit.

PR #3669, opened 2016-08-10
Disable pep8 warnings in matchers.py

seanh:


seanh:
Unfortunately with flake8 there's no way to disable a specific warning for the whole file

chdorner:
LGTM.

PR #3670, opened 2016-08-10
Rename TimeframeMatcher for consistency

seanh:
Rename TimeframeMatcher to make it consistent with other matcher classes
in matchers.py.


seanh:
I didn't move this into `matchers.py` because it's specific to the bucketing tests.

PR #3671, opened 2016-08-10
Disable PEP257 missing docstring warnings

seanh:
Although our coding standards claim that we follow PEP8 and PEP257, the
fact is that we don't put docstrings on everything, not even in new code
that we write, and these "missing docstring" warnings are creating a lot
of noise that's drowning out other, more interesting linter warnings.

### DIFF
--- a/.prospector.yaml
+++ b/.prospector.yaml
@@ -15,6 +15,14 @@ pep257:
     - D203 # "1 blank line required before class docstring" conflicts with
            # another pep257 rule D211 "No blank lines allowed before class
            # docstring"!
+
+    # "Missing docstring" warnings:
+    - D100
+    - D101
+    - D102
+    - D103
+    - D104
+    - D105
 pyflakes:
   run: false
 


nickstenning:
I'm not sure why we'd want to disable these? Why not write docstrings for public methods instead?

seanh:
We don't seem to be writing docstrings for public functions, even in new code.

I'm happy to start writing them if that's what you want to do.

I don't think pep257 supports any way to disable a particular warning for a particular line or file. We could disable all pep257's missing docstring warnings globally and then re-enable pylint's equivalent missing docstring warnings, though. Because pylint lets you turn off the warning on a per-line or per-file basis, for cases where we really don't want to write a docstring.

We have 880 linter warnings in the whole Python codebase (flake8 and prospector combined) of which 527 (60%) are missing docstring warnings.

nickstenning:
We should have docstrings for _public_ functions. We don't always remember to write them, but I don't think turning off the linter warnings is a good idea.

If I want to see the pydocstyle warnings that aren't about missing warnings, then I can just run:

```
pydocstyle --ignore=D100,D101,D102,D103,D104,D105 h
```

Gonna close this.

seanh:
What I personally want from the linters is this:

1. Have your text editor configured to automatically run the linters and show the warnings in the text editor (I use Syntastic)
2. Always either do what the linter says or configure it (either in its config file or with inline comments) so that there should be 0 warnings on master
3. The result if you do both 1 and 2 is sublime, because it has the following characteristics:
   1. The linters are always run automatically for you by your text editor, you never have to remember or think about them, run them yourself, or wait for them while they run
   2. You almost never hear from the linters as there aren't any violations, 99% of the time you forget they even exist
   3. On the rare occasions when you do hear from a linter (completely automatically a warning pops up next to a line of code in your editor) it's almost always an actual problem that you actually want to know about or fix now. Very occasionally it'll be a warning that you want to reconfigure or disable either globally or one-off.

In other words completely automated and all signal / no noise.

It's really nice to have this automatic "co-pilot" as you're coding that you almost never know is there, but that occasionally points out something useful to you.

I don't want to use command line arguments because I don't run the linters from the command line. I might be able to configure or hack Syntastic to run them with args but then those args would apply to anything I edit in vim not just h. So what I've ended up doing is adding `.landscape.yml` and `.flake8` files (in the `h` dir) containing my personal linter configs for h.  As a start I have:

`.landscape.yml`:

```yaml
inherits:
  - .prospector.yaml
pep257:
  disable:
    - D100  # Missing docstring in public module.
    - D101  # Missing docstring in public class.
    - D102  # Missing docstring in public method.
    - D103  # Missing docstring in public function.
    - D104  # Missing docstring in public package.
    - D105  # Missing docstring in magic method.
pyroma:
  run: false
```

`.flake8`:

```ini
[flake8]
ignore =
    # Function name should be lowercase.
    N802,
    # Argument name should be lowercase.
    N803,
```

I then ignore both files in `.git/info/exclude` so they don't clutter up my `git status` or accidentally get committed:

`.git/info/exclude`:

```.gitignore
.landscape.yml
.flake8
```

Both `.landscape.yml` and `.flake8` seem to inherit from / be combined with the existing project config files.

So now I can configure them to my heart's content.

One thing I haven't solved yet is that if you destroy and recreate your dev env (or run `pip sync` or something) then flake8 and prospector won't be installed in it, and syntastic will just silently not run them, and since I'm used to having them be completely automated and never thinking about them I won't notice. Might figure something out for this one day.

nickstenning:
> Might figure something out for this one day.

Just add them to `requirements-dev.in`?

seanh:
Well that would show up in `git status`, or you mean I should commit that `requirements-dev.in` change and send a pr?

seanh:
I see prospector is already in requirements-dev, no reason not to add flake8 then

PR #3672, opened 2016-08-10
Docker build tweaks

nickstenning:
This PR improves our Docker build by ensuring we don't install system deps which aren't needed at runtime in the image.

Until now, every commit invalidated the Docker image layer which contained the application dependencies. This meant that installing the system-level build deps and leaving them installed was preferable, despite the additional space they occupied, as a result of the decreased build time.

Now, we no longer need to invalidate the layer on every commit, so it makes more sense to recover the disk space by only installing the system build deps for as long as they are needed.

This reduces the size of the image by ~100MB.

As a bonus, I've upgraded us to Alpine 3.4.


chdorner:
LGTM.

PR #3673, opened 2016-08-10
Don't run `npm` when parsing the Makefile

nickstenning:
This results in warnings when the Makefile is run in environments without npm installed.


chdorner:
LGTM.

PR #3674, opened 2016-08-10
Tag Docker image with h version info

nickstenning:
This commit updates the Jenkinsfile so that the Docker image we build is tagged with the specific version we're building, e.g. version 20160810+g213e0a8 would be tagged as "20160810-g213e0a8".

This will make it easier to track what version of the software is running in what environment, by avoiding our reliance on "latest" and "auto" tags.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,21 +1,34 @@
 #!groovy
 
 node {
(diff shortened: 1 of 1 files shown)


chdorner:
LGTM.

PR #3675, opened 2016-08-10
Add a helper for generating SVG icons for use on the website

robertknight:
The helper generates icons using inline SVG, following an approach
described at https://github.com/blog/2112-delivering-octicons-with-svg

This gives us a nice way to do vector icons which can be customized
using CSS.

In order for icons to be colorable using CSS, they must set the 'fill'
or 'stroke' properties of paths to 'currentColor'

### DIFF
--- a/h/app.py
+++ b/h/app.py
@@ -65,6 +65,7 @@ def includeme(config):
 
     config.include('pyramid_jinja2')
     config.add_jinja2_extension('h.jinja_extensions.Filters')
+    config.add_jinja2_extension('h.jinja_extensions.SvgIcon')
     # Register a deferred action to setup the assets environment
     # when the configuration is committed.
     config.action(None, configure_jinja2_assets, args=(config,))
--- a/h/jinja_extensions.py
+++ b/h/jinja_extensions.py
@@ -1,10 +1,19 @@
 # -*- coding: utf-8 -*-
 
 import datetime
+from functools import partial
 import json
+import re
+
+try:
+    from xml.etree import cElementTree as ElementTree
+except ImportError:
+    from xml.etree import ElementTree
 
 from jinja2 import Markup
 from jinja2.ext import Extension
+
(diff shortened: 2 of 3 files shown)


robertknight:
The Jenkins build failed because libxml2-dev is not installed. We could find alternative ways to manipulate the `<svg>` output if the `lxml` dependency is a significant issue. Thoughts?

nickstenning:
It's not a significant issue -- we can certainly install lxml if necessary -- but I wonder if you even need it. I'd guess that this would work just fine with no additional dependencies:

```python
try:
    from xml.etree import cElementTree as ElementTree
except ImportError:
    from xml.etree import ElementTree

...
    tree = ElementTree.parse(loader(name))
...
```

robertknight:
> It's not a significant issue -- we can certainly install lxml if necessary -- but I wonder if you even
> need it. I'd guess that this would work just fine with no additional dependencies:

The issue with `ElementTree` is this:

```
>>> from xml.etree import ElementTree as ET
>>> ET.tostring(ET.fromstring('<svg xmlns="http://www.w3.org/2000/svg"></svg>'))
'<ns0:svg xmlns:ns0="http://www.w3.org/2000/svg" />'
```

I can get around it using the global `ET.register_namespace` helper, which is fine as long as nothing else in the codebase tries to register a different namespace with the same URI:

```py
def svg_icon(loader, name, css_class=''):
    ElementTree.register_namespace('', 'http://www.w3.org/2000/svg')
    tree = ElementTree.parse(loader(name))
    ...
```

nickstenning:
Registering the namespace seems fine. You can do that at module level.

robertknight:
> Registering the namespace seems fine. You can do that at module level.

I had a quick look at the implementation, its very cheap and replaces any existing registration so I'm doing it in the `svg_icon` function directly to keep everything close together.

nickstenning:
Honestly, it would be better to have it at top-scope so that it's _clear_ that it's a stateful action. But ¯\_(ツ)_/¯.

PR #3676, opened 2016-08-10
Run each alembic migration in its own transaction

chdorner:
This is a config setting that we should have enabled a long time ago
(and the alembic default should be turned on).
It ensures that each migration is wrapped in its own transaction. This
is especially important when we run two migrations at the same time, the
first one adding a column, and the second one using that column.

We also don't need to wrap the `context.run_migrations()` call in a
transaction anymore since alembic takes care of that now.

### DIFF
--- a/h/migrations/env.py
+++ b/h/migrations/env.py
@@ -60,10 +60,10 @@ def run_migrations_offline():
     script output.
 
     """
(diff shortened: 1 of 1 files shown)


nickstenning:
LGTM.

PR #3677, opened 2016-08-11
Enable submitting create group form via xhr

seanh:


tests/h/accounts/views_test.py
```diff
@@ -10,6 +10,8 @@
 from h import accounts
 from h.accounts import views
 
+from tests.h import conftest
```
nickstenning: You should never need to import the conftest file, and doing so is a surefire recipe for circular imports. Just expose the things you want to use in tests as fixtures.

h/form.py
```diff
@@ -89,6 +95,78 @@ def configure_environment(config):  # pragma: no cover
...
+
+    :param form: the form that was submitted
+    :type form: deform.form.Form
+
+    """
+    if not request.is_xhr:
+        return non_xhr_result
+
+    request.override_renderer = 'string'
+    return form.render()
```
robertknight: Since no argument is being passed to `render()`, this will return the form without any of the fields filled in. If we naively apply the result on the client (eg. with `formElement.innerHTML = <updated form>`) this will result in all the input fields being empty, instead of showing their current values.

In the legacy forms on the site, the current form values are not included when the form is rendered, but in new designs (eg. As seen in the mocks at https://trello.com/c/KIU3NXPe/411-inline-editing-enhancement-for-forms) the forms show the current field values when they are initially displayed.

On this basis I would suggest that render should be passed the appstruct that was saved.

tests/h/accounts/views_test.py
seanh: Changed these to fixtures

h/form.py
seanh: Are you sure?

seanh: If I change `test_submit_create_group_form_with_xhr_returns_partial_html_snippet()` to submit `group_form['name'] = 'abc'` (this is an invalid group name - too short) and then inspect `res.body`, I see `<input type="text" name="name" value="abc" ... `, the error message "Shorter than minimum length 4" is also there.

nickstenning:
This is pretty hefty PR. Can I suggest splitting this into at least two different changesets.

1. Move the `to_xhr_response` and `handle_form_submission` helpers to their new homes.

2. Make the necessary changes to the groups views.

nickstenning:
Also, please do add a description to your PRs, however cursory. 

Not adding a description pretty much guarantees that the only people who will review it are the people who already know why you're doing it. By adding a well-written description that explains the motivation and approach for the PR, you're maximising the chances that somebody can pick up and review the PR without prior knowledge.

For bonus points, include the descriptions in the commit messages and then summarise them in the PR description.

seanh:
Split into <https://github.com/hypothesis/h/pull/3700> and <https://github.com/hypothesis/h/pull/3701>

PR #3678, opened 2016-08-11
New navigation bar

robertknight:
_This depends on #3675_

This implements the new navigation bar on Activity Pages and Accounts pages for mobile and desktop. The navigation bar is implemented as a [panel](http://pyramid-layout.readthedocs.io/en/latest/layouts.html#using-panels) using `pyramid_layout` which provides a way to define components consisting of views + templates across multiple pages, without needing to add all the context data to renderer globals or clutter the templates with complex expressions/logic. In future this also provides a nice path towards having chunks of a page which can be rendered as part of a complete page or as just a chunk which is applied to an existing page by the client.

----
- [x] Tests for dropdown menu controller
- [x] Check whether `request.authenticated_user.groups` is guaranteed to be sorted in alphabetical order or whether we need to sort these ourselves (*Yes it is - see the relationship definition in `h.groups.models`)


h/static/scripts/dropdown-menu-controller.js
```diff
@@ -0,0 +1,35 @@
+'use strict';
```
nickstenning: Could we possibly start putting non-entrypoint JavaScript files in a directory of their own? Having one big flat folder full of JavaScript makes it very difficult to understand how the code is structured.

nickstenning:
The whole of `panels.py` appears to be untested?

nickstenning:
This looks good -- I think there are a number of visual issues that we'll need to work through but there's no reason to block merging this on those. Just add some tests for the panels file and we'll get this merged.

h/static/scripts/dropdown-menu-controller.js
robertknight: Agreed. I'll do this as a follow-up for all the controllers.

nickstenning:
LGTM.

PR #3680, opened 2016-08-12
Improve CSS asset watcher

robertknight:
Improve the system watch auto-rebuilds CSS when input SASS files change:

 - Fix SASS build errors causing the Gulp asset build process to exit,
   requiring a restart of `make dev`. This only affects the 'watch-css'
   task. build-css will still exit with a non-zero status if it fails.

 - Fix CSS build not triggering when newly added dependencies change.

The approach here is to make the CSS build work much more like the
Browserify JS build:

 - Add a `create-style-bundle` module which provides a function that
   takes care of building style bundles and applying appropriate
   transformations.

 - Instead of watching a fixed list of files via a glob when the SASS
   build starts, use the dependency information from the SASS build to
   dynamically watch the files transitively imported by the entry file.

   This is similar to how `watchify` automatically picks up new
   dependencies in JS bundles.

   The result is that the build will now start watching for changes to
   CSS files when they are referenced by the entry file or one of its
   existing dependencies.

 - Simplify the vendor CSS build and remove the sourcemap and
   autoprefixer steps that are not required


nickstenning:
Am I missing something really obvious here...

```
./node_modules/.bin/node-sass \
  --watch \
  --recursive \
  --source-map true \
  --source-map-contents \
  --output build/styles/ \
  --include-path h/static/styles/ \
  h/static/styles
```

Doesn't this achieve basically the same thing, only without the several hundred lines of code?

nickstenning:
So, I've had a closer look and I see that of course this doesn't do autoprefixer, and it doesn't fix up the font URLs (why do we even need to do that, still?) but I really think the front-end build is getting a bit out of hand. Do we really need hundreds of lines of code just to copy some CSS from one place to another?

robertknight:
> Do we really need hundreds of lines of code just to copy some CSS from one place to another?

I understand the concern. On the other hand, this fits in better with the rest of our asset build (being usable from Gulp) and as you noted, makes it easy to do things like autoprefixer and URL rewriting.

The `create-style-bundle` module can also solve the same annoyances with the client repo, so what I would propose is moving it to a package in the frontend-toolkit repo. IIRC we talked about doing something similar a while back for the code which builds the Browserify bundle.

> So, I've had a closer look and I see that of course this doesn't do autoprefixer, and it doesn't
> fix up the font URLs (why do we even need to do that, still?) 

Just had a look at this. It looks like it is not something that we currently rely on. I did find at least one broken image URL reference along the way though, and it would be nice if URLs were checked at build time, something which could be quite easily added into this.

nickstenning:
> On the other hand, this fits in better with the rest of our asset build (being usable from Gulp)

Ok, but let's not use complexity to justify complexity. I could make a similar argument that a single `browserify` CLI invocation would replace 98% of what the Gulpfile does for our JS build.

I'm happy to merge this so we can get on to more important things, but I'd like to see the complexity of the build tooling trending *down* now that we aren't building three projects in one repository, not up.

robertknight:
> I'm happy to merge this so we can get on to more important things, but I'd like to see the
> complexity of the build tooling trending down now that we aren't building three projects in one repository, not up.

Noted. Merging this will make my life easier in the meantime :)

PR #3681, opened 2016-08-12
Remove unused CSS files

robertknight:
- Remove a 'robots.txt' file that was used when static assets were
   served from the 'h/static' folder

 - Remove KaTeX CSS files and fonts that are now part of the client
   assets

 - Remove Angular CSP support CSS file that is now part of the client
   assets


nickstenning:
> Remove a 'robots.txt' file that was used when static assets were served from the 'h/static' folder

It's always good to have a quick grep to verify the thing you're removing isn't used! This is [still in use](https://github.com/hypothesis/h/blob/79821aec9801acb59fff7590dc92a060b5c09380/h/views/main.py#L46-L50).

robertknight:
> It's always good to have a quick grep to verify the thing you're removing isn't used! This is still in use.

Oh good grief, PEBKAC. I've reverted that from this PR.

nickstenning:
LGTM.

PR #3682, opened 2016-08-12
Restructure JS files

robertknight:
This restructures the `h/static/scripts` dir to move everything that is not a bundle entry point into its own directory. Files are currently grouped into `controllers` (classes that upgrade elements on the page) and `base` (shared code for reading settings, setting up Raven etc.) that is used by multiple sections of the site.

I haven't re-arranged the 'test/' dir accordingly because we should agree on how we want to structure that first. I think we've got three options:

1. Mirror the way our Python tests are structured (eg. `tests/controllers/shiny-button-controller-test.js`)
2. Put the JS unit tests alongside their modules with a `.test.js` suffix, which is probably the most common JS convention (`controllers/shiny-button-controller.test.js`)
3. Follow the convention in the client/browser-extension repos of having a 'test' folder inside each dir (`controllers/test/shiny-button-controller-test.js`)

(2) or (3) would be my personal preference in terms of developer convenience and making it obvious which modules have tests but I think @nickstenning you expressed a preference for having a parallel dir structure in a PR comment recently.

----

- [x] Bikeshed layout of unit tests and move these files accordingly


nickstenning:
@robertknight Where you put the tests is entirely up to you.

robertknight:
I have updated the test structure, using the same layout as the `client` and `browser-extension` repos, since the layout in those repos proved convenient to work with.

robertknight:
Following discussion on Slack (see https://botbot.me/freenode/hypothes.is/msg/71389150/ ) I've re-arranged the tests again to approach (1) and @nickstenning reminded me that the browser-extension repo now has a top-level tests dir.

sheetaluk:
LGTM.

PR #3684, opened 2016-08-12
Remove Raven addons that are not required by the site

robertknight:
- Remove the Angular integration as the new site does not use Angular

 - Remove the source filename translation which was needed for the
   browser extension but is not required by the site which is always
   served from HTTP(S) URLs


nickstenning:
LGTM.

PR #3687, opened 2016-08-12
Close db connections opened by exception views

### LINKED ISSUE #3683: Submitting accounts forms with a bad CSRF token leaves database transactions open
[h.accounts.views.bad_csrf_token()](https://github.com/hypothesis/h/blob/689535f0e448449fa042fd5e78efa1fdee8c2a55/h/accounts/views.py#L105) (which is a `BadCSRFToken` exception view) opens database transactions that do not get closed.

For example add this test:

```diff
diff --git a/tests/functional/test_accounts.py b/tests/functional/test_accounts.py
index a84f368..8e7761d 100644
--- a/tests/functional/test_accounts.py
+++ b/tests/functional/test_accounts.py
@@ -46,6 +46,16 @@ class TestAccountSettings(object):
 
         assert res.content_type == 'text/plain'
 
+    def test_submit_email_form_without_csrf(self, app):
+        res = app.get('/account/settings')
+        email_form = res.forms['email']
+        email_form['email'] = 'new_email@example.com'
+        email_form['email_confirm'] = 'new_email@example.com'
+        email_form['password'] = 'pass'
+        res.forms['email']['csrf_token'] = None
+        for _ in range(50):
+            email_form.submit(xhr=True, status=403)
+
     def test_submit_password_form_without_xhr_returns_full_html_page(self,
                                                                      app):
         res = app.get('/account/settings')
```

and then `py.test tests/functional/test_accounts.py -svx` will freeze waiting to open a database connection.

The problem is that `bad_csrf_token()` calls `h.session.model()` which access the db. If you make this change to `model()` the freeze will go away:

```diff
diff --git a/h/session.py b/h/session.py
index 4ee8519..8076204 100644
--- a/h/session.py
+++ b/h/session.py
@@ -4,13 +4,10 @@
 def model(request):
     session = {}
     session['csrf'] = request.session.get_csrf_token()
-    session['userid'] = request.authenticated_userid
-    session['groups'] = _current_groups(request)
-    session['features'] = request.feature.all()
+    session['userid'] = None
+    session['groups'] = []
+    session['features'] = []
     session['preferences'] = {}
-    user = request.authenticated_user
-    if user and not user.sidebar_tutorial_dismissed:
-        session['preferences']['show_sidebar_tutorial'] = True
     return session
```

This isn't a solution of course but it demonstrates that accessing the db from an exception view is what causes the unclosed transactions.

The problem is that with pyramid_tm **you cannot access the database from an exception view, a NewResponse callback, or a finished callback** (`request.add_finished_callback()`). Because of the order in which `pyramid_tm` and exception views execute, `pyramid_tm`'s transaction has already been committed or aborted by the time the exception view executes. So any code that ia called during an exception view and directly or indirectly uses the db will get multiple problems, including:

* `DetachedInstanceError`s if the exception view tries to access certain properties of certain ORM objects. Because the session has already been committed, so ORM objects such as `request.authenticated_user` are now "detached" objects. If for example you access a lazy-loaded attribute of such a detached ORM object you'll get a `DetachedInstanceError`. This currently happens if any code (including templates) called during an exception view tries to read a feature flag, because reading feature flags accesses lazy loaded properties of `request.authenticated_user` (e.g. `.cohorts`). We don't currently have any instances of this problem that I know of in our code, but it's waiting to happen.

* Unclosed database transactions. Try to read something from the db, sqlalchemy implicitly creates a new transaction, `pyramid_tm` is already finished and won't close this transaction for you.

For more info about this problem in pyramid and pyramid_tm:

* https://github.com/Pylons/pyramid/issues/2622
* https://github.com/Pylons/pyramid/issues/2606
* https://github.com/Pylons/pyramid_tm/issues/40
* https://github.com/Pylons/pyramid_tm/issues/41
* https://github.com/Pylons/pyramid/issues/1336

seanh:
Make sure that db connections opened by exception views or by NewRequest
callbacks get closed.

Fixes #3683.

### DIFF
--- a/h/db/__init__.py
+++ b/h/db/__init__.py
@@ -78,6 +78,22 @@ def _session(request):
     else:
         zope.sqlalchemy.register(session, transaction_manager=tm)
 
+    # pyramid_tm doesn't always close the database session for us.
+    #
+    # For example if an exception view accesses the session and causes a new
+    # transaction to be opened, pyramid_tm won't close this connection because
+    # pyramid_tm's transaction has already ended before exception views are
+    # executed.
+    # Connections opened by NewResponse and finished callbacks aren't closed by
+    # pyramid_tm either.
+    #
+    # So add our own callback here to make sure db sessions are always closed.
+    #
+    # See: https://github.com/Pylons/pyramid_tm/issues/40
+    @request.add_finished_callback
+    def close_the_sqlalchemy_session(request):
+        session.close()
+
     return session
 
 


seanh:
This of course won't close db connections opened by finished callbacks that were registered after this finished callback! But that seems unlikely.

This does not fix the `DetachedInstanceError` problem with the feature flags client, only fixes the unclosed transactions problem.

seanh:
@nickstenning You mentioned a couple of times that there was an obvious bug with this line <https://github.com/hypothesis/h/blob/51c07a6920c09b5c765274cc96795fd04d4686f8/h/db/__init__.py#L77> but I'm afraid I don't know what that bug is?

It looks like if somehow when `request.db` is accessed for the first time in a request there is no `request.tm` then the session won't be registered with a transaction manager and therefore there'll be no transaction manager to close the session. But I don't know when it would happen that there'd be a request with no tm (this doesn't happen in the cases of `NewRequest` subscribers or exception views for example).

robertknight:
> You mentioned a couple of times that there was an obvious bug with this
> line https://github.com/hypothesis/h/blob/51c07a6920c09b5c765274cc96795fd04d4686f8/h/db/__init__.py#L77 but I'm afraid I don't know what that bug is?

Yes you do ;) - As your next sentence explains, if the request has no transaction manager _at the point when `_session` is invoked_ then the DB connection will silently get leaked. I don't know the answer about when/whether `_session` is ever being invoked when there is no transaction manager active though.

nickstenning:
@robertknight is spot on. You do know what the bug is, and you just fixed it.

nickstenning:
FWIW, I really dislike function names like `close_the_sqlalchemy_session`. There's no need to be so verbose when `cleanup` would do, especially given the excellent, detailed comment directly above.

Not going to stop me merging it, though.

seanh:
My thinking was that there's no need to be terse when we're never going to have to write a line of code that calls this function. Same reason I think terse function names aren't justified for test functions. But for normal functions that we're actually going to call, yes.

seanh:
I was wondering if there is ever a case when `request.tm` doesn't exist when `_session()` is called, though? I don't think that was what was happening in the case of exception views not closing sessions for example, that was a different problem. Is the try / except there even needed? It's not important, just something I was wondering about.

nickstenning:
Verbose != readable. The longer name is more distracting than it is helpful, regardless of whether it's ever called outside of this function.

nickstenning:
In the context of our web application you are probably right that `request.tm` will always be present, and the real bug was assuming that the presence of `request.tm` implied that somebody else would clean up the session.

I can check later how the websocket server works, but it probably doesn't need to use `pyramid_tm` even if it currently does so. The `try: ... except: ...` might not strictly be needed currently, but there's also no need to tightly couple `h.db` to `pyramid_tm`, which is why it is there.

PR #3689, opened 2016-08-14
Add annotation quote and text to presenter.

sheetaluk:
The annotation description property contains the quote
wrapped in `<blockquote>` as well as the text of the entered
annotation. Add separate properties for the quote and the
annotation text.

https://trello.com/c/TDZE8zFu/416-document-component

### DIFF
--- a/h/presenters.py
+++ b/h/presenters.py
@@ -20,16 +20,16 @@ def _format_document_link(href, title, link_text, host_or_filename):
 
     """
     if href and host_or_filename and host_or_filename in link_text:
-        host_or_filename = ""
+        host_or_filename = ''
     elif not href and title == host_or_filename:
-        title = ""
+        title = ''
 
     def truncate(content, length=55):
         """Truncate the given string to at most length chars."""
         if len(content) <= length:
             return content
         else:
-            return content[:length] + jinja2.Markup("&hellip;")
+            return content[:length] + jinja2.Markup('&hellip;')
 
     host_or_filename = truncate(host_or_filename)
     link_text = truncate(link_text)
@@ -63,9 +63,31 @@ class AnnotationHTMLPresenter(object):
         else:
             self.document = None
 
+    def _get_selection(self):
+        selectors = self.annotation.target_selectors
+        for selector in selectors:
+            if 'exact' in selector:
+                return selector['exact']
+
     @property
     def uri(self):
         return jinja2.escape(self.annotation.target_uri)
+
+    @property
+    def text_rendered(self):
+        """The body text of this annotation."""
+
+        return self.annotation.text_rendered
+
+    @property
+    def quote(self):
+        """The text in the document which this annotation refers to"""
+
+        selection = self._get_selection()
+        if selection:
+            return jinja2.escape(selection)
+
+        return ''
 
     @property
     def description(self):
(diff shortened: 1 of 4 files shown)


h/presenters.py
```diff
@@ -68,6 +68,29 @@ def uri(self):
         return jinja2.escape(self.annotation.target_uri)
 
     @property
+    def text(self):
```
robertknight: If you are going to use the `text_rendered` property then this isn't needed.

h/presenters.py
```diff
@@ -68,6 +68,29 @@ def uri(self):
...
+        """The body text of this annotation.
+
+        """
+
+        return  jinja2.escape(self.annotation.text)
+
+    @property
+    def quote(self):
+        """The target text of this annotation.
```
robertknight: See how other single-line API doc comments in the file are formatted, just `"""One line description."""` with no extra newlines

h/presenters.py
```diff
@@ -68,6 +68,29 @@ def uri(self):
...
+
+        """
+        def get_selection():
```
robertknight: This could be extracted into a `_quote()` helper method on the class and the `description()` method below refactored accordingly.

tests/h/presenters_test.py
```diff
@@ -32,6 +32,28 @@ def test_uri_returns_Markup(self):
...
 
+    def test_quote(self):
+        annotation = self._annotation(
+            annotation=mock.Mock(
+                target_selectors=[{'exact': 'selected text'}],
+                text="entered text"
+            )
+        )
+
+        assert annotation.quote == (
```
robertknight: You don't need to split an assert over multiple lines if it is <= 79 chars total

h/presenters.py
```diff
@@ -63,11 +63,34 @@ def __init__(self, annotation):
...
+                return selector["exact"]
+
     @property
     def uri(self):
         return jinja2.escape(self.annotation.target_uri)
 
     @property
+    def text_rendered(self):
+        """The body text of this annotation.
+
```
robertknight: The empty new line in the comment is not consistent with other single-line API comments in the file. Should just be `"""One line description."""`

h/presenters.py
```diff
@@ -63,11 +63,34 @@ def __init__(self, annotation):
...
+        """
+
+        return self.annotation.text_rendered
+
+    @property
+    def quote(self):
```
robertknight: In the front-end code at least, we require functions to be consistent about returning a value or not returning a value. Other functions in this file return an empty string if the value is not present, I would suggest doing the same here.

tests/h/presenters_test.py
```diff
@@ -32,6 +32,24 @@ def test_uri_returns_Markup(self):
...
+            "selected text")
+
+    def test_text_rendered(self):
+        annotation = self._annotation(
+            annotation=mock.Mock()
+        )
+
+        assert annotation.text_rendered == (annotation.annotation.text_rendered)
```
robertknight: The parens around the expression are not needed here.

h/presenters.py
```diff
@@ -63,11 +63,34 @@ def __init__(self, annotation):
...
+        """The target text of this annotation.
```
robertknight: I think this description is a little ambiguous. Perhaps "The text in the document which this annotation refers to"?

h/presenters.py
```diff
@@ -63,11 +63,34 @@ def __init__(self, annotation):
         else:
             self.document = None
 
+    def _get_selection(self):
+        selectors = self.annotation.target_selectors
+        for selector in selectors:
+            if "exact" in selector:
```
robertknight: Strings should use single quotes

robertknight:
LGTM

PR #3690, opened 2016-08-15
Add infrastructure for adjusting CSS styling depending on user agent capabilities

robertknight:
_This depends on https://github.com/hypothesis/h/pull/3682_

This PR implements the basic infrastructure for changing element styling depending on user agent capabilities (scripts supported, touch input available etc.). This consists of:

- A small script which is added inline to the `<head>` of each page that detects the user agent capabilities and adds `env-$flag` classes to the `<html>` element.
- `env-js-capable` and `env-touch` flags to indicate support for scripting and touch input respectively
- SASS mixins to enable selectors if these flags are set
- Changing the styling of form buttons to use the `env-touch` flag as a fallback method for enabling touch-friendly buttons in browsers that do not support CSS interaction media queries (Safari < 9.x, Firefox)

Even though the header script is pretty small, it is still built as a Browserify bundle because the overhead of that is minimal and it is convenient for testing and supporting dev/production builds.

### DIFF
--- a/h/templates/layouts/base.html.jinja2
+++ b/h/templates/layouts/base.html.jinja2
@@ -81,6 +81,10 @@
       </script>
     {% endif %}
 
+    <script>
+    {% include "h:../build/scripts/header.bundle.js" %}
+    </script>
+
     {% block base_tag %}{% endblock %}
   </head>
   {% block body_tag %}<body class="body">{% endblock %}
--- a/h/templates/layouts/base.html.jinja2
+++ b/h/templates/layouts/base.html.jinja2
@@ -81,9 +81,9 @@
       </script>
     {% endif %}
 
-    <script>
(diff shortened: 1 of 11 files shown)


h/templates/layouts/base.html.jinja2
```diff
@@ -81,6 +81,10 @@
       </script>
     {% endif %}
 
+    <script>
+    {% include "h:../build/scripts/header.bundle.js" %}
```
nickstenning: This isn't great if we're trying to make the site CSP-safe. I get why you've inlined this but I'd suggest that for now we load it with a `<script src>`.

nickstenning:
Other than my comment about inlining the header bundle, this looks nice!

nickstenning:
LGTM.

PR #3691, opened 2016-08-15
Style activity page search results.

sheetaluk:
Implement basic styling for the activity page search result list from the specs at:
https://trello.com/c/TDZE8zFu/416-document-component

### DIFF
--- a/h/static/styles/partials-v2/_annotation-card.scss
+++ b/h/static/styles/partials-v2/_annotation-card.scss
@@ -1,0 +1,56 @@
+// Annotation card
+// ---------------
+// Styles for the annotation card which is displayed in the
+// results of activity queries.
+
+.annotation-card {
+  @include font-normal;
+  border: 1px solid $grey-2;
+  list-style: none;
+  padding: 15px;
+  margin-bottom: 10px;
+  box-shadow: 0px 1px 1px 0px rgba(0, 0, 0, 0.1);
+  background: white;
+}
+
+.annotation-card:hover {
+  box-shadow: 0px 2px 3px 0px rgba(0, 0, 0, 0.15);
+
+  .annotation-card__quote {
+    border-left: $highlight 3px solid;
+    color: $grey-5;
+  }
+
+  .annotation-card__timestamp {
+    color: $grey-5;
+  }
+}
+
+.annotation-card__header {
+  margin-bottom: 15px;
+  display: flex;
+  flex-direction: row;
+  align-items: baseline;
+  justify-content: space-between;
+}
+
+.annotation-card__username {
+  color: #202020;
+  font-weight: bold;
+}
+
+.annotation-card__timestamp {
+  @include font-small;
+  color: $grey-4;
+}
+
+.annotation-card__quote {
+  padding: 0 10px;
+  overflow-wrap: break-word;
+  color: $grey-4;
+  font-family: sans-serif;
+  font-size: 12px;
+  font-style: italic;
+  letter-spacing: 0.1px;
+  border-left: 3px solid #d3d3d3;
+}
--- a/h/static/styles/partials-v2/_search.scss
+++ b/h/static/styles/partials-v2/_search.scss
@@ -1,0 +1,149 @@
+// Search page
+// -----------
+// Styles for the search page which displays the results
+// of activity queries.
+
+.search-result-container {
+  font-size: $normal-font-size;
+  margin-top: 20px;
+  color: $grey-6;
+}
+
+.search-result-list {
+  max-width: 950px;
+
+  // Remove default padding from <ol>
+  list-style: none;
+  padding-left: 0;
+  padding-right: 0;
+
+  // These margins are set so that the left edge of the search result list
+  // aligns with the left edge of the navbar content
+  margin-left: 30px;
+  margin-right: 30px;
+}
+
+.search-result__timeframe {
+  font-weight: bold;
+  padding-bottom: 10px;
+  border-bottom: 1px solid $grey-2;
+  margin-top: 30px;
+}
+
+// A group of search results
+.search-result-bucket {
+  display: flex;
+  flex-direction: row;
+  padding: 10px 10px;
+  background-color: $grey-2;
+  border-bottom: 1px solid $grey-3;
+
+  &:hover {
+    background-color: $grey-2;
+  }
+}
+
+.search-result-bucket__domain {
+  width: 120px;
+  color: $grey-4;
+}
+
+.search-result-bucket__content {
+  display: flex;
+  flex-direction: column;
+  flex-grow: 1;
+}
+
+.search-result-bucket__header {
+  display: flex;
+  flex-direction: row;
+}
+
+.search-result-bucket__title {
+  font-weight: bold;
+  flex-grow: 1;
+}
+
+.search-result-bucket__annotation-cards-container {
+  display: flex;
+}
+
+.search-result-bucket__annotation-cards {
+  flex-grow: 1;
+}
+
+.search-result-bucket__annotations-count-container {
+  width: 30px;
+  padding: 5px 0px;
+  background-color: $grey-2;
+  text-align: center;
+  border-radius: 2px;
+  float: right;
+}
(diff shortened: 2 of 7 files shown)


sheetaluk:
1. I have asked @conordelahunty to take a look at it.
2. I don't know if I have made the annotation_card macro correctly.
3. I'm yet to add the username to the annotation-card and also the annotation-card and username clickable.

h/static/styles/partials-v2/_annotation-card.scss
```diff
@@ -0,0 +1,60 @@
...
+  overflow-wrap: break-word;
+  color: $grey-4;
+  font-family: sans-serif;
+  font-size: 12px;
+  font-style: italic;
+  letter-spacing: 0.1px;
+  border-left: 3px solid #d3d3d3;
+}
+
+.annotation-card__text {
```
robertknight: This CSS selector is empty so it should just be removed

h/static/styles/partials-v2/_annotation-card.scss
```diff
@@ -0,0 +1,60 @@
...
+.annotation-card:hover {
+  box-shadow: 0px 2px 3px 0px rgba(0, 0, 0, 0.15);
+
+  .annotation-card__quote {
+    border-left: $highlight 3px solid;
+    color: $grey-5;
+  }
+
+  .annotation-card__timestamp {
+    color: $grey-5;;
```
robertknight: Double semi-colon

h/static/styles/partials-v2/_annotation-card.scss
```diff
@@ -0,0 +1,60 @@
...
+}
+
+.annotation-card__username {
+  color: #202020;
+  font-weight: bold;
+}
+
+.annotation-card__timestamp {
+  color: $grey-4;
+  font-size: $small-font-size;
```
robertknight: You can use the `font-small` mixin here to set the font-size and line-height together

h/static/styles/partials-v2/_annotation-card.scss
```diff
@@ -0,0 +1,60 @@
...
+  line-height: $small-line-height;
+}
+
+.annotation-card__quote {
+  padding: 0 1em;
```
robertknight: For consistency this should use `px` rather than `em` for the padding.

h/static/styles/partials-v2/_search.scss
```diff
@@ -0,0 +1,91 @@
...
+
+.search {
+  font-size: $normal-font-size;
+  margin-top: 20px;
+  color: $grey-6;
+}
+
+.search-results {
+  list-style: none;
+  width: 950px;
```
robertknight: Don't forget that everything we build for the new site needs to work on mobile. That means that it needs to work on devices with a screen size of 320px and up.

If you want to limit the maximum width, use `max-width` rather than `width` for this reason. Of course, we might need to use a completely different layout for smaller screens, in which case that can be done with media queries at the bottom of the file. See `form-input` for an example.

h/static/styles/partials-v2/_search.scss
```diff
@@ -0,0 +1,91 @@
+// Search page
+// -----------
+// Styles for the search page which displays the results
+// of activity queries.
+
+.search {
```
robertknight: CSS component names should generally be nouns, so `search-result-container` rather than `search` for example.

h/static/styles/partials-v2/_search.scss
```diff
@@ -0,0 +1,91 @@
...
+  font-size: $normal-font-size;
+  margin-top: 20px;
+  color: $grey-6;
+}
+
+.search-results {
```
robertknight: Again, component names should be nouns, so `search-result-list` perhaps.

h/static/styles/partials-v2/_search.scss
```diff
@@ -0,0 +1,91 @@
...
+  list-style: none;
+  width: 950px;
+}
+
+.search-result {
```
robertknight: I would probably change the naming here slightly because this component doesn't refer to a single search result (which would be one annotation card) but a _bucket_ of search results (to use the terminology used in the Python code) which all belong to the same document and were all updated in the same time frame. `search-result-bucket` perhaps?

There was a lengthy discussion on Slack a while back about how search is going to work. Conceptually the important thing that we (and users) need to understand is that it searches annotations and then groups/"buckets" the result by document and last-update time.

h/static/styles/partials-v2/_search.scss
```diff
@@ -0,0 +1,91 @@
...
+
+.search-result__domain {
+  width: 130px;
+  text-indent: 10px;
+  color: $grey-4;
+}
+
+.search-result__title {
+  font-weight: bold;
+  width: 695px;
```
robertknight: Be careful about using large fixed widths as these are likely to break on mobile

h/static/styles/partials-v2/_search.scss
```diff
@@ -0,0 +1,91 @@
...
+
+.search-result__annotation-cards-info {
+  display: flex;
+  flex-direction: row;
+  justify-content: space-around;
+  margin-top: 10px;
+}
+
+.search-result__annotation-cards {
+  width: 500px;
```
robertknight: Ditto comments about `width` vs `max-width` above

sheetaluk:
@robertknight To test on mobile:
http://localhost:5000/search?q=h

I also used the "Toggle device mode" on chrome dev tools.

robertknight:
I've had another look at this. The first comment I have is that responsiveness is not just about supporting a set of specific screen sizes but making sure that the layout works well across a range of sizes within some bounds.

As it stands at the moment, if I view the layout at 320px it looks fine. If I have a device that has a width of 400-500px the layout looks odd because there is more padding on the left than right. On devices that are 500-850px (ie. most small/medium sized tablets (!)) then the right-hand side of the annotation card is cut off. Between 850 and 990px everything is visible but there is extra space on the left.

As a general rule of thumb, its a good idea to test layouts on at least:

- Small phone (iPhone 5 preset)
- Bigger phone/phablet (Samsung Galaxy or iPhone 6 preset)
- Medium tablet (iPad Mini/Standard or Nexus 7/9)
- Desktop

h/static/styles/partials-v2/_search.scss
```diff
@@ -0,0 +1,128 @@
+// Search page
```
robertknight: Styles apply to components rather than pages, the difference being that a UI component might be re-used across multiple pages. For example, we'll probably have views in future that show single annotations or all annotations from a document and these would use the same components, though they may or may not be served at the `/search` URL.

So I would suggest here that the description should be something like "Annotation search result list".

h/static/styles/partials-v2/_search.scss
```diff
@@ -0,0 +1,128 @@
...
+.search-result__annotations-count-container {
+  width: 30px;
+  padding: 5px 0px;
+  background-color: $grey-2;
+  text-align: center;
+  border-radius: 2px;
+  float: right;
+}
+
+.search-result__annotations-count {
```
robertknight: s/annotations-count/annotation-cunt

h/static/styles/partials-v2/_search.scss
```diff
@@ -0,0 +1,128 @@
...
+.search-result__metadata-key {
+  color: $grey-4;
+}
+
+.search-result__metadata-val {
+  color: $grey-6;
+  margin-bottom: 20px;
+}
+
+@media screen and (max-width: $max-phone-width) {
```
robertknight: I find it helpful to have a description in the CSS file of how the responsive layout works from a user's perspective. See `_nav-bar.scss` or `_form-input.scss` for example.

h/static/styles/partials-v2/_search.scss
nickstenning: :joy:

robertknight: :facepalm:

robertknight:
LGTM

PR #3692, opened 2016-08-15
Account settings form styling: Change email and password

finitud:
This PR contains two parts:

* A small refactoring that extracts explicit calls to `form.render()` in the templates and places them in the view controller.

* Pre-populating the email field with the user's email, by calling `#render()` with an `appstruct` parameter in the `activity_pages` form.

### DIFF
--- a/h/templates/accounts/account.html.jinja2
+++ b/h/templates/accounts/account.html.jinja2
@@ -30,17 +30,19 @@
     {%- endif %}
     {% if feature('activity_pages') -%}
     <footer class="form-footer">
+      {% trans %}
+        If you would like to delete your account, please email us at
+        <a class="link" href="mailto:support@hypothes.is">support@hypothes.is</a> from your
+        registered email address, and we'll take it from there.
+      {% endtrans %}
+    </footer>
     {% else %}
     <legend>
-    {%- endif %}
       {% trans %}
         If you would like to delete your account, please email us at
         <a href="mailto:support@hypothes.is">support@hypothes.is</a> from your
         registered email address, and we'll take it from there.
       {% endtrans %}
-    {% if feature('activity_pages') -%}
-    </footer>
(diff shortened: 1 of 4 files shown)


h/templates/accounts/account.html.jinja2
```diff
@@ -5,20 +5,25 @@
...
 
+    {% if not feature('activity_pages') -%}
     <h2 class="form-heading">
       <span>{% trans %}Change your password{% endtrans %}</span>
     </h2>
-    {{ password_form.render() }}
+    {%- endif %}
+    {{ password_form }}
 
+    {% if not feature('activity_pages') -%}
```
robertknight: Not showing this matches the mocks but I think we should continue to provide users with information on what to do to delete their account. My guess is that this would be done with some text and a link in the footer of the form (see the `/signup` page for an example of what that looks like). Thoughts @conordelahunty ?

h/templates/layouts/account.html.jinja2
```diff
@@ -17,7 +17,6 @@
   {{ panel('navbar') }}
   {% endif %}
   <div class="content paper">
-    {% include "h:templates/includes/logo-header.html.jinja2" %}
```
robertknight: Doesn't need to be fixed in this PR but I'll note that the tabs now touch the navbar and need some vertical spacing above them. For the logged-out pages this is provided by the `form-header` class.

robertknight:
I have a query about what we do with the instructions on how to delete an account. Otherwise LGTM.

h/templates/layouts/account.html.jinja2
finitud: Ah, thanks. This spacing is why I had still marked the PR as WIP.

robertknight:
LGTM

PR #3693, opened 2016-08-16
Fix account settings tab bar styling

finitud:
The `form-header` class was being applied to the whole form. It has
been removed, since it was only being used for adding the proper spacing
before the tab bar, and a margin has been added to the `tabs` class
to correct this.


robertknight:
LGTM

PR #3694, opened 2016-08-16
Account settings form styling: Dev API token

finitud:


h/static/styles/partials-v2/_api-token.scss
```diff
@@ -0,0 +1,11 @@
+.api-token {
+  border: 1px solid $grey-3;
```
robertknight: The style guide doesn't specify this at all, but what I've been starting to do recently is organize properties into two blocks:

```
.selector {
  <properties that affect the box of the element (position, display, width, height etc.)>

  <properties that affect the internal styling of the element (border, background color, color, font-size etc.)>
}
```

h/templates/accounts/developer.html.jinja2
```diff
@@ -5,25 +5,55 @@
...
+    {% if not feature('activity_pages') -%}
     <h2 class="form-heading">
       <span>{% trans %}Generate your API token{% endtrans %}</span>
     </h2>
+    {% endif %}
 
     {% if token %}
+      {% if feature('activity_pages') %}
+        <p>
+        <label for="token">{% trans %}Your API token is:{% endtrans %}</label>
```
robertknight: Looking at the mocks, this label and the paragraph below should use `$grey-5`.

h/static/styles/partials-v2/_api-token.scss
```diff
@@ -0,0 +1,14 @@
...
+  border-radius: 3px;
+  display: block;
+  min-height: 50px;
+  width: 100%;
+
+  color: $grey-6;
+  text-align: center;
+}
+
+.api-token--text {
```
robertknight: The `--modifier` here would indicate a different version of `api-token`, perhaps to indicate a read only vs read/write token for example.

I think we probably want a `form-help-text` component, that can go in `_form.scss` and we can then use it for things like the hint text in https://trello.com/c/gGtgaRg9/637-form-field-hint-enhancement on touch devices.

h/static/styles/partials-v2/_api-token.scss
```diff
@@ -0,0 +1,14 @@
+.api-token {
+  border: 1px solid $grey-3;
+  border-radius: 3px;
+  display: block;
+  min-height: 50px;
+  width: 100%;
+
+  color: $grey-6;
+  text-align: center;
```
robertknight: The font size looks larger than the default in the mock at https://trello-attachments.s3.amazonaws.com/577a6fb9667c37f497050532/1440x800/e84807532e003d6a2d2ed386af011405/settings_-_developer_regenerate.png . I think this would be `$input-font-size`

h/templates/accounts/developer.html.jinja2
```diff
@@ -5,25 +5,55 @@
...
       <legend>
         {% trans %}You haven't generated an API token yet.{% endtrans %}
       </legend>
+      {%- endif %}
     {% endif %}
 
       <form method="POST">
+        {% if feature('activity_pages') %}
+          {% if token %}
+            <button type="submit" class="btn btn-danger"
```
robertknight: `btn-danger` doesn't exist in the new styles (it would be `btn--danger` as a variant of btn). Since the styling isn't any different for the "Regenerate your token" button we can just remove this class.

h/templates/accounts/developer.html.jinja2
```diff
@@ -5,25 +5,55 @@
...
+      {% if feature('activity_pages') %}
+        <p class="form-help-text">
+        <label for="token">{% trans %}Your API token is:{% endtrans %}</label>
+        </p>
+        <input id="token"
+               type="text"
+               value="{{ token }}"
+               readonly="readonly"
+               class="api-token">
+        <p class="help-block form-help-text">
```
robertknight: `grep` tells me that `help-block` is defined in three templates but not actually styled anywhere (except in `front-page.css` which is not used in any of the three pages that reference this class), so we can remove it.

robertknight:
LGTM

PR #3696, opened 2016-08-16
Extract signup email into `h.emails`

nickstenning:
The generation of the email we send to users at signup has, until now,
been hard-coded in `h.accounts.views`. This commit extracts this into
the generic `h.emails` package, adding an HTML format as a bonus.

### DIFF
--- a/h/emails/__init__.py
+++ b/h/emails/__init__.py
@@ -1,5 +1,5 @@
 # -*- coding: utf-8 -*-
 
-from h.emails import reply_notification
+from h.emails import reply_notification, signup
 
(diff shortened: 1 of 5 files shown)


nickstenning:
Note that this PR doesn't actually remove the old implementation of this email from `h.accounts.views`. That'll come in a subsequent PR.

robertknight:
LGTM. It feels a little bit icky to me that code for rendering a trivial email template is tied to the Pyramid request object and all that comes with it, but hey - we can refactor that if it becomes a problem.

PR #3697, opened 2016-08-16
Add migration for auth token "expires" column

nickstenning:
This PR adds a schema migration to add an "expires" datetime column to the token table. By merging this first as a nullable column we can ensure zero-downtime for subsequently merged/deployed model changes.


finitud:
LGTM 👍

PR #3698, opened 2016-08-16
Authentication token expiry

nickstenning:
This PR adds the ability for auth tokens to expire. We don't actually issue
expiring auth tokens just yet, but we'll need to for third-party accounts.

In order to do this we:

- add an `expires` column to the `Token` model
- add an `is_valid()` method to the `IAuthenticationToken` interface, and
  implement it for the two token types we have currently
- check `is_valid()` in the `TokenAuthenticationPolicy`

Current "dev tokens" will have NULL token expiry dates and thus will never be
considered expired.

~~_This PR depends on code from #3695 and should not be merged until #3697 has
been merged, deployed, and the schema migration run in production._~~

### DIFF
--- a/h/auth/interfaces.py
+++ b/h/auth/interfaces.py
@@ -9,3 +9,6 @@ class IAuthenticationToken(Interface):
     """Represents an authentication token."""
 
     userid = Attribute("""The userid to which this token was issued.""")
+
+    def is_valid(self):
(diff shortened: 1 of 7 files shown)


nickstenning:
Prerequisites now merged, deployed, and migration run. This is ready to go.

robertknight:
LGTM

PR #3700, opened 2016-08-16
Make handle_form_submission() generic

seanh:
Move `handle_form_submission()` into `h.form` so it can be imported but multiple `views.py` files that want to use it.

Also allow passing in a callable that provides the return value when the form submission is successful and the request is not an XHR request, instead of always redirecting to `request.url`. This makes `handle_form_submission()` usable by a wider variety of form pages.

Also some improvements to the test code.

### DIFF
--- a/h/form.py
+++ b/h/form.py
@@ -104,16 +104,22 @@ def handle_form_submission(request, form, on_success, on_failure):
     :param form: the form that was submitted
     :type form: deform.form.Form
 
-    :param on_success: A callback function to be called if the form validates
-        successfully. This function should carry out the action that the form
-        submission requests. For example for a change password form, this
-        function would change the user's password.
+    :param on_success:
+        A callback function to be called if the form validates successfully.
+
+        This function should carry out the action that the form submission
+        requests (for example for a change password form, this function would
+        change the user's password) and return the view callable result that
+        should be returned if this is not an XHR request.
+
+        If on_success() returns ``None`` then ``handle_form_submission()``
+        will return ``HTTPFound(location=request.url)`` by default.
     :type on_success: callable
 
-    :param on_failure: A callback function that will be called if form validation
-        fails in order to get the view callable result that should be returned.
-        Note that the result returned by on_failure() will *not* be used if the
-        request is an XHR request.
+    :param on_failure:
+        A callback function that will be called if form validation fails in
+        order to get the view callable result that should be returned if this is
+        not an XHR request.
     :type on_failure: callable
 
     """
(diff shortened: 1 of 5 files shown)


tests/h/accounts/views_test.py
```diff
@@ -241,24 +79,26 @@ def test_post_no_event_when_validation_fails(self,
...
-    def test_post_redirects_when_validation_succeeds(self, pyramid_config, pyramid_request):
+    def test_post_redirects_when_validation_succeeds(self,
+                                                     form_validating_to,
+                                                     pyramid_config,
+                                                     pyramid_request):
         pyramid_config.testing_securitypolicy(None)  # Logged out
         controller = views.AuthController(pyramid_request)
-        controller.form = form_validating_to(
-            {"user": FakeUser(username='cara')})
+        controller.form = form_validating_to({"user": FakeUser(username='cara')})
```
nickstenning: I'm just going to make the point again that including formatting changes in diffs like this is not helpful to reviewers. Please try and resist the temptation! 

Ideally a PR is either refactoring, in which case the goal of the reviewer is to identify changes that are not strictly behaviour-preserving, or a PR is changing behaviour, in which case the goal of the reviewer is to understand the changes that have been made.

tests/h/accounts/views_test.py
```diff
@@ -465,7 +315,8 @@ def test_post_generates_mail(self,
         reset_mail.assert_called_with(user, "faketoken", "http://example.com")
 
     @mock.patch('h.accounts.views.account_reset_email')
-    def test_post_sends_mail(self, reset_mail, mailer, pyramid_request):
+    def test_post_sends_mail(
+            self, reset_mail, form_validating_to, mailer, pyramid_request):
```
nickstenning: It's a small thing, but I don't really understand why half the time you're doing:

```python
def test_something(with,
                   many,
                   arguments):
```

and half the time doing


```python
def test_something(
    with, many, arguments):
```

Can we stick to the former?

h/form.py
```diff
@@ -89,6 +95,78 @@ def configure_environment(config):  # pragma: no cover
...
+        appstruct = form.validate(request.POST.items())
+    except deform.ValidationFailure:
+        result = on_failure()
+    else:
+        result = on_success(appstruct)
+
+        if result is None:
+            result = httpexceptions.HTTPFound(location=request.url)
+
+        request.session.flash(_("Success. We've saved your changes."),
```
nickstenning: I'm not entirely sure this is going to turn out to be the right place for this, but I guess we can deal with that when it becomes a problem.

nickstenning:
Thanks for splitting it out! Looks good to me. Feel free to push small cleanups to master if you wish.

tests/h/accounts/views_test.py
seanh: If the former doesn't fit within 80 chars then I fall back on the latter. In some cases though it may just be inconsistency

h/form.py
seanh: Could become an optional param too

robertknight:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;MxOJSA6b&#x2F;413-return-form-html-fragments-for-ajax-requests"> Return form HTML fragments for Ajax requests</a>

PR #3701, opened 2016-08-16
Enable submitting create group form via XHR

seanh:
If a `POST` request submitting the create new group form is an `XMLHttpRequest` (it has a `X-Requested-With: XMLHttpRequest` header) then return just the `<form> ... </form>` element as an HTML snippet, rather than returning the entire HTML page (in the case of an invalid form submission) or a redirect (in the case of an valid form submission).

Depends on https://github.com/hypothesis/h/pull/3700

### DIFF
--- a/tests/functional/test_groups.py
+++ b/tests/functional/test_groups.py
@@ -6,7 +6,7 @@ import pytest
 
 
 @pytest.mark.functional
-def test_group_page_includes_referrer_tag(app, db_session, factories):
+def test_group_page_includes_referrer_tag(app, db_session, factories, user):
     """
     The group read page should include a referrer tag.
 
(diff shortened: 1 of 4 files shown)


tests/functional/test_groups.py
```diff
@@ -6,7 +6,7 @@
 
 
 @pytest.mark.functional
-def test_group_page_includes_referrer_tag(app, db_session, factories):
+def test_group_page_includes_referrer_tag(app, db_session, factories, user):
```
nickstenning: I don't think this parameter is needed here?

nickstenning:
LGTM.

nickstenning:
Actually, while the code here is fine, just playing around with this makes me realise that there's no easy way (currently) for the JavaScript to tell whether or not the form submission was successful or not. In both cases the result is `200 OK`, with `Content-Type: text/plain`. It's possible/probably that we'll need to adjust this in future so that the front-end code can easily distinguish between these two outcomes.

robertknight:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;MxOJSA6b&#x2F;413-return-form-html-fragments-for-ajax-requests"> Return form HTML fragments for Ajax requests</a>

PR #3703, opened 2016-08-17
Allow toggling features on with a query string param

nickstenning:
This small commit allows you to toggle a feature flag on by setting a
__feature__[featurename] query parameter. This can be useful for
toggling a feature flag on for requests made while logged-out.


finitud:
Awesome! 👍

PR #3706, opened 2016-08-17
Add a UserSignupService

nickstenning:
In order to support API signups, we need to be able to sign up users from places other than `h.accounts.views`.

This PR adds a `UserSignupService` which acquires the responsibility for this task, and thus greatly simplifies the corresponding views in `h.accounts`.

### DIFF
--- a/tests/h/accounts/services_test.py
+++ b/tests/h/accounts/services_test.py
@@ -1,0 +1,122 @@
+# -*- coding: utf-8 -*-
+
+from __future__ import unicode_literals
+
+import mock
+import pytest
+
+from h.accounts.services import UserSignupService, user_signup_service_factory
+from h.models import Activation, Subscriptions, User
+
+
+class TestUserSignupService(object):
+    def test_signup_returns_user(self, svc):
+        user = svc.signup(username='foo', email='foo@bar.com', password='baz')
+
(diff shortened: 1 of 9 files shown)


tests/h/accounts/services_test.py
```diff
@@ -0,0 +1,122 @@
...
+        user = svc.signup(username='foo', email='foo@bar.com', password='baz')
+
+        signup_email.assert_called_once_with(id=user.id,
+                                             email='foo@bar.com',
+                                             activation_code=user.activation.code)
+
+    def test_signup_sends_email(self, mailer, svc):
+        svc.signup(username='foo', email='foo@bar.com', password='baz')
+
+        mailer.send.delay.assert_called_once_with(['test@example.com'],
```
robertknight: It confused me initially that the email was different than that passed to `signup` but I see that this is the hardcoded return value of the `signup_email` fake

robertknight:
LGTM. Much nicer.

tests/h/accounts/services_test.py
nickstenning: I suppose this is a bit confusing, but it's intentional -- the email should be sent to the email address returned by `signup_email`, not (necessarily) to the email provided to `.signup(...)`.

PR #3707, opened 2016-08-17
Add 'authority' column to user table

nickstenning:
This PR adds two migrations:

1. To add a nullable 'authority' column to the user table.

2. To fill the authority column with 'hypothes.is' for all rows in the user table.

This authority column will be used to distinguish between our own service users and the users of accounts generated by the "third-party accounts" system. In the mean time, this change actually makes handling of userids within the `h` code substantially simpler, by making it possible for a `User` object to know its own userid. Subsequent PRs will make these changes.

By deploying and running these migrations before we make the corresponding changes to the user model, we can ensure a zero-downtime rollout of that new code.


robertknight:
LGTM

PR #3708, opened 2016-08-17
Move activity page core logic into `h.activity.query`

nickstenning:
This is still pre-production code, but this PR:

1. Moves the core data retrieval for activity pages into its own module.
2. Improves the efficiency of retrieving group information.


robertknight:
LGTM

PR #3709, opened 2016-08-17
Add an 'unparse' function for string search queries

nickstenning:
This commit adds a function which can be used to transform a dict or MultiDict object representing a search query back into a string query. This will be useful for activity pages because it allows us to manipulate the user's query (by, for example, removing terms) and then turn it back into a string where it can be used in a URL.

### DIFF
--- a/src/memex/search/parser.py
+++ b/src/memex/search/parser.py
@@ -76,6 +76,26 @@ def parse(q):
     return MultiDict([m for m in parse_results if isinstance(m, Match)])
 
 
+def unparse(q):
+    """
+    Turn a dict-like object into a Lucene-like query string.
+
+    This can be considered the reverse of the
+    :py:func:`memex.search.parser.parse` function, as it can be used to
+    transform the MultiDict returned from that function back into a string
+    query.
+    """
+    terms = []
+
+    for key, val in q.iteritems():
+        if key == 'any':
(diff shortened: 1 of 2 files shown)


src/memex/search/parser.py
```diff
@@ -76,6 +76,31 @@ def parse(q):
...
+
+    This can be considered the reverse of the
+    :py:func:`memex.search.parser.parse` function, as it can be used to
+    transform the MultiDict returned from that function back into a string
+    query.
+    """
+    terms = []
+
+    try:
+        items = q.iteritems(multi=True)
```
robertknight: The `webob.multidict.MultiDict` object returned by `memex.search.parser.parse` does not support a 'multi' keyword argument. It looks like that is a feature of `werkzeug.datastructures.MultiDict`. Is there any reason that we would be using that type with this function?

robertknight:
Aside from the query about the `multi` argument to `iteritems` this LGTM.

src/memex/search/parser.py
nickstenning: That is a *very* good point. Good spot. Given all the tests are passing I guess that `iteritems` from `webob.multidict.MultiDict` does "the right thing". I'll check.

nickstenning: Fixed.

PR #3710, opened 2016-08-17
Add user/group activity pages and auto-redirects

nickstenning:
This PR adds user and group activity pages at

    /groups/{pubid}/search
    /users/{username}/search

which automatically filter the searches on these pages by the relevant group or user. These are not necessarily final URLs for these pages.

It also adds functionality to the main activity page route at /search so that if a query is entered containing a single user or group term, then the user is redirected to the relevant user or group activity page, with their query appropriately modified.

For example, a request to

     /search?q=group:abcd1234+tag:foo

will be automatically redirected to

     /groups/abcd1234/search?q=tag:foo

~~There are some open questions here about how the searches on these pages work -- specifically around how URLs such as~~

     /groups/abcd1234/search?q=group:efgh4567

~~should behave. The current behaviour is almost certainly not correct.~~

If a group is specified in the query to a group search page, such as

     /groups/abcd1234/search?q=group:efgh4567

then it will simply be ignored. Displaying the union of the results for two groups on a specialised group search page doesn't make a lot of sense.

~~_This PR depends on work that's been split out into #3708 and #3709 and will need rebasing once those have merged._~~

### DIFF
--- a/h/activity/__init__.py
+++ b/h/activity/__init__.py
@@ -5,3 +5,5 @@ def includeme(config):
     config.include('h.activity.views')
 
     config.add_route('activity.search', '/search')
+    config.add_route('activity.group_search', '/groups/{pubid}/search')
+    config.add_route('activity.user_search', '/users/{username}/search')
--- a/h/activity/query.py
+++ b/h/activity/query.py
@@ -5,11 +5,13 @@ from __future__ import unicode_literals
 from collections import namedtuple
 
 from memex.search import Search
+from memex.search import parser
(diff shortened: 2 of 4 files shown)


robertknight:
LGTM. Just need to resolve one minor query around #3709 and then this can be merged.

nickstenning:
Rebased and updated with what I think is a slightly more reasonable default behaviour for the "group query" to a group page.

robertknight:
The new behavior seems sensible to me. LGTM

PR #3711, opened 2016-08-18
Cleanup layout issues in the document component.

sheetaluk:
Fixed the following issues:
1. Long domain names overflowing into the neighbouring container.
2. Images in the annotation body overflowing out of the parent container.
3. Annotator usernames showing  `@domain`.
4. Not enough margin between a long title and the first annotation card under it.

Question:
Should we wrap a title which is longer than the width of an annotation card under it or does this look fine for now:
https://www.dropbox.com/s/oiz1h9s5q5xqy0v/Screenshot%202016-08-17%2021.15.20.png?dl=0


h/templates/activity/search.html.jinja2
```diff
@@ -26,7 +26,7 @@
   </div>
   <ul class="search-bucket-stats__val">
     {% for user in bucket.users %}
-      <li>{{ user.split(':')[1] }}</li>
+      <li>{{ user.split(':')[1].split('@')[0] }}</li>
```
robertknight: OK for the moment, in future we'll probably want to extract this out into a helper or pre-process `bucket.users` before rendering the result in the template.

h/static/styles/partials-v2/_annotation-card.scss
```diff
@@ -54,3 +54,9 @@
   letter-spacing: 0.1px;
   border-left: 3px solid #d3d3d3;
 }
+
+.annotation-card__text {
+  img {
```
robertknight: This works, although there are a few other tags that are also styled inconsistently with the stream/client such as `<code>` and `<blockquote>`.

In the client, the `markdown-body` class which displays annotation markdown gets most of its styling by importing the `styled-text` mixin, which still exists in this repository in `partials/_styled-text.scss`.

What we could do is:

1. Extract the mixin from `partials/_styled-text.scss` into `mixins/_styled-text.scss` and add that to `mixins/_all.scss`
2. Include the mixin in the `annotation-card__text` class.

This will mean that all of the styling for annotation cards is in one mixin file and if we make changes to it, we can easily update the copy in the client repository as well.

robertknight:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;TDZE8zFu&#x2F;416-document-component"> Document component</a>

h/static/styles/mixins/_styled-text.scss
```diff
@@ -84,7 +84,7 @@
   blockquote {
     @include font-normal;
 
-    border-left: 3px solid $gray-lighter;
+    border-left: 3px solid $grey-3;
```
sheetaluk: Since we don't have $gray-lighter in the palette any more, I have replaced it with $grey-3.

h/static/styles/mixins/_styled-text.scss
```diff
@@ -111,7 +111,7 @@
   pre code {
     padding: 10px;
     display: block;
-    background-color: $gray-lightest;
+    background-color: $grey-1;
```
sheetaluk: We don't have $gray-lightest in the palette anymore, so I have replaced it with $grey-1.

h/static/styles/partials-v2/_annotation-card.scss
```diff
@@ -52,5 +52,9 @@
   font-size: 12px;
   font-style: italic;
   letter-spacing: 0.1px;
-  border-left: 3px solid #d3d3d3;
+  border-left: 3px solid $grey-3;
```
sheetaluk: Using $grey-3 from our palette instead of \#d3d3d3.

h/static/styles/partials-v2/_base.scss
```diff
@@ -19,6 +19,7 @@ $highlight: #58cef4;
 // ----------
 
 $sans-font-family: "Helvetica Neue", Helvetica, Arial, "Lucida Grande", sans-serif !default;
+$mono-font-family: Open Sans Mono, Menlo, DejaVu Sans Mono, monospace !default;
```
sheetaluk: styled-text mixin uses $mono-font-family, so moved it over too.

robertknight:
LGTM

robertknight:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;TDZE8zFu&#x2F;416-document-component"> Document component</a>

PR #3712, opened 2016-08-18
Collapse document buckets until toggled by clicking document titles 

robertknight:
This implements the behavior described at https://trello.com/c/TDZE8zFu/416-document-component to collapse search result buckets until toggled by clicking the bucket titles.

~~While working on this I found it more natural to implement the 'expand by default' behavior by changing the way the environment classes on the `<html>` element work, to have an explicit `no JS` class and then style that in a selector that also covers the 'explicitly expanded' case.  The first commit in this PR makes that change and the second implements the collapsing-by-default for the search result buckets.  When testing this, you can flip between the initial states in JS/non-JS mode by setting the `<html>` element's classes to `env-js` or `env-no-js` respectively.~~ (We decided against this change in the end for the time being, see https://botbot.me/freenode/hypothes.is/msg/71520214/ )

### DIFF
--- a/h/static/scripts/base/environment-flags.js
+++ b/h/static/scripts/base/environment-flags.js
@@ -68,8 +68,6 @@ EnvironmentFlags.prototype.ready = function () {
   if (this._jsLoadTimeout) {
     clearTimeout(this._jsLoadTimeout);
   }
-  this.set('js-ready', true);
-  this.set('js-timeout', false);
 };
 
 module.exports = EnvironmentFlags;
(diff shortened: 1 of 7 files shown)


robertknight:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;TDZE8zFu&#x2F;416-document-component"> Document component</a>

robertknight:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;TDZE8zFu&#x2F;416-document-component"> Document component</a>

nickstenning:
LGTM.

PR #3713, opened 2016-08-18
Add the "authority" column to the user model

nickstenning:
Having now added the column directly to the database, it is safe to add reference to the authority column to the user model.

We ensure that the authority column is filled in for new users using the `default_authority` parameter to the `UserSignupService`, which defaults to `request.auth_domain`.

This means that in production, new users will be created with an authority of "hypothes.is" as expected.

This PR also includes a migration to add the NOT NULL constraint to the authority column.

### DIFF
--- a/h/accounts/models.py
+++ b/h/accounts/models.py
@@ -68,6 +68,14 @@ class User(Base):
                           nullable=False,
                           unique=True)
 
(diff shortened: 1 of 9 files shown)


robertknight:
LGTM

PR #3714, opened 2016-08-18
Make password checking an instance method

nickstenning:
It is not at all clear to me why this was implemented as a classmethod accepting an instance as its first argument, rather than an instance method.

Making this an instance method makes it substantially easier to use in tests, especially when the user object isn't mocked.


robertknight:
LGTM

PR #3715, opened 2016-08-18
Add a userid property to the User model

nickstenning:
Now that we store the authority part of the userid in the database, we can simply have this as a property of the model instances rather than requiring helper functions all over the place.


robertknight:
LGTM

PR #3716, opened 2016-08-18
Remove `h.util.user.userid_from_username`

nickstenning:
This helper function is no longer needed, as the user model now has its own "userid" property.

Now that the user model has its own "userid" property, this PR removes all uses of the aforementioned helper function.

_~~This PR depends on #3714 and #3715 and should not be merged until after those two. We should also wait until the associated `User.authority` migrations have been run in production.~~_

### DIFF
--- a/h/accounts/services.py
+++ b/h/accounts/services.py
@@ -5,7 +5,6 @@ from __future__ import unicode_literals
 from functools import partial
 
 from h import mailer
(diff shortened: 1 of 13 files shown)


nickstenning:
Updated, migrations run. Ready to go.

robertknight:
LGTM

PR #3718, opened 2016-08-18
Respond with 400 for invalid XHR form submissions

seanh:
For forms that use the handle_form_submission() helper to support
submitting forms via XHR, if the submitted form values are invalid then
set the response status to 400 rather than 200. This allows the client
to handle successful and unsuccessful form submission responses differently.

### DIFF
--- a/tests/functional/test_accounts.py
+++ b/tests/functional/test_accounts.py
@@ -30,7 +30,7 @@ class TestAccountSettings(object):
         email_form['email_confirm'] = 'new_email@example.com'
         email_form['password'] = 'pass'
 
-        res = email_form.submit(xhr=True)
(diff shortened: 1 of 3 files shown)


tests/functional/test_accounts.py
```diff
@@ -30,7 +30,7 @@ def test_submit_email_form_with_xhr_returns_partial_html_snippet(self,
         email_form['email_confirm'] = 'new_email@example.com'
         email_form['password'] = 'pass'
 
-        res = email_form.submit(xhr=True)
+        res = email_form.submit(xhr=True, status=200)
```
seanh: This line already would have crashed if the response status code was non-successful, here I'm just making it explicit that that is tested.

robertknight:
LGTM

PR #3719, opened 2016-08-18
Backend support for search paginator

seanh:
This pull request refactors `h.paginate` to work with both SQLAlchemy query objects and Elasticsearch search results.

`h.paginate.paginate()` is renamed to `paginate_query()`, but otherwise has the same interface. The only change that needs to be made to the `/admin/groups` and `/admin/features/cohorts` pages that were already using `paginate()` is that they now call `paginate_query()` instead.

A new `paginate()` function containing all the logic that is _not_ specific to handling SQLAlchemy query objects is then factored out of `paginate_query()`. `paginate_query()` now calls `paginate()` for the pagination logic, and only does the SQLAlchemy query handling itself.

A new `paginate_es()` is then added that calls `paginate()` and handles Elasticsearch results.

`paginate_es()` is then used to add a (completely unstyled) paginator control to the `/search` page.

This pull request does not do the HTML and CSS work to style the paginator control on the search page.

This pr does not add the "eliding" feature to the paginator. It does not produce a paginator control that looks like this: `1 ... 149 [150] 151 ... 300`. It will list all 300 pages instead.

This pr does not actually paginate the results on the `/search` page. Pagination does continue to work on the `/admin/groups` and `/admin/features/cohorts` pages, but on the `/search` page you will see the first page's result on every page.

This pr _does_ add one new feature to the paginator (other than handling Es results as input): the page links that it generates preserve any other URL params (such as search queries) that were already in the URL.

### DIFF
--- a/h/activity/views.py
+++ b/h/activity/views.py
@@ -10,6 +10,7 @@ from pyramid import httpexceptions
 from pyramid.view import view_config
 
 from h.activity import query
+from h.paginator import paginate
 
 
 @view_config(route_name='activity.search',
@@ -40,6 +41,7 @@ def search(request):
         'total': result.total,
         'aggregations': result.aggregations,
         'timeframes': result.timeframes,
+        'page': paginate(request, result.total),
     }
 
 
--- a/h/paginator.py
+++ b/h/paginator.py
@@ -22,11 +22,15 @@ def paginate(request, total, page_size):
     next_ = current_page + 1 if current_page < page_max else None
     prev = current_page - 1 if current_page > 1 else None
 
+    def url_for(page):
+        return request.current_route_path(_query={'page': page})
+
     return {
         'cur': current_page,
         'max': page_max,
         'next': next_,
         'prev': prev,
+        'url_for': url_for,
     }
 
 
--- a/h/paginator.py
+++ b/h/paginator.py
@@ -8,6 +8,28 @@ import math
 PAGE_SIZE = 20
 
 
+def paginate(request, total, page_size):
+    page_max = int(math.ceil(total / page_size))
+    page_max = max(1, page_max)  # There's always at least one page.
+
+    try:
+        current_page = int(request.params['page'])
+    except (KeyError, ValueError):
+        current_page = 1
+    current_page = max(1, current_page)
+    current_page = min(current_page, page_max)
+
+    next_ = current_page + 1 if current_page < page_max else None
+    prev = current_page - 1 if current_page > 1 else None
+
+    return {
+        'cur': current_page,
+        'max': page_max,
+        'next': next_,
+        'prev': prev,
+    }
+
+
 def paginate_query(wrapped=None, page_size=PAGE_SIZE):
     """
     Decorate a view function, providing basic pagination facilities.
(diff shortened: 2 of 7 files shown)


seanh:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;fbvVXbvF&#x2F;417-backend-support-for-pagination-control"> Backend support for pagination control</a>

h/activity/views.py
```diff
@@ -40,6 +41,7 @@ def search(request):
         'total': result.total,
         'aggregations': result.aggregations,
         'timeframes': result.timeframes,
+        'page': paginate_es(request, result),
```
seanh: I'm not sure that this is how this will actually work in the end (passing the already-fetched search result into `paginate_es()`) because if it's the paginator that figures out what the offset and limit will be, then we need to call the paginator _before_ actually executing the search query, because we need to apply the offset and limit to the query before executing it. But for now this gives the paginator what it needs to generate the template data needed to render the pagination control (it needs to know the URL params and the total number of search results).

tests/h/paginator_test.py
```diff
@@ -1,158 +1,199 @@
...
+    def test_prev(self, pyramid_request, page_param, expected):
+        pyramid_request.params = {'page': page_param}
+        assert paginate(pyramid_request, 600, 10)['prev'] == expected
+
+
+@pytest.mark.usefixtures('paginate')
+class TestPaginateQuery(object):
+
+    CURRENT_PAGE = 3
+    "The current page that will be returned by paginate()."
```
nickstenning: I think a normal comment would do fine here, given this is test code...

tests/h/paginator_test.py
```diff
@@ -1,158 +1,199 @@
...
+
+        """
+        wrapped(mock.sentinel.context, pyramid_request)
+
+        query.limit.assert_called_once_with(self.PAGE_SIZE)
+
+    def test_it_returns_the_query_results(self,
+                                          wrapped,
+                                          pyramid_request):
+        assert wrapped(mock.sentinel.context, pyramid_request)['results'] == (
```
nickstenning: If your assertion is split over two lines like this, perhaps it would be better to write:

```python
results = wrapped(mock.sentinel.context, pyramid_request)['results']

assert results == mock.sentinel.all
```

tests/h/paginator_test.py
```diff
@@ -1,158 +1,199 @@
 # -*- coding: utf-8 -*-
 
+import mock
 import pytest
+from webob.multidict import NestedMultiDict
```
nickstenning: This belongs in the subsequent commit.

h/paginator.py
```diff
@@ -92,3 +92,7 @@ def wrapper(context, request):
             'page': page,
         }
     return wrapper
+
+
+def paginate_es(request, result, page_size=PAGE_SIZE):
```
nickstenning: I don't really see what this is adding.

I'd suggest adding the `PAGE_SIZE` default to the `paginate` function and then just using that from the view function.

nickstenning:
This mostly looks awesome! Just a few minor quibbles -- comments inline on the relevant commits.

h/paginator.py
seanh: Yeah, this was sort of a placeholder, because in future we're gonna have to actually paginate the search results not just show a pagination control, that'll be more complicated, and presumably this function would get bigger. But it does seem pointless now

tests/h/paginator_test.py
seanh: Hmm, I've changed this, but I don't see why documenting attrs in tests should be different from in other code, and we do use docstrings not comments on test classes and methods that need them...

tests/h/paginator_test.py
seanh: Done

tests/h/paginator_test.py
nickstenning: If you want to document attrs in tests the same way as in the rest of the code then use doc comments:

```python
#: An attr that does stuff
stuff_doer = ...
```

tests/h/paginator_test.py
seanh: Done

h/paginator.py
seanh: Done

seanh:
All done I think

nickstenning:
LGTM.

PR #3724, opened 2016-08-18
Don't set passwords when not necessary in tests

nickstenning:
Setting a password on a user is necessarily a slow operation. Doing this many times for test users substantially slows down the running of the test suite.

This PR:

1. relaxes the current database constraints on the password fields (see 04c4d79)
2. removes all the places in tests where we're setting passwords for users unnecessarily

In my testing this results in a ~1.7× speedup for our current test suite:

Timings for three runs of `tox -e py27-h` before this change:

    36s, 36s, 36s

Timings for three runs of `tox -e py27-h` after this change:

    20s, 21s, 22s

Speedup = 36s/21s ≈ 1.7×

### DIFF
--- a/h/accounts/models.py
+++ b/h/accounts/models.py
@@ -149,14 +149,11 @@ class User(Base):
         session.delete(self.activation)
 
     # Hashed password
-    _password = sa.Column('password', sa.UnicodeText(), nullable=False)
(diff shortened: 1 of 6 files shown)


robertknight:
Great stuff. I see similar timings on master vs. this branch when running the tests. Presumably in future it will also be the case that third-party accounts or accounts using SSO will not have passwords assigned?

PR #3728, opened 2016-08-19
Fix SASS watch & rebuild (take 2)

robertknight:
This fixes CSS rebuilding in the `gulp watch` task while preserving the behavior change from before to avoid killing the asset build process if the SASS build fails.

The actual problem was that the `compileSass()` function was not returning the list of imported files after some last-minute refactoring, so only the root files in the bundle ever got watched. This PR fixes the problem by switching back to just using a dumb glob watch and rebuilding all style bundles whenever any `.scss` file changes, with a small optimization to avoid reprocessing vendor CSS when nothing changed (see second commit).

`watchify`-like smartness where a rebuild only happens when a bundle dependency changes is nice, but we can get away without that for the time being.

----

Testing:
  1. Run `make dev`
  2. Add, edit or remove style files in `h/static/styles` and check that `build-css` task runs
  3. Make changes with errors in style files (eg. Add an a new line with `{}` to any imported file) and check that asset build continues running

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -165,10 +165,12 @@ gulp.task('build-css', ['build-vendor-css'], function () {
   return Promise.all(styleBundleEntryFiles.map(buildStyleBundle));
 });
 
-gulp.task('watch-css', ['build-vendor-css'], function () {
-  styleBundleEntryFiles.forEach(function (input) {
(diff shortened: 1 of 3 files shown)


nickstenning:
LGTM.

PR #3717, opened 2016-08-18
Add a UserService for retrieving users

nickstenning:
With a view to replacing `accounts.get_user`, this PR introduces a UserService which will serve as a place to put basic user retrieval and manipulation routines.

PR #3729, opened 2016-08-19
Fix performance issues introduced with UserService

nickstenning:
The introduction of the user `userid` property and subsequently the `UserService` (in #3715 and #3717 respectively) had an unintended negative impact on the performance of the web application.

The reason for this is that `request.authenticated_userid` is not memoized (or "reified" in Pyramid parlance), so every read from that attribute resulted in a call to `h.auth.util.groupfinder`, which in turn resulted in a database query to retrieve the user object, mediated by the `UserService`.

This was true before, meaning that we have been retrieving the user object from the database multiple times per request for some time, but the addition of the `userid` property to the `User` model made the problem substantially worse because:

- the `authority` column wasn't indexed
- even if the `authority` column had been indexed, the `UserService` wouldn't have taken advantage of that index, instead emitting a very inefficient query

        # master
        $ wrk -c 1 -t 1 -d 60s -H 'Cookie: ...' https://localhost:5000/app
        Running 1m test @ https://localhost:5000/app
          1 threads and 1 connections
          Thread Stats   Avg      Stdev     Max   +/- Stdev
            Latency   376.17ms   93.69ms 970.41ms   71.97%
            Req/Sec     2.32      0.89     5.00     76.92%
          156 requests in 1.00m, 175.20KB read
        Requests/sec:      2.60
        Transfer/sec:      2.92KB
    
        # with UserService userid cache
        $ wrk -c 1 -t 1 -d 60s -H 'Cookie: ...' https://localhost:5000/app
        Running 1m test @ https://localhost:5000/app
          1 threads and 1 connections
          Thread Stats   Avg      Stdev     Max   +/- Stdev
            Latency   102.81ms   42.75ms 431.89ms   89.14%
            Req/Sec     9.62      3.67    20.00     78.17%
          552 requests in 1.00m, 619.92KB read
        Requests/sec:      9.19
        Transfer/sec:     10.32KB
    
        # with UserIDComparator
        $ wrk -c 1 -t 1 -d 60s -H 'Cookie: ...' https://localhost:5000/app
        Running 1m test @ https://localhost:5000/app
          1 threads and 1 connections
          Thread Stats   Avg      Stdev     Max   +/- Stdev
            Latency    89.64ms   39.52ms 318.01ms   93.90%
            Req/Sec    11.12      4.30    20.00     77.50%
          658 requests in 1.00m, 738.96KB read
        Requests/sec:     10.96
        Transfer/sec:     12.31KB
    
        # with both UserService userid cache and UserIDComparator
        $ wrk -c 1 -t 1 -d 60s -H 'Cookie: ...' https://localhost:5000/app
        Running 1m test @ https://localhost:5000/app
          1 threads and 1 connections
          Thread Stats   Avg      Stdev     Max   +/- Stdev
            Latency    50.77ms   28.00ms 320.11ms   94.54%
            Req/Sec    19.32      5.66    30.00     72.25%
          1144 requests in 1.00m, 1.25MB read
        Requests/sec:     19.04
        Transfer/sec:     21.38KB to fetch users by userid.


robertknight:
LGTM

PR #3730, opened 2016-08-19
Fix change email form crash

seanh:
validate_user() was renamed to check_password().


nickstenning:
I'm going to merge this so I don't break the forms when I deploy, but could you add a test for this?

PR #3734, opened 2016-08-19
Fix error when changing email to current email

seanh:
If the user submits the change email address form with their *current*
email address as the value for the new email address field then they
would get a "That email address is already in use" error message.

Fix this to a "We've saved your changes" success message.

Of course it doesn't make sense to change your email to its existing
value. But our way of looking at it is that the user asked for their
email to be set to a value, and it is set to that value, there's no
point in bothering the user with the fact that it was already set to
that value in the first place.


chdorner:
LGTM.

Issue #3732, opened 2016-08-19
Test needed for change email form crash (#3730)


seanh:
https://github.com/hypothesis/h/pull/3730#issuecomment-241052114

PR #3736, opened 2016-08-19
Add test for changing email with wrong password

seanh:
Fixes <https://github.com/hypothesis/h/issues/3732>.

This adds one test that would have caught the crash that <https://github.com/hypothesis/h/pull/3730> fixed, but there are no tests for the new `EmailChangeSchema`, more need to be added.


chdorner:
LGTM.

PR #3737, opened 2016-08-22
Replace use of `app.css` from client with a pre-built bundle for pages using legacy styles

robertknight:
This replaces the use of the `app.css` style bundle from the client on the account settings, signup, login, 404 and email unsubscription pages with a pre-built copy of app.css, named 'legacy-site.css'.

This will allow us to [remove the remaining CSS ](https://github.com/hypothesis/client/pull/80) in the client repository which is not actually used by the client itself and also a Jinja helper for loading client CSS assets into these pages.

----

Testing:

1. Turn off 'Activity Pages' feature flag
2. Visit account settings, login, signup and some URL that 404s, check that styling looks correct

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -130,6 +130,10 @@ gulp.task('build-vendor-css', function () {
     // page
     './h/static/styles/front-page.css',
 
+    // `legacy-site.css` is a pre-built bundle of legacy CSS used by the
+    // login and account settings pages
+    './h/static/styles/legacy-site.css',
+
     // Icon font
     './h/static/styles/vendor/icomoon.css',
     './node_modules/bootstrap/dist/css/bootstrap.css',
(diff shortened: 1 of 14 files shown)


nickstenning:
Ok, this looks fine. I guess you're ok with this file being unminified?

And I was going to point out that there's a bunch of other machinery we can remove but I guess that's not true as we still need `assets_client.ini` and the python stuff in `assets.py` in order to render `app.html`, true?

robertknight:
> Ok, this looks fine. I guess you're ok with this file being unminified?

Yes. We could pass the vendor CSS through a minifier but not much will be gained:

```
# (With one minor tweak to legacy-site.css to fix a parsing issue):

 ~/h/r/h > sass --scss build/styles/legacy-site.css | gzip | wc -c
10441
 ~/h/r/h > sass -t compressed --scss build/styles/legacy-site.css | gzip | wc -c
9005
```

> And I was going to point out that there's a bunch of other machinery we can remove but
> I guess that's not true as we still need assets_client.ini and the python stuff in assets.py in order
> to render app.html, true?

Yes, indeed.

PR #3745, opened 2016-08-22
Organize account schemas tests

seanh:
I want to add tests for the new `EmailChangeSchema` but I really want to add them in a `class TestEmailChangeSchema` so I can for example have per class fixtures. So organize the existing tests into classes first so that it'll be consistent.

Also tidied up the use of fixtures in these tests (define fixtures inside test classes when they're only relevant to that class, use class-level `usefixtures()` decorators...)

Also changed the names of many of the test methods to make the naming consistent.

I think this organization separates the tests for one schema from those for another more clearly, and also clarifies what our style / approach to testing these schemas is.

### DIFF
--- a/tests/h/accounts/schemas_test.py
+++ b/tests/h/accounts/schemas_test.py
@@ -29,351 +29,394 @@ class FakeInvalidSerializer(FakeSerializer):
         raise BadData("Invalid token")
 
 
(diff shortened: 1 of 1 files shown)


nickstenning:
LGTM.

PR #3746, opened 2016-08-22
User profile edit form

robertknight:
This implements the basic HTML version of the user profile edit form described at https://trello.com/c/HFGf3BGj/419-user-profile-edit-form-basics , with validation of the ORCID and URL fields.

### DIFF
--- a/h/accounts/schemas.py
+++ b/h/accounts/schemas.py
@@ -15,6 +15,7 @@ from h.accounts.models import (
     USERNAME_MIN_LENGTH,
     USERNAME_PATTERN,
 )
+from h.accounts import util
 
 _ = i18n.TranslationString
 log = logging.getLogger(__name__)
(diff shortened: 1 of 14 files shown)


h/accounts/schemas.py
```diff
@@ -323,6 +340,50 @@ def validator(self, node, value):
             raise exc
 
 
+def validate_url(value):
+    return wrap_validator(util.validate_url)(value)
```
nickstenning: It would probably be a little neater to use the `wrap_validator` function you've written as a decorator. You can actually do this right now:

```python
@wrap_validator
def validate_url(value):
    <body of util.validate_url>
```

But that wouldn't be ideal, as debugging the resulting `validate_url` would be tricky due to the way the decorator would change the function. To fix that you'd add one line to your `wrap_validator` decorator definition:

```python
import functools

def wrap_validator(validator):
    @functools.wraps(validator)
    def validate(value):
        ...
    return validate
```

h/accounts/util.py
```diff
@@ -0,0 +1,74 @@
...
+    if not re.match('https?', parsed_url.scheme):
+        raise ValueError('Links must have an "http" or "https" prefix')
+
+    if not parsed_url.netloc:
+        raise ValueError('Links must include a domain name')
+
+    return parsed_url.geturl()
+
+
+def orcid_checksum_digit(orcid):
```
nickstenning: This is a helper function so should probably go to the bottom of the file and be named `_ordcid_checksum_digit`.

h/accounts/util.py
```diff
@@ -0,0 +1,74 @@
...
+    """
+    Return the checksum digit for an ORCID identifier.
+
+    Translated from the example ISO 7064 checksum implementation at
+    http://support.orcid.org/knowledgebase/articles/116780-structure-of-the-orcid-identifier
+    """
+    total = 0
+    digits = [int(ch) for ch in orcid if ch.isdigit()]
```
nickstenning: I'm not sure `isdigit()` is quite what you're looking for here. In Python 3:

```
>>> '²'.isdigit()
True
```

(That's a Unicode superscript character.)

Given the hard restrictions on valid inputs to this function, I'd be inclined to do something like this in `validate_orcid`:

```python
orcid_normalised = orcid.replace('-', '')

if not _validate_iso7064_mod_11_2(orcid_normalised):
    raise ValueError('{} is not a valid ORCID'.format(orcid))
```

and then define

```python
def _validate_iso7064_mod_11_2(value):
    check = 0
    for n in value:
        check = (2 * check + int(10 if n == 'X' else n)) % 11
    return check == 1
```

h/accounts/util.py
```diff
@@ -0,0 +1,74 @@
...
+    for digit in digits:
+        total = (total + digit) * 2
+    remainder = total % 11
+    result = (12 - remainder) % 11
+
+    if result == 10:
+        return 'X'
```
robertknight: Done - https://github.com/hypothesis/h/pull/3746/files#diff-f6d167a4ba6d4ac46716fadbbd0fddffR26

h/accounts/util.py
```diff
@@ -0,0 +1,77 @@
...
+    Returns the normalized ORCID if successfully parsed or raises a ValueError
+    otherwise.
+    """
+    ORCID_REGEX = '\A[0-9]{4}-[0-9]{4}-[0-9]{4}-[0-9]{3}[0-9X]\Z'
+
+    if not re.match(ORCID_REGEX, orcid):
+        raise ValueError('The format of this ORCID is incorrect'.format(orcid))
+
+    if _orcid_checksum_digit(orcid[:-1]) != orcid[-1:]:
+        raise ValueError('{} is not a valid ORCID'.format(orcid))
```
nickstenning: Repeating the value looks a bit odd in the form.

nickstenning:
LGTM.

PR #3749, opened 2016-08-22
Search results pager

sheetaluk:
Implement pagination states to better display search results.
    
    Always show at most 3 pages of results to the left and the
    right of the current page.
    Show an ellipsis on each side that has more than 3 pages of
    search results.
    
    Test scenarios:
    1) 5 or less pages of search results.
       >> All 5 page numbers are listed. Clicking through each page
          doesn't change the appearance of the list.
    2) 6 or more pages of search results.
       >> The current page and up to 3 page numbers to the left and
          right of the current page are shown. Any more pages are
          represented as an ellipsis char on either side.
    3) 1 page of search results.
       >> No pagination is displayed.
    4) No search query.
       >> No pagination is displayed
    
    Trello:
    https://trello.com/c/mANzbvei/418-pagination-control

PS: I split this branch into two. The other one `search-results-pager-styling` (https://github.com/hypothesis/h/pull/3750) deals with initial styling for the paginator.


h/templates/includes/paginator.html.jinja2
```diff
@@ -1,33 +1,47 @@
 <nav>
-  <ul class="pagination">
+  <ul class="search-results-pager">
```
nickstenning: This should probably not say `search-results` if this is to be a reusable component.

seanh:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;fbvVXbvF&#x2F;417-backend-support-for-pagination-control"> Backend support for pagination control</a>

sheetaluk:
At the moment the ellipses are not clickable. Do we want them to be?

robertknight:
I see an exception when visiting `http://localhost:5000/search` (note the lack of a query string parameter): `UndefinedError: 'page' is undefined`

robertknight:
> At the moment the ellipses are not clickable. Do we want them to be?

What would ellipses link to? Ideas are welcome but I can't see it being that useful for them to be links myself.

tests/h/paginator_test.py
```diff
@@ -19,6 +19,38 @@ def test_current_page_defaults_to_1(self, pyramid_request):
         assert page['cur'] == 1
 
     @pytest.mark.parametrize('page_param,expected', [
+      # If the current page is the first page.
```
robertknight: 👍 I like tests in this style, where test cases are essentially ASCII representations of what the UI should look like in different scenarios

robertknight:
Aside from the exception I reported above when there is no search query, this looks great!

robertknight:
In testing this, the pager control works fine but changing pages didn't seem to actually do anything - I always get the first page of results back. I see the same issue on the production Hypothesis service, so it looks like this isn't an issue in this PR.

sheetaluk:
@robertknight:
Fixed the `UndefinedError: 'page' is undefined` exception.

Also, pagination hasn't been implemented yet. That is why we keep seeing the same set of results returned.

robertknight:
LGTM

PR #3750, opened 2016-08-22
Search results pager styling

sheetaluk:
Initial styling for search results pagination.
https://trello.com/c/mANzbvei/418-pagination-control

I split the original branch branch `search-results-pager` into 2, to create 2 separate PRs - this one for the style related changes and the other one (https://github.com/hypothesis/h/pull/3749) for pagination states.

### DIFF
--- a/h/static/styles/partials-v2/_pager.scss
+++ b/h/static/styles/partials-v2/_pager.scss
@@ -1,0 +1,45 @@
+.pager {
+  display: flex;
+  flex-direction: row;
+  margin-top: 50px;
+  font-weight: bold;
+  justify-content: center;
+}
+
+.pager__item {
+  border-radius: 2px;
+  color: $grey-6;
+  padding: 7px 10px 7px 10px;
+}
+
+.pager__item:hover {
+  background-color: $grey-3;
+}
+
+.pager__item--begin {
+  background-color: $grey-3;
+  margin-right: 10px;
+}
+
+.pager__item--end {
+  background-color: $grey-3;
+  margin-left: 10px;
+}
+
+.pager__item--end:not(.is-disabled):hover,
+.pager__item--begin:not(.is-disabled):hover {
+  background-color: $brand;
+  color: $white;
+}
+
+.pager__item--link {
+  color: inherit;
+}
+
+.pager__item.is-highlighted {
+  background-color: $grey-3;
+}
+
+.pager__item.is-disabled {
+  color: $grey-4;
+}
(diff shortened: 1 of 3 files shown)


h/static/styles/partials-v2/_pager.scss
```diff
@@ -0,0 +1,36 @@
...
+}
+
+.pager__item {
+  border-radius: 2px;
+  color: $grey-6;
+  padding: 7px 10px 7px 10px;
+}
+
+.item-highlight {
+  background-color: #ececec;
```
robertknight: This should use color variables from the palette (this is `$grey-2`)

h/templates/includes/paginator.html.jinja2
```diff
@@ -1,33 +1,47 @@
 <nav>
-  <ul class="pagination">
+  <ul class="pager">
     {% if page.prev is none %}
-      <li class="disabled">
-        <span aria-hidden="true">&laquo;</span>
+      <li class="item-disabled pager__item item-begin">
+        <span aria-hidden="true">&#x3c;</span>
```
robertknight: As far as I am aware, all of our source files use UTF-8 encoding, so I think it should be possible to just use the `<` character literally, rather than a hex escape.

h/static/styles/partials-v2/_pager.scss
```diff
@@ -0,0 +1,36 @@
...
+  justify-content: center;
+}
+
+.pager__item {
+  border-radius: 2px;
+  color: $grey-6;
+  padding: 7px 10px 7px 10px;
+}
+
+.item-highlight {
```
robertknight: The `item-*` classes should be renamed to fit out conventions:

`component__element` for elements
`is-$state` for state modifiers, in which case they should always be written as a combination selector along with the element that they style, eg `.pager__item.is-highlighted { ... }` rather than `is-highlighted { ... }`. This avoids the risk that they apply to the wrong element.

h/static/styles/partials-v2/_pager.scss
```diff
@@ -0,0 +1,36 @@
...
+  background-color: #ececec;
+}
+
+.item-begin {
```
robertknight: You could use `:first-of-type` and `:last-of-type` or `:first-child` / `:last-child` pseudo-element selectors here rather than introducing a new class.

robertknight:
This is looking pretty close, there are a few differences I noticed from [the mocks](https://trello-attachments.s3.amazonaws.com/5797dfe1b768b1347fbea9b0/800x750/17f75b93fe170ecfc63eb5d2f0d59d06/Pagination_-_measurements_and_states.png):

- [ ] Hover states (red background) for pages and items
- [ ] The background color of the selected page and start/end items is darker (`$grey-3` rather than `$grey-2`) in the mocks

robertknight:
LGTM

PR #3753, opened 2016-08-23
Remove a couple of unused customization points in the base template

robertknight:
No templates in the service are currently defining `base_tag`,
`body_tag` or `templates` blocks. The first and last of these were used
only by the Hypothesis client, which no longer uses `base.html`.
`body_tag` was used by a previous attempt at a new homepage.


nickstenning:
:+1:

PR #3754, opened 2016-08-23
Remove separate salt when setting a new password

nickstenning:
If a user with an old-style (double-salted) password walked through the
password reset process, they'd end up with a new-style password without
the `salt` column being reset and they'd be unable to login.

This fixes that issue.


chdorner:
LGTM.

PR #3757, opened 2016-08-23
Strip POST payload data from Sentry reports

nickstenning:
As useful as it can be, sending POST payloads to Sentry will inevitably result in us sending unhashed user passwords to a third-party service. Not a great idea.

The SanitizePasswordsProcessor is already on by default. This commit adds the RemovePostDataProcessor, which strips the POST payload data from error reports, as SanitizePasswordsProcessor is not smart enough to handle multipart forms.

### DIFF
--- a/h/sentry.py
+++ b/h/sentry.py
@@ -14,6 +14,11 @@ from raven.utils.wsgi import get_environ
 
 from h import __version__
 
+PROCESSORS = (
+    'raven.processors.SanitizePasswordsProcessor',
+    'raven.processors.RemovePostDataProcessor',
+)
+
 
 def http_context_data(request):
     return {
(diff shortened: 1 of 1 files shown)


h/sentry.py
```diff
@@ -14,6 +14,11 @@
 
 from h import __version__
 
+PROCESSORS = (
+    'raven.processors.SanitizePasswordsProcessor',
+    'raven.processors.RemovePostDataProcessor',
```
robertknight: I'd suggest that it would probably be a good idea to add a comment to indicate why we are removing POST data here, given that the issue with multipart forms is non-obvious.

nickstenning: `git log -S RemovePostDataProcessor` will give you that information straight away ;)

robertknight: Meh. I suppose ;)

robertknight:
I wonder if stripping all POST data might end up being too blunt, but hey, we can write a custom processor if we find its a problem in future.

PR #3758, opened 2016-08-23
Add invalid session HTML error

### LINKED ISSUE #3751: Changing password twice with same form results in raw JSON output.
### Steps to reproduce
1. Go to change password form in account settings
2. Enter new password
3. Submit
4. Reuse same form with yet another password change

### Expected behaviour
After I save the first password change, there is a banner that says: "Success. We've saved your changes."   And I'm still on the form.  It seems clear that I should be able to continue to make changes.  However, apparently the session (token) is no longer valid.  I should not remain on what looks like an active form page if the form is not actually active.

### Actual behaviour
I get a nasty looking stream of raw JSON

### Additional details
Screenshot:

screenshot 2016-08-22 16 11 17

chdorner:
We used to render the JSON error response, even for non-XHR requests, this adds a very basic error page with a link to login (which then redirects to the most likely place the user came from).

For testing (and checking that the "design" looks good enough):

1. Login
2. Go to `/account/settings`
3. Open a second tab
4. Log out in the second tab
5. Log back in in the second tab
6. Submit the password reset form in the first tab
7. tada

Fixes #3751

### DIFF
--- a/h/accounts/views.py
+++ b/h/accounts/views.py
@@ -38,7 +38,7 @@ def ajax_payload(request, data):
 
 
 @json_view(context=BadCSRFToken)
-def bad_csrf_token(context, request):
+def bad_csrf_token_json(context, request):
     request.response.status_code = 403
     reason = _('Session is invalid. Please try again.')
     return {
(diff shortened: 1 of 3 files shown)


seanh:
I think we're going to want a better looking page at some point, but this is better than JSON

PR #3760, opened 2016-08-23
Scroll expanded search results into view

robertknight:
When expanding search results, scroll the expanded annotation cards into
view. This fixes an issue where after expanding a bucket at the bottom
of the search result list, the user had to manually scroll the page in
order to actually read the newly exposed annotation cards.

 * Add `scroll-into-view` dependency and use it to scroll expanded
   search result buckets

 * Expose controller instances on upgraded elements for use in tests and
   debugging

### DIFF
--- a/h/static/scripts/base/upgrade-elements.js
+++ b/h/static/scripts/base/upgrade-elements.js
@@ -11,6 +11,9 @@ function isJSDisabled(document) {
  * If `root` lives in a document whose URL contains the query string parameter
  * `nojs=1` then `upgradeElements()` will return immediately.
  *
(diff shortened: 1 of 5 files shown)


seanh:
This is more of a "scroll into the middle of the viewport" than just "scroll into view". In some cases when the expanded annotations would have been in view anyway, this will scroll either up or down in order to scroll them into the middle of the window. Is that okay?

robertknight:
> Is that okay?

I think it is an improvement on the current behavior on master. It will be easy to replace `scroll-into-view` with some alternative logic in future that is perhaps a little more sophisticated.

PR #3762, opened 2016-08-23
Add tests for EmailChangeSchema

seanh:
Add tests for the new EmailChangeSchema class, and fix a couple of bugs
that the tests found.

### DIFF
--- a/tests/h/accounts/schemas_test.py
+++ b/tests/h/accounts/schemas_test.py
@@ -377,27 +377,138 @@ class TestLegacyEmailChangeSchema(object):
 @pytest.mark.usefixtures('user_model')
 class TestEmailChangeSchema(object):
 
-    def test_it_is_invalid_if_password_wrong(self,
-                                             db_session,
-                                             factories,
-                                             pyramid_csrf_request,
-                                             user_model):
-        # There isn't already an account with the email address that we're
-        # trying to change to.
-        user_model.get_by_email.return_value = None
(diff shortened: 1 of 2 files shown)


tests/h/accounts/schemas_test.py
```diff
@@ -377,27 +377,138 @@ def test_it_is_invalid_if_password_wrong(self,
...
         with pytest.raises(colander.Invalid) as exc:
             schema.deserialize({
                 'email': 'foo@bar.com',
-                'password': 'flibble'  # Not the correct password!
+                'password': 'WRONG'  # Not the correct password!
             })
 
-        assert 'password' in exc.value.asdict()
+        assert exc.value.asdict() == {
+            'password': 'Incorrect password. Please try again.'}
```
seanh: I realize that asserting the actual error messages is a little brittle, but it's not too bad (you change an error message, you have to update one test) and while writing these tests I did come across cases where `Invalid` would be raised but for some reason other than what the test was trying to test for (so the test was not actually testing what it thought it was) and where an error message would be raised on the right field but it would be the wrong error message. It used to be that if you tried to change your email address, and entered "a" in the password field, you could get a "password too short" error.

chdorner: I usually don't mind these test cases. It is much better to include the error message than to not be exactly sure which error happened.

chdorner:
LGTM.

PR #3767, opened 2016-08-24
Add a simple counter view for debugging

nickstenning:
Adds a simple view which makes changes to the database using SQLAlchemy.

This will be removed when it has served its purpose.


nickstenning:
(I've manually added the relevant table for this view.)

PR #3770, opened 2016-08-24
Setup ES2015 transpilation

robertknight:
Since the client has been moved to a separate repo and we don't have the concern about mixing three languages in the same codebase, I think we're now in a position to start using modern JS in the H service.

The first commit here adds the config, the other two commits change a couple of the controllers to be JS classes to make sure that everything is working.

~~I'm using TypeScript because I've used it quite a bit previously for this and I would at some point in future like to start taking advantage of its type checking facilities, although that will be much more useful in the larger client/browser-extension codebases. For the moment the code is still straight up ES2015. Alternative options are to use Babel (probably more popular in the ecosystem overall, but if you want type checking you need to bring additional tools) or stick with ES5.~~ _This PR is now using Babel with the ES2015 preset so we get can started, until we've had further team discussion about our tools. See this comment: https://github.com/hypothesis/h/pull/3770#issuecomment-244039325_


nickstenning:
Needs a rebase.

nickstenning:
And actually, while it does need a rebase, I'm not comfortable with us using the TypeScript compiler for this.

Any decision to adopt TypeScript features in our codebase is one I'd like to see discussed amongst the team, and if there's broad agreement, then we can have an ADR.

ES2015 I'm absolutely fine with, because it's not anything new and shiny, it's just JavaScript, and all we're doing is working around the fact that not all browsers support it yet. But if we're going that route we should taking the path of least resistance, which at this point, from my estimation, is Babel, not TS.

robertknight:
> Any decision to adopt TypeScript features in our codebase is one I'd like to see discussed
> amongst the team, and if there's broad agreement, then we can have an ADR.

OK, that's fair enough. I do think type checking is very valuable, so I'll write up something about the rationale and the different options as a starting point for discussion.

nickstenning:
By all means do so, but for now I'd suggest let's move forward with Babel as an ES2015 transpiler.

nickstenning:
It looks to me like you're unconditionally applying the `babelify` transform when running the tests, but only applying it to specific entrypoints when running `gulp build`. Given that ES2015 is a strict superset of ES5, why not add the transform to `package.json` and remove the custom config from `karma.config.js` and `gulpfile.js`?

robertknight:
> Given that ES2015 is a strict superset of ES5, why not add the transform to package.json
> and remove the custom config from karma.config.js and gulpfile.js?

OK. I'll note that not feeding code that doesn't need to be transformed through Babel makes the build slightly faster (~2.5s vs ~3.2s for the `build-js` task on my system). The difference is pretty small here though. In the client repo I think it was much more substantial last time I checked, because of things like the large `annotator.js` vendor script.

gulpfile.js
```diff
@@ -87,7 +87,7 @@ var bundleBaseConfig = {
 var bundles = [{
   // Public-facing website
   name: 'site',
-  entry: './h/static/scripts/site',
```
nickstenning: This is is now an unnecessary change. Please get into the habit of reviewing your own diffs to spot stuff like this!

nickstenning:
LGTM.

PR #3771, opened 2016-08-24
Add a base class for controllers

robertknight:
In preparation for implementing more complex controllers such as the one
that will support inline form editing, add a `Controller` base class
which provides shared functionality and helps to codify/reduce boilerplate in patterns that are common across controllers.

 - Setting the state of the controller and updating the DOM to reflect
   that

 - Replacing the controlled Element's content with a new version
   rendered on the server

 - Getting references to the various elements within the controlled
   element which the controller needs to interact with

 * Add Controller base class and rewrite DropdownMenuController and
   SearchBucketController to use it

 * Simplify tests using functionality from the Controller base class

I'm aware that there is a danger of inventing our own UI framework here, so in the meantime I'm keeping an eye open for existing solutions which are a good fit for our needs.

### DIFF
--- a/h/static/scripts/base/controller.js
+++ b/h/static/scripts/base/controller.js
@@ -1,0 +1,82 @@
+'use strict';
+
+/**
+ * Search the DOM tree starting at `el` and return a map of "data-ref" attribute
+ * values to elements.
+ *
+ * Multiple controllers may need to refer to the same element, so `data-ref`
+ * supports a space-separated list of reference names.
+ */
+function findRefs(el, map) {
+  if (!el.dataset) {
+    return map;
+  }
+  map = map || {};
+
+  var refs = (el.dataset.ref || '').split(' ');
+  refs.forEach(function (ref) {
+    map[ref] = el;
+  });
+
+  for (var i=0; i < el.children.length; i++) {
+    var node = el.children[i];
+    if (node.nodeType === Node.ELEMENT_NODE) {
+      findRefs(node, map);
+    }
+  }
+
+  return map;
+}
+
+/**
+ * Base class for controllers that upgrade elements with additional
+ * functionality.
+ *
+ * - Child elements with `data-ref="$name"` attributes are exposed on the
+ *   controller as `this.refs.$name`.
+ * - The element passed to the controller is exposed via the `element` property
+ * - The controllers attached to an element are accessible via the
+ *   `element.controllers` array
+ *
+ * The controller maintains internal state in `this.state`, which can only be
+ * updated by calling (`this.setState(changes)`). Whenever the internal state of
+ * the controller changes, `this.update()` is called to sync the DOM with this
+ * state.
+ *
+ * @param {Element} element - The DOM Element to upgrade
+ */
+function Controller(element) {
+  if (!element.controllers) {
+    element.controllers = [this];
+  } else {
+    element.controllers.push(this);
+  }
+
+  this.state = {};
+  this.element = element;
+  this.refs = findRefs(element);
+}
+
+/**
+ * Set the state of the controller.
+ *
+ * This will merge `changes` into the current state and call the `update()`
+ * method provided by the subclass to update the DOM to match the current state.
+ */
+Controller.prototype.setState = function (changes) {
+  var prevState = this.state;
+  this.state = Object.freeze(Object.assign({}, this.state, changes));
+  this.update(this.state, prevState);
+};
+
+/**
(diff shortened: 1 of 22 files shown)


robertknight:

h/static/scripts/base/controller.js
```diff
@@ -0,0 +1,61 @@
...
+  var prevState = this.state;
+  this.state = Object.freeze(Object.assign({}, this.state, changes));
+  this.update(this.state, prevState);
+};
+
+/**
+ * Replace the HTML content of the element with an updated version from the
+ * server.
+ */
+Controller.prototype.reload = function (html) {
```
robertknight: This is not used in this PR but is used as part of the inline form editing PR

h/static/scripts/base/controller.js
```diff
@@ -0,0 +1,69 @@
...
+function Controller(element) {
+  if (!element.controllers) {
+    element.controllers = [this];
+  } else {
+    element.controllers.push(this);
+  }
+
+  this.state = {};
+  this.element = element;
+  this.refs = dom.jsElements(element);
```
seanh: I wonder if we can come up with better name for this than refs? children? childElements?

h/static/scripts/base/controller.js
```diff
@@ -0,0 +1,69 @@
...
+}
+
+/**
+ * Set the state of the controller.
+ *
+ * This will merge `changes` into the current state and call the `update()`
+ * method provided by the subclass to update the DOM to match the current state.
+ */
+Controller.prototype.setState = function (changes) {
```
seanh: It might be better if this were called `updateState()` or maybe `mergeState()`? To me `set` implies that `state` is completely replaced with the new state that's passed in.

h/static/scripts/controllers/character-limit-controller.js
```diff
@@ -1,29 +1,28 @@
...
-  }
-
-  updateCounter();
-  input.addEventListener('input', updateCounter);
+function CharacterLimitController(element) {
+  Controller.call(this, element);
 
-  counter.classList.add('is-ready');
+  var self = this;
+  this.refs.characterLimitInput.addEventListener('input', function () {
```
seanh: This means that grepping for "character-limit-input" won't actually find all uses of it. It might be better if it were `this.refs['character-limit-input']` for that reason. Being able to grep for something and find all uses is really useful.

robertknight: You can grep for all uses of `refs` in a particular controller easily enough. An alternative way to address this would be to use a `data-ref` attribute rather than a class name, or just avoid hyphenating the class name.

h/static/scripts/base/controller.js
robertknight: I wondered about this, although that could also be confused with referring to the element's direct children. The terminology (`setState`, `refs`) is borrowed from React in a bid to hopefully be a) familiar (at least to some front-end devs) and b) concise. We use the same `setState` pattern in a few places in the client and browser extension.

h/static/scripts/controllers/character-limit-controller.js
seanh: Either of those alternatives sounds good to me.

I did wonder about the use of `js-*` CSS classes for JS stuff. Isn't CSS supposed to be for styling? Don't element attributes make more sense to use for JS?

But either way, I'd like any alternative that doesn't involve one name having two different spellings

robertknight:
I've just realised there is an issue with the reload in that element upgrades are not applied after the element's markup is replaced by the new version from the server.

nickstenning:
Any reason we're not using event delegation for these things?

h/static/scripts/controllers/character-limit-controller.js
robertknight: Done - I've replaced the use of `js-` classes with `data-ref` attributes to identify sub-elements within a controlled element.

robertknight:
> Any reason we're not using event delegation for these things?

It just hasn't seemed necessary yet. I'll have a look through in case there is anything that can be simplified with it.

h/static/scripts/controllers/character-limit-controller.js
seanh: Ah, so now it's _always_ referenced as `characterLimitInput` in code and never `character-limit-input`? Works for me

seanh:
I'm happy with this code now, will just test each of the components that got changed..

PR #3772, opened 2016-08-24
Cleanup form CSS component names

robertknight:
* Rename the `form` class to `form-container` since it represents the
   container that centers a form within a page, not the form itself and
   we'll probably want to use different form containers in future for
   other forms.

 * Update form-actions sub-elements to use BEM-structured names


robertknight:

seanh:
The merge of this got reverted because it broke the functional tests. Reopened this as https://github.com/hypothesis/h/pull/3783

PR #3774, opened 2016-08-24
Inline form editing

robertknight:
**_Depends on: #3771 and #3772_**

This implements an enhancement that enables inline editing for forms (see ) and enables it for the 'Edit Profile' and 'Edit group' forms. Support for the 'Account' and 'Notifications' forms will come separately as these require some additional work.

Forms opt into inline editing by passing the `inline_editing` flag when calling `form.render()`

See https://trello.com/c/KIU3NXPe/411-inline-editing-enhancement-for-forms

----
- [x] Reset form state after user makes a change that results in a validation error and then cancels editing the form

### DIFF
--- a/h/accounts/views.py
+++ b/h/accounts/views.py
@@ -566,10 +566,11 @@ class EditProfileController(object):
             self.request,
             self.form,
             on_success=self._update_user,
-            on_failure=self._template_data)
+            on_failure=self._template_data,
+            inline_editing=True)
 
     def _template_data(self):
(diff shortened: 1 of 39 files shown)


robertknight:

robertknight:
The initial approach taken here is starting to feel pretty messy and hard to test. I'm currently exploring a different approach to inline form editing whereby the server includes a JSON representation of the form in the initial HTML and we then render on the client. In response to an XHR request, the server returns the updated JSON representation of the form.

PR #3776, opened 2016-08-24
Add blacklist for group names

### LINKED ISSUE #3775: Group read page doesn't work for certain group names
### Steps to reproduce
1. Go to `/groups/new` and create a group with the name "leave" or "edit"
2. Click Create

### Expected behaviour
It should create the group and redirect to the group read page.

### Actual behaviour
For group name "leave" it creates the group and immediately removes the logged-in user from the members and then rendering a 404.
For group name "edit" it creates the group and shows the new edit form.

### Additional details
We should introduce a blacklist of group names that "slugify" to "leave" or "edit".

chdorner:
Since the URL structure is /{pubid}/{slug}, and we have two special URLs
(/{pubid}/leave and /{pubid}/edit) we have to stop allowing group names
that end up with slugs that are "edit" or "leave".

This implements a blacklist similar to what we do for usernames, the
error message says that the chosen group name is not allowed.

Fixes #3775

### DIFF
--- a/h/groups/schemas.py
+++ b/h/groups/schemas.py
@@ -3,6 +3,8 @@
 import colander
 import deform
 
+import slugify
+
 from h import i18n
 from h.accounts.schemas import CSRFSchema
 from h.groups.models import (
@@ -14,6 +16,15 @@ from h.groups.models import (
 
 _ = i18n.TranslationString
 
+GROUPSLUG_BLACKLIST = set(['edit', 'leave'])
+
+
+def unblacklisted_group_name_slug(node, value, blacklist=GROUPSLUG_BLACKLIST):
+    """Colander validator that ensures the "slugified" group name is not blacklisted."""
+    if slugify.slugify(value).lower() in blacklist:
+        raise colander.Invalid(node, _("Sorry, this group name is not allowed. "
+                                       "Please choose another one."))
+
 
 class GroupSchema(CSRFSchema):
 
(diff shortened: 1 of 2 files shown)


robertknight:
AIUI the slug is purely used to produce a friendly-looking URL. In that case, couldn't we just change the slug that is generated if the resulting URL would conflict with a defined route?

Adding a blacklisting mechanism seems elaborate and I'm also sure that someone will add a new `/groups/` route at some point in future and forget to update the blacklist.

robertknight:
Alternatively, we could introduce some way of namespacing the slugs and the routes so that they cannot conflict. Some ideas:

* Add an underscore or other char in front of slugs (eg. https://hypothes.is/groups/ID/_my-group)
* Add an underscore or other char in front of actions (eg. https://hypothes.is/groups/ID/_edit)
* Put group action routes under a `/actions` path (eg. https://hypothes.is/groups/ID/actions/edit)

nickstenning:
Briefly: I think this approach is fine, but there's no need to read the blacklist from a file at this stage. We went that route for the usernames simply because the size of the username blacklist makes putting it in a source file silly.

And maybe just add a comment next to the group routes advising people to update the blacklist.

chdorner:
Okay, I changed the blacklist to just be a hardcoded set instead of reading from a file. This should be ready to review/merge now.

robertknight:
I still think it would be better to have a standardized way to differentiate slugs from verbs that map to routes that doesn't require knowing the list of routes, but this will do for the moment.

Issue #3777, opened 2016-08-24
Form field error messages overlap scroll bars


seanh:
![screenshot from 2016-08-24 19-16-49](https://cloud.githubusercontent.com/assets/22498/17942464/62dd3658-6a2f-11e6-91e2-bc3d12fb3fe3.png)

seanh:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;D5kUsmPr&#x2F;671-move-activity-pages-error-messages-under-form-fields"><img src="https:&#x2F;&#x2F;github.trello.services&#x2F;images&#x2F;trello-icon.png" width="12" height="12"> Move activity pages error messages under form fields</a>

chdorner:
This does not happen anymore:
![2016-09-29-125654_518x190_scrot](https://cloud.githubusercontent.com/assets/112063/18951210/45e7582c-8644-11e6-9b68-299a89efaffe.png)

Closing.

seanh:
@robertknight @sean-roberts Do either of you know what's happened here? ^ It's good that the error message no longer overlaps the scrollbar, and I notice that the user's text no longer overlaps with the error message either, but now it seems that the error message squashes down the user's text into about half the width of the space available.

Anyway, iirc we decided to [move the error messages outside of the form fields](https://trello.com/c/D5kUsmPr/671-move-activity-pages-error-messages-under-form-fields) which should hopefully fix this.

sean-roberts:
yeah, I am not sure what changed to cause the new behavior but, as you sad, it will be gone with that other ticket.

PR #3778, opened 2016-08-24
Add character limit enhancement to form fields

seanh:
Add a character limit counter JavaScript enhancement to textareas, and
apply this enhancement to the "Description" text area on the user
profile edit form.

Any `<textarea>` whose Deform widget (defined in the textarea's deform
schema, in the `schema.py` file) has a `max_length` argument will
automatically get the character counter enhancement applied.

**Terminology**:

This enhancement is actually called the **character limit** enhancement
because its job is to enhance a textarea with UX affordances to do with
the fact that the textarea has a character limit. The UX affordances of
a character limit enhancement could include a character counter, or they
could be done some other way.

The **character counter** is a particular `<span>` element that this
particular character limit enhancement manages, that displays a count of
characters entered so far / character limit.

**Implementation**:

* In `textarea.jinja2`, when the `<textarea>` has a `max_length`, we add
  a new `<span>` to Deform `<textarea>`s. This `<span>` is actually a
  _sibling_ of the `<textarea>` not a child because child elements of
  `<textarea>`s are not displayed.

* By default this span contains a static informational message about the max
  length: **Up to 250 characters**.

  Which is the fallback, no-JavaScript behaviour. It lets users know that
  there is a character limit, but does not tell them how many characters
  they've entered so far, nor whether they're over the limit, nor does
  it prevent them from entering too many characters (this last one is
  desirable - we don't want to prevent them).

* Whether JavaScript is enabled or not, if the user submits a textarea
  with too many characters the server will respond with the error
  message **Longer than maximum length 250** so the user does get some
  feedback that they've gone over the limit (but this still does not
  tell them how far over the limit they've gone, or which portion of
  their text is over the limit).

  This server error message could perhaps be improved by telling the
  user how many characters they submitted.

  When JavaScript is enabled it's good that the character limit 250 is
  repeated in both the error message and character counter. The error
  message may trigger the user to realise what that 300/250 counter
  means, if they hadn't already.

* When the browser supports JavaScript we use the `"env-js-capable"` CSS
  class to immediately hide the `<span>` (`display: none`). This
  prevents a momentary flash of the "Up to 250 characters" message
  before the JavaScript kicks in.

  If for some reason the JavaScript fails to load then the
  `"env-js-timeout"` class gets added, and we use this to override
  `"env-js-capable"` and re-show the `<span>` containing the fallback
  static message (`display: initial`).

* The new JavaScript `CharacterLimitController` in
  `character-limit-controller.js` kicks in and replaces the text of the
  `<span>` with a **0/250** counter, then adds the `"is-ready"` class to
  the `<span>` which overrides the `"env-js-capable"` and re-shows the
  span (`display: initial`).

* The character limit from the `max_length` argument in the Python code
  is passed into the Jinja2 template by Deform, and the template passes it to
  `CharacterLimitController` as the value of a `data-maxlength`
  attribute on the `<textarea>` element.

* `CharacterLimitController` finds the character counter `<span>` using
  a `js-character-counter` CSS class that's put on the span (in the
  template) for this purpose.

* `CharacterLimitController` uses the `"input"` event which I believe is
  supported by all the browsers that we support, and that is triggered
  when the user types, deletes (with backspace or delete), selects text
  and deletes, pastes, undoes and redos in the textarea.

  When this event is fired it's just a simple case of counting the
  number of characters in the textarea and updating the contents of the
  `<span>`.

* We also add the `"form-input__character-counter--error"` CSS class
  when too many characters are entered, and this makes the character
  counter turn both red and bold. (Bold is used in additional to color
  to make the change colorblind-visible.) And we remove this class when
  the text becomes short enough again.

* Nothing prevents the user from entering too many characters into the
  form, or from submitting the form with too many characters in it.

* If the textarea is pre-rendered with some text already in it the counter is still correct (this happens for example when you enter too much text and submit the form, and the server responds by re-rendering the form with an error message and with your text still in the textarea).

### DIFF
--- a/h/static/scripts/controllers/character-limit-controller.js
+++ b/h/static/scripts/controllers/character-limit-controller.js
@@ -1,0 +1,29 @@
+'use strict';
+
+function CharacterLimitController(element) {
+  var counter = element.parentElement.querySelector(
+    '.js-character-limit-counter');
+  var input = element.querySelector('.js-character-limit-input');
+  var maxlength = parseInt(input.dataset.maxlength);
+
+  counter.textContent = '';
+
+  function updateCounter() {
+    var length = input.value.length;
+
+    counter.textContent = length + '/' + maxlength;
+
+    if (length > maxlength) {
+      counter.classList.add('is-too-long');
+    } else {
+      counter.classList.remove('is-too-long');
+    }
+  }
+
+  updateCounter();
+  input.addEventListener('input', updateCounter);
+
+  counter.classList.add('is-ready');
+}
+
+module.exports = CharacterLimitController;
--- a/h/static/scripts/tests/controllers/character-limit-controller-test.js
+++ b/h/static/scripts/tests/controllers/character-limit-controller-test.js
@@ -1,0 +1,155 @@
+'use strict';
+
+var CharacterLimitController = require('../../controllers/character-limit-controller');
+var util = require('./util');
+
+describe('CharacterLimitController', function () {
+
+  var containerEl;
+
+  afterEach(function () {
+    if (containerEl) {
+      containerEl.remove();
+      containerEl = null;
+    }
+  });
+
+  /**
+   * Make a <textarea> with the character limit controller enhancement
+   * applied and return the various parts of the component.
+   */
+  function component(value, maxlength) {
+    maxlength = maxlength || 250;
+
+    var template = '<div class="js-character-limit">';
+    template += '<textarea class="js-character-limit-input" data-maxlength="' + maxlength + '">';
+    if (value) {
+      template += value;
+    }
+    template += '</textarea><span class="js-character-limit-counter">Foo</span>';
+    template += '</div>';
+
+    containerEl = util.setupComponent(document, template, {
+      '.js-character-limit': CharacterLimitController,
+    });
+    var divEl = containerEl.querySelector('.js-character-limit');
+    var counterEl = containerEl.querySelector('.js-character-limit-counter');
+    var textarea = containerEl.querySelector('textarea');
+    var ctrl = divEl.controllers[0];
+
+    return {
+      containerEl: containerEl,
+      counterEl: counterEl,
+      textarea: textarea,
+      ctrl: ctrl,
+    };
+  }
+
+  it('adds the ready class', function () {
+    var counterEl = component().counterEl;
+
+    assert.equal(counterEl.classList.contains('is-ready'), true);
+  });
+
+  it('shows the counter initially even if textarea empty', function () {
+    var counterEl = component().counterEl;
+
+    assert.equal(counterEl.innerHTML, '0/250');
+  });
+
+  it('shows the counter if the element has pre-rendered text', function () {
+    var counterEl = component('pre-rendered').counterEl;
+
+    assert.equal(counterEl.innerHTML, '12/250');
+  });
+
+  it('continues to show the container after text deleted', function() {
+    var parts = component();
+    var counterEl = parts.counterEl;
+    var textarea = parts.textarea;
+
+    // Trigger the counter to be shown.
+    textarea.value = 'Some text';
+    textarea.dispatchEvent(new Event('input'));
+
+    // Delete all the text in the textarea,
+    textarea.value = '';
+    textarea.dispatchEvent(new Event('input'));
+
+    assert.equal(counterEl.innerHTML, '0/250');
+  });
+
+  it('reads the max length from the data-maxlength attribute', function () {
+    var counterEl = component('foo', 500).counterEl;
+
+    assert.equal(counterEl.innerHTML, '3/500');
+  });
+
+  it('updates the counter when text is added on "input" events', function () {
+    var parts = component();
+    var textarea = parts.textarea;
+    var counterEl = parts.counterEl;
+
+    textarea.value = 'testing';
+    textarea.dispatchEvent(new Event('input'));
+
+    assert.equal(counterEl.innerHTML, '7/250');
+  });
+
+  it('updates the counter when text is removed on "input" events', function () {
+    var parts = component('Testing testing');
+    var textarea = parts.textarea;
+    var counterEl = parts.counterEl;
+
+    // Make the text shorter.
+    textarea.value = 'Testing';
+    textarea.dispatchEvent(new Event('input'));
+
+    assert.equal(counterEl.innerHTML, '7/250');
+  });
+
+  it('does not add error class when no pre-rendered text', function() {
+    var counterEl = component(null, 5).counterEl;
+
+    assert.equal(counterEl.classList.contains('is-too-long'),
+                 false);
+  });
+
+  it('does not add error class when pre-rendered text short enough', function() {
+    var counterEl = component('foo', 5).counterEl;
+
+    assert.equal(counterEl.classList.contains('is-too-long'),
+                 false);
+  });
+
+  it('adds an error class to the counter when pre-rendered value too long', function() {
+    var counterEl = component('Too long', 5).counterEl;
+
+    assert.equal(counterEl.classList.contains('is-too-long'),
+                 true);
+  });
+
+  it('adds an error class to the counter when too much text entered', function() {
+    var parts = component(null, 5);
+    var counterEl = parts.counterEl;
+    var textarea = parts.textarea;
+
+    textarea.value = 'too long';
+    textarea.dispatchEvent(new Event('input'));
+
+    assert.equal(counterEl.classList.contains('is-too-long'),
+                 true);
+  });
+
+  it('removes error class from counter when text reduced', function() {
+    var parts = component('too long', 6);
+    var counterEl = parts.counterEl;
+    var textarea = parts.textarea;
+
+    textarea.value = 'short';
+    textarea.dispatchEvent(new Event('input'));
+
+    assert.equal(counterEl.classList.contains('is-too-long'),
+                 false);
+  });
+});
--- a/h/static/styles/partials-v2/_form-input.scss
+++ b/h/static/styles/partials-v2/_form-input.scss
@@ -50,6 +50,38 @@
     color: $grey-5;
   }
 
+  .form-input__character-counter {
+    @include font-normal;
+
+    // Show fallback "counter" if loading js times out.
+    .env-js-timeout & {
+      display: initial;
+    }
+
+    // Immediately hide counter if js capable - prevents momentary flash of
+    // unenhanced fallback counter before js kicks in.
+    .env-js-capable & {
+      display: none;
+    }
+
+    // The js controller adds this class when it's ready - show the now
+    // enhanced counter.
+    &.is-ready {
+      display: initial;
+    }
+
+    &.is-too-long {
+      color: $brand;
+      font-weight: bold;
+    }
+
+    position: absolute;
+    bottom: 10px;
+    right: $h-padding;
+
+    color: $grey-5;
+ }
+
   .form-input__hint {
     @include font-normal;
 
--- a/h/accounts/schemas.py
+++ b/h/accounts/schemas.py
@@ -353,7 +353,11 @@ class EditProfileSchema(CSRFSchema):
         colander.String(),
         missing=None,
         validator=colander.Length(max=250),
-        widget=deform.widget.TextAreaWidget(rows=2, cols=60),
+        widget=deform.widget.TextAreaWidget(
+            cols=60,
+            max_length=250,
+            rows=2,
+        ),
         title=_('Description'))
 
     location = colander.SchemaNode(
--- a/h/static/scripts/site.js
+++ b/h/static/scripts/site.js
@@ -8,6 +8,7 @@ if (settings.raven) {
 
 require('./polyfills');
 
+var CharacterLimitController = require('./controllers/character-limit-controller');
 var CreateGroupFormController = require('./controllers/create-group-form-controller');
 var DropdownMenuController = require('./controllers/dropdown-menu-controller');
 var FormSelectOnFocusController = require('./controllers/form-select-onfocus-controller');
@@ -15,6 +16,7 @@ var SearchBucketController = require('./controllers/search-bucket-controller');
 var upgradeElements = require('./base/upgrade-elements');
 
 var controllers = {
+  '.js-character-limit': CharacterLimitController,
   '.js-create-group-form': CreateGroupFormController,
   '.js-dropdown-menu': DropdownMenuController,
   '.js-select-onfocus': FormSelectOnFocusController,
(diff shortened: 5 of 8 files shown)


seanh:
With JavaScript disabled in the browser:

untitled screencast - edited 42

seanh:
With JavaScript:

untitled screencast - edited 43

seanh:
**Known issues**:

1. Server error message overlaps scrollbar. Separate issue: <https://github.com/hypothesis/h/issues/3777>
   
   screenshot from 2016-08-24 19-16-49

2. The server wraps your text. Separate issue: <https://github.com/hypothesis/h/issues/3779>

   screenshot from 2016-08-24 19-58-47

3. Character counter overlaps scrollbar

   screenshot from 2016-08-24 20-00-14

4. User profile description textarea is too small. Separate issue: https://github.com/hypothesis/h/issues/3780

h/static/scripts/tests/controllers/character-limit-controller-test.js
```diff
@@ -0,0 +1,144 @@
...
+    var textarea = containerEl.querySelector('.js-character-limit');
+    var ctrl = textarea.controllers[0];
+
+    return {
+      containerEl: containerEl,
+      counterEl: counterEl,
+      textarea: textarea,
+      ctrl: ctrl,
+    };
+  }
```
seanh: I'm using a helper function here instead of a `beforeEach()` because I have different tests that want to pass different args in to change _how the component gets created_. Creating test fixtures in `beforeEach()` just tightly couples tests to each other, I'd argue this is a better pattern.

seanh:
P.S. I've tested this on my iPhone 5S, seems fine on mobile except for the various known issues with the forms mentioned above.

seanh:
On Twitter, the character count does not change (e.g. does not turn red and bold) when you go over the limit, instead they dim out and disable the tweet button:

screenshot from 2016-08-24 20-25-25

We can't do this with our current forms as we don't have a save button that applies only to the textarea to dim out. Once the modal dialogs thing is done we could do this though, and then perhaps we won't need the counter to turn red and bold.

Also note that twitter shows a single number: the number of chars remaining. It counts down and goes negative when you go over. This seems fairly bad to me, not really that clear why Twitter is displaying "-7". I think "147/140" is clearer.

seanh:
Twitter also highlights what part of your text is over the limit, which is really good:

screenshot from 2016-08-24 20-29-56

I don't think you can do that with a normal textarea. Suggest that if we want that it should be a separate Trello card. It's not mentioned on the current card.

seanh:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;k7OGmSEm&#x2F;424-character-limit-enhancement-for-form-textareas-and-inputs"> Character-limit enhancement for form textareas and inputs</a>

seanh:
Note this works with `<textarea>`s only. The Trello card also requires `<input>`s. That will be a separate PR.

robertknight:
> This <span> is actually a sibling of the <textarea> not a child because child
> elements of <textarea>s are not displayed.

FWIW the spec terminology is that `<textarea>` is an _[escapable raw text elment](https://www.w3.org/TR/html5/syntax.html#escapable-raw-text-elements)_

h/static/scripts/controllers/character-limit-controller.js
```diff
@@ -0,0 +1,27 @@
+'use strict';
+
+function CharacterLimitController(element) {
+  var maxlength = parseInt(element.getAttribute('data-maxlength'));
```
robertknight: The standard way to read data properties is via the `dataset` attribute: `element.dataset.maxlength` (note that hyphens in the name are converted to `camelCase`)

h/static/scripts/tests/controllers/character-limit-controller-test.js
robertknight: This is fine. Its not a binary choice - just do what makes sense for a given test case and is consistent with the rest of the code.

h/static/scripts/tests/controllers/character-limit-controller-test.js
```diff
@@ -0,0 +1,144 @@
...
+
+  it('adds the ready class', function () {
+    var counterEl = component().counterEl;
+
+    assert.equal(counterEl.classList.contains('is-ready'), true);
```
robertknight: The component is not being removed from the DOM after the test finishes, which can be done via `Element#remove`

h/static/scripts/controllers/character-limit-controller.js
```diff
@@ -0,0 +1,27 @@
...
+  var counter = element.parentElement.querySelector('.js-character-counter');
+
+  counter.innerHTML = '';
```
robertknight: If you just want to set the text of a node, use `Element#textContent`, which avoids the risk of introducing a vector for loading arbitrary HTML into the page.

h/static/styles/partials-v2/_form-input.scss
```diff
@@ -50,6 +50,38 @@
...
+    }
+
+    position: absolute;
+    bottom: 10px;
+    right: $h-padding;
+
+    color: $grey-5;
+  }
+
+  .form-input__character-counter--error {
```
robertknight: `--foo` classes are for _static_ modifications to a class, eg. `btn--cancel` or `btn--danger`. For dynamic _states_ applied via JS, use `is-$state` classes with combination selectors, like `.is-ready` above. See https://github.com/hypothesis/frontend-toolkit/blob/master/docs/css-style-guide.md#naming

h/static/scripts/controllers/character-limit-controller.js
```diff
@@ -0,0 +1,27 @@
+'use strict';
+
+function CharacterLimitController(element) {
+  var maxlength = parseInt(element.getAttribute('data-maxlength'));
+  var counter = element.parentElement.querySelector('.js-character-counter');
```
robertknight: Controllers should avoid referencing elements outside of the sub-tree that they are upgrading. I would suggest fixing this by adding the controller on a class in `mapping_item.jinja2` which defines the container for the `<textarea>`.

h/static/scripts/controllers/character-limit-controller.js
```diff
@@ -0,0 +1,27 @@
...
+
+  counter.innerHTML = '';
+
+  function updateCounter() {
+    var length = element.value.length;
+
+    counter.innerHTML = length + '/' + maxlength;
```
robertknight: Again, use `textContent` here unless you really want to set HTML and are sure that it is sanitized.

h/static/scripts/tests/controllers/character-limit-controller-test.js
```diff
@@ -0,0 +1,144 @@
...
+var CharacterLimitController = require('../../controllers/character-limit-controller');
+var util = require('./util');
+
+describe('CharacterLimitController', function () {
+
+  /**
+   * Make a <textarea> with the character limit controller enhancement
+   * applied and return the various parts of the component.
+   */
+  function component(value, maxlength) {
```
robertknight: This is fine, though unfortunately verbose. Once my other open PRs land we'll be able to simplify this to something like:

```
function component(value, maxlength = 250)
  var template = `<textarea class="js-character-limit" data-maxlength="${maxlength}">${value}</textarea>
<span class="js-character-counter">Foo</span>`;

  return util.setupComponent(document, template, CharacterLimitController);
}

var ctrl = component("test", 42);
assert.equal(ctrl.refs.characterCounter.textContent, "4 / 42")
```

robertknight:
This is pretty good, just a few things I'd like to see addressed before merging this:

- [ ] Use `is-error` (or perhaps `is-too-long`) for the error state rather than a `--modifier` class
- [ ] Use `textContent` rather than `innerHTML`
- [ ] Avoid the controller referencing elements outside the upgraded element, eg. by moving the controller up to the containing `<div>`

robertknight:
> Bold is used in additional to color to make the change colorblind-visible.

Thoughtful 👍

h/static/scripts/controllers/character-limit-controller.js
seanh: Done

h/static/scripts/tests/controllers/character-limit-controller-test.js
seanh: Done

h/static/scripts/controllers/character-limit-controller.js
seanh: Done

h/static/styles/partials-v2/_form-input.scss
seanh: Done

h/static/scripts/controllers/character-limit-controller.js
seanh: Done

h/static/scripts/controllers/character-limit-controller.js
seanh: Done

h/static/scripts/tests/controllers/character-limit-controller-test.js
seanh: The problem is that if you've used `beforeEach()` to initialize text fixtures in your tests and you can get away with doing that because so far all the tests need the same fixtures initialized in the same way, then I come along later to edit a test or add a new one to that suite, and my new test needs the test fixtures initialized differently, then I can't add the test that I want to, not without refactoring first.

`beforeEach()` is fine for any code that you absolutely know that you want to execute before every test in the suite, including future tests not written yet. Something like resetting a database for example. But I'd say that creating test fixture objects that are actually going to be used in the tests never comes under this.

h/static/scripts/tests/controllers/character-limit-controller-test.js
seanh: It's a shame I had to make `containerEl` global so I could do this from an `afterEach()`, but Mocha doesn't provide any way for the `component()` helper to register a finalizer function, as far as I can see.

h/static/scripts/tests/controllers/character-limit-controller-test.js
seanh: Sounds good.

This is one area where CoffeeScript (with its destructuring assignment) really shined.

h/static/scripts/controllers/character-limit-controller.js
```diff
@@ -0,0 +1,29 @@
+'use strict';
+
+function CharacterLimitController(element) {
+  var counter = element.parentElement.querySelector(
+    '.js-character-limit-counter');
```
seanh: I renamed this so that all of the CSS classes that are only meant to be used internally by `CharacterLimitController` are `"js-character-limit-*"`.

h/static/scripts/controllers/character-limit-controller.js
```diff
@@ -0,0 +1,29 @@
...
+  var textarea = element.querySelector('.js-character-limit-input');
```
seanh: I'm using "input" here to refer to "whatever the input element is" be it a textarea or (later) an input or maybe in the future something else.

h/static/scripts/tests/controllers/character-limit-controller-test.js
robertknight: > This is one area where CoffeeScript (with its destructuring assignment) really shined.

ES2015 has destructuring as well, I believe inspired by CoffeeScript:

```js
const {foo, bar} = object;
const [first, second] = tuple;
```

seanh:
@robertknight all done

h/static/scripts/controllers/character-limit-controller.js
robertknight: Ah, my point was not so much about naming as boundaries. The idea is that a controller manages a particular DOM element and its children, so it shouldn't know or depend upon anything of the surrounding structure.

h/static/scripts/tests/controllers/character-limit-controller-test.js
```diff
@@ -0,0 +1,155 @@
...
+
+var CharacterLimitController = require('../../controllers/character-limit-controller');
+var util = require('./util');
+
+describe('CharacterLimitController', function () {
+
+  var containerEl;
+
+  afterEach(function () {
+    if (containerEl !== null) {
```
robertknight: Should just be `if (containerEl) { ... }`

h/static/scripts/tests/controllers/character-limit-controller-test.js
robertknight: > and my new test needs the test fixtures initialized differently, then I can't add the test that I want to, not without refactoring first.

In that case, just do the refactoring when you need to. I don't have any objections to factoring out this kind of thing into a helper function, I just don't think its always necessary.

h/static/scripts/tests/controllers/character-limit-controller-test.js
robertknight: mocha provides a fairly minimal set of hooks in its core, on the assumption that you'll roll your own additional facilities as needed, eg:

```js
var afterFns;

beforeEach(function () {
  afterFns = [];
});

afterEach(function () {
  afterFns.forEach(function (fn) { fn(); });
});

it('creates a thing which needs to be cleaned up', function () {
  var thing = createThing();

  afterFns.push(function () { thing.remove(); });
});
```

h/static/scripts/controllers/character-limit-controller.js
seanh: Sure, I moved the controller onto another DOM element as well, but it's probably a good CSS class naming pattern to use `"js-foo-controller"` and then `"js-foo-controller-*"` isn't it?

h/static/scripts/tests/controllers/character-limit-controller-test.js
seanh: Done

h/static/styles/partials-v2/_form-input.scss
```diff
@@ -50,6 +50,38 @@
...
+    }
+
+    position: absolute;
+    bottom: 10px;
+    right: $h-padding;
+
+    color: $grey-5;
+  }
+
+  .is-too-long {
```
robertknight: State classes should never be used on their own, always in combination with the class they modify, eg. `.form-input__character-counter.is-too-long`. Otherwise the `is-too-long` state here will apply to any component that happens to use a state with the same name (!)

robertknight: Or put this above underneath `is-ready` and use `&.is-too-long { ... }` to save typing

seanh: Done

h/static/scripts/controllers/character-limit-controller.js
robertknight: Actually I'd rather use a different mechanism or namespace to indicate controllers vs. classes that are there to provide handles for controllers to get references to elements.

I'm keeping an eye out for existing libraries or at least conventions that are a good fit for our circumstances, since I don't want to reinvent wheels if there are good round ones going.

robertknight:
LGTM

PR #3782, opened 2016-08-25
Support inline form editing for Notifications settings form

robertknight:
**_Depends on #3774_**

Implement inline editing for the Notifications settings form. This comes in two parts:

 1. Add support for checkbox and radio button fields in forms that use inline editing. The form is automatically submitted when a checkbox is toggled or a radio button is selected
 2. Change the notifications view to support XHR requests in the same way as other accounts views

### DIFF
--- a/h/accounts/views.py
+++ b/h/accounts/views.py
@@ -566,10 +566,11 @@ class EditProfileController(object):
             self.request,
             self.form,
             on_success=self._update_user,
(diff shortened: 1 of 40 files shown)


robertknight:
I'm closing this for the moment while I explore an alternative approach to implementing inline form editing. See https://github.com/hypothesis/h/pull/3774

Issue #3780, opened 2016-08-24
User profile description textarea is too small


seanh:
You're allowed to enter up to 250 chars, but the textarea is too small to hold 250 chars without growing a scrollbar.

![screenshot from 2016-08-24 20-02-58](https://cloud.githubusercontent.com/assets/22498/17944017/c61eff20-6a35-11e6-9230-691caec4f268.png)

Issue #3779, opened 2016-08-24
Server hard wraps textarea text


seanh:
![untitled screencast - edited 44](https://cloud.githubusercontent.com/assets/22498/17943842/08b03a12-6a35-11e6-99fb-8c9e769c1bfd.gif)

nickstenning:
Fixed by #3784.

PR #3784, opened 2016-08-25
Make user description field bigger

seanh:
This makes it just big enough to hold the max 250 chars on desktop. It'll still be too small on smaller screens where the width is less. Fixes https://github.com/hypothesis/h/issues/3780

screenshot from 2016-08-25 16-28-47

### DIFF
--- a/h/templates/deform/textarea.jinja2
+++ b/h/templates/deform/textarea.jinja2
@@ -1,5 +1,4 @@
-
-<textarea
+<textarea {%- if field.widget.rows -%}rows="{{ field.widget.rows }}"{% endif %}
 {% include "includes/common_attrs.jinja2" %}>{{ cstruct }}</textarea>
 {% if field.widget.max_length %}
   <span class="form-input__character-counter" data-ref="characterLimitCounter">
--- a/h/templates/deform/textarea.jinja2
+++ b/h/templates/deform/textarea.jinja2
@@ -1,4 +1,4 @@
-<textarea {%- if field.widget.rows -%}rows="{{ field.widget.rows }}"{% endif %}
+<textarea {% if field.widget.rows -%}rows="{{ field.widget.rows }}"{% endif %}
 {% include "includes/common_attrs.jinja2" %}>{{ cstruct }}</textarea>
 {% if field.widget.max_length %}
   <span class="form-input__character-counter" data-ref="characterLimitCounter">
--- a/h/accounts/schemas.py
+++ b/h/accounts/schemas.py
@@ -362,7 +362,6 @@ class EditProfileSchema(CSRFSchema):
         missing=None,
         validator=colander.Length(max=250),
         widget=deform.widget.TextAreaWidget(
-            cols=60,
             max_length=250,
             rows=4,
         ),
--- a/h/accounts/schemas.py
+++ b/h/accounts/schemas.py
@@ -364,7 +364,7 @@ class EditProfileSchema(CSRFSchema):
         widget=deform.widget.TextAreaWidget(
             cols=60,
             max_length=250,
-            rows=2,
+            rows=4,
         ),
         title=_('Description'))
 


seanh:
I think this also fixes https://github.com/hypothesis/h/issues/3779 by removing the cols arg

h/templates/deform/textarea.jinja2
```diff
@@ -1,3 +1,3 @@
-<textarea
+<textarea {%- if field.widget.rows -%}rows="{{ field.widget.rows }}"{% endif %}
```
robertknight: This generates `<textarearows` on my system and breaks layout

robertknight:
On my system this still results in an input that scrolls with a 250 char limit. Also, I'm not sure that we want an input field this large because that encourages a longer description than I think we want people to write.

My inclination would be to leave the description field at two rows by default, but leave in the support for specifying the number in the schema.

We can work on fixing the overlapping labels separately.

seanh:
Fixed the broken HTML / missing space. Not sure how that got in.

How does this scroll on your system, but not on mine? Screenshot? I could make it another row bigger to give it a safety buffer.

I think we should make the textarea big enough (on a typical desktop setup) to contain the maximum chars without scrolling. We can either do this by making the textarea bigger, or making the maximum smaller. Reasons:

1. I think showing a character counter that counts up to 250 kind of encourages you to enter up to 250 characters. If we then make the textarea too small to contain that many characters because we actually want to discourage entering that many - then I think we're sending mixed signals.

2. I think it's awkward and annoying to make the user (if they've used the maximum allowed four rows) have to move a scrollbar up and down by one cm in order to read or edit their text, especially when there's plenty of room on the page for the textarea to be just a little bigger and avoid this:

   untitled screencast - edited 45

Alternative suggestions:

If we want the real maximum to be 250 chars but the _encouraged_ maximum to be something less (e.g. 150) there may be other ways that we can achieve this. For example: the character counter could already turn red and bold when they go over 150. Or the character counter could actually say 0/150 instead of 0/250, and go red when it goes over 150, but not actually prevent the user from entering and submitting more than 150, and as long as it was less then 250 the server would actually accept it.

Or we could just reduce the limit (both in the counter and the actually enforced server-side limit) to 150. This would be my preference - it's simpler and clearer to simply have a maximum limit, rather than a somehow encouraged limit and then a higher actual limit.

nickstenning:
Needs a rebase.

seanh:
Rebased

nickstenning:
> How does this scroll on your system, but not on mine? Screenshot? I could make it another row bigger to give it a safety buffer.

Because there isn't a single size that is correct for 250 characters of a variable-width font:

> ''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''

vs

> MMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMM

4 rows seems to work for most (English) text, so I'm just going to merge this for now.

robertknight:
> Because there isn't a single size that is correct for 250 characters of a variable-width font:

@seanh - In addition to this, also remember that font metrics vary between platforms. We are not bundling our own fonts (which is good IMO, because that makes loading much faster and web fonts can have really problematic FOUC issues), but the flipside is that font metrics can vary quite significantly depending on the font that actually gets used.

PR #3785, opened 2016-08-26
Show tooltip when user hovers on the orcid help icon in the profile edit form.

sheetaluk:
Show info icons on mobile and JS enabled browsers.
Do not show tooltips on non-js or JS time outs.
Instead, display a hint text.

Test:
`http://localhost:5000/account/profile`
`http://localhost:5000/account/profile (with JS disabled)`
`http://localhost:5000/account/profile?__env__=js-timeout`
`http://localhost:5000/account/profile (mobile)`

https://trello.com/c/gGtgaRg9/425-form-field-hint-enhancement

### DIFF
--- a/h/static/images/icons/info_icon.svg
+++ b/h/static/images/icons/info_icon.svg
@@ -1,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<svg width="12px" height="12px" viewBox="0 0 12 12" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
+    <g id="Page-1" stroke="none" stroke-width="1" fill="none" fill-rule="evenodd">
+        <g id="Artboard" fill="#A6A6A6">
+            <path d="M6,12 C9.3137085,12 12,9.3137085 12,6 C12,2.6862915 9.3137085,0 6,0 C2.6862915,0 0,2.6862915 0,6 C0,9.3137085 2.6862915,12 6,12 Z M5,2 L7,2 L7,4 L5,4 L5,2 Z M5,5 L7,5 L7,10 L5,10 L5,5 Z" id="Combined-Shape"></path>
+        </g>
+    </g>
+</svg>
--- a/h/static/scripts/controllers/tooltip-controller.js
+++ b/h/static/scripts/controllers/tooltip-controller.js
@@ -1,0 +1,64 @@
+'use strict';
+
+/**
+ * A custom tooltip similar to the one used in Google Docs which appears
+ * instantly when activated on a target element.
+ *
+ * The tooltip is displayed and hidden by setting its target element.
+ *
+ *  var tooltip = new Tooltip(document.body);
+ *  tooltip.setState({target: aWidget}); // Show tooltip
+ *  tooltip.setState({target: null}); // Hide tooltip
+ *
+ * The tooltip's label is derived from the target element's 'aria-label'
+ * attribute.
+ *
+ * @param {Element} el - The container for the tooltip.
+ */
+function TooltipController(el) {
+  var self = this;
+
+  // With mouse input, show the tooltip on hover. On touch devices we rely on
+  // the browser to synthesize 'mouseover' events to make the tooltip appear
+  // when the host element is tapped and disappear when the host element loses
+  // focus.
+  // See http://www.codediesel.com/javascript/making-mouseover-event-work-on-an-ipad/
+  el.addEventListener('mouseover', function () {
+    self.setState({target: el});
+  });
+
+  el.addEventListener('mouseout', function () {
+    self.setState({target: null});
+  });
+
+  this.setState = function (state) {
+    this.state = Object.freeze(Object.assign({}, this.state, state));
+    this.render();
+  };
+
+  this.render = function () {
+    if (!this.state.target) {
+      this._el.style.visibility = 'hidden';
+      return;
+    }
+
+    var target = this.state.target;
+    var label = target.getAttribute('aria-label');
+    this._labelEl.textContent = label;
+
+    Object.assign(this._el.style, {
+      visibility: '',
+      bottom: 'calc(100% + 5px)',
+    });
+  };
+
+  this._el = el.ownerDocument.createElement('div');
+  this._el.innerHTML = '<span class="tooltip-label js-tooltip-label"></span>';
+  this._el.className = 'tooltip';
+  el.appendChild(this._el);
+  this._labelEl = this._el.querySelector('.js-tooltip-label');
+
+  this.setState({});
+}
+
+module.exports = TooltipController;
--- a/h/static/scripts/tests/controllers/tooltip-controller-test.js
+++ b/h/static/scripts/tests/controllers/tooltip-controller-test.js
@@ -1,0 +1,41 @@
+'use strict';
+
+var TooltipController = require('../../controllers/tooltip-controller');
+
+describe('TooltipController', function () {
+  var targetEl;
+  var template;
+  var testEl;
+  var tooltipEl;
+
+  before(function () {
+    template = '<div class="form-input__hint-icon js-tooltip"' +
+     'aria-label="Test"></div>';
+  });
+
+  beforeEach(function () {
+    testEl = document.createElement('div');
+    testEl.innerHTML = template;
+
+    targetEl = testEl.querySelector('div');
+    new TooltipController(targetEl);
+
+    tooltipEl = testEl.querySelector('.tooltip');
+  });
+
+
+  it('appears when the target is hovered', function () {
+    targetEl.dispatchEvent(new Event('mouseover'));
+    assert.equal(tooltipEl.style.visibility, '');
+  });
+
+  it('sets the label from the target\'s "aria-label" attribute', function () {
+    targetEl.dispatchEvent(new Event('mouseover'));
+    assert.equal(tooltipEl.textContent, 'Test');
+  });
+
+  it('disappears when the target is unhovered', function () {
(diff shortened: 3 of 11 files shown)


h/static/scripts/controllers/tooltip-controller.js
```diff
@@ -0,0 +1,68 @@
...
+ *  var tooltip = new Tooltip(document.body);
+ *  tooltip.setState({target: aWidget}); // Show tooltip
+ *  tooltip.setState({target: null}); // Hide tooltip
+ *
+ * The tooltip's label is derived from the target element's 'aria-label'
+ * attribute.
+ *
+ * @param {Element} el - The container for the tooltip.
+ */
+function TooltipController(el) {
```
robertknight: Once https://github.com/hypothesis/h/pull/3771 lands this should be refactored to use the `Controller` base class.

h/static/scripts/tests/controllers/tooltip-controller-test.js
```diff
@@ -0,0 +1,51 @@
+'use strict';
+
+var TooltipController = require('../../controllers/tooltip-controller');
+
+// helper to dispatch a native event to an element
+function sendEvent(element, eventType) {
```
robertknight: We upgraded to PhantomJS 2.x since this was originally written. PhantomJS 2.x _does_ support the _Event_ constructor, so you can just do:

```js
element.dispatchEvent(new Event('mouseover'))
```

For example

h/static/styles/partials-v2/_form-input.scss
```diff
@@ -80,16 +80,31 @@
...
+  .form-input__hint-icon {
+    display: none;
+    position: absolute;
+    top: 10px;
+    left: 55px;
+    cursor: help;
+
+    .env-js-capable & {
+      display: block;
+    }
```
robertknight: There is one extra case you need to deal with here which is where the browser is JS capable but loading scripts fails or takes too long. You can simulate this by appending `__env__=js-timeout` at the end of the URL. Search for uses of `env-js-timeout` to see how we handle this elsewhere in the code.

h/static/images/icons/info_icon.svg
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<svg width="12px" height="12px" viewBox="0 0 12 12" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
+    <!-- Generator: Sketch 39.1 (31720) - http://www.bohemiancoding.com/sketch -->
+    <title>Artboard</title>
+    <desc>Created with Sketch.</desc>
+    <defs></defs>
+    <g id="Page-1" stroke="none" stroke-width="1" fill="none" fill-rule="evenodd">
+        <g id="Artboard" fill="#A6A6A6">
```
robertknight: Protip: As an alternative to hard-coding the color, you can use `currentColor` and you can then set the icon color from CSS via the `color` property.

h/static/styles/partials-v2/_form-input.scss
```diff
@@ -80,16 +80,31 @@
...
     color: $grey-5;
- }
+  }
+
+  .form-input__hint-icon {
+    display: none;
+    position: absolute;
+    top: 10px;
+    left: 55px;
+    cursor: help;
```
robertknight: The `help` cursor matches the spec, but on my Ubuntu system this unfortunately shows a large '?' cursor which partially obscures the label. I would be inclined to change it to just an arrow. We can address this after the merge.

robertknight:
This is looking good. Just a few minor things to address:

- [ ] Use `new Event` rather than `sendEvent()` (see my comments)
- [ ] The icon's positioning looks a little too high to me and it needs a little bit more spacing to its left
- [ ] Handle the JS load timeout case (see my comments)

h/static/styles/partials-v2/_form-input.scss
```diff
@@ -80,16 +80,31 @@
     right: $h-padding;
 
     color: $grey-5;
- }
+  }
+
+  .form-input__hint-icon {
+    display: none;
+    position: absolute;
```
robertknight: Using `absolute` positioning here means that the icon's positioning does not adapt to the length of the label, so this positioning is only correct for the label 'ORCID' and if the user agent is rendering the text with the same font metrics that you are testing with.

One way to fix this would be to put the icon inside the `<label>`

robertknight:
Ah, found one major issue - the icon positioning needs to be changed so that it adapts to the length of the field label.

robertknight:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;gGtgaRg9&#x2F;425-form-field-hint-enhancement"> Form field hint enhancement</a>

h/static/images/icons/info_icon.svg
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<svg width="12px" height="12px" viewBox="0 0 12 12" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
+    <!-- Generator: Sketch 39.1 (31720) - http://www.bohemiancoding.com/sketch -->
+    <title>Artboard</title>
+    <desc>Created with Sketch.</desc>
+    <defs></defs>
```
seanh: Should we delete these 4 lines?

h/static/scripts/tests/controllers/tooltip-controller-test.js
```diff
@@ -0,0 +1,41 @@
+'use strict';
+
+var TooltipController = require('../../controllers/tooltip-controller');
+
+describe('TooltipFormController', function () {
```
robertknight: The test suite name needs updating to match the name of the class being tested

h/static/styles/partials-v2/_form-input.scss
```diff
@@ -173,4 +197,11 @@
       max-width: unset;
     }
   }
+
+  @include touch-input {
+    .form-input__label > .form-input__hint-icon {
+      min-height: 44px;
```
robertknight: I'd suggest we define a constant for this (eg. `$touch-target-size`) in `_base.scss` and use it here and in `_btn.scss`

h/static/styles/partials-v2/_tooltip.scss
```diff
@@ -0,0 +1,74 @@
...
+}
+
+.tooltip-label {
+  // Make the label a positioned element so that it appears _above_ the
+  // tooltip's arrow, which partially overlaps the content of the tooltip.
+  position: relative;
+}
+
+@include touch-input {
+  .form-input__label .tooltip {
```
robertknight: Components are meant to be reusable and isolated from their surrounding context. This means that we shouldn't style the tooltip in a specific context (eg. when used on a form input) within the file that defines the styles for the component. There are different ways to address the problem:

1. Try to restructure the styling / component in such a way that you don't need the contextual styles. Here for example, the '50px' is actually used because the styling is tied to the width of the "ORCID" label. We should try to find a way of setting up the tooltip so that the margins are independent of this
2. If you do need multiple variations of a component, eg. Danger / Cancel / Normal buttons, you can use `component--modifier` styles.
3. If you really do need some context-specific adjustment, you can put the element in a wrapper `<span>` or `<div>`. The search bar in the navbar for example is wrapped in a `nav-bar__search` element which handles positioning of the search bar within the navbar, so that the search input can be reused elsewhere.

h/static/styles/partials-v2/_tooltip.scss
```diff
@@ -0,0 +1,74 @@
...
+    position: inherit;
+    margin-top: -53px;
+  }
+
+  .form-input__label .tooltip:before {
+    margin-left: 50px;
```
robertknight: This breaks if I make the label text longer, eg. change ORCID to "This is a longer label"

h/static/styles/partials-v2/_tooltip.scss
```diff
@@ -0,0 +1,74 @@
...
+  right: calc(100% - 17px);
+}
+
+.tooltip-label {
+  // Make the label a positioned element so that it appears _above_ the
+  // tooltip's arrow, which partially overlaps the content of the tooltip.
+  position: relative;
+}
+
+@include touch-input {
```
robertknight: I thought the intention was to display the label on touch devices rather than rely on the hint icon? When viewing the page with http://localhost:5000/account/profile?__env__=touch I still see the tooltip icon rather than the hint text below the field.

h/static/images/icons/info_icon.svg
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<svg width="12px" height="12px" viewBox="0 0 12 12" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
+    <g id="Page-1" stroke="none" stroke-width="1" fill="none" fill-rule="evenodd">
```
robertknight: I see you've removed the `<title>` and `<desc>`. That's fine, though not strictly required. The `svg_icon()` helper will actually strip out the `title` tag itself. The idea is that we can add SVGs generated by tools like Sketch to the source tree without any manual edits to the file.

h/static/scripts/controllers/tooltip-controller.js
```diff
@@ -0,0 +1,68 @@
...
+  var self = this;
+
+  el.addEventListener('mouseover', function () {
```
robertknight: I assumed that this wouldn't work on mobile but it turns out that iOS Safari at least does synthesize `mouseover` and `mouseout` events when the element is clicked.

robertknight:
LGTM.

Issue #3742, opened 2016-08-22
Locally-annotated PDF files show "None" in place of the domain name


seanh:
![download 4](https://cloud.githubusercontent.com/assets/22498/17850985/b27f3870-6857-11e6-8b97-fddfca792a8a.png)

PR #3789, opened 2016-08-26
Fix display of local PDF domain name on search page

seanh:
When the annotated document has no domain name because it was a
locally-annotated PDF file, just display nothing rather than displaying
the word "None".

Fixes https://github.com/hypothesis/h/issues/3742

This replaces "None" with just an empty space:

screenshot from 2016-08-26 12-25-57

### DIFF
--- a/tests/h/activity/bucketing_test.py
+++ b/tests/h/activity/bucketing_test.py
@@ -88,14 +88,16 @@ class TestDocumentBucket(object):
         bucket = bucketing.DocumentBucket(document)
         assert bucket.domain == 'www.example.com'
 
-    def test_init_sets_None_domain_when_no_uri_is_set(self, db_session, document):
+    def test_init_sets_domain_to_local_file_when_no_uri_is_set(self,
+                                                               db_session,
+                                                               document):
         docuri_pdf = factories.DocumentURI(uri='urn:x-pdf:fingerprint',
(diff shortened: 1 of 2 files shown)


seanh:
@conordelahunty ^

conordelahunty:
I think we solved this problem on the older version of the groups page by saying "Local file" rather than leaving it blank?

seanh:
@conordelahunty Ah yes, it's left blank on the stream but on the groups page it says local file. Changed this to say local file:

screenshot from 2016-08-26 15-10-02

tests/h/activity/bucketing_test.py
```diff
@@ -88,14 +88,14 @@ def test_init_sets_the_domain_from_the_extracted_uri(self, db_session, document)
         bucket = bucketing.DocumentBucket(document)
         assert bucket.domain == 'www.example.com'
 
-    def test_init_sets_None_domain_when_no_uri_is_set(self, db_session, document):
+    def test_init_sets_empty_string_domain_when_no_uri_is_set(self, db_session, document):
```
nickstenning: Test name needs updating.

seanh: Fixed and rebased

nickstenning:
LGTM.

PR #3790, opened 2016-08-26
Make the entire document bucket header clickable

### LINKED ISSUE #3723: Document bucket click area is invisible, too small
When I hover the mouse over a collapsed document bucket, the entire rectangular area that gets highlighted in grey should be clickable to in order to expand the document. In fact, only a smaller invisible rectangle within is actually clickable.

untitled screencast - edited 33

seanh:
Make the entire part of a document bucket that is visible when the
bucket is collapsed (domain name, document title, and number of
annotations) clickable in order to expand/collapse the bucket, instead
of having to click on the invisible flex box containing the document
title.

Fixes #3723.

### DIFF
--- a/h/static/styles/partials-v2/_search.scss
+++ b/h/static/styles/partials-v2/_search.scss
@@ -32,9 +32,6 @@
 
 // A group of search results
 .search-result-bucket {
-  display: flex;
-  flex-direction: row;
-  padding: 10px 10px;
   background-color: $grey-2;
   border-bottom: 1px solid $grey-3;
 
@@ -60,11 +57,14 @@
   display: flex;
   flex-direction: column;
   flex-grow: 1;
+  padding-left: 10px;
+  padding-right: 10px;
 }
 
 .search-result-bucket__header {
   display: flex;
   flex-direction: row;
+  padding: 10px 10px;
 
   cursor: pointer;
   user-select: none;
@@ -78,6 +78,7 @@
 
 .search-result-bucket__annotation-cards-container {
   display: flex;
+  padding-left: 130px;
   .env-js-capable & {
     display: none;
   }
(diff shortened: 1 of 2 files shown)


seanh:
The entire header row is clickable (it doesn't work in the GIF, but the mouse cursor turns into the "you are hovering over something clickable" cursor whenever it's over any part of the header row. The content row is _not_ clickable to collapse the bucket as this might get in the way of clicking on other clickable things (such as links) inside the expanded content.

untitled screencast - edited 48

nickstenning:
Needs a rebase.

seanh:
Rebased

h/static/styles/partials-v2/_search.scss
```diff
@@ -90,20 +107,6 @@
   flex-grow: 1;
 }
 
-.search-result-bucket__annotations-count-container {
```
nickstenning: _Please_ resist the urge to make changes that aren't related to your PR! I just spent two minutes trying to work out how this change was related to the intent of the PR, before realising that you've just reordered some CSS.

There's nothing wrong with reordering and tidying, but _please please please_ put it in a separate PR.

h/static/styles/partials-v2/_search.scss
```diff
@@ -49,35 +46,55 @@
...
 }
 
+.search-result-bucket__header {
+  padding: 10px 10px;
+
+  display: flex;
+  flex-direction: row;
+
+  cursor: pointer;
+  user-select: none;
```
nickstenning: Not sure we need to prevent the user from selecting the text in the header, do we? This seems like it could be super annoying.

nickstenning: I've just worked out that you didn't add this, so please ignore this comment. I reiterate my point from [this comment](https://github.com/hypothesis/h/pull/3790/files#r77322163).

nickstenning:
Ok, now I can actually tell what's going on, this looks good. _Please_ don't mix non-functional changes with functional changes, for my sanity if nothing else!

PR #3791, opened 2016-08-26
Hook up search pagination

### LINKED ISSUE #3766: Search results pager does not work
### Steps to reproduce
1. Go to https://hypothes.is/users/robertknight/search?q=&page=1
2. Use the links at the bottom of the search results page to navigate to the second, third, forth etc. pages

### Expected behaviour
The search results should change as I go from one page to the next

### Actual behaviour
I see the same results (the first page) regardless of the `page` parameter

chdorner:
Even though we already set the `page` parameter, it didn't get passed
down to the search query. To do that we also have to convert the page
number to a pair of limit/offset. This can all be done transparently in
the `query` module, while the page size definition should stay in the
view because it's being used for the pagination as well as for the
query.

Fixes #3766

### DIFF
--- a/h/activity/query.py
+++ b/h/activity/query.py
@@ -94,12 +94,18 @@ def check_url(request, query, unparse=parser.unparse):
         raise HTTPFound(location=redirect)
 
 
-def execute(request, query):
+def execute(request, query, page_size):
     search = Search(request)
     search.append_filter(TopLevelAnnotationsFilter())
     for agg in aggregations_for(query):
         search.append_aggregation(agg)
 
+    query = query.copy()
+    page = request.params.get('page', '')
+    if not page:
+        page = '1'
+    query['limit'] = page_size
+    query['offset'] = (int(page) - 1) * page_size
     search_result = search.run(query)
     result = ActivityResults(total=search_result.total,
                              aggregations=search_result.aggregations,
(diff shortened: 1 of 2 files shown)


chdorner:
I've realised that we don't have proper tests yet for the `h.activity.views.search` and `h.activity.query.execute` methods. Considering that I wouldn't get them done today, I think that's one for another Friday.

chdorner:
This is probably not going to be the final pagination solution, because you will get a different amount of document buckets back on each page (because we currently paginate on annotations), but this will at least allow us to click around a bit more than before.

h/activity/query.py
```diff
@@ -94,12 +94,18 @@ def check_url(request, query, unparse=parser.unparse):
...
+def execute(request, query, page_size):
     search = Search(request)
     search.append_filter(TopLevelAnnotationsFilter())
     for agg in aggregations_for(query):
         search.append_aggregation(agg)
 
+    query = query.copy()
+    page = request.params.get('page', '')
+    if not page:
+        page = '1'
```
seanh: I think this can just be a one-liner `page = request.params.get('page', '1')`?

seanh:
I'll take this one on and finish it off, going to look into adding tests..

seanh:
Gonna merge this now, will send a separate PR with a couple of tests and the change I suggested above

PR #3805, opened 2016-08-30
Generate tooltip DOM with preact

robertknight:
**Depends on #3771**

This tidies up the tooltip implementation and replaces manual DOM node creation with [preact](https://preactjs.com/guide/getting-started). This makes the tooltip implementation a little simpler and cleaner, but more importantly it serves as an example of how to build components that generate their own UI on the client as opposed to simple controllers which just listen to events and set styles on markup already rendered by the server. This will be useful when we come to implement the search bar or other more complex client-side additions in future.

### DIFF
--- a/h/static/scripts/controllers/tooltip-controller.js
+++ b/h/static/scripts/controllers/tooltip-controller.js
@@ -1,18 +1,28 @@
 'use strict';
 
 var inherits = require('inherits');
+var preact = require('preact');
 
 var Controller = require('../base/controller');
+
+var h = preact.h;
+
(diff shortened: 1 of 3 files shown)


seanh:
Needs a rebase now

nickstenning:
I have to be honest, I'm a little dubious about this. It seems like a an 18KiB hammer to crack a (tiny) nut.

The old code was explicit and probably understandable to pretty much anyone who is familiar with the DOM. The new code assumes an understanding of react...

nickstenning:
I'm going to close this for now.

PR #3806, opened 2016-08-31
Pin zope.interface to 4.2.0

robertknight:
Pin zope.interface until
https://github.com/zopefoundation/zope.interface/issues/53 is fixed


jamadden:
zope.interface 4.3.1 has been released to PyPI to fix zopefoundation/zope.interface#53.

robertknight:
Thanks for the quick response @jamadden

PR #3807, opened 2016-09-01
Fix Safari and IE element upgrades

robertknight:
In Safari 9.x, `findRefs()` ran into an issue where
`Element#dataset#name` could sporadically return `undefined` even though
the element has a `data-ref` attribute. See
https://bugs.webkit.org/show_bug.cgi?id=161454

IE 10 lacks support for `Element#dataset` and also `ParentNode#children`
for <svg> elements, so use `querySelectorAll()` to find matching descendants.

The revised implementation does not look for `data-ref` on the root node
since that is already available via `this.element` in controllers

### DIFF
--- a/h/static/scripts/base/controller.js
+++ b/h/static/scripts/base/controller.js
@@ -7,22 +7,18 @@
  * Multiple controllers may need to refer to the same element, so `data-ref`
  * supports a space-separated list of reference names.
  */
-function findRefs(el, map) {
-  if (!el.dataset) {
-    return map;
-  }
-  map = map || {};
+function findRefs(el) {
+  var map = {};
 
-  var refs = (el.dataset.ref || '').split(' ');
-  refs.forEach(function (ref) {
-    map[ref] = el;
-  });
-
-  for (var i=0; i < el.children.length; i++) {
-    var node = el.children[i];
-    if (node.nodeType === Node.ELEMENT_NODE) {
-      findRefs(node, map);
-    }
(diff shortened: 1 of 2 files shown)


h/static/scripts/base/controller.js
```diff
@@ -7,21 +7,22 @@
...
 
-  for (var i=0; i < el.children.length; i++) {
-    var node = el.children[i];
+  // Use `Node#childNodes` rather than `ParentNode#children` because
+  // IE lacks support for `#children` on <svg> elements
+  for (var i=0; i < el.childNodes.length; i++) {
+    var node = el.childNodes[i];
     if (node.nodeType === Node.ELEMENT_NODE) {
-      findRefs(node, map);
+      Object.assign(map, findRefs(node));
```
robertknight: An alternative to a recursive search here would be to use `el.querySelectorAll('*[data-ref]')`. The current approach has the slight advantage of working on detached trees (ie. Elements that are not part of the Document) which is mildly useful in tests. I don't mind changing it either way.

nickstenning: I'm not sure I follow. Surely you can use `el.querySelectorAll` on a detached element:

```javascript
var el = document.createElement('div');  # => <div></div>
el.innerHTML = '<span id="a" data-ref="foo"><span id="b" data-ref="bar"></span></span><span id="c" data-ref="baz"></span>'
el.querySelectorAll('*[data-ref]'); # => [span#a, span#b, span#c]
```

nickstenning:
LGTM.

PR #3809, opened 2016-09-01
Add activity query execute tests and fix a couple of bugs

seanh:
I think `execute()` is doing too much and has too many dependencies, but in this PR I haven't refactored it, only added tests that should be helpful when we do come to refactor it, and that are helpful now for fixing bugs.


nickstenning:
LGTM.

PR #3812, opened 2016-09-01
Use capital letters and full stops

seanh:
In all "hint" tooltip texts on form fields.


robertknight:
LGTM. I don't think punctuation makes sense for _all tooltips in general_. Most native UI tooltips don't use them. For the slightly specialized case of lengthier form hints it seems reasonable though.

PR #3814, opened 2016-09-01
Inline form editing

robertknight:
**Depends on https://github.com/hypothesis/h/pull/3895**

----

This implements inline form editing and enables it for the Edit Profile and Edit Group forms.

**UX Notes**

* When input fields in forms with inline editing are focused, the form submission buttons are displayed below the focused field and the area outside that field is dimmed, until the user either saves changes or cancels editing.
* Keyboard focus handling changes depending on whether the current field has unsaved changes or not. If the field has no unsaved changes, the user can navigate through the fields in the form just by hitting the tab/shift+tab keys as they normally would. When a field becomes _dirty_ (by a user making changes), the current field becomes keyboard modal and focus stays in the field until the user saves or cancels changes.

**Implementation Overview**

The basic approach is to capture the initial rendered version of the form when the form is upgraded. When the user makes changes, the form is marked as dirty. When the user cancels changes, the form is replaced with the markup for the original version and upgrades to child elements are reapplied. When the user submits the form, the markup is replaced by the updated version from the server and upgrades to child elements are re-applied. If submission failed with a validation error, focus returns to the field that was being edited.

If the form submission fails with a non-validation (non-400) error, the current field remains in the editing state and an error message is displayed next to it.

**Implementation Notes**

Forms opt in to inline editing by setting the `use_inline_editing` attribute on the `deform.Form`. This results in the `js-form` class being added to the rendered form which triggers an upgrade using `FormController` on the client.

Two new features have been added to the current controller system to support the implementation:

* Support for communication between parent and child controllers (eg. forms and form inputs) via children broadcasting events upwards (`Controller#trigger`) and parents listening for them (`Controller#on`).
* Support for replacing the content of a controlled element with new markup from the server and re-applying upgrades (`Controller#reload`)

### DIFF
--- a/h/static/scripts/base/controller.js
+++ b/h/static/scripts/base/controller.js
@@ -1,31 +1,12 @@
 'use strict';
 
-/**
- * Search the DOM tree starting at `el` and return a map of "data-ref" attribute
- * values to elements.
- *
- * Multiple controllers may need to refer to the same element, so `data-ref`
- * supports a space-separated list of reference names.
- */
-function findRefs(el) {
-  var map = {};
+var { findRefs } = require('../util/dom');
 
-  var descendantsWithRef = el.querySelectorAll('[data-ref]');
-  for (var i=0; i < descendantsWithRef.length; i++) {
-    // Use `Element#getAttribute` rather than `Element#dataset` to support IE 10
-    // and avoid https://bugs.webkit.org/show_bug.cgi?id=161454
-    var refEl = descendantsWithRef[i];
-    var refs = (refEl.getAttribute('data-ref') || '').split(' ');
-    refs.forEach(function (ref) {
-      map[ref] = refEl;
-    });
-  }
-
-  return map;
-}
-
-/**
+/*
  * @typedef {Object} ControllerOptions
+ * @property {Function} [reload] - A function that replaces the content of
+ *           the current element with new markup (eg. returned by an XHR request
+ *           to the server) and returns the new root Element.
  */
 
 /**
@@ -85,6 +66,15 @@ class Controller {
    */
   forceUpdate() {
     this.update(this.state, this.state);
+  }
+
+  /**
+   * Listen for events dispatched to the root element passed to the controller.
+   *
+   * This a convenience wrapper around `this.element.addEventListener`.
+   */
+  on(event, listener, useCapture) {
+    this.element.addEventListener(event, listener, useCapture);
   }
 }
 
--- a/h/static/scripts/base/upgrade-elements.js
+++ b/h/static/scripts/base/upgrade-elements.js
@@ -1,6 +1,18 @@
 'use strict';
 
 /**
+ * Mark an element as having been upgraded.
+ */
+function markReady(element) {
+  var HIDE_CLASS = 'is-hidden-when-loading';
+  var hideOnLoad = Array.from(element.querySelectorAll('.' + HIDE_CLASS));
+  hideOnLoad.forEach(function (el) {
+    el.classList.remove(HIDE_CLASS);
+  });
+  element.classList.remove(HIDE_CLASS);
+}
+
+/**
  * Upgrade elements on the page with additional functionality
  *
  * Controllers attached to upgraded elements are accessible via the `controllers`
@@ -37,6 +49,7 @@ function upgradeElements(root, controllers) {
         new ControllerClass(el, {
           reload: reload.bind(null, el),
         });
+        markReady(el);
       } catch (err) {
         console.error('Failed to upgrade element %s with controller', el, ControllerClass, ':', err.toString());
 
--- a/h/static/scripts/base/upgrade-elements.js
+++ b/h/static/scripts/base/upgrade-elements.js
@@ -13,12 +13,30 @@
  *        order to upgrade it.
  */
 function upgradeElements(root, controllers) {
+  // A helper which replaces the content (including the root element) of
+  // an upgraded element with new markup and re-applies element upgrades to
+  // the new root element
+  function reload(element, html) {
+    if (typeof html !== 'string') {
+      throw new Error('Replacement markup must be a string');
+    }
+    var container = document.createElement('div');
+    container.innerHTML = html;
+    upgradeElements(container, controllers);
+
+    var newElement = container.children[0];
+    element.parentElement.replaceChild(newElement, element);
+    return newElement;
+  }
+
   Object.keys(controllers).forEach(function (selector) {
     var elements = Array.from(root.querySelectorAll(selector));
     elements.forEach(function (el) {
       var ControllerClass = controllers[selector];
       try {
-        new ControllerClass(el);
+        new ControllerClass(el, {
+          reload: reload.bind(null, el),
+        });
       } catch (err) {
         console.error('Failed to upgrade element %s with controller', el, ControllerClass, ':', err.toString());
 
--- a/h/static/scripts/controllers/form-controller.js
+++ b/h/static/scripts/controllers/form-controller.js
@@ -1,0 +1,204 @@
+'use strict';
+
+var Controller = require('../base/controller');
+var { findRefs, setElementState } = require('../util/dom');
+var submitForm = require('../util/submit-form');
+
+/**
+ * A controller which adds inline editing functionality to forms
+ */
+class FormController extends Controller {
+  constructor(element, options) {
+    super(element, options);
+
+    setElementState(this.refs.cancelBtn, {hidden: false});
+    this.refs.cancelBtn.addEventListener('click', event => {
+      event.preventDefault();
+      this.cancel();
+    });
+
+    // List of groups of controls that constitute each form field
+    this._fields = Array.from(element.querySelectorAll('.js-form-input'))
+      .map(el => {
+        var parts = findRefs(el);
+        return {container: el, input: parts.formInput};
+      });
+
+    this.on('focus', event => {
+      var field = this._fields.find(field => field.input === event.target);
+      if (!field) {
+        return;
+      }
+
+      // Enforce that the current field retains focus while it has unsaved
+      // changes
+      if (this.state.dirty &&
+          this.state.editingField &&
+          this.state.editingField !== field) {
+        this.state.editingField.input.focus();
+        return;
+      }
+
+      this.setState({editingField: field});
+    }, true /* capture - focus does not bubble */);
+
+    this.on('input', () => {
+      this.setState({dirty: true});
+    });
+
+    this.on('keydown', event => {
+      event.stopPropagation();
+      if (event.key === 'Escape') {
+        this.cancel();
+      }
+    });
+
+    // Ignore clicks outside of the active field when editing
+    this.refs.formBackdrop.addEventListener('mousedown', event => {
+      event.preventDefault();
+      event.stopPropagation();
+    });
+
+    // When the user tabs outside of the form, cancel editing
+    this.on('blur', () => {
+      // Add a timeout because `document.activeElement` is not updated until
+      // after the event is processed
+      setTimeout(() => {
+        // If the user has made changes to the active element, then keep focus
+        // on the active field, otherwise allow them to move to the previous /
+        // next fields by tabbing
+        if (this.state.dirty && !this._isEditingFieldFocused()) {
+          this.state.editingField.input.focus();
+        } else if (!this.element.contains(document.activeElement)) {
+          this.setState({editingField: null});
+        }
+      }, 0);
+    }, true /* capture - 'blur' does not bubble */);
+
+    // Setup AJAX handling for forms
+    this.on('submit', event => {
+      event.preventDefault();
+      this.submit();
+    });
+
+    this.setState({
+      // True if the user has made changes to the field they are currently
+      // editing
+      dirty: false,
+      // The group of elements (container, input) for the form field currently
+      // being edited
+      editingField: null,
+      // Markup for the original form. Used to revert the form to its original
+      // state when the user cancels editing
+      originalForm: this.element.outerHTML,
+      // Flag that indicates a save is currently in progress
+      saving: false,
+      // Error that occurred while submitting the form
+      submitError: '',
+    });
+  }
+
+  update(state, prevState) {
+    if (prevState.editingField &&
+        state.editingField !== prevState.editingField) {
+      setElementState(prevState.editingField.container, {editing: false});
+    }
+
+    if (state.editingField) {
+      // Display Save/Cancel buttons below the field that we are currently
+      // editing
+      state.editingField.container.parentElement.insertBefore(
+        this.refs.formActions,
+        state.editingField.container.nextSibling
+      );
+      setElementState(state.editingField.container, {editing: true});
+    }
+
+    var isEditing = !!state.editingField;
+    setElementState(this.element, {editing: isEditing});
+    setElementState(this.refs.formActions, {
+      hidden: !isEditing,
+      saving: state.saving,
+    });
+    setElementState(this.refs.formSubmitError, {
+      visible: state.submitError.length > 0,
+    });
+    this.refs.formSubmitErrorMessage.textContent = state.submitError;
+  }
+
+  /**
+   * Perform an AJAX submission of the form and replace it with the rendered
+   * result.
+   */
+  submit() {
+    var originalForm = this.state.originalForm;
+
+    var activeInputId;
+    if (this.state.editingField) {
+      activeInputId = this.state.editingField.input.id;
+    }
+
+    this.setState({saving: true});
+
+    return submitForm(this.element).then(response => {
+      this.options.reload(response.form);
+    }).catch(err => {
+      if (err.form) {
+        // The server processed the request but rejected the submission.
+        // Display the returned form which will contain any validation error
+        // messages.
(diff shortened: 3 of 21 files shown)


robertknight:
I know this PR is pretty large at the moment. Once we're happy that the fundamental functionality and UX behaviors then I can make this smaller by splitting out the generic additions to the element enhancement/upgrade code into a separate PR.

h/templates/deform/form.jinja2
```diff
@@ -5,39 +5,50 @@
...
+{% if field.use_inline_editing %}
+{% set form_controller_class = 'js-form' %}
+{% endif %}
 <form id="{{ field.formid }}"
       action="{{ field.action }}"
       method="{{ field.method }}"
       enctype="multipart/form-data"
       accept-charset="utf-8"
-      class="form {% if field.css_class %}{{ field.css_class }} {% endif %}">
+      class="form {{ field.css_class or '' }} {{ form_controller_class }}">
```
nickstenning: Personally I think this is kind of icky. It's a bit nicer to keep the conditionals where they're used and use Jinja's whitespace control to keep the output reasonable:

```jinja2
<form ...
     class="form
            {%- if field.css_class %} {{ field.css_class}}{% endif -%}
            {%- if field.use_inline_editing %} js-form{% endif -%}
     ">
```

h/static/scripts/tests/util/submit-form-test.js
```diff
@@ -0,0 +1,59 @@
+'use strict';
+
+var submitForm = require('../../util/submit-form');
+
+var lastRequest;
+function FakeXMLHttpRequest() {
```
nickstenning: You are aware that sinon [has it's own fake XHR module](http://sinonjs.org/docs/#server)? We don't have to write everything ourself.

h/static/scripts/controllers/form-input-controller.js
```diff
@@ -0,0 +1,32 @@
...
+
+var inherits = require('inherits');
+
+var Controller = require('../base/controller');
+var setElementState = require('../util/dom').setElementState;
+
+/**
+ * Controller for form input fields.
+ */
+function FormInputController(element) {
```
nickstenning: Perhaps I'm missing something, but why go to all the trouble to add `Controller#trigger` and `Controller#on`, when you could just use event delegation in `FormController` to set up the same handlers?

h/static/scripts/base/upgrade-elements.js
```diff
@@ -1,6 +1,18 @@
 'use strict';
 
 /**
+ * Mark an element as having been upgraded.
+ */
+function markReady(element) {
+  var HIDE_CLASS = 'u-js-hide-when-loading';
+  var hideOnLoad = Array.from(element.querySelectorAll('.' + HIDE_CLASS));
```
nickstenning: Doesn't this break [your own rules](https://github.com/hypothesis/frontend-toolkit/blob/master/docs/css-style-guide.md#classes-used-in-code) about which classes it's ok to reference from JavaScript?

In general, I think we should probably not be using CSS classes for purposes other than styling. For stuff like this, a data attribute is probably a clearer way of doing it.

h/static/scripts/controllers/form-input-controller.js
robertknight: In https://github.com/hypothesis/h/pull/3815 I make use of this encapsulation to map different native events to the same synthetic event depending on input type.

A caveat specific to focus events is that `focus` and `blur` do not bubble, although you can intercept them before they actually trigger in the capture phase.

h/static/scripts/util/submit-form.js
```diff
@@ -0,0 +1,37 @@
...
+  var formData = new FormData(formEl);
+  var req = new XMLHttpRequest();
+  req.open('POST', formEl.action, true /* async */);
+
+  // Flag this as an XHR request so that the server can respond accordingly
+  req.setRequestHeader('X-Requested-With', 'XMLHttpRequest');
+
+  req.send(formData);
+
+  return new Promise(function (resolve, reject) {
```
nickstenning: Please god let's be done with code like this and just use fetch.

h/templates/deform/form.jinja2
```diff
@@ -5,39 +5,50 @@
...
+      <span data-ref="formSubmitErrorMessage"></span>
+    </div>
     <div class="{{ form_message_class }}">
       {%- if field.footer %}{{ field.footer | safe }}{% endif -%}
     </div>
     {% if feature('activity_pages') %}
     <div class="u-stretch"></div>
     {% endif %}
-    <div class="{{ form_buttons_class }}">
+    <div class="{{ form_buttons_class }}" data-ref="formButtons">
```
nickstenning: What's this for?

h/templates/deform/form.jinja2
```diff
@@ -5,39 +5,50 @@
...
       {%- for button in field.buttons -%}
         <button id="{{ field.formid + button.name }}"
                 name="{{ button.name }}"
                 type="{{ button.type }}"
-                class="btn{% if button.css_class %} {{ button.css_class }}{% endif %}"
+                class="form-actions__btn btn{% if button.css_class %} {{ button.css_class }}{% endif %}"
                 value="{{ _(button.value) }}"
+                data-ref="formBtn"
```
nickstenning: What's this for?

h/static/scripts/base/upgrade-elements.js
robertknight: > Doesn't this break your own rules about which classes it's ok to reference from JavaScript?

This is a bit unusual in that we _do_ style this class, _do_ want to use it on its own (so not in the context of a specific component) and also want to reference it from JS. Probably the least ugly way to break the existing rules is to make this a state class (`is-hidden-when-loading`) and allow it to be used along with any element.

Another thought I had was to create a custom wrapper element:
```
<hide-when-loading>
... content ...
</hide-when-loading>
```

h/templates/deform/form.jinja2
robertknight: Ah sorry, this is left over from an earlier iteration. Same for `formBtn`. I'll remove both.

h/static/scripts/util/submit-form.js
robertknight: 👍

nickstenning:
I haven't gone through this with a fine toothcomb, but here are my observations.

The UX seems good. I'm happy you've considered how to navigate and use the form from a keyboard.

The implementation seems (to me) pretty complicated, all things considered. I'm not as close to the problem as you are, of course, but here are a few observations:

- I'm not sure reusing the form's existing submit and cancel buttons is really necessary -- I suspect it adds a fair bit of complexity to the FormController over just hiding the originals and inserting/removing your own at the appropriate point. Perhaps this will introduce a little bit of markup duplication -- I think that's ok.
- The division of labour between FormController and FormInputController seems like overkill to me: as mentioned in a line comment, simply providing event handlers for inputs in the FormController would seem to me about as complicated, except then:
    - you don't need to introduce the concept of parent/child relationships between controllers
    - you don't need to implement `childControllers`, `trigger` or `on`
- I think this is a clear case where event delegation will simplify the code. Instead of having to implement `reload` and call it at all the right times, simply move the controller onto a _container element_ rather than the form itself, and put all event handlers on the container so that you can simply swap out contents without having to rebind event handlers.

robertknight:
> I'm not sure reusing the form's existing submit and cancel buttons is really necessary --
> I suspect it adds a fair bit of complexity to the FormController

The amount of code that deals with the form actions is pretty small, and re-using the existing actions panel avoids potential bugs where form buttons have different attributes in the basic HTML/enhanced cases. It also means that labels etc. set on deform widget properties are respected.

> The division of labour between FormController and FormInputController seems like overkill to me

There were a couple of motivations for this separation:

1. I think it makes it easier to understand what constitutes a UI component if the boundary is the same across the template, styling and logic. In this case, form input fields (`form-input.scss`, `mapping-item.html.jinja2`, `FormInputController`).
2. In addition to handling events from input fields, we also need to update the state of the `<input>` _and its container_ as it is selected/deselected. Encapsulating the details of handling events and updating the state of items for individual form fields in a separate controller felt cleaner to me. I made use of this in the follow up PR that adds support for inline editing for checkboxes and radio buttons to hide some browser-specific event handling details from the form: https://github.com/hypothesis/h/pull/3815/commits/b1a28861d88f6386d0451c9d550d11976eb63e6f .

> I think this is a clear case where event delegation will simplify the code.
> Instead of having to implement reload and call it at all the right times, simply move the controller
> onto a container element rather than the form itself, and put all event handlers on the container
> so that you can simply swap out contents without having to rebind event handlers.

That's an interesting idea which I hadn't considered, but on the other hand since we need to re-apply upgrades to all elements in the replacement markup, it seems easier just to have a mechanism which can replace everything from a given root. It also means that after the reload, we start off with a clean state of fresh HTML and a fresh controller instance, which seems less likely to me to lead to eg. mistakes where a controller holds onto a reference to an old DOM node.

robertknight:
I think this is ready for another round of review now.

robertknight:
Looking at the codecov report this still needs a test for the change to how flash messages are handled. I'll do that.

robertknight:
The client-side code for handling XHR form submissions has been split out into https://github.com/hypothesis/h/pull/3895

h/static/scripts/base/controller.js
```diff
@@ -86,6 +106,95 @@ class Controller {
...
+  /**
+   * Broadcast an event to event handlers registered on the current element and
+   * parent elements using `on`.
+   *
+   * @param {string|Event} event - The name of an event, or the event to emit
+   * @param {Object} [data] - Optional data to attach to the event as `event.data`
+   */
+  trigger(event, data) {
+    if (typeof event === 'string') {
+      event = new Event(event);
```
nickstenning: It's not a good idea to change the type of a variable halfway through a function. Let's call the argument here `eventType`.

h/static/scripts/base/controller.js
```diff
@@ -25,6 +25,23 @@ function findRefs(el) {
...
 /**
+ * Dispatch an event to all event handlers of an element's controller registered
+ * using `Controller#on`
+ */
+function dispatchEvent(element, event) {
+  if (!element.controllers) {
+    return;
+  }
+  element.controllers.forEach(ctrl => {
+    ctrl._listeners.forEach(listener => {
```
nickstenning: I don't think an array is the appropriate data structure here. Let's have `_listeners` be a mapping from event type to array of listeners, then you don't need to use `forEach` here.

h/static/scripts/base/controller.js
```diff
@@ -86,6 +106,95 @@ class Controller {
...
+
+  /**
+   * Return controllers attached to child elements which are instances of
+   * `ControllerClass`
+   *
+   * @return {Array<ControllerClass>} Array of controllers attached to children
+   *         of `this.element` which are instances of `ControllerClass`
+   */
+  childControllers(ControllerClass) {
+    var children = this.element.querySelectorAll('*');
```
nickstenning: This doesn't seem right to me. This code means that if we were to have a single controller with a single child controller, the code would slow down linearly in the size of the DOM of the parent controller.

As suggested earlier, let's either remove the concept of child controllers for now, or explicitly keep track of children.

h/static/scripts/base/controller.js
```diff
@@ -86,6 +106,95 @@ class Controller {
...
+   * receive events for the replaced element. Instead the new controller for the
+   * replaced element is returned and can be used to transfer state from the
+   * existing controller.
+   *
+   * @param {string} html - The new markup for the element
+   * @return {Controller} - The new controller instance of the same type as this
+   *         controller on the element that replaces `this.element`
+   */
+  reload(html) {
+    if (!this.options.reload) {
```
nickstenning: Wait, so this function is just a wrapper around a function which is passed in from outside the controller? That's a really odd inversion of control. Either the controller base class should know how to apply upgrades itself, and use that functionality here, or `reload` itself should be part of the `upgrade-elements` module, and not part of a controller.

h/static/scripts/controllers/form-input-controller.js
```diff
@@ -0,0 +1,30 @@
...
+
+/**
+ * Controller for form input fields.
+ */
+class FormInputController extends Controller {
+  constructor(element, options) {
+    super(element, options);
+
+    this.refs.formInput.addEventListener('focus', () => {
+      this.trigger('form-input:focus');
```
nickstenning: I stand by my earlier assessment that the division of labour here between `FormInputController` and `FormController` is very awkward.

The only thing this controller does is mirror DOM events to our own homebrew event system, and mirror the FormController's `editing` state boolean onto this controller. I just don't see how that's making anything any easier, or any clearer.

robertknight:
> Currently, submitting an invalid form results in the form resetting to the "unfocused" state, where the buttons aren't displayed.

That is a bug. If you submit the form by pressing enter, the field remains focused. If you submit the form by clicking on the 'Save' button, this does not happen.

h/static/scripts/base/controller.js
nickstenning: Hmm, I've just reread this and noticed that you've got the ability to pass in an event or a string. But is any code here using the ability to pass in an event? If not that functionality shouldn't be in this PR, IMO.

h/static/scripts/base/controller.js
robertknight: This code is now gone as I've removed the synthetic event handling for the moment. TBQH this is unlikely to make much of a difference but I'll remember it if we do decide we need something similar in future.

h/static/scripts/base/controller.js
robertknight: This is now gone and `FormController` calls `this.options.reload` directly.

h/static/scripts/base/controller.js
robertknight: This is now gone.

h/static/scripts/base/controller.js
robertknight: This function is now gone.

robertknight:
Following our discussion on Slack, this has been revised as follows:

- `FormInputController` is gone. Instead `FormController` loops over the form field containers in its constructor and creates a list of `{container, input}` objects, where `container` and `input` are the form field containers and the `<input>` element respectively
- The synthetic event support is gone for the moment
- The pre-existing `findRefs` helper has been extracted into the `dom-util` module so it can be re-used inside `FormController` and tests have been added. You and Sean both asked about a better name for this before, I'd suggest s/refs/parts (as in "a named part of a component"). If we're agreed I'll change that as a separate PR.

robertknight:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;KIU3NXPe&#x2F;411-inline-editing-enhancement-for-edit-profile-form"> Inline editing enhancement for Edit Profile form</a>

h/static/scripts/base/upgrade-elements.js
```diff
@@ -13,12 +13,30 @@
...
  */
 function upgradeElements(root, controllers) {
+  // A helper which replaces the content (including the root element) of
+  // an upgraded element with new markup and re-applies element upgrades to
+  // the new root element
+  function reload(element, html) {
+    if (typeof html !== 'string') {
+      throw new Error('Replacement markup must be a string');
+    }
+    var container = document.createElement('div');
```
nickstenning: Rather than assuming that controllers are always `div`s, it would probably be nice to do `element.cloneNode(false)` here instead.

h/static/scripts/controllers/form-controller.js
```diff
@@ -0,0 +1,204 @@
...
+
+    // List of groups of controls that constitute each form field
+    this._fields = Array.from(element.querySelectorAll('.js-form-input'))
+      .map(el => {
+        var parts = findRefs(el);
+        return {container: el, input: parts.formInput};
+      });
+
+    this.on('focus', event => {
+      var field = this._fields.find(field => field.input === event.target);
```
nickstenning: You really don't like maps, do you? 😉

Given this is the only place you use `this._fields`, and you're doing an O(N) lookup in it, it would make more sense to have `this._fields` be a mapping between the input element and its container `js-form-input` field.

Alternatively, and even more simply, you could just not maintain this data structure at all, and instead just find the parent `.js-form-input` field when you receive a focus event:

```javascript
var input = event.target;
var field = getClosestParent(event.target, '.js-form-input');

...

function getClosestParent(element, selector) {
  while ((element = element.parentNode) && element.nodeType == Node.ELEMENT_NODE) {
    if (element.matches(selector)) {
      return element;
    }
  }
  return null;
}
```

robertknight: > You really don't like maps, do you? 😉

I think it is more that arrays are very convenient to work with in JS for a variety of reasons, so it becomes a bit of a habit :)

> Alternatively, and even more simply, you could just not maintain this data structure at all,
> and instead just find the parent .js-form-input field when you receive a focus event:

True. There is actually now a standardized [Element#closest](https://developer.mozilla.org/en-US/docs/Web/API/Element/closest) DOM function for this.

nickstenning: Heheh, the polyfill for IE/Edge is basically identical to my hastily written code above (except that `Element#closest` will match the current element as well:

https://github.com/jonathantneal/closest/blob/master/closest.js

PR #3815, opened 2016-09-01
Inline form editing for Notifications form

robertknight:
~~**_Depends on https://github.com/hypothesis/h/pull/3814_**~~

This implements inline editing for the Notifications form. The first commit implements auto-submission of forms which have inline editing enabled when a user toggles a checkbox or selects a radio item. The second commit enables XHR submission and inline editing for the Notifications settings form.

### DIFF
--- a/h/templates/deform/checkbox_choice.jinja2
+++ b/h/templates/deform/checkbox_choice.jinja2
@@ -7,11 +7,11 @@
            class="form-checkbox__label">
       <input type="checkbox"
              name="checkbox"
+             data-ref="formInput"
              value="{{ value }}"
              id="{{ field.oid }}-{{ loop.index0 }}"
-             class="form-checkbox__input {% if field.widget.css_class -%}
-               {{ field.widget.css_class }}
-             {% endif -%}"
+             class="form-checkbox__input
+             {%- if field.widget.css_class %} {{ field.widget.css_class }}{% endif -%}"
              {%- if field.error -%}
              aria-invalid="true"
              {% endif -%}
--- a/h/accounts/views.py
+++ b/h/accounts/views.py
@@ -588,7 +588,8 @@ class NotificationsController(object):
         self.request = request
         self.schema = schemas.NotificationsSchema().bind(request=self.request)
         self.form = request.create_form(self.schema,
-                                        buttons=(_('Save changes'),))
+                                        buttons=(_('Save'),),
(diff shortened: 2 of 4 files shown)


robertknight:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;bWW4niWC&#x2F;433-inline-editing-for-notifications-settings-form"> Inline editing for Notifications settings form</a>

nickstenning:
Rebased.

h/templates/deform/checkbox_choice.jinja2
```diff
@@ -7,9 +7,11 @@
            class="form-checkbox__label">
       <input type="checkbox"
              name="checkbox"
+             data-ref="formInput"
              value="{{ value }}"
              id="{{ field.oid }}-{{ loop.index0 }}"
-             class="form-checkbox__input {% if field.widget.css_class -%}
+             class="form-checkbox__input
+             {% if field.widget.css_class -%}
```
nickstenning: Hmm. I really would like us to be a bit more careful with whitespace controls in templates ([as mentioned previously](https://github.com/hypothesis/h/pull/3827#r77492424)).

It's really not too hard to arrange for a template to satisfy both a) be readable and b) emit minimal additional whitespace when rendered. Here, I'd suggest the following:

```jinja2
             class="form-checkbox__input
             {%- if field.widget.css_class %} {{ field.widget.css_class }}{% endif -%}"
```

nickstenning:
This seems to work mostly fine, but the flash of "modal translucent background" when the field is toggled is a bit disconcerting. Anything we can do about that?

It's also not great that we don't currently have any feedback to the use to indicate that the change was saved. Do we have concrete plans to address that?

robertknight:
> This seems to work mostly fine, but the flash of "modal translucent background" when the field is toggled is a bit disconcerting. Anything we can do about that?

This does function as a _saving_ indicator at the moment though, although it is subtle and a bit weird if saving is very quick.

> It's also not great that we don't currently have any feedback to the use to indicate that the change was saved. Do we have concrete plans to address that?

No. I think on both this point and indicating the saving status, I could use some design input.

h/templates/deform/checkbox_choice.jinja2
robertknight: Done

nickstenning:
LGTM.

PR #3817, opened 2016-09-02
Improve search bucket behavior in the JS load timeout case

robertknight:
When a search result bucket is upgraded after a JS load timeout, we want
the results to remain initially expanded, but they should still toggle
when clicked. Previously the toggle functionality was broken after an
eventual load.

This commit fixes the problem by removing the static `env-js-timeout` styling
and instead passing the timeout state to the search bucket controller,
which can use it to set the initial expanded state of the results.

 * Fix search bucket toggling failing after JS load timeout

 * Fix search bucket background being white instead of grey when
   expanded if mouse was not hovered over bucket.

   The `is-expanded` state was being set on the wrong element in
   `SearchBucketController#update`

### DIFF
--- a/h/static/scripts/base/upgrade-elements.js
+++ b/h/static/scripts/base/upgrade-elements.js
@@ -3,10 +3,6 @@
 /**
  * Upgrade elements on the page with additional functionality
  *
- * `upgradeElements()` provides a hook to test a page without JS enhancements.
- * If `root` lives in a document whose URL contains the query string parameter
- * `nojs=1` then `upgradeElements()` will return immediately.
- *
  * Controllers attached to upgraded elements are accessible via the `controllers`
  * property on the element.
  *
--- a/h/static/scripts/controllers/search-bucket-controller.js
+++ b/h/static/scripts/controllers/search-bucket-controller.js
@@ -5,22 +5,44 @@ var scrollIntoView = require('scroll-into-view');
 var Controller = require('../base/controller');
 var setElementState = require('../util/dom').setElementState;
 
-class SearchBucketController extends Controller {
-  constructor(element) {
-    super(element);
+/**
+ * @typedef Options
+ * @property {EnvironmentFlags} [envFlags] - Environment flags. Provided as a
+ *           test seam.
+ * @property {Function} [scrollTo] - A function that scrolls a given element
+ *           into view. Provided as a test seam.
+ */
 
-    this.scrollTo = scrollIntoView;
+/**
+ * Controller for buckets of results in the search result list
+ */
+class SearchBucketController extends Controller {
+  /**
+   * @param {Element} element
+   * @param {Options} options
+   */
+  constructor(element, options) {
+    super(element, options);
+
+    this.scrollTo = this.options.scrollTo || scrollIntoView;
 
     this.refs.header.addEventListener('click', () => {
       this.setState({expanded: !this.state.expanded});
     });
+
+    var envFlags = this.options.envFlags || window.envFlags;
+
+    this.setState({
+      expanded: !!envFlags.get('js-timeout'),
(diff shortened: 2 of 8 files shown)


nickstenning:
This seems like a slightly convoluted approach to me. Why doesn't the relevant controller just look at `window.envFlags && window.envFlags.get('js-timeout')` itself?

nickstenning:
Also, needs a rebase.

robertknight:
> Why doesn't the relevant controller just look at window.envFlags &&
> window.envFlags.get('js-timeout') itself? 

I think it is nicer to avoid a dependency on global state like this in the controller, it doesn't add that much plumbing and it makes testing straightforward.

nickstenning:
I'm not sure I agree. You've avoided a reference to an unavoidable global variable, at the cost of a coupling between components that isn't needed by any other controllers.

Maybe a suitable compromise would be to pass the entire environment flags instance to every controller?

robertknight:
> Maybe a suitable compromise would be to pass the entire environment flags instance to every controller?

OK. I'm making this change although this PR isn't quite ready yet.

h/static/scripts/site.js
```diff
@@ -29,12 +29,13 @@ var controllers = {
...
 };
 
-var doUpgrade = !window.envFlags || window.envFlags.get('js-capable');
-
-if (doUpgrade) {
-  upgradeElements(document.body, controllers);
-}
-
 if (window.envFlags) {
+  upgradeElements(document.body, controllers, {
```
nickstenning: So you're no longer checking the `js-capable` flag? Doesn't this mean that you can no longer simulate no-js-capable mode using the query string?

h/static/scripts/base/upgrade-elements.js
```diff
@@ -1,12 +1,15 @@
...
+ * @typedef {Object} UpgradeOptions
+ * @property {EnvironmentFlags} [envFlags] - EnvironmentFlags instance that can
+ *           be used by controllers to customize behavior depending on the
+ *           browser environment (eg. touch, whether a JS load timeout occurred)
+ */
+
+/**
  * Upgrade elements on the page with additional functionality
  *
- * `upgradeElements()` provides a hook to test a page without JS enhancements.
```
nickstenning: Perhaps move this to environment-flags.js rather than just removing it?

h/static/scripts/site.js
robertknight: Indeed. That's a mistake.

h/static/scripts/controllers/search-bucket-controller.js
```diff
@@ -6,21 +6,28 @@ var Controller = require('../base/controller');
 var setElementState = require('../util/dom').setElementState;
 
 class SearchBucketController extends Controller {
-  constructor(element) {
-    super(element);
+  constructor(element, options) {
+    super(element, options);
 
-    this.scrollTo = scrollIntoView;
+    this.scrollTo = this.options.scrollTo || scrollIntoView;
```
robertknight: Indeed. `scrollTo` is just a test seam. I'll note that.

robertknight:
> I would expect that adding "env=no-js-capable" or "env=js-timeout" would both result in pages
> where the document components started expanded, but this doesn't appear to be the case.

PEBKAC. For the case of `js-timeout`, I mixed up the flag names and was testing with `__env__=timeout` rather than `__env__=js-timeout`. Currently fixing that.

h/static/scripts/base/upgrade-elements.js
robertknight: Done, plus I've also documented the __env__ URL parameter.

nickstenning:
LGTM.

PR #3818, opened 2016-09-02
Pin pytest to version 3.0.1

robertknight:
Pin pytest version until https://github.com/pytest-dev/pytest/issues/1905 is resolved as it
caused memex tests to fail (eg. https://travis-ci.org/hypothesis/h/jobs/157031235)


nickstenning:
Thanks!

PR #3819, opened 2016-09-02
Put search bucket CSS in markup order

nickstenning:
This commit was extracted from #3790 and makes it easier to edit CSS and markup side-by-side by putting the classes in the order in which they are used.


seanh:
Thanks

Issue #3813, opened 2016-09-01
JavaScript tooltip size is wrong when text too short


seanh:
![screenshot from 2016-09-01 14-54-24](https://cloud.githubusercontent.com/assets/22498/18178324/ccb696b8-7075-11e6-8423-b1525b7e5cf3.png)

@robertknight said:

> The solution might be to move the tooltip's black background from the `.tooltip` element, to the actual label inside it, since the label is an inline element which will grow to fit its content.
> 
> Whereas block-level elements are sized horizontally to fit their containing block

PR #3820, opened 2016-09-02
Increase tooltip hover target

seanh:
Enable hovering anywhere over the entire "ORCID (i)" label (for example) rather than having to hover over the (i) icon itself, which is a very small hover target:

peek 2016-09-02 11-36

This is more accessible (larger hover target is easier to hit) and also
more consistent with normal HTML title-attribute tooltips which are
usually attached to words such as <label> or <a> elements and not (just)
to icons.

You no longer need to hover over the (i) icon to reveal the tooltip, but
the icon still serves the purpose of giving the user a hint that there
is a tooltip here.

I also tweaked the positioning of the tooltip's arrow to make it closer (to my eye) to the center of the "i", although I still don't think it's exactly right.

Screenshots of all three tooltips now:

screenshot from 2016-09-02 11-35-41
screenshot from 2016-09-02 11-34-58
screenshot from 2016-09-02 11-34-31

(The "This will be emailed to you" tooltip still has a whitespace bug but that's a [separate issue](https://github.com/hypothesis/h/issues/3813))

### DIFF
--- a/h/static/styles/partials-v2/_form-input.scss
+++ b/h/static/styles/partials-v2/_form-input.scss
@@ -200,12 +200,4 @@
       max-width: unset;
     }
   }
-
-  // Make hit target for field hint icon larger on mobile
-  @include touch-input {
-    .form-input__label > .form-input__hint-icon {
-      min-height: $touch-target-size;
-      min-width: $touch-target-size;
-    }
-  }
 }
--- a/h/static/styles/partials-v2/_tooltip.scss
+++ b/h/static/styles/partials-v2/_tooltip.scss
@@ -35,6 +35,7 @@
   padding-top: 4px;
   padding-bottom: 4px;
   position: absolute;
+  left: calc(100% - 12px);
   width: 220px;
   z-index: $zindex-tooltip;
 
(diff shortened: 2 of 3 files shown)


seanh:
On mobile you actually have to touch the (i) icon to reveal the tooltip, it's far too small for a touch target, so this should help there as well.

robertknight:
> On mobile you actually have to touch the (i) icon to reveal the tooltip, it's far too small for a touch target, so this should help there as well.

Actually that's not true, the tooltip has a touch target width of 44px on mobile to resolve this problem.

In principle, this is a nice improvement. Testing with Chrome dev tools' mobile preview however I see this:

tooltip-too-far-right

This is because of the hidden extra width added to the tooltip on mobile, which you could simply remove given that is is now fixed in an alternative way.

> I also tweaked the positioning of the tooltip's arrow to make it closer (to my eye)
> to the center of the "i", although I still don't think it's exactly right.

This means that the tooltip now embeds assumptions about the content it will be positioned on top of. A better solution IMO would be to specify a sub-element that the tooltip should target. eg. via `data-ref="tooltipTarget"`. The tooltip could then position itself on top of the middle of that element.

I don't think it is necessary to do that now however. If you can resolve the positioning issue above, I'd be happy to merge this.

seanh:
Yeah it would be better if the tooltip position could be computed dynamically.

For now - fixed the alignment on mobile by removing the touch target resizing.

robertknight:
LGTM

PR #3821, opened 2016-09-02
Remove some unused CSS

nickstenning:
This appears to have been introduced in e53316a, but it wasn't even used
there, so I guess it's ok to get rid?


robertknight:
Yes, indeed. Thanks!

PR #3823, opened 2016-09-02
Retry mailer tasks on SMTP failure

nickstenning:
We are seeing evidence that occasionally the Mandrill SMTP server disconnects our mailer, or perhaps that the network between our application and Mandrill occasionally drops:

  https://sentry.io/hypothesis/prod/issues/152186124/

This commit adds retry logic to the mailer task to ensure that in the event of any kind of network or SMTP error, the task is requeued for another delivery attempt later.

With the current configuration we will attempt to deliver mail four times (i.e. 3 retries) over a period of approximately 20 minutes, using an exponential backoff.

### DIFF
--- a/h/mailer.py
+++ b/h/mailer.py
@@ -5,6 +5,8 @@ A module for sending email.
 This module defines a Celery task for sending emails in a worker process.
 """
 
+import smtplib
+
 import pyramid_mailer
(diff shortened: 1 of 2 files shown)


robertknight:
LGTM

PR #3824, opened 2016-09-04
Convert remaining controllers to ES2015 classes

robertknight:
Convert remaining controllers to ES2015 classes that extend Controller,
except for `CreateGroupFormController`, since the whole group creation
form is going to be replaced shortly by a new version which uses the
standard form layout and behavior.

 * Convert ES3/5-style classes to ES2015 syntax

 * Convert event listener callbacks to use arrow functions,
   removing the need to alias `this` to `self`

 * Remove 'inherits' dependency, since all classes that have a base
   class have now been converted to ES2015 `class X extends Y` syntax

Module imports and var declarations are still using ES5 + CommonJS
syntax, for consistency with the other controller classes. These can be
converted using an automated tool later.


nickstenning:
LGTM.

PR #3825, opened 2016-09-04
Convert EnvironmentFlags to an ES2015 class

robertknight:
Use ES2015 class syntax + arrow functions in EnvironmentFlags


nickstenning:
LGTM.

PR #3826, opened 2016-09-04
Remove jQuery bundle from site JS pages

robertknight:
The new site JS does not use jQuery. Checked by searching for uses of `jQuery` and `$` in `h/static/scripts`

Saves ~40KB min+gzip.


nickstenning:
Wooop!

PR #3827, opened 2016-09-04
Fix behavior of legacy 'Create Group' form

robertknight:
- Add missing function parameter to the event listener for the
   'Tell me more about groups' link

 - Do not show the character counter on text input fields unless
   activity_pages feature flag is enabled

### DIFF
--- a/h/templates/deform/mapping_item.jinja2
+++ b/h/templates/deform/mapping_item.jinja2
@@ -4,18 +4,21 @@
 {% set field_label_class = 'form-input__label' %}
 {% set field_error_list_class = '' %}
 {% set field_error_item_class = 'form-input__error-item' %}
+{% set show_char_counter = field.widget.template in ('textinput', 'textarea') and
+                           field.widget.max_length %}
 {% else %}
 {% set field_class = 'form-group form-field' %}
 {% set field_error_state_class = 'form-field-error' %}
 {% set field_label_class = 'form-label' %}
 {% set field_error_list_class = 'form-error-list' %}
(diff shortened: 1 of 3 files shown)


h/templates/deform/mapping_item.jinja2
```diff
@@ -4,18 +4,21 @@
...
 {% set field_error_list_class = 'form-error-list' %}
 {% set field_error_item_class = 'form-error' %}
+{% set show_char_counter = False %}
 {% endif %}
 
 {%- if not field.widget.hidden -%}
 <div class="{{ field_class }}
             {% if field.error %}{{ field_error_state_class }}{% endif %}
-            {% if field.widget.template in ('textinput', 'textarea') and field.widget.max_length -%}js-character-limit{% endif %}"
+            {% if show_char_counter %} js-character-limit {%- endif %}"
```
nickstenning: I think we need to pay a little more care and attention to how we're using Jinja2 whitespace controls in situations like this.

Either we're using the newlines to separate attribute values, in which case:

```jinja2
class="{{ field_class }}
       {% if foo %}foo{% endif %}
       {% if bar %}bar{% endif %}"
```

or, preferably, we get rid of the newlines and all the indent whitespace and explicitly add whitespace in order to separate attribute values:

```jinja2
class="{{ field_class }}
       {%- if foo %} foo{% endif -%}
       {%- if bar %} bar{% endif -%}"
```

nickstenning:
I'm going to merge this so I can deploy, but if you could address the comments above in a followup commit that would be great.

PR #3833, opened 2016-09-06
Fix losing group / user filter when searching

chdorner:
When using the search form on the group / user search page we POSTed the
form to `/search` which lost the filter on the group / user. This now
checks on which page the user is and then uses that as the form action,
defaulting to `/search`.

### DIFF
--- a/h/views/panels.py
+++ b/h/views/panels.py
@@ -31,6 +31,11 @@ def navbar(context, request):
             "?q=user:{}".format(request.authenticated_user.username))
         username = request.authenticated_user.username
 
+    if request.matched_route.name in ['activity.group_search', 'activity.user_search']:
+        search_url = request.current_route_url()
+    else:
+        search_url = request.route_url('activity.search')
+
     return {
         'settings_menu_items': [
             {'title': _('Account details'), 'link': request.route_url('account')},
@@ -44,6 +49,7 @@ def navbar(context, request):
             {'title': _('Create new group'), 'link': request.route_url('group_create')},
         'username': username,
         'username_link': stream_url,
+        'search_link': search_url,
         'q': request.params.get('q', ''),
     }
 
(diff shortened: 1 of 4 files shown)


chdorner:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;iKjbl16J&#x2F;446-user-group-search-page-loses-user-group-filter"> User &#x2F; group search page loses user &#x2F; group filter</a>

h/views/panels.py
```diff
@@ -31,6 +31,15 @@ def navbar(context, request):
             "?q=user:{}".format(request.authenticated_user.username))
         username = request.authenticated_user.username
 
+    if request.matched_route.name == 'activity.group_search':
+        search_url = request.route_url('activity.group_search',
+                                       pubid=request.matchdict['pubid'])
```
nickstenning: This is ok, but could probably be simplified a bit with `request.current_route_url()`.

tests/h/views/panels_test.py
```diff
@@ -52,13 +53,45 @@ def test_navbar_includes_search_query(pyramid_request):
     assert result['q'] == 'tag:question'
 
 
+@pytest.mark.usefixtures('routes')
+def test_navbar_includes_search_url_when_on_user_search(pyramid_request):
+    pyramid_request.authenticated_user = None
+
+    type(pyramid_request.matched_route).name = PropertyMock(return_value='activity.user_search')
```
nickstenning: I _think_ you should be able to set `request.path` in a test rather than manually mocking `matched_route`. If you've registered the route as you have below, that should work fine.

chdorner: This doesn't seem to work. It looks like Pyramid sets the `matched_route` in the router which doesn't get called in our tests.

nickstenning: Huh. Okdoke.

PR #3834, opened 2016-09-06
Refactor navbar tests

chdorner:
**Depends on #3833**

Refactoring the tests to use a test class, a special request fixture with the `authenticated_user` property, and fixing a PEP warning about asserting `None`.


nickstenning:
LGTM. Just waiting for Travis.

PR #3835, opened 2016-09-07
Allow to specify activity page size via parameter

chdorner:
This will allow us to experiment with different page sizes and how the
response times react on staging and production.


nickstenning:
👍

PR #3839, opened 2016-09-08
Add NewRelic function traces in activity pages code

chdorner:
To get better insight on where most time is spent when serving requests.

I've tested this with a development app in NewRelic: https://rpm.newrelic.com/accounts/1385283/applications/22942021/transactions


nickstenning:
:+1:

PR #3840, opened 2016-09-08
Move Activation and User models into h.models package

nickstenning:
Start the process of moving models to a central location.


chdorner:
LGTM.

PR #3841, opened 2016-09-08
Use specified page size for pagination as well

chdorner:
It used to still use the default page size (20) to calculate the
pagination, which resulted in index out of bounds errors.


nickstenning:
LGTM.

PR #3843, opened 2016-09-08
Don't crash when there are no matching annotations

### LINKED ISSUE #3842: Activity search page crashes if page arg too high
On the `/search` page if the shown results have n pages, then just change the value of the `page` param in the URL to `n+1` and it'll crash.

seanh:
The bucketing code expects to receive a list of annotations, not a
sqlalchemy query object. Sending it a query object crashes when there
are 0 annotations that match the query, because the bucketing code's
test for this relies on an empty list being falsey (whereas a sqlalchemy
query with no results is truthy).

This could happen for example if the ?page=n argument in the query
string is larger than the number of pages of matching annotations.

Fixes #3842


chdorner:
LGTM.

PR #3845, opened 2016-09-08
Add an AuthClient model for third-party accounts

nickstenning:
This PR adds an AuthClient model/table, which holds the ids and secrets
associated with publisher accounts, which are -- more generally -- a kind of
OAuth client.


chdorner:
Rebased, waiting for Travis.

chdorner:
LGTM.

PR #3847, opened 2016-09-08
Make user unique constraints compound with authority

nickstenning:
In order to be able to create users for other authorities with any username, we need to only enforce username uniqueness constraints per-authority.

Update the user model so that we enforce uniqueness on the

    (email, authority)
    (uid, authority)
    (username, authority)

tuples rather than simply on email, uid, and username.


chdorner:
LGTM. Just tested it with my local DB, and then with a new production dump from today

PR #3851, opened 2016-09-09
Disable signup button on form submit

### LINKED ISSUE #3830: Prevent double-submitting of user registration form
It's possible to create an `IntegrityError` by double-clicking on the submit button on the user account registration form: https://sentry.io/hypothesis/prod/issues/154374560/activity/

Suggest solution: disable the submit button the first time it's clicked.

Presumably this much be done in JavaScript? And the no-js fallback behaviour is to just not have this feature (and continue to allow potential multiple submitting)?

What happens if the request is submitted but the browser doesn't receive a response. Do we _want_ to allow resubmitting of the form when that happens?

chdorner:
This is to prevent users from double-clicking the signup button and thus
sending two requests in quick succession which will probably result in
a SQLAlchemy integrity error as the username/email is already being
used when trying to commit the transaction.

Fixes #3830

### DIFF
--- a/h/accounts/views.py
+++ b/h/accounts/views.py
@@ -339,7 +339,9 @@ class SignupController(object):
         self.request = request
         self.schema = schemas.RegisterSchema().bind(request=self.request)
         self.form = request.create_form(self.schema,
-                                        buttons=(_('Sign up'),),
+                                        buttons=(deform.Button(title=_('Sign up'),
+                                                               css_class='js-signup-btn'),),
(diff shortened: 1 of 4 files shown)


nickstenning:
Ideally we'd style disabled form controls to signal to the user that they're disabled. This is also probably applicable more generally than just on the signup form. But it does the job for now. LGTM.

chdorner:
@nickstenning Yes, especially in the new design. Cc @conordelahunty

PR #3852, opened 2016-09-12
Show suggestion box of search facets when the search bar is in focus.

sheetaluk:
When the user focuses on the activity page search box a suggestion box is shown underneath which contains possible query terms the user can enter.
When the input field loses focus the sugestion box is hidden.
The user can use the up/down arrows and the enter key to select suggestions.
The user can use the mouse to select suggestions from the dropdown box.
The user can use a combination of the up and down arrows and the mouse to
highlight a facet.

https://trello.com/c/I6s1UK1R/434-show-suggestion-box-under-search-box-on-focus

### DIFF
--- a/h/static/scripts/controllers/search-bar-controller.js
+++ b/h/static/scripts/controllers/search-bar-controller.js
@@ -1,0 +1,180 @@
+'use strict';
+
+var Controller = require('../base/controller');
+var setElementState = require('../util/dom').setElementState;
+
+/**
+ * Controller for the search bar.
+ */
+class SearchBarController extends Controller {
+  constructor(element) {
+    super(element);
+
+    this._input = this.refs.searchBarInput;
+    this._dropdown = this.refs.searchBarDropdown;
+    this._dropdownItems = Array.from(
+      element.querySelectorAll('[data-ref="searchBarDropdownItem"]'));
+
+    var getActiveDropdownItem = () => {
+      return element.querySelector('.js-search-bar-dropdown-menu-item--active');
+    };
+
+    var clearActiveDropdownItem = () => {
+      var activeItem = getActiveDropdownItem();
+      if (activeItem) {
+        activeItem.classList.remove('js-search-bar-dropdown-menu-item--active');
+      }
+    };
+
+    var updateActiveDropdownItem = element => {
+      clearActiveDropdownItem();
+      element.classList.add('js-search-bar-dropdown-menu-item--active');
+    };
+
+    var selectFacet = facet => {
+      this._input.value = this._input.value + facet;
+
+      closeDropdown();
+
+      setTimeout(function() {
+        this._input.focus();
+      }.bind(this), 0);
+    };
+
+    var getPreviousSiblingElement = element => {
+      if (!element) {
+        return null;
+      }
+
+      do {
+        element = element.previousSibling;
+      } while (element && element.nodeType != 1);
+
+      return element;
+    };
+
+    var getNextSiblingElement = element => {
+      if (!element) {
+        return null;
+      }
+
+      do {
+        element = element.nextSibling;
+      } while (element && element.nodeType != 1);
+
+      return element;
+    };
+
+    var closeDropdown = () => {
+      clearActiveDropdownItem();
+      this.setState({open: false});
+      this._input.removeEventListener('keydown', setupListenerKeys,
+        true /*capture*/);
+    };
+
+    var openDropdown = () => {
+      this.setState({open: true});
+      this._input.addEventListener('keydown', setupListenerKeys,
+        true /*capture*/);
+    };
+
+    var setupListenerKeys = event => {
+      const ENTER_KEY_CODE = 13;
+      const UP_ARROW_KEY_CODE = 38;
+      const DOWN_ARROW_KEY_CODE = 40;
+
+      var activeItem = getActiveDropdownItem();
+      var handlers = {};
+
+      var handleEnterKey = event => {
+        event.preventDefault();
+
+        if (activeItem) {
+          var facet =
+            activeItem.
+              querySelector('[data-ref="searchBarDropdownItemTitle"]').
+              innerHTML.trim();
+          selectFacet(facet);
+        }
+      };
+
+      var handleUpArrowKey = event => {
+        updateActiveDropdownItem(getPreviousSiblingElement(activeItem) ||
+          this._dropdownItems[this._dropdownItems.length - 1]);
+      };
+
+      var handleDownArrowKey = event => {
+        updateActiveDropdownItem(getNextSiblingElement(activeItem) ||
+          this._dropdownItems[0]);
+      };
+
+      handlers[ENTER_KEY_CODE] = handleEnterKey;
+      handlers[UP_ARROW_KEY_CODE] = handleUpArrowKey;
+      handlers[DOWN_ARROW_KEY_CODE] = handleDownArrowKey;
+
+      var handler = handlers[event.keyCode];
+      if (handler) {
+        handler(event);
+      }
+    };
+
+    var handleClickOnItem = event => {
+      var facet =
+        event.currentTarget.
+          querySelector('[data-ref="searchBarDropdownItemTitle"]').
+          innerHTML.trim();
+      selectFacet(facet);
+    };
+
+    var handleHoverOnItem = event => {
+      updateActiveDropdownItem(event.currentTarget);
+    };
+
+    var handleClickOnDropdown = event => {
+      // prevent clicking on a part of the dropdown menu itself that
+      // isn't one of the suggestions from closing the menu
+      event.preventDefault();
+    };
+
+    var handleFocusOutside = event => {
+      if (!element.contains(event.target) ||
+        !element.contains(event.relatedTarget)) {
+        this.setState({open: false});
+      }
+      closeDropdown();
+    };
+
+    var handleFocusOnInput = () => {
+      if (this._input.value.trim().length > 0) {
+        closeDropdown();
+      } else {
+        openDropdown();
+      }
+    };
+
+    Object.keys(this._dropdownItems).forEach(function(key) {
+      var item = this._dropdownItems[key];
+      if(item && item.addEventListener) {
+        item.addEventListener('mouseover', handleHoverOnItem,
+          true);
+        item.addEventListener('mousedown', handleClickOnItem,
+          true);
+      }
+    }.bind(this));
+
+    this._dropdown.addEventListener('mousedown', handleClickOnDropdown,
+      true /*capture*/);
+    this._input.addEventListener('focusout', handleFocusOutside,
+      true /*capture*/);
+    this._input.addEventListener('input', handleFocusOnInput,
+      true /*capture*/);
+    this._input.addEventListener('focusin', handleFocusOnInput,
+      true /*capture*/);
+  }
+
+  update(state) {
+    setElementState(this._dropdown, {open: state.open});
+  }
+}
+
+module.exports = SearchBarController;
--- a/h/templates/panels/navbar.html.jinja2
+++ b/h/templates/panels/navbar.html.jinja2
@@ -1,5 +1,48 @@
 {% from '../includes/dropdown_menu.html.jinja2' import dropdown_menu %}
 
+{# Dropdown displaying search facet options. #}
+{% macro search_bar_dropdown() %}
+<div class="search-bar__dropdown-menu-container" data-ref="searchBarDropdown">
+  <h4 class="search-bar__dropdown-menu-header">
+    Narrow your search
+  </h4>
+  <ul class="search-bar__dropdown-menu">
+    <li class="search-bar__dropdown-menu-item" data-ref="searchBarDropdownItem">
+      <span class="search-bar__dropdown-menu-title" data-ref="searchBarDropdownItemTitle">
+        user:
+      </span>
+      <span class="search-bar__dropdown-menu-explanation">
+        search by username
+      </span>
+    </li>
+    <li class="search-bar__dropdown-menu-item" data-ref="searchBarDropdownItem">
+      <span class="search-bar__dropdown-menu-title" data-ref="searchBarDropdownItemTitle">
+        tag:
+      </span>
+      <span class="search-bar__dropdown-menu-explanation">
+        search for annotations with a tag
+      </span>
+    </li>
+    <li class="search-bar__dropdown-menu-item" data-ref="searchBarDropdownItem">
+      <span class="search-bar__dropdown-menu-title" data-ref="searchBarDropdownItemTitle">
+        url:
+      </span>
+      <span class="search-bar__dropdown-menu-explanation">
+        see all annotations on a page
+      </span>
+    </li>
+    <li class="search-bar__dropdown-menu-item" data-ref="searchBarDropdownItem">
+      <span class="search-bar__dropdown-menu-title" data-ref="searchBarDropdownItemTitle">
+        group:
+      </span>
+      <span class="search-bar__dropdown-menu-explanation">
+        show annotations created in a group you are a member of
+      </span>
+    </li>
+  </ul>
+</div>
+{% endmacro %}
+
 {% block content %}
 {% if feature('activity_pages') %}
 <header class="nav-bar">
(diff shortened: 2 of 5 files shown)


h/templates/panels/navbar.html.jinja2
```diff
@@ -1,16 +1,60 @@
...
+      <span class="search-bar__dropdown-menu-explanation">
+        see all annotations on a page
+      </span>
+    </li>
+    <li class="search-bar__dropdown-menu-item js-search-bar-dropdown-item">
+      <span class="search-bar__dropdown-menu-title js-search-bar-dropdown-item-title">
+        group:
+      </span>
+      <span class="search-bar__dropdown-menu-explanation">
+        show annotations created in a group you are a member of.
```
seanh: shouldn't be a full stop here

seanh:
Is it possible to prevent clicking on a part of the dropdown menu itself that isn't one of the suggestions from closing the menu?

peek 2016-09-12 13-19

seanh:
When typing, I wonder if the dropdown menu should stay open as long as what the user has typed so far matches one of the suggestions? Could be a separate pull request. Here's the interaction I'm talking about: I start typing "user:" and the menu closes as soon as I type the "u", it might be good if it stayed open (filtering down to show only the user and url suggestions, and then only the user suggestion) until I type the ":" then it should close.

peek 2016-09-12 13-24

seanh:
Another one that could be for a separate PR, should the menu open again after I type a space? For example in this GIF I've finished enter a "user:seanh" facet, and at this point I could now enter another user, url, group or tag facet, but the suggestions menu doesn't open again to help me:

peek 2016-09-12 13-27

seanh:
Apart from my first suggestion (to prevent clicking on a part of the dropdown menu itself that isn't one of the suggestions from closing the menu), which is a minor detail, this seems to work really well, I can't find any problems with it. Of my two other suggestions one should be a separate PR, and the other we'll get for free when we do one of the later Trello cards. So this is looking really good. Will take a look at the code now..

conordelahunty:
@nickstenning Just checking with you. This doesn't include anything about the 'Narrow your search' menu filtering when you type. Do we have that behaviour captured on one of the cards you wrote? I know we have something for the suggestions menu filtering.

h/static/scripts/controllers/search-bar-controller.js
```diff
@@ -0,0 +1,139 @@
...
+    var menuLen = this._dropdownItems.length;
+
+    var updateActiveDropdownItem = function (index) {
+      clearActiveDropdownItem();
+      self._dropdownItems[index].classList.add('js-search-bar-dropdown-menu-item--active');
+    };
+
+    var clearActiveDropdownItem = function () {
+      if (element.querySelector('.js-search-bar-dropdown-menu-item--active')) {
+        element.querySelector('.js-search-bar-dropdown-menu-item--active').
```
seanh: Can you avoid calling `querySelector()` twice by saving its result in a local variable?

h/static/scripts/controllers/search-bar-controller.js
```diff
@@ -0,0 +1,139 @@
...
+        classList.remove('js-search-bar-dropdown-menu-item--active');
+      }
+    };
+
+    var onOpenDropdown = function () {
+      self.setState({open: true});
+      self._input.addEventListener('keydown', setupListenerKeys,
```
seanh: Did you consider using the `'input'` event instead? That will trigger any time the contents of the `<input>` are changed whether by typing with the keyboard or pasting with the mouse etc, and it _won't_ trigger when a keyboard key is pressed but it doesn't change the contents.

h/static/scripts/controllers/search-bar-controller.js
```diff
@@ -0,0 +1,139 @@
...
+  constructor(element) {
+    super(element);
+
+    this._input = element.querySelector('.js-search-bar-input');
+    this._dropdown = element.querySelector('.js-search-bar-dropdown');
+    this._dropdownItems = element.querySelectorAll('.js-search-bar-dropdown-item');
+
+    var self = this;
+    var index = -1;
+    var menuLen = this._dropdownItems.length;
```
seanh: I think you should just access this as `this._dropdownItems.length` when you need it, rather than saving it in a class-level variable. Removes an unnecessary variable, but also it should continue working when, in future versions of this menu, the number of dropdown items can change dynamically

h/static/scripts/controllers/search-bar-controller.js
```diff
@@ -0,0 +1,139 @@
...
+ */
+class SearchBarController extends Controller {
+  constructor(element) {
+    super(element);
+
+    this._input = element.querySelector('.js-search-bar-input');
+    this._dropdown = element.querySelector('.js-search-bar-dropdown');
+    this._dropdownItems = element.querySelectorAll('.js-search-bar-dropdown-item');
+
+    var self = this;
```
seanh: Why not do this right after the `super()`? Then the three `this.*` lines above could use `self` as well

h/static/scripts/controllers/search-bar-controller.js
```diff
@@ -0,0 +1,139 @@
...
+    var index = -1;
```
seanh: Is it possible to get rid of this index variable? I think that when the up or down arrow is pressed, the code can find which menu item in the DOM that currently has the "active" CSS class on it, and then the menu item before or after that one, and then call `updateActiveDropdownItem()` passing it the DOM element for the new menu item to be selected. I think the code will be simpler if it doesn't have an `index` int to keep up to date.

h/static/scripts/controllers/search-bar-controller.js
```diff
@@ -0,0 +1,139 @@
...
+
+    var handleClickOnItem = event => {
+      var facet = event.currentTarget.querySelector('.js-search-bar-dropdown-item-title').innerHTML.trim();
+      onSelectFacet(facet);
+    }
+
+    var handleHoverOnItem = event => {
+      clearActiveDropdownItem();
+      index = Array.prototype.slice.call(this._dropdownItems).indexOf(event.currentTarget)
+      event.currentTarget.classList.add('js-search-bar-dropdown-menu-item--active');
```
seanh: I think this should call `updateActiveDropdownItem()` so that there's only one function that adds and removes the `'--active'` class

h/static/scripts/controllers/search-bar-controller.js
```diff
@@ -0,0 +1,139 @@
...
+    }
+
+    Object.keys(this._dropdownItems).forEach(function(key) {
+      var item = self._dropdownItems[key];
+      // FIXME
```
seanh: ?

seanh:
Ok, I've reviewed the `SearchBarController` class, I think it's generally quite clear and found it easy to understand, and it works very well. I think it would be good to try and simplify some parts of it, some suggestions: 

* I think getting rid of the `index` variable should simplify things (see comment above)
* I think there should only be one function that deals with adding and/or removing of the `'--active'` class, and all other functions should just call it. This should remove some duplication.
* I think all the event listeners should be added and removed in one place

h/static/scripts/controllers/search-bar-controller.js
```diff
@@ -0,0 +1,139 @@
...
+var setElementState = require('../util/dom').setElementState;
+
+/**
+ * Controller for the search bar.
+ */
+class SearchBarController extends Controller {
+  constructor(element) {
+    super(element);
+
+    this._input = element.querySelector('.js-search-bar-input');
```
robertknight: We've changed the way that elements referenced by controllers are identified to avoid confusion with the identifiers that trigger upgrades.

The new method is to add a `data-ref="name"` attribute on the element and then use `this.refs.name` inside the controller. eg. `data-ref="searchBarInput"`, `this.refs.searchBarInput`

h/static/scripts/controllers/search-bar-controller.js
```diff
@@ -0,0 +1,139 @@
...
+      }
+    };
+
+    var onOpenDropdown = function () {
+      self.setState({open: true});
+      self._input.addEventListener('keydown', setupListenerKeys,
+        true /*capture*/);
+    };
+
+    var onCloseDropdown = function () {
```
robertknight: In new (ES2015) code we should be consistent about using arrow functions rather than function expressions., ie. `var onCloseDropdown = () => { ... }`. This avoids the need to use `self` because the value of `this` is captured from the outer scope.

nickstenning:
> This doesn't include anything about the 'Narrow your search' menu filtering when you type.

Right, I think this is a "bug" in how the card is written. Typing something that is a prefix match for one or more of the options in the menu should result in the menu being narrowed to show the terms that match.

Happy for this to happen in a followup PR, but it should be done before the cards is marked as delivered.

sheetaluk:
>> Right, I think this is a "bug" in how the card is written. Typing something that is a prefix match for one or more of the options in the menu should result in the menu being narrowed to show the terms that match.

Will make this a followup PR.

sheetaluk:
@seanh:
>> When typing, I wonder if the dropdown menu should stay open as long as what the user has typed so far matches one of the suggestions?

This should fix the above issue.
https://github.com/hypothesis/h/pull/3852#issuecomment-246722373

sheetaluk:
@seanh 
>> Did you consider using the 'input' event instead? That will trigger any time the contents of the <input> are changed whether by typing with the keyboard or pasting with the mouse etc, and it won't trigger when a keyboard key is pressed but it doesn't change the contents.

Yes. But when a user presses the up/down arrow keys, we are not updating the contents of the input box until they press 'enter' which means the input box value doesn't change. So I went with the keydown event instead.

sheetaluk:
@conordelahunty Do you think we should allow the user to close the dropdown by pressing 'esc'?

h/static/scripts/controllers/search-bar-controller.js
```diff
@@ -0,0 +1,143 @@
...
+  constructor(element) {
+    super(element);
+
+    this._input = this.refs.searchBarInput;
+    this._dropdown = this.refs.searchBarDropdown;
+    this._dropdownItems = element.querySelectorAll('[data-ref="searchBarDropdownItem"]');
+
+    var getActiveElementIndex = () => {
+      return Array.prototype.slice.call(this._dropdownItems).
+        indexOf(element.querySelector('.js-search-bar-dropdown-menu-item--active'));
```
seanh: This confused me for a second (why not just do `this._dropdownItems.indexOf(...)`?) but I realised that it's because `this._dropdownItems` is a DOM `NodeList` not a JavaScript array, and `NodeList`s, don't have `.indexOf()`, and `Array.prototype.slice.call(this._dropdownItems)` converts the `NodeList` into an array.

With ES6 I think you can convert the `NodeList` into an array like this instead: `return [...this._dropdownItems].indexOf(...);`.

Also, I wonder if this should just be done at the top of the file when `this._dropdownItems` is first defined? Then it would always be an array and we wouldn't have to convert it in multiple places:

```javascript
this._dropdownItems = element.querySelectorAll('[data-ref="searchBarDropdownItem"]');
// Turn this._dropdownItems from a NodeList into an array:
this._dropdownItems = [...this._dropdownItems];
```

seanh: Ah, and @robertknight points out that there's also [Array.from()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/from)

h/static/scripts/controllers/search-bar-controller.js
```diff
@@ -0,0 +1,143 @@
...
+ */
+class SearchBarController extends Controller {
+  constructor(element) {
+    super(element);
+
+    this._input = this.refs.searchBarInput;
+    this._dropdown = this.refs.searchBarDropdown;
+    this._dropdownItems = element.querySelectorAll('[data-ref="searchBarDropdownItem"]');
+
+    var getActiveElementIndex = () => {
```
seanh: Is it really necessary to get the _index_ of the active element? Or could this be `getActiveElement()` instead, and then the code that uses it could work with the element itself rather than the element's index?

For example `updateActiveDropdownItem(index)` could become `updateActiveDropdownItem(element)` (where `element` is the element that should become the new active element). 

The code in `setupListenerKeys()` that gets the active element's index, increments it, and passes the new index to `updateActiveDropdownItem(index)` could instead get the active element, find its next sibling (`element.nextSibling` I think), and pass the next sibling element to `updateActiveDropdownItem(element)`.

I think this should allow the use of an `index` variable to be completely removed.

h/static/scripts/controllers/search-bar-controller.js
```diff
@@ -0,0 +1,143 @@
...
+      return Array.prototype.slice.call(this._dropdownItems).
+        indexOf(element.querySelector('.js-search-bar-dropdown-menu-item--active'));
+    };
+
+    var clearActiveDropdownItem = () => {
```
seanh: I think there's some inconsistent naming here: `getActiveElementIndex()` and then `clearActiveDropdownItem()`. Need to decide on "active element" vs "active dropdown item" and use the same one everywhere.

h/static/scripts/controllers/search-bar-controller.js
```diff
@@ -0,0 +1,143 @@
...
+        activeItem.classList.remove('js-search-bar-dropdown-menu-item--active');
+      }
+    };
+
+    var updateActiveDropdownItem = index => {
+      clearActiveDropdownItem();
+      this._dropdownItems[index].classList.add('js-search-bar-dropdown-menu-item--active');
+    };
+
+    var onSelectFacet = facet => {
```
seanh: I don't think this should be an `on*()` function because it's not a function that is called automatically on an event (e.g. `onEnterKeyPressed()` or something), rather it's called explicitly by our code in a couple of places, so I'd name the function after what it does, maybe just `selectFacet()`.

h/static/scripts/controllers/search-bar-controller.js
```diff
@@ -0,0 +1,143 @@
...
+      this._input.value = this._input.value + facet;
+
+      onCloseDropdown();
```
seanh: This one as well - seems to be an action rather than an event handler, I'd call it `closeDropdown()`.

h/static/scripts/controllers/search-bar-controller.js
```diff
@@ -0,0 +1,143 @@
...
+    };
+
+    var onCloseDropdown = () => {
+      clearActiveDropdownItem();
+      this.setState({open: false});
+      this._input.removeEventListener('keydown', setupListenerKeys,
+        true /*capture*/);
+    };
+
+    var onOpenDropdown = () => {
```
seanh: I'd call this `openDropdown()`

h/static/scripts/controllers/search-bar-controller.js
```diff
@@ -0,0 +1,143 @@
...
+
+    var handleHoverOnItem = event => {
+      var index = Array.prototype.slice.call(this._dropdownItems).indexOf(event.currentTarget);
+      updateActiveDropdownItem(index);
+    };
+
+    var handleClickOnDropdown = event => {
+      event.preventDefault();
+    };
+    this._dropdown.addEventListener('mousedown', handleClickOnDropdown, true);
```
seanh: I think this could do with a comment explaining why

seanh:
Looking even better and functionally-speaking I think it's perfect now. I just think it could do with a bit more simplification work to completely remove the use of an `index` variable if possible, and a few naming tweaks to make the naming of functions a bit clearer and more consistent.

h/static/scripts/controllers/search-bar-controller.js
```diff
@@ -0,0 +1,151 @@
...
+        } while (element && element.nodeType != 1);
+        return element;
+    };
+
+    var setupListenerKeys = event => {
+      var activeItem = element.querySelector('.js-search-bar-dropdown-menu-item--active');
+
+      switch (event.keyCode) {
+      case 13:
+        // enter key
```
seanh: Instead of comments I think these would be better as constants at the top of the function: `const ENTER_KEY_CODE = 13;`, then the code is nicely self-documenting.

h/static/scripts/controllers/search-bar-controller.js
```diff
@@ -0,0 +1,151 @@
...
+      case 38:
+        // up arrow key
+        updateActiveDropdownItem(getPreviousSiblingElement(activeItem) || this._dropdownItems[this._dropdownItems.length - 1]);
+        break;
+      case 40:
+        // down arrow key
+        updateActiveDropdownItem(getNextSiblingElement(activeItem) || this._dropdownItems[0]);
+        break;
+      default:
+        break;
```
seanh: I think instead of a long `switch` (which will potentially grow longer in the future) this might be better as a separate function for handling each key:

```javascript
var handlers = {};

var handleEnterKey = event => { ...};
handlers.ENTER_KEY_CODE = handleEnterKey;

var handleUpArrow = event => { ...};
handlers.UP_ARROW_KEY_CODE = handleUpArrow;

...

handlers[event.keyCode](event);
```

This would also mean that the handler functions could easily be called directly, for easy testing.

PR #3857, opened 2016-09-12
Denormalize document web uri

chdorner:
There are only a few document meta and document URI records that we actually need to display. So this pull request is to denormalize the document's first `http` or `https` URI onto the document itself when a new document gets created, or when an existing document gets updated and its missing the `web_uri`. We can save ourselves some extra queries on the document_uri table when reading from the database.


chdorner:
This will have to wait until #3856 is merged and run in production.

PR #3859, opened 2016-09-12
Replace sphinxcontrib-httpdomain API docs with OpenAPI

nickstenning:
This PR replaces our rather clunky existing API documentation with API docs
built from a data definition of our API using the OpenAPI format (AKA Swagger).

This allows us to directly reference JSON Schema objects to define the inputs
and outputs from our API, as well as give users much clearer examples of
conforming inputs and expected responses.


robertknight:
🎉

robertknight:
I got an error about a missing `sphinx_rtd_theme` package when trying to build the docs, and this package isn't listed in `requirements-dev.in`. Presumably it should be?

After installing the package, running `make html` in the docs dir worked fine. This is a great improvement over the previous docs.

nickstenning:
If anyone wants to see what this looks like without building it themselves: https://h.readthedocs.io/en/swagger-api-docs/api/

nickstenning:
For some additional context here, there are basically three major players in the "structured API documentation" world at the moment:

- [OpenAPI/Swagger](http://swagger.io/)
- [RAML](http://raml.org/)
- [API Blueprint](https://apiblueprint.org/)

The basic characteristics of these can be summarised as follows.

- RAML and Swagger are both YAML/JSON based API descriptions. Swagger leans heavily on JSON Schema, whereas RAML is more of its own thing. API Blueprint is a set of standards for formatting Markdown documents.
- RAML and Swagger are, by their nature, quite prescriptive in how they model APIs. They make assumptions about the ways in which requests map to responses which might not always be true. API Blueprint is lower-level, making it possible to describe pretty much anything which can be built with HTTP.
- RAML and Swagger both provide ways to break up API description documents into multiple files. RAML's is a custom syntax (`!include ...`) which is supported by only some of the RAML toolchain. Swagger's is based on JSON Pointer (`{"$ref": ...}`) and seems to be widely supported by Swagger tooling. API Blueprint appears to have no native support for this, but it could be built using Markdown transclusion tools such as [Hercule](https://github.com/jamesramsay/hercule).
- The actual documentation view/explore tools are best (in terms of user experience) for API Blueprint and Swagger. Both of these are supported by [Apiary](https://apiary.io/), but there are some limitations to Apiary's Swagger support. The two self-hostable API consoles I tried for RAML either [didn't support splitting up the API definition file](http://raml2html.leanlabs.io/) or [loaded a 29MiB unminifiable(!) JavaScript file](https://github.com/mulesoft/api-console).
- Swagger is potentially [convertible to API Blueprint](https://github.com/Humanstate/swagger2-markdown) without too much effort.
- Swagger and API Blueprint are the formats supported by [Dredd](https://dredd.readthedocs.io/en/latest/) an impressive-looking HTTP API contract testing tool.

For these reasons, I've chosen to go with Swagger for the time being. I'm very happy for us to reevaluate in the future.

PR #3862, opened 2016-09-13
Disallow empty document titles

seanh:
This prevents creation of `None`, `''`, whitespace-only strings and strings with leading and/or trailing whitespace as document titles, and removes and fixes these kinds of document titles that are already in the db. This will make it a lot easier for activity pages and future features to display documents without lots of edge cases where the display looks wrong - a document's title is always either `None` or a non-empty string.

On master, I can create documents with whitespace-only titles:

```bash
$ http POST 'http://localhost:5000/api/annotations' Authorization:'Bearer 6879-***' uri='foo' document:='{"title": " "}' | jq .document.title
[
  " "
]
```

Or with leading and/or trailing whitespace on the document title:

```bash
http POST 'http://localhost:5000/api/annotations' Authorization:'Bearer 6879-***' uri='bar' document:='{"title": " Title "}' | jq .document.title
[
  " Title "
]
```

It doesn't actually seem to be possible to create a `None` or empty string document title by posting to this API, but instances of both do somehow exist in our production database:

```sql
postgres=# SELECT count(*) FROM document_meta WHERE type = 'title' AND '' = ANY(value);
 count 
-------
   881
(1 row)
```

```python
In [3]: len([d for d in request.db.query(models.DocumentMeta).filter_by(type=u'title') if None in d.value])
Out[3]: 335
```

There aren't any `document_meta` rows whose `value` cell is an empty array, but after deleting `None` and `''` and whitespace-only strings from the arrays there will be, so the migrations remove those as well.

I've tested the db migrations here on prod data and looked at all the deleted and changed document metas - looks as expected.

Anyway, on this branch none of these kinds of document titles will be created anymore:

```bash
http POST 'http://localhost:5000/api/annotations' Authorization:'Bearer 6879-***' uri='fi' document:='{"title": " "}' | jq .document
{}
```

```bash
http POST 'http://localhost:5000/api/annotations' Authorization:'Bearer 6879-***' uri='fy' document:='{"title": " Leading whitespace"}' | jq .document
{
  "title": [
    "Leading whitespace"
  ]
}
```

```bash
http POST 'http://localhost:5000/api/annotations' Authorization:'Bearer 6879-***' uri='fo' document:='{"title": "Trailing whitespace "}' | jq .document
{
  "title": [
    "Trailing whitespace"
  ]
}
```

```bash
http POST 'http://localhost:5000/api/annotations' Authorization:'Bearer 6879-***' uri='fumb' document:='{"title": ""}' | jq .document
{}
```

```bash
http POST 'http://localhost:5000/api/annotations' Authorization:'Bearer 6879-***' uri='fum' document:='{"title": null}' | jq .document
{}
```

The validation that achieves this is in `parse_document_claims.py` rather than in the model. I did try to add a second layer of validation to the model code as well but it was complicated by two factors: 1. I only want to filter out these kinds of values when the `type` is `'title'` and 2. `value` is an array which can be initialized, set, appended to, inserted into etc and all cases would need to be validated (but only when `type` is `'title'`).

### DIFF
--- a/src/memex/parse_document_claims.py
+++ b/src/memex/parse_document_claims.py
@@ -101,7 +101,10 @@ def document_metas_from_data(document_data, claimant):
                 if not isinstance(value, list):
                     value = [value]
 
-                value = [v.strip() for v in value]
+                value = [v.strip() for v in value if v and v.strip()]
+
+                if not value:
+                    continue
 
                 document_meta_dicts.append({
                     'type': '.'.join(keypath),
--- a/src/memex/parse_document_claims.py
+++ b/src/memex/parse_document_claims.py
@@ -101,6 +101,8 @@ def document_metas_from_data(document_data, claimant):
                 if not isinstance(value, list):
                     value = [value]
 
+                value = [v.strip() for v in value]
+
                 document_meta_dicts.append({
                     'type': '.'.join(keypath),
                     'value': value,
--- a/src/memex/parse_document_claims.py
+++ b/src/memex/parse_document_claims.py
@@ -101,13 +101,18 @@ def document_metas_from_data(document_data, claimant):
                 if not isinstance(value, list):
                     value = [value]
 
-                value = [v.strip() for v in value if v and v.strip()]
+                type_ = '.'.join(keypath)
 
-                if not value:
-                    continue
+                if type_ == 'title':
+                    # We don't allow None, empty strings, whitespace-only
+                    # strings, leading or trailing whitespaces, or empty arrays
+                    # in document title values.
+                    value = [v.strip() for v in value if v and v.strip()]
+                    if not value:
+                        continue
 
                 document_meta_dicts.append({
-                    'type': '.'.join(keypath),
+                    'type': type_,
                     'value': value,
                     'claimant': claimant,
                 })
--- a/tests/memex/parse_document_claims_test.py
+++ b/tests/memex/parse_document_claims_test.py
@@ -201,11 +201,15 @@ class TestDocumentMetasFromData(object):
         # Leading and trailing whitespace gets stripped from document titles.
         (
             {
-                'title': ['   My Document', 'My Document   ', ' My Document '],
+                'title': ['   My Document',
+                          'My Document   ',
+                          ' My Document ',
+                          '\nMy Document\n\n'],
             },
             {
                 'type': 'title',
-                'value': ['My Document', 'My Document', 'My Document']
+                'value': ['My Document', 'My Document', 'My Document',
+                          'My Document']
             }
         ),
     ])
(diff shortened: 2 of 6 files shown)


seanh:
Needs a db migration to clean up the various kinds of empty titles we already have in our DB (`None1, leading and trailing whitespace, whitespace-only strings, and empty strings).

src/memex/parse_document_claims.py
```diff
@@ -101,7 +101,10 @@ def transform_meta_(document_meta_dicts, items, path_prefix=None):
                 if not isinstance(value, list):
                     value = [value]
 
-                value = [v.strip() for v in value]
+                value = [v.strip() for v in value if v and v.strip()]
+
+                if not value:
+                    continue
```
chdorner: Isn't this now being applied to all meta objects, and not just the ones of type `title`?

tests/memex/parse_document_claims_test.py
```diff
@@ -261,6 +265,36 @@ def test_document_metas_from_data_with_multiple_metadata_claims(self):
...
+
+    def test_document_metas_from_data_ignores_whitespace_only_titles(self):
+        """It should ignore whitespace-only document titles."""
+        for title in (' ', [' ', ' '], '\n\n  \n'):
+            document_data = {'title': title}
+
+            document_metas = parse_document_claims.document_metas_from_data(
+                document_data, 'http://example/claimant')
+
+            assert document_metas == []
```
chdorner: Considering my comment above, and if this behaviour should only apply to titles, then I think it would be good to have a test which shows that we don't ignore whitespace-only values for other types than `title`.

chdorner:
Probably not worth changing at this point, but I think it would have been better for the database to do all of these 4 migrations in one with something along the lines of:

```
ids_to_delete = []
for document_meta in session.query(DocumentMeta).filter_by(type='title'):
    new_titles = []
    for title in document_meta.value:
        if title is None:
            continue

        stripped = title.strip()
        if stripped:
            new_titles.append(stripped)

    if new_titles:
        document_meta.value = new_titles
    else:
        ids_to_delete.append(document_meta.id)

session.query(DocumentMeta).filter(DocumentMeta.id.in_(ids_to_delete)).delete()
```

tests/memex/parse_document_claims_test.py
```diff
@@ -196,7 +196,18 @@ class TestDocumentMetasFromData(object):
...
+        # Leading and trailing whitespace gets stripped from document titles.
+        (
+            {
+                'title': ['   My Document', 'My Document   ', ' My Document '],
+            },
+            {
+                'type': 'title',
+                'value': ['My Document', 'My Document', 'My Document']
+            }
+        ),
```
seanh: Sure, can do. Will just u'\n\ and u'\t' be enough?

src/memex/parse_document_claims.py
seanh: Hmm, good catch, I'll put this behind an `if`

tests/memex/parse_document_claims_test.py
seanh: Yep

tests/memex/parse_document_claims_test.py
chdorner: Yeah, maybe also add one of the more esoteric ones `u"\u2004"`

seanh: Done

src/memex/parse_document_claims.py
seanh: Done

tests/memex/parse_document_claims_test.py
seanh: Done

seanh:
@chdorner All done I think?

PR #3864, opened 2016-09-14
Execute query in _fetch_annotations

chdorner:
This way we get a more accurate breakout in New Relic of where time is
spent.


nickstenning:
LGTM.

PR #3865, opened 2016-09-14
Add a feature flag for deferring application of real-time updates in the client

robertknight:
Add a feature flag in preparation for implementing https://trello.com/c/tRZ2H7iH/452-defer-applying-realtime-updates in the client

### DIFF
--- a/h/features/models.py
+++ b/h/features/models.py
@@ -13,6 +13,8 @@ log = logging.getLogger(__name__)
 
 FEATURES = {
     'activity_pages': "Show the new activity pages?",
+    'defer_realtime_updates': ("Require a user action before applying real-time"
+                               " updates to annotations in the client?"),
     'orphans_tab': "Show the orphans tab to separate anchored and unanchored annotations?",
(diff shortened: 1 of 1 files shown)


h/features/models.py
```diff
@@ -13,6 +13,8 @@
 
 FEATURES = {
     'activity_pages': "Show the new activity pages?",
+    'defer_realtime_updates': """Require a user action before applying real-time
+                                 updates to annotations in the client?""",
```
robertknight: Is `textwrap.dedent` the idiomatic way to remove the extra spacing in the string here? It doesn't make a visible difference because the whitespace is collapsed down to a single space when this string is rendered in the feature flags page of the admin control panel.

nickstenning: In this case I'd be more inclined to do:

```python
    'defer_realtime_updates': ("Require a user action before applying real-time "
                               "updates to annotations in the client?"),
```

robertknight: 👍

Issue #3867, opened 2016-09-14
Direct link failing on an arXiv PDF


dwhly:
### Steps to reproduce
1. Click: https://hyp.is/4WFeIjFfEeaY6PfnfB_nDQ/arxiv.org/pdf/1312.5602v1.pdf

### Expected behaviour
User should be taken to the corresponding annotation

### Actual behaviour
Error: "Sorry, but it looks like this annotation was made on a document that is not publicly available"

### Browser/system information
About this version
Version:
0.42.0
User agent:
Mozilla/5.0 (Macintosh; Intel Mac OS X 10_10_5) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/52.0.2743.116 Safari/537.36
URL:
http://arxiv.org/pdf/1312.5602.pdf
PDF fingerprint:
71b82f824e88b33c359deaa8f9af44a1
Username:
dwhly
Date:
14 Sep 2016 15:03:30 -0700

chdorner:
This is a regression introduced with e45573b162a45420428bbad58e6262808ffc3e4f.
The annotation's `target[source]` is a `urn:x-pdf:...` URI, in which case bouncer [tries to load the document links](https://github.com/hypothesis/bouncer/blob/master/bouncer/util.py#L64-L73), but we stopped rendering these in the `DocumentJSONPresenter` because the client didn't need the links anymore.

I think the solution here is to provide a specific `DocumentSearchIndexPresenter` which adds the document links again, and use that when indexing an annotation.

chdorner:
This is fixed on production now (the reindex is about to finish within the next minutes, the above mentioned bouncer link already works).

dwhly:
Thank you!!!

PR #3869, opened 2016-09-15
Index document web_uri

chdorner:
This is part one of fixing #3867. It will add the new `web_uri` value to the search index. Part two will be changing bouncer to read from that field instead of looping through the document links (that don't exist anymore).


chdorner:
The bouncer counterpart is at https://github.com/hypothesis/bouncer/pull/32

PR #3872, opened 2016-09-16
Load activity replies

chdorner:
This is just searching Elasticsearch for the replies, and then loading them from the database. It isn't showing them in any way, it isn't even building up the conversation threads. Searching and loading the replies will help us figure out how fast this will be in production.

### DIFF
--- a/h/activity/query.py
+++ b/h/activity/query.py
@@ -8,7 +8,6 @@ from memex.search import Search
 from memex.search import parser
 from memex.search.query import (
     TagsAggregation,
-    TopLevelAnnotationsFilter,
     UsersAggregation,
 )
 import newrelic.agent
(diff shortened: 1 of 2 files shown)


h/activity/query.py
```diff
@@ -140,9 +143,21 @@ def aggregations_for(query):
 
 
 @newrelic.agent.function_trace()
+def fetch_annotations(session, ids, reply_ids):
+    def load_documents(query):
+        return query.options(subqueryload(Annotation.document))
+
+    annotations = storage.fetch_ordered_annotations(
```
nickstenning: It's worth keeping in mind that `fetch_ordered_annotations` does a bunch of work in Python that can be done by Postgres if the annotations are ordered by a column in the database. This may turn out to be irrelevant -- just flagging it up.

PR #3873, opened 2016-09-16
Mark 'selection_tabs' as pending removal

nickstenning:
This feature flag has been gone from the client for quite a while now. We could probably just remove it, but for the sake of safety we'll follow the standard procedure.


robertknight:
LGTM

PR #3875, opened 2016-09-16
Move Blocklist model into h.models package

nickstenning:
As part of an effort to move all models into the h.models package, this moves the Blocklist model out of h.badge and into h.models.


chdorner:
Rebased.

PR #3876, opened 2016-09-16
Move Feature/FeatureCohort models into h.models package

nickstenning:
As part of an effort to move all models into the h.models package, this moves the Feature and FeatureCohort models out of h.features and into h.models.


chdorner:
Rebased.

PR #3877, opened 2016-09-16
Move Group model into h.models package

nickstenning:
As part of an effort to move all models into the h.models package, this moves the Group model out of h.groups and into h.models.


chdorner:
Rebased.

PR #3881, opened 2016-09-16
Move NIPSA flag to a column on User

nickstenning:
This PR uses the database table created and populated in #3880, and adds a `nipsa` column to the User model. It also updates the NIPSA service and the relevant admin views to manage the NIPSA flag based on this column rather than on a separate table.

Database migrations to update the defaults on the `nipsa` column and make the column non-nullable will come in a subsequent commit.

By way of a bonus, this commit fixes an extant bug with NIPSA, where flagging a user with a mixed-case username would not correctly update the `nipsa` flag on their historical annotations.

_**N.B.** This PR is based on #3880 and will need rebasing once that merges._


chdorner:
Marked as WIP because we need to run migration ddb5f0baa429 on the prod database first.

h/nipsa/services.py
```diff
@@ -15,47 +15,41 @@ def __init__(self, session):
...
-    def flagged_userids(self):
+    def flagged_users(self):
         """
-        A list of all the NIPSA'd user IDs.
+        A list of all the NIPSA'd users.
 
         :rtype: list of unicode strings
         """
-        return [u.userid for u in self.session.query(models.NipsaUser)]
+        return self.session.query(User).filter_by(nipsa=True)
```
chdorner: This will be quite a bit slower than what it was before, but it's probably alright since we don't call this method very often. New Relic will help us identify this once it becomes an issue.

PR #3882, opened 2016-09-19
Narrow dropdown items for facets as user types.

sheetaluk:
When the user navigates with the up and down arrow, only cycle through
the visible items.
When the user's input does not match any of the facets, close the dropdown.
When the user enters a : close the dropdown.

https://trello.com/c/I6s1UK1R/434-show-suggestion-box-under-search-box-on-focus

### DIFF
--- a/h/static/scripts/controllers/search-bar-controller.js
+++ b/h/static/scripts/controllers/search-bar-controller.js
@@ -32,7 +32,7 @@ class SearchBarController extends Controller {
     };
 
     var selectFacet = facet => {
-      this._input.value = this._input.value + facet;
+      this._input.value = facet;
 
       closeDropdown();
 
@@ -41,41 +41,100 @@ class SearchBarController extends Controller {
       }.bind(this), 0);
     };
 
-    var getPreviousSiblingElement = element => {
+    var isHidden = element => {
+      return element &&
+        (element.nodeType !== 1 ||
+          !element.classList ||
+          element.classList.contains('is-hidden'));
+    };
+
+    var getPreviousVisibleSiblingElement = element => {
       if (!element) {
         return null;
       }
 
       do {
         element = element.previousSibling;
-      } while (element && element.nodeType != 1);
-
+      } while (isHidden(element));
       return element;
     };
 
-    var getNextSiblingElement = element => {
+    var getNextVisibleSiblingElement = element => {
       if (!element) {
         return null;
       }
 
       do {
         element = element.nextSibling;
-      } while (element && element.nodeType != 1);
+      } while (isHidden(element));
 
       return element;
     };
 
+    var showAllDropdownItems = () => {
+      this._dropdownItems.forEach(function(dropdownItem) {
+        dropdownItem.classList.remove('is-hidden');
+      });
+    };
+
     var closeDropdown = () => {
       clearActiveDropdownItem();
+      showAllDropdownItems();
       this.setState({open: false});
       this._input.removeEventListener('keydown', setupListenerKeys,
         true /*capture*/);
     };
 
     var openDropdown = () => {
+      clearActiveDropdownItem();
+
       this.setState({open: true});
+
       this._input.addEventListener('keydown', setupListenerKeys,
         true /*capture*/);
+    };
+
+    var getVisibleDropdownItems = () => {
+      return this._dropdown.querySelectorAll('li:not(.is-hidden)');
+    };
+
+    /** Show items that match the word and hide ones that don't. */
+    var setVisibleDropdownItems = word => {
+      this._dropdownItems.forEach(function(dropdownItem) {
+        var dropdownItemTitle =
+          dropdownItem.querySelector('[data-ref="searchBarDropdownItemTitle"]').
+            innerHTML.trim();
+        if (dropdownItemTitle.indexOf(word) < 0) {
+          dropdownItem.classList.add('is-hidden');
+        } else {
+          dropdownItem.classList.remove('is-hidden');
+        }
+      });
+    };
+
+    var getTrimmedInputValue = () => {
+      return this._input.value.trim();
+    };
+
+    var maybeOpenOrCloseDropdown = () => {
+      var word = getTrimmedInputValue();
+      var shouldOpenDropdown = true;
+
+      // If there are no visible items that match the word close the dropdown
+      if (getVisibleDropdownItems().length < 1) {
+        shouldOpenDropdown = false;
+      }
+
+      // If the word has a ':' don't show dropdown
+      if (word.indexOf(':') > -1) {
+        shouldOpenDropdown = false;
+      }
+
+      if (shouldOpenDropdown) {
+        openDropdown();
+      } else {
+        closeDropdown();
+      }
     };
 
     var setupListenerKeys = event => {
@@ -85,6 +144,8 @@ class SearchBarController extends Controller {
 
       var activeItem = getActiveDropdownItem();
       var handlers = {};
+
+      var visibleDropdownItems = getVisibleDropdownItems();
 
       var handleEnterKey = event => {
         event.preventDefault();
@@ -98,14 +159,14 @@ class SearchBarController extends Controller {
         }
       };
 
-      var handleUpArrowKey = event => {
-        updateActiveDropdownItem(getPreviousSiblingElement(activeItem) ||
-          this._dropdownItems[this._dropdownItems.length - 1]);
+      var handleUpArrowKey = () => {
+        updateActiveDropdownItem(getPreviousVisibleSiblingElement(activeItem) ||
+          visibleDropdownItems[visibleDropdownItems.length - 1]);
       };
 
-      var handleDownArrowKey = event => {
-        updateActiveDropdownItem(getNextSiblingElement(activeItem) ||
-          this._dropdownItems[0]);
+      var handleDownArrowKey = () => {
+        updateActiveDropdownItem(getNextVisibleSiblingElement(activeItem) ||
+          visibleDropdownItems[0]);
       };
 
       handlers[ENTER_KEY_CODE] = handleEnterKey;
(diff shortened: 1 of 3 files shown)


seanh:
I think the way the query-suggestion filtering is implemented is as specified on this Trello card <https://trello.com/c/glm03Us8/436-add-tag-suggestions-to-suggestion-box>, but it does lead to a couple of odd cases.

First weird thing: I enter "g" and "tag" is still available as an option, but at this point there's nothing else that I can type that won't make the tag option disappear, e.g. typing "a" or "t" will make it disappear:

peek 2016-09-19 16-13

That one's maybe not _so_ weird. This second one is a little worse: I enter "g", now the only I can get to "tag:" by using the keyboard is to move the cursor back to the start and type the "ta" in front of the g, but this makes the tag suggestion disappear and then reappear again:

peek 2016-09-19 16-14

Some alternative possibilities might be:

1. The query has to be a prefix match of the suggestion. "t", "ta" and "tag" will all match "tag:" but "g" and "ag" will not. This eliminates the two cases above, but it does mean that the user has to enter prefix matches, which might make the suggestions less helpful.

2. A suggestion is show as long as all of the letters in the query are also in the suggestion and are in the same order as in the query, but they don't have to be consecutive. So all of these would match the "tag:" suggestion:

  * "g"
  * "ag"
  * "tag"
  * "ta"
  * "t"
  * "tg"

  The letters are always in the right order, even if sometimes there are missing letters in-between.

   But these would not match it:

   * "ga"
   * "gat"
   * "at"
   * "gt"

   Because the letters are in the wrong order.

3. A suggestion is shown as long as all of the letters in the query are also in the suggestion, regardless of order _or_ consecutiveness. So all of the examples given in (2) above would match. Only when you type one letter other than "t", "a" or "g" would the "tag:" suggestion disappear (optionally it could also disappear if you typed two t's, two a's or two g's). This might be easier to implement than 2.

@nickstenning I think you wrote the Trello card, any thoughts on the above? I'm mostly trying to avoid the same suggestions disappearing and then reappearing again as you type (the example shown in the second gif), the idea is that adding letters to the query should only ever reduce the number of suggestions that match or leave it the same, never increase them.

h/static/scripts/controllers/search-bar-controller.js
```diff
@@ -73,11 +78,56 @@ class SearchBarController extends Controller {
     };
 
     var openDropdown = () => {
+      clearActiveDropdownItem();
+
+      // Reset dropdown items to not hidden
+      Object.keys(this._dropdownItems).forEach(function(key) {
+        this._dropdownItems[key].classList.remove('is-hidden');
+      }.bind(this));
```
seanh: Is it possible to write this more simply? Does `this_.dropdownItems.forEach(function(dropdownItem) {dropdownItem.classList.remove('is-hidden')})` not work?

h/static/scripts/controllers/search-bar-controller.js
```diff
@@ -73,11 +78,56 @@ class SearchBarController extends Controller {
...
+
+      if (word.length > 0) {
+        openDropdown();
+
+        // if the word has a ':' don't show dropdown
+        if (word.indexOf(':') > -1) {
+          closeDropdown();
+        }
+
+        // Get items that match the word
```
seanh: This looks like it actually shows items that match the word, and hides ones that don't, rather than just getting items

h/static/scripts/controllers/search-bar-controller.js
```diff
@@ -73,11 +78,56 @@ class SearchBarController extends Controller {
...
+
       this._input.addEventListener('keydown', setupListenerKeys,
         true /*capture*/);
     };
 
+    var getVisibleDropdownItems = () => {
+      return this._dropdown.querySelectorAll('li:not(.is-hidden)');
+    };
+
+    var maybeOpenOrCloseDropdown = () => {
```
seanh: I noticed that this function is actually doing two things: 1. Maybe either opening or closing the dropdown and 2. Showing and hiding all the dropdown items. I wonder if it should be split into two functions?

h/static/scripts/controllers/search-bar-controller.js
```diff
@@ -73,11 +78,56 @@ class SearchBarController extends Controller {
...
+      var word = this._input.value.trim();
+
+      if (word.length > 0) {
+        openDropdown();
+
+        // if the word has a ':' don't show dropdown
+        if (word.indexOf(':') > -1) {
+          closeDropdown();
+        }
```
seanh: It might be good to simplify this function a little so that it always either calls `openDropdown()` once or calls `closeDropdown()` once. At the moment I think it can under certain circumstances call `openDropdown()` and then `closeDropdown()`.

h/static/scripts/controllers/search-bar-controller.js
```diff
@@ -166,9 +218,9 @@ class SearchBarController extends Controller {
       true /*capture*/);
     this._input.addEventListener('focusout', handleFocusOutside,
       true /*capture*/);
-    this._input.addEventListener('input', handleFocusOnInput,
+    this._input.addEventListener('input', handleFocusinOnInput,
       true /*capture*/);
-    this._input.addEventListener('focusin', handleFocusOnInput,
+    this._input.addEventListener('focusin', handleInputOnInput,
```
seanh: Looks like the wrong handler functions on the wrong events! (Which doesn't break anything since they both do the same thing)

h/static/scripts/controllers/search-bar-controller.js
```diff
@@ -41,26 +41,31 @@ class SearchBarController extends Controller {
...
       }
 
       do {
         element = element.previousSibling;
-      } while (element && element.nodeType != 1);
-
+      } while (element &&
+        (element.nodeType !== 1 ||
+          !element.classList ||
+          element.classList.contains('is-hidden')));
```
seanh: It's not super obvious what this boolean expression means and it's repeated twice, might be worth moving it into a helper function named something like `isVisible(element)`?

h/static/scripts/controllers/search-bar-controller.js
```diff
@@ -41,43 +41,101 @@ class SearchBarController extends Controller {
       }.bind(this), 0);
     };
 
-    var getPreviousSiblingElement = element => {
+    var isVisible = element => {
+      return element &&
+        (element.nodeType !== 1 ||
+          !element.classList ||
+          element.classList.contains('is-hidden'));
+    };
```
seanh: Wait, should this be called `isHidden()`?

h/static/scripts/controllers/search-bar-controller.js
```diff
@@ -41,43 +41,101 @@ class SearchBarController extends Controller {
...
+      // If the word has a ':' don't show dropdown
+      if (word.indexOf(':') > -1) {
+        shouldOpenDropdown = false;
+      }
+
+      if (shouldOpenDropdown) {
+        openDropdown();
+      } else {
+        closeDropdown();
+      }
```
seanh: Can this function be written something like this?

```javascript
if (getVisibleDropdownItems().length < 1 || word.indexOf(':') > -1) {
  closeDropdown();
} else }
  openDropdown();
}
```

h/static/scripts/controllers/search-bar-controller.js
```diff
@@ -41,43 +41,101 @@ class SearchBarController extends Controller {
...
+          dropdownItem.classList.remove('is-hidden');
+        }
+      });
+    };
+
+    var maybeOpenOrCloseDropdown = () => {
+      var word = this._input.value.trim();
+      var shouldOpenDropdown = true;
+
+      setVisibleDropdownItems(word);
```
seanh: I think it should be up to `handleInputOnInput()` to call `setVisibleDropdownItems()` and then call `maybeOpenOrCloseDropdown()`. That way calling `maybeOpenOrCloseDropdown()` does only one thing - either opens or closes the dropdown. Rather than updating the visibility of the dropdown items and then opening or closing the dropdown.

I _think_ `handleFocusinOnInput()` should only need to call `maybeOpenOrCloseDropdown()`, and not `setVisibleDropdownItems()`? It doesn't need to update the visibility.

You'll end up with `this._input.value.trim()` repeated twice in `setVisibleDropdownItems()` and then in `maybeOpenOrCloseDropdown()`, so it might be worth moving that into a helper.

h/static/scripts/controllers/search-bar-controller.js
```diff
@@ -41,43 +41,101 @@ class SearchBarController extends Controller {
...
       do {
         element = element.nextSibling;
-      } while (element && element.nodeType != 1);
+      } while (isVisible(element));
 
       return element;
     };
 
+    var unHideDropdownItems = () => {
+      // Reset dropdown items to not hidden
```
seanh: This comment is probably unnecessary

seanh: `showAllDropdownItems()` might be a better name for this?

h/static/scripts/controllers/search-bar-controller.js
```diff
@@ -41,43 +41,101 @@ class SearchBarController extends Controller {
...
       this._input.addEventListener('keydown', setupListenerKeys,
         true /*capture*/);
     };
 
+    var getVisibleDropdownItems = () => {
+      return this._dropdown.querySelectorAll('li:not(.is-hidden)');
+    };
+
+    var setVisibleDropdownItems = word => {
+      // Show items that match the word and ones that don't
```
seanh: Typo, I think this comment should read: `// Show items that match the word and hide ones that don't.` Also it might be worth using a [jsdoc](http://usejsdoc.org/) comment here? We've got those in other js code.

nickstenning:
> @nickstenning I think you wrote the Trello card, any thoughts on the above?

Yes! It's a nice behaviour, but let's leave it for another day. For now, let's stick to the filtering we spec'd in the card.

nickstenning:
One query I had on the basis of the gifs above (and I've just checked out the branch and verified it myself) is that this doesn't currently implement the sorting of results as specified on the card:

> The matching candidates should be displayed in order, from lowest score to highest.

Where the "score" is the value returned by `.indexOf(query)`. That is, if I type in "g", then "group:" should display before "tag:" because the score of "group:" is 0, and the score of "tag:" is 2.

seanh:
Hmm, Sorting could be a good thing to do in a separate follow-up PR.

sheetaluk:
Sort of results can be implemented as part of:
https://trello.com/c/glm03Us8/436-add-tag-suggestions-to-suggestion-box

PR #3884, opened 2016-09-19
Add authticket table migration

chdorner:
This is the migration for the table that will be used as the storage for the authentication tickets coming up in the next pull request.

### DIFF
--- a/h/migrations/versions/1e88c31d8b1a_add_authticket_table.py
+++ b/h/migrations/versions/1e88c31d8b1a_add_authticket_table.py
@@ -1,0 +1,46 @@
+"""
+Add authticket table
+
+Revision ID: 1e88c31d8b1a
+Revises: f9d3058bec5f
+Create Date: 2016-09-16 12:22:26.480404
+"""
+
+from __future__ import unicode_literals
+
+from alembic import op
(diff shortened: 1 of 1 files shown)


h/migrations/versions/1e88c31d8b1a_add_authticket_table.py
```diff
@@ -0,0 +1,45 @@
...
+                    sa.Column('expires',
+                              sa.DateTime,
+                              nullable=False),
+                    sa.Column('user_id',
+                              sa.Integer(),
+                              nullable=False),
+                    sa.Column('user_userid',
+                              sa.UnicodeText(),
+                              nullable=False),
+                    sa.ForeignKeyConstraint(['user_id'], ['user.id']))
```
nickstenning: Shall we add on `ondelete='cascade'` here?

chdorner: Yes! Forgot about that

chdorner: Done

nickstenning:
LGTM.

PR #3885, opened 2016-09-19
Use new ticket auth policy

chdorner:
**This depends on #3884 being merged and run in production [✔ · DONE]**

This is replacing the authentication state storage previously done with `pyramid_redis_sessions` with `pyramid_authsanity`. This is step 1 of removing `pyramid_redis_sessions` from our code base.

When a user logs in they will get a randomly generated ticket id, we then store this id alongside the user id in the new `authticket` table, and the ticket id will get sent back to the client via the `auth` cookie. This should speed up most requests by about 10-30ms, because `pyramid_redis_sessions` was way to chatty with Redis, setting a new TTL every single time the `authenticated_userid` was being called.

The old session storage is valid for seven days, when a logged-in user accesses the site, but doesn't have a new authentication ticket yet, we will authenticate them via the old session policy and then create an auth ticket for them. After seven days we will be able to remove this fallback code and only use the new auth ticket system. This way we ensure that no logged-in user would suddenly get logged out.

### DIFF
--- a/h/auth/__init__.py
+++ b/h/auth/__init__.py
@@ -6,10 +6,12 @@ import logging
 
 from pyramid.authentication import (RemoteUserAuthenticationPolicy,
                                     SessionAuthenticationPolicy)
+import pyramid_authsanity
 from pyramid_multiauth import MultiAuthenticationPolicy
 
 from h.auth.policy import AuthenticationPolicy, TokenAuthenticationPolicy
 from h.auth.util import groupfinder
+from h.security import derive_key
 
 __all__ = (
     'DEFAULT_POLICY',
@@ -20,7 +22,18 @@ log = logging.getLogger(__name__)
 
 PROXY_POLICY = RemoteUserAuthenticationPolicy(environ_key='HTTP_X_FORWARDED_USER',
                                               callback=groupfinder)
+# We currently have three ways of authenticating a user:
+# 1. session - finds the authenticated userid in the session
+# 2. ticket - finds the authenticated user in the database through auth tickets
+# 3. token - finds the authenticated user in the database through tokens (API)
+#
+# We are currently in the progress of migrating the session policy to the ticket
+# policy, for non-API requests. This is only possible by having both of them
+# running for a certain time period. When a requests comes in that does not
+# authenticate with the ticket policy, but does with the session policy, then
+# we migrate this session over to use the new auth tickets.
 SESSION_POLICY = SessionAuthenticationPolicy(callback=groupfinder)
+TICKET_POLICY = pyramid_authsanity.AuthServicePolicy()
 TOKEN_POLICY = TokenAuthenticationPolicy(callback=groupfinder)
 
 DEFAULT_POLICY = AuthenticationPolicy(api_policy=TOKEN_POLICY,
@@ -40,6 +53,17 @@ def auth_domain(request):
 def includeme(config):
     global DEFAULT_POLICY
     global WEBSOCKET_POLICY
+
+    # Set up authsanity
+    config.register_service_factory('.services.auth_ticket_service_factory',
+                                    iface='pyramid_authsanity.interfaces.IAuthService')
+    settings = config.registry.settings
+    settings['authsanity.source'] = 'cookie'
+    settings['authsanity.cookie.max_age'] = 2592000
+    settings['authsanity.cookie.httponly'] = True
+    settings['authsanity.secret'] = derive_key(settings['secret_key'],
+                                               b'h.auth.cookie_secret')
+    config.include('pyramid_authsanity')
 
     if config.registry.settings.get('h.proxy_auth'):
         log.warn('Enabling proxy authentication mode: you MUST ensure that '
--- a/h/auth/__init__.py
+++ b/h/auth/__init__.py
@@ -37,8 +37,9 @@ TICKET_POLICY = pyramid_authsanity.AuthServicePolicy()
 TOKEN_POLICY = TokenAuthenticationPolicy(callback=groupfinder)
 
 DEFAULT_POLICY = AuthenticationPolicy(api_policy=TOKEN_POLICY,
-                                      fallback_policy=SESSION_POLICY)
-WEBSOCKET_POLICY = MultiAuthenticationPolicy([TOKEN_POLICY, SESSION_POLICY])
+                                      fallback_policy=TICKET_POLICY,
+                                      migration_policy=SESSION_POLICY)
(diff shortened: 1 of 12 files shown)


h/auth/__init__.py
```diff
@@ -41,6 +44,15 @@ def includeme(config):
     global DEFAULT_POLICY
     global WEBSOCKET_POLICY
 
+    config.register_service_factory('.services.auth_ticket_service_factory',
+                                    iface=pyramid_authsanity.interfaces.IAuthService)
+
+    # Set up authsanity
+    settings = config.registry.settings
+    settings['authsanity.source'] = 'cookie'
+    settings['authsanity.secret'] = derive_key(settings['secret_key'], b'h.auth')
```
nickstenning: We can probably be a bit more specific than `b'h.auth'`. Let's call it something like `b'h.auth.cookie_secret'`.

h/auth/__init__.py
```diff
@@ -41,6 +44,15 @@ def includeme(config):
     global DEFAULT_POLICY
     global WEBSOCKET_POLICY
 
+    config.register_service_factory('.services.auth_ticket_service_factory',
+                                    iface=pyramid_authsanity.interfaces.IAuthService)
+
+    # Set up authsanity
```
nickstenning: Might be worth moving this comment up a couple of lines, to indicate that the `auth_ticket_service_factory` is part of the configuration.

h/auth/__init__.py
```diff
@@ -41,6 +44,15 @@ def includeme(config):
     global DEFAULT_POLICY
     global WEBSOCKET_POLICY
 
+    config.register_service_factory('.services.auth_ticket_service_factory',
+                                    iface=pyramid_authsanity.interfaces.IAuthService)
```
nickstenning: The interface can be a dotted string here so you can avoid importing `pyramid_authsanity` altogether:

```python
    config.register_service_factory('.services.auth_ticket_service_factory',
                                    iface='pyramid_authsanity.interfaces.IAuthService')
```

h/auth/__init__.py
```diff
@@ -21,6 +23,7 @@
 PROXY_POLICY = RemoteUserAuthenticationPolicy(environ_key='HTTP_X_FORWARDED_USER',
                                               callback=groupfinder)
 SESSION_POLICY = SessionAuthenticationPolicy(callback=groupfinder)
+TICKET_POLICY = pyramid_authsanity.AuthServicePolicy()
```
nickstenning: As mentioned earlier, I think it would be good to add a paragraph here to explain what all these millions of authentication policies do and how the migration is going to happen.

h/auth/policy.py
```diff
@@ -7,14 +7,28 @@
...
     def authenticated_userid(self, request):
         if _is_api_request(request):
             return self.api_policy.authenticated_userid(request)
-        return self.fallback_policy.authenticated_userid(request)
+
+        userid = self.fallback_policy.authenticated_userid(request)
+        if userid is not None:
+            return userid
+
+        if self.migration_policy is None:
```
nickstenning: As before, just add some comments here to explain what's going on for somebody who doesn't have the context in their heads.

h/auth/__init__.py
```diff
@@ -41,6 +44,15 @@ def includeme(config):
     global DEFAULT_POLICY
     global WEBSOCKET_POLICY
 
+    config.register_service_factory('.services.auth_ticket_service_factory',
+                                    iface=pyramid_authsanity.interfaces.IAuthService)
+
+    # Set up authsanity
+    settings = config.registry.settings
+    settings['authsanity.source'] = 'cookie'
```
nickstenning: We probably also want to set `authsanity.cookie.max_age` and `authsanity.cookie.httponly` in order preserve behaviour with the current implementation.

h/auth/__init__.py
chdorner: We still need to import `pyramid_authsanity` for instantiating the ticket policy though?

nickstenning: Oh, yes, so you do. Probably still nicer to use a dotted string in places you can so there isn't a perceived import dependency where one isn't needed.

h/auth/worker.py
```diff
@@ -0,0 +1,19 @@
+# -*- coding: utf-8 -*-
+
+from __future__ import unicode_literals
```
chdorner: #3893

h/models/auth_ticket.py
```diff
@@ -0,0 +1,39 @@
...
+    id = sa.Column(sa.UnicodeText(), primary_key=True)
+
+    #: The datetime when this ticket expires
+    expires = sa.Column(sa.DateTime, nullable=False)
+
+    user_id = sa.Column(sa.Integer,
+                        sa.ForeignKey('user.id', ondelete='cascade'),
+                        nullable=False)
+
+    #: The user who's auth ticket it is
```
nickstenning: whose

chdorner: haha, fixed

h/auth/services.py
```diff
@@ -0,0 +1,114 @@
...
+
+        Raises ``AuthTicketNotLoadedError`` when auth ticket has not been
+        loaded yet, which signals the auth policy to call ``verify_ticket``.
+        """
+
+        if self._userid is None:
+            raise AuthTicketNotLoadedError('auth ticket is not loaded yet')
+
+        if self._user is None:
+            self._user = self.usersvc.fetch(self._userid)
```
nickstenning: Wait, why are we loading the user here? Wasn't the point to avoid having to hit the user table just to return the userid?

chdorner: Oh, that's what you meant, okay, can remove that. It won't affect much in terms of queries, other parts of the code will load the user, but I can definitely remove it from here.

h/auth/services.py
```diff
@@ -0,0 +1,114 @@
...
+
+        return self._user.userid
+
+    def groups(self):
+        """Returns security principals of the logged-in user."""
+
```
nickstenning: I'd throw AuthTicketNotLoadedError here as well if `self._userid is None`.

chdorner: But this will not be caught by `pyramid_authsanity` and will result in a 500, which might be okay?

nickstenning: That's fine because `pyramid_authsanity` only ever calls `authsvc.groups()` after first calling `authsvc.userid()` (and, if necessary, `authsvc.verify_ticket(..., ...)`).

chdorner:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;oPLkR1wv&#x2F;456-replace-pyramid-redis-sessions-for-authentication"> Replace pyramid_redis_sessions for authentication</a>

PR #3888, opened 2016-09-19
build env in containers using compose

cove:
Notes:
- Added Dockerfile.devenv to h, client and browser-extension (separate PRs)
- [cove/devenv](github.com/cove/devenv) is top level project to build and run the env using compose (need to transfer repo over)
- Switched prod docker build to deb since sass [doesn't compile under apline linux](https://github.com/sass/node-sass/issues/808https://github.com/sass/node-sass/issues/808) since it uses ulibc, the size is bigger with but the trade off is dev and prod are almost identical now
- All tests pass, [except for one](https://gist.github.com/cove/92962fadaf5af245bf0e7eaa162ea645), I'm not sure why
- Wasn't clear which versions of redis, pgsql et al. were needed so I guess, would be good to pin those down so it's tracked
- Didn't update the docs yet, wanted to get feedback here first
- Updated makefile targets to be a little more consistent across projects (standarize on make dev and make test)
- There is a race condition when starting up the containers while its dbs etc come online, i think it's been [worked around](https://github.com/cove/h/blob/c14n-devenv/Makefile#L44), but if it fails to start, retry, might need to add more automatic retries. 
- The dev envs run as root, so if you need to modify the files in your checkout you might need to sudo. it's a little tricky to work around this since you don't know what the uid of the user starting the env is without some hackery, seems like a minor quirk right now.
- If a python lib is updated, you'll need to rebuild the container since it's just in the container.
- The client and server pull the latest official binary dist since the official node container is using wheezy and jessie the the current stable.
- Running the h python tests under [docker on osx is very slow unfortunately](https://forums.docker.com/t/file-access-in-mounted-volumes-extremely-slow-cpu-bound/8076/158), so they'd need to be run on travis I think. The npm tests run well however.  

Here's how to try quickly
1) clone branch c14n-devenv and the devenv

```
 cd mydevenv/
 git clone  -b c14n-devenv git@github.com:cove/browser-extension.git
 git clone  -b c14n-devenv git@github.com:cove/h.git
 git clone  -b c14n-devenv git@github.com:cove/client.git
 git clone  git@github.com:cove/devenv.git
```

2) install docker :)
3) rebuild containers 

```
  cd devenv/
  make rebuild
```

4) start up devenv

```
  make up
```

You should then be able to connect to http://localhost:3000, run "make tests" in the devenv checkout, and update the client src.

Happy to make any changes. :)

### DIFF
--- a/docker-compose-common.env
+++ b/docker-compose-common.env
@@ -1,0 +1,19 @@
+TARGET_ENV=development
+
+# npm/assetwatcher
+HOME=/var/lib/hypothesis
+
+# webserver
+APP_URL=http://localhost:5000
+REDIS_HOST=redis
+ELASTICSEARCH_HOST=elasticsearch
+
+# webserver/websocket
+DATABASE_URL=postgresql://postgres@postgres/htest
+BROKER_URL=amqp://guest:guest@rabbit:5672/
+
+# postgres, db to create on startup
+POSTGRES_DB=htest
+
+# tests
+TEST_DATABASE_URL=postgresql://postgres@postgres/htest
--- a/docker-compose-common.yml
+++ b/docker-compose-common.yml
@@ -1,0 +1,10 @@
+version: '2'
+services:
+  common:
+    build: .
+    # NB: you have to export UID for it to be visable here
+    user: $UID
+    restart: always
+    env_file: docker-compose-common.env
+    volumes:
+     - .:/var/lib/hypothesis
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,57 @@
+version: '2'
+services:
+  webserver:
+    extends:
+     file: docker-compose-common.yml
+     service: common
+    # export MODEL_CREATE_ALL so it's picked up by sub-processes during migrations
+    command: /bin/sh -c "export MODEL_CREATE_ALL='true' SEARCH_AUTOCONFIG='true' && gunicorn --paste conf/development-app.ini"
+    ports:
+     - '5000:5000'
+    links:
+     - postgres
+     - elasticsearch
+     - rabbit
+     - redis
+     - websocket
+  websocket:
+    extends:
+     file: docker-compose-common.yml
+     service: common
+    command: gunicorn --reload --paste conf/development-websocket.ini
+    ports:
+     - '5001:5001'
+    links:
+     - postgres
+     - rabbit
+  worker:
+    extends:
+     file: docker-compose-common.yml
+     service: common
+    command: hypothesis --dev celery worker --autoreload
+    links:
+     - postgres
+  assetwatch:
+    extends:
+     file: docker-compose-common.yml
+     service: common
+    command: /bin/sh -c "npm install && npm rebuild node-sass && gulp watch"
+  postgres:
+    image: postgres:9.6
+    env_file: docker-compose-common.env
+    ports:
+     - '5432:5432'
+  elasticsearch:
+    image: nickstenning/elasticsearch-icu
+    ports:
+     - '9200:9200'
+     - '9300:9300'
+  rabbit:
+    image: rabbitmq:3.6.5-management
+    ports:
+     - '5672:5672'
+     - '15672:15672'
+  redis:
+    image: redis:3.2.3
+    ports:
+     - '6379:6379'
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -12,7 +12,6 @@ services:
      - postgres
      - elasticsearch
      - rabbit
-     - redis
      - websocket
   websocket:
     extends:
@@ -51,7 +50,3 @@ services:
     ports:
      - '5672:5672'
      - '15672:15672'
-  redis:
-    image: redis:3.2.3
-    ports:
-     - '6379:6379'


nickstenning:
Hey! Thanks for contributing to Hypothesis. This is definitely something we are interested in seeing, in particular because it allows us to orchestrate the external services as well as the various processes and daemons that comprise the h web service.

In order to get this merged, I'm going to suggest a number of changes to this PR, most of which are aimed at reducing the number of things we're changing at once, and some of which are aimed at making better use of the existing infrastructure (both from h itself and from docker-compose).

1. I don't think we need to change the Dockerfile to build against `debian/jessie`. While the node-sass build toolchain does have some problems [it's possible to build node-sass on Alpine](https://gist.github.com/nickstenning/537f40af1658096d067cf7fbe7a48d20). Indeed, our repository [already contains](https://github.com/hypothesis/h/tree/master/vendor) a prebuilt node-sass binary for Alpine, so you don't need to fix this problem yourself.
2. Let's focus, for now, on getting a working docker compose setup just for the `h` repository, without worrying about the client or the browser extension. We can always integrate the client and the browser extension later, but for now at least I think we should hold off on https://github.com/hypothesis/client/pull/121 and https://github.com/hypothesis/browser-extension/pull/8.
3. If at all possible, I'd like to avoid having a completely separate Dockerfile for the development environment. We could possibly do this by using a [Docker build argument](https://docs.docker.com/engine/reference/builder/#/arg) t

cove:
Hey Nick, Robert, thanks for the feedback here.

3. This is a judgment call type of thing IMO. The ARG idea would be a trade off between using tooling to diff two Dockerfiles to see what changed vs. referring back to a build system when reasoning about the inputs at build time to see why an image was different. I think I'd almost lean towards leaving the dev data in assuming it doesn't create a perf or security risk, and then making stripping it out an optimization down the line. But I'll take a look at adding a switch.

2./4.  I think I didn't quite follow your comment about the devserver due to lack of familiarity with the code (sorry!), but I see what you mean. My thought was that it would be nice to have a single button to push to get full env setup w/o having to install anything locally. I ran into some confusion on which version of node which caused me some friction during setup, so I wanted to see if I could help lower the bar for others. I'll see what I can do to deprecate the devserver however, thanks for the pointers. :)

cove:
Lmk if something like this is what you had in mind:
https://gist.github.com/cove/933717be6f38cf876ce511784853f93d

nickstenning:
Hi @cove. I took a look at the gist and have two pieces of feedback:

- I don't think we need a different Dockerfile for each container. We can simply override the `command` run in each container.
- There's a lot of duplication in that `docker-compose.yml`. I mentioned in the comment above Compose's ability to [define and extend services](https://docs.docker.com/compose/extends/#/extending-services), which should help avoid most of that boilerplate. See the [warehouse `docker-compose.yml`](https://github.com/pypa/warehouse/blob/master/docker-compose.yml) for an example.

cove:
making this commit much smaller, so it's just the docker compose config for review. if this looks ok, my thought is that we create another makefile target for containers to it doesn't interrupt anyone if there are any issue, then if it looks ok, switch make dev to the containerized version. thoughts?

nickstenning:
This branch is currently showing 18 commits in the diff. You may want to rebase this on current master.

cove:
reduced it to 1 commit to make it simpler

cove:
hey @nickstenning let me know if you have any other thoughts here. thx!

nickstenning:
Hi @cove!

First of all, sorry for the long delay since we last reviewed this. I've spent a couple of hours this morning playing with this setup, and I'm afraid I'm still reluctant to merge this as it is right now.

This is through no fault of yours. The main issue is about container boot ordering. Depending on which way you look at it, the problems is either a) that Compose doesn't provide a way of expressing boot order dependencies between containers, or b) that the h application isn't tolerant of its backing services (Postgres/Elasticsearch) not being available at boot.

The result (for me, at least) is that it's nearly impossible to get this thing to start up properly. Most of the time Postgres and Elasticsearch don't start fast enough and the application crashes. Yes, you could probably work around this by starting the postgres and elasticsearch containers separately, but I feel like if we're merging a Compose file, then `docker compose up` needs to Just Work :tm:.

There are a couple of additional nagging issues, too, which aren't really Compose's fault, but it's hard to see how to fix them just now. For example, we have to choose between 

a) keeping `node_modules/` on the host (as you've done), which makes it possible to share those files between containers but makes it impossible to then reuse that `node_modules/` directory on a different platform (which is a problem if you're running Compose on OS X, say) or 
b) leaving `node_modules/` in the container, which seems nicer, but poses problems in development, as we then need the asset watcher process to be running in the same container as the web process. It seems we might be able to solve this problem with a data volume, but I'm not sure.

If you've got any thoughts on the above, I'd be interested to hear them.

cove:
hey @nickstenning 

Re: Boot order, yes on "b" and definitely should Just Work, that's what I saw as being critical to lowering the bar for future contributions. IMO the fix here is that h needs to be updated to fail its healthcheck until the backend services come online and return 5xx to the clients in the meantime, instead it exits right now. Compose does restart it automatically however, and after a few restarts eventually everything would converge for me. It might take a while esp. if npm is installing packages for the first time.

Re: node_modules, my thought was that it should go in the container as in option "b" here too, but ran into npm fussiness and decided on the host was a reasonable first step vs. fighting with npm.  It's not a problem for the watcher process to run in the same container or to junk have another watcher container that share the same volume. That works fine as long as they're on the same host.

nickstenning:
Ok, well you're probably right that the h process shouldn't just bail if pg and es are unavailable. The only problem is that in development these services are expected to be up at boot so that we can automatically create tables and/or indices. I'll have a think about what we can do here.

cove:
As I recall Compose should still continue to restart h until its deps come up, and should ultimately work, despite the poor ux of having to wait out the startup errors.

nickstenning:
Hi @cove, thanks for your work on this. We've just added a `docker-compose.yml`, currently just for spinning up the services (Postgres, Elasticsearch, Rabbit), and we'll continue to evaluate whether moving the whole development environment to Compose is possible and desirable.

If it's ok with you I'll close this for now. Thanks again for nudging us towards using Compose!

P.S. we've made some changes recently to `h` which eliminate some of the problems we identified earlier in this thread.

nickstenning:
https://github.com/hypothesis/h/commit/749136cc01b6a135682226e17a494019f428b57a

cove:
@nickstenning oh great, glad you found something that's working for your team. 👍

PR #3891, opened 2016-09-20
Don't use an ORM object for request.auth_token

seanh:
request.auth_token is sometimes used outside of the request's sqlalchemy
transaction. For example after the transaction has been committed, or
after it has been invalidated and rolled back. Since request.auth_token
is a sqlalchemy ORM object trying to access any of its attributes at
these times will result in a DetachedInstanceError.

This commit avoids this problem by introducing a new h.auth.token.Token
which is a non-ORM wrapper class for the h.models.Token class, and using
this new class instead of the model class for request.auth_token.

For example this HTTP POST:

    http POST 'http://localhost:5000/api/annotations' Authorization:'Bearer 6879-***' uri='http://example.com' document:='{"dclink": [{"identifier": "doi:foo"}]}'

would trigger a separate issue which causes the document model code to
crash with a ProgrammingError from sqlalchemy as it tries to insert a
dict into the document_meta.value column (which is an array of
strings). Sqlalchemy crashes, which aborts the sqlalchemy transaction.

The Sentry crash reporting code then kicks in and tries to report the
ProgrammingError but crashes with a DetachedInstanceError of its own.
Excerpt from the traceback:

    File "/home/seanh/Projects/hypothesis/h/h/views/exceptions.py", line 50, in json_error
    _handle_exc(request)
    File "/home/seanh/Projects/hypothesis/h/h/views/exceptions.py", line 65, in _handle_exc
    request.sentry.captureException()
    ...
    File "/home/seanh/Projects/hypothesis/h/h/sentry.py", line 59, in _get_request_client
    client.user_context(user_context_data(request))
    File "/home/seanh/Projects/hypothesis/h/h/sentry.py", line 37, in user_context_data
    'id': request.authenticated_userid,
    ...
    File "/home/seanh/.virtualenvs/h/local/lib/python2.7/site-packages/pyramid/security.py", line 344, in authenticated_userid
    return policy.authenticated_userid(self)
    File "/home/seanh/Projects/hypothesis/h/h/auth/policy.py", line 17, in authenticated_userid
    File "/home/seanh/.virtualenvs/h/local/lib/python2.7/site-packages/pyramid/authentication.py", line 69, in authenticated_userid
    userid = self.unauthenticated_userid(request)
    File "/home/seanh/Projects/hypothesis/h/h/auth/policy.py", line 84, in unauthenticated_userid
    if not token.is_valid():
    File "/home/seanh/Projects/hypothesis/h/h/auth/models.py", line 50, in is_valid
    if self.expires is None:
    ...
    DetachedInstanceError: Instance <Token at 0x7fc96a388190> is not bound to a Session; attribute refresh operation cannot proceed

The issue is that request.auth_token is a sqlalchemy ORM object
(an h.auth.models.Token instance) and it is added using:

    config.add_request_method(auth_token, reify=True)

reify=True means that the first time something accesses
request.auth_token the actual auth_token() function will be called to
fetch the Token object from the db, but from then on Pyramid will cache
the object and just return the cached instance for each subsequent
access of request.auth_token.

After the model code raises the ProgrammingError the sqlalchemy session
to which this cached Token ORM object belongs to is aborted, which means
that the Token object is "detached" (in sqlalchemy's terminology) and
any attempt to access its properties will raise DetachedInstanceError.

As the traceback above shows, the crash reporting code in h/sentry.py
indirectly ends up accessing this auth token and getting such a
DetachedInstanceError.

The fix is to not use an ORM object for request.auth_token.


nickstenning:
LGTM.

PR #3893, opened 2016-09-20
Cleanup expired authtickets

chdorner:
**This depends on #3885 being merged and will need a rebase**

We should clean up expired auth tickets from the database, this adds a celery task that is being run periodically with `celery beat`.

This will also need some OPS work in the playbooks repo to get a celery beat (exactly one) process up and running.

### DIFF
--- a/h/celery.py
+++ b/h/celery.py
@@ -10,6 +10,7 @@ integrates it with the Pyramid application by attaching a bootstrapped fake
 
 from __future__ import absolute_import
 
+from datetime import timedelta
 import logging
 import os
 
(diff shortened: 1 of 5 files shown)


h/celery.py
```diff
@@ -37,9 +38,15 @@
     # want for all of our queues, but it makes the failure-mode behaviour of
     # Celery the same as our old NSQ worker:
     CELERY_ACKS_LATE=True,
+    CELERYBEAT_SCHEDULE={
```
nickstenning: Sorry to be, uh, _me_, but `CELERYBEAT` sorts before `CELERY_ACCEPT_CONTENT` ;)

nickstenning:
At some point in the future I guess we probably do want to move the schedule to the database, but for now, as you say, there's no harm in not running this exactly every hour.

The only possible problem with that (if we don't bind-mount the shelve file into the container) is that deploying every 59 minutes would result in the task never getting run.

chdorner:
Fixed the sorting. Remember, this also depends on https://github.com/hypothesis/playbook/pull/70

chdorner:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;oPLkR1wv&#x2F;456-replace-pyramid-redis-sessions-for-authentication"> Replace pyramid_redis_sessions for authentication</a>

nickstenning:
LGTM.

PR #3895, opened 2016-09-21
Implement XHR form submission client

robertknight:
_Split out from https://github.com/hypothesis/h/pull/3814_

This implements the client-side logic for handling XHR form submissions and changes the form submission handling code to avoid rendering flash messages for XHR submissions.

### DIFF
--- a/h/static/scripts/tests/util/submit-form-test.js
+++ b/h/static/scripts/tests/util/submit-form-test.js
@@ -1,0 +1,63 @@
+'use strict';
+
+var fetchMock = require('fetch-mock');
+
+var submitForm = require('../../util/submit-form');
+
+describe('submitForm', function () {
+  var FORM_URL = 'http://example.org/things';
+
+  function mockResponse(response) {
+    fetchMock.post(FORM_URL, response);
+  }
+
+  function createForm() {
+    var form = document.createElement('form');
+    form.action = FORM_URL;
+    form.method = 'POST';
+    form.innerHTML = '<input name="field" value="value">';
+    return form;
+  }
+
+  it('submits the form data', function () {
+    var form = createForm();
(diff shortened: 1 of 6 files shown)


h/static/scripts/tests/util/submit-form-test.js
```diff
@@ -0,0 +1,98 @@
...
+var submitForm = require('../../util/submit-form');
+
+/**
+ * Fake `fetch()` implementation.
+ *
+ * This does not mock the global `window.fetch` function but provides a mock
+ * via `this.fetch` which can be passed as a dependency to functions that make
+ * network requests.
+ */
+class FakeFetch {
```
nickstenning: Sinon already provides a full XHR stubbing toolkit, so let's not introduce assumptions into our tests about exactly *how* we utilise XHR, but instead stub requests/responses and ensure they result in the right behaviour.

robertknight: A caveat is that sinon's stubbing toolkit doesn't work if we ever run our tests in a browser which _does_ support fetch natively, such as headless Chrome for example.

I personally think that faking the Fetch API directly, but re-using the real [Response](https://developer.mozilla.org/en-US/docs/Web/API/Response/Response) class, is nicer, as it involves less indirection/magic.

nickstenning: Ok, I didn't realise that Chrome `fetch` wasn't a wrapper around XMLHttpRequest. In that case let's not implement something [that already exists](https://www.npmjs.com/package/fetch-mock) for ourselves.

robertknight: Done. I did look at `fetch-mock` in the past and dismissed it but I can't recall why. I took another look again just now and it worked fine.

Issue #3887, opened 2016-09-19
Documents annotated with Via have via.hypothes.is as their domain name


seanh:
![screenshot from 2016-09-19 20-00-13](https://cloud.githubusercontent.com/assets/22498/18645102/c47a35a4-7ea3-11e6-8815-fe90053b87d8.png)

nickstenning:
I don't think this is true in general, because I've just made a bunch of annotations through via and they don't show up this way. There must be specific circumstances under which we (incorrectly) end up storing the via prefix on the annotations.

seanh:
Hmm, just made a test private annotation on prod via and it does show up like this. I'll see if I can reproduce locally

chdorner:
#3910 fixed this.

Issue #3860, opened 2016-09-13
None is shown on search page when document has no title


seanh:
If the document has `None` for its title then `"None"` is displayed on the search page:

![screenshot from 2016-09-13 14-45-25](https://cloud.githubusercontent.com/assets/22498/18476137/d2d6f4a8-79c0-11e6-8444-6d5fb37d9f8f.png)

chdorner:
Exciting! (we have a few more of these)

seanh:
@chdorner Was going to ask you if this is really an issue? The model can't guarantee that a document does actually have a title, can it? So it's always a possibility that the title can be `None`, and the presentation code has to deal with this? Or can we do something about this in the model?

chdorner:
You're right, there isn't much we can do to guarantee a title. But it shouldn't show "None" :)

seanh:
Yeah, I think something like "Untitled document" should be fine

seanh:
This was fixed by https://github.com/hypothesis/h/pull/3908

Issue #3858, opened 2016-09-12
Local files have "None" under address on activity search page


seanh:
![screenshot from 2016-09-12 16-15-05](https://cloud.githubusercontent.com/assets/22498/18444701/3666c552-7912-11e6-99b4-86a32425486d.png)

Issue #3822, opened 2016-09-02
Document buckets look wrong when document title is an empty string


seanh:
![title](https://cloud.githubusercontent.com/assets/22498/18204614/9c6f620a-7115-11e6-8a58-bdd5cd8ba0be.png)

seanh:
I _think_ that this one happens when the title is `''`, because if the title is `None` then you get <https://github.com/hypothesis/h/issues/3860> instead

seanh:
This was fixed by https://github.com/hypothesis/h/pull/3862

PR #3896, opened 2016-09-21
Fix display of documents without titles on activity search page

seanh:
This will show you the problem on master:

```bash
http POST 'http://localhost:5000/api/annotations' Authorization:'Bearer 6879-***' uri='http://example'
```

Then visit `/search` and see the untitled document displayed badly. Then checkout this branch and see the fix.

### DIFF
--- a/h/activity/bucketing.py
+++ b/h/activity/bucketing.py
@@ -10,6 +10,7 @@ import newrelic.agent
 from pyramid import i18n
 
 from h._compat import urlparse
+from h import presenters
 
 
 _ = i18n.TranslationStringFactory(__package__)
@@ -22,7 +23,9 @@ class DocumentBucket(object):
         self.users = set()
         self.uri = None
 
-        self.title = document.title
+        presented_document = presenters.DocumentHTMLPresenter(document)
+
+        self.title = presented_document.title_or_untitled
 
         if document.web_uri:
             parsed = urlparse.urlparse(document.web_uri)
--- a/h/presenters.py
+++ b/h/presenters.py
@@ -168,7 +168,7 @@ class AnnotationHTMLPresenter(object):
     def title(self):
         """Return a title for this annotation."""
         if self.document:
-            return self.document.title
+            return self.document.title_or_filename_or_uri
         else:
             return ''
 
@@ -304,7 +304,8 @@ class DocumentHTMLPresenter(object):
 
         """
         return _format_document_link(
-            self.href, self.title, self.link_text, self.hostname_or_filename)
+            self.href, self.title_or_filename_or_uri, self.link_text,
+            self.hostname_or_filename)
 
     @property
     def link_text(self):
@@ -323,11 +324,11 @@ class DocumentHTMLPresenter(object):
         Markup object so it doesn't get double-escaped.
 
         """
-        title = jinja2.escape(self.title)
+        title = jinja2.escape(self.title_or_filename_or_uri)
 
-        # Sometimes self.title is the annotated document's URI (if the document
-        # has no title). In those cases we want to remove the http(s):// from
-        # the front and unquote it for link text.
+        # Sometimes self.title_or_filename_or_uri is the annotated document's
+        # URI (if the document has no title). In those cases we want to remove
+        # the http(s):// from the front and unquote it for link text.
         lower = title.lower()
         if lower.startswith('http://') or lower.startswith('https://'):
             parts = urlparse.urlparse(title)
@@ -337,9 +338,9 @@ class DocumentHTMLPresenter(object):
             return title
 
     @property
-    def title(self):
+    def title_or_filename_or_uri(self):
         """
-        Return a title for this document.
+        Return a title for this document or fall back on filename or URI.
 
         Return the document's title or if the document has no title then return
         its filename (if it's a file:// URI) or its URI for non-file URIs.
--- a/h/presenters.py
+++ b/h/presenters.py
@@ -3,10 +3,14 @@ from __future__ import unicode_literals
 import urlparse
 import urllib2
 from dateutil import parser
+from pyramid import i18n
 
 import jinja2
 
 from h._compat import text_type
+
+
+_ = i18n.TranslationStringFactory(__package__)
 
 
 def _format_document_link(href, title, link_text, host_or_filename):
@@ -362,6 +366,11 @@ class DocumentHTMLPresenter(object):
             return jinja2.escape(urllib2.unquote(self.filename))
         else:
             return jinja2.escape(urllib2.unquote(self.uri))
+
+    @property
+    def title_or_untitled(self):
+        """Return this document's title or 'Untitled Document'."""
+        return self.document.title or  _("Untitled Document")
 
     @property
     def uri(self):
--- a/tests/h/activity/bucketing_test.py
+++ b/tests/h/activity/bucketing_test.py
@@ -32,18 +32,17 @@ class timeframe_with(object):  # noqa: N801
             n=len(self.document_buckets))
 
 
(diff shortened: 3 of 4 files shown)


nickstenning:
Why would we not just do this? This PR introduces a *lot* of code and lot of new paths through the presenter that I don't think are really justified.

```diff
diff --git a/h/templates/activity/search.html.jinja2 b/h/templates/activity/search.html.jinja2
index 336d1d4..f9e3107 100644
--- a/h/templates/activity/search.html.jinja2
+++ b/h/templates/activity/search.html.jinja2
@@ -44,7 +44,13 @@
     <div class="search-result-bucket__domain">
       {{ bucket.domain }}
     </div>
-    <div class="search-result-bucket__title">{{ bucket.title }}</div>
+    <div class="search-result-bucket__title">
+      {% if bucket.title %}
+        {{ bucket.title }}
+      {% else %}
+        {% trans %}Untitled document{% endtrans %}
+      {% endif %}
+    </div>
     <div class="search-result-bucket__annotations-count">
       <div class="search-result-bucket__annotations-count-container">
         {{ bucket.annotations_count }}
```

seanh:
We wanted to show either the title or "Untitled Document", but the existing `DocumentHTMLPresenter.title` property would return the title or, if there wasn't a title, fall back on the filename or, if there wasn't a filename, fall back on the URI. So in the case of the simple template code snippet you posted - it would only display "Untitled Document" for a document with no title, filename or URI. You would get numerical PDF identifiers in place of document bucket titles, or filenames in place of the titles (when the filename is already shown to the right - so it displays the same filename twice right next to each other), etc.

What I've done in this PR is:

1. Rename `DocumentHTMLPresenter.title` to `DocumentHTMLPresenter.title_of_filename_or_uri`, other parts of the code that are still using this behaviour still use it.

2. Add `DocumentHTMLPresenter.title_or_untitled` which just does what you've done in your template snippet except in Python code

I'm not sure this does introduce a lot of new code / new paths through the code? The diff looks big because of the renamed method. But the only _new_ code in this  is the `DocumentHTMLPresenter.title_or_untitled` property which is just `return self.document.title or  _("Untitled Document")`. And I guess the fact that the bucketing code now uses `DocumentHTMLPresenter` is also new, but it's going to need to use that for more upcoming changes as well.

nickstenning:
If that's true, why was it displaying `None` before for an annotation which clearly had a URI?

nickstenning:
Basically, I'm asking this: why do we need two different concepts of a document title? Can't we just have one?

nickstenning:
> If that's true, why was it displaying None before for an annotation which clearly had a URI?

I just got a chance to look at this and it's because if you just use `document.title` then you're not using the presenter at all, so "You would get numerical PDF identifiers in place of document bucket titles, or URLs in place of the titles" isn't true.

Let's fix this one in the simplest way possible (i.e. don't use the presenter) and then evaluate the situation if we later decide we do need to use a presenter here.

seanh:
Previously the bucketing code was just using `Document.title` and not using `DocumentHTMLPresenter`, so for a document with no title it would render `None`. I suppose it's not strictly necessary in order to fix this bug but I want to move the bucketing over to using `DocumentHTMLPresenter` as part of fixing the various display bugs with document buckets, `DocumentHTMLPresenter` seems like the right place for this kind of logic to go, and there should be existing code in there that it can re-use (although in the case of document titles it turned out that it couldn't just use the existing `DocumentHTMLPresenter.title` field).
 
By two different concepts of a document title, you're referring to `title_or_filename_or_uri` and `title_or_untitled`?

There are some parts of the code that _want_ to display the title falling back to the filename or the URI, for example the legacy groups pages use this (and I remember some lengthy back-and-forth discussions deciding exactly what should be displayed on that page for different kinds of documents) and I didn't want to change / break the behaviour of those pages.

At the same time for this new page we don't want that behaviour, we want the title or untitled behaviour.

When we remove the legacy pages, if no code is left using `title_or_filename_or_uri` then we can of course remove it.

But in general I don't think we necessarily want to have just one concept of how to display a title. Different pages with different designs may want to display titles in different ways. On one page it might make sense to fall back on the filename or URL. On another page, maybe the filename is already displayed next to the title anyway and it wouldn't make sense to repeat the filename again as the title. Or maybe on another page the title is shown in a large title font where showing a URL in place of the title would look really ugly. Etc. Basically different designs may call for titles to be displayed in different ways so I think in general we should be prepared for multiple `DocumentHTMLPresenter.title_*` properties. But lets put them in `DocumentHTMLPresenter` so that they're all in one place and so that when two page designs can reuse the same display logic they can reuse the code.

seanh:
I'd like the bucketing code to use `DocumentHTMLPresenter` if you don't mind. The very next bucket display bug fix that I have in progress (the one where `via.hypothes.is` is displayed as the domain name) was going to work by adding `DocumentHTMLPresenter.web_uri` that wraps `Document.web_uri` and strips `https://via.hypothes.is`, so the `bucketing.py` will need to use `DocumentHTMLPresenter` for that anyway (although that's for the `uri` field not `title`).

I count ~7 display logic issues on the new `/search` page (<https://github.com/hypothesis/h/issues/3887>, <https://github.com/hypothesis/h/issues/3860>, <https://github.com/hypothesis/h/issues/3858>, <https://github.com/hypothesis/h/issues/3822>, <https://github.com/hypothesis/h/pull/3789>, <https://github.com/hypothesis/h/issues/3742>, <https://github.com/hypothesis/h/issues/3739>) and I think this comes down to the fact that there are just too many variations and possibilities with our annotation and document data and if you try to render those in any kind of design, you will miss things and make mistakes. Admittedly not _all_ of these should be fixed in `DocumentHTMLPresenter`. But I think it seems smart to add reusable, tested logic in `DocumentHTMLPresenter` and use/reuse it as much as possible, rather than each page coding its own. My general approach to these has been: 1. Eliminate possibilities by tightening validation and cleaning up the db; then 2. Move display logic into `DocumentHTMLPresenter`, handle the remaining cases in one testable place there; then 3. Where use of `DocumentHTMLPresenter` isn't appropriate just fix it with per-page logic.

nickstenning:
Ok, I can see why you want to push this into the presenter, but I think the `title_or_untitled` vs `title_or_filename_or_uri` thing is a bit messy.

So here's what I'd suggest:

- the groups pages actually use `DocumentHTMLPresenter#link`, so move the "title or filename or uri" logic into that method (or a helper called by that method)
- have `DocumentHTMLPresenter#title` just return the title (or empty string, if there is no title)
- move the "Untitled document" into the template, so that decisions about what exactly is appropriate to render in a given context for a document without a title are left up to the place with the most context, rather than us ending up in a situation where we need a `title_or_something` method for each variant.

nickstenning:
(This approach also has the advantage that you can break it up into two PRs: 1) refactor the relationship between the link and title properties, 2) use the presenter and the title property for activity pages.)

seanh:
`Document.title` is already always either `None` or a non-empty, non-whitespace string, so a `DocumentHTMLPresenter.title` that returns either the `Document.title` or `""` would be adding very little.

The thing about having a `DocumentHTMLPresenter.title_or_untitled` that actually returns the `"Untitled Document"`, compared to letting the page or template transform `None` into `"Untitled Document"` itself, is that it guards against future inconsistencies where some pages will display "Untitled Document" and some "Untitled document" and some "Unknown Document", and some will forget this case entirely and just display an empty string, some will internationalize it and some will fail to, etc.

This compromise approach is going to be more work than either of our original suggestions and may end up being a worst of both worlds. I don't want to get stuck in the mud here so I've just done <https://github.com/hypothesis/h/pull/3908/>

PR #3897, opened 2016-09-21
Fix documents without URIs on activity search page

### LINKED ISSUE #3858: Local files have "None" under address on activity search page

seanh:
Not all documents have a URI (e.g. locally-annotated PDF files don't
have one). Instead of displaying "Address: None" for such documents on
the /search page, just don't display the address field at all.
(This is the same thing that the search page does for e.g. documents
with no tags.)

Fixes #3858


seanh:
screenshot from 2016-09-21 15-08-12

PR #3898, opened 2016-09-21
Remove deprecated session auth

chdorner:
**This should not be merged until 2016-09-28 to allow all users to migrate to the new auth system**


chdorner:
Will restructure this

PR #3899, opened 2016-09-21
Remove deprecated session authentication

chdorner:
**This should not be merged until 2016-09-28 to allow all users to migrate to the new auth system**

This PR removes the fallback to the session authentication policy and only uses the ticket policy for non-API requests. This is to remove the migration we added in #3885.

### DIFF
--- a/h/accounts/__init__.py
+++ b/h/accounts/__init__.py
@@ -36,18 +36,6 @@ def authenticated_user(request):
     user_service = request.find_service(name='user')
     user = user_service.fetch(request.authenticated_userid)
 
-    # If the authenticated user doesn't exist in the db then log them out.
-    # This happens when we delete a user account but the user still has a
-    # valid session for that account (because we don't invalidate sessions
-    # when we delete user accounts).
-    #
-    # FIXME: switch to authentication ticket based sessions so that we *can*
-    # invalidate sessions when deleting users. Throwing a 302 in the middle of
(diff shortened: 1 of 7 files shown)


h/accounts/__init__.py
```diff
@@ -36,18 +36,6 @@ def authenticated_user(request):
...
 
-    # If the authenticated user doesn't exist in the db then log them out.
-    # This happens when we delete a user account but the user still has a
-    # valid session for that account (because we don't invalidate sessions
-    # when we delete user accounts).
-    #
-    # FIXME: switch to authentication ticket based sessions so that we *can*
-    # invalidate sessions when deleting users. Throwing a 302 in the middle of
-    # an arbitrary request is NOT safe (e.g. POST requests).
-    if request.authenticated_userid and not user:
```
nickstenning: If we're going to remove this we should probably also ensure that when users are renamed any authtickets they have open are removed from the database, as this case isn't covered by our `on delete cascade`.

chdorner: Good point, I thought that this will be automatically handled since we check for the username, but that's the username we stored in the cookie during the creation of the auth ticket, so that's not going to work.
It is now deleting all auth tickets in the rename user service.

chdorner:

nickstenning:
Just to confirm: LGTM.

PR #3900, opened 2016-09-21
Switch to a cookie-based session

chdorner:
~~**This depends on #3899 being merged and possibly deployed, and then needs to be rebased**~~

Now that we don't store the authentication state in the session anymore, we can switch to a cookie-based session which in turn allows us to get rid of `pyramid_redis_sessions` (in the next PR).

### DIFF
--- a/h/session.py
+++ b/h/session.py
@@ -1,5 +1,9 @@
 # -*- coding: utf-8 -*-
 
+from pyramid.session import SignedCookieSessionFactory
+
+from h.security import derive_key
+
 
 def model(request):
     session = {}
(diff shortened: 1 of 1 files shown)


h/session.py
```diff
@@ -55,9 +59,13 @@ def _current_groups(request):
...
-        "redis.sessions.cookie_max_age": 2592000,
-        "redis.sessions.timeout": 604800,
-    })
-    config.include('pyramid_redis_sessions')
+    settings = config.registry.settings
+
+    factory = SignedCookieSessionFactory(
+        secret=derive_key(settings['secret_key'], b'h.session.cookie_secret'),
+        hashalg='sha512',
+        max_age=2592000,
```
nickstenning: The lovely thing about getting rid of this is that we can now *drastically* reduce the session and cookie lifetimes, because it doesn't have anything to do with login.

I'd suggest removing `max_age`, and allowing session cookies to be browser-scoped, and reducing `timeout` to, say, an hour?

chdorner: Yup, I initially just took whatever we had. Changed it now :+1:

h/session.py
```diff
@@ -55,9 +59,12 @@ def _current_groups(request):
...
-    config.add_settings({
-        "redis.sessions.cookie_httponly": True,
-        "redis.sessions.cookie_max_age": 2592000,
-        "redis.sessions.timeout": 604800,
-    })
-    config.include('pyramid_redis_sessions')
+    settings = config.registry.settings
+
+    factory = SignedCookieSessionFactory(
+        secret=derive_key(settings['secret_key'], b'h.session.cookie_secret'),
```
nickstenning: Maybe just add a comment here that by default `derive_key` generates a 64-byte (512-bit) secret, which is the correct length for a SHA512-based HMAC as specified by `hashalg='sha512'`, but if `hashalg` changes, we may need to derive a longer secret.

chdorner: Done

chdorner:

nickstenning:
LGTM.

PR #3901, opened 2016-09-21
Remove pyramid_redis_sessions and redis

chdorner:
**This depends on #3900 being merged and possibly deployed, and then needs to be rebased**

Now that we store the session in a cookie instead of in redis, we can remove the dependency on `pyramid_redis_sessions` and Redis itself.


chdorner:

nickstenning:
LGTM. Will wait for Travis.

PR #3908, opened 2016-09-22
Fix display of document buckets with no title

seanh:
When a document has no title display "Untitled document"
(internationalized) on its document bucket, instead of "None".

### DIFF
--- a/h/templates/activity/search.html.jinja2
+++ b/h/templates/activity/search.html.jinja2
@@ -46,7 +46,13 @@
     <div class="search-result-bucket__domain">
       {{ bucket.domain }}
     </div>
-    <div class="search-result-bucket__title">{{ bucket.title }}</div>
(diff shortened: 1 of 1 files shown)


nickstenning:
I don't mind if it returns `None` or empty string. That wasn't the point. My input was on two independent points:

- The "untitled document" string does not, IMO, belong in a presenter, because then there's no way in a template to distinguish between a document with no title and a document with a title of "untitled document". Presenters should really mostly be being used to simplify and normalise the view of some data, not to replace templates.
- I don't want us to have two different kinds of title, hence my request to move the "title or filename or URL" functionality into the only place it's actually currently used.

In any case, your new PR looks good for now and we can revisit this discussion when we need to make more extensive changes to the activity pages rendering.

nickstenning:
Posted that on the wrong PR, sorry...

PR #3909, opened 2016-09-22
Fix a failing functional test

seanh:
The test was accessing the model directly and creating annotations with no `target_uri`. This can't happen in production because [CreateAnnotationSchema won't allow it](https://github.com/hypothesis/h/blob/9105875105f4d29fa380442ce7dfd73846533547/src/memex/schemas.py#L178) ([update doesn't allow it either](https://github.com/hypothesis/h/blob/9105875105f4d29fa380442ce7dfd73846533547/src/memex/schemas.py#L237)).

Possibly the db column should also be `NOT NULL`

### DIFF
--- a/tests/functional/test_api.py
+++ b/tests/functional/test_api.py
@@ -23,11 +23,9 @@ class TestAPI(object):
 
 
 @pytest.fixture
-def annotation(db_session):
-    from h.models import Annotation
-    ann = Annotation(userid='acct:testuser@localhost',
-                     groupid='__world__',
-                     shared=True)
-    db_session.add(ann)
+def annotation(db_session, factories):
+    ann =  factories.Annotation(userid='acct:testuser@localhost',
+                                groupid='__world__',
+                                shared=True)
     db_session.commit()
     return ann


nickstenning:
Under what circumstances does this test fail? It seems to be working on master (Jenkins runs the functional suite).

seanh:
This fails on master for me locally:

```bash
py.test tests/functional/
```

But curiously `tox -e functional tests/functional/` does not fail.

The difference turns out to be that I have `BOUNCER_URL` set in my shell env, whereas `BOUNCER_URL` is not set in the env that tox runs py.test in. So when the tests are run by tox, `incontext_link()` bails out: <https://github.com/hypothesis/h/blob/master/h/links.py#L19>. But when the tests are run outside of tox, `incontext_link()` goes on to try to do string stuff with the annotation's `None` `target_uri` and crashes: <https://github.com/hypothesis/h/blob/master/h/links.py#L31>.

nickstenning:
To better simulate production, I'd suggest setting the bouncer URL in https://github.com/hypothesis/h/blob/master/tests/functional/conftest.py#L9-L19.

PR #3910, opened 2016-09-22
Strip via.hypothes.is from document bucket titles

seanh:
If someone uses the https://via.hypothes.is/ proxy to annotate
http://www.example.com/ then their annotation can end up saved in our db
as an annotation of https://via.hypothes.is/http://www.example.com/
(which is the URL for http://www.example.com/ proxied through
https://via.hypothes.is/) rather than saved as an annotation of
http://www.example.com/.

On the /search page, don't display via.hypothes.is as the domain for
these https://via.hypothes.is/http://www.example.com/ documents, instead
display www.example.com.

Also don't display the address on the document bucket as
https://via.hypothes.is/http://www.example.com/ but just
http://www.example.com/.

### DIFF
--- a/tests/h/presenters_test.py
+++ b/tests/h/presenters_test.py
@@ -364,6 +364,56 @@ class TestDocumentHTMLPresenter(object):
 
         assert self.presenter(title=None).title == ""
 
+    def test_web_uri_returns_document_web_uri(self):
+        """
+        It just returns Document.web_uri for non-via web_uris.
+
+        If Document.web_uri is a string that doesn't start with
+        https://via.hypothes.is/ then DocumentHTMLPresenter.web_uri should
+        just return Document.web_uri.
+
+        """
+        non_via_uri = 'http://example.com/page'
+
+        assert self.presenter(web_uri=non_via_uri).web_uri == non_via_uri
+
+    def test_web_uri_returns_None_if_document_web_uri_is_None(self):
+        assert self.presenter(web_uri=None).web_uri is None
+
+    @pytest.mark.parametrize('via_url', (
+                             'https://via.hypothes.is',
+                             'https://via.hypothes.is/'))
+    def test_web_uri_returns_via_front_page(self, via_url):
+        """It doesn't strip https://via.hypothes.is if that's the entire URL."""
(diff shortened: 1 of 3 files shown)


seanh:
Needs tests. Needs to be more careful and only strip `https://via.hypothes.is/` if there is an `http://www.example.com/` after it.

dwhly:
THANK YOU!

tests/h/presenters_test.py
```diff
@@ -364,6 +364,54 @@ def test_title_when_no_document_title_no_filename_and_no_uri(self,
...
+
+        assert self.presenter(web_uri=non_via_uri).web_uri == non_via_uri
+
+    def test_web_uri_returns_None_if_document_web_uri_is_None(self):
+        assert self.presenter(web_uri=None).web_uri is None
+
+    def test_web_uri_returns_via_front_page(self):
+        """It doesn't strip https://via.hypothes.is if that's the entire URL."""
+        for via_url in ('https://via.hypothes.is', 'https://via.hypothes.is/'):
+            assert self.presenter(web_uri=via_url).web_uri == via_url
```
chdorner: These should just be parametrized:
```
@pytest.mark.parametrize('via_url', ['https://via.hypothes.is', 'https://via.hypothes.is/'])
def test_web_uri_returns_via_front_page(self, via_url):
    assert self.presenter(web_uri=via_url).web_uri == via_url
```

tests/h/presenters_test.py
```diff
@@ -364,6 +364,54 @@ def test_title_when_no_document_title_no_filename_and_no_uri(self,
...
+
+        If Document.web_uri is https://via.hypothes.is/<path>, for any <path>
+        (whether path is a URL or not), DocumentHTMLPresenter.web_uri just
+        returns <path> with the https://via.hypothes.is/ prefix removed.
+
+        """
+        for path in ('foo', 'http://example.com'):
+            uri = 'https://via.hypothes.is/' + path
+
+            assert self.presenter(web_uri=uri).web_uri == path
```
chdorner: Same as above, these should be parametrized.

seanh: Done!

PR #3911, opened 2016-09-22
Implement inline editing for "Change Email" and "Change Password" forms.

robertknight:
**~~Depends on https://github.com/hypothesis/h/pull/3814~~**
**~~Depends on https://github.com/hypothesis/h/pull/3950~~**

----

This implements inline editing for the accounts forms. Compared to the other forms with inline editing, these have a couple of differences:

1. There are a set of fields which are initially hidden, but then shown when a visible field is focused - eg. in the case of the Change Email form, the Confirm Password field.
2. When any of the fields in these forms are focused, all of the fields should be editable - that is, focusable via keyboard/mouse and raised above the backdrop

The first commit implements support for these features in forms with inline editing. The second commit enables inline editing for the Change Email/Change Password forms and hides the Confirm Password, New Password and Confirm New Password fields by default.

I also found a bug where it was possible for the page to end up in a state with focus ping-ponging continually between the two forms when tabbing from one to the other. This is fixed by the third commit which simplifies and removes asynchronicity from the logic that handles tabbing out of the form.

Compared to the mocks, I made one UX change which is that the password field is labelled "Current Password" rather than just "Password" to avoid ambiguity.

I have a reservation which is that when editing a field, I don't think there is enough contrast between the fields that are not editable _behind_ the modal backdrop and the ones that are active. I'll talk with Conor about this and address as a follow-up.

### DIFF
--- a/h/static/scripts/controllers/form-controller.js
+++ b/h/static/scripts/controllers/form-controller.js
@@ -15,7 +15,7 @@ function shouldAutosubmit(type) {
  * editing the form.
  */
 function isHiddenField(el) {
-  return el.getAttribute('data-hide-until-active');
+  return el.dataset.hideUntilActive;
 }
 
 /**
@@ -35,7 +35,11 @@ class FormController extends Controller {
     this._fields = Array.from(element.querySelectorAll('.js-form-input'))
       .map(el => {
         var parts = findRefs(el);
-        return {container: el, input: parts.formInput};
+        return {
+          container: el,
+          input: parts.formInput,
+          label: parts.label,
+        };
       });
 
     this.on('focus', event => {
@@ -133,7 +137,7 @@ class FormController extends Controller {
     var isEditing = state.editingFields.length > 0;
     setElementState(this.element, {editing: isEditing});
     setElementState(this.refs.formActions, {
-      hidden: !isEditing || shouldAutosubmit(state.editingField.input.type),
+      hidden: !isEditing || shouldAutosubmit(state.editingFields[0].input.type),
       saving: state.saving,
     });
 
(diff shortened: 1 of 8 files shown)


nickstenning:
Rebased.

nickstenning:
screen shot 2016-09-23 at 11 01 31

This looks odd to me:

- What is "Current password"?
- Why is it empty?
- What would clicking into it do?

Also, we're asking for password confirmation in both cases, but it's the second field when changing my email, and the first field when changing my password? That doesn't make any sense.

The way this should work is that by default the user should see this:

screen shot 2016-09-23 at 11 04 05

And when they click into the password field they should see this:

screen shot 2016-09-23 at 11 06 07

robertknight:
[Inline editing for Account settings form](https://trello.com/c/y0319hHD/432-inline-editing-for-account-settings-form)

robertknight:
> Also, we're asking for password confirmation in both cases, but it's the second field when changing my email, and the first field when changing my password?

That matches the previous design of the form, and it does make sense in that it is a more natural flow to ask what the user's password currently is _and then_ what they would like it to be.

I've just looked at Twitter, GitHub and Slack for example and they all arrange the Current Password / New Password / Confirm new password forms in that order.

robertknight:
OK, what I propose to do is:

1. Change the initial display of the Change Password form to display a "Password" field with a dummy label
2. Keep the field ordering once the form is active as it is now

nickstenning:
Needs a rebase.

robertknight:
This now implements a revised UX where, when the Change Password form is inactive, the current password field is labelled "Password" and displays a string of bullets as a placeholder in case the UA doesn't auto-fill the field. Once the user focuses the field, the label changes to "Current password" and the placeholder is removed.

This works as intended on desktop (Chrome, Firefox) but saving changes is not reliable on iOS. I'm currently investigating that.

nickstenning:
Marking this as WIP.

robertknight:
While working on this I encountered a raft of issues related to focus handling across different browsers and also when there are multiple forms on the page, as is the case for the inline accounts forms.

I've split the work to fix that out into a separate PR which should be tested/reviewed/discussed first - https://github.com/hypothesis/h/pull/3950

h/static/scripts/controllers/form-controller.js
```diff
@@ -11,6 +11,14 @@ function shouldAutosubmit(type) {
 }
 
 /**
+ * Return true if a form field should be hidden until the user starts
+ * editing the form.
+ */
+function isHiddenField(el) {
```
sheetaluk: Would it be possible to add an @example (http://usejsdoc.org/tags-example.html) here?

h/static/scripts/controllers/form-controller.js
```diff
@@ -92,33 +108,65 @@ class FormController extends Controller {
...
+
     setElementState(this.refs.formSubmitError, {
       visible: state.submitError.length > 0,
     });
     this.refs.formSubmitErrorMessage.textContent = state.submitError;
+
+    this._updateFields(state);
+  }
+
+  _updateFields(state) {
```
sheetaluk: It would be good to have docstring tags describing the @param here.

h/static/scripts/controllers/form-controller.js
robertknight: While I think `@example` is useful for public APIs, IMO adding it to all internal functions would end up making the code harder to read - because you end up adding many lines that don't actually do anything. I will add a type annotation for `el` though.

robertknight:
Thanks for the review @sheetaluk - I've updated `FormController` with additional API documentation; see the last commit.

sheetaluk:
LGTM

PR #3916, opened 2016-09-23
Remove legacy features view

nickstenning:
This is no longer needed by the client and hasn't been for a long time.


robertknight:
LGTM

PR #3920, opened 2016-09-23
Don't wrap URLs unnecessarily

### LINKED ISSUE #3741: URLs wrapped unnecessarily
In these screenshots URLs wrap onto the next line when they don't need to

download 1

* * *

download 3

seanh:
Don't unnecessarily wrap URLs shown inside document buckets - expand the
div that the URL is wrapped within to fill all of the horizontal space
available to it.

Fixes #3741.

On master:

screenshot from 2016-09-23 16-45-13

screenshot from 2016-09-23 16-44-35

On this branch:

screenshot from 2016-09-23 16-46-01

screenshot from 2016-09-23 16-48-15

### DIFF
--- a/h/static/styles/partials-v2/_search.scss
+++ b/h/static/styles/partials-v2/_search.scss
@@ -106,7 +106,7 @@
 // Card displaying stats about a group of annotations in search results
 .search-bucket-stats {
   @include font-normal;
-  width: 210px;
+  width: 260px;
   margin-left: 30px;
   word-wrap: break-word;
 }


h/static/styles/partials-v2/_search.scss
```diff
@@ -106,7 +106,7 @@
 // Card displaying stats about a group of annotations in search results
 .search-bucket-stats {
   @include font-normal;
-  width: 210px;
+  width: 260px;
```
robertknight: Ideally we want the annotation card to have as much space as it needs up to 500px and then assign the leftover space to the stats box. In testing this again just now it looks like there have been some regressions at smaller screen sizes. For the moment I would just suggest changing this to `max-width` so that the box can be resized smaller than 260px if necessary.

seanh: Ah, I hadn't thought to test this with different screen sizes! Now I wonder if this bit of layout needs a larger refactoring, as you say there seems to be multiple problems with smaller screen sizes.

Here it is with `max-width:260px` at the maximum width:

screenshot from 2016-09-26 11-15-25

But it does start to look quite bad as the window gets smaller (stats retaining its width even though I used `max-width` not `width`), and the actual annotation card shrinking), this is as bad as it gets:

screenshot from 2016-09-26 11-15-53

If the window gets any smaller than that then it hits a breakpoint where the stats move below the annotations. Notice the apparently pointless big empty gap on the left hand side, annotation card pointlessly squished (when there is a big empty gap to the left of it), and misalignment of the annotation card and the title:

screenshot from 2016-09-26 11-16-15

This all gets worse as the window gets smaller:

screenshot from 2016-09-26 11-21-41

seanh:
Closing this in favour of a bigger fix of these document bucket layouts

PR #3921, opened 2016-09-23
Centralise route definitions in h/routes.py

nickstenning:
Our route naming (and even our URL selection) has at times in the past been a little inconsistent and messy. Partly in order to fix this, and partly in order to enable us to move all views into h.views, this commit centralises the definition of the application's routes into a single file.

Hopefully this makes it easier to tell where a URL comes from and easier to ensure consistent naming styles across the application in future.


tests/h/routes_test.py
```diff
@@ -0,0 +1,75 @@
...
+
+from h.routes import includeme
+
+
+def test_includeme():
+    config = Mock(spec_set=['add_route'])
+
+    includeme(config)
+
+    # This may look like a ridiculous test, but the cost of keeping it
```
robertknight: > This may look like a ridiculous test

My thoughts exactly, although I suppose it does at least make sure that the code runs without error.

robertknight:
LGTM

PR #3924, opened 2016-09-26
Fix ragged document bucket title left-alignment

seanh:
On master:

peek 2016-09-26 12-10

On this branch:

hy

### DIFF
--- a/h/static/styles/partials-v2/_search.scss
+++ b/h/static/styles/partials-v2/_search.scss
@@ -54,6 +54,8 @@
 }
 
 .search-result-bucket__domain {
+  flex-grow: 0;
+  flex-shrink: 0;
   width: 120px;
   color: $grey-4;
   margin-right: 10px;


seanh:
This is how this used to work (from <https://github.com/hypothesis/h/commit/32d4f11fb9e2dcd39c79a6db404367cd09440e11>):

peek 2016-09-26 12-38

1. Left-alignment of titles is not ragged
2. There is a breakpoint at which the titles go underneath the domain names

On master, the ragged left-alignment was introduced and the breakpoint was lost (producing the behaviour you see in the first GIF at the top of this page). Both issues were introduced by <https://github.com/hypothesis/h/commit/ec8c98593ea0e4276d4b714949e4356cfc8e5c3e> which restructured the HTML.

seanh:
I think we still have the original CSS code present that was making the document title go under the domain name below a certain screen size: <https://github.com/hypothesis/h/blob/b75b1b20c6197d6212e517025076c7a686653fe3/h/static/styles/partials-v2/_search.scss#L132>, but that CSS has not been working since <https://github.com/hypothesis/h/commit/ec8c98593ea0e4276d4b714949e4356cfc8e5c3e>.

seanh:
I've pushed https://github.com/hypothesis/h/pull/3925 to remove the no-longer-used media query. If we can merge both this PR and https://github.com/hypothesis/h/pull/3925 then I can send another PR to re-implement the media query fully working again.

seanh:
Tested in Chrome and Firefox (both with different window sizes) and on iPhone 5S (Safari, iOS 10.0.2) - looks good everywhere. The above-mentioned broken media query is sorely needed on small screens, will send that in a follow-up PR.

robertknight:
LGTM

PR #3925, opened 2016-09-26
Remove an unused CSS media query

seanh:
What this media query used to do (before <https://github.com/hypothesis/h/commit/ec8c98593ea0e4276d4b714949e4356cfc8e5c3e> broke it):

peek 2016-09-26 12-38

Now on master it only adds 10px of vertical margin, which is pointless:

peek 2016-09-26 13-04

On this branch it's gone completely:

peek 2016-09-26 13-07

A follow-up PR will be needed to re-implement what this media query did but make it work fully again.

### DIFF
--- a/h/static/styles/partials-v2/_search.scss
+++ b/h/static/styles/partials-v2/_search.scss
@@ -127,18 +127,6 @@
   margin-bottom: 20px;
 }
 
-// On large tablets and below, display bucket contents beneath rather than
-// alongside the domain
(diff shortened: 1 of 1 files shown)


seanh:
Gave this a quick test on an iPhone 5S and it seems fine (the layout is a mess on small screens currently but this doesn't seem to make it any worse)

robertknight:
LGTM

PR #3926, opened 2016-09-26
Reimplement document bucket media query

seanh:
Depends on <https://github.com/hypothesis/h/pull/3924> and <https://github.com/hypothesis/h/pull/3925>.

On master the domain in a document bucket header remains to the left of the title and annotations count as the screen shrinks, which looks bad on small screens:

screenshot from 2016-09-26 14-06-33

This is a regression. On this branch the domain moves above the title and annotations count on small screens, as it used to do:

peek 2016-09-26 13-57

screenshot from 2016-09-26 14-00-21

Tested in Chrome, Firefox and Safari on iOS.

### DIFF
--- a/h/static/styles/partials-v2/_search.scss
+++ b/h/static/styles/partials-v2/_search.scss
@@ -62,6 +62,11 @@
   word-wrap: break-word;
 }
 
+.search-result-bucket__title-and-annotations-count {
+  display: flex;
+  flex-grow: 1;
+}
+
 .search-result-bucket__title {
   font-weight: bold;
   flex-grow: 1;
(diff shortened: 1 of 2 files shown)


h/static/styles/partials-v2/_search.scss
```diff
@@ -62,6 +61,12 @@
   word-wrap: break-word;
 }
 
+.search-result-bucket__title-and-annotations-count {
+  display: flex;
+  flex-grow: 1;
+  width: 100%;
```
robertknight: What is the `width` property for here? The width is controlled by flexbox layout of the parent element.

seanh: Good point! It doesn't seem to do anything. I've removed it

robertknight: Actually I was mistaken, this rule _does_ make sense when this element is part of a column, in which case flexbox controls its size along the vertical axis but not the horizontal axis. The result of this is that the annotation count ends up next to the right edge of the title instead of the right edge of the list.

robertknight:
LGTM. We definitely need to get some regression tests in place for this.

PR #3928, opened 2016-09-26
Add a preliminary "create user" API endpoint

nickstenning:
In order to allow third-party accounts clients to furnish their users with grant tokens that allow them to use our API, we need those users to first exist in our database.

This commit adds a relatively straightforward "create user" API endpoint that checks the client's credentials (which must be provided as an HTTP Basic username and password) and creates a user from provided parameters if they validate.

At this stage, I haven't hooked up any useful validation of the payload provided by the HTTP client, but that doesn't matter in production, as we have no AuthClients in the database and so all requests to the endpoint will simply get 403 responses anyway.

In a subsequent pull request I plan to hook up the JSON Schema added in eb4ee79, which will provide proper validation of the request payload and better error messages.

~~_**N.B.** This PR depends on #3923 and will need to be rebased once that has merged._~~

### DIFF
--- a/h/views/api_users.py
+++ b/h/views/api_users.py
@@ -1,0 +1,66 @@
+# -*- coding: utf-8 -*-
+
+from __future__ import unicode_literals
+
+import hmac
+
+import sqlalchemy as sa
+
+from h.auth.util import basic_auth_creds
+from h.exceptions import ClientUnauthorized
+from h.models import AuthClient
+from h.util.view import json_view
+
+
+@json_view(route_name='api.users', request_method='POST')
+def create(request):
+    """
+    Create a user.
+
+    This API endpoint allows authorised clients (those able to provide a valid
+    Client ID and Client Secret) to create users in their authority. These
+    users are created pre-activated, and are unable to log in to the web
+    service directly.
+    """
+    client = _request_client(request)
+    payload = request.json_body
+
+    user_props = {
+        'authority': client.authority,
+        'username': payload['username'],
+        'email': payload['email'],
+    }
+
+    user_signup_service = request.find_service(name='user_signup')
+    user = user_signup_service.signup(require_activation=False, **user_props)
+
+    return {
+        'authority': user.authority,
+        'email': user.email,
+        'userid': user.userid,
+        'username': user.username,
+    }
+
+
+def _request_client(request):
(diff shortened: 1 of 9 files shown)


chdorner:
Rebased

tests/h/exceptions_test.py
```diff
@@ -20,3 +20,15 @@ def test_custom_status_code(self):
         exc = APIError('some message', status_code=418)
 
         assert exc.status_code == 418
+
+
+class TestClientUnauthorized(object):
+    def test_message(self):
+        exc = ClientUnauthorized()
+
+        assert 'credentials are invalid' in exc.message
```
chdorner: Why not test for `exc.message == 'Client credentials are invalid.'`?

h/views/api_users.py
```diff
@@ -0,0 +1,66 @@
...
+    # because the resulting code may be subject to a timing attack.
+    client_id, client_secret = creds
+    try:
+        client = request.db.query(AuthClient).get(client_id)
+    except sa.exc.StatementError:  # client_id is malformed
+        raise ClientUnauthorized()
+    if client is None:
+        raise ClientUnauthorized()
+
+    if not hmac.compare_digest(client.secret, client_secret):
```
chdorner: I was just reading about time attacks. So in order to not leak any information about the existence of a client id, shouldn't we compare the provided secret against some dummy string when we can't find the client in the database? (See the [timing attack example of the Unix `login` program](https://hyp.is/Wbf9VoSXEeaNDV81Wyudww/en.wikipedia.org/wiki/Timing_attack))

tests/h/exceptions_test.py
nickstenning: Because  `client credentials are invalid`, `Client credentials are invalid!` or `The supplied credentials are invalid.` would all be perfectly acceptable alternative error messages. Just trying to minimise the chances that small copy changes break the test.

h/views/api_users.py
nickstenning: There's already timing information leaked by the SQL query about client existence, and there's not an awful lot we can do about that.

What we can do is prevent people from being able to work out the client secret byte by byte through timing analysis. That's what this is for.

PR #3929, opened 2016-09-26
Ensure all annotations have a document (1/3)

chdorner:

We add the new `document_id` column on the `annotation` table, note that at this moment we still allow `NULL` values because we still need to run the data migration. And there needs to be code that sets the `document_id` under normal circumstances.


chdorner:

PR #3850, opened 2016-09-08
Improve logged-out forms error messages

seanh:
I haven't done anything with the error messages that the browser generates, only the ones from our server. I've done the logged-out forms only, there are still more forms available when you login.

screenshot from 2016-09-08 18-35-26
screenshot from 2016-09-08 18-34-55
screenshot from 2016-09-08 18-34-09
screenshot from 2016-09-08 18-32-56
screenshot from 2016-09-08 18-32-28
screenshot from 2016-09-08 18-28-45
screenshot from 2016-09-08 18-28-05
screenshot from 2016-09-08 18-27-02
screenshot from 2016-09-08 18-25-11
screenshot from 2016-09-08 18-24-27
screenshot from 2016-09-08 18-23-54
d73929ee-75ec-11e6-843b-4a1dc65ad7ee
screenshot from 2016-09-08 18-02-08
screenshot from 2016-09-08 18-01-43
screenshot from 2016-09-08 18-01-00
screenshot from 2016-09-08 18-00-33
screenshot from 2016-09-08 17-56-43
screenshot from 2016-09-08 17-55-59
screenshot from 2016-09-08 17-47-00
screenshot from 2016-09-08 17-45-55
screenshot from 2016-09-08 17-43-40
screenshot from 2016-09-08 17-43-16
screenshot from 2016-09-08 17-42-42
screenshot from 2016-09-08 17-42-18
screenshot from 2016-09-08 17-39-32
screenshot from 2016-09-08 17-38-06
screenshot from 2016-09-08 17-36-33
screenshot from 2016-09-08 16-55-21
screenshot from 2016-09-08 16-53-43
screenshot from 2016-09-08 16-52-19
screenshot from 2016-09-08 16-51-03

PR #3930, opened 2016-09-26
Ensure all annotations have a document (2/3)

chdorner:
**~~This depends on #3929 being merged and the migration being run on production.~~**

This pull request makes sure we always set the `Annotation.document_id` when:

* creating annotations (`memex.storage.create_annotation`)
* updating annotations (`memex.storage.update_annotation`)
* merging documents (`memex.document.merge_documents`)

Note: I ran the data migration locally with a recent production dump and it took around 30 minutes to complete.

### DIFF
--- a/src/memex/models/document.py
+++ b/src/memex/models/document.py
@@ -404,9 +404,11 @@ def merge_documents(session, documents, updated=None):
 
 
 def update_document_metadata(session,
-                             annotation,
+                             target_uri,
                              document_meta_dicts,
-                             document_uri_dicts):
+                             document_uri_dicts,
+                             created=None,
+                             updated=None):
     """
     Create and update document metadata from the given annotation.
 
@@ -414,8 +416,8 @@ def update_document_metadata(session,
     and deleted in the database as required by the given annotation and
     document meta and uri dicts.
 
-    :param annotation: the annotation that the document metadata comes from
-    :type annotation: memex.models.Annotation
+    :param target_uri: the target_uri of the annotation from which the document metadata comes from
+    :type target_uri: unicode
 
     :param document_meta_dicts: the document metadata dicts that were derived
         by validation from the "document" dict that the client posted
@@ -425,35 +427,46 @@ def update_document_metadata(session,
         validation from the "document" dict that the client posted
     :type document_uri_dicts: list of dicts
 
+    :param created: Date and time value for the new document records
+    :type created: datetime.datetime
+
+    :param updated: Date and time value for the new document records
+    :type updated: datetime.datetime
+
     """
+    if created is None:
+        created = datetime.utcnow()
+    if updated is None:
+        updated = datetime.utcnow()
+
     documents = Document.find_or_create_by_uris(
         session,
-        annotation.target_uri,
+        target_uri,
         [u['uri'] for u in document_uri_dicts],
-        created=annotation.created,
-        updated=annotation.updated)
+        created=created,
+        updated=updated)
 
     if documents.count() > 1:
         document = merge_documents(session,
                                    documents,
-                                   updated=annotation.updated)
+                                   updated=updated)
     else:
         document = documents.first()
 
-    document.updated = annotation.updated
+    document.updated = updated
 
     for document_uri_dict in document_uri_dicts:
         create_or_update_document_uri(
             session=session,
             document=document,
-            created=annotation.created,
-            updated=annotation.updated,
+            created=created,
+            updated=updated,
             **document_uri_dict)
 
     for document_meta_dict in document_meta_dicts:
         create_or_update_document_meta(
             session=session,
             document=document,
-            created=annotation.created,
-            updated=annotation.updated,
+            created=created,
+            updated=updated,
             **document_meta_dict)
--- a/src/memex/models/document.py
+++ b/src/memex/models/document.py
@@ -12,6 +12,7 @@ from sqlalchemy.ext.hybrid import hybrid_property
 
 from memex.db import Base
 from memex.db import mixins
+from memex.models import Annotation
 from memex.uri import normalize as uri_normalize
 from memex._compat import urlparse
 
@@ -381,6 +382,7 @@ def merge_documents(session, documents, updated=None):
 
     master = documents[0]
     duplicates = documents[1:]
+    duplicate_ids = [doc.id for doc in duplicates]
 
     for doc in duplicates:
         for _ in range(len(doc.document_uris)):
@@ -393,10 +395,14 @@ def merge_documents(session, documents, updated=None):
             m.document = master
             m.updated = updated
 
-        session.delete(doc)
-
     try:
         session.flush()
+        session.query(Annotation) \
+            .filter(Annotation.document_id.in_(duplicate_ids)) \
+            .update({Annotation.document_id: master.id}, synchronize_session='fetch')
+        session.query(Document) \
+            .filter(Document.id.in_(duplicate_ids)) \
+            .delete(synchronize_session='fetch')
     except sa.exc.IntegrityError:
         raise ConcurrentUpdateError('concurrent document merges')
 
--- a/src/memex/models/document.py
+++ b/src/memex/models/document.py
@@ -433,6 +433,8 @@ def update_document_metadata(session,
     :param updated: Date and time value for the new document records
     :type updated: datetime.datetime
 
+    :returns: the matched or created document
+    :rtype: memex.models.Document
     """
     if created is None:
         created = datetime.utcnow()
@@ -470,3 +472,5 @@ def update_document_metadata(session,
             created=created,
             updated=updated,
             **document_meta_dict)
+
+    return document
--- a/src/memex/storage.py
+++ b/src/memex/storage.py
@@ -120,15 +120,18 @@ def create_annotation(request, data):
     annotation = models.Annotation(**data)
     annotation.created = created
     annotation.updated = updated
-    request.db.add(annotation)
 
-    models.update_document_metadata(
+    document = models.update_document_metadata(
         request.db,
         annotation.target_uri,
         document_meta_dicts,
         document_uri_dicts,
         created=created,
         updated=updated)
+    # FIXME: use `document` setter once the relationship changed to use the document_id column
+    annotation.document_id = document.id
+
+    request.db.add(annotation)
 
     return annotation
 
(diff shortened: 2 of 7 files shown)


chdorner:

src/memex/models/document.py
```diff
@@ -433,6 +433,8 @@ def update_document_metadata(session,
     :param updated: Date and time value for the new document records
     :type updated: datetime.datetime
 
+    :returns: the matched or created document
+    :rtype: memex.models.Annotation
```
nickstenning: s/Annotation/Document/

src/memex/storage.py
```diff
@@ -120,15 +120,17 @@ def create_annotation(request, data):
...
+    document = models.update_document_metadata(
         request.db,
         annotation.target_uri,
         document_meta_dicts,
         document_uri_dicts,
         created=created,
         updated=updated)
+    annotation.document_id = document.id
+
+    request.db.add(annotation)
```
nickstenning: There's no need to move the call to `.add`, and it's probably easier to see that it's been called if it stays closest to the call to the `Annotation` constructor.

src/memex/views.py
```diff
@@ -158,6 +158,7 @@ def create(request):
     schema = schemas.CreateAnnotationSchema(request)
     appstruct = schema.validate(_json_payload(request))
     annotation = storage.create_annotation(request, appstruct)
+    request.db.flush()
```
nickstenning: Hm?

src/memex/storage.py
```diff
@@ -120,15 +120,17 @@ def create_annotation(request, data):
...
 
-    models.update_document_metadata(
+    document = models.update_document_metadata(
         request.db,
         annotation.target_uri,
         document_meta_dicts,
         document_uri_dicts,
         created=created,
         updated=updated)
+    annotation.document_id = document.id
```
nickstenning: Maybe add a `FIXME` here to come back and update this to just do `annotation.document = document` once we've updated the relationship property?

src/memex/storage.py
```diff
@@ -173,12 +173,12 @@ def update_annotation(session, id_, data):
...
-            annotation.target_uri,
-            document_meta_dicts,
-            document_uri_dicts,
-            updated=updated)
+        document = models.update_document_metadata(session,
+                                                   annotation.target_uri,
+                                                   document_meta_dicts,
+                                                   document_uri_dicts,
+                                                   updated=updated)
+        annotation.document_id = document.id
```
nickstenning: As in the previous commit, maybe add a `FIXME` here to come back and update this to just do `annotation.document = document` once we've updated the relationship property.

src/memex/models/document.py
```diff
@@ -12,6 +12,7 @@
```
nickstenning: The commit message says `annotation_id` instead of `document_id`.

src/memex/models/document.py
```diff
@@ -393,10 +395,14 @@ def merge_documents(session, documents, updated=None):
             m.document = master
             m.updated = updated
 
-        session.delete(doc)
-
     try:
         session.flush()
+        session.query(Annotation) \
+            .filter(Annotation.document_id.in_(duplicate_ids)) \
```
nickstenning: Again, this can I think be simplified (and `duplicate_ids` removed) when you update the `document` relationship on `Annotation` to just `Annotation.document.in_(duplicates)`. Maybe worth making a note that says as much.

src/memex/storage.py
chdorner: The next PR will require the `document_id` to be set. And the call to `update_document_metadata` invokes at least one flush which then creates the annotation in the database with `document_id=NULL`.

src/memex/views.py
chdorner: So, we used to add the annotation object to the session before we updated the document meta data, which meant that it was automatically being flushed to the DB and thus it would have an id that is not `None`. But since that's not happening anymore (see my comment above), the annotation isn't flushed at this point (there's no need for it at this point, not like for the document related records which we need to query inside `update_document_metadata`). 
We then emit the created event, and [the reply notification subscribers tries to load the annotation from the database](https://github.com/hypothesis/h/blob/master/h/subscribers.py#L47) but it does have a `NULL` id at this point which then blows up.
Would it be nicer to move the flush into the `create_annotation` function? Doesn't make much sense to me since we don't care about it being flushed inside that function or any functions that we call from `create_annotation` (that's how update works for example).

src/memex/models/document.py
chdorner: I can change the filter to use `Annotation.document.in_()`, but I think we still need the `duplicate_ids` because we need to change the way we delete the duplicate documents. We have a foreign-key constraint on `annotation.document_id`, so we can't delete a document (or have it accidentally flushed to the db) while the annotations haven't changed yet. Plus this way we only do one query for changing the `document_id` of all Annotations that belong to one of the duplicate documents.

chdorner: Or rather, I will add the FIXME comment to use the relationship for filtering, but ... (see above for the rest)

src/memex/views.py
chdorner: As discussed in Slack, the annotation had an ID rather accidentally because of how the code was structured (objects added to the session and later some other part of the code flushed the session). I've removed the flush from the view and added it explicitly as an API contract of `create_annotation` (906770ce782f910222f92aa46683bd7799906245).

chdorner:
All comments have been fixed or responded to.

src/memex/models/document.py
```diff
@@ -393,29 +395,35 @@ def merge_documents(session, documents, updated=None):
...
+            .update({Annotation.document_id: master.id}, synchronize_session='fetch')
```
seanh: Should there be a test that `merge_documents()` updates `Annotation.document_id`s? (I don't think I saw one)

tests/memex/storage_test.py
```diff
@@ -407,10 +434,10 @@ def fetch_annotation(patch):
...
 
 @pytest.fixture
-def models(patch):
-    models = patch('memex.storage.models', autospec=False)
-    models.Annotation.return_value.is_reply = False
-    return models
+def patched_annotation(patch):
+    klass = patch('memex.storage.models.Annotation')
+    klass.return_value.is_reply = False
+    return klass
```
chdorner: The issue was that we patched the whole `models` module, so even the `update_document_metadata.assert_called_once_with` was mocked out and always truthy. I think the only fix is to not patch the whole `models` module. Unless you have a better way to make sure that these tests can fail (by commenting out lines in the implementation)?

src/memex/models/document.py
chdorner: That is a great point! I've added a bunch of annotation records in the tests for merge documents, but seem to have missed that there isn't a test that checks that! Will add right now.

tests/memex/storage_test.py
seanh: Normally, this would work:

```python
import mock
m = mock.MagicMock()  # or mock.Mock()
m.update_document_metadata  # <Mock name='mock.update_document_metadata' id='123'>
m.update_document_metadata.assert_called_once_with('foo')  # AssertionError: Expected 'update_document_metadata' to be called once. Called 0 times.
```

But the `models` fixture in this test file does not work like that, as you say `update_document_metadata.assert_called_once_with('anything')` doesn't raise. But I don't understand why it's apparently not behaving like I expect a mock object to.

src/memex/models/document.py
chdorner: There is one now: https://github.com/hypothesis/h/pull/3930/files#diff-d0e476880deffcb29098651d7d4c5fc3R629
Thanks for spotting this!

tests/memex/storage_test.py
seanh: This seems to work for me, on master:

```diff
diff --git a/tests/memex/storage_test.py b/tests/memex/storage_test.py
index ece6cb5..32863fb 100644
--- a/tests/memex/storage_test.py
+++ b/tests/memex/storage_test.py
@@ -94,8 +94,7 @@ class TestExpandURI(object):
         ]
 
 
-@pytest.mark.usefixtures('models',
-                         'update_document_metadata')
+@pytest.mark.usefixtures('models')
 class TestCreateAnnotation(object):
 
     def test_it_fetches_parent_annotation_for_replies(self,
@@ -181,8 +180,8 @@ class TestCreateAnnotation(object):
 
     def test_it_updates_the_document_metadata_from_the_annotation(self,
                                                                   models,
-                                                                  pyramid_request,
-                                                                  update_document_metadata):
+                                                                  pyramid_request):
+        models.update_document_metadata.assert_called_once_with('foo')
         annotation_data = self.annotation_data()
         annotation_data['document']['document_meta_dicts'] = (
             mock.sentinel.document_meta_dicts)
@@ -191,7 +190,7 @@ class TestCreateAnnotation(object):
 
         storage.create_annotation(pyramid_request, annotation_data)
 
-        update_document_metadata.assert_called_once_with(
+        models.update_document_metadata.assert_called_once_with(
             pyramid_request.db,
             models.Annotation.return_value,
             mock.sentinel.document_meta_dicts,
```

The added assertion fails as I'd expect it to. (And if I remove the added assertion then the test passes as I'd expect it to.)

Or in other words: if you've already patched `model` then you don't need to patch `update_document_metadata`, just use the `model.update_document_metadata` of your existing `model` mock.

The reason I'm keen on this is that `patch('memex.storage.models')` patches a top-level name in the module under test (`memex.storage`). _Only_ `memex.storage` will see this mock object in place of `models`, any other code that happens to be run during the test and that also uses `memex.storage.models` will see the real `models` module.

Whereas `patch('memex.storage.models.update_document_metadata')` will replace the `update_document_metadata()` function in the `memex.storage.models` module with a mock, and _any_ code run during the test that is accessing `memex.storage.models.update_document_metadata()` will see a mock instead of the real function, _not_ just the code under test.

I don't think that creates a problem in this case, but it has created difficult to debug issues elsewhere, for example [breaking the test factories](https://github.com/hypothesis/h/pull/3850/commits/bb7fbb03adfede15db249fe09912cbcc8810a9af). So I think it's a good idea to _always patch names on the module under test's own namespace, and never patch names within objects imported into the module under test_.

seanh: I think that what was happening, that was causing `assert_called_once_with('anything')` to always return a truthy mock object and never raise, was that `memex.storage.models` was first being patched and replaced with a mock, and _then_ the `update_document_metadata` fixture was being called and patching `'memex.storage.models.update_document_metadata'`, which was _already_ a mock, so you get a mock of a mock which breaks mock's `assert_called_once_with()`.

chdorner: I like the method you describe in the diff, I replaced my commit with this one :+1:

chdorner:
Merging (after getting the OK from @nickstenning in Slack)

PR #3931, opened 2016-09-26
Ensure all annotations have a document (3/3)

chdorner:
~~**This depends on #3930  being merged and the migration being run on production.**~~

This pull request makes sure that we don't allow `NULL` values in the `annotation.document_id` column. It also changes the `Annotation.document` relationship to use the new `document_id` column instead of the old way of joining over the `document_uri` table.

### DIFF
--- a/tests/common/factories.py
+++ b/tests/common/factories.py
@@ -7,18 +7,29 @@ from __future__ import unicode_literals
 import base64
 import os
 from datetime import (datetime, timedelta)
-import random
 
 import factory
 import faker
-from sqlalchemy import orm
 
 from h import models
-from memex import models as api_models
+
+from ..memex import factories as memex_factories
 
 
 FAKER = faker.Factory.create()
 SESSION = None
+
+Annotation = memex_factories.Annotation
+Document = memex_factories.Document
+DocumentMeta = memex_factories.DocumentMeta
+DocumentURI = memex_factories.DocumentURI
+
+
+def set_session(value):
+    global SESSION
+
+    SESSION = value
+    memex_factories.SESSION = value
 
 
 class ModelFactory(factory.alchemy.SQLAlchemyModelFactory):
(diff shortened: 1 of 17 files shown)


chdorner:

tests/common/factories.py
```diff
@@ -176,14 +177,15 @@ def document_meta_dict(**kwargs):
...
-        api_models.update_document_metadata(
+        document = api_models.update_document_metadata(
             orm.object_session(self),
             self.target_uri,
             document_meta_dicts=document_meta_dicts,
             document_uri_dicts=document_uri_dicts,
             created=self.created,
             updated=self.updated,
         )
+        self.document_id = document.id
```
nickstenning: `self.document = document`?

tests/h/cli/commands/normalize_uris_test.py
```diff
@@ -131,33 +131,27 @@ def test_it_deletes_duplicate_document_meta_objects(req):
...
-                                     _target_uri_normalized='http://example.net')
-
-    req.db.add_all([annotation_1, annotation_2])
-    req.db.flush()
+def test_it_normalizes_annotation_target_uri(req, factories, db_session):
+    annotation_1 = factories.Annotation(userid='luke', target_uri='http://example.org/')
+    annotation_1._target_uri_normalized = 'http://example.org'
+    annotation_2 = factories.Annotation(userid='luke', target_uri='http://example.net/')
+    annotation_2._target_uri_normalized = 'http://example.net'
+    db_session.flush()
```
chdorner: Sure. But this flush here is so that the `_target_uri_normalized` values are getting flushed.
This is the only (and ugly) way of making sure the database fixtures have old normalized URIs, unfortunately.

nickstenning: Oh, sorry. I picked the wrong one to comment on. In this case, yes...

tests/h/cli/commands/normalize_uris_test.py
```diff
@@ -167,15 +161,13 @@ def test_it_reindexes_changed_annotations(req, index):
...
-                                     _target_uri='http://example.net/',
-                                     _target_uri_normalized='http://example.net')
-
-    req.db.add_all([annotation_1, annotation_2])
-    req.db.flush()
+def test_it_skips_reindexing_unaltered_annotations(req, index, factories, db_session):
+    annotation_1 = factories.Annotation(userid='luke',
+                                        target_uri='http://example.org/')
+    annotation_2 = factories.Annotation(userid='luke',
+                                        _target_uri='http://example.net/')
```
nickstenning: Is this correct? In the above examples you changed `_target_uri` to `target_uri`, but not here.

tests/common/factories.py
```diff
@@ -104,6 +104,7 @@ class Meta:  # pylint: disable=no-init, old-style-class
     target_uri = factory.Faker('uri')
     text = factory.Faker('paragraph')
     userid = factory.Faker('user_name')
+    document = factory.SubFactory(Document)
```
nickstenning: Is there a reason we're not importing the `Annotation` and `Document` factories from `tests.memex.factories` rather than duplicating them?

tests/h/cli/commands/normalize_uris_test.py
chdorner: Just came across the annotation factory while fixing another comment, we already do `force_flush` that factory.

nickstenning: :tada: ok, good! For some reason I was convinced we didn't.

tests/h/cli/commands/normalize_uris_test.py
chdorner: Should be the same as above, fixed it.

tests/common/factories.py
chdorner: I was just trying this out and it only works if we can use the same session which might require some work.

nickstenning:
LGTM.

PR #3932, opened 2016-09-26
Fix wrapping of long document bucket titles

### LINKED ISSUE #3927: Long words in document titles break document bucket layout
download

seanh:
If a document bucket had a long word in its title, for example
Pitt_Comp_Literacy_Culture_Distant_Publics_Development_Rhetoric_and_the_Subject_of_Crisis.pdf,
this word was not being wrapped and was overflowing out of the document
bucket and expanding the width of the entire page (especially on mobile).

This commit makes browsers wrap within word if necessary to prevent
overflow.

Fixes #3927

Before:

download

After:

screenshot from 2016-09-26 16-27-03

screenshot from 2016-09-26 16-30-57

Tested on Chrome, Firefox and iOS Safari

### DIFF
--- a/h/static/styles/partials-v2/_search.scss
+++ b/h/static/styles/partials-v2/_search.scss
@@ -64,12 +64,19 @@
 .search-result-bucket__title-and-annotations-count {
   display: flex;
   flex-grow: 1;
+  min-width: 0;  // Without this min-width document titles containing really
+                 // long words won't be wrapped (overflow-wrap: break-word
+                 // won't work), see https://github.com/hypothesis/h/pull/3932
 }
 
 .search-result-bucket__title {
   font-weight: bold;
   flex-grow: 1;
   margin-right: 5px;
+  overflow-wrap: break-word;
+  min-width: 0;  // Without this min-width document titles containing really
+                 // long words won't be wrapped (overflow-wrap: break-word
+                 // won't work), see https://github.com/hypothesis/h/pull/3932
 }
 
 .search-result-bucket__annotations-count {


robertknight:
The downside of this change is that instead of wrapping at word boundaries, all titles can now wrap anywhere within the word.

The usual solution to this involves the [overflow-width](https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-wrap) property. See eg. https://github.com/hypothesis/client/commit/39b4ea8e06faa7a5040818ab5ceca9752b5b2641

robertknight:
@seanh - To answer your question on Slack about `overflow-width` not working, this wasn't obvious to me but you need to specify a `min-width` explicitly, otherwise it defaults to the intrinsic width of the content, which in this case means that the long title does not wrap.

See http://stackoverflow.com/a/36150764/434243 , this minimal test case: https://gist.github.com/robertknight/cec3453c046705bfc9c8e5b8d2153b56 and the definition of the "auto" value for the "min-width" property in the spec: https://drafts.csswg.org/css-flexbox-1/#min-size-auto

robertknight:
I've added a suggestion in https://github.com/hypothesis/h/pull/3937#discussion_r82785795 for how a helper could be added to fix this and document how to achieve the desired effect, since it isn't obvious.

seanh:
I've force-pushed a version that uses `overflow-wrap`, long words are wrapped and long titles composed of short words are wrapped at word boundaries not within words:

peek 2016-10-12 13-28

We have a flexbox within a flexbox, so to get this working I had to add `min-width: 0` to two flex elements. An `@mixin` like the one you linked to could be used in the `.search-result-bucket__title` CSS rule but that alone wouldn't work, you'd still need to add a `min-width: 0` to the `.search-result-bucket__title-and-annotations-count` rule.

It's not actually necessary for `.search-result-bucket__title-and-annotations-count` to have the `overflow-wrap: break-word`, but I could add a mixin with both the `min-width` and `overflow-wrap` rules and then use that mixin in both `.search-result-bucket__title` and `.search-result-bucket__title-and-annotations-count`?

robertknight:
> but I could add a mixin with both the min-width and overflow-wrap rules and then use that
> mixin in both .search-result-bucket__title and .search-result-bucket__title-and-annotations-count?

Hmm, perhaps. This PR is fine as-is for the moment, we can refactor in future.

PR #3934, opened 2016-09-27
Improve input field styling on iOS

### LINKED ISSUE #3740: Focusing the search field on mobile zooms the page
Putting the keyboard focus in the search field causes the browser to unnecessarily zoom in a little, obscuring parts of the page

download 2

robertknight:
* Use `appearance: none` on text input controls to remove the
   drop shadow which iOS adds by default

 * Set font size to 16px on touch devices. This makes the text more
   comfortable to read but also avoids iOS zooming into the field when
   selected

Fixes #3740


nickstenning:
LGTM.

PR #3935, opened 2016-09-27
Set default auth domain to localhost in development

robertknight:
This ensures that the default auth domain is the same for a development
instance regardless of which IP/domain the web service is accessed
through. This enables login to work if the service is accessed via its
IP from a different device on the same network or via ngrok.

The AUTH_DOMAIN environment variable still takes precedence if set.


nickstenning:
I think we may revisit how AUTH_DOMAIN works in future, but for now, LGTM.

PR #3933, opened 2016-09-26
Fix search bucket content alignment

seanh:
Fix the horizontal alignment of search result bucket contents on small
screens.

There was a 130px left margin inside the search result bucket contents
which, on small screens, takes up most of the screen width and squashes
the actual contents into a small space.

Make this margin disappear on small screens.

On master:

peek 2016-09-26 19-41

On this branch:

peek 2016-09-26 19-39

The only problem is that I think when it's this size we might want the stats to be left-aligned with the annotation card, which I can't seem to do with this flexbox approach:

screenshot from 2016-09-26 19-42-50

We may need to find a different approach for this reason?

PR #3936, opened 2016-09-27
Fix search bucket content alignment

seanh:
Fix the horizontal alignment of search result bucket contents on small
screens.

There was a 130px left margin inside the search result bucket contents
which, on small screens, takes up most of the screen width and squashes
the actual contents into a small space.

Make this margin disappear on small screens, and also centre the search
bucket contents within the search result bucket div that contains it,
when the search result bucket contents are not as wide as the search
result bucket div.

peek 2016-09-27 10-41

### DIFF
--- a/h/static/styles/partials-v2/_search.scss
+++ b/h/static/styles/partials-v2/_search.scss
@@ -165,6 +165,7 @@
 
   .search-result-bucket__annotation-cards-container {
     flex-direction: column;
+    padding-left: 0;
   }
 
   .search-bucket-stats {


robertknight:
If the annotations have only a short quote and body, they do not expand to fill the available space:

thin-annotation-card

seanh:
Hmm, damn, I don't know how to fix this.

1. After the small screen media query kicks in, we want that 130px left margin to disappear
2. After the small screen media query kicks in but the document bucket is still wider than an annotation card, we want the annotation card to be centered within the document bucket
3. As the document bucket gets thinner than an annotation card, we want the annotation cards to squish to stay within the document bucket
4. We want all annotation cards to always be the same width as each other, regardless of their contents
5. We want the stats to be below the annotation cards and left-aligned with them

<https://github.com/hypothesis/h/pull/3933> sacrificed 5 (stats were centered, not left-aligned with cards. This PR sacrifices 4 (which seems worse than sacrificing 5). I can't find a way to get all 5 at once?

robertknight:
Rebased

robertknight:
> After the small screen media query kicks in but the document bucket is still wider
>than an annotation card, we want the annotation card to be centered within the document bucket

Testing out a few configurations, I don't think this is desirable. IMO it looks better if everything has the same left alignment once the layout of buckets switches to a column.

In that case, you just need the `padding-left: 0` property and you can drop the `margin-*: auto` properties.

seanh:
> > After the small screen media query kicks in but the document bucket is still wider
> > than an annotation card, we want the annotation card to be centered within the document bucket
> 
> Testing out a few configurations, I don't think this is desirable. IMO it looks better if everything has the
> same left alignment once the layout of buckets switches to a column.
> 
> In that case, you just need the `padding-left: 0` property and you can drop the `margin-*: auto` 
> properties.

Done (and force-pushed). This is how it works on this branch now: 

peek 2016-10-12 14-24

And the problem you found with short annotations not keeping the right width is gone:

peek 2016-10-12 14-28

Tested on mobile (in Chrome dev tools) as well, seems to work fine, the left-alignment looks a little odd on iPad-sized screens imo but I'm happy to go with it.

robertknight:
LGTM

PR #3937, opened 2016-09-27
Don't wrap document bucket addresses unnecessarily

### LINKED ISSUE #3741: URLs wrapped unnecessarily

seanh:
peek 2016-09-27 11-14
(But do be sure to wrap them whenever the URL is too wide for the
document bucket, as the flexbox that is the URL's parent element is
resized.)

Fixes #3741.

### DIFF
--- a/h/static/styles/partials-v2/_search.scss
+++ b/h/static/styles/partials-v2/_search.scss
@@ -115,9 +115,10 @@
 // Card displaying stats about a group of annotations in search results
 .search-bucket-stats {
   @include font-normal;
-  width: 210px;
   margin-left: 30px;
   word-wrap: break-word;
+  min-width: 0;  // Make sure the .search-bucket-stats element's content wraps
+                 // as its flexbox parent element is resized.
 }
 
 .search-bucket-stats__key {


h/static/styles/partials-v2/_search.scss
```diff
@@ -115,9 +115,10 @@
 // Card displaying stats about a group of annotations in search results
 .search-bucket-stats {
   @include font-normal;
-  width: 210px;
   margin-left: 30px;
   word-wrap: break-word;
+  min-width: 0;  // Make sure the .search-bucket-stats element's content wraps
```
robertknight: The comment explains what this line does but not the (IMO) much less obvious question of _how_. Given that we are pairing `word-wrap: break-word` and `min-width: N` declarations together in several elements here, it might be a good idea to extract the two into a mixin.

eg.

```scss
@mixin wrap-long-words(min-width: 0) {
  overflow-wrap: break-word;
  // Set an explicit min width, otherwise it will default to the
  // element's content size. See http://stackoverflow.com/a/36150764/434243
  min-width: $min-width;
}

.search-bucket-stats {
  @include wrap-long-words(0);
}
```

robertknight: Note that `word-wrap` is a widely supported but non-standard alias for `overflow-wrap`. See https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-wrap

seanh:
I'm just going to close this, there are (still) multiple problems with the document bucket bodies with production data as the window size gets smaller, but I don't think the fix here helps (or at least it introduces more problems while it might fix one). Will try again

PR #3938, opened 2016-09-27
Remove old and unused scripts

nickstenning:
These include:

- tools for converting CoffeeScript to JavaScript (there is no more CoffeeScript in this repository)
- a tool for deleting "releases" in Sentry (we no longer create releases for h in Sentry)
- a tool for creating a "release" of h on GitHub (we no longer do this)
- a tool for tagging the "hypothesis/hypothesis" image correctly (this image is autotagged by Jenkins)


robertknight:
LGTM. The decaf script will still be useful in the client repo but I can get that back when I need it.

PR #3939, opened 2016-09-27
Upgrade Raven.js to 3.7.0

sean-roberts:
Upgrading Raven.js on h so we can keep it consistent with Raven.js in the client. 

I manually tested sending errors to the dev project. That worked fine. See: https://sentry.io/hypothesis/dev/issues/163330433/


nickstenning:
LGTM.

PR #3945, opened 2016-09-28
Use consistent names in panels.py

nickstenning:
For consistency with search_url, rename username_link to username_url.

This'll need rebasing after #3944 merges.


chdorner:
Rebased

PR #3946, opened 2016-09-28
Install the same version of Node we use in production

nickstenning:
This should also[1] fix the issues we've been seeing with intermittent build failures on Travis.

[1]: https://github.com/npm/registry/issues/10#issuecomment-216959451


robertknight:
LGTM

PR #3947, opened 2016-09-28
Only update auth tickets expiry column after a threshold

chdorner:
Before we did an `UPDATE ... RETURNING` query, which updated the auth
tickets `expires` column on every single request. This can lead to
locking problems when one auth ticket is used in a lot of concurrent
requests. Which can lead to the `UPDATE ... RETURNING` query having to
wait until the previous request has been finished and its transaction
committed.

The proposed solution in this commit is to split this up into a `SELECT`
query for checking if the provided ticket is valid, then calculate the
new expiry date, but only set it if the current expiry date is older
with some additional threshold. This threshold is currently set at one
minute.

This also has the effect that in the worst case that users can get
logged out after 6 days, 23 hours, and 59 minutes, instead of 7 days.
This difference however is small enough so that we can ignore it.

### DIFF
--- a/h/auth/services.py
+++ b/h/auth/services.py
@@ -7,12 +7,16 @@ import datetime
 from pyramid_authsanity import interfaces
 import sqlalchemy as sa
 from zope import interface
-from zope.sqlalchemy import mark_changed
 
 from h.models import AuthTicket
 from h.auth.util import principals_for_user
 
 TICKET_TTL = datetime.timedelta(days=7)
+
+# We only want to update the `expires` column when the tickets `expires` is at
+# least one minute smaller than the potential new value. This prevents that we
+# update the `expires` column on every single request.
+TICKET_REFRESH_INTERVAL = datetime.timedelta(minutes=1)
 
 
 class AuthTicketNotLoadedError(Exception):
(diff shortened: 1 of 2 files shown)


h/auth/services.py
```diff
@@ -7,13 +7,17 @@
...
 
 from h.models import AuthTicket
 from h.auth.util import principals_for_user
 
 TICKET_TTL = datetime.timedelta(days=7)
 
+# We only want to update the `expires` column when the tickets `expires` is at
+# least one minute smaller than the potential new value. This prevents that we
+# update the `expires` column on every single request.
+NEW_EXPIRY_THRESHOLD = datetime.timedelta(minutes=1)
```
nickstenning: I know it's pernickety, but `NEW_EXPIRY_THRESHOLD` doesn't make a lot of sense to me as a name. Perhaps `TICKET_REFRESH_INTERVAL`?

h/auth/services.py
```diff
@@ -61,26 +65,26 @@ def verify_ticket(self, principal, ticket_id):
...
-        if self._userid:
-            mark_changed(self.session)
-            return True
-
-        return False
+        ticket = self.session.query(AuthTicket) \
+            .filter(AuthTicket.id == ticket_id,
+                    AuthTicket.user_userid == principal,
+                    AuthTicket.expires > sa.func.now()) \
+            .first()
```
nickstenning: Given the PK constraint on `AuthTicket.id` it might be clearer to do `.one_or_none()` here.

h/auth/services.py
```diff
@@ -61,26 +65,26 @@ def verify_ticket(self, principal, ticket_id):
...
+            return False
+
+        self._userid = ticket.user_userid
+
+        # We don't want to update the `expires` column of an auth ticket on
+        # every single request. So here we check that the difference would be
+        # bigger than configured in the `NEW_EXPIRY_THRESHOLD` constant, and
+        # only then update the auth ticket.
+        new_expires = datetime.datetime.utcnow() + TICKET_TTL
+        if (new_expires - ticket.expires) > NEW_EXPIRY_THRESHOLD:
```
nickstenning: Perhaps it would be simpler and clearer to check `if utcnow() - ticket.updated > NEW_EXPIRY_THRESHOLD:`?

chdorner: Sure, I was thinking of doing it like that, but decided against it in case the `updated` column gets updated due to another column getting changed?

chdorner: Done

nickstenning:
LGTM.

PR #3948, opened 2016-09-29
Periodically remove expired auth tokens

chdorner:
Similar to the worker that removes expired auth tickets, this adds a new
worker which runs every hour and deletes expired auth tokens from the
database.

### DIFF
--- a/h/celery.py
+++ b/h/celery.py
@@ -37,7 +37,11 @@ celery.conf.update(
         'delete-expired-authtickets': {
             'task': 'h.auth.worker.delete_expired_auth_tickets',
             'schedule': timedelta(hours=1)
-        }
+        },
+        'delete-expired-tokens': {
+            'task': 'h.auth.worker.delete_expired_tokens',
+            'schedule': timedelta(hours=1)
+        },
     },
     CELERY_ACCEPT_CONTENT=['json'],
     # Enable at-least-once delivery mode. This probably isn't actually what we
--- a/h/models/token.py
+++ b/h/models/token.py
@@ -39,8 +39,8 @@ class Token(Base, mixins.Timestamps):
     #: A NULL value in this column indicates a token that does not expire.
     expires = sqlalchemy.Column(sqlalchemy.DateTime, nullable=True)
 
-    def __init__(self, userid):
-        self.userid = userid
+    def __init__(self, **kwargs):
+        super(Token, self).__init__(**kwargs)
         self.regenerate()
 
     @classmethod
--- a/tests/common/factories.py
+++ b/tests/common/factories.py
@@ -115,3 +115,12 @@ class AuthTicket(ModelFactory):
     @factory.lazy_attribute
     def user_userid(self):
         return self.user.userid
+
+
+class Token(ModelFactory):
+
+    class Meta:  # pylint: disable=no-init, old-style-class
+        model = models.Token
+        force_flush = True
+
+    userid = factory.LazyAttribute(lambda _: ('acct:' + FAKER.user_name() + '@example.com'))
--- a/tests/h/auth/worker_test.py
+++ b/tests/h/auth/worker_test.py
@@ -7,7 +7,7 @@ from datetime import (datetime, timedelta)
 import pytest
 
 from h.auth import worker
-from h.models import AuthTicket
+from h.models import (AuthTicket, Token)
 
 
 @pytest.mark.usefixtures('celery')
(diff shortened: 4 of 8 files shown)


chdorner:
[Periodically clear out expired access tokens](https://trello.com/c/QiTyqaP7/8-periodically-clear-out-expired-access-tokens)

tests/h/auth/worker_test.py
```diff
@@ -34,8 +34,34 @@ def test_it_leaves_valid_tickets(self, db_session, factories):
...
-    def celery(self, patch, db_session):
-        cel = patch('h.auth.worker.celery', autospec=False)
-        cel.request.db = db_session
-        return cel
+
+@pytest.mark.usefixtures('celery')
+class TestDeleteExpiredTokens(object):
+    def test_it_removes_expired_tokens(self, db_session, factories):
+        tokens = [
+            Token('luke', expires=datetime(2014, 5, 6, 7, 8, 9)),
```
nickstenning: You've got the `factories` fixture in the params here, but aren't using it. Probably best to add a factory for tokens and use it here?

tests/h/models/token_test.py
```diff
@@ -0,0 +1,23 @@
...
+
+    def test_init_sets_expires(self):
+        expires = datetime.now()
+        assert Token('luke', expires=expires).expires == expires
+
+    def test_init_sets_expires_to_None_when_missing(self):
+        assert Token('luke').expires is None
+
+    def test_init_generates_value(self):
+        assert Token('lukek').value is not None
```
nickstenning: `assert Token('lukek').value` would cover both these cases.

h/celery.py
```diff
@@ -37,7 +37,11 @@
         'delete-expired-authtickets': {
             'task': 'h.auth.worker.delete_expired_auth_tickets',
             'schedule': timedelta(hours=1)
-        }
+        },
+        'delete-expired-authtokens': {
```
nickstenning: Probably `delete-expired-tokens` for consistency with the model name.

h/models/token.py
```diff
@@ -39,8 +39,9 @@ class Token(Base, mixins.Timestamps):
     #: A NULL value in this column indicates a token that does not expire.
     expires = sqlalchemy.Column(sqlalchemy.DateTime, nullable=True)
 
-    def __init__(self, userid):
+    def __init__(self, userid, expires=None):
```
nickstenning: Hmm. This is fine, but we should probably fix this properly -- i.e. not override the method signature of the default sqlalchemy constructor, and update uses of `Token(userid)` to do `Token(userid=userid)` instead.

chdorner: Sure, but that can happen in a separate PR as this one isn't introducing the override of the constructor.

nickstenning: Absolutely it can (I said as much in the review summary) 👍

chdorner: Had to do it anyway to not have to faff around with the new factory.

tests/common/factories.py
```diff
@@ -115,3 +115,12 @@ class Meta:  # pylint: disable=no-init, old-style-class
...
         return self.user.userid
+
+
+class Token(ModelFactory):
+
+    class Meta:  # pylint: disable=no-init, old-style-class
+        model = models.Token
+        force_flush = True
+
+    userid = factory.LazyAttribute(lambda _: ('acct' + FAKER.user_name() + '@example.com'))
```
nickstenning: Missing colon after `acct` here.

tests/h/auth/worker_test.py
```diff
@@ -34,8 +34,28 @@ def test_it_leaves_valid_tickets(self, db_session, factories):
...
+        worker.delete_expired_tokens()
+        assert db_session.query(Token).count() == 0
+
+    def test_it_leaves_valid_tickets(self, db_session, factories):
+        factories.Token(expires=datetime(2014, 5, 6, 7, 8, 9))
+        factories.Token(expires=(datetime.utcnow() + timedelta(hours=1)))
+
+        assert db_session.query(Token).count() == 2
+        worker.delete_expired_tokens()
+        assert db_session.query(Token).count() == 1
```
nickstenning: Could we add a test that a token with `expires=None` isn't deleted, because those are developer tokens and it would be a bit embarrassing if we blew them all away.

tests/common/factories.py
chdorner: Argh, sorry, should have spotted this. Fixed.

tests/h/auth/worker_test.py
chdorner: That is a brilliant idea! Added :)

PR #3949, opened 2016-09-29
Allow submitting an empty search form

### LINKED ISSUE #3903: Clearing the input of search does not fallback to all results
### Steps to reproduce
1. type a search value
2. submit
3. clear value
4. submit

### Expected behaviour
I would see the completely unfiltered version of the feed

### Actual behaviour
It does not accept allow an empty value to submit

chdorner:
**This depends on #3956 being merged, and will need a rebase after that.**

We handle key events differently when the search suggestions dropdown is
open. It used to always prevent the default event behaviour, even when
no suggestion from the dropdown was selected, this prevented a user from
submitting an empty form, which can be useful to return to a "search
all" query.

This change only prevents the default key event behaviour if a
suggestion is selected, otherwise it lets the event bubble up and thus
submit the form.

Fixes #3903.

### DIFF
--- a/h/static/scripts/tests/controllers/search-bar-controller-test.js
+++ b/h/static/scripts/tests/controllers/search-bar-controller-test.js
@@ -23,9 +23,11 @@ describe('SearchBarController', function () {
   var dropdown;
   var dropdownItems;
   var ctrl;
+  var form;
 
   before(function () {
-    template = '<input data-ref="searchBarInput">' +
+    template = '<form>' +
+      '<input data-ref="searchBarInput">' +
       '<div data-ref="searchBarDropdown">' +
       '<div>Narrow your search</div>' +
       '<ul>' +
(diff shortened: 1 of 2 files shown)


nickstenning:
I've rebased this in @chdorner's absence.

h/static/scripts/tests/controllers/search-bar-controller-test.js
```diff
@@ -23,9 +23,11 @@ describe('SearchBarController', function () {
   var dropdown;
   var dropdownItems;
   var ctrl;
+  var form;
 
   before(function () {
-    template = '<input data-ref="searchBarInput">' +
+    template = '<form>' +
```
robertknight: Protip: You can use template literals (the JS equivalent of Python's `"""strings"""`) as an easier way to write multi-line strings:

```js
var template = `<form>
...
</form>`
```

robertknight:
LGTM

PR #3950, opened 2016-09-29
Rework modal focus handling for forms

robertknight:
The previous method of ensuring that focus remained within the field
being edited and the form's action buttons relied on asynchronously
reverting focus changes after a blur.

This had several problems:

 - There was a brief flash while focus moved to a new "disallowed"
   element and back again

 - If the page contained multiple forms, it was possible for
   the page to get into a state where focus would continually
   ping-pong back between two forms each trying to retain focus

One solution is to use `FocusEvent#relatedTarget` in the `blur` handler,
but this is not supported by Firefox < 48 or IE and `#relatedTarget` is null in some
cases in Safari. Another would be the `focusout` event, but this
is not supported by Firefox either, and `#relatedTarget` may also be null there.

This commit instead adds a helper for implementing modal UI which instead intercepts `focus` events at the document level use a `useCapture` event listener. This method is compatible with all the browsers we support.

### DIFF
--- a/h/static/scripts/base/controller.js
+++ b/h/static/scripts/base/controller.js
@@ -76,6 +76,14 @@ class Controller {
   on(event, listener, useCapture) {
     this.element.addEventListener(event, listener, useCapture);
   }
+
+  /**
+   * Handler which is invoked when the controller's element is about to be
+   * removed.
+   *
+   * This can be used to clean up subscriptions, timeouts etc.
+   */
+  beforeRemove() {}
 }
 
 module.exports = Controller;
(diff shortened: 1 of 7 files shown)


seanh:
@robertknight Can you give steps or videos for how to reproduce the two issues with the current implementation? I can't reproduce them on master, not sure exactly what I'm looking for

robertknight:
@seanh - The second problem does not occur on master because all pages using inline editing currently only have a single form. This fix is a prerequisite for #3911 . The first issue is quite subtle and you may not see it depending on the browser.

On master, the main thing to check is that modal focus handling for forms continues to work:

1. Clicking outside an active field should have no effect
2. Tabbing outside of an active field _that has no unsaved changes_ should hide the form's modal backdrop and re-enable interacting with the rest of the page.
3. When an active field _has unsaved changes_, tab focus should be restricted to that field and the Save/Cancel controls

seanh:
Looks good, sorry for the delay

PR #3952, opened 2016-09-29
Don't generate duplicate form field IDs

nickstenning:
On multi-form pages, we need to pass a counter into the form constructor to avoid generating duplicate form field IDs. See:

    http://docs.pylonsproject.org/projects/deform/en/latest/api.html#deform.Field


robertknight:
LGTM

PR #3956, opened 2016-09-30
Make SearchBarController tests implementation-agnostic

nickstenning:
The existing SearchBarController tests depend heavily on the specific implementation details of the SearchBarController -- that is, they assume the use of the 'input' event as opposed to other alternatives ('keyup', 'change', etc.) that would in principle also work.

Coupling tests to implementation in this way can make refactoring very difficult -- one is forced to change the tests at the same time as the code. Even more importantly, ensuring correct behaviour in response to user-triggered events is substantially more complicated than it would at first appear, because each user event (such as a click) typically triggers a predefined sequence of related events (in this case, mousedown, mouseup, focus, click) which need to be simulated as well.

To pick a pathological example to illustrate this point: adding a 'keyup' event handler to the SearchBarController which destroyed the entire component would clearly break the real controller, but would have had no effect on this test suite.

In order to address these issues, this commit rewrites the SearchBarController test suite to use `syn`, a synthetic event library, which takes care of most of these considerations.

Doing this has exposed at least one bug in the implementation, where the `handleInputOnInput` function calls `updateActiveDropdownItem` to update the "active" dropdown item in response to a keypress, and then immediately calls `maybeOpenOrCloseDropdown`, which (indirectly) calls `clearActiveDropdownItem`.

---

Subsequent commits do the following:

- Use the `focus`/`blur` events rather than `focusin`/`focusout`. This fixes the search bar in Firefox.
- Makes the `openDropdown`/`closeDropdown` functions idempotent, fixing the issue described in the last paragraph above.

### DIFF
--- a/h/static/scripts/tests/controllers/search-bar-controller-test.js
+++ b/h/static/scripts/tests/controllers/search-bar-controller-test.js
@@ -1,8 +1,22 @@
 'use strict';
 
+var syn = require('syn');
+
 var SearchBarController = require('../../controllers/search-bar-controller');
 
-describe('SearchBarController', function () {  
+function center(element) {
+  let rect = element.getBoundingClientRect();
+  return {
+    pageX: rect.left + (rect.width / 2),
+    pageY: rect.top + (rect.height / 2),
+  };
+}
+
+function isActiveItem(element) {
+  return element.classList.contains('js-search-bar-dropdown-menu-item--active');
+}
+
+describe('SearchBarController', function () {
   var template;
   var testEl;
   var input;
@@ -42,6 +56,7 @@ describe('SearchBarController', function () {
   beforeEach(function () {
     testEl = document.createElement('div');
     testEl.innerHTML = template;
+    document.body.appendChild(testEl);
 
     ctrl = new SearchBarController(testEl);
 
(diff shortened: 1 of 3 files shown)


h/static/scripts/tests/controllers/search-bar-controller-test.js
```diff
@@ -51,125 +66,135 @@ describe('SearchBarController', function () {
   });
 
   afterEach(function () {
-    ctrl = null;
+    document.body.removeChild(testEl);
```
robertknight: In PhantomJS 2.x the easiest way to remove an element from the DOM is with [element.remove()](https://developer.mozilla.org/en-US/docs/Web/API/ChildNode/remove)

robertknight:
In general I think decoupling the tests from the implementation - both in terms of event handling and UI structure is a good idea. One issue I see with the implementation is that the mouse move steps for the "highlights the correct facet for a combination of mouseover, up and down arrow keys" test is really slow:

```js
START:
LOG: 'pressed down key', 7
LOG: 'moved mouse', 520
LOG: 'pressed up arrow', 527
LOG: 'moved mouse again', 1039
LOG: 'pressed down key again', 1048
  SearchBarController
    ✔ highlights the correct facet for a combination of mouseover, up and down arrow keys (slow: 1.049 secs)
```

robertknight:
Ah - it looks like `move()` has a duration argument which defaults to 500ms: https://github.com/bitovi/syn/blob/master/src/drag.js#L299

robertknight:
LGTM.

PR #3957, opened 2016-09-30
Turn search query terms into lozenges as user types. 

sheetaluk:
A phrase is determined to be a valid query term if, when the user
types a space character, the phrase is of the one of the following
forms:
    sometext
    "some text with spaces"
    'some text with spaces'
    <keyword>:sometext
    <keyword>:"some text with spaces"
    <keyword>:'some text with spaces'

After a lozenge is created, pressing the enter button when there are no active
dropdown items highlighted, submits the search form with the query parameter
"q" set to the query terms represented by the lozenges and any unfinished text
within the search box joined with a space character.

A lozenge can be deleted by pressing the "X" button on the right of each lozenge.

When a search page loads any query terms already in the hdden search box are turned
into lozenges.

https://trello.com/c/Iv4mWcIs/19-turn-search-query-terms-into-lozenges-as-user-types

Things to consider for a followup PR:
max-width and max-height for each lozenge.
max-width for the lozenges after which we wrap them around to the next line.
deleting lozenges from the keyboard

### DIFF
--- a/h/static/scripts/controllers/lozenge-controller.js
+++ b/h/static/scripts/controllers/lozenge-controller.js
@@ -1,0 +1,37 @@
+'use strict';
+
+var Controller = require('../base/controller');
+
+/**
+ * Create a lozenge with options.content as its content and append it to containerEl.
+ *
+ * A lozenge is made of two parts - the lozenge content and
+ * the 'x' button which when clicked removes the lozenge
+ * from the container and executes the delete callback provided.
+ *
+ * var lozenge = new Lozenge(containerEl, {
+ *   content: content,
+ *   deleteCallback: deleteCallback,
+ * });
+ */
+class LozengeController extends Controller {
+  constructor(containerEl, options) {
+    super(containerEl, options);
+    var lozengeEl = document.createElement('div');
+    lozengeEl.innerHTML =
+      '<div class="js-lozenge__content lozenge__content">'+
+      options.content+
+      '</div>' +
+      '<div class="js-lozenge__close lozenge__close">x</div>';
+    lozengeEl.classList.add('lozenge');
+    lozengeEl.classList.add('js-lozenge');
+    containerEl.appendChild(lozengeEl);
+
+    lozengeEl.querySelector('.js-lozenge__close').addEventListener('mousedown', () => {
+      lozengeEl.remove();
+      options.deleteCallback();
+    });
+  }
+}
+
+module.exports = LozengeController;
--- a/h/static/scripts/controllers/search-bar-controller.js
+++ b/h/static/scripts/controllers/search-bar-controller.js
@@ -2,6 +2,8 @@
 
 var Controller = require('../base/controller');
 var setElementState = require('../util/dom').setElementState;
+var LozengeController = require('./lozenge-controller');
+var SearchTextParser = require('../util/search-text-parser');
 
 /**
  * Controller for the search bar.
@@ -10,11 +12,12 @@ class SearchBarController extends Controller {
   constructor(element) {
     super(element);
 
-    this._input = this.refs.searchBarInput;
     this._dropdown = this.refs.searchBarDropdown;
     this._dropdownItems = Array.from(
       element.querySelectorAll('[data-ref="searchBarDropdownItem"]'));
-
+    this._input = this.refs.searchBarInput;
+    this._inputHidden = this.refs.searchBarInputHidden;
+    this._lozengeContainer = this.refs.searchBarLozenges;
     var getActiveDropdownItem = () => {
       return element.querySelector('.js-search-bar-dropdown-menu-item--active');
     };
@@ -29,16 +32,6 @@ class SearchBarController extends Controller {
     var updateActiveDropdownItem = element => {
       clearActiveDropdownItem();
       element.classList.add('js-search-bar-dropdown-menu-item--active');
-    };
-
-    var selectFacet = facet => {
-      this._input.value = facet;
-
-      closeDropdown();
-
-      setTimeout(function() {
-        this._input.focus();
-      }.bind(this), 0);
     };
 
     var isHidden = element => {
@@ -82,16 +75,22 @@ class SearchBarController extends Controller {
       clearActiveDropdownItem();
       showAllDropdownItems();
       this.setState({open: false});
-      this._input.removeEventListener('keydown', setupListenerKeys);
     };
 
     var openDropdown = () => {
       if (this.state.open) { return; }
       clearActiveDropdownItem();
-
       this.setState({open: true});
+    };
 
-      this._input.addEventListener('keydown', setupListenerKeys);
+    var selectFacet = facet => {
+      this._input.value = facet;
+
+      closeDropdown();
+
+      setTimeout(function() {
+        this._input.focus();
+      }.bind(this), 0);
     };
 
     var getVisibleDropdownItems = () => {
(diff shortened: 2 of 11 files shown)


h/static/scripts/tests/controllers/search-bar-controller-test.js
```diff
@@ -9,9 +10,13 @@ describe('SearchBarController', function () {
   var dropdown;
   var dropdownItems;
   var ctrl;
+  var clock;
 
   before(function () {
-    template = '<input data-ref="searchBarInput">' +
+    template = 
+      '<div class="search-bar__lozenges" data-ref="searchBarLozenges"></div>' +
```
robertknight: You can use template strings now to make writing templates over multiple lines easier. Have a look at `form-controller-test.js`

h/static/scripts/tests/controllers/search-bar-controller-test.js
```diff
@@ -172,4 +180,38 @@ describe('SearchBarController', function () {
...
+    input.dispatchEvent(new Event('focusin'));
+    input.value = testStr;
+    input.dispatchEvent(new Event('input'));
+    input.dispatchEvent(e1);
+    assert.equal(testEl.querySelector('.lozenge-content').textContent, testStr);
+  });
+*/
+
+  unroll('should create lozenge on space', function (fixture) {
+    var e1 = new Event('keydown');
```
robertknight: I would suggest that we take advantage of the `syn` library introduced in https://github.com/hypothesis/h/pull/3956 to make sure that input is done in a more realistic way, and with less code.

h/static/scripts/controllers/search-bar-controller.js
```diff
@@ -137,16 +140,45 @@ class SearchBarController extends Controller {
       }
     };
 
+    var lozengify = content => {
+      var lozengeEl = LozengeHelper.createLozenge(content);
+      this._lozenges.appendChild(lozengeEl);
+      new LozengeController(lozengeEl);
```
sean-roberts: this can be done in the createLozenge function - so that searchbarcontroller does not need to know about it

h/static/styles/partials-v2/_lozenge.scss
```diff
@@ -0,0 +1,23 @@
...
+  display: flex;
+  background-color: $grey-3;
+  color: $grey-6;
+  border-radius: 2px;
+  margin-left: 10px;
+  margin-top: 15px;
+  margin-bottom: 15px;
+}
+
+.lozenge-content {
```
robertknight: Element styles should use `component__element` naming rather than `component-element`, as per the [naming rules](https://github.com/hypothesis/frontend-toolkit/blob/master/docs/css-style-guide.md#naming) in our CSS style guide.

h/static/scripts/controllers/lozenge-controller.js
```diff
@@ -0,0 +1,28 @@
...
+var LozengeHelper = require('../util/lozenge-helper');
+
+class LozengeController extends Controller {
+  constructor(el) {
+    super(el);
+    el.querySelector('.js-lozenge-close').addEventListener('mousedown', () => {
+      // remove the DOM element
+      el.remove();
+      // update input field
+      this._inputHidden = document.querySelector('.js-search-bar').querySelector('.js-search-bar__input-hidden');
```
robertknight: This couples the lozenge controller to the structure of the document. What I would suggest doing instead is passing in callbacks (eg. An `onClick` function) via the `options` argument to the constructor (controllers now take two arguments, see `controller.js`)

h/static/scripts/controllers/search-bar-controller.js
```diff
@@ -156,6 +188,16 @@ class SearchBarController extends Controller {
               querySelector('[data-ref="searchBarDropdownItemTitle"]').
               innerHTML.trim();
```
sean-roberts: for readibility, it would be better to create two variables, the `titleEl` and then the `facet` so that it is not a 4 line assignment

sean-roberts: handleClickOnItem has the same problem - I know this wasn't a change with your PR

h/static/scripts/controllers/search-bar-controller.js
```diff
@@ -203,13 +241,15 @@ class SearchBarController extends Controller {
         this.setState({open: false});
       }
       closeDropdown();
+      this._input.removeEventListener('keydown', setupListenerKeys,
```
sean-roberts: I am not sure what you are expecting this to do. It's not actually going to remove the listener nor invoke setupListenerKeys.

See: https://jsfiddle.net/sean_roberts/swmhkjtm/

h/static/scripts/controllers/lozenge-controller.js
```diff
@@ -0,0 +1,28 @@
...
+      var queryTerms = LozengeHelper.getQueryTerms(this._inputHidden.value);
+
+      queryTerms = queryTerms.map(function(term) {
+        if (term !== el.querySelector('.lozenge-content').textContent) {
+          return term;
+        }
+      });
+      this._inputHidden.value = queryTerms.join(' ');
+      // submit form
+      document.querySelector('.js-search-bar').querySelector('form').submit();
```
sean-roberts: We need to let the search controller know that we have deleted a lozenge but we don't need to directly invoke the submit of the search bar or update the hidden input valu as these are concerns of the searchbarcontroller.

h/static/scripts/controllers/search-bar-controller.js
nickstenning: It does remove the listener. Your JSFiddle doesn't work because you're not passing the same function to `removeEventListener`.

https://jsfiddle.net/swmhkjtm/1/

sean-roberts: You're right. I did that on purpose because I thought that's what the code was doing. When I wrote that I thought setupListenerKeys was a function assigning other event listeners. My oversight. Perhaps renaming this to something like `handleKeys` would help make that more obvious

seanh:
Assigned this to myself, I've been talking to Sheetal about this one so I'll review it when it's ready again

h/static/scripts/controllers/search-bar-controller.js
```diff
@@ -137,16 +140,81 @@ class SearchBarController extends Controller {
...
+
+      //clear input
+      this._input.value = '';
+
+      // add to state
+      this.setState({newQueryParam: content});
+    };
+
+    var lozengifyQueryParams = () => {
+      var queryTerms = SearchTextParser.getQueryTerms(this._inputHidden.value);
```
seanh: Nice!

h/static/scripts/controllers/lozenge-controller.js
```diff
@@ -0,0 +1,25 @@
+'use strict';
+
+var Controller = require('../base/controller');
+
+class LozengeController extends Controller {
```
seanh: I got some feedback on this from Rob and SeanR: It's fine for `SearchBarController` to pass in the parent element to `LozengeController`, though less consistent with the other controllers it is a more convenient API from the calling code's point of view. So you can leave this the way it is if you want, but maybe just rename the `el` argument to something like `containerEl` so it's clear that it's the containing element rather than the lozenge element itself?

h/static/scripts/util/search-text-parser.js
```diff
@@ -0,0 +1,123 @@
...
+ *
+ * @param {String} queryString A string of query terms.
+ *
+ * @returns {Array} queryTerms An array of individual query terms.
+ *
+ * @example
+ * // returns ['foo', 'key:"foo bar"', 'gar']
+ * getQueryTerms('foo key:"foo bar" gar')
+ */
+function getQueryTerms(queryString) {
```
seanh: As we discussed I think something like `getLozengeValues()` might be a better name for this

h/static/scripts/controllers/search-bar-controller.js
```diff
@@ -137,16 +137,113 @@ class SearchBarController extends Controller {
...
+     */
+    var setupLozengeListenerKeys = event => {
+      const SPACE_KEY_CODE = 32;
+
+      var handleSpaceKey = () => {
+        var word = getTrimmedInputValue();
+        if (SearchTextParser.shouldLozengify(word)) {
+          lozengify(word);
+        }
+      };
```
seanh: Optionally this block (get text from the input field; make a lozenge out of that text; and then empty the input field) could move into a `lozengifyInputField()` function

seanh:
This should be fixed in a separate PR (I'll be making GitHub issues for these), but after clicking an `x` to delete a lozenge (which triggers a page reload), it's possible to click another `x` to delete another lozenge _before_ the page reload happens, and then when the reload does happen the second deleted lozenge reappears:

peek 2016-10-18 18-49

(The fix is probably to disable all the other `x`s, as soon as one `x` is clicked.)

seanh:
The search bar has gotten taller on this branch, needs to be fixed before merge I think. Master:

screenshot from 2016-10-18 18-57-01

This branch:

screenshot from 2016-10-18 18-57-27

seanh:
Lozenges wrap onto two lines if they contain two separate words, even if they aren't long. Can be fixed in a separate PR.

screenshot from 2016-10-18 19-02-25

seanh:
It's possible to get into a state where the text you've typed can't be lozengified, like so:

peek 2016-10-18 19-11

This was the only way I could find to break the lozengification rules!

h/static/scripts/tests/controllers/search-bar-controller-test.js
```diff
@@ -221,30 +231,37 @@ describe('SearchBarController', function () {
...
+        done();
+      });
+  });
+
+  it('should create a lozenge when the user presses space', function (done) {
+    syn
+      .click(input)
+      .type('bar')
+      .type('[space]', () => {
+        assert.equal(testEl.querySelectorAll('.js-lozenge__content')[1].textContent, 'bar');
```
seanh: I think a few more new tests for `search-bar-controller.js` are needed, maybe:

* Test that when there is text in the hidden input when the controller is instantiated, then it lozengifies that text and empties the hidden input. Test with various kinds of texts in the hidden input (single lozenge, multiple lozenges, facets, quotes...). For example there is a bug that if I load this URL: <http://localhost:5000/search?q=%22foo+bar>, which contains an unclosed search term, then a lozenge for that search term is created, and weird things happen from there, for example if I just keep submitting the form by pressing `Enter` then it keeps adding `+`s to the query:

peek 2016-10-18 19-18

Or if I type something into the search and either lozengify it or don't, on `Enter` it'll get added to the first, unclosed lozenge:

peek 2016-10-18 19-21

peek 2016-10-18 19-22

* Test what happens when the form is submitted (updating the hidden input value from the lozenges then submitting the form). Test with various states: no lozenges, one lozenge, multiple lozenges, no unlozengified text in the input, some unlozengified text in the input, the text in the input is a single word or is a quoted multi-word or is a quoted multi-word with no closing quote.

   For example, it's possible to get into the "unclosed lozenge" situation above like this: Type `"foo` into the search but don't type the closing quote and don't press `Space`. Then press `Enter`. Now you have an unclosed `"foo` lozenge that will "eat" any other lozenges on `Enter`:

peek 2016-10-18 19-26

* I'd add more tests for creating lozenges when `Space` is pressed, e.g. if there's an unclosed quote it should _not_ create one.

h/static/styles/partials-v2/_nav-bar.scss
```diff
@@ -87,7 +92,7 @@
 @media screen and (max-width: $max-phone-width) {
   // Display search bar full-width underneath nav links on phone-sized screens
   .nav-bar {
-    height: 90px;
+    height: 120px;
```
seanh: Did you mean to change the height here, is that in the design?

sheetaluk: I did because in the mockup it looked like there should be 15px ish of margin betwen the search input box and the lozenges.

h/static/scripts/tests/controllers/search-bar-controller-test.js
seanh: > For example there is a bug that if I load this URL: <http://localhost:5000/search?q=%22foo+bar>, which contains an unclosed search term, then a lozenge for that search term is created, and weird things happen from there, for example if I just keep submitting the form by pressing `Enter` then it keeps adding `+`s to the query:
> 
> peek 2016-10-18 19-18
> 
> Or if I type something into the search and either lozengify it or don't, on `Enter` it'll get added to the first, unclosed lozenge:
> 
> peek 2016-10-18 19-21
> 
> peek 2016-10-18 19-22

This is fixed now

seanh: > For example, it's possible to get into the "unclosed lozenge" situation above like this: Type `"foo` into the search but don't type the closing quote and don't press `Space`. Then press `Enter`. Now you have an unclosed `"foo` lozenge that will "eat" any other lozenges on `Enter`:
> 
> peek 2016-10-18 19-26

This is also fixed now

h/static/scripts/controllers/search-bar-controller.js
```diff
@@ -137,16 +137,112 @@ class SearchBarController extends Controller {
...
 
-    var setupListenerKeys = event => {
+    /**
+     * Updates the hidden input field with the consolidated
+     * search terms from the lozenges.
+     */
+    var updateHiddenInputValue = () => {
+      var newValue = '';
+      Array.from(this._lozenges.querySelectorAll('.js-lozenge__content'), function(loz) {
+        newValue = newValue + loz.textContent + ' ';
```
seanh: > This should be fixed in a separate PR, but I think there's a teeny bug that, after something is lozengified, one more space than necessary is added into the input. Watch me move the cursor back by one space after creating a lozenge:
> 
> peek 2016-10-18 18-29

This is fixed now

seanh:
:dancers:

PR #3959, opened 2016-09-30
Remove unnecessary Pyramid boilerplate

nickstenning:
Calling `config.scan(__name__)` at the root of the `h.views` package guarantees that all the submodules will be included and processed. See:

  http://docs.pylonsproject.org/projects/pyramid/en/latest/api/config.html#pyramid.config.Configurator.scan

As a result, there is no need to explicitly `.include(...)` views modules, and thus they don't need `includeme` functions.

### DIFF
--- a/h/views/__init__.py
+++ b/h/views/__init__.py
@@ -3,9 +3,3 @@
 
 def includeme(config):
     config.scan(__name__)
-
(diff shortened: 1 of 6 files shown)


robertknight:
LGTM

PR #3961, opened 2016-09-30
Move badge views into h.views package

nickstenning:
As part of an effort to move all views into the h.views package, this moves the badge view out of h.badge and into h.views.


robertknight:
LGTM

PR #3962, opened 2016-09-30
Move activity views into h.views package

nickstenning:
As part of an effort to move all views into the h.views package, this moves the activity views out of h.activity and into h.views.


robertknight:
Rebased

robertknight:
LGTM

PR #3963, opened 2016-09-30
Move feeds views into h.views package

nickstenning:
As part of an effort to move all views into the h.views package, this moves the feeds views out of h.feeds and into h.views.

This PR also adds some missing tests for the feeds views.


robertknight:
Rebased

PR #3964, opened 2016-09-30
Move groups views into h.views package

nickstenning:
As part of an effort to move all views into the h.views package, this moves the groups views out of h.groups and into h.views.


robertknight:
LGTM

PR #3966, opened 2016-09-30
Move notification views into h.views package

nickstenning:
As part of an effort to move all views into the h.views package, this moves the notification views out of h.notification and into h.views.


robertknight:
LGTM

PR #3969, opened 2016-10-11
Strip spaces from username forms in admin

### LINKED ISSUE #3627: Rename user field silent error for names with a space at the end.
### Steps to reproduce
1. Go to admin form for user
2. Enter a new username with an extra space at the end
3. Press submit

### Expected behaviour
User's account will be changed to username w/ space removed-- or form should error to complain appropriately

### Actual behaviour
Name appeared to change (because no error was issued), but the new name wasn't created.

Tell us which browser, browser version, and operating system you were using when
you encountered this issue.

### Additional details
When you double click the desired new username in a zendesk request, and Ctrl-C to copy the result, most browsers add a space at the end of the selection (i.e. you were selecting the word plus the space at the end, not just the word).  Pasting that selection into the admin field preserves the errant space.

lenazun:
Strips spaces from usernames used in different files of the admin tools.  Fixes #3627

### DIFF
--- a/h/admin/views/admins.py
+++ b/h/admin/views/admins.py
@@ -24,7 +24,7 @@ def admins_index(request):
              permission='admin_admins')
 def admins_add(request):
     """Make a given user an admin."""
-    username = request.params['add']
+    username = request.params['add'].strip()
     user = models.User.get_by_username(request.db, username)
     if user is None:
         request.session.flash(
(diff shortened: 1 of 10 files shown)


lenazun:
Hola @chdorner! I've added tests to most of these functions. There are two cases in which I'd appreciate your advice.
1. not sure how to test for this on features_test.py.  
2. the function for changing a user's username has no test. Should that should come in a different PR?

chdorner:
These tests are looking great!

1. As discussed in Slack, something similar to https://github.com/hypothesis/h/blob/master/tests/h/admin/views/features_test.py#L99-L112
2. Of course!

PR #3970, opened 2016-10-12
Wrap long words in annotation bodies

seanh:
On the /search page long words (e.g. URLs) in annotation bodies were not
wrapping and were overflowing the annotation card. Wrap these to the
width of the annotation card.

Before:

screenshot from 2016-10-12 14-36-06

After:

screenshot from 2016-10-12 14-38-18

### DIFF
--- a/h/static/styles/partials-v2/_annotation-card.scss
+++ b/h/static/styles/partials-v2/_annotation-card.scss
@@ -57,4 +57,5 @@
 
 .annotation-card__text {
   @include styled-text;
+  overflow-wrap: break-word;
 }


robertknight:
LGTM. I'll let @sean-roberts close this one since he started reviewing it.

judell:
This doesn't completely solve the problem. Example:

image

See https://hypothes.is/users/chunnan?q=uri%3Ahttps%3A%2F%2Fcommonfund.nih.gov%2Fsites%2Fdefault%2Ffiles%2FRM-17-026_CommonsPilotPhase.pdf. 

The problem is that the the class attribute annotation-card__text, which has the overflow styling, is only applied to the first p element in the text.

judell:
@dawa, the example I I showed you:

image

is the same unresolved issue as noted above. I think we need to apply `overflow-wrap:break-word` to the `.search-result-bucket__annotation-cards-container.is-expanded` class instead of to `.annotation-card__text`

`

robertknight:
@judell did you mean to assign Dawa to a pull request that has been closed? If you did, can you file (or find an existing) GitHub issue to track the remaining work so we can keep track of its status.

dawariley:
A secondary 'overflowing a container' issue  found here: https://github.com/hypothesis/h/issues/4566.

PR #3971, opened 2016-10-13
Add RFC7523 compliant OAuth access token endpoint

chdorner:
This endpoint handles the exchange of a JWT authorization grant to an
API access token as described in [RFC7523](https://tools.ietf.org/html/rfc7523).

According to the OAuth 2.0 ([RFC6749](https://tools.ietf.org/html/rfc6749)) spec the token endpoint needs to handle `POST` requests. We currently already have a cookie-based token exchange implemented under `/api/token`, but this one is only called via `GET`. So I made sure that the current implementation only listens to `GET` and started the new implementation in a different views file only listening to `POST`.

Partners will generate JWT tokens for their users and sign them with
their client secret. They then pass that JWT token as a configuration to
the Hypothesis client (that part still needs work), and it will then
exchange the JWT authorization grant for an access token via the `POST
/api/token` endpoint.

RFC7523 [requires the use of the RS256 algorithm](https://hyp.is/1izOopB8EeaKcHPHrD9aYw/tools.ietf.org/html/rfc7523) to sign the JWT token,
the implementation in this commit still uses thi HS256 algorithm. In the
future, authclients would generate a public/private keypair and share
the public part with us.

### DIFF
--- a/h/views/api_auth.py
+++ b/h/views/api_auth.py
@@ -1,0 +1,33 @@
+# -*- coding: utf-8 -*-
+
+from __future__ import unicode_literals
+
+from h.auth.services import TOKEN_TTL
+from h.exceptions import OAuthTokenError
+from h.util.view import json_view
+
+
+@json_view(route_name='token', request_method='POST')
+def access_token(request):
+    svc = request.find_service(name='oauth')
+
+    user = svc.verify_jwt_bearer(
+        assertion=request.POST.get('assertion'),
+        grant_type=request.POST.get('grant_type'))
+    token = svc.create_token(user)
+
+    return {
+        'access_token': token.value,
+        'token_type': 'bearer',
+        'expires_in': TOKEN_TTL.total_seconds(),
+    }
+
+
+@json_view(context=OAuthTokenError)
+def api_token_error(context, request):
+    """Handle an expected/deliberately thrown API exception."""
+    request.response.status_code = context.status_code
+    resp = {'error': context.type}
+    if context.message:
+        resp['error_description'] = context.message
+    return resp
--- a/h/auth/__init__.py
+++ b/h/auth/__init__.py
@@ -65,6 +65,9 @@ def includeme(config):
         WEBSOCKET_POLICY = MultiAuthenticationPolicy([TOKEN_POLICY,
                                                       PROXY_POLICY])
 
+    config.register_service_factory('.services.oauth_service_factory',
+                                    name='oauth')
+
     # Set the default authentication policy. This can be overridden by modules
     # that include this one.
     config.set_authentication_policy(DEFAULT_POLICY)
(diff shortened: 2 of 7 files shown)


chdorner:

h/views/api_auth.py
```diff
@@ -0,0 +1,114 @@
...
+                              description='user with userid described in subject could not be found')
+
+    expires_in = timedelta(hours=1)
+    token = models.Token(userid=user.userid, expires=(utcnow() + expires_in))
+    request.db.add(token)
+
+    return {
+        'access_token': token.value,
+        'token_type': 'bearer',
+        'expires_in': expires_in.seconds,
```
nickstenning: I think you probably want to use `expires_in.total_seconds()` here, because while this will currently work, it'll return zero if you were to change `expires_in` to `timedelta(days=1)`.

h/views/api_auth.py
```diff
@@ -0,0 +1,114 @@
...
+def access_token(request):
+    grant_token = _validate_request(request)
+    userid = _extract_userid(request, grant_token)
+
+    user = request.db.query(models.User).filter_by(userid=userid).one_or_none()
+    if user is None:
+        raise OAuthTokenError('invalid_grant',
+                              description='user with userid described in subject could not be found')
+
+    expires_in = timedelta(hours=1)
```
nickstenning: It's probably worth pulling this out into a module "constant".

h/views/api_auth.py
```diff
@@ -0,0 +1,114 @@
...
+        self.status_code = status_code
+        super(OAuthTokenError, self).__init__(message)
+
+
+@json_view(route_name='token', request_method='POST')
+def access_token(request):
+    grant_token = _validate_request(request)
+    userid = _extract_userid(request, grant_token)
+
+    user = request.db.query(models.User).filter_by(userid=userid).one_or_none()
```
nickstenning: It's probably a good idea to use `UserService.fetch` to retrieve users by userid.

h/views/api_auth.py
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+
+@json_view(context=OAuthTokenError)
+def api_token_error(context, request):
+    """Handle an expected/deliberately thrown API exception."""
+    request.response.status_code = context.status_code
+    resp = {'error': context.message}
+    if context.description:
+        resp['error_description'] = context.description
```
nickstenning: I think the name for this that we use elsewhere is `reason`.

seanh: I think `'error_description'` is mandated by RFC7523 though..

chdorner: Sean is right (RFC6749, OAuth, to be exact).
https://tools.ietf.org/html/rfc6749#appendix-A.8

chdorner:
I've addressed most comments here and shuffled a few things around. I will implement the CORS decorator tomorrow and then create at least two new PR's so we can split this up a bit.

chdorner:
Closing this in favour of a series of smaller pull requests.

PR #3972, opened 2016-10-13
Add opengraph tags to the H homepage

lenazun:
This adds the basic opengraph tags to the H homepage.  I added them directly here because @robertknight advised it this way, given that they're static and they are not too many, plus there's no need to optimize for size in this case.

I also added the new H logo to the static images folder, cause it's about time.


h/templates/home.html.jinja2
```diff
@@ -11,6 +11,10 @@
   <link rel="chrome-webstore-item"
         href="{{ chrome_extension_link }}">
   <link href="{{ rss_feed_url }}" rel="alternate" type="application/rss+xml">
+  <meta property="og:url"                content="{{ base_url }}" />
+  <meta property="og:title"              content="Hypothes.is | Annotate the web" />
```
lenazun: True! thank you for pointing that out :)

PR #3973, opened 2016-10-13
Fix document bucket responsive layout bugs

seanh:
This pull request fixes a number of issues with the layout of document bucket bodies on different screen sizes.

Fixes https://github.com/hypothesis/h/issues/3741

# Before

Stats overrun the document bucket (this also creates a horizontal scrollbar):

screenshot from 2016-10-13 19-02-34

In other cases the stats retain their width and the annotation card is shrunk thinner than we want it to be (it appears to depend on the contents of the annotation cards and/or stats, whether this or the above happens):

screenshot from 2016-10-13 19-03-02

Stats are wrapped unnecessarily:

screenshot from 2016-10-13 19-02-00

screenshot from 2016-10-13 19-01-16

# After

peek 2016-10-13 18-56

peek 2016-10-13 18-55

screenshot from 2016-10-13 18-54-09

screenshot from 2016-10-13 18-53-47

screenshot from 2016-10-13 18-53-19

screenshot from 2016-10-13 18-53-01

screenshot from 2016-10-13 18-52-37

screenshot from 2016-10-13 18-52-19

screenshot from 2016-10-13 18-51-04

screenshot from 2016-10-13 18-50-42

### DIFF
--- a/h/static/styles/partials-v2/_search.scss
+++ b/h/static/styles/partials-v2/_search.scss
@@ -140,6 +140,11 @@
   margin-bottom: 20px;
 }
 
+.search-bucket-stats__address,
+.search-bucket-stats__username {
+  word-break: break-all;
+}
+
 // On large tablets and below, display bucket titles beneath rather than
 // alongside the domain name.
 @media screen and (max-width: $tablet-width + 100px) {
--- a/h/static/styles/partials-v2/_search.scss
+++ b/h/static/styles/partials-v2/_search.scss
@@ -182,6 +182,10 @@
     flex-direction: column;
   }
 
+  .search-result-bucket__annotation-cards {
+    width: 100%;
+  }
+
   .search-bucket-stats {
     margin-left: 5px;
     margin-top: 10px;
--- a/h/static/styles/partials-v2/_search.scss
+++ b/h/static/styles/partials-v2/_search.scss
@@ -137,8 +137,8 @@
   margin-bottom: 20px;
 }
 
-// On large tablets and below, display bucket contents beneath rather than
-// alongside the domain
+// On large tablets and below, display bucket titles beneath rather than
+// alongside the domain name.
 @media screen and (max-width: $tablet-width + 100px) {
     .search-result-bucket__header {
       flex-direction: column;
--- a/h/static/styles/partials-v2/_search.scss
+++ b/h/static/styles/partials-v2/_search.scss
@@ -124,7 +124,6 @@
 // Card displaying stats about a group of annotations in search results
 .search-bucket-stats {
   @include font-normal;
-  width: 210px;
   margin-left: 30px;
   word-wrap: break-word;
 }
--- a/h/static/styles/partials-v2/_search.scss
+++ b/h/static/styles/partials-v2/_search.scss
@@ -160,6 +160,14 @@
     }
 }
 
+// On large tablets and below left-align annotation cards in order to make more
+// space for the stats to the right of the annotation card.
+@media screen and (max-width: $tablet-width + 150px) {
+  .search-result-bucket__annotation-cards-container {
+    padding-left: 0;
+  }
+}
+
 // On normal tablets and below, display annotation stats below annotation list
 @media screen and (max-width: $tablet-width) {
 
@@ -172,7 +180,6 @@
 
   .search-result-bucket__annotation-cards-container {
     flex-direction: column;
-    padding-left: 0;
   }
 
   .search-bucket-stats {
--- a/h/static/styles/partials-v2/_search.scss
+++ b/h/static/styles/partials-v2/_search.scss
@@ -116,6 +116,7 @@
 .search-result-bucket__annotation-cards {
   flex-grow: 1;
   max-width: 500px;
+  flex-shrink: 0;
   padding-left: 0;
   padding-right: 0;
 }
--- a/h/static/styles/partials-v2/_search.scss
+++ b/h/static/styles/partials-v2/_search.scss
@@ -126,6 +126,9 @@
   @include font-normal;
   margin-left: 30px;
   word-wrap: break-word;
+  min-width: 0;  // Without this min-width stats containing really
+                 // long words won't be wrapped (word-wrap: break-word won't
+                 // work).
 }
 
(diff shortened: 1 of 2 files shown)


h/static/styles/partials-v2/_search.scss
```diff
@@ -100,9 +100,6 @@
 }
 
 .search-result-bucket__content {
-  display: flex;
-  flex-direction: column;
-  flex-grow: 1;
```
seanh: `search-result-bucket__content` only contains one element (`search-result-bucket__annotation-cards-container`) so this flex layout seems to be unnecessary. I removed it (in a separate commit).

h/static/styles/partials-v2/_search.scss
```diff
@@ -117,18 +114,20 @@
 }
 
 .search-result-bucket__annotation-cards {
-  flex-grow: 1;
   max-width: 500px;
+  flex-shrink: 0;
```
seanh: Stop the annotation cards from shrinking below 500px (shrink the stats instead).

h/static/styles/partials-v2/_search.scss
```diff
@@ -117,18 +114,20 @@
 }
 
 .search-result-bucket__annotation-cards {
-  flex-grow: 1;
```
seanh: I'm not sure why this was here, we don't want annotation cards to grow wider than 500px

h/static/styles/partials-v2/_search.scss
```diff
@@ -117,18 +114,20 @@
...
   max-width: 500px;
+  flex-shrink: 0;
   padding-left: 0;
   padding-right: 0;
 }
 
 // Card displaying stats about a group of annotations in search results
 .search-bucket-stats {
   @include font-normal;
-  width: 210px;
```
seanh: This is what was causing the unnecessary wrapping of stats, let flexbox expand the element to fill the width available instead

h/static/styles/partials-v2/_search.scss
```diff
@@ -117,18 +114,20 @@
...
   margin-left: 30px;
   word-wrap: break-word;
+  min-width: 0;  // Without this min-width stats containing really
+                 // long words won't be wrapped (word-wrap: break-word won't
+                 // work).
```
seanh: `word-wrap: break-word` was already here but it wasn't working because `min-width: 0` is needed as well.

h/static/styles/partials-v2/_search.scss
```diff
@@ -172,7 +179,10 @@
 
   .search-result-bucket__annotation-cards-container {
     flex-direction: column;
-    padding-left: 0;
+  }
+
+  .search-result-bucket__annotation-cards {
+    width: 100%;
```
seanh: Shrink annotation card on small (<500px) screens from 500px (wider than the screen) to 100% of the space available (ie as wide as the screen minus margins and padding)

h/static/styles/partials-v2/_search.scss
```diff
@@ -117,18 +114,20 @@
...
   padding-left: 0;
   padding-right: 0;
 }
 
 // Card displaying stats about a group of annotations in search results
 .search-bucket-stats {
   @include font-normal;
-  width: 210px;
   margin-left: 30px;
   word-wrap: break-word;
```
seanh: I've left this as `word-wrap` because `overflow-wrap` [doesn't have great browser support](http://caniuse.com/#feat=wordwrap) (maybe I should do a separate PR to replace all `overflow-wrap`s with `word-wrap`? We have a few)

robertknight: My preferred solution would be to let our CSS building tooling deal with mapping standardized property names to non-standard ones, as we do with properties that need vendor prefixing. Looks like this was requested upstream in autoprefixer but not actually implemented: https://github.com/postcss/autoprefixer/issues/558

A PR that either converts all uses of `overflow-wrap` to `word-wrap` would be OK with me though - since realistically support for the legacy name is never going to be removed from the web platform.

sean-roberts: you can, but I really don't think it's worth the time (little as it may be)

h/static/styles/partials-v2/_search.scss
sean-roberts: That is needed so that it will fill in the area when the card contents do not push it to the full width. So if you have a card with only two words making up the annotation, it will leave the card as a small box.

h/static/styles/partials-v2/_search.scss
sean-roberts: I would suggest using `word-break: break-all` in addition to break-word. It will not try to calculate good endings and simple roll the text over to the next row, middle of word or not. Doing that has the added benefit of consistency through the stats per section and it won't have the awkward short row for a url like `alsdfjalskdfjlaskdjflaskdfjalskdfj-b-laksdfjlaksjdfoaisdjflaksdjf` where b would be left by itself. That said, I know it applies to things like author names as well. Of which, I imagine is far less likely to be long compared to URIs. I'll leave that up to you

seanh: Good idea. `word-wrap: break-all` seems like a good idea for URIs, for example without:

screenshot from 2016-10-18 11-08-06

With:

screenshot from 2016-10-18 11-14-12

I'd say that we should also use `break-all` for usernames since usernames can't contain spaces anyway (although as you say this is far less likely to ever actually make a difference for usernames).

Tags _can_ contain spaces so I think stick with `word-break: normal` for those (though I haven't extensively checked to see how this works with production usage of tags, which probably various widely between different groups of users).

I'll add a commit for this...

h/static/styles/partials-v2/_search.scss
seanh: You're right (god this is tricky), previously I found that on _some_ production data the annotation card would grow _too_ big (beyond its `500px`) and squash the stats, but now (after re-adding this `flex-grow: 1`) I don't seem to be able to reproduce that anymore (maybe it's because we have a `max-width: 500px` now). Anyway, I'll remove this deletion

seanh:
@sean-roberts think that's all the review comments responded to

PR #3977, opened 2016-10-19
Limit current /api/token view to GET only (OAuth Token 1/n)

chdorner:
The current token endpoint is only called via [HTTP GET]. Specifying
this in the view config will allow us to implement the new JWT OAuth
token exchange view as a separate view and after some time remove the
old cookie-based token exchange handling.

[HTTP GET]: https://github.com/hypothesis/client/blob/22181c80ce08319ac646eb35a82873dc85992cdb/h/static/scripts/auth.js#L40


chdorner:

PR #3978, opened 2016-10-19
Remove Token.userid uniqueness constraint (OAuth Token 2/n)

chdorner:
<del>**This will need rebasing once #3977 is merged.**</del> (rebased)

We will have OAuth tokens soon which means that a user can and
definitely will have more than one token at a time.

Also ensure that `Token.get_by_userid` always returns the most recent
token.

_The migration and code changes can be deployed at the same time here since we're just removing the uniqueness constraint._

### DIFF
--- a/h/models/token.py
+++ b/h/models/token.py
@@ -28,8 +28,7 @@ class Token(Base, mixins.Timestamps):
                            primary_key=True)
 
     userid = sqlalchemy.Column(sqlalchemy.UnicodeText(),
-                               nullable=False,
-                               unique=True)
+                               nullable=False)
 
     value = sqlalchemy.Column(sqlalchemy.UnicodeText(),
                               nullable=False,
@@ -45,7 +44,10 @@ class Token(Base, mixins.Timestamps):
 
     @classmethod
     def get_by_userid(cls, session, userid):
-        return session.query(cls).filter(cls.userid == userid).first()
+        return (session.query(cls)
+                .filter_by(userid=userid)
+                .order_by(cls.created.desc())
+                .first())
 
     def regenerate(self):
         self.value = self.prefix + _token()
--- a/tests/h/models/token_test.py
+++ b/tests/h/models/token_test.py
@@ -8,3 +8,15 @@ from h.models import Token
 class TestToken(object):
     def test_init_generates_value(self):
         assert Token().value
+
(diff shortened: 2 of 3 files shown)


chdorner:

seanh:
Rebased

h/models/token.py
```diff
@@ -45,7 +44,10 @@ def __init__(self, **kwargs):
 
     @classmethod
     def get_by_userid(cls, session, userid):
-        return session.query(cls).filter(cls.userid == userid).first()
+        return session.query(cls) \
+            .filter_by(userid=userid) \
+            .order_by(cls.created.desc()) \
+            .first()
```
seanh: By the way I think this style might be better:

```python
return (session.query(cls)
  .filter_by(userid=userid)
  .order_by(cls.created.desc())
  .first())
```

because PEP8 says to avoid \'s for line continuation whenever possible. It's less characters (just an extra `(` and `)` instead of 3+ `\`'s), and it's easier to edit as you don't have to add a new `\` if you add a new line.

tests/h/models/token_test.py
```diff
@@ -8,3 +8,17 @@
...
     def test_init_generates_value(self):
         assert Token().value
+
+    def test_get_by_userid_filters_by_userid(self, db_session, factories):
+        token_1 = factories.Token()
+        token_2 = factories.Token()
+
+        assert token_1 != token_2
+
+        Token.get_by_userid(db_session, token_2.userid) == token_2
```
seanh: Wait, should this be an `assert ...`?

tests/h/models/token_test.py
```diff
@@ -8,3 +8,17 @@
 class TestToken(object):
     def test_init_generates_value(self):
         assert Token().value
+
+    def test_get_by_userid_filters_by_userid(self, db_session, factories):
+        token_1 = factories.Token()
+        token_2 = factories.Token()
+
+        assert token_1 != token_2
```
seanh: So this isn't what you're actually testing, but you're just checking that the two tokens are different because they have to be for the test below to work, right?

Is it possible that the `Token()` factory could, rarely, return two equal tokens and this test would fail?

I think you can ensure that they're always different by doing:

```python
token_1 = factories.Token(userid='user1')
token_2 = factories.Token(userid='user2')
```

chdorner:
Addressed the issues with the tests and the style issue.

PR #3979, opened 2016-10-19
Add token.authclient_id migration (OAuth Token 3/n)

chdorner:
<del>**This will need rebasing once #3978 is merged.**</del> (rebased)

This way we can track which authclient is responsible for which tokens,
how many active users an authclient has, etc.

It is also useful for the future when everybody will be able to create
authclients, because when an authclient gets deleted all tokens need to
be removed to make sure that they can't be used anymore for making
requests.

_This is only the migration as the code depending on this new column cannot be deployed at the same time._


chdorner:

PR #3980, opened 2016-10-19
Add Token.authclient and refactor dev tokens (OAuth Token 4/n)

chdorner:
**<del>This will need rebasing once #3979 is merged</del>** (rebased) - **this should also not be merged before #3979 is deployed and the migration has been run!**

This adds the relationship between tokens and authclients. At the same time, this will change how we identify a developer token to when the `authclient_id` column is `NULL`. The second commit in this pull request reflects that change.

### DIFF
--- a/h/models/token.py
+++ b/h/models/token.py
@@ -53,9 +53,9 @@ class Token(Base, mixins.Timestamps):
         self.regenerate()
 
     @classmethod
-    def get_by_userid(cls, session, userid):
+    def get_dev_token_by_userid(cls, session, userid):
         return (session.query(cls)
-                .filter_by(userid=userid)
+                .filter_by(userid=userid, authclient=None)
                 .order_by(cls.created.desc())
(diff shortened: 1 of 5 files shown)


chdorner:

chdorner:

h/models/token.py
```diff
@@ -38,6 +39,15 @@ class Token(Base, mixins.Timestamps):
     #: A NULL value in this column indicates a token that does not expire.
     expires = sqlalchemy.Column(sqlalchemy.DateTime, nullable=True)
 
+    _authclient_id = sqlalchemy.Column('authclient_id',
+                                       postgresql.UUID(),
```
seanh: Do you think there's any chance that we might want to use these authclient IDs in URLs one day / do you think we should use `URLSafeUUID` here? (I already missed this in the DB migration in the previous PR anyway...)

PR #3981, opened 2016-10-19
Add service for exchanging RFC7523 authorization grants (OAuth Token 5/n)

chdorner:
**<del>This will need rebasing once #3980 is merged</del>.** (rebased)

This service will receive an assertion JWT bearer token and a grant_type
and will exchange that for an API token.

Partners will generate JWT tokens for their users and sign them with
their client secret. They then pass that JWT token as a configuration to
the Hypothesis client (that part still needs work), and it will then
exchange the JWT authorization grant for an access token via the `POST
/api/token` endpoint, which will be implemented in the next commit.

It adheres to the behaviour described in [RFC7523], except for the
hashing algorithm. RFC7523 requires the use of the RS256 algorithm to
sign the JWT token, the implementation in this commit still uses the
HS256 algorithm. In the future, authclients would generate a
public/private keypair and share the public part with us.

[RFC7523]: https://tools.ietf.org/html/rfc7523

### DIFF
--- a/tests/h/auth/services_test.py
+++ b/tests/h/auth/services_test.py
@@ -3,13 +3,17 @@
 from __future__ import unicode_literals
 
 from datetime import (datetime, timedelta)
+from calendar import timegm
 
+import jwt
 import mock
 import pytest
 
 from h import models
 from h.accounts.services import UserService
 from h.auth import services
+from h.exceptions import OAuthTokenError
+from h._compat import text_type
 
 
 class TestAuthTicketService(object):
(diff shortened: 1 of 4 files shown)


chdorner:

tests/h/auth/services_test.py
```diff
@@ -206,9 +210,308 @@ def test_it_provides_user_service(self, pyramid_request, user_service):
         assert svc.usersvc == user_service
 
 
+class TestOAuthServiceVerifyJWTBearer(object):
+    """ Tests for ``OAuthService.verify_jwt_bearer`` """
```
seanh: There's a couple of prospector / linter warning about this docstring too (and the one below)

tests/h/auth/services_test.py
```diff
@@ -206,9 +210,308 @@ def test_it_provides_user_service(self, pyramid_request, user_service):
...
+    def test_it_provides_user_service(self, pyramid_request, user_service):
+        svc = services.oauth_service_factory(None, pyramid_request)
+        assert svc.usersvc == user_service
+
+    def test_it_provides_request_domain(self, pyramid_request):
+        pyramid_request.domain = 'example.org'
+        svc = services.oauth_service_factory(None, pyramid_request)
+        assert svc.domain == 'example.org'
+
+
```
seanh: Nice, really simple, clear and _detailed_ tests

PR #3982, opened 2016-10-19
Refactor h.auth.services code (OAuth Token 6/n)

chdorner:
**<del>This will need rebasing once #3981 is merged</del>.** (rebased)

These two commits refactor some of the code in `h.auth.services` to make testing of the `AuthTicketService` a bit nicer (due to a module level `utcnow` introduced with the `OAuthService`).

PR #3983, opened 2016-10-19
Add new POST /api/token endpoint (OAuth Token 7/n)

chdorner:
**<del>This will need rebasing once #3982 is merged</del>.** (rebased)

This implements the `POST` endpoint which exchanges an RFC7523 JWT grant token for an OAuth token by using the `OAuthService` that we introduced in #3981.

_This endpoint will need to support CORS as @nickstenning mentioned in #3971. I will do that in a follow up pull request._

### DIFF
--- a/h/views/api_auth.py
+++ b/h/views/api_auth.py
@@ -1,0 +1,33 @@
+# -*- coding: utf-8 -*-
+
+from __future__ import unicode_literals
+
+from h.auth.services import TOKEN_TTL
+from h.exceptions import OAuthTokenError
+from h.util.view import json_view
+
+
+@json_view(route_name='token', request_method='POST')
+def access_token(request):
+    svc = request.find_service(name='oauth')
+
+    user, authclient = svc.verify_jwt_bearer(
+        assertion=request.POST.get('assertion'),
+        grant_type=request.POST.get('grant_type'))
+    token = svc.create_token(user, authclient)
+
+    return {
+        'access_token': token.value,
+        'token_type': 'bearer',
+        'expires_in': TOKEN_TTL.total_seconds(),
+    }
+
+
+@json_view(context=OAuthTokenError)
+def api_token_error(context, request):
+    """Handle an expected/deliberately thrown API exception."""
+    request.response.status_code = context.status_code
+    resp = {'error': context.type}
+    if context.message:
+        resp['error_description'] = context.message
+    return resp
--- a/tests/h/views/api_auth_test.py
+++ b/tests/h/views/api_auth_test.py
@@ -1,0 +1,80 @@
+# -*- coding: utf-8 -*-
+
+from __future__ import unicode_literals
+
+import mock
+import pytest
+
+from h import models
+from h.accounts.services import user_service_factory
+from h.auth.services import oauth_service_factory, TOKEN_TTL
+from h.exceptions import OAuthTokenError
+from h.views import api_auth as views
+
+
+@pytest.mark.usefixtures('user_service', 'oauth_service')
+class TestAccessToken(object):
+    def test_it_verifies_the_jwt_bearer(self, pyramid_request, oauth_service):
+        pyramid_request.POST = {'assertion': 'the-assertion', 'grant_type': 'the-grant-type'}
+
+        views.access_token(pyramid_request)
+
+        oauth_service.verify_jwt_bearer.assert_called_once_with(
+            assertion='the-assertion', grant_type='the-grant-type'
+        )
+
+    def test_it_creates_a_token(self, pyramid_request, oauth_service):
+        views.access_token(pyramid_request)
+
+        oauth_service.create_token.assert_called_once_with(
+            mock.sentinel.user, mock.sentinel.authclient)
+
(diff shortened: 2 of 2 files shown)


chdorner:

seanh:
Usage:

```python
In [1]: # Create an authclient in the h database.

In [2]: session.add(models.AuthClient(authority=u'seanh', secret=u'mysecret'))

In [3]: request.tm.commit()

In [4]: # Now someone with the authclient's secret can make an HTTP request to get an auth ticket.

In [5]: import requests

In [6]: # You need the dev token of an existing h account in order to make the request.

In [7]: dev_token = '6879-24679686935ce83fed6efcf7c98c3d45'

In [8]: import jwt

In [9]: auth_client_id = session.query(models.AuthClient).first().id

In [10]: requests.post('http://localhost:5000/api/token', headers={'Authorization': 'Bearer ' + dev_token}, data={'grant_type': 'urn:ietf:params:oauth:grant-type:jwt-bearer
    ...: ', 'assertion': jwt.encode({'iss': auth_client_id, 'aud': 'localhost', 'sub': 'acct:seanh@localhost'}, 'mysecret')}).content
Out[10]: '{"access_token": "6879-6ef0a8c4f832c152f2b40975d3978d23", "token_type": "bearer", "expires_in": 3600.0}'

# This access token can now be used to access the API, for example to create an annotation:

In [15]: requests.post('http://localhost:5000/api/annotations', headers={'Authorization': 'Bearer 6879-6ef0a8c4f832c152f2b40975d3978d23'}, data=json.dumps({'uri': 'http://e
    ...: xample.com'})).content
Out[15]: '{"updated": "2016-10-20T15:29:47.865384+00:00", "group": "__world__", "target": [{"source": "http://example.com"}], "links": {"json": "http://localhost:5000/api/annotations/CPf7qpbaEeagSVvtmuVTZA", "html": "http://localhost:5000/a/CPf7qpbaEeagSVvtmuVTZA", "incontext": "http://localhost:8000/CPf7qpbaEeagSVvtmuVTZA/example.com"}, "tags": [], "text": "", "created": "2016-10-20T15:29:47.865384+00:00", "uri": "http://example.com", "user": "acct:seanh@localhost", "document": {}, "id": "CPf7qpbaEeagSVvtmuVTZA", "permissions": {"read": ["acct:seanh@localhost"], "admin": ["acct:seanh@localhost"], "update": ["acct:seanh@localhost"], "delete": ["acct:seanh@localhost"]}}'

```

chdorner:
> Question: it looks as if someone with an auth client's id and secret can get an api token for any user. Do we want to restrict to only users belonging to that auth client?

I think of it like this:

* JWT OAuth grant: Only for users with the same authority as the client
* Any other ("normal") OAuth flow: users with `hypothes.is` authority

I thought that I already locked-down the JWT OAuth flow to only users with the same authority, if this isn't the case then I consider that a bug and should fix it. (Tracked here: #3999)

> Question: it looks as if auth client secrets are stored in plain text in the database. Wondering if those should be hashed and salted or something?

It is usually beneficial to be able to show the client id and secret in some sort of authclient management panel (like if you create a Twitter OAuth app). I'm not sure if encryption for the secret is needed, maybe we should talk about this again when @nickstenning is back. We can easily move to a solution where we do store it encrypted before we create clients for other people to use.

I've also created an issue for when the passed-in client_id is not a valid UUID: #4000

I'd suggest merging this PR if you're happy with my answers, and then we fix the linked issues separately.

h/views/api_auth.py
```diff
@@ -0,0 +1,33 @@
+# -*- coding: utf-8 -*-
+
+from __future__ import unicode_literals
+
+from h.auth.services import TOKEN_TTL
+from h.exceptions import OAuthTokenError
+from h.util.view import json_view
+
+
+@json_view(route_name='token', request_method='POST')
```
chdorner: It currently has to be a POST form request, yes. This can either be `application/x-www-form-urlencoded` or `multipart/form-data`.
Is it worth supporting JSON POST bodies as well? I'm not sure (I'm not against it, but I rather don't implement it if it isn't needed).
Maybe @sean-roberts has some input on this?

_I'm sure you already know this, but a request to `/api/token` does not need the developer token (because you do include it in your example)._

sean-roberts: My main opinion is to be consistent. It will make our endpoints more predictable and less prone to needless cycles on small differences like encoding. That said, if we expect this to be used by other orgs, we should support JSON bodies as well. I have no data on this but most posting libraries are not `application/x-www-form-urlencoded` by default - they tend to be JSON body by default. Adding a layer of abstraction before interacting with inputs should handle this in a way that is reusable for any API we expect to open to the public.

chdorner: For now this will only be accessible for us, since 3rd parties will generate a JWT and pass that to the client, the client then takes the JWT and calls this endpoint.
I don't see why we wouldn't open this up to other developers, but that's still some months off.

chdorner: Alright, I just found the documentation again. Access Token requests according to [RFC6749](https://tools.ietf.org/html/rfc6749#section-4.1.3) need to be `application/x-www-form-urlencoded`. So I won't make this endpoint work with JSON.

PR #4002, opened 2016-10-25
Make search page work without js

### LINKED ISSUE #3990: Search bar doesn't work without javascript
Turn off javascript in your browser, go to the `/search` page, enter a search and submit it - nothing happens

seanh:
Make the `/search` page work without JavaScript.

Fixes #3990.

Get the `/search` page working without JavaScript again.

Before: when JavaScript wasn't available the user's search text would not be submitted in the `q=` param in the URL, and would disappear from the search bar on form submit:

peek 2016-10-25 13-08

After:

peek 2016-10-25 13-09

When JavaScript _is_ available the search bar should continue to work exactly as before, all searching and lozengification etc should be the same.

Previously what happened was this:

1. The server rendered the `/search` page with two `<input>`s in the search bar `<form>`: an empty, nameless but visible `<input>`, and a hidden `<input type="hidden" name="q">`. The server rendered any search text the user had submitted into the hidden input.

2. At controller initialization time, `SearchBarController` would move the text from the `<input type="hidden" name="q">` into lozenges and/or into the visible `<input>`.

3. The user would type into the visible input. Some of this text might be lozengified by `SearchBarController` and moved from the visible input into lozenges. Some might remain in the visible input. The empty `<input type="hidden" name="q">` remained empty as the user typed.

4. On form submit `SearchBarController` would read text from the lozenges and the visible `<input>` and write it into the `<input type="hidden" name="q">` and then submit the form, so this text constructed from the lozenges and the visible input would be put into the hidden input at the last second submitted as the `q` parameter.

The problem is that when JavaScript is not available, the text just remains in the visible `<input>` and the `<input type="hidden" name="q">` remains empty, so an empty string is always submitted as the `q` parameter and search doesn't work. It always submits an empty search query.

In addition, there is no JavaScript to copy the text from the `<input type="hidden" name="q">` into the visible `<input>` on page load, so the user's search query appears to disappear from the search bar on form submit as well.

What happens now:
    
1. The server renders the `/search` page with a single, visible `<input name="q">` containing any search text the user submitted. There is no hidden input.

2. The user types into the visible input. Some of this text might be lozengified by `SearchBarController` and moved from the visible input into lozenges. Some might remain in the visible input.

3. On form submit `SearchBarController` reads text from the lozenges and the visible `<input>` and writes it into a _new_ `<input type="hidden" name="q">` which it inserts into the DOM immediately before submitting the form. `SearchBarController` also moves the `name="q"` HTML attribute from the visible to the newly-inserted hidden `<input>` at the last second, before submitting the form.
    
   (The reason for inserting a hidden input like this rather than just setting the value of the visible input is so that users don't see a flash as the search text changes for a moment before the page reloads.)
    
When JavaScript is not available, this works exactly like a normal basic form would do with a single input.
    
Changes:
    
* In the server-side HTML template, the `value="{{ q }}"` is moved from the hidden `<input>` back on to the visible `<input>` where it used to be.

   This means that, when JavaScript is not available, when the search page is
reloaded after submitting the search form, the search text that the user has typed remains in the visible `<input>` (rendered there by the server) instead of disappearing (because the server was rendering it into the hidden `<input>` instead).
    
   As a result of this, `SearchBarController`'s `lozengifyHiddenInput()` becomes `lozengifyInput()`, since the search text that it needs to lozengify on controller initialization is now in the `<input>` not in the hidden `<input>`.

* In the server-side HTML template, the `name="q"` attribute is moved from the hidden `<input>` back on to the visible `<input>` (where it used to be). This means that the search bar works without JavaScript again: the user types text into the visible `<input>` and hits `Enter` to submit it as the `q` parameter.

* The hidden input is removed from the search bar form (removed from the server-side HTML template). `SearchBarController` now inserts the hidden input into the DOM at the last second when it is needed - just before submitting the form.

* `this._inputHidden` is removed from `SearchBarController`. It just creates a hidden input when it needs it, doesn't need to keep track of it in a class-level variable.

* `SearchBarController`'s `updateHiddenInputValue()` becomes `insertHiddenInput()`, a method that both creates a hidden input and inserts it into the DOM, _and_ sets the `value` of that hidden `<input>` to the search query that we want to submit according to the lozenges and any text in the visible `<input>`.

* A `submitForm()` method is added that inserts the hidden input and submits the form (called both on `Enter` and on deleting a lozenge)

* `SearchTextParser.getLozengeValues()` now returns an object containing two properties: the array of lozenge values and the "incomplete input value" (any un-lozengifiable text, e.g. an unclosed quoted phrase, that should remain in the `<input>` and not be lozengified).

   This was previously two separate functions `getLozengeValues()` and `getIncompleteInputValue()`. This pair of functions is now called in one place only - in `lozengifyInput()` - and in that place both the lozenge and the remaining text are needed so it makes sense for the one function just to return both.

Known issues:

Before this commit there was already a "flash of no lozenges". The search bar, as rendered by the server, didn't contain any lozenges (or any search text - it was empty). When `SearchBarController` runs it inserts the lozenges and search text into the DOM. This results in a flash where you can see the empty search bar for a moment before the lozenges and search text are inserted.

By moving the un-lozengified search text, as rendered initially by the server, from the hidden input into the visible input, this commit would have made this flash even worse. Instead of seeing an empty `<input>` for a moment you would see an `<input>` containing un-lozengified text, until `SearchBarController` lozengifies it.

To avoid this, some CSS involving `env-js-capable` and `env-js-timeout` has been added to hide the visible input until after lozengification has happened. This returns us to seeing an empty search bar for a moment before the lozenges appear.

`visibility: hidden` (rather than, say, `display: none`) is used to hide the `<input>` so that the input still takes up the same space that it would have taken, affecting the layout of the other still-visible elements on the page in the same way. Otherwise (e.g. with `display: none`), the search bar would shrink in height for a second until the `<input>` was displayed.

### DIFF
--- a/h/static/scripts/controllers/search-bar-controller.js
+++ b/h/static/scripts/controllers/search-bar-controller.js
@@ -16,7 +16,6 @@ class SearchBarController extends Controller {
     this._dropdownItems = Array.from(
       element.querySelectorAll('[data-ref="searchBarDropdownItem"]'));
     this._input = this.refs.searchBarInput;
-    this._inputHidden = this.refs.searchBarInputHidden;
     this._lozengeContainer = this.refs.searchBarLozenges;
     var getActiveDropdownItem = () => {
       return element.querySelector('.js-search-bar-dropdown-menu-item--active');
@@ -137,16 +136,45 @@ class SearchBarController extends Controller {
     };
 
     /**
-     * Updates the hidden input field with the consolidated
-     * search terms from the lozenges.
+     * Insert a hidden <input> into the search <form>.
+     *
+     * The value of the hidden <input> is a search string constructed from
+     * the lozenges plus any text currently in the visible <input>.
+     *
+     * The name="q" attribute is moved from the visible <input> on to the
+     * hidden <input> so that when the <form> is submitted it's the value of
+     * the _hidden_ input, not the visible one, that is submitted as the
+     * q parameter.
+     *
      */
-    var updateHiddenInputValue = () => {
-      var newValue = '';
-      Array.from(this._lozengeContainer.querySelectorAll('.js-lozenge__content'), function(loz) {
-        newValue = newValue + loz.textContent + ' ';
+    var insertHiddenInput = () => {
+      var hiddenInput = document.createElement('input');
+      hiddenInput.type = 'hidden';
+
+      Array.from(this._lozengeContainer.querySelectorAll('.js-lozenge__content')).forEach((loz) => {
+        hiddenInput.value = hiddenInput.value + loz.textContent + ' ';
       });
-      newValue = newValue + getTrimmedInputValue();
-      this._inputHidden.value = newValue;
+      hiddenInput.value = hiddenInput.value + getTrimmedInputValue();
+
+      // When JavaScript isn't enabled this._input is submitted to the server
+      // as the q param. With JavaScript we submit hiddenInput instead.
+      hiddenInput.name = this._input.name;
+      this._input.removeAttribute('name');
+
+      this.refs.searchBarForm.appendChild(hiddenInput);
+    };
+
+    /**
+     * Submit the user's search query to the server.
+     *
+     * We build a search query out of the lozenges plus any text in
+     * this._input. To avoid a potential flash of text if we were to update
+     * this._input with the search terms from the lozenge before submitting it,
+     * we update a hidden input and submit that instead.
+     */
+    var submitForm = () => {
+      insertHiddenInput();
+      this.refs.searchBarForm.submit();
     };
 
     /**
@@ -158,11 +186,10 @@ class SearchBarController extends Controller {
      */
     var addLozenge = content => {
       var deleteCallback = () => {
-        updateHiddenInputValue();
         this._lozengeContainer.querySelectorAll('.js-lozenge').forEach(function(loz) {
           loz.classList.add('is-disabled');
         });
-        this.element.querySelector('form').submit();
+        submitForm();
       };
 
       new LozengeController(
@@ -175,22 +202,16 @@ class SearchBarController extends Controller {
     };
 
     /**
-     * Add incomplete search query terms which start with a quote but are missing the end quote
-     * to the input field from the hidden input field on page load.
+     * Create lozenges for the search query terms already in the input field on
+     * page load.
      */
-    var addIncompleteSearchTermToInput = () => {
-      this._input.value = SearchTextParser.getIncompleteInputValue(this._inputHidden.value);
-    };
-
-    /**
-     * Create lozenges for the search query terms already in
-     * the hidden input field on page load.
-     */
-    var lozengifyHiddenInput = () => {
-      var queryTerms = SearchTextParser.getLozengeValues(this._inputHidden.value);
-      queryTerms.forEach(function(term) {
-        addLozenge(term);
+    var lozengifyInput = () => {
+      var {lozengeValues, incompleteInputValue} = SearchTextParser.getLozengeValues(this._input.value);
+      lozengeValues.forEach(function(value) {
+        addLozenge(value);
       });
+      this._input.value = incompleteInputValue;
+      this._input.style.visibility = 'visible';
     };
 
     /**
@@ -262,8 +283,7 @@ class SearchBarController extends Controller {
               innerHTML.trim();
           selectFacet(facet);
         } else {
-          updateHiddenInputValue();
-          this.element.querySelector('form').submit();
+          submitForm();
         }
       };
 
@@ -331,8 +351,7 @@ class SearchBarController extends Controller {
     this._input.addEventListener('input', handleInputOnInput);
     this._input.addEventListener('focus', handleFocusinOnInput);
 
-    lozengifyHiddenInput();
-    addIncompleteSearchTermToInput();
+    lozengifyInput();
   }
 
   update(state) {
--- a/h/static/scripts/controllers/search-bar-controller.js
+++ b/h/static/scripts/controllers/search-bar-controller.js
@@ -207,9 +207,7 @@ class SearchBarController extends Controller {
      */
     var lozengifyInput = () => {
       var {lozengeValues, incompleteInputValue} = SearchTextParser.getLozengeValues(this._input.value);
-      lozengeValues.forEach(function(value) {
-        addLozenge(value);
-      });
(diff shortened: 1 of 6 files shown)


h/static/scripts/controllers/search-bar-controller.js
```diff
@@ -137,18 +136,48 @@ class SearchBarController extends Controller {
...
+     * the _hidden_ input, not the visible one, that is submitted as the
+     * q parameter.
+     *
      */
-    var updateHiddenInputValue = () => {
-      var newValue = '';
-      Array.from(this._lozengeContainer.querySelectorAll('.js-lozenge__content'), function(loz) {
-        newValue = newValue + loz.textContent + ' ';
+    var insertHiddenInput = () => {
+      var hiddenInput = document.createElement('input');
```
seanh: I'm inserting the hidden input immediately before submitting the form. @robertknight mentioned that it's more conventional for JavaScript controllers to do their DOM modification at controller initialization time. There are a couple of advantages to doing this at form submit time instead, though:

* No empty hidden input sitting in the DOM not doing anything (there's literally no value text going into it as the user types and things get lozengified, _until_ the moment of form submit)

* You can have a single `insertHiddenInput()` method that does it all, instead of needing one method to insert the hidden input into the DOM at controller init time, and then a second method to set the value of the hidden input at form submit time

* `hiddenInput` can be a local variable of this function (I was able to delete a class-level variable)

seanh:
@sheetaluk might be a good person to review this?

h/static/scripts/controllers/search-bar-controller.js
```diff
@@ -137,18 +136,48 @@ class SearchBarController extends Controller {
...
+      hiddenInput.type = 'hidden';
+
+      Array.from(this._lozengeContainer.querySelectorAll('.js-lozenge__content')).forEach((loz) => {
```
robertknight: Stylistic nit. Arrow functions with a single argument should omit braces around the argument.

h/static/scripts/util/search-text-parser.js
```diff
@@ -106,39 +122,7 @@ function getLozengeValuesAndIncompleteSearchTerms(queryString) {
   };
 }
 
-/**
- * Function which returns individual lozenge from a string.
- *
- * @param {String} queryString A string of query terms.
```
robertknight: Nit: Primitive types are written with a lowercase name in JSDoc (eg. `string`, `number`, `boolean`, `null`, `undefined`).

robertknight:
Change here looks good. The conceptual change here seems pretty obvious and straightforward to me, so as much I appreciate the detailed explanation, I think it probably could have been shortened to a paragraph :)

----

An overall concern I have related to this is that we need to make implementing basic HTML and upgraded versions of components that each work reliably easier. We've got plenty of refinements we need to do on the UX for the 98%+ of our userbase who get the normal experience.

For example, as it stands right now 20% of our users will see this:

firefox-search

h/static/scripts/controllers/search-bar-controller.js
seanh: Is that right? I wasn't getting a linter warning. Looked through the airbnb guide, closest I could find was [If your function takes a single argument *and doesn’t use braces*, omit the parentheses. Otherwise, always include parentheses around arguments. ](https://github.com/airbnb/javascript#arrows--one-arg-parens). I think this would come under the otherwise because it does use braces. And indeed if I remove the braces then, if I'm using eslint-config-airbnb, eslint complains _Expected parens around arrow function having a body with curly braces_.

By the way should we change [our eslint config](https://github.com/hypothesis/h/blob/master/.eslintrc) to `"extends": "airbnb"` or make eslint-config-hypothesis extend airbnb? (And then also add instructions, or link to instructions, for installing eslint-config-airbnb as it's a bit tricky.) I can send a PR for that if you want. eslint-config-airbnb has **176** warnings about `search-bar-controller.js` which is a _new_ file!

h/static/scripts/controllers/search-bar-controller.js
```diff
@@ -175,22 +203,16 @@ class SearchBarController extends Controller {
...
+     * Create lozenges for the search query terms already in the input field on
+     * page load.
      */
-    var lozengifyHiddenInput = () => {
-      var queryTerms = SearchTextParser.getLozengeValues(this._inputHidden.value);
-      queryTerms.forEach(function(term) {
-        addLozenge(term);
+    var lozengifyInput = () => {
+      var {lozengeValues, incompleteInputValue} = SearchTextParser.getLozengeValues(this._input.value);
+      lozengeValues.forEach(function(value) {
```
seanh: Hmm, this is existing code (I just changed the variable name) but I may as well put in this tweak while I'm at it...

h/static/scripts/util/search-text-parser.js
seanh: Good to know. I may send a follow-up PR that fixes them all. I don't suppose this is lintable?

h/static/scripts/controllers/search-bar-controller.js
robertknight: _Note for anyone else looking at this - we had a discussion about this on Slack and hashed out a plan there._

h/static/scripts/util/search-text-parser.js
robertknight: > I don't suppose this is lintable?

[Yes](http://eslint.org/docs/rules/valid-jsdoc#prefertype)

h/static/scripts/util/search-text-parser.js
```diff
@@ -79,7 +79,23 @@ function shouldLozengify(phrase) {
   return true;
 }
 
-function getLozengeValuesAndIncompleteSearchTerms(queryString) {
+/**
+ * Return an array of lozenge values from the given string.
+ *
+ * @param {String} queryString A string of query terms.
+ *
+ * @returns {Object} An object with two properties: lozengeValues is an array
```
robertknight: There is a better way to document object shapes in JSDoc, search for uses of `@typedef` in the client or h codebases for examples.

robertknight: eg.

```js
/**
 * @typedef LozengeValues
 * @property {string[]} lozengeValues - Description
 * @property {string} incompleteInputValue - Description
 */

/**
 * ...
 * @return {LozengeValues}
 */
```

robertknight:
LGTM - I have some small nits which can be addressed separately.

PR #4003, opened 2016-10-25
Decouple search bar lozenge tests

seanh:
Decouple `SearchBarController`'s lozenge tests.

A big part of what we want to test when testing `SearchBarController` is what it does on controller initialization. For example, it should split any text in the search bar into lozenges, creating lozenges divs in the DOM. Any lozengified search terms should be removed from the input. It should leave any unfinished search terms in the search input (for example, strings with unclosed quotes in them). This all needs to be tested with lots of different initial values from the server in the `<input>`: empty, simple string, multiple lozenges, different types of lozengez unclosed lozenge (different types), and combinations of these.

This is not compatible with a `beforeEach()` function that creates an `<input>` with one value, the same for every test, and initializes the controller on it before the test's code is executed.

So replace that with a helper function that each test calls instead, as has been done in several places elsewhere.

(This pattern should be used in all JavaScript controller tests imo, in Python decoupling test fixtures is best practice and done throughout our code.)

This will pave the way for adding lots more `SearchBarController` tests for what it does when initialized on lots of different values.

I think this also makes the tests simpler and easier to read.

### DIFF
--- a/h/static/scripts/tests/controllers/search-bar-controller-test.js
+++ b/h/static/scripts/tests/controllers/search-bar-controller-test.js
@@ -1,6 +1,7 @@
 'use strict';
 
 var syn = require('syn');
+var util = require('./util');
 
 var SearchBarController = require('../../controllers/search-bar-controller');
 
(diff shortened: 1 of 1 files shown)


h/static/scripts/tests/controllers/search-bar-controller-test.js
```diff
@@ -251,78 +252,98 @@ describe('SearchBarController', function () {
...
-
-      input = ctrl.refs.searchBarInput;
-      inputHidden = ctrl.refs.searchBarInputHidden;
-    });
 
     afterEach(function () {
-      document.body.removeChild(testEl);
+      if (ctrl) {
+        ctrl.element.remove();
+        ctrl = null;
```
seanh: I really wish that `ctrl` didn't have to be a global variable just so that we can do this cleanup. Is there a better way?

robertknight: Its not a global, its a variable local to the test suite. Its simple enough. You _could_ make the `component()` function take a callback (the unit test code) which took `ctrl` as an argument and then performed cleanup after the callback returned, but I don't think that would add any value here.

sheetaluk: +1

sheetaluk:
LGTM.

PR #4006, opened 2016-10-27
Fix flash of expanded document buckets

seanh:
Document buckets on the /search page are expanded at first (as the HTML
and CSS is delivered from the server) and then when some JavaScript
(search-bucket-controller.js) runs it collapses them all.

So the buckets are all expanded initially and then they collapse, and on
slower connections this change is noticeable:

on-master-with-js

To fix this, this commit changes it so that:

* Without any JavaScript at all (e.g. browser doesn't support js at all)
  the HTML and CSS delivered from the server are such that all of the
  buckets are _expanded_:

   no-js

* If the browser _is_ JS capable then as soon as environment-flags.js
  runs (this is run by a `<script>` in the HTML's `<head>`, so it blocks
  rendering and runs _before_ the browser renders the  HTML `<body>`) it
  adds the "env-js-capable" CSS class. This class collapses all the
  buckets.

  So in a js-capable browser the buckets will be _collapsed_ when the
  document is first rendered, no flash of expanded document buckets will
  be seen, even on a slow connection:
   
   on-this-branch-with-js

* If the browser is js-capable and environment-flags.js runs and adds
  the env-js-capable class, but then the site.bundle.js class
  (which contains search-bucket-controller.js) fails to load, then
  environment-flags.js will add the env-js-timeout CSS class. This class
  expands the document buckets.

  So in a js-capable browser, if loading site.bundle.js failed, the
  buckets will be _expanded_.

  On slow connections when loading site.bundle.js fails, this can cause
  the buckets to be rendered as collapsed at first and then expand, so a
  "flash of collapsed document buckets":

   time out

* In a js-capable browser in which site.bundle.js has loaded correctly,
  clicking on a document bucket's header will add the is-expanded CSS
  class which expands the bucket (and clicking again removes the class).

  Previously this worked the other way round: clicking would _remove_
  the is-hidden class.

### DIFF
--- a/h/static/scripts/controllers/search-bucket-controller.js
+++ b/h/static/scripts/controllers/search-bucket-controller.js
@@ -38,7 +38,7 @@ class SearchBucketController extends Controller {
   }
 
   update(state, prevState) {
-    setElementState(this.refs.content, {hidden: !state.expanded});
+    setElementState(this.refs.content, {expanded: state.expanded});
     setElementState(this.element, {expanded: state.expanded});
 
     // Scroll to element when expanded, except on initial load
--- a/h/static/scripts/tests/controllers/search-bucket-controller-test.js
+++ b/h/static/scripts/tests/controllers/search-bucket-controller-test.js
@@ -33,9 +33,19 @@ describe('SearchBucketController', function () {
     ctrl.element.remove();
   });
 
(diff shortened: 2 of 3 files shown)


robertknight:
LGTM - The flash was quite noticeable before

PR #4007, opened 2016-10-27
Refactor User.get_by_(username/email)

chdorner:
In preparation for the validation layer of `POST /api/users`, this PR adds the necessary `authority` arguments to both `User.get_by_username` and `User.get_by_email`. In addition to this change, the admin interfaces that allow for a user's username or email input will now also have a second field for the authority (prefilled with the currently configured authentication domain, in our case `hypothes.is`).

### DIFF
--- a/h/accounts/schemas.py
+++ b/h/accounts/schemas.py
@@ -49,7 +49,7 @@ def get_blacklist():
 def unique_email(node, value):
     '''Colander validator that ensures no user with this email exists.'''
     request = node.bindings['request']
-    user = models.User.get_by_email(request.db, value)
+    user = models.User.get_by_email(request.db, value, request.auth_domain)
     if user and user.userid != request.authenticated_userid:
         msg = _("Sorry, an account with this email address already exists.")
         raise colander.Invalid(node, msg)
@@ -180,7 +180,7 @@ class ForgotPasswordSchema(CSRFSchema):
 
         request = node.bindings['request']
         email = value.get('email')
-        user = models.User.get_by_email(request.db, email)
+        user = models.User.get_by_email(request.db, email, request.auth_domain)
 
         if user is None:
             err = colander.Invalid(node)
--- a/h/accounts/schemas.py
+++ b/h/accounts/schemas.py
@@ -58,7 +58,7 @@ def unique_email(node, value):
 def unique_username(node, value):
     '''Colander validator that ensures the username does not exist.'''
     request = node.bindings['request']
-    user = models.User.get_by_username(request.db, value)
+    user = models.User.get_by_username(request.db, value, request.auth_domain)
     if user:
         msg = _("This username is already taken.")
         raise colander.Invalid(node, msg)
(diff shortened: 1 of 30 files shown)


seanh:
I'm seeing this warning:

    /home/seanh/.virtualenvs/h/local/lib/python2.7/site-packages/sqlalchemy/sql/sqltypes.py:185: SAWarning: Unicode type received non-unicode bind param value 'localhost'. (this warning may be suppressed after 10 occurrences)

I think it's because `request.auth_domain` is a byte string

seanh:
For some reason when I register a new user account, the activation email isn't being printed to the terminal. Investigating why...

seanh:
> For some reason when I register a new user account, the activation email isn't being printed to the terminal. Investigating why...

Hmm, false alarm, for some reason I wasn't getting activation emails locally for a while there but now it seems to be working again! :|

seanh:
I found an issue with the `/admin/staff` page. Removing staff users doesn't seem to work on this branch. When adding users, I can't specify the authority (seems to default to `request.auth_domain`).

seanh:
I think authority needs to be added to the `/admin/users` page as well. You can search for a user by username or email, but can't specify which authority you're searching for, and it doesn't tell you the authority of the user that it found either.

Both of these pages could be done in follow-up PRs?

seanh:
Done testing now, all seems to work well apart from the issues above

chdorner:
Thanks for the thorough testing. Since the next PR depends on this I'd rather fix the issues in here. I will start working on it.

chdorner:
> One thing I thought is that it would be nice if the default authority was in the model get_by_username() and get_by_email() functions themselves, then all the other code wouldn't need to pass the authority argument (it would just default to the default) unless it actually wanted to fetch a non-default authority. But I guess the problem is that the default authority is request.auth_domain, and we don't want model code to depend on the request object?

Yes, exactly. The default is stored within the pyramid request (or can be fetched through a util method but that one needs the registry). So there is no way to make the model code default to the correctly configured auth domain without having access to the request.

> I was wondering whether a few more models tests would be worth having? For get_by_username() and get_by_email(), test that it returns the right user when there are multiple users with the same authority but different username and email, and when there are users with the same username and email but different authorities.

Done. I just updated the test fixtures to include multiple users and then check that the correct one is fetched instead of adding more tests.

> And the other thing I was wondering was whether the tests for the admins, nipsa and features views ought to test what happens when you submit a non-default authority? I didn't see a test for that.

Done.

> I'm seeing this warning ...

That is fixed now. I'm making sure that `request.auth_domain` is always `h._compat.text_type`.

> I found an issue with the /admin/staff page. Removing staff users doesn't seem to work on this branch.

I forgot about `/admin/staff`, that page now also supports specifying the authority.

> I think authority needs to be added to the /admin/users page as well.

Done. This could benefit from some manual testing locally (especially deleting, renaming, activating users). It works for me, but always good to have other people test it out locally as well.

seanh:
Ok, I'll give this another test...

seanh:
@chdorner Oops, looks like `bin/hypothesis --dev user admin seanh` needs to be updated

chdorner:
> Oops, looks like `bin/hypothesis --dev user admin seanh` needs to be updated

Done. It is now even possible to specify the authority when a non-default is needed `--authority foo.com`.

seanh:
Great work Christof, thanks for sweating the details on this one

PR #4008, opened 2016-10-27
Create user api validation

chdorner:
~This depends on #4007 being merged and then needs a rebase.~

This PR adds validation of the `POST /api/users` endpoint. It first validates the structure of the JSON input according to the same schema as we publish in the documentation, that the passed-in `authority` is the same as the one of the authenticated client, and then also adds uniqueness validation (username/email within the same authority).

### DIFF
--- a/h/accounts/schemas.py
+++ b/h/accounts/schemas.py
@@ -17,6 +17,7 @@ from h.models.user import (
     USERNAME_MIN_LENGTH,
     USERNAME_PATTERN,
 )
+from h.schemas import JSONSchema
 
 _ = i18n.TranslationString
 log = logging.getLogger(__name__)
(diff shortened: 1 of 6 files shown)


h/accounts/schemas.py
```diff
@@ -404,5 +405,34 @@ class NotificationsSchema(CSRFSchema):
...
+    schema = {
+        'type': 'object',
+        'properties': {
+            'authority': {
+                'type': 'string',
+                'format': 'hostname',
+            },
+            'username': {
+                'type': 'string',
+                'minLength': 3,
```
nickstenning: At least until we get round to moving the actual schemas into data files, I reckon it would probably be wise to use `USERNAME_MIN_LENGTH` and `USERNAME_MAX_LENGTH` here. I think they're already available in this module.

h/views/api_users.py
```diff
@@ -23,17 +25,17 @@ def create(request):
...
-        'authority': client.authority,
-        'username': payload['username'],
-        'email': payload['email'],
-    }
+    schema = schemas.CreateUserAPISchema()
+    appstruct = schema.validate(request.json_body)
 
-    user_signup_service = request.find_service(name='user_signup')
-    user = user_signup_service.signup(require_activation=False, **user_props)
+    _check_authority(client, appstruct)
```
nickstenning: I think this approach is absolutely fine for now, but one option that would keep the validation more self-contained would be to pass the client into the schema constructor and then override `validate`. That is, the class would look like:

```python 
class CreateUserAPISchema(JSONSchema):
    schema = ...

    def __init__(self, client):
        self.client = client
        super(CreateUserAPISchema, self).__init__()

    def validate(self, data):
        appstruct = super(CreateUserAPISchema, self).validate(data)
        if self.client.authority != data.get('authority'):
            ...
        return appstruct
```

PR #4011, opened 2016-10-27
Extract autosuggestion dropdown logic into a separate control

### LINKED ISSUE #3992: Dropdown menu should open after lozengification
As soon as I put the keyboard cursor into the (empty) search box, the dropdown appears. But after I type a word and hit space to lozengify it, the dropdown does not appear until I've typed one more letter:

peek 2016-10-20 19-09

sean-roberts:
The first commit pulls out the requirements that we need for a reusable suggestion component into its own controller and uses that component on the searchbar controller.

The thought behind the architecture is that it should be reusable for any future need of an autosuggestion feature (which is really common to need this in many places). It's worth noting that I made it reusable enough to scale well without making too many assumptions about the usage of the component. For example, the component offers a way to add the list and a way for the consumer to decide how the list should be reduced on interaction with the input. This means that there are no data structure limitations on the list items or how it can be filtered.

Also, you will see there is no default styling for the autosuggest controller elements. Once we have more than one use of the component, we can extract the reusable styles into a default set of styling for the components - once we have the pattern we want to reflect.

I tested pretty thoroughly in unit tests and in smoke tests in various browsers. You will see that I have a few fixes for regression bugs from previous master commits in here as well.

also: Fixes #3992

### DIFF
--- a/h/static/scripts/controllers/autosuggest-dropdown-controller.js
+++ b/h/static/scripts/controllers/autosuggest-dropdown-controller.js
@@ -1,0 +1,442 @@
+'use strict';
+
+const Controller = require('../base/controller');
+const setElementState = require('../util/dom').setElementState;
+const updateHelper = require('../util/update-helpers.js');
+
+const ENTER = 13;
+const UP = 38;
+const DOWN = 40;
+
+/**
+ * Controller for adding autosuggest control to a piece of the page
+ */
+class AutosuggestDropdownController extends Controller {
+
+  /*
+   * @typedef {Object} ConfigOptions
+   * @property {Function} renderListItem - called with every item in the list
+   *   after the listFilter function is called. The return value will be the final
+   *   html that is set in the list item DOM.
+   * @property {Function} listFilter - called at initialization, focus of input,
+   *   and input changes made by user. The function will receieve the full list
+   *   and the current value of the input. This is meant to be an pure function
+   *   that will return a filtered list based on the consumer's domain needs.
+   * @properly {Function} onSelect - called once the user has made a selection of an
+   *   item in the autosuggest. It will receive the item selected as the only argument.
+   * @property {Object} [classNames] - this is the enumerated list of class name
+   *   overrides for consumers to customize the UI.
+   *   Possible values: { container, list, item, activeItem, header }
+   * @property {String} [header] - value of the header label at top of suggestions
+   */
+
+  /**
+   * @param {HTMLInputElement} inputElement that we are responding to in order to provide
+   *    suggestions. Note, we will add the suggestion container as a sibling to this
+   *    element.
+   * @param {ConfigOptions} configOptions are used to set the initial set of items and the header
+   *    as well as providing the hooks for updates and callbacks
+   *
+   */
+  constructor(inputElement, configOptions) {
+
+    super(inputElement, configOptions);
+
+    if (!configOptions.renderListItem){
+      throw new Error('Missing renderListItem callback in AutosuggestDropdownController constructor');
+    }
+
+    if (!configOptions.listFilter){
+      throw new Error('Missing listFilter function in AutosuggestDropdownController constructor');
+    }
+
+    if (!configOptions.onSelect){
+      throw new Error('Missing onSelect callback in AutosuggestDropdownController constructor');
+    }
+
+    // set up our element class attribute enum values
+    // Note, we currently are not doing anything with the default
+    // classes, but we have them if we wanted to give something for a default
+    // styling.
+    if (configOptions.classNames){
+      this.options.classNames.container = configOptions.classNames.container || 'autosuggest__container';
+      this.options.classNames.list = configOptions.classNames.list || 'autosuggest__list';
+      this.options.classNames.item = configOptions.classNames.item || 'autosuggest__list-item';
+      this.options.classNames.activeItem = configOptions.classNames.activeItem || 'autosuggest__list-item--active';
+      this.options.classNames.header = configOptions.classNames.header || 'autosuggest__header';
+    }
+
+    // renaming simply to make it more obvious what
+    // the element is in other contexts of the controller
+    this._input = this.element;
+
+    // initial state values
+    this.setState({
+      visible: false,
+      header: configOptions.header || '',
+
+      // working list that are displayed to use
+      list: [],
+
+      // rootList is the original set that the filter
+      // will receive to determine what should be shown
+      rootList: [],
+    });
+
+    this._setList(configOptions.list);
+  }
+
+  update(newState, prevState){
+
+    // if our prev state is empty then
+    // we assume that this is the first update/render call
+    if (!('visible' in prevState)){
+
+      // create the elements that make up the component
+      this._renderContentContainers();
+      this._addTopLevelEventListeners();
+    }
+
+    if (newState.visible !== prevState.visible){
+      // updates the dom to change the class which actually updates visibilty
+      setElementState(this._suggestionContainer, { open: newState.visible});
+    }
+
+    if (newState.header !== prevState.header){
+      this._header.innerHTML = newState.header;
+    }
+
+    let listChanged = updateHelper.listIsDifferent(newState.list, prevState.list);
+
+    if (listChanged){
+      this._renderListItems();
+    }
+
+    // list change detection is needed to persist the
+    // currently active elements over to the new list
+    if (newState.activeId !== prevState.activeId || listChanged){
+
+      let currentActive = this._getActiveListItemElement();
+
+      if (prevState.activeId && currentActive){
+        currentActive.classList.remove(this.options.classNames.activeItem);
+      }
+
+      if (newState.activeId && newState.list.find((item)=>item.__suggestionId === newState.activeId)){
+        this._listContainer
+          .querySelector(`[data-suggestion-id="${newState.activeId}"]`)
+          .classList.add(this.options.classNames.activeItem);
+      }
+    }
+  }
+
+  /**
+   * sets what would be the top header html
+   *  to give context of what the suggestions are for
+   *
+   * @param  {string} header Html to place in header. You can pass plain text
+   *  as well.
+   */
+  _setHeader(header){
+    this.setState({
+      header,
+    });
+  }
+
+  /**
+   * update the current list
+   *
+   * @param  {Array} list The new list.
+   */
+  _setList(list){
+
+    if (!Array.isArray(list)){
+      throw new TypeError('setList requires an array first argument');
+    }
+
+    this.setState({
+      rootList: list.map((item)=>{
+        return Object.assign({}, item, {
+          // create an id that lets us direction map
+          // selection to arbitrary item in list.
+          // This allows lists to pass more than just the required
+          // `name` property to know more about what the list item is
+          __suggestionId: Math.random().toString(36).substr(2, 5),
+        });
+      }),
+    });
+
+    this._filterListFromInput();
+  }
+
+  /**
+   * we will run the consumer's filter function
+   *  that is expected to be a pure function that will receive the
+   *  root list (the initial list or list made with setList) and
+   *  the input's current value. that function will return the array items,
+   *  filtered and sorted, that will be set the new working list state and
+   *  be rerendered.
+   */
+  _filterListFromInput(){
+    this.setState({
+      list: this.options.listFilter(this.state.rootList, this._input.value) || [],
+    });
+  }
+
+  /**
+   * hit the consumers filter function to determine
+   *   if we still have list items that need to be shown to the user.
+   */
+  _filterAndToggleVisibility(){
+    this._filterListFromInput();
+
+    this._toggleSuggestionsVisibility(/*show*/ this.state.list.length > 0);
+  }
+
+  /**
+   * lookup the active element, get item from
+   *  object from list that was passed in, and invoke the onSelect callback.
+   *  This is process to actually make a selection
+   */
+  _selectCurrentActiveItem(){
+
+    const currentActive = this._getActiveListItemElement();
+    const suggestionId = currentActive && currentActive.getAttribute('data-suggestion-id');
+    const selection = this.state.list.filter((item)=>{ return item.__suggestionId === suggestionId;})[0];
+
+    if (selection){
+      this.options.onSelect(selection);
+      this._toggleSuggestionsVisibility(/*show*/false);
+      this.setState({
+        activeId: null,
+      });
+    }
+  }
+
+  /**
+   * update the list item dom elements with
+   *  their "active" state when the user is hovering.
+   *
+   * @param  {bool} hovering are we hovering on the current element
+   * @param  {Event} event    event used to pull the list item being targeted
+   */
+  _toggleItemHoverState(hovering, event){
+
+    let currentActive = this._getActiveListItemElement();
+    let target = event.currentTarget;
+
+    if ( hovering && currentActive && currentActive.contains(target)){
+      return;
+    }
+
+    this.setState({
+      activeId: hovering ? target.getAttribute('data-suggestion-id') : null,
+    });
+  }
+
+  /**
+   * this function piggy backs on the setElementState
+   *  style of defining element state in its class. Used in combination with the
+   *  this.options.classNames.container the consumer has easy access to what the visibility state
+   *  of the container is.
+   *
+   * @param  {bool} show should we update the state to be visible or not
+   */
+  _toggleSuggestionsVisibility(show) {
+
+    // keeps the internal state synced with visibility
+    this.setState({
+      visible: !!show,
+    });
+  }
+
+  /**
+   * @returns {HTMLElement}  the active list item element
+   */
+  _getActiveListItemElement(){
+    return this._listContainer.querySelector('.' + this.options.classNames.activeItem);
+  }
+
+  /**
+   * navigate the list, toggling the active item,
+   *  based on the users arrow directions
+   *
+   * @param  {bool} down is the user navigating down the list?
+   */
+  _keyboardSelectionChange(down){
+
+    const currentActive = this._getActiveListItemElement();
+    let nextActive;
+
+    // we have a starting point, navigate on siblings of current
+    if (currentActive){
+
+      if (down){
+        nextActive = currentActive.nextSibling;
+      } else {
+        nextActive = currentActive.previousSibling;
+      }
+
+    // we have no starting point, let's navigate based on
+    // the directional expectation of what the first item would be
+    } else if (down){
+      nextActive = this._listContainer.firstChild;
+    } else {
+      nextActive = this._listContainer.lastChild;
+    }
+
+    this.setState({
+      activeId: nextActive ?  nextActive.getAttribute('data-suggestion-id') : null,
+    });
+  }
+
+  /**
+   * build the DOM structure that makes up
+   *  the suggestion box and content containers.
+   */
+  _renderContentContainers(){
+
+    // container of all suggestion elements
+    this._suggestionContainer = document.createElement('div');
+    this._suggestionContainer.classList.add(this.options.classNames.container);
+
+
+    // child elements that will be populated by consumer
+    this._header = document.createElement('h4');
+    this._header.classList.add(this.options.classNames.header);
+    this._setHeader(this.state.header);
+    this._suggestionContainer.appendChild(this._header);
+
+    this._listContainer = document.createElement('ul');
+    this._listContainer.classList.add(this.options.classNames.list);
+    this._suggestionContainer.appendChild(this._listContainer);
+
+    // put the suggestions adjacent to the input element
+    // firefox does not support insertAdjacentElement
+    if (HTMLElement.prototype.insertAdjacentElement){
+      this._input.insertAdjacentElement('afterend', this._suggestionContainer);
+    } else {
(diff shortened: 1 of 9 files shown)


h/static/scripts/controllers/search-bar-controller.js
```diff
@@ -186,7 +168,7 @@ class SearchBarController extends Controller {
      */
     var addLozenge = content => {
       var deleteCallback = () => {
-        this._lozengeContainer.querySelectorAll('.js-lozenge').forEach(function(loz) {
+        Array.from(this._lozengeContainer.querySelectorAll('.js-lozenge')).forEach(function(loz) {
```
sean-roberts: Bug fix

h/static/styles/partials-v2/_search-bar.scss
```diff
@@ -60,7 +63,8 @@
   display: none;
   padding: 20px 0 0 0;
   position: absolute;
-  margin-top: 52px;
+  top: 52px;
```
sean-roberts: bug fix for safari box alignment

robertknight: Can you explain a little more?

h/static/scripts/controllers/autosuggest-dropdown-controller.js
```diff
@@ -0,0 +1,431 @@
...
+    this._renderContentContainers();
+
+    this._addTopLevelEventListeners();
+
+    this._setList(configOptions.list);
+
+  }
+
+  // required to satisfy controller.js interface
+  update(){}
```
robertknight: This function is called with the current and previous states whenever `setState()` is called, so the idea is that all DOM updates in response to state changes should happen here. It is effectively the `render()` method. See `FormController` for an example.

robertknight:
There are quite a lot of ESLint complaints for the autosuggest code. Have you got the [ESLint package](https://atom.io/packages/linter-eslint) enabled in Atom? Having live feedback makes the linter much nicer to work with.

h/static/scripts/controllers/autosuggest-dropdown-controller.js
```diff
@@ -0,0 +1,431 @@
...
+/**
+ * Controller for adding autosuggest control to a piece of the page
+ */
+class AutosuggestDropdownController extends Controller {
+
+  /**
+   * @param {HTMLElement} inputElement that we are responding to in order to provide
+   *    suggestions. Note, we will add the suggestion container as a sibling to this
+   *    element.
+   * @param {Object} configOptions are used to set the initial set of items and the header
```
robertknight: I'd suggest using an `@typedef` to document the supported options here as we do in `controller.js`

h/static/scripts/controllers/autosuggest-dropdown-controller.js
```diff
@@ -0,0 +1,431 @@
+'use strict';
+
+const Controller = require('../base/controller');
+const setElementState = require('../util/dom').setElementState;
+
+let CONTAINER_CLASS = '';
```
robertknight: Since these are global variables, if two instances of the autosuggest component are created and both use different styles, whichever is created last will win and set the styles used by the other instance - at least for dynamic updates.

h/static/scripts/tests/controllers/autosuggest-dropdown-controller-test.js
```diff
@@ -0,0 +1,412 @@
...
+    });
+
+
+    it('opens and closes based on focus status', function(done){
+
+      new AutosuggestDropdownController(input, defaultConfig);
+
+      let suggestionContainer = container.querySelector('.' + defaultConfig.classNames.container);
+
+      assert.isFalse(suggestionContainer.classList.contains('is-open'), 'basline hidden');
```
robertknight: I think it can often be helpful to abstract out the tests for the visual state of a component into a helper (see eg. the `isEditing()` and `isSaving()` functions in `form-controller-test.js`). This then requires fewer changes in the tests if you later change how you implement the style.

h/static/scripts/controllers/autosuggest-dropdown-controller.js
```diff
@@ -0,0 +1,431 @@
...
+    this._header.innerHTML = header;
+  }
+
+
+  /**
+   * _setList - update the current list
+   *
+   * @param  {Array} list The new list.
+   */
+  _setList(list){
```
robertknight: This this is a private method, what is the API for replacing the list contents with the suggestions (eg. when the user types `tag:`) going to be? I would have expected either a public instance method or a call to `component.setState({list: newSuggestions})`

h/static/scripts/controllers/autosuggest-dropdown-controller.js
sean-roberts: good catch!

h/static/scripts/controllers/autosuggest-dropdown-controller.js
sean-roberts: sure, makes sense

h/static/scripts/controllers/autosuggest-dropdown-controller.js
sean-roberts: Just before I submitted the PR, I removed the public hooks to the select methods, like this one, that would be needed. I did that so to not conflate the needs of my next work with the needs of the refactor. So they will likely be coming back post refactor

h/static/scripts/controllers/autosuggest-dropdown-controller.js
sean-roberts: So, originally I was using this method for that, then it grew to be too awkward. But it evolved a lot since that first pass, so it's probably ok to have some calls into those dom update methods. However, this isn't a verbose framework so putting all details of dom in this function is not practical and will make it much harder to maintain. I will update with what makes sense and we can go from there.

sean-roberts:
> Wherever we populate the DOM with user-provided content, we need to be sure that it is guaranteed to be suitably escaped or set it via a method (eg. textContent) that doesn't treat the input as HTML.

This is not a concern of the work that I have done here. For expressiveness, the autosuggest openly allows html to be passed in so to make future changes much easier. If you can point out a specific case in the code where that is not the case, let me know.

> Where possible, I would suggest following the pattern where DOM updates only happen in the constructor and update() methods, so that the implementation is consistent with how eg. inline forms and tooltips work. This avoids bugs where the state gets out of sync with what is displayed.

I'll see if I can make a sensible change that can follow that paradigm.

> There are many lint warnings and little stylistic differences compared to other code. I don't like nitpicking but being consistent will make it easier for everyone to jump in on reviews and concentrate on the substance of the changes. Specific things:

There should be enforcement on these things in the tooling - not code review alone. Otherwise, it's just wasting time in reviews.

sean-roberts:
@robertknight changes made!

h/static/scripts/controllers/autosuggest-dropdown-controller.js
```diff
@@ -0,0 +1,454 @@
...
+    if(configOptions.classNames){
+      this.options.classNames.container = configOptions.classNames.container || 'autosuggest__container';
+      this.options.classNames.list = configOptions.classNames.list || 'autosuggest__list';
+      this.options.classNames.item = configOptions.classNames.item || 'autosuggest__list-item';
+      this.options.classNames.activeItem = configOptions.classNames.activeItem || 'autosuggest__list-item--active';
+      this.options.classNames.header = configOptions.classNames.header || 'autosuggest__header';
+    }
+
+    // renaming simply to make it more obvious what
+    // the element is in other contexts the control
```
robertknight: Slightly odd wording in this comment

h/static/scripts/controllers/autosuggest-dropdown-controller.js
```diff
@@ -0,0 +1,454 @@
...
+
+  /**
+   * _setList - update the current list
+   *
+   * @param  {Array} list The new list.
+   */
+  _setList(list){
+
+    if(!Array.isArray(list)){
+      console.error('setList requires an array first argument');
```
robertknight: I would suggest throwing a `TypeError` rather than just logging an error here. That way the error will get reported back to us via Raven.

h/static/scripts/controllers/autosuggest-dropdown-controller.js
```diff
@@ -0,0 +1,454 @@
...
+
+  /**
+   * _listDidChange - compare two list arrays and decide if they have changed
+   *
+   * @param  {Array} listA
+   * @param  {Array} listB
+   * @returns {bool}       the result of comparing if the two
+   *   arrays seem like they have changed. True if they have changed
+   */
+  _listDidChange(listA, listB){
```
robertknight: This doesn't need to be a method, it could just be extracted into a helper function at the top of the file.

h/static/scripts/controllers/autosuggest-dropdown-controller.js
```diff
@@ -0,0 +1,454 @@
...
+   */
+  constructor(inputElement, configOptions) {
+
+    super(inputElement, configOptions);
+
+    if(!configOptions.renderListItem){
+      throw new Error('Missing renderListItem callback in AutosuggestDropdownController constructor');
+    }
+
+    if(!configOptions.listReducer){
```
robertknight: I think it would be a little clearer what this function was for if it used the term `filter` rather than `reduce` - cf. `Array.filter()` (returns a subset of items that match a predicate) vs. `Array.reduce()` (a more generic method which returns a single value by applying a reduce/fold function over the contents of the list, eg. for summing up a list of values).

h/static/scripts/controllers/autosuggest-dropdown-controller.js
```diff
@@ -0,0 +1,454 @@
...
+   * @properly {Function} onSelect - called once the user has made a selection of an
+   *   item in the autosuggest. It will receive the item selected as the only argument.
+   * @property {Object} [classNames] - this is the enumerated list of class name
+   *   overrides for consumers to customize the UI.
+   *   Possible values: { container, list, item, activeItem, header }
+   * @property {String} [header] - value of the header label at top of suggestions
+   */
+
+  /**
+   * @param {HTMLElement} inputElement that we are responding to in order to provide
```
robertknight: Presumably this needs to be an `HTMLInputElement` rather than just an `HTMLElement`?

h/static/scripts/controllers/autosuggest-dropdown-controller.js
```diff
@@ -0,0 +1,454 @@
...
+    }
+
+    if(this._listDidChange(newState.list, prevState.list)){
+      this._renderListItems();
+    }
+
+  }
+
+  /**
+   * _setHeader - sets what would be the top bolded header text
```
robertknight: Small nit about comment style in the client and H codebase. We generally use a full sentence for the first line and don't mention the function/method name in the comment.

See comments in the other controllers for examples.

h/static/scripts/controllers/autosuggest-dropdown-controller.js
```diff
@@ -0,0 +1,454 @@
...
+   *  to give context of what the suggestions are for
+   *
+   * @param  {string} header Text to place in header
```
robertknight: Is this HTML or plain text? From the docs I would interpret it as plain text but looking at the code it treats this as HTML.

h/static/scripts/controllers/autosuggest-dropdown-controller.js
```diff
@@ -0,0 +1,454 @@
...
+   */
+  _setList(list){
+
+    if(!Array.isArray(list)){
+      console.error('setList requires an array first argument');
+    }
+
+    this.setState({
+      rootList: list.map((item)=>{
+        return Object.assign(item, {
```
robertknight: This mutates the input item, is that intended, or did you mean `Object.assign({}, item, ...)`?

h/static/scripts/controllers/autosuggest-dropdown-controller.js
```diff
@@ -0,0 +1,454 @@
...
+    this._setHeader(this.state.header);
+    this._suggestionContainer.appendChild(this._header);
+
+    this._listContainer = document.createElement('ul');
+    this._listContainer.classList.add(this.options.classNames.list);
+    this._suggestionContainer.appendChild(this._listContainer);
+
+    // put the suggestions adjacent to the input element
+    // firefox does not support insertAdjacentElement
+    if(HTMLElement.prototype.insertAdjacentElement){
```
robertknight: I would be inclined to keep just the Firefox-safe code path here, that way you've got less to test.

robertknight: Oh, and FWIW I'll note that Firefox >= 48 [does support this](https://developer.mozilla.org/en-US/docs/Web/API/Element/insertAdjacentElement), but that's too new for us to rely on.

h/static/scripts/controllers/autosuggest-dropdown-controller.js
```diff
@@ -0,0 +1,454 @@
...
+      li.classList.add(this.options.classNames.item);
+      li.setAttribute('data-suggestion-id', listItem.__suggestionId);
+
+      // persist the active state of selections
+      // through the rendering process
+      if(listItem.__suggestionId === currentActiveSelectorId){
+        li.classList.add(this.options.classNames.activeItem);
+      }
+
+      // this should use some sort of event delegation if
```
robertknight: Controller's have a convenience `this.on(eventName, callback)` you can use if you wish. I agree with the comment here, given how few elements we're going to display, event delegation is not worth worrying about.

h/static/scripts/controllers/autosuggest-dropdown-controller.js
```diff
@@ -0,0 +1,454 @@
...
+      // we find we want to expand this to lists with *a lot* of items in it
+      // But for now this binding has no real affect on small list perf
+      li.addEventListener('mouseenter', this._toggleItemHoverState.bind(this, /*hovering*/true));
+      li.addEventListener('mouseleave', this._toggleItemHoverState.bind(this, /*hovering*/false));
+      li.addEventListener('mousedown', (event)=>{
+        // for situations like mobile, hovering might not be
```
robertknight: Has the mobile UX been tested? Do you think any design changes are needed for small screens?

h/static/scripts/controllers/autosuggest-dropdown-controller.js
```diff
@@ -0,0 +1,454 @@
...
+  /**
+   * _addTopLevelEventListeners - The events that can be set on a "global" or top
+   *  level scope, we are going to set them here.
+   */
+  _addTopLevelEventListeners(){
+
+    // we need to use mousedown instead of click
+    // so we can beat the blur event which can
+    // change visibility/target of the active event
+    document.addEventListener('mousedown', (event)=>{
```
robertknight: This subscription should be cleaned up in the `beforeRemove()` callback (see controller.js)

h/static/scripts/controllers/search-bar-controller.js
```diff
@@ -12,128 +12,109 @@ class SearchBarController extends Controller {
...
-    };
+    var selectFacet = facet => {
+      this._input.value = facet;
 
-    var isHidden = element => {
-      return element &&
-        (element.nodeType !== 1 ||
-          !element.classList ||
-          element.classList.contains('is-hidden'));
+      setTimeout(function() {
```
robertknight: Use an arrow function here to avoid the `bind`?

h/static/scripts/tests/controllers/autosuggest-dropdown-controller-test.js
```diff
@@ -0,0 +1,413 @@
+'use strict';
+
+const syn = require('syn');
+const unroll = require('../util').unroll;
+
```
robertknight: Style nit. Should have just one blank line between sections.

h/static/scripts/tests/controllers/autosuggest-dropdown-controller-test.js
```diff
@@ -0,0 +1,413 @@
...
+// mouseenter and mouseleave events.
+const mouseMove = (()=>{
+
+  let _lastMovePos;
+
+  return (posObj) => {
+
+    if(_lastMovePos){
+      const prevEl = document.elementFromPoint(_lastMovePos.pageX, _lastMovePos.pageY);
+      const leaveEvent = document.createEvent( 'Events' );
```
robertknight: PhantomJS 2.x supports the Event constructor, so you can do `new Event('type', options)` rather than using createEvent + initEvent which I think is more readable.

h/static/scripts/tests/controllers/autosuggest-dropdown-controller-test.js
```diff
@@ -0,0 +1,413 @@
...
+    it('opens and closes based on focus status', function (done) {
+
+      new AutosuggestDropdownController(input, defaultConfig);
+
+      assert.isFalse(isSuggestionContainerVisible(), 'basline hidden');
+
+      syn
+        .click(input, () => {
+
+          assert.isTrue(isSuggestionContainerVisible(), 'focus show');
```
robertknight: The last argument to `assert` should be a short message that explains what the error was. eg. "Suggestions not visible when input focused") - something that makes it easy to understand what was wrong if the test fails.

h/static/scripts/tests/controllers/autosuggest-dropdown-controller-test.js
```diff
@@ -0,0 +1,413 @@
...
+
+          assert.isTrue(list.childNodes[0].classList.contains(defaultConfig.classNames.activeItem));
+          assert.equal(list.childNodes[0].querySelector('.a-title').textContent.trim(),
+            defaultConfig.list[1].title);
+
+          done();
+        });
+    });
+
+    it('can keyboard select', function(done){
```
robertknight: Description is perhaps a little terse - "Updates selection when keys are pressed"?

h/static/scripts/tests/controllers/autosuggest-dropdown-controller-test.js
```diff
@@ -0,0 +1,413 @@
...
+          assert.isTrue(list.childNodes[3].classList.contains(defaultConfig.classNames.activeItem));
+        })
+        .type('[up][up][up]', ()=>{
+          assert.lengthOf(list.querySelectorAll('.' + defaultConfig.classNames.activeItem), 1);
+          assert.isTrue(list.childNodes[0].classList.contains(defaultConfig.classNames.activeItem));
+
+
+          mouseMove(center(list.childNodes[2]));
+
+          assert.lengthOf(list.querySelectorAll('.' + defaultConfig.classNames.activeItem), 1);
```
robertknight: Doesn't block merge, but I often find it helpful to extract UI queries like this into helper functions, for two reasons:

1. If you change the UI, you only have one place to update in the tests
2. When reading the tests, it is easier to understand something like `assert.equal(countSelectedItems(), 1)` than `assert.lengthOf(list.querySelectorAll('.' + defaultConfig.classNames.activeItem), 1);`

h/static/styles/partials-v2/_search-bar.scss
robertknight: Ah OK. I see the same in Firefox. I don't suppose you know why there is a difference between Chrome and Safari/Firefox here?

h/static/scripts/controllers/autosuggest-dropdown-controller.js
sean-roberts: It's not an upgrade controller so that beforeRemove isn't applicable. It's also not really needed at this stage.

h/static/scripts/controllers/search-bar-controller.js
sean-roberts: Yeah we can, this is a carry over.

h/static/scripts/controllers/autosuggest-dropdown-controller.js
sean-roberts: Sure, originally it was more reduce like but receiving the full list felt like a nicer api than a reduce cb.

sean-roberts:
@robertknight updates made. But I have to say, the majority of the suggestions were pretty nit-picky and not much more than "here's how I would have done this" which isn't helpful a helpful review nor does it get the work done better. I did a handful of the suggestions as to not bikeshed atm, but we have to get better about this.

robertknight:
I've just tested the updated branch and search submission appears to be broken. Steps to reproduce:

1. Focus search box, type "test"
2. Press Space
3. Press Enter

Expected: A search is performed with the query string set to "q=test"
Actual: A search is performed with an empty query string

robertknight:
> But I have to say, the majority of the suggestions were pretty nit-picky and not much more
> than "here's how I would have done this" which isn't helpful a helpful review nor does it get the work done better.

Please understand that I'm not trying to be difficult. I'd group my comments into two categories - those which highlight things that I thought there was a possibility that you might not have been aware of (eg. the fact that PhantomJS 2.x supports `new Event` rather than the legacy style) and those where there is a clear style difference from how the rest of the code has been written (eg. spacing around control structures). The latter category is something where I hope we can agree it is valuable to have consistency, but also that we shouldn't need to spend time on it in code reviews - tools should just capture those aspects.

Anyhow, lets get the function issue (https://github.com/hypothesis/h/pull/4011#issuecomment-257923217 ) resolved and then I think this can be merged.

sean-roberts:
Not able to reproduce on Chrome or FF

PR #4012, opened 2016-10-28
Improve the group invite link field

seanh:
Some improvements to the group invite field on the group page. This is the old group page (`/groups/abcxyz/my-group`), but the same group invite field is needed on the new group sidebar, and I intend to component-ize and reuse this code.

Changes:
- Text tweak
- Twitter, Facebook etc buttons removed
- The link text is now read-only so you can't edit it!
- Text is autoselected when you click on it:
  
  autoselect
- Added a clipboard icon which copies the text when clicked:
  
  copy
  
  GitHub uses a tooltip to show the _copied_ notification:
  
  github
  
  Unfortunately our own implementation of these kind of JavaScript tooltips is buggy and couldn't be used for this. I didn't want to have to go and fix that problem now, so I went for the different approach to success notification above.
  
  Note that the GitHub button also "looks clicky" when you click it, ours doesn't. Maybe we can fix that.
  
  This shouldn't happen, but if for some reason the copy fails (`document.execCommand('copy')` returns `false`):
  
  copyfail
  
  I haven't dealt with cases where `document.execCommand` doesn't exist.
- If JavaScript isn't available there's no copy button:
  
  nojs
  
  (the autoselect doesn't work either)
- The click target size of the copy button is as big as its corner of the input field, you don't actually have to click on the (small) icon:
  
  screenshot from 2016-10-28 21-07-58


seanh:
There's some small design tweaks to be done to this: https://hypothes-is.slack.com/archives/design/p1477934451000450

seanh:
As [discussed in Slack](https://hypothes-is.slack.com/archives/design/p1477934451000450) I've made some design tweaks to make this match with the share control in the client: clipboard icons turns black on hover (had to inline the SVG into the HTML), text copied success and failure notification messages changed, time that notification message is displayed for reduce from 3 to 1s.

I didn't do the white box around the notification text.

Ultimately, of course, it's going to be difficult to keep the exact details of UI components like this in sync between the website and the client, as long as they're implemented by completely separate code.

I think this PR is finished for now anyway.

sheetaluk:
LGTM

PR #4014, opened 2016-10-31
Style lozenges with known named query terms differently from other ty…

sheetaluk:
…pes of lozenges.

https://trello.com/c/d5iFjTeQ/3-customize-search-lozenge-display-for-named-search-terms

### DIFF
--- a/h/static/scripts/controllers/lozenge-controller.js
+++ b/h/static/scripts/controllers/lozenge-controller.js
@@ -3,6 +3,7 @@
 var escapeHtml = require('escape-html');
 
 var Controller = require('../base/controller');
+var searchTextParser = require('../util/search-text-parser');
 
 /**
  * Create a lozenge with options.content as its content and append it to containerEl.
@@ -20,9 +21,21 @@ class LozengeController extends Controller {
   constructor(containerEl, options) {
     super(containerEl, options);
     var lozengeEl = document.createElement('div');
+    var lozengeMarkup = escapeHtml(options.content);
+
+    if (searchTextParser.hasKnownNamedQueryTerm(options.content)) {
+      var queryTerm = searchTextParser.getLozengeFacetNameAndValue(options.content);
+      lozengeMarkup = '<span class="lozenge__facet-name">' +
+        escapeHtml(queryTerm.facetName) +
+        '</span>' +
+        ':' +
+        '<span class="lozenge__facet-value">' +
+        escapeHtml(queryTerm.facetValue) +
+        '</span>';
+    }
     lozengeEl.innerHTML =
       '<div class="js-lozenge__content lozenge__content">'+
-      escapeHtml(options.content)+
+      lozengeMarkup +
       '</div>' +
       '<div class="js-lozenge__close lozenge__close">' +
       '<img alt="Delete lozenge" src="/assets/images/icons/lozenge-close.svg">' +
--- a/h/static/scripts/tests/controllers/lozenge-controller-test.js
+++ b/h/static/scripts/tests/controllers/lozenge-controller-test.js
@@ -26,6 +26,40 @@ describe('LozengeController', function () {
     assert.equal(lozengeContentEl.textContent, opts.content);
   });
 
+  it('creates a new lozenge for a known named query term inside the container provided', function () {
+    var facetNameEl;
+    var facetValueEl;
+
+    el = document.createElement('div');
+    opts = {
+      content: 'user:foo',
+    };
+
+    new LozengeController(el, opts);
+    lozengeEl = el.querySelector('.js-lozenge');
+    lozengeContentEl = lozengeEl.querySelector('.js-lozenge__content');
+    facetNameEl = lozengeContentEl.querySelector('.lozenge__facet-name');
+    facetValueEl = lozengeContentEl.querySelector('.lozenge__facet-value');
+
+    assert.equal(facetNameEl.textContent, 'user');
+    assert.equal(facetValueEl.textContent, 'foo');
+  });
+
+  it('does not create a new lozenge for named query term which is not known', function () {
+    el = document.createElement('div');
+    opts = {
+      content: 'foo:bar',
+    };
+
+    new LozengeController(el, opts);
+    lozengeEl = el.querySelector('.js-lozenge');
+    lozengeContentEl = lozengeEl.querySelector('.js-lozenge__content');
+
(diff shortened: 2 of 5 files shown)


h/static/scripts/controllers/lozenge-controller.js
```diff
@@ -1,6 +1,7 @@
 'use strict';
 
 var Controller = require('../base/controller');
+var SearchTextParser = require('../util/search-text-parser');
```
robertknight: Only constructor function names / class names should be capitalized. If you are importing a module use `camelCase`

h/static/scripts/controllers/lozenge-controller.js
```diff
@@ -18,9 +19,21 @@ class LozengeController extends Controller {
   constructor(containerEl, options) {
     super(containerEl, options);
     var lozengeEl = document.createElement('div');
+    var lozengeMakeup = options.content;
```
robertknight: Presumably you mean `lozengeMarkup`?

h/static/scripts/controllers/lozenge-controller.js
```diff
@@ -18,9 +19,21 @@ class LozengeController extends Controller {
...
+
+    if(SearchTextParser.hasKnownNamedQueryTerm(options.content)) {
+      var queryTerm = SearchTextParser.getLozengeFacetNameAndValue(options.content);
+      lozengeMakeup = '<span class="lozenge__facet-name">' +
```
robertknight: Injecting user-controlled content into HTML markup enables trivial XSS attacks: [See this demo](http://localhost:5000/search?q=tag%3A%22%3Cimg%20src=%27http://i.giphy.com/TOWeGr70V2R1K.gif%27%3E%22+)

h/static/scripts/util/search-text-parser.js
```diff
@@ -122,7 +116,73 @@ function getLozengeValues(queryString) {
...
+function hasKnownNamedQueryTerm(queryTerm) {
+  const knownNamedQueryTerms = ['user',
+    'url',
+    'group',
+    'tag'];
+
+  var facetName;
+
+  facetName = getLozengeFacetNameAndValue(queryTerm).facetName;
+  if (knownNamedQueryTerms.indexOf(facetName) >= 0) {
```
robertknight: The `if` statement is not required here, you can just write this as `return knownNamedQueryTerms.indexOf(facetName) >= 0)`

h/static/scripts/util/search-text-parser.js
```diff
@@ -122,7 +116,73 @@ function getLozengeValues(queryString) {
...
+  }
+  return false;
+}
+
+/**
+ * Return an object with the facet name and value for a given query term.
+ *
+ * @param {String} queryTerm The query term string.
+ *
+ * @returns {Object} An object with two properties:
```
robertknight: The return type doesn't match the description if the query term does not contain a facet separator. There are several ways you could address this. One would be to change the return type to `null|Object` where the result is null if there is no facet. The other would be to always return an object with `facetName` and `facetValue` properties but make `facetName` an empty string if there is no separator. My inclination would be to use the latter style as this will probably result in fewer paths through the code.

h/static/scripts/tests/controllers/lozenge-controller-test.js
```diff
@@ -26,6 +26,40 @@ describe('LozengeController', function () {
...
+    var facetValueEl;
+
+    el = document.createElement('div');
+    opts = {
+      content: 'user:foo',
+    };
+
+    new LozengeController(el, opts);
+    lozengeEl = el.querySelector('.js-lozenge');
+    lozengeContentEl = lozengeEl.querySelector('.js-lozenge__content');
```
robertknight: I would suggest extracting these `querySelector()` calls into helper functions. This way individual tests can be more concise and you don't have to update every single test if you change the UI structure of the component.

h/static/scripts/controllers/lozenge-controller.js
```diff
@@ -18,9 +19,21 @@ class LozengeController extends Controller {
   constructor(containerEl, options) {
     super(containerEl, options);
     var lozengeEl = document.createElement('div');
+    var lozengeMarkup = options.content;
+
+    if(searchTextParser.hasKnownNamedQueryTerm(options.content)) {
```
robertknight: Style nit - there should be a space after `if`. Our lint config needs updating to check that.

h/static/scripts/util/search-text-parser.js
```diff
@@ -122,7 +116,84 @@ function getLozengeValues(queryString) {
   };
 }
 
+/**
+ * Return true if the input query term string has a known named
+ * query term.
+ *
+ * @param {String} queryTerm The query term string.
```
robertknight: Nit: For consistency with the rest of our documentation, primitive types should be written in lowercase: `string`, `boolean`, `number`. In other words, what `typeof value` returns for primitive types.

sheetaluk:
>> There is a security issue here that needs to be fixed, plus a set of smaller suggestions.

Fixed as part of this PR:
https://github.com/hypothesis/h/pull/4017

h/static/scripts/controllers/lozenge-controller.js
```diff
@@ -20,9 +21,21 @@ class LozengeController extends Controller {
   constructor(containerEl, options) {
     super(containerEl, options);
     var lozengeEl = document.createElement('div');
+    var lozengeMarkup = escapeHtml(options.content);
```
robertknight: Since you are escaping the markup when generating the HTML, you don't need to do it here as well. In fact, that could possibly result in the wrong output if `getLozengeFacetNameAndValue` doesn't expect escaped input.

h/static/scripts/util/search-text-parser.js
```diff
@@ -122,7 +116,82 @@ function getLozengeValues(queryString) {
   };
 }
 
+/**
+ * Return true if the input query term string has a known named
+ * query term.
+ *
+ * @param {string} queryTerm The query term string.
+ *
+ * @returns {bool} True if the query term string has a known named
```
robertknight: s/bool/boolean for consistency

h/static/scripts/controllers/lozenge-controller.js
robertknight: Edit: I didn't read this closely enough, ignore me :P

robertknight:
LGTM

PR #4016, opened 2016-10-31
Ensure that API token exchange happens within the same authority

### LINKED ISSUE #3999: Only allow JWT OAuth token exchange with the same authority
### Steps to reproduce
1. Create a local user account (`acct:bob@localhost`)
2. Create an authclient with the authority `example.org`
3. Exchange a JWT auth grant token for `acct:bob@localhost` with the created client
### Expected behaviour

It should render an error saying that JWT auth grant token exchange only works when the authclient and user authority are the same.
### Actual behaviour

It happily returns a token for any user with any authority.

chdorner:
The authenticated authclient's authority has to match with the users
authority for the JWT grant token exchange to work.

This was a bug @seanh found when reviewing #3983.

Fixes #3999.

### DIFF
--- a/h/auth/services.py
+++ b/h/auth/services.py
@@ -181,6 +181,10 @@ class OAuthService(object):
             raise OAuthTokenError('user with userid described in subject could not be found',
                                   'invalid_grant')
 
+        if user.authority != authclient.authority:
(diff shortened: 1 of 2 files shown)


seanh:
Cheers! Always good to test that the actual error message was correct as well

PR #4020, opened 2016-11-02
Lint frontend code via a Travis job

robertknight:
This adds a separate job to the Travis build matrix which lints the frontend code, following the same approach that we use for the client repository.

Why Travis instead of Hound etc? It allows us to run arbitrary tasks/configs during the lint, it has proven reliable in the client repo and it doesn't add additional infrastructure to worry about. Furthermore it is more obvious what versions of tools are being used and how developers can run the same lint tasks locally.

Since our codebase is lint clean (after a couple of fixes in the second commit in this PR), the task runs on all frontend code. If we add new lint rules in future, we can configure the task to only run on code that changed in the branch, or we can enable new lint rules on a per directory basis, or we can do what we did in the client repo and fix all existing instances of an issue when a new rule is added.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,6 +24,13 @@ matrix:
       node_js: '6.2'
       before_install: npm install gulp-cli
       script: gulp test
+
+    # Lint frontend code
+    - env: ACTION=frontend-lint
+      language: node_js
+      node_js: '6.2'
+      script: gulp lint
+
 cache:
   directories:
     - node_modules
--- a/h/static/scripts/tests/controllers/dropdown-menu-controller-test.js
+++ b/h/static/scripts/tests/controllers/dropdown-menu-controller-test.js
@@ -3,10 +3,12 @@
 var DropdownMenuController = require('../../controllers/dropdown-menu-controller');
 var util = require('./util');
 
-var TEMPLATE = ['<div class="js-dropdown-menu">',
+var TEMPLATE = [
+  '<div class="js-dropdown-menu">',
   '<span data-ref="dropdownMenuToggle">Toggle</span>',
   '<span data-ref="dropdownMenuContent">Menu</span>',
-  '</div>'].join('\n');
+  '</div>',
+].join('\n');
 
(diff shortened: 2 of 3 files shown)


seanh:
[What a failed run looks like](https://travis-ci.org/hypothesis/h/jobs/172553952), [and a successful one](https://travis-ci.org/hypothesis/h/jobs/172559788)

seanh:
Rebased. Looks good to me, great to see progress on this! Will wait for all the checks to pass after the rebase.

Does this mean we can turn off Hound now?

robertknight:
> Does this mean we can turn off Hound now?

Thanks for the reminder, yes I believe we can. We're not using it for any back-end linting are we? (I'm guessing not, since it doesn't list Python as a supported language on their website)

seanh:
I don't think we're using Hound for Python. I haven't seen it making any comments, its Python support is in beta, we don't have it configured to use our custom Flake8 configuration, it doesn't even support our other Python linter (prospector). Actually I think you may have to explicitly enable Python in your .hound.yml and we haven't. Send a PR to remove it?

chdorner:
I would have added this at the end of the `gulp test` build, the lint only takes about 3-4 seconds, while the whole build takes ~40 seconds. And adding a new build also means longer waiting times for us when we push code to Travis.

robertknight:
@chdorner - For the client repo we split test/lint into separate jobs to avoid conflating correctness failures (failing tests) and consistency/style errors (failing lint checks).

The lint job can run concurrently with the test job, so the overall cycle time shouldn't increase, unless it is frequently the case that Travis ends up scheduling the jobs in serial. We could also make the job smarter by exiting early if no front-end changes were made in a given branch.

chdorner:
> The lint job can run concurrently with the test job, so the overall cycle time shouldn't increase, unless it is frequently the case that Travis ends up scheduling the jobs in serial.

This is my only worry, the more jobs the more likely things will run in serial on Travis.

PR #4021, opened 2016-11-02
First version of group search page sidebar

seanh:
This adds a first version of the group sidebar on the `/groups/<pubid>/search` page. See <https://github.com/hypothesis/product-backlog/issues/12> for requirements.

What this includes:

- [x] Sidebar shows only on the `/groups/<pubid>/search` page
- [x] Sidebar does not show if no group with the given pubid exists
- [x] Sidebar does not show if the user isn't logged in or isn't a member of the group
- [x] Sidebar contains group name, description, num. annotations, created date, num. members, list of members sorted by num. annotations and num. annotations for each member
- [x] Multi-paragraph group descriptions are handled
- [x] Group name and description are escaped
- [x] Edit group link if you are the group's creator
- [x] Leave group button if you are _not_ the group's creator
- [x] Confirm dialog before leaving group (js only), the text that this dialog uses is the same as when leaving a group in the client sidebar

What this does not include:

* No tags in the sidebar, that's a [separate card](https://github.com/hypothesis/product-backlog/issues/10)
* No group invite link in the sidebar, that's to be done in a [follow-up PR](https://github.com/hypothesis/h/pull/4037)
* The members are not clickable, that's to be done in a follow-up PR
* No responsive design, full-sized desktop layout only, responsive is to be done in a [follow-up PR](https://github.com/hypothesis/h/pull/4025)

Known issues:

* The "<- Back to the groups page" link on the edit group page takes you back to the old group page, not to the page that you came from.

* <del>When you click on the _Leave group_ button any search times that you've typed that have 1. Been turned into lozenges by the JavaScript and 2. Not been submitted to the server yet (so they aren't in the URL yet) will be lost when the page reloads. I think this needs to be solved but propose to do it in a later PR.</del> (This is fixed now that <https://github.com/hypothesis/h/pull/4026> is merged.)

*  Hard line breaks in group descriptions:
   
   ```
   This is the first line
   This is the second line, of the first paragraph
   ```
   
   Will produce two separate paragraphs (two `<p>` elements), just as it would if there was an empty line between the two lines:
   
   ```
   This is the first paragraph
   
   This is the second paragraph
   ```

### DIFF
--- a/h/static/scripts/controllers/confirm-submit-controller.js
+++ b/h/static/scripts/controllers/confirm-submit-controller.js
@@ -1,0 +1,28 @@
+'use strict';
+
+var Controller = require('../base/controller');
+
+/* Turn a normal submit element into one that shows a confirm dialog.
+ *
+ * The element's form will only be submitted if the user answers the
+ * confirmation dialog positively.
+ *
+ */
+class ConfirmSubmitController extends Controller {
+  constructor(element, options) {
+    super(element);
+
+    var window_ = options.window || window;
+
+    element.addEventListener('click', (event) => {
+      if (!window_.confirm(element.dataset.confirmMessage)) {
+        event.preventDefault();
+        event.stopPropagation();
+        event.stopImmediatePropagation();
+        return;
+      }
+    }, /*capture*/ true);
+  }
+}
+
+module.exports = ConfirmSubmitController;
--- a/h/static/scripts/tests/controllers/confirm-submit-controller-test.js
+++ b/h/static/scripts/tests/controllers/confirm-submit-controller-test.js
@@ -1,0 +1,83 @@
+'use strict';
+
+var ConfirmSubmitController = require('../../controllers/confirm-submit-controller');
+var util = require('./util');
+
+describe('ConfirmSubmitController', function () {
+
+  var ctrl;
+
+  afterEach(function () {
+    if (ctrl) {
+      ctrl.element.remove();
+      ctrl = null;
+    }
+  });
+
+  /**
+   * Make a <button type="submit"> with the confirm submit controller
+   * enhancement applied and return the various parts of the component.
+   *
+   */
+  function component(windowConfirmReturnValue) {
+    var confirmMessage = "Are you sure you want to leave the group 'Test Group'?";
+    var template = `<button type="submit"
+      class="js-confirm-submit"
+      data-confirm-message="${confirmMessage}">
+    `;
+
+    var fakeWindow = {
+      confirm: sinon.stub().returns(windowConfirmReturnValue),
+    };
+
+    ctrl = util.setupComponent(document,
+                               template,
+                               ConfirmSubmitController,
+                               {window: fakeWindow});
+
+    return {
+      ctrl: ctrl,
+      fakeWindow: fakeWindow,
+      confirmMessage: confirmMessage,
+    };
+  }
+
+  function fakeEvent() {
+    var event = new Event('click');
+    event.preventDefault = sinon.stub();
+    event.stopPropagation = sinon.stub();
+    event.stopImmediatePropagation = sinon.stub();
+    return event;
+  }
+
+  it('shows a confirm dialog using the text from the data-confirm-message attribute', function () {
+    var {ctrl, fakeWindow, confirmMessage} = component(true);
+
+    ctrl.element.dispatchEvent(new Event('click'));
+
+    assert.calledOnce(fakeWindow.confirm);
+    assert.calledWithExactly(fakeWindow.confirm, confirmMessage);
+  });
+
+  it('prevents form submission if the user refuses the confirm dialog', function () {
+    var ctrl = component(false).ctrl;
+    var event = fakeEvent();
+
+    ctrl.element.dispatchEvent(event);
+
+    assert.called(event.preventDefault);
+    assert.called(event.stopPropagation);
+    assert.called(event.stopImmediatePropagation);
+  });
+
+  it('allows form submission if the user confirms the confirm dialog', function () {
+    var ctrl = component(true).ctrl;
+    var event = fakeEvent();
+
+    ctrl.element.dispatchEvent(event);
+
+    assert.isFalse(event.preventDefault.called);
+    assert.isFalse(event.stopPropagation.called);
+    assert.isFalse(event.stopImmediatePropagation.called);
+  });
+});
--- a/h/static/styles/partials-v2/_search-result-sidebar.scss
+++ b/h/static/styles/partials-v2/_search-result-sidebar.scss
@@ -52,6 +52,11 @@
   text-decoration: underline;
 }
 
+.search-result-sidebar__leave-button::-moz-focus-inner {
+  padding: 0;
+  border-width: 0;
+}
+
 .search-result-sidebar-title__annotations-count {
   font-size: $normal-font-size;
   font-weight: normal;
--- a/h/static/styles/partials-v2/_search-result-sidebar.scss
+++ b/h/static/styles/partials-v2/_search-result-sidebar.scss
@@ -1,0 +1,60 @@
+.search-result-sidebar {
+  width: 300px;
+}
+
+.search-result-sidebar__title {
+  margin-top: 0;
+  margin-bottom: 26px;
+}
+
+.search-result-sidebar__title,
+.search-result-sidebar__subtitle {
+  color: $brand;
+}
+
+.search-result-sidebar__section {
+  margin-bottom: 26px;
+}
+
+.search-result-sidebar__section p:first-of-type {
+  margin-top: 0; 
+}
+
+.search-result-sidebar__section p:last-of-type {
+  margin-bottom: 0;
+}
+
+.search-result-sidebar__dt {
+  font-weight: bold;
+  float: left;
+  clear: left;
+  margin-right: 3px;
+}
+
+.search-result-sidebar__dd {
+  float: left;
+  margin-left: 0;
+}
+
+.search-result-sidebar__link {
+  color: inherit;
+  text-decoration: underline;
+}
+
+.search-result-sidebar__link:hover {
+  text-decoration: underline;
+}
+
+.search-result-sidebar-title__annotations-count {
+  font-size: $normal-font-size;
+  font-weight: normal;
+}
+
+.search-result-sidebar__annotations-count {
+  color: $grey-5;
+}
+
+.search-result-sidebar__username {
+  font-weight: bold;
+  margin-right: 3px;
+}
--- a/h/static/styles/partials-v2/_search-result-sidebar.scss
+++ b/h/static/styles/partials-v2/_search-result-sidebar.scss
@@ -45,6 +45,13 @@
   text-decoration: underline;
 }
 
+.search-result-sidebar__leave-button {
+  background: none;
+  border: none;
+  padding: 0;
+  text-decoration: underline;
+}
+
 .search-result-sidebar-title__annotations-count {
   font-size: $normal-font-size;
   font-weight: normal;
--- a/h/static/styles/partials-v2/_search.scss
+++ b/h/static/styles/partials-v2/_search.scss
@@ -4,8 +4,9 @@
 // of activity queries.
 
 .search-result-container {
+  display: flex;
   font-size: $normal-font-size;
-  margin-top: 20px;
+  margin-top: 50px;
   color: $grey-6;
 }
 
@@ -28,6 +29,10 @@
   padding-bottom: 10px;
   border-bottom: 1px solid $grey-2;
   margin-top: 30px;
+}
+
+.search-result__timeframe:first-child {
+  margin-top: 0;
 }
 
 // A group of search results
--- a/h/static/styles/partials-v2/_search.scss
+++ b/h/static/styles/partials-v2/_search.scss
@@ -11,6 +11,7 @@
 }
 
 .search-result-list {
+  flex-basis: 950px;
   max-width: 950px;
 
   // Remove default padding from <ol>
--- a/h/templates/activity/search.html.jinja2
+++ b/h/templates/activity/search.html.jinja2
@@ -135,14 +135,18 @@
           {% endif %}
           {% if show_group_leave_button %}
             <li>
-              <button class="search-result-sidebar__leave-button"
-                      form="search-bar"
-                      formmethod="POST"
-                      type="submit"
-                      name="group_leave"
-                      value="{{ group.pubid }}">
-                  {% trans %}Leave this group{% endtrans %}
-              </button>
+              {#- This form element is needed for Internet Explorer because it
+                  doesn't support the `form` attribute. See commit message. #}
+              <form method="POST">
+                <button class="search-result-sidebar__leave-button"
+                        form="search-bar"
+                        formmethod="POST"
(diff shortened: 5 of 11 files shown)


h/static/scripts/controllers/confirm-submit-controller.js
```diff
@@ -0,0 +1,25 @@
...
+ *
+ * The element's form will only be submitted if the user answers the
+ * confirmation dialog positively.
+ *
+ */
+class ConfirmSubmitController extends Controller {
+  constructor(element) {
+    super(element);
+
+    element.onclick = (event) => {
```
sean-roberts: The problem that you are having is rooted in how this controller handles the events specifically propagating your click interaction. There are two ways to fix this:

```
// option 1
element.onclick = (event) => {
  event.preventDefault();
  if (window.confirm(element.dataset.confirmMessage)) {
    element.onclick = null;
    setTimeout(()=>{
        element.click();
    }, 0);
  }
};
```

```
// option 2
element.addEventListener('click', (e)=>{
  if (!window.confirm(element.dataset.confirmMessage)) {
    e.preventDefault();
    e.stopPropagation();
    e.stopImmediatePropagation();
    return;
  }
}, /*capture*/true);
```

Option 1 works because it is pulling the click event into another control flow. Why that fixes this is not immediately obvious but I am not a huge fan of this approach to begin with (I have seen this breakdown on android as well)

Option 2 takes this an opposite approach of what you currently have. Your current implementation says "I will trigger the continue to happen if they accept confirmation" and the proposed solution says "I will trigger a shut down of the event if they do not confirm." So combined with use capture phase, we will trigger the confirmation and if they fail we do not let anything else hear about the event. I would go this route as it doesn't have to worry about any event bindings and triggering.


Side note: If you ever need to override a method like onclick or on{EventName} from an abstract helper, make sure you don't break any other possible overrides from other abstractions.
```
// something along these lines would not cause any conflicts with other overrides
// potentially sneaking in some painfully hard to find bugs
(()=>{
  var prevHandler = element.onclick;
  element.onclick = function(e){
    // .... logic here
    if(prevHandler){
      prevHandler(e);
    }
  }
})();
```

seanh: *Huge* thanks, this fixes the issue and thanks for the explanation

seanh:
Ok, I think all this needs now is tests, and then I'll add a better PR description and it'll be ready for review...

h/static/scripts/controllers/confirm-submit-controller.js
```diff
@@ -0,0 +1,28 @@
...
+ *
+ * The element's form will only be submitted if the user answers the
+ * confirmation dialog positively.
+ *
+ */
+class ConfirmSubmitController extends Controller {
+  constructor(element, options) {
+    super(element);
+
+    var window_ = options.window || window;
```
seanh: This is a "test seam" (I saw this done with `window` elsewhere in the code as well, guessing by Rob)

h/static/scripts/controllers/confirm-submit-controller.js
```diff
@@ -0,0 +1,28 @@
...
+
+    element.addEventListener('click', (event) => {
+      if (!window_.confirm(element.dataset.confirmMessage)) {
+        event.preventDefault();
+        event.stopPropagation();
+        event.stopImmediatePropagation();
+        return;
+      }
+    }, /*capture*/ true);
```
seanh: Thanks, Sean Roberts!

h/static/scripts/controllers/confirm-submit-controller.js
```diff
@@ -0,0 +1,28 @@
...
+
+var Controller = require('../base/controller');
+
+/* Turn a normal submit element into one that shows a confirm dialog.
+ *
+ * The element's form will only be submitted if the user answers the
+ * confirmation dialog positively.
+ *
+ */
+class ConfirmSubmitController extends Controller {
```
seanh: As with another controller I wrote recently (copy clipboard button), this isn't a "React-style" controllers. It's not using update, state, etc. This is partially because I haven't gotten the hang of / internalised that pattern yet. But it's also because these controllers are so trivial, needing only a constructor function that adds an event listener, that it doesn't seem necessary.

h/templates/activity/search.html.jinja2
```diff
@@ -95,6 +95,82 @@
...
+              <a class="search-result-sidebar__link"
+                 href="{{ group_edit_url }}">
+                {% trans %}Edit group{% endtrans %}
+              </a>
+            </li>
+          {% endif %}
+          {% if show_group_leave_button %}
+            <li>
+              {#- This form element is needed for Internet Explorer because it
+                  doesn't support the `form` attribute. See commit message. #}
```
seanh: These comments get removed by Jinja2 in the HTML that the server renders the template to

h/static/scripts/controllers/confirm-submit-controller.js
robertknight: This is absolutely fine, the pattern isn't really applicable/useful in this context. It _is_ useful when the component has a non-trivial amount of state to manage, in which case it provides a pattern for how to store and update that state and how to keep the DOM in sync with that state.

h/templates/panels/navbar.html.jinja2
```diff
@@ -15,6 +15,13 @@
         {{ svg_icon('search', 'search-bar__icon') }}
         <div class="search-bar__lozenges" data-ref="searchBarLozenges"></div>
         <input class="search-bar__input" data-ref="searchBarInput" name="q" value="{{ q }}" placeholder="Search…" autocomplete="off">
+
+        {#- This seemingly pointless <input> is actually needed to make the
```
seanh: The empty line before this comment will also be stripped from the HTML by Jinja2

h/templates/panels/navbar.html.jinja2
```diff
@@ -15,6 +15,13 @@
...
+            form work correctly. If you remove this input then when the user
+            presses Enter in the search bar input, the browser will act as if
+            the user had clicked one of the form's other submit elements
+            (e.g. the leave group button). #}
+        <input type="submit" style="display: none;">
```
seanh: I actually felt that an inline style was appropriate in this one-off case?

h/static/scripts/controllers/confirm-submit-controller.js
robertknight: A couple of people have asked me about usage of const/let/var recently. I suggest that in new source files we follow the Airbnb style - `const` where possible, `let` otherwise. There are tools to automate updating existing code and I'll look into using those so we're consistent everywhere.

h/views/activity.py
```diff
@@ -42,9 +42,81 @@ def search(request):
...
+        raise httpexceptions.HTTPNotFound()
+
+    result = search(request)
+
+    pubid = request.matchdict['pubid']
+
+    try:
+        group = request.db.query(models.Group).filter_by(pubid=pubid).one()
+    except exc.NoResultFound:
+        return result
```
seanh: We don't want to 404 if the group that the user searches for does not exist (or if it does exist but the user is a member of it). Anyone can type `group:foo<Enter>` into the search box, and an empty search results page seems better than 404ing them.

seanh: It would probably be better if the groups service had a method for getting a group by its pubid, rarther than the view function accessing sqlalchemy itself here, but it doesn't and I'm not 100% sure that adding it won't be controversial and I'm in a bit of a hurry so I'm not fixing that here.

h/templates/panels/navbar.html.jinja2
robertknight: The only problem with this is that it prevents using CSP to disable inline styles as a mechanism to guard against style injection vulnerabilities in future. We have a set of utility classes with `u-*` prefixes defined for this kind of use case in `partials-v2/_util.scss`

h/views/activity.py
```diff
@@ -42,9 +42,81 @@ def search(request):
...
+
+    result['group'] = {
+        'created': group.created.strftime('%B, %Y'),
+        'description': group.description,
+        'name': group.name,
+        'pubid': group.pubid,
+    }
+
+    result['show_group_leave_button'] = (
+        group.creator != request.authenticated_user)
```
seanh: Another thing that I chose not to fix right now because the details seems likely to be controversial and I'm in a bit of a hurry: it might be nicer if the Pyramid permissions system can tell us whether the authenticated user has "permission" to leave the group (as it tells us whether they have permission to "admin" the group below), rather than encoding the "group creators cannot leave the group (but members can)+ business logic here in the view function.

h/views/activity.py
```diff
@@ -42,9 +42,81 @@ def search(request):
...
+    if request.has_permission('admin', group):
+        result['group_edit_url'] = request.route_url('group_edit', pubid=pubid)
+
+    return result
+
+
+@view_config(route_name='activity.group_search',
+             request_method='POST',
+             renderer='h:templates/activity/search.html.jinja2',
+             request_param='group_leave')
```
seanh: This means that this view function will be called if the browsers POSTs to the "activity.group_search URL (`/groups/<pubid>/search`) _and_ the post includes a param called "group_leave". GET requests to the same URL go to the `group_search()` function above. In future we will add more POST requests with different params instead of "group_leave", for example for adding user or tag facets into the search, and new functions like this one will be added to handle those.

h/static/scripts/controllers/confirm-submit-controller.js
seanh: Lets get the airbnb eslint config in place, then it will tell me to do this.

h/static/styles/partials-v2/_search.scss
```diff
@@ -4,12 +4,14 @@
 // of activity queries.
 
 .search-result-container {
+  display: flex;
```
seanh: This introduces a problem on small screens (small enough that the document buckets must shrink to smaller than their default size), the container ends up being wider than the screen and you get a horizontal scrollbar.

Fortunately setting `flex-direction: column;` on small screens, which I think I'm going to do in the follow-up PR that adds the responsive layout for this, fixes it. So I'll fix this in the follow-up PR.

seanh:
> When you click on the _Leave group_ button any search times that you've typed that have 1. Been turned into lozenges by the JavaScript and 2. Not been submitted to the server yet (so they aren't in the URL yet) will be lost when the page reloads. I think this needs to be solved but propose to do it in a later PR.

I think the solution to this, at least in non-IE browsers, will be for `SearchBarController` to insert a hidden input into the search bar form at controller initialization time and to update the value of that hidden input whenever it lozengifies a term. Then if any button outside of  `SearchBarController` submits the search bar form, all the lozenge values will already be in the form, without any need for communication between different js controllers or anything like that. This is how Sheetal had it working at first, but I got her to change it as I thought it was unnecessary. The way it works now is that `SearchBarController` doesn't insert its hidden `<input>` into the form until the moment that the form is submitted by `SearchBarController`, but this of course doesn't work if the form is submitted by a button that's outside of `SearchBarController`.

We're going to be adding a lot of buttons that will need this. I'll send a separate PR to fix it.

h/templates/panels/navbar.html.jinja2
seanh: Done

seanh:
@sean-roberts ping

seanh:
Just rebased this on master, no changes

h/static/scripts/tests/controllers/confirm-submit-controller-test.js
```diff
@@ -0,0 +1,85 @@
...
+      ctrl = null;
+    }
+  });
+
+  /**
+   * Make a <button type="submit"> with the confirm submit controller
+   * enhancement applied and return the various parts of the component.
+   *
+   */
+  function component(windowConfirmReturnValue) {
```
robertknight: Protip (no change required here): You can use destructuring as a slightly nicer way to do boolean parameters in ES6: `function component({windowConfirmReturnValue})`. Usage: `component({windowConfirmReturnValue: true})`. This is mostly useful if you have several such parameters.

h/static/scripts/tests/controllers/confirm-submit-controller-test.js
```diff
@@ -0,0 +1,84 @@
...
+    assert.calledWithExactly(fakeWindow.confirm, confirmMessage);
+  });
+
+  it('prevents form submission if the user refuses the confirm dialog', function () {
+    var ctrl = component(false).ctrl;
+    var event = fakeEvent();
+
+    ctrl.element.dispatchEvent(event);
+
+    assert.isTrue(event.preventDefault.called);
```
robertknight: Using `assert.called` rather than `assert.isTrue(stub.called)` will give you easier to read error messages if the check fails.

h/static/styles/partials-v2/_search-result-sidebar.scss
```diff
@@ -0,0 +1,67 @@
...
+
+.search-result-sidebar__section {
+  margin-bottom: 26px;
+}
+
+.search-result-sidebar__section p:first-of-type {
+  margin-top: 0; 
+}
+
+.search-result-sidebar__section p:last-child {
```
robertknight: Nit: Any reason this isn't symmetrical (`p:last-of-type`)?

h/templates/panels/navbar.html.jinja2
robertknight: 👍 For comments explaining non-obvious things.

h/views/activity.py
robertknight: Group creators can leave their own group in the client, so we shouldn't stop them from doing so here. If we decide we want to change the policy around leaving a group, we should do so in a way that makes it consistent whether leaving from the client or the service.

h/views/activity.py
robertknight: I had a query here as to why we were not reusing the existing view for leaving a group. Slack discussion: https://hypothes-is.slack.com/archives/frontend/p1478619891001151

h/static/scripts/tests/controllers/confirm-submit-controller-test.js
```diff
@@ -0,0 +1,84 @@
...
+   * enhancement applied and return the various parts of the component.
+   *
+   */
+  function component(windowConfirmReturnValue) {
+    var confirmMessage = "Are you sure you want to leave the group 'Test Group'?";
+    var template = [
+      '<button type="submit"',
+      '        class="js-confirm-submit"',
+      '        data-confirm-message="' + confirmMessage + '">',
+    ].join('\n');
```
seanh: Done

h/static/scripts/tests/controllers/confirm-submit-controller-test.js
seanh: Done

h/static/styles/partials-v2/_search-result-sidebar.scss
seanh: Done

h/views/activity.py
seanh: Done

robertknight:
group-sidebar

There is a small amount of padding around the 'Leave this group' button in Firefox.

LGTM otherwise and happy for this to be fixed separately.

seanh:
Thanks for the Firefox fix Rob, pushed it

PR #4023, opened 2016-11-03
Adds Help link to homepage

lenazun:
This adds a new link to the home page directing people to the new Help Center.


lenazun:
Yeah also this front page is not long for this world.

PR #4024, opened 2016-11-04
Fix extra space at end of search queries

### LINKED ISSUE #3985: Extra + at end of search query
Open the `/search` page, type "foo bar<Space>" into the search bar (it's important that you type the second space so that "bar" gets lozengified), hit `Enter`, the URL is `http://localhost:5000/search?q=foo+bar+`.

Expected: `http://localhost:5000/search?q=foo+bar`

chdorner:
Instead of looping through the lozenges and appending the content and a
trailing space, we just get an array of the lozenge contents and join
it. This way we avoid the extra space at the end.

Fixes #3985.

### DIFF
--- a/h/static/scripts/controllers/search-bar-controller.js
+++ b/h/static/scripts/controllers/search-bar-controller.js
@@ -132,10 +132,10 @@ class SearchBarController extends Controller {
       var hiddenInput = document.createElement('input');
       hiddenInput.type = 'hidden';
 
-      Array.from(this._lozengeContainer.querySelectorAll('.js-lozenge__content')).forEach((loz) => {
-        hiddenInput.value = hiddenInput.value + loz.textContent + ' ';
-      });
-      hiddenInput.value = hiddenInput.value + getTrimmedInputValue();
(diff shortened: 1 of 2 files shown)


seanh:
This has a conflict with <https://github.com/hypothesis/h/pull/4026>, I'll review and merge this once <https://github.com/hypothesis/h/pull/4026> is merged.

robertknight:
@seanh , @chdorner - This needs rebasing.

seanh:
I'm on it. As it happened https://github.com/hypothesis/h/pull/4026 changed this part of the code and actually fixed the bug, but Christof's fix is more elegant, so I am going to rebase and merge this anyway..

seanh:
Ah, it turns out this actually introduces a bug. In the case where there are some lozenges and there is some unlozengified text (so `getTrimmedInputValue()` returns a non-empty string) then the value that this puts in the hidden input will be missing a space in-between the last lozenge and the trimmed input value. (New tests that have been added to master since Christof wrote this PR catch this.)

The code currently on master doesn't have this bug and fixes the bug this was fixing.

Sorry Christof

PR #4025, opened 2016-11-04
Add responsive layout to the group search page sidebar

seanh:
<del>This is marked as WIP because it requires https://github.com/hypothesis/h/pull/4021 to be merged first and then will need a rebase.</del> (Done)

This adds responsive layout to the sidebar on the group search page. This is an MVP of this responsive layout, when you click on the "More info" button to open the sidebar it does a full page load and it does not show any animated "slide in" of the sidebar. It can be enhanced with JavaScript later (and this no-js behaviour kept for when js is not available), but I propose that this no-js version gets us 99% and is probably fine to ship with activity pages v1.

The way it works is:

* A CSS media query means that when the screen is too small to show both the search results and the sidebar together then only the search results are shown and the sidebar is `display: none;`

* Another media query means that on such small screens an additional `<nav>` containing the group name and a "More info" button is shown above the search results, instead of the sidebar

* Clicking on the "More info" button does a full page reload and adds an `&more_info` param to the URL. This param causes the server to render a version of the page that, if the screen is too small to show both the search results and the sidebar together, shows only the sidebar rather than only the search results (this is just a case of toggling a CSS class on/off using a Jinja2 `{% if %}`). It also replaces the "More info" button with a "Back" button.

* The "Back" button also does a full page reload and just removes the `&more_info` param so you get the first version of the page again.

### Mockup

Note that this mockup also contains a bunch of other new components that are not part of this PR:

30c697e2-9d3b-11e6-9193-c8995babfebf

#### Phone

screenshot from 2016-11-04 17-45-31

screenshot from 2016-11-04 17-46-00

#### Tablet

screenshot from 2016-11-04 17-46-33

screenshot from 2016-11-04 17-46-49

#### Desktop

When the screen is large enough the sidebar appears on the side as normal:

screenshot from 2016-11-04 17-47-27

### Transition

GIF of the transition to and from the "More info" page on my dev machine (we'll see how fast this is on production):

peek 2016-11-04 15-44

### Resizing the window

Whether you're on the search results or the sidebar version of the page, if you make the window bigger you will get the search results and the sidebar. This is relevant not only to people who resize windows (or maximize or fullscreen) but to bookmarking and then opening a bookmark, opening from browser history, sharing a link with someone else, using various OS and browser features to share tabs between phone and desktop, etc. There are many ways in which the `&more_info` version of the URL may get opened on a larger screen.

peek 2016-11-04 17-38

### Known issues

* In Internet Explorer, any search query is lost when you click on the "More info" or "Back" button. This does not happen in other browsers. I don't think it can be fixed in IE except by using JavaScript.

### DIFF
--- a/h/static/styles/partials-v2/_pager.scss
+++ b/h/static/styles/partials-v2/_pager.scss
@@ -1,7 +1,10 @@
 .pager {
   display: flex;
   flex-direction: row;
-  margin: 40px auto;
+  margin-top: 0;
+  margin-bottom: 40px;
+  margin-left: auto;
+  margin-right: auto;
   font-weight: bold;
   justify-content: center;
 }
--- a/h/static/styles/partials-v2/_search.scss
+++ b/h/static/styles/partials-v2/_search.scss
@@ -7,9 +7,7 @@
   display: none;
   align-items: center;
   justify-content: space-between;
-
   margin-bottom: 40px;
-  margin-right: 10px;
 }
 
 .search-result-nav__title {
@@ -242,5 +240,16 @@
   .search-bucket-stats {
     margin-left: 5px;
     margin-top: 10px;
+  }
+
+  .search-result-container {
+    // Align the left edge of the search result list with the left edge of the
+    // navbar content.
+    margin-left: 10px;
+    margin-right: 10px;
+  }
+
+  .search-result-list {
+    margin-right: 0;
   }
 }
--- a/h/static/styles/partials-v2/_search.scss
+++ b/h/static/styles/partials-v2/_search.scss
@@ -8,6 +8,10 @@
   font-size: $normal-font-size;
   margin-top: 50px;
   color: $grey-6;
+
+  // Align the left edge of the search result list with the left edge of the
+  // navbar content.
+  margin-left: 30px;
 }
 
 .search-result-list {
@@ -19,10 +23,8 @@
   padding-left: 0;
   padding-right: 0;
 
-  // These margins are set so that the left edge of the search result list
-  // aligns with the left edge of the navbar content
-  margin-left: 30px;
   margin-right: 30px;
+  margin-bottom: 40px;
 }
 
 .search-result__timeframe {
@@ -174,9 +176,17 @@
     }
 }
 
+// As the screen size gets smaller, just before it would have introduced a
+// horizontal scrollbar move the sidebar below the search results.
+@media screen and (max-width: $tablet-width + 225px) {
+  .search-result-container {
+    flex-direction: column;
+  }
+}
+
 // On large tablets and below left-align annotation cards in order to make more
 // space for the stats to the right of the annotation card.
-@media screen and (max-width: $tablet-width + 150px) {
+@media screen and (max-width: 1127px) {
   .search-result-bucket__annotation-cards-container {
     padding-left: 0;
   }
--- a/h/static/styles/partials-v2/_search.scss
+++ b/h/static/styles/partials-v2/_search.scss
@@ -230,13 +230,6 @@
 // On normal tablets and below, display annotation stats below annotation list
 @media screen and (max-width: $tablet-width) {
 
-  // Reduce margins on tablet and below to match navbar left/right margins
-  // on these screen sizes
-  .search-result-list {
-    margin-left: 10px;
-    margin-right: 10px;
-  }
-
   .search-result-bucket__annotation-cards-container {
     flex-direction: column;
   }
--- a/h/static/styles/partials-v2/_search.scss
+++ b/h/static/styles/partials-v2/_search.scss
@@ -3,6 +3,32 @@
 // Styles for the search page which displays the results
 // of activity queries.
 
+.search-result-nav {
+  display: none;
+  align-items: center;
+  justify-content: space-between;
+
+  margin-bottom: 40px;
+  margin-right: 10px;
+}
+
+.search-result-nav__title {
+  color: $brand;
+  margin: 0;
+}
+
+.search-result-nav__button {
+  background-color: $grey-2;
+  border: none;
+  color: inherit;
+  font-weight: bold;
+  padding-top: 10px;
+  padding-bottom: 10px;
+  padding-left: 15px;
+  padding-right: 15px;
+  text-decoration: none;
+}
+
 .search-result-container {
   display: flex;
   font-size: $normal-font-size;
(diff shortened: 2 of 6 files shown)


robertknight:
The line height could use adjusting slightly. If the group name wraps, the lines run into one another:

group-profile-page-mobile

h/static/styles/partials-v2/_search.scss
```diff
@@ -230,13 +230,6 @@
 // On normal tablets and below, display annotation stats below annotation list
 @media screen and (max-width: $tablet-width) {
 
-  // Reduce margins on tablet and below to match navbar left/right margins
-  // on these screen sizes
-  .search-result-list {
-    margin-left: 10px;
-    margin-right: 10px;
-  }
```
robertknight: The purpose of this was to _reduce_ the margins of the search result list on small (phone) sized devices. The result should be that the left and right edges of annotation cards are horizontally aligned with the navbar at the top.

Failing to do this wastes horizontal space on small screens.

tests/h/views/activity_test.py
```diff
@@ -208,6 +225,46 @@ def search(self, patch):
         return search
 
 
+@pytest.mark.usefixtures('routes')
+class TestGroupSearchMoreInfo(object):
+
+    def test_it_redirects_to_group_search(self, pyramid_request):
+        """It should redirect and preserve the search query param."""
+        pyramid_request.matchdict['pubid'] = 'test_pubid'
+        pyramid_request.POST = {'q': 'foo bar'}
```
robertknight: Shouldn't the POST body include a `more_info` param here?

h/static/styles/partials-v2/_search.scss
```diff
@@ -181,6 +207,15 @@
...
   .search-result-container {
     flex-direction: column;
+    margin-top: 25px;
+  }
+
+  .search-result-nav {
+    display: flex;
+  }
+
+  .hide-on-small-screens {
```
robertknight: Hmm ... this class doesn't follow our naming conventions - where every class is defined in a file that starts with a prefix of the class name.

h/static/styles/partials-v2/_search.scss
seanh: Fixed

tests/h/views/activity_test.py
seanh: It actually doesn't make any difference in the test. The `more_info` param is only used by Pyramid to select which view callable to call, the view callable itself doesn't actually use it. And in the test, we don't use Pyramid's view selection mechanism, we just call the view function directly.

But it makes the test more realistic so I'll add it anyway

h/templates/activity/search.html.jinja2
```diff
@@ -83,7 +83,28 @@
   {{ panel('navbar') }}
 
   <div class="search-result-container">
-    <ol class="search-result-list">
+    {% if group %}
```
seanh: Done

h/static/styles/partials-v2/_search.scss
seanh: Done

seanh:
@robertknight all done I think

seanh:
I found one bug which I'd like to open a separate issue for - the clickable usernames don't work on mobile (!):

peek 2016-11-10 14-34

Pyramid is probably selecting the wrong view, hopefully easy enough to fix

h/templates/activity/search.html.jinja2
```diff
@@ -2,6 +2,33 @@
 
 {% from '../includes/annotation_card.html.jinja2' import annotation_card %}
 
+{% macro search_result_nav(group_name) %}
+<nav class="search-result-nav">
```
robertknight: This makes it much easier to follow the overall structure 👍

h/static/styles/partials-v2/_pager.scss
```diff
@@ -1,7 +1,10 @@
 .pager {
   display: flex;
   flex-direction: row;
-  margin: 40px auto;
```
robertknight: _NTS: This is just expanding shorthand into multiple properties_

h/templates/activity/search.html.jinja2
```diff
@@ -2,6 +2,33 @@
...
+  {%- if more_info %}
+    <form method="POST"> {# This <form> is needed for IE because it doesn't
```
robertknight: We're starting to repeat this pattern in quite a few places. You might want to factor it into a macro.

h/static/styles/partials-v2/_search.scss
```diff
@@ -174,9 +201,26 @@
...
+
+  .search-result-hide-on-small-screens {
+    display: none;
+  }
+}
+
 // On large tablets and below left-align annotation cards in order to make more
 // space for the stats to the right of the annotation card.
-@media screen and (max-width: $tablet-width + 150px) {
+@media screen and (max-width: 1127px) {
```
seanh: I guess I don't really get the point of using `$tablet-width` here, given that we're then going to have to add an arbitrary hardcoded value to it. 1127px is the exact right width at which this needs to come into effect

h/templates/activity/search.html.jinja2
seanh: Hmm, yeah

h/views/activity.py
```diff
@@ -84,12 +84,41 @@ def group_search(request):
...
+    result['more_info'] = 'more_info' in request.params
+
     return result
 
 
 @view_config(route_name='activity.group_search',
              request_method='POST',
              renderer='h:templates/activity/search.html.jinja2',
+             request_param='more_info')
+def group_search_more_info(request):
```
robertknight: I suspect quite a bit of this is going to end up duplicated when the user profile sidebar is implemented. If we can find a way to avoid too much duplication when that is implemented it would be helpful.

h/templates/activity/search.html.jinja2
```diff
@@ -105,7 +136,7 @@
     </ol>
 
     {% if group %}
-    <aside class="search-result-sidebar">
+    <aside class="search-result-sidebar {%- if not more_info %} search-result-hide-on-small-screens{% endif %}">
```
seanh: I don't know yet whether we're gonna want to change how the list of members is computed. The sidebar could certainly become a simple macro already though. Although yes - follow up PR. We're accruing several refactorings that I want to do in follow-up PRs, once the rush to get this feature complete and into people's hands for testing is over

h/views/activity.py
seanh: Sure. This file is one place where I'm making technical debt that I want to refactor once the rush to get all the features in is over. Grouping all these tightly related views into a Pyramid view class will remove a lot of boilerplate and duplication, and something can probably be done to avoid user - group sidebar duplication as well

PR #4026, opened 2016-11-04
Update hidden input live

seanh:
Change SearchBarController to insert its hidden input into the DOM at
controller initialization and keep its value up to date "live" as text
is lozengified, lozenges are deleted, and text is entered into the
visible <input>.

Previously SearchBarController didn't generate the value for the hidden
input, and didn't insert the hidden input into the form, until the form
was submitted.

The problem with this is that we now want to add various new submit
buttons to this form that are outside of SearchBarController's element,
and those submit buttons would end up submitting an empty q value as
SearchBarController would have removed search text from the <input> and
turned it into lozenges, but since SearchBarController was not
submitting the form it would not insert its hidden input (containing the
values from the lozenges).

As long as SearchBarController always keeps the value of its hidden
input up to date then the form can be submitted in the normal way
(without JavaScript) at any time by any submit button and the full value
of the query will be submitted.

SearchBarController no longer does anything at submit time.

### DIFF
--- a/h/static/scripts/controllers/search-bar-controller.js
+++ b/h/static/scripts/controllers/search-bar-controller.js
@@ -117,10 +117,7 @@ class SearchBarController extends Controller {
 
 
     /**
-     * Insert a hidden <input> into the search <form>.
-     *
-     * The value of the hidden <input> is a search string constructed from
-     * the lozenges plus any text currently in the visible <input>.
+     * Insert a hidden <input> with an empty value into the search <form>.
      *
      * The name="q" attribute is moved from the visible <input> on to the
      * hidden <input> so that when the <form> is submitted it's the value of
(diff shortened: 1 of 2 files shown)


seanh:
@sheetaluk Can you review this?

I should add some tests since it's now a requirement that `SearchBarController` update the hidden input live and not just before form submit, and we don't have any tests for that. But I've run out of time this week.

h/static/scripts/controllers/search-bar-controller.js
```diff
@@ -132,30 +129,31 @@ class SearchBarController extends Controller {
...
-     * this._input with the search terms from the lozenge before submitting it,
-     * we update a hidden input and submit that instead.
      */
-    var submitForm = () => {
-      insertHiddenInput();
-      this.refs.searchBarForm.submit();
+    var updateHiddenInput = () => {
+      let newValue = '';
+      Array.from(this._lozengeContainer.querySelectorAll('.js-lozenge__content')).forEach((loz) => {
+        newValue = newValue + loz.textContent + ' ';
```
chdorner: I've changed this in #4024, we'll have to coordinate the two pull requests.

seanh: Yes, if you don't mind I'd like to hang on to <https://github.com/hypothesis/h/pull/4024>, I'll review and merge it after this one is merged. The reason is that this PR is blocking a lot of feature work on activity pages that is due at the end of this sprint.

seanh:
Ok, I've added tests to this now

h/static/scripts/controllers/search-bar-controller.js
```diff
@@ -226,23 +225,20 @@ class SearchBarController extends Controller {
           // appended to the container element.
           event.preventDefault();
           this._input.value = '';
+          updateHiddenInput();
         }
       };
 
-      var handleEnterKey = (event) => {
```
robertknight: Note to self: We don't need this any more because the hidden input is always kept up to date.

PR #4031, opened 2016-11-07
Update "embedding" docs

nickstenning:
This commit removes some half-written documentation about how to
override the sidebar application, and updates the documentation about
configuring the sidebar embed to be current.


seanh:
Much nicer

PR #4032, opened 2016-11-07
Add a "For Publishers" section to the docs

nickstenning:
This is primarily so I have somewhere sensible to put documentation of the JWT format used for third-party accounts.


chdorner:
LGTM.

PR #4033, opened 2016-11-07
Support CORS requests for POST /api/token endpoint

chdorner:
I'm using a similar CORS policy to the one we use for the memex endpoints, except that we (for now) only allow the `Authorization` and `Content-Type` headers which should be enough for the h side of the API.

### DIFF
--- a/h/util/view.py
+++ b/h/util/view.py
@@ -11,7 +11,8 @@ cors_policy = cors.policy(
         'Authorization',
         'Content-Type',
     ),
-    allow_methods=('HEAD', 'GET', 'POST', 'PUT', 'DELETE'))
+    allow_methods=('HEAD', 'GET', 'POST', 'PUT', 'DELETE'),
+    allow_preflight=True)
 
 
 def handle_exception(request):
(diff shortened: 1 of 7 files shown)


chdorner:
Good point Robert! It looks like some of the memex views were supporting preflight requests rather by accident because the request methods were not defined in the view config `memex.views.index`, `memex.views.search`, etc.) and others were returning a 404 because they specifically defined request methods but `OPTIONS` wasn't one of them.

I've changed the memex CORS policy to be able to respond to OPTIONS requests automatically (configurable, default is off for the moment), and then I changed the two decorators (h and memex) to always add OPTIONS to the request methods.

robertknight:
LGTM. As discussed on Slack, this still doesn't quite enable third-party browser based clients to actually do useful things like create annotations because the `OPTIONS` preflight request that is made when performing a `POST /api/annotations` request fails with a 404 before the CORS decorator is even called.

Still, this is a prerequisite that gets us closer to supporting such clients. See https://hypothes-is.slack.com/archives/public/p1478860818007179 for more discussion.

PR #4034, opened 2016-11-07
Tag suggestions

sean-roberts:
Work for: https://trello.com/c/glm03Us8/1-add-tag-suggestions-to-suggestion-box

Having a component need both static data and dynamic data, I opted to set a global object window.h to namespace anything that we need like that. This is meant to make it available and not assume that the script is loaded already. Once the script is loaded, it will pick it up and render the list items. I did not see an example for which to base a solution for dynamic data - so if there is a pattern that we would prefer, let me know.

Also, I figured there would be a nice way jinja template handle to pull in a script file on demand. I looked and tried a lot of different things based on what our templates already do, and I could not come out with something that wasn't hacky as hell. So, I opted to just put the script tag in the template and we will figure out a more elegant solution if this pattern/need is more common than just this one use case. 

The tests and the unicode pattern that is used is pulled straight from unicode.coffee in the client repo.


robertknight:
> I did not see an example for which to base a solution for dynamic data - so if there is a pattern that we would prefer, let me know.

We have a couple of existing approaches to getting dynamic data into the rendered HTML:

1) Using HTML attributes (eg. with either base-64 or quote-escaped markup) on the controller's root element (eg. `data-suggestions="<blob of escaped data, eg in JSON format>"`)
2) Using `<script>` tags containing JSON data. See how `app_config` is rendered in `app.html.jinja2`. The `<script>` tag is given a known class which the consumer of the data then uses to find and load it (see `settings.js` in the client)

For the second approach, there are two reasons for using a script tag containing JSON data rather than an inline JS script:

1. It makes correct escaping easy. You can either generate a string of JSON in Python code and mark it as escaped using `variable | safe` in a template, or just pipe a variable (dict, list etc.) into the `to_json` filter in a template using `variable | to_json`.
2. It enables us to block inline JS scripts using Content Security Policy on the production website as a security measure.

robertknight:
> Also, I figured there would be a nice way jinja template handle to pull in a script file on demand.
> I looked and tried a lot of different things based on what our templates already do, and I could
> not come out with something that wasn't hacky as hell.

You didn't miss anything - we don't do any lazy loading of scripts in the client or on the site yet, although we have several use cases for it (eg. on the client only loading KaTeX and unorm when actually needed).

h/static/scripts/controllers/search-bar-controller.js
```diff
@@ -4,6 +4,12 @@ var Controller = require('../base/controller');
 var LozengeController = require('./lozenge-controller');
 var AutosuggestDropdownController = require('./autosuggest-dropdown-controller');
 var SearchTextParser = require('../util/search-text-parser');
+var stringUtil = require('../util/string');
+var escapeHtml = require('escape-html');
```
robertknight: In both the Python and JS code, we follow a pattern of putting vendor imports at the top of the file, followed by a blank line, then local requires.

h/static/scripts/controllers/search-bar-controller.js
```diff
@@ -207,12 +117,174 @@ class SearchBarController extends Controller {
...
+        title: escapeHtml(item.tag), // make safe
+        matchOn: stringUtil.fold(stringUtil.normalize(item.tag)),
+        usageCount: item.count || 0,
+      });
+    });
+
+
+    this._suggestionsHandler = new AutosuggestDropdownController( this._input, {
+
+      list: [].concat(explanationList, tagsList),
```
robertknight: `concat` returns a copy of the array rather than mutating the original, so `explanationList.concat(tagsList)` would be equivalent here.

h/static/scripts/tests/controllers/search-bar-controller-test.js
```diff
@@ -83,6 +88,79 @@ describe('SearchBarController', function () {
         });
     });
 
+    it('it shows suggestions for tags when face is selected', function (done) {
```
robertknight: Typo: face => "tags facet"?

h/static/scripts/tests/util/string-test.js
```diff
@@ -18,4 +18,56 @@ describe('util/string', function () {
       assert.equal(stringUtil.unhyphenate('-foo-bar-baz'), 'FooBarBaz');
     });
   });
+
+  describe('stringUtil helpers', function(){
+
+    it('removes hungarian marks', function(){
```
robertknight: Note to self: These were ported across from the tests in `unicode-test.coffee` in the client.

h/static/scripts/util/string.js
```diff
@@ -19,7 +24,39 @@ function unhyphenate(name) {
...
+ *
+ * @param  {String}
+ * @returns {String}
+ */
+function normalize(str){
+
+  // This require is coming from a vendor bundle
+  // that is loaded globally on the running webpage
+  // not a require in the node context
+  if(!require || !require('unorm') || !String.prototype.normalize){
```
robertknight: Couple of issues:

1. If `require` is not defined, this would result in a `ReferenceError` exception because of strict mode. Given that we are using CommonJS, I think we can just assume the existence of require as a function here, so we can just remove that check.

2. If `require()` fails here, I think it would throw an exception rather than returning false (but please check, I may be mistaken), so you'd need to use `try..catch`.

h/static/scripts/controllers/search-bar-controller.js
```diff
@@ -207,12 +117,174 @@ class SearchBarController extends Controller {
...
+    ].map((item)=>{ return Object.assign(item, { type: FACET_TYPE}); });
+
+    // tagSuggestions are made available by the scoped template data.
+    // see search.html.jinja2 for definition
+    let tagSuggestions = window.h && window.h.tagSuggestions;
+
+    let tagsList = (tagSuggestions || []).map((item)=>{
+      return Object.assign(item, {
+        type: TAG_TYPE,
+        title: escapeHtml(item.tag), // make safe
```
robertknight: This is fine, although I'm concerned that there will be a mixup with escaping in future, where something escaped for HTML is incorrectly used in a different context or some transformation gets applied to the string which breaks the escaping (on the Python side, this hazard is avoided by wrapping escaped strings using a special `Markup` class).

h/static/scripts/controllers/search-bar-controller.js
sean-roberts: I can update. But I did this intentionally because I think it reads more easily and leaves the slight lack of tribal knowledge question "is tagsList is being added to the explanationList?" out of the way

sean-roberts:
As mentioned in slack the losing of search term values was a regression in master which got merged into this branch. I fixed it in this branch though

h/static/scripts/controllers/search-bar-controller.js
```diff
@@ -207,12 +117,183 @@ class SearchBarController extends Controller {
...
+
+    // tagSuggestions are made available by the scoped template data.
+    // see search.html.jinja2 for definition
+    let tagSuggestions = document.querySelector('.js-tag-suggestions');
+
+    if(tagSuggestions){
+      try{
+        tagSuggestions = JSON.parse(tagSuggestions.innerHTML.trim());
+      }catch(e){
+        console.error('Could not parse .js-tag-suggestions JSON content', e);
```
robertknight: If finding the tag succeeds but parsing the JSON fails this will lead to some confusing errors below because `.map` will be called on an `Element`. I would suggest pulling this out into a helper function which takes the selector for the suggestions tag and returns `Object|Array|null`

h/static/scripts/util/string.js
```diff
@@ -19,7 +24,45 @@ function unhyphenate(name) {
...
+ * @param  {String}
+ * @returns {String}
+ */
+function normalize(str){
+
+  // This require is coming from a vendor bundle
+  // that is loaded globally on the running webpage
+  // not a require in the node context
+  try{
+    require('unorm');
```
robertknight: Great, this is much clearer and paves the way to loading unorm only if needed.

h/templates/activity/search.html.jinja2
```diff
@@ -82,6 +82,14 @@
 
   {{ panel('navbar') }}
 
+  {% for url in asset_urls("search_js") %}
+    <script src="{{ url }}"></script>
```
robertknight: Unfortunately this is still going to block HTML parsing during download. If we add the `defer` attribute here and on the script tags which depend on it (in `base.html.jinja2`) this will avoid the problem since deferred scripts execute after HTML parsing and in-order.

Perhaps it would be best to revisit this after this PR is merged though.

h/static/scripts/tests/controllers/search-bar-controller-test.js
```diff
@@ -83,6 +92,82 @@ describe('SearchBarController', function () {
         });
     });
 
+    it('it shows suggestions for tags when facet is selected', function (done) {
```
robertknight: Hmm... we're testing several behaviors in a single test case here. We generally try to only check one behavior in a particular `it()` test, which tends to make debugging easier when tests fail. Where several test cases share common setup, that can be grouped into `beforeEach()` within a nested `describe()` block.

h/static/scripts/controllers/autosuggest-dropdown-controller.js
```diff
@@ -84,6 +84,10 @@ class AutosuggestDropdownController extends Controller {
     });
 
     this._setList(configOptions.list);
+
+
+    // Public API
+    this.setHeader = this._setHeader;
```
robertknight: Any reason not to just rename `_setHeader` to `setHeader`?

h/static/scripts/controllers/search-bar-controller.js
```diff
@@ -1,9 +1,15 @@
 'use strict';
 
+var escapeHtml = require('escape-html');
```
robertknight: Need a new line after the vendor bundle list.

h/static/scripts/controllers/autosuggest-dropdown-controller.js
sean-roberts: I'd prefer to not mix internals and public api. Or rather, make it super obvious which is which

h/static/scripts/controllers/search-bar-controller.js
sean-roberts: I just changed the logic to not reuse a variable. I don't like doing that for this reason- didn't think about it while jumping around though.

h/templates/activity/search.html.jinja2
sean-roberts: I will add defer to this script as it is not required that we have it immediately

h/static/scripts/controllers/search-bar-controller.js
sean-roberts: lol I did that and then removed it thinking you would say not to have two lines :P

sean-roberts:
@robertknight updated. Should be good to go now

robertknight:
I'm still seeing an issue where the hidden "q" input is not always updated after accepting a suggestion. 

**Steps to reproduce:**

1. Create an annotation with the tag "aaa"
2. Go to `/search`, click in the search box and type "tag:a"
3. Press [down], then [enter]

**Expected:** Hidden "q" input has value "tag:aaa"
**Actual:** Hidden "q" input has value "tag:a"

My initial efforts to write a failing test for this were frustrated by the fact that 'syn' triggers the "input" event in cases where real browsers do not (See https://github.com/bitovi/syn/issues/119 and https://github.com/bitovi/syn/issues/120). In both cases that causes the "input" event handler in `SearchBarController` to be called after accepting a tag suggestion when it isn't in real browsers.

h/static/scripts/tests/controllers/search-bar-controller-test.js
```diff
@@ -83,6 +134,61 @@ describe('SearchBarController', function () {
         });
     });
 
+    describe('it allows tag value suggestions', function () {
```
robertknight: This is easier to follow and extend now that the individual test cases have been split up 👍

h/static/scripts/tests/controllers/search-bar-controller-test.js
```diff
@@ -83,6 +134,61 @@ describe('SearchBarController', function () {
...
+      beforeEach(function(){
+        // we need to setup the env vars before invoking controller
+        teardown();
+        addTagSuggestions();
+        setup();
+
+        sinon.stub(testEl.querySelector('form'), 'submit');
+      });
+
+      it('shows tag suggestsions', function(done){
```
robertknight: Typo: 'suggestsions'

h/templates/activity/search.html.jinja2
```diff
@@ -82,6 +82,14 @@
 
   {{ panel('navbar') }}
 
+  {% for url in asset_urls("search_js") %}
+    <script src="{{ url }}" defer></script>
```
robertknight: If you use `defer` here you'll also need to add it to other scripts which depend on this one - ie. the "site_js" bundle. The browser guarantees to execute deferred scripts in the order that it encounters them, but only once parsing of all HTML (including parsing and evaluation of any "ordinary" blocking `<script>` tags) is complete.

Hence the console error message about `require('unorm')` failing in this branch.

h/static/scripts/util/string.js
```diff
@@ -19,7 +24,45 @@ function unhyphenate(name) {
...
+
+/**
+ * Utility function to normalize a string based on our standard pattern of NFKD
+ *
+ * @param  {String}
+ * @returns {String}
+ */
+function normalize(str){
+
+  // This require is coming from a vendor bundle
```
robertknight: Probably worth adding a brief comment that "unorm" provides a `String.prototype.normalize` polyfill

h/templates/activity/search.html.jinja2
sean-roberts: You're right. Throwing a console.error in combination with the compromise that the module might not actually be there is not a good plan. I am just going to back out the defer approach now and move on here.

robertknight:
Populating the hidden input with accepted suggestions and loading unorm now works, however handling of Unicode normalization and folding when typing in the search box isn't.

**Steps to reproduce:**

1. Create an annotation with the tag "éffort" (note accented "é")
2. In the Hypothesis client, search for either "tag:effort" or "tag:éffort" - both should filter the annotations in the sidebar to the annotation created in step (1)
3. On `/search`, type in "tag:e" (note use of non-accented "é").

Expected: The tag "éffort" is suggested and available to select
Actual: The tag "éffort" disappears from the suggestion list after typing "e"

PR #4036, opened 2016-11-08
Add preliminary documentation of JWT authorization grants

nickstenning:


docs/publishers/authorization-grant-tokens.rst
```diff
@@ -0,0 +1,145 @@
...
+This section contains complete example code for generating a JWT in various
+common programming environments.
+
+Python
+``````
+
+We recommend using `PyJWT <https://pyjwt.readthedocs.io/en/latest/>`_::
+
+   import datetime
+   import pyjwt
```
robertknight: I presume this should be `import jwt`, since you're using `jwt.encode` below?

robertknight:
LGTM. I fixed a trivial typo (`import pyjwt` => `import jwt`)

docs/publishers/authorization-grant-tokens.rst
```diff
@@ -0,0 +1,145 @@
...
+
+   require 'jwt'
+
+   # IMPORTANT: replace these values with those for your client account!
+   CLIENT_AUTHORITY = 'customwidgets.com'
+   CLIENT_ID        = '4a2fa3b4-c160-4436-82d3-148f602c9aa8'
+   CLIENT_SECRET    = '5SquUVG0Tpg57ywoxUbPPgjtK0OkX1ttipVlfBRRrpo'
+
+   def generate_grant_token(username)
+     now = Time.now.to_i
```
chdorner: This should be `Time.now.utc.to_i`

nickstenning: I don't think it makes any difference. `to_i` turns the time into the number of seconds since the epoch, which is defined as UTC.

PR #4037, opened 2016-11-08
Group invite component

seanh:
Add a new reusable group invite "panel" and add it to the group search sidebar.

screenshot from 2016-11-08 14-12-59

### DIFF
--- a/h/static/images/icons/copy_to_clipboard.svg
+++ b/h/static/images/icons/copy_to_clipboard.svg
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<svg class="share-link-field__clipboard-image" version="1.1"
+<svg class="group-invite__clipboard-image" version="1.1"
   xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
   <g stroke="none" stroke-width="1" fill-rule="evenodd">
     <g>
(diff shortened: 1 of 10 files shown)


seanh:
This is WIP because it depends on https://github.com/hypothesis/h/pull/4021

robertknight:
Vertical alignment of the copy icon is correct in Chrome but slightly off in Firefox:

firefox-invite-panel

h/static/images/icons/copy_to_clipboard.svg
```diff
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<svg class="share-link-field__clipboard-image" version="1.1"
```
robertknight: _The old class name is not referenced in the service code._

h/static/styles/partials-v2/_group-invite.scss
```diff
@@ -0,0 +1,63 @@
...
+  margin-bottom: 5px;
+
+  width: 13px;
+  height: 16px;
+
+  fill: $grey-4;
+}
+
+.group-invite__clipboard-image:hover {
+  fill: $grey-7;
```
robertknight: This works in Chrome but not Firefox. It looks like the `fill` rule in the non-hovered state does work, but the `:hover` pseudo-class is not triggered.

I haven't investigated deeply, but this hack works around the problem by testing for the button rather than the `<svg>` element being hovered:

```diff
-.group-invite__clipboard-image:hover {
+button:hover > .group-invite__clipboard-image {
   fill: $grey-7;
 }
```

seanh: Done

seanh:
@robertknight all done I think

robertknight:
LGTM

PR #4038, opened 2016-11-08
Add username from url as a lozenge to the nav bar.

sheetaluk:
User and Group profile pages are not adding the facet to the search box on load.
This is because a query 'q' from which lozenges are created and which contains a
single group or user term is redirected to the specific group or user search page
with that term removed.

https://github.com/hypothesis/product-backlog/issues/16

### DIFF
--- a/h/templates/panels/navbar.html.jinja2
+++ b/h/templates/panels/navbar.html.jinja2
@@ -4,6 +4,19 @@
 
 {% if feature('activity_pages') %}
 
+{% macro lozenge(facetName, facetValue) %}
+  <div class="lozenge">
+    <div class="lozenge__content">
+      <span class="lozenge__facet-name">
+        {{ facetName }}:
+      </span>
+      <span class="lozenge__facet-value">
+        {{ facetValue }}
+      </span>
+    </div>
+  </div>
+{% endmacro %}
+
 <header class="nav-bar">
   <div class="nav-bar__content">
     <a href="/" title="Hypothesis homepage" class="nav-bar__logo-container"><!--
@@ -15,6 +28,14 @@
             id="search-bar"
             action="{{ search_url }}">
         {{ svg_icon('search', 'search-bar__icon') }}
+        <div class="search-bar__lozenges">
+          {% if opts['search_username'] %}
+            {{ lozenge('user', opts['search_username']) }}
+          {% endif %}
+          {% if opts['search_groupname'] %}
+            {{ lozenge('group', opts['search_groupname']) }}
+          {% endif %}
+        </div>
         <div class="search-bar__lozenges" data-ref="searchBarLozenges"></div>
         <input class="search-bar__input" data-ref="searchBarInput" name="q" value="{{ q }}" placeholder="Search…" autocomplete="off">
 
--- a/h/views/activity.py
+++ b/h/views/activity.py
@@ -20,9 +20,6 @@ PAGE_SIZE = 200
 @view_config(route_name='activity.search',
              request_method='GET',
              renderer='h:templates/activity/search.html.jinja2')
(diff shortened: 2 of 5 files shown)


h/static/styles/partials-v2/_lozenge.scss
```diff
@@ -1,3 +1,4 @@
+.special-lozenge,
```
robertknight: I thought the plan was to just use the existing `lozenge` class and avoid defining a new `special-lozenge` class, since we don't need to customize the appearance at this stage.

h/templates/panels/navbar.html.jinja2
```diff
@@ -13,6 +13,20 @@
             id="search-bar"
             action="{{ search_url }}">
         {{ svg_icon('search', 'search-bar__icon') }}
+        <div class="search-bar__special-lozenges">
```
robertknight: I would suggest pulling this out into a helper macro to clearly indicate that it is a separate component, and also adding a comment to the effect that it is the same markup as generated by `LozengeController` on the client.

h/templates/panels/navbar.html.jinja2
```diff
@@ -13,6 +13,20 @@
             id="search-bar"
             action="{{ search_url }}">
         {{ svg_icon('search', 'search-bar__icon') }}
+        <div class="search-bar__lozenges">
+          {% if opts['search_username'] %}
```
robertknight: I would suggest we extract this into a helper `macro()` and add a comment to the effect that it is a static version of the UI generated by `LozengeController`

sheetaluk:
@robertknight done!

tests/h/views/activity_test.py
```diff
@@ -266,6 +261,35 @@ def pyramid_request(self, group, pyramid_request):
...
         pyramid_config.add_route('activity.search', '/search')
 
+class TestUserSearch(object):
+    def test_it_returns_404_when_feature_turned_off(self,
+                                                    pyramid_request):
+        pyramid_request.feature.flags['search_page'] = False
+
+        with pytest.raises(httpexceptions.HTTPNotFound):
+            activity.user_search(pyramid_request)
+
```
robertknight: Should only have one new line between methods, as opposed to the two blank lines between top level items in a file

h/views/activity.py
```diff
@@ -54,18 +53,22 @@ def search(request):
...
     if not request.feature('search_page'):
         raise httpexceptions.HTTPNotFound()
 
+    opts = {}
+
     result = search(request)
 
     pubid = request.matchdict['pubid']
 
+    opts['search_groupname'] = request.matchdict['pubid']
```
robertknight: This says group name but you're using the pubid here, from the user's POV, they should see the name rather than the pubid on the lozenge.

sheetaluk: Yes, thats part of this card: https://trello.com/c/2b1dIFpS/5-customise-search-lozenge-display-for-group-terms

h/views/activity.py
```diff
@@ -14,15 +14,14 @@
 from h.paginator import paginate
 from h import util
 
+from h import models
```
robertknight: This duplicates another import just above (line 12)

robertknight:
LGTM

sean-roberts:
There is no way to remove the lozenges created here

h/templates/panels/navbar.html.jinja2
sean-roberts: Unless there is any reason not to @sheetaluk, I am going to move this logic of iterating and adding lozenges into the .search-bar__lozenges container below. It has the `data-ref="searchBarLozenges"` that is used in the search controller and tracked with this._lozengeContainer.

PR #4040, opened 2016-11-09
Make usernames in search sidebar clickable

seanh:
Make clicking on one of the usernames in the sidebar on the `/groups/<pubid>/search` page add a facet for that user into the current search query. Clicking on a username adds a user facet to the query, without losing any existing parts of the query (even unsubmitted query words), and reloads the page to update the search results. Once the search is faceted by a user, clicking on the username in the sidebar again _removes_ the facet (same as clicking on the facet's x in the search bar):

peek 2016-11-09 19-05

* * *

There are a number of problems with this list of usernames and annotation counts in the sidebar (and tags, when added, will have the same issues). The problems emerge from the statistics that the server-side code already makes available to the page and that are supposed to be used to render the sidebar. The problems _aren't_ introduced by the code in this PR, this PR just brings them to light. I will describe the problems below, and I think we probably want to make some changes to address these, but I would beg that this current pull request (which does not address any of these problems) be merged first so that we can deploy it behind the feature flag and then demonstrate it to and discuss it with the right people and decide what to do.

The problems that I see are:

My understanding of the requirements for the group search sidebar (and also the user search sidebar still to come) has been that the various stats displayed were to come from the stats that the server-side code already makes available to the page. The `search()` and `group_search()` views make available:

* The total number of annotations that matched the search
* A list of all the users that have created annotations that matched the search (also the number of users that have annotations that matched the search is the length of this list)
* For each user, the number of their annotations that matched the search
* Also the group name, description, date that the group was created on

screenshot from 2016-11-09 19-25-53

All of these stats are based on _the set of annotations that matched the search_ though, which means the stats are probably not what users will assume they are. For example:

* Users that are members of the group but that haven't created any shared annotations in the group will _not_ be listed
* Users that have created shared annotations in the group but who are no longer members of the group _will_ be listed

It gets worse when you start searching for annotations within the group. For example if I go to `/groups/<pubid>/search` and then type "metrics" into the search box and hit Enter I will end up on `/groups/<pubid>/search?q=metrics` and:

* The total number of annotations shown in the sidebar counts only annotations that belong to the group _and that match the search "metrics"_
* The list of all the users contains only those users that have created an annotation in the group _that matches the search "metrics"
* The number of annotations next to each username will count only those annotations created by that user in the group _and that match the search metrics_

Another example: go to `/groups/<pubid>/search` and click on the username **seanh** in the sidebar. It adds a user:seanh facet into the search query, so you end up at `groups/<pubid>/search?q=user%3A%22seanh%22`. _Now only a single user, seanh, is listed in the list of the group's members in the sidebar_. Because seanh is, obviously, the only user who has created an annotation that was created by seanh.

This raises the question of what should happen if I click on "seanh" in the sidebar when the search is already faceted by seanh, since the sidebar button is supposed to add a search facet into the query, but the query already contains one. For now I've made it _toggle_ the facet into/out of the query, so it'll remove it.

### DIFF
--- a/h/views/activity.py
+++ b/h/views/activity.py
@@ -1,8 +1,6 @@
 # -*- coding: utf-8 -*-
 
-"""
-Activity pages views.
-"""
+"""Activity pages views."""
 
 from __future__ import unicode_literals
 
--- a/h/views/activity.py
+++ b/h/views/activity.py
@@ -9,6 +9,7 @@ from __future__ import unicode_literals
 from pyramid import httpexceptions
 from pyramid.view import view_config
 from sqlalchemy.orm import exc
+from memex.search import parser
 
 from h import models
 from h.activity import query
@@ -44,6 +45,8 @@ def search(request):
 
     for user in result.aggregations.get('users', []):
         user['username'] = util.user.split_user(user['user'])['username']
+        user['userid'] = user['user']
+        user['faceted_by'] = _faceted_by_user(request, user['username'], q)
         del user['user']
 
     return {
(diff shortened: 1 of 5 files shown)


h/views/activity.py
```diff
@@ -115,3 +115,60 @@ def group_leave(request):
...
+        raise httpexceptions.HTTPNotFound()
+
+    userid = request.params['toggle_user_facet']
+    username = util.user.split_user(userid)['username']
+
+    new_params = request.params.copy()
+
+    del new_params['toggle_user_facet']
+
+    search_words = new_params.get('q', '').split()
```
robertknight: The query may contain quoted terms. I think you may want to look at using the existing functions for extracting the search query from the request in `h/activity/query.py`.

h/views/activity.py
```diff
@@ -115,3 +115,60 @@ def group_leave(request):
...
+
+def _query_facet_for_user(username):
+    """
+    Return a search query facet for the given username.
+
+    Returns query facets suitable for adding in to the search bar input.
+    For example for "seanh" return 'user:"seanh"'.
+
+    """
+    return 'user:"{username}"'.format(username=username)
```
robertknight: This assumes that the username is always quoted. Will that be the case?

h/views/activity.py
seanh: Hmm. Good point. `query,py` does seem to understand unquoted, single-quoted and double-quoted group facets and do its redirect correctly, so there must be something in there that I can use...

chdorner: fyi: we handle quotes that are not closed as literal quotes. So `group:"foo bar` will become `{"group": "\"foo", "any": "bar"}`.

chdorner:
I just want to second Robert's point about "Try to use the existing query parsing/unparsing code if you can rather than rolling new code for it here.".
We have a proper query parser, just splitting the query by whitespace is not enough to mimic how we parse queries.

h/views/activity.py
seanh: Ok, I've replaced this with use of `memex.search.parser.parse()` instead. It works and avoids duplicating code for parsing and generating search query strings. There is a factoring issue that I think should be dealt with in a later PR:

`h.activity.query.extract()` already uses `memex.search.parser.parse()`, so it'd be nice if `h.views.activity` could just call `h.activity.query.extract()` instead of also using `memex.search.parser.parse()` itself. The problem is that as well as calling `parse()`, `extract()` also does some things that these views that just construct a new query string and redirect to it don't want: on group and user pages it extracts the group or user from the URL _and adds it into the query_.

h/views/activity.py
seanh: This is gone now that we're using the existing `parse()` and `unparse()` instead

seanh:
@robertknight All done I think

robertknight:
LGTM

PR #4041, opened 2016-11-10
Redirect group overview page to group search page

robertknight:
For users with the search_page feature enabled, redirect requests for
the legacy group overview page to the new group profile page which is a
variant of the search page.

This results in the navbar's dropdown group links and links to the
group's page from the client land on the new group profile page.

### DIFF
--- a/h/views/groups.py
+++ b/h/views/groups.py
@@ -122,6 +122,11 @@ def read(group, request):
         request.override_renderer = 'h:templates/groups/join.html.jinja2'
         return {'group': group}
 
+    # Redirect to new group overview page if search page is enabled
+    if request.feature('search_page'):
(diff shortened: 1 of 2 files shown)


h/views/groups.py
```diff
@@ -122,6 +122,11 @@ def read(group, request):
         request.override_renderer = 'h:templates/groups/join.html.jinja2'
         return {'group': group}
 
+    # Redirect to new group overview page if search page is enabled
+    if request.feature('search_page'):
+        url = request.route_path('activity.group_search', pubid=group.pubid)
+        return HTTPSeeOther(url)
```
chdorner: Hmm, I thought you'd need to `raise` these exceptions in Pyramid, but it looks like we always return `HTTPSeeOther` exceptions.

PR #4042, opened 2016-11-10
Make usernames link to new user activity page

robertknight:
This makes usernames in several places link to the new user activity page instead of the stream, when the search_page feature flag is enabled:

 * The username in the navbar on the home page
 * Usernames in annotation cards in search results
 * Usernames linked in annotation cards from the client

In order to achieve this:
 * The existing `/u/{username|account}` URL now redirects to `/users/{username}/search` if the search_page flag is enabled
 * The homepage now uses the `/u/{username}` route instead of generating the stream URL manually
 * The navbar and search result pages now use the `activity.user_search` route instead of generating URLs manually

### DIFF
--- a/h/templates/includes/annotation_card.html.jinja2
+++ b/h/templates/includes/annotation_card.html.jinja2
@@ -2,7 +2,7 @@
 
 <li class="annotation-card">
   <header class="annotation-card__header">
-    <a href="{{ request.route_url('stream') }}?q=user:{{ annotation.username }}"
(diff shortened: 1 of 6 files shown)


dwhly:
Should we move /u/ > /users/ or /groups/ > /g/ ??   Assume /u/ will continue to be aliased either way.

chdorner:
@dwhly This is just to bring people for whom the feature flag is enabled to the new pages wherever they click in the UI. I think there are plans to change `/users/.../search` to `/users/...` before the launch, but others will know better.

PR #4043, opened 2016-11-10
Add suggestions for users current group

sean-roberts:
Suggestions of groups and resolving groups search value vs display value

https://trello.com/c/S11Vh6C0/2-add-group-suggestions-to-suggestion-box

https://trello.com/c/2b1dIFpS/5-customise-search-lozenge-display-for-group-terms

https://trello.com/c/X3PGovwH/4-customize-search-term-entry-for-groups

### DIFF
--- a/h/static/scripts/controllers/search-bar-controller.js
+++ b/h/static/scripts/controllers/search-bar-controller.js
@@ -8,9 +8,20 @@ var AutosuggestDropdownController = require('./autosuggest-dropdown-controller')
 var SearchTextParser = require('../util/search-text-parser');
 var stringUtil = require('../util/string');
 
+
 const FACET_TYPE = 'FACET';
 const TAG_TYPE = 'TAG';
+const GROUP_TYPE = 'GROUP';
 const MAX_SUGGESTIONS = 5;
+
+
+/**
+ * Normalize a string for use in comparisons of user input with a suggestion.
+ * This causes differences in unicode composition and combining characters/accents to be ignored.
+ */
+var normalizeStr = function(str){
+  return stringUtil.fold(stringUtil.normalize(str));
+};
 
 /**
  * Controller for the search bar.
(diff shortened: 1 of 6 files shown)


robertknight:
Found an issue with escaping:

1. Create a group named `<*>Haskell fans<*>`
2. Focus the search box and type "group:"
3. Select the group created in step (1)

**Expected:** Lozenge created with the text `group:<*>Haskell fans<*>`
**Actual:** Lozenge created with the text: `group:"&lt;*&gt;Haskell fans&lt;*&gt;"`

robertknight:
I've just started reviewing this. One piece of general feedback I have is that understanding the high-level flow in `SearchBarController` is starting to get pretty difficult because the constructor function is now ~400 lines long (!). Anything that can be done to break it into smaller chunks would be helpful.

Some ideas, I'm sure you have others:

1. Pull out long anonymous functions which don't access any of the internal state of the controller out into helper functions
2. Separate out the logic for reading data rendered into the page into a separate function or a separate module
3. Separate out logic for filtering and ranking suggestions based on some input query into a separate function or module

h/static/scripts/controllers/search-bar-controller.js
```diff
@@ -23,10 +24,143 @@ class SearchBarController extends Controller {
...
+      return explanationList.concat(tagsList, groupsList);
+    })();
+
     var getTrimmedInputValue = () => {
       return this._input.value.trim();
     };
 
+
+    /**
+     * given a lozenge value for a group: find the pubid, if we can
```
robertknight: I found it difficult to parse this comment and understand what the inputs/outputs of this function are.

My understanding reading through the code is that it accepts user input of the form `group:{pubid|name}` and returns `{ input: "group:{pubid}", display: "group:{name}" }` if the group with the corresponding pubid or name is found or `{ input: <original input>, display: <original input> }` otherwise.

h/static/scripts/controllers/search-bar-controller.js
```diff
@@ -23,10 +24,143 @@ class SearchBarController extends Controller {
...
+
+      const matchVal = stringUtil.fold(stringUtil.normalize(groupVal)).toLowerCase();
+
+      // NOTE: We are pushing a pubid to lowercase here. These ids are created by us
+      // in a random generation case-sensistive style. Theoretically, that means
+      // casting to lower could cause overlaps of values like 'Abc' and 'aBC' - making
+      // them equal to us. Since that is very unlikely to occur for one user's group
+      // set, the convenience of being defensive about bad input/urls is more valuable
+      // than the risk of overlap.
+      const matchByPubid = this._suggestionsMap.find((item)=>{
```
sean-roberts: that is correct it's needed to resolve existing group query lozenge values - which will be pubid more often than not.

h/static/scripts/controllers/search-bar-controller.js
```diff
@@ -23,10 +24,143 @@ class SearchBarController extends Controller {
...
+        return item.type === GROUP_TYPE && item.pubid.toLowerCase() === matchVal;
+      });
+
+      if(matchByPubid){
```
robertknight: There is a mix of spacing styles here. I'm going to add the relevant config to ESLint shortly and we can make the existing code consistent with `eslint --fix` (so it doesn't need to be corrected in this PR), but to clarify, the style should be:

- A space before and after the condition in `if/while/for/select` blocks (`if (...) {`)
- A space before/after the parens for `else` (`} else {`)
- A space after the arg list for arrow functions (`(args...) => {`)

h/static/scripts/controllers/search-bar-controller.js
```diff
@@ -61,8 +195,17 @@ class SearchBarController extends Controller {
      */
     var updateHiddenInput = () => {
       let newValue = '';
-      Array.from(this._lozengeContainer.querySelectorAll('.js-lozenge__content')).forEach((loz) => {
-        newValue = newValue + loz.textContent + ' ';
+      Array.from(this._lozengeContainer.querySelectorAll('.lozenge')).forEach((loz) => {
+        // static lozenges are meant to be
```
robertknight: Incomplete comment

h/static/scripts/controllers/search-bar-controller.js
```diff
@@ -74,7 +217,8 @@ class SearchBarController extends Controller {
      *
      * @param {string} content The search term
      */
-    var addLozenge = content => {
+    var addLozenge = (content, configOverrides)=> {
```
robertknight: `configOverrides` param needs API doc comment

h/static/scripts/controllers/search-bar-controller.js
```diff
@@ -83,27 +227,52 @@ class SearchBarController extends Controller {
...
           deleteCallback: deleteCallback,
-        }
+        }, configOverrides || {})
       );
     };
 
     /**
      * Create lozenges for the search query terms already in the input field on
-     * page load.
+     * page load and update lozenges that are already in teh lozenges container
```
robertknight: Typo: teh/the

h/static/scripts/tests/controllers/search-bar-controller-test.js
```diff
@@ -138,6 +193,97 @@ describe('SearchBarController', function () {
...
+          .click(input)
+          .type('group:', () => {
+            assert.isTrue(dropdown.classList.contains('is-open'));
+
+            let titles = getItemTitles();
+
+            assert.lengthOf(titles, 5, 'we should be enforcing the 5 item max');
+          })
+          .type('[backspace][backspace][backspace][backspace][backspace][backspace]', () => {
+            assert.deepEqual(getItemTitles(), [ 'user:', 'tag:', 'url:', 'group:' ], 'tags go away as facet is removed');
```
robertknight: "tags go away" => "group suggestions disappear"?

h/static/scripts/tests/controllers/search-bar-controller-test.js
```diff
@@ -138,6 +193,97 @@ describe('SearchBarController', function () {
...
+            done();
+          });
+      });
+
+      it('orders groups by indexOf score', function(done){
```
robertknight: What is "indexOf score"? Ideally tests should be phrased in terms of the behavior that the end user sees so that during any refactoring, it is obvious what the consequences of a failing test are from a user's POV.

h/static/scripts/tests/controllers/search-bar-controller-test.js
```diff
@@ -374,5 +530,60 @@ describe('SearchBarController', function () {
           done();
         });
     });
+
+    describe('resolving initial group lozenge and input value', function(){
```
robertknight: I found this difficult to parse. Do you mean "converting server-rendered group lozenges to client-rendered group lozenges"?

tests/h/views/activity_test.py
```diff
@@ -8,7 +8,7 @@
 
 from h.views import activity
 
-@pytest.mark.usefixtures('paginate', 'query')
+@pytest.mark.usefixtures('paginate', 'query', 'pyramid_request_mocked')
```
robertknight: We do something like this in a few other places in the codebase except the usual pattern is to "override" the `pyramid_request` fixture for a suite of tests using something like this:

```py
class TestSomething(object):
  def test_it_does_something(self, pyramid_request):
    ...
  @pytest.fixture
  def pyramid_request(pyramid_request):
    # Customize request for this suite of tests
    return pyramid_request
```

sean-roberts: I am not familiar at all with what you're suggesting or really what the code is doing. renaming and removing the useFixture reference did not override it. I'm afraid for this suggestion, if it's worth holding up the merge, you may need to do the code change yourself so I can understand/learn what you are talking about

h/static/scripts/controllers/search-bar-controller.js
sean-roberts: I am updating the jsdoc

h/static/scripts/controllers/search-bar-controller.js
```diff
@@ -8,10 +8,20 @@ var AutosuggestDropdownController = require('./autosuggest-dropdown-controller')
...
 const FACET_TYPE = 'FACET';
 const TAG_TYPE = 'TAG';
+const GROUP_TYPE = 'GROUP';
 const MAX_SUGGESTIONS = 5;
 
+
+/**
+ * standardized normalization pattern
+ */
+var normaliseStr = function(str){
```
robertknight: Nit: Should use US English spelling for consistency ("normalize"). Would also benefit from a comment about why normalization is done, eg. "Normalize a string for use in comparisons of user input with a suggestion. This causes differences in unicode composition and combining characters/accents to be ignored."

h/static/scripts/controllers/search-bar-controller.js
```diff
@@ -23,10 +33,152 @@ class SearchBarController extends Controller {
...
+     *  or the group name. The result will be an object to identify what
+     *  is the search input term to use and what value can be displayed
+     *  to the user. If there is no match, the input and display will be
+     *  the original input value.
+     *
+     *  @param {String} groupLoz - ex: "group:value"
+     *  @returns {Object} represents the values to display and use for inputVal
+     *    {
+     *      display: {String}, // like group:"friendly name"
+     *      input: {String}    // like group:pid1234
```
robertknight: 👍 - Much clearer

h/static/scripts/controllers/search-bar-controller.js
sean-roberts: lol.

h/static/scripts/tests/controllers/search-bar-controller-test.js
```diff
@@ -250,6 +397,15 @@ describe('SearchBarController', function () {
       return ctrl.refs.searchBarLozenges.querySelectorAll('.js-lozenge');
     }
 
+    /**
+     * return the array of all of lozenges values
+     */
+    const getLozengeValues = function (){
+      return Array.from(ctrl.refs.searchBarLozenges.querySelectorAll('.lozenge')).map((el)=>{
+        return el.querySelector('.lozenge__content').innerText;
```
robertknight: Doesn't prevent merge but I'll note that [`innerText` is a non-standard property](https://developer.mozilla.org/en-US/docs/Web/API/Node/innerText).

robertknight:
Works nicely. As I mentioned in a few comments I did see a bunch of style nits, but all fixable with `eslint --fix` so https://github.com/hypothesis/h/pull/4055 will take care of that.

PR #4045, opened 2016-11-10
Add user sidebar to user search page

seanh:
* Sidebar appears on `/users/<username>/search` page only
* Display name is used, falls back on username if no display name
* Num. annotations and date joined are shown
* Description is shown, only if user has one
* Location is shown, if user has one
* Link is shown (domain name part only) and hyperlinked, if user has one
* ORCID is shown and linked to ORCID site if user has one
* Edit profile link is shown if you're logged-in and on your own user search page
* If you go to a `/users/<username>/search` for a `<username>` that doesn't exist you just get no sidebar, same as group sidebar
* The mobile layout with more info / back buttons works the same way as the group sidebar
* As with the groups sidebar there's no tags component yet

### DIFF
--- a/h/accounts/services.py
+++ b/h/accounts/services.py
@@ -48,23 +48,38 @@ class UserService(object):
         def flush_cache(session):
             self._cache = {}
 
-    def fetch(self, userid):
+    def fetch(self, userid_or_username, authority=None):
         """
-        Fetch a user by userid, e.g. 'acct:foo@example.com'
+        Fetch a user by userid or by username and authority.
+
+        Takes *either* a userid *or* a username and authority as arguments.
+        For example::
+
+          user_service.fetch('acct:foo@example.com')
+
+        or::
+
+          user_service.fetch('foo', 'example.com')
 
         :returns: a user instance, if found
         :rtype: h.models.User or None
+
         """
-        parts = util.user.split_user(userid)
-        username = parts['username']
-        authority = parts['domain']
+        if authority is not None:
+            username = userid_or_username
+        else:
+            userid = userid_or_username
+            parts = util.user.split_user(userid)
+            username = parts['username']
+            authority = parts['domain']
 
         # The cache is keyed by (username, authority) tuples.
         cache_key = (username, authority)
 
         if cache_key not in self._cache:
             self._cache[cache_key] = (self.session.query(User)
-                                      .filter_by(userid=userid)
+                                      .filter_by(username=username)
+                                      .filter_by(authority=authority)
                                       .one_or_none())
 
         return self._cache[cache_key]
--- a/h/accounts/services.py
+++ b/h/accounts/services.py
@@ -10,6 +10,7 @@ from h import mailer
 from h._compat import text_type
 from h.emails import signup
 from h.models import Activation, Subscriptions, User
+from h import util
 
 
 class LoginError(Exception):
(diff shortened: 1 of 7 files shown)


h/views/activity.py
```diff
@@ -92,31 +95,76 @@ def group_search(request):
...
+
+    result['user'] = {
+        'name': user.display_name or user.username,
+        'num_annotations': result['total'],
+        'description': user.description,
+        'registered_date': user.registered_date.strftime('%B, %Y'),
+        'location': user.location,
+        'uri': user.uri,
+        'domain': domain(user),
+        'orcid': user.orcid,
```
seanh: In the future we might want a `UserHTMLPresenter` class in `h.presenters` for all this

h/views/activity.py
```diff
@@ -92,31 +95,76 @@ def group_search(request):
...
+    username = request.matchdict['username']
+
+    result = search(request)
+
+    result['opts'] = {'search_username': username}
+    result['more_info'] = 'more_info' in request.params
+
+    user = request.find_service(name='user').fetch(
+        'acct:{username}@{authority}'.format(username=username,
+                                             authority=request.auth_domain))
```
seanh: This is a bit of technical debt (that some view code should need to know how to construct a userid from a username). We used to have a helper function for this but `fdd39757b62308680ba55c98557abeb1bdf52bb0` removed it. There us a `.userid` property on the `User` model instead. The problem is that the user service does not have a method for fetching a user by username and authority, it requires the userid, so I can't get the user model object without having the userid first, unless I access the model code directly instead of going through the service.

I think the solution is to add a `fetch_by_username(username, authority)` to the user service

seanh: Hmm, the user service also caches user objects by userid, and of course if you only have the username and authority (and need the model object in order to get the userid) then you can't look up the model object in the cache by userid. So we may need a two-keyed cache that can look up objects by userid or bu (username, authority). (This definitely seems like something for a separate PR)

seanh: I've done this now but suggest that I push it as a separate PR once this one is merged

seanh:
Tags (not clickable yet) in the groups sidebar have just been merged into master, and I've just rebased this onto master. Added the tags into the user sidebar as well. And added that user service refactor that I mentioned.

h/accounts/services.py
```diff
@@ -47,19 +48,41 @@ def __init__(self, default_authority, session):
         def flush_cache(session):
             self._cache = {}
 
-    def fetch(self, userid):
+    def fetch(self, userid_or_username, authority=None):
         """
-        Fetch a user by userid, e.g. 'acct:foo@example.com'
+        Fetch a user by userid or by username and authority.
+
+        Takes *either* a userid *or* a usernamen and authority as arguments.
```
robertknight: Typo - "usernamen"

seanh: Fixed

seanh:
Tests should be fixed now...

h/accounts/services.py
```diff
@@ -47,19 +48,41 @@ def __init__(self, default_authority, session):
...
-                                   .one_or_none())
+        if authority is not None:
+            username = userid_or_username
+        else:
+            userid = userid_or_username
+            parts = util.user.split_user(userid)
+            username = parts['username']
+            authority = parts['domain']
+
+        # The cache is keyed by (username, authority) tuples.
```
robertknight: It would be helpful to include this documentation at the point where `self._cache` is "declared" (since you don't pre-declare fields in Python, the point of first initialization in the constructor is a good alternative)

h/templates/activity/search.html.jinja2
```diff
@@ -105,6 +105,191 @@
...
+<section class="search-result-sidebar__section">
+  <h3 class="search-result-sidebar__subtitle">
+    {% trans %}Tags{% endtrans %}
+    <span class="search-result-sidebar__subtitle-count">
+      {{ aggregations['tags'] | length }}
+    </span>
+  </h3>
+  <div class="search-result-tags">
+    {% for tag in aggregations.tags %}
+      <div class="search-result-tag js-tag">
```
robertknight: It doesn't look like there are any controllers registered for the `.js-tag` selector?

seanh: Hmm, nor is there any js related to tags afaik. Good spot

robertknight:
LGTM. I'm going to merge this now but have a few pieces of UX feedback which can be addressed afterwards:

1. The ORCID ID link needs a tooltip
2. The user profile link does not display if the user did not enter a scheme in the Edit Profile form. For example, I have my link set to "github.com/robertknight".
3. I think the Link field should display both the domain and the path, since I imagine that many users will not have their own custom domain for their homepage
4. I would suggest that the description would look better if justified so it didn't have a ragged right edge. What is implemented here is consistent with the mocks though. Thoughts on this point @conordelahunty ?

PR #4046, opened 2016-11-11
Retain original query term while redirecting search url to group_search or user_search.…

sheetaluk:
When the query terms contain both `group` and `user` terms, the `group` query term
gets popped during the redirect and eventually gets lost.
This change fixes that problem.

https://github.com/hypothesis/product-backlog/issues/36

### DIFF
--- a/h/activity/query.py
+++ b/h/activity/query.py
@@ -84,7 +84,7 @@ def check_url(request, query, unparse=parser.unparse):
                                       pubid=pubid,
                                       _query={'q': unparse(query)})
 
-    if _single_entry(query, 'user'):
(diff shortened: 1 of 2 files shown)


seanh:
Hmm, is the problem that, if the query contains _both_ a single user facet, then both the `if _single_entry(query, 'group'):` and the `if _single_entry(query, 'user'):` blocks get executed? And therefore both the group and the user get popped from the query. I'm wondering if changing that second `if` to an `else if` would fix it

robertknight:
Also, don't forget to add tests for this please.

robertknight:
LGTM

PR #4047, opened 2016-11-11
Display tags on search results page.

sheetaluk:
https://github.com/hypothesis/product-backlog/issues/10

### DIFF
--- a/h/static/styles/partials-v2/_search.scss
+++ b/h/static/styles/partials-v2/_search.scss
@@ -181,6 +181,32 @@
   word-break: break-all;
 }
 
+.search-result-tags {
+  display: flex;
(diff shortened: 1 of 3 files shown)


h/static/styles/partials-v2/_search.scss
```diff
@@ -181,6 +181,32 @@
   word-break: break-all;
 }
 
+.search-result-tags {
+  display: flex;
+  flex-wrap: wrap;
```
seanh: Ah of course, flexbox is perfect for this!

seanh:
Ok, I've pushed a couple of commits just to move the tags into the sidebar and to i18n the _Tags_ title. Looks great

PR #4055, opened 2016-11-11
Fix inconsistencies in front-end code style

robertknight:
When the use of ES2015+ was initially enabled in the JS codebase, it was done so without a clear plan for how we would ensure consistent code style. Consequently, we're now mixing arrow and regular functions, let/var/const etc. Additionally, not all of our spacing rules were codified in the lint config.

Since then, I proposed and we've agreed to broadly use the [Airbnb style](https://github.com/airbnb/javascript) as a basis, since it was consistent with our existing style and is very widely used. This PR is a series of commits which fixes these inconsistencies using the same approach that was used for the client repo. The resulting code is much closer to the Airbnb style.

Each commit:

1. Enables an ESLint rule following the appropriate section of the Airbnb guide
2. Uses `eslint --fix h/static/scripts` to fix all violations in the existing codebase

The changes included here are:

1. Consistent spacing around keywords (if/else/function etc.)
2. Consistent spacing before blocks
3. Consistent use of const/let/var (use const everywhere possible, let otherwise)
4. Consistent use of arrow functions (always use them for anonymous functions, parens around the arguments if there is a block, no parens otherwise if the function has just one argument)

In future we'll fold at least the ES5-compatible config changes into the [shared config](https://github.com/hypothesis/frontend-toolkit/tree/master/packages/eslint-config-hypothesis).

When reviewing this, the easiest way is just to go through each commit and look at the commit message plus the first few lines of each diff.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,6 +1,7 @@
 {
   "extends": "hypothesis",
   "rules": {
+    "arrow-parens": ["error", "as-needed", { "requireForBlockBody": true }],
     "arrow-spacing": "error",
     "keyword-spacing": "error",
     "space-before-blocks": "error"
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,6 +1,7 @@
 {
   "extends": "hypothesis",
   "rules": {
+    "arrow-spacing": "error",
     "keyword-spacing": "error",
     "space-before-blocks": "error"
   }
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,6 +1,7 @@
 {
   "extends": "hypothesis",
   "rules": {
+    "keyword-spacing": "error",
     "space-before-blocks": "error"
   }
 }
--- a/.eslintrc
+++ b/.eslintrc
@@ -3,8 +3,10 @@
   "rules": {
     "arrow-parens": ["error", "as-needed", { "requireForBlockBody": true }],
     "arrow-spacing": "error",
+    "no-var": "error",
     "keyword-spacing": "error",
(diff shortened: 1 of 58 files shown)


robertknight:
For future reference, here is how I rebase a PR like this which touches a lot of code using automated tools, driven by a config file in the root of the tree.

```sh
git checkout frontend-airbnb-style
git fetch
git branch -f master origin/master

# Rebase on master, mark each commit as "edit"
git rebase --interactive master

# For each commit
git checkout HEAD h/static/scripts # Check out HEAD before the automated code change was applied
./node_modules/.bin/eslint --fix h/static/scripts # Apply the automated code change
git add h/static/scripts # Add updated modified code to commit
gulp test # Changes should be safe, but run tests as an check
```

.eslintrc
```diff
@@ -1,3 +1,12 @@
 {
-  "extends": "hypothesis"
+  "extends": "hypothesis",
+  "rules": {
```
robertknight: These rules are all taken from the "eslint: <rule name>" mentioned in the appropriate section of the Airbnb guide. The one exception is `{ "destructuring": "all" }` which avoids warnings if you do:

```js
let { fieldWhichIsLaterModified, fieldWhichIsNotLaterModified } = object;
```

Instead of:

```js
const { fieldWhichIsNotLaterModified } = object;
let { fieldWhichIsLaterModified } = object;
```

I felt this was just a little bit too pedantic and more importantly ESLint can't yet fix such cases automatically.

PR #4062, opened 2016-11-14
Add and remove tags from the side bar.

sheetaluk:
https://github.com/hypothesis/product-backlog/issues/10


seanh:
Great work! Works perfectly.

I've discovered a bug that any unsubmitted search terms that are in the search box are lost when you click on a tag in the sidebar, but that bug also exists when clicking on usernames in the sidebar, I'll send a separate PR to fix it.

PR #4078, opened 2016-11-15
Fix sidebar buttons in IE

seanh:


seanh:
I think this gets all the buttons working in both IE and other browsers. In IE any unsubmitted query terms will be lost when clicking on any button, but submitted terms will be preserved. In other browsers unsubmitted terms will be preserved as well. This could be fixed in IE with a bit of JavaScript (but I don't think it's fixable in IE without JavaScript).

seanh:
@robertknight seems like you'd be a good person to review this?

robertknight:
LGTM. Tested in IE 11, Firefox, Chrome

PR #4081, opened 2016-11-15
Fix form buttons on "more info" page

### LINKED ISSUE #4054: Clicking on usernames and tags in the sidebar does not work when on the "more info" page
e0cc6bbe-a752-11e6-9ed9-6645acdb887f

See https://github.com/hypothesis/h/pull/4025#issuecomment-259705303

seanh:
When on the "More info" page (so &more_info is in the URL) various
buttons (usernames for adding a user facet to the search query, tag
names for adding tags into the search query, the leave group button...)
were not working.

The problem is to do with query parameters in the URL and post
parameters in the body of a POST request. When you click on a tag for
example, this does a POST request form submission with `q` (the search
query entered so far) and `toggle_tag_facet` in the post body. BUT
&more_info is in the URL of the current page, so that is also submitted
to the server as a URL query parameter.

Server-side we use Pyramid's request_param view config predicate to
decide which view function to call, and because &more_info is present in
the URL it calls the search_more_info() function instead of
toggle_tag_facet(). Pyramid view config does not distinguish between URL
query params and post body params - there us a single request_param view
predicate and it looks for the named param in either the URL query
string or the post body.

The fix is to remove the URL query string from all of the form submission POSTs
that happen when you click on these buttons. Since we're never
(intentionally) using the URL query, but always want to use the post
body, having the URL query be present is just a hazard.

The query string can be removed by having the server set the `action`
attribute of the forms (or `formaction` in the case of buttons that are
outside of their form).

Fixes #4054.

### DIFF
--- a/h/templates/activity/search.html.jinja2
+++ b/h/templates/activity/search.html.jinja2
@@ -5,12 +5,13 @@
 {% macro search_result_nav(name) %}
 {# This <form> is needed for IE because it doesn't support the `form`
    attribute. See commit message. #}
-<form method="POST">
+<form method="POST" action="{{ request.path_url }}">
   <input type="hidden" name="q" value="{{ q }}">
   <nav class="search-result-nav">
     {%- if more_info %}
       <button form="search-bar"
               formmethod="POST"
+              formaction="{{ request.path_url }}"
               name="back"
               class="search-result-nav__button">
         {% trans %}Back{% endtrans %}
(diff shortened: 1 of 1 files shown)


seanh:
@robertknight as discussed on hangout!

seanh:
Clearly there's some duplication of the form and button elements in the template here, that could do with some refactoring. This PR fixes the bug though.

robertknight:
LGTM

PR #4083, opened 2016-11-16
Fix 'Copy to clipboard' link on iOS 10 and improve mobile UX of 'Copy to clipboard' field

### LINKED ISSUE #4049: Copying group link not possible on iOS
### Steps to reproduce
1. On an iDevice with iOS 10, go to a group profile page `/groups/<ID>/search`
2. Tap on the "More Info" link if the sidebar is not visible
3. Attempt to copy the group profile link

### Observed issues

1. Tapping the copy button fails, although I believe iOS 10 should support copying to the clipboard in the context of a user gesture
2. Focusing the group link input field zooms in the display because the font size is < 16px
3. You can't easily select the contents of the group invite widget because it is read only

robertknight:
This fixes several usability issues with the group link on iOS:

1. The "Copy" button did not work because `HTMLInputElement.prototype.select()` does not work properly on iOS (See first commit)
2. The field had several mobile UX issues - the copy button was too small and the font size of the input was also too small (See second commit)
3. Copying text from a read-only input field, in the event that the "Copy" button fails to work (iOS < 10), is hard in Mobile Safari. Therefore I made the field read-write in Mobile Safari.

Fixes #4049

### DIFF
--- a/h/static/scripts/tests/controllers/copy-button-controller-test.js
+++ b/h/static/scripts/tests/controllers/copy-button-controller-test.js
@@ -1,0 +1,51 @@
+'use strict';
+
+const CopyButtonController = require('../../controllers/copy-button-controller');
+const { setupComponent } = require('./util');
+
+describe('CopyButtonController', () => {
+  const template = `<div>
+    <input data-ref="input">
+    <button data-ref="button"></button>
+  </div>`;
+
+  let copiedText;
+  let copySuccess;
+  let ctrl;
+
+  beforeEach(() => {
+    copySuccess = false;
+    copiedText = '';
+    sinon.stub(document, 'execCommand', (command) => {
+      if (command === 'copy') {
+        copiedText = document.getSelection().toString();
+        return copySuccess;
+      }
+      return false;
+    });
+    ctrl = setupComponent(document, template, CopyButtonController);
+    ctrl.refs.input.value = 'Text to copy';
+  });
+
+  afterEach(() => {
+    document.execCommand.restore();
+  });
+
+  it('copies text when button is clicked', () => {
+    copySuccess = true;
+    ctrl.refs.button.click();
+    assert.equal(copiedText, 'Text to copy');
+  });
+
+  it('displays success message if text was copied', () => {
+    copySuccess = true;
+    ctrl.refs.button.click();
+    assert.include(ctrl.refs.input.value, 'Link copied');
+  });
+
+  it('displays alternative message if text was not copied', () => {
+    copySuccess = false;
+    ctrl.refs.button.click();
+    assert.include(ctrl.refs.input.value, 'Copying link failed');
+  });
+});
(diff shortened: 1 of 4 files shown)


h/static/styles/partials-v2/_group-invite.scss
```diff
@@ -9,22 +9,16 @@
...
+  padding-right: 22px;  /* Make space for the clipboard icon. */
+  padding-top: 0;
   margin-top: 12px;
   margin-bottom: 8px;
   border: 1px solid $grey-3;
   border-radius: 2px;
   width: 100%;
   height: 31px;
-
-  .env-js-capable & {
```
robertknight: I removed these to reduce the number of states to reason about to just two - mobile and non-mobile. Otherwise we'd have 6 to deal with and test (mobile|non-mobile * {JS,no-JS,JS timeout}. In general for this reason, I would suggest that we _only_ add non-JS/timeout selectors if required to make the component usable, not for purely aesthetic reasons.

Otherwise we introduce additional complexity, which increases the likelihood of bugs, which means a worse experience for all our users.

robertknight: For reference, the plain HTML version of this component now looks like this:

group-invite-plain-html

h/static/scripts/tests/controllers/copy-button-controller-test.js
```diff
@@ -0,0 +1,63 @@
+'use strict';
```
seanh: Yay tests :)

seanh:
In testing activity pages on iPhone, I'm seeing a variety of small issues to do with things not seeming to be quite the right size or having the wrong padding or not quite being positioned correctly etc. Generally the UI all looks really nice and is quite nicely usable on mobile though.

Anyway, tested this clipboard button on both desktop and iPhone. Clipboard icon looks incorrectly vertically aligned to me on iPhone but I assume that's not this PR's fault. Copying does work, either by clicking the icon (which is a bit too small though) or by selecting and copying the text.

Since the text is editable you can delete it all and if you do that then click the clipboard icon you will get the copying link failed error message.

Clipboard icon seems to retain its black focus colour after you've clicked it.

This is a big improvement though

seanh:
Great start on fixing lots of activity pages bugs @robertknight , I don't know how you've got the energy, I need a few days to recover first!

h/templates/panels/group_invite.html.jinja2
```diff
@@ -5,8 +5,7 @@ Sharing the link below lets people join this group:
 <div class="group-invite__container js-copy-button">
   <input class="group-invite__input js-select-onfocus"
          data-ref="input"
-         value="{{ group_url }}"
-         readonly>
+         value="{{ group_url }}">
```
robertknight: Yes indeed, although I think it is better that the user gets a form input that they can accidentally edit than one which they can't copy the text from at all (if scripts fail to load on Mobile Safari).

Issue #4069, opened 2016-11-14
Annotation cards in activity pages not showing their groups


dwhly:
### Steps to reproduce
1. Go to activty pages
2. Search for annotations that belong to a group
3. Click to expand the document

### Expected behaviour
Annotations inside groups should indicate that they are 

### Actual behaviour
The annotation cards that show won't indicate that they are within a group

ajpeddakotla:
We can reuse the style from the stream for this feature.

robertknight:
This is already implemented.

PR #4086, opened 2016-11-16
Show group info in annotation cards.

sheetaluk:
https://github.com/hypothesis/h/issues/4069


robertknight:
group-link

Couple of thoughts:

- The group name here is to the right of rather than above the username as it is on the stream, was this an intentional difference?
- You need to override the default color for the `:visited` pseudo-class, that's why I see the out-of-place purple in the screenshot above
- The spacing between the username and group name is perhaps a little large. This could either be fixed by reducing it or putting the group name below as on the stream

robertknight:
LGTM

PR #4088, opened 2016-11-17
Autofocus search input when user types a letter, number or backspace

robertknight:
Copy the behavior seen on Google.com where the search field is
autofocused when the user types an alphanumeric character or presses
Backspace, if no other input on the page has focus.

This allows the user to start typing a search query as soon as the page loads
without having to focus the input field first.

The reason for doing this over just automatically focusing the search input on page load is to avoid obscuring the top results on the page with the autosuggest box which is shown when the search input gains focus.

### DIFF
--- a/h/static/scripts/controllers/input-autofocus-controller.js
+++ b/h/static/scripts/controllers/input-autofocus-controller.js
@@ -1,0 +1,39 @@
+'use strict';
+
+const Controller = require('../base/controller');
+
+function isWordChar(event) {
+  return event.key.match(/^\w$/) && !event.ctrlKey && !event.altKey;
+}
+
+/**
+ * Automatically focuses an input field when the user presses a letter, number
+ * or backspace if no other element on the page has keyboard focus. The field's
+ * focus can also be blurred by pressing Escape.
+ *
+ * This provides behavior similar to Google.com where the user can "type" in the
+ * search box even if it is not focused.
+ */
+class InputAutofocusController extends Controller {
+  constructor(element) {
+    super(element);
+
+    this._onKeyDown = (event) => {
+      if (document.activeElement === document.body) {
+        if (isWordChar(event) || event.key === 'Backspace') {
+          element.focus();
+        }
+      } else if (document.activeElement === element && event.key === 'Escape') {
+        element.blur();
+      }
+    };
+
+    document.addEventListener('keydown', this._onKeyDown);
+  }
+
+  beforeRemove() {
+    document.removeEventListener('keydown', this._onKeyDown);
+  }
+}
+
+module.exports = InputAutofocusController;
--- a/h/static/scripts/polyfills.js
+++ b/h/static/scripts/polyfills.js
@@ -24,6 +24,10 @@ try {
   require('js-polyfills/url');
 }
 
+// KeyboardEvent.prototype.key
+// (Native in Chrome >= 51, Firefox >= 23, IE >= 9)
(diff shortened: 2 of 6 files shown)


h/static/scripts/polyfills.js
```diff
@@ -24,6 +24,10 @@ try {
   require('js-polyfills/url');
 }
 
+// KeyboardEvent.prototype.key
+// (Native in Chrome >= 51, Firefox >= 23, IE >= 9)
+require('keyboardevent-key-polyfill');
```
robertknight: KeyboardEvent.prototype.key is the modern standards-based way to get the character or special key that was pressed. It is available for both 'keydown' and 'keypress' events. Unfortunately it was only recently supported in Chrome and is still not supported in Safari. Fortunately it looks like a basic polyfill is quite small.

seanh:
In some browsers (e.g. Firefox, and I think also Chrome if you have certain plugins installed) just typing when nothing has keyboard focus does something, for example searching within the page, or searching for / selecting links. This breaks that on the `/search` page, causing you to type into the search box instead. Google also does this, but DuckDuckGo doesn't (it just seems to focus the search box once on page load). Bing does it. Yahoo doesn't. Anyway, I'm happy for us to do it.

seanh:
If I just start typing, and say I type a "g", then the dropdown menu will show but it'll only show the groups suggestion.If I type b it won't show any suggestions. So I think a benefit of this not actually focusing the input _until_ you type is that the input doesn't look focused (e.g. no cursor in it), so the user is still encouraged to manually put the focus in the input at which point they'll see the dropdown with all suggestions. At least they're still encouraged to do this until they realise that they don't have to. I don't think it does any harm to the discoverability of the available suggestions.

seanh:
When the input is not focused type a letter that does not match any of the four suggestions in the dropdown menu, and the dropdown briefly shows and then disappears again:

peek 2016-11-21 12-31

Seems like something that could be fixed separately in dropdown menu controller.

seanh:
I guess the best thing about this is that after deleting a lozenge, or adding a lozenge from the sidebar etc, after the page reloads you can just keep typing without having to click in the input again every time. It doesn't _look_ like you can keep typing (input does not look focused, has no cursor in it), but you can.

robertknight:
> In some browsers (e.g. Firefox, and I think also Chrome if you have certain plugins installed) just typing when nothing has keyboard focus does something, for example searching within the page, or searching for / selecting links

This is a very good point. My aim here was to be conservative in which keys can trigger autofocus and to follow Google's behavior on the basis that their implementation is widely battle-tested.

h/static/scripts/controllers/input-autofocus-controller.js
```diff
@@ -0,0 +1,39 @@
...
+const Controller = require('../base/controller');
+
+function isWordChar(event) {
+  return event.key.match(/^\w$/) && !event.ctrlKey && !event.altKey;
+}
+
+/**
+ * Automatically focuses an input field when the user presses a letter, number
+ * or backspace if no other element on the page has keyboard focus. The field's
+ * focus can also be blurred by pressing Escape.
```
seanh: I wonder what happens if you put two of these on the same page at once :)

robertknight: I think what will happen is that whichever one was initialized first will have its event handler run when a key is typed, that will cause a change in focus and then the second event handler will do nothing because the body no longer has focus.

PR #4089, opened 2016-11-17
Instrument elasticsearch queries to statsd

chdorner:
We have two functions that send a request to ElasticSearch, this commit
adds an optional instrumentation code path to these two functions. When
the Search client gets initialized with a stats client, then we'll time
how long it takes the request to finish, and also add counters for
successful, timed-out, and errored search queries.

These new metrics are:

* `memex.search.query` (in graphite as `stats.$host.timers.memex.search.query`)
* `memex.search.query.success` (in graphite as `stats.$host.counters.memex.search.query.success`)
* `memex.search.query.timeout` (in graphite as `stats.$host.counters.memex.search.query.timeout`)
* `memex.search.query.error` (in graphite as `stats.$host.counters.memex.search.query.error`)

### DIFF
--- a/h/activity/query.py
+++ b/h/activity/query.py
@@ -157,7 +157,7 @@ def fetch_annotations(session, ids, reply_ids):
 
 @newrelic.agent.function_trace()
 def _execute_search(request, query, page_size):
-    search = Search(request, separate_replies=True)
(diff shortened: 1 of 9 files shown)


robertknight:
LGTM

PR #4090, opened 2016-11-17
Make document links in URL buckets clickable

### LINKED ISSUE #4058: Document URLs in document buckets aren't linked
screenshot from 2016-11-11 20-07-22

The URL under _Address_ on the right here isn't linked. Possibilities: link to the document in Via? Or just link to the URL as shown, without h?

robertknight:
This implements the document links as shown in the "URLs" section of the bucket in the mocks at https://projects.invisionapp.com/share/B88RCLRFN#/screens

For each bucket, there are two links at the top:
- A prettified version of the URL which goes to the URL without the Hypothesis overlay activated
- A "Visit annotations in context" link which is a direct link to the first annotation in the bucket. When the user lands on that URL, they can click "Show all annotations" to see the rest of the annotations from that bucket

If the user happens to be looking at a specific group, then this will result in the overlay automatically switching to the appropriate group.

The implementation here changes the label from "Address" to "URLs" as in the mock and tweaks the presentation of the section labels in the bucket sidebar, but does not currently include the icons.

Fixes #4058 

bucket-url-links

### DIFF
--- a/h/activity/bucketing.py
+++ b/h/activity/bucketing.py
@@ -10,6 +10,7 @@ import newrelic.agent
 from pyramid import i18n
 
 from h._compat import urlparse
+from h import links
 from h import presenters
 
 
@@ -40,6 +41,19 @@ class DocumentBucket(object):
     @property
     def annotations_count(self):
         return len(self.annotations)
+
+    def incontext_link(self, request):
+        """
+        Return a link to view this bucket's annotations in context.
+
+        The bouncer service and Hypothesis client do not currently provide
+        direct links to view a document's annotations without specifying a
+        specific annotation, so here we just link to the first annotation in the
+        document.
+        """
+        if len(self.annotations) == 0:
+            return None
+        return links.incontext_link(request, self.annotations[0])
 
     def append(self, annotation):
         self.annotations.append(annotation)
--- a/h/activity/query.py
+++ b/h/activity/query.py
@@ -123,11 +123,12 @@ def execute(request, query, page_size):
     # Add group information to buckets and present annotations
     for timeframe in result.timeframes:
         for bucket in timeframe.document_buckets.values():
-            for index, annotation in enumerate(bucket.annotations):
-                bucket.annotations[index] = {
+            bucket.presented_annotations = []
+            for annotation in bucket.annotations:
+                bucket.presented_annotations.append({
                     'annotation': presenters.AnnotationHTMLPresenter(annotation),
                     'group': groups.get(annotation.groupid)
-                }
+                })
 
     return result
 
--- a/h/links.py
+++ b/h/links.py
@@ -5,7 +5,23 @@ Provides links to different representations of annotations.
 """
 
 
-from h._compat import urlparse
+from h._compat import urlparse, url_unquote
+
+
+def pretty_link(url):
+    """
+    Return a nicely formatted version of a URL.
+
+    This strips off 'visual noise' from the URL including common schemes
+    (HTTP, HTTPS), domain prefixes ('www.') and query strings.
+    """
+    parsed = urlparse.urlparse(url)
+    if parsed.scheme not in ['http', 'https']:
+        return url
+    netloc = parsed.netloc
+    if netloc.startswith('www.'):
+        netloc = netloc[4:]
+    return url_unquote(netloc + parsed.path)
 
 
 def html_link(request, annotation):
--- a/h/static/styles/partials-v2/_search.scss
+++ b/h/static/styles/partials-v2/_search.scss
@@ -84,6 +84,11 @@
 
   cursor: pointer;
   user-select: none;
+
+  // The in-context arrow icon is hidden until the bucket header is hovered
+  &:hover .search-result-bucket__incontext-icon {
+    visibility: visible;
+  }
 }
 
 .search-result-bucket__domain {
@@ -93,6 +98,21 @@
   color: $grey-4;
   margin-right: 10px;
   word-wrap: break-word;
+}
+
+.search-result-bucket__domain-link {
+  color: inherit;
+  &:visited {
+    color: inherit;
+  }
+}
+
+.search-result-bucket__incontext-icon {
+  // The in-context arrow icon is hidden until the bucket header is hovered
+  visibility: hidden;
+  width: 9px;
+  height: 9px;
+  color: $grey-4;
 }
 
 .search-result-bucket__title-and-annotations-count {
--- a/h/static/styles/partials-v2/_search.scss
+++ b/h/static/styles/partials-v2/_search.scss
@@ -169,7 +169,9 @@
 }
 
 .search-bucket-stats__key {
-  color: $grey-4;
+  margin-bottom: 2px;
+  color: $grey-6;
+  font-weight: bold;
 }
 
 .search-bucket-stats__val {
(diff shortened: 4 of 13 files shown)


robertknight:
There are some design changes in https://github.com/hypothesis/h/issues/4058#issuecomment-261578927 that need to be implemented.

h/activity/bucketing.py
```diff
@@ -41,6 +42,19 @@ def __init__(self, document, annotations=None):
...
         return len(self.annotations)
 
+    def incontext_link(self, request):
+        """
+        Return a link to view this bucket's annotations in context.
+
+        The bouncer service and Hypothesis client do not currently provide
+        direct links to view a document's annotations without specifying a
+        specific annotation, so here we just link to the first annotation in the
+        document.
```
seanh: Nice, this kind of comment documenting the _why_ is good I think

h/activity/query.py
```diff
@@ -123,11 +123,12 @@ def execute(request, query, page_size):
     # Add group information to buckets and present annotations
     for timeframe in result.timeframes:
         for bucket in timeframe.document_buckets.values():
+            bucket.presented_annotations = []
             for index, annotation in enumerate(bucket.annotations):
-                bucket.annotations[index] = {
+                bucket.presented_annotations.append({
```
seanh: This is odd. I wonder why `index` and `enumerate()` were used here in the first place? As you've noticed it's not necessary to use `index`. I think you can actually change the `for index, annotation in enumerate(bucket.annotations):` to `for annotation in bucket.annotations:` as well?

h/links.py
```diff
@@ -5,7 +5,23 @@
...
+    This strips off 'visual noise' from the URL including common schemes
+    (HTTP, HTTPS), domain prefixes ('www.') and query strings.
+    """
+    parsed = urlparse.urlparse(url)
+    if parsed.scheme not in ['http', 'https']:
+        return url
+    netloc = parsed.netloc
+    if netloc.startswith('www.'):
+        netloc = netloc[4:]
+    return url_unquote(netloc + parsed.path)
```
seanh: Hmm. I'm a little confused about when we should be adding code to `h.links` and when it should go in `h.presenters.AnnotationHTMLPresenter` or `DocumentHTMLPresenter`. Is there code in one of those presenter classes that could be reused here? (Or that should be changed to call this new function instead?)

h/templates/activity/search.html.jinja2
```diff
@@ -35,10 +35,20 @@
...
   {% if bucket.uri %}
     <div class="search-bucket-stats__key">
-        Address
+        URL
     </div>
-    <div class="search-bucket-stats__val search-bucket-stats__address">
-        {{ bucket.uri }}
+    <div class="search-bucket-stats__val search-bucket-stats__url">
+        <div><a class="search-bucket-stats__link"
+                href="{{ bucket.uri }}">{{ pretty_link(bucket.uri) }}</a></div>
```
seanh: Are there cases where a document bucket either doesn't have a URI or doesn't have one that we can / should link to? Locally-annotated files for example. I know that `DocumentHTMLPresenter.web_uri` always tries to return an http(s) URL for a document, and I think it'll return `None` if the document doesn't have an http(s) URL and will also strip `https://via.hypothes.is/` from the start

tests/h/links_test.py
```diff
@@ -76,6 +76,19 @@ def test_incontext_link_skips_uri_for_pdfs_with_no_document(pyramid_request):
...
+    ('http://notsecure.com', 'notsecure.com'),
+    ('https://secure.com', 'secure.com'),
+    ('ftp://not_http', 'ftp://not_http'),
+    ('http://www.google.com', 'google.com'),
+    ('http://site.com/with-a-path', 'site.com/with-a-path'),
+    ('https://site.com/with-a-path?q=and-a-query-string', 'site.com/with-a-path'),
+    ('http://site.com/path%20with%20spaces','site.com/path with spaces'),
+])
+def test_pretty_link(uri, formatted):
+    assert links.pretty_link(uri) == formatted
```
seanh: Nice, very simple yet thorough test

h/static/styles/partials-v2/_search.scss
```diff
@@ -95,6 +100,28 @@
   word-wrap: break-word;
 }
 
+// The domain name in the bucket header is a clickable link on the left-hand
+// side in desktop layouts. On smaller screens it spans the full width of the
+// header and becomes plain text to avoid accidental clicks.
```
seanh: Another good comment

h/templates/activity/search.html.jinja2
```diff
@@ -264,7 +280,7 @@
           </dt>
           <dd class="search-result-sidebar__dd">
             <a href="{{ user.uri }}" class="search-result-sidebar__user-link">
-              {{ user.domain }}
+              {{ pretty_link(user.uri) }}
```
seanh: I think `user.uri` may sometimes be `""` or `None` or just a non-URL string. Does `pretty_link()` deal with these okay?

h/activity/query.py
robertknight: What this originally used to do was change `bucket.annotations` from being a `List[model.Annotation]` to `List[Dict]`. That was confusing both when reading the code and in trying to reason about what could be done with the `annotations` property in `DocumentBucket`. Here I've changed the code to build up the `List[Dict]` in a separate property.

h/links.py
robertknight: This code is a low-level utility that formats URLs for display. The presenters might use this code if they have a need to format links in the same way, though I don't think they do currently.

h/templates/activity/search.html.jinja2
robertknight: See `DocumentBucket` initialization. `bucket.uri` will either be an HTTP(S) URL or `None`, in which case this whole section is not rendered.

h/templates/activity/search.html.jinja2
robertknight: Good catch. I thought the `uri` property was converted to an absolute URL before being stored. That isn't the case. If the user enters say "twitter.com/modelstudent" as their link the URL here is `$HOST/users/username/twitter.com/modelstudent`

robertknight:
> We probably don't want it to do the expand / collapse?

Indeed.

> What was the reason for opening the annotation in context in a new tab? I think links that open new tabs are generally considered a bit of an annoyance, most of the time.

My thinking was to avoid losing the user's place in the search result list, similar to how links in your Facebook or Twitter feeds are opened externally.

seanh:
> > What was the reason for opening the annotation in context in a new tab? I think links that open new 
> > tabs are generally considered a bit of an annoyance, most of the time.
> My thinking was to avoid losing the user's place in the search result list, similar to how links in your 
> Facebook or Twitter feeds are opened externally.

Ah ok. I guess it depends whether activity pages is more like twitter, or more like google! Fair enough, I'm happy to go with this.

h/templates/activity/search.html.jinja2
seanh: Ah ok, cool

h/activity/query.py
seanh: Oh yeah, it was modifying the list that it was iterating over! I agree with your change

tests/h/activity/bucketing_test.py
```diff
@@ -179,6 +180,22 @@ def test_eq_title_mismatch(self, document):
...
         assert not bucket_1 == bucket_2
 
+    def test_incontext_link_returns_link_to_first_annotation(self, document, patch):
+        incontext_link = patch('h.links.incontext_link')
+        bucket = bucketing.DocumentBucket(document)
+        ann = factories.Annotation()
+        bucket.append(ann)
+        request = Mock()
+
+        assert bucket.incontext_link(request) == incontext_link.return_value
```
seanh: Does this test what the method name says it tests? I don't see how it tests that it was the _first_ annotation that was passed to `incontext_link()`

robertknight: True, it only proves that `bucket.incontext_link` returned a value from _a call_ to `incontext_link`, not the value from `incontext_link(ann)`.

PR #4092, opened 2016-11-17
Choose login redirect URL based on feature flags for logged-in user

### LINKED ISSUE #4070: Direct the user to their profile page after login
When a user logs into Hypothesis from `https://hypothes.is/login`, they are currently directed to `https://hypothes.is/stream` after login. 

We should instead direct them to their user profile page after they login.

robertknight:
Previously the URL to redirect to after login was determined at the
start of request processing before authentication completed and
consequently it used the feature flags for _logged out_ users.

This makes the request redirect based on the flags for the logged in
user by deferring retrieval of the login URL until after login and
clearing the feature flag cache just before the login is processed.

No new unit tests added here because the original bug depends on the
interactions of several services and behavior of the feature flag client, so it would
need an integration test.

Fixes #4070

### DIFF
--- a/h/views/accounts.py
+++ b/h/views/accounts.py
@@ -102,8 +102,6 @@ class AuthController(object):
                                         buttons=(_('Log in'),),
                                         footer=form_footer)
 
-        self.login_redirect = self.request.params.get(
-            'next', _login_redirect_url(self.request))
         self.logout_redirect = self.request.route_url('index')
 
     @view_config(request_method='GET')
@@ -125,7 +123,7 @@ class AuthController(object):
 
         user = appstruct['user']
         headers = self._login(user)
-        return httpexceptions.HTTPFound(location=self.login_redirect,
+        return httpexceptions.HTTPFound(location=self._login_redirect(),
                                         headers=headers)
 
     @view_config(route_name='logout',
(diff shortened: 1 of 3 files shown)


h/views/accounts.py
```diff
@@ -139,15 +137,24 @@ def logout(self):
 
     def _redirect_if_logged_in(self):
         if self.request.authenticated_userid is not None:
-            raise httpexceptions.HTTPFound(location=self.login_redirect)
+            raise httpexceptions.HTTPFound(location=self._login_redirect())
+
+    def _login_redirect(self):
+        return self.request.params.get('next', _login_redirect_url(self.request))
```
chdorner: According to the linked Github issue, this should be `/users/:username/search`, but it is currently `/search`

robertknight: Hmm... you're right. That's a change in behavior I missed; sending the user to the replacement for the stream with a specific query rather than just "the stream". I'm dubious about this since it means we'll most likely be sending new users to a blank page after they login for the first time. I'll flag it up in #design

robertknight:
@chdorner - Fixed. Users are now directed to their profile page as per the spec.

PR #4094, opened 2016-11-18
Match facet types case insensitive in search bar

### LINKED ISSUE #4080: Matching of facet type is case sensitive
### Steps to reproduce
1. Go to `/search`, focus input field
2. Type "Tag:" (note upper case 'T') or "User:"

### Expected behaviour
Search suggestions should appear

### Actual behaviour
No search suggestions appear

Firefox v52, iOS 10

### Additional details

I came across this issue because in iOS Safari, the search field by default makes the first letter of the user's input upper case, so if they arrive at `/search` and type "tag:" the input will contain "Tag:".

While we should be able to give the browser a hint that it should not capitalize the first letter of the input in the search input, matching of the facet term should be case-insensitive.

chdorner:
We show a list of tag or group suggestions when a user types in "tag:"
or "group:". But this doesn't work when the user types in "Tag:" or
"GROUP:". This commit makes sure that we match the facet type case
insensitive by lowercasing the string before we match that it starts
with "tag:" or "group:".

Fixes #4080.

### DIFF
--- a/h/static/scripts/controllers/search-bar-controller.js
+++ b/h/static/scripts/controllers/search-bar-controller.js
@@ -324,9 +324,10 @@ class SearchBarController extends Controller {
         currentInput = (currentInput || '').trim();
 
         let typeFilter = FACET_TYPE;
-        if (currentInput.indexOf('tag:') === 0) {
+        const inputLower = currentInput.toLowerCase();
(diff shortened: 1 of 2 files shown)


robertknight:
LGTM

PR #4095, opened 2016-11-18
Fix removing current tag facet from tag sidebar

### LINKED ISSUE #4076: Unicode mixup when adding and removing tags
### Steps to reproduce
1. Go to `/users/<username>/search` for a user who has created annotations with at least two different tags
2. Click on a tag in the sidebar to add it to the search query
3. Click on a _different_ tag in the sidebar to add it to the search query
4. Click on the tag added in step (3) _in the sidebar_ to remove it from the search query

### Expected behaviour
Search query reverts to the query seen after step (2)

### Actual behaviour

The lozenge value for the tag added in step (2) is displayed as `tag:[u'<value>']`:

tag-unicode-bug

chdorner:
When the current search query includes "tag:foo tag:bar" and the user is
clicking on the tag "foo" in the sidebar (which toggles the tag), we end
up with a search query that looks like: "tag:[u'bar']".

This is because we're building up a list of tag facets, removing the one
the user wants to remove, and then set the list of remaining tags on the
parameters (which is a MultiDict). What we need to do is loop through
the list of remaining tags and add them one by one to the MultiDict
parameters.

Fixes #4076.

### DIFF
--- a/h/views/activity.py
+++ b/h/views/activity.py
@@ -318,10 +318,9 @@ def toggle_tag_facet(request):
         # so remove that tag facet.
         tag_facets = _tag_facets(request, parsed_query)
         tag_facets.remove(tag)
-        if tag_facets:
-            parsed_query['tag'] = tag_facets
-        else:
-            del parsed_query['tag']
+        del parsed_query['tag']
+        for tag_facet in tag_facets:
+            parsed_query.add('tag', tag_facet)
     else:
         # The search query is not yet faceted by the given tag, so add a facet
         # for the tag.
--- a/tests/h/views/activity_test.py
+++ b/tests/h/views/activity_test.py
@@ -684,6 +684,16 @@ class TestToggleTagFacet(object):
             'http://example.com/users/foo/search'
             '?q=foo+bar')
 
+    def test_it_preserves_query_when_removing_one_of_multiple_tag_facets(self,
+                                                                         pyramid_request):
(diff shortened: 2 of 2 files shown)


h/views/activity.py
```diff
@@ -319,7 +319,8 @@ def toggle_tag_facet(request):
         tag_facets = _tag_facets(request, parsed_query)
         tag_facets.remove(tag)
         if tag_facets:
-            parsed_query['tag'] = tag_facets
+            for tag_facet in tag_facets:
+                parsed_query['tag'] = tag_facet
```
robertknight: Shouldn't this add a tag facet to the parsed query rather than replacing it?

If I do the following, I get an unexpected result:

1. Create an annotation with 3 tags
2. Go to /users/username/search and click each of the three tags in the sidebar in succession. This should result in `username:<name> tag:<first> tag:<second> tag:<third>` in the search input
3. Remove the last tag from the query by clicking it again in the sidebar

**Expected** - Only the last tag is removed from the query
**Actual** - The first tag added is also removed from the query.

chdorner: Yeah, your're right. This whole thing should be rewritten. This works:
```
        tag_facets = _tag_facets(request, parsed_query)
        tag_facets.remove(tag)
        del parsed_query['tag']
        if tag_facets:
            for tag_facet in tag_facets:
                parsed_query.add('tag', tag_facet)

```

I'll also change the test to have 3 tags

robertknight: Hmm... I can still reproduce the problem with the latest build of this branch.

PR #4096, opened 2016-11-18
Set ES clustername to build version on Jenkins

chdorner:
We are having an issue that sometimes when two builds are running at the same time we get errors from the integration tests saying that the index already exists when it tries to create the index. This is due to the fact that two elasticsearch docker containers will discover each other and join the same cluster. The easiest way to avoid this is to set the cluster name to something unique, like the build version.


nickstenning:
LGTM.

PR #4098, opened 2016-11-18
Add methods for managing aliased indices to Client

nickstenning:
Adds get_aliased_index and update_aliased_index methods to the Elasticsearch client which can be used to interrogate and update index aliases for the open search index.

### DIFF
--- a/src/memex/search/client.py
+++ b/src/memex/search/client.py
@@ -1,6 +1,9 @@
 # -*- coding: utf-8 -*-
 
-import elasticsearch
+from elasticsearch import Elasticsearch
+from elasticsearch.exceptions import NotFoundError
+
+__all__ = ('Client',)
 
 
 class Client(object):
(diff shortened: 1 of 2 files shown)


src/memex/search/client.py
```diff
@@ -15,13 +18,44 @@ class Client(object):
...
+
+    def get_aliased_index(self):
+        """
+        Fetch the name of the underlying index.
+
+        Returns ``None`` if the index is concrete.
+        """
+        try:
+            result = self.conn.indices.get_alias(name=self.index)
+        except NotFoundError:  # self.index is a concrete index
```
chdorner: It seems to throw `NotFoundError` for a missing alias as well, does that matter?

nickstenning: That's the `NotFoundError` we're catching, so I'm not sure what you're asking. We're checking to see if `self.index` is an alias.

chdorner: The docstring says that it returns `None` if the index is concrete, but it will also return `None` when the alias or index does not exist at all. Same thing with the code comment on this line.

PR #4099, opened 2016-11-21
Reindex into new index when aliased

nickstenning:
Adds support for reindexing into a brand-new index when the current index is aliased.

Dropping support for unaliased indices will come later.

### DIFF
--- a/src/memex/search/index.py
+++ b/src/memex/search/index.py
@@ -2,10 +2,12 @@
 """Functions for updating the search index."""
 
 from __future__ import unicode_literals
-from collections import namedtuple
+
+import binascii
 import itertools
 import logging
-from memex._compat import xrange
+import os
+from collections import namedtuple
 
 import elasticsearch
 from elasticsearch import helpers as es_helpers
@@ -13,8 +15,9 @@ from sqlalchemy.orm import subqueryload
 
 from memex import models
 from memex import presenters
+from memex._compat import xrange
 from memex.events import AnnotationTransformEvent
-
+from memex.search.config import configure_index
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


src/memex/search/index.py
```diff
@@ -79,11 +82,35 @@ def delete(es, annotation_id):
...
-    indexing.index_all()
+    """Reindex all annotations into a new index, and update the alias."""
+
+    aliased = es.get_aliased_index() is not None
+
+    # If the index we're using is an alias, then we index into a brand new
+    # index and update the alias at the end.
+    if aliased:
+        new_index = es.index + '-' + _random_id()
+        configure_index(es, index=new_index)
```
chdorner: I believe this will create an index with `number_of_replicas` set to 1 by default, which in our case (current) situation will make the cluster turn yellow because it can't assign the replica shards. Or am I missing something? (I'm also not really sure how to fix this)

src/memex/search/index.py
```diff
@@ -92,25 +119,16 @@ class BatchIndexer(object):
...
-    def __init__(self, session, es_client, request):
+    def __init__(self, session, es_client, request, target_index=None):
         self.session = session
         self.es_client = es_client
         self.request = request
 
         # By default, index into the open index
-        self._target_index = self.es_client.index
-
-    def index_all(self):
```
nickstenning: Simply that it didn't really seem like there was any reason that the `BatchIndexer` needed to know about anything other than where it was indexing to.

src/memex/search/index.py
nickstenning: We can sort this with an [index template](https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-templates.html) in production.

src/memex/search/index.py
chdorner: I see this retrying functionality more part of a batch indexer, mainly because of the "batch" part of it where I don't want to care about transient errors of individual items in the batch

tests/memex/search/index_test.py
```diff
@@ -93,43 +93,100 @@ def log(self, patch):
         return patch('memex.search.index.log')
 
 
-@pytest.mark.usefixtures('BatchIndexer', 'BatchDeleter')
+@pytest.mark.usefixtures('BatchDeleter', 'BatchIndexer', 'configure_index')
 class TestReindex(object):
-    def test_it_indexes_all_annotations(self, BatchIndexer):
-        index.reindex(mock.sentinel.session, mock.sentinel.es, mock.sentinel.request)
+    def test_indexes_annotations(self, es, indexer):
```
chdorner: I don't really see a reason to change the function name from "test_it_..." to "test_...", but that would be nicer in a separate commit for easier reviewing.

src/memex/search/index.py
nickstenning: Agreed, but the actual functionality of `BatchIndexer` is currently split between a) code to return a stream of annotations as fast as possible from a query, and b) code to manage indexing (including, possibly, retries). I'll try and clean this up in subsequent PRs.

PR #4100, opened 2016-11-21
Show groups members based on database

### LINKED ISSUE #4067: List of group members is incorrect
Even when the group facet is the only thing in the search query, the members list in the groups sidebar is not a list of all the group's members. It's a list of all the creators of the annotations that matched the search query - the shared annotations in the group. So:

* Group members who haven't created a shared annotation in the group are missing
* People who have created a shared annotation in the group and then left the group, are still listed as group members

chdorner:
Only get the count of annotations based on the users aggregation.
We used to get the list of group members from the aggregation as well,
but that included members who annotated but left the group, and missed
current members who didn't annotate yet.

Fixes #4067.

### DIFF
--- a/h/views/activity.py
+++ b/h/views/activity.py
@@ -40,12 +40,6 @@ def search(request):
     # Fetch results
     result = query.execute(request, q, page_size=page_size)
 
-    for user in result.aggregations.get('users', []):
-        user['username'] = util.user.split_user(user['user'])['username']
-        user['userid'] = user['user']
-        user['faceted_by'] = _faceted_by_user(request, user['username'], q)
-        del user['user']
-
     groups_suggestions = []
 
     if request.authenticated_user:
@@ -86,6 +80,21 @@ def group_search(request):
     if request.authenticated_user not in group.members:
         return result
 
+    def user_annotation_count(aggregation, userid):
+        for user in aggregation:
+            if user['user'] == userid:
+                return user['count']
+        return 0
+
+    q = query.extract(request)
+    users_aggregation = result.get('aggregations', {}).get('users', [])
+    members = [{'username': u.username,
+                'userid': u.userid,
+                'count': user_annotation_count(users_aggregation, u.userid),
+                'faceted_by': _faceted_by_user(request, u.username, q)}
+               for u in group.members]
+    members = sorted(members, key=lambda k: k['username'].lower())
+
     result['group'] = {
         'created': group.created.strftime('%B, %Y'),
         'description': group.description,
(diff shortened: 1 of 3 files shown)


seanh:
I'm seeing no annotations count next to some users who _do_ have shared annotations in the group, trying to figure out why...

chdorner:
@seanh Maybe because the users don't have any annotations that match the current search? The count is still coming from the aggregation, just the list of users is from the database rather than the aggregation.

seanh:
I don't think that's it. If I go to my group, nothing else in the query except the group, there's no number next to dwhyly, but dhwyly does have some annotations which are shown if you filter the search by him. Also seanh is listed twice:

peek 2016-11-21 17-35

chdorner:
It's going to be hard to debug this group on your local machine. Can you check that it doesn't have anything to do with the fact that those other users might have "hypothes.is" as their authority, and your user (the one you're presumably logged-in as?) has "localhost" as the authority, while the auth domain is "localhost"? Because we use the userid when finding the users in the aggregation.

seanh:
Oh, this is a hypothes.is vs localhost auth domain issue, with production data locally. Sorry Christof, let me fix up my db and re-test

chdorner:
I hope it is, that would be the easiest fix to this problem :)

seanh:
The code looks perfect, I just want to finish reindexing my local search index so I can give it a good test...

h/views/activity.py
```diff
@@ -40,12 +40,6 @@ def search(request):
     # Fetch results
     result = query.execute(request, q, page_size=page_size)
 
-    for user in result.aggregations.get('users', []):
-        user['username'] = util.user.split_user(user['user'])['username']
-        user['userid'] = user['user']
-        user['faceted_by'] = _faceted_by_user(request, user['username'], q)
-        del user['user']
-
```
seanh: Ah, you replaced this with a dict comprehension below, and also removed the need for `split_user()`. Nice refactoring!

h/views/activity.py
```diff
@@ -86,6 +80,21 @@ def group_search(request):
...
+        return 0
+
+    q = query.extract(request)
+    users_aggregation = result.get('aggregations', {}).get('users', [])
+    members = [{'username': u.username,
+                'userid': u.userid,
+                'count': user_annotation_count(users_aggregation, u.userid),
+                'faceted_by': _faceted_by_user(request, u.username, q)}
+               for u in group.members]
+    members = sorted(members, key=lambda k: k['username'].lower())
```
seanh: Was it already sorted like this before?

seanh:
There is a bug to do with deleting user facets but I don't think your code introduced it, just revealed it, I'll open an issue for it

h/views/activity.py
seanh: Interesting. So as the search query and matching annotations counts changed, would the order of the users change as the page reloaded? Anyway, I think alphabetical order is good, nice and consistent!

PR #4103, opened 2016-11-22
Fix removing current user facet from group members sidebar

### LINKED ISSUE #4101: Deleting a user facet results in a bad facet
If the search query is already faceted by a user then clicking on that user in the sidebar again _removes_ them from the query. If there is also at least one other user in the query then this results in a bad facet. For example `user:chdorner` turns into `user:['chdorner']`:

peek 2016-11-21 18-24

chdorner:
When the current search query includes "user:fred user:mary" and the user is
clicking on the username "fred" in the sidebar (which toggles the user facet),
we end up with a search query that looks like: "user:[u'mary']".

This is because we're building up a list of user facets, removing the one
the user wants to remove, and then set the list of remaining user facets on the
parameters (which is a MultiDict). What we need to do is loop through
the list of remaining user facets and add them one by one to the MultiDict
parameters.

Fixes #4101.

(This is the same bug and fix as with the tag facet, see #4095)

### DIFF
--- a/h/views/activity.py
+++ b/h/views/activity.py
@@ -294,10 +294,9 @@ def toggle_user_facet(request):
         # so remove that user facet.
         username_facets = _username_facets(request, parsed_query)
         username_facets.remove(username)
-        if username_facets:
-            parsed_query['user'] = username_facets
-        else:
-            del parsed_query['user']
+        del parsed_query['user']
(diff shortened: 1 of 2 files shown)


robertknight:
LGTM

Issue #4102, opened 2016-11-21
Search results area width changes when there are no results


seanh:
Watch the sidebar move:

![peek 2016-11-21 18-26](https://cloud.githubusercontent.com/assets/22498/20495231/183e319a-b018-11e6-8a32-8eff8d8dee57.gif)

PR #4105, opened 2016-11-22
Add basic "zero state" when no search results

seanh:
On the /search page, if there are no search results, show a grey box
that says "No annotations matched your search."

If on the `/groups/<pubid>/search` page and there is no search query
(other than the group) then the message changes to 'The group "{name}
have not made any annotations yet." When there is a search query other
than just the group the original  "No annotations matched your search"
message is still used.

If on the `/users/<username>/search` page and there is no search query
(other than the user) then the message changes to "{name} hasn't made
any annotations yet". If this is the search page of the authorized user
then "You haven't made any annotations yet" is used. When there is a search
query other than just the user the original  "No annotations matched your
search" message is still used.

The message that's used when you're on your own user page and have no annotations is going to be changed to this one https://cloud.githubusercontent.com/assets/8598901/20438202/f9daacc8-adb6-11e6-815c-3a563000f7fd.png in a following PR.

This also incidentally fixes https://github.com/hypothesis/h/issues/4102

### DIFF
--- a/h/static/styles/partials-v2/_search.scss
+++ b/h/static/styles/partials-v2/_search.scss
@@ -40,17 +40,36 @@
   margin-right: 30px;
 }
 
-.search-result-list {
+.search-result-container--empty {
+  /* The default height of the search container is calculated to so that it
+     stretches to the bottom of the viewport and no further.
+     The px value here is the height of the stuff above the
+     search-result-container. */
+  height: calc(100vh - 134px);
+  margin-bottom: 0;
+}
+
+.search-result-list,
+.search-result-zero {
   flex-basis: 950px;
   max-width: 950px;
+  margin-right: 30px;
+}
 
+.search-result-list {
   // Remove default padding from <ol>
   list-style: none;
   padding-left: 0;
   padding-right: 0;
-
-  margin-right: 30px;
   margin-bottom: 40px;
+}
+
+.search-result-zero {
+  background: $grey-2;
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  padding: 10px;
 }
 
 .search-result__timeframe {
@@ -271,6 +290,14 @@
     margin-top: 25px;
   }
 
+  .search-result-container--empty {
+    /* The default height of the search container is calculated to so that it
+       stretches to the bottom of the viewport and no further.
+       The px value here is the height of the stuff above the
+       search-result-container. */
+    height: calc(100vh - 155px);
+  }
+
   .search-result-nav {
     display: flex;
   }
@@ -279,7 +306,8 @@
     display: none;
   }
 
-  .search-result-list {
+  .search-result-list,
+  .search-result-zero {
     margin-right: 0;
   }
 }
--- a/h/templates/activity/search.html.jinja2
+++ b/h/templates/activity/search.html.jinja2
@@ -334,25 +334,31 @@
     {{groups_suggestions | to_json}}
   </script>
 
-  <div class="search-result-container">
+  <div class="search-result-container {%- if not timeframes %} search-result-container--empty{% endif %}">
     {% if group %}
       {{ search_result_nav(group.name) }}
     {% elif user %}
       {{ search_result_nav(user.name) }}
     {% endif %}
 
-    <ol class="search-result-list {%- if more_info %} search-result-hide-on-small-screens{% endif %}">
-      {% for timeframe in timeframes %}
-        <li class="search-result__timeframe">
-          {{ timeframe.label }}
-        </li>
-        <li>
-          {% for bucket in timeframe.document_buckets.values() %}
-            {{ search_result_bucket(bucket) }}
-          {% endfor %}
-        </li>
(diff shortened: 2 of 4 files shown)


seanh:
I've tested this in desktop Chrome and Firefox (including different device sizes in dev tools) and on iOS Safari. Not tested on IE, Edge or desktop Safari. Tested it both with groups with very short sidebars (e.g. a brand new group with no tags and one member) and with very tall sidebars (e.g. a group with 100 members), and on desktop (where sidebar and search results are shown side by side) and mobile (where search results are shown alone, and sidebar is on a separate page).

h/views/activity.py
```diff
@@ -74,6 +76,8 @@ def user_link(userid):
         'total': result.total,
         'user_link': user_link,
         'username_from_id': username_from_id,
+        # The message that is displayed (only) if there are no search result.
+        'zero_message': _('No annotations matched your search.'),
```
seanh: I thought it was simpler if this value was always present, regardless of whether there's search results or not. The template only renders it if there's no search results.

h/views/activity.py
```diff
@@ -174,6 +183,15 @@ def domain(user):
     if request.authenticated_user == user:
         result['user']['edit_url'] = request.route_url('account_profile')
 
+    if not result.get('q'):
+        if request.authenticated_user == user:
+            result['zero_message'] = _("You haven't made any annotations yet.")
```
seanh: This is the one that will be getting replaced with something more elaborate, in a following PR

seanh:
screenshot from 2016-11-22 18-09-50

screenshot from 2016-11-22 18-10-43

h/templates/activity/search.html.jinja2
```diff
@@ -334,25 +334,31 @@
...
+          </li>
+          <li>
+            {% for bucket in timeframe.document_buckets.values() %}
+              {{ search_result_bucket(bucket) }}
+            {% endfor %}
+          </li>
+        {% endfor %}
+      </ol>
+    {% else %}
+      <div class="search-result-zero {%- if more_info %} search-result-hide-on-small-screens{% endif %}">
```
seanh: I might want to refactor the HTML a bit here, is there's duplication in the application of the `search-result-hide-on-small-screens` class.

seanh:
It's very simple but I guess I should add some tests for the changes to the three Python methods...

h/views/activity.py
```diff
@@ -130,6 +134,11 @@ def user_annotation_count(aggregation, userid):
 
     result['more_info'] = 'more_info' in request.params
 
+    if not result.get('q'):
+        result['zero_message'] = Markup(
+            _('The group &ldquo;{name}&rdquo; have not made any annotations '
```
robertknight: Can we just use unicode chars here rather than `&ldquo;`?

robertknight:
The typography looks on the small side to me, but I'm happy if you want to get the initial version merged and deployed and then ping @conordelahunty for feedback.

h/views/activity.py
seanh: Yeah. And it'll be more testable here, once I actually write some tests!

h/views/activity.py
seanh: I've put the unicode characters in. I've done the message tweak you suggested as well, which differs from in Conor's mockup. @conordelahunty if _has not_ is US and _have not_ is British, this seems a clear rule to follow as Rob suggests?

h/static/styles/partials-v2/_search.scss
```diff
@@ -40,19 +40,38 @@
   margin-right: 30px;
 }
 
-.search-result-list {
+.search-result-container--empty {
+  /* The default height of the search container is calculated to so that it
+     stretches to the bottom of the viewport and no further.
+     The px value here is the height of the stuff above the
+     search-result-container. */
+  height: calc(100vh - 124px);
```
seanh: I've changed this (and the other one for mobile) so that, on Chrome, there's a 10px gap at the bottom of the viewport. Should be enough to mean no scroll bar in other browsers I think

seanh: I'm still seeing a vertical scrollbar on Safari iOS, even though I don't see it on iPhone in Chrome dev tools, but the page seems to be _way_ longer on Safari iOS for some reason, I propose to leave it

seanh:
Regarding the typography requirements, I just used the normal body font size, which I think is what the mockup uses:

f4c3e4f2-adb6-11e6-821e-2788089567de

_All_ of the text on the page looks larger in the mockup than on the actual site to me, but I don't think it's to scale, I think the intention was this text just be normal body font size anyway.

Happy to see it changed though if you want to suggest bigger text to Conor.

seanh:
I'm just gonna add some Python tests to this, then I'll be done with it

seanh:
Ok, tests added, @robertknight this can be reviewed agsin

robertknight:
LGTM

PR #4106, opened 2016-11-22
Fix static asset version mismatch

robertknight:
I suspect that Sentry may sometimes be producing incorrect stacktraces when it processes reports because when it fetches the original JS and sourcemap for a URL it may request `/asset/scripts/foo.js?v1` as specified in the stacktrace but end up getting `/asset/scripts/foo.js?v2` which is the current version.

This PR fixes the general problem that a client can request a particular version of an asset but end up getting a different version by checking the version in the query string against the version in the manifest file. If they differ, the response is 404-ed. The benefit for this approach rather than baking the version into the file path is that it doesn't require any changes to the JS build for the client or server.

In order to keep old versions of assets around for the benefit of Sentry and other clients, we simply set a long max-age on requests so CDNs downstream will retain it in their caches for a long time.

A side benefit of this PR is that it enables browser caching of client and server assets during local development, making repeated loads of the client and server pages faster.

### DIFF
--- a/tests/h/assets_test.py
+++ b/tests/h/assets_test.py
@@ -3,9 +3,12 @@
 from sys import version_info
 from StringIO import StringIO
 
-from mock import patch
+from mock import Mock, patch
+from pyramid import httpexceptions
+from pyramid.response import Response
+import pytest
 
-from h.assets import Environment
+from h.assets import Environment, create_assets_view
 
 if version_info.major == 2:
     open_target = '__builtin__.open'
@@ -34,6 +37,12 @@ def _fake_open(path):
         return StringIO(BUNDLE_INI)
     elif path == 'manifest.json':
         return StringIO(MANIFEST_JSON)
+
+
+def _fake_static_view(file_path, cache_max_age, use_subpath):
+    def fake_view(context, request):
+        return Response('Content of ' + request.path)
+    return fake_view
 
 
 @patch(open_target, _fake_open)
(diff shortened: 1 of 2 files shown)


tests/h/assets_test.py
```diff
@@ -58,6 +67,22 @@ def test_environment_generates_bundle_urls(mtime):
     ]
 
 
+@patch(open_target, _fake_open)
+@patch('os.path.getmtime')
+def test_environment_version_returns_version_if_asset_exists(mtime):
```
robertknight: There is some boilerplate in the tests for `Environment` that could be removed if the tests were grouped into a class. I will save this for a follow-up PR.

nickstenning:
I'll take a closer look at this in a few hours, but it seems like manually implementing the version check for a cachebuster in the query string is a rather roundabout way of achieving something that you could get for free if you put the cachebuster in the pathinfo.

nickstenning:
See also: http://docs.pylonsproject.org/projects/pyramid/en/latest/api/static.html#pyramid.static.ManifestCacheBuster

robertknight:
> I'll take a closer look at this in a few hours, but it seems like manually implementing the version check 
> for a cachebuster in the query string is a rather roundabout way of achieving something that you
> could get for free if you put the cachebuster in the pathinfo.

Right now it is quite convenient because it works for both client and server static assets (of all types) without any added complexity to the frontend build and it works equally well for development and production.

An alternative would be to use a Gulp plugin to add cache busters to file paths as it generates them. This would need to be done per asset type I think and it would also be nice not to leave the `build/` directory cluttered with old versions of the asset. Probably not that difficult though using something like [gulp-hash](https://www.npmjs.com/package/gulp-hash)

nickstenning:
> without any added complexity to the frontend build

Perhaps not, but it adds substantial (and [incidental](http://infiniteundo.com/post/19910636303/intrinsic-and-incidental-complexity) complexity) to this code. I'm not likely to be swayed by an argument that says "X doesn't add complexity over here" if X adds complexity to the system regardless.

I'm also not really sure what complexity you think would be needed in the frontend build. We already generate cache-busting tokens in the manifest, so how much complexity could really be added by moving them to the file basename rather than the query string?

Note that we don't need to write the files anywhere different on disk if we simply replace the default static view with a view that uses the manifest to rewrite the requests. That would still be simpler and much less prone to cache errors caused by cachebust tokens in the query string than the approach here. See: http://docs.pylonsproject.org/projects/pyramid/en/latest/narr/assets.html#advanced-static

Lastly, let's not consider that we need the asset serving to be identical for client and service assets. As we've discussed many times now, the serving of the client assets in the service currently requires far too much hard-coding of bundle contents -- I'd like to see that move into the client itself.

robertknight:
See Slack discussion around an alternative which would involve a custom view callable for static assets: https://hypothes-is.slack.com/archives/public/p1479900316007886

robertknight:
I made a start on this, just noting a couple of related issues along the way, which were able to occur because of the current implementation:

* Currently not all of the static assets are listed in the manifest. Specifically sourcemaps for CSS and JS files are not listed. This applies to both the client and the server.
* Not all assets are served with cache-busted paths. Specifically SVG files are not.

robertknight:
Pyramid's `static_view` used to accept an argument (`cachebust_match`) that could turn a cache-busted path back into a non-cachebusted one so that file paths didn't need to change on disk. That was deprecated/undocumented - see https://github.com/Pylons/pyramid/issues/2675. The way they recommend converting cache-busted URLs back to non-cache busted ones is via a Tween. See https://github.com/Pylons/pyramid/issues/2675#issuecomment-230544321

robertknight:
Sourcemap URLs are currently not cache-busted. In order to fully fix the issue from the original description, we need to make sure that Sentry uses both the correct source and the correct sourcemap when processing reports. That can either be done by adding cachebusters to the sourcemap files or by uploading sourcemaps to Sentry as part of the build.

robertknight:
I'm going to close this at the moment. Investigating this in more detail revealed other issues that need to be fixed first, see https://github.com/hypothesis/h/issues/4117.

Issue #3987, opened 2016-10-20
Flash on deleting lozenges


seanh:
When you delete a lozenge, there is sometimes a noticeable "flash" when all the remaining lozenges seem to disappear for a sec and then reappear

![peek 2016-10-20 18-35](https://cloud.githubusercontent.com/assets/22498/19570969/12326dbc-96f4-11e6-9ac6-21d922884feb.gif)

sean-roberts:
This is worse than just a flash. The flash is the result of a delay between clicking and the page getting reloaded. This also introduces a race condition of the page being reloaded and the input being used to type a new parameter. Really annoying UX

seanh:
I'm going to be rewriting how deleting lozenges work, coming soon...

nickstenning:
Looks like this is fixed.

PR #4107, opened 2016-11-22
Adding max bounds to lozenge growth and proper wrapping of lozenges

### LINKED ISSUE #3988: Too many lozenges overflow the search box
screenshot from 2016-10-20 18-37-14

Or one really long one:

screenshot from 2016-10-20 18-37-46

This also makes it difficult to put the keyboard focus into the search box.

I'm not sure what the desired design for this is?

sean-roberts:
Fixes #3988 

PR adds the maximum width to lozenges that is responsive to the width of the container form. 

We talked about wanting to have the search box remain inline with lozenges if the start wrapping. This means we need to put them inside of the same flex parent which is the lozenge container. There were changes needed to make sure we keep the elements in the right order - that's where properly injecting elements changes come from.

### DIFF
--- a/h/static/scripts/controllers/lozenge-controller.js
+++ b/h/static/scripts/controllers/lozenge-controller.js
@@ -20,8 +20,10 @@ const searchTextParser = require('../util/search-text-parser');
 class LozengeController extends Controller {
   constructor(containerEl, options) {
     super(containerEl, options);
-    const lozengeEl = document.createElement('div');
+    const lozTempContainer = document.createElement('div');
+    let lozengeEl = document.createElement('div');
     let lozengeMarkup = escapeHtml(options.content);
+    const currentLozenges = containerEl.querySelectorAll('.lozenge');
 
     if (searchTextParser.hasKnownNamedQueryTerm(options.content)) {
       const queryTerm = searchTextParser.getLozengeFacetNameAndValue(options.content);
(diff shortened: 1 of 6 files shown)


h/static/styles/partials-v2/_search-bar.scss
```diff
@@ -6,13 +6,14 @@
...
   display: flex;
+  flex-wrap: wrap;
 
   // scope absolute positioned elements
   position: relative;
 }
 
 .search-bar__input {
-  width: 100%;
+  flex: 1;
```
sean-roberts: allows input to fill container

robertknight:
On small screens the navbar has a fixed height so the search input ends up below the navbar:

h/static/scripts/controllers/lozenge-controller.js
```diff
@@ -42,7 +43,17 @@ class LozengeController extends Controller {
...
     lozengeEl.classList.add('js-lozenge');
-    containerEl.appendChild(lozengeEl);
+
+    // append the lozenge after the last lozgenge child
+    // or as the very first element in the container.
+    // We are doing this over appendChild because we don't want to limit
+    // the ability for the container to also hold other elements like an
+    // input field - allowing them to flow together in the same box model
+    if (currentLozenges && currentLozenges.length > 0) {
+      currentLozenges[currentLozenges.length - 1].insertAdjacentElement('afterend', lozengeEl);
```
robertknight: A test would be helpful here.

robertknight:
On tablets and above, I think this works pretty well:

screen shot 2016-11-23 at 09 16 08

robertknight:
On small screens in the common case where there is just one user or group term this often ends up truncating the lozenge to make enough room for the search input, which looks odd. Can you think of a way to address this?

sean-roberts:
@robertknight changes made. 

Regarding the split of mobile search: https://github.com/hypothesis/h/pull/4107#issuecomment-262467375

The leaving room for search is exactly the intent we have decided to build for atm. That said, I did increase the max width from 35% to 55% which feels a little better.

robertknight:
Sweet. This is working nicely.

There can be a slight flicker on page load as the height of the navbar changes before and after the lozenges are created but we can fix that separately.

sean-roberts:
Yep, the flicker is something brought up in #3987. This brings it out more for sure

PR #4119, opened 2016-11-24
Create aliased Elasticsearch indexes by default

nickstenning:
In the development environment the application will automatically create a search index if it doesn't find one. In production the same behaviour can be triggered by a call to

    hypothesis initdb

This commit updates the code so that it creates a randomly named index (the name is based off the supplied index name) such as

    hypothesis-abcd1234

and creates an alias for the supplied index name pointing to the concrete index. This makes it much easier to reindex into a new index without having to redeploy or restart the application.

This commit also removes quite a bit of code that is no longer needed now that "documents" aren't stored in the search index, and adds tests for the config functionality.

_**N.B.** This branch is based on #4118 and will need rebasing once that's merged._

### DIFF
--- a/src/memex/search/config.py
+++ b/src/memex/search/config.py
@@ -9,7 +9,10 @@ these settings in an Elasticsearch instance.
 """
 
 from __future__ import unicode_literals
+
+import binascii
 import logging
+import os
 
 from elasticsearch.exceptions import NotFoundError, RequestError
 
@@ -109,7 +112,7 @@ ANNOTATION_MAPPING = {
     }
 }
 
-ANNOTATION_ANALYSIS = {
+ANALYSIS_SETTINGS = {
     'char_filter': {
         'strip_scheme': {
             'type': 'pattern_replace',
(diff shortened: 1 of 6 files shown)


nickstenning:
Marked as WIP pending #4118 and rebase.

chdorner:
Rebased

src/memex/search/config.py
```diff
@@ -241,6 +237,19 @@ def update_aliased_index(client, new_target):
     })
 
 
+def update_index_settings(client):
```
chdorner: This isn't being called anywhere. The old code was called during `configure_index`. Is this just something that we want to do manually? Or did you forget to call it somewhere? For example when returning early in `init` when the index already exists.

nickstenning: Yeah, sorry, I should have added a comment about this. I think we should add a CLI command to run this manually. Doing it automatically doesn't make a lot of sense to me, especially if it's only going to happen in dev -- seems like it'll just end up as a way to catch people out.

chdorner: Agreed. This way we handle it the same way as we do database migrations 👍

Issue #4004, opened 2016-10-25
New user and group search URLs are confusing


seanh:
[@dwhly said](https://github.com/hypothesis/product-backlog/issues/12#issuecomment-256103074):

> I think the URLs here are getting strange.  I know this is late in the game, but worth asking.
> 
> Why are we implementing a URL of the format `/groups/xyz123/search`  
> 
> Why don't we keep this simple and have URLs of the following structure:
> `/search?q=group:xyz123`   for the explicit search syntax.
> `/groups/xyz123/my-test-group` for the invite link, and after you've joined the group as an alias for
> `/search?q=group:xyz123`  
> 
> But:
> `/groups/xyz123/search?q=tag:foo`
> For what is essentially:
> `/search?q=group:xyz123&tag:foo`
> 
> Just seems odd to me.  I don't get the reasoning behind it.  
> 
> The activity page will be the home for the group.  So special case 
> `/search?q=group:xyz123`  and show `/groups/xyz123/my-test-group` instead.
> 
> If the person then clicks on a tag, switch to the multiple search term format:
> `/search?q=group:xyz123&tag:foo`

seanh:
> The activity page will be the home for the group.  So special case 
> `/search?q=group:xyz123`  and show `/groups/xyz123/my-test-group` instead.
> 
> If the person then clicks on a tag, switch to the multiple search term format:
> `/search?q=group:xyz123&tag:foo`

Could be a good solution. What happens if I'm on `/groups/xyz123/my-test-group` and, rather than clicking on a tag or any other facet link on the page, I just type some keywords into the search box and hit Enter? Does that switch me to `/search?q=group:xyz123&some-keywords`?

Does the group's sidebar shown when on `/search?q=group:xyz123...` or only when on `/groups/xyz123/my-test-group`?

dwhly:
> What happens if I'm on `/groups/xyz123/my-test-group` and, rather than clicking on a tag or any other facet link on the page, I just type some keywords into the search box and hit Enter? Does that switch me to `/search?q=group:xyz123&some-keywords` ?

That was my notion.

My conception of this is that:
1) It's nice to have simple, short vanity URLs for groups, users, etc that people can share, and that people think of as the home for things.
2) it's nice that the 'sharing URL' is the same as the 'home URL'.  which is what we have now w/ the current group landing pages-- so that it doesn't change when you go there, and so that the thing you grab in the address bar is always the same thing you share.
[...]

PR #4120, opened 2016-11-24
Group views.activity into classes

seanh:
Group the view functions in h.views.activity into classes.

This doesn't make any behavioural changes to the code, just:

* Groups the views into classes, reducing some duplication and
  boilerplate, especially in view configuration.

* The check for whether the feature flag is enabled now only needs to be
  done in one place.

* Lots of references to request become self.request

* Some functions renamed, e.g. group_search() -> GroupSearchController.search()

* Group and user search use super() to call the core search view

* The tests are arranged so that it's now on test class per view
  controller class. In some cases this means multiple methods are tested
  in one class, previously it was one test class per function to test

### DIFF
--- a/tests/h/views/activity_test.py
+++ b/tests/h/views/activity_test.py
@@ -12,62 +12,79 @@ from h.views import activity
 
 
 @pytest.mark.usefixtures('paginate', 'query', 'routes')
-class TestSearch(object):
-    def test_it_returns_404_when_feature_turned_off(self, pyramid_request):
+class TestSearchController(object):
+
+    def test_init_returns_404_when_feature_turned_off(self, pyramid_request):
         pyramid_request.feature.flags['search_page'] = False
 
         with pytest.raises(httpexceptions.HTTPNotFound):
-            activity.search(pyramid_request)
+            activity.SearchController(pyramid_request)
 
-    def test_it_checks_for_redirects(self, pyramid_request, query):
-        activity.search(pyramid_request)
+    def test_search_checks_for_redirects(self, controller, pyramid_request, query):
+        controller.search()
 
         query.check_url.assert_called_once_with(pyramid_request,
                                                 query.extract.return_value)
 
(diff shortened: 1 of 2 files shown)


seanh:
It would be good if this could be reviewed quickly, since anything that changes `h/views/activity.py` is going to clash with this. The diff is huge because it indents and moves around a lot of code - it doesn't actually make any behavioral changes, just organizing into classes.

I need to make some changes to `h.views.activity`, especially to the routing configuration, to do <https://github.com/hypothesis/h/issues/4004>. The way `h.views.activity` is on master makes this a pain to do and will mean that I'll only be making the mess worse. So this pull request makes the necessary changes easier to make, by reducing boilerplate and duplication, especially in the view configurations.

robertknight:
I'm still looking through this at the moment but an initial concern I have is that it binds the functionality of the user and group sidebars ever tighter into the `/groups/<pubid>/` and `/users/<username>` routes, which I don't think was the intent of the design.

I wonder if it would make more sense to pull the user and group sidebars into their own panels which are just rendered from activity pages via `{{ panel(...) }}` calls, similar to what we do with the navbar.

tests/h/views/activity_test.py
```diff
@@ -344,182 +519,211 @@ def test_it_redirects_to_user_search(self, pyramid_request):
...
 
+    @pytest.mark.usefixtures('delete_lozenge_request')
+    def test_delete_lozenge_returns_a_redirect(self, controller):
+        result = controller.delete_lozenge()
 
+        assert isinstance(result, httpexceptions.HTTPSeeOther)
 
-@pytest.mark.usefixtures('groups_service', 'routes')
-class TestGroupLeave(object):
+        # This tests that the location redirect to is correct and also that
```
robertknight: Typo "redirect to is correct"

robertknight:
Given the size of the diff I haven't examined every line closely and I still think it might be worth looking into extracting the sidebars out into their panels in future. LGTM.

PR #4121, opened 2016-11-25
Use consistent markup for client and server rendered lozenges

robertknight:
_This is ready for review but marked as WIP because of an issue seen in Firefox (but not Chrome) where removing a client-rendered lozenge does not result in the search bar form being submitted_

The client-side implementation of lozenges used a different DOM-structure than
server-rendered lozenges. This caused several problems:

 - The client JS code for lozenges could not be used with server-rendered
   lozenges.

 - The client DOM structure failed at accessibility by using a `<div>` rather
   than a `<button>` for the close button - hence it did not support keyboard
   focus, identify as a button for screen readers etc.

 - LozengeController did not follow the patterns and conventions of other
   controllers, which all work by taking an existing server-rendered element as
   input and enhancing it with behavior.

Fix this by rendering a `<template>` [1] for client-generated lozenges in the markup
from the server and cloning that as the basis for new lozenges in client code.

Sharing the structure will also make a few other things easier in future:
* We can fix the current flash of "Delete lozenge" text as the delete icon loads by replacing it with an inline SVG
* We can eliminate the FOUC in the navbar on page load by server-rendering the lozenges that correspond to the initial search query

[1] https://developer.mozilla.org/en/docs/Web/HTML/Element/template

### DIFF
--- a/h/static/scripts/controllers/lozenge-controller.js
+++ b/h/static/scripts/controllers/lozenge-controller.js
@@ -21,6 +21,13 @@ class LozengeController extends Controller {
   constructor(element, options) {
     super(element, options);
 
+    // Work-around for HTMLFormElement#submit() failing in Firefox if a submit
+    // button removes itself during click event handler.
+    //
+    // See https://bugzilla.mozilla.org/show_bug.cgi?id=494755#c4 and
+    // https://bugzilla.mozilla.org/show_bug.cgi?id=586329
+    this.refs.deleteButton.type = 'button';
+
     let facetName = '';
     let facetValue = options.content;
 
--- a/h/static/scripts/controllers/lozenge-controller.js
+++ b/h/static/scripts/controllers/lozenge-controller.js
@@ -1,70 +1,65 @@
 'use strict';
 
-const escapeHtml = require('escape-html');
-
 const Controller = require('../base/controller');
 const searchTextParser = require('../util/search-text-parser');
+const { setElementState } = require('../util/dom');
 
 /**
- * Create a lozenge with options.content as its content and append it to containerEl.
+ * A lozenge representing a single search term.
  *
- * A lozenge is made of two parts - the lozenge content and
- * the 'x' button which when clicked removes the lozenge
- * from the container and executes the delete callback provided.
+ * A lozenge consists of two parts - the lozenge content and the 'x' button
+ * which when clicked calls the `deleteCallback` handler passed in the
+ * controller's options.
  *
- * var lozenge = new Lozenge(containerEl, {
- *   content: content,
- *   deleteCallback: deleteCallback,
+ * const lozenge = new Lozenge(element, {
+ *   content,
+ *   deleteCallback,
  * });
  */
 class LozengeController extends Controller {
-  constructor(containerEl, options) {
-    super(containerEl, options);
-    const lozTempContainer = document.createElement('div');
-    let lozengeEl = document.createElement('div');
-    let lozengeMarkup = escapeHtml(options.content);
-    const currentLozenges = containerEl.querySelectorAll('.lozenge');
+
+  constructor(element, options) {
+    super(element, options);
+
+    let facetName = '';
+    let facetValue = options.content;
 
     if (searchTextParser.hasKnownNamedQueryTerm(options.content)) {
       const queryTerm = searchTextParser.getLozengeFacetNameAndValue(options.content);
-      lozengeMarkup = '<span class="lozenge__facet-name">' +
-        escapeHtml(queryTerm.facetName) +
-        '</span>' +
-        ':' +
-        '<span class="lozenge__facet-value">' +
-        escapeHtml(queryTerm.facetValue) +
-        '</span>';
-    }
-    lozengeEl.innerHTML =
-      '<div class="js-lozenge__content lozenge__content">'+
-      lozengeMarkup +
-      '</div>' +
-      '<div class="js-lozenge__close lozenge__close">' +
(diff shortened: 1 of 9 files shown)


h/static/styles/partials-v2/_elements.scss
```diff
@@ -23,3 +23,9 @@ input[type=text],
 textarea {
   appearance: none;
 }
+
+// Hide contents of <template> elements in browsers that do not support
+// this element natively (eg. IE).
+template {
```
robertknight: `<template>` elements are inert and do not render anything in browsers that support it natively. In browsers that don't, `<template>` behaves like an ordinary unrecognized Element.

h/templates/panels/navbar.html.jinja2
```diff
@@ -24,6 +25,9 @@
 {% endmacro %}
 
 <header class="nav-bar">
+  <template id="lozenge-template">
```
robertknight: This is a template for client rendered lozenges. On the client, we clone this each time a new lozenge is created.

h/static/scripts/controllers/lozenge-controller.js
```diff
@@ -1,70 +1,65 @@
...
- * var lozenge = new Lozenge(containerEl, {
- *   content: content,
- *   deleteCallback: deleteCallback,
+ * const lozenge = new Lozenge(element, {
+ *   content,
+ *   deleteCallback,
  * });
  */
 class LozengeController extends Controller {
-  constructor(containerEl, options) {
```
robertknight: This controller used to take the container of the lozenges as input, which was weird because none of the controllers work like this. Now it works like other controllers, taking either a server-rendered lozenge or one cloned from a template on the client as input.

h/static/scripts/controllers/search-bar-controller.js
```diff
@@ -236,10 +243,19 @@ class SearchBarController extends Controller {
      */
     const addLozenge = (content) => {
 
+      const lozengeEl = cloneTemplate(this.options.lozengeTemplate);
```
robertknight: This stamps out a new lozenge from the template (as if it were server-rendered), insert it after the previous lozenge and then upgrade it.

h/static/scripts/tests/controllers/lozenge-controller-test.js
```diff
@@ -1,65 +1,32 @@
 'use strict';
 
 const LozengeController = require('../../controllers/lozenge-controller');
+const lozengeTemplate = require('./lozenge-template');
+const { setupComponent } = require('./util');
 
 describe('LozengeController', () => {
```
robertknight: Now that LozengeController behaves like other Controller subclasses, the tests can be greatly simplified.

h/static/scripts/tests/controllers/lozenge-template.js
```diff
@@ -0,0 +1,19 @@
+'use strict';
+
+// Template copied from navbar.html.jinja2
```
robertknight: For now I've just copied the server template from `navbar.html.jinja2` rather than try to actually use the same markup, although that wouldn't be super hard to do in future.

h/static/scripts/controllers/lozenge-controller.js
```diff
@@ -1,70 +1,65 @@
...
+    if (facetName) {
+      facetName += ':';
+    }
+    this.refs.facetName.textContent = facetName;
+    this.refs.facetValue.textContent = state.facetValue;
+  }
+
+  inputValue() {
+    if (this.state.facetName) {
+      return this.state.facetName + ':' + this.state.facetValue;
```
seanh: It's a small worry that this `:` is being inserted and appended in multiple places

h/static/scripts/controllers/search-bar-controller.js
```diff
@@ -27,13 +28,16 @@ const normalizeStr = function(str) {
...
  */
 class SearchBarController extends Controller {
-  constructor(element) {
-    super(element);
+  constructor(element, options = {}) {
+    super(element, options);
+
+    if (!options.lozengeTemplate) {
+      options.lozengeTemplate = document.querySelector('#lozenge-template');
+    }
```
seanh: Why are there two ways of passing in the lozenge template?

h/static/scripts/controllers/search-bar-controller.js
```diff
@@ -214,15 +224,12 @@ class SearchBarController extends Controller {
      */
     const updateHiddenInput = () => {
       let newValue = '';
-      Array.from(this._lozengeContainer.querySelectorAll('.js-lozenge')).forEach((loz) => {
-
-        let inputValue = loz.querySelector('.js-lozenge__content').textContent;
-
+      lozenges().forEach((loz) => {
+        let inputValue = loz.inputValue();
```
seanh: Oh, nice, so `LozengeController` is now responsible for saying what contribution its lozenge should make to the search input. Nice

h/static/scripts/controllers/search-bar-controller.js
```diff
@@ -236,10 +243,19 @@ class SearchBarController extends Controller {
      */
     const addLozenge = (content) => {
 
+      const lozengeEl = cloneTemplate(this.options.lozengeTemplate);
+      const currentLozenges = this.element.querySelectorAll('.lozenge');
```
seanh: Shouldn't this be `.js-lozenge`?

seanh:
@robertknight Ping me when you're happy that it's working in Firefox

seanh:
I'm happy for you to merge this once your Firefox fix is in, unless you want me to review that fix

PR #4124, opened 2016-11-25
Fix FOUC in lozenges on page load

robertknight:
Use an inline SVG rather than an async-loaded image for the lozenge
delete icon to avoid a layout flicker and flash of `` alt text on
page load.

The padding tweak is needed to ensure correct vertical centering of the
icon - tested in Firefox and Chrome.

### DIFF
--- a/h/static/styles/partials-v2/_lozenge.scss
+++ b/h/static/styles/partials-v2/_lozenge.scss
@@ -27,7 +27,7 @@
   @include reset-button;
 
   font-weight: bold;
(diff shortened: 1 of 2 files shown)


nickstenning:
Does this mean that there's now no hint to screenreaders about the purpose of this button?

robertknight:
> Does this mean that there's now no hint to screenreaders about the purpose of this button?

Good point. I have added a `title` attribute on the `<button>` itself.

nickstenning:
LGTM.

PR #4125, opened 2016-11-25
Add migration for `group.authority` column

chdorner:
This is just adding a new database column. This has to be deployed and run on production before the next pull request is merged.

Being able to choose an authority for a group is the first step in making publisher groups possible. The second step will be a bunch of flags (`readable_by`, `writeable_by`, `joinable_by`, each with a set of possible values: `None`, `"authority"`, `"members"`, `"world"`), with which we can have a group that is scoped by the publishers authority and will automatically allow users with the same authority to write to it.


nickstenning:
LGTM.

PR #4126, opened 2016-11-25
Add group authority field and use it

chdorner:
~**This is based on #4125 and needs rebasing once that one is merged. The migration in #4125 also needs to run in production first.**~

This pull request uses the Group authority field in the code base by writing the value of `request.auth_domain` when a new group gets created. This also includes a migration that sets all current groups authority field to "hypothes.is" and then disallows `NULL` values.

### DIFF
--- a/h/groups/services.py
+++ b/h/groups/services.py
@@ -22,7 +22,7 @@ class GroupsService(object):
         self.user_fetcher = user_fetcher
         self.publish = publish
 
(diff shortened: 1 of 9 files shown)


chdorner:
Rebased and the migration has been run on production. This can be reviewed now

nickstenning:
LGTM.

PR #4127, opened 2016-11-25
Implement updated design for search result bucket sidebar

robertknight:
Implement the design from https://github.com/hypothesis/h/issues/4058#issuecomment-261954000

 * Move in-context link to top of sidebar and URL section to bottom
 * Add icons to the left of the heading for each section

_**Note: This is not using the final icons for the Tags or URL sections, I will submit another PR to update the icons when I have them**_

**Spec:**

**Current screenshot (Ubuntu):**

bucket-sidebar-v3

### DIFF
--- a/h/static/images/icons/account.svg
+++ b/h/static/images/icons/account.svg
@@ -5,7 +5,7 @@
     <desc>Created with Sketch.</desc>
     <defs></defs>
     <g id="Page-1" stroke="none" stroke-width="1" fill="none" fill-rule="evenodd" sketch:type="MSPage">
-        <g id="Artboard-1-Copy" sketch:type="MSArtboardGroup" fill="#000000">
+        <g id="Artboard-1-Copy" sketch:type="MSArtboardGroup" fill="currentColor">
             <circle id="Oval-1" sketch:type="MSShapeGroup" cx="8" cy="4" r="3"></circle>
             <path d="M8,15 C11,15 14,14.4329966 14,12 C14,10.5670034 10,8 8,8 C6,8 2,10.5670034 2,12 C2,14.4329966 5,15 8,15 Z" id="Oval-1-Copy" sketch:type="MSShapeGroup"></path>
         </g>
--- a/h/static/styles/partials-v2/_search.scss
+++ b/h/static/styles/partials-v2/_search.scss
@@ -225,7 +225,18 @@
                  // work).
 }
 
+$stats-icon-column-width: 20px;
+
+.search-bucket-stats__icon {
+  width: 12px;
+  margin-right: $stats-icon-column-width - 12px;
+}
+
 .search-bucket-stats__key {
+  display: flex;
+  flex-direction: row;
+  align-items: center;
+
   margin-bottom: 2px;
   color: $grey-6;
   font-weight: bold;
@@ -233,7 +244,8 @@
 
 .search-bucket-stats__val {
   color: $grey-6;
-  margin-bottom: 20px;
+  margin-bottom: 15px;
+  margin-left: $stats-icon-column-width;
 }
 
 .search-bucket-stats__url,
@@ -249,11 +261,10 @@
 }
 
 .search-bucket-stats__incontext-link {
-  display: block;
-  margin-top: 5px;
-
   color: $grey-6;
-  font-weight: bold;
+  &:hover {
+    text-decoration: underline;
+  }
 }
 
 .search-bucket-stats__incontext-icon {
--- a/h/templates/activity/search.html.jinja2
+++ b/h/templates/activity/search.html.jinja2
@@ -39,7 +39,7 @@
       <a class="search-bucket-stats__incontext-link"
          href="{{ bucket.incontext_link(request) }}"
          target="_blank">
-         Visit annotations in context
+         {% trans %}Visit annotations in context{% endtrans %}
        </a>
     </div>
     <div class="search-bucket-stats__val"></div>
@@ -47,7 +47,7 @@
   {% if bucket.tags %}
     <div class="search-bucket-stats__key">
       {{ svg_icon('up-right-arrow', css_class='search-bucket-stats__icon') }}
-      Tags
+      {% trans %}Tags{% endtrans %}
     </div>
     <ul class="search-bucket-stats__val">
       {% for tag in bucket.tags %}
@@ -58,7 +58,7 @@
   {% endif %}
   <div class="search-bucket-stats__key">
     {{ svg_icon('account', css_class='search-bucket-stats__icon') }}
-    Annotators
+    {% trans %}Annotators{% endtrans %}
   </div>
   <ul class="search-bucket-stats__val">
     {% for user in bucket.users %}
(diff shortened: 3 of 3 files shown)


h/templates/activity/search.html.jinja2
```diff
@@ -33,26 +33,20 @@
...
-             target="_blank">
-             Visit annotations in context {{ svg_icon('up-right-arrow', css_class='search-bucket-stats__incontext-icon') }}
-           </a>
-        </div>
-        {% endif %}
+      {{ svg_icon('up-right-arrow', css_class='search-bucket-stats__icon') }}
+      <a class="search-bucket-stats__incontext-link"
+         href="{{ bucket.incontext_link(request) }}"
+         target="_blank">
+         Visit annotations in context
```
seanh: Should this be inside `{% trans %} ... {% endtrans %}` tags? I don't think we have any agreed on i18n standards, and of course we don't localise h into any other languages, but I've been trying to remember to use `_()` in Python and `{% trans %}` in Jinja2 for user-visible strings. There's some docs on `{% trans %}` etc here: http://docs.ckan.org/en/latest/contributing/string-i18n.html#internationalizating-strings-in-jinja2-templates

seanh: I see the uninternationalised string wa already here though

h/static/images/icons/account.svg
```diff
@@ -5,9 +5,9 @@
     <desc>Created with Sketch.</desc>
     <defs></defs>
     <g id="Page-1" stroke="none" stroke-width="1" fill="none" fill-rule="evenodd" sketch:type="MSPage">
-        <g id="Artboard-1-Copy" sketch:type="MSArtboardGroup" fill="#000000">
+        <g id="Artboard-1-Copy" sketch:type="MSArtboardGroup" fill="currentColor">
```
seanh: Does `fill="currentColor` enable the colour of the user icon to be changed by CSS? I was fiddling with `color` and `fill` CSS classes on some of the elements in dev tools but couldn't seem to get it to change

h/static/styles/partials-v2/_search.scss
```diff
@@ -225,15 +225,27 @@
                  // work).
 }
 
+$stats-icon-column-width: 20px;
```
seanh: Ah, cool. I probably should have used variables in some of the CSS I've written recently :)

h/static/styles/partials-v2/_search.scss
```diff
@@ -249,11 +261,10 @@
 }
 
 .search-bucket-stats__incontext-link {
-  display: block;
-  margin-top: 5px;
-
   color: $grey-6;
-  font-weight: bold;
+  &:hover {
+    text-decoration: underline;
```
seanh: Just noting that, unless I'm missing the pattern, we seem to be quite inconsistent about whether links are underlined or not and whether they get an underline when hovered or not. I think 3/4 possible combinations of those two are present on the search page alone! Not this PR's fault though, I think it has already happened

h/templates/activity/search.html.jinja2
```diff
@@ -33,26 +33,20 @@
 {# Card displaying statistics about a bucket of annotations. #}
 {% macro search_bucket_stats(bucket) %}
 <div class="search-bucket-stats">
-  {% if bucket.uri %}
+  {% if bucket.uri and bucket.incontext_link(request) %}
```
seanh: Seems like a good idea

h/templates/activity/search.html.jinja2
```diff
@@ -33,26 +33,20 @@
...
-        </div>
-        {% endif %}
+      {{ svg_icon('up-right-arrow', css_class='search-bucket-stats__icon') }}
+      <a class="search-bucket-stats__incontext-link"
+         href="{{ bucket.incontext_link(request) }}"
+         target="_blank">
+         Visit annotations in context
+       </a>
     </div>
+    <div class="search-bucket-stats__val"></div>
```
seanh: So the "Visit annotations in context" search bucket stats key doesn't actually have a value to go with it like the others do, but it requires a val div anyway for layout purposes?

I guess you could also wrap each key / value pair in a containing `<section>` or something and put the bottom margin on that instead of on the value div, but this may be simpler.

h/templates/activity/search.html.jinja2
```diff
@@ -33,26 +33,20 @@
...
   {% endif %}
   {% if bucket.tags %}
     <div class="search-bucket-stats__key">
+      {{ svg_icon('up-right-arrow', css_class='search-bucket-stats__icon') }}
```
seanh: `svg_icon()` is great

h/templates/activity/search.html.jinja2
```diff
@@ -73,6 +68,17 @@
       </li>
     {% endfor %}
   </ul>
+  {% if bucket.uri %}
+    <div class="search-bucket-stats__key">
+      {{ svg_icon('up-right-arrow', css_class='search-bucket-stats__icon') }}
+      URL
```
seanh: I guess maybe this one doesn't need i18n!

seanh:
I wonder if the addition of a new column containing icons has reduced the amount of space available for the stats text itself now? It gets squeezed quite thin before we hit the breakpoint where the annotation cards jump to the left. We may need to move that breakpoint earlier now?

screenshot from 2016-11-28 11-10-51

h/static/images/icons/account.svg
robertknight: Yes. `currentColor` causes the fill to use the inherited value of the `color` CSS property.

h/templates/activity/search.html.jinja2
robertknight: It probably ought to be. I'm guessing @lenazun would have a better idea if "URL" would ever be changed by translation.

h/templates/activity/search.html.jinja2
robertknight: You're right - the empty `<div>` here is just to achieve layout consistency.

h/static/styles/partials-v2/_search.scss
robertknight: Perhaps raise this in #design with the concrete examples that you can see?

h/templates/activity/search.html.jinja2
robertknight: You're right - it should be marked for translation.

h/templates/activity/search.html.jinja2
robertknight: Thank GitHub for the idea ;)

h/static/styles/partials-v2/_search.scss
robertknight: Please do :) Making the rationale for particular sizes in CSS clear makes it easier to avoid breaking things when refactoring.

robertknight:
Testing layout across a few common device sizes, the layout issue partially affects the iPad layout, but I think a bit more thought is required to figure out how to resolve it. I'll address that separately.

screen shot 2016-11-28 at 12 16 41

Issue #4117, opened 2016-11-24
Cache busting issues with static assets


robertknight:
## Current issues with static assets and cache busting

- **Static asset URLs use query strings as cache busters**. These are used when generating the asset but ignored when serving it. Consequently, it is possible that client may request 'foo.js' version 'v1' via a request for '/assets/foo.js?v1' but actually get a newer version 'v2'. This is a problem for Sentry if it tries to process an exception report for Hypothesis release v.X after Hypothesis v.X+1 has been released.
- **Sourcemaps URLs do not include cache-busting tokens**, so Sentry may end up trying to process an exception for '/assets/foo.js?v2' by requesting '/assets/foo.js.map' and actually end up using a cached sourcemap (where the sourcemap has been cached either by Sentry itself, or CloudFlare) for an earlier release.
- **Static assets are served with a short cache expiration of 4 hours**, despite the fact that assets are immutable (provided they are cache-busted). This is a small inefficiency that may lead clients to download resources they already have.
- **Not all resources are served with cache-busting tokens in the URL**. eg. SVG images are currently served without. This may lead to users seeing old versions of assets.

## Proposed changes

- Fix manifest.json generation in client and server to include all assets (see https://github.com/hypothesis/h/pull/4106#issuecomment-262760621)
- Fix image asset serving to include cache-busting tokens (see https://github.com/hypothesis/h/pull/4106#issuecomment-262760621)
- Modify manifest generation and asset serving to use path rather than query-string cache busting tokens. This will require either a custom static view callable in the Python code to map the cache-busted URL to the path on disk or a tween to map the request's subpath to a version without the cache busting token. See https://github.com/Pylons/pyramid/issues/2675#issuecomment-230544321
- Rewrite sourcemap URLs in JS bundles to use cache-busted URLs
- Set the `cache_max_age` parameter in the `static_view` to a large value so that clients do not re-download assets that they already have.

robertknight:
Status update for May 2023: The code for serving static assets was extracted into the [h-assets](https://github.com/hypothesis/h-assets) package. That still uses query strings as cache busters, but it does [validate them](https://github.com/hypothesis/h-assets/blob/7561d45a8d9ae83b754fba00f440e2423591b1d6/src/h_assets/view.py#L15) and return a 404 if the cache buster query string is invalid.

This resolves the first problem where Sentry could end up receiving a bug report for version X and end up fetching the static asset for a later release Y.

However this problem is still open:

[...]

PR #4128, opened 2016-11-28
Use cache-busted URLs for image assets

robertknight:
Replace hard-coded '/asset' URL paths in templates with cache-busted
URLs generated from the site's static asset manifest.

 * Add `asset_url` helper for retrieving the cache-busted URL for an
   asset with a given path and use it in templates.

 * Fix missing entries in asset manifest by including subdirectories
   (eg. 'images/icons/*.svg') and all extensions (eg. sourcemaps)

See #4117

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -208,7 +208,7 @@ gulp.task('watch-images', function () {
   gulp.watch(imageFiles, ['build-images']);
 });
 
-var MANIFEST_SOURCE_FILES = 'build/@(fonts|images|scripts|styles)/*.@(js|css|woff|jpg|png|svg)';
+var MANIFEST_SOURCE_FILES = 'build/@(fonts|images|scripts|styles)/**/*.*';
 
 /**
  * Generate a JSON manifest mapping file paths to
--- a/h/assets.py
+++ b/h/assets.py
@@ -85,12 +85,16 @@ class Environment(object):
         Returns the URLs at which all files in a bundle are served,
         read from the asset manifest.
         """
-        manifest = self.manifest.load()
         bundles = self.bundles.load()
 
-        def asset_url(path):
-            return '{}/{}'.format(self.assets_base_url, manifest[path])
-        return [asset_url(path) for path in bundles[bundle]]
+        return [self.url(path) for path in bundles[bundle]]
+
+    def url(self, path):
+        """
+        Return the cache-busted URL for an asset with a given path.
+        """
+        manifest = self.manifest.load()
+        return '{}/{}'.format(self.assets_base_url, manifest[path])
 
 
 def _add_cors_header(wrapped):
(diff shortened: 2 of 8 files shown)


h/assets.py
```diff
@@ -147,3 +151,5 @@ def includeme(config):
     # integration can be configured in app.py
     config.registry['assets_env'] = assets_env
     config.registry['assets_client_env'] = assets_client_env
+
+    config.add_request_method(assets_env.asset_url, 'asset_url')
```
nickstenning: These asset urls are used only in the templates so perhaps a neater way of exposing this function there would be to follow the pattern in [app.py](https://github.com/hypothesis/h/blob/f671cedcc615c0b55e7c21a7c5fe63fe7051c7fc/h/app.py#L20) and add the `asset_url` function as a Jinja2 global.

h/assets.py
```diff
@@ -85,12 +85,16 @@ def urls(self, bundle):
...
         """
-        manifest = self.manifest.load()
         bundles = self.bundles.load()
 
-        def asset_url(path):
-            return '{}/{}'.format(self.assets_base_url, manifest[path])
-        return [asset_url(path) for path in bundles[bundle]]
+        return [self.asset_url(path) for path in bundles[bundle]]
+
+    def asset_url(self, path):
```
nickstenning: This class is `h.assets.Environment`, so `asset_url` is possibly a bit redundant. The method above is called `urls`, so maybe call this one `url`?

h/assets.py
```diff
@@ -85,12 +85,16 @@ def urls(self, bundle):
...
+        """
+        Return the cache-busted URL for an asset with a given path.
+        """
+        manifest = self.manifest.load()
```
nickstenning: I think we want to be a little careful here, as previously we were doing one call to `load()` per bundle, and now we're doing one call to load per file (including when those files are referenced from a bundle).

The manifest is cached, sure, but this is still going to do at least one `stat()` call, which isn't ideal if we're doing it dozens of times during a single request. Especially in production, where we know the manifest isn't going to change, it might be nice to load it just once.

gulpfile.js
```diff
@@ -208,7 +208,7 @@ gulp.task('watch-images', function () {
   gulp.watch(imageFiles, ['build-images']);
 });
 
-var MANIFEST_SOURCE_FILES = 'build/@(fonts|images|scripts|styles)/*.@(js|css|woff|jpg|png|svg)';
```
nickstenning: This seems to already cover the relevant file extensions, so do we need to change it?

robertknight: It doesn't cover sourcemaps (`.{js,css}.map`) or subdirectories (`images/icons/foo.svg`)

nickstenning: Ok!

nickstenning:
Ok, changes LGTM. I'd like you to review the performance of the application when this goes out and ensure that all the extra `stat()` calls we'll be doing don't have an adverse effect.

PR #4129, opened 2016-11-28
Update pyramid dependency for memex

chdorner:
We already ran this with Pyramid 1.7 in production for ~2 months (since 4ad66a7904b7a0ceb21480e81432dd4164b8cc86) without upgrading the memex dependency. This now fixes that (and gets a new version of pyramid-tm).


nickstenning:
LGTM.

PR #4130, opened 2016-11-28
Factor message handlers out of main `handle_message` function

nickstenning:
This is mostly preparatory work for improvements to the WebSocket API. In this commit we:

- Move the business logic used for handling websocket messages into individually-tested handler functions.

- Improve the error handling behaviours:
     + Close the connection on invalid JSON input
     + Only catch expected errors when parsing filter messages.

- Standardise the incoming message format (all messages should have a "type" field -- those using "messageType" or lacking type fields should be updated.)

### DIFF
--- a/h/streamer/websocket.py
+++ b/h/streamer/websocket.py
@@ -14,6 +14,10 @@ from h.streamer import filter
 
 log = logging.getLogger(__name__)
 
+# Mapping incoming message type to handler function. Handlers are added inline
+# below.
+MESSAGE_HANDLERS = {}
+
 # An incoming message from a WebSocket client.
 Message = namedtuple('Message', ['socket', 'payload'])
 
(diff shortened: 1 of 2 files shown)


h/streamer/websocket.py
```diff
@@ -14,6 +14,10 @@
 
 log = logging.getLogger(__name__)
 
+# Mapping incoming message type to handler function. Handlers are added inline
+# below.
+MESSAGE_HANDLERS = {}
```
chdorner: Hmm, I personally think it would be nicer if the handlers are added here, then I don't have to search through the code to see which message types we support. What's the reason to have them inline after the function definition?

tests/h/streamer/websocket_test.py
```diff
@@ -94,23 +93,88 @@ def test_socket_send_json_skips_when_terminated(fake_environ, fake_json, fake_so
...
+        socket = mock.Mock(spec_set=['close'])
+        message = websocket.Message(socket, payload='{"foo":"bar"}')
+
+        websocket.handle_message(message)
+
+        unknown_handler.assert_called_once_with(socket,
+                                                {"foo": "bar"},
+                                                session=None)
+
+    def test_uses_unknown_handler_for_unknown_type(self, unknown_handler):
```
chdorner: Might be good to add a test for when there is no `type` key in the payload? It currently calls the "unknown handler" as well, might be beneficial to have a test for that so we don't accidentally change this behaviour in the future.

chdorner:
It seems to work well, we should obviously give it another test run on staging before pushing it out to production. But other than the two questions posted inline: 👍

tests/h/streamer/websocket_test.py
nickstenning: Yep! That's what the test above this one does :)

chdorner: Haha, yes. missed that one

PR #4131, opened 2016-11-28
Upgrade SQLAlchemy to 1.1

chdorner:
Version 1.1 provides better handling of the enum datatype which is needed for the next chunk of work for publisher groups.


nickstenning:
LGTM.

PR #4132, opened 2016-11-28
Add a 'Back' link to the footer of the accounts forms

robertknight:
Add a 'Back' link to the footer of the accounts forms

Add a Back link at the bottom of the accounts forms that takes the user
back to the previous page, if they came from the 'Edit Profile' link in
the sidebar.

Fixes hypothesis/product-backlog#42


h/templates/panels/back_link.html.jinja2
```diff
@@ -0,0 +1,4 @@
+{% if back_location and back_label %}
+<a class="link link--footer"
```
nickstenning: Having `link--footer` in here seems a bit odd if this is supposed to be a "reusable" panel...?

PR #4139, opened 2016-11-29
Change group route

seanh:
Fixes https://github.com/hypothesis/h/issues/4004

<del>This is marked as WIP because it depends on <https://github.com/hypothesis/h/pull/4136>.</del>

This pull request changes the `/groups/{pubid}/search` URL to `/groups/{pubid}/{slug}`.

This is not nearly as simple as it may sound because `/groups/{pubid}/{slug}` is already a heavily overloaded URL (with different behaviours depending on whether you're logged in or not, a member of the group or not, whether the activity_pages or search_page feature flags are on or not, redirects when the slug is missing or incorrect...).  The `/groups/{pubid}/search` also has many different behaviours (for just getting the page and for submitting the search form and clicking various buttons in the sidebar) and these are now all piled onto the `/groups/{pubid}/{slug}` URL as well.  

In addition, the existing  `/groups/{pubid}/{slug}` views are in `h.views.groups` whereas the `/groups/{pubid}/search` are in `h.views.activity`, and the groups ones use traversal whereas the activity ones use URL dispatch.

Merging these two URLs into one has consequences for the structure and factoring of the codebase where the views for the two URLs (`h.views.groups` and `h.views.activity`) are concerned.

## Behaviour

The new behaviour that this pull request implements is:

* If you aren't logged in then `/groups/{pubid}/{slug}` shows the "Login to join group" page. Either the classic or the new version of this page, depending on the activity_pages feature flag.  

* If you are logged in, but aren't a member of the group, then `/groups/{pubid}/{slug}` shows the "Join group" page. Either the classic or the new version of this page, depending on the activity_pages feature flag.

* If you're already a member of the group then `/groups/{pubid}/{slug}` shows the group's activity search page if the search_page feature flag is on or the classic group page if it's not.

* `/groups/{pubid}`, `/groups/{pubid}/` and `/groups/{pubid}/{something_else}` all redirect to `/groups/{pubid}/{slug}` in all cases

* `/groups/{pubid}[/[{anything}]]` 404s if no group with the given pubid exists, in all cases (previously `/groups/{does_not_exist}/search` would show an empty search page, _not_ a 404, but `/groups/{does_not_exist}[/[{anything_else}]]` would 404)

* Each of these different "versions" of the page is implemented as a separate view function or method, and we use Pyramid routing to select the correct view depending on the situation. We don't have Python logic in the view functions saying "if the user is a member then X else Y" etc.

* The `/search` page no longer redirects to the group's page if you search for a group that doesn't exist. This is because we don't want to send users to a 404 page.

## Pyramid routing

The `/groups/{pubid}/search` page was implemented by the `"activity.group_search"` route and the views in `h.views.activity.GroupSearchController` and `h.views.activity.GroupUserSearchController` .

The `/groups/{pubid}/{slug}` page, which is already used on master for the "Login to join group", "Join group", and classic group home pages, was implemented by the `"group_read"` route and the views in `h.views.groups`.

It's not possible for both the `"activity.group_search"` and the `"group_read"` route to use the `/groups/{pubid}/{slug}` URL without overriding each other in unwanted ways. (Even if it is possible, this would be very brittle and not easily testable routing.)

In this pull request this has been reconciled by having the group activity search page views in `h.views.activity` share the `"group_read"` route with the other group views in `h.views.groups`.  The `"activity.group_search"` route is no longer used. The view configurations on the various views that use the `"group_read"` route have been tightened up to correctly and more explicitly declare which views should be called under which circumstances:

* `h.views.activity.GroupSearchController` (the new group activity search page) is only called when the user is a member of the group and the `'search_page'` feature is on.

  Once activity pages is released and the `'search_page'` feature flag is removed we can remove the `'search_page'` predicate from this.

* `h.views.groups.read()` (the classic group home page) is only called when the user is a member of the group and the `search_page` feature is _not_ on.

  Once activity pages is released and the `'search_page'` feature flag is removed we can delete this view function.

* `h.views.groups.GroupJoinController()` is only called when the user is logged in and is _not_ a member of the group

* `h.views.groups.read_unauthenticated()` is only called when the user is _not_ logged in.

## Traversal

The `"activity.group_search"` route (which has been deleted) used [Pyramid's URL dispatch mechanism](http://docs.pylonsproject.org/projects/pyramid//en/latest/narr/urldispatch.html) to look up and call the view for a request.

The `"group_read"` route, which is now shared by the previous `"group_read"` views and the views that used to use the `"activity.group_search"` route, uses [Pyramid's traversal mechanism](http://docs.pylonsproject.org/projects/pyramid//en/latest/narr/hellotraversal.html).

So the group activity search page views in `h.views.activity.GroupSearchController` now use traversal not URL dispatch, to match with the existing `"group_read"` views.

This means that the views no longer need to look up the group in the database, nor deal with the case when the group doesn't exist, because traversal handles this for them.

## Known issues

* The user activity page's route is called `"activity.user_search"` whereas the group activity page's route is `"group_read"`.

* The views for the `"group_read"` route are spread across two modules `h.views.activity` and `h.views.groups`.

  `h.views.activity.GroupSearchController` shares the `"group_read"` route with the other group views in `h.views.groups`. On the other hand, it shares much code with the other activity search views in `h.views.activity`.

In a follow-up pull request I think these issues could be resolved by:

* Moving all the group views from `h.views.groups` into `h.views.activity` and renaming the `"group_read"` route to `"activity.group_search"` to match with `"activity.user_search"`. Or

* Move `h.views.activity.GroupSearchController` into `h.views.groups` so that all of the group views are together in one place.  Also move `h.views.activity.UserSearchController` into a new `h.views.users` module to mirror this organisation.

  `GroupSearchController` and `UserSearchController` would both still inherit from `h.views.activity.SearchController`, which is implemented in a different module from either of the inheriting classes, but maybe that's okay.

  Several shared private helper functions in `h.views.activity` would also need to be made available to `GroupSearchController` and `UserSearchController` which are now in different files. Making these helper functions into methods of `SearchController` might work, or move them into a helpers module.

### DIFF
--- a/h/views/activity.py
+++ b/h/views/activity.py
@@ -20,6 +20,7 @@ from h.links import pretty_link
 from h.paginator import paginate
 from h import util
 from h.util.user import split_user
+from h.views import custom_predicates
 
 
 PAGE_SIZE = 200
@@ -90,27 +91,23 @@ class SearchController(object):
         }
 
 
-@view_defaults(route_name='activity.group_search',
-               renderer='h:templates/activity/search.html.jinja2')
+@view_defaults(route_name='group_read',
+               renderer='h:templates/activity/search.html.jinja2',
+               custom_predicates=[custom_predicates.has_read_permission])
 class GroupSearchController(SearchController):
-    """View callables unique to the "activity.group_search" route."""
+    """View callables unique to the "group_read" route."""
+
+    def __init__(self, group, request):
+        super(GroupSearchController, self).__init__(request)
+        self.group = group
 
     @view_config(request_method='GET')
     def search(self):
         result = super(GroupSearchController, self).search()
 
-        pubid = self.request.matchdict['pubid']
-        result['opts'] = {'search_groupname': pubid}
+        result['opts'] = {'search_groupname': self.group.name}
 
-        try:
-            group = (self.request.db.query(models.Group)
-                     .filter_by(pubid=pubid).one())
-        except exc.NoResultFound:
-            return result
-
-        result['opts']['search_groupname'] = group.name
-
-        if self.request.authenticated_user not in group.members:
+        if self.request.authenticated_user not in self.group.members:
             return result
 
         def user_annotation_count(aggregation, userid):
@@ -128,30 +125,30 @@ class GroupSearchController(SearchController):
                     'faceted_by': _faceted_by_user(self.request,
                                                    u.username,
                                                    q)}
-                   for u in group.members]
+                   for u in self.group.members]
         members = sorted(members, key=lambda k: k['username'].lower())
 
         result['group'] = {
-            'created': group.created.strftime('%B, %Y'),
-            'description': group.description,
-            'name': group.name,
-            'pubid': group.pubid,
+            'created': self.group.created.strftime('%B, %Y'),
+            'description': self.group.description,
+            'name': self.group.name,
+            'pubid': self.group.pubid,
             'url': self.request.route_url('group_read',
-                                          pubid=group.pubid,
-                                          slug=group.slug),
+                                          pubid=self.group.pubid,
+                                          slug=self.group.slug),
             'members': members,
         }
 
-        if self.request.has_permission('admin', group):
-            result['group_edit_url'] = self.request.route_url('group_edit',
-                                                              pubid=pubid)
+        if self.request.has_permission('admin', self.group):
+            result['group_edit_url'] = self.request.route_url(
+                'group_edit', pubid=self.group.pubid)
 
         result['more_info'] = 'more_info' in self.request.params
 
         if not result.get('q'):
             result['zero_message'] = Markup(_(
                 'The group “{name}” has not made any annotations yet.').format(
-                    name=Markup.escape(group.name)))
+                    name=Markup.escape(self.group.name)))
 
         return result
 
@@ -221,10 +218,29 @@ class GroupSearchController(SearchController):
         new_params['q'] = parser.unparse(parsed_query)
 
         location = self.request.route_url(
-            'activity.group_search', pubid=self.request.matchdict['pubid'],
+            'group_read', pubid=self.group.pubid, slug=self.group.slug,
             _query=new_params)
 
         return httpexceptions.HTTPSeeOther(location=location)
+
+    @view_config(request_method='POST',
+                 request_param='more_info')
+    def more_info(self):
+        return _more_info(self.group, self.request)
+
+    @view_config(request_method='POST',
+                 request_param='back')
+    def back(self):
+        return _back(self.group, self.request)
+
+    @view_config(request_param='delete_lozenge')
+    def delete_lozenge(self):
+        return _delete_lozenge(self.request)
+
+    @view_config(request_method='POST',
+                 request_param='toggle_tag_facet')
+    def toggle_tag_facet(self):
+        return _toggle_tag_facet(self.group, self.request)
 
 
 @view_defaults(route_name='activity.user_search',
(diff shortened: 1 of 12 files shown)


h/views/activity.py
```diff
@@ -90,27 +92,25 @@ def user_link(userid):
         }
 
 
-@view_defaults(route_name='activity.group_search',
-               renderer='h:templates/activity/search.html.jinja2')
+@view_defaults(route_name='group_read',
+               renderer='h:templates/activity/search.html.jinja2',
+               custom_predicates=[custom_predicates.has_read_permission,
+                                  custom_predicates.feature('search_page')])
```
seanh: We only have to state this once, and it means that _all_ view methods in this class will only be called if the user is a member of the group ("has read permission" on the context object, which is the group) and the search_page feature flag is on.

h/views/activity.py
```diff
@@ -90,27 +92,25 @@ def user_link(userid):
...
 class GroupSearchController(SearchController):
-    """View callables unique to the "activity.group_search" route."""
+    """View callables unique to the "group_read" route."""
+
+    def __init__(self, group, request):
+        super(GroupSearchController, self).__init__(request)
+        self.group = group
+        check_slug(self.group, self.request)
```
seanh: So all the view methods will redirect to the correct slug if it's wrong or missing

h/views/activity.py
```diff
@@ -90,27 +92,25 @@ def user_link(userid):
...
+@view_defaults(route_name='group_read',
+               renderer='h:templates/activity/search.html.jinja2',
+               custom_predicates=[custom_predicates.has_read_permission,
+                                  custom_predicates.feature('search_page')])
 class GroupSearchController(SearchController):
-    """View callables unique to the "activity.group_search" route."""
+    """View callables unique to the "group_read" route."""
+
+    def __init__(self, group, request):
+        super(GroupSearchController, self).__init__(request)
```
seanh: `SearchController.__init__()` raises `HTTPNotFound` of the search_page feature flag is off, so all view methods in this class will 404 if the flag is off (but of course due to the `@view_defaults` above they would not be selected by Pyramid if the flag was off anyway...)

h/views/activity.py
```diff
@@ -90,27 +92,25 @@ def user_link(userid):
...
         result = super(GroupSearchController, self).search()
 
-        pubid = self.request.matchdict['pubid']
-        result['opts'] = {'search_groupname': pubid}
-
-        try:
-            group = (self.request.db.query(models.Group)
-                     .filter_by(pubid=pubid).one())
-        except exc.NoResultFound:
-            return result
```
seanh: Now that we're using traversal the view no longer needs to look up the group in the db (it's available as `self.group`) or handle the case where the group doesn't exist (Pyramid won't call the view in this case)

h/views/groups.py
```diff
@@ -112,23 +114,12 @@ def _update_group(self, appstruct):
 @view_config(route_name='group_read',
              request_method='GET',
              renderer='h:templates/groups/share.html.jinja2',
-             effective_principals=security.Authenticated)
+             effective_principals=security.Authenticated,
+             custom_predicates=[custom_predicates.has_read_permission,
+                                not_(custom_predicates.feature('search_page'))])
```
seanh: This is the view function for the classic (pre-activity pages) group page. Tighten up the view config on this so that's it's only called if the user is a member of the group and the search_page feature flag is off.

Note that this single view function used to handle the group URL when the user was or was not a member of the group (as long as they were logged in). This has now been split out so that this view is _only_ for logged in users who _are_ members of the group.

h/views/groups.py
```diff
@@ -112,23 +114,12 @@ def _update_group(self, appstruct):
...
-    # 'read' permission on the group. In this case, we show them the join
-    # page.
-    if not request.has_permission('read'):
-        request.override_renderer = 'h:templates/groups/join.html.jinja2'
-        return {'group': group}
-
-    # Redirect to new group overview page if search page is enabled
-    if request.feature('search_page'):
-        url = request.route_path('activity.group_search', pubid=group.pubid)
-        return HTTPSeeOther(url)
```
seanh: We no longer redirect to the group `/search` page - instead the search page is shown at the same URL (by changing the view config rules to have Pyramid select a different view function than this one)

seanh:
Rebased and ready for review

h/views/activity.py
seanh: Hmm, although actually some of them are `POST`s...

h/views/groups.py
```diff
@@ -112,23 +114,12 @@ def _update_group(self, appstruct):
 @view_config(route_name='group_read',
              request_method='GET',
              renderer='h:templates/groups/share.html.jinja2',
-             effective_principals=security.Authenticated)
+             effective_principals=security.Authenticated,
+             custom_predicates=[custom_predicates.has_read_permission,
```
nickstenning: The `custom_predicates` kwarg was deprecated in Pyramid 1.5. To quote the Pyramid docs:

> Deprecated since version 1.5: This value should be a sequence of references to custom predicate callables. Use custom predicates when no set of predefined predicates do what you need. Custom predicates can be combined with predefined predicates as necessary. Each custom predicate callable should accept two arguments: context and request and should return either True or False after doing arbitrary evaluation of the context and/or the request. The predicates argument to this method and the ability to register third-party view predicates via pyramid.config.Configurator.add_view_predicate() obsoletes this argument, but it is kept around for backwards compatibility.

seanh:
Closing this as I'm re-sending it as a series of smaller pull requests...

PR #4143, opened 2016-11-29
Backfill group access flags and set them on new groups

chdorner:
This PR adds a migration to set the group access flags for all the private groups that are already in the database. It also introduces a new `type_` parameter to the `GroupsService.create` method. Currently we have two group types: `private` and `publisher`. Based on the value of that parameter we set the access flags of the new group.

For more information about the access flags and the different types of groups, see #4133.

### DIFF
--- a/h/migrations/versions/8ae9d103551f_backfill_group_access_flags.py
+++ b/h/migrations/versions/8ae9d103551f_backfill_group_access_flags.py
@@ -1,0 +1,48 @@
+"""
+Backfill group access flags
+
+Revision ID: 8ae9d103551f
+Revises: af88524994ce
+Create Date: 2016-11-29 12:51:31.247598
+"""
+
+from __future__ import unicode_literals
+
+import enum
+
+from alembic import op
+import sqlalchemy as sa
+
+
+revision = '8ae9d103551f'
+down_revision = 'af88524994ce'
+
+
+class JoinableBy(enum.Enum):
+    authority = 'authority'
+joinable_type = sa.Enum(JoinableBy, name='group_joinable_by')
+
+
+class ReadableBy(enum.Enum):
+    members = 'members'
+readable_type = sa.Enum(ReadableBy, name='group_readable_by')
+
+
+class WriteableBy(enum.Enum):
+    members = 'members'
+writeable_type = sa.Enum(WriteableBy, name='group_writeable_by')
+
+group_table = sa.table('group',
+                       sa.Column('joinable_by', joinable_type, nullable=True),
+                       sa.Column('readable_by', readable_type, nullable=True),
+                       sa.Column('writeable_by', writeable_type, nullable=True))
+
+
+def upgrade():
+    op.execute(group_table.update().values(joinable_by=JoinableBy.authority,
+                                           readable_by=ReadableBy.members,
+                                           writeable_by=WriteableBy.members))
+
+
+def downgrade():
+    pass
--- a/h/models/group.py
+++ b/h/models/group.py
@@ -1,5 +1,6 @@
 # -*- coding: utf-8 -*-
 
+import enum
 import sqlalchemy as sa
 from pyramid import security
 from sqlalchemy.orm import exc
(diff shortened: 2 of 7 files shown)


h/migrations/versions/8ae9d103551f_backfill_group_access_flags.py
```diff
@@ -0,0 +1,48 @@
...
+import sqlalchemy as sa
+
+
+revision = '8ae9d103551f'
+down_revision = 'af88524994ce'
+
+
+class JoinableBy(enum.Enum):
+    authority = 'authority'
+joinable_type = sa.Enum(JoinableBy, name='group_joinable_by')
```
seanh: Ok, so it's nice and clear that this migration script is only setting groups to joinable by 'authority', since that's the only value in its version of the enum. And it needs to have its own copy of the enum because we don't know what the real `h.models.group.JoinableBy` will look like (or if it'll even exist) in the future versions of the code that may be deployed when this migration script is run

h/models/group.py
```diff
@@ -49,6 +65,21 @@ class Group(Base, mixins.Timestamps):
...
+
+    #: Which type of user is allowed to read annotations in this group,
+    #: possible values are: authority, members, world
+    readable_by = sa.Column(sa.Enum(ReadableBy, name='group_readable_by'),
+                            nullable=True)
+
+    #: Which type of user is allowed to write to this group, possible values
+    #: are: authority, members
+    writeable_by = sa.Column(sa.Enum(WriteableBy, name='group_writeable_by'),
+                             nullable=True)
```
seanh: Do you think it might be better for data integrity for `joinable_by`, `readable_by` and `writeable_by` to be Python `@property`s that aren't stored in the db but are just implemented in Python and return different values depending on a `type` field that is stored in the db? And `type` would be an enum that, for now, has I guess only one possible value `"private"` or maybe two `"private"` and `"publisher"`? And of course if we want to make use of Pyramid's ACL system `Group` can have an `__acl__()` method that works based on these and any other properties of the group (like `memex.models.Annotation.__acl__()`).

I'm thinking some of the advantages might be that, like with the current `Annotation` model, it won't be possible at the database level for an invalid value to be stored (for example a group that is joinable by no one but only readable by its members etc). We can guarantee that, based on the more restricted properties that are actually stored in the db, the `__acl__()` method will only generate the valid possible ACLs and not any contradictory ones. And it'll be easier to look in the db or model and find all the "private" groups or all the "publisher" groups by just looking at the `type` column. Adding a new type of group would just be a case of adding a new value to the type column's enum and extending the `__acl__()` method to support it.

seanh:
Oops, I see that a PR adding the new columns to the db has already been merged, I didn't realise this one needed to be re-based.

seanh:
Ok, in that case I wanna test this a little more tomorrow morning and I still wonder about the storage question I asked but since that's actually a question about the previous PR which has already been merged, it shouldn't hold up merging this one...

nickstenning:
On the question of whether we should store a single token in the database that contains a group type name, rather than the `{readable,writeable,joinable}_by` columns...

I think this is a side-effect of how we've been talking about groups internally, as "private"/"publisher"/"public" groups. The problem is that these are opaque tokens which don't explain much to anyone not already familiar with past conversations about how they work or what they're for. For example, which of the following is more immediately understandable to a new developer:

```SQL
SELECT * FROM "group" WHERE readable_by = 'world' AND writeable_by = 'members';
```

or

```SQL
SELECT * FROM "group" WHERE type = 'public';
```

Moreover, for "public groups" we've discussed both a) groups that only members can write to but anyone can read from, and b) groups that anyone can write to and anyone can read from. Do we need to devise two distinct names for these types of groups?

In my way of thinking, we have just one feature: "groups". We're talking about evolving its capabilities over time. I think it'll be much easier to understand if the code keeps the relevant properties of a group independent of one another rather than bundling them all up under an arbitrarily chosen name.

(One last consideration: I highly doubt it's best to ask users in the UI whether they want to create a "public" or "private" group, because they may have preconceptions about these terms that don't map to our chosen meanings. Instead, they'll create a group, and then configure the groups privacy and membership settings independently of one another.)

seanh:
Ok, tested and works nicely

seanh:
Just trying to figure out why tests are failing locally for me...

seanh:
Alright, works now

seanh:
@nickstenning That makes good sense to me. I do think it's a -1 that, at least at the db level, it seems possible for invalid / nonsensical combinations of values to be stored. But other than that I think it sounds right, great points.

(We could probably prevent that at the db level using check constraints but I wouldn't recommend it, just do it in Python)

PR #4144, opened 2016-11-29
Show the only me icon for annotations which are private.

sheetaluk:
https://github.com/hypothesis/product-backlog/issues/47


dwhly:
Assuming this means we're going to pull the group name down here onto this line too, as this is the other key way that the sidebar deviates from the stream.

sheetaluk:
@dwhly yes.

robertknight:
LGTM 👍

PR #4145, opened 2016-11-29
Report errors for invalid messages from the client

nickstenning:
This PR adds error handling to the websocket API. Messages from the client which are invalid will be replied to, but ONLY if those messages contain a message ID.

This change should be entirely backwards compatible, as the error message replies won't be sent to our existing client, which doesn't set message IDs.

### DIFF
--- a/h/streamer/websocket.py
+++ b/h/streamer/websocket.py
@@ -18,8 +18,28 @@ log = logging.getLogger(__name__)
 # below.
 MESSAGE_HANDLERS = {}
 
+
 # An incoming message from a WebSocket client.
-Message = namedtuple('Message', ['socket', 'payload'])
+class Message(namedtuple('Message', [
+    'socket',
+    'payload',
+])):
+    def reply(self, payload, ok=True):
+        """
+        Send a response to this message.
+
+        Sends a reply message back to the client, with the passed `payload`
+        and reporting status `ok`.
+        """
+        reply_to = self.payload.get('id')
+        # Short-circuit if message is missing an ID or has a non-numeric ID.
+        if not isinstance(reply_to, (int, float)):
+            return
+        data = {}
+        data.update(payload)  # avoid modifying the payload in-place
+        data['ok'] = ok
+        data['reply_to'] = reply_to
+        self.socket.send_json(data)
 
 
 class WebSocket(_WebSocket):
--- a/h/streamer/websocket.py
+++ b/h/streamer/websocket.py
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 
 from collections import namedtuple
+import copy
 import json
 import logging
 import weakref
@@ -35,8 +36,7 @@ class Message(namedtuple('Message', [
         # Short-circuit if message is missing an ID or has a non-numeric ID.
         if not isinstance(reply_to, (int, float)):
             return
-        data = {}
-        data.update(payload)  # avoid modifying the payload in-place
+        data = copy.deepcopy(payload)
         data['ok'] = ok
         data['reply_to'] = reply_to
         self.socket.send_json(data)
@@ -127,7 +127,10 @@ def handle_message(message, session=None):
 def handle_client_id_message(message, session=None):
     """A client telling us its client ID."""
     if 'value' not in message.payload:
-        # FIXME: send an error message to the client
+        message.reply({'type': 'error',
+                       'error': {'type': 'invalid_data',
+                                 'description': '"value" is missing'}},
+                      ok=False)
         return
     message.socket.client_id = message.payload['value']
 MESSAGE_HANDLERS['client_id'] = handle_client_id_message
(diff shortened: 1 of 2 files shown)


h/streamer/websocket.py
```diff
@@ -18,8 +18,28 @@
...
+        """
+        Send a response to this message.
+
+        Sends a reply message back to the client, with the passed `payload`
+        and reporting status `ok`.
+        """
+        reply_to = self.payload.get('id')
+        # Short-circuit if message is missing an ID or has a non-numeric ID.
+        if not isinstance(reply_to, (int, float)):
+            return
```
chdorner: It probably outweighs the drawbacks because the goal is to change the client and to get rid of the old websocket API. But this code would be nicer if it would raise an error and the caller has to make sure to only reply when there's an id in the request. The way the code is written now will probably introduce some situations where it's hard to debug why the client isn't getting a reply. I would imagine that once we remove the old API we would immediately error / close when we get a request without an id, without even trying to do anything on the server.

h/streamer/websocket.py
```diff
@@ -18,8 +18,28 @@
...
+    def reply(self, payload, ok=True):
+        """
+        Send a response to this message.
+
+        Sends a reply message back to the client, with the passed `payload`
+        and reporting status `ok`.
+        """
+        reply_to = self.payload.get('id')
+        # Short-circuit if message is missing an ID or has a non-numeric ID.
+        if not isinstance(reply_to, (int, float)):
```
chdorner: I'd suspect that some implementations could use a UUID for the message ids, why do you want to limit the type to only integers and floats?

h/streamer/websocket.py
```diff
@@ -18,8 +18,28 @@
...
+            return
+        data = {}
+        data.update(payload)  # avoid modifying the payload in-place
```
chdorner: Yes, we only update toplevel items, but to avoid weird issues popping up in the future I would use `copy.deepcopy` instead. 

The way it is written now allows sub-items to be changed in-place:

```
payload = {'foo': {'bar': 'baz'}}
data = {}
data.update(payload)
data['foo']['bar'] = 'qux'
print(payload) # -> {'foo': {'bar': 'qux'}}
```

h/streamer/websocket.py
```diff
@@ -18,8 +18,28 @@
...
 MESSAGE_HANDLERS = {}
 
+
 # An incoming message from a WebSocket client.
-Message = namedtuple('Message', ['socket', 'payload'])
+class Message(namedtuple('Message', [
+    'socket',
+    'payload',
+])):
+    def reply(self, payload, ok=True):
```
chdorner: This might be personal preference, but the parameter `ok` isn't as obvious to me than using `error` (with default `False`).

tests/h/streamer/websocket_test.py
```diff
@@ -13,6 +14,47 @@
...
+    @pytest.mark.parametrize('ok', [True, False])
+    def test_reply_adds_ok(self, ok, socket):
+        """Adds an appropriate `ok` field to the sent message."""
+        message = websocket.Message(socket=socket, payload={'id': 123})
+
+        message.reply({'foo': 'bar'}, ok=ok)
+
+        socket.send_json.assert_called_once_with({'ok': ok,
+                                                  'reply_to': 123,
+                                                  'foo': 'bar'})
```
chdorner: It might be good to add a test for each of these reply fields that it overrides `ok` and `reply_to` in the payload.

h/streamer/websocket.py
nickstenning: I'm not sure about this. To me having `{"error": false}` in a successful message back to the client (hopefully the 90% case) is more likely to cause confusion than `{"ok": true}`, and I think the kwarg here should correspond directly to what we send to the client.

(Not to mention the fact that the `error` key is already in use and contains the actual error data.)

h/streamer/websocket.py
nickstenning: If we allow strings then we're not bounding the size of the responses we send unless we add additional validation to ensure IDs aren't humongous. Restricting to numbers seems pretty reasonable -- I can't think of a scenario in which you'd need anything other than a counter or high-resolution timer here.

h/streamer/websocket.py
nickstenning: Why would that be nicer? That means that the responsibility for extracting and validating the ID has to be repeated in each and every message handler, no? That seems to me to be *more* likely to introduce bugs, and thus "situations where it's hard to debug why the client isn't getting a reply"...

As for whether we'd immediately close if we got a message without an ID... I don't think so. It's potentially a nice way for the client to signal it doesn't care about the response.

For now the rule is simple: if you send a numeric ID, you get a response. If you don't, you don't.

h/streamer/websocket.py
nickstenning: I'm happy to change this, but I think using deepcopy here is probably overkill.

tests/h/streamer/websocket_test.py
nickstenning: Yep, agreed. Will do.

nickstenning: Added.

nickstenning:
Updated, pending your final verdict on the deepcopy question. 👮

h/streamer/websocket.py
chdorner: It'll potentially avoid weird situations in the future when we start (for whatever reason) to override sub-items and not just top level ones.

nickstenning: Ok. Personally I'm still not sure the processing cost is worth it in this situation but I've updated the PR.

PR #4146, opened 2016-11-29
Update icons for 'Tag' and 'URL' headings in search result buckets

### LINKED ISSUE #4058: Document URLs in document buckets aren't linked

robertknight:
Update search result bucket sidebar with tag and link icons that match the design at
https://github.com/hypothesis/h/issues/4058#issuecomment-261954000

tag-link-icons

Fixes #4058


robertknight:
If you look at the URL icon in Firefox, the top of the chain link gets very slightly cut off, though I think it is acceptable. The GitHub blog post that describes the approach we use for rendering icons does [mention a related issue](https://hyp.is/zyRYtLbdEeaoz7Osq-hP_A/github.com/blog/2112-delivering-octicons-with-svg).

PR #4147, opened 2016-11-29
Add the search results total annotation count

sean-roberts:
fixes [#55](https://github.com/hypothesis/product-backlog/issues/55)

Adds the annotation count to the top left of the search results page if there is some sort of search criteria (at least one lozenge). 

Note, on mobile there is room to rework our use of vertical whitespace on the user profile page. Because we put the user name in bold and the big button for more info.. _then_ we add this annotations bit. It is fine for now, though.

The renaming was needed to keep some logical grouping with the many container level elements in the UI.

### DIFF
--- a/h/templates/activity/search.html.jinja2
+++ b/h/templates/activity/search.html.jinja2
@@ -348,18 +348,26 @@
     {% endif %}
 
     {% if timeframes %}
-      <ol class="search-result-list {%- if more_info %} search-result-hide-on-small-screens{% endif %}">
-        {% for timeframe in timeframes %}
+
+      <div class="search-results">
+
+        {% if q or user or group %}
+        <div class="search-results__total"><b>{% trans %}Annotations{% endtrans %}</b> {{total}}</div>
+        {% endif %}
+
+        <ol class="search-results__list {%- if more_info %} search-result-hide-on-small-screens{% endif %}">
+          {% for timeframe in timeframes %}
           <li class="search-result__timeframe">
             {{ timeframe.label }}
           </li>
           <li>
             {% for bucket in timeframe.document_buckets.values() %}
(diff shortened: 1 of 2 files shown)


h/templates/activity/search.html.jinja2
```diff
@@ -348,18 +348,26 @@
     {% endif %}
 
     {% if timeframes %}
-      <ol class="search-result-list {%- if more_info %} search-result-hide-on-small-screens{% endif %}">
-        {% for timeframe in timeframes %}
+
+      <div class="search-results">
+
+        {% if q or user or group %}
+        <div class="search-results__total"><b>{% trans %}Annotations{% endtrans %}</b> {{total}}</div>
```
robertknight: Is the `<b>` here purely for styling? If so, I'd suggest moving that to `font-weight: bold` in `search-results__total`.

h/templates/activity/search.html.jinja2
```diff
@@ -348,18 +348,26 @@
     {% endif %}
 
     {% if timeframes %}
-      <ol class="search-result-list {%- if more_info %} search-result-hide-on-small-screens{% endif %}">
-        {% for timeframe in timeframes %}
+
+      <div class="search-results">
```
robertknight: Hmm, I wonder if there is a better name for this class to distinguish it from `.search-result-container`. If that is hard to come by, it would be helpful to add some comments in `search.scss` to clarify the difference between the two.

sean-roberts: I will add a comment in search.scss. I spent at least 20 mins solely thinking through how to rename it without refactoring all of the naming. Given how people interact with CSS updating and debugging, the current naming and a comment should suffice.

robertknight:
LGTM

PR #4148, opened 2016-11-30
Add documentation for the "real time" API

nickstenning:
This PR adds a couple of message types to the websocket protocol (a PING/PONG and a simple mechanism of interrogating auth status) and adds some initial documentation for the protocol.

I have deliberately not documented any of the actually useful message types at the moment because I think we need to audit them and decide whether they're going to be part of the public API or whether they need reimplementing.

### DIFF
--- a/docs/realtime.rst
+++ b/docs/realtime.rst
@@ -1,0 +1,117 @@
+The Hypothesis Real Time API
+============================
+
+.. warning::
+
+   This document describes an API that is in the early stages of being
+   documented and refined for public use. Details may change, and your systems
+   may break in the future.
+
+In addition to the `HTTP API <http://h.readthedocs.io/en/latest/api/>`_
+Hypothesis has a WebSocket-based API that allows developers to receive near
+real-time notifications of annotation events.
+
+Overview
+--------
+
+.. note::
+
+   At the moment, use of the Real Time API is limited to server-side clients.
+   JavaScript applications running in the browser **will not be able to use this
+   API** due to security restrictions. We hope to lift this restriction in the
+   future.
+
+To use the Real Time API, you should open a WebSocket connection to the
+following endpoint::
+
+    wss://hypothes.is/ws
+
+Communication with this endpoint consists of JSON-encoded messages sent from
+client to server and vice versa.
+
+Server messages
+---------------
+
+Each messages from the server will be either an :term:`event` or a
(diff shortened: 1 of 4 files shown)


docs/realtime.rst
```diff
@@ -0,0 +1,117 @@
...
+
+   At the moment, use of the Real Time API is limited to server-side clients.
+   JavaScript applications running in the browser **will not be able to use this
+   API** due to security restrictions. We hope to lift this restriction in the
+   future.
+
+To use the Real Time API, you should open a WebSocket connection to the
+following endpoint::
+
+    wss://hypothes.is/ws
```
nickstenning: One thing worth thinking about is that this is actually a pretty implausible way of managing the websocket in the long term, because it requires all these long-lived connections to go through this one DNS address.

Slack works around this problem by requiring users to make an HTTP API call which returns a "capability" endpoint URL. This gives a couple of nice properties:

- WebSockets don't need any authentication at all -- any auth is provided by the capability URL.
- Sharding WS connections across domain names is easy.
- The API call can be responsible for any necessary rate-limiting or black/whitelisting, thus simplifying the WebSocket server.

robertknight: Ah, interesting. Thanks for mentioning this.

docs/realtime.rst
```diff
@@ -0,0 +1,117 @@
...
+
+   event
+      An event is sent to clients as a result of an action taken elsewhere in
+      the system. For example: if an annotation is made which matches one of the
+      client's subscriptions, the client will receive an event message. All
+      event messages have a ``type`` field.
+
+   reply
+      A reply is sent in response to a message sent by the client. All replies
+      have an ``ok`` field which indicates whether the server successfully
```
robertknight: Can you add documentation for the format of errors here as well? It looks like we have a standard format for including errors in responses.

docs/realtime.rst
```diff
@@ -0,0 +1,117 @@
+The Hypothesis Real Time API
+============================
+
+.. warning::
+
+   This document describes an API that is in the early stages of being
+   documented and refined for public use. Details may change, and your systems
+   may break in the future.
+
+In addition to the `HTTP API <http://h.readthedocs.io/en/latest/api/>`_
```
robertknight: Perhaps make this an HTTPS link?

PR #4149, opened 2016-11-30
Show direct links on annotation cards.

sheetaluk:


tests/h/activity/query_test.py
```diff
@@ -379,6 +379,21 @@ def test_it_returns_each_annotations_group(self,
...
+        result = execute(pyramid_request, MultiDict(), self.PAGE_SIZE)
+
+        presented_annotations = []
+        for timeframe in result.timeframes:
+            for bucket in timeframe.document_buckets.values():
+                presented_annotations.extend(bucket.presented_annotations)
+
+        for annotation in annotations:
+            for presented_annotation in presented_annotations:
+                assert presented_annotation['incontext_link'] == links.incontext_link.return_value
```
seanh: It's an awkward test to write but here it might be better to test that 1. `incontext_link()` is called once for each annotation, and 2. Each `presented_annotation['incontext_link']` contains the value that `incontext_link()` returned for that annotation. One way to do that is to replace `links.incontext_link()` with a fake function that, when called with a given annotation, returns a value that's predictable based on the annotation that was given.

This can be done using [mock's side_effect property](http://www.voidspace.org.uk/python/mock/mock.html#mock.Mock.side_effect) which allows you to nominate a function to be called when a mock is called, and the mock will return whatever that functions returns.

Something like this:

```diff
diff --git a/tests/h/activity/query_test.py b/tests/h/activity/query_test.py
index 0e7a085..2fbb7b6 100644
--- a/tests/h/activity/query_test.py
+++ b/tests/h/activity/query_test.py
@@ -383,6 +383,13 @@ class TestExecute(object):
                                                   annotations,
                                                   links,
                                                   pyramid_request):
+        def incontext_link(request, annotation):
+            assert request == pyramid_request, (
+                "It should always pass the request to incontext_link")
+            # Return a predictable per-annotation value for the incontext link.
+            return 'incontext_link_for_annotation_{id}'.format(id=id)
+        links.incontext_link.side_effect = incontext_link
+
         result = execute(pyramid_request, MultiDict(), self.PAGE_SIZE)
 
         presented_annotations = []
@@ -392,7 +399,8 @@ class TestExecute(object):
 
         for annotation in annotations:
             for presented_annotation in presented_annotations:
-                assert presented_annotation['incontext_link'] == links.incontext_link.return_value
+                assert presented_annotation['incontext_link'] == (
+                    'incontext_link_for_annotation_{id}'.format(id=id))
 
     def test_it_returns_the_total_number_of_matching_annotations(
             self, pyramid_request):
```

h/activity/query.py
```diff
@@ -130,7 +131,8 @@ def execute(request, query, page_size):
             for annotation in bucket.annotations:
                 bucket.presented_annotations.append({
                     'annotation': presenters.AnnotationHTMLPresenter(annotation),
-                    'group': groups.get(annotation.groupid)
+                    'group': groups.get(annotation.groupid),
+                    'incontext_link':links.incontext_link(request, annotation)
```
seanh: PEP 8, there should be a space here: `'incontext_link': links.incontext_link(request, annotation)`

h/templates/activity/search.html.jinja2
```diff
@@ -122,7 +122,7 @@
     <div class="search-result-bucket__annotation-cards-container" data-ref="content">
       <ol class="search-result-bucket__annotation-cards">
         {% for result in bucket.presented_annotations %}
-          {{ annotation_card(result.annotation, request, result.group) }}
+          {{ annotation_card(result.annotation, request, result.incontext_link, result.group) }}
```
seanh: If you do want to reduce the number of arguments here you could just pass `result` instead of the three `result.*`'s

h/templates/includes/annotation_card.html.jinja2
```diff
@@ -29,5 +29,10 @@
   <section class="annotation-card__text">
     {{ annotation.text_rendered|safe }}
   </section>
+  <footer class="annotation-card__footer">
+    <a class="annotation-card__incontext-link" href="{{ incontext_link }}" target="_blank" title="{{ incontext_link }}">
```
seanh: I wonder if the `title` should be `"{% trans %}Visit annotation in context{% endtrans %}"`?

seanh: Did you mean to put the `"annotation-card__incontext-link"` CSS class on both the `<a>` and the `<svg>`?

tests/h/activity/query_test.py
```diff
@@ -379,6 +379,21 @@ def test_it_returns_each_annotations_group(self,
             else:
                 assert False
 
+    def test_it_returns_each_annotations_incontext_link(self,
+                                                  annotations,
+                                                  links,
+                                                  pyramid_request):
```
seanh: PEP 8: These three params need to be indented to line up with the `self`

h/templates/includes/annotation_card.html.jinja2
seanh: `incontext_link` is sometimes `None` (the `links.incontext_link()` function returns `None` for some annotations that can't be linked to), I think you might need an `{% if %}` here to not render the link in that case?

h/templates/includes/annotation_card.html.jinja2
```diff
@@ -39,5 +39,12 @@
   <section class="annotation-card__text">
     {{ annotation.text_rendered|safe }}
   </section>
+  <footer class="annotation-card__footer">
+    {% if incontext_link %}
+      <a href="{{ incontext_link }}" target="_blank" title="Visit link in context">
```
seanh: `"Visit link in context"` should be `"{% trans %}Visit link in context{% endtrans %}"` so that it gets internationalized. Some docs about this here http://docs.ckan.org/en/latest/contributing/string-i18n.html

PR #4150, opened 2016-11-30
Use traversal for group search page

seanh:
Second in a sequence of PRs to refactor the groups and activity views so that we can easily change the `/groups/<pubid>/search` URL to share `/groups/<pubid>/<slug>` with the other group pages.

This means that:

* `/groups/<pubid_that_does_not_exist>/search` is now a 404, same as `/users/<username_that_does_not_exist>`, instead of an empty search page

* Searching for `group:pubid_that_does_not_exist` will no longer redirect you to `/groups/<pubid_that_does_not_exist>/search`. Again, same as for users

### DIFF
--- a/h/activity/query.py
+++ b/h/activity/query.py
@@ -80,9 +80,11 @@ def check_url(request, query, unparse=parser.unparse):
 
     if _single_entry(query, 'group'):
         pubid = query.pop('group')
-        redirect = request.route_path('activity.group_search',
-                                      pubid=pubid,
-                                      _query={'q': unparse(query)})
+        group = request.db.query(Group).filter_by(pubid=pubid).one_or_none()
+        if group:
+            redirect = request.route_path('activity.group_search',
+                                          pubid=group.pubid,
+                                          _query={'q': unparse(query)})
 
     elif _single_entry(query, 'user'):
         username = query.pop('user')
--- a/h/views/activity.py
+++ b/h/views/activity.py
@@ -95,22 +95,17 @@ class SearchController(object):
 class GroupSearchController(SearchController):
     """View callables unique to the "activity.group_search" route."""
 
+    def __init__(self, group, request):
+        super(GroupSearchController, self).__init__(request)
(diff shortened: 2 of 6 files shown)


h/views/activity.py
```diff
@@ -95,22 +95,17 @@ def user_link(userid):
...
-        pubid = self.request.matchdict['pubid']
-        result['opts'] = {'search_groupname': pubid}
-
-        try:
-            group = (self.request.db.query(models.Group)
-                     .filter_by(pubid=pubid).one())
-        except exc.NoResultFound:
-            return result
-
-        result['opts']['search_groupname'] = group.name
```
seanh: No longer need to fetch the `Group` from the db itself, or worry about when the group doesn't exist

h/views/activity.py
```diff
@@ -165,16 +160,9 @@ def leave(self):
         browser to the search page.
 
         """
-        pubid = self.request.POST['group_leave']
-
-        try:
-            group = (self.request.db.query(models.Group)
-                     .filter_by(pubid=pubid).one())
-        except exc.NoResultFound:
-            raise httpexceptions.HTTPNotFound()
```
seanh: No longer need to fetch the `Group` from the db itself, or worry about when the group doesn't exist

h/activity/query.py
```diff
@@ -80,9 +81,15 @@ def check_url(request, query, unparse=parser.unparse):
 
     if _single_entry(query, 'group'):
         pubid = query.pop('group')
-        redirect = request.route_path('activity.group_search',
-                                      pubid=pubid,
-                                      _query={'q': unparse(query)})
+
+        try:
+            group = request.db.query(Group).filter_by(pubid=pubid).one()
+        except exc.NoResultFound:
```
nickstenning: It's a small thing, but you'd save yourself an import here and a couple of lines if you used `.one_or_none()`.

seanh: Done

PR #4151, opened 2016-11-30
Only display back links in specific cases

nickstenning:
Previously, this code would display a generic "Back" link if any HTTP "Referer" was set, resulting in a back link being displayed when clicking around between different panes of the account area.

This commit restricts the display of the "back" link to only the two cases identified in the original card[1].

[1]: https://github.com/hypothesis/product-backlog/issues/42


nickstenning:
One for you, @robertknight.

robertknight:
I was wondering about the merits of the generic case when I implemented it. Happy to remove it for the moment.

PR #4152, opened 2016-12-01
Factor out GroupUserSearchController

seanh:
Replace it with private helper methods instead. This is a refactoring
with no functional change. It should make it easier to make future
changes to these view classes by allowing view defaults to be specified
more easily.

`GroupUserSearchController` was a class containing view methods that each
had two `@view_config`s on: one for the `"activity.group_search"` route
and one for `"activity.user_search"`. This avoided duplicating the Python
code in the view methods, which is shared by the group and user search
routes.

The problem with this is that it ends up duplicating view config
declarations - `@view_defaults` can't be used effectively on
`GroupUserSearchController` because it can only contain defaults that
apply to both the group search and the user search routes for every view
method in the class. Additionally the view defaults needed for the group
search and user search routes are already on `@view_defaults`
declarations on `GroupSearchController` and `UserSearchController`
respectively.

Instead move all of `GroupUserSearchController`'s methods to private
helper functions: `_more_info(request)` etc. Then for each of the
private helper functions add a corresponding method to _both_
`GroupSearchController` _and_ `UserSearchController`:
`GroupSearchController.more_info(self)` etc. These duplicated methods
are just one-liners that call the helper methods (avoiding duplication
of Python code) and that have `@view_config`s on them. These
`@view_config`s can rely on the `@view_defaults` of
`GroupSearchController` and `UserSearchController`, avoiding duplication
of view config declarations.

We can now add more view defaults to `UserSearchController` or
`GroupSearchController` separately and have them apply to all the view
methods they contain.

### DIFF
--- a/tests/h/views/activity_test.py
+++ b/tests/h/views/activity_test.py
@@ -137,6 +137,8 @@ class TestSearchController(object):
 @pytest.mark.usefixtures('groups_service', 'routes', 'search')
 class TestGroupSearchController(object):
 
+    """Tests unique to GroupSearchController."""
+
     def test_init_returns_404_when_feature_turned_off(self,
                                                       group,
                                                       pyramid_request):
@@ -365,6 +367,42 @@ class TestGroupSearchController(object):
         assert isinstance(result, httpexceptions.HTTPSeeOther)
         assert result.location == 'http://example.com/search?q=foo+bar+gar'
 
+    def test_more_info_redirects_to_group_search(self,
+                                                 controller,
+                                                 group,
+                                                 pyramid_request):
+        """It should redirect and preserve the search query param."""
+        pyramid_request.matched_route = mock.Mock()
+        pyramid_request.matched_route.name = 'activity.group_search'
+        pyramid_request.POST = {'q': 'foo bar', 'more_info': ''}
+
+        result = controller.more_info()
+
+        assert isinstance(result, httpexceptions.HTTPSeeOther)
+        assert result.location.startswith(
+            'http://example.com/groups/{pubid}/search?'.format(
+                pubid=group.pubid))
+        # The order of the params vary (because they're in an unordered dict)
+        # but they should both be there.
+        assert 'more_info=' in result.location
+        assert 'q=foo+bar' in result.location
+
+    def test_back_redirects_to_group_search(self,
+                                            controller,
+                                            group,
+                                            pyramid_request):
+        """It should redirect and preserve the search query param."""
+        pyramid_request.matched_route = mock.Mock()
+        pyramid_request.matched_route.name = 'activity.group_search'
+        pyramid_request.POST = {'q': 'foo bar', 'back': ''}
+
+        result = controller.back()
+
+        assert isinstance(result, httpexceptions.HTTPSeeOther)
+        assert result.location == (
+            'http://example.com/groups/{pubid}/search?q=foo+bar'.format(
+                pubid=group.pubid))
+
     @pytest.mark.usefixtures('toggle_user_facet_request')
     def test_toggle_user_facet_returns_a_redirect(self, controller):
         result = controller.toggle_user_facet()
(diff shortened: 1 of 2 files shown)


tests/h/views/activity_test.py
```diff
@@ -753,6 +670,116 @@ def user(self, factories):
...
+
+    """Tests common to both GroupSearchController and UserSearchController."""
+
+    @pytest.mark.usefixtures('delete_lozenge_request')
+    def test_delete_lozenge_returns_a_redirect(self, controller):
+        result = controller.delete_lozenge()
+
+        assert isinstance(result, httpexceptions.HTTPSeeOther)
+
+        # This tests that the location redirect to is correct and also that
```
robertknight: Typo: "redirect" => "redirected"

robertknight:
LGTM. It appears that there is a way to share view configuration between multiple view classes, using the `lift` and `onlyliftedfrom` decorators:

```diff
diff --git a/h/views/activity.py b/h/views/activity.py
index fa0cd21..28db55b 100644
--- a/h/views/activity.py
+++ b/h/views/activity.py
@@ -12,6 +12,7 @@ from pyramid.view import view_config
 from pyramid.view import view_defaults
 from sqlalchemy.orm import exc
 from memex.search import parser
+from venusian import lift, onlyliftedfrom
 
 from h import models
 from h.activity import query
@@ -90,9 +91,29 @@ class SearchController(object):
         }
 
 
+@onlyliftedfrom() # Do not register these methods as views directly
+class SidebarMixin(object):
+    @view_config(request_method='POST', request_param='more_info')
+    def more_info(self):
+        return _more_info(self.request)
+
+    @view_config(request_method='POST', request_param='back')
+    def back(self):
+        return _back(self.request)
+
+    @view_config(request_param='delete_lozenge')
+    def delete_lozenge(self):
+        return _delete_lozenge(self.request)
+
+    @view_config(request_method='POST', request_param='toggle_tag_facet')
+    def toggle_tag_facet(self):
+        return _toggle_tag_facet(self.request)
+
+
+@lift() # Import views from mixins
 @view_defaults(route_name='activity.group_search',
                renderer='h:templates/activity/search.html.jinja2')
-class GroupSearchController(SearchController):
+class GroupSearchController(SidebarMixin, SearchController):
     """View callables unique to the "activity.group_search" route."""
 
     def __init__(self, group, request):
@@ -214,26 +235,11 @@ class GroupSearchController(SearchController):
 
         return httpexceptions.HTTPSeeOther(location=location)
 
-    @view_config(request_method='POST', request_param='more_info')
-    def more_info(self):
-        return _more_info(self.request)
-
-    @view_config(request_method='POST', request_param='back')
-    def back(self):
-        return _back(self.request)
-
-    @view_config(request_param='delete_lozenge')
-    def delete_lozenge(self):
-        return _delete_lozenge(self.request)
-
-    @view_config(request_method='POST', request_param='toggle_tag_facet')
-    def toggle_tag_facet(self):
-        return _toggle_tag_facet(self.request)
-
 
+@lift() # Import views from mixins
 @view_defaults(route_name='activity.user_search',
                renderer='h:templates/activity/search.html.jinja2')
-class UserSearchController(SearchController):
+class UserSearchController(SidebarMixin, SearchController):
     """View callables unique to the "activity.user_search" route."""
 
     def __init__(self, user, request):
@@ -278,22 +284,6 @@ class UserSearchController(SearchController):
 
         return result
 
-    @view_config(request_method='POST', request_param='more_info')
-    def more_info(self):
-        return _more_info(self.request)
-
-    @view_config(request_method='POST', request_param='back')
-    def back(self):
-        return _back(self.request)
-
-    @view_config(request_param='delete_lozenge')
-    def delete_lozenge(self):
-        return _delete_lozenge(self.request)
-
-    @view_config(request_method='POST', request_param='toggle_tag_facet')
-    def toggle_tag_facet(self):
-        return _toggle_tag_facet(self.request)
-
 
 def _parsed_query(request):
     """
```

I don't think it is particularly worthwhile at the moment, but possible useful for future reference.

seanh:
That's pretty interesting, Rob. Did you test that the different view _defaults_ on `GroupSearchController` and `UserSearchController` get applied, when one of the inherited views from `SidebarMixin` is called?

I guess the drawback of this is that you can't have the view _config_ (as opposed to view defaults) of a given view, say `more_info()`, be different in `UserSearchController` than `GroupSearchController`. Whereas if `UserSearchController` and `GroupSearchController` both have their own `more_info()` method (that just calls a helper function) then, at the expense of duplicating the `@view_config` arguments that are the same, you can also have `@view_config` arguments that differ between the two classes.

The helper function approach could of course use a base class and have `UserSearchController` and `GroupSearchController` inherit from it as well. The base class would define `more_info()` etc but without any `@view_config`s on them. The sub classes would add their own `more_info()` with an `@view_config` and user `super()`. It's not necessary to use a base class but it does group related methods together nicely.

The lift approach seems like it reduces duplication and boilerplate the most, as long as you always want the view configs (if not the defaults) of the inherited methods to be the same, which in this case we actually do.

tests/h/views/activity_test.py
seanh: Fixed this on master

PR #4154, opened 2016-12-01
Factor group join into separate view

seanh:
Factor the group join page (the view for the 'group_read' route when the
user is logged-in but isn't a member of the group) into a separate view
callable.

Previously h.views.groups.read() was called for the 'group_read' route
when the user was logged-in, whether they're a member of the group or
not.

In the future the classic group page that this read() view provides is
going to be used only when the 'search_page' feature flag is off, and a
completely separate view (h.views.activity.GroupSearchController) is
going to be used when the flag is on. We'll add a view config predicate
to read() so that it's not called if the flag is on.

But we still want the join group page to be called even if the feature
flag is on, so we can't have this read() function (which is not going to
be called when the flag is on anymore) also being responsible for the
join page.

Additionally there was a separate view function, join(), which was
called when the user is logged-in but isn't a member of the group and
POSTs to the group_read route. It's a little odd that GET requests in
the "join group" scenario (group_read route, logged-in, not a member)
are handled by read() (which also handles the read group scenario)
whereas POST requests in the join group scenario are handled by a
separate join() function.

So refactor things so that:

* read() is only for the group read page that you see when you're a
  member of the group

* The new JoinController is for the join page that you see when you
  aren't a member, and has separate methods for GET and POST to this page

Note that there's still one more separate view function for this route as
well, which this commit doesn't touch: read_unauthenticated() (for when you're
not logged in).

### DIFF
--- a/h/views/groups.py
+++ b/h/views/groups.py
@@ -2,6 +2,7 @@
 
 import deform
 from pyramid import security
+from pyramid.config import not_
 from pyramid.httpexceptions import (HTTPMovedPermanently, HTTPNoContent,
                                     HTTPSeeOther)
 from pyramid.view import view_config, view_defaults
@@ -112,17 +113,11 @@ class GroupEditController(object):
 @view_config(route_name='group_read',
              request_method='GET',
              renderer='h:templates/groups/share.html.jinja2',
-             effective_principals=security.Authenticated)
+             effective_principals=security.Authenticated,
+             has_permission='read')
 def read(group, request):
     """Group view for logged-in users."""
     _check_slug(group, request)
-
-    # If the current user is not a member of the group, they will not have the
-    # 'read' permission on the group. In this case, we show them the join
-    # page.
-    if not request.has_permission('read'):
-        request.override_renderer = 'h:templates/groups/join.html.jinja2'
-        return {'group': group}
 
     # Redirect to new group overview page if search page is enabled
     if request.feature('search_page'):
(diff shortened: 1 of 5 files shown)


h/views/groups.py
```diff
@@ -112,18 +113,12 @@ def _update_group(self, appstruct):
 @view_config(route_name='group_read',
              request_method='GET',
              renderer='h:templates/groups/share.html.jinja2',
-             effective_principals=security.Authenticated)
+             effective_principals=security.Authenticated,
+             has_permission='read')
```
seanh: We make it so that this is only called if you have read permission, so that JoinGroupController below will be called if you don't

h/views/groups.py
```diff
@@ -112,18 +113,12 @@ def _update_group(self, appstruct):
...
 def read(group, request):
     """Group view for logged-in users."""
     _check_slug(group, request)
 
-    # If the current user is not a member of the group, they will not have the
-    # 'read' permission on the group. In this case, we show them the join
-    # page.
-    if not request.has_permission('read'):
-        request.override_renderer = 'h:templates/groups/join.html.jinja2'
-        return {'group': group}
```
seanh: JoinGroupController handles this now

seanh:
h.views.groups could do with a fair bit more cleaning up. I think it would probably be better if all the views for the 'group_read' route were together in one class, and the 'group_leave' route may not be necessary (just use the 'group_read' route with a 'leave' request param). But in this PR I'm doing only the tidy up that I need for the changes that this series of PRs is driving at - I need the group read page and the group join page to be separate view callables so that I can change their view config arguments separately

h/views/groups.py
```diff
@@ -149,25 +144,43 @@ def read_unauthenticated(group, request):
     return {'group': group}
 
 
-@view_config(route_name='group_read_noslug', request_method='GET')
-def read_noslug(group, request):
-    _check_slug(group, request)
+@view_defaults(route_name='group_read',
```
robertknight: So, there are three different `group_read` cases here, each guarded by a set of predicates that determines whether they apply:

- User logged in, has permission to read annotations in the group: Redirect to search page if that feature is enabled, show legacy group overview page otherwise
- User logged in, does not have permissions to read annotations in the group: Show group join page
- User logged out: Show group join page. The user is invited to log in first, after which they are redirected back to the join page

seanh: Yep. I like to use the view config as much as possible to make Pyramid choose the right view, instead of having branching logic in Python. Makes the view functions simpler

robertknight:
LGTM

PR #4157, opened 2016-12-02
Set up Travis job to lint Python code using flake8

robertknight:
This PR sets up Python code linting using flake8 using the same approach that we use in the client and service for front-end code, namely:

1. Add a lint config and make task that runs the linter, with all current violations whitelisted
2. Add a Travis job that runs the linter
3. Fix the violations in a series of commits, each of which removes one entry from the whitelist and fixes or suppresses the corresponding issues in the `h/` or `src/` directories.

In the interests of keeping the diff smaller, this PR only lints the H service and memex sources, excluding the tests and database migrations.

Annoyingly, it appears that flake8 does not currently support customized whitelists on a per-subdirectory basis like ESLint does, which is how I would have preferred to alter the whiltelists for the tests and DB migrations. What we can do in future however is add additional commands to the `lint-python` job which check those directories using a different config.

**An alternative** to this would be to use [Hound CI's beta Python support](https://github.com/hypothesis/h/pull/4157), however I think the approach here has several advantages:
- One less piece of infrastructure (since we're already using Travis for other things)
- It is obvious to new contributors how they can run the same lint checks locally as Travis does
- We can customize the lint tools that are run, the version that is used etc.
- We're already using this approach for front-end code in the client and service, where it has been working well.


.flake8
```diff
@@ -0,0 +1,7 @@
+[flake8]
+ignore =
+    # E126 - continuation line over-indented for hanging indent
+    E126,
+    # N813 - camelcase imported as lowercase
```
robertknight: N813 is allowed because of this frequent line: `from h.i18n import TranslationString as _`

src/memex/models/__init__.py
```diff
@@ -29,4 +29,5 @@
     'DocumentMeta',
     'DocumentURI',
     'merge_documents',
+    'update_document_metadata',
```
robertknight: This makes it visible to flake8 that `update_document_metadata` is imported here in order to re-export it.

chdorner:
I think we have discussed this at least at the developer meetup in Berlin and didn't reach a conclusion about automatically linting Python code. I like some of the cleanup you've done, but I'm not sure we really want a failing Travis build for linting violations. In addition to that, it's only for memex and h source code, and not for the tests. Are we comfortable saying that we have a different set of code style expectations in the tests vs the implementation?

It might be worth discussing the pros and cons (again) and make a decision on this as a team.

robertknight:
> I like some of the cleanup you've done, but I'm not sure we really want a failing Travis build for linting violations

There was the same concern when linting was added to the client/service frontend code. We addressed that there by making it is a separate job in Travis, so that it is distinct from a test failure and provides quick feedback (<1min). I've done the same thing here.

> In addition to that, it's only for memex and h source code, and not for the tests. 

The reason I omitted the tests here was to keep the number of changes in this PR down and to get feedback on the approach sooner, rather than because tests necessarily need different treatment. Using the same config here for the tests, there are 88 issues flagged up - so it wouldn't take that long to address them.

> It might be worth discussing the pros and cons (again) and come to a decision on this as a team.

Absolutely. For context, the original reason I looked into this was because style violations, unused imports etc. had been manually flagged up several times on a number of recent PRs.

.flake8
seanh: I think there's already flake8 configuration in the `setup.cfg` file, could this config go in there as well?

Makefile
```diff
@@ -40,6 +40,18 @@ test: node_modules/.uptodate
...
+.PHONY: lint
+lint: lint-js lint-python
+
+.PHONY: lint-js
+lint-js:
+	$(GULP) lint
+
+.PHONY: lint-python
+lint-python:
+	flake8 h src --exclude h/migrations/versions
```
seanh: Might be unnecessary given the flake9 config in `setup.cfg`?

nickstenning:
I'm afraid I share @chdorner's concerns here, but I want to emphasis one thing first: I **completely agree** that we should have automated reporting of linter violations. Unfortunately, I don't think Travis is the right place for this, for two reasons:

1. Linter feedback should not fail the build. If it fails the build, you're confusing a very important signal ("the build failed") with a broadly unimportant one ("your code is not lint-clean").
2. Running a linter in Travis is an all-or-nothing affair. You can't (easily) lint only changes introduced by the PR or branch you're interested in.

These are both problems, but 2. is by far the bigger problem. If you can only enable a linter rule if the entire codebase is lint-clean, then you have to do these big lint cleanups as you've done in this pull request. These are *dangerous*. You yourself discovered this on this branch when you followed the linter's advice and removed an "unused" import that wasn't unused. Yes, the linter identified a small problem (the relevant import wasn't in `__all__`) but the change it encouraged you to make was *incorrect* and broke the software. Luckily, our tests caught that one, but this will not always be the case.

That's why I am now, as I have advocated before, in favour of only enabling automated linting tools if they can lint the *changed code*, such as Hound.

seanh:
For what it's worth, I'm in favor of this pull request:

* As this PR itself amply demonstrates doing this will help to keep the code tidier

* It makes it more accessible for people who aren't already experienced Python developers to contribute to the backend code. All of our own frontend developers have been making flake8 violations when contributing Python code recently.

* It's nice to have something quickly and automatically point out code formatting and other flake8-detectable issues because it helps us to keep the code tidy and correct while also not having humans go back and forth over these issues in code review which is time wasting and distracting. Code reviews should provoke discussions, spending time on basic formatting issues won’t help with that goal.

* It avoids devs getting into arguments about code formatting and other flake8-able issues

* It helps us not to break the rule of no surprises in code review. While technically our dev docs do say that we use PEP8 and flake8, so a pull review author shouldn't be surprised to see changes requested if they submit flake8 violations, the fact is it's hard to keep an encyclopedia of all this stuff in your head, especially for devs less experienced with Python, recent pull requests on h demonstrate that most devs do not run flake8 locally before submitting a pr and this will only get worse as team size increases, and wouldn't you rather be thinking about the more important things? If a dev can submit a WIP PR and see that make lint passes then they know that basic stuff is right and they won't have any surprises of that sort when it gets reviewed

* I don't think it costs much time if you push a PR and have a build fail for a lint violation, to fix that violation and re-push. I think this costs less time than back and forth about lint issues in code review

* Automation. Let the tools deal with the automated things, spend your time thinking and talking about the more important things

nickstenning:
@seanh I agree with everything you've said. Everything you've said is arguing for automated reporting of linter violations, not specifically for doing it on travis and failing the build as a result.

seanh:
I'm not sure that I understand why failing the build is a big problem. If the build passes on master, and fails on your branch, then presumably you have a problem on your branch, and you can just fix it.

I do agree that something that works in the way that Hound works would be better. Landscape, for the brief time that it was working for us, was great. But unfortunately I haven't seen Hound (or any similar tool) work quickly or reliably for us? I think I recall that Hound didn't let us specify our own flake8 config either (and it doesn't support prospector at all). That said I'd still be happy with something like Hound if it quickly and reliably ran on each PR.

Failing that I think I prefer Rob's solution over having to do human back-and-forth about flake8 issues in code review and having new flake8 issues make it into master.

seanh:
Ah, apparently [Hound does let you provide your own flake8 config](https://houndci.com/configuration#python). Its Python support is still beta. I wonder if it works any better yet?

nickstenning:
Hound still has [some issues](https://github.com/houndci/python/issues/5) with parsing certain configuration options for flake8, but I actually opened [a PR](https://github.com/houndci/python/pull/7) the other day to address this.

seanh:
Hmm, no commits to Hound's Python support code for over a year. But the main hound repo looks active, and their python repo is small, and we can send our own prs to it.

Anyway, for my part I'd be content with either this PR or Hound.

robertknight:
I do want to respond to the comments above, but I'd prefer to focus on getting activity pages shipped first. I'm marking this as WIP until Friday.

nickstenning:
Hound have merged https://github.com/houndci/python/pull/7 and deployed it to their production service, so I've re-enabled Hound's Python support with https://github.com/hypothesis/h/commit/075a07f9128f2b714cd9fe8eb0af81175c7931e9.

Let's run with that for now and see how we get on.

nickstenning:
See also: https://github.com/hypothesis/h/pull/4238

robertknight:
Thanks for sorting this Nick!

PR #4159, opened 2016-12-02
Add user authority to its principals

chdorner:
This will allow us to implement new ACLs for groups where joinable/writeable can be allowed based on the group's and user's authority.


nickstenning:
LGTM.

PR #4160, opened 2016-12-02
Change group activity search url

seanh:
Fixes <https://github.com/hypothesis/h/issues/4004>. See commit messages.

### DIFF
--- a/h/views/activity.py
+++ b/h/views/activity.py
@@ -8,6 +8,7 @@ import urlparse
 
 from jinja2 import Markup
 from pyramid import httpexceptions
+from pyramid import security
 from pyramid.view import view_config
 from pyramid.view import view_defaults
 from memex.search import parser
@@ -88,10 +89,13 @@ class SearchController(object):
         }
 
 
-@view_defaults(route_name='activity.group_search',
-               renderer='h:templates/activity/search.html.jinja2')
+@view_defaults(route_name='group_read',
+               renderer='h:templates/activity/search.html.jinja2',
+               effective_principals=security.Authenticated,
+               has_feature_flag='search_page',
+               has_permission='read')
 class GroupSearchController(SearchController):
-    """View callables unique to the "activity.group_search" route."""
+    """View callables unique to the "group_read" route."""
 
     def __init__(self, group, request):
         super(GroupSearchController, self).__init__(request)
@@ -207,7 +211,7 @@ class GroupSearchController(SearchController):
         new_params['q'] = parser.unparse(parsed_query)
 
         location = self.request.route_url(
-            'activity.group_search', pubid=self.group.pubid,
+            'group_read', pubid=self.group.pubid, slug=self.group.slug,
             _query=new_params)
 
         return httpexceptions.HTTPSeeOther(location=location)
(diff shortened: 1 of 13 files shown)


h/views/groups.py
```diff
@@ -114,14 +114,15 @@ def _update_group(self, appstruct):
              request_method='GET',
              renderer='h:templates/groups/share.html.jinja2',
              effective_principals=security.Authenticated,
+             has_feature_flag=not_('search_page'),
              has_permission='read')
```
seanh: Just want to highlight something from the commit message here - `has_feature_flag` is not strictly necessary here, without it Pyramid will still find `GroupSearchController` when the feature flag is _on_, but I like to be clear and explicit. See commit message.

h/views/groups.py
```diff
@@ -114,14 +114,15 @@ def _update_group(self, appstruct):
...
 def read(group, request):
     """Group view for logged-in users."""
-    _check_slug(group, request)
+    check_slug(group, request)
```
seanh: I'm not super happy about changing an `h.views.groups` private helper function into a public one and having `h.views.activity` import it but it's a quick way to get the job done and I want to refactor these views more later so that the group_read views aren't spread across two separate files.

tests/h/views/activity_test.py
```diff
@@ -147,6 +147,27 @@ def test_init_returns_404_when_feature_turned_off(self,
...
+        the ones in h.views.activity do as well.
+
+        """
+        pyramid_request.matchdict['slug'] = 'wrong'
+
+        with pytest.raises(httpexceptions.HTTPMovedPermanently) as exc:
+            controller.search()
+
+        assert exc.value.location == '/groups/{pubid}/{slug}'.format(
+            pubid=group.pubid, slug=group.slug)
```
seanh: If all of the group_read views were in a single controller class, which is how I think we ought to refactor it later, then the controller's `__init__()` could do the redirecting of `GET` requests in one place and we could test it in one place as well

h/views/activity.py
```diff
@@ -88,17 +90,22 @@ def user_link(userid):
         }
 
 
-@view_defaults(route_name='activity.group_search',
-               renderer='h:templates/activity/search.html.jinja2')
+@view_defaults(route_name='group_read',
+               renderer='h:templates/activity/search.html.jinja2',
+               effective_principals=security.Authenticated,
+               has_feature_flag='search_page',
+               has_permission='read')
```
seanh: I really like the clarity of this, to be honest. (Although later I would like to look into writing tests for the routing)

seanh: `feature_flag` might be a better name than `has_feature_flag`, as it matches better with the naming style of Pyramid's built-in predicates. On the other hand currently all of our custom predicates are named `has_*` and it's nice to distinguish custom from built-in predicates. (Also `has_permission` can't be `permission` because there's already a built in, non-predicate argument with that name.) So I'm not sure. The names can always be improved later!

robertknight: Yes, this is nice - although as per the discussion in #backend earlier, eliminating the activity pages feature flag ASAP will simplify the possible combinations somewhat.

robertknight:
LGTM

PR #4161, opened 2016-12-02
Fix RSS and Atom validator errors

### LINKED ISSUE #3976: h rss feed doesn't pass feedvalidator check
### Steps to reproduce
1. http://validator.w3.org/feed/check.cgi?url=https%3A%2F%2Fhypothes.is%2Fstream.rss%3Furi%3Dhttp%3A%2F%2Fblog.jonudell.net
### Expected behaviour

Feed should validate
### Actual behaviour

It doesn't, the validator reports a couple of problems.

chdorner:
Check the individual commit messages for more information. This makes sure to pass the [w3 feed validator](http://validator.w3.org/feed/check.cgi) for both our RSS and Atom feeds.

Fixes #3976.

### DIFF
--- a/h/feeds/render.py
+++ b/h/feeds/render.py
@@ -65,8 +65,6 @@ def render_rss(request, annotations, rss_url, html_url, title, description):
     :rtype: pyramid.response.Response
 
     """
-    request.response.content_type = "application/rss+xml"
-
     def annotation_url(annotation):
         """Return the HTML permalink URL for the given annotation."""
         return request.route_url('annotation', id=annotation.id)
@@ -76,5 +74,7 @@ def render_rss(request, annotations, rss_url, html_url, title, description):
         rss_url=rss_url, html_url=html_url, title=title,
         description=description)
 
-    return renderers.render_to_response(
+    response = renderers.render_to_response(
         'h:templates/rss.xml.jinja2', {"feed": feed}, request=request)
+    response.content_type = 'application/rss+xml'
+    return response
--- a/h/feeds/render.py
+++ b/h/feeds/render.py
@@ -25,8 +25,6 @@ def render_atom(request, annotations, atom_url, html_url, title, subtitle):
     :rtype: pyramid.response.Response
 
     """
-    request.response.content_type = "application/atom+xml"
-
     def annotation_url(annotation):
         """Return the HTML permalink URL for the given annotation."""
         return request.route_url('annotation', id=annotation.id)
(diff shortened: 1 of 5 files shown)


h/feeds/render.py
```diff
@@ -40,8 +38,10 @@ def annotation_api_url(annotation):
         annotation_url=annotation_url, annotation_api_url=annotation_api_url,
         html_url=html_url, title=title, subtitle=subtitle)
 
-    return renderers.render_to_response(
+    response = renderers.render_to_response(
         'h:templates/atom.xml.jinja2', {"feed": feed}, request=request)
+    response.content_type = "application/atom+xml"
+    return response
```
seanh: Huh. I wonder if we could add a failing / passing test for this? Seems like it'd just have to assert against the returned `response.content_type`?

h/feeds/rss.py
```diff
@@ -9,18 +13,16 @@
 _ = i18n.TranslationStringFactory(__package__)
 
 
-def _pubDate_string_from_annotation(annotation):
-    """Return a correctly-formatted pubDate string for the given annotation.
+def _pubDate_string(timestamp):
+    """Return a RFC822-formatted pubDate string for the given timestamp.
```
seanh: Should this be RFC**2**822?

h/feeds/rss.py
```diff
@@ -1,4 +1,8 @@
 """Functions for generating RSS feeds."""
+
+from calendar import timegm
+from email.utils import formatdate
```
seanh: Great use of the standard library. I'm also learning about new standard lib things from your PRs!

h/templates/rss.xml.jinja2
```diff
@@ -22,7 +22,7 @@ render the feed to RSS XML.
        <title>{{ item.title }}</title>
        <description>{{ item.description|safe }}</description>
        <pubDate>{{ item.pubDate }}</pubDate>
-       <guid isPermaLink="true">{{ item.guid }}</guid>
+       <guid isPermaLink="false">{{ item.guid }}</guid>
```
seanh: Hmm, fair enough, since these are not actually URLs then I suppose they can't really be called permalinks

h/feeds/rss.py
chdorner: Ah! The feed validator refers to the older RFC 822 which got superseded by RFC2822. The date format has stayed the same. I will update the comment to say RFC 2822, that's also what the Python docs for `email.utils.formatdate` say.

h/feeds/render.py
chdorner: Done

PR #4162, opened 2016-12-02
Show tags for each annotation card.

sheetaluk:
https://github.com/hypothesis/product-backlog/issues/56


h/static/styles/partials-v2/_annotation-card.scss
```diff
@@ -43,6 +43,27 @@
...
 
+.annotation-card__tags {
+  display: flex;
+  flex-direction: row;
+  flex-wrap: wrap;
+}
+
+.annotation-card__tag {
+  text-decoration: none;
+  border: 1px solid #d3d3d3;
```
robertknight: We should always try to use colors from the color palette.

h/templates/includes/annotation_card.html.jinja2
```diff
@@ -1,47 +1,53 @@
...
-      {{ annotation.quote|safe }}
+      {{ presented_annotation.annotation.quote|safe }}
     </section>
   {% endif %}
   <section class="annotation-card__text">
-    {{ annotation.text_rendered|safe }}
+    {{ presented_annotation.annotation.text_rendered|safe }}
+  </section>
+  <section class="annotation-card__tags">
+    {% for tag in presented_annotation.annotation.annotation.tags%}
```
robertknight: I wonder if it would make sense to use the same component for tags in the sidebar and in annotation cards.

robertknight: > {% for tag in presented_annotation.annotation.annotation.tags%}

All the other properties of the real Annotation object are accessed via a forwarding property in the AnnotationHTMLPresenter class. I'd suggest we should do the same for tags for consistency.

h/templates/includes/annotation_card.html.jinja2
```diff
@@ -1,47 +1,53 @@
-{% macro annotation_card(annotation, request, incontext_link, group=none) -%}
-
+{% macro annotation_card(presented_annotation, request, tag_link) -%}
```
robertknight: It would be helpful here to have some documentation about the expected types of these arguments.

robertknight:
> Here is a screenshot of an annotation card with lots of tags and some really long tags.

Ah, I don't see a screenshot :P I have verified that the tags get elided correctly locally though.

h/static/styles/partials-v2/_annotation-card.scss
```diff
@@ -43,6 +43,27 @@
...
+  flex-direction: row;
+  flex-wrap: wrap;
+}
+
+.annotation-card__tag {
+  text-decoration: none;
+  border: 1px solid $grey-3;
+  border-radius: 2px;
+  padding: 0 5px 2px;
+  color: #969696;
```
robertknight: These two colors here should use the palette if possible as well.

h/templates/includes/annotation_card.html.jinja2
sheetaluk: For now lets leave it as is, but possibly as a followup PR.

h/templates/includes/annotation_card.html.jinja2
```diff
@@ -1,47 +1,67 @@
-{% macro annotation_card(annotation, request, incontext_link, group=none) -%}
+{#
+  Annotation card component.
 
+  Render an annotation card which has a username, timestamp,
+  an annotation quote, description and a list of tags.
```
robertknight: This description is very helpful 👍

h/templates/includes/annotation_card.html.jinja2
```diff
@@ -1,47 +1,67 @@
...
+
+  Usage:
+    {{ annotation_card(presented_annotation, request, tag_link) }}
+
+  :param presented_annotation: The annotation object which includes all the fields to display.
+  :param request: The request param containing the route_url method.
+  :param tag_link: Method which generates a link for a given tag.
```
robertknight: Just for future reference, the Python term for an object which can be called/invoked (such as _tag_link_) is a _callable_ - which can be either a stand-alone function or a method in a class.

robertknight:
LGTM

Issue #4116, opened 2016-11-24
Searching within a user or group is broken on iOS


seanh:
I can only reproduce this on iOS (iPhone 5) Safari, not in Chrome (not even in dev tools iPhone 5 device mode):

1. Go to a user or group's page:

![img_0637](https://cloud.githubusercontent.com/assets/22498/20600279/b3439390-b24a-11e6-9574-6b497dbe6105.PNG)

2. Type something else after the user or group name in the search and submit. In this case I typed "goo":

![img_0638](https://cloud.githubusercontent.com/assets/22498/20600300/d3e1c6a8-b24a-11e6-9a94-a21f5183a583.PNG)

Problems:

* My "goo" lozenge is truncated after one letter, but that's a [separate issue](https://github.com/hypothesis/h/issues/4115)
* The "user:Sean Hammond" lozenge has disappeared (and I'm no longer on the user page, it has redirected back to `/search`, which it doesn't do in other browsers)

chdorner:
Just investigating this. The same happens on a group search.

chdorner:
Ah! So this is the path that Safari on iOS calls after hitting Enter in the search form: `/users/chdorner?delete_lozenge=&q=goo`

PR #4163, opened 2016-12-02
Fix user and group search pages on iOS Safari

seanh:
Fix https://github.com/hypothesis/h/issues/4116

When submitting the search form using the keyboard on the
/groups/{pubid}/{slug} page or the /users/{username} page on mobile
Safari, the group or user lozenge was being incorrectly deleted and the
browser incorrectly redirected to the /search page. On other browsers
this was not happening.

There is an `<input type="submit">` that's hidden with CSS that is the
submit button that's submitted when you hit Enter in the search form.
The problem is that mobile Safari was not finding this submit button,
skipping over it and instead finding the delete button on the user or
group lozenge and submitting that instead.

The reason turns out to be that display: none; was used in CSS to hide
the hidden input, and that causes Safari to ignore it for form
submission purposes. (This seems like a bug in mobile Safari - CSS
probably shouldn't affect form behavior.)

The fix is to use visibility: hidden; instead. That then breaks the
search box layout, so add position: absolute; to fix it again.

### DIFF
--- a/h/templates/panels/navbar.html.jinja2
+++ b/h/templates/panels/navbar.html.jinja2
@@ -44,7 +44,7 @@
             presses Enter in the search bar input, the browser will act as if
             the user had clicked one of the form's other submit elements
             (e.g. the leave group button). #}
-        <input type="submit" class="u-hidden">
+        <input type="submit" class="nav-bar__search-hidden-input">
 
         {{ svg_icon('search', 'search-bar__icon') }}
         <div class="search-bar__lozenges" data-ref="searchBarLozenges">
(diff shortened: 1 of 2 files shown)


h/templates/panels/navbar.html.jinja2
```diff
@@ -44,7 +44,7 @@
             presses Enter in the search bar input, the browser will act as if
             the user had clicked one of the form's other submit elements
             (e.g. the leave group button). #}
-        <input type="submit" class="u-hidden">
```
seanh: "u-hidden" isn't used anywhere anymore but it seems useful to have / inevitable that we will need it one day so I left it in the CSS

robertknight:
I found a related upstream bug: https://bugs.webkit.org/show_bug.cgi?id=64733

According to this, the issue affects IE as well.

Issue #4141, opened 2016-11-29
Remove the unnecessary `?q=` from activity search URLs when there's no search query


seanh:
It sometimes appears, not sure exactly when and when not.

This was split out from https://github.com/hypothesis/h/issues/4004

dwhly:
Also, is the "q=" necessary even when we do need a query string?

nickstenning:
> we're not currently using it on the stream.

Yes, we are! And yes, it is necessary: https://en.wikipedia.org/wiki/Query_string

seanh:
I think I'm a bit too nackered at the end of the week to really think about this, but the situations in which this `?q=` appears are now reduced: https://hypothes-is.slack.com/archives/public/p1480699670008861 But you can still get it by, for example:

* Going to `/search` and hitting `Enter` in the search field without typing in anything (this seems not worth fixing to me, and I think it could only be fixed by some js to prevent submission of an empty form)

* Clicking to delete the last lozenge on the `/search` page

* Clicking to delete the second-last lozenge (after the user or group lozenge) on a user or group search page

* Clicking to delete the user or group lozenge on a user or group search page when that's the only lozenge present and getting redirected back to `/search?q=`

* On mobile, on a user or group search page, clicking the more info and back buttons when there's no search query

On user and group pages "clicking to delete lozenge" means clicking on either the x on the lozenge or a tag or username in the sidebar.

So an empty `?q=` may appear in a variety of ways when you're clicking and searching around on the search page. But it no longer appears when you click on a simple link to the search page or to a user or group's page.

So I think the severity of this issue has been reduced somewhat at least

PR #4164, opened 2016-12-02
Don't put empty ?q= params on URLs

seanh:
Avoid putting empty ?q= params at the end of activity search page URLs, when redirecting the browser

Fix https://github.com/hypothesis/h/issues/4141

* When clicking to delete the first user or group lozenge on a user or group search, if the search query is empty, redirect to `/search` instead of `/search?q=`

* When clicking _Back_ button on a user or group search on mobile, if the search query is empty, redirect to the user or group's page _without_ a trailing `?q=`

  The _More info_ button _does_ still produce an empty `q`, you get `?q=&more_info`. This is because the browser submits the search form with an empty `q` field, it's the _browser_ that requests the `?q=&more_info` URL, and since the server no longer needs to do any redirect for the _More info_ button it doesn't have a chance to remove it. I think the only way to fix this would be to add in a server-side redirect just for the sake of removing the `?q=` or to use JavaScript to remove an empty q from the form submission, so I haven't bothered.

* When leaving a group, if the search query is empty, you get redirected to `/search` not `/search?q=`

* When clicking on tags or users in the sidebar on a user or group search page to remove the second-last lozenge (other than the first user or group lozenge) you get redirected to the user or group's page _without_ a trailing `?q=`

* Go to the `/search page` (or a user or group search page) and hit `Enter` in the search field without typing any search - you will get to `/search?q=`. I think the only way to fix this would be to have JavaScript remove an empty q from the form submission and/or to have the server do a redirect to `/search`, so I haven't bothered.

* Click on the `(x)` button to delete the last dynamic (JavaScript-created) lozenge on `/search` or a user or group search page, and you will get a trailing `?q=`. This is because it's JavaScript, not the server, that handles deleting the javascript-created lozenge from the search query when you click an `(x)`. So this would need to be fixed in JavaScript or by refactoring to get the server to delete these lozenges instead.

### DIFF
--- a/h/views/activity.py
+++ b/h/views/activity.py
@@ -170,7 +170,7 @@ class GroupSearchController(SearchController):
         groups_service.member_leave(self.group,
                                     self.request.authenticated_userid)
 
-        new_params = self.request.POST.copy()
+        new_params = _copy_params(self.request, self.request.POST.copy())
         del new_params['group_leave']
         location = self.request.route_url('activity.search', _query=new_params)
 
@@ -193,7 +193,7 @@ class GroupSearchController(SearchController):
         userid = self.request.params['toggle_user_facet']
         username = util.user.split_user(userid)['username']
 
-        new_params = self.request.params.copy()
+        new_params = _copy_params(self.request)
 
         del new_params['toggle_user_facet']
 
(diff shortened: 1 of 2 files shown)


seanh:
This fixes the issue but the attempt at code sharing breaks some buttons because it assumes POST requests but the page is actually inconsistent about whether it uses GET or POST for different buttons. Should probably fix that inconsistency first...

seanh:
There might be a way to do this with less duplication by having the `SearchController` base class detect `GET` requests with an empty `q` param, maybe in the `__init__()` method, and do a redirect to the same URL without the `q`. It would result in more redirects (in some cases the other view methods would redirect to a URL with an empty `?q=` that, when requested, would result in a redirect to the same URL without the `q`), but the code would be less duplicative.

PR #4166, opened 2016-12-05
Fix tags aggregation

chdorner:
Analyzed fields do affect aggregations as they return buckets of the
tokenized field values, not the raw string ([see here]). Thus we are adding a new
field to the search index that contains the non-analyzed list of tag
strings.

Similar to the `user_raw` field that was introduced in 3514315.

[see here]: https://www.elastic.co/guide/en/elasticsearch/guide/1.x/aggregations-and-analysis.html#aggregations-and-analysis

### DIFF
--- a/src/memex/presenters.py
+++ b/src/memex/presenters.py
@@ -104,6 +104,7 @@ class AnnotationSearchIndexPresenter(AnnotationBasePresenter):
             'uri': self.annotation.target_uri,
             'text': self.text,
             'tags': self.tags,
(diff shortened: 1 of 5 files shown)


chdorner:
Marked WIP because I need to check with @nickstenning first about the reindex strategy. I think we still have to disable writes to the index while the reindex is working. And this change needs a full reindex immediately after being deployed.

nickstenning:
LGTM.

PR #4169, opened 2016-12-05
Make all search page requests use get

seanh:
This will unblock https://github.com/hypothesis/h/pull/4164

Make all the forms and buttons on the search (and group and user search)
pages use GET rather than POST requests.

GET is more semantically correct for submitting a search form (because
the requests are idempotent / they don't change anything on the server). This
is also more consistent as some of the search form fields and buttons
were using GET and some POST - now they're all GET.

The one button on the search page that remains a POST request is the
"Leave group" button on the group search page - this is semantically
correct as a POST.

For most buttons this is simply a change from "GET" to "POST" and the
Python code now needs to look at request.params instead of request.POST.

The one exception is the "More info" button - this button no longer
requires a server-side redirect that it had, so that Python code is
removed.

This does result in one behavior change - when on the more info page, if
you submit the search form (by hitting enter in the search box or by
clicking on a tag or username) when the page reloads you will be back on
the search results page looking at the updated results. This is arguably
an improvement.

### DIFF
--- a/h/templates/activity/search.html.jinja2
+++ b/h/templates/activity/search.html.jinja2
@@ -10,7 +10,6 @@
   <nav class="search-result-nav">
     {%- if more_info %}
       <button form="search-bar"
-              formaction="{{ request.path_url }}"
               name="back"
               class="search-result-nav__button">
         {% trans %}Back{% endtrans %}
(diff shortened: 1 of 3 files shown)


h/templates/activity/search.html.jinja2
```diff
@@ -213,7 +207,6 @@
                   type="submit"
                   form="search-bar"
                   formmethod="POST"
```
seanh: The group leave button is deliberately still a POST

robertknight:
LGTM

PR #4170, opened 2016-12-06
Raise Elasticsearch timeout for reindex CLI command

chdorner:
Since I'm always the one that forgets to set this env variable, now it
is automatically doing it for me.


nickstenning:
:+1:

PR #4171, opened 2016-12-06
Add `postgresql` to homebrew requirements

chdorner:
I've just set up the dev environment on my MacBook and ran into this issue when installing the python dependencies. Even though the install docs use PostgreSQL in a Docker container, we still need to install the homebrew package to be able to install the psycopg2 python package.


nickstenning:
👍

PR #4172, opened 2016-12-06
Increase limit of tags in the groups activity pages from 10 to 50.

sheetaluk:
Fixes https://github.com/hypothesis/product-backlog/issues/46

### DIFF
--- a/h/activity/query.py
+++ b/h/activity/query.py
@@ -142,11 +142,11 @@ def execute(request, query, page_size):
 
 
 def aggregations_for(query):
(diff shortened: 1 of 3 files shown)


dwhly:
@sheetaluk: from the issue, "Also, add language to indicate what that max number is, but only if the limit has been hit. "Top X most frequent tags"

dwhly:
So, right under "Top 50 tags"  ..
Maybe "Only the top 50 shown"

PR #4173, opened 2016-12-06
Fix minimum annotation card width

robertknight:
**Before:**

ann-card-short

**After:**

ann-with-flex-basis

The fix is to use `flex-basis` to set the initial ("preferred") width of an annotation card to 500px on desktop rather than using the size of the card's contents.

### DIFF
--- a/h/static/styles/partials-v2/_search.scss
+++ b/h/static/styles/partials-v2/_search.scss
@@ -201,11 +201,14 @@
   padding-bottom: 5px;
 }
 
+// The container for the contents of a search result bucket.
 .search-result-bucket__content {
   padding-left: 10px;
   padding-right: 10px;
 }
 
+// The contents of a search result bucket which is shown when the bucket is
+// expanded.
 .search-result-bucket__annotation-cards-container {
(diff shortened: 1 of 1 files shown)


h/static/styles/partials-v2/_search.scss
```diff
@@ -367,6 +375,10 @@ $stats-icon-column-width: 20px;
 
   .search-result-bucket__annotation-cards {
     width: 100%;
+
+    // Set the preferred height of the annotation card list based on its
+    // contents.
+    flex-basis: auto;
```
robertknight: On smaller screens the bucket sidebar and card list are arranged vertically and `flex-basis` controls the height, which we want to be based on the card's contents.

seanh:
Yay! I hated seeing this

h/static/styles/partials-v2/_search.scss
seanh: _Nice_ catch, this would have been easy to miss

seanh:
I'm really happy to see this, this card width bug was really bad for the overall feel of the pages

PR #4174, opened 2016-12-06
Share link on results

### LINKED ISSUE #4113: No share link to annotation on annotation card
Need to add the share icon that allows users to share the annotation over social channels on the activity pages implementation of the annotation card. 

**See Mockup:** 
screen shot 2016-11-23 at 4 19 39 pm

**Current production view**
screen shot 2016-11-23 at 4 20 51 pm

sean-roberts:
Fixes #4113 

The intent for my changes was to make it easy for use to extract the share-widget module into it's on encapsulated module that can be used in client, h, etc. It only goes as far as to make the pieces that will be pulled in more isolated from the h app itself.

The diff looks large but there were lots of files (icons, importing stuff) that are brand new but largely a carry over from what worked in the client (besides the JS since before it was angular based).


h/templates/includes/annotation_card.html.jinja2
```diff
@@ -62,7 +62,14 @@
   <footer class="annotation-card__footer">
     {% if presented_annotation.incontext_link %}
       <a href="{{ presented_annotation.incontext_link }}" target="_blank" title="{% trans %}Visit annotation in context{% endtrans %}">
-        {{ svg_icon('up-right-arrow', 'annotation-card__incontext-link') }}
+        {{ svg_icon('up-right-arrow', 'annotation-card__footer-link annotation-card__incontext-link') }}
+      </a>
+      <a href="#" share-widget-config='{
+        "url": "{{presented_annotation.incontext_link}}",
+        "private": {%if presented_annotation.annotation.shared and not presented_annotation.group %}true{%else%}false{%endif%},
```
robertknight: This conditional is the wrong way round - `private` should be true if the annotation is _not_shared.

h/templates/includes/share_widget.html.jinja2
```diff
@@ -0,0 +1,55 @@
+{% macro share_widget() -%}
+
+<div class="share-widget">
+  <div class="share-widget-target">
+    <span class="share-widget-target__label">Share:</span>
+    <a share-target-href="https://twitter.com/intent/tweet?url={href}&hashtags=annotated"
+      target="_blank"
```
robertknight: Hmm, we open each of these share links in a new window but I notice that the Facebook, Twitter and G+ UIs that appear all look like they are designed to appear in a popup, and the example buttons at the bottom of https://dev.twitter.com/web/tweet-button do

robertknight:
The UI is pretty solid. Here are a few issues I noted in testing. Some of these also apply to the sidebar's implementation, but perhaps lets use this as an opportunity to get them sorted:

- The "Only me" indicator is shown on _public_ annotations instead of _private_ ones.
- Of the links in the footer of the card, I think the one that should be given primacy is the "Visit in context" link, therefore I would suggest that would put it as the right-most link on the card
- The Twitter, Facebook and G+ links all open new tabs, although the UI that appears in each case looks like it is intended to be displayed in a small popup window, and the example at the bottom of the Twitter dev page (https://dev.twitter.com/web/tweet-button ) does indeed appear in a popup.
- On iOS, tapping the "Copy" button zooms into the widget, whereas it doesn't do this for the existing group invite link.
- On iOS, tapping outside the share link dialog does not hide it.
- The footer buttons have quite small hit targets - I'd suggest that we increase their size on mobile (search for uses of the `touch-input` mixin in our styling for existing examples).

In terms of the technical implementation - the basic separation between link and share widget is fine, and it makes sense that you're continuing to re-use the existing infrastructure for templating, icons and the copy button.

I think I understand the motivation for exposing the actual widget as `window.h.shareWidget` but honestly, I think the whole discussion about making this a re-usable component is a bit of a distraction at the moment - there are a whole bunch of issues we'd need to deal with (styling, icons, packaging, theming) - and that is assuming it even makes sense to use the same UI for this in the sidebar and AP. As it stands at the moment, I think it would be better to follow the conventions/patterns that we already have in the code and then figure out whether/what/how to share parts with the sidebar separately in future.

h/static/images/icons/google-plus.svg
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<svg width="22px" height="14px" viewBox="0 0 22 14" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
+    <!-- Generator: Sketch 41.2 (35397) - http://www.bohemiancoding.com/sketch -->
+    <title>logo_google_plus_32px_clr copy 6</title>
+    <desc>Created with Sketch.</desc>
+    <defs></defs>
+    <g id="Page-1" stroke="none" stroke-width="1" fill="none" fill-rule="evenodd">
+        <g id="logo_google_plus_32px_clr-copy-6" fill="#A6A6A6">
```
robertknight: For the other monochrome SVG icons, I've replaced references to a hard-coded fill color with `currentColor`, so that the icon inherits its color from the CSS `color` property of the containing `<svg>` element.

h/static/images/icons/mail.svg
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<svg width="18px" height="14px" viewBox="0 0 18 14" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
+    <!-- Generator: Sketch 41.2 (35397) - http://www.bohemiancoding.com/sketch -->
+    <title>Rectangle 93 Copy 12</title>
+    <desc>Created with Sketch.</desc>
+    <defs></defs>
+    <g id="Page-1" stroke="none" stroke-width="1" fill="none" fill-rule="evenodd">
+        <path d="M2.57142857,2 L0,2 L0,14 L18,14 L18,2 L15.4285714,2 L18,0 L0,0 L2.57142857,2 L0,2 L9,9 L18,2 L15.4285714,2 L9,7 L2.57142857,2 Z" id="Rectangle-93-Copy-12" fill="#A6A6A6"></path>
```
robertknight: Same comment as https://github.com/hypothesis/h/pull/4174/files#r91279058 for the hard-coded color here and in the other icons.

h/static/scripts/controllers/share-widget-controller.js
```diff
@@ -0,0 +1,211 @@
+'use strict';
+
+const Controller = require('../base/controller');
+
+const TEMPLATE_SELECTOR = '.share-widget-template';
```
robertknight: In our CSS style guide, we specify that all classes referenced by code should use a `js-` prefix, except in tests: https://github.com/hypothesis/frontend-toolkit/blob/master/docs/css-style-guide.md#classes-used-in-code

h/static/scripts/controllers/share-widget-controller.js
```diff
@@ -0,0 +1,211 @@
...
+const PRIVATE_MSG_SELECTOR = '.share-widget-msg-private';
+const GROUP_MSG_SELECTOR = '.share-widget-msg-group';
+
+const ARROW_PADDING_RIGHT =  16;
+const ARROW_PADDING_BOTTOM =  5;
+
+
+// Travel up the nodes parents to find a matching
+// element with the share-widget configuration values
+const _getParentWithConfig = function(node) {
```
robertknight: We should be consistent about using arrow functions here. This ought to have been flagged up by ESLint, will have to look into why that did not happen.

h/static/scripts/controllers/share-widget-controller.js
```diff
@@ -0,0 +1,211 @@
...
+    // share widgets. It will only actually run once
+    // but having a widget at all on the page and all
+    // other share-widget components will be handled
+    // through psuedo event delegation.
+    this._setupGlobalWidget();
+  }
+
+  _setupGlobalWidget() {
+
+    if (!window.h || !window.h.shareWidget) {
```
robertknight: Rather than polluting the `window` namespace, we could just make this a property of the `ShareWidget` class.

h/static/scripts/controllers/share-widget-controller.js
```diff
@@ -0,0 +1,211 @@
...
+
+      window.h = Object.assign({}, window.h, {
+
+        shareWidget: {
```
robertknight: If `shareWidget` is a class rather than just a bag of properties I suggest we follow our conventions and make it an actual `class` at the top of the file using standard method syntax rather than `function` expressions.

h/static/scripts/controllers/share-widget-controller.js
```diff
@@ -0,0 +1,211 @@
...
+              groupMessage.style.display = 'block';
+            }
+          },
+
+
+          /**
+           * Given a node, update the template, repostion the widget properly,
+           *  and make it visible.
+           *
+           * @param  {HTMLNode} node The trigger node that we will place the widget
```
robertknight: `HTMLNode` isn't a thing in the HTML spec. Did you mean `Element`?

h/static/scripts/controllers/share-widget-controller.js
```diff
@@ -0,0 +1,211 @@
...
+              return;
+            }
+
+            this._renderWidgetTemplate(config);
+
+            // offsets affecting height need to be updated after variable replacement
+            const widgetOffsets =  _getOffset(this._widget);
+            const nodeOffset = _getOffset(node);
+
+            // console.log(this._widget, nodeOffset.top, widgetOffsets.height, ARROW_PADDING_BOTTOM);
```
robertknight: Debug statement left in by mistake.

h/static/scripts/controllers/share-widget-controller.js
```diff
@@ -0,0 +1,211 @@
...
+
+            // on initialize we need to reset container visbility
+            this.hide();
+
+
+            window.document.body.addEventListener('click', (event) => {
+              const target = event.target;
+
+              // do nothing if we are clicking inside of the widget
+              if (containerElement.contains(target) || containerElement === target) {
```
robertknight: This surprised me, but [Element#contains](https://dom.spec.whatwg.org/#dom-node-contains) is _inclusive_ so `someElement.contains(someElement)` returns true and you don't need the second branch of the `if` statement here.

h/static/scripts/polyfills.js
```diff
@@ -31,3 +31,23 @@ require('keyboardevent-key-polyfill');
...
+// from https://developer.mozilla.org/en/docs/Web/API/Element/matches
+if (!Element.prototype.matches) {
+  Element.prototype.matches =
+    Element.prototype.matchesSelector ||
+    Element.prototype.mozMatchesSelector ||
+    Element.prototype.msMatchesSelector ||
+    Element.prototype.oMatchesSelector ||
+    Element.prototype.webkitMatchesSelector ||
+    function(s) {
+      const matches = (this.document || this.ownerDocument).querySelectorAll(s);
```
robertknight: Hmm, this seems a little bit inefficient as it searches the whole document rather than just the relevant subtree. But hey, we can fix that it if becomes a problem.

h/static/scripts/site.js
```diff
@@ -20,6 +20,7 @@ const SearchBarController = require('./controllers/search-bar-controller');
 const SearchBucketController = require('./controllers/search-bucket-controller');
 const SignupFormController = require('./controllers/signup-form-controller');
 const TooltipController = require('./controllers/tooltip-controller');
+const ShareWidgetController = require('./controllers/share-widget-controller');
```
robertknight: Nit: The items here and keys in the `controllers` array below should be sorted alphabetically.

h/static/scripts/tests/controllers/share-widget-controller-test.js
```diff
@@ -0,0 +1,154 @@
...
+    </button>
+    <button id="btn4" class="js-share-widget"
+      share-widget-config='{"url":"url4", "private":true, "group":true}'>
+      <span><b>btn</b></span>
+    </button>
+  `;
+
+  let container;
+
+  const getTriggers = function() {
```
robertknight: These should be arrow functions

h/static/scripts/tests/controllers/share-widget-controller-test.js
```diff
@@ -0,0 +1,154 @@
...
+      .click(btns[1], () => {
+        assert.isFalse(widgetIsVisble(), 'hidden after clicking another trigger while it is open');
+      })
+      .click(btns[3].querySelector('b'), () => {
+        assert.isTrue(widgetIsVisble(), 'opens when clicking elements inside of trigger');
+        done();
+      });
+  });
+
+  it('updates template properly per the configs', (done) => {
```
robertknight: Can we rephrase this in terms of how a user would see it? eg. "Share widget displays correctly for current share link". The reason being to make it more obvious what is broken if the test fails in a certain environment or as a result of refactoring.

h/static/styles/partials-v2/_annotation-card.scss
```diff
@@ -68,12 +68,22 @@
...
   flex-direction: row;
   justify-content: flex-end;
+  align-items: baseline;
 }
 
-.annotation-card__incontext-link {
-  height: 10px;
-  width: 10px;
-  color: $grey-5;
+.annotation-card__footer .annotation-card__footer-link {
```
robertknight: Do we need the descendant selector here or can we just use the `.annotation-card__footer-link`?

h/static/styles/partials-v2/_share-widget.scss
```diff
@@ -0,0 +1,138 @@
...
+  height:0;
+  width:0;
+}
+
+.share-widget.is-open {
+  display: flex;
+  flex-direction: column;
+}
+
+.share-widget {
```
robertknight: I presume much of this is adapted from the client's styling. It would be useful to have some brief (even just one line) comments above the main classes in this file indicating elements of the share dialog they are.

h/static/styles/site-v2.scss
```diff
@@ -25,6 +25,7 @@
 @import 'partials-v2/search-result-sidebar';
 @import 'partials-v2/tooltip';
 @import 'partials-v2/tabs';
+@import 'partials-v2/share-widget';
```
robertknight: Nit. The list of components should be sorted alphabetically.

h/static/styles/partials-v2/_share-widget.scss
```diff
@@ -0,0 +1,138 @@
...
+
+.share-widget-link__btn--copy {
+  margin-top: 5px;
+  margin-bottom: 5px;
+  width: 13px;
+  height: 16px;
+}
+
+.share-widget-action * {
+  fill: $grey-4;
```
robertknight: I'm afraid this isn't documented property in the comments for the `svg_icon` helper, but the intent (and how our other SVG icons work) is that you set a `color` property on the containing `<svg>` element and then use `fill="currentColor"` or `stroke="currentColor"` in the SVG itself to apply that color.

h/static/scripts/polyfills.js
sean-roberts: we don't support browsers that would use that function anyway. I am pulling in the whole polyfill for completeness. But it is the naming consistency that is needed.

sean-roberts:
So, the biggest difference now is the approach of the controller. Originally, the controllers were basically assigned to the trigger buttons. Instead, it is now assigned to a single node - wrapped around the template here, that will invoke the attachment of a global listener. This feels much cleaner without the weird indirection of triggers setting up a single global controller.

As discussed in slack, I made font bigger on small resolutions to prevent zoom on ios. The other issue of tapping away from the dialog does not close it on ios is only partially true. Ios doesn't bubble mouse events to things that do not have a `cursor:pointer` property. And assigning that to something like the body has effects of the whole body gets a highlight color on click. So, since this is not unique to my component (dropdowns have the same issue), let's solve this some other time.

Everything else though, has been addressed.

robertknight:
The unhovered/hover for the copy-to-clipboard button at the bottom of the group sidebar is broken, it shows in a dark state when hovered and not-hovered. In master it is only darkened when hovered.

h/templates/includes/annotation_card.html.jinja2
```diff
@@ -62,7 +62,14 @@
   <footer class="annotation-card__footer">
     {% if presented_annotation.incontext_link %}
       <a href="{{ presented_annotation.incontext_link }}" target="_blank" title="{% trans %}Visit annotation in context{% endtrans %}">
-        {{ svg_icon('up-right-arrow', 'annotation-card__incontext-link') }}
+        {{ svg_icon('up-right-arrow', 'annotation-card__footer-link annotation-card__incontext-link') }}
+      </a>
+      <a href="#" share-widget-config='{
```
robertknight: I wasn't sure if new lines were legal in HTML attribute values but apparently [they are](http://stackoverflow.com/questions/8047042/)

h/static/scripts/tests/controllers/share-widget-controller-test.js
```diff
@@ -0,0 +1,144 @@
...
+
+  const getTriggers = () => {
+    return Array.from(container.querySelectorAll('[share-widget-config]'));
+  };
+
+  const widgetIsVisble = () => {
+    return container.querySelector('.js-share-widget').style.visibility === 'visible';
+  };
+
+  const urlsReflectUpdate = (addedValue) => {
```
robertknight: It wasn't that clear to me from the function and argument names here what this does. `shareLinksMatchUrl = (url) => {...}`?

h/static/styles/partials-v2/_annotation-card.scss
```diff
@@ -68,12 +68,22 @@
...
-  width: 10px;
-  color: $grey-5;
+.annotation-card__footer-link {
+  margin-left: 15px;
+  color: $grey-4;
+  height: 16px;
+  width: 16px;
+}
+
+.annotation-card__footer .annotation-card__incontext-link {
```
robertknight: I don't think we need this descendant selector, given that the incontext link is only going be shown in the footer. Even if we did, the convention would be to use a modifier class in most cases (ie. `some-thing--variation` rather than `some-parent some-thing`)

h/static/styles/partials-v2/_share-widget.scss
```diff
@@ -0,0 +1,152 @@
...
+  text-decoration: none;
+  font-size: 20px;
+  cursor: pointer;
+}
+
+.share-widget-target__icon:hover {
+  color: $grey-7;
+}
+
+@media screen and (max-width: $max-phone-width) {
```
robertknight: AFAIK the 16px size exists because of the input method as well as the screen, so this is needed for tablets (eg. iPad) as well.

To be safe, in other cases we set this font size for all touch devices. See https://github.com/hypothesis/h/blob/master/h/static/styles/partials-v2/_form-input.scss#L212 for an example of the way we do this elsewhere.

h/static/scripts/controllers/share-widget-controller.js
```diff
@@ -0,0 +1,212 @@
...
+    }
+    if (node.hasAttribute(CONFIG_ATTR)) {
+      configNode = node;
+      break;
+    }
+  }
+  return configNode;
+};
+
+const _getOffset = (el) => {
```
robertknight: The `_` prefix on this helper function isn't needed. We use the `_` prefix on private class methods/variables because there isn't a way to avoid exporting them.

h/static/scripts/controllers/share-widget-controller.js
```diff
@@ -0,0 +1,212 @@
...
+      }
+
+      // hide the widget if the click was not handled by
+      // clicking on the triggers or widget itself
+      if (this._widgetVisible) {
+        this.hide();
+      }
+    };
+
+    window.document.body.addEventListener('click', this._handler);
```
robertknight: We usually access the document object via just `document` rather than `window.document`, it is a little terser.

h/static/scripts/polyfills.js
```diff
@@ -40,3 +40,23 @@ require('keyboardevent-key-polyfill').polyfill();
 // Fetch API
 // https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API
 require('whatwg-fetch');
+
+// Quick polyfill to normalize the naming of matches function
+// from https://developer.mozilla.org/en/docs/Web/API/Element/matches
+if (!Element.prototype.matches) {
```
robertknight: If we use the Element.closest API as suggested in https://github.com/hypothesis/h/pull/4174/files#r91678734 , we can remove this polyfill and replace it with `element-closest`.

h/static/scripts/site.js
```diff
@@ -36,6 +37,7 @@ const controllers = {
   '.js-search-bucket': SearchBucketController,
   '.js-signup-form': SignupFormController,
   '.js-tooltip': TooltipController,
+  '.js-share-widget': ShareWidgetController,
```
robertknight: Nit: This list of selectors should be sorted alphabetically.

h/static/scripts/tests/controllers/share-widget-controller-test.js
```diff
@@ -0,0 +1,144 @@
...
+      });
+  });
+
+  it('displays correctly for current share link', (done) => {
+
+    const btns = getTriggers();
+
+    syn
+      .click(btns[0], () => {
+        assert.isTrue(urlsReflectUpdate('url1'));
```
robertknight: I think it makes tests more useful and easier to understand if the test data (ie. the URL in this case) follows the same format as actual data that might be used. In other words, this should be a real URL - since the test needs to check that URL encoding works correctly.

h/static/styles/partials-v2/_share-widget.scss
```diff
@@ -0,0 +1,152 @@
...
+}
+
+.share-widget-target__icon:hover {
+  color: $grey-7;
+}
+
+@include touch-input {
+  // prevent zoom on smaller devices
+  .share-widget-link input {
+    font-size: 16px;
```
robertknight: This should use the `$touch-input-font-size` variable rather than hard-coding 16px. Although the value isn't likely to change soon, the variable has some useful documentation above it that explains where and why it should be used.

h/static/scripts/controllers/share-widget-controller.js
```diff
@@ -0,0 +1,212 @@
...
+  detach() {
+    window.document.body.removeEventListener('click', this._handler);
+  }
+}
+
+
+// Note, this seemlingly meaningless class is here so that we can map
+// our controller paradigm to the invokaction of the appropriate events. And leave
+// the ShareWidget compontent itself abstracted from this paradigm.
+class ShareWidgetController extends Controller {
```
robertknight: > And leave the ShareWidget compontent itself abstracted from this paradigm.

I don't think we need this really, the `Controller` class is very little code. We could just include it as part of any package of reusable code that is shared with the client or other projects. In the interests of getting this shipped - let's leave that discussion for later though.

What would be useful here is a JSDoc comment explaining what a `ShareWidgetController` is.

h/static/scripts/tests/controllers/share-widget-controller-test.js
```diff
@@ -0,0 +1,144 @@
+'use strict';
+
+const syn = require('syn');
+
+const ShareWidgetController = require('../../controllers/share-widget-controller');
+
+describe('ShareWidgetController', () => {
+
+  const widgetTemplate = `
```
robertknight: Worth adding a note here that this duplicates the `share_widget.html.jinja2` template. Right now duplicating the templates in tests doesn't cause that much pain, but we might want to reconsider in future.

h/static/styles/partials-v2/_share-widget.scss
```diff
@@ -0,0 +1,152 @@
+// Share widget
+// ---------------
+// The widget that allows social linking and the ability to
+// copy a url for use elsewhere.
+
+
+// container for widget should not be displayed
+.js-share-widget {
```
robertknight: There is an existing `is-hidden-when-loading` CSS class in `_util.scss` that you can use to hide elements on the page before the JS runs. This class is automatically removed once the element has been upgraded.

To use it, all you need to do is:

```diff
--- a/h/templates/activity/search.html.jinja2
+++ b/h/templates/activity/search.html.jinja2
@@ -415,7 +415,7 @@
       {{ user_sidebar(user) }}
     {% endif %}
 
-    <div class="js-share-widget">
+    <div class="js-share-widget is-hidden-when-loading">
       {{ share_widget() }}
     </div>
```

h/templates/includes/annotation_card.html.jinja2
robertknight: I haven't closely looked at keyboard or screen-reader accessibility for this, and it isn't too much of an issue since the incontext link is available as a plain link right text next to it. However, I think there are a couple of basic things we can at least do:

- Set a `title` attribute on the link
- Set the `aria-haspopup="true"` attribute on the link

robertknight:
OK, we're getting close. I have a couple of small UX queries:

1. Clicking a share link on one card while one is open on another doesn't show the share link on the one just clicked. This looks like it is intentional decision here but I think it would be more convenient and conventional for clicking on a menu - see the footer links on twitter cards for example.
2. We should set at least some basic label/role attributes on the share link for accessibility

The implementation is easier to follow now. I've mentioned a few simplifications that I think we can make and have added a bunch of small comments on the code/tests. I think the ShareWidget/ShareWidgetController distinction can be avoided in future, but let's keep this as-is for the moment and get it merged.

h/static/styles/partials-v2/_share-widget.scss
sean-roberts: we are explicitly using visibility so that we can easily calculate height and width without displaying the element. With display:none, getBoundingClientRect returns 0. That utility uses display none so it's not helpful here

sean-roberts:
@robertknight updated!

h/static/scripts/controllers/share-widget-controller.js
```diff
@@ -0,0 +1,200 @@
...
+const ARROW_PADDING_BOTTOM =  5;
+
+let shareWidgetAttached = false;
+
+const getOffset = (el) => {
+  el = el.getBoundingClientRect();
+  return {
+    // adjust for top left of the document
+    left: el.left + window.scrollX,
+    top: el.top + window.scrollY,
```
robertknight: I forgot this when I looked before, but `window.scrollX` and `window.scrollY` are not available in IE 10/11. Changing this to `window.pageXOffset` and `window.pageYOffset` respectively works in IE 11 and newer browsers.

h/static/styles/partials-v2/_share-widget.scss
```diff
@@ -0,0 +1,154 @@
+// Share widget
+// ---------------
+// The widget that allows social linking and the ability to
+// copy a url for use elsewhere.
+
+
+.js-share-widget {
+  // we are explicitly using visibility so that we can
+  // easily calculate height and width without displaying the element.
+  // display:none causes 0 return value on getBoundingClientRect
```
robertknight: As I understand it, the purpose of this styling is to hide the widget before the JS runs, since in the `ShareWidget` constructor you call `hide()` anyway? If I remove this styling and add the `is-hidden-when-loading` class to the `<div>` with the `.js-share-widget`, everything still works as far as I can see. Am I missing something?

sean-roberts:
@robertknight updated

PR #4175, opened 2016-12-07
Fix search results right margin

seanh:
The contents of the nav bar header and of the search results container (which contains the search results and the sidebar) are supposed to be left- and right-aligned with each other but the right margins are actually different, the  nav bar contents are more indented:

screenshot from 2016-12-07 11-26-30

This PR aligns those two right margins correctly:

screenshot from 2016-12-07 12-03-46

### DIFF
--- a/h/static/styles/partials-v2/_base.scss
+++ b/h/static/styles/partials-v2/_base.scss
@@ -56,6 +56,8 @@ $touch-target-size: 44px;
 // ----
 $search-result-container-margin-left: 30px;
 $search-result-container-margin-right: 60px;
+$search-result-container-margin-left-on-small-screens: 10px;
+$search-result-container-margin-right-on-small-screens: 10px;
 
 // Include all mixins, making them available to the stylesheet. N.B. Mixins
 // should *never* output CSS, so this is safe.
--- a/h/static/styles/partials-v2/_base.scss
+++ b/h/static/styles/partials-v2/_base.scss
@@ -54,10 +54,10 @@ $touch-target-size: 44px;
 
 // Misc
 // ----
-$search-result-container-margin-left: 30px;
-$search-result-container-margin-right: 60px;
-$search-result-container-margin-left-on-small-screens: 10px;
-$search-result-container-margin-right-on-small-screens: 10px;
+$content-margin-left: 30px;
+$content-margin-right: 60px;
+$content-margin-left-on-small-screens: 10px;
+$content-margin-right-on-small-screens: 10px;
 
 // Include all mixins, making them available to the stylesheet. N.B. Mixins
 // should *never* output CSS, so this is safe.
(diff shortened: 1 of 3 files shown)


h/static/styles/partials-v2/_base.scss
```diff
@@ -52,6 +52,11 @@ $max-phone-width: 500px;
 $tablet-width: 768px;
 $touch-target-size: 44px;
 
+// Misc
+// ----
+$search-result-container-margin-left: 30px;
+$search-result-container-margin-right: 60px;
```
seanh: Wasn't sure where to put these variables, since they need to be accessed by two other files, there didn't seem to be any place other than `_base.scss`

h/static/styles/partials-v2/_nav-bar.scss
```diff
@@ -23,7 +25,6 @@
 }
 
 .nav-bar__logo {
-  margin-left: 30px;
```
seanh: The left and right margins were on the first and last children of the nav bar header, I've moved them onto the container element instead.

h/static/styles/partials-v2/_search.scss
```diff
@@ -38,8 +38,8 @@
 
   // Align the left edge of the search result list with the left edge of the
   // navbar content.
```
robertknight: This comment should perhaps be updated to say "left and right edges"

h/static/styles/partials-v2/_nav-bar.scss
robertknight: That makes things clearer 👍

h/static/styles/partials-v2/_base.scss
robertknight: OK, let's just keep an eye on this in future and not let it become a dumping ground.

h/static/styles/partials-v2/_search.scss
seanh: Done

seanh:
Just pushed a new commit with a fix for the margins on small screens. On master:

master

On this branch:

fixed

seanh:
So we now have two pairs of rules - for the margins on large screens and for the margins on small screens - which need to match up in `_search.scss` and `_nav-bar.scss`. Introducing variables helps, but it'd also be nicer to have these dependent rules in the same file. But we'd need a `_search-and-nav-bar.scss` or something. Maybe just a `_shared.scss`.

h/static/styles/partials-v2/_base.scss
```diff
@@ -52,6 +52,13 @@ $max-phone-width: 500px;
...
+$search-result-container-margin-left-on-small-screens: 10px;
+$search-result-container-margin-right-on-small-screens: 10px;
```
robertknight: On second thoughts, the name is a bit misleading since this margin applies to both the navigation and the content of the page and it should apply to _all_ pages, not just search results. `content-{left,right}-margin-{small,large}`?

seanh: Hmm, fair enough...

robertknight:
LGTM

PR #4177, opened 2016-12-07
Fix pagination control alignment

seanh:
Horizontally align the pagination control with the center of the search
results, rather than with the center of the page contents (which makes
it off-center relative to the search results).

Before:

screenshot from 2016-12-07 14-17-48

After:

screenshot from 2016-12-07 14-18-07


seanh:
@sean-roberts Done

Issue #4018, opened 2016-11-01
Search page content is not centered on page


seanh:
Once the group and user sidebars are in on the search page, let's center the page contents (including the contents of the nav bar at the top):

![screenshot from 2016-11-01 17-12-25](https://cloud.githubusercontent.com/assets/22498/19899041/7daea1f8-a056-11e6-8983-80cc54708d2c.png)

sean-roberts:
It's a pretty significant design decision to center content like that. We need to make sure we are doing this correctly and are weighing the UX impacts it has.

seanh:
This was a request from Conor so I think it has been thought through / was always the intention. The idea is to take the combined width (on full-size screens) of the search results, sidebar, and gap in-between them, make that the max width of the page and center it. Also use the same max width for the contents of the nav bar at the top of the page, and center those contents within the nav bar. Seems fairly sensible?

PR #4178, opened 2016-12-07
Center search page content

seanh:
<del>Depends on https://github.com/hypothesis/h/pull/4175</del>

Fixes https://github.com/hypothesis/h/issues/4018

When the window is wider than about 1355px then most of the page contents get pushed all the way along to the left side of the window but the account and groups menus etc get pushed all the way along to the right.

Instead, centre the contents of both the nav bar and the main page, as in the designs.

Before:

screenshot from 2016-12-07 14-22-27

After:

screenshot from 2016-12-07 14-22-48

### DIFF
--- a/h/static/styles/partials-v2/_base.scss
+++ b/h/static/styles/partials-v2/_base.scss
@@ -54,10 +54,10 @@ $touch-target-size: 44px;
 
 // Misc
 // ----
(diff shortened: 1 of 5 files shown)


robertknight:
LGTM. Let's see what reaction this provokes ;)

PR #4184, opened 2016-12-07
Check annotation write permission from group's access flags

chdorner:
This is changing the way we check if the current user is allowed to create an annotation in the given group. The old code just checked if the user is a member (if `group:{pubid}` is in `request.effective_principals`). Now that we have group access flags we need to change this behaviour. At the same time the separation between the knowledge of groups in h and memex was always a bit undefined. Memex at least knew about the `group` string and had some knowledge about authorization. This should now change as well by introducing a new configuration function that will get the current request and a groupid, and this function can be implemented by h and passed into memex. The default implementation of that function is permissive, that way we can easily allow applications to use memex without having to support the concept of groups.

See the individual commit messages for more information.

### DIFF
--- a/h/groups/__init__.py
+++ b/h/groups/__init__.py
@@ -1,5 +1,9 @@
 # -*- coding: utf-8 -*-
 
+from h.groups.auth import annotation_group_write_permitted
+
 
 def includeme(config):
+    config.set_memex_group_write_permitted(annotation_group_write_permitted)
+
     config.register_service_factory('.services.groups_factory', name='groups')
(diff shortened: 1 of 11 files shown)


h/groups/__init__.py
```diff
@@ -1,5 +1,9 @@
 # -*- coding: utf-8 -*-
 
+from h.groups.auth import annotation_group_write_permitted
+
 
 def includeme(config):
+    config.set_memex_group_write_permitted(annotation_group_write_permitted)
```
chdorner: What is the best way to test this? Not having this call tested makes me a bit nervous since the default permission then is permissive.

nickstenning: I'd suggest testing the `includeme` with a stub configurator (as we do, for example, in `routes_test.py`).

chdorner:
Closing in favour of #4200 which is using the Pyramid ACL system.

PR #4190, opened 2016-12-07
Skip rendering annotation text when None

### LINKED ISSUE #4183: Annotation cards on activity pages show "None" if they have no text
### Steps to reproduce
1. See annotation at, for example: https://hypothes.is/users/judell?q=tag%3Aimages+tag%3Ahypothesisrelated

### Expected behaviour
Annotations without text should not display anything in the "text" area.

### Actual behaviour
Annotations have a literal "None" in the card.

chdorner:
Jinja seems to be rendering the string "None" instead of rendering
nothing.

Fixes #4183.

### DIFF
--- a/h/templates/includes/annotation_card.html.jinja2
+++ b/h/templates/includes/annotation_card.html.jinja2
@@ -50,7 +50,9 @@
     </section>
   {% endif %}
   <section class="annotation-card__text">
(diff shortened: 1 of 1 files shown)


nickstenning:
Perhaps we should fix this further up the stack, but it'll certainly do for now.

PR #4191, opened 2016-12-07
Send authenticated user ID in error reports from client

robertknight:
Provide the authenticated user ID in reports to Sentry from client code.

This allows us to associate JavaScript errors with the user who experienced them. In Sentry reports, this will show up like this:

sentry-logged-in-user

For logged-out users, the value appears as "None".

### DIFF
--- a/h/static/scripts/site.js
+++ b/h/static/scripts/site.js
@@ -5,7 +5,6 @@ const settings = require('./base/settings')(document);
 if (settings.raven) {
   const raven = require('./base/raven');
   raven.init(settings.raven);
-  raven.setUserInfo({userid: settings.raven.userid});
 }
 
 require('./polyfills');
--- a/h/static/scripts/site.js
+++ b/h/static/scripts/site.js
@@ -3,7 +3,9 @@
 // Configure error reporting
 const settings = require('./base/settings')(document);
 if (settings.raven) {
-  require('./base/raven').init(settings.raven);
+  const raven = require('./base/raven');
+  raven.init(settings.raven);
+  raven.setUserInfo({userid: settings.raven.userid});
 }
 
 require('./polyfills');
--- a/h/templates/layouts/base.html.jinja2
+++ b/h/templates/layouts/base.html.jinja2
@@ -66,7 +66,8 @@
         {
           "raven": {
             "dsn": "{{ request.sentry.get_public_dsn('https') }}",
-            "release": "{{ h_version }}"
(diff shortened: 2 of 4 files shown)


h/static/scripts/site.js
```diff
@@ -3,7 +3,9 @@
 // Configure error reporting
 const settings = require('./base/settings')(document);
 if (settings.raven) {
-  require('./base/raven').init(settings.raven);
+  const raven = require('./base/raven');
+  raven.init(settings.raven);
+  raven.setUserInfo({userid: settings.raven.userid});
```
nickstenning: I'm pretty sure you want to use the key `id`, not `userid`, as otherwise Sentry won't display it properly.

h/templates/layouts/base.html.jinja2
```diff
@@ -66,7 +66,8 @@
         {
           "raven": {
             "dsn": "{{ request.sentry.get_public_dsn('https') }}",
-            "release": "{{ h_version }}"
+            "release": "{{ h_version }}",
+            "userid": {{ request.authenticated_userid | to_json }}
```
nickstenning: It would seem more consistent to do `"{{ request.authenticated_userid }}"` here instead of using `to_json`.

h/static/scripts/site.js
nickstenning: Is there a reason this line wouldn't go in `base/raven` along with the usage of `settings.raven.release`?

nickstenning: Lastly, you should call `setUserInfo()` for logged-out users, not `setUserInfo({id: null})`.

h/templates/layouts/base.html.jinja2
nickstenning: Oh, sorry, this is because it might be `None`.

h/static/scripts/site.js
robertknight: > Is there a reason this line wouldn't go in base/raven along with the usage of settings.raven.release?

Not in this context. In the client setting the user info is separate from initializing Raven because the user's ID is not known when the client boots. That is not the case here however.

robertknight: > I'm pretty sure you want to use the key id, not userid, as otherwise Sentry won't display it properly

You're right. Sentry will still display the userid on reports if the key "userid" is used, but won't treat it as a unique identifier for the user.

nickstenning:
LGTM.

PR #4194, opened 2016-12-08
Fix inline form submission in IE

robertknight:
For `<form>` elements that have an empty "action" attribute, IE submits
the form to the correct location when using normal submission but
returns the wrong URL when reading the form's `action` property. This
URL was used when submitting the form via fetch().

One way to fix this would be to always specify an absolute URL for the
form in the deform template. I've opted to add a workaround in JS to
avoid the code breaking if it is ever used with forms not rendered by
deform.

### DIFF
--- a/h/static/scripts/util/submit-form.js
+++ b/h/static/scripts/util/submit-form.js
@@ -30,7 +30,7 @@ class FormSubmitError extends Error {
  * @param {HTMLFormElement} form
  */
 function formUrl(form) {
-  if (form.getAttribute('action').length > 0) {
+  if (form.getAttribute('action')) {
     return form.action;
   } else {
     // `form.action` returns an absolute URL created by resolving the URL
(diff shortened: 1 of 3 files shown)


h/static/scripts/util/submit-form.js
```diff
@@ -25,6 +25,24 @@ class FormSubmitError extends Error {
 }
 
 /**
+ * Return the URL which a form should be submitted to.
+ *
+ * @param {HTMLFormElement} form
+ */
+function formUrl(form) {
+  if (form.getAttribute('action').length > 0) {
```
nickstenning: Won't this throw a `TypeError` if the action attribute is missing?

h/templates/deform/form.jinja2
```diff
@@ -6,7 +6,7 @@
 {% set form_message_class = 'form-actions-message' %}
 {% endif %}
 <form id="{{ field.formid }}"
-      action="{{ field.action }}"
+      {% if field.action %} action="{{ field.action }}" {% endif %}
```
nickstenning: Could you fix the whitespace control here so this doesn't add a line if the action isn't set?

nickstenning:
LGTM.

PR #4195, opened 2016-12-08
Fix search bar in Internet Explorer and Edge

robertknight:
This fixes input in the search bar in Internet Explorer and Edge.

1. Internet Explorer needed a polyfill for `Element.prototype.remove` - triggered when you click on the delete button on a lozenge.

2. Microsoft Edge supports `<template>` elements (used to render a template for the lozenge on the server) but only supports `firstElementChild` on `Element` objects, not `DocumentFragment` objects as returned by `template.content`.

Fixes https://github.com/hypothesis/product-backlog/issues/75


seanh:
Works!

PR #4197, opened 2016-12-08
Use ActivityResults container object in templates

nickstenning:
This is a small refactoring which changes how the search results are passed into templates. This will make it easier to extract the rendering of the search results into independent partial templates or panels.

### DIFF
--- a/h/views/activity.py
+++ b/h/views/activity.py
@@ -50,7 +50,7 @@ class SearchController(object):
             page_size = PAGE_SIZE
 
         # Fetch results.
-        result = query.execute(self.request, q, page_size=page_size)
+        results = query.execute(self.request, q, page_size=page_size)
 
         groups_suggestions = []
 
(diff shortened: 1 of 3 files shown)


h/views/activity.py
```diff
@@ -75,14 +75,12 @@ def user_link(userid):
...
-            'aggregations': result.aggregations,
+            'search_results': results,
             'groups_suggestions': groups_suggestions,
-            'page': paginate(self.request, result.total, page_size=page_size),
+            'page': paginate(self.request, results.total, page_size=page_size),
             'pretty_link': pretty_link,
             'q': self.request.params.get('q', ''),
             'tag_link': tag_link,
-            'timeframes': result.timeframes,
-            'total': result.total,
```
chdorner: I can't seem to add comments to lines that are not in the initial set of the diff.

Line 408 in the search template still uses the old `total`, that needs to change to `search_results.total` as well

chdorner:
Pushed the fix. Waiting for Travis and will then merge.

robertknight:
Rebased. Just waiting on Travis.

h/views/activity.py
nickstenning: Thanks for spotting this one!

chdorner:
Oh, I must have forgotten to merge it yesterday. Thanks for taking over @robertknight

PR #4198, opened 2016-12-08
Redirect stream links for tags to the search page if the feature flag…

sheetaluk:
… is enabled.

https://github.com/hypothesis/product-backlog/issues/63

### DIFF
--- a/h/views/main.py
+++ b/h/views/main.py
@@ -57,6 +57,15 @@ def robots(context, request):
 
 @view_config(route_name='stream')
 def stream(context, request):
+    if request.feature('search_page'):
+        q = request.params.get('q', '').split(':', 1)
+        if len(q) >= 2 and q[0] == 'tag':
+            tag = q[1]
+            if ' ' in tag:
+                tag = '"' + tag + '"'
+            query = {'q': 'tag:{}'.format(tag)}
+            location = request.route_url('activity.search', _query=query)
+            raise httpexceptions.HTTPFound(location=location)
     atom = request.route_url('stream_atom')
     rss = request.route_url('stream_rss')
     return render_app(request, {
--- a/tests/h/views/main_test.py
+++ b/tests/h/views/main_test.py
@@ -39,7 +39,7 @@ def test_og_no_document(render_app_html, pyramid_request):
     assert any(test(d) for d in kwargs['extra']['meta_attrs'])
 
 
-@pytest.mark.usefixtures('routes')
+@pytest.mark.usefixtures('render_app', 'routes')
 class TestStreamUserRedirect(object):
 
     def test_it_redirects_to_stream(self, pyramid_request):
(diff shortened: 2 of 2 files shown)


h/views/main.py
```diff
@@ -57,14 +57,24 @@ def robots(context, request):
...
-    atom = request.route_url('stream_atom')
-    rss = request.route_url('stream_rss')
-    return render_app(request, {
-        'link_tags': [
-            {'rel': 'alternate', 'href': atom, 'type': 'application/atom+xml'},
-            {'rel': 'alternate', 'href': rss, 'type': 'application/rss+xml'},
-        ]
-    })
+    if request.feature('search_page'):
+        q = request.params['q'].split(':')
```
chdorner: I think this will break when there is no "q" parameter

h/views/main.py
```diff
@@ -57,14 +57,24 @@ def robots(context, request):
...
+        if q[0] == 'tag':
```
chdorner: Similar to above, if the "q" parameter exists but without a colon then this will break.

chdorner: (And the next line)

h/views/main.py
```diff
@@ -57,14 +57,24 @@ def robots(context, request):
...
+            tag = q[1]
+            if tag.find(' ') > -1:
+                tag = "'" + tag + "'"
+            query = {'q': 'tag:{}'.format(tag)}
+            location = request.route_url('activity.search', _query=query)
+            raise httpexceptions.HTTPFound(location=location)
+    else:
```
chdorner: This will fail when the feature flag is on but the query string doesn't start with "tag:", in that case it will not respond with what we expect it to respond. The easiest is to just remove the `else`, because the next block should happen as long as the previous block didn't raise an error

h/views/main.py
```diff
@@ -57,14 +57,24 @@ def robots(context, request):
...
-            {'rel': 'alternate', 'href': rss, 'type': 'application/rss+xml'},
-        ]
-    })
+    if request.feature('search_page'):
+        q = request.params['q'].split(':')
+        if q[0] == 'tag':
+            tag = q[1]
+            if tag.find(' ') > -1:
+                tag = "'" + tag + "'"
+            query = {'q': 'tag:{}'.format(tag)}
```
chdorner: Nice making sure that the different behaviour with whitespace in the stream and activity pages is cared for!

h/views/main.py
```diff
@@ -57,6 +57,15 @@ def robots(context, request):
 
 @view_config(route_name='stream')
 def stream(context, request):
+    if request.feature('search_page'):
+        q = request.params.get('q', '').split(':')
+        if len(q) == 2 and q[0] == 'tag':
+            tag = q[1]
+            if ' ' in tag:
+                tag = "'" + tag + "'"
```
chdorner: This works, although I think my personal preference is to show users double quotes, this renders `tag:'foo bar'`.
What do you think?

h/views/main.py
```diff
@@ -57,6 +57,15 @@ def robots(context, request):
 
 @view_config(route_name='stream')
 def stream(context, request):
+    if request.feature('search_page'):
+        q = request.params.get('q', '').split(':')
+        if len(q) == 2 and q[0] == 'tag':
```
chdorner: I just noticed that this isn't bullet-proof. I clicked on a tag that is "hypothesis:education", which then obviously didn't redirect because the length should be 3. Maybe we can change this to be `len(q) >= 2`?

tests/h/views/main_test.py
```diff
@@ -39,6 +39,7 @@ def test_og_no_document(render_app_html, pyramid_request):
     assert any(test(d) for d in kwargs['extra']['meta_attrs'])
 
 
+@pytest.mark.usefixtures('render_app')
```
chdorner: This should just be one-line: `@pytest.mark.usefixtures('render_app', 'routes')`

tests/h/views/main_test.py
```diff
@@ -69,8 +129,16 @@ def test_it_extracts_username_from_account_id(self, pyramid_request):
...
     def routes(self, pyramid_config):
+        pyramid_config.add_route('activity.search', '/search')
         pyramid_config.add_route('activity.user_search', '/user/{username}')
         pyramid_config.add_route('stream', '/stream')
+        pyramid_config.add_route('stream_atom', '/stream.atom')
+        pyramid_config.add_route('stream_rss', '/stream.rss')
+
+
+@pytest.fixture
+def render_app(patch):
```
chdorner: You can put this fixture inside the `TestStreamUserRedirect` class since it isn't used by any test outside of that class.

PR #4199, opened 2016-12-08
Make JSON data payloads embedded in search results page HTML-safe

robertknight:
Escape '<', '>' and '&' when serializing values to JSON that is embedded in HTML.

Fixes hypothesis/product-backlog#82


chdorner:
Marking as WIP because of [this](https://hypothes-is.slack.com/archives/frontend/p1481214347000225) Slack discussion.

nickstenning:
LGTM.

PR #4200, opened 2016-12-08
Upgrade group ACL: Write

chdorner:
This PR will change all group related authorization to use the new access flags.

This is a work in progress.


h/models/group.py
```diff
@@ -85,11 +84,15 @@ class Group(Base, mixins.Timestamps):
...
-    def __init__(self, name, authority, creator, description=None):
+    def __init__(self, name, authority, creator, description=None,
+                 joinable_by=None, readable_by=None, writeable_by=None):
         self.name = name
         self.authority = authority
         self.description = description
         self.creator = creator
+        self.joinable_by = joinable_by
+        self.readable_by = readable_by
+        self.writeable_by = writeable_by
```
nickstenning: Yeah, this isn't your fault, but this is getting a bit silly now. We should get rid of the explicit args /kwargs here and use `super`. Perhaps a job for a later commit, but ORM objects have constructors already. Not really sure why this one got overridden.

chdorner: I know, this can be fixed later :) I guess the reason is that the creator will automatically added to the members, plus not using `super` :)

src/memex/__init__.py
```diff
@@ -3,6 +3,12 @@
 __all__ = ('__version__',)
 __version__ = '0.39.0+dev'
 
+GROUPFINDER_KEY = 'memex.groupfinder'
+
+
+def set_groupfinder(config, func):
+    config.registry[GROUPFINDER_KEY] = func
```
nickstenning: Looks good, but as a matter of convention Pyramid configuration directives should usually accept "dotted-string" arguments as it makes it much easier to avoid import cycles. There's a `maybe_dotted` method on the configurator instance to help with this.

src/memex/storage.py
```diff
@@ -11,6 +11,7 @@
 
 from pyramid import i18n
 
+from memex import GROUPFINDER_KEY
```
nickstenning: This is ok, but I think it would be better to encapsulate the registry storage within a module dedicated to that task, say `memex.groups`, and provide a helper function like `memex.groups.find(...)` which pulls the relevant thing out of the registry.

src/memex/storage.py
```diff
@@ -108,14 +109,16 @@ def create_annotation(request, data):
...
     # The user must have permission to create an annotation in the group
-    # they've asked to create one in.
-    if data['groupid'] != '__world__':
-        group_principal = 'group:{}'.format(data['groupid'])
-        if group_principal not in request.effective_principals:
+    # they've asked to create one in. If the application didn't configure
+    # a groupfinder we will allow writing this annotation without any
+    # further checks.
+    groupfinder = request.registry.get(GROUPFINDER_KEY)
+    if groupfinder is not None:
```
nickstenning: Again, this is ok, and I'd be happy to merge it, but the beauty of using a context object is that it doesn't matter what the object is, so you can set a default groupfinder which just returns an instance of `PermissiveGroupContext`:

```python
class PermissiveGroupContext(object):
    __acl__ = [(security.Allow, security.Everyone, security.ALL_PERMISSIONS)]
```

or something.

nickstenning: Note also that the way pyramid's configuration system works, you can actually *call* `memex_set_groupfinder` in `memex/groups.py`'s includeme in order to set the default. Because `h` includes this module (indirectly) it will then be able to override that call with its own call to `memex_set_groupfinder`.

tests/memex/__init___test.py
```diff
@@ -0,0 +1,14 @@
+# -*- coding: utf-8 -*-
+
+from __future__ import unicode_literals
+
+import mock
+
+import memex
+
+
+def test_set_groupfinder(pyramid_config):
```
nickstenning: Luckily (because files called `__init___test.py` give me the heebie-jeebies) I don't think this test is really necessary. See my comment above about a `memex/groups.py` and its `includeme`.

h/models/group.py
chdorner: I just fixed it in a commit previous to this, so this commit will just contain the changes to the group factory.

src/memex/__init__.py
chdorner: Good to know! Fixed.

tests/memex/__init___test.py
chdorner: Sure, the reason why I did that is that I'm a bit nervous because of memex' default-permissive behaviour when the groupfinder isn't set. But I'm just writing a functional test for that and then I will remove these.

nickstenning: That's fair enough. I think that concern is reason enough to make the default context something like this:

```
class DefaultGroupContext(object):
    def __init__(self, name):
        self.name = name

    def __acl__(self):
        if self.name == '__world__':
            return [(security.Allow, security.Authenticated, 'write'),
                    (security.Allow, security.Everyone, 'read'),
                    security.DENY_ALL]
        return [security.DENY_ALL]

File: src/memex/storage.py (hunk 3 above)
Comment (chdorner) [inline src/memex/storage.py]: Done

File: src/memex/storage.py (hunk 4 above)
Comment (chdorner) [inline src/memex/storage.py]: Done

File: tests/memex/__init___test.py (hunk 5 above)
Comment (chdorner) [inline tests/memex/__init___test.py]: Done

**Comment by chdorner** [comment]
All comments should be addressed. This will still need manual testing of all occurrences where we need to check group permissions that are not "write".

**Comment by chdorner** [comment]
I've been testing that the join/search page always gets rendered correctly:

The join page gets rendered when the request does not have the permission to read.
Private groups with the `hypothes.is` authority (`readable_by=members`):
* User has `hypothes.is` authority, is member: read permission ✔️ ➡️ render search page
* User has `hypothes.is` authority, is not member: read permission ❌ ➡️ render join page
* User has different authority (doesn't work, but anyway): read permission ❌ ➡️ render join page (wrong, but doesn't happen, because non-hypothes.is users can't login on the website)

Publisher groups with the `example.com` authority (`readable_by=world`):
* User has `hypothes.is` authority: read permission ✔️ ➡️ render search page (no annotations yet, because read isn't implemented with the new access flags)
* User has `example.com` authority: doesn't work, can't login on the website

**Comment by chdorner** [comment]
There shouldn't be any changes to the `read` permission since we still set the principal to `group:{pubid}` only when the user is a member.

`read` permission for groups with readable_by other than members still needs work.

Other people should test as well, but to me this means that this PR will not mix up any of the other group permission.

File: src/memex/storage.py [hunk 6]
```diff
@@ -108,14 +109,14 @@ def create_annotation(request, data):
...
-        group_principal = 'group:{}'.format(data['groupid'])
-        if group_principal not in request.effective_principals:
-            raise schemas.ValidationError('group: ' +
-                                          _('You may not create annotations '
-                                            'in groups you are not a member '
-                                            'of!'))
+    # they've asked to create one in. If the application didn't configure
+    # a groupfinder we will allow writing this annotation without any
+    # further checks.
+    group = groups.find(request, data['groupid'])
```
nickstenning: Do we need to handle the situation in which the groupid is invalid here?

tests/h/models/groups_test.py
```diff
@@ -155,15 +147,57 @@ def test_documents_when_group_has_no_documents(group):
...
 
 
-def test_acl(group, factories):
-    group.pubid = 'testing-pubid'
-    group.creator = factories.User(username='luke', authority='foobar.org')
+class TestGroupACL(object):
+    def test_authority_joinable(self, group):
+        group.joinable_by = JoinableBy.authority
+
+        assert (security.Allow, 'authority:example.com', 'join') in group.__acl__()
```
nickstenning: I think these tests should probably be implemented using an instance of `ACLAuthorizationPolicy`, and doing `assert policy.permits(group, ['someprincipal'], 'join')` because asserting that the (allow/deny, principal, permission) triple is in the ACL doesn't really mean much if the first element of the ACL is `security.DENY_ALL`.

tests/memex/groups_test.py
```diff
@@ -0,0 +1,85 @@
...
+from __future__ import unicode_literals
+
+import mock
+from pyramid import security
+import pytest
+
+from memex import groups
+
+
+class TestPermissiveGroupContext(object):
```
nickstenning: This name needs updating now we changed it to `DefaultGroupContext`.

tests/memex/storage_test.py
```diff
@@ -218,6 +258,14 @@ def test_it_does_not_crash_if_no_text_or_tags(self, pyramid_request):
 
         storage.create_annotation(pyramid_request, data)
 
+    @pytest.fixture
+    def default_groupfinder(self, pyramid_config):
+        pyramid_config.registry[groups.GROUPFINDER_KEY] = groups.default_groupfinder
```
nickstenning: Hmm, yeah, this is a bit odd. Redoing the contents of `memex.groups` `includeme` in the tests is probably something we'd like to avoid if possible, but I can't think of a simpler way right now.

src/memex/storage.py
chdorner: The groupfinder specifically returns `None`, and `has_permission` on `None` will return `False`. So an API client will get a 400 error: `{"status": "failure", "reason": "group: You may not create annotations in the specified group!"}`

That's the correct behaviour in my opinion, no?

chdorner: Ah, now I can reproduce. I tried with a third-party account and it returned that error I posted. Looking into it now

chdorner: As discussed in Slack, this is due to [this](https://github.com/hypothesis/h/blob/master/h/resources.py#L15) when the authenticated user is an admin, and that Pyramid uses the request context when the passed-in one is `None`.
The easiest fix for now is to check for `None` as well as `has_permission`.

Fixed

tests/h/models/groups_test.py
chdorner: Ah yes, much safer. I was loosely basing these tests on what we do for the annotation model. But they always check the full list of permissions, which doesn't have the same issue.
Fixed

nickstenning:
LGTM.

PR #4201, opened 2016-12-08
Remove a couple of unneeded uses of the Jinja2 "safe" filter

nickstenning:
This is a scary filter to see in a template. In general, values should be clearly marked as safe by the code that handles them *before* they get to the templates.

This commit removes the use of the "safe" Jinja2 filter from a couple of places, and adds some comments explaining when and where user input gets escaped.

### DIFF
--- a/h/presenters.py
+++ b/h/presenters.py
@@ -75,9 +75,16 @@ class AnnotationHTMLPresenter(object):
 
     @property
     def text_rendered(self):
-        """The body text of this annotation."""
+        """
+        The body text of this annotation.
 
-        return self.annotation.text_rendered
+        This return value of this field is marked safe because it is rendered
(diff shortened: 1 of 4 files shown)


h/presenters.py
```diff
@@ -75,9 +75,14 @@ def uri(self):
...
-        """The body text of this annotation."""
+        """
+        The body text of this annotation.
 
-        return self.annotation.text_rendered
+        This return value of this field is marked safe because it is rendered
+        to HTML on write by :py:func:`memex.markdown.render`, which must take
+        care of all necessary escaping.
+        """
+        return jinja2.Markup(self.annotation.text_rendered)
```
nickstenning: ARGH.

nickstenning: Fixed.

robertknight:
LGTM.

Issue #4074, opened 2016-11-15
Navbar logo stretched in Internet Explorer


robertknight:
The navbar logo is vertically stretched in Internet Explorer (tested with IE 11):

![ie-navbar-logo](https://cloud.githubusercontent.com/assets/2458/20298637/e4dcaf9c-ab0e-11e6-9bb4-280b1c4c4f8a.png)

PR #4203, opened 2016-12-09
Fix vertical stretch of logo SVG in IE

seanh:
In IE the SVG image of the h logo was being vertically stretched to be
as tall as its containing element. Add an align-items CSS rule to
explicitly tell it not to do this.

Fixes https://github.com/hypothesis/h/issues/4074


robertknight:
LGTM

PR #4204, opened 2016-12-09
Remove unwanted search bar margins on mobile

seanh:
On mobile devices the search bar has 10px left and right margins which
means that it has 10px more whitespace to the left and right of it than
any of the other elements on the screen, creating a ragged look.

Remove these margins so that the search bar's left and right borders are
left and right aligned with the rest of the elements on the screen.

Before:

before

After:

after


robertknight:
LGTM

Issue #3808, opened 2016-09-01
Character counter limit not visible in IE 11


robertknight:
### Steps to reproduce
1. In IE 11, login and go to the `/account/profile` page
2. Look at the bottom right corner of the "Description" field
### Expected behaviour

There should be a counter displaying the number of remaining characters
### Actual behaviour

No character counter is visible
### Additional details

The CSS relies on `display: initial` to show the counter once the enhancement loads. CSS initial values are [not supported in IE 10/11](http://caniuse.com/#feat=css-initial-value)

PR #4205, opened 2016-12-09
Fix character limit counter in IE

seanh:
The character limit counter wasn't appearing in IE because IE doesn't
support display: initial. Instead hardcode the value (display: block)
that display: initial actually results in for this element in other
browsers.

Fixes https://github.com/hypothesis/h/issues/3808


robertknight:
LGTM

Issue #4206, opened 2016-12-09
Form field error messages not visible in IE


seanh:
In IE, enter an invalid value into a form and submit it. Expected: error message. Result: no error message.

IE 11:

![screenshot from 2016-12-09 13-48-53](https://cloud.githubusercontent.com/assets/22498/21051071/699914e0-be16-11e6-9f2d-410bf0af6128.png)

Chrome (how it should look):

![screenshot from 2016-12-09 13-49-10](https://cloud.githubusercontent.com/assets/22498/21051076/7242c078-be16-11e6-87f8-921f4122e04e.png)

(The incorrect wrapping of the user's text in these screenshots is a [separate issue](https://github.com/hypothesis/h/issues/3777), as is the error message and character counter overlapping the scrollbar. Also the circle icon in the bottom-right in the second screenshot is <https://grammarly.com/>, sorry.)

PR #4207, opened 2016-12-09
Fix form field error messages in IE

seanh:
Form field error messages where not being shown in IE because it doesn't
support display: initial. In other browsers display: initial actually
results in display: list-item for these error messages, so hardcode
display: list-item instead so that it works in IE.

Fixes https://github.com/hypothesis/h/issues/4206


robertknight:
LGTM.

PR #4208, opened 2016-12-09
Use Node 6.7 on Travis

chdorner:
We are building our production docker images with Node 6.7, so we should also use this version on Travis.


chdorner:
Well, this isn't the best day for this PR while Travis isn't able to download the node source code :)

robertknight:
LGTM.

PR #4209, opened 2016-12-09
Move panels to h.panels package

nickstenning:
This commit tidies up our use of "panels". While similar to views, these are different things, so I've moved them to their own package.

In addition, we should probably only be using panels in situations where we need to do additional computation beyond simply rendering a template, so I've removed the "group_invite" panel and just inlined it back into the one place it was being used.

(This PR is in preparation for moving some of the activity pages code out into reusable panels with a view to simplifying `search.html.jinja2` and improving the performance of activity pages in general.)


h/templates/activity/search.html.jinja2
```diff
@@ -251,7 +251,19 @@
...
+    </h3>
+    Sharing the link below lets people join this group:
+    <div class="group-invite__container js-copy-button">
+      <input class="group-invite__input js-select-onfocus"
+             data-ref="input"
+             value="{{ group.url }}">
+      <button class="group-invite__clipboard-button"
+              data-ref="button"
+              title="Copy to clipboard">
+      {{ svg_icon('copy_to_clipboard') }}
```
robertknight: Closing `</button>` tag is missing

robertknight: Ah, I see this problem was present in the place where this was moved from. Still, now is a good time to fix it.

robertknight:
LGTM

h/templates/activity/search.html.jinja2
nickstenning: https://github.com/hypothesis/h/commit/3d5143973ef1bc9922e8cc01e0f430bb6bbe23fb

PR #4210, opened 2016-12-09
Move responsibility for rendering suggestions data to navbar panel

nickstenning:
This commit moves responsibility for rendering the JSON suggestions data from the activity pages views themselves to the "navbar" panel, which is the component that actually uses this data.

This fixes the issue of the groups suggestions not being available on pages which show the search box but do not have an active search (such as the account pages), but more generally it's a step towards including all "search state" context in a single "search" object which is passed to reusable panels such as this one.

~~~_**N.B.** This PR depends on #4209 and will need a rebase after that's merged._~~~


nickstenning:
Marked as WIP for dependency on prior PR.

robertknight:
LGTM

PR #4211, opened 2016-12-12
Add tooltip to only me icon

seanh:


h/templates/includes/annotation_card.html.jinja2
```diff
@@ -36,10 +36,12 @@
       {% endif %}
       {% if not presented_annotation.annotation.shared %}
         <span class="annotation-card__is-private">
-          {{ svg_icon('lock', 'annotation-card__is-private-icon') }}
-          {% if not presented_annotation.group %}
-            Only me
-          {% endif %}
+          <span title="{% trans %}This annotation is visible only to you.{% endtrans %}">
```
robertknight: I wondered initially if this should have a period or not. It is however a full sentence, unlike most tooltips which are fragments, so I think it makes sense that it does.

seanh:
Done

robertknight:
LGTM

PR #4212, opened 2016-12-12
Fix 'How to get started' layout in Firefox

robertknight:
This makes the layout of the "How to get started" view seen by new users after logging in for the first time closer to the mocks at https://cloud.githubusercontent.com/assets/22498/20572409/3f1e6a50-b1a3-11e6-867d-fbda27ff4838.png . Note that the link styling still does not match the mock (note the spacing at the bottom). I'll address that in a follow-up PR.

 * Fix list item contents appearing underneath item numberings by
   removing the paragraph tag used to create spacing between items and
   instead setting line-height on list items.

 * Fix difference in whitespace at the start of list items between
   Firefox and Chrome by removing whitespace around Jinja tags.

 * Adjust padding between list item marker and content.
   See http://stackoverflow.com/a/26336092/434243

----

**Before:**

zero-state-firefox

**After:**

zero-state-firefox-after

### DIFF
--- a/h/templates/activity/search.html.jinja2
+++ b/h/templates/activity/search.html.jinja2
@@ -374,38 +374,38 @@
         {% if zero_message == '__SHOW_GETTING_STARTED__' %}
           <div>
             <h2 class="search-result-zero__title">
-              {% trans %}How to get started{% endtrans %}
+              {%- trans %}How to get started{% endtrans -%}
             </h2>
             <ol class="search-result-zero__list">
-              <li><p>
-                <a href="https://chrome.google.com/webstore/detail/hypothesis-web-pdf-annota/bjfhmglciegochdpefhhlphglcehbmek">
-                  {% trans %}Install our Chrome extension{% endtrans %}
+              <li class="search-result-zero__list-item"><!--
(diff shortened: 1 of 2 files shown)


h/templates/activity/search.html.jinja2
```diff
@@ -374,38 +374,38 @@
...
             <h2 class="search-result-zero__title">
-              {% trans %}How to get started{% endtrans %}
+              {%- trans %}How to get started{% endtrans -%}
             </h2>
             <ol class="search-result-zero__list">
-              <li><p>
-                <a href="https://chrome.google.com/webstore/detail/hypothesis-web-pdf-annota/bjfhmglciegochdpefhhlphglcehbmek">
-                  {% trans %}Install our Chrome extension{% endtrans %}
+              <li class="search-result-zero__list-item"><!--
+                !--><a href="https://chrome.google.com/webstore/detail/hypothesis-web-pdf-annota/bjfhmglciegochdpefhhlphglcehbmek">
```
seanh: Should this HTML comment be here?

seanh:
Looks great, I quickly tested it in Chrome, Firefox, Edge, IE and mobile Safari. Just have the one question about an empty HTML comment

PR #4214, opened 2016-12-12
Deduplicate link styling

robertknight:
Use variants of the `link` component style for all text links in search
result pages instead of defining a different class for each one.

The only visual change as a result of this commit is that links with
solid underlines now display the underline underneath the decenders, as
in this mock:
https://cloud.githubusercontent.com/assets/22498/20572409/3f1e6a50-b1a3-11e6-867d-fbda27ff4838.png #

**Before**

zero-state-firefox-after

**After**

underline-below-descender

### DIFF
--- a/h/static/styles/partials-v2/_link.scss
+++ b/h/static/styles/partials-v2/_link.scss
@@ -1,19 +1,39 @@
 // Text links
 // ----------
-// Styling for text links. Note that these are not used for links to pages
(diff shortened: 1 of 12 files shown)


nickstenning:
Looks mostly good to me, but you might want to address line-height here:

screen shot 2016-12-13 at 10 46 53

PR #4215, opened 2016-12-13
Make icon logo reusable

seanh:
[The upcoming footer](https://github.com/hypothesis/product-backlog/issues/87) has a red h logo on a black background. In order to make the current h logo svg reusable for this case we need to:

1. Be able to change the colour of the image using CSS
2. Be able to use the image on different background colours (currently the "h" part is transparent and shows the colour of the site behind it)

This PR changes the SVG file to make it reusable this way and changes navbar.html.jinja2 to use the image in the new way.

A following PR will reuse the image file in the new footer.

See the individual commit messages for more details.


h/static/images/icons/logo.svg
```diff
@@ -4,7 +4,9 @@
     <title>Rectangle 2 Copy 17</title>
     <desc>Created with Sketch.</desc>
     <defs></defs>
+    <rect fill="#ffffff" stroke="none" width="17.14407" height="16.046612"
+          x="3.8855932" y="3.9449153" />
```
seanh: See commit message. I put an opaque white rect behind the "h"-shaped hole so that the "h" always appears white regardless of which background colour the svg file is rendered against.

h/static/images/icons/logo.svg
```diff
@@ -4,7 +4,9 @@
...
+        <path d="M0,2.00494659 C0,0.897645164 0.897026226,0 2.00494659,0 L21.9950534,0 C23.1023548,0 24,0.897026226 24,2.00494659 L24,21.9950534 C24,23.1023548 23.1029738,24 21.9950534,24 L2.00494659,24 C0.897645164,24 0,23.1029738 0,21.9950534 L0,2.00494659 Z M9,24 L12,28 L15,24 L9,24 Z M7.00811294,4 L4,4 L4,20 L7.00811294,20 L7.00811294,15.0028975 C7.00811294,12.004636 8.16824717,12.0097227 9,12 C10,12.0072451 11.0189302,12.0606714 11.0189302,14.003477 L11.0189302,20 L14.0270431,20 L14.0270431,13.1087862 C14.0270433,10 12,9.00309038 10,9.00309064 C8.01081726,9.00309091 8,9.00309086 7.00811294,11.0019317 L7.00811294,4 Z M19,19.9869002 C20.1045695,19.9869002 21,19.0944022 21,17.9934501 C21,16.892498 20.1045695,16 19,16 C17.8954305,16 17,16.892498 17,17.9934501 C17,19.0944022 17.8954305,19.9869002 19,19.9869002 Z" id="Rectangle-2-Copy-17" fill="currentColor"></path>
```
seanh: The only change on this line is `fill="currentColor"`

h/templates/panels/navbar.html.jinja2
```diff
@@ -30,8 +29,7 @@
   </template>
   <div class="nav-bar__content">
     <a href="/" title="Hypothesis homepage" class="nav-bar__logo-container"><!--
-      !--><img alt="Hypothesis logo" class="nav-bar__logo"
-               src="{{ asset_url('images/logo.svg') }}"></a>
+      !-->{{ svg_icon('hypothesis-logo', 'nav-bar__logo') }}</a>
```
seanh: The "Hypothesis logo" alt text is gone. The `<a>` that this `` was on already has a `title`.

h/static/images/icons/logo.svg
seanh: Alternatively I guess you could put a white h on top of the speech bubble shape instead of having an h-shaped hole in the speech bubble. That would be harder to do from what we have now but maybe I should do it...

h/static/images/icons/logo.svg
robertknight: This is fine, though an alternative would be to put the color on the containing `<g>` element's fill.

h/templates/panels/navbar.html.jinja2
seanh: Oops. fixed

robertknight:
**Warning: Rabbit-hole ahead!**

In Chrome the rendering looks the same before and after this PR. In Firefox on my Linux system the SVG rendering unfortunately became noticeably blurry afterwards.

**master**

firefox-before

**This PR**

firefox-after

The reason this is happening is because the left/top edges of the SVG are not at pixel boundaries but rather at half-pixel boundaries. For the logo at the top, this happens because:

- The navbar is centered and has a width that is an odd number of pixels. Hence `margin-left: auto` calculates the left edge of the navbar to be at a half-pixel position.
- The height of the navbar content is an odd number of pixels, and the logo is vertically centered.

It looks to me as though Chrome is snapping the SVG to a pixel boundary when rendering, because it renders sharply despite this. Neither Firefox nor Edge do this however.

I tried a few methods to fix this:

- Setting the [shape-rendering](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/shape-rendering) attribute on the `<svg>` - no dice.
- Shifting the image by setting `position:relative;left:0.5px;top:0.5px;` in order to align the left/top edge to a pixel boundary. This resulted in sharp rendering in Firefox and didn't regress the rendering in Chrome.
- The same transformation to the image itself in Inkscape. This resulted in a sharp image in Firefox and a blurry image in Chrome.
- Applying relative positioning to the element in JS to ensure that the element's top/left edge on screen aligned to a pixel boundary. This gave the best results in Chrome, Firefox, Edge and IE.

robertknight:
Quick update from Slack discussion. Nick found this Bugzilla item - https://bugzilla.mozilla.org/show_bug.cgi?id=608812

Also, it turns out that adding `transform: translateZ(0)` appears to resolve the issue in Firefox, though I'm not clear on why that is yet:

```diff
diff --git a/h/jinja_extensions.py b/h/jinja_extensions.py
index 928f38c..df81239 100644
--- a/h/jinja_extensions.py
+++ b/h/jinja_extensions.py
@@ -103,6 +103,12 @@ def svg_icon(loader, name, css_class=''):
     if css_class:
         root.set('class', css_class)
 
+    # Trigger snapping of the <svg>'s left and top edges to pixel boundaries
+    # in Firefox.
+    # See also https://bugzilla.mozilla.org/show_bug.cgi?id=608812 and
+    # https://github.com/hypothesis/h/pull/4215#issuecomment-267012849
+    root.set('style', 'transform: translateZ(0px)')
+
     # If the SVG has its own title, ignore it in favor of the title attribute
     # of the <svg> or its containing element, which is usually a link.
     title_el = root.find('{{{}}}title'.format(SVG_NAMESPACE_URI))
```

robertknight:
LGTM. @seanh - I subsequently discovered that _any_ active transform [triggers pixel snapping](https://bugzilla.mozilla.org/show_bug.cgi?id=739176#c4) in Firefox, so `translateX` and `translateY` also work and I _think_ that avoids the issue I was concerned about with creating new layers. I'm going to merge this now and we can revise in future in light of that.

Issue #4168, opened 2016-12-05
Annotation count is shown twice on user and group search pages


seanh:
![screenshot from 2016-12-05 14-20-21](https://cloud.githubusercontent.com/assets/22498/20889227/19524cfa-bafa-11e6-9730-428332229e6a.png)

Note that this is the same number twice, it's not (currently) two different numbers, see https://github.com/hypothesis/h/issues/4066

dwhly:
From a discussion in slack, change the count of search results to read "nn Matching Annotations".  Change the count within the user or group sidepanel to always show the total annotations that the user or group has made.

seanh:
I don't think the count of _all_ the user's or group's annotations is available to the templates yet (the count shown in the sidebar is just the matching ones), so the Python code'll need to add those counts to the template data, it'll have to get them from the DB.

chdorner:
Just a quick note: If we need to do additional queries we need to profile to see how long they take and how much time they add to the overall request processing.

seanh:
@chdorner Would a feature flag for this change work? Then we can turn it on and off and test in production

chdorner:
I would test it out locally first, also running `EXPLAIN ANALYZE ...` and making sure we have all the necessary indices set. If it doesn't add any significant time we should just get it out without a feature flag.
If it does make the request too slow we have to decide how important that count is to us and comparing it to how much work it is to build a caching layer into this.

ajpeddakotla:
2 points

ajpeddakotla:
I'm reopening this because there are some edge cases and some general confusion from a user perspective that we should clear up. @sheetaluk and I had a convo about this and here's what we came up with (also included this in the PR): 

______
So after talking @sheetaluk, there needs to be a finer grained approach to counts of annotations that we are showing in both the sidebar and the search results. 

So if we breakdown the different types of annotations that can be viewed, and who they can be viewed by, it looks something like this (for now - this will change with publisher groups). 

Empty |Private | Shared | Public
------ | ------ | ------ | ------
Me | Y | Y | Y
Person I am in one or more groups with | N | Y (but only in groups that I am a part of) | Y
Stranger | N | N | Y

Right now, what's been implemented - we're only showing public annotations which are not private or shared. 

What we'd like to do, because showing the finer detail helps better inform the user of the types of annotations they are making is to: 

**1. In the sidebar:** 
Breakdown the count to include (for users viewing their own profile):
Shared annotations: 
Private annotations: 
Public annotations: 

[...]

PR #4216, opened 2016-12-13
Show total shared annotations in the group and user side bar.

sheetaluk:
Fixes: https://github.com/hypothesis/h/issues/4168

### DIFF
--- a/h/accounts/services.py
+++ b/h/accounts/services.py
@@ -9,7 +9,7 @@ import sqlalchemy
 from h import mailer
 from h._compat import text_type
 from h.emails import signup
-from h.models import Activation, Subscriptions, User
+from h.models import Activation, Annotation, Subscriptions, User
 from h import util
 
 
@@ -114,6 +114,16 @@ class UserService(object):
             return user
 
         return None
+
+    def annotation_count(self, userid):
+        """
+        Return the count of shared annotations for this user.
+        """
+
+        return (
+            self.session.query(Annotation)
+            .filter_by(userid=userid, groupid='__world__',  shared=True)
+            .count())
 
 
 class UserSignupService(object):
--- a/h/groups/services.py
+++ b/h/groups/services.py
@@ -3,7 +3,7 @@
 from functools import partial
 
 from h import session
-from h.models import Group
+from h.models import Annotation, Group
 from h.models.group import JoinableBy, ReadableBy, WriteableBy
 
 GROUP_ACCESS_FLAGS = {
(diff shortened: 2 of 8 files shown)


nickstenning:
I see @chdorner has already been tagged as a reviewer so I'll let him do the honours, but I'll just note that we should probably wrap the count functions up in a service, and not as properties on the User and Group models -- as per [ADR 2](https://github.com/hypothesis/h/blob/master/docs/arch/adr-002.rst) (use services for business logic which touches multiple models, in order to avoid circular dependencies, as well as general model bloat).

tests/h/models/user_test.py
```diff
@@ -289,6 +289,19 @@ def test_it_filters_by_authority(self, db_session, users):
...
 
+    def test_annotation_count_returns_count_of_shared_annotations_for_user(self, db_session, users):
+        user = users['meredith']
+
+        for i in range(50):
+            annotation(db_session,
+                       document(db_session,
+                                'http://example.com/document_' + str(i)),
+                       user.userid,
+                       shared=True)
```
chdorner: I think you can just use the factories here, which saves you quite some typing and will be more future proof when we change the annotation/document models.
Which means that you won't need the annotation and document fixture but can just do `factories.Annotation(user=user.userid, shared=True)` in the test above where you use it.

h/accounts/services.py
```diff
@@ -115,6 +115,16 @@ def login(self, username_or_email, password):
...
         return None
 
+    def annotation_count(self, userid):
+        """
+        Return the count of shared annotations for this user.
+        """
+
+        return (
+            self.session.query(Annotation)
+            .filter_by(userid=userid, shared=True)
```
chdorner: Ah! One thing I totally missed in the last review (sorry), but just noticed now when I tested is that we have to filter by `groupid='__world__'` as well in this case. As we don't want to expose how many annotations a user made in other groups. This also means that the test should create a bunch of annotations in a group that is not `__world__` to test that we don't count them.

sheetaluk:
@chdorner ok, all done. ready for re-review :)

ajpeddakotla:
@chdorner thanks for catching the cases I failed to mention - not counting private annotations or annotations made in other groups.

ajpeddakotla:
So after talking @sheetaluk, there needs to be a finer grained approach to counts of annotations that we are showing in both the sidebar and the search results. 

So if we breakdown the different types of annotations that can be viewed, and who they can be viewed by, it looks something like this (for now - this will change with publisher groups). 

Empty |Private | Shared | Public
------ | ------ | ------ | ------
Me | Y | Y | Y
Person I am in one or more groups with | N | Y (but only in groups that I am a part of) | Y
Stranger | N | N | Y

Right now, what's been implemented - we're only showing public annotations which are not private or shared. 

What we'd like to do, because showing the finer detail helps better inform the user of the types of annotations they are making is to: 

**1. In the sidebar:** 
Breakdown the count to include (for users viewing their own profile):
Shared annotations: 
Private annotations: 
Public annotations: 

For users viewing someone else's profile: 
Public annotations: 
Shared annotations (if they share a common group with that user)

**2. In the search results:** 
If I'm viewing my own profile. 
Shared
Private
Public annotations (if I'm viewing my own profile)

If I'm viewing someone else's annotations
Shared (if they share a common group with me)
Public

Issue #4179, opened 2016-12-07
Console error when toggling notification settings


nickstenning:
### Steps to reproduce
1. Go to your user [notification settings page](https://hypothes.is/account/settings/notifications).
2. Quickly toggle the "Email me when..." checkbox on and off

### Expected behaviour
The thing should not emit warnings/errors in the console.

### Actual behaviour
We see a `TypeError: Cannot read property 'length'` error in the browser console.
<img width="632" alt="screen shot 2016-12-07 at 16 38 05" src="https://cloud.githubusercontent.com/assets/3602/20974290/fa8dccba-bc9b-11e6-9732-caa914cd0408.png">

ajpeddakotla:
notes from sprint planning: A significant refactoring doesn't need to be done here to write the tests. The work still be done is test and merge. 

Points given: 2

PR #4217, opened 2016-12-13
Notifications Toggle Error

sean-roberts:
Fixes: #4179

While incorrectly going down the path of testing what happens on the submit-form utility when we throw an error, I came across the fact that we were not cleaning up our global fetch usage. 

Regarding the actual error. I don't think it's at all helpful to present an actual error to the user in the form but we need to have something to make it obvious that something went wrong. So I added a generic error there so we/they have something to tell us if it needs support.

### DIFF
--- a/h/static/scripts/controllers/form-controller.js
+++ b/h/static/scripts/controllers/form-controller.js
@@ -243,7 +243,7 @@ class FormController extends Controller {
         // If there was an error processing the request or the server could
         // not be reached, display a helpful error
         this.setState({
-          submitError: err.reason,
+          submitError: err.reason || 'There was a problem saving changes.',
           saving: false,
         });
       }
--- a/h/static/scripts/tests/controllers/form-controller-test.js
+++ b/h/static/scripts/tests/controllers/form-controller-test.js
@@ -229,9 +229,16 @@ describe('FormController', () => {
   });
 
   it('displays an error if form submission fails without returning a new form', () => {
-    fakeSubmitForm.returns(Promise.reject({status: 500, reason: 'Internal Server Error'}));
+    fakeSubmitForm.returns(Promise.reject({status: 501, reason: 'Internal Server Error'}));
     return submitForm().then(() => {
       assert.equal(submitError(), 'Internal Server Error');
+    });
+  });
+
+  it('Displays a generic error message when an internal error occurs inside of submit form', () => {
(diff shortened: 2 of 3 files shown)


robertknight:
>  I don't think it's at all helpful to present an actual error to the user in the form but we need to have something to make it obvious that something went wrong

It depends. If we take one of the most common cases of the network request failing due to a connectivity issue, does that result in a vaguely useful error message in say Chrome and Firefox?

h/static/scripts/controllers/form-controller.js
```diff
@@ -243,7 +243,7 @@ class FormController extends Controller {
         // If there was an error processing the request or the server could
         // not be reached, display a helpful error
         this.setState({
-          submitError: err.reason,
+          submitError: err.reason || 'Form Processing Error',
```
robertknight: I think we should tweak the wording here to sound a little more friendly. Perhaps "There was a problem saving changes."?

h/static/scripts/tests/controllers/form-controller-test.js
```diff
@@ -229,12 +229,19 @@ describe('FormController', () => {
...
   it('displays an error if form submission fails without returning a new form', () => {
-    fakeSubmitForm.returns(Promise.reject({status: 500, reason: 'Internal Server Error'}));
+    fakeSubmitForm.returns(Promise.reject({status: 501, reason: 'Internal Server Error'}));
     return submitForm().then(() => {
       assert.equal(submitError(), 'Internal Server Error');
     });
   });
 
+  it('Displays a generic error message when an internal error occurs inside of submit form', () => {
+    fakeSubmitForm.returns(Promise.reject('An internal issue'));
```
robertknight: No change needed in this context but I'll just note for future that we try to follow the convention that promises are always rejected with Error-derived objects or Error-like objects (so regardless of what went wrong, you can access the error via the `message` property and in most cases, find the original via the `stack` property).

I did find this mentioned in a useful guide to Promise conventions somewhere but I couldn't find it just now.

h/static/scripts/controllers/form-controller.js
sean-roberts: sure though I was trying to match the tone of the other errors "Form submission failed" "Form validation failed" etc. But I am fine with what you suggested. I will update and merge

PR #4220, opened 2016-12-14
Add Annotation.shared field to the search index

chdorner:
This will allow us to update the `AuthFilter` to only filter out
private annotations of other users. In addition to that we will add a
new filter that checks for the correct group permissions.

This is step one from hypothesis/product-backlog#95.


nickstenning:
LGTM. Will wait for Travis.

PR #4221, opened 2016-12-14
Use a single model metadata instance for h and memex

nickstenning:
Previously each of h and memex had its own independent model base class and associated `sqlalchemy.MetaData` instance. This is problematic if we want to add relationships between models defined in h and models defined in memex, because the SQLAlchemy mapper for h models doesn't know anything about memex models and vice versa.

This PR changes all that by ensuring that there is only one model base class, and it lives in h. We must manually set the base class for memex models before importing any memex model modules. This is handled by the `h.models` package.

### DIFF
--- a/src/memex/db/__init__.py
+++ b/src/memex/db/__init__.py
@@ -1,36 +1,16 @@
 # -*- coding: utf-8 -*-
 
-from sqlalchemy import MetaData
-from sqlalchemy.ext import declarative
+#: The base class used by memex's models.
+#:
+#: This must be configured by the including application using
+#: :py:func:`memex.db.set_base`.
+Base = None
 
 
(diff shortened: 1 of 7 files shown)


src/memex/db/__init__.py
```diff
@@ -1,40 +1,26 @@
 # -*- coding: utf-8 -*-
 
-from sqlalchemy import MetaData
-from sqlalchemy.ext import declarative
+#: The base class used by memex's models.
+#:
+#: This must be configured by the including application using
+#: :py:func:`memex.db.set_base`.
+Base = None
```
chdorner: It might make sense to have a very minimal example app using memex soon. We're adding a lot of config here and there without experiencing how this would look like in a very minimal application that uses memex.

PR #4225, opened 2016-12-15
Move annotation ACL to memex.resources.AnnotationResource

nickstenning:
As part of introducing new kinds of group, we need to be able to have an annotation's ACL depend on the ACL of the group it is in. This is problematic if the ACL lives on the Annotation model because:

1. In order to retrieve the group we need to use memex.groups.find, which takes the request as its first argument.
2. The __acl__ method does not have access to the request.
3. There's no easy way to pass the request to the constructor of a SQLAlchemy mapped class.

The most straightforward way to solve this problem seems to be to introduce a wrapper class, AnnotationResource, which acts as the request context for views operating on annotations, and thus which holds the ACL for the annotation.

This PR does not change any of the access control rules for annotations, it merely introduces the AnnotationResource wrapper class and updates the various views that previously received Annotation objects as context to receive AnnotationResource objects instead.

### DIFF
--- a/h/routes.py
+++ b/h/routes.py
@@ -50,7 +50,7 @@ def includeme(config):
     # Annotations & stream
     config.add_route('annotation',
                      '/a/{id}',
-                     factory='memex.resources:AnnotationFactory',
+                     factory='memex.resources:AnnotationResourceFactory',
                      traverse='/{id}')
     config.add_route('stream', '/stream')
     config.add_route('stream.user_query', '/u/{user}')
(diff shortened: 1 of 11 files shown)


nickstenning:
One thing worth noting: at the moment I don't *think* the streamer needs any changes because it doesn't actually use the Annotation ACL to decide whether the websocket session has perms on the annotation, it uses the rendered `permissions` object:

- https://github.com/hypothesis/h/blob/fa162f630670e4fc28f0dc9356ae59a2ae7c9c99/h/streamer/messages.py#L161-L163
- https://github.com/hypothesis/h/blob/fa162f630670e4fc28f0dc9356ae59a2ae7c9c99/src/memex/presenters.py#L234-L249

PR #4226, opened 2016-12-15
Fix crashing login page

seanh:
With the activity_pages feature flag on, when logged out, the /login
page is crashing because it's trying to access an SVG file that has been
moved. Update it to use the file's new path.


nickstenning:
LGTM.

PR #4227, opened 2016-12-15
Inline logo svg file

seanh:
For consistency's sake, always use an inline svg when using the logo.svg file. I don't think there's any actual reason to do this here, other than consistency.

### DIFF
--- a/h/static/styles/partials-v2/_masthead.scss
+++ b/h/static/styles/partials-v2/_masthead.scss
@@ -7,3 +7,7 @@
   padding-right: 15px;
   padding-top: 15px;
 }
+
+.masthead-logo {
+  color: $grey-6;
+}
--- a/h/templates/includes/logo-header.html.jinja2
+++ b/h/templates/includes/logo-header.html.jinja2
@@ -1,7 +1,7 @@
 <header class="masthead">
   {% if feature('activity_pages') %}
   <a href="/" title="Hypothesis homepage"><!--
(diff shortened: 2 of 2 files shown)


h/static/styles/partials-v2/_masthead.scss
```diff
@@ -7,3 +7,7 @@
   padding-right: 15px;
   padding-top: 15px;
 }
+
+.masthead-logo {
+  color: $black;
```
seanh: This is a link so by browser default it's purple

h/templates/includes/logo-header.html.jinja2
```diff
@@ -1,7 +1,7 @@
 <header class="masthead">
   {% if feature('activity_pages') %}
   <a href="/" title="Hypothesis homepage"><!--
-    !--><img alt="Hypothesis logo" class="masthead-logo" src="{{ asset_url('images/icons/logo.svg') }}"></a>
+    !-->{{ svg_icon('logo', 'masthead-logo', translate_z=True) }}</a>
```
seanh: Same as we do in the other place that we use this icon file

h/static/styles/partials-v2/_masthead.scss
robertknight: Presumably this should be the same color as the logo when it appears in the navbar (ie. a very dark grey)?

PR #4228, opened 2016-12-15
Wrap page contents in <main> element

seanh:
Wrap the main contents of every page that uses the base.html.jinja2 layout in a
`<main>` element.

This is going to enable us to use `display: flex`
(with `flex-direction: column`) on the `<body>` element. We need `<body>` to
be a flex container so that we can add a sticky, variable-height footer to some
pages, using this technique: https://philipwalton.github.io/solved-by-flexbox/

Normally making `<body>` a flex container would mess up the widths of the
content elements of some pages, because a flex container treats widths
differently to a non-flex container. (In particular, it doesn't seem to treat
`max-width` in the same way.) By wrapping the contents of every page in a
`<main>` we ensure that this `<main>`, and not the contents of the page, will
be subject to the `<body>`'s new flex layout, and avoid any unwanted changes to
the widths of page contents.

Using `<main>` also adds some semantic information.

### DIFF
--- a/h/templates/layouts/base.html.jinja2
+++ b/h/templates/layouts/base.html.jinja2
@@ -78,7 +78,12 @@
     {% endfor %}
   </head>
   <body class="body">
-    {% block content %}{% endblock %}
+    {% block header %}{% endblock %}
+
+    <main>
+      {% block content %}{% endblock %}
+    </main>
+
     {% block scripts %}
     {% for url in asset_urls("site_js") %}
     <script src="{{ url }}"></script>
--- a/h/templates/accounts/forgot_password.html.jinja2
+++ b/h/templates/accounts/forgot_password.html.jinja2
@@ -4,6 +4,10 @@
 {% set style_bundle = 'legacy_site_css' %}
 {% endif %}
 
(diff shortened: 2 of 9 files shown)


h/templates/layouts/base.html.jinja2
```diff
@@ -78,7 +78,12 @@
     {% endfor %}
   </head>
   <body class="body">
-    {% block content %}{% endblock %}
+    {% block header %}{% endblock %}
+
+    <main>
+      {% block content %}{% endblock %}
+    </main>
+
```
seanh: Here is where you would put a `{% block footer %}` and have child templates override it to provide a footer if they want to, and the CSS class to make that footer always stick to the bottom could be here in the base template. But that's for a follow-up PR.

seanh:
Although it's not ready for a pull request yet, I've pushed [my branch with the new activity pages footer](https://github.com/hypothesis/h/compare/add-footer-to-activity-pages) that exhibits the problem (that turning `<body>` into a flex box breaks the width of the main page content of multiple pages) that _this_ pr solves.

PR #4230, opened 2016-12-16
Memex demo tweaks

nickstenning:
Minor tweaks to the memex package that are needed to be able to use it more easily independently from the rest of h:

- fix a missing include of `pyramid_services`
- allow automatic provision of a default model base class when needed
- move the handling of an `h.*` setting into h itself

### DIFF
--- a/src/memex/db/__init__.py
+++ b/src/memex/db/__init__.py
@@ -1,5 +1,11 @@
 # -*- coding: utf-8 -*-
 
+import logging
+
+from sqlalchemy.ext.declarative import declarative_base
+
+log = logging.getLogger(__name__)
+
 #: The base class used by memex's models.
 #:
 #: This must be configured by the including application using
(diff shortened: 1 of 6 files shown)


src/memex/db/__init__.py
```diff
@@ -20,7 +26,14 @@ def init(engine, base=None, should_create=False, should_drop=False):
         base.metadata.create_all(engine)
 
 
-def set_base(basecls):
+def set_base(basecls=None):
```
chdorner: This is exactly why I was bringing up a memex demo app last time, I was thinking that always having to provide a base class might be cumbersome for some more minimal implementations.

src/memex/search/__init__.py
```diff
@@ -43,12 +43,10 @@ def includeme(config):
...
     # reread the settings.
-    config.registry['es.client'] = _get_client(settings)
+    config.registry['es.client'] = client = _get_client(settings)
     config.add_request_method(
         lambda r: r.registry['es.client'],
         name='es',
         reify=True)
 
-    # If requested, automatically configure the index
-    if asbool(settings.get('h.search.autoconfig', False)):
```
chdorner: Seems like this crept in at some point, nice to see it gone.

PR #4232, opened 2016-12-16
Change text for annotation count in user sidebar from shared annotati…

sheetaluk:
…ons ro public annotations.

Change the text that appears next to the annotation count in the user side bar to "Public annotations" instead of "shared annotations" because shared could imply non-public but annotations shared within groups. Public annotations refer to annotations created in the public channel, which is what the count represents.

### DIFF
--- a/h/templates/activity/search.html.jinja2
+++ b/h/templates/activity/search.html.jinja2
@@ -276,7 +276,7 @@
       <dl>
         {% if user.annotation_count %}
           <dt class="search-result-sidebar__dt">
-            {% trans %}Shared annotations:{% endtrans %}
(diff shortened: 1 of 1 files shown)


robertknight:
It would be helpful to include some brief context for the change in the PR description or commit message to make it easier for reviewers to understand what is going on. Questions that immediately came to mind:

- Has the way that the count is calculated changed recently?
- Does the count differ depending on whether I am looking at the sidebar for my profile vs someone else's?
- Does the count differ depending on what groups I am in and what groups the user I am looking at is in?

robertknight:
LGTM.

PR #4233, opened 2016-12-16
Annotation 'read' permissions delegate to their group

nickstenning:
Instead of hard-coding the principals allowed the annotation 'read' permission based on our existing types of group, delegate the permissions through to the underlying group resource returned by memex.groups.find.

With only two kinds of group, namely,

1. Public (everyone can read)
2. Private (members can read)

computing the annotation ACL was straightforward.

But now we are looking to introduce new kinds of groups where every user with a given authority can read the annotations in a group, or everyone can read annotations in a group despite it not being the "__world__" group.

Now we have to actually fetch the group in question and delegate the 'read' permission on the annotation to those principals permitted the 'read' permission on the containing group.

### DIFF
--- a/src/memex/resources.py
+++ b/src/memex/resources.py
@@ -1,7 +1,10 @@
 # -*- coding: utf-8 -*-
 
+from __future__ import unicode_literals
+
 from pyramid import security
 
+from memex import groups
 from memex import storage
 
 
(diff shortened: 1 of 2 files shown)


nickstenning:
Tagging hypothesis/product-backlog#96.

PR #4234, opened 2016-12-16
Add footer to activity pages

seanh:
Fixes https://github.com/hypothesis/product-backlog/issues/87

## Mockup

de5fd91a-c07a-11e6-96af-7e912506fe0f

## Screenshots

Full size:

screenshot from 2016-12-16 16-06-40

iPad pro:

screenshot from 2016-12-16 16-08-17

iPad:

screenshot from 2016-12-16 16-08-40

iPhone 5:

screenshot from 2016-12-16 16-09-33

# Zero state

I changed the design of the search page zero state, as the CSS that tries to stretch the grey box to the bottom of the viewport gets in the way of the footer, I decided just to remove this feature rather than fix it. Before:

screenshot from 2016-12-16 16-14-49

After:

screenshot from 2016-12-16 16-15-18

## Notes

The footer appears on the search page and user account pages, but not on the group edit pages or logged-out forms. Same as the nav bar.


h/app.py
```diff
@@ -104,6 +104,34 @@ def includeme(config):
...
+        'hypothesis-twitter', 'https://twitter.com/hypothes_is', static=True)
+    config.add_route('jobs', '/jobs/', static=True)
+    config.add_route('press', '/press/', static=True)
+    config.add_route('privacy', '/privacy/', static=True)
+    config.add_route('roadmap', '/roadmap/', static=True)
+    config.add_route('team', '/team/', static=True)
+    config.add_route('terms-of-service', '/terms-of-service/', static=True)
+    config.add_route(
+        'wordpress-plugin', 'https://wordpress.org/plugins/hypothesis/',
+        static=True)
```
seanh: Not all of these routes are used in the new footer of course, I just cleaned up all the hrefs on the home page so that none of them are hardcoded anymore

h/static/styles/partials-v2/_footer.scss
```diff
@@ -0,0 +1,93 @@
+// The width (not including margins) of the brand component.
+$brand-width: 113px;
+
+// The amount of horizontal space between items in the footer.
+$spacing: 30px;
+
+.site-footer {
```
robertknight: The name of the main component in a file should be the same as the file name.

h/static/styles/partials-v2/_footer.scss
```diff
@@ -0,0 +1,93 @@
...
+  flex-wrap: wrap;
+  flex-basis: calc(100% - #{$brand-width + $spacing});
+}
+
+.site-footer__row {
+  display: flex;
+  align-items: center;
+}
+
+.u-push-right {
```
robertknight: Utilities should go in the existing `_util.scss` file.

h/static/styles/partials-v2/_search.scss
```diff
@@ -46,15 +46,6 @@
   margin-right: auto;
 }
 
-.search-result-container--empty {
-  /* The default height of the search container is calculated to so that it
-     stretches to the bottom of the viewport and no further.
-     The px value here is the height of the stuff above the
-     search-result-container. */
-  height: calc(100vh - 134px);
-  margin-bottom: 0;
```
robertknight: Agreed with the comment in the PR description - removing this makes sense.

h/panels/footer.py
```diff
@@ -0,0 +1,13 @@
...
+
+"""The footer displayed at the bottom of most pages."""
+
+from __future__ import unicode_literals
+
+from pyramid_layout.panel import panel_config
+
+
+@panel_config(name='footer', renderer='h:templates/panels/footer.html.jinja2')
+def footer(context, request):
```
seanh: It's probably not going to have any logic I guess, but I just did it as a panel for consistency with the nav bar. Could be changed to an include though (the I wouldn't need to add a Python test to keep codecov happy!)

dwhly:
At some point, we'll want to tighten up the mishmashed menu layout at narrower browser widths.  Not sure if there's anything to be done right now tho.

h/app.py
nickstenning: These should go with the rest of the routes in `h/routes.py`.

robertknight:
In Safari on my iPhone 5 the footer overflows horizontally:



----

seanh:
@dwhly You mean that _Blog_, _Jobs_ and _Terms of Service_ should be left-aligned with each other in a column?

screenshot from 2016-12-16 16-09-33

It was actually quite difficult to achieve that, while still having the layout work in the other desired ways for other screen sizes, and having it so that future adds, edits or deletes of footer links wouldn't potentially badly break the layout at certain screen sizes. Someone can always try to improve on it later

h/panels/footer.py
nickstenning: 👍 if it doesn't have any python code in the panel method, there's no point it being a panel -- just shove it in `templates/includes/`.

h/templates/home.html.jinja2
```diff
@@ -9,7 +9,7 @@
   {{ super() }}
   <link rel='canonical' href='{{ base_url }}'>
```
nickstenning: It would be nice to clean up these last few references to `base_url` so we can get rid of it -- I don't think there's any reason we shouldn't just use `request.route_url('index')` here.

seanh:
Hmm, rats, the overflow on iPhone is annoying. It doesn't happen in the iPhone simulator in Chrome dev tools. I'll try to fix it. Have pushed fixes for the rest of the changes

robertknight:
On Chrome stable on my Moto G4 and in the simulator, I see the footer overlap the bottom of the annotation list in a vertical layout:





Meanwhile in a horizontal layout there is too much vertical space at the bottom:

h/static/styles/partials-v2/_footer.scss
```diff
@@ -0,0 +1,65 @@
+// The amount of horizontal space between items in the footer.
+$spacing: 30px;
```
robertknight: Just a note for future reference - A potential hazard here is that SASS `@import` is dumb textual inclusion - so you could end up accidentally referencing a generically named variable like this in another file. A trick I have used in the client to avoid this is:

```scss
@at-root {
  $local-variable-name: value;

  // rest of the file here
}
```

Where `@at-root` starts a scope but does not alter the output CSS.

h/templates/home.html.jinja2
seanh: Ran out of time so leaving this for follow up

PR #4241, opened 2016-12-19
Add focus state to lozenge removal button

robertknight:
Add focus state to lozenge removal button so that it is obvious when a
particular lozenge's remove button has keyboard focus (also applies when remove button is pressed).

Lozenge where close button has keyboard focus:

----

**Before:**

lozenge-focus-before

**After:**

lozenge-focus-after


nickstenning:
Nice polish. :+1:

PR #4242, opened 2016-12-19
Mark annotations as deleted in search index

chdorner:
Part of hypothesis/product-backlog#106.

This makes sure that we don't delete annotations from the search index, but rather mark them as deleted. These annotations then need to be filtered out when we query the index.

### DIFF
--- a/src/memex/search/index.py
+++ b/src/memex/search/index.py
@@ -59,10 +59,10 @@ def index(es, annotation, request):
 
 def delete(es, annotation_id):
     """
-    Delete an annotation from the search index.
+    Mark an annotation as deleted in the search index.
 
-    If no annotation with the given annotation's ID exists in the search index,
-    just log the resulting elasticsearch exception (don't crash).
+    This will write a new body that only contains the ``deleted`` boolean field
+    with the value ``true``.
 
     :param es: the Elasticsearch client object to use
     :type es: memex.search.Client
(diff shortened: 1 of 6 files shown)


chdorner:
Just looking at the issues Jenkins found with the functional tests.

src/memex/search/index.py
```diff
@@ -59,10 +59,10 @@ def index(es, annotation, request):
 
 def delete(es, annotation_id):
     """
-    Delete an annotation from the search index.
+    Mark an annotation as deleted in the search index.
 
-    If no annotation with the given annotation's ID exists in the search index,
-    just log the resulting elasticsearch exception (don't crash).
+    This will write a new body that only contains the ``deleted`` boolean field
+    with the value ``true``.
```
nickstenning: It would be great to add a line or two here about _why_ we're doing this.

src/memex/search/query.py
```diff
@@ -200,6 +200,19 @@ def __call__(self, params):
...
+
+    """
+    A filter that only returns non-deleted documents.
+
+    Documents are not getting deleted from the index, they only get marked as
+    deleted.
+    """
+
+    def __call__(self, _):
+        return {"bool": {"must_not": {"exists": {"field": "deleted"}}}}
```
nickstenning: I see the `missing` filter has been deprecated in more recent Elasticsearch versions. 👍

src/memex/search/index.py
chdorner: Done on master.

PR #4244, opened 2016-12-19
Add a "deleted" column to the annotation table

nickstenning:
Deleting annotations on the web thread makes many things more complicated than they should be, particularly the streamer.

Adds a "deleted" column to the annotation table, which will be used to mark annotations as deleted for later cleanup.


nickstenning:
Wrong branch -- 1 second.

nickstenning:
Ok, I think we should be good now.

PR #4245, opened 2016-12-19
Use op_type "create" when reindexing

chdorner:
This is part of hypothesis/product-backlog#106.

The indexer workers will soon write to both the current and the new index during a re-index operation. To ensure that the re-indexer doesn't override any annotations already in the index with potentially old data we will set the Elasticsearch `op_type` to `"create"`. This means that indexing a document that already exists in the index will fail with an error. Since these errors are expected we will not return them to the caller.

### DIFF
--- a/src/memex/search/index.py
+++ b/src/memex/search/index.py
@@ -108,10 +108,11 @@ class BatchIndexer(object):
     the search index.
     """
 
-    def __init__(self, session, es_client, request, target_index=None):
+    def __init__(self, session, es_client, request, target_index=None, op_type='index'):
         self.session = session
         self.es_client = es_client
         self.request = request
+        self.op_type = op_type
 
         # By default, index into the open index
         if target_index is None:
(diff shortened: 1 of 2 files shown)


src/memex/search/index.py
```diff
@@ -141,13 +143,16 @@ def index(self, annotation_ids=None):
         errored = set()
         for ok, item in indexing:
             if not ok:
-                errored.add(item['index']['_id'])
+                status = item[self.op_type]
+                if self.op_type != 'create' and \
```
nickstenning: I'm not sure this conditional is right...

```python
was_index_op = self.op_type != 'create'
was_other_error = 'document already exists' not in status['error']

if was_index_op and was_other_error:
    record(error)
```

Doesn't that mean that *all* errors thrown on 'create' will be swallowed?

I'd think that something like this would be clearer:

```
was_doc_exists_err = 'document already exists' in status['error']
if self.op_type == 'create' and was_doc_exists_err:
    continue
```

chdorner: Fixed.

tests/memex/search/index_test.py
```diff
@@ -271,6 +303,27 @@ def fake_streaming_bulk(*args, **kwargs):
...
+        indexer = index.BatchIndexer(db_session, es, pyramid_request,
+                                     op_type='create')
+
+        ann_success_1, ann_success_2 = factories.Annotation(), factories.Annotation()
+        ann_fail_1, ann_fail_2 = factories.Annotation(), factories.Annotation()
+
+        def fake_streaming_bulk(*args, **kwargs):
+            for ann in args[1]:
+                if ann.id in [ann_fail_1.id, ann_fail_2.id]:
+                    error = 'DocumentAlreadyExistsException[[index-name][1] [annotation][gibberish]: ' \
```
chdorner: Done

nickstenning:
LGTM.

PR #4246, opened 2016-12-19
Centralise tasks

nickstenning:
In preparation for adding a "purge deleted annotations" task, which wouldn't have a logical place to live in `h`, this PR centralises all the Celery task definitions in their own package, much as we've done with models and views.

The only non-trivial move here is `h.indexer`, where the Celery task definitions have moved but the Pyramid configuration remains in `h.indexer`.

I've also removed what appears to be an unused `reindex` task.


h/nipsa/__init__.py
```diff
@@ -11,7 +10,7 @@ def includeme(config):
                           'memex.events.AnnotationTransformEvent')
 
     # Register the NIPSA service
-    config.register_service_factory(services.nipsa_factory, name='nipsa')
+    config.register_service_factory('.services.nipsa_factory', name='nipsa')
```
nickstenning: This is necessary to prevent an import cycle, because:

- `h.tasks.nipsa` imports `h.nipsa.search`
- `h.nipsa` imports `h.nipsa.services`
- `h.nipsa.services` imports `h.tasks.nipsa`

Switching to a dotted-string import here fixes the issue, but it's worth noting that this would also be solved by moving all services into a `services` package. I expect we'll do that soon, too...

nickstenning:
I ran into some pretty strange things while testing this, so make sure you remove `.pyc` files and do a full restart of the Celery workers and app.

PR #4247, opened 2016-12-19
Show private and group annotation count if user is looking at their o…

sheetaluk:
…wn profile.

We want the user to see their total number of private annotations as well as the number of shared annotations they have made in groups while viewing their own profile.

I have the following questions:
1. Does the query to retrieve annotations also include notes?
2. Is there a better way to retrieve the total number of shared annotations across groups for a user?
3. Should we use the same feature flag 'total_shared_annotations' be used for private and group annotations as well?

https://github.com/hypothesis/h/issues/4168

### DIFF
--- a/h/services/annotation_stats.py
+++ b/h/services/annotation_stats.py
@@ -1,0 +1,37 @@
+# -*- coding: utf-8 -*-
+
+import sqlalchemy as sa
+
+from h.models import Annotation
+
+
+class AnnotationStatsService(object):
+    """A service for retrieving annotation stats for users and groups."""
+
+    def __init__(self, session):
+        self.session = session
+
+    def user_annotation_counts(self, userid):
+        """Return the count of annotations for this user."""
+
+        annotations = self.session.query(Annotation).filter_by(userid=userid).options(sa.orm.load_only('groupid', 'shared')).subquery()
+        grouping = sa.case([
+            (sa.not_(annotations.c.shared), 'private'),
+            (annotations.c.groupid == '__world__', 'public'),
+        ], else_='group')
+        return dict(self.session.query(grouping, sa.func.count(annotations.c.id)).group_by(grouping).all())
+
+    def group_annotation_count(self, pubid):
+        """
+        Return the count of shared annotations for this group.
+        """
+
+        return (
+            self.session.query(Annotation)
+            .filter_by(groupid=pubid, shared=True)
+            .count())
+
+
+def annotation_stats_factory(context, request):
+    """Return an AnnotationStatsService instance for the passed context and request."""
+    return AnnotationStatsService(session=request.db)
--- a/h/views/activity.py
+++ b/h/views/activity.py
@@ -130,16 +130,19 @@ class GroupSearchController(SearchController):
                    for u in self.group.members]
         members = sorted(members, key=lambda k: k['username'].lower())
 
-        annotation_count = None
+        group_annotation_count = None
         if self.request.feature('total_shared_annotations'):
-            annotation_count = self.request.find_service(name='group').annotation_count(self.group.pubid)
+            group_annotation_count = self.request.find_service(name='annotation_stats').group_annotation_count(self.group.pubid)
+
+        result['stats'] = {
+            'annotation_count': group_annotation_count,
+        }
 
         result['group'] = {
             'created': self.group.created.strftime('%B, %Y'),
             'description': self.group.description,
             'name': self.group.name,
             'pubid': self.group.pubid,
-            'annotation_count': annotation_count,
             'url': self.request.route_url('group_read',
                                           pubid=self.group.pubid,
                                           slug=self.group.slug),
@@ -258,11 +261,17 @@ class UserSearchController(SearchController):
 
         annotation_count = None
         if self.request.feature('total_shared_annotations'):
-            annotation_count = self.request.find_service(name='user').annotation_count(self.user.userid)
+            user_annotation_counts = self.request.find_service(name='annotation_stats').user_annotation_counts(self.user.userid)
+            annotation_count = user_annotation_counts['public']
+            if self.request.authenticated_userid == self.user.userid:
+                annotation_count = annotation_count + user_annotation_counts['private'] + user_annotation_counts['group']
+
+        result['stats'] = {
+            'annotation_count': annotation_count,
+        }
 
         result['user'] = {
             'name': self.user.display_name or self.user.username,
-            'annotation_count': annotation_count,
             'description': self.user.description,
             'registered_date': self.user.registered_date.strftime('%B, %Y'),
             'location': self.user.location,
--- a/tests/h/services/annotation_stats_test.py
+++ b/tests/h/services/annotation_stats_test.py
@@ -10,7 +10,7 @@ from h.services.annotation_stats import annotation_stats_factory
 
 
 class TestAnnotationStatsService(object):
-    def test_user_annotation_counts_returns_count_of_annotations_for_user(self, svc, factories, db_session):
+    def test_user_annotation_counts_returns_count_of_annotations_for_user(self, svc, factories):
         userid = '123'
         for i in range(3):
             factories.Annotation(userid=userid, shared=True)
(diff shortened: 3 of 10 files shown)


h/accounts/services.py
```diff
@@ -115,16 +115,36 @@ def login(self, username_or_email, password):
...
+        """
+        Return the count of shared annotations this user has made in groups.
+        """
+
+        return (
+            self.session.query(Annotation)
+            .filter_by(userid=userid, shared=True)
+            .count())
+
+    def private_annotation_count(self, userid):
```
nickstenning: Given that each of these methods just queries the Annotation table, it seems to me that there's no particular reason these need to stay on the `UserService`, and it would be clearer if there was a separate `StatsService` for this.

Second, every round-trip to the database adds latency, so if we can avoid doing three, that would be good. Here's a quick sketch of what doing this in one query would look like:

```python
import sqlalchemy as sa
...

annotations = session.query(m.Annotation).filter_by(userid=userid).subquery()
grouping = sa.case([
    (sa.not_(annotations.c.shared), 'private'), 
    (annotations.c.groupid == '__world__', 'public'),
], else_='group')
results = session.query(sa.func.count(annotations.c.id), grouping).group_by(grouping).all()
```

h/views/activity.py
```diff
@@ -256,13 +256,22 @@ def domain(user):
...
         if self.request.feature('total_shared_annotations'):
-            annotation_count = self.request.find_service(name='user').annotation_count(self.user.userid)
+            if self.request.authenticated_userid == self.user.userid:
+                private_annotation_count = self.request.find_service(name='user').private_annotation_count(self.user.userid)
+                group_annotation_count = self.request.find_service(name='user').all_groups_annotation_count(self.user.userid) - public_annotation_count
 
         result['user'] = {
             'name': self.user.display_name or self.user.username,
-            'annotation_count': annotation_count,
+            'public_annotation_count': public_annotation_count,
```
nickstenning: It would be good to wrap the stats here up into a single `stats` object that gets passed into the template, rather than making this view method more and more complicated for every new stat we pass in. This would simplify the tests, as well.

tests/h/views/activity_test.py
```diff
@@ -594,25 +594,101 @@ def test_search_passes_the_display_name_to_the_template_if_the_user_has_one(
...
+                                                                                                    user,
+                                                                                                    user_service,
+                                                                                                    pyramid_request):
         pyramid_request.feature.flags['total_shared_annotations'] = False
         controller.search()
 
-        assert user_service.annotation_count.called == False
+        assert user_service.public_annotation_count.called is False
+
+    def test_search_does_not_pass_the_private_annotation_count_to_the_template_when_user_is_different_from_authenticated_user(self,
```
nickstenning: Some of these test names are getting pretty silly :joy:

It's ok for the test names to be a bit shorter, and then add a docstring to the test method explaining to the reader what it's supposed to be testing.

h/services/annotation_stats.py
```diff
@@ -0,0 +1,35 @@
+import sqlalchemy as sa
+
+from h.models import Annotation
+
+
+class AnnotationStatsService(object):
+    """A service for retrieving annotation stats for a user."""
```
nickstenning: It's not just for a user, so let's update this.

h/services/annotation_stats.py
```diff
@@ -0,0 +1,35 @@
+import sqlalchemy as sa
```
nickstenning: Could you make sure this has ` # -*- coding: utf-8 -*-` at the top of the file?

tests/h/services/annotation_stats_test.py
```diff
@@ -0,0 +1,70 @@
...
+            factories.Annotation(userid=user.userid, groupid='abc', shared=True)
+
+        results = svc.user_annotation_counts(user.userid)
+
+        assert results['public'] == 3
+        assert results['private'] == 2
+        assert results['group'] == 4
+
+    @pytest.mark.usefixtures('group')
+    def test_annotation_count_returns_count_of_shared_annotations_for_group(self, svc, db_session, group, factories):
```
nickstenning: I don't think you actually need to create a group here, so the `group` fixture is kind of redundant. Just make up a groupid and use it for the annotations.

tests/h/views/activity_test.py
```diff
@@ -594,25 +592,40 @@ def test_search_passes_the_display_name_to_the_template_if_the_user_has_one(
...
+                                                                                                   pyramid_request,
+                                                                                                   search,
+                                                                                                   user):
         pyramid_request.feature.flags['total_shared_annotations'] = False
-        controller.search()
+        result = controller.search()['stats']
+
+        assert result['annotation_count'] is None
+
+    def test_search_does_not_pass_the_user_annotation_counts_to_the_template_when_user_is_different_from_auth_user(self,
```
nickstenning: I still think some of these test names are a bit long and unwieldy. Far more readable to just call the test something short and use a docstring to explain the case you're testing here.

tests/h/views/activity_test.py
```diff
@@ -671,6 +684,7 @@ def test_search_returns_the_user_zero_message_to_the_template(
 
     def test_search_shows_the_getting_started_box_when_on_your_own_page(
             self, controller, search, user):
+
```
nickstenning: Hello spurious space 😂 !

tests/h/views/activity_test.py
```diff
@@ -921,8 +943,9 @@ def routes(pyramid_config):
 
 
 @pytest.fixture
-def search(patch):
+def search(patch, pyramid_config):
     search = patch('h.views.activity.SearchController.search')
+    pyramid_config.register_service(FakeAnnotationStatsService(), name='annotation_stats')
```
nickstenning: This should really be in its own fixture, not attached to the `search` fixture :)

tests/h/services/annotation_stats_test.py
```diff
@@ -0,0 +1,69 @@
...
+        for i in range(4):
+            factories.Annotation(userid=user.userid, groupid='abc', shared=True)
+
+        results = svc.user_annotation_counts(user.userid)
+
+        assert results['public'] == 3
+        assert results['private'] == 2
+        assert results['group'] == 4
+
+    def test_annotation_count_returns_count_of_shared_annotations_for_group(self, svc, db_session, group, factories):
```
nickstenning: The `group` param isn't used/needed here now, right?

tests/h/services/annotation_stats_test.py
```diff
@@ -0,0 +1,69 @@
...
+        pubid = 'abc123'
+        for i in range(3):
+            factories.Annotation(groupid=pubid, shared=True)
+        for i in range(2):
+            factories.Annotation(groupid=pubid, shared=False)
+
+        assert svc.group_annotation_count(pubid) == 3
+
+    @pytest.fixture
```
nickstenning: Ditto, this fixture isn't needed any more, because you're just using a static pubid of 'abc123'.

tests/h/services/annotation_stats_test.py
```diff
@@ -0,0 +1,69 @@
...
+import mock
+import pytest
+
+from h.services.annotation_stats import AnnotationStatsService
+from h.services.annotation_stats import annotation_stats_factory
+
+
+class TestAnnotationStatsService(object):
+    @pytest.mark.usefixtures('user')
+    def test_user_annotation_counts_returns_count_of_annotations_for_user(self, svc, user, factories, db_session):
```
nickstenning: Actually the same thing is true here. You don't need the user fixture at all. Just make up a userid, much as you've done for the group pubid.

tests/h/services/annotation_stats_test.py
```diff
@@ -0,0 +1,69 @@
...
+        assert svc.group_annotation_count(pubid) == 3
+
+    @pytest.fixture
+    def group(self, user, factories):
+        return factories.Group(name='Donkey Trust',
+                               authority='foobar.com',
+                               creator=user)
+
+    @pytest.fixture
+    def user(self, db_session, factories):
```
nickstenning: And that means this isn't needed either...

chdorner:
The two test functions in `tests/h/services/annotation_stats_test.py` received the `db_session` fixture but didn't use it. I just removed that and am now waiting for the tests. I will merge after that.

PR #4248, opened 2016-12-19
Don't delete annotations synchronously

nickstenning:
Removing the annotation from the database synchronously (i.e. in the DELETE request) is problematic, as it complicates the process of streaming that delete to WebSocket users.

This commit adds a "deleted" property to the Annotation model (added to the table in 11f8aba) and updates the AnnotationResource ACL so that nobody has any permissions on a deleted annotation.

This ensures that the annotation cannot be returned in any direct-access request. The preexisting search index code will ensure that the annotation is not returned in search results.

This PR also adds a celerybeat task to clean up these deleted annotations on an hourly basis.


nickstenning:
Seeing as this adds a task which, if incorrect, could blow away our entire annotation table, I'm very happy for a slow and steady review here... 😂

PR #4249, opened 2016-12-19
Store new index name during reindex

chdorner:
This is part of hypothesis/product-backlog#106.

When we start a reindex, this makes sure that we store the name of the index in the `setting` table (introduced in #4243). This is done in preparation for the indexer workers to be able to write to both the current and the new index at the same time.


tests/common/factories.py
```diff
@@ -133,3 +133,13 @@ class Meta:  # pylint: disable=no-init, old-style-class
         force_flush = True
 
     userid = factory.LazyAttribute(lambda _: ('acct:' + FAKER.user_name() + '@example.com'))
+
+
+class Setting(ModelFactory):
+
+    class Meta:  # pylint: disable=no-init, old-style-class
+        model = models.Setting
+        force_flush = True
```
nickstenning: Is there a reason this has `force_flush = True`? Ideally none of our factories would have this and we'd flush manually in the tests when necessary...

tests/h/services/settings_test.py
```diff
@@ -0,0 +1,69 @@
...
+    def test_delete_deletes_existing_setting(self, db_session, svc, factories):
+        setting = factories.Setting()
+
+        svc.delete(setting.key)
+        db_session.flush()
+
+        assert db_session.query(Setting).get(setting.key) is None
+
+    def test_delete_is_noop_when_setting_missing(self, db_session, svc, factories):
+        # create a random setting
```
nickstenning: Does this serve a purpose here?

h/services/settings.py
```diff
@@ -0,0 +1,48 @@
...
+        setting = self._fetch(key)
+
+        if setting is not None:
+            self.session.delete(setting)
+
+    def _fetch(self, key):
+        return self.session.query(Setting).get(key)
+
+
+def settings_factory(context, request):
```
nickstenning: In the new `h.settings` package I think we can probably stick to one service per module, so what do you think about reducing stutter and calling these factory functions just `factory` so that the setup looks like:

```
config.register_service_factory('.settings.factory', name='settings')
```

h/indexer.py
```diff
@@ -13,27 +13,40 @@
...
+SETTING_NEW_INDEX = u'reindex.new_index'
+
 
 def reindex(session, es, request):
     """Reindex all annotations into a new index, and update the alias."""
 
     if get_aliased_index(es) is None:
         raise RuntimeError('cannot reindex if current index is not aliased')
 
+    settings = request.find_service(name='settings')
```
nickstenning: In this case I'd suggest that it's probably nicer to actually construct the service yourself using `session` rather than using `request.find_service`. That way we're not introducing another dependency on the `request` object here.

h/indexer.py
```diff
@@ -13,27 +13,40 @@
...
-        log.debug('failed to index {} annotations, retrying...'.format(
-            len(errored)))
-        errored = indexer.index(errored)
-        if errored:
-            log.warn('failed to index {} annotations: {!r}'.format(
-                len(errored),
-                errored))
+    try:
+        settings.put(SETTING_NEW_INDEX, new_index)
+        request.tm.commit()
```
nickstenning: Hmm. Ideally this would use `session.commit()` rather than `request.tm.commit()`, but I can see how that's not going to work right now...

...gah. Not your problem, and it'll do for now.

PR #4250, opened 2016-12-20
Reindex write to both indices

chdorner:
~~**Depends on #4249 to be merged and then this needs a rebase.**~~

This is the last piece of the puzzle that is zero-downtime re-indexing (hypothesis/product-backlog#106).

This PR:
* Allows to specify an index name for `memex.search.index.index` and `memex.search.index.delete`, it defaults to `request.es.index`
* Makes sure that the indexer workers for adding and deleting annotations are calling doing their operation twice, once with the default index, and once with the new index name that is stored in the `setting` table

---

This is making sure that certain race-conditions don't break the new index during a re-index. So testing this manually is a bit cumbersome. I've made sure that I have two annotations in my database (and index), one for updating its text, and one for deleting, both during the re-index at specific points. You will have to make a note of the ids of these two annotations (in the examples below I'm using `annotation-id-update`, and `annotation-id-delete`).

Apply this diff to `src/memex/search/index.py` (and replace the annotation ids):
```diff
diff --git a/src/memex/search/index.py b/src/memex/search/index.py
index 15b013d..f6b67fc 100644
--- a/src/memex/search/index.py
+++ b/src/memex/search/index.py
@@ -144,6 +144,16 @@ class BatchIndexer(object):
         event = AnnotationTransformEvent(self.request, data)
         self.request.registry.notify(event)
 
+        if annotation.id == 'annotation-id-update':
+            import time
+            print('now update annotation-id-update - sleeping...')
+            time.sleep(30)
+
+        if annotation.id == 'annotation-id-delete':
+            import time
+            print('now delete annotation-id-delete - sleeping...')
+            time.sleep(30)
+
         return (action, data)
 
     def _stream_all_annotations(self, chunksize=2000):
```
This will make sure that right after it renders each of the two test annotations (but before sending it to Elasticsearch),  the re-index process will pause for 30 seconds allowing you to either edit the annotation's text or to delete it.

Open these tabs (make sure to replace the annotation ids):

* https://localhost:5000/a/annotation-id-update
* https://localhost:5000/a/annotation-id-delete
* http://localhost:9200/hypothesis/annotation/annotation-id-update
* http://localhost:9200/hypothesis/annotation/annotation-id-delete

Then follow these steps:

Experience the bug:

1. Check out the `master` branch
2. Run a re-index without making any changes, to make sure you start with a correct index
3. Run `bin/hypothesis --dev search reindex`
4. Wait for console print "now update annotation annotation-id-update - sleeping..."
5. Edit the text of `annotation-id-update` in the first browser tab
6. Reload the third browser tab to see that the text change has made it into the current index
7. Wait for console print "now delete annotation annotation-id-delete - sleeping..."
8. Delete annotation `annotation-id-delete` in the second browser tab
9. Reload the fourth browser tab to see that the annotation is marked as deleted in the current index
10. Wait for the re-index to finish (the `hypothesis` alias has been swapped to the new index)
11. Reload the third browser tab: notice that the text changed to its initial state 💣 💥 
12. Reload the fourth browser tab: notice that the deleted annotation is back again 💣 💥 

Experience the fix:

1. Check out this branch (`reindex-write-to-both-indices`)
2. Re-create the test annotation that will be deleted (as it is now gone from the database)
3. Replace all occurrences of the old "to-delete-annotation-id" to the new id
4. Run a re-index without making any changes, to make sure you start with a correct index
5. Run `bin/hypothesis --dev search reindex`
6. Wait for console print "now update annotation annotation-id-update - sleeping..."
7. Edit the text of `annotation-id-update` in the first browser tab
8. Reload the third browser tab to see that the text change has made it into the current index
9. Wait for console print "now delete annotation annotation-id-delete - sleeping..."
10. Delete annotation `annotation-id-delete` in the second browser tab
11. Reload the fourth browser tab to see that the annotation is marked as deleted in the current index
12. Wait for the re-index to finish (the `hypothesis` alias has been swapped to the new index)
13. Reload the third browser tab: notice that the text is still correct ✔️ 🎉 
14. Reload the fourth browser tab: notice that the deleted annotation is missing from the index ✔️ 🎉

### DIFF
--- a/h/indexer/__init__.py
+++ b/h/indexer/__init__.py
@@ -1,0 +1,12 @@
+# -*- coding: utf-8 -*-
+
+from h.indexer.reindexer import reindex
+
+__all__ = (
+    'reindex',
+)
+
+
+def includeme(config):
+    config.add_subscriber('h.indexer.subscribers.subscribe_annotation_event',
+                          'memex.events.AnnotationEvent')
--- a/src/memex/search/index.py
+++ b/src/memex/search/index.py
@@ -57,7 +57,7 @@ def index(es, annotation, request, target_index=None):
     )
 
 
-def delete(es, annotation_id):
+def delete(es, annotation_id, target_index=None):
     """
     Mark an annotation as deleted in the search index.
 
@@ -72,9 +72,15 @@ def delete(es, annotation_id):
         delete from the search index
     :type annotation_id: str
 
+    :param target_index: the index name, uses default index if not given
+    :type target_index: unicode
     """
+
+    if target_index is None:
+        target_index = es.index
+
     es.conn.index(
-        index=es.index,
+        index=target_index,
         doc_type=es.t.annotation,
         body={'deleted': True},
         id=annotation_id)
--- a/src/memex/search/index.py
+++ b/src/memex/search/index.py
@@ -9,7 +9,7 @@ from collections import namedtuple
 
 import elasticsearch
 from elasticsearch import helpers as es_helpers
-from sqlalchemy.orm import subqueryload
+from sqlalchemy.orm import load_only, subqueryload
 
 from memex import models
 from memex import presenters
@@ -152,15 +152,18 @@ class BatchIndexer(object):
         # the database while still supporting eagerloading of associated
         # document data.
 
-        updated = self.session.query(models.Annotation.updated). \
+        updated = self._base_query(). \
+                options(load_only('updated')). \
                 execution_options(stream_results=True). \
+                filter_by(deleted=False). \
                 order_by(models.Annotation.updated.desc()).all()
 
         count = len(updated)
         windows = [Window(start=updated[min(x+chunksize, count)-1].updated,
                           end=updated[x].updated)
                    for x in xrange(0, count, chunksize)]
-        basequery = self._eager_loaded_query().order_by(models.Annotation.updated.asc())
+        basequery = self._eager_loaded_query(). \
+            order_by(models.Annotation.updated.asc())
 
         for window in windows:
             in_window = models.Annotation.updated.between(window.start, window.end)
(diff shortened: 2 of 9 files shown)


src/memex/search/index.py
```diff
@@ -143,13 +154,16 @@ def _stream_all_annotations(self, chunksize=2000):
...
+                filter_by(deleted=False). \
                 order_by(models.Annotation.updated.desc()).all()
 
         count = len(updated)
         windows = [Window(start=updated[min(x+chunksize, count)-1].updated,
                           end=updated[x].updated)
                    for x in xrange(0, count, chunksize)]
-        basequery = self._eager_loaded_query().order_by(models.Annotation.updated.asc())
+        basequery = self._eager_loaded_query(). \
+            filter_by(deleted=False). \
```
nickstenning: This works ok, but rather than duplicating the filter in three places it might be nice to isolate the base query somewhere...

```python
def _base_query(session):
    return session.query(models.Annotation).filter_by(deleted=False)
```

I think the only wrinkle with that approach is reproducing your first query. Luckily I think that should be pretty straightforward, using `sqlalchemy.orm.load_only`:

```python
updated = (_base_query(self.session)
           .options(load_only('updated'))
           .execution_options(...)
           ...)
```

etc.

h/indexer/__init__.py
```diff
@@ -0,0 +1,12 @@
+# -*- coding: utf-8 -*-
```
nickstenning: Probably worth correcting the commit message to say "package" rather than "module".

h/indexer/__init__.py
```diff
@@ -0,0 +1,12 @@
...
+
+from h.indexer.reindexer import reindex
```
nickstenning: It doesn't look like any refers to "h.indexer.reindex", so this (and the `__all__`) probably aren't needed.

chdorner: The [CLI code](https://github.com/hypothesis/h/blob/334c5b33b8ea80c6ef01fffb3f793c0faae28459/h/cli/commands/search.py#L31) is.

src/memex/search/index.py
chdorner: Done (not pushed)

h/indexer/__init__.py
nickstenning: Ok!

PR #4251, opened 2016-12-21
Fix reindexing's unholy memory consumption

nickstenning:
The reindexing process started consuming slightly ridiculous quantities (GB) of memory, because in b0914fe we switched from loading a list of all the updated times of annotations to loading a list of all Annotation objects, albeit with most column values deferred.

While switching back to the prior approach would fix the problem for now, it is still O(N) in the number of annotations, just with a smaller constant factor.

This commit fixes the problem by switching to an altogether different way of calculating the windows we use to stream all annotations. This version is adapted from a recipe supplied by the SQLAlchemy maintainers:

  https://bitbucket.org/zzzeek/sqlalchemy/wiki/UsageRecipes/WindowedRangeQuery

This has the advantage of doing the window calculations on the server, so keeps the memory in use much more strongly bounded.

### DIFF
--- a/src/memex/search/index.py
+++ b/src/memex/search/index.py
@@ -3,20 +3,22 @@
 
 from __future__ import unicode_literals
 
-import itertools
 import logging
 from collections import namedtuple
 
-import elasticsearch
+import sqlalchemy as sa
 from elasticsearch import helpers as es_helpers
-from sqlalchemy.orm import load_only, subqueryload
+from sqlalchemy.orm import subqueryload
 
 from memex import models
 from memex import presenters
-from memex._compat import xrange
 from memex.events import AnnotationTransformEvent
+from memex.util.query import column_windows
 
 log = logging.getLogger(__name__)
+
+ES_CHUNK_SIZE = 100
+PG_WINDOW_SIZE = 2000
 
 
 class Window(namedtuple('Window', ['start', 'end'])):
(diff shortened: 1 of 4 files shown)


tests/memex/conftest.py
```diff
@@ -17,6 +17,7 @@
 from sqlalchemy.orm import sessionmaker
 
 from memex import db
+from memex import models  # noqa: ensure that model base class is set
```
nickstenning: This is here so that if you run a single memex test module that *doesn't* depend on models, you can still use the database when the `db_engine` fixture calls `init`. `query_test.py`, added in this commit, is an example of a test module which uses the database but not the models.

nickstenning:
For reference: on master reindexing quickly consumes many GB of memory on my machine:
 
screen shot 2016-12-21 at 10 04 45

(This is with only a few 1000 annotations reindexed).

And here's the memory usage on this branch after indexing more than 400K annotations:

screen shot 2016-12-21 at 12 36 37

src/memex/search/index.py
```diff
@@ -146,43 +150,35 @@ def _prepare(self, annotation):
...
 
-    def _eager_loaded_query(self):
-        return self._base_query().options(
-            subqueryload(models.Annotation.document).subqueryload(models.Document.document_uris),
-            subqueryload(models.Annotation.document).subqueryload(models.Document.meta)
-        )
+def _filtered_annotations(session, ids):
+    annotations = (_eager_loaded_annotations(session)
+                   .execution_options(stream_results=True)
+                   .filter(models.Annotation.id.in_(ids)))
```
chdorner: I think this needs to filter out deleted annotations as well, since the batch indexer can be called with a list of annotation ids. We should prevent that from indexing deleted annotations.

nickstenning: Oh, actually, sorry, I guess it is, because I got rid of _base_query.

chdorner: > because I got rid of _base_query
Exactly 😄

chdorner:
Waiting for the tests to pass

PR #4253, opened 2016-12-21
Log indexing status

nickstenning:
Adds a small generator wrapper which reports the indexing progress as it goes.

_**N.B.** This is currently based on #4251, so will need a rebase after that merges._


chdorner:
Rebased

PR #4255, opened 2016-12-21
Remove special-case handling of deletes in streamer

nickstenning:
Until now it has been necessary to special-case the handling of "delete" activity when it flows through the streamer, because by the time the message gets to the streamer code, the annotation has been deleted.

As of e2e6dda this is no longer true: deleted annotations remain in the database for up to 70 minutes (a job runs every hour but has a 10-minute buffer, so an annotation deleted 9m59s before the job runs will have to wait until the following run), which should be plenty of time for the streamer to process the message.

This PR removes the special handling in `h.streamer.messages` and the support on the memex side in `memex.events.AnnotationEvent`.

### DIFF
--- a/h/streamer/messages.py
+++ b/h/streamer/messages.py
@@ -83,19 +83,12 @@ def handle_annotation_event(message, sockets, session):
     id_ = message['annotation_id']
     annotation = storage.fetch_annotation(session, id_)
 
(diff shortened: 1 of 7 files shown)


chdorner:
Tested this out locally and it works

PR #4256, opened 2016-12-21
Move services to new `h.services` package

chdorner:
There are two services that didn't move:

* ~`h.auth.services.AuthTicketService` - because it needs to be registered before we include authsanity (which happens when `h.auth` is included).~ I must have misread the error messages, it works.
* `memex.links.LinksService` - well, because it's memex

All other services have moved to `h.services`


tests/h/views/activity_test.py
```diff
@@ -376,24 +376,24 @@ def test_search_passes_the_annotation_count_to_the_template(self,
...
                                                                                          group,
-                                                                                         groups_service,
+                                                                                         group_service,
                                                                                          pyramid_request):
         pyramid_request.authenticated_user = group.members[-1]
         pyramid_request.feature.flags['total_shared_annotations'] = False
         controller.search()
 
-        assert groups_service.annotation_count.called == False
+        assert group_service.annotation_count.called == False
```
chdorner: Argh, this isn't even my fault! :)

chdorner:
Might be worth waiting for #4247, depending on how close that one is to being merged.

nickstenning:
Hmm, although I think you might still need to make a change to the websocket server...

PR #4257, opened 2016-12-21
Remove form schemas used by non-AP email change form

robertknight:
**Follow up to https://github.com/hypothesis/h/pull/4254**

Remove the schema for the legacy email change form that was used by the
pre-Activity Pages version of the site.

### DIFF
--- a/h/views/accounts.py
+++ b/h/views/accounts.py
@@ -488,9 +488,6 @@ class AccountController(object):
         self.request = request
 
         email_schema = schemas.EmailChangeSchema().bind(request=request)
-        save_email_label = _('Save')
-        save_password_label = _('Save')
-
         password_schema = schemas.PasswordChangeSchema().bind(request=request)
 
         # Ensure deform generates unique field IDs for each field in this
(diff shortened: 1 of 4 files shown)


h/views/accounts.py
```diff
@@ -487,15 +487,9 @@ class AccountController(object):
...
-
-        if request.feature('activity_pages'):
-            email_schema = schemas.EmailChangeSchema().bind(request=request)
-            save_email_label = _('Save')
-            save_password_label = _('Save')
-        else:
-            email_schema = schemas.LegacyEmailChangeSchema().bind(request=request)
+        email_schema = schemas.EmailChangeSchema().bind(request=request)
+        save_email_label = _('Save')
+        save_password_label = _('Save')
```
chdorner: We probably don't need these to variables anymore and can just pass `_('Save')` to the buttons argument in the two `create_form` calls further down.

robertknight:
> Looks good. Just one small change that was probably done when we added the feature flag which should be reverted.

Done. See https://github.com/hypothesis/h/pull/4257/commits/95c3bcbeb478037df29a611f4e77da70333916f7

chdorner:
Thanks. LGTM

PR #4259, opened 2016-12-21
Measure coverage of test code as well

nickstenning:
It can be very useful to see and check that the coverage of test code is as we expect. Without this feedback, it's easier for bugs like that fixed in https://github.com/hypothesis/h/pull/4258 to creep in.

While this does somewhat artificially inflate our coverage numbers, I think the value of the feedback is worthwhile.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -22,8 +22,8 @@ deps =
 passenv =
     TEST_DATABASE_URL
 commands =
-    h: coverage run --parallel --source h -m pytest {posargs:tests/h/}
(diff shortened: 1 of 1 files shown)


chdorner:
Is codecov correct here? 99.19% code coverage is probably a bit more than "somewhat artificially inflate our coverage numbers"

nickstenning:
Yep, you're right, I had `--source a --source b` rather than `--source a,b`, so it was reporting only the coverage for the tests.

PR #4263, opened 2016-12-22
Filter on readable groups based on request

chdorner:
This adds a new filter to all memex search queries which makes sure that we do a terms filter on the group field with all the groups that the user has access to, which are:

* `__world__`
* All groups the user is a member of
* All groups with `readable_by=ReadableBy.world`

This is part of hypothesis/product-backlog#95.

### DIFF
--- a/h/groups/__init__.py
+++ b/h/groups/__init__.py
@@ -1,4 +1,8 @@
 # -*- coding: utf-8 -*-
 
+from h.groups import search
+
+
 def includeme(config):
     config.memex_set_groupfinder('h.groups.util.fetch_group')
+    config.memex_add_search_filter(search.GroupAuthFilter)
--- a/h/groups/__init__.py
+++ b/h/groups/__init__.py
@@ -1,8 +1,6 @@
 # -*- coding: utf-8 -*-
 
-from h.groups import search
-
 
 def includeme(config):
     config.memex_set_groupfinder('h.groups.util.fetch_group')
-    config.memex_add_search_filter(search.GroupAuthFilter)
+    config.memex_add_search_filter('h.groups.search.GroupAuthFilter')
--- a/h/groups/search.py
+++ b/h/groups/search.py
@@ -1,0 +1,29 @@
+# -*- coding: utf-8 -*-
+
+from __future__ import unicode_literals
+
+from h.models import Group
+from h.models.group import ReadableBy
+
+
+class GroupAuthFilter(object):
+    """
+    A memex search filter that filters out groups the request isn't authorized to read.
+    """
+
+    def __init__(self, request):
+        self.authenticated_user = request.authenticated_user
+        self.session = request.db
+
+    def __call__(self, _):
+        groups = set(['__world__'])
+
+        if self.authenticated_user:
+            for group in self.authenticated_user.groups:
+                groups.add(group.pubid)
+
+        world_readable_groups = self.session.query(Group.pubid).filter_by(readable_by=ReadableBy.world)
+        for group in world_readable_groups:
+            groups.add(group[0])
+
+        return {'terms': {'group': list(groups)}}
--- a/h/groups/search.py
+++ b/h/groups/search.py
@@ -2,9 +2,6 @@
 
 from __future__ import unicode_literals
 
-from h.models import Group
-from h.models.group import ReadableBy
-
 
 class GroupAuthFilter(object):
     """
@@ -14,16 +11,8 @@ class GroupAuthFilter(object):
     def __init__(self, request):
         self.authenticated_user = request.authenticated_user
         self.session = request.db
+        self.group_service = request.find_service(name='group')
 
     def __call__(self, _):
-        groups = set(['__world__'])
-
-        if self.authenticated_user:
-            for group in self.authenticated_user.groups:
-                groups.add(group.pubid)
-
-        world_readable_groups = self.session.query(Group.pubid).filter_by(readable_by=ReadableBy.world)
-        for group in world_readable_groups:
-            groups.add(group[0])
-
-        return {'terms': {'group': list(groups)}}
+        groups = self.group_service.groupids_readable_by(self.authenticated_user)
+        return {'terms': {'group': groups}}
--- a/h/services/group.py
+++ b/h/services/group.py
@@ -2,8 +2,10 @@
 
 from functools import partial
 
+import sqlalchemy as sa
+
 from h import session
-from h.models import Annotation, Group
+from h.models import Annotation, Group, User
 from h.models.group import JoinableBy, ReadableBy, WriteableBy
 
 GROUP_ACCESS_FLAGS = {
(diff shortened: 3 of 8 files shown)


h/groups/__init__.py
```diff
@@ -1,4 +1,8 @@
 # -*- coding: utf-8 -*-
 
+from h.groups import search
+
+
 def includeme(config):
     config.memex_set_groupfinder('h.groups.util.fetch_group')
+    config.memex_add_search_filter(search.GroupAuthFilter)
```
chdorner: Since this line is very important, I feel that it should be tested somehow. But there is no good way of putting fixture data into the search index for functional tests. What are our alternative ways of ensuring that we always set this filter?

tests/h/services/group_test.py
```diff
@@ -175,6 +175,17 @@ def test_annotation_count_returns_count_of_shared_annotations_for_group(self, db
 
         assert svc.annotation_count(group.pubid) == 3
 
+    def test_world_readable_pubids(self, db_session, factories, users):
+        # Groups that are members readable - should not be returned
+        [factories.Group(readable_by=ReadableBy.members) for _ in xrange(2)]
+        # Groups that are world readable - should be returned
+        world_readable = [factories.Group(readable_by=ReadableBy.world) for _ in xrange(3)]
```
chdorner: I guess I can import `h._compat.xrange`, but doesn't this imply that Hound runs flake8 with Python 3?

h/groups/__init__.py
nickstenning: I'd suggest we unit test this the same way `pyramid_services` suggests you test service registration: https://github.com/mmerickel/pyramid_services#testing-examples.

Specifically: add a request method and configuration directive `memex_get_search_filters`, and then test the includeme explicitly:

```python
def test_includeme_adds_groupauthfilter(pyramid_config):
    pyramid_config.include('h.groups')
    
    assert search.GroupFilter in pyramid_config.memex_get_search_filters()
```

(And then we can use that (in request method form) in `memex.search.core.default_querybuilder`, too.)

h/groups/search.py
```diff
@@ -0,0 +1,23 @@
...
+        self.group_service = request.find_service(name='group')
+
+    def __call__(self, _):
+        groups = set(['__world__'])
+
+        if self.authenticated_user:
+            for group in self.authenticated_user.groups:
+                groups.add(group.pubid)
+
+        for pubid in self.group_service.world_readable_pubids():
```
nickstenning: This is ok, and we can definitely merge it, but it would seem more logical for me for the group service to expose a `.groups_readable_by(user)` method and encapsulate the logic in the service rather than here. That way you could also fetch all the groups in one go:

```python
def groups_readable_by(user, query_options=None):
    readable = (Group.readable_by = ReadableBy.world)

    if user is not None:
        readable_member = sa.and_(Group.readable_by = ReadableBy.members, Group.members.has(user))
        readable = sa.or_(readable, readable_member)

    query = self.session.query(Group).filter(readable)

    if query_options is not None:
        query = query.options(*query_options)

    return query
```

...and then in the filter...

```python
groups = group_service.groups_readable_by(self.authenticated_user, 
                                          query_options=[sa.orm.load_only('pubid')])
```

etc.

tests/h/services/group_test.py
nickstenning: Yeah, it does, unfortunately. But hey, it works for us because yes, you should include xrange from the compat module.

h/groups/__init__.py
nickstenning: Also, to avoid more circular import dependencies, it might be nice to add a `config.maybe_dotted` to `memex_add_search_{filter,matcher}` while you're here, so you can get rid of the import in this module.

h/groups/search.py
```diff
@@ -0,0 +1,23 @@
+# -*- coding: utf-8 -*-
+
+from __future__ import unicode_literals
+
+
+class GroupAuthFilter(object):
```
nickstenning: It would be good to have a docstring here to explain what this filter does.

h/groups/search.py
```diff
@@ -0,0 +1,23 @@
...
+    def __init__(self, request):
+        self.auth_domain = request.auth_domain
```
nickstenning: Is this used?

h/groups/search.py
nickstenning: I'm not entirely sure whether `Group.members.has(user)` will generate sensible SQL here, but there are a couple of equivalent ways we could do this.

nickstenning: I think it should probably be something like `Group.members.any(User.id == user.id)`, actually.

nickstenning: No, just had a play and it should actually be `m.Group.members.contains(user)`, which avoids a subquery by just joining against the membership table.

chdorner: With my local database (dump from a few months ago) using `contains` returns 29'460 results, which then SQLAlchemy processes and returns 10 groups. I think the subquery with `any` might be the best way.

Also, we need to add an index to `group.readable_by`.

nickstenning: Huh? Are you sure?

```python
In [4]: readable_member = sa.and_(m.Group.readable_by == ReadableBy.members, m.Group.members.contains(user))

In [5]: print readable_member
"group".readable_by = :readable_by_1 AND "group".id = user_group_1.group_id AND :param_1 = user_group_1.user_id
```

Unless you're a member of 29,460 groups, I can't see how that query could return that many rows?

chdorner: After profiling these options we're going for `Group.members.any` as it's well fast enough.

h/groups/search.py
chdorner: Done

tests/h/services/group_test.py
```diff
@@ -9,6 +9,7 @@
 from h.models.group import JoinableBy, ReadableBy, WriteableBy
 from h.services.group import GroupService
 from h.services.group import groups_factory
+from h._compat import xrange
```
chdorner: 😭

h/services/group.py
```diff
@@ -102,6 +104,24 @@ def annotation_count(self, pubid):
...
+        If the passed-in user is ``None``, this returns the list of
+        world-readable groups.
+        """
+        readable = (Group.readable_by == ReadableBy.world)
+
+        if user is not None:
+            readable_member = sa.and_(Group.readable_by == ReadableBy.members, Group.members.any(User.id == user.id))
+            readable = sa.or_(readable, readable_member)
+
+        ids = [record[0] for record in self.session.query(Group.pubid).filter(readable)]
```
nickstenning: Small thing, but maybe use `record.pubid` here instead of `record[0]`? It's a little more readable.

chdorner: Oh, I didn't know about this! I always thought it just returns a normal tuple.

Fixed.

tests/h/groups/__init___test.py
```diff
@@ -0,0 +1,23 @@
...
+    includeme(pyramid_config)
+    assert GroupAuthFilter in pyramid_config.memex_get_search_filters()
+
+
+@pytest.fixture
+def pyramid_config(pyramid_config):
+    pyramid_config.add_directive('memex_set_groupfinder', lambda c, f: None)
+
+    filters = []
+    pyramid_config.add_directive('memex_add_search_filter',
```
nickstenning: Oh, right. I think I'd probably `pyramid_config.include('memex')` in preference to this, but no need to change that at the moment.

chdorner: It would have required too much knowledge about memex, because memex tried to initialise the search index which failed when you don't have ES running on localhost:9200

nickstenning: Guh.

nickstenning: (Which means "Guh, memex" rather than "Guh, Christof", to be clear... 😝)

nickstenning:
LGTM.

PR #4265, opened 2016-12-22
Add index to group.readable_by column

chdorner:


h/migrations/versions/e10ce4472966_add_index_to_group_readable_by.py
```diff
@@ -0,0 +1,23 @@
...
+
+from alembic import op
+
+
+revision = 'e10ce4472966'
+down_revision = 'f0f42ffaa27d'
+
+
+def upgrade():
+    op.create_index(op.f('ix__group__readable_by'), 'group', ['readable_by'], unique=False)
```
nickstenning: As a general matter of hygiene we should be using `CREATE INDEX CONCURRENTLY` for this. Have a look at previous migrations which added indices.

(Otherwise we lock the group table until the index is built).

nickstenning:
LGTM.

PR #4266, opened 2016-12-22
Remove non-Activity Pages version of logged out accounts forms

robertknight:
Also remove uses of the `activity_pages` flag in the navbar and masthead, since all pages using them have now had their own use of the `activity_pages` flag removed.

### DIFF
--- a/tests/functional/test_accounts.py
+++ b/tests/functional/test_accounts.py
@@ -18,7 +18,7 @@ class TestAccountSettings(object):
 
         res = email_form.submit().follow()
 
-        assert res.body.startswith('<!DOCTYPE html>')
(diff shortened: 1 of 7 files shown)


tests/functional/test_accounts.py
```diff
@@ -18,7 +18,7 @@ def test_submit_email_form_without_xhr_returns_full_html_page(self, app):
 
         res = email_form.submit().follow()
 
-        assert res.body.startswith('<!DOCTYPE html>')
+        assert res.text.startswith('<!DOCTYPE html>')
```
robertknight: This is needed because these tests now fetch the Activity Pages version of the site, which has some non-ASCII characters in parts of the template (eg. the navbar).

PR #4270, opened 2016-12-22
Remove styles for pre-Activity Pages version of the website

robertknight:
The first commit removes the style bundle for the previous version of the website ("site.scss") and the styles in the partials directory that were only used by that bundle.

The second commit renames the site-v2.scss bundle to site.scss, since there is now only the new version of the website left.

There are still a number of pages using other legacy style bundles - the 403/404/5xx error pages, home page and `/docs/help` page.


sean-roberts:
LGTM. code clean up looks fine and I did some smoke testing of various pages and found no issues.

PR #4280, opened 2017-01-06
Collapse document expansion from base of expanded form.

sheetaluk:
Collapsing a document which has several annotations means having to
scroll back up the header and then clicking it. This commit gives the user
the ability to collapse an open document with annotations, from the
bottom also.

Fixes https://github.com/hypothesis/product-backlog/issues/79

### DIFF
--- a/h/static/styles/partials-v2/_search.scss
+++ b/h/static/styles/partials-v2/_search.scss
@@ -230,6 +230,8 @@ $annotation-card-width: 500px;
 // Card displaying stats about a group of annotations in search results
 .search-bucket-stats {
   @include font-normal;
+  display: flex;
+  flex-direction: column;
   margin-left: 30px;
   word-wrap: break-word;
   min-width: 0;  // Without this min-width stats containing really
@@ -277,6 +279,22 @@ $stats-icon-column-width: 20px;
   top: 1px;
   height: 9px;
   width: 9px;
+}
+
+.search-bucket-stats__collapse-view {
+  @include reset-button;
+  color: $grey-6;
+  margin-bottom: 15px;
+  text-align: left;
+}
+
+.search-bucket-stats__collapse-view-icon {
+  margin-right: 4px;
+  width: 12px;
+}
+
+.search-bucket-stats__collapse-view:focus {
+  outline:0;
 }
 
 // On large tablets and below, display bucket titles beneath rather than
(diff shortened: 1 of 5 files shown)


h/templates/activity/search.html.jinja2
```diff
@@ -76,6 +76,11 @@
            target="_blank">{{ pretty_link(bucket.uri) }}</a>
     </div>
   {% endif %}
+  <div class="u-stretch">
+  </div>
+  <a data-ref="collapseView" class="search-bucket-stats__collapse-view">
```
sean-roberts: I would not use anchor elements without hrefs for this. Use `<button>`s when you need a non document linking click target.

h/static/styles/partials-v2/_search.scss
```diff
@@ -279,6 +281,20 @@ $stats-icon-column-width: 20px;
   width: 9px;
 }
 
+.search-bucket-stats-filler {
+  flex-grow: 1;
+}
+
+.search-bucket-stats__collapse-view {
+  @include link;
+  @include link-underline(dotted);
```
sean-roberts: There is a mock here https://github.com/hypothesis/product-backlog/issues/79#issuecomment-267441967 and it seems that the implemented design does not match it. Note, this comment: https://github.com/hypothesis/product-backlog/issues/79#issuecomment-267371896 has the icon link as well.

ajpeddakotla:
@sheetaluk One question here - for UX purposes, can we have the collapse indicator show only when there’s enough height for it to make sense? For example, if a user has only one card open, it doesn't make much sense to show the collapse indicator. 

But if a user has more than one card open (n>1), we should show the indicator.

sheetaluk:
@ajpeddakotla Even with one card, it is possible for the list of tags and annotators in the sidebar to grow long enough to need the "collapse view" functionality from the bottom. I would suggest to keep it the way it is for now.

dwhly:
@sheetaluk can you screenshot what it looks like now w/ the shortest possible card?  ty!

sheetaluk:
@dwhly 

Shortest:
https://www.dropbox.com/s/ewsii78dur4kv0n/Screenshot%202017-01-10%2011.48.52.png?dl=0

Card with some tags and 1 annotator:
https://www.dropbox.com/s/2z5tbzh6gn2ot5b/Screenshot%202017-01-10%2011.49.38.png?dl=0

dwhly:
Thank you! 👍   cc @jeremydean

sean-roberts:
LGTM.

PR #4284, opened 2017-01-09
Update Elasticsearch plugin link in dev setup docs

fatbusinessman:
The previous link led to a 404 page. As far as I can tell, this new link is entirely equivalent, and the Elasticsearch developers just changed their URLs.


nickstenning:
:+1:

PR #4285, opened 2017-01-09
Use annotation stats service in admin user view

chdorner:
There was a bug with the old code which counted deleted annotations,
instead of fixing that we're now just using the annotation stats service
which does the query for us.


nickstenning:
LGTM.

PR #4286, opened 2017-01-09
Rework AuthFilter

chdorner:
We have moved the filtering on the groups into the new groups filter
that lives in `h` (#4263). The memex auth filter will now only check for the
`shared` field. It filters out private annotations of other users, if
the request is done on behalf of a logged-in user. Otherwise it filters
out all private annotations.

This is part of hypothesis/product-backlog#95.

### DIFF
--- a/src/memex/search/query.py
+++ b/src/memex/search/query.py
@@ -111,30 +111,29 @@ class AuthFilter(object):
     """
     A filter that selects only annotations the user is authorised to see.
 
-    Only annotations where the 'read' permission contains one or more of the
-    user's effective principals will pass through this filter.
+    Only annotations that are shared, or private annotations made
+    by the logged-in user will pass through this filter.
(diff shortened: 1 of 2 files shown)


src/memex/search/query.py
```diff
@@ -111,30 +111,32 @@ class AuthFilter(object):
...
-        return {'terms': {'permissions.read': principals}}
+        public_filter = {'term': {'shared': True}}
+
+        userid = self.request.authenticated_userid
+        if userid is None:
+            return public_filter
+
+        return {'or': [
+            public_filter,
+            {'and': [
```
nickstenning: There's no particular need for this `and` here, because the `{'term': {'shared': False}}` is implied. It would be just as clear to have `{'or': [public_filter, {'term': {'user': userid}}]}`.

PR #4289, opened 2017-01-10
Fix link syntax in ADR 002

fatbusinessman:
The links appear to be written using Markdown’s link syntax, which doesn't display as intended in `.rst` files.


nickstenning:
👌

PR #4293, opened 2017-01-11
Refactor memex groupfinder

chdorner:
This changes the groupfinder from a configurable callable to a service which allows us greater flexibility how the finder is initialized and how it can access other data (`auth_domain` for example).

### DIFF
--- a/h/services/groupfinder.py
+++ b/h/services/groupfinder.py
@@ -1,0 +1,28 @@
+# -*- coding: utf-8 -*-
+
+from __future__ import unicode_literals
+
+from zope.interface import implementer
+
+from memex.interfaces import IGroupService
+
+from h import models
+from h.groups.util import WorldGroup
+
+
+@implementer(IGroupService)
+class GroupfinderService(object):
+    def __init__(self, session, auth_domain):
+        self.session = session
+        self.auth_domain = auth_domain
+
+    def find(self, id_):
+        if id_ == '__world__':
+            return WorldGroup(self.auth_domain)
+
+        # TODO: caching
+        return self.session.query(models.Group).filter_by(pubid=id_).one_or_none()
+
+
+def groupfinder_service_factory(context, request):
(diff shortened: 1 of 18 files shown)


src/memex/storage.py
```diff
@@ -75,7 +74,7 @@ def fetch_ordered_annotations(session, ids, query_processor=None):
     return anns
 
 
-def create_annotation(request, data):
+def create_annotation(request, data, group_service):
```
nickstenning: The docstring needs updating here.

src/memex/resources.py
```diff
@@ -16,17 +16,19 @@ def __getitem__(self, id):
         annotation = storage.fetch_annotation(self.request.db, id)
         if annotation is None:
             raise KeyError()
-        return AnnotationResource(self.request, annotation)
+
+        group_service = self.request.find_service(IGroupService)
```
nickstenning: Nice. The division between `AnnotationResourceFactory` and `AnnotationResource` means that the latter is easily usable on its own (i.e. without a request.)

h/services/groupfinder.py
```diff
@@ -0,0 +1,28 @@
...
+from zope.interface import implementer
+
+from memex.interfaces import IGroupService
+
+from h import models
+from h.groups.util import WorldGroup
+
+
+@implementer(IGroupService)
+class GroupfinderService(object):
```
nickstenning: It might be worth adding a note here to say something like "Ideally this would be called the GroupService to match nomenclature in memex. FIXME: rename/split existing GroupService and rename this."

src/memex/resources.py
chdorner: Which was exactly the goal to make the [streamer code work](https://github.com/hypothesis/h/pull/4299/files#diff-23fa267bf8100c9a7401f260e0e347d4R150).

PR #4295, opened 2017-01-11
Fix renaming user to equivalent username

seanh:
It was not possible to rename a user, e.g. "bob.smith", to a new
username that reduces to the same uid, e.g. "Bob.Smith", because the
rename user service would think that the username "Bob.Smith" already
exists.

Fix it to allow renaming "bob.smith" to "Bob.Smith" but *don't* allow
renaming a different user, e.g. "fred", to a username such as
"Bob.Smith" that's equivalent to the other "bob.smith" user's username.

### DIFF
--- a/h/admin/views/users.py
+++ b/h/admin/views/users.py
@@ -86,7 +86,7 @@ def users_rename(request):
 
     try:
         svc = request.find_service(name='rename_user')
-        svc.check(new_username, request.auth_domain)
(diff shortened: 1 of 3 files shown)


seanh:
This means that you can now harmlessly rename "fred" to "fred" (on master you would get a "Another user already has the username "fred"" error message)

nickstenning:
LGTM.

PR #4297, opened 2017-01-12
Use unanalyzed `user_raw` field for AuthFilter

chdorner:
The `user` field in the search index is lowercased, while the AuthFilter
does not lower-case the user's id when it passes it in. This is
preventing users from seeing their own private annotations.

The fix is to use the unanalyzed `user_raw` field which closes ways of
exploiting this behaviour by manging to create users with the same
username as another, but different upper- and lowercase letters and thus
managing to see another users' private annotations.

### DIFF
--- a/src/memex/search/query.py
+++ b/src/memex/search/query.py
@@ -132,7 +132,7 @@ class AuthFilter(object):
 
         return {'or': [
             public_filter,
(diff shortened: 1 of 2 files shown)


nickstenning:
LGTM.

PR #4299, opened 2017-01-12
Annotation render read permission

chdorner:
This changes the memex annotation presenters to get an `memex.resources.AnnotationResource` passed-in instead of just a `memex.models.Annotation`, allowing us to get the principals allowed to read the annotation (which on the annotation resource takes the group ACL into account), and thus overriding the read permission in the JSON output to `__world__` when an annotation is shared in a group that is world readable.

### DIFF
--- a/h/streamer/messages.py
+++ b/h/streamer/messages.py
@@ -10,11 +10,15 @@ from h.realtime import Consumer
 from memex import presenters
 from memex import storage
 from memex.links import LinksService
+from memex.resources import AnnotationResource
 from h.auth.util import translate_annotation_principals
 from h.services.nipsa import NipsaService
+from h.services.groupfinder import GroupfinderService
 from h.streamer import websocket
 import h.sentry
 import h.stats
+
+from h._compat import text_type
 
 log = logging.getLogger(__name__)
 
@@ -55,7 +59,7 @@ def process_messages(settings, routing_key, work_queue, raise_error=True):
         raise RuntimeError('Realtime consumer quit unexpectedly!')
 
 
-def handle_message(message, session, topic_handlers):
+def handle_message(message, settings, session, topic_handlers):
     """
     Deserialize and process a message from the reader.
 
@@ -76,10 +80,10 @@ def handle_message(message, session, topic_handlers):
     # to stop connections being added or dropped during iteration, and if that
     # happens Python will throw a "Set changed size during iteration" error.
     sockets = list(websocket.WebSocket.instances)
-    handler(message.payload, sockets, session)
+    handler(message.payload, sockets, settings, session)
 
 
-def handle_annotation_event(message, sockets, session):
+def handle_annotation_event(message, sockets, settings, session):
     id_ = message['annotation_id']
     annotation = storage.fetch_annotation(session, id_)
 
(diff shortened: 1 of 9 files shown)


h/streamer/messages.py
```diff
@@ -140,8 +147,8 @@ def _generate_annotation_event(message, socket, annotation, user_nipsad):
     base_url = socket.registry.settings.get('h.app_url',
                                             'http://localhost:5000')
     links_service = LinksService(base_url, socket.registry)
-    serialized = presenters.AnnotationJSONPresenter(annotation,
-                                                    links_service).asdict()
+    resource = AnnotationResource(annotation, group_service)
```
nickstenning: I think you could create the `GroupfinderService` here (`settings` is available as `socket.registry.settings`), which would obviate the need to pass settings into all the functions above.

src/memex/presenters.py
```diff
@@ -6,10 +6,13 @@
 import collections
 import copy
 
+from pyramid import security
+
 
 class AnnotationBasePresenter(object):
-    def __init__(self, annotation, links_service):
-        self.annotation = annotation
+    def __init__(self, annotation_resource, links_service):
```
nickstenning: This seems a bit odd to me. Or, to be more accurate, the resulting call pattern looks odd:

```python
annotation = storage.fetch_annotation(id_)
link_service = request.find_service(...)
group_service = request.find_service(...)
resource = AnnotationResource(annotation, group_service)
presenter = AnnotationJSONPresenter(resource, link_service)
```

I think it would be more natural if the presenter accepted the group_service and constructed the resource itself:

```python
annotation = storage.fetch_annotation(id_)
link_service = request.find_service(...)
group_service = request.find_service(...)
presenter = AnnotationJSONPresenter(resource, group_service, link_service)
```

src/memex/presenters.py
```diff
@@ -106,7 +128,6 @@ def asdict(self):
             'tags': self.tags,
             'tags_raw': self.tags,
             'group': self.annotation.groupid,
-            'permissions': self.permissions,
```
nickstenning: Yay!

nickstenning: Can we add a FIXME next to the relevant item in the mapping so we remember to remove it?

h/streamer/messages.py
chdorner: We aren't taking advantage of the caching mechanism of the group service if we create it here, we would effectively load the same group once for each connected client, instead of just loading and caching it once, and then reusing it.

src/memex/presenters.py
chdorner: Isn't it much nicer to change the presenter to always use the resource rather than the model? This way the presenter does not need to know anything about the group service, it just uses the ACL of the passed-in resource.

src/memex/presenters.py
chdorner: Done

h/streamer/messages.py
nickstenning: Caching the group object within the handling of a single message is fine, but caching between messages is bound to end in disaster -- if someone changes the sharing settings of a group that's cached we'll have to restart the websocket server before it starts calculating permissions correctly again.

src/memex/presenters.py
nickstenning: I'm not sure.

It doesn't do anything with the group service now. But it would do if you moved the creation of the resource object into it.

At the moment I think it's cleaner for the presenter to have a small set of services it depends on, otherwise you end up in weird situations like the one you have in the `update_annotation` view where you have a context object which is an `AnnotationResource`, but it's the wrong one, so you now have to create a new one just to pass it into the presenter.

h/streamer/messages.py
chdorner: As discussed in [Slack](https://hypothes-is.slack.com/archives/backend/p1484318794001939), this does not cache groups in-between annotation events, but when generating the responses for each individual clients for one annotation events.

nickstenning:
I'm going to merge this, but perhaps spend some time looking into the refactorings we discussed, as the streamer in particular is getting a bit unmanageable.

PR #4300, opened 2017-01-13
Use the best DocumentURI for Document.web_uri

seanh:
Denormalize the "best" DocumentURI.uri that we have into the
Document.web_uri field, instead of just taking the first one.

This fixes an issue that, depending on what order the DocumentURIs are
processed in, a rel-shortlink URI (for example) might be used instead of
a self-claim or canonical URI. We want to display the best available URI
for a document, a self-claim or canonical one if possible, to users.

The first time a given document is annotated consider all of the
DocumentURIs that we've generated from the document metadata sent by the
client and:

- Use a self-claim URI if there is one
- Use a rel-canonical URI if there's no self-claim
- Use any other http(s) URI if there's no self-claim or canonical
- Set web_uri to None if there are not http(s) URIs for the document

I'll need to send a follow-up PR to update the existing web_uris in the production db.

### DIFF
--- a/src/memex/models/document.py
+++ b/src/memex/models/document.py
@@ -58,9 +58,6 @@ class Document(Base, mixins.Timestamps):
         Set self.web_uri to None if there's no http(s) DocumentURIs.
 
         """
-        if self.web_uri is not None:
-            return
-
         def first_http_url(type_=None):
             """
             Return this document's first http(s) URL of the given type.
--- a/src/memex/models/document.py
+++ b/src/memex/models/document.py
@@ -32,7 +32,7 @@ class Document(Base, mixins.Timestamps):
     #: The denormalized value of the first DocumentMeta record with type title.
     title = sa.Column('title', sa.UnicodeText())
 
-    #: The denormalized value of the first http(s) DocumentURI
+    #: The denormalized value of the "best" http(s) DocumentURI for this Document.
     web_uri = sa.Column('web_uri', sa.UnicodeText())
 
     # FIXME: This relationship should be named `uris` again after the
@@ -48,6 +48,41 @@ class Document(Base, mixins.Timestamps):
 
     def __repr__(self):
         return '<Document %s>' % self.id
+
+    def update_web_uri(self):
+        """
+        Update the value of the self.web_uri field.
+
+        Set self.web_uri to the "best" http(s) URL from self.document_uris.
+
+        Set self.web_uri to None if there's no http(s) DocumentURIs.
+
+        """
+        if self.web_uri is not None:
+            return
+
+        def first_http_url(type_=None):
+            """
+            Return this document's first http(s) URL of the given type.
+
+            Return None if this document doesn't have any http(s) URLs of the
+            given type.
+
+            If no type is given just return this document's first http(s)
+            URL, or None.
+
+            """
+            for document_uri in self.document_uris:
+                uri = document_uri.uri
+                if type_ is not None and document_uri.type != type_:
+                    continue
+                if urlparse.urlparse(uri).scheme not in ['http', 'https']:
+                    continue
+                return document_uri.uri
+
+        self.web_uri = (first_http_url(type_='self-claim') or
+                        first_http_url(type_='rel-canonical') or
+                        first_http_url())
 
     @classmethod
     def find_by_uris(cls, session, uris):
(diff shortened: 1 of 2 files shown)


tests/memex/models/document_test.py
```diff
@@ -833,6 +893,43 @@ def test_it_saves_all_the_document_uris(self,
...
 
+    def test_it_sets_document_web_uri(self,
+                                      annotation,
+                                      best_document_uri_from,
+                                      Document,
+                                      factories,
+                                      session):
+        document_ = mock.Mock(web_uri=None)
+        Document.find_or_create_by_uris.return_value.count.return_value = 1
+        Document.find_or_create_by_uris.return_value.first.return_value = document_
```
seanh: This mocking is ugly but it's how the existing tests in this class work

src/memex/models/document.py
```diff
@@ -409,6 +405,34 @@ def merge_documents(session, documents, updated=None):
     return master
 
 
+def best_document_uri_from(document_uris):
```
seanh: I guess just `best_uri_from(docuris)` might be a better name

nickstenning: I'm not sure `best` is particularly helpful without qualification. Best for what? Identity? Location?

src/memex/models/document.py
```diff
@@ -409,6 +405,34 @@ def merge_documents(session, documents, updated=None):
...
+
+        if urlparse.urlparse(uri).scheme in ['http', 'https']:
+            if type_ == 'self-claim':
+                return uri
+            elif type_ == 'rel-canonical':
+                canonical_uri = canonical_uri or uri
+            else:
+                other_uri = other_uri or uri
+
+    return canonical_uri or other_uri
```
nickstenning: This seems ok, but personally I find this code a little hard to follow. Having the return statements at two different levels is a bit confusing. Do you think it might be clearer to do something like this?

```python
def _first_url_of_type(uris, type_=None):
    for u in uris:
        if type_ is not None and u.type != type_:
            continue
        if urlparse.urlparse(u.uri).scheme not in ['http', 'https']:
            continue
        return u.uri

return (_first_url_of_type(document_uris, 'self-claim') or 
        _first_url_of_type(document_uris, 'rel-canonical') or
        _first_url_of_type(document_uris))
```

src/memex/models/document.py
```diff
@@ -463,13 +487,19 @@ def update_document_metadata(session,
...
+        document_uris.append(
+            create_or_update_document_uri(
+                session=session,
+                document=document,
+                created=created,
+                updated=updated,
+                **document_uri_dict)
+        )
+
+    if not document.web_uri:
```
nickstenning: I know this is consistent with the previous code, but shouldn't this denormalised field be updated every time the set of URIs is updated?

src/memex/models/document.py
seanh: I actually rejected this approach because it loops over the uris (up to) 3 times. But it's not likely to be a long list of uris so that's not likely to matter, so if you think it's clearer then sure I can change it

nickstenning: Why is looping over the URIs a problem? This only happens on write, right?

src/memex/models/document.py
```diff
@@ -281,15 +281,11 @@ def create_or_update_document_uri(session,
...
-    if not document.web_uri:
-        parsed = urlparse.urlparse(uri)
-        if parsed.scheme in ['http', 'https']:
-            document.web_uri = uri
-
     try:
         session.flush()
     except sa.exc.IntegrityError:
         raise ConcurrentUpdateError('concurrent document uri updates')
+    return docuri
```
seanh: Hmm, yeah I don't think so, I just didn't think of `document.document_uris`, I'll give it a try

src/memex/models/document.py
seanh: You mean that (effectively) web_uri would be updated every time a new annotation of this document is created (or updated probably)? I guess the downside would be potential confusion caused by document domain names changing (e.g. on activity pages), but overall I think we probably should update them yes. We have [the same problem with document titles](https://github.com/hypothesis/product-backlog/issues/127) as well.

I think this is a separate issue but it should be a very simple change to make so I can tag it on to this pr if you want..

src/memex/models/document.py
seanh: Yep, it only happens on write

seanh:
Gonna go for lunch now. I've made all the requested changes, but I think the tests can be deduplicated so will do that after lunch.

seanh:
Ok, all done @nickstenning

seanh:
Crumbs, actually the commit history got a bit messed up, hold on a sec...

seanh:
Ok fixed

PR #4301, opened 2017-01-13
Update setup instructions to bind ports locally

fatbusinessman:
By default, the `-p` flags to `docker run` open up ports on all interfaces on the host machine. Unless there's some clever Docker networking going on (which I don't think there is), this means that following the default instructions opens up Elasticsearch, RabbitMQ and Postgres servers on the developer's machine, all with default (or no) authentication settings.

Binding these ports to local connections only by default seems safer – albeit a little more verbose – and doesn't appear to break anything.

### DIFF
--- a/docs/developing/install/website.rst
+++ b/docs/developing/install/website.rst
@@ -97,9 +97,9 @@ installed on:
 
    .. code-block:: bash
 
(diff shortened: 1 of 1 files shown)


seanh:
Works here, well spotted

PR #4302, opened 2017-01-16
Pass the PYTEST_ADDOPTS env var to pytest

seanh:
Configure tox to pass the PYTEST_ADDOPTS environment variable through to
pytest. This enables developers to set their own PYTEST_ADDOPTS
environment variable containing their default options that they always
want to be passed to pytest, for example:

    export PYTEST_ADDOPTS="--verbose --exitfirst --failed-first --showlocals"

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -21,6 +21,7 @@ deps =
     h: -rrequirements.txt
 passenv =
     TEST_DATABASE_URL
+    PYTEST_ADDOPTS
 commands =
     h: coverage run --parallel --source h,tests/h -m pytest {posargs:tests/h/}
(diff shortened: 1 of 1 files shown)


tox.ini
```diff
@@ -21,6 +21,7 @@ deps =
     h: -rrequirements.txt
 passenv =
     TEST_DATABASE_URL
+    PYTEST_ADDOPTS
```
nickstenning: For consistency let's pass this through to all pytest environments (so, `functional` too).

seanh: Done

seanh:
I'm currently writing up some notes about pytest and h which I'll post somewhere. I'm not really sure where the best place to document things like this would be, but in the sphinx docs seems like as good a place as any to me

PR #4303, opened 2017-01-16
Run gunicorn with --workers 7 in dev

seanh:
Run gunicorn with the --workers=7 argument for the web process in
`make dev` and `hypothesis devserver`.

This fixes a problem that page loads would frequently take 20 seconds or
longer in development environments.

The hypothesis behind this is that browsers will (at least sometimes)
open up to 6 TCP connections to one domain at once even if they aren't
going to use all 6 right away, and sometimes this can block all of
gunicorn's workers. Giving gunicorn 7 workers seems to keep it out of
trouble. See Slack discussion starting here:

https://hypothes-is.slack.com/archives/backend/p1484241382001768

### DIFF
--- a/h/cli/commands/devserver.py
+++ b/h/cli/commands/devserver.py
@@ -79,7 +79,7 @@ def devserver(https, web, ws, worker, assets, beat):
         m.add_process('web',
                       'MODEL_CREATE_ALL=true '
                       'SEARCH_AUTOCONFIG=true '
-                      'gunicorn --reload --paste conf/development-app.ini %s' % gunicorn_args)
+                      'gunicorn --reload --paste conf/development-app.ini --workers 7 %s' % gunicorn_args)
 
     if ws:
         m.add_process('ws', 'gunicorn --reload --paste conf/development-websocket.ini %s' % gunicorn_args)


robertknight:
This solves the problem but each worker has a pretty hefty memory cost. That might be annoying for some developers.

On my system:

Base/"background" system memory usage: 2.0GB
After running `make dev` with this branch and waiting for a few seconds: 3.2GB
After running `make dev` with `--worker-class gevent` instead and waiting for a few seconds: 2.7GB

`gevent` solves the problem but has the downside of `--worker-class gevent` is that it doesn't match the production environment. Perhaps an alternative would be to run a proxy in front of the sync worker - which would mirror how `nginx` functions in production?

seanh:
Hmm, I don't think we should use gevent if it differs from prod, and the memory usage is almost as high anyway.

The proxy idea might be interesting but I don't know of anything off hand (although I suppose we could actually use dockerized nginx in dev...)

If no quick solution is apparent I think most dev systems can handle an extra 1.2GB of memory usage (for example I have 16GB of memory, well over half of which is always empty, I think 8GB would do), it's certainly less bad than 20-30 second page load times, so if nothing comes up maybe we should consider merging this?

robertknight:
> If no quick solution is apparent I think most dev systems can handle an extra 1.2GB of memory usage

Err, yikes. My laptop "only" has 6GB of RAM, so that's a _lot_ for me. This is equivalent to running another heavily loaded browser or a small VirtualBox VM.

> it's certainly less bad than 20-30 second page load times, so if nothing comes up maybe
> we should consider merging this for now?

Hmm, gevent is a much lighter-weight option though. Are there significant caveats with it in development? 75b69dd6460601487a1ea49f9dedde47a1a43346 gives the original rationale for switching to sync workers, but I think that only applies to the production environment which has higher load. Thoughts @nickstenning ?

> The proxy idea might be interesting but I don't know of anything off hand

Something like https://tinyproxy.github.io/ perhaps? I've had a brief look at using that as a reverse proxy with this config:

```
LogFile "tinyproxy.log"
MaxClients 10
MaxSpareServers 10
MinSpareServers 2
PidFile "tinyproxy.pid"
Port 5060
ReverseMagic Yes
ReverseOnly Yes
ReversePath "/" "http://localhost:5000/"
StartServers 2
```
The initial page load works but most links are incorrect because the `Host` header is not preserved when the proxy makes the upstream request.

robertknight:
Slack discussion: https://hypothes-is.slack.com/archives/backend/p1484653449002012

TL;DR - Running nginx in development might not be too hard to do and would reduce dev <-> production disparity, let's explore that option.

seanh:
I'm gonna close this for now. Here's a [Slack discussion of where we got to with nginx](https://hypothes-is.slack.com/archives/backend/p1484753267002314). TLDR: maybe it can be done but it certainly seems a bit difficult and time consuming to figure out how.

We've discovered that the extremely slow page load times only seem to happen in Chrome **with http**, [running h over SSL](http://h.readthedocs.io/en/latest/developing/ssl/) seems to avoid it (so far). So that's what I'm doing now

PR #4304, opened 2017-01-18
Add CLI commands for managing publisher accounts.

robertknight:
This adds additional CLI commands to make it easy to create all the pieces needed for publisher accounts:

1. `authclient add` and `authclient secret` to create new authorities and get their client ID and secret
2. `groups add_publisher_group` to create a new group with the appropriate permissions to serve as the main group for a publisher
3. An `--authority` argument to `users add`
 
Together, these commands can be used to create all the pieces needed to test out "publisher" accounts locally:

```sh
(h)  ~/h/r/h > ./bin/hypothesis --dev authclient add --name eLIFE --authority elifesciences.org
OAuth client for elifesciences.org created
Client ID: a0c56d00-df05-11e6-bf77-1bb7ff4be90f
Client Secret: jsfV4fkSlFpR70GHgVwgK48Va4vFDdenDk_kExvUnBs
(h)  ~/h/r/h > ./bin/hypothesis --dev user add --authority elifesciences.org --username admin --email admin@localhost --password admin
admin created
(h)  ~/h/r/h > ./bin/hypothesis --dev groups add_publisher_group --authority elifesciences.org --creator admin --name eLIFE
```

_nb. These changes are all in the one PR because it originally grouped all this functionality into one command. After discussion, we split up the pieces by type._


h/cli/commands/publisher.py
```diff
@@ -0,0 +1,73 @@
...
+def add(ctx, name, authority):
+    """
+    Create a new publisher account.
+
+    This consists of an authority, an admin user and the main group for the
+    publisher's annotations.
+    """
+    request = ctx.obj['bootstrap']()
+
+    # Create a new authority.
```
chdorner: Nitpick, but this comment should say "Create a new auth client", which is the correct term.

h/cli/commands/publisher.py
```diff
@@ -0,0 +1,73 @@
...
+@click.group()
+def publisher():
+    """Manage publisher accounts."""
+
+
+@publisher.command()
+@click.option('--name', prompt=True, help="The name of the publisher")
+@click.option('--authority', prompt=True, help="The authority (domain name) for the publisher")
+@click.pass_context
+def add(ctx, name, authority):
```
chdorner: It might be helpful to display a warning message at the beginning stating that this CLI command is only intended for development environments. At least until we can create groups without creators, because that admin user is just a really nasty hack to make it work when I sent you the code snippet.

h/cli/commands/publisher.py
```diff
@@ -0,0 +1,73 @@
...
+    """
+    Create a new publisher account.
+
+    This consists of an authority, an admin user and the main group for the
+    publisher's annotations.
+    """
+    request = ctx.obj['bootstrap']()
+
+    # Create a new authority.
+    authclient = models.AuthClient(name=name, authority=authority)
```
chdorner: I'm not sure if these `name` and `authority` variables contain unicode strings or `str`. We should use `h._compat.text_type` in case they're not `unicode` yet (also when they're used later on).

h/cli/commands/publisher.py
```diff
@@ -0,0 +1,73 @@
...
+    group_svc = request.find_service(name='group')
+    group_svc.create(name=name, authority=authority, userid=creator.userid,
+                     type_='publisher')
+
+    id = authclient.id
+    secret = authclient.secret
+
+    request.tm.commit()
+
+    click.echo('Publisher account for {authority} created'.format(authority=authority), err=True)
```
chdorner: Why is this `err=True`, I'm guessing this will send it to `stderr`, do we want that over `stdout`? (same with the call below this one)

h/cli/commands/publisher.py
```diff
@@ -0,0 +1,73 @@
...
+def secret(ctx, authority):
+    """
+    Display client ID and secret.
+
+    Display the client ID and secret for making API requests to manage the
+    users associated with a publisher.
+    """
+    request = ctx.obj['bootstrap']()
+
+    # Create a new authority
```
chdorner: This comment should be deleted because it's not true 😄

h/cli/commands/publisher.py
```diff
@@ -0,0 +1,73 @@
...
+    authclient = request.db.query(models.AuthClient).filter(
+                     models.AuthClient.authority == authority).first()
```
chdorner: Similar to above, if `authority` isn't a `unicode` yet, then we should convert it with `h._compat.text_type`.

h/cli/commands/publisher.py
robertknight: According to https://hyp.is/NnyczN2qEeayc1f9U4OdYw/click.pocoo.org/latest/arguments/ , string arguments are always unicode.

h/cli/commands/publisher.py
robertknight: This was just blindly adapted from the `h/cli/commands/user.py`. I suppose this is because the output was just informational in that context. I'll switch to stdout here.

h/cli/commands/publisher.py
robertknight: 🤦‍♂️

h/cli/commands/publisher.py
robertknight: As noted above, string arguments are unicode in click.

h/cli/commands/publisher.py
robertknight: Done

nickstenning:
AFAIK we haven't referred to these things as "publisher accounts" anywhere else in the code. They're called "auth clients" in the database, and we should be consistent about what we're calling them within the one codebase.

h/cli/commands/authclient.py
```diff
@@ -0,0 +1,61 @@
...
+
+
+@authclient.command()
+@click.argument('authority')
+@click.pass_context
+def secret(ctx, authority):
+    """
+    Display client ID and secret.
+
+    Display the ID and secret for an OAuth client associated with a given
```
robertknight: Strictly speaking the model permits multiple OAuth clients for a given authority. For practical purposes however, we're only going to have one per authority as far as I understand.

tests/h/cli/commands/authclient_test.py
```diff
@@ -0,0 +1,59 @@
...
+
+from h import models
+from h.cli.commands import authclient as authclient_cli
+
+
+class TestAddCommand(object):
+
+    def test_it_creates_an_authclient(self, cli, cliconfig, db_session):
+        authclient = self._add_authclient(cli, cliconfig, db_session)
+        assert authclient
```
chdorner: It might be better to do: `assert authclient is not None` because that's what SQLAlchemy's `first()` returns if it can't find anything, otherwise any truthy value would pass this test.

tests/h/cli/commands/authclient_test.py
```diff
@@ -0,0 +1,59 @@
...
+
+    def test_it_prints_the_client_id_and_secret(self, cli, cliconfig, db_session, echo):
```
chdorner: Not sure how important it is to test this, if you really want to then we should do some regex matching, because changing something that isn't important shouldn't fail this test. Also, I think the `click` context is storing stdout, so you can just read it from there instead of asserting that the method has been called.

tests/h/cli/commands/authclient_test.py
```diff
@@ -0,0 +1,59 @@
+# -*- coding: utf-8 -*-
+
+import mock
+import pytest
+
+from h import models
+from h.cli.commands import authclient as authclient_cli
+
+
+class TestAddCommand(object):
```
chdorner: It might be good to add another test to make sure that the name is set correctly.

chdorner: It might be good to add another test to make sure that the name is set correctly.

tests/h/cli/commands/authclient_test.py
```diff
@@ -0,0 +1,59 @@
...
+        return authclient
+
+
+class TestSecretCommand(object):
+    def test_it_prints_the_client_id_and_secret(self, authclient, cli, cliconfig, echo):
+        result = cli.invoke(authclient_cli.secret, [u'partner.org'], obj=cliconfig)
+
+        assert result.exit_code == 0
+
+        echo.assert_called_with('ID: {}\nSecret: {}'.format(authclient.id, authclient.secret))
```
chdorner: Similar to above, it might be better to test this with regex matching.

h/cli/commands/groups.py
```diff
@@ -0,0 +1,31 @@
...
+
+import click
+
+
+@click.group()
+def groups():
+    """Manage groups."""
+
+
+@groups.command()
```
chdorner: If you change this to `@groups.command('add-publisher-group')` then we don't have to deal with ugly underscores when using the CLI :)

chdorner: If you change this to `@groups.command('add-publisher-group')` then we don't have to deal with ugly underscores when using the CLI :)

h/cli/commands/user.py
```diff
@@ -14,18 +14,24 @@ def user():
...
     """Create a new user."""
     request = ctx.obj['bootstrap']()
 
     signup_service = request.find_service(name='user_signup')
 
-    user = signup_service.signup(username=username,
-                                 email=email,
-                                 password=password)
-    user.activate()
+    signup_args = {}
```
chdorner: It took me a while to figure out why you did it like this (because you don't want to set the authority when it's `None`). Maybe it's easier to understand if you change this to:

```
signup_args = {
    'username': username,
    'email': email,
    'password': password,
    'require_activation': False
}
if authority:
    signup_args['authority'] = authority

signup_service.signup(**signup_args)
```

tests/h/cli/commands/authclient_test.py
robertknight: Err... any truthy value of `authclient` would also pass `assert authclient is not None`.

tests/h/cli/commands/authclient_test.py
robertknight: I added this test after the initial implementation incorrectly printed `None` as the client ID and secret because it called `authclient.{id, secret}` before the object had actually been saved to the DB. I'll check to see if the output can be read / match from `result` instead of having to patch echo.

tests/h/cli/commands/authclient_test.py
chdorner: Right, so my reasoning was a bit vague.
The documentation clearly [states that `first()`](http://docs.sqlalchemy.org/en/rel_1_1/orm/query.html?highlight=first#sqlalchemy.orm.query.Query.first) that it returns the first result or `None`. So we should check that it's not `None` instead of a truthy value.

tests/h/cli/commands/authclient_test.py
chdorner: Yep, you can access the output string here: http://click.pocoo.org/6/api/#click.testing.Result.output

tests/h/cli/commands/authclient_test.py
robertknight: Done. I've opted to make it another assertion in the same test as the simplest way to check that the result is "an object with properties X and Y with values foo and bar"

robertknight:
As suggested, I have removed the `authclient secret` command because it did not have any real value over just performing a `SELECT * FROM authclient WHERE authority = 'publisher.org'` query using psql.

h/cli/__init__.py
```diff
@@ -3,7 +3,6 @@
 
 import functools
 import logging
-import os
```
robertknight: This import was unused.

PR #4305, opened 2017-01-18
Adding support for client_ga_tracking_id for client property analytics

sean-roberts:
Part of the work for: https://github.com/hypothesis/product-backlog/issues/117

Adds support for telling the client about client_ga_tracking_id as the "googleAnalytics" field in the hypothesis settings.

This PR's job is just to pipe the property down to the client

### DIFF
--- a/h/templates/app.html.jinja2
+++ b/h/templates/app.html.jinja2
@@ -34,20 +34,6 @@
     {% for url in app_js_urls %}
     <script src="{{ url }}"></script>
     {% endfor %}
-
-    <!-- Analytics !-->
-    {% if ga_tracking_id %}
-      <!-- Google Analytics -->
-      <script async src='//www.google-analytics.com/analytics.js'></script>
-      <script>
-       window.ga=window.ga||function(){(ga.q=ga.q||[]).push(arguments)};ga.l=+new Date;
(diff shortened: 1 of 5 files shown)


sean-roberts:
@robertknight can you check this one out and add client_ga_tracking_id for prod with a value of: "UA-26026798-5"

robertknight:
https://github.com/hypothesis/playbook/pull/88 adds `GOOGLE_ANALYTICS_CLIENT_TRACKING_ID=UA-26026798-5` to the configuration for the production environment.

The stage environment currently does not have a GA tracking ID set at all. We probably want to create a separate property for the client on staging / the stage extension in order to test analytics changes before they are deployed.

sean-roberts:
updated

h/templates/app.html.jinja2
```diff
@@ -34,20 +34,6 @@
     {% for url in app_js_urls %}
     <script src="{{ url }}"></script>
     {% endfor %}
-
-    <!-- Analytics !-->
-    {% if ga_tracking_id %}
```
robertknight: If we are removing this then we can also remove the `ga_tracking_id` argument to `render_app_html`.

PR #4306, opened 2017-01-19
Remove issue-based branch naming convention

fatbusinessman:
From a look through the extant branches on [`client`](https://github.com/hypothesis/client) and [`h`](https://github.com/hypothesis/h), this convention doesn’t appear to be in wide use in either repository (I did see [one branch in the client](https://github.com/hypothesis/client/compare/54-fix-crash-when-annotating-first-word-of-line-in-pdf) from six months ago).

### DIFF
--- a/docs/developing/submitting-a-pr.rst
+++ b/docs/developing/submitting-a-pr.rst
@@ -30,12 +30,6 @@ Some things to remember when submitting or reviewing a pull request:
   Please minimize issue gardening by using the GitHub syntax for closing
   issues with commit messages.
 
-- Name your branch in a way that makes it easier to follow the history back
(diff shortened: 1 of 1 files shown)


robertknight:
The most recent guidance on branch names comes from Nick in https://groups.google.com/a/list.hypothes.is/forum/#!searchin/eng/git$20branch$20names%7Csort:relevance/eng/Rme-s1WP9lY/5I7yT2cHCAAJ , specifically:

> Can I suggest we stick to relatively short, descriptive, dash-delimited names for branches? There's no need with a team our size to prefix every branch name with your name, and I'm not a fan of putting Trello card IDs or GitHub issues numbers in branch names1.

robertknight:
LGTM

PR #4308, opened 2017-01-20
Consider group join permission when rendering group views

chdorner:
These changes will make sure that we always show the correct search/join/404 page depending on the logged-in user (or not) and the group's `joinable_by` value.

It wasn't possible to do this change and still have the group search (activity) page and join controller separated as we had before, because we would have to check for `not read permission and join permission`, which doesn't work with our `has_permission` predicate, or any other method, except for creating a new custom view predicate just for this use case which doesn't sound like a good solution to me.

This also means that the group joining and leaving is now in the same place.

### DIFF
--- a/h/views/activity.py
+++ b/h/views/activity.py
@@ -88,7 +88,6 @@ class SearchController(object):
 @view_defaults(route_name='group_read',
                renderer='h:templates/activity/search.html.jinja2',
                effective_principals=security.Authenticated,
-               has_permission='read',
                request_method='GET')
 class GroupSearchController(SearchController):
     """View callables unique to the "group_read" route."""
@@ -99,6 +98,10 @@ class GroupSearchController(SearchController):
 
     @view_config(request_method='GET')
     def search(self):
+        result = self._check_access_permissions()
+        if result is not None:
+            return result
+
         check_slug(self.group, self.request)
 
         result = super(GroupSearchController, self).search()
@@ -232,6 +235,13 @@ class GroupSearchController(SearchController):
     @view_config(request_param='toggle_tag_facet')
     def toggle_tag_facet(self):
         return _toggle_tag_facet(self.request)
+
+    def _check_access_permissions(self):
+        if not self.request.has_permission('read', self.group):
+            if self.request.has_permission('join', self.group):
+                self.request.override_renderer = 'h:templates/groups/join.html.jinja2'
+                return {'group': self.group}
+            raise httpexceptions.HTTPNotFound()
 
 
 @view_defaults(route_name='activity.user_search',
--- a/h/views/activity.py
+++ b/h/views/activity.py
@@ -162,6 +162,27 @@ class GroupSearchController(SearchController):
         return result
 
     @view_config(request_method='POST',
+                 request_param='group_join')
+    def join(self):
+        """
+        Join the given group.
+
+        This adds the authenticated user to the given group and redirect the
+        browser to the search page.
+        """
+        if not self.request.has_permission('join', self.group):
+            raise httpexceptions.HTTPNotFound()
+
+        groups_service = self.request.find_service(name='group')
+        groups_service.member_join(self.group,
+                                   self.request.authenticated_userid)
+
+        url = self.request.route_url('group_read',
+                                     pubid=self.group.pubid,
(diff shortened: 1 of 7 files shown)


h/views/activity.py
```diff
@@ -99,6 +98,10 @@ def __init__(self, group, request):
 
     @view_config(request_method='GET')
     def search(self):
+        result = self._check_access_permissions()
+        if result is not None:
+            return result
```
nickstenning: I reckon we probably do want to test this branch?

h/views/activity.py
```diff
@@ -233,6 +236,13 @@ def delete_lozenge(self):
     def toggle_tag_facet(self):
         return _toggle_tag_facet(self.request)
 
+    def _check_access_permissions(self):
+        if not self.request.has_permission('read', self.group):
```
nickstenning: This is just fine, but it's worth noting that strictly I think the `self.group` argument here isn't needed, as the default context is the request context, which here is the group already, right?

tests/h/views/activity_test.py
```diff
@@ -181,19 +181,13 @@ def test_search_returns_group_info_if_user_a_member_of_group(self,
...
-        pyramid_request.has_permission.assert_called_once_with('admin', group)
-
     def test_search_does_not_show_the_edit_link_to_group_members(self,
                                                                  controller,
                                                                  group,
                                                                  pyramid_request):
-        pyramid_request.has_permission = mock.Mock(return_value=False)
+        def fake_has_permission(permission, context=None):
+            return permission != 'admin'
+        pyramid_request.has_permission = mock.Mock(side_effect=fake_has_permission)
```
nickstenning: In an ideal world, we wouldn't mock out has_permission here, we'd provide a stub context object with an appropriate ACL and let the view function worry about how to test that.

Just a comment -- no need to change anything.

h/views/activity.py
```diff
@@ -162,6 +162,27 @@ def user_annotation_count(aggregation, userid):
         return result
 
     @view_config(request_method='POST',
+                 request_param='group_join')
```
nickstenning: As discussed briefly on Slack, I think that all the alternate views in here with `underscored_params` are more than a little ugly.

In future I think we probably want to look into replacing those that represent actions (such as join, leave) with their own URLs (perhaps `/groups/:pubid/:slug/join` and `/groups/:pubid/:slug/leave`). Most of the others should, I think, be reimplemented on the client-side and don't need server-side support (remove a facet, toggle "More info", etc.)

h/views/activity.py
```diff
@@ -162,6 +162,27 @@ def user_annotation_count(aggregation, userid):
...
+    def join(self):
+        """
+        Join the given group.
+
+        This adds the authenticated user to the given group and redirect the
+        browser to the search page.
+        """
+        if not self.request.has_permission('join', self.group):
```
nickstenning: Same comment here about the fact that the `self.group` param is probably not needed as `self.group` is already the request context.

h/views/groups.py
```diff
@@ -112,6 +112,10 @@ def _update_group(self, appstruct):
 def read_unauthenticated(group, request):
     """Group view for logged-out users, allowing them to join the group."""
     check_slug(group, request)
+
+    if group.joinable_by is None:
```
nickstenning: I'm not sure I understand the commit message here. Even for logged out users there is a list of effective principals: `['system.Everyone']`. You should be able to use `request.has_permission('join')` to much the same effect?

h/views/activity.py
chdorner: Totally agree, I think it'll also make it much  more obvious what's happening by having two specific routes for joining and leaving.

h/views/activity.py
chdorner: Yep, I added it explicitly to make sure it's harder to introduce a bug where a user could join a non-joinable group.

h/views/activity.py
chdorner: That would not be the worst idea ;) Done!

h/views/activity.py
chdorner: See comment above

chdorner:
I think I fixed all the ones that need fixing now, otherwise please shout.

PR #4310, opened 2017-01-20
Remove old features in a Celery task (not at boot)

nickstenning:
We remove old feature flags from the database so that if we later accidentally reuse a feature name it doesn't unexpectedly come up switched on.

At the moment this is handled at application boot time, which means that the application fails to boot if the database isn't available. Instead, this commit adds a celery task to perform the removal, and an entry in the Celery Beat schedule to run it every six hours.

(This is one step towards allowing the application to boot without a database available.)


robertknight:
LGTM

PR #4311, opened 2017-01-20
Make memex.search._get_client public

nickstenning:
This is used by tests and will shortly be used by a CLI tool, so it probably shouldn't remain a module private function.

### DIFF
--- a/tests/functional/conftest.py
+++ b/tests/functional/conftest.py
@@ -62,8 +62,8 @@ def init_db(db_engine):
 
 @pytest.fixture(scope='session', autouse=True)
 def init_elasticsearch():
-    from memex.search import init, _get_client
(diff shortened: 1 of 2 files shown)


tests/functional/conftest.py
```diff
@@ -12,9 +12,6 @@
     'es.host': os.environ.get('ELASTICSEARCH_HOST', 'http://localhost:9200'),
     'es.index': 'hypothesis-test',
     'h.app_url': 'http://localhost',
-    'h.db.should_create_all': False,
-    'h.db.should_drop_all': False,
-    'h.search.autoconfig': False,
```
robertknight: What are these changes for?

nickstenning: Gah! They should be in #4312, sorry.

PR #4312, opened 2017-01-20
Simplify database and search index initialization

### LINKED ISSUE #4290: Race condition when auto-configuring search index
I’ve happened across a bug that can manifest when setting up a development server for the first time, whereby multiple search indices get created and assigned the same alias. This effectively stops any annotations being indexed, as Elasticsearch doesn’t know which index to use.

### Steps to reproduce
1. Clear out all search indices (for instance, with `curl -X DELETE 'http://localhost:9200/*'`)
2. If you want to make the race condition reliably reproducible, apply this patch:
    ```diff
    diff --git a/src/memex/search/config.py b/src/memex/search/config.py
    index a4fad51..99610bd 100644
    --- a/src/memex/search/config.py
    +++ b/src/memex/search/config.py
    @@ -181,6 +181,9 @@ def init(client):
         if client.conn.indices.exists(index=client.index):
             return
     
    +    from time import sleep
    +    sleep(5)
    +
         # Otherwise we create a new randomly-named index and alias it. This allows
         # us to straightforwardly reindex when we need to.
         concrete_index = configure_index(client)

    ```

3. Run `make devserver`
4. Go to `http://localhost:5000/docs/help` and annotate something.
5. Refresh the page

### Expected behaviour

The annotation should still be visible on the page.

### Actual behaviour

The annotation doesn’t display after a page refresh.

### Additional details

If the index aliases are in a healthy state, a request to `http://localhost:9200/_cat/aliases` should result in a single line, something like:

    hypothesis hypothesis-c8af8a7c - - - 

If the condition has been triggered, the same request will return multiple lines:

    hypothesis hypothesis-c8af8a7c - - - 
    hypothesis hypothesis-e82e3a39 - - - 
    hypothesis hypothesis-9e2abe0e - - -

nickstenning:
This commit removes the automatic (and hence racy) initialisation of database tables and search index, in favour of a `hypothesis init` command which must be called explicitly.

This simplifies the integration of memex into h (no more overriding of memex's automatic search index initialization), and various other parts of the application boot code, at the cost of requiring an explicit initialisation of the database before first use.

In addition, this commit adds support for detecting if a database is managed by Alembic, and skips automatic database initialisation in that case. This should prevent a number of confusing behaviours which crop up for developers when new tables are added to the database.

For development environments, this should remain largely transparent to the user, thanks to an additional call made by the Makefile's `dev` target.

The following settings are removed with this commit:

- MODEL_CREATE_ALL (h.db.should_create_all)
- MODEL_DROP_ALL (h.db.should_drop_all)
- SEARCH_AUTOCONFIG (h.search.autoconfig)

Fixes #4290.

~~_**N.B.** This depends on #4311, and will need to be rebased after that has merged._~~

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -26,6 +26,7 @@ clean:
 ## Run the development H server locally
 .PHONY: dev
 dev: build/manifest.json .pydeps
+	@bin/hypothesis --dev init
 	@bin/hypothesis devserver
 
 ## Build hypothesis/hypothesis docker image
--- a/h/app.py
+++ b/h/app.py
@@ -112,11 +112,6 @@ def includeme(config):
     # - we can override behaviour from `memex` if necessary.
     config.include('memex', route_prefix='/api')
 
-    # If search autoconfig is false, then override memex.search.init with a
-    # no-op.
-    if not asbool(config.registry.settings.get('h.search.autoconfig')):
-        config.action('memex.search.init', lambda: None)
-
     # Override memex group service
     config.register_service_factory('h.services.groupfinder.groupfinder_service_factory',
                                     iface='memex.interfaces.IGroupService')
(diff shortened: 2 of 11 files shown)


robertknight:
LGTM. I've tested this with and without a pre-existing database and search index.

PR #4313, opened 2017-01-20
Remove the 'UA-' prefix from the Google Analytics tracking ID

robertknight:
The prefix is part of the ID, so it has been moved to the configuration
for the service in the playbook.

See https://github.com/hypothesis/playbook/pull/88#issuecomment-274102054

This needs to be deployed together with https://github.com/hypothesis/playbook/pull/89


nickstenning:
LGTM.

PR #4314, opened 2017-01-20
Add db migration to update all existing web_uris

seanh:
b240b8f303b67ba42ae78fbc067babbf6bf79b23 changed the logic that is used
to set Document.web_uri based on the Document.document_uris. This commit
adds a db migration that updates all existing Document.web_uris using
the new logic.

### DIFF
--- a/h/migrations/versions/9f5e274b202c_update_all_document_web_uris.py
+++ b/h/migrations/versions/9f5e274b202c_update_all_document_web_uris.py
@@ -1,0 +1,116 @@
+"""
+Update all Document.web_uris.
+
+Revision ID: 9f5e274b202c
+Revises: e10ce4472966
+Create Date: 2017-01-20 16:07:03.442975
+"""
+
+from __future__ import unicode_literals
+
+from collections import namedtuple
+import logging
+
+from alembic import op
+import sqlalchemy as sa
+from sqlalchemy.ext.declarative import declarative_base
+from sqlalchemy.orm import sessionmaker
+from sqlalchemy.orm import subqueryload
+
+from memex._compat import urlparse
+
+
+revision = '9f5e274b202c'
+down_revision = 'e10ce4472966'
+
+
+Base = declarative_base()
+Session = sessionmaker()
+log = logging.getLogger(__name__)
+
+
+class Window(namedtuple('Window', ['start', 'end'])):
+    pass
+
+
+class Document(Base):
+    __tablename__ = 'document'
+    id = sa.Column(sa.Integer, primary_key=True)
+    updated = sa.Column(sa.DateTime)
+    web_uri = sa.Column(sa.UnicodeText())
+    document_uris = sa.orm.relationship('DocumentURI',
+                                        backref='document',
+                                        order_by='DocumentURI.updated.desc()')
+
+    def updated_web_uri(self):
+        def first_http_url(type_=None):
+            for document_uri in self.document_uris:
+                uri = document_uri.uri
+                if type_ is not None and document_uri.type != type_:
+                    continue
+                if urlparse.urlparse(uri).scheme not in ['http', 'https']:
+                    continue
+                return document_uri.uri
+
+        return (first_http_url(type_='self-claim') or
+                first_http_url(type_='rel-canonical') or
+                first_http_url())
+
+
+class DocumentURI(Base):
+    __tablename__ = 'document_uri'
+    id = sa.Column(sa.Integer, primary_key=True)
+    updated = sa.Column(sa.DateTime)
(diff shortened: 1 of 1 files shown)


seanh:
I'm gonna do a quick analysis of all the web_uris that this changes...

seanh:
This db migration changes the web_uri's of 59,542 documents in production. Of course with that many it's impossible to look at the, all, or even look at all the domains of the changed documents, and know that all of the changes are improvements and that no links will be broken. But I had a quick look at the domains that will have the most changed web_uris from this migration:

34,155 of the 59,542 changed documents that previously had no web_uri, but now do. I didn't look at these but presumably that's an improvement.

Of the remaining 25,387 documents where a non-empty web_uri is changed to a different non-empty web_uri:

* 5318 were `via.hypothes.is` URLs. For example <https://via.hypothes.is/http://www.josiefraser.com/2016/09/alt-c-2016-keynote-in-the-valley-of-the-trolls/> changes to <http://www.josiefraser.com/2016/09/alt-c-2016-keynote-in-the-valley-of-the-trolls/>`. This would cause it to be displayed on activity pages as `josiefraser.com` instead of as `via.hypothes.is`, but I think we already have code elsewhere to strip the `via.hypothes.is` anyway, that code is probably unnecessary now.

* 1016 are `wp.me` URLs, for example <http://wp.me/p13vSu-MK> changes to <http://allthingsanalytics.com/2015/06/26/big-data-and-the-future-of-work/>. So it'll now show as `allthingsanalytics.com` rather than `wp.me` (this affects many different domains)

* 898 are `en.wikipedia.org` URLs. There are lots of variations on this but it all looks like different WordPress URLs reducing to more canonical ones.

  A lot of these are changing a wikipedia edit URL, e.g. <https://en.wikipedia.org/w/index.php?title=Weingarten_Rights&action=edit>, to the normal page URL <https://en.wikipedia.org/wiki/Weingarten_Rights>.

  There are some changes that strip `#` fragments, and maybe some other stuff in here.

* 709 are `public-api.wordpress.com`, example <https://public-api.wordpress.com/oembed/1.0/?format=json&url=http%3A%2F%2Fchristianlauersen.net%2F2015%2F08%2F11%2Fthe-fall-of-the-library-fortress%2F&for=wpcom-auto-discovery> changes to <http://christianlauersen.net/2015/08/11/the-fall-of-the-library-fortress/>. So it will now display as `christianlauersen.net` instead of as `public-api.wordpress.com`. Affects many domains.

* 554 are `nytimes.com`, these all seem to be removing unnecessary queries, example <http://www.nytimes.com/2015/07/31/opinion/david-brooks-two-cheers-for-capitalism.html?rref=collection%2Fcolumn%2Fdavid-brooks&action=click&contentCollection=opinion&region=stream&module=stream_unit&contentPlacement=3&pgtype=collection> changes to <http://www.nytimes.com/2015/07/31/opinion/david-brooks-two-cheers-for-capitalism.html>

* 300 are `medium.com`, and it looks like it just strips unnecessary fragments and queries.

* `mobile.nytimes.com` URLs change to `nytimes.com`

* `scalar.usc.edu` URLs get queries removed from them

* `courses.canelalearning.com` URLs ending with `?p=<num>` get changed to what looks like more canonical URLs

I looked at several more domains that had 100+ web_uris changed, they're all the same sort of thing, all look like improvements to me. Of course there's a large number of domains that have < 100 annotated documents whose web_uris get changed, too many to look through.

I would bet that this migration _does_ introduce some problems with some domains, but I'd also bet that it fixes a lot more problems.

h/migrations/versions/9f5e274b202c_update_all_document_web_uris.py
```diff
@@ -0,0 +1,116 @@
...
+                     server_default='')
+    document_id = sa.Column(sa.Integer,
+                            sa.ForeignKey('document.id'),
+                            nullable=False)
+
+
+def upgrade():
+    session = Session(bind=op.get_bind())
+
+    windows = _fetch_windows(session)
```
chdorner: This is totally fine for this migration, but in case you haven't seen it, we've added a more efficient way of doing windowed queries with the new [`memex.util.query.column_windows`](https://github.com/hypothesis/h/blob/master/src/memex/util/query.py#L8), and an example of how it is used is in the [`memex.search.index.BatchIndexer`](https://github.com/hypothesis/h/blob/master/src/memex/search/index.py#L158-L171).

seanh: Ooh, nice, I just copied this one from one of the existing migrations, but I'll use the new way next time

PR #4315, opened 2017-01-20
Refactor annotation presenters to get links from resource

chdorner:
This will allow for a better API when creating a presenter, as the
resource already has the group service, so it's better to also pass the
links service to the resource instead of directly to the presenter.

This way the presenter only needs to know about the resource and nothing
else.

This refactoring came out of discussions around #4299.


nickstenning:
Looks great!

PR #4317, opened 2017-01-23
API profile endpoint (for authenticated H users)

fatbusinessman:
This is a first step towards hypothesis/product-backlog#119 – so far it only replicates the existing behaviour of the `/app` endpoint for logged-in users (minus the CSRF token), but the next steps are to:

* Get this working for third-party users, once I’ve got a better understanding of how third-party groups work (including whether there’s an equivalent to the `__world__` group).
* Add an `authority` parameter so we can show first– and third-party groups
* Get it working for anonymous users and remove the auth requirement from the view decorator

At this stage, I’m mostly looking for guidance on whether I’m following our team conventions, whether I seem to be heading in a sensible direction, whether I’m likely to head down any nasty rabbit holes, or whether there are any handy tricks I seem to have missed. I’m intending to smush a load of the commits around before merging, to get rid of some of my stream-of-consciousness commit messages while leaving enough context for future developers (who may or may not be me).

### DIFF
--- a/h/routes.py
+++ b/h/routes.py
@@ -62,6 +62,7 @@ def includeme(config):
 
     # API (other than those provided by memex)
     config.add_route('badge', '/api/badge')
+    config.add_route('api.profile', '/api/profile')
     config.add_route('token', '/api/token')
     config.add_route('api.users', '/api/users')
 
--- a/h/session.py
+++ b/h/session.py
@@ -11,11 +11,20 @@ def model(request):
     session['userid'] = request.authenticated_userid
     session['groups'] = _current_groups(request)
     session['features'] = request.feature.all()
-    session['preferences'] = {}
-    user = request.authenticated_user
-    if user and not user.sidebar_tutorial_dismissed:
-        session['preferences']['show_sidebar_tutorial'] = True
+    session['preferences'] = _user_preferences(request.authenticated_user)
     return session
+
+
+def profile(request):
+    """
+    Return a representation of the current user's information and settings.
+    """
+    profile = {}
+    profile['userid'] = request.authenticated_userid
+    profile['groups'] = _current_groups(request)
+    profile['features'] = request.feature.all()
+    profile['preferences'] = _user_preferences(request.authenticated_user)
+    return profile
 
 
 def pop_flash(request):
@@ -56,6 +65,13 @@ def _current_groups(request):
                                      slug=group.slug),
         })
     return groups
+
+
+def _user_preferences(user):
+    preferences = {}
+    if user and not user.sidebar_tutorial_dismissed:
+        preferences['show_sidebar_tutorial'] = True
+    return preferences
 
 
 def includeme(config):
--- a/h/views/api_profile.py
+++ b/h/views/api_profile.py
@@ -1,0 +1,15 @@
+# -*- coding: utf-8 -*-
+
+from __future__ import unicode_literals
+
+from pyramid import security
+
+from h import session as h_session
+from h.util.view import json_view
+
+
+@json_view(route_name='api.profile',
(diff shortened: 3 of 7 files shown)


h/routes.py
```diff
@@ -64,6 +64,7 @@ def includeme(config):
     config.add_route('badge', '/api/badge')
     config.add_route('token', '/api/token')
     config.add_route('api.users', '/api/users')
+    config.add_route('api.profile', '/api/profile')
```
chdorner: We usually keep things ordered alphabetically (@nickstenning will always leave a comment when he sees something that isn't correctly ordered 😉). Although this block of code is already in a bad order, so not really your fault at all 🙂

h/views/api_profile.py
```diff
@@ -0,0 +1,11 @@
+from pyramid import security
```
chdorner: New Python files should always start with this magic comment: `# -*- coding: utf-8 -*-`

And to aid the efforts of moving our code base to Python 3, I usually also add: `from __future__ import unicode_literals`.

tests/h/views/api_profile_test.py
```diff
@@ -0,0 +1,10 @@
+import mock
+
+from h.views import api_profile
+
+
+@mock.patch('h.session.profile')
```
chdorner: This is fine, just in case you didn't come across it yet, in recent months we opted for explicitely creating fixtures, with one of the advantages being you can give it a name yourself:

```python
@pytest.fixture
def profile(patch):
    return patch('h.session.profile')
```

Note: This also uses the `patch` fixture that is defined in `tests/h/conftest.py`.

h/session.py
```diff
@@ -11,13 +11,27 @@ def model(request):
...
-        session['preferences']['show_sidebar_tutorial'] = True
+    session['preferences'] = _user_preferences(request.authenticated_user)
     return session
 
 
+def profile(request):
+    """
+    Return a representation of the current user's information and settings.
+
+    This is intended for use in a third-party setting, where the client will
```
robertknight: "for use in a third-party setting" is a little ambiguous. Initially we will be using it in the client when using third-party _accounts_, but I expect in the not-too-distant future we will use OAuth for first-party accounts as well. This API is also likely to be useful to third-party _clients_.

h/routes.py
fatbusinessman: Ah, ok: good to know. For the sake of clarity, when the route names and the paths have different sort orders, which one should take precedence? As far as I can tell, all the blocks that are already sorted have naming conventions that mean it’s not possible to tell.

nickstenning: I think in this case it doesn't really matter. Either makes sense, so choose whichever involves less moving around.

tests/h/session_test.py
```diff
@@ -69,8 +69,28 @@ def test_model_show_sidebar_tutorial(
         assert 'show_sidebar_tutorial' not in preferences
 
 
+def test_profile_userid_unauthenticated(unauthenticated_request):
```
chdorner: It'll be good to expand the unit tests a bit. There are tests for all keys in the model tests which we should probably duplicate. Because the `model` function will probably be deleted in the coming months, so just duplicating the tests would make the process of deleting that one easier.

fatbusinessman: Sounds reasonable. I had a quick go at parameterizing the tests based on their target, just to satisfy my own curiosity, and I agree that duplicating the tests feels cleaner.

chdorner: Yeah, usually parameterizing is the way to go, but since we know that the `model` tests will be removed over the long term I don't mind duplicating the tests. 👍

PR #4318, opened 2017-01-23
Run nginx and all (most) processes in container

nickstenning:
The easiest way for us to migrate to Skyliner involves us having a single deployable artifact built from this repository which spins up everything we need to run the application.

This commit does two important things:

1. Adds nginx (and an nginx configuration) to the Docker image, allowing us to run both web server and websocket server on a single port.

2. Installs [supervisord][1] to manage the various processes which must run concurrently and sets up supervisor to be the default command for the image.

There are a couple of additional conveniences added, namely:

- A helper script called `logprefix` which emulates the line-prefixing behaviour of honcho so that it's easy to tell which log line comes from which subprocess.

- A helper script called `start-nginx` which is responsible for replacing the fallback upstream if the `FALLBACK_UPSTREAM` environment variable is set. This will be used to forward appropriate requests to WordPress for the time being, and we hope will be removable in due course.

Observations and caveats:

- This commit shouldn't prevent us from continuing to deploy as we currently do, with the one caveat that we'll need to explicitly set the command for the `web` container.

- Setting the WEB_CONCURRENCY environment variable for the container will currently adjust the number of workers started for *both* the web application and the websocket server. This may result in unintended behaviour.

- We don't currently run a Celery beat process in the container. This will need to be managed separately for the time being.

- These changes add only about ~5MB to the image size.

[1]: http://supervisord.org/

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,6 +6,7 @@ RUN apk-install \
     ca-certificates \
     libffi \
     libpq \
+    nginx \
     python \
     py-pip \
     nodejs \
@@ -14,6 +15,9 @@ RUN apk-install \
 # Create the hypothesis user, group, home directory and package directory.
 RUN addgroup -S hypothesis && adduser -S -G hypothesis -h /var/lib/hypothesis hypothesis
 WORKDIR /var/lib/hypothesis
+
+# Ensure nginx state and log directories writeable by unprivileged user.
+RUN chown -R hypothesis:hypothesis /var/log/nginx /var/lib/nginx
 
 # Copy minimal data to allow installation of dependencies.
 COPY src/memex/__init__.py ./src/memex/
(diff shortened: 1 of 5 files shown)


nickstenning:
Marked as WIP at least until https://github.com/hypothesis/playbook/pull/90 has merged.

chdorner:
I will test this out thoroughly either tonight or otherwise tomorrow

chdorner:
Oh, I just saw that it's still marked as WIP. I will not merge yet in this case.

nickstenning:
I think you've identified a real issue with the Celery workers, though -- I think they're going to be fighting over the schedule file. Will have a look into that.

nickstenning:
Okay, I've updated this and changed two things:

1. There's only one top-level Celery worker process, which will (as per the Celery defaults) listen on all configured queues. Celery will default to preforking N worker processes, where N is the number of CPUs on the system. This should provide reasonable default response times for handling of indexer tasks.

2. I've turned what was the `logprefix` helper into a small program called `logger` which implements the Supervisor event listener protocol. This allows us to straightforwardly distinguish process STDOUT/STDERR in the overall output, and also fixes an issue where process output was being buffered regardless of whether `PYTHONUNBUFFERED` was set in the container.

nickstenning:
Removed WIP as I think this is safe to merge now.

PR #4319, opened 2017-01-23
Identify appType of bookmarklet script

sean-roberts:


robertknight:
LGTM. I'm not entirely sure about the `appType` name as what we're talking about here really is the system or service used to deliver the application. I'll leave bikeshedding about that for another time though.

PR #4321, opened 2017-01-24
Add refresh tokens to api tokens

seanh:
Adds `refresh_token` to the `Token` model and ensures that *if* an `expires` time is given when creating a token then a `refresh_token` is also generated for it.

This is marked WIP because I need to write a db migration pr that will need to be merged before this one.

### DIFF
--- a/h/models/token.py
+++ b/h/models/token.py
@@ -9,6 +9,7 @@ import os
 import sqlalchemy
 from sqlalchemy.dialects import postgresql
 
+from h import security
 from h.auth.interfaces import IAuthenticationToken
 from h.db import Base
 from h.db import mixins
@@ -60,7 +61,7 @@ class Token(Base, mixins.Timestamps):
         self.regenerate()
 
         if expires:
-            self.refresh_token = _token()
+            self.refresh_token = security.token_urlsafe()
 
     @classmethod
     def get_dev_token_by_userid(cls, session, userid):
--- a/h/models/token.py
+++ b/h/models/token.py
@@ -24,9 +24,6 @@ class Token(Base, mixins.Timestamps):
     #: to, for example, one of the short-lived JWTs that the client uses).
     prefix = u'6879-'
 
-    #: A prefix that identifies a token as a refresh token.
-    refresh_token_prefix = u'7980-'
-
     id = sqlalchemy.Column(sqlalchemy.Integer,
                            autoincrement=True,
                            primary_key=True)
@@ -63,7 +60,7 @@ class Token(Base, mixins.Timestamps):
         self.regenerate()
 
         if expires:
-            self.refresh_token = self.refresh_token_prefix + _token()
+            self.refresh_token = _token()
 
     @classmethod
     def get_dev_token_by_userid(cls, session, userid):
--- a/h/models/token.py
+++ b/h/models/token.py
@@ -56,11 +56,11 @@ class Token(Base, mixins.Timestamps):
     #: A NULL value means it is a developer token.
     authclient = sqlalchemy.orm.relationship('AuthClient')
 
-    def __init__(self, expires=None, **kwargs):
-        super(Token, self).__init__(expires=expires, **kwargs)
+    def __init__(self, **kwargs):
+        super(Token, self).__init__(**kwargs)
         self.regenerate()
 
-        if expires:
+        if self.expires:
             self.refresh_token = security.token_urlsafe()
 
     @classmethod
--- a/h/models/token.py
+++ b/h/models/token.py
@@ -24,6 +24,9 @@ class Token(Base, mixins.Timestamps):
     #: to, for example, one of the short-lived JWTs that the client uses).
     prefix = u'6879-'
 
+    #: A prefix that identifies a token as a refresh token.
+    refresh_token_prefix = u'7980-'
+
     id = sqlalchemy.Column(sqlalchemy.Integer,
                            autoincrement=True,
                            primary_key=True)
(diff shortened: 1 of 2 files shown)


chdorner:
I've merged, deployed, and ran the database migration

seanh:
Thanks @chdorner , this is ready for review now then

h/models/token.py
```diff
@@ -24,6 +24,9 @@ class Token(Base, mixins.Timestamps):
     #: to, for example, one of the short-lived JWTs that the client uses).
     prefix = u'6879-'
 
+    #: A prefix that identifies a token as a refresh token.
+    refresh_token_prefix = u'7980-'
```
chdorner: I was just checking with @nickstenning about why we have these prefixes, and he said they're just for humans to be able to recognize what type of token it is. In this case I don't think we need a prefix for the refresh tokens, as there will never be used anywhere else other than when refreshing an expired access token.

tests/h/models/token_test.py
```diff
@@ -2,12 +2,40 @@
...
+        expires = one_hour_from_now()
+        authclient = 'example.com'
+
+        access_token = Token(userid=userid,
+                             expires=expires,
+                             authclient=authclient)
+
+        assert access_token.userid == userid
+        assert access_token.value
+        assert access_token.value.startswith("6879-")
```
chdorner: Didn't we want to change the prefix for normal access tokens? I seem to remember that there was an issue for it, but I can't find it right now.

h/models/token.py
```diff
@@ -48,10 +58,13 @@ class Token(Base, mixins.Timestamps):
...
     authclient = sqlalchemy.orm.relationship('AuthClient')
 
-    def __init__(self, **kwargs):
-        super(Token, self).__init__(**kwargs)
+    def __init__(self, expires=None, **kwargs):
+        super(Token, self).__init__(expires=expires, **kwargs)
         self.regenerate()
 
+        if expires:
+            self.refresh_token = self.refresh_token_prefix + _token()
```
nickstenning: I don't think the refresh token needs a prefix. The main purpose of the token prefix is for a human to be able to quickly and easily identify that the string they're looking at is an API token. The refresh token should never be used by a human -- it's a string that only machines have to worry about.

In addition, I'd suggest that for ease of identifying that this is doing the correct thing, we use `h.security.token_urlsafe` to generate the refresh tokens. (Perhaps we want to switch to using this for API tokens at some point in the future, too.)

tests/h/models/token_test.py
seanh: [I wanted to change it but we decided not to bother](https://hypothes-is.slack.com/archives/backend/p1484923659002492), I think that'd be a separate pr anyway though

h/models/token.py
seanh: I guess so. If refresh tokens have a prefix then we can easily identify mistakes such as using the refresh token in the place where the access token is supposed to be, that sort of thing. And there's no chance, however small, of a randomly generated refresh token happening to begin with one of the other token prefixes. Can remove it if you want though.

h/models/token.py
```diff
@@ -48,10 +58,13 @@ class Token(Base, mixins.Timestamps):
     #: A NULL value means it is a developer token.
     authclient = sqlalchemy.orm.relationship('AuthClient')
 
-    def __init__(self, **kwargs):
-        super(Token, self).__init__(**kwargs)
+    def __init__(self, expires=None, **kwargs):
```
chdorner: On second thought, I'm not a big fan of overriding SQLAlchemy model constructors too much, [we've done some recent work to remedy this in the `Group` model](https://github.com/hypothesis/h/commit/84ed007b2a151381299497789f015e087c3f31d9).
Can't you just change the `if` further down to: `if kwargs.get('expires')`?

In general I think we should try and leave the model constructors as much as possible, at least we should never mess with the signature. It's okay like in this case to do additional work (generating token, maybe generating a refresh token).

seanh: I can certainly pull `expires` out of `kwargs` instead as you suggest. Does this make any difference to the method signature though? (Both ways, `expires` is an optional arg.)

Also wondering what you'll think of https://github.com/hypothesis/h/pull/4323/commits/7a2d697ae769675c5aa335c25745e9f4a4e764a6 (which is in [another pr](https://github.com/hypothesis/h/pull/4323/commits))

tests/h/models/token_test.py
chdorner: Right okay, I'm fine not to bother, just couldn't remember what happened after that discussion.

h/models/token.py
chdorner: > If refresh tokens have a prefix then we can easily identify mistakes such as using the refresh token in the place where the access token is supposed to be, that sort of thing

We can easily do this by just the token having a prefix and not the refresh token. It would also eliminate complexity from the model.

seanh: If refresh tokens _don't_ have a prefix then it's possible (albeit unlikely) for a randomly generated refresh token to start with the developer or access token prefix, and so look to a human like a different type of token. Anyway, this is unlikely, and people really seem to dislike this refresh token prefix, so I'll remove it.

chdorner: Hmm, I don't think that `binascii.hexlify(os.urandom(16))` will ever return a string with a dash?

h/models/token.py
seanh: Both done (well not for the access tokens yet)

h/models/token.py
chdorner: I think that SQLAlchemy models have a specific constructor which we shouldn't change. Adding the optional `expires` is okay, but it doesn't add anything because you can just read it from `kwargs`. I wouldn't make the `userid` a required parameter though, because it adds complexity that needs to be kept in our heads of how the constructors look like for the different models. And since we have a database-level constraint that checks that it is not missing there isn't a data integrity issue. The error message is also pretty clear stating that the value of the required column is missing.

h/models/token.py
```diff
@@ -52,6 +60,9 @@ def __init__(self, **kwargs):
         super(Token, self).__init__(**kwargs)
         self.regenerate()
 
+        if self.expires:
```
chdorner: This is even nicer than `kwargs.get('expires')` 👍

PR #4322, opened 2017-01-24
Remove check for CSRF token in `GET /api/token` endpoint

robertknight:
Since this endpoint does not modify any data and does not support CORS, there is no need to check the CSRF token that I can see.

Removing the CSRF check eliminates the need for the client to fetch the session state (in a separate API call) before it fetches an API token during the initial load. Removing this dependency will make implementing 3rd party authentication in the client simpler, until we can remove the legacy cookie-based auth.

I presume the CSRF check is there because historically (before 5b46ddc9000fcf4038ab27b4cde1dba22037dfe9) this endpoint used to support POST as well, so could have modified user data in theory.

I have tested this with a client (embed and Chrome extension) modified to stop sending the CSRF token.

Before I made this change, I hadn't realized that the `GET /api/token` endpoint returns a valid token even for unauthenticated users and the client relies on that behavior.

### DIFF
--- a/h/views/client.py
+++ b/h/views/client.py
@@ -48,15 +48,8 @@ def annotator_token(request):
 
     The token can be used in the Authorization header in subsequent requests to
     the API to authenticate the user identified by the
-    request.authenticated_userid of the _current_ request.
+    request.authenticated_userid of the _current_ request, which may be None.
     """
-    try:
-        # The client must supply the CSRF token in the X-CSRF-Token request
-        # header.
-        session.check_csrf_token(request)
-    except exceptions.BadCSRFToken:
-        raise httpexceptions.HTTPUnauthorized()
-
     return generate_jwt(request, 3600)
 
 
--- a/h/views/client.py
+++ b/h/views/client.py
@@ -8,9 +8,6 @@ Views which exist either to serve or support the JavaScript annotation client.
 
 from __future__ import unicode_literals
 
-from pyramid import exceptions
-from pyramid import httpexceptions
-from pyramid import session
 from pyramid.view import view_config
 
 from h import client
--- a/tests/h/views/client_test.py
+++ b/tests/h/views/client_test.py
@@ -3,22 +3,16 @@
 from __future__ import unicode_literals
 
 import pytest
-from pyramid import exceptions
-from pyramid import httpexceptions
 
 from h.views import client
 
-annotator_token_fixtures = pytest.mark.usefixtures('generate_jwt', 'session')
 
-
-@annotator_token_fixtures
 def test_annotator_token_calls_generate_jwt(generate_jwt, pyramid_request):
     client.annotator_token(pyramid_request)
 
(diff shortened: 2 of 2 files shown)


robertknight:
@nickstenning asked me to look into the history to clarify why this endpoint used a CSRF token for authentication. As far as I can ascertain, that goes back to 72b527dfee4c2222b71af0816d4c84db62b31956 which was an attempt to abstract authentication from the main application. It introduced a generic `/api/token` endpoint which took an assertion and returned an access token. This endpoint needed an assertion specified in the query string and the CSRF was conveniently available.

Since then, the abstraction involved in authentication has been stripped away on the client and service and we have a better idea of what _the future_™ looks like (a more standard OAuth flow). In this legacy endpoint, it looks clear to me that the CSRF token check isn't needed.

In terms of the impact on the client, I realized that getting rid of this check isn't as essential as I thought yesterday, but it still allows a few dozen lines of code/tests to be removed.

nickstenning:
Ok, I think I'm convinced that this is safe. The endpoint is cookie-authenticated and not CORS-enabled or JSONP-enabled, so there should be no problem here.

h/views/client.py
```diff
@@ -48,15 +48,8 @@ def annotator_token(request):
...
-    request.authenticated_userid of the _current_ request.
+    request.authenticated_userid of the _current_ request, which may be None.
     """
-    try:
-        # The client must supply the CSRF token in the X-CSRF-Token request
-        # header.
-        session.check_csrf_token(request)
-    except exceptions.BadCSRFToken:
-        raise httpexceptions.HTTPUnauthorized()
-
```
nickstenning: There are three unused imports at the top of this file as a result of these changes.

tests/h/views/client_test.py
```diff
@@ -12,23 +12,6 @@
...
-    session.check_csrf_token.assert_called_once_with(pyramid_request)
-
-
-@annotator_token_fixtures
-def test_annotator_token_raises_Unauthorized_if_check_csrf_token_raises(
-        pyramid_request,
-        session):
-    session.check_csrf_token.side_effect = exceptions.BadCSRFToken
-
-    with pytest.raises(httpexceptions.HTTPUnauthorized):
```
nickstenning: Ditto with the unused imports. You also have an unused `session` fixture in this file now.

h/views/client.py
robertknight: Thanks for pointing these out. It would be helpful if Hound could flag these up. flake8 did locally once I enabled syntastic.

tests/h/views/client_test.py
robertknight: Fixed

h/views/client.py
nickstenning: It's not possible to make a line comment on a line that isn't in the diff.

nickstenning:
LGTM.

PR #4323, opened 2017-01-24
Make userid mandatory

seanh:
Depends on https://github.com/hypothesis/h/pull/4321

You can't create a Token without a userid (you can create the Python
object fine, but if you add it to a db session and then try to flush or
commit that session you will get an IntegrityError - userid is not
nullable in the database).

Make the userid arg a positional arg on the Token class so that its
mandatoriness is explicit and so that you can't create an invalid ORM
object.

### DIFF
--- a/h/models/token.py
+++ b/h/models/token.py
@@ -24,6 +24,9 @@ class Token(Base, mixins.Timestamps):
     #: to, for example, one of the short-lived JWTs that the client uses).
     prefix = u'6879-'
 
+    #: A prefix that identifies a token as a refresh token.
+    refresh_token_prefix = u'7980-'
+
     id = sqlalchemy.Column(sqlalchemy.Integer,
                            autoincrement=True,
                            primary_key=True)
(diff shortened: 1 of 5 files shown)


nickstenning:
Can you expand on the problem that this is trying to solve? A token without a userid isn't valid *for persistence*, sure, but aren't you going to discover that the first time you try and save one?

seanh:
On master, the `IntegrityError` from psycopg2 won't come until you flush or commit the db session, which could be far away from the bug (where a `Token` without a `userid` was created and added to the session). So the problem that I'm trying to fix here is the potential for difficult to find bugs to creep in.

On this branch you'll get a `TypeError` as soon as you try to init a `Token` without a `userid`, and the traceback will show you the line of code where the bug is.

nickstenning:
My concern is that in general overriding SQLAlchemy constructors causes more pain than it's worth. I'm not really convinced that whatever gains are to be had by making this mandatory at construction time are greater than the loss of consistency between sqla models.

seanh:
Do you mean consistency in that arguments usually have to be passed as named rather than positional args to sqla constructors? `Token(userid='foo')` rather than `Token('foo')`?

nickstenning:
Do we still want/need this change?

seanh:
I guess not

PR #4325, opened 2017-01-25
Use token_urlsafe() to generate API tokens

seanh:
Use h.security.token_urlsafe() to generate access and dev API tokens, as
we already do for refresh tokens.

### DIFF
--- a/h/models/token.py
+++ b/h/models/token.py
@@ -71,9 +71,4 @@ class Token(Base, mixins.Timestamps):
                 .first())
 
     def regenerate(self):
-        self.value = self.prefix + _token()
-
-
-def _token():
-    """Return a random string suitable for use in an API token."""
-    return binascii.hexlify(os.urandom(16))
+        self.value = self.prefix + security.token_urlsafe()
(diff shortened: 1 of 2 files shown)


seanh:
Just following up on https://github.com/hypothesis/h/pull/4321#discussion_r97760470

h/models/token.py
```diff
@@ -71,9 +71,4 @@ def get_dev_token_by_userid(cls, session, userid):
                 .first())
 
     def regenerate(self):
-        self.value = self.prefix + _token()
-
-
-def _token():
-    """Return a random string suitable for use in an API token."""
-    return binascii.hexlify(os.urandom(16))
+        self.value = self.prefix + security.token_urlsafe()
```
chdorner: Although it isn't the nicest way of testing, it might make sense to add one new test that checks that we're using `security.token_urlsafe()`, which basically tests that whatever we're using, it will be safe to put it in the URL.

seanh:
Added tests

nickstenning:
Coming to this a little late, but @seanh when you get a moment would you update the example tokens in the documentation so they look like the ones we now generate?

PR #4326, opened 2017-01-25
Return refresh tokens along with access tokens

seanh:
When a client asks for a short-lived API access token, also return the
access token's refresh token along with the access token itself in the
response.

This is compatible with the OAuth 2.0 access token response in RFC 6749.

### DIFF
--- a/h/views/api_auth.py
+++ b/h/views/api_auth.py
@@ -16,11 +16,16 @@ def access_token(request):
         grant_type=request.POST.get('grant_type'))
     token = svc.create_token(user, authclient)
 
-    return {
+    response = {
         'access_token': token.value,
         'token_type': 'bearer',
         'expires_in': TOKEN_TTL.total_seconds(),
     }
+
+    if token.refresh_token:
+        response['refresh_token'] = token.refresh_token
+
+    return response
 
 
 @cors_json_view(context=OAuthTokenError)
--- a/tests/h/views/api_auth_test.py
+++ b/tests/h/views/api_auth_test.py
@@ -28,35 +28,32 @@ class TestAccessToken(object):
         oauth_service.create_token.assert_called_once_with(
             mock.sentinel.user, mock.sentinel.authclient)
 
-    def test_it_returns_an_oauth_compliant_response(self, factories, pyramid_request, oauth_service):
-        token = factories.Token()
-        oauth_service.create_token.return_value = token
-
+    def test_it_returns_an_oauth_compliant_response(self, pyramid_request, token):
         response = views.access_token(pyramid_request)
 
         assert response['access_token'] == token.value
         assert response['token_type'] == 'bearer'
         assert response['expires_in'] == TOKEN_TTL.total_seconds()
 
-    def test_it_returns_the_refresh_token_if_the_token_has_one(self, factories, pyramid_request, oauth_service):
-        token = factories.Token()
-        refresh_token = token.refresh_token = 'test_refresh_token'
-        oauth_service.create_token.return_value = token
+    def test_it_returns_the_refresh_token_if_the_token_has_one(self, pyramid_request, token):
+        token.refresh_token = 'test_refresh_token'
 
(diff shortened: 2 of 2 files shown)


h/views/api_auth.py
```diff
@@ -20,6 +20,7 @@ def access_token(request):
         'access_token': token.value,
         'token_type': 'bearer',
         'expires_in': TOKEN_TTL.total_seconds(),
+        'refresh_token': token.refresh_token,
```
chdorner: I'm thinking we should maybe only set the `refresh_token` key when the value of it is not `None` or empty. A good example is that the way the test is written, the response will include: `"refresh_token": null`.

Maybe something along the lines of:
```python
payload = {
    'access_token': token.value,
    'token_type': 'bearer',
    'expires_in': TOKEN_TTL.total_seconds()
}
if token.refresh_token is not None:
    payload['refresh_token'] = token.refresh_token
return payload
```

seanh: Access tokens (any tokens that have an expires, such as those created by the oauth service) always have refresh tokens.

It worries me a little that we only use two types of token (dev tokens with no expires or refresh, and access tokens with both an expires and a refresh) but we're letting some database complexity leak out into services and views (the fact that at the db level, because for efficiency reasons we use a single table for both types of token, the table could contain other combinations, e.g. a token with an expires but no refresh or a refresh but no expires).

The model code is written such that a token always has neither an expires nor a refresh, or it has both. Should the services that use the model be able to assume this? And the oauth service is written so that the tokens it returns always have expires and refresh. Should the views that use the service be able to assume this?

I'm happy to add this check if you still want it though

chdorner: I don't see this as database complexity leak out into other parts of the code. It's domain object complexity, there are tokens with and without expires / refresh tokens.

In the future we might want to whitelist specific clients that can get non-expiring tokens, as an example. This is just the inherent complexity of API tokens, and this view we're talking about here is responsible for returning any type of token that was created through the OAuth 2 token endpoint.

According to the [RFC](https://tools.ietf.org/html/rfc6749#section-5.1) it's optional for the authorization server to issue and return a refresh token, which to me means that we shouldn't render the `refresh_token` key if the token does not contain refreshing capabilities.

I would maybe even say that doing it this way makes the response generation for the token view token type agnostic, it renders the information is has. To play this even further, we might want to also only render `expires_in` when it is set, and get the time delta from `token.expires - token.created`.

seanh: Thanks Christof, I'm off today (at least this morning) but agree with your take on it

seanh: Ok, I've pushed this change.

I noticed that the view code always returns an `expires_in` even if the token does not have an expiry time. It assumes that the oauth service always returns tokens that expire. Sounds like it would be worth making this also check whether or not the token has an expiry time? (Although then we'd have to change how `expires_in` is calculated).

<pre><code>
response = {
    'access_token': token.value,
    'token_type': 'bearer',
    <strong>'expires_in': TOKEN_TTL.total_seconds()</strong>,
}
</code></pre>

chdorner: Yes, I think it would be a good idea to remove this hard-coded way of returning the `expires_in` and calculate the amount of time until the token expires and return that. And for non-expiring tokens it should probably not include this key in the response (depending on what the RFC says).

seanh: Hmm, all I could find was this: https://tools.ietf.org/html/rfc6749#section-4.2.2

`expires_in` is RECOMMENDED, and if it's left out then we SHOULD provide the expiration time "by other means" or document its default value. The spec seems like it might be written to assume that all tokens expire.

My reading of that, though, is that we could omit `expires_in` for non-expiring tokens and then document, in our API docs, that when there's no `expires_in` the default is that the token doesn't expire.

So as far as the code goes, we can just change it to omit `expires_in` from the response if the token has no expires.

PR #4327, opened 2017-01-25
Optional AWS Auth v4 request signing

chdorner:
This is basically a copy of hypothesis/bouncer#34.

This commit enables AWS Auth v4 request signing if all three of the following environment variables are set:

    ELASTICSEARCH_AWS_ACCESS_KEY_ID
    ELASTICSEARCH_AWS_SECRET_ACCESS_KEY
    ELASTICSEARCH_AWS_REGION

This allows us to make requests to an AWS Elasticsearch cluster without knowing in advance the public IP from which those requests will be made.

When these environment variables are not set, the configuration is exactly as before.


nickstenning:
LGTM. Thanks for writing the missing tests for `get_client`!

PR #4330, opened 2017-01-26
Support passing tokens with HTTP Basic Auth for API and WebSocket 

chdorner:
@robertknight mentioned that the [WebSocket JavaScript API](https://developer.mozilla.org/en-US/docs/Web/API/WebSocket) does not support passing any custom headers to the request. The way we initially thought we would authenticate websocket connections is the same way as we do API requests, with a `Authorization: Bearer {token}` header.

What does work with the WebSocket JavaScript API is passing the HTTP basic auth username and password in the URL (`wss://username:password@hypothes.is/ws`), which lead us to implement another way of authenticating requests by passing the token as the HTTP basic auth password.
Note: Some implementations (Firefox, curl) do not support an empty username, in this case the client should just send a dummy string in the username (e.g. `wss://X:{token}@hypothes.is/ws`).

### DIFF
--- a/h/auth/tokens.py
+++ b/h/auth/tokens.py
@@ -46,10 +46,9 @@ class LegacyClientJWT(object):
     Exposes the standard "auth token" interface on top of legacy tokens.
     """
 
-    def __init__(self, body, key, audience=None, leeway=240):
+    def __init__(self, body, key, leeway=240):
         self.payload = jwt.decode(body,
                                   key=key,
-                                  audience=audience,
                                   leeway=leeway,
                                   algorithms=['HS256'])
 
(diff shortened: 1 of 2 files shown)


chdorner:
Marking WIP as browsers apparently don't support this, we should have investigated more yesterday before jumping on this solution. Anyway, @robertknight is doing some research first.

robertknight:
After some additional testing of different browsers using a tiny Go server, it looks like basic auth isn't going to work for us.

1. Current versions of Firefox and Chrome _will_ send an `Authorization` header with a base64-encoded version of the user and password embedded in the URL but not in the initial HTTP handshake request. Instead the server must respond with a 401 status and `WWW-Authenticate` field in the header if no auth is present and the browser will then retry with HTTP Basic credentials.
2. IE and Edge do not appear to support sending basic auth credentials derived from the URL at all.

chdorner:
Closing, the solution to the browser issues with basic auth is in #4334.

PR #4331, opened 2017-01-27
Set X-Forwarded-Proto header for WordPress proxying in nginx config

chdorner:
Making the WordPress-generated URLs use the correct protocol on Skyliner we need to forward the correct protocol of the initial request here to the WordPress Skyliner application. I will probably have to do something similar with the [nginx](https://github.com/hypothesis/wordpress/blob/master/nginx.conf) configuration in our WordPress docker image.

### DIFF
--- a/bin/start-nginx
+++ b/bin/start-nginx
@@ -8,7 +8,7 @@ CONF_TEMPLATE=/etc/nginx/nginx.conf.tpl
 CONF_DEST=/etc/nginx/nginx.conf
 
 if [ -n "$FALLBACK_UPSTREAM" ]; then
-    FALLBACK_TEXT="proxy_set_header Host \$host; proxy_pass $FALLBACK_UPSTREAM;"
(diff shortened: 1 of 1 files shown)


nickstenning:
Ah, not sure this is where you want to make the change. I'm not sure, but I'd guess the internal ALB in front of WordPress will override this (it's going to be an HTTP-only ALB), so you may need to hardcode it in the WordPress nginx config (i.e. `proxy_set_header X-Forwarded-Proto https;`.

Might be worth checking if the internal ALB respects incoming `X-Forwarded-Proto` or overrides it, because forwarding it all the way through like this _is_ probably safer.

chdorner:
I've found an easier way of doing this, especially because we're the only ones using our own wordpress docker container, I can just force WordPress to generate https urls by setting `$_SERVER['HTTPS'] = 'on';` in the `wp-config.php`.

Closing this.

PR #4333, opened 2017-01-27
Extend `/api/profile` to logged-in third-party users

fatbusinessman:
As the next stage of hypothesis/product-backlog#119, this extends the `/api/profile` endpoint to work for authenticated users belonging to a third party. Part of this involves extracting an `AuthorityGroupService` to list the public groups under an authority, so that the `session` module no longer needs hard-coded knowledge about the `__world__` pseudo-group.

### DIFF
--- a/h/groups/util.py
+++ b/h/groups/util.py
@@ -7,7 +7,7 @@ from pyramid import security
 
 class WorldGroup(object):
     """
-    A Group object for the __world__ group, it only implements __acl__.
+    A Group object for the __world__ group.
 
     This is so we don't have to store a __world__ group in the database.
     """
@@ -21,3 +21,15 @@ class WorldGroup(object):
             (security.Allow, 'authority:{}'.format(self.auth_domain), 'write'),
             security.DENY_ALL,
         ]
+
+    @property
+    def name(self):
+        return 'Public'
+
+    @property
+    def pubid(self):
+        return '__world__'
+
+    @property
+    def is_public(self):
+        return True
--- a/h/session.py
+++ b/h/session.py
@@ -2,7 +2,6 @@
 
 from pyramid.session import SignedCookieSessionFactory
 
-from h.groups import util
 from h.security import derive_key
 
 
@@ -51,19 +50,12 @@ def _current_groups(request):
 
     user = request.authenticated_user
     authority = user.authority if user else request.auth_domain
+    authority_groups = (request.find_service(name='authority_group')
+                        .public_groups(authority=authority))
 
-    groups = (_authority_groups(request.auth_domain, authority) +
-              _user_groups(user))
+    groups = authority_groups + _user_groups(user)
 
     return [_group_model(request.route_url, group) for group in groups]
-
-
-def _authority_groups(auth_domain, authority):
-    """Return the default groups associated with an authority."""
-    if authority == auth_domain:
-        return [util.WorldGroup(auth_domain)]
-    else:
-        return []
 
 
 def _user_groups(user):
--- a/h/session.py
+++ b/h/session.py
@@ -47,10 +47,12 @@ def _current_groups(request):
     This list is meant to be returned to the client in the "session" model.
 
     """
-    groups = [
-        {'name': 'Public', 'id': '__world__', 'public': True},
-    ]
+
     user = request.authenticated_user
+    authority = user.authority if user else request.auth_domain
+
+    groups = _authority_groups(request.auth_domain, authority)
+
     if user is None:
         return groups
     for group in sorted(user.groups, key=_group_sort_key):
@@ -62,6 +64,14 @@ def _current_groups(request):
                                      slug=group.slug),
         })
     return groups
+
+
+def _authority_groups(auth_domain, authority):
+    """Return the default groups associated with an authority."""
+    if authority == auth_domain:
+        return [{'name': 'Public', 'id': '__world__', 'public': True}]
+    else:
+        return []
 
 
 def _user_preferences(user):
--- a/h/session.py
+++ b/h/session.py
@@ -2,6 +2,7 @@
 
 from pyramid.session import SignedCookieSessionFactory
 
+from h.groups import util
 from h.security import derive_key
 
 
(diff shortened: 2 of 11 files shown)


h/session.py
```diff
@@ -47,23 +47,39 @@ def _current_groups(request):
...
     """
-    groups = [
-        {'name': 'Public', 'id': '__world__', 'public': True},
-    ]
+
     user = request.authenticated_user
-    if user is None:
-        return groups
-    for group in sorted(user.groups, key=_group_sort_key):
+    authority = user.authority if user else request.auth_domain
```
chdorner: If I remember the discussions correctly, the client will need to pass us the authority for which it wants the profile (and the groups).

If I'm not mistaken, an unauthenticated request on a publisher site will return the `__world__` group here, instead of the publisher group which is word-readable as well.

chdorner: Diff changed! I'll copy this to the new location.

h/session.py
```diff
@@ -47,21 +48,43 @@ def _current_groups(request):
     This list is meant to be returned to the client in the "session" model.
 
     """
-    groups = [
-        {'name': 'Public', 'id': '__world__', 'public': True},
-    ]
+
     user = request.authenticated_user
+    authority = user.authority if user else request.auth_domain
```
chdorner: If I remember the discussions correctly, the client will need to pass us the authority for which it wants the profile (and the groups).

If I'm not mistaken, an unauthenticated request on a publisher site will return the __world__ group here, instead of the publisher group which is word-readable as well.

fatbusinessman: Correct on both counts. I haven’t yet tweaked this to accommodate unauthenticated users (and the `authority` parameter that goes along with it), which is why the `effective_principals` is still in place on the view.

tests/functional/test_api.py
```diff
@@ -88,6 +100,30 @@ def user_with_token(user, db_session, factories):
 
 
 @pytest.fixture
+def auth_client(db_session, factories):
+    auth_client = factories.AuthClient(authority='thirdparty.example.org')
+    db_session.add(auth_client)
```
nickstenning: FWIW you [don't need to explicitly add](https://github.com/hypothesis/h/blob/b8024a0fc0ea37bd2fdc54d2d3adba087aff116d/tests/common/factories.py#L50) ORM objects created through the `factories` fixture.

h/session.py
```diff
@@ -47,21 +48,43 @@ def _current_groups(request):
...
+
+    return [_group_model(request.route_url, group) for group in groups]
+
+
+def _authority_groups(auth_domain, authority):
+    """Return the default groups associated with an authority."""
+    if authority == auth_domain:
+        return [util.WorldGroup(auth_domain)]
+    else:
+        return []
```
robertknight: For a logged-in third party user, this currently results in an empty list of groups being returned to the client because the user is not an explicit member of a publisher group (as returned by `user.groups`).

See [Slack discussion](https://hypothes-is.slack.com/archives/public/p1485943172003668)

fatbusinessman:
This needs a bit of tweaking for the behaviour of publisher groups: turns out they don’t work in quite the way I thought they did (in that they are _unjoinable_, rather than _automatically joined_).

h/groups/util.py
```diff
@@ -21,3 +21,15 @@ def __acl__(self):
             (security.Allow, 'authority:{}'.format(self.auth_domain), 'write'),
             security.DENY_ALL,
         ]
+
+    @property
```
robertknight: There is a comment above about this group only implementing `__acl__` which needs updating, if these properties are considered part of the interface that groups must implement.

tests/functional/test_api.py
```diff
@@ -61,6 +61,18 @@ def test_profile_api(self, app, user_with_token):
         assert res.json['userid'] == user.userid
         assert [group['id'] for group in res.json['groups']] == ['__world__']
 
+    def test_third_party_profile_api(self, app, third_party_user_with_token):
```
robertknight: 👍  having a few more functional tests for the API.

tests/h/session_test.py
```diff
@@ -85,14 +141,106 @@ def test_profile_sorts_groups():
...
+def test_world_group_is_public_in_profile(fake_user, auth_domain):
+    request = mock.Mock(authenticated_user=fake_user, auth_domain=auth_domain)
+
+    profile = session.profile(request)
+    world_group = [g for g in profile['groups'] if g['id'] == '__world__'][0]
+
+    assert world_group['public'] is True
+
+
+def test_private_group_is_not_public_in_profile(auth_domain):
```
robertknight: Some of these tests for `session.profile` look like they are repeating the tests for `session.model` quite a bit. Given that `h.session` is exactly the same as `h.profile` except for the addition of the `csrf` token, perhaps you could simplify the `session` tests to just checking that they return the value of `profile` with that one parameter added?

fatbusinessman: That’s certainly an option, although in a [prior discussion with @chdorner](https://github.com/hypothesis/h/pull/4317/commits/5d386431d9f9f995299fab29d2ad9cbdcac728ce#r97614858) we came to the conclusion that duplicating the tests was a reasonable option, under the premise that we’re hoping to get rid of `session.model` in the future.

robertknight: That seems reasonable. I'll leave it up to you.

tests/functional/test_api.py
```diff
@@ -87,6 +101,34 @@ def user_with_token(user, db_session, factories):
...
+@pytest.fixture
+def third_party_user(auth_client, db_session, factories):
+    user = factories.User(authority=auth_client.authority)
+    db_session.commit()
+    return user
+
+
+@pytest.fixture
+def publisher_group(auth_client, db_session, factories):
+    group = factories.PublisherGroup(authority=auth_client.authority)
```
robertknight: Hmm, groups _have_ to have a creator in the current model, yet none is specified here and this test passes.

fatbusinessman: The `PublisherGroup` factory inherits from the `Group` factory, so I think it’s picking up the [`creator` definition from there](https://github.com/hypothesis/h/pull/4333/files#diff-149b85565f69bed7e895157dfc37e09aL103).

robertknight: Ah, I see. Thanks for the pointer.

PR #4334, opened 2017-01-27
Token auth for WebSocket API

chdorner:
This will add support for authenticating websocket connections with a query string parameter, e.g. `wss://hypothes.is/ws?access_token={token}`. This token can be a new OAuth 2 token, a developer token, or a legacy client JWT token.

After some more research, @robertknight found out that HTTP basic auth support for WebSockets is not fully supported by all browsers, so this supersedes #4330.

I had to move quite a bit of things around for a nice implementation of this, as we can't import any models in the Python files that contain the authentication policies.

I suggest @robertknight should try out his client branch with this, and since there are quite a lot of authz related changes it would be great to get at least @nickstenning's eyes on this, others are welcome as well.

### DIFF
--- a/conf/development-websocket.ini
+++ b/conf/development-websocket.ini
@@ -8,6 +8,10 @@ origins:
 # Use gevent-compatible transport for the Sentry client
 raven.transport: gevent
 
+# Set a default persistent secret for development. DO NOT copy this into a
+# production settings file.
+h.client_secret: nosuchsecret
+
 secret_key: notverysecretafterall
 
 # SQLAlchemy configuration -- See SQLAlchemy documentation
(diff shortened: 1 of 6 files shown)


robertknight:
I have tested this branch with https://github.com/hypothesis/client/pull/200 and this diff to disable the fallback to cookie auth:

```diff
diff --git a/h/auth/__init__.py b/h/auth/__init__.py
index ecbab58..230e28a 100644
--- a/h/auth/__init__.py
+++ b/h/auth/__init__.py
@@ -27,7 +27,7 @@ TOKEN_PARAM_POLICY = TokenParameterAuthenticationPolicy(callback=groupfinder, de
 
 DEFAULT_POLICY = AuthenticationPolicy(api_policy=TOKEN_POLICY,
                                       fallback_policy=TICKET_POLICY)
-WEBSOCKET_POLICY = MultiAuthenticationPolicy([TOKEN_PARAM_POLICY, TOKEN_POLICY, TICKET_POLICY])
+WEBSOCKET_POLICY = MultiAuthenticationPolicy([TOKEN_PARAM_POLICY])
```

This now works as expected.

chdorner:
Closing in favour of #4345

PR #4335, opened 2017-01-30
Make the User factory into a ModelFactory

fatbusinessman:
This factory was originally created as a subclass of the base `Factory` (see commit a928eb5e) rather than our custom `ModelFactory`, which meant it would not add the objects it generated to the current session. This is different behaviour to the other factories, which can be a little confusing. We also have the `build` method, which achieves the same outcome.

I've gone through the tests that invoke this factory, and I think I've got them all working as intended. Where the tests rely on patching out method calls, I've assumed that this means the behaviour of not adding the new objects to the session is intended, and changed the call to a `build` call. Where the factory is immediately followed by `session.add` calls, I've removed them as they now happen automatically. In other cases, I've mostly assumed that adding the new object to the session or not has no effect on the test, and have left the method as it stands.

### DIFF
--- a/tests/common/factories.py
+++ b/tests/common/factories.py
@@ -70,14 +70,9 @@ class AuthClient(ModelFactory):
     secret = factory.LazyAttribute(lambda _: unicode(FAKER.sha256()))
 
 
-class User(factory.Factory):
+class User(ModelFactory):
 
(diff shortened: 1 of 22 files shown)


nickstenning:
Thanks for cleaning up this mess! :cake:

PR #4336, opened 2017-01-30
Add Token.get_by_refresh_token() classmethod

seanh:
This is needed to implement the OAuth 2 refresh token request in which
the client submits a refresh token (only) and we need to find the
corresponding API token in order to make and return a new token for the
same user and authclient.

### DIFF
--- a/h/models/token.py
+++ b/h/models/token.py
@@ -64,6 +64,19 @@ class Token(Base, mixins.Timestamps):
             self.refresh_token = security.token_urlsafe()
 
     @classmethod
+    def get_by_refresh_token(cls, session, refresh_token):
+        """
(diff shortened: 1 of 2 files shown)


nickstenning:
I'd much rather see us adding query functionality to an appropriate service than adding endless `get_by_*` fetchers to the models. See, for example, the AuthTicket service and its corresponding model:

- https://github.com/hypothesis/h/blob/1953457d7f88c96b42cdc5c8bfa69db1e1b3e7f7/h/services/auth_ticket.py
- https://github.com/hypothesis/h/blob/1953457d7f88c96b42cdc5c8bfa69db1e1b3e7f7/h/models/auth_ticket.py

seanh:
Ok, closing this

PR #4337, opened 2017-01-30
Add Token.expired property

seanh:
This'll be used by OAuthService to check whether a token has expired
when verifying it.

### DIFF
--- a/h/models/token.py
+++ b/h/models/token.py
@@ -63,6 +63,11 @@ class Token(Base, mixins.Timestamps):
         if self.expires:
             self.refresh_token = security.token_urlsafe()
 
+    @property
+    def expired(self):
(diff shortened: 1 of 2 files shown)


nickstenning:
Again, sorry for coming to this late, but both reference to `datetime.datetime.now()` here should be `datetime.datetime.utcnow()` instead.

fatbusinessman:
> Again, sorry for coming to this late, but both references to `datetime.datetime.now()` here should be `datetime.datetime.utcnow()` instead.

Oh yes—`now` returns a naïve object by default, doesn’t it? Oops. 😕

seanh:
D'oh, I always forget that, I'll send a quick PR to correct it...

PR #4338, opened 2017-01-30
Allow to set statsd prefix

chdorner:
This should be the last piece before we can thoroughly test `h` on skyliner.

The biggest downside here is that starting supervisor will fail if the `ENV`, `APP`, or `HOSTNAME` environment variables are not set. This is not an issue for us, because all of them are set by Skyliner by default, but it makes it harder for other people to run the docker container with its default settings.

I've tried multiple ways to achieve this with Skyliner environment variables and the docker env file, but couldn't get the hostname into the string. Similarly, I've been trying to check if I can get supervisor to not fail when one of the env variables is not set, but this doesn't seem to be an option as they basically just use the Python string interpolation (`"%(ENV_APP)s" % {"ENV_APP": "foobar"}`) and there is no support for a default value without having access to the dictionary.

### DIFF
--- a/gunicorn.conf.py
+++ b/gunicorn.conf.py
@@ -22,11 +22,9 @@ if not os.environ.get('GUNICORN_STATS_DISABLE', None):
         _port = os.environ.get('STATSD_PORT', '8125')
         statsd_host = '{}:{}'.format(_host, _port)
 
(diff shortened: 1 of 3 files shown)


nickstenning:
LGTM.

PR #4340, opened 2017-01-31
Update Token code and tests to use UTC timestamps

fatbusinessman:
As the timestamp is stored in UTC in the database, and `datetime.now` returns an object that doesn't understand timezones, this may cause time offset bugs, depending on the server's configuration.

This is a fix to #4337, as pointed out by @nickstenning.

### DIFF
--- a/h/models/token.py
+++ b/h/models/token.py
@@ -66,7 +66,7 @@ class Token(Base, mixins.Timestamps):
     @property
     def expired(self):
         """True if this token has expired, False otherwise."""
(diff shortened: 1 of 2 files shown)


nickstenning:
This is :+1:

Just for clarity, though, `datetime.datetime.utcnow()` returns a "naive", TZ-less object as well, but if the server is running on a non-UTC timezone it will still be a UTC timestamp. As you say, by convention *all* timestamps handled by the application and stored in the database are UTC.

PR #4341, opened 2017-01-31
Update example dev API token in docs

seanh:
Make it look like what real dev tokens look like now that they're
generated using token_urlsafe().


seanh:
@nickstenning

nickstenning:
:+1: thank you!

PR #4343, opened 2017-01-31
Use 'example.com' as authority in functional tests

fatbusinessman:
The functional tests default to an authority of `localhost`, whereas users, groups and auth clients are created using `example.com`. This causes some problems in the profile endpoint, where some of the logic involves checking the user's authority against the `auth_domain` of the request.

For the sake of consistency, going with `example.com` everywhere seems reasonable.


seanh:
:+1:

PR #4344, opened 2017-01-31
Refresh token request

seanh:
This pr adds support for getting a new access token by using the refresh_token that came with a not-yet-expired access token:

```
POST /api/token
grant_type=refresh_token&refresh_token=***
```

Using a refresh token does not expire or do anything else to the old access token, which will continue to work (until it expires). A follow-up PR could do something here.

* Small OAuthService refactor to make way for refresh tokens
* Add verifying refresh_token requests, as well as the existing jwt-bearer requests, to OAuthService. The /api/token view code doesn't need to change
* Some deduplication of test boilerplate
* Added functional tests, to prove to myself that getting and using access tokens and refresh tokens actually works. This is probably a good thing to have a few functional tests for.

Details in the commit messages.

### DIFF
--- a/h/services/oauth.py
+++ b/h/services/oauth.py
@@ -41,6 +41,7 @@ class OAuthService(object):
 
         verifiers = {
             'urn:ietf:params:oauth:grant-type:jwt-bearer': self._verify_jwt_bearer,
+            'refresh_token': self._verify_refresh_token,
         }
 
         try:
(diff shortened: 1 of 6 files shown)


h/services/oauth.py
```diff
@@ -21,7 +21,38 @@ def __init__(self, session, user_service, domain):
...
+
+        :raises OAuthTokenError: if the request is invalid
+
+        """
+        grant_type = body.get('grant_type')
+
+        verifiers = {
+            'urn:ietf:params:oauth:grant-type:jwt-bearer': self._verify_jwt_bearer,
+            'refresh_token': self._verify_refresh_token,
+        }
```
chdorner: Nice! I like how you did this.

h/services/oauth.py
```diff
@@ -88,13 +114,40 @@ def verify_jwt_bearer(self, assertion, grant_type):
...
+
+        if not refresh_token:
+            raise OAuthTokenError('required refresh_token parameter is missing',
+                                  'invalid_request')
+
+        if type(refresh_token) != text_type:
+            raise OAuthTokenError('refresh_token is invalid', 'invalid_refresh')
+
+        token = (self.session.query(models.Token)
+                 .filter_by(refresh_token=refresh_token)
```
chdorner: We should add an index to the `refresh_token` column.

tests/functional/test_oauth.py
```diff
@@ -0,0 +1,146 @@
...
+        app.get(
+            '/api/profile',
+            headers={'Authorization': str('Bearer {}'.format(access_token))},
+        )
+
+    def test_request_fails_if_access_token_wrong(self, app, authclient,
+                                                 userid):
+        app.get(
+            '/api/profile',
+            headers={'Authorization': str('Bearer {}'.format('wrong'))},
```
chdorner: Nitpick, but this would read nicer to just do `str('Bearer wrong')`.

chdorner:
Tested it locally and it works great 👍

h/services/oauth.py
seanh: Done

tests/functional/test_oauth.py
seanh: Done

PR #4345, opened 2017-01-31
Query parameter token auth for WebSocket clients

chdorner:
This will add support for authenticating websocket connections with a query string parameter, e.g. wss://hypothes.is/ws?access_token={token}. This token can be a new OAuth 2 token, a developer token, or a legacy client JWT token.

After some more research, @robertknight found out that HTTP basic auth support for WebSockets is not fully supported by all browsers, so this supersedes #4330.

I had to move quite a bit of things around for a nice implementation of this, as we can't import any models in the Python files that contain the authentication policies.

### DIFF
--- a/h/services/auth_token.py
+++ b/h/services/auth_token.py
@@ -1,0 +1,63 @@
+# -*- coding: utf-8 -*-
+
+from __future__ import unicode_literals
+
+import jwt
+
+from h import models
+from h.auth.tokens import LegacyClientJWT, Token
+
+
+class AuthTokenService(object):
+    def __init__(self, session, client_secret, host_url):
+        self._session = session
+        self._client_secret = client_secret
+        self._host_url = host_url
+
+        self._validate_cache = {}
+
+    def validate(self, token_str):
+        """
+        Load and validate a token.
+
+        This will return a token object implementing
+        ``h.auth.interfaces.IAuthenticationToken``, or ``None`` when the token
+        cannot be found, is not a legacy JWT token, or is not valid.
+
+        :param token_str: the token string
+        :type token_str: unicode
+
+        :returns: the token object, if found and valid, or ``None``.
+        """
+
+        if token_str in self._validate_cache:
+            return self._validate_cache[token_str]
+
+        token = self._fetch_token(token_str)
+        self._validate_cache[token_str] = token
+        if token is not None and token.is_valid():
+            return token
+        return None
+
+    def _fetch_token(self, token_str):
+        token_model = (self._session.query(models.Token)
+                       .filter_by(value=token_str)
+                       .one_or_none())
+        if token_model is not None:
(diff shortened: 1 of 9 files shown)


chdorner:
@robertknight I tested against client master and the token auth in the parameters seems to be working. but it might be good if you could test again, just to make sure

robertknight:
@chdorner - I've just tested this PR with the master build of the client and it looks to be working as expected. I have put up a [client PR](https://github.com/hypothesis/client/pull/204) to make use of the WS's "whoami/whoyouare" messages to verify that the WebSocket userid matches the client's expected userid.

h/services/auth_token.py
```diff
@@ -0,0 +1,62 @@
...
+        ``h.auth.interfaces.IAuthenticationToken``, or ``None`` when the token
+        cannot be found, is not a legacy JWT token, or is not valid.
+
+        :param token_str: the token string
+        :type token_str: unicode
+
+        :returns: the token object, if found and valid, or ``None``.
+        """
+
+        if token_str in self._validate_cache:
```
robertknight: Hmm ... are there any scenarios in which this cache needs to be explicitly invalidated?

tests/h/auth/policy_test.py
```diff
@@ -124,38 +125,74 @@ def test_unauthenticated_userid_is_none_if_no_token(self, pyramid_request):
...
 
         assert result is None
 
-    def test_authenticated_userid_uses_callback(self, fake_token, pyramid_request):
+    def test_unauthenticated_userid_returns_userid_from_query_params_token(self, pyramid_request):
+        """When the path is `/ws` then we look into the query string parameters as well."""
+
+        policy = TokenAuthenticationPolicy()
+        pyramid_request.GET['access_token'] = 'valid123'
+        pyramid_request.path = '/ws'
```
robertknight: An alternative to checking a hard-coded path might be to check one of the headers in the [upgrade request](https://developer.mozilla.org/en-US/docs/Web/API/WebSockets_API/Writing_WebSocket_servers#Client_Handshake_Request) such as the `Sec-WebSocket-Version` header, which browser-based clients are not able to set themselves. This would work if WS connections were ever served from a different path. Thoughts?

tests/h/auth/tokens_test.py
```diff
@@ -36,100 +35,79 @@ def test_token_with_past_expiry_is_not_valid(self):
 
 VALID_TOKEN_EXAMPLES = [
     # Valid
-    lambda a, k: jwt.encode({'aud': a, 'exp': _seconds_from_now(3600)},
```
robertknight: I've just checked the client code and it looks like we don't ever look at the `aud` property, so this should be fine.

robertknight:
LGTM

PR #4347, opened 2017-01-31
Don't return expires_in if token doesn't expire

seanh:
This doesn't change how expires_in is calculated but just omits it for tokens that don't expire.


seanh:
@chdorner

PR #4349, opened 2017-02-01
Revert "Remove unused group leave view"

robertknight:
This reverts commit eb0baaa39721e9aef7c88ce6225284486fb3633b.

This view is still used by the client.

Fixes hypothesis/client#203

----

To avoid this confusion in future, I suggest we should provide a proper API-based method for leaving a group. I'm restoring this route for the moment since existing clients in the wild depend on it.

### DIFF
--- a/h/views/groups.py
+++ b/h/views/groups.py
@@ -124,10 +124,13 @@ def read_noslug(group, request):
     check_slug(group, request)
 
 
+# FIXME: This view is only used by the client, it needs to be refactored
+#        into a proper API endpoint under the /api namespace.
 @view_config(route_name='group_leave',
              request_method='POST',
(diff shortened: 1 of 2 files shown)


h/views/groups.py
```diff
@@ -124,6 +124,17 @@ def read_noslug(group, request):
     check_slug(group, request)
 
 
+@view_config(route_name='group_leave',
+             request_method='POST',
+             effective_principals=security.Authenticated)
+def leave(group, request):
+    """Route for leaving a group. Used by the Hypothesis client."""
```
chdorner: Can you please add another comment with a big FIXME that explains the real solution of providing a proper API endpoint?

chdorner: Done

PR #4350, opened 2017-02-01
Add API endpoint for updating user preferences

chdorner:
This is the backend work for hypothesis/product-backlog#133.

This adds a new API endpoint for updating user preferences, we currently only have `show_sidebar_tutorial`, so it will work with a request like:

```
PATCH /api/profile
{
  "preferences": {
    "show_sidebar_tutorial": false
  }
}
```

### DIFF
--- a/h/services/user.py
+++ b/h/services/user.py
@@ -7,6 +7,8 @@ import sqlalchemy
 from h.models import Annotation, User
 from h import util
 
+UPDATE_PREFS_ALLOWED_KEYS = set(['show_sidebar_tutorial'])
+
 
 class LoginError(Exception):
     pass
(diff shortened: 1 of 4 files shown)


robertknight:
LGTM. On the client side, the changes needed to use this are:

1. Add the endpoint to the set of API methods in `src/sidebar/store.js`
2. Modify the `dismissSidebarTutorial` export in `src/sidebar/session.js` to be a function which calls the API method (eg. `store.profile.update`) and calls the `update` function with the response to update the local application state.

chdorner:
@robertknight thanks for the hints, I decided not to look at them and come up with my own implementation. I basically came to the exact same conclusions as your list 👍

PR #4351, opened 2017-02-02
Add fake request class to remove test duplication

fatbusinessman:
A lot of these tests were building up the same set of nested mock objects as one another, with different sets of groups and features.  For the sake of simplicity (and at @nickstenning's suggestion), this sets up a `FakeRequest` class that sets up the request methods that the `model` and `profile` functions expect, and that knows how to customise its own internals in a few set ways. This means the user fixtures are no longer referenced and can be removed.

I found that this approach didn't work very nicely with the heavily-parametrized tests around user preferences, so I split these into separate tests for authenticated and unauthenticated requests, which removed a good chunk of the complexity.

(Further context: I pulled this change out of #4333 as a discrete reviewable chunk.)

### DIFF
--- a/tests/h/session_test.py
+++ b/tests/h/session_test.py
@@ -11,62 +11,45 @@ class FakeGroup(object):
         self.slug = pubid
 
 
-def test_model_sorts_groups():
-    fake_user = mock.Mock()
-    fake_user.groups = [
+def test_model_sorts_groups(authenticated_request):
+    authenticated_request.set_groups([
         FakeGroup('c', 'Group A'),
         FakeGroup('b', 'Group B'),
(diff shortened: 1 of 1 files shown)


nickstenning:
:+1: nice improvement. 

One general comment to consider for the future: wherever possible it's a good idea to use stub objects that are as close as possible to the real thing in preference to generic `mock.Mock` instances. Sometimes that'll be made easier by factoring out business logic into more functional components (either just functions or a service) which accept values and return values. Just be wary of getting carried away with `mock.Mock` objects. They make it very easy to write tests that will continue to pass even when the values you're passing into the system under test (in the real code) change.

PR #4353, opened 2017-02-03
Remove cookie-based authentication for WebSocket

robertknight:
This PR removes cookies as an authentication method for the WebSocket, restricting authentication to access tokens passed in the header or via a query string parameter. As a result, it is possible to remove the origin restrictions.

* The first commit changes the auth policy for the WebSocket to token-only
* The second commit removes the origin checks for the WS and associated tests
* The third commit updates the documentation for the Real Time API

The latest Hypothesis client (v0.53, released on 02/02/17) always uses tokens for authentication. Earlier clients only used cookies. If earlier clients try to connect after this change is deployed, they will simply be connected to the WS as unauthenticated clients and will consequently not alert the user to non-public updates or update the groups list in real time.

### DIFF
--- a/h/auth/__init__.py
+++ b/h/auth/__init__.py
@@ -6,7 +6,6 @@ import logging
 
 from pyramid.authentication import RemoteUserAuthenticationPolicy
 import pyramid_authsanity
-from pyramid_multiauth import MultiAuthenticationPolicy
 
 from h.auth.policy import AuthenticationPolicy, TokenAuthenticationPolicy
 from h.auth.util import auth_domain, groupfinder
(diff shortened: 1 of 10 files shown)


robertknight:
I believe you were interested in use of the Real Time API by third party clients @judell

nickstenning:
There's some documentation to update, too:`docs/realtime.rst`

h/auth/__init__.py
```diff
@@ -6,7 +6,6 @@
 
 from pyramid.authentication import RemoteUserAuthenticationPolicy
 import pyramid_authsanity
-from pyramid_multiauth import MultiAuthenticationPolicy
```
nickstenning: I think that removes our dependency on `pyramid_multiauth`.

robertknight: Indeed it does. I've removed the dependency in https://github.com/hypothesis/h/pull/4353/commits/3bdc8ec22e13330ffbe6e2afc83de9780ea2fe8b

nickstenning:
Are we good to merge and deploy this now, @robertknight?

robertknight:
> Are we good to merge and deploy this now, @robertknight?

Yes

PR #4354, opened 2017-02-03
Add /_status health check endpoint

chdorner:
This will be used by Skyliner to determine if a deploy was successful or not.

The endpoint itself is trying to contact Postgres, ElasticSearch, and RabbitMQ. With ElasticSearch it also checks that the cluster status is green.

Some questions:

* We should add tests, but short of mocking out the methods the view is calling, is there another, better way?
* I should probably check for the ElasticSearch cluster status to not be "red" instead of checking that it's "green" ("yellow" I think happens when it reshuffles shards, we don't want to be considered as down during this time). Correct?

### DIFF
--- a/h/views/status.py
+++ b/h/views/status.py
@@ -1,0 +1,58 @@
+# -*- coding: utf-8 -*-
+
+from __future__ import unicode_literals
+
+import logging
+
+from h.celery import celery
+from h.exceptions import APIError
+from h.util.view import json_view
+
+
+log = logging.getLogger(__name__)
+
+
+@json_view(route_name='status')
+def status(request):
+    _check_database(request)
+    _check_search(request)
+    _check_celery()
+    return {'status': 'okay'}
+
+
+def _check_database(request):
+    try:
+        request.db.execute('SELECT 1')
+    except Exception as exc:
+        log.exception(exc)
+        raise APIError('Database connection failed')
+
+
+def _check_search(request):
+    try:
+        info = request.es.conn.cluster.health()
+        if info['status'] not in ('green', 'yellow'):
+            raise APIError('Search cluster state is %s' % info['status'])
+    except Exception as exc:
+        log.exception(exc)
+        raise APIError('Search connection failed')
+
+
+def _check_celery():
+    try:
+        result = celery.control.ping(timeout=0.25)
+        if not result:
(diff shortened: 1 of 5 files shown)


nickstenning:
Looks good. And yes to your last question, [as per bouncer](https://github.com/hypothesis/bouncer/blob/fe400bb9507af754e3b18f3c2b93030a702572be/bouncer/views.py#L134-L143).

h/views/status.py
```diff
@@ -0,0 +1,46 @@
...
+            raise APIError('Search cluster state is %s' % info['status'])
+    except Exception as exc:
+        log.exception(exc)
+        raise APIError('Search connection failed')
+
+
+def _check_realtime(request):
+    try:
+        request.realtime.connection.release()
+        request.realtime.connection.connect()
```
nickstenning: Shouldn't these two lines be the other way around? You probably also want `release` to be in a `finally:` block?

chdorner: In case it's already connected (which it seems to be), it doesn't try to reconnect. Releasing it first and then connecting afterwards was the only way I could get it to fail when I stopped rabbitmq.

This makes me suspicious about how we're closing connections to rabbit during a request in general though.

nickstenning: Hmm. I've checked this out locally and I can't reproduce that behaviour: just a call to `connect()` seems to be enough to fail the healthcheck when Rabbit goes away.

That said, just checking we can connect -- while better than nothing -- doesn't tell us all that much. We could, alternatively, check that the return value from `h.celery.celery.control.ping()` is non-empty? That checks that a) we can make a connection, b) Rabbit is processing messages, and c) we have >0 workers running and handling tasks. What do you think?

nickstenning: Oh and:

> This makes me suspicious about how we're closing connections to rabbit during a request in general though.

It actually makes me suspicious about how we're opening connections to Rabbit, too. We should probably have a threadlocal connection pool and it sounds like we don't.

chdorner: As discussed in Slack, I had some reservations about what happens to `ping` requests when the workers are already busy because of a long job queue, but it turns out that `ping` control requests are not enqueued as jobs, so we don't have to worry about this. This is done now 👍

nickstenning:
LGTM!

PR #4355, opened 2017-02-07
Add feature flag and config setting to load client from an external URL

robertknight:
Add a feature flag (`use_client_boot_script`) and a configuration setting (`h.client_url`) which causes the service to use the client's boot script instead of using the boot code baked into the service itself. When the flag is enabled:

- The `/embed.js` view returns a redirect to the client's entry point at `https://unpkg.com/hypothesis` or the custom location set with the `h.client_url` setting.
- The `app.html` view returns a minimal HTML page which just includes the client URL.

Together, these changes will enable us to deploy and serve the client from a CDN rather than baking it into the Docker image for the service.

----

**Manual testing:**

 - Check out this branch
 - In the client repo, check out https://github.com/hypothesis/client/pull/229.
- In the client repo, set the `H_SERVICE_URL` env var to `http://localhost:5000` and run `gulp watch`
- In the h repo, set the the `CLIENT_URL` env var to `http://localhost:3001/hypothesis` and run `make dev`
- Go to `/docs/help` and `/stream` and verify that the client loads. Also verify that the client's CSS and JS assets are loaded from `http://localhost:3001`.

### DIFF
--- a/conf/app.ini
+++ b/conf/app.ini
@@ -19,7 +19,6 @@ use: call:h.app:create_app
 # If not provided, both default to a random URL-safe base64-encoded string.
 #h.client_id:
 #h.client_secret:
-h.client_url: https://unpkg.com/hypothesis
 
 # SQLAlchemy configuration -- See SQLAlchemy documentation
 sqlalchemy.url: postgresql://postgres@localhost/postgres
--- a/conf/app.ini
+++ b/conf/app.ini
@@ -19,6 +19,7 @@ use: call:h.app:create_app
 # If not provided, both default to a random URL-safe base64-encoded string.
 #h.client_id:
 #h.client_secret:
+h.client_url: https://unpkg.com/hypothesis
 
 # SQLAlchemy configuration -- See SQLAlchemy documentation
 sqlalchemy.url: postgresql://postgres@localhost/postgres
--- a/h/templates/app.html.jinja2
+++ b/h/templates/app.html.jinja2
@@ -9,7 +9,7 @@
      only ever displays a single route.
      #}
     <base target="_top" href="/" />
-  {% for url in app_css_urls %}
+    {% for url in app_css_urls %}
     <link rel="stylesheet" href="{{ url }}">
     {% endfor %}
     {% for attrs in meta_attrs -%}
@@ -34,6 +34,6 @@
     {% for url in app_js_urls %}
     <script src="{{ url }}"></script>
     {% endfor %}
-    
+
   </body>
 </html>
--- a/h/templates/app.html.jinja2
+++ b/h/templates/app.html.jinja2
@@ -9,7 +9,7 @@
      only ever displays a single route.
      #}
     <base target="_top" href="/" />
-    {% for url in app_css_urls %}
+  {% for url in app_css_urls %}
     <link rel="stylesheet" href="{{ url }}">
     {% endfor %}
     {% for attrs in meta_attrs -%}
@@ -34,6 +34,6 @@
     {% for url in app_js_urls %}
     <script src="{{ url }}"></script>
     {% endfor %}
-
+    
   </body>
 </html>
--- a/h/templates/embed.js.jinja2
+++ b/h/templates/embed.js.jinja2
@@ -1,4 +1,4 @@
-{% if client_boot_url %}
+{% if client_boot_script %}
 (function () {
   var configEl = document.createElement('script');
   configEl.className = 'js-hypothesis-settings';
@@ -9,9 +9,7 @@
   });
   document.head.appendChild(configEl);
 
-  var bootScriptEl = document.createElement('script');
-  bootScriptEl.src = '{{ client_boot_url }}';
-  document.head.appendChild(bootScriptEl);
+  {{ client_boot_script | safe }}
 })();
 {% else %}
 (function () {
--- a/h/templates/embed.js.jinja2
+++ b/h/templates/embed.js.jinja2
@@ -1,17 +1,3 @@
-{% if client_boot_script %}
-(function () {
-  var configEl = document.createElement('script');
-  configEl.className = 'js-hypothesis-settings';
-  configEl.type = 'application/json';
-  configEl.textContent = JSON.stringify({
-    assetRoot: '{{ client_asset_root }}',
-    sidebarAppUrl: '{{ app_html_url }}',
-  });
-  document.head.appendChild(configEl);
-
-  {{ client_boot_script | safe }}
-})();
-{% else %}
 (function () {
 // Detect presence of Hypothesis in the page
 var appLinkEl = document.querySelector('link[type="application/annotator+html"]');
@@ -71,4 +57,3 @@ install();
 
 return {installedURL: baseUrl.href};
 })();
-{% endif %}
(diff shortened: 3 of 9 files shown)


robertknight:
WIP because it needs more tests.

tests/h/views/client_test.py
```diff
@@ -19,6 +20,52 @@ def test_annotator_token_returns_token(generate_jwt, pyramid_request):
...
+        ctx = client.embed(pyramid_request_)
+        assert ctx['client_boot_url'] is None
+
+    def test_sets_content_type(self, pyramid_request_):
+        client.embed(pyramid_request_)
+        assert pyramid_request_.response.content_type == 'text/javascript'
+
+
+@pytest.fixture
+def pyramid_request_(pyramid_request):
```
robertknight: This fixture needs to be named differently than the `pyramid_request` fixture which it extends, as otherwise the registry gets reset to a new object before individual test cases actually run.

nickstenning:
I think it would make more sense to have the feature flag operate at the view level and actually request from upstream (and then serve with appropriate cache headers), rather than having embed.js be a bootstrapper for a bootstrapper.

robertknight:
This is currently blocked on https://github.com/hypothesis/h/pull/4363 which simplifies the client views a lot and makes them easier to test.

robertknight:
Slack discussion: https://hypothes-is.slack.com/archives/public/p1487328023004988

conf/app.ini
```diff
@@ -19,6 +19,7 @@ use: call:h.app:create_app
 # If not provided, both default to a random URL-safe base64-encoded string.
 #h.client_id:
 #h.client_secret:
+h.client_url: https://unpkg.com/hypothesis
```
nickstenning: We're slowly but surely phasing out use of these configuration files, so perhaps we could put the default value in an appropriate module in the application?

h/templates/app.html.jinja2
```diff
@@ -9,7 +9,7 @@
      only ever displays a single route.
      #}
     <base target="_top" href="/" />
-  {% for url in app_css_urls %}
+    {% for url in app_css_urls %}
```
nickstenning: Doesn't seem like this diff is relevant to these changes, is it?

h/views/client.py
```diff
@@ -125,11 +162,18 @@ def absolute_asset_urls(bundle_name):
         return [urlparse.urljoin(base_url, url)
                 for url in assets_env.urls(bundle_name)]
 
-    return {
-        'app_html_url': request.route_url('sidebar_app'),
-        'inject_resource_urls': (absolute_asset_urls('inject_js') +
-                                 absolute_asset_urls('inject_css'))
-    }
+    client_boot_url = _resolve_client_boot_url(request)
```
nickstenning: Rather than adding complexity to this view function, it probably makes more sense to have a separate view function which uses the `has_feature_flag` view predicate:

```python
from pyramid.config import not_

...

@view_config(route_name='embed',
             renderer='h:templates/embed.js.jinja2',
             has_feature_flag=not_('use_client_boot_script'))
def embed(request):
    # current view

@view_config(route_name='embed',
             has_feature_flag='use_client_boot_script')
def embed_redirect(request):
    # redirect
```

h/views/client.py
```diff
@@ -125,11 +162,18 @@ def absolute_asset_urls(bundle_name):
...
+
+    if client_boot_url:
+        rsp = HTTPFound(location=client_boot_url)
+        rsp.cache_control.max_age = 60 * 5  # 5 minutes
```
nickstenning: Let's use the `http_cache` arg to `view_config` instead, as that will ensure we set both `Expires` and `C-C` headers.

h/views/client.py
```diff
@@ -85,6 +112,7 @@ def sidebar_app(request, extra=None):
     ga_client_tracking_id = settings.get('ga_client_tracking_id')
 
     ctx = _app_html_context(api_url=request.route_url('api.index'),
+                            client_boot_url=_resolve_client_boot_url(request),
```
nickstenning: Is there a reason we need to actually resolve the URL here? Can we use the value of the setting without making any HTTP requests?

h/templates/embed.js.jinja2
```diff
@@ -1,3 +1,17 @@
+{% if client_boot_script %}
+(function () {
+  var configEl = document.createElement('script');
+  configEl.className = 'js-hypothesis-settings';
+  configEl.type = 'application/json';
+  configEl.textContent = JSON.stringify({
+    assetRoot: '{{ client_asset_root }}',
+    sidebarAppUrl: '{{ app_html_url }}',
```
robertknight: The defaults are now hardcoded to correct values in the client, with one caveat - the `/app.html` URL can only be hardcoded to one value - so if we remove this config it won't be possible to load the QA version of `/app.html` on the QA version of the service.

h/templates/app.html.jinja2
robertknight: No, I think it stems from an indentation correction made during an earlier version of this PR.

conf/app.ini
robertknight: Ah, I didn't know we were phasing these out. I can do that.

h/views/client.py
robertknight: The client will have an `/embed.js` => `https://unpkg.com/hypothesis@0.X.Y` redirect cached locally but won't have an `https://unpkg.com/hypothesis` => `https://unpkg.com/hypothesis@0.X.Y` redirect cached locally because the service proxies the original redirect. So not resolving the URL would require the client to perform an extra request.

In an earlier version of this PR, I just had `/app.html` include `/embed.js` directly (which the client _will_ have cached). Come to think of it, I think I could just revert back to doing that.

h/views/client.py
```diff
@@ -61,8 +94,12 @@ def _app_html_context(assets_env, api_url, service_url, sentry_public_dsn,
             'googleAnalytics': ga_client_tracking_id
         })
 
-    app_css_urls = assets_env.urls('app_css')
-    app_js_urls = assets_env.urls('app_js')
+    if client_url:
+        app_js_urls = [client_url]
```
nickstenning: I thought you were going to use `request.route_path('embed')` here to avoid having to do an HTTP call every time we render `app.html` -- which is notably not served with sensible cache headers at the moment.

h/views/client.py
```diff
@@ -132,6 +176,20 @@ def absolute_asset_urls(bundle_name):
     }
 
 
+@view_config(route_name='embed',
+             has_feature_flag='use_client_boot_script',
+             http_cache=60 * 5)
```
nickstenning: We want this to be cached by the CDN, so need to specify the `public` C-C token: `http_cache=(300, {'public': True})` as with robots.txt.

nickstenning: Hmm, although we should probably not do that while it's controlled by a feature flag. Leave this one for now and we'll follow up once we've deployed it.

nickstenning: Actually, no, it's fine. We send `Vary: Cookie` so `C-C: public` should be fine here.

nickstenning:
Two small changes needed: 

1. to the Cache-Control headers for `/embed.js` so that it can be cached by the CDN and not just by user-agents.
2. to `/app.html` so that we don't put an external HTTP request on the critical path.

h/views/client.py
robertknight: Good catch. Done.

h/views/client.py
robertknight: Ah, yes indeed. Done.

PR #4356, opened 2017-02-07
Remove a test for a case that never occurs

fatbusinessman:
The `model` function should never be invoked for third-party requests, so this test is only meaningful for the `profile` function, where there is an equivalent test.


robertknight:
LGTM

PR #4357, opened 2017-02-07
Support unauthenticated `profile` requests

fatbusinessman:
This functionality is necessary so that we can display a list of public groups and use a default set of feature flags before a user has logged in. This needs to work both for the default authority and for a third party.

There was some discussion around whether to expect the `authority` parameter, even for authenticated requests, and fail if the value did not match the user's authority. The main argument behind this was to provide separation between the different URLs used for profiles from different authorities. However, in the interests of unblocking the client component of this work, and given a piece of suggested upcoming investigation into providing entirely separate API namespaces for third parties, we've gone with the simplest option for the moment.

This is the final piece of hypothesis/product-backlog#119.

This depends on #4356, as the removed test breaks with this new functionality.

### DIFF
--- a/h/session.py
+++ b/h/session.py
@@ -9,21 +9,35 @@ def model(request):
     session = {}
     session['csrf'] = request.session.get_csrf_token()
     session['userid'] = request.authenticated_userid
-    session['groups'] = _current_groups(request)
+    session['groups'] = _current_groups(request, request.auth_domain)
     session['features'] = request.feature.all()
     session['preferences'] = _user_preferences(request.authenticated_user)
     return session
 
 
-def profile(request):
+def profile(request, authority=None):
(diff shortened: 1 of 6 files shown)


robertknight:
Ah, I see. The failing test in question was removed in https://github.com/hypothesis/h/pull/4356

robertknight:
Rebased

robertknight:
Functional tests need updating - see Jenkins output.

chdorner:
Since the functional test was updated to have the tests back to green, and I need these changes to properly fix the OAuth integration tests to call the new `/api/debug-token` endpoint instead, I will merge this.

I looked at the code and it looks good to me, Robert has tested it with the client.

👍

PR #4358, opened 2017-02-08
Run collectd inside docker container for system metrics

chdorner:
In order to get system metrics from h running on Skyliner, we need to
run collectd inside the container.

This adds the exact same collectd configuration as we currently use on
the old infrastructure instances, except for the removal of the nptd
plugin which doesn't work due to [a
bug](https://github.com/collectd/collectd/issues/932).

The `start-collectd` script will configure collectd to send the metrics
to graphite if the `GRAPHITE_HOST` environment variable is set,
otherwise it will configure the basic `write_log` plugin.
Similar to that it tries to get the EC2 instance id through the metadata
API, if it isn't available it will fallback to the local hostname
(calling `hostname`).

### DIFF
--- a/bin/start-collectd
+++ b/bin/start-collectd
@@ -1,0 +1,36 @@
+#!/bin/sh
+
+: ${GRAPHITE_HOST:=}
+: ${GRAPHITE_PORT:="2003"}
+
+set -eu
+
+CONF_TEMPLATE=/etc/collectd/collectd.conf.tpl
+CONF_DEST=/etc/collectd/collectd.conf
+EXTRA_CONF_DIR=/etc/collectd/collectd.conf.d
+
+if [ -n "$GRAPHITE_HOST" ]; then
+    HOSTNAME=$(wget -O - -T 1 http://169.254.169.254/1.0/meta-data/instance-id 2>/dev/null || hostname)
+    echo "Hostname \"${HOSTNAME}\"" > "$EXTRA_CONF_DIR/hostname.conf"
+
+    cat >"$EXTRA_CONF_DIR/graphite.conf" <<EOM
+LoadPlugin write_graphite
+<Plugin write_graphite>
+  <Node "default">
+    Host "${GRAPHITE_HOST}"
+    Port "${GRAPHITE_PORT}"
+    Protocol "tcp"
+    LogSendErrors true
+    Prefix "collectd."
+    StoreRates true
+    AlwaysAppendDS false
+    EscapeCharacter "_"
+  </Node>
+</Plugin>
+EOM
+
+    exec collectd -f -C $CONF_DEST
+else
+  echo "INFO: GRAPHITE_HOST not provided so collectd will not be started"
+  exit 0
+fi
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,6 +4,7 @@ MAINTAINER Hypothes.is Project and contributors
 # Install system build and runtime dependencies.
 RUN apk-install \
     ca-certificates \
+    collectd \
     libffi \
     libpq \
     nginx \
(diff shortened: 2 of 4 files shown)


bin/start-collectd
```diff
@@ -0,0 +1,23 @@
...
+
+HOSTNAME=$(wget -O - --tries 1 --timeout 0.2 http://169.254.169.254/1.0/meta-data/instance-id 2>/dev/null || hostname)
+
+if [ -n "$GRAPHITE_HOST" ]; then
+    sed -e "s/{{ hostname }}/$HOSTNAME/g" \
+        -e "s/{{ graphite_host }}/$GRAPHITE_HOST/g" \
+        -e "s/{{ graphite_port }}/$GRAPHITE_PORT/g" "$CONF_TEMPLATE"
+else
+  sed -e "s/{{ hostname }}/$HOSTNAME/g" \
+      -e "s/LoadPlugin write_graphite/LoadPlugin write_log/g" \
```
chdorner: @nickstenning try this one out locally and let me know if you like it. The problem with collectd is that if there is no output configured it will send a bunch of error messages on every interval. This will just use the `write_log` output which will write all the metrics to stdout, which is a bit annoying as well.

bin/start-collectd
```diff
@@ -0,0 +1,23 @@
...
+
+: ${GRAPHITE_HOST:=}
+: ${GRAPHITE_PORT:="2003"}
+
+set -eu
+
+CONF_TEMPLATE=/etc/collectd/collectd.conf.tpl
+CONF_DEST=/etc/collectd/collectd.conf
+
+HOSTNAME=$(wget -O - --tries 1 --timeout 0.2 http://169.254.169.254/1.0/meta-data/instance-id 2>/dev/null || hostname)
```
nickstenning: I don't think this'll work (with `--tries` and `--timeout`) on Alpine, as I think `wget` is the busybox version of `wget` by default. Had a quick noodle and I think the best you'll manage is probably ` wget -O - -T 1 http://...`.

bin/start-collectd
nickstenning: Ok, that makes sense, but I think there might be a better option here, namely to just have this bootstrap script print a notice like `INFO: GRAPHITE_HOST not provided so collectd will not be started` and then just exit with a success returncode. In supervisor's default configuration it will only autorestart processes that exit with a nonzero status code, so this should work just fine.

After all, if we're not sending the metrics anywhere we might as well just not run collectd.

bin/start-collectd
```diff
@@ -0,0 +1,23 @@
...
+
+if [ -n "$GRAPHITE_HOST" ]; then
+    sed -e "s/{{ hostname }}/$HOSTNAME/g" \
```
nickstenning: I have a suggestion which might make this a bit nicer. Instead of doing this hacky thing like I did for nginx, perhaps we could use an

```
<Include "/etc/collectd/collectd.conf.d">
  Filter "*.conf"
</Include>
```

And then rather than using sed, do something like:

```shell
if [ -n "$GRAPHITE_HOST" ]; then
  cat >/etc/collectd/collectd.conf.d/graphite.conf <<EOM
<Plugin write_graphite>
	<Node "default">
		Host "${GRAPHITE_HOST}"
		Port "${GRAPHITE_PORT}"
		Protocol "tcp"
		LogSendErrors true
		Prefix "collectd."
		StoreRates true
		AlwaysAppendDS false
		EscapeCharacter "_"
	</Node>
</Plugin>
EOM
fi
```

Of course, you'll need to tweak the bits that you `chown` in the Dockerfile to be the `.d` directory rather than the root config file, but I think that'll work...

In retrospect I probably could have and should have done something like this for nginx.

bin/start-collectd
chdorner: Oh, you're right! And I never saw it failing because, well the metadata endpoint isn't available in my local testing. Also seems that we can't go under 1 second timeout, seems like the busybox version of `wget` doesn't support fractions.

bin/start-collectd
chdorner: Ah, I didn't know that supervisor doesn't try to restart a process if it exits with 0 or 2. But, I had to set the `startsecs` down to 0 as well, because:

> Even if a process exits with an “expected” exit code (see `exitcodes`), the start will still be considered a failure if the process exits quicker than `startsecs`.

This is much better 👍

bin/start-collectd
chdorner: This is much better, should have done this in the first place rather than copying the way you did the nginx config :)

bin/start-collectd
```diff
@@ -0,0 +1,36 @@
+#!/bin/sh
+
+: ${GRAPHITE_HOST:=}
+: ${GRAPHITE_PORT:="2003"}
+
+set -eu
+
+CONF_TEMPLATE=/etc/collectd/collectd.conf.tpl
```
nickstenning: This is unused now.

nickstenning:
LGTM. One line you can remove in the `start-collectd` script.

PR #4359, opened 2017-02-08
Simplify embed.js view

robertknight:
This is a follow-up to https://github.com/hypothesis/h/pull/4355 which removes an unnecessary separation between `h.client.embed_context` and `h.views.client.embed`. The separation dated back to when `h.client` was used to build the browser extension.


robertknight:
I'm closing this as I think it will be easier to do the refactoring here and to the _app.html_ views _before_ adding support for using the client boot script.

PR #4360, opened 2017-02-08
Add endpoint for debugging OAuth tokens

chdorner:
This endpoint will return some information about the token at hand. This
only works with OAuth 2 and developer tokens, not with our legacy client
JWT tokens. Since this will mostly be used by other developers, I don't
think we need to add support for our legacy client tokens to this
endpoint.

Example request with successful response for developer token:

```
GET /api/debug-token
Authorization: Bearer {developer token}

HTTP/1.1 200 OK
{
  "userid": "acct:chdorner@localhost",
  "issued_at": "2016-12-21T18:13:24.970395+00:00",
  "expired": false,
  "expires_at": null
}
```

Example request with successful response for OAuth 2 token:

```
GET /api/debug-token
Authorization: Bearer {developer token}

HTTP/1.1 200 OK
{
  "userid": "acct:fred@partner-1.org",
  "client": {
    "id": "5a2a8d1a-9074-11e6-b888-d705d9289341",
    "name": "Partner 1"
  },
  "issued_at": "2017-02-08T14:18:56.041786+00:00",
  "expires_at": "2017-02-08T15:18:49.868218+00:00"
  "expired": false,
}
```

There are two error cases.

1. The token is missing in the Authentication HTTP header:

```
GET /api/debug-token

HTTP/1.1 401 Unauthorized
{
  "error_description": "Bearer token is missing in Authorization HTTP header",
  "error": "missing_token"
}
```

2. The token either does not exist, or is expired:

```
GET /api/debug-token
Authorization: Bearer {expired token}

HTTP/1.1 401 Unauthorized
{
  "error_description": "Bearer token does not exist or is expired",
  "error": "missing_token"
}
```

### DIFF
--- a/h/services/auth_token.py
+++ b/h/services/auth_token.py
@@ -30,7 +30,10 @@ class AuthTokenService(object):
         """
 
         if token_str in self._validate_cache:
-            return self._validate_cache[token_str]
+            token = self._validate_cache[token_str]
+            if token is not None and token.is_valid():
+                return token
+            return None
 
         token = self._fetch_auth_token(token_str)
         self._validate_cache[token_str] = token
--- a/h/services/auth_token.py
+++ b/h/services/auth_token.py
@@ -32,16 +32,31 @@ class AuthTokenService(object):
         if token_str in self._validate_cache:
             return self._validate_cache[token_str]
 
-        token = self._fetch_token(token_str)
+        token = self._fetch_auth_token(token_str)
         self._validate_cache[token_str] = token
         if token is not None and token.is_valid():
             return token
         return None
 
-    def _fetch_token(self, token_str):
-        token_model = (self._session.query(models.Token)
-                       .filter_by(value=token_str)
-                       .one_or_none())
+    def fetch(self, token_str):
+        """
+        Fetch and return a token.
+
+        This returns a ``h.models.Token`` in comparison to what ``validate``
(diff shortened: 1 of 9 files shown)


h/services/auth_token.py
```diff
@@ -30,18 +30,24 @@ def validate(self, token_str):
...
         self._validate_cache[token_str] = token
         if token is not None and token.is_valid():
             return token
         return None
 
-    def _fetch_token(self, token_str):
-        token_model = (self._session.query(models.Token)
-                       .filter_by(value=token_str)
-                       .one_or_none())
+    def fetch(self, token_str):
```
fatbusinessman: Given this is a new public method we’re adding to the service, should it have a docstring too? Pretty much what’s in the commit message, from the look of it.

h/services/auth_token.py
```diff
@@ -30,18 +30,24 @@ def validate(self, token_str):
         """
 
         if token_str in self._validate_cache:
-            return self._validate_cache[token_str]
+            token = self._validate_cache[token_str]
+            if token is not None and token.is_valid():
```
fatbusinessman: Ah yes – well spotted. 👍

h/services/auth_token.py
chdorner: Done 👍

PR #4363, opened 2017-02-09
Simplify client views and add tests

robertknight:
The logic for the views that serve the client (`/embed.js` and `/app.html`) was split across two modules `h.client` and `h.views.client` and not well tested.

The `h.client` module was originally created for rendering the app.html and
embed.js files during the browser extension build. In that environment
there is no request object, hence the separation.

Since the browser extension build has now been moved to a separate repo,
the client views can be simplified by moving all client view logic to
`h.views.client` and making the views return context dicts rather than
rendering templates directly. Add tests which take advantage of this and simplify some existing ones.

Additionally the `widget` route has been renamed to `sidebar_app` to be
more descriptive.


tests/h/views/main_test.py
```diff
@@ -12,36 +12,34 @@
...
     document_title.return_value = 'WikiHow — How to Make a ☆Starmap☆'
 
-    render_app_html.return_value = '<html></html>'
-    main.annotation_page(context, pyramid_request)
-    args, kwargs = render_app_html.call_args
-    test = lambda d: 'foo' in d['content'] and 'Starmap' in d['content']
-    assert any(test(d) for d in kwargs['extra']['meta_attrs'])
+    ctx = main.annotation_page(context, pyramid_request)
+
+    def test(d):
```
robertknight: I changed `test` here because flake8 recommends defining a nested func rather than doing `var_name = lambda ...`. IMO the result is cleaner so I did.

h/views/client.py
```diff
@@ -50,23 +116,27 @@ def annotator_token(request):
...
-        app_html_url=request.route_url('widget'),
-        base_url=request.route_url('index'))
-    return request.response
+@view_config(route_name='embed',
+             renderer='h:templates/embed.js.jinja2')
+def embed(request):
+    request.response.content_type = 'text/javascript'
+
+    assets_env = request.registry['assets_client_env']
+    base_url = request.route_url('index')
```
seanh: Both names come from the preexisting code, but this is called `service_url` above, I wonder if it'd be worth renaming it to `service_url` here as well? Or even rename both of them to `index_url` to match the route name?

h/views/client.py
```diff
@@ -8,30 +8,96 @@
...
-        ga_client_tracking_id=request.registry.settings.get('ga_client_tracking_id'),
-        extra=extra)
-    request.response.text = html
-    return request.response
+    assets_env = request.registry['assets_client_env']
+    api_url = request.route_url('api.index')
+    service_url = request.route_url('index')
+    sentry_public_dsn = client_sentry_dsn
+    websocket_url = request.registry.settings.get('h.websocket_url')
+    ga_client_tracking_id = request.registry.settings.get('ga_client_tracking_id')
```
seanh: There's some duplication of this code for getting variables out of the `request` object (the `embed()` view below pulls out `assets_env` and `service_url` in the same way that this code does). I wonder if this block could be moved to a helper function that takes the `request` and returns a dict of all these?

tests/h/views/client_test.py
```diff
@@ -19,6 +22,80 @@ def test_annotator_token_returns_token(generate_jwt, pyramid_request):
...
+            'http://example.com/assets/client/inject.css',
+            ]
+
+    def test_it_sets_content_type(self, pyramid_request_):
+        client.embed(pyramid_request_)
+        assert pyramid_request_.response.content_type == 'text/javascript'
+
+
+@pytest.fixture
+def pyramid_request_(pyramid_request):
```
robertknight: Good question - we had a discussion on Slack and Nick suggested a cleaner solution which I am now implementing. See https://hypothes-is.slack.com/archives/public/p1486725096004540

h/routes.py
```diff
@@ -69,7 +69,7 @@ def includeme(config):
 
     # Client
     config.add_route('session', '/app')
-    config.add_route('widget', '/app.html')
+    config.add_route('sidebar_app', '/app.html')
```
seanh: Slightly better name here :)

seanh:
@robertknight I'll let you make any more changes that you want and then feel free to merge if yourself

h/views/client.py
robertknight: `base_url` and `service_url` happen to have the same URL but they have different meanings. The first is the _root URL to which asset URLs are relative_, the second is the URL which the client should use to construct URLs that link into pages in the service.

h/views/client.py
robertknight: These variables are not really necessary, I've simplified this code a bit to move the values into the place where the kwargs are passed to `_app_html_context`

tests/h/views/main_test.py
robertknight: The purpose of this test is to make sure that the right `<link>` and `<meta>` tags are rendered when you visit `/a/<annotation ID>` pages, depending on the data available in the annotation. Probably easier to clarify by changing the test name as much as this lambda. I'm going to leave it for the moment to get the rest of the work here merged.

PR #4364, opened 2017-02-09
Enable collectd nginx plugin and nginx's status page

chdorner:
The nginx status page is only going to be accessible locally inside the
docker container, collectd is then sending the data off with the other
metrics to graphite.


nickstenning:
LGTM.

PR #4365, opened 2017-02-09
Add initial documentation on writing tests

fatbusinessman:
This initially sprang out of some confusion on my part, where I assumed – based on the [tests for the session module](https://github.com/hypothesis/h/blob/4db28c5da0ca91b765c11d2d6bbc6c7a4231b52f/tests/h/session_test.py) that we wanted to use mocks heavily in our codebase.

This is by no means complete, and may well not be an accurate representation of how we collectively want to work. Comments, musings and enthusiastic agreement or dissent are all welcome.

### DIFF
--- a/docs/developing/testing.rst
+++ b/docs/developing/testing.rst
@@ -11,3 +11,23 @@ our tests, which covers some of the tools we use (``tox`` and ``pytest``) and
 some of the testing techniques they provide (factories and parametrization).
 
 .. _guide to getting started: https://www.seanh.cc/posts/running-the-h-tests
+
+Using mock objects
+------------------
+
+The ``mock`` library lets us construct fake versions of our objects to help with
+testing. While this can make it easier to write fast, isolated tests, it also
+makes it easier to write tests that don't reflect reality.
+
+Where practical, we prefer to write tests using (in order of preference):
+
+1. Real objects
+2. Fake objects
+3. Mock objects constructed with ``spec_set`` or ``autospec`` – these objects
+   will only respond to the same methods as the real objects they replace,
+   which reduces the risk of mocks that do not behave like their real
+   counterparts
+4. Generic mock objects
+
+This is a deliberate decision that, when we have to choose between test
+isolation and test fidelity, we usually want to choose test fidelity.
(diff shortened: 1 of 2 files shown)


docs/developing/testing.rst
```diff
@@ -0,0 +1,41 @@
...
+This section covers writing tests for the ``h`` codebase.
+
+Getting started
+---------------
+
+Sean Hammond has written up a `guide to getting started`_ running and writing
+our tests, which covers some of the tools we use (``tox`` and ``pytest``) and
+some of the testing techniques they provide (factories and parametrization).
+
+.. _guide to getting started: https://www.seanh.cc/posts/running-the-h-tests
```
seanh: Feel free to steal from this if there's anything useful in there, but I wouldn't bother linking to it necessarily

nickstenning:
I think this is fantastic! I'm really glad you wrote this up, and so clearly, too.

The only part of the whole thing I'd consider tweaking is the final paragraph. You say

```
This is a deliberate decision that, when we have to choose between test
isolation and test fidelity, we usually want to choose test fidelity.
```

which is certainly true, but I think is so plainly stated that it could be misleading. For example, it implies that we'd choose test fidelity even if using real objects made the tests ridiculously slow, which is sometimes a big part of the decision to use stubs or fakes.

What would you say to something like this:

```rst
In an ideal world, we would always be able to use real objects instead of stubs or mocks, but sometimes this can result in:

- complicated test setup code
- slow tests
- coupling of test assertions to non-interface implementation details

For new code, it's usually a good idea to design the code so that testing with "real" rather than stubs or mocking is easy. It can help to make extensive use of `value objects`_ in tested interfaces (using ``collections.namedtuple`` from the standard library) and apply the `functional core, imperative shell`_ pattern.

For older code which doesn't make testing so easy, or for code that is part of the "imperative shell" (see link in previous paragraph) it can sometimes be hard to test what you need without resorting to stubs or mock objects, and that's fine.

.. _value objects: https://martinfowler.com/bliki/ValueObject.html
.. _functional core, imperative shell: https://www.destroyallsoftware.com/talks/boundaries
```

docs/developing/testing.rst
fatbusinessman: I did consider that, but pretty much all of that guide would have been useful to me starting out, so there’s a good chance that most of it will be useful to any given new starter. Given that, I figured I’d link rather than copy it wholesale.

fatbusinessman:
> What would you say to something like this:

I would say yes; added.

(I’m especially happy to see you mention the “functional core, imperative shell” pattern.)

nickstenning:
Yeesh! Sorry this stayed open so long.

PR #4366, opened 2017-02-10
Fix statsd prefix ec2 instance

chdorner:
This removes the rather crude handling of the `STATSD_PREFIX` where we add the hostname of the instance at the end of the configured (or empty) prefix (done [here](https://github.com/hypothesis/h/blob/0c716a0ffdf58ff74a98cc4bdcf82b2975b92eab/gunicorn.conf.py#L26-L27) and [here](https://github.com/hypothesis/h/blob/0c716a0ffdf58ff74a98cc4bdcf82b2975b92eab/h/config.py#L115-L117)).

To achieve this we add a new shell script `init-env` which wraps supervisor. It first tries to get the instance id from the EC2 metadata API and sets the `INSTANCE_ID` environment variable to this value (or empty when we're not running in a non-EC2 environment). Then it appends either the instance id, or `hostname` to at the end of the already configured (or empty) `STATSD_PREFIX` environment variable.

This way we can get rid of any special `STATSD_PREFIX` handling from the Python code and only deal with it in the helper scripts.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -64,4 +64,4 @@ ENV PYTHONPATH /var/lib/hypothesis:$PYTHONPATH
 
 # Start the web server by default
 USER hypothesis
-CMD ["supervisord", "-c" , "conf/supervisord.conf"]
+CMD ["init-env", "supervisord", "-c" , "conf/supervisord.conf"]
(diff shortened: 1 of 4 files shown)


nickstenning:
:+1:

PR #4368, opened 2017-02-13
Limit celery ping replies to one answer

chdorner:
Otherwise it tries to read messages until the given timeout is reached.
This should get our response times for the status endpoint down a bit.

This might return wrong data while we're running both the old infrastructure and Skyliner at the same time, but as soon as we turn off the old one this should always be correct.


nickstenning:
Nice!

PR #4369, opened 2017-02-13
Limit OAuth grant token lifetime

fatbusinessman:
This is part of hypothesis/product-backlog#138. It’s very much a work in progress so far (the magic constants in particular need extracting), and I’d like some more tests around missing properties, but if someone could at least confirm that this is the right fix in the right place, I’d appreciate it.

### DIFF
--- a/tests/h/services/oauth_test.py
+++ b/tests/h/services/oauth_test.py
@@ -224,6 +224,26 @@ class TestOAuthServiceVerifyJWTBearerRequest(object):
 
         assert exc.value.type == 'invalid_grant'
 
+    def test_missing_expiry(self, svc, claims, authclient, jwt_bearer_body):
+        del claims['exp']
+        jwt_bearer_body['assertion'] = self.jwt_token(claims, authclient.secret)
+
+        with pytest.raises(OAuthTokenError) as exc:
+            svc.verify_token_request(jwt_bearer_body)
+
+        assert exc.value.type == 'invalid_grant'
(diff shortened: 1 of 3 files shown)


tests/h/services/oauth_test.py
```diff
@@ -209,6 +209,66 @@ def test_it_raises_when_client_authority_does_not_match_userid(self, svc, db_ses
...
+
+        assert exc.value.type == 'invalid_grant'
+        assert 'JWT is missing claim {}'.format(claim_name) in exc.value
+
+    @pytest.mark.parametrize('claim_name,delta',
+                             [['iat', timedelta(minutes=-5)],
+                              ['nbf', timedelta(minutes=-5)],
+                              ['exp', timedelta(minutes=5)]])
+    def test_string_timestamp(self, svc, claims, authclient, jwt_bearer_body, claim_name, delta):
+        claims[claim_name] = unicode(self.epoch(delta=delta))
```
chdorner: I'm not sure if you've come across it yet, since we try to write code that works in Python 2 and 3 you can use `h._compat.text_type`

fatbusinessman: Ah, yes – I’d seen that dotted around the codebase before, but I hadn’t made the connection that was what it was for. Thanks for the tip. 👍

fatbusinessman:
I’ve taken a different tack on this, which seems to work much more cleanly. While I’m taking a bunch of the code from this pull request, I think the new approach is different enough to warrant a new PR, so I’m closing this one.

PR #4370, opened 2017-02-13
Don't collect disk usage metrics for weird mountpoints

nickstenning:
These single-file mountpoints are an artifact of the Docker engine, and so aren't worth reporting to graphite.


chdorner:
Looks good, just testing it out locally.

PR #4375, opened 2017-02-16
Add PATCH support for annotation update API endpoint

chdorner:
The current update annotations endpoint mixes the `PUT` request method with `PATCH` characteristics. This PR changes this, but we can't just remove the `PUT` request method as there might still be API clients out there that are using `PUT`, so we support both for now.

This also adds a statsd counter that gets incremented every time we get a request with the `PUT` request method, this will allow us to determine when we can remove `PUT` support.

Backlog ticket: hypothesis/product-backlog#147

### DIFF
--- a/docs/_extra/api/hypothesis.yaml
+++ b/docs/_extra/api/hypothesis.yaml
@@ -131,8 +131,18 @@ paths:
           description: Annotation not found or no permission to view
           schema:
             $ref: '#/definitions/Error'
-    put:
+    patch:
       summary: Update an annotation
+      description: |
+        This endpoint is available under both the `PATCH` and `PUT`
+        request methods. Both endpoints have PATCH-characteristics
+        as defined in [RFC5789](https://tools.ietf.org/html/rfc5789#section-1),
+        meaning the request body does not have to include the whole annotation
+        object.
+
+        New implementations should use the `PATCH` request method, and existing
+        implementations continue to work under `PUT` but should switch to `PATCH`.
+
       operationId: updateAnnotation
       parameters:
         - name: id
--- a/src/memex/views.py
+++ b/src/memex/views.py
@@ -254,6 +254,9 @@ def read_jsonld(context, request):
             description='Update an annotation')
 def update(context, request):
     """Update the specified annotation with data from the PATCH payload."""
+    if request.method == 'PUT' and hasattr(request, 'stats'):
+        request.stats.incr('memex.api.deprecated.put_update_annotation')
+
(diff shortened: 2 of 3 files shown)


src/memex/views.py
```diff
@@ -248,12 +248,16 @@ def read_jsonld(context, request):
...
-            request_method='PUT',
+            request_method=('PATCH', 'PUT'),
             permission='update',
             link_name='annotation.update',
             description='Update an annotation')
 def update(context, request):
-    """Update the specified annotation with data from the PUT payload."""
+    """Update the specified annotation with data from the PATCH payload."""
+    if request.method == 'PUT' and hasattr(request, 'stats'):
+        stats = getattr(request, 'stats')
```
fatbusinessman: Out of interest, what’s the reasoning behind using `getattr(request, 'stats')` here instead of `request.stats`? Is it for symmetry with the `hasattr` on the line above, or is there a difference between the two that I’ve missed?

docs/_extra/api/hypothesis.yaml
```diff
@@ -131,8 +131,18 @@ paths:
           description: Annotation not found or no permission to view
           schema:
             $ref: '#/definitions/Error'
-    put:
+    patch:
       summary: Update an annotation
+      description: |
+        This endpoint is available under both the `PATCH` and `PUT`
+        request methods. Both endpoints have PATCH-characteristics
```
fatbusinessman: Should the “PATCH” on this line have backticks around it too?

chdorner: I don't know, it might look at bit weird: `PATCH`-characteristics

src/memex/views.py
chdorner: You are right, I didn't change it when I added the `hasattr`. Fixed.

docs/_extra/api/hypothesis.yaml
fatbusinessman: That reads more clearly to me, but that’s totally subjective personal taste. I don’t have strong feelings about this, so go with whichever version you prefer.

nickstenning: <GRAMMAR-TROLL>It may solve both your concerns to know that a hyphen is probably unnecessary here.</GRAMMAR-TROLL>

nickstenning: OMG GITHUB STRIPPED MY `<GRAMMAR-TROLL>` TAGS!

robertknight:
I've tested this with the client, works as expected.

PR #4376, opened 2017-02-16
Add __ne__ method to all matchers

nickstenning:
Matcher objects should behave correctly with the '!=' operator, too. Add a common superclass for all matchers which defines __ne__ to return the complement of __eq__.

### DIFF
--- a/tests/common/matchers.py
+++ b/tests/common/matchers.py
@@ -32,7 +32,16 @@ import re
 from pyramid import httpexceptions
 
 
-class any_callable(object):  # noqa: N801
+class Matcher(object):
+
+    def __eq__(self, other):
+        raise NotImplementedError('subclasses should provide an __eq__ method')
+
+    def __ne__(self, other):
+        return not self.__eq__(other)
+
+
+class any_callable(Matcher):  # noqa: N801
     """An object __eq__ to any callable object."""
 
     def __eq__(self, other):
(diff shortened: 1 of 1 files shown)


fatbusinessman:
I have a question. Is this:

* a current problem that means existing tests are passing when they shouldn’t, or aren’t passing when they should?
* something you’re anticipating we’ll need in an upcoming piece of work? For instance, being able to say `assert_called_once_with(please_anything_but(instance_of(int))`?
* a precautionary measure?
* purely for the sake of correctness/completeness?

nickstenning:
Sorry, your question is a valid one and I should have provided that context in the PR.

I encountered the missing `__ne__` method while working on a PR I'll submit shortly (to add a "send test email" view to the admin panel) and figured it would be easier to review split out.

tests/common/matchers.py
```diff
@@ -32,15 +32,24 @@
...
 
 
-class any_callable(object):  # noqa: N801
+class Matcher(object):
+
+    def __eq__(self, other):
+        raise NotImplementedError('subclasses should provide an __eq__ method')
+
+    def __ne__(self, other):
+        return not self.__eq__(other)
```
fatbusinessman: Is the use of `self.__eq__` rather than `==` a matter of style (to explicitly link it to the method above) or does it serve an additional purpose, like stopping Python trying to invert the arguments?

nickstenning:
> shouldn’t that be a common superclass?

Gah! Thank you for spotting that.

tests/common/matchers.py
nickstenning: It's just to make it explicit that it's returning the complement of a function you've defined. I suppose it would probably also reduce the risk of introducing a bug if you somehow managed to define an intransitive equality operator.

nickstenning: (But I wasn't really thinking about it that hard...)

PR #4377, opened 2017-02-16
Add an admin page which can send test mails

nickstenning:
Adds a section to the admin panel which can be used to test email sending. Any staff user can send test emails quickly and easily without having to sign up test accounts or awkwardly log into two accounts and send replies.

This is prompted by a mail outage yesterday, and part of a broader wish to add better tools to h for [testing in production](https://www.theguardian.com/info/developer-blog/2016/dec/20/testing-in-production-rethinking-the-conventional-deployment-pipeline).

### DIFF
--- a/tests/h/admin/views/mailer_test.py
+++ b/tests/h/admin/views/mailer_test.py
@@ -1,0 +1,78 @@
+# -*- coding: utf-8 -*-
+
+from __future__ import unicode_literals
+
+import pytest
+
+from pyramid.httpexceptions import HTTPSeeOther
+
+from h.admin.views import mailer as views
+
+
+class TestMailerIndex(object):
+
+    def test_when_no_taskid(self, pyramid_request):
+        result = views.mailer_index(pyramid_request)
+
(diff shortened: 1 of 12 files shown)


tests/h/admin/views/mailer_test.py
```diff
@@ -0,0 +1,78 @@
...
+@pytest.fixture
+def testmail(patch):
+    test = patch('h.admin.views.mailer.test')
+    test.generate.side_effect = lambda _, r: ([r], 'TEST', 'text', 'html')
+    return test
+
+
+@pytest.fixture
+def routes(pyramid_config):
+    pyramid_config.add_route('admin_mailer', '/adm/mailer')
```
robertknight: Is 'adm' a typo here or intentional?

nickstenning: It's deliberate -- the view should be using the defined routes to get its URLs, not hardcoding them. Keeping the routes different in tests from in the "real" application is one way of checking that.

robertknight:
LGTM

PR #4378, opened 2017-02-16
Report celery task failure to the log in dev

### LINKED ISSUE #4291: Tasks fail silently when running dev server
### Steps to reproduce
1. Have a task that fails. I encountered this as part of #4290, but you could force a failure with this patch:
    ```diff
    diff --git a/h/tasks/indexer.py b/h/tasks/indexer.py
    index 1dd2e3f..dd8aed1 100644
    --- a/h/tasks/indexer.py
    +++ b/h/tasks/indexer.py
    @@ -10,6 +10,7 @@ from memex.search.index import delete

     @celery.task
     def add_annotation(id_):
    +    raise RuntimeError('Ohnoes!')
         annotation = storage.fetch_annotation(celery.request.db, id_)
         if annotation:
             index(celery.request.es, annotation, celery.request)
    ```

2. Start up the app with `make devserver`
3. Trigger the task (in my example, by creating an annotation)

### Expected behaviour
Something should crop up in the `worker` logs for the app, mentioning that the task threw an exception.

### Actual behaviour
The task fails silently. Even if running with `DEBUG`-level logs on the worker, there is no description of the exception thrown.

### Additional details

I’m not 100% sure whether this qualifies as a bug or a feature request, but the lack of reporting on failed tasks here makes other bugs (e.g. #4290) rather harder to track down.

nickstenning:
As reported in #4291, celery task failure isn't reported in development. Add a `task_failure` signal handler which logs exceptions thrown in tasks.

Fixes #4291.

### DIFF
--- a/h/celery.py
+++ b/h/celery.py
@@ -122,6 +122,19 @@ def transaction_abort(sender, **kwargs):
     sender.app.request.tm.abort()
 
 
+@signals.task_failure.connect
+def report_failure(sender, task_id, args, kwargs, einfo, **kw):
+    """Report a task failure to the console in development."""
+    if not sender.app.request.debug:
+        return
+    log.error('task failure: %s (%s) called with args=%s, kwargs=%s',
+              sender.name,
+              task_id,
+              args,
+              kwargs,
+              exc_info=einfo.exc_info)
+
+
 def start(argv, bootstrap):
     """Run the Celery CLI."""
     # We attach the bootstrap function directly to the Celery application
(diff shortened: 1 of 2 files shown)


seanh:
Works here, code is clear and simple and nicely formatted. I would have created a `log` fixture that `patch`es `h.celery.log`, instead of patching it in each test. Out of interest, any reason why it wasn't done that way?

nickstenning:
> Out of interest, any reason why it wasn't done that way?

None in particular! Two duplicated lines didn't seem like enough friction to break it out into a fixture, that's all.

PR #4379, opened 2017-02-17
Refactor logic for OAuth grant tokens

fatbusinessman:
This is some preparatory work to make it easier to check token timestamps as part of hypothesis/product-backlog#138, and generally to make the logic in `_verify_jwt_bearer` a little cleaner and easier to understand.

### DIFF
--- a/h/services/oauth.py
+++ b/h/services/oauth.py
@@ -71,14 +71,7 @@ class OAuthService(object):
         """
         token = GrantToken(body.get('assertion'))
 
-        authclient = None
-        try:
-            authclient = self.session.query(models.AuthClient).get(token.issuer)
-        except sa.exc.StatementError as exc:
-            if str(exc.orig) == 'badly formed hexadecimal UUID string':
-                pass
-            else:
-                raise
+        authclient = self._get_authclient_by_id(token.issuer)
         if not authclient:
             raise OAuthTokenError('given JWT issuer is invalid', 'invalid_grant')
 
@@ -94,6 +87,15 @@ class OAuthService(object):
                                   'invalid_grant')
 
         return (user, authclient)
+
+    def _get_authclient_by_id(self, client_id):
+        try:
+            return self.session.query(models.AuthClient).get(client_id)
+        except sa.exc.StatementError as exc:
+            if str(exc.orig) == 'badly formed hexadecimal UUID string':
+                return None
+            else:
+                raise
 
     def _verify_refresh_token(self, body):
         refresh_token = body.get('refresh_token')
--- a/h/services/oauth.py
+++ b/h/services/oauth.py
@@ -69,22 +69,11 @@ class OAuthService(object):
         :returns: a tuple with the user and authclient
         :rtype: tuple
         """
-        assertion = body.get('assertion')
-
-        if not assertion or type(assertion) != text_type:
-            raise OAuthTokenError('required assertion parameter is missing',
-                                  'invalid_request')
-        token = assertion
-
-        unverified_claims = self._decode(token, verify=False)
-
-        client_id = unverified_claims.get('iss', None)
-        if not client_id:
-            raise OAuthTokenError('grant token issuer is missing', 'invalid_grant')
+        token = GrantToken(body.get('assertion'))
 
         authclient = None
         try:
-            authclient = self.session.query(models.AuthClient).get(client_id)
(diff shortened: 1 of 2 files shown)


fatbusinessman:
There’s one line in the diff that isn’t covered by the tests (and presumably wasn’t before). I’ve put together a test that will cover it, but I think I’m about to put in stricter controls on the values we accept – effectively adding type-checking to all the claims – which will mean that test no longer applies.

The test:

```diff
diff --git a/tests/h/services/oauth_test.py b/tests/h/services/oauth_test.py
index 542ca1c..e18863b 100644
--- a/tests/h/services/oauth_test.py
+++ b/tests/h/services/oauth_test.py
@@ -8,6 +8,7 @@ from calendar import timegm
 import jwt
 import mock
 import pytest
+import sqlalchemy as sa
 
 from h import models
 from h.exceptions import OAuthTokenError
@@ -82,6 +83,13 @@ class TestOAuthServiceVerifyJWTBearerRequest(object):
         assert exc.value.type == 'invalid_grant'
         assert 'issuer is missing' in exc.value.message
 
+    def test_invalid_jwt_issuer(self, svc, claims, authclient, jwt_bearer_body):
+        claims['iss'] = 72
+        jwt_bearer_body['assertion'] = self.jwt_token(claims, authclient.secret)
+
+        with pytest.raises(sa.exc.StatementError):
+            svc.verify_token_request(jwt_bearer_body)
+
     def test_missing_authclient_with_given_jwt_issuer(self, svc, authclient, db_session, jwt_bearer_body):
         db_session.delete(authclient)
         db_session.flush()
```

h/services/oauth.py
```diff
@@ -161,10 +144,51 @@ def create_token(self, user, authclient):
...
+    the token is a correctly-formatted JSON Web Token, and that it contains all
+    the required claims in the right formats. Some of this processing is
+    deferred to the `jwt` module, but that doesn't handle all the fields we
+    want to validate.
+
+    """
+
+    def __init__(self, token):
+        if not token or type(token) != text_type:
+            raise OAuthTokenError('required assertion parameter is missing',
```
chdorner: I found it a bit weird to have this error message here, because the implementation of this class doesn't know about an assertion parameter, it just gets the token string. But it's probably good enough for now as long as we don't use the `GrantToken` class anywhere else than here.

h/services/oauth.py
```diff
@@ -161,10 +144,51 @@ def create_token(self, user, authclient):
...
+
+    @property
+    def issuer(self):
+        iss = self._claims.get('iss', None)
+        if not iss:
+            raise OAuthTokenError('grant token issuer is missing', 'invalid_grant')
+        return iss
+
+    @property
+    def subject(self):
```
chdorner: One thing I'm missing is that the old code was very clear that it first decodes the claims without verifying for getting the issuer, and then decoded the claims again with verification for reading the other claims (by using `unverified_claims` and `claims` variable names).
Since this isn't a general purpose grant token class and is only being used here, we might want to track the verification state and raise an error if we try to access the subject (or any other non-issuer) claim without first verifying the grant token?

What do you think?

fatbusinessman: That’s an excellent point. I had an idea about a potentially neat way to address it (bbf4360); let me know what you think.

h/services/oauth.py
```diff
@@ -69,7 +69,11 @@ def _verify_jwt_bearer(self, body):
         :returns: a tuple with the user and authclient
         :rtype: tuple
         """
-        token = GrantToken(body.get('assertion'))
+        if 'assertion' not in body:
+            raise OAuthTokenError('required assertion parameter is missing',
+                                  'invalid_request')
+
+        token = GrantToken(body['assertion'])
```
chdorner: 👍 I like this solution

h/services/oauth.py
```diff
@@ -177,14 +178,24 @@ def issuer(self):
...
-        sub = self._claims.get('sub', None)
-        if not sub:
-            raise OAuthTokenError('JWT subject is missing', 'invalid_grant')
-        return sub
+    def verified(self, key, audience):
+        return VerifiedGrantToken(self._token, key, audience)
 
-    def verify(self, key, audience):
+
+class VerifiedGrantToken(GrantToken):
```
chdorner: I like this! Taking your OO approach to the next level 👍

PR #4381, opened 2017-02-20
Limit OAuth grant token lifetime (take 2)

fatbusinessman:
This is another implementation of #4369, using the new grant token classes. It’s part of hypothesis/product-backlog#138.

To mitigate the security risks if a grant token is compromised, we're now making sure that these tokens will only be honoured if they are short-lived – namely, that they have a start time (`nbf`) and an expiry time (`exp`) and that the difference between them is no more than 10 minutes.

As part of this, the format validation on these timestamps is now more restrictive. The `pyjwt` library is pretty permissive, allowing `None` values and timestamp values as strings (e.g. `"1487339382"`).  Given that RFC 7519 [insists that these values MUST be numbers][1], we can reasonably reject values that are not number types.

[1]: https://tools.ietf.org/html/rfc7519#section-4.1.4

### DIFF
--- a/tests/functional/test_oauth.py
+++ b/tests/functional/test_oauth.py
@@ -3,6 +3,7 @@
 
 from __future__ import unicode_literals
 
+import calendar
 import datetime
 
 import jwt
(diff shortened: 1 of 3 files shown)


tests/h/services/oauth_test.py
```diff
@@ -210,6 +210,65 @@ def test_it_raises_when_client_authority_does_not_match_userid(self, svc, db_ses
...
+            svc.verify_token_request(jwt_bearer_body)
+
+        assert exc.value.type == 'invalid_grant'
+        assert 'JWT is missing claim {}'.format(claim_name) in exc.value
+
+    @pytest.mark.parametrize('claim_name,delta',
+                             [['nbf', timedelta(minutes=-5)],
+                              ['exp', timedelta(minutes=5)]])
+    def test_string_timestamp(self, svc, claims, authclient, jwt_bearer_body, claim_name, delta):
+        claims[claim_name] = unicode(self.epoch(delta=delta))
```
fatbusinessman: Gah! Somehow I managed to revert this. One sec…

fatbusinessman: …there we go. Nothing to see here.

tests/functional/test_oauth.py
```diff
@@ -122,17 +123,31 @@ def test_you_cannot_use_a_refresh_token_to_authenticate_api_requests(
...
         return response.json_body
 
+    def epoch(self, delta=None):
+        """Get a Unix timestamp for the current time, with optional offset."""
+        timestamp = datetime.datetime.utcnow()
+
+        if delta is not None:
+            timestamp = timestamp + delta
+
+        return calendar.timegm(timestamp.utctimetuple())
```
nickstenning: Oh Python...

nickstenning:
Works nicely :)

nickstenning:
One thing I'd like to flag up as followup: could we standardise the error messages returned here. Sometimes we say "grant token", sometimes we say "JWT", and sometimes we say "JWT token" (ew).

Could we unify these on "grant token", say?

fatbusinessman:
> Could we unify these on "grant token", say?

Yeah, that sounds good to me. How safely can we change these without worrying about current clients relying on the existing error messages?

PR #4384, opened 2017-02-21
Update class name for config script tag in sidebar app

robertknight:
**This needs to be merged before https://github.com/hypothesis/client/pull/243 is deployed to the service**

https://github.com/hypothesis/client/pull/243 changes the client to use
the same class name, `js-hypothesis-config`, for config script tags in
both the sidebar app and host page.

This updates the service code to use the expected name.

### DIFF
--- a/h/templates/app.html.jinja2
+++ b/h/templates/app.html.jinja2
@@ -25,8 +25,11 @@
   <body>
     <hypothesis-app></hypothesis-app>
 
+    {# The `js-hypothesis-settings` class should be removed once
+       https://github.com/hypothesis/client/pull/243 is released.
+     #}
     <!-- App Configuration !-->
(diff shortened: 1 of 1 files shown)


nickstenning:
As mentioned over on https://github.com/hypothesis/client/pull/243, it might be better to simply *add* the preferred classname here with a `FIXME` to remove the old one once the client changes have been released. That way neither change is blocked on the other.

robertknight:
> As mentioned over on hypothesis/client#243, it might be better to simply add the preferred classname 
>  That way neither change is blocked on the other.

I'm fine with doing that, but I'll note that there will still be a dependency. If the client reads _only_ from tags with the preferred class name, then we need to deploy the (trivial) change to the service first that adds this class name before releasing the updated client.

PR #4389, opened 2017-02-24
Add documentation for the `/api/profile` endpoint

fatbusinessman:
While this endpoint probably still counts as slightly experimental, I think it's probably better documented than not.

This is some follow-up work for hypothesis/product-backlog#119.

### DIFF
--- a/docs/_extra/api/hypothesis.yaml
+++ b/docs/_extra/api/hypothesis.yaml
@@ -197,6 +197,27 @@ paths:
           description: Annotation not found or no permission to delete
           schema:
             $ref: '#/definitions/Error'
+  /profile:
+    get:
+      summary: Get the user's profile
+      description: >
+        This can either be accessed with an API key, in which case it will show
+        the authenticated user's profile, or without, in which case it will show
+        information for a logged-out profile, including only public groups for
+        the current authority.
+      parameters:
+        - name: authority
+          in: query
+          description: >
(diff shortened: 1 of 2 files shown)


fatbusinessman:
Some questions for the person who reviews this:

* Did I get Swagger right? It appears to work, but I haven’t really poked at Swagger before, so I might well have missed something
* Do we need more documentation about the way this particular bit of the API behaves, given the number of different ways it can work?
* Would it be worth the extra complexity to have tests checking whether the actual API responses validate against the schema, or would that be overcomplicating matters?

nickstenning:
> Did I get Swagger right? It appears to work, but I haven’t really poked at Swagger before, so I might well have missed something

LGTM.

> Do we need more documentation about the way this particular bit of the API behaves, given the number of different ways it can work?

This is a good start. I don't think adding more documentation here is the most important place we can be spending our time.

> Would it be worth the extra complexity to have tests checking whether the actual API responses validate against the schema, or would that be overcomplicating matters?

In general terms, I think there are two things we probably want to do at some point:

- load and use the JSON schemas we use for the API documentation in the application itself and use them where input validation is required, rather than [having them duplicated](https://github.com/hypothesis/h/blob/0261bc4306940bd52744c72e16d8894edd282190/src/memex/schemas.py#L52-L156).
- make example requests against API endpoints that validate the response format against the documented response format, perhaps using these JSON schema files.

PR #4392, opened 2017-02-27
Change default client URL to cdn.hypothes.is

robertknight:
Change the default client URL to https://cdn.hypothes.is/hypothesis and remove
the logic for resolving the redirect which is not required since
`https://cdn.hypothes.is/hypothesis` returns the latest version of the
boot script rather than a redirect.


nickstenning:
LGTM.

PR #4394, opened 2017-02-28
Add environment name to Sentry client

nickstenning:
Skyliner provides the ENV environment variable for our application, which contains either "prod" or "qa" depending on the environment.

Sentry supports partitioning reports based on the environment within a single application. Adding this metadata to the `Client` allows us to migrate QA and Prod onto a single application in Sentry.


nickstenning:
> For my own curiosity, is the extra bit of work to move everything to a common DSN somewhere in one of our repos, or within Skyliner itself?

Good question. The DSN is configured automatically from the `SENTRY_DSN` environment variable, so this will just be a config change in Skyliner.

PR #4395, opened 2017-03-01
Match dot- and case-variants when looking up by username

nickstenning:
We've had a number of support tickets from users who have created their accounts with usernames such as "JoeBloggs", and are later confused because they can't log in as "joebloggs".

This gets particularly confusing when, having had the login form tell them "User does not exist," they go and try and sign up again, at which point the signup form tells them that the username is taken.

This commit addresses this at the ORM level, by ensuring that all lookups against the "username" field are actually performed against the normalised "uid" field. That is, a query such as

```python
request.db.query(models.User).filter_by(username='Joe.Bloggs')
```

which would previously have resulted in a WHERE clause of

```sql
WHERE username = 'Joe.Bloggs'
```

will now result in a WHERE clause more like

```sql
WHERE uid = 'joebloggs'
```

where the term on the RHS has been run through the same `_username_to_uid` function that is used to generate the "uid" field in the first place.

Making this change in the User model rather than higher in the code could be argued to be a bit surprising if a developer isn't already familiar with this behaviour. On balance I think it's safer to put it here, as I can't think of any valid reason for wanting to look up a user by a literal username match rather than using this more liberal behaviour.

### DIFF
--- a/h/models/user.py
+++ b/h/models/user.py
@@ -41,6 +41,19 @@ class UserIDComparator(Comparator):
                        val['domain'] == self.authority)
 
 
+class UIDComparator(Comparator):
+    """
+    Custom comparator for the User.username property.
+
+    This ensures that lookups by username are actually always done against the
+    :py:attr:`~h.models.user.User.uid` property. This ensures that username
+    normalisation is consistently applied when retrieving users by username
+    throughout the application.
+    """
+    def __eq__(self, other):
+        return self.__clause_element__() == _username_to_uid(other)
+
+
 class UserFactory(object):
     """Root resource for routes that look up User objects by traversal."""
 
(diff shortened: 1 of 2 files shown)


nickstenning:
I don't think this observation should necessarily block this from merging, but it has occurred to me that we might want to consider getting rid of the `uid` column in favour of a unique index on `replace(lower(username), '.', '')` -- it might simplify the user model somewhat, and would also make it more obvious what to do with `User.username` when used in an expression like `session.query(User).filter(User.username == User.display_name)`.

nickstenning:
I've just had a quick look into the approach I suggest above, and it's actually probably about as easy as this, although will involve some migrations. I'm going to close this PR as there's no point asking people to review this and then another thing which achieves the same ends in a nicer way...

PR #4397, opened 2017-03-02
Match dot- and case-variants when looking up by username

nickstenning:
We've had a number of support tickets from users who have created their accounts with usernames such as "JoeBloggs", and are later confused because they can't log in as "joebloggs".

This gets particularly confusing when, having had the login form tell them "User does not exist," they go and try and sign up again, at which point the signup form tells them that the username is taken.

This commit addresses this at the database and ORM level, by ensuring that all for comparisons against the username column, both sides of the comparison are normalised first. For example

```python
request.db.query(models.User).filter_by(username='Joe.Bloggs')
```

which would previously have resulted in a WHERE clause of

```sql
WHERE username = 'Joe.Bloggs'
```

will now result in a WHERE clause more like

```sql
WHERE lower(replace(username,     '.', '')) =
      lower(replace('Joe.Bloggs', '.', ''))
```

In order to make this feasible, I've made some changes to the user model. The separate `uid` column is gone, replaced by a unique functional index on the tuple:

    lower(replace(username, '.', '')), authority

This simultaneously achieves three ends:

1. It makes lookups by normalised username fast.
2. It ensures uniqueness of (normalised username, authority)
3. It (together with some changes to the UserIDComparator) makes lookups by userid fast.

With this in place, we're also able to remove the index on `authority` alone.

_**N.B.** This depends on #4396 being deployed and migrations run before merging._

### DIFF
--- a/h/models/user.py
+++ b/h/models/user.py
@@ -5,8 +5,9 @@ import re
 
 import sqlalchemy as sa
 from sqlalchemy.ext.hybrid import Comparator, hybrid_property
+from sqlalchemy.ext.declarative import declared_attr
 
-from h._compat import text_type
+from h._compat import string_types, text_type
 from h.db import Base
 from h.security import password_context
 from h.util.user import split_user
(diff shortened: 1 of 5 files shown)


h/models/user.py
```diff
@@ -59,27 +107,30 @@ def __getitem__(self, username):
...
 class User(Base):
     __tablename__ = 'user'
-    __table_args__ = (
-        sa.UniqueConstraint('email', 'authority'),
-        sa.UniqueConstraint('uid', 'authority'),
-        sa.UniqueConstraint('username', 'authority'),
-    )
+
+    @declared_attr
+    def __table_args__(cls):
```
nickstenning: This has to become a `declared_attr`, so that the `ix__user__user` index can reference the `username` column.

h/models/user.py
```diff
@@ -18,27 +19,74 @@
...
 
+def _normalise_username(username):
+    # We normalize usernames by dots and case in order to discourage attempts
+    # at impersonation.
+    return sa.func.lower(sa.func.replace(username,
+                                         sa.text("'.'"),
+                                         sa.text("''")))
+
+
+class UsernameComparator(Comparator):
```
fatbusinessman: Am I right in surmising that the reason you need to implement this is a full-on `Comparator` rather than using the more concise `@username.expression` is so that it’ll also convert the other side of the lookup? So that something like `query(User).filter(User.username == 'Enterprise.Synergy')` will still work?

I think that’s what you mean in the comment, but I just wanted to check my understanding that `@username.expression` wouldn’t do that for you.

h/models/user.py
```diff
@@ -18,27 +19,74 @@
...
+            (lower(replace('luis.silva', '.', '')), 'example.com')
     """
     def __init__(self, username, authority):
         self.username = username
         self.authority = authority
 
+    def __clause_element__(self):
+        return sa.tuple_(_normalise_username(self.username), self.authority)
+
     def __eq__(self, other):
```
fatbusinessman: I _think_ I understand what this is now doing, and the different types it will deal with for `other`, but is it worth adding a note here explaining why this has the two different branches (for `string_types` and for everything else)?

h/models/user.py
nickstenning: That's correct, but only half of the story. Here's the other half. Consider if we had the following code:

```python
@hybrid_property
def username(self):
    return self._username

@username.expression
def username(self):
    return _normalise_username(self._username)
```

Then a query like `session.query(User.username)` would return *normalised* usernames, because `.expression` gets used wherever an expression is needed, including in `SELECT` clauses.

Using a distinct `.expression` and `.comparator` means that you can control comparisons separately, so with the code on this branch you can do something like:

```python
session.query(User.username).filter(User.username.like('david%'))
```

And you'll get back unnormalised usernames that match the filter, where the filter is evaluated against *normalised* usernames.

h/models/user.py
nickstenning: 👍 good suggestion. I've added a docstring to this function.

nickstenning:
Thanks for the review, @fatbusinessman. Perhaps @chdorner could look at this this morning so we can get it out?

PR #4399, opened 2017-03-03
Remove Hypothesis client from service

robertknight:
**~~Do not merge until Monday 2017-03-06~~ _Go for it!_**

This PR removes the ability for the service to serve the Hypothesis client directly in favor of always loading the client from an external URL, defaulting to cdn.hypothes.is

The changes are:
 1. Remove code paths in client views (/app.html, /embed.js) for serving client assets directly, keeping just the versions which use the externally hosted boot script
 2. Remove the client asset environment which supported the code paths in (1)
 3. Remove the 'hypothesis' npm package which provided the assets used by (2)


nickstenning:
There's also stuff in the Dockerfile which can be simplified now.

robertknight:
> There's also stuff in the Dockerfile which can be simplified now.

Just [this line](https://github.com/hypothesis/h/blob/master/Dockerfile#L54) or is there anything else?

seanh:
There's still some mentions of `<script src="https://hypothes.is/embed.js" async></script>` in the `docs` directory. Looks like this 302s to `cdn.hypothes.is/hypothesis` now. Do we want to leave it like that or we we want to change the docs to say `<script src="cdn.hypothes.is/hypothesis"` directly?

robertknight:
I think we can leave it like that for now and reconsider whether to change the _publicly advertised_ entry point in future. cdn.hypothes.is is an "implementation detail" for most users at the moment, unless they have strict CSP policies.

seanh:
So it's deliberate that we ourselves even continue to use the `/embed.js` URL that 302s to the CDN, instead of just using the CDN URL directly? For example in `/app.html` it's still `<script src="/embed.js"></script>` and not `<script src="https://cdn.hypothes.is/hypothesis"></script>`

seanh:
Should `use_client_boot_script` be added to `FEATURES_PENDING_REMOVAL`?

seanh:
Can the `assets_client` route be removed from `routes.py`?

seanh:
I think I share the concern that you mentioned in Slack about potential confusion with dev envs. It looks like by default `https://localhost:5000/docs/help` will make a request to `localhost/embed.js` which will 302 to `https://cdn.hypothes.is/hypothesis`. For some reason I'm also seeing a second `embed.js` request, this time to `hypothes.is/embed.js`, as well. And then the sidebar gets loaded from cdn.hypothes.is, I login to it with my _production_ hypothesis account and it sends annotations to the production hypothesis API. Definitely seems a bit confusing

seanh:
Hmm, I think I'm gonna have to figure out how to run the client repo's `gulp watch` over HTTPS :). I still can't run the local dev web service over HTTP (it still looks up Chrome horrifically due to gunicorn running out of workers), and when running h over HTTPS pages like `/docs/help` can't embed my local dev client unless that is also HTTPS. (For now I can run h over HTTP and just increase the number of gunicorn workers so it doesn't lock up.)

seanh:
Regarding the `/embed.js` URL vs the CDN URL - I wonder if it might be worth coming up with two distinct names for these two URLs and using them consistently?

For example in `views/client.py` there's `DEFAULT_CLIENT_URL = 'https://cdn.hypothes.is/hypothesis'` (and also the `h.client_url` setting and env var that can override this), so it seems like "the client URL" is the CDN URL. But then further down in the file there is `client_url = request.route_path('embed')`, another variable named client url but holding a different URL (a URL which, when loaded, 302s to the first client URL).

I think it might be a little confusing.

I think "client url" is a good name for the CDN URL and the `DEFAULT_CLIENT_URL` / `h.client_url` settings, and maybe just "embed url" should be used consistently as the name for the embed.js URL (and its associated route and view callable). The embed URL is the one that people embed into pages, the client URL is the one where the client is actually served.

robertknight:
> Should use_client_boot_script be added to FEATURES_PENDING_REMOVAL?

Done.

> Can the assets_client route be removed from routes.py?

Done.

> But it seems less bad than embedding the prod client in a dev web service

The plan for this is that the `/docs/help` page will by default load the _production client_ but use the _local service for storage_. This means that you can develop the web service without needing to set up a build of the client. The way this will be done is by having pages such as `/docs/help` include a config script tag which causes the client to load `app.html` from the local service (which includes the API endpoints from the local service) but use the production client assets. Does this make sense to you?

I'd like to implement this as a separate change from this PR however.

robertknight:
> Hmm, I think I'm gonna have to figure out how to run the client repo's gulp watch over HTTPS

It is possible to make an express server use SSL. For the purposes of testing this out I suggest just configuring gunicorn to use a different type of worker locally.

seanh:
> The plan for this is that the `/docs/help` page will by default load the _production client_ but use the
> _local service for storage_. This means that you can develop the web service without needing to set up a 
> build of the client. The way this will be done is by having pages such as `/docs/help` include a config 
> script tag which causes the client to load `app.html` from the local service (which includes the API 
> endpoints from the local service) but use the production client assets. Does this make sense to you?

Yeah, that makes sense. So you want to merge it as is (dev site embeds prod client that talks to prod service) for now?

robertknight:
> So you want to merge it as is (dev site embeds prod client that talks to prod service) for now?

Yes please, I'll follow up with a small PR to resolve the issue for `/docs/help` separately.

seanh:
Ok, this sounds good to me, just waiting for a response to my question about whether we ourselves want to be using embed.js or just using the CDN URL directly? And to my suggestion about naming the embed.js and the CDN URLs more distinctly in the code.

seanh:
I'm still seeing references to `app_js` in some tests, though I guess they don't do any harm

seanh:
Ok, looks good to me. Great code surgery @robertknight, loads of code and complexity removed here, and also releasing, deploying and setting up dev envs for everything is becoming steadily simpler and more decoupled

PR #4400, opened 2017-03-03
Remove unnecessary _app_html_context function

robertknight:
**Depends on https://github.com/hypothesis/h/pull/4399**

This removes an unnecessary `_app_html_context` helper function from the client views in favor of inlining its contents into the one place where it was used and a `url_with_path` helper this is also no longer required because the way we generate the service URL is now always guaranteed to include the path.

### DIFF
--- a/h/views/client.py
+++ b/h/views/client.py
@@ -13,7 +13,6 @@ import json
 from pyramid.httpexceptions import HTTPFound
 from pyramid.view import view_config
 
-from h._compat import urlparse
 from h import session as h_session
 from h.auth.tokens import generate_jwt
 from h.util.view import json_view
@@ -24,13 +23,6 @@ from h import __version__
 DEFAULT_CLIENT_URL = 'https://cdn.hypothes.is/hypothesis'
 
 
-def url_with_path(url):
-    if urlparse.urlparse(url).path == '':
-        return '{}/'.format(url)
-    else:
-        return url
-
-
 def _client_url(request):
     """
     Return the configured URL for the client.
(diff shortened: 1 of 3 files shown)


seanh:
I'll get this after 4399 merges

h/views/client.py
```diff
@@ -24,36 +23,34 @@
...
-def _app_html_context(api_url, service_url, sentry_public_dsn,
-                      websocket_url, auth_domain, ga_client_tracking_id,
-                      client_url):
+@view_config(route_name='sidebar_app',
+             renderer='h:templates/app.html.jinja2')
+def sidebar_app(request, extra=None):
     """
-    Returns a dict of asset URLs and contents used by the sidebar app
-    HTML tempate.
+    Return the HTML for the Hypothesis client's sidebar application.
```
seanh: Don't worry about it here but I think being more careful about these names may be something to discuss in this client technical debt meeting. In the code (mostly in the client code but also sometimes in h) we have multiple different things called app, and multiple called sidebar, and now here is "sidebar app" :)

I think there are at least three distinct things: **1.** The `embed.js` script when it's used in the parent document (called `src/annotator/` in the client repo) **2.** The `app.html` page that contains the `<hypothesis-app>`, a `"js-hypothesis-config"` object, and the `<script src="/embed.js">` **3.** The actual AngularJS app that's delivered in `embed.js` (the second time that script is run) and rooted at the `<hypothesis-app>` element. There may be more bits that I'm forgetting. But we need to come up with distinct names for each component and then use them consistently.

h/views/client.py
```diff
@@ -24,36 +23,34 @@
...
+
+    :param extra: A dict of optional properties specifying link tags and meta
+                  attributes to be included on the page.
     """
 
-    # the serviceUrl parameter must contain a path element
-    service_url = url_with_path(service_url)
+    settings = request.registry.settings
+    client_url = request.route_path('embed')
```
seanh: Can we rename this to `embed_url` or `embed_js_url`?

h/views/client.py
```diff
@@ -74,33 +71,11 @@ def _app_html_context(api_url, service_url, sentry_public_dsn,
             'googleAnalytics': ga_client_tracking_id
         })
 
-    return {
+    ctx = {
         'app_config': json.dumps(app_config),
         'client_url': client_url,
```
seanh: Same rename here too

seanh:
...and the rename is already done

PR #4401, opened 2017-03-03
Remove client dev install docs

seanh:
There are new client dev install docs in the client repo now, remove these old and out of date ones from readthedocs.

Note: This means that the [website dev install page](http://h.readthedocs.io/en/latest/developing/install/website/) will now appear at <http://h.readthedocs.io/en/latest/developing/install/> (where there was previously an index page linking to separate website and client install pages). The `developing/install/website/` and `developing/install/client/` URLs won't work anymore. (We could put stub pages there.)


robertknight:
> The developing/install/website/ and developing/install/client/ URLs won't work anymore. (We could put stub pages there.)

There are a handful of useful annotations on https://h.readthedocs.io/en/latest/developing/install/website/ which will break if we do nothing. If we can fix it so that those annotations transfer that would be ideal.

seanh:
After this has merged, in the readthedocs admin, we can specify a [page redirect](http://docs.readthedocs.io/en/latest/user-defined-redirects.html#page-redirects) that will cause the old pages to 302 to the new page. But I don't think that will cause h to move the annotations. I can't think of any way in RTD that we can get the annotations moved. We could of course do a db migration on the production db to move them.

robertknight:
@seanh - I'm going to merge this PR now and open a separate issue to discuss redirects. We could solve the problem specifically for this URL but it applies to the web as a whole, so we should have a think about how we could make H handle it automatically.

seanh:
Sounds good - cheers Rob.

PR #4404, opened 2017-03-06
Remove unused indices and columns from user table

nickstenning:
As of c9a5aeb several indices (or constraints) on the "user" table are
redundant or unnecessary. Namely:

1. uq__user__uid, which guarantees the uniqueness of the (uid, authority) tuple, is more-or-less directly replaced by ix__user__userid.

2. uq__user__username, which guarantees the uniqueness of the (username, authority) tuple, is no longer needed as it if two usernames satisfy ix__user__userid's uniqueness requirement they are clearly unique. We also do not do lookups against the raw "username" so this isn't needed for performance reasons.

3. ix__user__authority, which indexed the authority column. Any queries which need to query against authority can now use the compound index provided by ix__user__userid, e.g.

       postgres=# explain select count(*) from "user" where authority = 'test.foobar.org';
                                               QUERY PLAN
       ---------------------------------------------------------------------------------------------
       Aggregate  (cost=2249.51..2249.52 rows=1 width=8)
         ->  Index Only Scan using ix__user__userid on "user" (cost=0.42..2249.51 rows=1 width=0)
               Index Cond: (authority = 'test.foobar.org'::text)
       (3 rows)

Also removes the `uid` column.

### DIFF
--- a/h/migrations/versions/18dfed902c9e_remove_unused_user_indices.py
+++ b/h/migrations/versions/18dfed902c9e_remove_unused_user_indices.py
@@ -1,0 +1,42 @@
+"""
+Remove unused user indices
+
+Revision ID: 18dfed902c9e
+Revises: 7e2443f8d7d6
+Create Date: 2017-03-03 12:13:16.122752
+"""
+
+from __future__ import unicode_literals
+
+from alembic import op
+import sqlalchemy as sa
+
+
+revision = '18dfed902c9e'
+down_revision = '7e2443f8d7d6'
+
+
+def upgrade():
+    op.drop_constraint('uq__user__uid', 'user')
+    op.drop_constraint('uq__user__username', 'user')
+    op.drop_index(op.f('ix__user__authority'), 'user')
+
+
+def downgrade():
+    op.execute('COMMIT')
+
+    op.create_index(op.f('ix__user__authority'), 'user', ['authority'],
+                    postgresql_concurrently=True)
+
+    # We add the index back first, and then create a constraint from the
+    # index, to avoid holding the table locked for the entire time the index
+    # is building.
+    op.create_index(op.f('uq__user__uid'), 'user', ['uid', 'authority'],
+                    unique=True,
+                    postgresql_concurrently=True)
+    op.execute(sa.text('ALTER TABLE "user" ADD CONSTRAINT uq__user__uid UNIQUE USING INDEX uq__user__uid'))
+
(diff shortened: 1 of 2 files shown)


fatbusinessman:
> Any queries which need to query against authority can now use the compound index provided by ix__user__userid

I was a little confused by this, as I’m used to a world where this only works if you’re using the first _n_ columns of the index. I took a look at [the Postgres docs](https://www.postgresql.org/docs/8.2/static/indexes-multicolumn.html), and found:

> A multicolumn B-tree index can be used with query conditions that involve any subset of the index's columns, but the index is most efficient when there are constraints on the leading (leftmost) columns.

So that looks like using this index is going to be less efficient than directly indexing `authority`; is that fine? I don’t know how often this crops up, or how much of a performance hit this will produce on 80,000 or so rows.

fatbusinessman:
> uq__user__username, which guarantees the uniqueness of the (username, authority) tuple, is no longer needed as it if two usernames satisfy ix__user__userid's uniqueness requirement they are clearly unique.

I’m not seeing a uniqueness requirement on `ix__user__userid` (or the `uq` prefix which I’m guessing goes with it); am I missing something?

nickstenning:
Yeesh. You're right, that's missing from the migrations. We should probably fix that ;)

nickstenning:
Marked as WIP pending #4405.

nickstenning:
Remove the WIP label as this should now be ok to go. 

> So that looks like using this index is going to be less efficient than directly indexing authority; is that fine? I don’t know how often this crops up, or how much of a performance hit this will produce on 80,000 or so rows.

Yep, thank you for checking that. As it is I don't know of any cases where we filter users by authority alone, so I think we'll be ok here. If we detect slowness in New Relic then we can always put it back.

fatbusinessman:
Ok, this all looks sensible to me now, although when I try to run the migration to remove the `uid` column, I get this error:

```
sqlalchemy.exc.InternalError: (psycopg2.InternalError) cannot drop table "user" column uid because other objects depend on it
DETAIL:  view metrics_user_edu depends on table "user" column uid
```

I’m guessing this is something to do with our Metabase instance. Is this something that would be fixed if I had a more recent data dump?

nickstenning:
Yep, that's a view I added to fix a problem in Metabase. Should be gone as of this morning's dump -- I replaced it with a [CTE](https://www.postgresql.org/docs/current/static/queries-with.html).

PR #4405, opened 2017-03-07
Fix userid index in database

nickstenning:
Unfortunately, the migration in c9a5aeb12 was incorrect, as it failed to specify that the index to be added was unique.

This migration fixes the problem by moving the old index out of the way, recreating it as unique, and then dropping the old index.

Note that the index is still referred to using a normal index prefix ("ix__") rather than a unique constraint prefix ("uq__") because unique constraints cannot be created using functional indices.

### DIFF
--- a/h/migrations/versions/7e2443f8d7d6_make_userid_index_unique.py
+++ b/h/migrations/versions/7e2443f8d7d6_make_userid_index_unique.py
@@ -1,0 +1,42 @@
+"""
+Make userid index unique
+
+Revision ID: 7e2443f8d7d6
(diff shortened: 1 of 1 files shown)


nickstenning:
> At least once I read through the Postgres docs and realised that a unique constraint was its own distinct thing.

Right -- a unique constraint is defined by its effect (no two non-null entries can have the same value) but uses a unique index under the hood to achieve this end. That's strictly considered an implementation detail in Postgres land, so using a unique index to achieve a unique constraint in this way is vaguely frowned upon. But only very vaguely.

PR #4406, opened 2017-03-07
Make the /docs/help page use the local Hypothesis service for annotation storage

robertknight:
By default the service uses the production client hosted at cdn.hypothes.is but
we still want pages hosted by the service to use the local service for annotation storage and the 'app.html' content.

Both of these things can be achieved by configuring the client on the `/docs/help` page to use the sidebar app HTML page served by the local service.

We might also want to do this for the bookmarklet but I'll make that a separate change.

### DIFF
--- a/h/templates/help.html.jinja2
+++ b/h/templates/help.html.jinja2
@@ -7,12 +7,16 @@
   {{ super() }}
   <script type="application/json" class="js-hypothesis-config">
     {
-      "firstRun": true, {# TODO: remove after 2016-09-01, by which time old clients will be gone. #}
       "openSidebar": true,
       "openLoginForm": true
     }
   </script>
   {% if not is_onboarding %}
+    <script type="application/json" class="js-hypothesis-config">
+    {
+      "sidebarAppUrl": "{{ request.route_url('sidebar_app') }}"
+    }
+    </script>
     <script async defer src="{{ embed_js_url }}"></script>
   {% endif %}
 {% endblock %}


h/templates/help.html.jinja2
```diff
@@ -7,12 +7,16 @@
   {{ super() }}
   <script type="application/json" class="js-hypothesis-config">
     {
-      "firstRun": true, {# TODO: remove after 2016-09-01, by which time old clients will be gone. #}
       "openSidebar": true,
       "openLoginForm": true
     }
   </script>
   {% if not is_onboarding %}
+    <script type="application/json" class="js-hypothesis-config">
```
robertknight: This is in a separate config block so that we only set the sidebar app URL if the client is being embedded on the page.

seanh:
I think this highlights a potential issue with the current config system, but that's a separate problem that I hope we can discuss soon.

The prod client talking to local annotation service that this PR sets up works. What I'm seeing is this:

* `localhost:5000/docs/help` contains a `<script src="localhost:5000/embed.js">`
* `localhost:5000/embed.js` redirects to `https://cdn.hypothes.is/hypothesis`, so the JavaScript code that gets injected into the parent page is the prod code 
* The prod code in the parent page creates an `<iframe src="localhost:5000/app.html">`
* `localhost:5000/app.html` contains another `<script src="localhost:5000/embed.js">`
* `localhost:5000/embed.js` redirects to `https://cdn.hypothes.is/hypothesis` again, so the JavaScript code that gets loaded in the iframe is also the prod code
* That prod iframe code is configured to use the localhost annotations API, so you see requests to `localhost:5000/api/search` for example (it also logs in to the localhost annotation service, so for example the `/api/token` `GET` request is to localhost)

PR #4409, opened 2017-03-07
Clear user service cache on transaction end

### LINKED ISSUE #4398: Detached instance when loading session model with invalid CSRF token
This is the issue for tracking the following error reports:
* https://sentry.io/hypothesis/h/issues/227367217/
* https://sentry.io/hypothesis/h/issues/227367212/
* https://sentry.io/hypothesis/h/issues/227727131/
* https://sentry.io/hypothesis/h/issues/227727128/

### Steps to reproduce
I did not yet find a way to reproduce this.

### Expected behaviour
No exception.

### Actual behaviour
💥 `Parent instance <User at 0x7f91ddb75d10> is not bound to a Session; lazy load operation of attribute 'groups' cannot proceed`

### Additional details
The errors are coming from at least the Chrome extension and via, I haven't seen embed yet but we can assume that this happens for all types of environments the clients is running.

Following the Sentry error report and the code it looks like this is happening:

1. User submits the login in sidebar
2. Pyramid thinks the CSRF token is wrong and throws `BadCSRFToken`
3. [`h.views.accounts.bad_csrf_token_json`](https://github.com/hypothesis/h/blob/aec5e9511654d4f409c2ef3c5b2326ffcff35df6/h/views/accounts.py#L62) is getting called
4. The view renders the session model by calling [`h.sessions.model`](https://github.com/hypothesis/h/blob/aec5e9511654d4f409c2ef3c5b2326ffcff35df6/h/session.py#L8-L15)
5. It tries to [load the groups](https://github.com/hypothesis/h/blob/aec5e9511654d4f409c2ef3c5b2326ffcff35df6/h/session.py#L58-L71)
6. It succeeds to load the authority group
7. It tries to [load the user's groups](https://github.com/hypothesis/h/blob/aec5e9511654d4f409c2ef3c5b2326ffcff35df6/h/session.py#L74-L78) - and this is the place where `request.authenticated_user` is not `None`, but a detached instance so `request.authenticated_user.groups` explodes.

I don't really know why the client is posting the login form when the user is already logged-in because we check the CSRF token before we handle the login request. So the user must already have a valid auth ticket. This is also the reason why I'm struggling to reproduce it.

The user itself is detached because any exception raised will rollback the transaction and detach any objects that are cached in the database session.

chdorner:
We previously cleared the user service cache on `after_commit` and
`after_rollback` SQLAlchemy events. This caused issues where the view
was raising an exception, because SQLAlchemy only fires the
`after_rollback` event for an explicit rollback.

pyramid_tm and zope.sqlalchemy will close the database transaction when
the view raises an exception. A call to `SessionTransaction.close` does
implicitly send a `ROLLBACK` SQL query to the database, but it doesn't
emit the `after_rollback` event that we're relying on in order to clear
out the user service cache.
This results in a exception views potentially accessing detached
database objects which in turn throws another exception, which then
returns a generic 500 error back to the user.

For following all of this along at home:

1. pyramid_tm starts a transaction at the beginning of a request:
https://github.com/Pylons/pyramid_tm/blob/1.1.1/pyramid_tm/__init__.py#L86
2. pyramid_tm catches exceptions and if the error is not retryable,
aborts the transaction:
https://github.com/Pylons/pyramid_tm/blob/1.1.1/pyramid_tm/__init__.py#L116
3. The `abort` that pyramid_tm sends is finishing the transaction:
https://github.com/zopefoundation/zope.sqlalchemy/blob/0.7.7/src/zope/sqlalchemy/datamanager.py#L91
4. Finishing a transaction in turn will close it:
https://github.com/zopefoundation/zope.sqlalchemy/blob/0.7.7/src/zope/sqlalchemy/datamanager.py#L87

Whereas an explicit rollback works as we expected it to:
1. `sqlalchemy.orm.session.SessionTransaction.rollback` checks if it
needs to send a `ROLLBACK` SQL query to the database, and if it does...
(https://bitbucket.org/zzzeek/sqlalchemy/src/ce17fd4bf361e6cd08b6e698e49c118643abe724/lib/sqlalchemy/orm/session.py?at=rel_1_1_4&fileviewer=file-view-default#session.py-490)
2. calls `_rollback_impl` which emits the `after_rollback` event
(https://bitbucket.org/zzzeek/sqlalchemy/src/ce17fd4bf361e6cd08b6e698e49c118643abe724/lib/sqlalchemy/orm/session.py?at=rel_1_1_4&fileviewer=file-view-default#session.py-532).

Fixes #4398.

### DIFF
--- a/h/services/user.py
+++ b/h/services/user.py
@@ -40,10 +40,17 @@ class UserService(object):
         self._cache = {}
 
         # But don't allow the cache to persist after the session is closed.
-        @sqlalchemy.event.listens_for(session, 'after_commit')
-        @sqlalchemy.event.listens_for(session, 'after_rollback')
-        def flush_cache(session):
-            self._cache = {}
+        #
+        # This makes sure that under all circumstances whenever the transaction
+        # ends the cache gets cleared. We used to only clear the cache on
+        # `after_commit` and `after_rollback`, but the latter only get called
+        # on explicit rollbacks, a rollback due to an exception does not fire
+        # off this event.
+        @sqlalchemy.event.listens_for(session, 'after_transaction_end')
+        def flush_cache(session, transaction):
+            # We only clear the cache when the top-level transaction finishes.
+            if transaction.parent is None:
+                self._cache = {}
 
     def fetch(self, userid_or_username, authority=None):
         """
--- a/h/services/user.py
+++ b/h/services/user.py
@@ -2,8 +2,6 @@
 
 from __future__ import unicode_literals
 
-import sqlalchemy
-
 from h.models import Annotation, User
 from h import util
 
@@ -40,17 +38,9 @@ class UserService(object):
         self._cache = {}
 
         # But don't allow the cache to persist after the session is closed.
-        #
-        # This makes sure that under all circumstances whenever the transaction
-        # ends the cache gets cleared. We used to only clear the cache on
-        # `after_commit` and `after_rollback`, but the latter only get called
-        # on explicit rollbacks, a rollback due to an exception does not fire
-        # off this event.
-        @sqlalchemy.event.listens_for(session, 'after_transaction_end')
-        def flush_cache(session, transaction):
-            # We only clear the cache when the top-level transaction finishes.
-            if transaction.parent is None:
-                self._cache = {}
+        @util.db.on_transaction_end(session)
+        def flush_cache():
+            self._cache = {}
 
     def fetch(self, userid_or_username, authority=None):
         """
--- a/h/util/db.py
+++ b/h/util/db.py
@@ -1,0 +1,35 @@
+# -*- coding: utf-8 -*-
+
+from __future__ import unicode_literals
+
+import sqlalchemy
+
+
+def on_transaction_end(session):
+    """
+    Decorator for a function which should run after a top-level transaction ended.
+
+    Transactions that are either implicitly or explicitly committed or rolled back will be
+    closed at the end of a Pyramid view. This is here for cleaning up caches so that
+    code after the view, exception views for example, will not be able to access
+    detached instances.
+
+    Example usage:
+
+    .. code-block:: python
+
+       @util.db.on_transaction_end(session)
+       def flush_cache():
+           self._cache = {}
+
+    """
+    def decorate(func):
+        def _handler(_, transaction):
+            # We only clear the cache when the top-level transaction finishes.
+            if transaction.parent is None:
+                func()
+
+        sqlalchemy.event.listen(session, 'after_transaction_end', _handler)
+        return func
+
+    return decorate
--- a/h/services/groupfinder.py
+++ b/h/services/groupfinder.py
@@ -3,11 +3,11 @@
 from __future__ import unicode_literals
 
 from zope.interface import implementer
-import sqlalchemy
 
 from memex.interfaces import IGroupService
 
 from h import models
+from h import util
 from h.groups.util import WorldGroup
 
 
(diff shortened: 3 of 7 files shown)


h/services/user.py
```diff
@@ -40,10 +40,18 @@ def __init__(self, default_authority, session):
...
+        #
+        # This makes sure that under all circumstances whenever the transaction
+        # ends the cache gets cleared. We used to only clear the cache on
+        # `after_commit` and `after_rollback`, but the latter only get called
+        # on explicit rollbacks, a rollback due to an exception does not fire
+        # off this event.
+        @sqlalchemy.event.listens_for(session, 'after_transaction_end')
+        def flush_cache(session, transaction):
+            if transaction.parent is None:
+                # We only clear the cache when the top-level transaction
```
nickstenning: This comment would make more sense to me above the conditional.

nickstenning:
BTW, if it makes testing easier, you might want to pull out the "after_transaction_end" stuff into a utility decorator in `h/utils/something.py`, and then use that decorator in the user service and groupfinder service. That way you can just test that the services use the right decorator, and you only need to do the crazy stuff with the `db_session` fixture once.

nickstenning:
And, frankly, you might not need to do it at all, because at that point you're really going to be testing SQLAlchemy...

chdorner:
I've refactored this into a util function, which is tested by emitting the appropriate event in the tests which allows us to only test the implementation and not SQLAlchemy. In addition to that, I changed the groupfinder service as well.

This should be ready to review again.

h/util/db.py
```diff
@@ -0,0 +1,36 @@
...
+def on_transaction_end(session):
+    """
+    Decorator for a function which should run after a top-level transaction ended.
+
+    Transaction that are either implicitly or explicity committed or rolled back will be
+    closed at the end of a Pyramid view. This is here for cleaning up caches so that
+    code after the view, exception views for example, will not be able to access
+    detached instances.
+
+    The wrapped function will receive the database session as ``session`` and the
```
nickstenning: Is there any reason to pass these through at the moment? I think I'd lean towards not doing so unless there's a need, as it will at least prevent unused argument warnings.

h/util/db.py
```diff
@@ -0,0 +1,36 @@
...
+    Example usage:
+
+    .. code-block:: python
+
+       @util.db.on_transaction_end(session)
+       def flush_cache(session, transaction):
+           self._cache = {}
+
+    """
+    def wrapped(func):
```
nickstenning: The naming here is a bit confusing, because `wrapped` is actually a decorator, not a wrapped function. Calling this `decorate` would probably be a bit clearer, and is [sort of conventional](https://github.com/zzzeek/sqlalchemy/blob/c04870ba7b8098c7d408ad66f60efe7229496fde/lib/sqlalchemy/event/api.py#L92-L126).

h/util/db.py
```diff
@@ -0,0 +1,36 @@
...
+        @sqlalchemy.event.listens_for(session, 'after_transaction_end')
+        def after_transaction_end(current_session, transaction):
+            # We only clear the cache when the top-level transaction finishes.
+            if transaction.parent is None:
+                func(current_session, transaction)
```
nickstenning: So this is a decorator, which means it should return the function that it's wrapping. In this particular case that's easy, because you don't actually need to modify the function at all (that's not we made it a decorator), so you can use [a similar pattern to the `listens_for` decorator itself](https://github.com/zzzeek/sqlalchemy/blob/c04870ba7b8098c7d408ad66f60efe7229496fde/lib/sqlalchemy/event/api.py#L92-L126).

So maybe something like:

```python
def on_transaction_end(session):
    def decorate(func):
        def _handler(_, transaction):
            if transaction.parent is None:
                func()
        sqlalchemy.event.listen(session, 'after_transaction_end', _handler)
        return func
    return decorate
```

h/util/db.py
```diff
@@ -0,0 +1,35 @@
...
+from __future__ import unicode_literals
+
+import sqlalchemy
+
+
+def on_transaction_end(session):
+    """
+    Decorator for a function which should run after a top-level transaction ended.
+
+    Transaction that are either implicitly or explicity committed or rolled back will be
```
nickstenning: *Transactions

nickstenning: *explicitly

PR #4410, opened 2017-03-07
Add Referrer-Policy and X-XSS-Protection headers

nickstenning:
This commit adds a [Referrer-Policy](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Referrer-Policy) and [X-XSS-Protection](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-XSS-Protection) header to every response.

Referrer-Policy aims to address problems of URL leakage -- users visiting HTTPS web pages expect the URLs they visit to be protected, which they're not unless the Referer information is appropriately redacted. A value of "strict-origin-when-cross-origin" means that:

1. pages on the same origin will receive a full Referer header

2. pages on a different secure origin will receive an origin-only Referer header (i.e. the URL path is redacted)

3. pages on a different insecure origin will receive no Referer header.

This commit also adds a strict X-XSS-Protection header. This will be largely unnecessary when we deploy a strict CSP policy, but can nonetheless protect users from reflected XSS in older (pre-CSP) browsers.

### DIFF
--- a/h/app.py
+++ b/h/app.py
@@ -62,6 +62,7 @@ def includeme(config):
     config.add_tween('h.tweens.csrf_tween_factory')
     config.add_tween('h.tweens.auth_token')
     config.add_tween('h.tweens.content_security_policy_tween_factory')
+    config.add_tween('h.tweens.security_header_tween_factory')
 
     config.add_renderer('csv', 'h.renderers.CSV')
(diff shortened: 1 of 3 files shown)


nickstenning:
Oh yeah, I forgot I hadn't written any tests for this. 🤦‍♂️

PR #4413, opened 2017-03-09
Flag table migration

chdorner:
This is part 1 of hypothesis/product-backlog#164.

This migration should be deployed out to production and run before the next PR, which includes the model, should be merged.


nickstenning:
LGTM.

PR #4414, opened 2017-03-09
Flag model

chdorner:
~**The migration in #4413 has be to deployed out to production and run before this can be merged.**~

This is part 2 of hypothesis/product-backlog#164.

### DIFF
--- a/h/models/flag.py
+++ b/h/models/flag.py
@@ -1,0 +1,43 @@
+# -*- coding: utf-8 -*-
+
+from __future__ import unicode_literals
+
+import sqlalchemy as sa
+
+from memex.db import types
+
+from h.db import Base
+from h.db.mixins import Timestamps
+
+
+class Flag(Base, Timestamps):
+    """
+    A flag.
+
+    A user can flag up content, in this case just annotations,
+    which are then being moderated by group moderators.
+    """
+
+    __tablename__ = 'flag'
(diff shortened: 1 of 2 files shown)


chdorner:
I ran the migration on production, this can now be merged.

h/models/flag.py
```diff
@@ -0,0 +1,43 @@
...
+
+from memex.db import types
+
+from h.db import Base
+from h.db.mixins import Timestamps
+
+
+class Flag(Base, Timestamps):
+    """
+    A flag.
```
nickstenning: Hahaha :joy:

As much as I like terse documentation, I think it would probably good to assume a little less context here. Something like:

```
"""
A flag representing a user request for moderator attention.

Users can "flag" annotations if they believe that the annotation in question violates the 
content policy of the group or service, or otherwise needs moderator attention.
"""
```

h/models/flag.py
```diff
@@ -0,0 +1,43 @@
...
+
+    A user can flag up content, in this case just annotations,
+    which are then being moderated by group moderators.
+    """
+
+    __tablename__ = 'flag'
+    __table_args__ = (sa.UniqueConstraint('annotation_id', 'user_id'),)
+
+    #: The id of the flag.
```
nickstenning: Not sure this comment adds much.

PR #4415, opened 2017-03-09
Don't use an inline script to configure Google Analytics

nickstenning:
Using an inline script to set up Google Analytics prevents us from turning on a restrictive Content-Security-Policy.

Instead, dump the tracking ID into a meta tag if it's available, and retrieve it and set up the "ga" command queue in header.js.

An earlier commit removes some unnecessary configurability from our GA usage.

### DIFF
--- a/h/templates/layouts/base.html.jinja2
+++ b/h/templates/layouts/base.html.jinja2
@@ -49,16 +49,10 @@
           href="{{ asset_url('images/favicons/favicon.ico') }}">
 
     {% if ga_tracking_id %}
-      <!-- Google Analytics -->
-      <script async src='//www.google-analytics.com/analytics.js'></script>
-      <script>
-       window.ga=window.ga||function(){(ga.q=ga.q||[]).push(arguments)};ga.l=+new Date;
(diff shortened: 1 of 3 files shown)


h/templates/layouts/base.html.jinja2
```diff
@@ -49,16 +49,10 @@
...
-      <script>
-       window.ga=window.ga||function(){(ga.q=ga.q||[]).push(arguments)};ga.l=+new Date;
-       ga('create', '{{ga_tracking_id}}', '{{ga_cookie_domain}}');
-       {% block ga_pageview %}
-       ga('send', 'pageview');
-       {% endblock %}
-      </script>
-      <!-- End Google Analytics -->
+    <meta name="google-analytics-tracking-id" content="{{ ga_tracking_id }}">
+    <script async src="https//www.google-analytics.com/analytics.js"></script>
```
robertknight: Missing ':' after 'https'

nickstenning: Ooh! Good spot.

PR #4416, opened 2017-03-09
Allow CSP opt-outs on a per-view basis

nickstenning:
There are a lot of CSP exceptions being generated from `/app.html` (the client sidebar page) many of which appear to be the result of adware or misbehaving browser extensions.

It may well be a good idea to have CSP enabled on this page, but in order to see more clearly if there are CSP issues on the rest of the site, suppress CSP temporarily on this page.

We do this by first converting the CSP support from a tween to a view deriver, and then adding a view option (`csp_insecure_optout`) which can be used to opt a view out of CSP protection.

Probably best reviewed commit by commit.


h/viewderivers.py
```diff
@@ -0,0 +1,46 @@
...
+    ])
+
+    if info.registry.settings.get('csp.report_only', False):
+        header_name = 'Content-Security-Policy-Report-Only'
+    else:
+        header_name = 'Content-Security-Policy'
+
+    # Views can set ``csp_insecure_optout=True`` in their view options to
+    # disable CSP for the view.
+    if info.options.get('csp_insecure_optout'):
```
robertknight: I think it would be clearer if this line was moved up to just after the global CSP setting check, so the function looks like:

```
if not csp_enabled_globally:
  return view

if not csp_enabled_for_view:
  return view

# Prepare and return the wrapper view
```

nickstenning: Agreed! Pushed that change.

h/views/client.py
```diff
@@ -31,7 +31,8 @@ def _client_url(request):
 
 
 @view_config(route_name='sidebar_app',
-             renderer='h:templates/app.html.jinja2')
+             renderer='h:templates/app.html.jinja2',
+             csp_insecure_optout=True)
```
robertknight: The stream and single-page app views will still fail CSP checks because of some client-side issues. For example, annotations with LaTeX markup will try to render inline styles. Perhaps we should add this flag to the stream and single annotation views as well for the moment?

nickstenning: :+1:

PR #4418, opened 2017-03-10
Remove CSP-violating inline styles from search page

nickstenning:
These inline styles violate our tentative Content-Security-Policy, and are apparently easily removed.


robertknight:
LGTM. Tested by comparing the appearance of the user and group sidebar with and without the `@include pie-clearfix` line. Without the layout is obviously broken.

PR #4419, opened 2017-03-10
Upgrade to raven v6.0.0

nickstenning:


nickstenning:
As a general rule new versions of raven have in the past sent better metadata about what happened either side of a reported exception. My motivation for keeping this up to date is that, combined with wanting to ensure we don't end up with a version of raven that's incompatible with the hosted version of Sentry that we use.

PR #4422, opened 2017-03-10
Slim down the healthcheck endpoint

nickstenning:
Having the healthcheck/status endpoint talk to all the backing services is not the greatest idea in the world, on reflection.

The problem (apart from the fact that it's quite slow) is that it tightly couples the health of the service to the health of the backing services. If ElasticSearch is temporarily unavailable, a failing healthcheck will take the entire site offline as all instances are removed from the load balancer. By contrast, a healthcheck that returns only the health of the web application will allow those parts of the site that don't depend on ElasticSearch to keep working.

The counter-argument is that if a single instance is having networking issues, it is useful to have it removed from the load balancer so it doesn't keep serving requests.

This commit slims down the healthcheck endpoint to only check Postgres. Very little of the site currently works without a database so this seems like a reasonable compromise for now.

### DIFF
--- a/h/views/status.py
+++ b/h/views/status.py
@@ -4,7 +4,6 @@ from __future__ import unicode_literals
 
 import logging
 
-from h.celery import celery
 from h.exceptions import APIError
 from h.util.view import json_view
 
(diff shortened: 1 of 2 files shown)


fatbusinessman:
Is this information still valuable, even if it’s not at the same endpoint we use to decide whether to route traffic to the application? If, as you say, we have some kind of operational problems, would it still be useful to be able to access some kind of extended status page that says “yeah, I’m working, but just so you know, I can’t seem to find my search engine any more”?

chdorner:
@fatbusinessman it's usually easy to find out the issue if one of the backend services is down, either a look at the Grafana graphs or the server logs (even easier now with Skyliner) should immediately tell you if that's the case.

fatbusinessman:
Ok; cool. Thanks, @chdorner.

(Graphs to the rescue!)

PR #4424, opened 2017-03-13
Remove embedding docs

seanh:
These have moved to the client docs:
http://h.readthedocs.io/projects/client/en/latest/publishers/embedding/

I'll add a readthedocs page redirect from the deleted URL to the new
one: http://docs.readthedocs.io/en/latest/user-defined-redirects.html#page-redirects

I've also deleted an old embedding.rst stub page which was just a link
to the (now also deleted) publishers/embedding.rst page, leftover from
the last time this page got moved. I'll also add a readthedocs page
redirect for this one.


seanh:
I've added redirects in the h readthedocs project's admin, so http://h.readthedocs.io/en/latest/publishers/embedding/ and http://h.readthedocs.io/en/latest/embedding/ should now both redirect to the client docs.

fatbusinessman:
> http://h.readthedocs.io/en/latest/publishers/embedding/ and http://h.readthedocs.io/en/latest/embedding/ should now both redirect to the client docs.

Do you mean they should be redirecting _now_, or they should redirect once this PR gets merged? Because they don’t seem to be redirecting yet.

seanh:
They should be redirect now, I'm not sure why they aren't, let me try again

seanh:
Ok, I think the redirects will only start working after this is merged: http://docs.readthedocs.io/en/latest/user-defined-redirects.html#implementation

fatbusinessman:
Yup, looks like you were right. I just checked both these URLs post-merge, and they both redirect to suitable places in the client docs.

Nice work finding out this was an option. 👍

PR #4425, opened 2017-03-13
Add intersphinx mapping to client repo

seanh:
These means the :ref:`...` cross-references (and other kinds of
cross-references, like :js:func:, :option:, :envvar:, ...) can refer to
things defined in the client repo's docs, and Sphinx will generate URLs
to the client docs.

There aren't actually and refs in these docs that make use of this yet
but it's good to have it here.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -266,6 +266,6 @@ texinfo_documents = [
 # How to display URL addresses: 'footnote', 'no', or 'inline'.
 #texinfo_show_urls = 'footnote'
 
-
-# Example configuration for intersphinx: refer to the Python standard library.
-intersphinx_mapping = {'http://docs.python.org/': None}
+intersphinx_mapping = {
+    'client': ('https://h.readthedocs.io/projects/client/en/latest/', None),
+}


docs/conf.py
```diff
@@ -268,4 +268,7 @@ def setup(app):
 
 
 # Example configuration for intersphinx: refer to the Python standard library.
```
fatbusinessman: Probably worth removing this comment line, as this isn’t an example configuration any more.

docs/conf.py
```diff
@@ -268,4 +268,7 @@ def setup(app):
...
-intersphinx_mapping = {'http://docs.python.org/': None}
+intersphinx_mapping = {
+    'http://docs.python.org/': None,
```
fatbusinessman: If we’re no longer leaving this as the default example, do we still need this line? I’m not sure I know enough about `intersphinx` to know what (if anything) it’s doing.

seanh: This lets use cross-reference to things defined in the official Python documentation, although it uses the old `intersphinx_mapping` format, and I don't think we're using it at all. I'll see if it can be removed without problems, Sphinx prints compilation warnings for any broken refs so it's easy to find out if we have any refs that use this...

seanh: It looks like we didn't have any cross-refs to the Python docs so I've removed it

docs/conf.py
seanh: Removed

PR #4426, opened 2017-03-13
Add link to client's own docs

seanh:
Clarify that h.readthedocs.io is the docs site for the web service only,
and provide a link to the client's own docs now that they're separate.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -6,7 +6,7 @@ that serves most of the https://hypothes.is/ website, including the web
 annotations API at https://hypothes.is/api/.
 The `Hypothesis client <https://github.com/hypothesis/client>`_
 is a browser-based annotator that is a client for the web service's API, see
-`the client's own documentation site <http://h.readthedocs.io/projects/client/>`_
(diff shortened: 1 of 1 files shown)


docs/index.rst
```diff
@@ -1,9 +1,16 @@
-Hypothesis
-==========
+Welcome to Hypothesis Web Service’s documentation!
```
fatbusinessman: This is a bit pedantic, I realise, but you refer here to “Hypothesis Web Service”, rather than “the Hypothesis web service” below (with the added “the” and the lack of capitals). I’d recommend standardising on one: my personal preference would be for the latter (for consistency with “the Hypothesis client” elsewhere, for one thing).

seanh: Oh, yeah, that's a typo

seanh: The capitals are because this one's a title though, but _the_ shouldn't be missing

seanh: And why is documentation not capitalised good lord

seanh: Corrected now

PR #4430, opened 2017-03-13
Rename Sphinx project to "Hypothesis Web Service"

seanh:
In the top left of the h docs (top of the sidebar) it says "The Hypothesis Annotation Framework" (also in other places, like the title of the PDF docs).

Rename the Sphinx project from "The Hypothesis Annotation Framework
Documentation" to "The Hypothesis Web Service Documentation". This is
the documentation for the web service only, the client (and browser
extensions etc) have their own separate documentations. And I'm not sure
there was ever really an annotation framework.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 # pylint: disable=invalid-name
 #
-# The Hypothesis Annotation Framework documentation build configuration file, created by
+# The h documentation build configuration file, created by
 # sphinx-quickstart on Fri Oct 12 19:21:42 2012.
 #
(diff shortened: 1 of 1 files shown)


fatbusinessman:
Given the discussion over in #4426, should we still be capitalising all these instances of “Web Service”?

seanh:
I _believe_ these ones are all titles (title in the sidebar, title of the man page, title of the PDF...) so they should be capitalised. There is one Python comment in the `conf.py` file that probably has the wrong capitalisation but I'm ignoring that.

seanh:
I'm gonna change this to just "h" as per https://github.com/hypothesis/h/pull/4434...

nickstenning:
LGTM.

PR #4431, opened 2017-03-13
Flag endpoint

chdorner:
This is the implementation of hypothesis/product-backlog#206.

This adds a new endpoint at `POST /api/flags/annotations/{id}` which will create a flag for the authenticated user and the given annotation id.

### DIFF
--- a/h/views/api_flags.py
+++ b/h/views/api_flags.py
@@ -1,0 +1,43 @@
+# -*- coding: utf-8 -*-
+
+from __future__ import unicode_literals
+
+from pyramid import security
+from pyramid.httpexceptions import HTTPNoContent, HTTPNotFound
+
+from memex.views import api_config
+
+from h.schemas import ValidationError
+
+
+@api_config(route_name='api.flags',
+            request_method='POST',
+            link_name='flag.create',
+            description='Flag an annotation for review.',
+            effective_principals=security.Authenticated)
+def create(context, request):
+    annotation = _fetch_annotation(context, request)
+    svc = request.find_service(name='flag')
+    svc.create(request.authenticated_user, annotation)
+    return HTTPNoContent()
+
+
+def _fetch_annotation(context, request):
+    try:
+        annotation_id = request.json_body.get('annotation')
+
+        if not annotation_id:
+            raise ValueError()
+    except ValueError:
+        raise ValidationError('annotation id is required')
+
+    not_found_msg = 'cannot find annotation with id %s' % annotation_id
+
+    try:
+        resource = context[annotation_id]
+        if not request.has_permission('read', resource):
+            raise HTTPNotFound(not_found_msg)
+
+        return resource.annotation
+    except KeyError:
+        raise HTTPNotFound(not_found_msg)
--- a/tests/common/factories.py
+++ b/tests/common/factories.py
@@ -143,3 +143,13 @@ class Setting(ModelFactory):
 
     key = factory.LazyAttribute(lambda _: FAKER.domain_word())
     value = factory.LazyAttribute(lambda _: FAKER.catch_phrase())
+
+
+class Flag(ModelFactory):
+
+    class Meta:  # pylint: disable=no-init, old-style-class
+        model = models.Flag
+        force_flush = True
+
+    user = factory.SubFactory(User)
+    annotation = factory.SubFactory(Annotation)
--- a/tests/h/services/flag_test.py
+++ b/tests/h/services/flag_test.py
@@ -1,0 +1,41 @@
+# -*- coding: utf-8 -*-
+
+from __future__ import unicode_literals
+
+import pytest
+
+from h.services import flag
+from h._compat import xrange
+
+
+@pytest.mark.usefixtures('flags')
+class TestFlagServiceFlagged(object):
+    def test_it_returns_true_when_flag_exists(self, svc, flags):
+        user = flags[-1].user
+        annotation = flags[-1].annotation
+
+        assert svc.flagged(user, annotation) is True
+
+    def test_it_returns_false_when_flag_does_not_exist(self, svc, factories):
+        user = factories.User()
+        annotation = factories.Annotation(userid=user.userid)
+
+        assert svc.flagged(user, annotation) is False
+
+    @pytest.fixture
+    def flags(self, factories):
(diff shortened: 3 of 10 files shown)


tests/h/services/flag_test.py
```diff
@@ -0,0 +1,67 @@
...
+
+    def test_it_returns_false_when_flag_does_not_exist(self, svc, factories):
+        user = factories.User()
+        annotation = factories.Annotation(userid=user.userid)
+
+        assert svc.flagged(user, annotation) is False
+
+    @pytest.fixture
+    def flags(self, factories):
+        return [factories.Flag() for _ in xrange(3)]
```
chdorner: I fixed this by importing `from h._compat import xrange`, I'm guessing that Github/hound is not updating this comment because the fix is on another line.

fatbusinessman:
To re-raise a [point on the original spike](https://hypothes.is/a/Szb0hP3dEeaymfPLucwEOA) that we decided to discuss later:

What’re your thoughts on having this be a `POST` of, say, `{"id": "123456"}` to `/api/flags/annotations`, so that it‘s then `POST`ing a representation of a “flag” resource, and more closely mirrors what we’re expecting back from a `GET` request?

nickstenning:
FWIW I've had a quick look at this and have two comments:

1. Really nice, clean, well-documented implementation. Makes me really happy to see commits like this. A++++ would review again, etc. 😁 

2. Like @fatbusinessman, I'm not so sure about the the `.../annotations/<aid>` thing. In fact I'd probably go a bit further than he did above. Given the resource-orientation of the annotations API it would seem more consistent to me to treat "flags" themselves as resources which have some properties:

      ```
      POST /api/flags

      {"annotation": "Th47uAgdEeabdkfdiAthVg"}
      ```
   (I'd suggest not using `"id"` as we may wish to expose a flag ID for removing flags in the future.)

   Notably I also think specifying the type of flagged object in the URL isn't necessary at this stage. If we need to flag additional kinds of domain object in the future we can still do that through a single "flags" API, but by specifying additional data in the flag payload. (e.g. `{"user": "acct:chdorner@hypothes.is"}` or whatever).

h/views/api_flags.py
```diff
@@ -0,0 +1,20 @@
...
+
+from pyramid import security
+from pyramid.httpexceptions import HTTPNoContent
+
+from memex.views import api_config
+
+
+@api_config(route_name='api.flag_annotation',
+            request_method='POST',
+            link_name='api.flag_annotation',
```
nickstenning: What's this for?

nickstenning: Ahhh, sorry, [answered my own question](https://github.com/hypothesis/h/blob/9a0a77f92b4f17253ce3b659a59460bb1db8355d/src/memex/views.py#L69-L85). I missed this change.

h/views/api_flags.py
```diff
@@ -0,0 +1,41 @@
...
+
+from memex.resources import AnnotationResourceFactory
+from memex.views import api_config
+
+from h.schemas import ValidationError
+
+
+@api_config(route_name='api.flags',
+            request_method='POST',
+            link_name='flags.create',
```
nickstenning: I think `flags` here should be singular to be consistent?

h/views/api_flags.py
```diff
@@ -0,0 +1,41 @@
...
+            effective_principals=security.Authenticated)
+def create(request):
+    annotation = _fetch_annotation(request)
+    svc = request.find_service(name='flag')
+    svc.create(request.authenticated_user, annotation)
+    return HTTPNoContent()
+
+
+def _fetch_annotation(request):
+    annotation_id = request.json_body.get('annotation')
```
nickstenning: I think this needs to handle the case that accessing `request.json_body` throws a `ValueError` if the body isn't JSON.

h/views/api_flags.py
```diff
@@ -0,0 +1,41 @@
...
+    if not annotation_id:
+        raise ValidationError('annotation id is required')
+
+    not_found_msg = 'cannot find annotation with id %s' % annotation_id
+
+    try:
+        factory = AnnotationResourceFactory(request)
```
nickstenning: Rather than doing this here, I think it might be cleaner to leave the route with `factory='memex.resources. AnnotationResourceFactory'` but just omit the `traverse=` argument. Then in the view you can get the annotation as `resource = context[annotation_id]`.

tests/common/factories.py
```diff
@@ -143,3 +143,13 @@ class Meta:  # pylint: disable=no-init, old-style-class
 
     key = factory.LazyAttribute(lambda _: FAKER.domain_word())
     value = factory.LazyAttribute(lambda _: FAKER.catch_phrase())
+
+
+class Flag(ModelFactory):
+
+    class Meta:  # pylint: disable=no-init, old-style-class
+        model = models.Flag
+        force_flush = True
```
nickstenning: Is there a good reason to have `force_flush = True` here?

tests/h/services/flag_test.py
nickstenning: Probably, yes. TBH just using `range` here would be more than fine.

fatbusinessman:
> I also wonder slightly whether we should be a bit more lenient about input validation here. If someone tries to flag an annotation that's just been deleted (for example) I don't know that the client should show an error. Could we just return "204 No Content" regardless of whether we successfully created the flag?

Does it necessarily follow that, if the API returns a `4xx` code, the client _has_ to show an error? Or can the API return some flavour of `4xx` response – because it hasn’t successfully completed the request – and let the client interpret that response and decide whether it’s worth displaying to the user?

Always returning a `2xx` response, whether or not we fulfil the request, feels like it could silence genuinely harmful errors and make them harder to diagnose.

nickstenning:
No, it doesn't, and that's a fine way to proceed. We just need to make sure that we follow through on that when the flagging feature makes it into the client. :+1:

tests/common/factories.py
chdorner: Not really, we seem to have a bit of a mess of how we handle model factories. Most of the model factories set `force_flush`, except for `User` and `AuthTicket`. So I went with the majority and set it.

chdorner:
This should be ready again 👍

nickstenning:
LGTM.

PR #4432, opened 2017-03-14
Reintegrate memex (Pt 1)

nickstenning:
This PR makes a start on the work described at https://notes.wtk.io/2017/03/08/reintegrating-memex.

I think it should be possible to do this piece by piece thanks to the relatively loose coupling that Pyramid affords.


nickstenning:
Done a little more testing here and it seems there's something funky going on with the token endpoint. Marking as WIP.

nickstenning:
Fixed by ensuring that the API root URL is `/api/` rather than `/api`, to match past behaviour.

nickstenning:
Oh joy, more problems.

nickstenning:
Yeah, I actually did call it `api_annotations.py` initially, but then I realised that that file also contains the API index view and the search view. So I was working on the basis that we could split it apart later. Sound ok?

chdorner:
> So I was working on the basis that we could split it apart later. Sound ok?

Depending on when you were planning to do this maybe add a FIXME to the file?

PR #4435, opened 2017-03-14
Remove more instances of "web service" from docs

seanh:
We're just calling it "h" not "the Hypothesis web service".


nickstenning:
:+1:

PR #4436, opened 2017-03-14
Generate an email notification when an annotation is flagged and send…

sheetaluk:
… to the group's administrator.

Fixes https://github.com/hypothesis/product-backlog/issues/180

### DIFF
--- a/h/templates/emails/flag_notification.txt.jinja2
+++ b/h/templates/emails/flag_notification.txt.jinja2
@@ -1,0 +1,1 @@
+An annotation  has been flagged. Please visit {{ incontext_link }} for more details.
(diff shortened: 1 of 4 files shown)


seanh:
This looks perfect to me. There's no code that calls this code in production yet, but there will be soon, and I don't see any reason why this can't merge now

h/templates/emails/flag_notification.txt.jinja2
```diff
@@ -0,0 +1 @@
+An annotation  has been flagged. Please visit {{ incontext_link }} for more details.
```
robertknight: Tiny typo. Looks like two spaces instead of one after "annotation" @sheetaluk . Same with the HTML template - although that won't make any difference to the rendered output.

PR #4437, opened 2017-03-14
Move memex.presenters into h.presenters

nickstenning:
Move the various Annotation and Document presenter classes into the newly-created `h.presenters` package. No presenter code has changed here, so the apparent size of the diff is a little misleading. The presenters have been split up into individual modules so that there is one presenter class defined by each module in `h.presenters`.

To assist making this move the following things have had to change at the same time:

- the 'json' and 'jsonld_id' link definitions have moved into `h.links`
- the 'utc_iso8601' helper function has moved into `h.util.datetime`
- Jinja2 has been added to the list of memex dependencies. This is because I've had to introduce a reverse dependency from memex on the `h.presenters` module, which now includes HTML presenters, and so we need Jinja2 in the memex test environment.

_*N.B.* This depends on the previous PRs in the series (#4432, #4433) and so will need a rebase after those merge._


tests/h/presenters/annotation_jsonld_test.py
```diff
@@ -0,0 +1,195 @@
...
+        presenter = AnnotationJSONLDPresenter(resource)
+
+        assert presenter.id == 'http://fake-link/jsonld_id'
+
+    def test_id_passes_annotation_to_link_service(self, group_service, fake_links_service):
+        annotation = mock.Mock(id='foobar')
+        resource = AnnotationResource(annotation, group_service, fake_links_service)
+
+        presenter = AnnotationJSONLDPresenter(resource)
+        _ = presenter.id
```
nickstenning: Thanks, Hound. Thound.

PR #4439, opened 2017-03-16
Add flag index endpoint

chdorner:
Implements hypothesis/product-backlog#207.

This adds the `GET /api/flags` endpoint which can optionally be filtered with one or more `uri` parameters, or with the `group` filter.
The `uri` parameter will take the document equivalence into account.

### DIFF
--- a/h/services/flag.py
+++ b/h/services/flag.py
@@ -50,6 +50,19 @@ class FlagService(object):
                            annotation=annotation)
         self.session.add(flag)
 
+    def list(self, user):
+        """
+        Return a list of flags made by the given user.
+
+        :param user: The user to filter flags on.
+        :type user: h.models.User
+
+        :returns: list of flags (``h.models.Flag``)
+        :rtype: list
+        """
+
+        return self.session.query(models.Flag).filter_by(user=user)
+
 
 def flag_service_factory(context, request):
     return FlagService(request.db)
--- a/h/services/flag.py
+++ b/h/services/flag.py
@@ -2,7 +2,10 @@
 
 from __future__ import unicode_literals
 
+from memex import uri
+
 from h import models
+from h import storage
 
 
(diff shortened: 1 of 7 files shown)


h/views/api_flags.py
```diff
@@ -22,6 +22,23 @@ def create(context, request):
     return HTTPNoContent()
 
 
+@api_config(route_name='api.flags',
+            request_method='GET',
+            link_name='flag.index',
+            description='List a users flagged annotations for review.',
+            effective_principals=security.Authenticated)
```
nickstenning: Might it be friendly to have a version of this endpoint for logged-out users which just returns the empty list? At the moment `GET /api/flags` without authentication will return a 404.

chdorner: I'd rather stay consistent with the rest of our API and respond with a 404 instead of providing a dummy response.

h/services/flag.py
```diff
@@ -50,6 +52,44 @@ def create(self, user, annotation):
...
+
+        if group is not None:
+            joined_annotation = True
+            query = query.join(models.Annotation) \
+                         .filter(models.Annotation.groupid == group)
+
+        if uris:
+            expanded = []
+            for uri in uris:
+                expanded.extend(storage.expand_uri(self.session, uri))
```
chdorner: Fixed.

h/views/api_flags.py
nickstenning: Yep, ok. We can probably make this better by distinguishing between 403s that are ok to return to the user and 403s that should be returned as 404s in the API. Probably something that'll be easier to do after we've reintegrated memex.

chdorner: I agree. In cases where it's okay to reveal the existence of a resource, we should return a 403 instead of a 404. We can do that change API-wide though.

PR #4440, opened 2017-03-17
Add a feature flag for filtering highlights based on visible annotations in sidebar

robertknight:
We are investigating changing the visible highlights in an annotated document based on the current filter settings in the sidebar. Add a feature flag so we can test out behavior and performance impact with a select group of users.

See https://github.com/hypothesis/client/issues/217 for more context.

CC @greebie


nickstenning:
LGTM.

PR #4441, opened 2017-03-17
Replace inline script for homepage Via form

### LINKED ISSUE #4438: Inline script on homepage conflicts with strict CSP policy
### Steps to reproduce

Go to https://hypothes.is , enter a URL in the form field next to the "Annotate!" button and click "Annotate!"

### Expected behaviour
No CSP reports.

### Actual behaviour
CSP error is reported in the console due to an inline `onsubmit` event handler.

nickstenning:
The form on the homepage which sends people to Via used an inline event handler which violates our Content Security Policy.

Replace the onsubmit handler with code provided in legacy-site.js.

This commit also:

- removes an apparently unused data- attribute in the HTML
- renames the input from "search" to "url"
- finds the input through the form rather than by document ID
- sends users to the correct Via URL, saving them a redirect

Fixes #4438.

### DIFF
--- a/h/static/scripts/legacy-site.js
+++ b/h/static/scripts/legacy-site.js
@@ -30,3 +30,14 @@ if (chromeextInstaller) {
     event.preventDefault();
   });
 }
+
+const viaForm = document.querySelector('.js-via-form');
+if (viaForm) {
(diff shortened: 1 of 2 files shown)


robertknight:
LGTM

PR #4442, opened 2017-03-20
Refactor session and WorldGroup tests

chdorner:
This just refactors the tests in `tests/h/groups/util_test.py` and `tests/h/session_test.py` to use test classes.

The next pull request will add some more tests to both of these classes, so using test classes will make them a bit more readable. The tests themselves haven't actually changed, I just moved them around, sometimes changed the test name, and obviously added the `self` parameter to all of them.

### DIFF
--- a/tests/h/session_test.py
+++ b/tests/h/session_test.py
@@ -12,223 +12,213 @@ class FakeGroup(object):
         self.is_public = is_public
 
 
-def test_model_sorts_groups(authenticated_request):
-    authenticated_request.set_groups([
-        FakeGroup('c', 'Group A'),
-        FakeGroup('b', 'Group B'),
-        FakeGroup('a', 'Group B'),
-    ])
-    session_model = session.model(authenticated_request)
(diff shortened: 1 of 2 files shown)


tests/h/session_test.py
```diff
@@ -12,223 +12,213 @@ def __init__(self, pubid, name, is_public=False):
...
+        ])
+        session_model = session.model(authenticated_request)
 
-    ids = [group['id'] for group in session_model['groups']]
+        ids = [group['id'] for group in session_model['groups']]
 
-    assert ids == ['__world__', 'c', 'a', 'b']
+        assert ids == ['__world__', 'c', 'a', 'b']
 
+    def test_world_group_is_included(self, authenticated_request):
```
fatbusinessman: What was the reason for changing this from `is_public` to `is_included`? Because the test still checks whether the group is public.

chdorner: Ah that was a typo. Fixed.

PR #4444, opened 2017-03-21
Expose moderation role in `/api/profile`

chdorner:
We decided that we will have two separate endpoints for getting the flags status for a given URL and group, one for normal group users, and one for moderators. Because of this, the client will need to know if the authenticated user has moderation abilities for a given group, we can return this flag in the response of `/api/profile`.

It will look similar to this:
```json
{
  "preferences": {},
  "userid": "acct:chdorner@localhost",
  "groups": [
    {
      "public": true,
      "name": "Public",
      "id": "__world__"
    },
    {
      "url": "https://localhost:5000/groups/redacted/chdorner-testing",
      "public": false,
      "is_moderator": true,
      "name": "chdorner-testing",
      "id": "redacted"
    },
    {
      "url": "https://localhost:5000/groups/redacted/hypothesis-reading",
      "public": false,
      "name": "Hypothesis Reading",
      "id": "redacted"
    }
  ],
  "authority": "localhost",
  "features": {
    "total_shared_annotations": true,
    "defer_realtime_updates": true,
    "orphans_tab": true,
    "filter_highlights": false
  }
}
```

Note that the `is_moderator` key only exists when the user is a moderator, otherwise it will be missing.

### DIFF
--- a/h/groups/util.py
+++ b/h/groups/util.py
@@ -33,3 +33,7 @@ class WorldGroup(object):
     @property
     def is_public(self):
         return True
+
+    @property
+    def creator_id(self):
+        return None
--- a/h/session.py
+++ b/h/session.py
@@ -68,7 +68,7 @@ def _current_groups(request, authority):
 
     groups = authority_groups + _user_groups(user)
 
-    return [_group_model(request.route_url, group) for group in groups]
+    return [_group_model(request.route_url, group, user) for group in groups]
 
 
 def _user_groups(user):
@@ -78,8 +78,12 @@ def _user_groups(user):
         return sorted(user.groups, key=_group_sort_key)
 
 
-def _group_model(route_url, group):
-    model = {'name': group.name, 'id': group.pubid, 'public': group.is_public}
+def _group_model(route_url, group, user):
+    is_moderator = user is not None and user.id == group.creator_id
+    model = {'name': group.name,
+             'id': group.pubid,
+             'public': group.is_public,
+             'is_moderator': is_moderator}
 
     # We currently want to show URLs for secret groups, but not for publisher
     # groups, and not for the `__world__` group (where it doesn't make sense).
@@ -89,6 +93,7 @@ def _group_model(route_url, group):
         model['url'] = route_url('group_read',
                                  pubid=group.pubid,
                                  slug=group.slug)
+
     return model
 
 
--- a/tests/h/groups/util_test.py
+++ b/tests/h/groups/util_test.py
@@ -26,6 +26,9 @@ class TestWorldGroup(object):
     def test_is_public(self, group):
         assert group.is_public
 
+    def test_creator_id(self, group):
+        assert group.creator_id is None
+
     @pytest.fixture
     def group(self):
         return util.WorldGroup('example.com')
--- a/tests/h/session_test.py
+++ b/tests/h/session_test.py
@@ -5,11 +5,12 @@ from h import session
 
 
 class FakeGroup(object):
-    def __init__(self, pubid, name, is_public=False):
+    def __init__(self, pubid, name, is_public=False, creator_id=None):
         self.pubid = pubid
         self.name = name
         self.slug = pubid
         self.is_public = is_public
+        self.creator_id = creator_id
 
 
 class TestModel(object):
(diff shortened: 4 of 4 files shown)


chdorner:
Also note that the `is_moderator` flag will be included for any group that the current user is a creator, no matter if it's a private group under the `hypothes.is` authority, or the third-party publisher group.

h/session.py
```diff
@@ -89,6 +89,10 @@ def _group_model(route_url, group):
         model['url'] = route_url('group_read',
                                  pubid=group.pubid,
                                  slug=group.slug)
+
+    if user is not None and user.id == group.creator_id:
```
nickstenning: So this does two things that I think we'd like to avoid: it links the concept of a group creator with the concept of a group moderator outside the group model, and it makes explicit reference to primary keys present on the model.

I think it might be equally convenient and more future-friendly to have the Group model expose a `moderators` property which returns a list of users. For now, the implementation can be as simple as:

```@property
def moderators(self):
    return [self.creator]
```

And then the check here would be:

```if user is not None and user in group.moderators:```

Which would a) make the intent much clearer, and b) extend straightforwardly to groups having 0 or multiple moderators.

File: h/session.py [hunk 2]
```diff
@@ -89,6 +89,10 @@ def _group_model(route_url, group):
...
+        model['is_moderator'] = True
```
nickstenning: I think given that the `public` property is currently a boolean, I'd lean towards calling this `moderator` rather than `is_moderator`.

nickstenning: I'd also like to hear some input from @robertknight, @sean-roberts, or @sheetaluk on omitting this field if the user isn't a moderator of the group. My inclination would be to return `"moderator": false` in those cases, as it simplifies the work of any API client.

h/groups/util.py
```diff
@@ -33,3 +33,7 @@ def pubid(self):
     @property
     def is_public(self):
         return True
+
+    @property
+    def creator_id(self):
+        return None
```
nickstenning: Adding stuff like this here is fine for now -- we can merge this PR without addressing this comment -- but I really don't think it's going to be hard to move the `__public__` group into the database now we have an explicit `hypothesis init` command. So perhaps you could do that as a followup so we don't have to keep implementing functionality here as well?

h/session.py
robertknight: > My inclination would be to return "moderator": false in those cases, as it simplifies the work of any API client.

👍 - It makes no practical difference in JS because we'll just test the falsyness of `group.moderator` but for languages that don't have such lax treatment of missing properties it will be nicer and conceptually its better because it simplifies the type of this property from `boolean|undefined` to just `boolean`.

h/session.py
chdorner: One of the things I wanted to avoid is having to fire off 1 SQL query per group, because the `creator` relationship is not loaded. Which plays into the same issue as your comment on `WorldGroup.creator_id`.
I agree that having a `Group.moderators` property is nicer, but is it really worth doing a SQL query for each of the groups on every request to this endpoint?

h/session.py
chdorner: That's fine with me. Fixed.

h/groups/util.py
chdorner: Putting `__world__` into the database is something that we should probably test quite a bit more than just as a small follow-up PR. I'd rather have it as a proper card in a sprint than something we just do on the side and then it'll turn into a moderately sized chunk of work.

h/session.py
nickstenning: Yeah, that's a good point.

Looking further up this file, I see this:

```python
     authority_groups = (request.find_service(name='authority_group')
                         .public_groups(authority=authority))
  
      groups = authority_groups + _user_groups(user)
```

Which to me seems to me to hint that this file is doing a bunch too much work already. Perhaps let's move the retrieval of the list of groups for this endpoint into its own service, which can optimise the query it uses for its own purposes:

```python
(session.query(models.Group)
     .options(joinedload(models.Group.creator).load_only("id"))
```

I'm guessing that would allow getting rid of the `authority_group` service?

h/groups/util.py
nickstenning: That's fine. Make a card for it. But we should definitely do it!

h/session.py
sean-roberts: 👍 I agree with Rob's feedback here. And add that, imo, apis that have the "it's false/not supported because it is missing" fields are harder to pick up and start working with in general for lots of reasons.

That said, consistency on that end is probably the most important thing. So, this should set the precedent for all fields going forward and we should make sure that all existing fields do the same - though I think that is the case.

h/session.py
chdorner: Well, if I'm honest I think pretty much all of the code in `h/session.py` should live in a service that returns a data model, and the view should call a presenter that returns the JSON. Leaving only the `includeme` that sets up the session cookie.

I can start this process by creating a `SessionService` that for now just returns a list of groups for a given user, and in the future we can move more and more of the business logic into that service.

nickstenning: Agree completely, although I'm not sure "SessionService" makes a lot of sense -- it's not really got anything to do with the session these days.

nickstenning:
I guess on the basis of our current thinking this change won't be necessary? I'll close this. Feel free to reopen if I've got the wrong end of the stick.

PR #4446, opened 2017-03-24
Annotation formatters and exposing flag state

chdorner:
This introduces a new layer that is annotation formatters. They can be hooked into the rendering pipeline to allow a bit more code separation. For example, it doesn't make much sense to make the annotation rendering code know about all the details of flagging and nipsa.

### DIFF
--- a/h/formatters/annotation_flag.py
+++ b/h/formatters/annotation_flag.py
@@ -1,0 +1,64 @@
+# -*- coding: utf-8 -*-
+
+from __future__ import unicode_literals
+
+from zope.interface import implementer
+
+from h import models
+from h.formatters.interfaces import IAnnotationFormatter
+
+
+@implementer(IAnnotationFormatter)
+class AnnotationFlagFormatter(object):
+    """
+    Formatter for exposing a user's annotation flags.
+
+    If the passed-in user has flagged an annotation, this formatter will
+    add: `"flagged": true` to the payload, otherwise `"flagged": false`.
+    """
+
+    def __init__(self, session, authenticated_user=None):
+        self.session = session
+        self.authenticated_user = authenticated_user
+
+        # Local cache of flags. We don't need to care about detached
+        # instances because we only store the annotation id and a boolean flag.
+        self._cache = {}
+
+    def preload(self, ids):
+        if self.authenticated_user is None:
+            return
+
+        query = self.session.query(models.Flag) \
+                            .filter(models.Flag.annotation_id.in_(ids),
+                                    models.Flag.user == self.authenticated_user)
+
+        flags = {f.annotation_id: True for f in query}
+
+        # Set flags which have not been found explicitly to False to indicate
+        # that we already tried to load them.
+        missing_ids = set(ids) - set(flags.keys())
+        missing = {id_: False for id_ in missing_ids}
+        flags.update(missing)
+
+        self._cache.update(flags)
+        return flags
+
+    def format(self, annotation):
+        flagged = self._load(annotation.id)
+        return {'flagged': flagged}
+
+    def _load(self, id_):
+        if self.authenticated_user is None:
+            return False
+
+        if id_ in self._cache:
+            return self._cache[id_]
+
+        flag = self.session.query(models.Flag) \
+                           .filter_by(annotation_id=id_,
+                                      user=self.authenticated_user) \
+                           .one_or_none()
+
+        self._cache[id_] = (flag is not None)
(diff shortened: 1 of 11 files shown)


h/formatters/annotation_flag.py
```diff
@@ -0,0 +1,57 @@
...
+from __future__ import unicode_literals
+
+from zope.interface import implementer
+
+from h import models
+from h.formatters.interfaces import IAnnotationFormatter
+
+
+@implementer(IAnnotationFormatter)
+class AnnotationFlagFormatter(object):
```
fatbusinessman: I’d find a top-level docstring really useful here, particularly coming back to this in the future. I had to look through the source code to confirm that `AnnotationFlagFormatter` means “formatter for annotations that includes information about flags”.

h/formatters/annotation_flag.py
```diff
@@ -0,0 +1,57 @@
...
+    def __init__(self, session, authenticated_user=None):
+        self.session = session
+        self.authenticated_user = authenticated_user
+
+        # Local cache of flags. We don't need to care about detached
+        # instances because we only store the annotatino id and a boolean flag.
```
fatbusinessman: Typo: “annotatino”

h/formatters/annotation_flag.py
```diff
@@ -0,0 +1,57 @@
...
+        if self.authenticated_user is None:
+            return
+
+        query = self.session.query(models.Flag) \
+                            .filter(models.Flag.annotation_id.in_(ids),
+                                    models.Flag.user == self.authenticated_user)
+
+        flags = {f.annotation_id: True for f in query}
+
+        # Set flags which have not been found explicitely to False to indicate
```
fatbusinessman: Typo: “explicitely”

h/formatters/interfaces.py
```diff
@@ -0,0 +1,38 @@
+# -*- coding: utf-8 -*-
+
+from __future__ import unicode_literals
+
+from zope.interface import Interface
+
+
+class IAnnotationFormatter(Interface):
```
fatbusinessman: I think a top-level comment might be helpful here, explaining what the interface is for as a whole and how it should work. That might be a good place to put the bits of documentation about how the two methods should work together, rather than documenting it separately in each method.

h/services/annotation_json_presentation.py
```diff
@@ -0,0 +1,61 @@
...
+
+        # preload formatters, so they can optimize database access
+        for formatter in self.formatters:
+            formatter.preload(annotation_ids)
+
+        return [self.present(
+                    resources.AnnotationResource(ann, self.group_svc, self.links_svc))
+                for ann in annotations]
+
+    def get_presenter(self, annotation_resource):
```
chdorner: My thinking is that this method could be useful for other use cases, as the canonical way of getting a presenter with the appropriate formatters applied. I'm not sure if we need to make everything private that isn't currently called from the outside. What do you think?

h/formatters/annotation_flag.py
chdorner: Added.

h/formatters/annotation_flag.py
chdorner: Fixed.

h/formatters/annotation_flag.py
chdorner: Fixed.

h/formatters/interfaces.py
chdorner: Done.

fatbusinessman: _Much_ nicer. 👍 🍰

h/services/annotation_json_presentation.py
fatbusinessman: I guess there could be other use cases here, but I’m not quite sure what they would be, since we’ve covered the cases for “present a single thing” and “present a whole bunch of things”. To me, the fact that we can instantiate a single presenter object for an annotation is an implementation detail, and one that we could quite conceivably end up swapping out at a later date. From the perspective of what’s currently calling the service, the service *is* what’s doing the presenting.

In most places, I’d agree that we can get away without a hard split between private and public methods. In this case, though, it’s a service, so I think there’s more benefit to making it extra clear what’s in the service’s interface and what isn’t.

chdorner: I'm not sure I agree, but it's not worth holding this up for another feedback cycle.  	e3a8360 fixes this.

PR #4447, opened 2017-03-24
Summarise our current API design approach

fatbusinessman:
Following on from a discussion on Wednesday (2017-03-22), we appear to have settled on an approach for how we structure our current API around resources, particular around nesting them under one another.


fatbusinessman:
I still need to smush the commits around, but I think this is ready for a lower-level review now.

PR #4449, opened 2017-03-27
Move admin views into h.views

nickstenning:
This moves the last large chunk of views into the `h.views` package. This simplifies the initialisation and registration of views, and makes it easier to find views within the codebase.


tests/h/views/admin_admins_test.py
```diff
@@ -3,23 +3,23 @@
 from __future__ import unicode_literals
 
 import mock
-from pyramid import httpexceptions
 import pytest
+from pyramid import httpexceptions
```
fatbusinessman: What’s your sorting algorithm here? Is is that all `from foo import bar` statements come after all `import baz` statements?

nickstenning: Yes, although in practice my sorting algorithm is "whatever [isort](https://github.com/timothycrosley/isort) says".

nickstenning:
> was there a specific reason you chose modules prefixed with admin_, rather than nesting everything under a views.admin package?

Only really the converse: there's very little gained other by nesting these in another package, and we already have a flat structure for, say, `api_*` views.

PR #4450, opened 2017-03-27
Make the test for flag URL filters ignore order

fatbusinessman:
Because the service does not guarantee to return the flags in any specific order, this seems to cause occasional test failures when Postgres returns the flags in the "reverse" order.

There is a custom `unordered_list` matcher defined, which I believe would work equally well, but in this case I converted the lists to `set`s, for consistency with another test in the same class.


nickstenning:
LGTM.

PR #4451, opened 2017-03-27
Send mail to group admin when group reader flags an annotation.

sheetaluk:
https://github.com/hypothesis/product-backlog/issues/180

### DIFF
--- a/h/views/api_flags.py
+++ b/h/views/api_flags.py
@@ -6,6 +6,10 @@ from pyramid import security
 from pyramid.httpexceptions import HTTPNoContent
 
 from h.views.api import api_config
+from h.emails import flag_notification
+from h import links
+from h.interfaces import IGroupService
+from h.tasks import mailer
 
 
 @api_config(route_name='api.annotation_flag',
@@ -17,4 +21,20 @@ from h.views.api import api_config
 def create(context, request):
     svc = request.find_service(name='flag')
     svc.create(request.user, context.annotation)
+
+    _email_group_admin(request, context.annotation)
+
     return HTTPNoContent()
+
+
+def _email_group_admin(request, annotation):
+    group_service = request.find_service(IGroupService)
+    group = group_service.find(annotation.groupid)
+
+    incontext_link = links.incontext_link(request, annotation)
+    if incontext_link is None:
+        incontext_link = annotation.target_uri
+
+    if group.creator is not None:
+        send_params = flag_notification.generate(request, group.creator.email, incontext_link)
+        mailer.send.delay(*send_params)
--- a/tests/h/groups/util_test.py
+++ b/tests/h/groups/util_test.py
@@ -26,6 +26,9 @@ class TestWorldGroup(object):
     def test_is_public(self, group):
         assert group.is_public
 
+    def test_creator(self, group):
+        assert group.creator is None
+
     @pytest.fixture
     def group(self):
         return util.WorldGroup('example.com')
--- a/tests/h/views/api_flags_test.py
+++ b/tests/h/views/api_flags_test.py
@@ -10,7 +10,7 @@ from pyramid.httpexceptions import HTTPNoContent
 from h.views import api_flags as views
 
 
-@pytest.mark.usefixtures('flag_service')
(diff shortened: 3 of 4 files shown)


h/views/api_flags.py
```diff
@@ -16,8 +20,18 @@
...
     svc = request.find_service(name='flag')
     svc.create(request.authenticated_user, annotation)
+
+    group_service = request.find_service(IGroupService)
+    group = group_service.find(annotation.groupid)
+
+    incontext_link = links.incontext_link(request, annotation)
+
+    send_params = flag_notification.generate(request, group.creator.email, incontext_link)
+    mailer.send.delay(*send_params)
```
chdorner: I think it might be nice to move the new code here into a private method to better structure the code, the tests can still be the same as we don't test private functions individually. This method should probably just get the request and the annotation, it can then fetch the group, the in-context link and send off the email.

tests/h/views/api_flags_test.py
```diff
@@ -86,6 +86,31 @@ def fake_has_permission(action, _):
 
         assert 'cannot find annotation' in exc.value.message
 
+    def test_passes_info_to_flag_notification_email(self,
+                                                    pyramid_request,
+                                                    group_service,
+                                                    flag_notification_email,
+                                                    incontext_link):
+        pyramid_request.json_body = {'annotation': annotation.id}
+        context = FakeAnnotationResourceFactory(annotation)
```
chdorner: Hmm, I'm guessing the `annotation` parameter was missing from this method, and now hound doesn't resolve this comment, right?

h/views/api_flags.py
```diff
@@ -16,8 +20,18 @@
...
+
     svc = request.find_service(name='flag')
     svc.create(request.authenticated_user, annotation)
+
+    group_service = request.find_service(IGroupService)
+    group = group_service.find(annotation.groupid)
+
+    incontext_link = links.incontext_link(request, annotation)
+
+    send_params = flag_notification.generate(request, group.creator.email, incontext_link)
```
chdorner: In case the annotation was made in the `__world__` group then the group service returns a [`WorldGroup`](https://github.com/hypothesis/h/blob/master/h/groups/util.py#L8) instance, which doesn't have a creator (it doesn't even have a `creator` property at the moment), which will then produce an exception here.

For now, maybe you can add a `creator` property in the `WorldGroup` class which just returns `None`, and then here check if `group.creator is not None`, and otherwise don't send an email. We really need to move the `__world__` group into the database to avoid these things, but that's needs a bit of testing work, which is why it hasn't been done yet.

tests/h/views/api_flags_test.py
sheetaluk: Yes. I briefly considered deleting the comment but didn't :)

tests/h/groups/util_test.py
```diff
@@ -25,6 +25,10 @@ def test_pubid(self, group):
 
     def test_is_public(self, group):
         assert group.is_public
+        assert group.pubid == '__world__'
+
+    def test_creator(self, group):
+        assert group.creator == None
```
chdorner: This seems to work, but it should be `assert group.creator is None` to make Hound happy 🐶

chdorner:
This is looking great now. Just one thing which I @sheetaluk is already checking, which is that when there is no `BOUNCER_URL` set, the incontext link is `None` and the rendered email says: `... Please visit None for more.`.
It might make sense to provide a fallback otherwise this is good to go. I'm approving my review for now.

chdorner:
We quickly talked about this today and decided for now to just return the target uri of the annotation when the incontext link is empty.

tests/h/groups/util_test.py
```diff
@@ -25,6 +25,10 @@ def test_pubid(self, group):
 
     def test_is_public(self, group):
         assert group.is_public
+        assert group.pubid == '__world__'
```
chdorner: This seems to be a remnant from the merge conflict. This assertion already exists in the previous test method, so this can be safely deleted.

sheetaluk:
Merging after @chdorner 's pre-approval.

PR #4453, opened 2017-03-29
Remove now unused `GET /api/flags` endpoint

chdorner:
__This depends on #4446 being merged first and then needs a rebase.__

We decided to include the `"flagged": true/false` items in the
annotation bodies, so this endpoint is not needed anymore.


chdorner:
Rebased. @nickstenning you can take another look

nickstenning:
LGTM.

PR #4454, opened 2017-03-29
Move create annotation flag endpoint

chdorner:
__This depends on #4453 being merged first and then needs a rebase__

We decided to move the `POST /api/flags` endpoint to `PUT /api/annotations/<annid>flag`, which is what this PR is doing.

A PR for the client which is using this new endpoint is coming up shortly.


chdorner:
Rebased. @nickstenning you can take another look.

nickstenning:
LGTM. I think the choice of API convention is at least partly validated by how much nicer this code now is.

PR #4455, opened 2017-03-29
Add a caching helper for services to use

nickstenning:
While doing some refactoring unrelated to the content of this commit (converting the feature flag client into a service) I found myself implementing caching functionality and tests for it in Yet Another Service.

Rather than have each service that needs to cache stuff for the duration of the database transaction implement and test its own caching behaviour, it makes sense to introduce a general purpose tool to achieve the same ends.

This commit introduces a decorator, `lru_cache_in_transaction`, which is a thin wrapper around the `lru_cache` decorator from Python >= 3.3, with the added bonus that it clears its internal cache when the database transaction
ends. It converts both the GroupfinderService and the UserService to use this new helper.

Given the way that the caching helper works, I think it will be sufficiently hard to use incorrectly without breaking the core functionality of the service in question that I have removed the tests that check caching for each individual service.

### DIFF
--- a/tests/h/util/db_test.py
+++ b/tests/h/util/db_test.py
@@ -2,10 +2,59 @@
 
 from __future__ import unicode_literals
 
+import random
+
 import mock
 import pytest
 
-from h.util.db import on_transaction_end
+from h.util.db import lru_cache_in_transaction, on_transaction_end
+
+
+class TestLRUCacheInTransaction(object):
+
+    def test_caches_during_transaction(self, db_session, mock_transaction):
+        """
+        Return values should be cached during the transaction, and the cache
+        cleared when the transaction ends.
+        """
+        @lru_cache_in_transaction(db_session)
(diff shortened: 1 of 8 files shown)


nickstenning:
Worth noting that this replaces the caching behaviour in each of these places with a 128-item LRU cache instead of an unbounded dictionary, but I can't foresee that causing any problems. If reviewers feel that this is a risky change, then we can always add a `maxsize=None` parameter and switch back to an unbounded cache.

tests/h/util/db_test.py
```diff
@@ -2,10 +2,59 @@
...
+import random
+
 import mock
 import pytest
 
-from h.util.db import on_transaction_end
+from h.util.db import lru_cache_in_transaction, on_transaction_end
+
+
+class TestLRUCacheInTransaction(object):
```
chdorner: Using `random` make these tests pretty 🍠  👍

PR #4457, opened 2017-03-30
Convert feature flag client into a service

nickstenning:
Move the feature flag client from its own package (`h.features`) into a service (`h.services.feature`), simplifying it along the way.

The aims of this change are:

- to reduce database load by only loading feature flags on request paths that actually query them (we currently load feature flags preemptively on nearly every request)
- eliminate the possibility of feature flag code accessing stale ORM mapped objects after their associated database session has closed
- simplify the management of the feature flag cache outside of the request/response cycle (i.e. in worker processes)

In order to achieve these ends I've done the following:

1. The bulk of the functionality of `h.features.Client` has moved into a service which is responsible for lazily fetching feature flag data on demand and caching it for the duration of the database transaction. The service makes use of the new `lru_cache_in_transaction` helper to manage caching of the data.

2. The convenience of the `request.feature` property is split into a separate helper class, `FeatureRequestProperty`, which simplifies access to the service, which can now answer questions about feature flag status for any user.

3. As a result of the above changes, we no longer need to preload feature flags at the start of every request, and we no longer need to manually clear the feature flag cache at the end of each task handler in Celery.

_**N.B.** This depends on #4455 and will need a rebase after that has merged._


nickstenning:
By way of a demonstration of the caching behaviour:

```ipython
$ DEBUG_QUERY=1 hypothesis --dev shell

Environment:

In [1]: request.feature('orphans_tab')
2017-03-30 10:48:48,947 [30387] [sqlalchemy.engine.base.Engine:INFO] select version()
2017-03-30 10:48:48,947 [30387] [sqlalchemy.engine.base.Engine:INFO] {}
2017-03-30 10:48:48,950 [30387] [sqlalchemy.engine.base.Engine:INFO] select current_schema()
2017-03-30 10:48:48,950 [30387] [sqlalchemy.engine.base.Engine:INFO] {}
2017-03-30 10:48:48,952 [30387] [sqlalchemy.engine.base.Engine:INFO] SELECT CAST('test plain returns' AS VARCHAR(60)) AS anon_1
2017-03-30 10:48:48,953 [30387] [sqlalchemy.engine.base.Engine:INFO] {}
2017-03-30 10:48:48,954 [30387] [sqlalchemy.engine.base.Engine:INFO] SELECT CAST('test unicode returns' AS VARCHAR(60)) AS anon_1
2017-03-30 10:48:48,954 [30387] [sqlalchemy.engine.base.Engine:INFO] {}
2017-03-30 10:48:48,957 [30387] [sqlalchemy.engine.base.Engine:INFO] show standard_conforming_strings
2017-03-30 10:48:48,957 [30387] [sqlalchemy.engine.base.Engine:INFO] {}
2017-03-30 10:48:48,959 [30387] [sqlalchemy.engine.base.Engine:INFO] BEGIN (implicit)
2017-03-30 10:48:48,960 [30387] [sqlalchemy.engine.base.Engine:INFO] SELECT feature.id AS feature_id, feature.name AS feature_name, feature.everyone AS feature_everyone, feature.admins AS feature_admins, feature.staff AS feature_staff
FROM feature
2017-03-30 10:48:48,960 [30387] [sqlalchemy.engine.base.Engine:INFO] {}
Out[1]: True

In [2]: request.feature('orphans_tab')
Out[2]: True

In [3]: request.feature('orphans_tab')
Out[3]: True

In [4]: request.tm.commit()
2017-03-30 10:48:56,186 [30387] [sqlalchemy.engine.base.Engine:INFO] ROLLBACK

In [5]: request.feature('orphans_tab')
2017-03-30 10:48:57,247 [30387] [sqlalchemy.engine.base.Engine:INFO] BEGIN (implicit)
2017-03-30 10:48:57,248 [30387] [sqlalchemy.engine.base.Engine:INFO] SELECT feature.id AS feature_id, feature.name AS feature_name, feature.everyone AS feature_everyone, feature.admins AS feature_admins, feature.staff AS feature_staff
FROM feature
2017-03-30 10:48:57,248 [30387] [sqlalchemy.engine.base.Engine:INFO] {}
Out[5]: True

In [6]: request.feature('orphans_tab')
Out[6]: True
```

h/celery.py
```diff
@@ -98,12 +98,6 @@ def bootstrap_worker(sender, **kwargs):
 
 
 @signals.task_prerun.connect
-def reset_feature_flags(sender, **kwargs):
```
chdorner: Just to check. We don't need this anymore because the LRU cache clears automatically when a transaction ends and celery tasks run in their own transaction?

h/views/accounts.py
```diff
@@ -142,18 +142,12 @@ def _login_redirect(self):
         return self.request.params.get('next', _login_redirect_url(self.request))
 
     def _login(self, user):
-        # Clear any cached feature flags
-        self.request.feature.clear()
```
chdorner: Ah nice. I first thought that we still need those in case we would access the features before the `authenticated_user` is set. But you changed the service to evaluate the feature state on every call by caching the feature flags, and not the state of the flag for the current user 👍

h/services/feature.py
```diff
@@ -0,0 +1,124 @@
...
+    def all(self):
+        """Get the status of all feature flags for the current user."""
+        return self.svc.all(user=self.request.authenticated_user)
+
+
+class FeatureService(object):
+    """
+    Manages access to feature flag status.
+
+    This service is exposed on the request object as `request.feature` for
```
fatbusinessman: Is this part of the docstring supposed to be here, or does it predate you introducing the helper class?

h/celery.py
nickstenning: Exactly: we're covered by [these lines](https://github.com/hypothesis/h/blob/8da2ea3f4ede59a12b91d928c22357237a97bbd5/h/celery.py#L113-L122).

chdorner: Yeah, this is so much nicer now 👍

h/services/feature.py
nickstenning: Hmm, good point. That can probably go now. Will fix.

nickstenning: Done!

h/views/accounts.py
nickstenning: TBH I'm not even sure this was ever a good idea. For the entirety of the request that actually logs the user in, `request.authenticated_userid` is `None` (I think...), and so I'm not even clear that this worked as intended.

PR #4458, opened 2017-03-30
Add migration for annotation nipsa table.

chdorner:
This table is being used for hiding individual annotations from being
shown to other users than the creator, and the group moderators.

### DIFF
--- a/h/migrations/versions/50df3e6782aa_add_annotation_moderation_table.py
+++ b/h/migrations/versions/50df3e6782aa_add_annotation_moderation_table.py
@@ -1,0 +1,44 @@
+"""
+add annotation moderation table
+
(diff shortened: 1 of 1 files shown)


nickstenning:
LGTM. The only question I have is whether we want to continue calling this NIPSA as it evolves into a substantially more multifaceted feature? In particular I was wondering if wanted to call it `annotation_moderation` or something a bit less opaque?

Very happy to merge as-is and leave the question of name as a thing to think about in future.

chdorner:
We decided to think about the renaming of nipsa and decide on Monday on a new term. There's no rush for this migration to go out anyway, it's only a small part of a pretty big card.

nickstenning:
LGTM.

PR #4461, opened 2017-03-31
Add `PUT /api/annotations/<annid>/hide` endpoint

chdorner:
__Before merge:__
- [x] __The database migration in #4458 has to be deployed and run in production__
- [x] __this branch needs to be rebased__

This is part of hypothesis/product-backlog#239.

This is the endpoint the client will use when a group moderator decides to hide an annotation (nipsa).


h/services/annotation_moderation.py
```diff
@@ -0,0 +1,37 @@
...
+from h import models
+
+
+class AnnotationModerationService(object):
+    def __init__(self, session):
+        self.session = session
+
+    def create(self, annotation):
+        """
+        Create an annotation nipa flag.
```
nickstenning: This docstring needs an update. More generally, it might not be a bad idea to name the service's methods in a way that's more obvious for _consumers_ of the service, rather than just descriptive of what happens internally. Specifically, I'm thinking something like `.hide(annotation)`. What do you reckon?

h/services/annotation_moderation.py
```diff
@@ -0,0 +1,37 @@
...
+        is a no-op.
+
+        :param annotation: The annotation to hide from others.
+        :type annotation: h.models.Annotation
+        """
+
+        query = self.session.query(models.AnnotationModeration) \
+                            .filter_by(annotation=annotation)
+
+        if query.count() > 0:
```
fatbusinessman: Out of interest, is there a clean way to do this in a single query? I’m thinking of something akin to Postres’ [`ON CONFLICT DO NOTHING`](https://www.postgresql.org/docs/9.5/static/sql-insert.html#SQL-ON-CONFLICT) behaviour.

I’m assuming that Postgres’ transactions protect us from the race condition in here, so this is more to satisfy my own curiosity about what SQLAlchemy’s ORM can do.

chdorner: I'm actually pulling this out into a `hidden` method and improving it by using `self.session.query(exists().models.AnnotationNipsa.annotation_id == annotation_id)).scalar()`, which is better than `count()`.

chdorner: Do you know how the `ON CONFLICT DO NOTHING` works with SQLAlchemy?

nickstenning: I'm pretty sure there's no (built-in) way of using `ON CONFLICT` with sqla, I'm afraid.

nickstenning: But an `EXISTS` followed by an `INSERT` is more than fine for this rarely-used endpoint.

h/services/annotation_moderation.py
chdorner: Yup, I had the same idea went I went out for lunch. Fixed both.

PR #4471, opened 2017-04-04
Remove password hashing functionality from the User model

nickstenning:
This completes the work to move all password checking and updating
responsibilities to the UserPasswordService by:

- removing the custom setter for the `password` property on the User model
- updating the name of the underlying column property from `_password` to `password` and changing the UserPasswordService so that it uses the new name
- updating a couple of places in the functional tests that assumed they could set plaintext passwords using the User constructor

_**N.B.** This PR depends on #4468, #4469, and #4470, and will need rebasing after those have merged._

### DIFF
--- a/h/accounts/schemas.py
+++ b/h/accounts/schemas.py
@@ -12,7 +12,6 @@ from h.accounts import util
 from h.services.user import UserNotActivated
 from h.models.user import (
     EMAIL_MAX_LENGTH,
-    PASSWORD_MIN_LENGTH,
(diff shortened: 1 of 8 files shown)


h/accounts/schemas.py
```diff
@@ -22,6 +21,7 @@
 _ = i18n.TranslationString
 log = logging.getLogger(__name__)
 
+PASSWORD_MIN_LENGTH = 2  # FIXME: this is ridiculous
```
chdorner: 😂

PR #4472, opened 2017-04-04
Add additional validation of target selectors

nickstenning:
Ensure that incoming target selector properties are:

- an array
- of objects
- each of which has a type property
- which is a string

This should prevent a repeat of https://sentry.io/hypothesis/h/issues/239837148/.


chdorner:
🚀 ⌚️ 

LGTM. Feel free to merge when all the tests pass

PR #4473, opened 2017-04-04
Mark moderated annotations as nipsa'd in search index

chdorner:
This is part of hypothesis/product-backlog#239

For now, this is done the same way we already check if the user is
nipsa'd, and then add a `"nipsa": true` item to the search index
payload.

Ideally, we would also get rid of the `AnnotationTransformEvent` system
and build on top of the recently introduced formatters (or something
similar to that). But since the card for annotation moderation is
already quite big I've opted for extending the current system instead.

### DIFF
--- a/h/nipsa/subscribers.py
+++ b/h/nipsa/subscribers.py
@@ -4,7 +4,7 @@ from __future__ import unicode_literals
 
 
 def transform_annotation(event):
-    """Add a {"nipsa": True} field on moderated annotations or whose users are flagged."""
+    """Add a {"nipsa": True} field on moderated annotations or those whose users are flagged."""
     payload = event.annotation_dict
 
     nipsa = _user_nipsa(event.request, payload)
--- a/h/nipsa/subscribers.py
+++ b/h/nipsa/subscribers.py
@@ -1,9 +1,24 @@
 # -*- coding: utf-8 -*-
 
+from __future__ import unicode_literals
+
 
 def transform_annotation(event):
-    """Add a {"nipsa": True} field on annotations whose users are flagged."""
-    annotation = event.annotation_dict
-    nipsa_service = event.request.find_service(name='nipsa')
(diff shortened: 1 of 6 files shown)


h/nipsa/subscribers.py
```diff
@@ -1,9 +1,24 @@
 # -*- coding: utf-8 -*-
 
+from __future__ import unicode_literals
+
 
 def transform_annotation(event):
     """Add a {"nipsa": True} field on annotations whose users are flagged."""
```
nickstenning: Docstring here needs a quick update.

h/services/annotation_moderation.py
```diff
@@ -2,13 +2,29 @@
...
+        Check if an annotation id is hidden.
+
+        :param annotation_id: The id of the annotation to check.
+        :type annotation: unicode
+
+        :returns: true/false boolean
+        :rtype: bool
+        """
+        query = self.session.query(exists().where(
+            models.AnnotationModeration.annotation_id == annotation_id))
```
nickstenning: This clearly works, but TBH I find this form of the exists query quite hard to read. You can also use `Query.exists()`, which would save you an import and work like this:

```python
q = self.session.query(models.AnnotationModeration).filter_by(annotation_id=annotation_id)
return self.session.query(q.exists()).scalar()
```

I'm certainly not going to fuss over this one. I just personally find this form a little easier to read.

h/views/api_moderation.py
```diff
@@ -19,4 +20,8 @@ def create(context, request):
 
     svc = request.find_service(name='annotation_moderation')
     svc.hide(context.annotation)
+
+    event = events.AnnotationEvent(request, context.annotation.id, 'update')
```
nickstenning: Hmmm. Not hmmm at your code, but hmmm at needing to do this any time we update an annotation in a way that might require a reindex. I guess we probably need to find a better way of doing this at some point.

For now, it might be worth adding a short comment just to explain why this event needs to be fired.

PR #4474, opened 2017-04-04
Document CLIENT_URL

seanh:
And start a page at which more environment variables can also be
documented.

This allows any page of the docs to link to an env var like this:

```
See the :envvar:`CLIENT_URL` env var.
```

This will:

1. Render the envvar mention as a `<code class="std-envvar">`
   (among other CSS classes), this is then styled specially by the
   Sphinx theme

2. Automatically hyperlink it to the env var's definition on the
   env vars page

3. Generate a warning if an envvar with that name isn't defined
   (if the -n arg to sphinx-build is used)

I think this will also allow docs in the client repo to link to env vars
defined in the h repo docs in this way too, because we have intersphinx
set up (there's already a couple of CLIENT_URL references in the client
docs).

In the future we might be able to use Sphinx's autodoc extension to
generate this envvars page from docstrings in the code.

### DIFF
--- a/docs/developing/envvars.rst
+++ b/docs/developing/envvars.rst
@@ -1,0 +1,10 @@
+Environment Variables
+=====================
+
+This section documents the environment variables supported by h.
+
+.. envvar:: CLIENT_URL
+
+   The URL at which the Hypothesis client code is hosted.
+   This is the URL to the client entrypoint script, by default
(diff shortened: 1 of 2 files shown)


seanh:
Done, the default URL is now duplicated between code and docs, but I think the solution to that is going to be to get autodoc pulling the env vars, their docstrings, and their default value out of the source code.

PR #4475, opened 2017-04-04
Use flag service for annotation flag formatter

fatbusinessman:
Now that we have a service to query for flag state, we can add a method to it to list subsets of flagged annotations and simplify the `AnnotationFlagFormatter`, removing its knowledge of the session and the model logic.

To make it clear that the behaviour of the formatter hasn't changed, I've left the tests as they are, and introduced a test dependency on the `FlagService`. This might be something to simplify in a later commit, faking out the `FlagService`, or it may be worth leaving in place to function as an integration test.

### DIFF
--- a/h/formatters/annotation_flag.py
+++ b/h/formatters/annotation_flag.py
@@ -17,8 +17,8 @@ class AnnotationFlagFormatter(object):
     add: `"flagged": true` to the payload, otherwise `"flagged": false`.
     """
 
-    def __init__(self, session, user=None):
-        self.session = session
+    def __init__(self, flag_service, user=None):
+        self.flag_service = flag_service
         self.user = user
 
         # Local cache of flags. We don't need to care about detached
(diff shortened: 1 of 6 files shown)


h/formatters/annotation_flag.py
```diff
@@ -29,36 +29,26 @@ def preload(self, ids):
...
-
-        # Set flags which have not been found explicitly to False to indicate
-        # that we already tried to load them.
-        missing_ids = set(ids) - set(flags.keys())
-        missing = {id_: False for id_ in missing_ids}
-        flags.update(missing)
+        flagged_ids = self.flag_service.all_flagged(user=self.user,
+                                                    annotation_ids=ids)
 
+        flags = {id_: (id_ in flagged_ids) for id_ in ids}
```
chdorner: I'm not sure if I'm missing something with this diff, but I did a whole dance with figuring out missing ids and adding them as `False` to the internal cache so that `_load` doesn't try to load it again from the database. Is that gone now? Because that would multiply the number of SQL queries we would do, by a lot.

fatbusinessman: That’s not gone, no: this line does the same thing. It uses the `ids` parameter (so _all_ the ids, not just the flagged ones) for the keys of `flags`, and uses `(id_ in flagged_ids)` for the Boolean values.

```python
>>> ids = [1, 2, 3]
>>> flagged_ids = [2]
>>> {id_: (id_ in flagged_ids) for id_ in ids}
{1: False, 2: True, 3: False}
```

h/services/flag.py
```diff
@@ -28,6 +28,25 @@ def flagged(self, user, annotation):
...
+        :type user: h.models.User
+
+        :param annotation_ids: The IDs of the annotations to check.
+        :type annotation_ids: sequence of unicode
+
+        :returns The subset of the IDs that the given user has flagged.
+        :rtype set of unicode
+        """
+        query = self.session.query(models.Flag.annotation_id) \
+                            .filter(models.Flag.annotation_id.in_(annotation_ids),
```
nickstenning: A very small thing, but it's probably worth including a short-circuit condition along the lines of:

```python
if not annotation_ids:
    return []
```

This is because passing empty `IN` clauses to databases doesn't always behave the way you might expect:

```python
>>> print(session.query(models.Flag.annotation_id).filter(models.Flag.annotation_id.in_([])))
SELECT flag.annotation_id AS flag_annotation_id
FROM flag
WHERE flag.annotation_id != flag.annotation_id
```

Uh...

```sql
postgres=# EXPLAIN SELECT flag.annotation_id AS flag_annotation_id FROM flag WHERE flag.annotation_id != flag.annotation_id;
                       QUERY PLAN
---------------------------------------------------------
 Seq Scan on flag  (cost=0.00..24.50 rows=1154 width=16)
   Filter: (annotation_id <> annotation_id)
(2 rows)
```

Uhhhh...!!!

(Yeah, so avoid `.in_(might_be_empty)`).

chdorner: SQLAlchemy even prints warning messages when this happens.

PR #4476, opened 2017-04-04
Use factory_boy's `create_batch` method

fatbusinessman:
This seems to show the intent more clearly than list comprehensions, and also means we don't need to think about `range` versus `xrange`.

I left one instance in place in `tests/h/services/rename_user_test.py`, because that was also extracting the ids from the objects it created, so we were going to need a list comprehension anyway.


nickstenning:
Nice! I didn't know about that at all!

PR #4479, opened 2017-04-04
Accept ADR 003

fatbusinessman:
No-one raised any objections to this approach, so as long as someone agrees with this enough to merge it, I think this counts as accepted.


nickstenning:
Yyyyarp. :+1: 🍆

PR #4484, opened 2017-04-05
Upgrade to cryptography 1.8

nickstenning:
Upgrade to the latest cryptography, mainly to ensure that we're staying current with security-critical libraries.

In particular, we appear to have entirely missed that cryptography 1.5.3 contained a security fix for an issue with HKDF, which we use for key derivation (CVE-2016-9243).


nickstenning:
We can ignore the safety-ci status check here, as we can fix that in subsequent PRs.

PR #4487, opened 2017-04-05
Ignore pidfiles

nickstenning:
Instances in production are reporting their version numbers as ".dirty" because of nginx and supervisor pidfiles which are placed in the root of the repository.

Ignore "*.pid" so that pidfiles don't contribute to the version dirty state.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,8 +24,8 @@ node_modules/
 *.pyc
 *.pyo
 
+*.pid
 *.log
 
 celerybeat-schedule
 celerybeat-schedule.*
-celerybeat.pid


nickstenning:
By doing something like:

```
$ h-ssh prod h
[...]
[ec2-user@ip-10-8-6-119 ~]$ docker ps
CONTAINER ID        IMAGE               COMMAND                  CREATED             STATUS              PORTS               NAMES
dda28e81ee49        h                   "init-env supervisord"   36 minutes ago      Up 36 minutes                           grave_bhabha
[ec2-user@ip-10-8-6-119 ~]$ docker exec -it grave_bhabha git status --short
?? nginx.pid
?? supervisord.pid
```

PR #4491, opened 2017-04-05
Return hidden annotation for group creators

chdorner:
Group creators will see annotations that have been hidden in their
groups, we do this by adding another clause to the nipsa filter where we
check the group pubids that the current user created.

### DIFF
--- a/tests/h/nipsa/search_test.py
+++ b/tests/h/nipsa/search_test.py
@@ -1,11 +1,51 @@
 # -*- coding: utf-8 -*-
 
+import mock
+import pytest
+
 from h.nipsa import search
 
 
-def test_nipsa_filter_filters_out_nipsad_annotations():
+@pytest.mark.usefixtures('group_service')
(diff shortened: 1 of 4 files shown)


tests/h/nipsa/search_test.py
```diff
@@ -1,11 +1,51 @@
...
+        assert f({}) == nipsa_filter.return_value
+
+    def test_call_passes_group_service(self, pyramid_request, nipsa_filter, group_service):
+        f = search.Filter(pyramid_request)
+
+        f({})
+
+        nipsa_filter.assert_called_once_with(group_service, mock.ANY)
+
+    def call_passes_request_user(self, pyramid_request, nipsa_filter):
```
nickstenning: This code isn't being executed because the test method doesn't start with `test_`.

chdorner: Oh no 😱 

Fixed now 👍

PR #4492, opened 2017-04-05
Show flag summary information to group moderators

fatbusinessman:
When presenting annotations for someone who is a moderator in their group, this will now include a `moderation` key in the annotation JSON. Currently it just shows the number of times each annotation has been flagged for the moderator’s attention.

My current plan:

- [x] Pull out the model logic into a service, akin to #4475 – this could go in a subsequent PR
- [x] Give the commits some hefty rejigging
- [x] Consider changing the name of the formatter class, since it could now be confused with the `AnnotationModeration` model
- [x] Decide whether to keep the functional test, remove it, or add to it

This work is for hypothesis/product-backlog#208.

### DIFF
--- a/h/formatters/annotation_moderation.py
+++ b/h/formatters/annotation_moderation.py
@@ -1,0 +1,67 @@
+# -*- coding: utf-8 -*-
+
+from __future__ import unicode_literals
+
+import sqlalchemy as sa
+from zope.interface import implementer
+
+from h import models
+from h.formatters.interfaces import IAnnotationFormatter
+
+
+@implementer(IAnnotationFormatter)
+class AnnotationModerationFormatter(object):
+    """
+    Formatter for exposing an annotation's moderation information.
+
+    If the passed-in user has permission to hide the annotation (if they are a
+    moderator of the annotation's group, for instance), this formatter will
+    add a `moderation` key to the payload, with a count of how many users have
+    flagged the annotation.
+    """
+
+    def __init__(self, session, user, has_permission):
+        self._session = session
+        self._user = user
+        self._has_permission = has_permission
+
+        # Local cache of flag counts. We don't need to care about detached
+        # instances because we only store the annotation id and a count.
+        self._cache = {}
+
+    def preload(self, ids):
+        if self._user is None:
+            return
+
+        if not ids:
+            return
+
+        query = self._session.query(sa.func.count(models.Flag.id).label('flag_count'),
+                                    models.Flag.annotation_id) \
+                             .filter(models.Flag.annotation_id.in_(ids)) \
+                             .group_by(models.Flag.annotation_id)
+
+        flag_counts = {f.annotation_id: f.flag_count for f in query}
+        missing_ids = set(ids) - set(flag_counts.keys())
+        flag_counts.update({id_: 0 for id_ in missing_ids})
+
+        self._cache.update(flag_counts)
+
+        return flag_counts
+
+    def format(self, annotation_resource):
+        if not self._has_permission('admin', annotation_resource.group):
+            return {}
+
+        flag_count = self._load(annotation_resource.annotation.id)
+        return {'moderation': {'flagCount': flag_count}}
+
+    def _load(self, id_):
+        if id_ in self._cache:
+            return self._cache[id_]
+
+        flag_count = self._session.query(sa.func.count(models.Flag.id)) \
+                                  .filter_by(annotation_id=id_) \
+                                  .scalar()
+        self._cache[id_] = flag_count
+        return flag_count
--- a/tests/h/formatters/annotation_moderation_test.py
+++ b/tests/h/formatters/annotation_moderation_test.py
@@ -1,0 +1,113 @@
+# -*- coding: utf-8 -*-
+
+from __future__ import unicode_literals
+
+from collections import namedtuple
+
+import pytest
+
+from h.formatters.annotation_moderation import AnnotationModerationFormatter
+
+FakeAnnotationResource = namedtuple('FakeAnnotationResource',
+                                    ['annotation', 'group'])
+
+
+class FakePermissionCheck(object):
+
+    def __init__(self):
+        self._permissions = {}
+
+    def add_permission(self, permission, context, granted):
+        self._permissions[(permission, context)] = granted
+
+    def __call__(self, permission, context):
+        return self._permissions[(permission, context)]
+
+
+class TestAnnotationModerationFormatter(object):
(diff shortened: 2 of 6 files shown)


h/formatters/annotation_moderation.py
```diff
@@ -0,0 +1,67 @@
...
+        self._cache.update(flag_counts)
+
+        return flag_counts
+
+    def format(self, annotation_resource):
+        if not self._has_permission('admin', annotation_resource.group):
+            return {}
+
+        flag_count = self._load(annotation_resource.annotation.id)
+        return {'moderation': {'flag_count': flag_count}}
```
nickstenning: Small thing, but I think we'll probably want `flagCount` here. We don't have a lot of multiword keys in the API output, the one exception being the selectors for the annotation, but those use `lowerCamelCase` rather than `lower_underscored`.

nickstenning:
This is looking pretty good to me! FWIW, my feedback on your to-do list:

- services good ✔️ 
- hefty rejigging good ✔️ 
- meh, don't worry too much about the name -- we have package namespaces and [they're cool](https://www.python.org/dev/peps/pep-0020/)
- perhaps we could work the functional test into a more generic functional test of the annotation API in totality? (i.e. prepare a couple of annotations with various data, and check the output is as we expect, rather than checking each feature individually)

nickstenning:
Oh, and I think having to pass `request.has_permission` around is a little unfortunate, but I can't immediately think of a better idea, so... ¯\_(ツ)_/¯

h/formatters/annotation_moderation.py
```diff
@@ -0,0 +1,67 @@
...
+        self._user = user
+        self._has_permission = has_permission
+
+        # Local cache of flag counts. We don't need to care about detached
+        # instances because we only store the annotation id and a count.
+        self._cache = {}
+
+    def preload(self, ids):
+        if self._user is None:
+            return
```
nickstenning: We probably also want an `if not ids` shortcut here, too.

h/formatters/annotation_moderation.py
```diff
@@ -0,0 +1,67 @@
...
+
+        query = self._session.query(sa.func.count(models.Flag.id).label('flag_count'),
+                                    models.Flag.annotation_id) \
+                             .filter(models.Flag.annotation_id.in_(ids)) \
+                             .group_by(models.Flag.annotation_id)
```
nickstenning: This is very much a take it or leave it comment, but I tend to think that using parens for multi-line queries is a bit neater than escaping newlines:

```python
        query = (self._session.query(sa.func.count(models.Flag.id).label('flag_count'),
                                     models.Flag.annotation_id)
                 .filter(models.Flag.annotation_id.in_(ids))
                 .group_by(models.Flag.annotation_id))
```

h/formatters/annotation_moderation.py
```diff
@@ -0,0 +1,67 @@
...
+
+        flag_counts = {f.annotation_id: f.flag_count for f in query}
+        missing_ids = set(ids) - set(flag_counts.keys())
```
nickstenning: Again, certainly not a blocker for merging this, but it might be slightly less work to do this the other way round. That is:

```python
flag_counts = {id_: 0 for id_ in ids}
flag_counts.update({f.annotation_id: f.flag_count for f in query})
```

h/formatters/annotation_moderation.py
```diff
@@ -0,0 +1,67 @@
...
+    def format(self, annotation_resource):
+        if not self._has_permission('admin', annotation_resource.group):
+            return {}
+
+        flag_count = self._load(annotation_resource.annotation.id)
+        return {'moderation': {'flagCount': flag_count}}
+
+    def _load(self, id_):
+        if self._user is None:
+            return False
```
nickstenning: It would be good to add a test for the negative branch here.

h/formatters/annotation_moderation.py
```diff
@@ -0,0 +1,67 @@
...
+
+        if id_ in self._cache:
+            return self._cache[id_]
```
nickstenning: Ditto -- we should cover the positive branch here.

tests/h/formatters/annotation_moderation_test.py
```diff
@@ -0,0 +1,97 @@
...
+
+    def __call__(self, permission, context):
+        return self._permissions[(permission, context)]
+
+
+class TestAnnotationModerationFormatter(object):
+    def test_preload_sets_flag_counts(self, db_session, flagged, unflagged, user):
+        formatter = AnnotationModerationFormatter(db_session,
+                                                  user,
+                                                  has_permission=None)
```
nickstenning: `has_permission=None` is  presumably a convenience, in the sense that you just happen not to be calling any methods in this test that would explode because you haven't passed it in?

h/formatters/annotation_moderation.py
fatbusinessman: Good point; added.

h/formatters/annotation_moderation.py
fatbusinessman: Good point. I personally prefer that too; I was following the convention I’ve seen elsewhere in the codebase, but I’d be happy to go with this one instead.

Given this is code I’m moving around in #4502, I’m going to take your suggestion, but over in the other pull request.

h/formatters/annotation_moderation.py
fatbusinessman: You’re right: it would. Again, I’ll address this over in #4502.

h/formatters/annotation_moderation.py
fatbusinessman: Having looked at it again, I realised that this path will never actually get run, because if the user is `None`, they’ll fail the permission check in `format` and this method won’t get called. I’ve removed the check.

h/formatters/annotation_moderation.py
fatbusinessman: Done.

tests/h/formatters/annotation_moderation_test.py
fatbusinessman: Correct. I’ve managed to eliminate most of these cases (except the one where the user is `None`) by extracting a fixture for the common case (of a user who has permission to see the `moderation` key).

fatbusinessman:
Ok. I think I’ve addressed all the feedback (thanks, @nickstenning!) and rebased onto `master` to accommodate #4499. Give or take Jenkins having a bit of a sad at the moment, I hope this is ready for review again (and, give or take some `autosquash` magicks that I’ve deferred to make review easier, for merging).

nickstenning:
LGTM pending a `rebase --autosquash`.

PR #4496, opened 2017-04-06
Change annotation formatters to accept a resource

fatbusinessman:
Some formatters (including the forthcoming one to add flag counts for moderators) are going to need to make decisions based on the current user's permissions, so needs access to the resource object.

This makes some of the tests for the existing formatter slightly less neat, as they need to wrap the annotation up in something that quacks like a resource. I also needed to add a new test to the presenter, as changing the interface didn't end up breaking any of the existing tests.

### DIFF
--- a/h/formatters/interfaces.py
+++ b/h/formatters/interfaces.py
@@ -13,13 +13,13 @@ class IAnnotationFormatter(Interface):
     without putting everything in the annotation presenter, thus allowing better
     decoupling of code.
 
-    The main method is ``format(annotation)`` which is expected to return a
-    dictionary representation based on the passed-in annotation. If the
-    formatter depends on other data it should be able to load it on-demand
+    The main method is ``format(annotation_resource)`` which is expected to
+    return a dictionary representation based on the passed-in annotation. If
+    the formatter depends on other data it should be able to load it on-demand
     for the given annotation.
 
     Since we are rendering lists of potentially hundreds of annotations in one
-    request formatters need to be able to optimize the fetching of additional
+    request, formatters need to be able to optimize the fetching of additional
     data (e.g. from the database). Which is why this interface defines the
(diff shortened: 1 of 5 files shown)


h/formatters/interfaces.py
```diff
@@ -34,12 +34,12 @@ def preload(ids):  # noqa: N805
...
 
-    def format(annotation):  # noqa: N805
+    def format(annotation_resource):  # noqa: N805
         """
         Presents additional annotation data that will be served to API clients.
 
-        :param annotation: The annotation object that needs presenting.
-        :type annotation: memex.models.Annotation
+        :param annotation_resource: The annotation that needs presenting.
+        :type annotation: memex.resources.AnnotationResource
```
nickstenning: `:type annotation:` should be `:type annotation_resource:`, I think.

h/formatters/interfaces.py
```diff
@@ -34,12 +34,12 @@ def preload(ids):  # noqa: N805
...
 
-    def format(annotation):  # noqa: N805
+    def format(annotation_resource):  # noqa: N805
         """
         Presents additional annotation data that will be served to API clients.
 
-        :param annotation: The annotation object that needs presenting.
-        :type annotation: memex.models.Annotation
+        :param annotation_resource: The annotation that needs presenting.
+        :type annotation_resource: memex.resources.AnnotationResource
```
chdorner: I just merged #4495, so this (and all other references to `memex.resources.AnnotationResource`) in this PR need to be `h.resources.AnnotationResource`.

nickstenning:
LGTM pending rebase, but seeing as @chdorner was in here recently I reckon it should get a review from him, too.

fatbusinessman:
Cool; all done and rebased onto the new `master` branch.

PR #4497, opened 2017-04-06
Move memex.links to h.services.links

nickstenning:
Moves the links service into the h package alongside other services. Happily, this services is relatively self-contained, and so this is a straight move from one place to another.

The only slight "gotcha" is that `h.services` must now be included before `h.links`, otherwise the `add_annotation_link_generator` configuration directive won't be defined.

This is part of a project to reintegrate the `memex` package into the main `h` package: https://notes.wtk.io/2017/03/08/reintegrating-memex.

_This and #4495 can be merged in either order, although there will likely be a small conflict in the `h.streamer.messages` imports which will need resolving regardless._

### DIFF
--- a/h/services/__init__.py
+++ b/h/services/__init__.py
@@ -28,7 +28,7 @@ def includeme(config):
     config.register_service_factory('.user_signup.user_signup_service_factory', name='user_signup')
 
     config.add_directive('add_annotation_link_generator',
-                         '.links._add_annotation_link_generator')
+                         '.links.add_annotation_link_generator')
     config.add_request_method('.feature.FeatureRequestProperty',
                               name='feature',
                               reify=True)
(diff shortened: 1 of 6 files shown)


h/services/__init__.py
```diff
@@ -26,6 +27,8 @@ def includeme(config):
     config.register_service_factory('.user_password.user_password_service_factory', name='user_password')
     config.register_service_factory('.user_signup.user_signup_service_factory', name='user_signup')
 
+    config.add_directive('add_annotation_link_generator',
+                         '.links._add_annotation_link_generator')
```
chdorner: Shouldn't we make this function public now that it's referenced here? Maybe we can just make the public function get the config instead of the registry directly.

nickstenning: Yes, probably. I'll have a look.

PR #4499, opened 2017-04-06
Expose moderated annotations with `hidden` key in payload.

chdorner:
This is part of hypothesis/product-backlog#239

This will make sure the client can either show a special banner for
group moderators to unhide the annotation, or if a moderated annotation
has replies we want to show blacked-out text, so this will signal that
to the client as well.

### DIFF
--- a/h/services/annotation_moderation.py
+++ b/h/services/annotation_moderation.py
@@ -23,6 +23,24 @@ class AnnotationModerationService(object):
                         .filter_by(annotation_id=annotation_id)
         return self.session.query(q.exists()).scalar()
 
+    def all_hidden(self, annotation_ids):
+        """
+        Check which of the given annotation ids is hidden.
+
+        :param annotation_ids: The ids of the annotations to check.
+        :type annotation: list of unicode
+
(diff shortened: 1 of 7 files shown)


h/services/annotation_moderation.py
```diff
@@ -23,6 +23,24 @@ def hidden(self, annotation_id):
...
 
+    def all_hidden(self, annotation_ids):
+        """
+        Check which of the given annotation ids is hidden.
+
+        :param annotation_ids: The ids of the annotations to check.
+        :type annotation: list of unicode
+
+        :returns The subset of the annotation ids that are hidden.
+        :rtype set of unicode
```
nickstenning: I think you're missing a couple of colons here: `:returns:` and `:rtype:`

chdorner: Ah right, my brain probably just doesn't want to add the trailing colons because it looks weird 😆

chdorner:
As mentioned on Slack, we always return `hidden: false` if the annotation author is the logged-in user. Once @robertknight tested hiding an annotation of another user it worked as expected.

PR #4500, opened 2017-04-06
Properly reindex annotations when un-/nipsa'ing a user

chdorner:
We used to go directly to the ElasticSearch index and add or remove the
`"nipsa": true` item to or from the annotation object.

Now that the nipsa flag either gets set based on the user nipsa flag, or
the annotation nipsa flag we have to properly re-index the users'
annotations so we go through the normal search indexing code path.

Considering that most users probably won't have anywhere close to >10k
annotations, I've decided to just load all the annotation ids of a user
in one query.


chdorner:
> Given that `add_nipsa` and `remove_nipsa` do identical things, would it perhaps make sense to get rid of `h.tasks.nipsa` entirely, and instead add a `reindex_user_annotations` task to `h.tasks.indexer`?

I was thinking about this. If we don't need the logging (maybe for debugging purposes?), then yes sure, that would be much nicer.

nickstenning:
Let's do that then. We can always move the logging up one level into the view function(s).

chdorner:
Ok, that's done now. There is no more `h.tasks.nipsa`, but now there is `h.tasks.indexer.reindex_user_annotations`.

Note (because I always forget): make sure to run `make clean` before testing this out locally, because we need to remove a bunch of `.pyc` files.

nickstenning:
Lovely stuff. 👍

PR #4501, opened 2017-04-06
Remove unused imports from flag service test

fatbusinessman:
The code that used the `xrange` import disappeared in commit 22cb7a1, and the code that used the `text_type` import disappeared in commit a83b508.

### DIFF
--- a/tests/h/services/flag_test.py
+++ b/tests/h/services/flag_test.py
@@ -6,7 +6,6 @@ import pytest
 
 from h.services import flag
 from h import models
-from h._compat import text_type, xrange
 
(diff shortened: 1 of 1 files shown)


tests/h/services/flag_test.py
```diff
@@ -6,7 +6,7 @@
 
 from h.services import flag
 from h import models
-from h._compat import text_type, xrange
+from h._compat import text_type
```
fatbusinessman: Well played, Hound. Well played.

nickstenning:
:+1:

PR #4502, opened 2017-04-06
Extract the flag-counting logic into a service

fatbusinessman:
Extracting flag-counting logic into a service means the `AnnotationModerationFormatter` doesn't need to know anything about how the flag models are laid out, and so doesn't need direct access to the database session.

### DIFF
--- a/h/services/annotation_json_presentation.py
+++ b/h/services/annotation_json_presentation.py
@@ -13,7 +13,7 @@ from h.interfaces import IGroupService
 
 
 class AnnotationJSONPresentationService(object):
-    def __init__(self, session, user, group_svc, links_svc, flag_svc, moderation_svc, has_permission):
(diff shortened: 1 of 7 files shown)


h/services/annotation_json_presentation.py
```diff
@@ -13,15 +13,15 @@
 
 
 class AnnotationJSONPresentationService(object):
-    def __init__(self, session, user, group_svc, links_svc, flag_svc, moderation_svc, has_permission):
+    def __init__(self, session, user, group_svc, links_svc, flag_svc, flag_count_svc, moderation_svc, has_permission):
```
chdorner: Not your fault, but this is getting increasingly cumbersome to manage, and the tests as well. In the past we didn't really want to pass a request to a service, but maybe we should think about an exception for this. Or come up with a different solution.

PR #4505, opened 2017-04-07
Optimise annotation moderation reindex

chdorner:
~**This is based on #4499 and needs a rebase once that one got merged.**~

The [recent changes to the nipsa subscriber](https://github.com/hypothesis/h/commit/c4b2bfcc22120dcff601a7ae27bd68bac9482095#diff-eb1d680a2a50ef1314ba44cd8edb1d12) meant that if we re-index batches of annotations (most notable when re-indexing from scratch) we did one query per annotation for checking the moderation state.

This PR adds a backref to the Annotation model (`Annotation.moderation`), and the `AnnotationModerationService.hidden` method changes the nipsa check to basically `(annotation.moderation is not None)`. This allows us to eager load this relationship and thus stop making a query per annotation.

### DIFF
--- a/h/events.py
+++ b/h/events.py
@@ -20,6 +20,7 @@ class AnnotationTransformEvent(object):
     annotation just before it is indexed or in other use-cases.
     """
 
-    def __init__(self, request, annotation_dict):
+    def __init__(self, request, annotation, annotation_dict):
         self.request = request
+        self.annotation = annotation
         self.annotation_dict = annotation_dict
--- a/h/models/annotation_moderation.py
+++ b/h/models/annotation_moderation.py
@@ -26,7 +26,10 @@ class AnnotationModeration(Base, Timestamps):
 
     #: The annotation which has been flagged.
     annotation = sa.orm.relationship('Annotation',
-                                     backref=sa.orm.backref('moderation', uselist=False))
+                                     backref=sa.orm.backref('moderation',
+                                                            uselist=False,
+                                                            cascade='all, delete-orphan',
+                                                            passive_deletes=True))
 
     def __repr__(self):
         return '<AnnotationModeration annotation_id=%s>' % self.annotation_id
--- a/h/models/annotation_moderation.py
+++ b/h/models/annotation_moderation.py
@@ -25,7 +25,8 @@ class AnnotationModeration(Base, Timestamps):
                               nullable=False)
 
     #: The annotation which has been flagged.
-    annotation = sa.orm.relationship('Annotation')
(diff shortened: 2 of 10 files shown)


chdorner:
I don't think the codecov checks are very meaningful in this case, or does anybody disagree? Here's the coverage diff: https://codecov.io/gh/hypothesis/h/pull/4505/diff

h/events.py
```diff
@@ -20,6 +20,7 @@ class AnnotationTransformEvent(object):
     annotation just before it is indexed or in other use-cases.
     """
 
-    def __init__(self, request, annotation_dict):
+    def __init__(self, request, annotation, annotation_dict):
         self.request = request
+        self.annotation = annotation
         self.annotation_dict = annotation_dict
```
nickstenning: This looks a bit odd now. It's not really your fault and I completely understand why this change is as it is, but it definitely seems wrong to be passing an annotation and a serialisation of the same annotation into this event.

h/models/annotation_moderation.py
```diff
@@ -25,7 +25,8 @@ class AnnotationModeration(Base, Timestamps):
                               nullable=False)
 
     #: The annotation which has been flagged.
-    annotation = sa.orm.relationship('Annotation')
+    annotation = sa.orm.relationship('Annotation',
+                                     backref=sa.orm.backref('moderation', uselist=False, cascade='all, delete, delete-orphan'))
```
nickstenning: Strictly, the presence of `ondelete='cascade'` in the FK defined above means that the `delete` cascade is redundant here (and potentially less efficient than letting the database do the work). What we probably want to do is:

- leave the database-side cascade as-is
- use `cascade='all, delete-orphan'` in the backref
- add `passive_deletes` to the backref, which will ensure that SQLA won't force the loading of `AnnotationModeration` rows when deleting annotations.

See [the call-out box in the SQLA docs for more information](http://docs.sqlalchemy.org/en/rel_1_1/orm/cascades.html#delete).

nickstenning:
Oh, sorry, I approved the PR but I think we should tidy up the cascade options before merge.

h/events.py
chdorner: Yup, I seem to remember mentioning it in one of the recent commit messages that we should use some sort of "annotation search" formatters, but that we should do it later. I will make sure to add a eng backlog card for it.

h/models/annotation_moderation.py
chdorner: 👍 Fixed.

nickstenning:
Yep! LGTM. 🌵

PR #4511, opened 2017-04-07
New CLI commands for user: password, delete

mgax:


mgax:
Ah, I was working off an old version, and the API for passwords has changed. Should be fixed now.

nickstenning:
LGTM. The only slight query I have is around use of the full `UserPasswordService` (rather than a stub) in the tests, but it's not a big deal.

mgax:
Cool, thanks for merging!

PR #4514, opened 2017-04-10
Remove the remainders of memex

nickstenning:
This commit removes most of the remainders of the memex package. Namely, it:

- Removes the `src/memex/` directory (and with it a duplicated `_compat` module)
- Removes the `tests/memex/` directory (and with it a duplicate set of test fixtures)
- Removes setup.py entirely, moving relevant omitted package specifications into `requirements.in`
- Removes a couple of logger configurations for the memex package
- Removes the dependency on memex from `requirements.in`
- Removes the inclusion of the memex pyramid module from `h/app.py`
- Updates the Dockerfile build steps


Dockerfile
```diff
@@ -23,8 +23,7 @@ WORKDIR /var/lib/hypothesis
 RUN chown -R hypothesis:hypothesis /var/log/nginx /var/lib/nginx
 
 # Copy minimal data to allow installation of dependencies.
-COPY src/memex/__init__.py ./src/memex/
-COPY README.rst setup.* requirements.txt ./
+COPY requirements.txt ./
```
chdorner: Is the removal of `README.rst` on purpose?

nickstenning: Yep -- it was only needed at this stage because it was referenced by `setup.py` it's added later by `COPY . .` anyway.

PR #4516, opened 2017-04-11
Import h.models while initialising database

nickstenning:
Rather than having each place that calls `h.db.init` have to ensure that the model metadata is correctly populated, it makes sense to move this responsibility into the `init` function itself.


robertknight:
LGTM.

PR #4517, opened 2017-04-11
Remove most references to the memex name

nickstenning:
This commit:

- renames some configuration directives that come from memex
- renames a couple of internal registry keys
- updates some code comments

to remove most of the remaining references to the "memex" name from the code.

The main remaining items are:

- clean up a duplication of APIError exception classes between h and (ex-)memex code
- rename some metrics

### DIFF
--- a/h/groups/__init__.py
+++ b/h/groups/__init__.py
@@ -2,4 +2,4 @@
 
 
 def includeme(config):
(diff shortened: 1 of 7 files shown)


robertknight:
LGTM

PR #4519, opened 2017-04-11
Conditionally redirect users from homepage

nickstenning:
Once our marketing website has moved fully off the main application domain, we'll serve redirects from the root URL that depend on the user's logged-in status.

If a user is logged in, they are redirected to their own profile page. If they're not, they are redirected to the marketing website.

The URL to which logged-out users are redirected is configured using the `HOMEPAGE_REDIRECT_URL` environment variable.

This change is feature-flagged (`homepage_redirects`).

~~_**N.B.** This PR is based on #4518 and will probably be easier to review once that's merged and this is rebased._~~

### DIFF
--- a/h/config.py
+++ b/h/config.py
@@ -89,6 +89,8 @@ SETTINGS = [
     # *not* toggle functionality based on this value. It is intended as a
     # label only.
     EnvSetting('h.env', 'ENV'),
+    # Where should logged-out users visiting the homepage be redirected?
+    EnvSetting('h.homepage_redirect_url', 'HOMEPAGE_REDIRECT_URL'),
     EnvSetting('h.proxy_auth', 'PROXY_AUTH', type=asbool),
     # Sentry DSNs for frontend code should be of the public kind, lacking the
     # password component in the DSN URI.
--- a/h/views/home.py
+++ b/h/views/home.py
@@ -33,3 +33,21 @@ def index(context, request):
         )
 
     return context
+
+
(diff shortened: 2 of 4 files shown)


h/views/home.py
```diff
@@ -33,3 +33,18 @@ def index(context, request):
...
     return context
+
+
+@view_config(route_name='index',
+             request_method='GET',
+             feature='homepage_redirects')
+def index_redirect(context, request):
+    settings = request.registry.settings
+    redirect = settings.get('h.homepage_redirect_url',
+                            'https://web.hypothes.is')
```
chdorner: I think it would make more sense to redirect to `/search` by default, otherwise I will always get redirected to another domain in local development which is confusing.

h/config.py
```diff
@@ -89,6 +89,8 @@
     # *not* toggle functionality based on this value. It is intended as a
     # label only.
     EnvSetting('h.env', 'ENV'),
+    # Where should logged-out users visiting the homepage be redirected?
+    EnvSetting('h.homepage_redirect_url', 'HOMEPAGE_REDIRECT_URL'),
```
chdorner: I've configured this on Skyliner.

h/views/home.py
nickstenning: Sounds reasonable. :+1:

nickstenning: Updated to do that instead.

PR #4520, opened 2017-04-11
Add support for falling through to web service

chdorner:
When the request contains the header key/value
`Hypothesis-Test-Migration: 1`, the fallback upstream will always be we
web service itself. This is to aid testing the migration of the
marketing website (WordPress) to the `web.hypothes.is` subdomain.

To aid this change we also get rid of the warning message response when
no fallback upstream URI is configured. In that case we just send the
request to the web service as well and let it respond with a 404 Not
Found response.

### DIFF
--- a/conf/nginx.conf.tpl
+++ b/conf/nginx.conf.tpl
@@ -26,6 +26,11 @@ http {
   upstream web { server unix:/tmp/gunicorn-web.sock fail_timeout=0; }
   upstream websocket { server unix:/tmp/gunicorn-websocket.sock fail_timeout=0; }
 
+  map $http_hypothesis_test_migration $default_upstream {
+    default $fallback_upstream;
+    1 http://web;
+  }
+
   server {
     listen 5000;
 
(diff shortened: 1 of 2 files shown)


conf/nginx.conf.tpl
```diff
@@ -81,7 +92,16 @@ http {
     }
 
     location / {
-      ${FALLBACK_DIRECTIVES}
+      proxy_pass $root_url;
```
chdorner: Duplicating this from the `location` directive above means that once we switch the marketing website we can just get rid of that directive.

conf/nginx.conf.tpl
```diff
@@ -26,6 +26,11 @@ http {
   upstream web { server unix:/tmp/gunicorn-web.sock fail_timeout=0; }
   upstream websocket { server unix:/tmp/gunicorn-websocket.sock fail_timeout=0; }
 
+  map $http_hypothesis_test_migration $root_url {
```
nickstenning: `$root_url` seems like a bit of an odd name for this. Maybe `$default_upstream` would make more sense?

conf/nginx.conf.tpl
```diff
@@ -39,6 +44,12 @@ http {
     rewrite ^/minutes/(\d+)/(.*) https://hypothesis-meeting-logs.s3.amazonaws.com/$1/$2 redirect;
     rewrite ^/minutes/(.*) https://shrub.appspot.com/hypothesis-meeting-logs/$1 redirect;
 
+
+    # Set the fallthrough upstream URL.
+    # We store the upstream endpoint in a variable to trigger dynamic resolution
+    # of the upstream hostname.
+    set $fallthrough_upstream ${FALLBACK_UPSTREAM};
```
nickstenning: Just to save my poor brain, can we refer to this as `fallback_upstream` for consistency?

conf/nginx.conf.tpl
nickstenning: :+1:

nickstenning:
Nice! Have a castle: 🏰

PR #4522, opened 2017-04-12
Pull formatters into the presenter constructor

fatbusinessman:
By pulling this into a constructor argument and adding the mystic underscore to take the `formatters` attribute out of the public interface, I'm hoping to achieve two things:

- Make it easier to pull formatters into a more general-purpose capability for all presenters.
- Stop us having to worry about extra formatters being added to a presenter after it's been constructed. We weren't doing this anywhere anyway, so making it so we don't even have to think about it seems helpful.

I've committed the cardinal Python sin here of putting a mutable default value into a constructor (the empty list of presenters). This saves having to assign it a default value within the method, and since we iterate over the argument and then discard it I don't think it's a risk, but we can change this if people have strong feelings.

### DIFF
--- a/h/presenters/annotation_json.py
+++ b/h/presenters/annotation_json.py
@@ -17,18 +17,22 @@ class AnnotationJSONPresenter(AnnotationBasePresenter):
 
     """Present an annotation in the JSON format returned by API requests."""
 
-    def __init__(self, annotation_resource):
+    def __init__(self, annotation_resource, formatters=None):
         super(AnnotationJSONPresenter, self).__init__(annotation_resource)
 
-        self.formatters = []
+        self._formatters = []
 
-    def add_formatter(self, formatter):
+        if formatters is not None:
+            for formatter in formatters:
+                self._add_formatter(formatter)
+
+    def _add_formatter(self, formatter):
         try:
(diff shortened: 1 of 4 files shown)


h/presenters/annotation_json.py
```diff
@@ -17,18 +17,21 @@ class AnnotationJSONPresenter(AnnotationBasePresenter):
 
     """Present an annotation in the JSON format returned by API requests."""
 
-    def __init__(self, annotation_resource):
+    def __init__(self, annotation_resource, formatters=[]):
```
nickstenning: I know you mention this in your PR message. Could we perhaps change this to `formatters=None` and a couple of extra lines below? Your code is fine, but seeing `foo=[]` in a parameter list is always going to cause people to have to stop and think, and thinking _hurts_ 🤕.

nickstenning:
I'm a moron. I meant to mark this as "changes requested" but failed.

h/presenters/annotation_json.py
fatbusinessman: Fair point, well made. Fixed.

nickstenning:
LGTM, for real this time.

PR #4524, opened 2017-04-12
Add the new `/api/links` endpoint

seanh:
Add the new `/api/links` endpoint and add a link to this endpoint to the existing `/api` endpoint. This new endpoint will be used by the client to get the URL templates for generating URLs to HTML pages of the web service. This is part of https://github.com/hypothesis/product-backlog/issues/215

### DIFF
--- a/h/views/api.py
+++ b/h/views/api.py
@@ -179,6 +179,32 @@ def index(context, request):
     }
 
 
+@api_config(route_name='api.links',
+            link_name='links',
+            description='URL templates for generating URLs for HTML pages')
+def links(context, request):
+    group_leave_url = request.route_url('group_leave', pubid='_id_')
+    group_leave_url = group_leave_url.replace('_id_', ':id')
+
+    tag_search_url = request.route_url('activity.search',
+                                       _query={'q': '_query_'})
+    tag_search_url = tag_search_url.replace('_query_', 'tag:":tag"')
+
+    user_url = request.route_url('stream.user_query', user='_user_')
+    user_url = user_url.replace('_user_', ':user')
+
+    return {
+        'account.settings': request.route_url('account'),
+        'forgot-password': request.route_url('forgot_password'),
+        'groups.leave': group_leave_url,
+        'groups.new': request.route_url('group_create'),
+        'help': request.route_url('help'),
+        'search.tag': tag_search_url,
+        'signup': request.route_url('signup'),
+        'user': user_url,
+    }
+
+
 @api_config(route_name='api.search',
             link_name='search',
             description='Search for annotations')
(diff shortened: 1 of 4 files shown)


h/views/api.py
```diff
@@ -179,6 +179,32 @@ def index(context, request):
...
+    group_leave_url = group_leave_url.replace('_id_', ':id')
+
+    tag_search_url = request.route_url('activity.search',
+                                       _query={'q': '_query_'})
+    tag_search_url = tag_search_url.replace('_query_', 'tag:":tag"')
+
+    user_url = request.route_url('stream.user_query', user='_user_')
+    user_url = user_url.replace('_user_', ':user')
+
+    return {
```
robertknight: I think it would be nice if this endpoint was self-documenting in a similar way to the `/api` endpoint. We could do this by keeping the key as the route name but changing the value to a `{ url, desc }` object, consistent with the `/api` endpoint. Thoughts?

h/views/api.py
```diff
@@ -179,6 +179,32 @@ def index(context, request):
...
+        'account.settings': request.route_url('account'),
+        'forgot-password': request.route_url('forgot_password'),
+        'groups.leave': group_leave_url,
```
robertknight: If we follow the suggestion above to add a description to each link, I would then explicitly call out the fact that `groups.leave` is deprecated.

h/views/api.py
seanh: We actually discussed this on Slack way back (one of the Slack convo's that the product-backlog issue for this links to, I think) because I wanted each link to be an object with `url` and `method`, but the `method` would always be `GET`, and these links aren't API endpoints like the `/api` ones are they're just links to be rendered as `href`s and the browser directed to, so we decided on a simple list of name -> URL. It's the same as what the github API uses.

What would be the benefit of adding `desc`s to these, beyond the already-existing descriptiveness of the name and the URL itself?

```json
account.settings: {
  url: "http://localhost:5000/account/settings",
  desc: "The account settings page"
}
```

I'm not sure it adds much.

I suppose we could add, in the `desc` or in an additional key, that `groups.leave` is deprecated. But I'm not sure anyone would notice it. A developer programs their client to use the links API once, later one deprecated is added to the `desc` of one of the links, but unless the developer is manually looking at the `/api/links` output again after having already finished programming their client to use it, they aren't likely to notice are they? (Of course developers writing _new_ clients might see it...)

seanh:
I've added the new feature flag to this but note that there's an issue over in the client PR with the client being able to actually use this feature flag

h/views/api.py
robertknight: @seanh - OK, if we've already had the discussion I don't have a strong enough opinion to justify rehashing it. Let's go with this as-is.

PR #4525, opened 2017-04-12
Update the "engines" entry in package.json

robertknight:
We haven't supported Node v0.10 for a long time. The current requirement is Node v6+ (the current stable/LTS release).

_Edit: An earlier version of this PR removed the field entirely. At @fatbusinessman 's suggestion I've revised it to just update the field_.

### DIFF
--- a/package.json
+++ b/package.json
@@ -67,7 +67,7 @@
     "websocket": "^1.0.22"
   },
   "engines": {
-    "node": "0.10.x"
+    "node": ">=6"
   },
   "repository": {
     "type": "git",


robertknight:
The vast majority of users wouldn't benefit because "engines" is ignored by default, you just get an [advisory message](https://docs.npmjs.com/files/package.json#engines) if you tried to install the "h" package as a dependency of another package.

However, if you can enable strict checking via:

```
npm config set engine-strict true
```

For these users, they would get a helpful error message. I suppose its a useful form of documentation as well.

I've changed the PR to update rather than remove the engines field.

nickstenning:
LGTM.

PR #4527, opened 2017-04-13
Generalise redirect tween

nickstenning:
We are shortly going to need to add a large number of redirects to the application. In preparation for this, this commit:

- moves the definition of redirects into a text file with a relatively simple format
- moves the parsing of this file and the looking up of redirects into functions provided by a `h.util.redirects` module
- adds support for various types of redirect not currently supported by the redirect tween

### DIFF
--- a/h/redirects
+++ b/h/redirects
@@ -1,0 +1,12 @@
+# Redirects to be checked before passing a request on to the application.
+#
+# Note that if this file is removed or incorrectly formatted, the application
+# will fail to boot. For details of the format, see h/util/redirects.py.
+
+# Legacy application URLs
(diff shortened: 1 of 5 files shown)


h/redirects
```diff
@@ -0,0 +1,7 @@
+# Legacy application URLs
```
chdorner: Since the application will fail to boot when this file is missing, maybe we should add a comment at the top saying `# For the application to boot, this file needs to exist, even if it's just empty`. Just so that when we (for some reason) in the future remove all redirects that we remember to keep the empty file around instead of just deleting it.

nickstenning: Sure thing!

PR #4528, opened 2017-04-13
Add WordPress redirects

nickstenning:
Add redirects for soon-to-be-previously-WordPress-hosted content.


nickstenning:
For reference, these come from the spreadsheet provided by @xolotl [over here](https://github.com/hypothesis/product-backlog/issues/248#issuecomment-293814699).

I've chosen to add slashless and slashed version of each of the exact redirects just to maximise the chances that people get to right place.

PR #4530, opened 2017-04-13
Rename some metrics with memex in their name

nickstenning:
Rename some statsd metric names to reflect the fact that `memex` is no longer a thing.


nickstenning:
Once this goes out we'll want to shuffle some whisper files around on `mon`.

PR #4531, opened 2017-04-18
Allow groups without a creator

chdorner:
As a preparation for moving the `__world__` group into the database,
which gets created automatically. While most instances of h should
have set a creator for this group, we don't want to make the
decision of who the creator will be.

### DIFF
--- a/h/migrations/versions/21b1ce37e327_allow_null_group_creator.py
+++ b/h/migrations/versions/21b1ce37e327_allow_null_group_creator.py
@@ -1,0 +1,23 @@
+"""
+Allow null group.creator
+
+Revision ID: 21b1ce37e327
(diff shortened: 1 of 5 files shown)


nickstenning:
Change looks good. Perhaps worth adding one or two very basic tests for the admin groups view at this point?

chdorner:
> Perhaps worth adding one or two very basic tests for the admin groups view at this point?

I would agree with you, but the only code I changed is the CSV export of the group which already is broken anyway because the request is timing out. So I opted for not adding any tests to that piece that will substantially change anyway (or we will just delete it perhaps because we now have other ways to view this data).

nickstenning:
Oh right, in which case don't worry about it. We should just kill off the report anyway.

PR #4532, opened 2017-04-18
Index the first of an annotation's references

fatbusinessman:
This is the first step toward indexing a top-level annotation's thread IDs (all annotations which cite the annotation as their top-level reference, so the entire tree of replies) into Elasticsearch, so that we don't end up hiding an entire thread when its root annotation gets hidden by a moderator.

A few notes on the expression for this index:

    [sa.text('("references"[1])')]

This is fairly heavily strewn with punctuation, and I feel I should explain what I understand of why. The double quotes around `references` are reasonably understandable: `references`, as well as being the name of the column, is a SQL keyword used when defining foreign keys, so it needs quoting as a column name. The extra parentheses I understand a little less well, but without them the migration fails with a syntax error:

    sqlalchemy.exc.ProgrammingError: (psycopg2.ProgrammingError) syntax error at or near ")"
    LINE 1: ...Y ix__annotation_thread_root ON annotation ("references"[1])
                                                                          ^
     [SQL: 'CREATE INDEX CONCURRENTLY ix__annotation_thread_root ON annotation ("references"[1])']

As for the "1" in the middle: Postgres uses 1-based indexing for its arrays, rather than 0-based indexing like most languages. Usually, SQLAlchemy's ORM hides this difference away from us, but we're not in the ORM world here.

### DIFF
--- a/h/migrations/versions/9389d52b037d_add_index_to_annotation_thread_root.py
+++ b/h/migrations/versions/9389d52b037d_add_index_to_annotation_thread_root.py
@@ -1,0 +1,27 @@
+"""
+Add index to annotation thread root
+
+Revision ID: 9389d52b037d
+Revises: b102c50b1133
+Create Date: 2017-04-18 12:36:53.842280
+"""
+
+from __future__ import unicode_literals
+
+import sqlalchemy as sa
+from alembic import op
+
+revision = '9389d52b037d'
+down_revision = 'b102c50b1133'
+
+
+def upgrade():
+    op.execute('COMMIT')
+    op.create_index(op.f('ix__annotation_thread_root'),
+                    'annotation',
+                    [sa.text('("references"[1])')],
+                    postgresql_concurrently=True)
+
+
+def downgrade():
+    op.drop_index(op.f('ix__annotation_thread_root'), 'annotation')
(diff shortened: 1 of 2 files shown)


h/migrations/versions/9389d52b037d_add_index_to_annotation_thread_root.py
```diff
@@ -0,0 +1,27 @@
+"""
+add_index_to_annotation_thread_root
```
nickstenning: It's a tiny thing, but this is usually a plain language sentence describing the change, such as "Add index to annotation thread root". At the moment this looks a bit funny:

```
$ hypothesis --dev migrate history | head -n10
b102c50b1133 -> 9389d52b037d (head), add_index_to_annotation_thread_root
50df3e6782aa -> b102c50b1133, Clean up moderation extra key
e554d862135f -> 50df3e6782aa, add annotation moderation table
5655d56d7c29 -> e554d862135f, Add index to flag.user_id
c322c57b49db -> 5655d56d7c29, Add flag table
18dfed902c9e -> c322c57b49db, Remove user uid column
7e2443f8d7d6 -> 18dfed902c9e, Remove unused user indices
faefe3b614db -> 7e2443f8d7d6, Make userid index unique
1a40e75a524d -> faefe3b614db, Make user uid nullable
02db2fa6ea98 -> 1a40e75a524d, Add normalised username index
```

PR #4533, opened 2017-04-19
Put __world__ group in database

chdorner:
~**This depends on #4531 and will need rebasing once that one is merged. The migration in #4531 also needs to be run on production before merging this.**~

This is part of hypothesis/product-backlog#272.

This gets rid of all special code paths where we handle the `__world__` group because this PR provides a database migration that will add this group to the database.

I'm not sure if it's worth deploying the database migration first? We can just run it in QA first to provide a smooth migration path in production.


tests/h/conftest.py
```diff
@@ -28,6 +28,7 @@
                                                 'postgresql://postgres@localhost/htest'))
 
 Session = sessionmaker()
+authority = 'example.com'
```
nickstenning: Given this isn't expected to change, it should probably be a "constant": `TEST_AUTHORITY = 'example.com'`

PR #4538, opened 2017-04-20
Remove the `thread` field from the search index

fatbusinessman:
This field, and its associated analyzer, are declared in the search config but never used.

As far as I can tell, this is a remnant of a previous way of representing threaded conversation in the search index. It's quite difficult to track down the history, though, given it dates back to 2014 (commit c12dc5bda593) and has endured multiple renames and refactorings since then. Still, it looks like it's now unused, so we can remove it to avoid confusion with the upcoming field to index the IDs of all replies and subreplies in an annotation's thread.

### DIFF
--- a/h/search/config.py
+++ b/h/search/config.py
@@ -105,10 +105,6 @@ ANNOTATION_MAPPING = {
         'document': {
             'enabled': False,  # not indexed
         },
-        'thread': {
-            'type': 'string',
-            'analyzer': 'thread'
-        },
         'group': {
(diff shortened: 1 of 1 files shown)


fatbusinessman:
> Will we need to do a reindex? Or will that wait until you add the new thread mapping?

Since I think we’ll need to do a full reindex to get all the thread IDs indexed, I suspect there’s no harm waiting until then, but I’m open to any more informed opinions.

PR #4541, opened 2017-04-24
Tweak new site redirects

nickstenning:
Having run through all the proposed redirects, I've made a few changes.

To avoid redirecting to a 404, I've updated:

- /annotations-to-the-next-president
- /new-about-page
- /nextprez

To avoid redirecting to a redirect, I've updated:

- /alpha (redirects to /)
- /what-is-it (redirects to /about)
- /who (redirects to /team)


xolotl:
LGTM! Thanks!

chdorner:
I will merge and deploy this since @nickstenning is on a train today.

PR #4543, opened 2017-04-24
Remove fallback and enable redirects to new site

nickstenning:
This commit removes the nginx configuration which allows for a "fallback" origin. This means that instead of sending non-whitelisted URLs to WordPress (as configured in production), they will now be handled by the application.

This has the effect of exposing the various redirects which have been prepared in the application and will "launch" the new blog and marketing website when released.

It also means that 404s will now be handled by the application.

This has been tested by deploying this branch to QA.

**N.B.** Do **not** merge until we are ready to launch the new site.


nickstenning:
This has already been reviewed by @chdorner. Merging.

PR #4546, opened 2017-04-25
Make logo point to new web.hypothes.is landing

sean-roberts:
we are using ?source=logo to help track users using this mechanism


chdorner:
Marked WIP because we need to wait until the new website launches.

PR #4550, opened 2017-04-27
Show hidden annotations with replies

fatbusinessman:
This is for hypothesis/product-backlog#231, and also deals with hypothesis/product-backlog#281 by always showing moderators the state of their own annotations.

For the indexing side that makes this possible, see #4540 and #4545.

Because we use top-level annotations to determine which replies to display, we need the search engine to return annotations that have replies, so hidden top-level annotations don't take the entire thread out with them. We then filter out the content itself in the presentation layer.


fatbusinessman:
Moderator view:

hidden-annotations-moderator-view

Non-moderator view:

hidden-annoations-muggle-view

PR #4553, opened 2017-05-06
Don't return same publisher group twice

seanh:
Don't return the same publisher group twice in the `/api/profile` response.

Steps to reproduce:

1. Follow the instructions at https://github.com/hypothesis/publisher-account-test-site
   to get the publisher account site setup
2. Login to the publisher test site as "admin" (the publisher group
   creator)

Result: You'll see the same Partner.org publisher group listed twice in
the groups dropdown menu in the client:

screenshot from 2017-05-06 22-56-31

Expected: You should only see the group once.

The problem appears to be that in the `_current_groups()` function the
publisher group appears in both the `authority_groups` and the
`_user_groups()`. The admin user, who is the creator of the publisher
group, is also a member of the group.

### DIFF
--- a/h/session.py
+++ b/h/session.py
@@ -66,7 +66,7 @@ def _current_groups(request, authority):
     authority_groups = (request.find_service(name='authority_group')
                         .public_groups(authority=authority))
 
-    groups = authority_groups + _user_groups(user)
+    groups = set(authority_groups + _user_groups(user))
 
     return [_group_model(request.route_url, group) for group in groups]
 


seanh:
Needs a test. Would be good if someone can confirm they can also reproduce this on master, too.

nickstenning:
This is fine, but we should probably also fix the underlying problem, which is that the current instructions result in the creator being a group member. "Publisher" groups shouldn't have members... (and being the user referenced in the `creator_id` field isn't the same as being a group member).

seanh:
This looks like the cause of the group creator becoming a member: <https://github.com/hypothesis/h/blob/c480243e112e2cd5f54b7b65e40a6891ca57fbfe/h/models/group.py#L92>, which I guess is actually desirable for non-publisher groups. So I think the `add-publisher-group` command needs to create the group a different way: https://github.com/hypothesis/h/blob/c480243e112e2cd5f54b7b65e40a6891ca57fbfe/h/cli/commands/groups.py#L28

Or alternatively we change the way that we're reusing the group's creator as its admin. Add group admin as a separate concept at the model level, and create publisher groups with an admin but no creator

nickstenning:
I'd suggest just not using the service to create the group in the CLI, and making one directly. Either that or modify the service so that it's possible to create groups without automatically adding the creator as a member.

chdorner:
> modify the service so that it's possible to create groups without automatically adding the creator as a member

This is now possible since #4531 and it's probably a good idea to split this up because certain group types should not have any members (when `joinable_by` is set to `None` or `authority`).

h/session.py
```diff
@@ -66,7 +66,7 @@ def _current_groups(request, authority):
     authority_groups = (request.find_service(name='authority_group')
                         .public_groups(authority=authority))
 
-    groups = authority_groups + _user_groups(user)
+    groups = set(authority_groups + _user_groups(user))
```
fatbusinessman: Given the discussion that’s already happened on this pull request, this might be a moot point by now, but does this version still preserve the ordering? As I understand it, the `set` type makes no promises about the order of its elements, so might this result in the groups being displayed out of order in the client?

nickstenning:
Did you have a plan to circle back to this @seanh?

seanh:
I completely forgot about this, but I can circle back to it yes, when I finish a couple other things first. I talked with @chdorner about this in SF and we agreed on a solution to how we wanted to implement the group creator not being a member, but I can't remember what we said, do you remember @chdorner ?

chdorner:
I think it was that we should move the part of the code where we add the creator as a member into the service or the view (not sure which one), but out of the constructor. Then we can also make sure to only add the creator as a member if the group supports memberships (as in, a private group, or not a publisher group).

seanh:
Created an issue to fix with a new PR: https://github.com/hypothesis/h/issues/4564

PR #4554, opened 2017-05-09
Accessible activity cards

judell:
This PR is intended to be a first step on a long road toward a more accessible product. It adds navigation roles and titles to key elements: the annotation quote and body, the count, the sidebar containers.

Screencast: http://jonudell.net/h/accessibility_06.mp4

Background: [H Accessibility](https://docs.google.com/document/d/1-XyRheGNgufrRVCv7AdshoGgfxsGI0hluAo2JuAkNeM/edit?usp=sharing)

That document also proposes tweaks to the injector (so highlights in the main doc are navigable/audible) and to the client (so cards in the sidebar are accessible in ways analogous to the behavior proposed here for activity pages).  But perhaps we should work through these first, then address the other cases where annotations are not consumable in a screenreader, and then think about how to tackle enabling screenreader-assisted creation of annotations.

### DIFF
--- a/h/static/scripts/controllers/search-bucket-controller.js
+++ b/h/static/scripts/controllers/search-bucket-controller.js
@@ -53,6 +53,8 @@ class SearchBucketController extends Controller {
     setElementState(this.refs.content, {expanded: state.expanded});
     setElementState(this.element, {expanded: state.expanded});
 
+    this.refs.title.setAttribute('aria-expanded', state.expanded.toString());
+
     // Scroll to element when expanded, except on initial load
     if (typeof prevState.expanded !== 'undefined' && state.expanded) {
       this.scrollTo(this.element);
--- a/h/static/scripts/controllers/search-bucket-controller.js
+++ b/h/static/scripts/controllers/search-bucket-controller.js
@@ -33,6 +33,11 @@ class SearchBucketController extends Controller {
       this.setState({expanded: !this.state.expanded});
     });
 
+    this.refs.title.addEventListener('click', (event) => {
+      this.setState({expanded: !this.state.expanded});
+      event.stopPropagation();
+    });
+
     this.refs.collapseView.addEventListener('click', () => {
       this.setState({expanded: !this.state.expanded});
     });
--- a/h/static/scripts/tests/controllers/search-bucket-controller-test.js
+++ b/h/static/scripts/tests/controllers/search-bucket-controller-test.js
@@ -8,6 +8,7 @@ const TEMPLATE = `<div class="js-search-bucket">
     <a data-ref="domainLink">foo.com</a>
   </div>
   <div data-ref="content"></div>
+  <a data-ref="title"></a>
   <button data-ref="collapseView"></button>
 </div>
 `;
--- a/h/static/scripts/tests/controllers/search-bucket-controller-test.js
+++ b/h/static/scripts/tests/controllers/search-bucket-controller-test.js
@@ -68,6 +68,14 @@ describe('SearchBucketController', () => {
     assert.calledWith(ctrl.scrollTo, ctrl.element);
   });
 
+  it('sets ARIA expanded state when expanded or collapsed', () => {
+    ctrl.refs.title.dispatchEvent(new Event('click'));
+    assert.equal(ctrl.refs.title.getAttribute('aria-expanded'), 'true');
+
+    ctrl.refs.title.dispatchEvent(new Event('click'));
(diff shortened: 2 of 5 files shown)


robertknight:
@judell - Now that I've had a chance to familiarise myself with VoiceOver on my Mac, I've had a look at this. These changes are great:

- Making the bucket title a link rather than just a `<div>` so it is clear that it can be clicked
- The addition of titles to various elements to better label them

What I don't follow is the reason for adding the "navigation" role to many elements. My understanding of the spec is that this role serves the same purpose as the [`<nav>` element](https://developer.mozilla.org/en/docs/Web/HTML/Element/nav) and should only appear once or a few times on the page - eg. for the global navigation at the top of the page, plus secondary navigation within the current page.

Compare for example the navigation landmarks that you see in Facebook:

img_1268


With what is seen in this PR:

img_1269

For the tags, annotators and URL sections in an annotation bucket, and the rest of the search results in general what I think we should do instead is to make better use of headings. If you agree, then perhaps what I would suggest is just to keep the labelling + bucket header link change in this PR but remove the "navigation" role. Improving the use of headings can be done separately.

Some other things I noticed, just with the search results, for future work:

- Search result buckets need additional attributes to indicate that they are expandable and whether they are currently expanded.
- A skip link at the start of the page that goes direct to the results might be useful.

judell:
> make better use of headings

That's just the feedback this needed! It reinforces something that Susan Gerhart told me in an interview (https://blog.jonudell.net/2008/06/30/from-seeing-to-hearing-a-conversation-with-susan-gerhart-about-assistive-technologies-for-the-sight-impaired/, https://blog.jonudell.net/2009/06/18/its-the-headings-stupid/) long ago. 

tl;dr "It's the headings, stupid!"

Thanks Rob, will do.

robertknight:
> (What's the tool that reveals/enumerates landmarks? That's super-helpful!)

I'm using a feature that is part of VoiceOver called the Web Rotor, which is activated with VO+U (where "VO" is Command+Option by default). In Windows 10's Narrator screen reader, it looks like there is an analogous feature, [Narrator views](https://hyp.is/r6ELcDeVEeehSb8BzY3hDQ/support.microsoft.com/en-gb/help/22808/windows-10-narrator-learning-basics). There is also the [Landmarks](https://hyp.is/8_JtjDeVEeemv4s82QLElg/support.microsoft.com/en-gb/help/22808/windows-10-narrator-learning-basics) view. If you find a better way, let me know :)

robertknight:
Sorry I haven't got back to you on this yet @judell - I'll have another look tomorrow. The investigation this lead to has been very insightful for me in the meantime 🙂

judell:
@robertknight No worries, I was just going to ask: This currently replaces the <nav> bits with <header> bits, but if there's a better way to do that I'll rewind to just the bucket title.

If you do think the <header> approach is valid here, might it also be valid as a substitute for <nav> [here](https://docs.google.com/document/d/1-XyRheGNgufrRVCv7AdshoGgfxsGI0hluAo2JuAkNeM/edit#heading=h.pg5r057ups4)?

> The investigation this led to has been very insightful for me

And me! It's often said every dev should have to watch users actually using their software. A corollary: Every dev should try to use their own software blindfolded. It's profoundly humbling.

robertknight:
Hi @judell - There was a misunderstanding here. I meant to use _headings_ (ie. `<hN>`) rather than a header for the search bucket contents. Otherwise you get something very similar to the screenshot I posted above. The use of the heading level is important as readers often provide quick links to skip between headers at the same level.

I had another look over it this morning, done another round of testing and pushed those as commits on this branch. Along the way I fixed a few other things:

- I added the "aria-expanded" state to the document title link so that A.T.s can indicate the expanded/collapsed state of the bucket when the title is focused
- The top level list structure of search results was wrong. All items in a bucket ended up being grouped into a single list item so the screen reader could not announce the number of results currently on screen.
- Use of section elements inside annotation cards wasn't really appropriate. Sections should have headers and they should be something that would make sense in an outline of the page. Compare https://gsnedders.html5.org/outliner/process.py?url=https%3A%2F%2Fhypothes.is%2Fsearch vs. https://gsnedders.html5.org/outliner/process.py?url=https%3A%2F%2Fbbc.co.uk%2Fnews

judell:
Thx @robertknight!

Now I'm wondering about title vs (or in addition to) aria-label, e.g.:

<h4 title="tags" 

<h4 title="tags" aria-label="tags"

<h4 aria-label="tags"

PR #4559, opened 2017-05-15
Make dropdown menus in navbar accessible

### LINKED ISSUE #4555: Navbar dropdown menus are not accessible
The dropdown menus in the navbar as seen at eg. https://hypothes.is/search are not accessible. When testing with VoiceOver in OS X 10.12 I encountered the following issues:

1. The "Settings" menu link is read as "down pointing small triangle"
1. There is no indication in VoiceOver that the Groups or Settings menus are popup menu triggers
1. When the menu is opened, the contents of the menu are not automatically focused, so navigating to the menu is difficult
1. Navigating the contents of the Groups and Settings menus does not announce the contents of the popup as a menu item

For examples of accessible dropdown menus, see the navbars in apps such as Gmail & Facebook.

robertknight:
This PR is a collection of accessibility fixes for the navigation menu in the top-right corner of the navbar:

- Use a `<nav>` element for the navigation links container so that they are grouped together and marked as a landmark when assistive technologies (eg. screen readers) traverse the page.
- Use `role` attributes to indicate the menu-ness of the dropdown menus and the current state of the menu (expanded/collapsed) to A.T.s
- Provide an accessible name for the "Settings" menu. I had assumed that the screen reader would read the title for the link but it didn't, instead VoiceOver at least read out the unicode dropdown arrow symbol as "dropdown arrow".

Fixes #4555

----

**Testing**:

1. Familiarise yourself with VoiceOver / Orca / Other Screen Reader enough to do basic navigation of web pages. VoiceOver has a built-in tutorial.
2. Try to navigate to the navbar links (username / Groups menu / Settings menu) using the screen reader, open and close the menus and select menu items. Compare output with master branch.

### DIFF
--- a/h/static/scripts/tests/controllers/dropdown-menu-controller-test.js
+++ b/h/static/scripts/tests/controllers/dropdown-menu-controller-test.js
@@ -5,7 +5,7 @@ const util = require('./util');
 
 const TEMPLATE = [
   '<div class="js-dropdown-menu">',
-  '<span data-ref="dropdownMenuToggle">Toggle</span>',
+  '<a href="#" data-ref="dropdownMenuToggle">Toggle</a>',
   '<span data-ref="dropdownMenuContent">Menu</span>',
   '</div>',
 ].join('\n');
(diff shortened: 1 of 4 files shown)


h/static/scripts/tests/controllers/dropdown-menu-controller-test.js
```diff
@@ -29,21 +29,32 @@ describe('DropdownMenuController', () => {
     return menuEl.classList.contains('is-open');
   }
 
+  function clickMenuLink() {
+    toggleEl.dispatchEvent(new Event('click', { cancelable: true }));
```
nickstenning: Perhaps this is ok, and I recognise that you're mainly just extracting this into a common helper function here rather than actually writing this code, but direct use of `dispatchEvent` to simulate user actions always makes me nervous. Don't we use [`syn`](https://www.npmjs.com/package/syn) for stuff like this?

robertknight: In most code we do. I think these tests were written before we started using syn.

PR #4560, opened 2017-05-15
Make search autocomplete accessible

### LINKED ISSUE #4556: Navbar search box autocomplete not well described by screen reader
The autocomplete box that appears when focusing the search box and typing eg. "tag:" _can_ be navigated to with a screen reader, but the experience is poor because the reader has no information about the relation between the box and the search bar. Testing with VoiceOver in OS X 10.12 I saw the following issues:

1. No indication that the search menu has a dropdown associated with it.
1. No indication that the items in the dropdown are menu or list items.

For an example of an autocomplete built with accessibility in mind, see https://github.com/alphagov/accessible-autocomplete . Also give the search inputs in the navbar of Google or Facebook a try with a screen reader.

robertknight:
This adds some basic ARIA markup to make the navbar search autocomplete popup accessible. This doesn't cover all the accessibility information that might be useful, but it at least makes the search box more "visible" when using the screen reader in a way that is similar to how the autocomplete list "appears" on eg. Google.

 * Mark the search input as a combobox with a popup. This is similar to
   how the search field is marked up on eg. Google.

 * Use the `listbox` and `option` roles for the listbox and menu items
   respectively

Fixes #4556 

----

Testing:

- When the search box is focused, the screen reader should announce it as a combobox and indicate that it has menu options associated with it. On Mac you'll be told to press VO+Space to show items, but this does nothing. That is expected and matches the behaviour on eg. Google and Facebook.
-  The suggestions in the dropdown menu should be grouped as together as a "list box" which you can navigate into or out of using the screen reader.

### DIFF
--- a/h/templates/panels/navbar.html.jinja2
+++ b/h/templates/panels/navbar.html.jinja2
@@ -34,7 +34,8 @@
       <form class="search-bar"
             data-ref="searchBarForm"
             id="search-bar"
-            action="{{ search_url }}">
+            action="{{ search_url }}"
+            role="search">
 
         {#- This seemingly pointless <input> is actually needed to make the
             form work correctly. If you remove this input then when the user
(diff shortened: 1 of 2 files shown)


h/templates/panels/navbar.html.jinja2
```diff
@@ -53,11 +54,15 @@
           {% endif %}
 
           <input class="search-bar__input js-input-autofocus"
+                 aria-autocomplete="list"
+                 aria-label="Search annotations"
```
nickstenning: This should be a translation string, I guess? `{{ _('Search annotations') }}`

Issue #4564, opened 2017-05-18
Profile API returns the same publisher group twice


seanh:
Steps to reproduce:

1. Follow the instructions at https://github.com/hypothesis/publisher-account-test-site
   to get the publisher account site setup
2. Login to the publisher test site as "admin" (the publisher group
   creator)

Result: You'll see the same Partner.org publisher group listed twice in
the groups dropdown menu in the client:

![screenshot from 2017-05-06 22-56-31](https://cloud.githubusercontent.com/assets/22498/25776179/85c736a2-32af-11e7-88c8-aedcc119bb3a.png)

Expected: You should only see the group once.

(This is a profile API bug, not a client bug.)

See also https://github.com/hypothesis/h/pull/4553

seanh:
I've taken the liberty of adding this issue to up next in the current sprint, so that I (or someone else) will remember to fix it when we get round to it. Since this only happens to the (one) publisher group admin it doesn't seem like a huge problem, but it is a user-visible bug.

PR #4565, opened 2017-05-22
Don't add publisher group creators as members

seanh:
Fixes https://github.com/hypothesis/h/issues/4564

### DIFF
--- a/h/services/group.py
+++ b/h/services/group.py
@@ -56,6 +56,8 @@ class GroupService(object):
                       creator=creator,
                       description=description)
 
+        group.members.append(creator)
+
         access_flags = GROUP_ACCESS_FLAGS.get(type_)
         if access_flags is None:
             raise ValueError('Invalid group type %s' % type_)
--- a/h/services/group.py
+++ b/h/services/group.py
@@ -56,13 +56,14 @@ class GroupService(object):
                       creator=creator,
                       description=description)
 
-        group.members.append(creator)
-
         access_flags = GROUP_ACCESS_FLAGS.get(type_)
         if access_flags is None:
             raise ValueError('Invalid group type %s' % type_)
         for attr, value in access_flags.iteritems():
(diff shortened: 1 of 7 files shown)


h/services/group.py
```diff
@@ -62,6 +62,9 @@ def create(self, name, authority, userid, description=None, type_='private'):
         for attr, value in access_flags.iteritems():
             setattr(group, attr, value)
 
+        if group.joinable_by is not None:
+            group.members.append(creator)
```
seanh: As I understand it `joinable_by` is something like "example.com", but there doesn't appear to be permissions code for asking "If this group joinable by this user" yet (otherwise I could have done `if group.is_joinable_by(creator)`) so I went with this. Alternatively this could also be `if type_ != publisher`.

chdorner: I think this is better than `if type_ != 'publisher'` because it's using the underlying logic and not the abstraction level that is the group types.

tests/h/views/activity_test.py
```diff
@@ -927,6 +927,11 @@ def group(factories):
     factories.Group()
 
     group = factories.Group()
+
+    # Add the group's creator as a member, as would normally be the case for
+    # (non-publisher) groups.
+    group.members.append(group.creator)
+
```
chdorner: Would it make sense to re-structure the group factories? We could get rid of the default `factories.Group` factory and create a `factories.PrivateGroup` which adds the creator to the members inside the factory, so we don't need to care about this in the tests themselves.

seanh: Ok, I've moved this into the `Group` factory. Do you think it needs to be renamed to `factories.PrivateGroup` as well?

chdorner: Probably not right now. We can look at renaming this when we add a new group type with its own factory. Thanks!

PR #4567, opened 2017-05-24
Pull preloading logic out of formatters

fatbusinessman:
This is still very much work in progress, and I’m unlikely to find time this week to finish up the last bits on the `AnnotationJSONPresentationService` and its tests, but feedback on the general approach is very welcome.


fatbusinessman:
I’m going to close this for now. We can resurrect this if and when we next change the formatters code, depending on whether it makes those changes easier.

PR #4568, opened 2017-05-25
Upgrade pyramid_authsanity to 1.0.0

nickstenning:
[Our fix](https://github.com/usingnamespace/pyramid_authsanity/pull/7) for an esoteric IE bug (where we'd set a correctly signed cookie instead of removing a cookie) is in the latest release of this package.

There are a couple of small changes to the generated requirements.txt that aren't directly related to this upgrade, but they all look harmless to me and are mostly the result of us acquiring direct dependencies on packages that were previously transitive dependencies (such as transaction, venusian and zope.interface).

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,7 +4,6 @@
 #
 #    pip-compile --output-file requirements.txt requirements.in
 #
-
 alembic==0.8.7
 amqp==1.4.9               # via kombu
 anyjson==0.3.3            # via kombu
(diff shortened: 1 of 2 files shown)


requirements.txt
```diff
@@ -70,14 +69,15 @@ requests==2.13.0          # via requests-aws4auth
...
 sqlalchemy==1.1.4
 statsd==3.2.1
-transaction==2.1.2        # via pyramid-mailer, pyramid-tm, repoze.sendmail, zope.sqlalchemy
+transaction==2.1.2
 translationstring==1.3    # via colander, deform, pyramid
 unidecode==0.4.19         # via python-slugify
 urllib3==1.16             # via elasticsearch
-venusian==1.0             # via pyramid
+venusian==1.0
+webencodings==0.5.1       # via html5lib
```
fatbusinessman: This one line appears to be a new addition; not quite sure what’s going on there.

nickstenning: Yep. While we had a `git+https://...` URL in here we had to edit this file manually, so I think we must have missed that `html5lib` got a new dependency when we upgraded it in b90a536aa.

PR #4572, opened 2017-06-05
Sort feature flags by name in admin UI

robertknight:
Sort feature flag list at /admin/features by name. This makes it a bit
easier to skim the list if there are several active flags.


robertknight:
> I was wondering whether it would be sensible to add sorting in at the model layer

I did consider this but depending on the transformations done at the view/presentation layer, this ordering can get lost. For example, the API layer sticks the feature names & flag states into a dict and renders that to JSON, but normal Python dicts don't preserve insertion order.

PR #4573, opened 2017-06-13
use tooltip to announce actions available behind timestamp link

judell:
Until we can more fully integrate the functions of the standalone /a/ID card, which is linked from the timestamp, whose tooltip currently just says `date`, this PR adjusts the tooltip to indicate that the timestamp's link is the gateway to functions (reply, edit, delete, flag) available there.

Background: A user accidentally annotated a Gmail page, found the annotation using /search, but was unable to find it in the sidebar. I don't know exactly why not. Group confusion? Orphan confusion? In any case, the direct path to the standalone annotation card is non-obvious and a more informative tooltip can make it slightly less so. Again, as an interim solution until the /search card can include those functions.

### DIFF
--- a/h/templates/includes/annotation_card.html.jinja2
+++ b/h/templates/includes/annotation_card.html.jinja2
@@ -20,7 +20,7 @@
         class="annotation-card__username">
         {{ presented_annotation.annotation.username }}
       </a>
-      <a title="date" href="{{ request.route_url('annotation', id=presented_annotation.annotation.id) }}"
+      <a title="reply, edit, delete, flag" href="{{ request.route_url('annotation', id=presented_annotation.annotation.id) }}"
         class="annotation-card__timestamp">
         {{ presented_annotation.annotation.updated.strftime('%d %b %Y') }}
       </a>


nickstenning:
Hmmm. Sorry for the delay in reviewing this at all. I've been a bit out of the loop.

We could certainly merge this, although it feels like a bit of a hack to me. My concern would only be that the point at which we start explicitly acknowledging that some of these features aren't available in the web service is the point at which people new to the team will start thinking that's a deliberate design choice rather than a limitation based on time constraints.

judell:
@nickstenning, @dawariley Maybe we can just think about it this way. The current title text, `date`, only recapitulates the link's label, it doesn't describe what the link does. And in fact, here's the full breakdown on the card:

| text  | title | link |
| ---- | ---- | ---- |
| VictorVenema | username | https://hypothes.is/users/VictorVenema |
| 12 Jul 2017 | date  |  https://hypothes.is/a/75HncmdCEeePCl-6BILhIw |
| in Public  |  group  |  https://hypothes.is/groups/__world__/public |
| reply icon  | Visit link in context  |  https://hyp.is/75HncmdCEeePCl-6BILhIw/nymag.com/daily/intelligencer/2017/07/climate-change-earth-too-hot-for-humans.html  |
| share icon  | share this annotation  | https://hypothes.is/search#  |

Arguably a more consistent approach would be:

| text  | title | link |
| ---- | ---- | ---- |
| VictorVenema | Visit user page | https://hypothes.is/users/VictorVenema |
| 12 Jul 2017 | Visit annotation to  reply, edit, delete, or flag |  https://hypothes.is/a/75HncmdCEeePCl-6BILhIw |
| in Public  |  Visit group page  |  https://hypothes.is/groups/__world__/public |
| reply icon  | Visit link in context  |  https://hyp.is/75HncmdCEeePCl-6BILhIw/nymag.com/daily/intelligencer/2017/07/climate-change-earth-too-hot-for-humans.html  |
| Share icon  | share this annotation  | https://hypothes.is/search#  |

dawariley:
Hi Jon, I need more context. As I understand there is a the standalone card: https://hypothes.is/a/75HncmdCEeePCl-6BILhIw and we want to introduce a tooltip from a time stamp? According to Nick we don't have much data to show in the tooltip due to implementation constraints. Can you please send a link or screenshot of the tooltip if it exists and what UX input you are needing exactly. Thanks. Dawa

judell:
@dawariley the tooltips on a standalone annotation page like https://hypothes.is/a/75HncmdCEeePCl-6BILhIw are a separate issue (i.e. could be improved, but not the subject of this PR).

This one is about the tooltips on the annotation cards at https://hypothes.is/search, you can look at any one of them, the two tables above compare the existing situation to the proposed situation.

I began with a suggestion that the date link on the /search card could, with a more informative tooltip, better indicate what the link actually does, namely take you to the /a/ID page where you can reply, edit, delete, or flag.

The second table proposes that, while we're at it, we can make all the tooltips consistent in describing the actions of the links they decorate.

nickstenning:
I'd be very happy with a PR that makes the changes you've proposed in the table above.

dawariley:
I've reviewed the issues outlined with Jon and here's what I think we need to do in order to improve this from a UX perspective.

Improving the tooltips to inform users that they need to click on date in order to navigate to a standalone card view where they can: reply, edit, delete or flag a card isn't the correct way to solve this and it's part of a bigger UX issue. Below is my outline of the problem as I understand it and the proposed solution to improve the UX.

Problem as I understand it: At the moment in order to access eply, edit, delete or flag buttons by drilling down to a second level standalone card view by clicking on the date. The way to get to this functionality is obscure and not easily discoverable. Second problem: no back button to allow users to get back up once they drill down to the standalone card view.

Proposed solution: Move reply, edit, delete or flag buttons to top level, remove access to the second card view: https://hypothes.is/a/5LzCLmvmEeeKXNc42wAa-w as this will become unnecessary when the edit/moderate/share buttons are moved to the top level card view.

judell:
@dawa your analysis is correct. However since the inner level is happening in one system, a client JS environment that's (more or less) common with the sidebar, and the outer level is happening another system, the server-based Python environment that delivers Activity Pages, it's not a simple thing to move those buttons. And this isn't important enough to spend more effort on now, vs many other things.

So my counter-proposals are:

1.  Forget about the original motivation, and just change the tooltips so that they more consistently and accurately reflect the actions on the existing links.

or

2. Drop the whole thing for now, and I'll close the issue.

dawariley:
I am leaning more towwards #2 - de-linking the date and adding another affordance to go to the standalone card. I do feel that otherwise we're delivering a broken experience as we discussed at the F2F, it might be better to disable it until it is implemented in the way users would expect.

judell:
Related: https://github.com/hypothesis/product-backlog/issues/378

seanh:
I'm gonna close this pull request for now, as we don't have time to prioritise this feature in the short term. The code will still be here for us when it's a better time to come back to this.

judell:
> it might be better to disable it until it is implemented in the way users would expect.

This user would like the existing affordance to have been more discoverable, but is happy to have it and has used it effectively.

https://twitter.com/ppival/status/954456133980143616

image

PR #4579, opened 2017-06-21
Use the pipeline library to orchestrate Jenkins build

nickstenning:
This commit mirrors similar changes made to the bouncer and via repository Jenkinsfiles. It:

- removes some of the logic associated with calculating build versions and building Docker images and instead uses functions provided by [our pipeline library][1]

- adds build stages to trigger deployment to QA and Prod environments as part of the build pipeline

This should be safe to deploy straight away, as it shouldn't affect Skyliner builds. Once it's merged, changes will start going to QA in Elastic Beanstalk automatically.

[1]: http://github.com/hypothesis/pipeline-library

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,73 +1,71 @@
 #!groovy
 
+@Library('pipeline-library') _
+
+def img
+
 node {
-    // -------------------------------------------------------------------------
-    stage 'Build'
+    stage('build') {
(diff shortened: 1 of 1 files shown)


robertknight:
LGTM from my limited knowledge of Groovy and what I gather from looking at the code for the pipeline library.

PR #4581, opened 2017-06-26
Add view-switcher feature flag

seanh:
See https://github.com/hypothesis/client/pull/465

### DIFF
--- a/h/models/feature.py
+++ b/h/models/feature.py
@@ -21,7 +21,9 @@ FEATURES = {
     'orphans_tab': "Show the orphans tab to separate anchored and unanchored annotations?",
     'search_for_doi': "Use DOI metadata when searching for annotations on the current page?",
     'total_shared_annotations': "Show the total number of shared annotations for users and groups?",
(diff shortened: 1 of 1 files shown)


h/models/feature.py
```diff
@@ -21,7 +21,9 @@
     'orphans_tab': "Show the orphans tab to separate anchored and unanchored annotations?",
     'search_for_doi': "Use DOI metadata when searching for annotations on the current page?",
     'total_shared_annotations': "Show the total number of shared annotations for users and groups?",
-    'overlay_highlighter': "Use the new overlay highlighter?"
+    'overlay_highlighter': "Use the new overlay highlighter?",
+    'view-switcher': 'Use the new-style [Annotations|Page Notes|Orphans] view switcher control at the top of the sidebar',
```
robertknight: Pedantic comment: The other flag names use underscores rather than hyphens. Can you do likewise?

seanh: Well spotted. If I change this I'll have to change the client PR that uses it as well though, and it's just a temporary feature flag...

PR #4582, opened 2017-06-27
Add feature flag to append cachebuster to client URL

robertknight:
Add a feature flag which causes a cachebuster to be appended to the
URL of the client's boot script when serving the embed. When enabled for a request, the client URL is returned as "https://cdn.hypothes.is/hypothesis?cachebuster=$TIMESTAMP".

This provides a way to bypass the CDN cache on a per-cohort basis (eg.
temporarily enable for a dev after a client release for testing), or on
a per-request basis (eg. create a bookmarklet which serves
`.../embed.js?__feature__[embed_cachebuster]` and sets the sidebar app
URL to `.../app.html?__feature__[embed_cachebuster]`).

There are other ways we could potentially do this, see [this issue](https://github.com/hypothesis/product-backlog/issues/331) for some ideas. I'm open to alternative suggestions.

### DIFF
--- a/h/models/feature.py
+++ b/h/models/feature.py
@@ -13,6 +13,8 @@ log = logging.getLogger(__name__)
 FEATURES = {
     'defer_realtime_updates': ("Require a user action before applying real-time"
                                " updates to annotations in the client?"),
+    'embed_cachebuster': ("Cache-bust client entry point URL to prevent browser/CDN from "
+                          "using a cached version?"),
     'filter_highlights': ("Filter highlights in document based on visible"
                           " annotations in sidebar?"),
     'flag_action': ("Enable user to flag inappropriate annotations in the "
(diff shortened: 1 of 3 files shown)


seanh:
Hmm, I guess feature flags are supposed to be temporary and this would be a permanent one. It seems more like a per-user config setting than a feature flag, except that we don't only want the setting to be visible to certain users. On the other hand even though this doesn't seem like quite what feature flags are supposed to be for, it looks like this would work just fine.

seanh:
@robertknight , @sean-roberts and I discussed this after standup - this merge this now, and we'll keep [the issue](https://github.com/hypothesis/product-backlog/issues/331) open as a place to discuss alternatives that we might move to one day

seanh:
We should update the release docs. I'm intending to do a release tomorrow now, so I can do that then

PR #4583, opened 2017-06-27
Expose OAuth feature flag & client ID to client

robertknight:
This adds a facility to expose an OAuth client ID to the client in the configuration for app.html and a feature flag indicating whether to use it or not. The feature flag is separate so we can turn it on/off, including allowing the client to override it, without needing to change the service environment.

Enabling it to be overridden it for testing purposes in client configuration will require a small client change (https://github.com/hypothesis/client/pull/470).

For convenient testing in development and production, the last commit in this PR configures the client to use OAuth when visiting `/docs/help` with the "client_oauth" flag enabled, which can be done using the existing `__feature__[<name>]` query param support. In other words, given a client with the change in https://github.com/hypothesis/client/pull/470, visiting `/docs/help?__feature__[client_oauth]` will load an embedded client configured to use OAuth.

### DIFF
--- a/h/config.py
+++ b/h/config.py
@@ -70,9 +70,19 @@ SETTINGS = [
                       message='use the AUTHORITY environment variable instead'),
     EnvSetting('h.authority', 'AUTHORITY'),
     EnvSetting('h.bouncer_url', 'BOUNCER_URL'),
+
+    # ID and secret used for the issuer when signing JWT tokens for legacy API
+    # auth.
     EnvSetting('h.client_id', 'CLIENT_ID'),
     EnvSetting('h.client_secret', 'CLIENT_SECRET'),
+
     EnvSetting('h.client_url', 'CLIENT_URL'),
+
+    # ID for the OAuth authclient that the embedded client should use when
+    # making requests to OAuth endpoints. As a public client, it does not have a
+    # secret.
+    EnvSetting('h.client_oauth_id', 'CLIENT_OAUTH_ID'),
+
     # Environment name, provided by the deployment environment. Please do
     # *not* toggle functionality based on this value. It is intended as a
     # label only.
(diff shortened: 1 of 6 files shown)


h/config.py
```diff
@@ -70,9 +70,19 @@
...
+    # auth.
     EnvSetting('h.client_id', 'CLIENT_ID'),
     EnvSetting('h.client_secret', 'CLIENT_SECRET'),
+
     EnvSetting('h.client_url', 'CLIENT_URL'),
+
+    # ID for the OAuth authclient that the embedded client should use when
+    # making requests to OAuth endpoints. As a public client, it does not have a
+    # secret.
+    EnvSetting('h.client_oauth_id', 'CLIENT_OAUTH_ID'),
```
seanh: I wonder if we should add the new env var here? http://h.readthedocs.io/en/latest/developing/envvars/ Of course most h env vars are not documented there but they probably should be

h/views/client.py
```diff
@@ -49,6 +49,16 @@ def sidebar_app(request, extra=None):
...
         'apiUrl': request.route_url('api.index'),
         'authDomain': request.authority,
+        'oauthClientId': settings.get('h.client_oauth_id'),
+
+        # The OAuth feature flag is included as part of the `app.html` config
+        # rather than being delivered via the "features" key in /api/profile so
+        # that it is available as soon as the client starts before
+        # API tokens are fetched.
+        #
+        # This should be removed once OAuth for first-party accounts is shipped.
```
seanh: I'd put `# TODO` here (has been done with other comments like this in h, you can grep it)

robertknight:
> That might work on any page and also with the Chrome ext, and might not require changes to help.html.jinja2 or a server-side feature flag

The flag is needed anyway so we can flip the feature on/off for everyone. For testing on other pages, the idea is that you would create a custom bookmarklet which sets the "oauthEnabled" property in the config it injects into the page, similar to how the default bookmarklet sets the "showHighlights" flag.

What I wanted to avoid was adding yet another way of passing config to the client 😉

PR #4586, opened 2017-06-29
Initial implementation of an OAuth authorization endpoint

robertknight:
Issue: https://github.com/hypothesis/product-backlog/issues/310
Corresponding client PR: https://github.com/hypothesis/client/pull/476

Add an `/oauth/authorize` page which allows a user to authorize an
OAuth client, such as the Hypothesis client, to access their data on the
service. When the client is authorized, the endpoint returns an auth
code to the client via `window.postMessage`.

This is an initial minimal implementation with several limitations:

 * The only supported response mode is "web_message", which is designed
   to be convenient for web app clients to use. In particular, the
   standard redirect response mode is not supported.

 * It only allows one client, the sidebar application served from the
   `/app.html` route, to use this endpoint. Once the OAuth client model
   has been extended to include an "origin" property, any registered
   client can be used providing that the request comes from the correct
   origin.

 * The "auth code" is in fact, just a JWT grant token of the same type
   that publishers can generate when they embed Hypothesis on their
   pages and have registered with us to use third party accounts.

### DIFF
--- a/h/services/oauth.py
+++ b/h/services/oauth.py
@@ -149,6 +149,25 @@ class OAuthService(object):
 
         return token
 
+    def create_grant_token(self, user, authclient):
+        """
+        Generate a JWT bearer grant token for a user.
+
+        :param user: The user to generate the token for.
+        :type user: h.models.User
+        :param authclient: The OAuth client that is going to use the token.
+        :type authclient: h.models.AuthClient
+        """
+        now = datetime.datetime.utcnow()
+        claims = {
+            'aud': self.domain,
+            'iss': authclient.id,
+            'sub': 'acct:{}@{}'.format(user.username, user.authority),
+            'nbf': now,
+            'exp': now + datetime.timedelta(minutes=5),
+        }
+        return jwt.encode(claims, authclient.secret, algorithm='HS256')
+
 
 class GrantToken(object):
     """
--- a/h/services/oauth.py
+++ b/h/services/oauth.py
@@ -75,7 +75,7 @@ class OAuthService(object):
 
         token = GrantToken(body['assertion'])
 
-        authclient = self._get_authclient_by_id(token.issuer)
+        authclient = self.get_authclient_by_id(token.issuer)
         if not authclient:
             raise OAuthTokenError('grant token issuer (iss) is invalid', 'invalid_grant')
 
@@ -93,7 +93,12 @@ class OAuthService(object):
 
         return (user, authclient)
 
-    def _get_authclient_by_id(self, client_id):
+    def get_authclient_by_id(self, client_id):
+        """
+        Return the AuthClient with the given ID.
+
+        Returns `False` if no such client exists.
+        """
         try:
             return self.session.query(models.AuthClient).get(client_id)
         except sa.exc.StatementError as exc:
--- a/h/services/oauth.py
+++ b/h/services/oauth.py
@@ -162,7 +162,7 @@ class OAuthService(object):
         claims = {
             'aud': self.domain,
             'iss': authclient.id,
-            'sub': 'acct:{}@{}'.format(user.username, user.authority),
+            'sub': user.userid,
             'nbf': now,
             'exp': now + datetime.timedelta(minutes=5),
         }
--- a/h/services/oauth.py
+++ b/h/services/oauth.py
@@ -97,7 +97,7 @@ class OAuthService(object):
         """
         Return the AuthClient with the given ID.
 
-        Returns `False` if no such client exists.
+        Returns `None` if no such client exists.
         """
         try:
             return self.session.query(models.AuthClient).get(client_id)
--- a/h/static/scripts/controllers/authorize-form-controller.js
+++ b/h/static/scripts/controllers/authorize-form-controller.js
@@ -1,0 +1,60 @@
+'use strict';
+
+const Controller = require('../base/controller');
+
+/**
+ * Controller for the OAuth authorization popup.
+ */
+class AuthorizeFormController extends Controller {
+  constructor(element) {
+    super(element);
+
+    this.on('submit', () => {
+      // Prevent multiple submission or clicking the "Cancel" button after
+      // clicking "Accept".
+      this.setState({ submitting: true });
+    });
+
+    this.refs.cancelBtn.addEventListener('click', () => {
+      window.close();
+    });
+
+    window.addEventListener('beforeunload', () => {
+      this._sendAuthCanceledMessage();
+    });
+  }
+
+  /**
+   * Notify the parent window that auth was canceled.
+   *
+   * This is necessary since there isn't a DOM event that a cross-origin opener
+   * window can listen for to know when a popup window is closed.
+   */
+  _sendAuthCanceledMessage() {
+    if (this.state.submitting) {
+      // User already pressed "Accept" button.
+      return;
+    }
+
+    if (window.opener) {
+      let state;
+      if (this.refs.stateInput) {
+        state = this.refs.stateInput.value;
+      }
+
+      // Since this message contains no private data, just set the origin to
+      // '*' (any).
+      window.opener.postMessage({
+        type: 'authorization_canceled',
+        state,
+      }, '*');
+    }
+  }
+
+  update() {
+    this.refs.cancelBtn.disabled = !this.state.submitting;
+    this.refs.acceptBtn.disabled = !this.state.submitting;
+  }
+}
+
+module.exports = AuthorizeFormController;
--- a/h/static/scripts/post-auth.js
+++ b/h/static/scripts/post-auth.js
@@ -1,0 +1,24 @@
+'use strict';
+
+/**
+ * Script which runs in the small HTML page served by the OAuth Authorization
+ * endpoint after successful authorization.
+ *
+ * It communicates the auth code back to the web app which initiated
+ * authorization.
+ */
+
+const settings = require('./base/settings')(document);
+
+if (window.opener) {
+  const msg = {
+    type: 'authorization_response',
+    code: settings.code,
+    state: settings.state,
+  };
+  window.opener.postMessage(msg, settings.origin);
+  window.close();
+} else {
+  console.error('The opening window has closed');
+}
+
--- a/h/views/api_auth.py
+++ b/h/views/api_auth.py
@@ -84,6 +84,10 @@ class OAuthAuthorizeController(object):
         if not authclient:
             raise HTTPBadRequest('Unknown client ID "{}"'.format(client_id))
 
+        if authclient.authority != self.request.authority:
+            raise HTTPBadRequest('Client "{}" not allowed to authorize "{}" users'.format(
+                                 client_id, self.request.authority))
+
         response_type = params.get('response_type')
         if response_type != 'code':
             raise HTTPBadRequest('Unsupported response type "{}"'
--- a/h/views/api_auth.py
+++ b/h/views/api_auth.py
@@ -2,9 +2,98 @@
 
 from __future__ import unicode_literals
 
+from pyramid.httpexceptions import HTTPBadRequest, HTTPFound
+from pyramid.view import view_config, view_defaults
+from pyramid import security
+
 from h.exceptions import OAuthTokenError
 from h.util.view import cors_json_view
 from h.util.datetime import utc_iso8601
+
+
+@view_defaults(route_name='oauth_authorize')
+class OAuthAuthorizeController(object):
+
+    def __init__(self, request):
+        self.request = request
+        self.oauth_svc = self.request.find_service(name='oauth')
+        self.user_svc = self.request.find_service(name='user')
+
+    @view_config(request_method='GET',
+                 renderer='h:templates/oauth/authorize.html.jinja2')
+    def get(self):
+        """
+        Check the user's authentication status and present the authorization
+        page.
+        """
+        self._check_params()
+
+        if self.request.authenticated_userid is None:
+            raise HTTPFound(self.request.route_url('login', _query={
+                              'next': self.request.url}))
+
+        params = self.request.params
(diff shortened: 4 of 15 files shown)


h/views/api_auth.py
```diff
@@ -2,11 +2,111 @@
...
+
+        params = self.request.params
+        user = self.user_svc.fetch(self.request.authenticated_userid)
+
+        return {'username': user.username,
+                'client_name': 'Hypothesis',
+                'client_id': params['client_id'],
+                'response_type': params['response_type'],
+                'response_mode': params['response_mode'],
+                'origin': params.get('origin'),
```
seanh: I'm wondering what origin is supposed to represent here and what it's going to be used for? Looks like if comes from the form post params (so anyone could set it to anything)

h/views/api_auth.py
```diff
@@ -2,11 +2,111 @@
...
+            origin = params.get('origin')
+            if origin is None:
+                raise HTTPBadRequest('"origin" must be specified when response_mode is "web_message"')
+
+            # The 'origin' property has not yet been implemented for auth
+            # clients, so we only accept requests from the sidebar app served at
+            # /app.html by the current service.
+            #
+            # TODO: Once authclients have an "origin" property, check that
+            #       instead.
```
seanh: Check that instead of `params['origin']`, or instead of `request.host_url`? Again I'm not totally clear on what's going on with `origin` in this PR

h/views/client.py
```diff
@@ -49,6 +49,9 @@ def sidebar_app(request, extra=None):
     app_config = {
         'apiUrl': request.route_url('api.index'),
         'authDomain': request.authority,
+
+        # OAuth config.
+        'oauthAuthorizeUrl': request.route_url('oauth_authorize'),
```
seanh: I wonder if this should go in [the links endpoint](https://github.com/hypothesis/h/blob/master/h/views/api.py#L153)? https://hypothes.is/api/links

h/services/oauth.py
```diff
@@ -149,6 +149,25 @@ def create_token(self, user, authclient):
...
+        """
+        now = datetime.datetime.utcnow()
+        claims = {
+            'aud': self.domain,
+            'iss': authclient.id,
+            'sub': 'acct:{}@{}'.format(user.username, user.authority),
+            'nbf': now,
+            'exp': now + datetime.timedelta(minutes=5),
+        }
+        return jwt.encode(claims, authclient.secret, algorithm='HS256')
```
robertknight: Right. Note that the final implementation will most likely not use JWTs but instead a one-time authorisation code. I'm re-using them at the moment because it lets us get the front-end part working before the model changes & migrations are done.

h/static/scripts/controllers/authorize-form-controller.js
```diff
@@ -0,0 +1,60 @@
...
+    window.addEventListener('beforeunload', () => {
+      this._sendAuthCanceledMessage();
+    });
+  }
+
+  /**
+   * Notify the parent window that auth was canceled.
+   *
+   * This is necessary since there isn't a DOM event that a cross-origin opener
+   * window can listen for to know when a popup window is closed.
```
robertknight: Yes. It will be whichever `Window` initiated the auth process.

h/static/scripts/post-auth.js
```diff
@@ -0,0 +1,24 @@
+'use strict';
+
+/**
+ * Script which runs in the small HTML page served by the OAuth Authorization
+ * endpoint after successful authorization.
+ *
+ * It communicates the auth code back to the web app which initiated
+ * authorization.
```
robertknight: Correct.

h/views/api_auth.py
robertknight: We might not need this any more. OAuth clients have redirect URIs or origins associated with them when they are registered. _If_ the service allows a client to have multiple redirect URIs/origins associated with it then this tells the service which specific URI or origin to send the response back to. What we're currently planning is that initially we'll only allow one redirect URI or origin for each auth client, in which case this param is redundant because we can determine it from the `client_id`.

h/views/api_auth.py
robertknight: The reason for using `request.host_url` here is that in this initial PR, we are only allowing OAuth clients hosted at the same URL as the service, because the "origin" property is not yet implemented so this endpoint cannot yet verify that the origin of the request is associated with whatever arbitrary origin was specified when the auth client was created.

Does this make sense?

h/views/client.py
robertknight: Ideally, yes.

However, there is currently an awkward circular dependency in the client that this URL is needed before authorization, but fetching `/api/links` in the client depends on first completing authorization. The solution I think will be to rework the client so that it can explicitly make certain requests without authenticating.

h/services/oauth.py
```diff
@@ -149,6 +149,25 @@ def create_token(self, user, authclient):
...
+        :param user: The user to generate the token for.
+        :type user: h.models.User
+        :param authclient: The OAuth client that is going to use the token.
+        :type authclient: h.models.AuthClient
+        """
+        now = datetime.datetime.utcnow()
+        claims = {
+            'aud': self.domain,
+            'iss': authclient.id,
+            'sub': 'acct:{}@{}'.format(user.username, user.authority),
```
robertknight: Thanks for the tip!

robertknight:
I think this is ready for another look @seanh . The main change since last time is the addition of tests for the `/oauth/authorize` endpoints in api_auth_test.py.

robertknight:
I think this is ready for another look @seanh . The main change since last time is the addition of tests for the `/oauth/authorize` endpoints in api_auth_test.py.

You'll need the latest version of the client PR (https://github.com/hypothesis/client/pull/476) to test this end-to-end.

tests/h/views/api_auth_test.py
```diff
@@ -15,6 +17,106 @@
...
+
+    @pytest.fixture
+    def oauth_service(self, pyramid_config, pyramid_request):
+        svc = mock.Mock(spec_set=oauth_service_factory(None, pyramid_request))
+        pyramid_config.register_service(svc, name='oauth')
+        return svc
+
+    @pytest.fixture
+    def user_service(self, pyramid_config, pyramid_request):
+        svc = mock.Mock(spec_set=user_service_factory(None, pyramid_request))
```
seanh: Would [`create_autospec`](http://www.voidspace.org.uk/python/mock/helpers.html?highlight=create_autospec#mock.create_autospec) work here? (And also for `oauth_service` above.) Something like:

```python
svc = mock.create_autospec(h.services.user.UserService, instance=True)
```

I think the differences would be that:

1. `create_autospec` recursively mocks the `UserService` class's methods as well so that the tests will crash if the code tries to call a method with the wrong number of arguments, or with a keyword argument that doesn't exist. (`spec_set` prevents calling a method that doesn't exist, but it doesn't prevent passing args that don't exist.)

2. You won't need to call `user_service_factory()` (rather, just pass the `UserService` class to `create_autospec()`), so you won't need to pass in the `pyramid_request` object. Simplifies the fixture code but also the passing `pyramid_request` into `oauth_service_factory()` here is a little misleading because, unless you look very closely, it looks as if the mock `OAuthService` that we're creating actually depends on that particular `pyramid_request` obj that's passed into it (e.g. whether that request is a GET or POST, what params it has), but I don't think it actually does depend on it.

3. You won't be instantiating a real `UserService` object, so these `OAuthAuthorizeController` unit tests won't crash because of bugs in `UserService`'s `__init__()` method.

If I understand your code correctly you're calling `user_service_factory()` in order to instantiate a real `UserService` object so you can pass that real `UserService` object to `spec_set` in order to create a mock user service object to use in the tests. And you need a `pyramid_request` object (with `.authority and `.db`) as  argument because the real `user_service_factory()` code will crash without it.

h/views/api_auth.py
```diff
@@ -2,11 +2,100 @@
...
+        response_type = params.get('response_type')
+        if response_type != 'code':
+            raise HTTPBadRequest('Unsupported response type "{}"'
+                                 .format(response_type))
+
+        response_mode = params.get('response_mode', 'query')
+        if response_mode != 'web_message':
+            raise HTTPBadRequest('Unsupported response mode "{}"'.format(response_mode))
+
+        return authclient
```
seanh: I wondered if `_check_params()` should just return the checked params (`response_mode` and `response_type`) because the methods that call `_check_params()` then end up having to get these out of the `request` again anyway

h/views/api_auth.py
```diff
@@ -2,11 +2,100 @@
...
+        if self.request.authenticated_userid is None:
+            raise HTTPFound(self.request.route_url('login', _query={
+                              'next': self.request.url}))
+
+        params = self.request.params
+        user = self.user_svc.fetch(self.request.authenticated_userid)
+
+        return {'username': user.username,
+                'client_name': 'Hypothesis',
+                'client_id': params['client_id'],
```
seanh: Can `client_id` be retrieved from the `authclient` obj that `_check_params()` returns instead?

h/views/api_auth.py
```diff
@@ -2,11 +2,100 @@
...
+        if not authclient:
+            raise HTTPBadRequest('Unknown client ID "{}"'.format(client_id))
+
+        response_type = params.get('response_type')
+        if response_type != 'code':
+            raise HTTPBadRequest('Unsupported response type "{}"'
+                                 .format(response_type))
+
+        response_mode = params.get('response_mode', 'query')
+        if response_mode != 'web_message':
```
seanh: It struck me as odd that `response_mode` defaults to `'query'` but then raises if it's not `'web_message'`. The `params.get('response_mode', 'query')` could be `params.get('response_mode')` with the same behaviour

h/services/oauth.py
```diff
@@ -93,7 +93,12 @@ def _verify_jwt_bearer(self, body):
 
         return (user, authclient)
 
-    def _get_authclient_by_id(self, client_id):
+    def get_authclient_by_id(self, client_id):
+        """
+        Return the AuthClient with the given ID.
+
+        Returns `False` if no such client exists.
```
seanh: Should be `None` I think?

tests/h/views/api_auth_test.py
```diff
@@ -79,7 +79,10 @@ def test_post_returns_state(self, post_auth_ctrl, pyramid_request):
         assert ctx['state'] == 'a_random_string'
 
     @pytest.fixture
-    def auth_ctrl(self, factories, pyramid_config, pyramid_request):
+    def auth_ctrl(self, pyramid_config, pyramid_request):
+        """
+        Configure a valid request for `OAuthAuthorizeController.get`.
```
seanh: This method actually returns a new `OAuthAuthorizeController` configured with a valid `GET` request, doesn't it? A `"Return a new ..."` docstring might be clearer.

h/views/api_auth.py
```diff
@@ -2,11 +2,100 @@
...
+        Depending on the "response_mode" parameter the auth code will be
+        delivered either via a redirect or via a `postMessage` call to the
+        opening window.
+        """
+        authclient = self._check_params()
+
+        user = self.user_svc.fetch(self.request.authenticated_userid)
+
+        # Create an "authorization code" for the response.
+        # This is in fact just a JWT grant token since auth codes have not yet
```
nickstenning: This is the only part about this that makes me raise an eyebrow slightly. 

Using JWTs as authorization codes isn't fundamentally problematic, but only if you actually support using them in that way. I'm guessing that the client is going to use the JWT grant type for the access token request, and not a `grant_type=code` request? We probably don't want people to look at that and think it's a supported way of authorizing clients using OAuth...

robertknight: My plan was that use of JWTs would be temporary, until auth codes are implemented, and that both here and in the client there would be comments making that clear. eg: https://github.com/hypothesis/client/blob/3f0416fed84e494f11e11a70f050bcab4a7e7e32/src/sidebar/oauth-auth.js#L107

Perhaps catch up about this on Slack?

h/services/oauth.py
robertknight: Yes, indeed.

h/views/api_auth.py
robertknight: The default post-authorization behaviour, [according to the spec](https://tools.ietf.org/html/rfc6749#section-4.1.2) should be to redirect to the pre-registered redirect URI associated with the client with the auth code and state in the query string.

The redirect flow isn't actually implemented yet, but I've set the default response_mode to "query" so it will be possible to use the auth endpoint without explicitly setting response_mode once the redirect flow is implemented.

tests/h/views/api_auth_test.py
```diff
@@ -15,6 +17,106 @@
 from h.views import api_auth as views
 
 
+@pytest.mark.usefixtures('routes', 'user_service', 'oauth_service')
+class TestOAuthAuthorizeController(object):
+    def test_get_verifies_client_id(self, auth_ctrl, pyramid_request,
```
robertknight: I've renamed and more importantly, also added checks that the authority of the client matches `request.authority`.

tests/h/views/api_auth_test.py
robertknight: Thanks @seanh - I didn't know about `create_autospec`. The tests worked fine using `create_autospec` but I started a discussion in #backend to clarify the reason for not using it originally.

h/views/api_auth.py
robertknight: I could, although I'm not sure it would save any code at the moment. I'll keep this in mind for future refactoring though.

robertknight:
Thanks for the reviews. I'm going to merge this in its current state but I'd like to catch up on Slack afterwards to confirm how we'll proceed with auth codes and JWTs.

judell:
This is a huge step forward!

PR #4584, opened 2017-06-27
return /api/search results as jsonld

judell:
# overview

A step in the direction of a more complete implementation of W3C-standard output.

Currently we support, e.g, https://hypothes.is/api/annotations/Dms34EbzEeebxbfPD5_8Rw.jsonld as the standard alternative to https://hypothes.is/api/annotations/Dms34EbzEeebxbfPD5_8Rw.

This change invokes the jsonld presenter for search results when an API query includes `&jsonld=true`

# sample output

```
{
	"rows": [{
		"body": [{
			"type": "TextualBody",
			"value": "",
			"format": "text/markdown"
		}, {
			"type": "TextualBody",
			"purpose": "tagging",
			"value": "FrontGroup"
		}, {
			"type": "TextualBody",
			"purpose": "tagging",
			"value": "SourceWatch"
		}, {
			"type": "TextualBody",
			"purpose": "tagging",
			"value": "Capital Research Center"
		}],
		"target": [{
			"source": "https://www.indianagazette.com/news/opinions/commentary-inside-the-liberal-dystopia,26312437/"
		}],
		"created": "2017-05-30T20:13:24.861874+00:00",
		"@context": "http://www.w3.org/ns/anno.jsonld",
		"creator": "acct:jon@localhost",
		"type": "Annotation",
		"id": "http://localhost:5000/a/b53cFEV0Eeev9cP7e7gJAw",
		"modified": "2017-05-30T20:13:24.861874+00:00"
	}, {
		"body": [{
			"type": "TextualBody",
			"value": "This may be a front group. [Investigate](http://www.sourcewatch.org/index.php/Capital_Research_Center), find additional sources, and leave research notes in the comments.",
			"format": "text/markdown"
		}, {
			"type": "TextualBody",
			"purpose": "tagging",
			"value": "FrontGroup"
		}, {
			"type": "TextualBody",
			"purpose": "tagging",
			"value": "SourceWatch"
		}, {
			"type": "TextualBody",
			"purpose": "tagging",
			"value": "Capital Research Center"
		}],
		"target": [{
			"source": "https://www.indianagazette.com/news/opinions/commentary-inside-the-liberal-dystopia,26312437/"
		}],
		"created": "2017-05-30T20:11:07.459662+00:00",
		"@context": "http://www.w3.org/ns/anno.jsonld",
		"creator": "acct:jon@localhost",
		"type": "Annotation",
		"id": "http://localhost:5000/a/HbPi4EV0Eeev9RdU_F__SQ",
[...]

PR #4587, opened 2017-07-07
Return API search results as JSONLD

judell:
Redo of https://github.com/hypothesis/h/pull/4584, conforming to our endpoint.extension style.

### DIFF
--- a/h/services/annotation_json_presentation.py
+++ b/h/services/annotation_json_presentation.py
@@ -13,7 +13,7 @@ from h.interfaces import IGroupService
 
 
 class AnnotationJSONPresentationService(object):
-    def __init__(self, session, user, group_svc, links_svc, flag_svc, flag_count_svc, 
+    def __init__(self, session, user, group_svc, links_svc, flag_svc, flag_count_svc,
                  moderation_svc, has_permission, json_flavor=None):
         self.session = session
         self.group_svc = group_svc
--- a/h/services/annotation_json_presentation.py
+++ b/h/services/annotation_json_presentation.py
@@ -13,10 +13,12 @@ from h.interfaces import IGroupService
 
 
 class AnnotationJSONPresentationService(object):
-    def __init__(self, session, user, group_svc, links_svc, flag_svc, flag_count_svc, moderation_svc, has_permission):
+    def __init__(self, session, user, group_svc, links_svc, flag_svc, flag_count_svc, 
+                 moderation_svc, has_permission, json_flavor=None):
         self.session = session
         self.group_svc = group_svc
         self.links_svc = links_svc
+        self.json_flavor = json_flavor
 
         def moderator_check(group):
             return has_permission('admin', group)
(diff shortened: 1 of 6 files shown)


h/services/annotation_json_presentation.py
```diff
@@ -48,16 +50,15 @@ def eager_load_documents(query):
                 for ann in annotations]
 
     def _get_presenter(self, annotation_resource):
-        return presenters.AnnotationJSONPresenter(annotation_resource,
-                                                  self.formatters)
+        if self.json_flavor == 'jsonld':
+            return presenters.AnnotationJSONLDPresenter(annotation_resource)
```
chdorner: This branch needs to be tested, we do that for the normal JSON presenter [here](https://github.com/hypothesis/h/blob/2f28db72cf8cb6764af6ddda067c1d1cd8b06751/tests/h/services/annotation_json_presentation_test.py#L44-L47) and [here](https://github.com/hypothesis/h/blob/2f28db72cf8cb6764af6ddda067c1d1cd8b06751/tests/h/services/annotation_json_presentation_test.py#L57-L62).

h/services/annotation_json_presentation.py
```diff
@@ -66,3 +67,25 @@ def annotation_json_presentation_service_factory(context, request):
...
+                                             json_flavor='jsonld')
+
+
+def _find_services(request):
+    group_svc = request.find_service(IGroupService)
+    links_svc = request.find_service(name='links')
+    flag_svc = request.find_service(name='flag')
+    flag_count_svc = request.find_service(name='flag_count')
+    moderation_svc = request.find_service(name='annotation_moderation')
+    return (group_svc, links_svc, flag_svc, flag_count_svc, moderation_svc)
```
chdorner: This is not so much your fault, as the JSON presentation service really just relies on too many other services. But it could be made a bit nicer by returning a map instead of a tuple because then the ordering doesn't matter when we try to get one or some services out of it.

h/views/api.py
```diff
@@ -175,28 +175,16 @@ def links(context, request):
 
 @api_config(route_name='api.search',
             link_name='search',
-            description='Search for annotations')
+            description='Search for annotations, return legacy json')
```
chdorner: I don't think this is necessary. We don't have any concrete plans as of yet to have the client use the jsonld format vs our own, so our own JSON format is not "legacy" by any means.

h/views/api.py
```diff
@@ -304,6 +292,29 @@ def _json_payload(request):
         raise PayloadError()
 
 
+def _call_search_lib(request, json_presenter=None):
```
chdorner: In my personal opinion, this would be a nicer being called `_execute_search`, `_search_annotations`, or something along those lines. But that's my personal preference, so feel free to ignore me.

chdorner: The `json_presenter` argument should not have a default value because the code in this function does not support a `None` value (`find_service(name=None)` will throw an exception).

h/views/api.py
```diff
@@ -175,28 +175,16 @@ def links(context, request):
...
-
-    if separate_replies:
-        out['replies'] = svc.present_all(result.reply_ids)
 
-    return out
+@api_config(route_name='api.search.jsonld',
+            link_name='search.jsonld',
+            description='Search for annotations, return jsonld')
+def search_jsonld(request):
+    return _call_search_lib(request, 'annotation_jsonld_presentation')
```
chdorner: This view is not being tested at all. Maybe we can do something clever with parametrizing the `TestSearch` class to run the same set of tests on both the `search` and `search_jsonld` view.

fatbusinessman:
This PR hasn’t had any activity in a few months, so I’m going to close it for now.

PR #4589, opened 2017-07-10
Use the default pyramid_mailer debug code

nickstenning:
Since https://github.com/hypothesis/h/commit/3e900fec77, it seems like our debug mailer was broken. Sorry! The interface for IMailer changed and I didn't notice. This seems like a great reason to not let people who aren't working with the code every day do dependency upgrades... =P

Anyway, this commit fixes the issue by getting rid of `h/debug.py` entirely and using the debug mailer which ships with pyramid_mailer instead. This debug mailer doesn't attempt to write emails to the console, and instead writes them as individual files to `$PWD/mail/`.

This might not be obvious to someone new to the codebase, so I've also made a few tweaks (to the Celery worker log level when run with `hypothesis devserver`, and to what gets logged by the mailer task) to point people to this directory when mail is sent in debug (i.e. development) mode.


h/cli/commands/devserver.py
```diff
@@ -80,7 +80,7 @@ def devserver(https, web, ws, worker, assets, beat):
         m.add_process('ws', 'gunicorn --name websocket --reload --paste conf/development-websocket.ini %s' % gunicorn_args)
 
     if worker:
-        m.add_process('worker', 'hypothesis --dev celery worker --autoreload')
+        m.add_process('worker', 'hypothesis --dev celery worker --autoreload -l INFO')
```
robertknight: Will this cause additional debug messages to be printed from other packages' logging code?

PR #4592, opened 2017-07-13
Database migrations for supporting OAuth 2 authorization code grant flow

chdorner:
This adds:

* New `authzcode` table which represents an OAuth 2 authorization code
* Adds a `redirect_uri` column to the `authclient` table, which is needed for the authorization code flow

### DIFF
--- a/h/migrations/versions/0d1b3fd8807c_create_authzcode_table.py
+++ b/h/migrations/versions/0d1b3fd8807c_create_authzcode_table.py
@@ -1,0 +1,36 @@
+"""
+Create authzcode table
+
+Revision ID: 0d1b3fd8807c
+Revises: b980b1a8f6af
+Create Date: 2017-07-10 12:04:28.328864
+"""
+
+from __future__ import unicode_literals
+
+import sqlalchemy as sa
+from alembic import op
+from sqlalchemy.dialects import postgresql
+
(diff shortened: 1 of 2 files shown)


h/migrations/versions/0d1b3fd8807c_create_authzcode_table.py
```diff
@@ -0,0 +1,36 @@
...
+        sa.Column('created', sa.DateTime(), server_default=sa.func.now(), nullable=False),
+        sa.Column('updated', sa.DateTime(), server_default=sa.func.now(), nullable=False),
+        sa.Column('expires', sa.DateTime(), nullable=False),
+        sa.Column('code', sa.UnicodeText, nullable=False),
+        sa.Column('authclient_id', postgresql.UUID(), nullable=False),
+        sa.Column('user_id', sa.Integer(), nullable=False),
+        sa.ForeignKeyConstraint(['user_id'], ['user.id']),
+        sa.ForeignKeyConstraint(['authclient_id'], ['authclient.id']),
+    )
+    op.create_index(op.f('ix__authzcode_code'), 'authzcode', ['code'], unique=True)
```
nickstenning: Tiny thing, but I think according to our naming convention this should be `uq__authzcode__code`.

PR #4593, opened 2017-07-17
Implement first part of authorization code flow

chdorner:
This creates two new endpoints (or changes the old, test endpoints) with
the help of oauthlib. It currently implements the standard authorization
endpoint described in [section 3.1 in RFC 6749][rfc-authz-endpoint].

Note: This only implements the necessary views to create an authorization code, there is no functionality to exchange the authorization code for an access token yet. Which will need some more work to make our JWT bearer exchange work with oauthlib.

`GET /oauth/authorize`
----------------------

This endpoint validates the data in the request parameters, typically a
`client_id`, `redirect_uri`, and `response_type`. These parameters are
checked againts the permissions stored on the `AuthClient` model.
When the request is valid it shows an authorize screen to the logged-in
user with information about the OAuth client requesting access to the
users' data.

`POST /oauth/authorize`
-----------------------

This endpoint is called when the user accepts the authorize screen that
is rendered in the previous endpoint.
The endpoint checks again that the requested data is consistent with
the permissions stored in the `AuthClient` model, and creates an
authorization code with a default set of scopes (`annotation:read`,
`annotation:write`), and then redirects the browser to the client's
preferred `redirect_uri` with the authorization code.

[rfc-authz-endpoint]: https://tools.ietf.org/html/rfc6749#section-3.1


tests/h/services/oauth_validator_test.py
```diff
@@ -0,0 +1,176 @@
...
+
+class TestFindClient(object):
+    def test_returns_client(self, svc, client):
+        assert svc.find_client(client.id) == client
+
+    def test_returns_none_for_invalid_client_id(self, svc):
+        assert svc.find_client('bogus') is None
+
+    def test_returns_none_when_not_found(self, svc, client):
+        id_ = unicode(uuid.uuid1())
```
chdorner: Yes yes yes Hound!

h/celery.py
```diff
@@ -42,6 +42,10 @@
             'task': 'h.tasks.cleanup.purge_expired_auth_tickets',
             'schedule': timedelta(hours=1)
         },
+        'purge-expired-authzcodes': {
+            'task': 'h.tasks.cleanup.purge_expired_authz_codes',
+            'schedule': timedelta(hours=1)
+        },
```
nickstenning: Just noting here that we need to update `h-periodic` with this, too.

h/models/__init__.py
```diff
@@ -39,6 +40,7 @@
     'Annotation',
     'AnnotationModeration',
     'AuthClient',
+    'AuthzCode',
```
nickstenning: Because I'm annoying, I will observe that this is one line too high :)

h/views/api_auth.py
```diff
@@ -2,102 +2,88 @@
...
-
-        Depending on the "response_mode" parameter the auth code will be
-        delivered either via a redirect or via a `postMessage` call to the
-        opening window.
-        """
-        authclient = self._check_params()
-
+        # We don't support scopes at the moment, but oauthlib does need a scope,
+        # so we're adding a hard-coded `annotation:read` and `annotation:write` scopes.
+        scopes = ['annotation:read', 'annotation:write']
```
nickstenning: Is this different from the return value of `get_default_scopes` in the request validator? Is there a way we can define default scopes just once?

h/views/api_auth.py
```diff
@@ -2,102 +2,88 @@
...
-        auth_code = self.oauth_svc.create_grant_token(user, authclient)
-
-        params = self.request.params
-
-        return {'code': auth_code,
-                # Once authclients have an Origin property, that should be used
-                # instead here.
-                'origin': self.request.host_url,
-                'state': params.get('state')}
+        headers, body, status = self.oauth.create_authorization_response(
```
nickstenning: Looks to me like we throw away `body` here, so I'd probably be inclined to do `headers, _, status = ...` to make that clear.

h/views/api_auth.py
```diff
@@ -2,102 +2,88 @@
...
+                self.request.url, scopes=scopes, credentials=credentials)
 
-    def _check_params(self):
-        """
-        Check parameters for the authorization request.
+        if 'Location' in headers:
+            return exception_response(status, location=headers['Location'])
```
nickstenning: I had to read this a few times before I realised that this is the "happy path," too. I think a comment explaining the use of `exception_response` here would probably help. Alternatively, it seems like we know we're going to be doing a redirect, so we could just `raise HTTPFound(location=...)` ourselves?

h/views/api_auth.py
```diff
@@ -2,102 +2,88 @@
...
+        else:
+            log.error('Created OAuth 2 authorization code, but got no redirect location')
```
nickstenning: Given that it seems like this should be a truly exceptional circumstance, it would make more sense to me if this were a `try: ... except: ...`, e.g.

```python
try:
    return exception_response(status, location=headers['Location'])
except KeyError:
    raise RuntimeError('created authorisation code for client "{}" but got no redirect location'.format(...)) 
```

Note that this way you'll get some useful debugging information in Sentry. `raise HTTPInternalServerError()` is going to result in a very bizarre looking error report.

h/views/api_auth.py
```diff
@@ -2,102 +2,88 @@
...
-            raise HTTPBadRequest('Unsupported response mode "{}"'.format(response_mode))
+@view_defaults(renderer='h:templates/oauth/error.html.jinja2')
+class OAuthAuthorizeErrorController(object):
+    def __init__(self, context, request):
+        self.context = context
+        self.request = request
 
-        return authclient
+    @view_config(context=InvalidRequestFatalError)
+    @view_config(context=InvalidRequestError)
```
nickstenning: Did you work out why `MissingResponseTypeError` isn't a subclass of `InvalidRequestFatalError`!?

h/celery.py
chdorner: Not so fast, I wanted to it after this is out. But since you insist: hypothesis/h-periodic#7 :stuck_out_tongue_winking_eye:

h/models/__init__.py
chdorner: Oh no! I missed that one. I specifically went over every changed file and ran `isort`, but I didn't look at this tuple!

h/views/api_auth.py
chdorner: Good point, I'm now reading it from the service.

h/views/api_auth.py
chdorner: Yeah, that's nicer. Thanks!

h/views/api_auth.py
chdorner: You're right, I changed it to `HTTPFound` which should make look less like it's the unhappy path.

h/views/api_auth.py
chdorner: I couldn't find anything. According to the standard, it is required and the request must fail when it's missing (see: https://hyp.is/74MpXmsLEee4vhcFDgNKQA/tools.ietf.org/html/rfc6749).

My best guess is that the oauthlib maintainers missed this line when they reviewed the PR which introduced this (https://github.com/idan/oauthlib/pull/358).

chdorner:
I've addressed your comments as fixup commits for easier reviewing. I will rebase this branch when you're happy with the fixes.

nickstenning:
Changes LGTM :+1:

chdorner:
Rebased, will merge as soon as the tests pass.

PR #4599, opened 2017-07-18
Allow auth clients without a secret

chdorner:
We will have support for both public and confidential OAuth clients (definition is [here](https://tools.ietf.org/html/rfc6749#section-2.1)). The best and easiest way to distinguish between these types is the presence of a client secret because a public OAuth client will never need a secret.

This should not be merged until #4598 is merged and the migration has been run on production.

### DIFF
--- a/h/cli/commands/authclient.py
+++ b/h/cli/commands/authclient.py
@@ -3,6 +3,7 @@
 import click
 
 from h import models
+from h.security import token_urlsafe
 
 
 @click.group()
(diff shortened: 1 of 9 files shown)


h/cli/commands/authclient.py
```diff
@@ -13,14 +14,18 @@ def authclient():
...
+def add(ctx, name, authority, type_):
     """
     Create a new OAuth client.
     """
+
     request = ctx.obj['bootstrap']()
 
     authclient = models.AuthClient(name=name, authority=authority)
+    if type_ == 'confidential':
+        authclient.secret = token_urlsafe()
```
chdorner: This would make more sense in a service, but since we already manually create the auth client in here and it's the only place where we currently create auth clients, I've opted for not pulling this out into a new service.

chdorner:
The migration ran on production, this has been rebased, and the tests finally pass. This is now ready for review.

PR #4601, opened 2017-07-20
Move business logic around Tokens into services

chdorner:
This is in preparation for replacing our custom OAuth implementation with one that uses oauthlib.

This moves developer token business logic:
* from model into new developer token service
* from accounts view into the new developer token service

And OAuth token business logic:
* from model into the existing OAuth service

### DIFF
--- a/h/services/oauth.py
+++ b/h/services/oauth.py
@@ -8,11 +8,15 @@ import jwt
 import sqlalchemy as sa
 
 from h import models
+from h import security
 from h.exceptions import OAuthTokenError
 from h._compat import text_type
 
 # TTL of an OAuth token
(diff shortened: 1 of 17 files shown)


h/services/oauth.py
```diff
@@ -148,7 +149,9 @@ def create_token(self, user, authclient):
         :rtype: h.models.Token
         """
         token = models.Token(userid=user.userid,
+                             value=security.token_urlsafe(),
```
nickstenning: I think it would be a good idea to have a prefix on this token type too so we can distinguish one random string of characters from another when debugging. Perhaps also on the refresh token.

PR #4602, opened 2017-07-20
Refactor JWT OAuth token exchange to use oauthlib

chdorner:
This change removes our custom OAuth implementation with one that is based on oauthlib. It still only works for the JWT bearer token exchange and refresh token exchange.

Broadly speaking this is:

1. adding a new grant type (`h.oauth.jwt_grant`) in the form oauthlib understands (responding to `create_token_response`and `validate_token_request`),
2. implements refresh token flow methods that oauthlib calls in the `h.services.oauth_validator` service,
3. replacing the class that tied everything together (`oauthlib.oauth2.WebApplicationServer`) with our own implementation (`h.services.oauth_provider`) that configures oauthlib the way we use OAuth in `h` (sets token generators, configures the jwt-bearer with the oauthlib `TokenEndpoint` endpoint),
4. replaces the `/api/token` view with one that uses oauthlib,
5. and removes the old oauth services which is not used anymore.


tests/h/services/oauth_provider_test.py
```diff
@@ -0,0 +1,53 @@
+# -*- coding: utf-8 -*-
+
+from __future__ import unicode_literals
+
+import mock
+import pytest
+
+from oauthlib.common import Request as OAuthRequest
```
chdorner: This is now actually being used in these tests, so ignore this comment.

chdorner:
This has been split out into four individual PRs:

* #4603 
* #4604 
* #4605 
* #4606

PR #4603, opened 2017-07-21
Add supporting code for using JWT bearer token exchange with oauthlib

chdorner:

This is adding a new grant type (`h.oauth.jwt_grant`) in the form oauthlib understands (responding to `create_token_response` and `validate_token_request`).

_This is split out from #4602._

### DIFF
--- a/h/oauth/errors.py
+++ b/h/oauth/errors.py
@@ -1,0 +1,21 @@
+# -*- coding: utf-8 -*-
+
+from __future__ import unicode_literals
+
+from oauthlib.oauth2 import InvalidGrantError
+
+
+class MissingJWTGrantTokenClaimError(InvalidGrantError):
+    def __init__(self, claim, claim_description=None):
+        if claim_description:
+            self.description = "Missing claim '{}' ({}) from grant token.".format(claim, claim_description)
+        else:
+            self.description = "Missing claim '{}' from grant token.".format(claim)
+
+
+class InvalidJWTGrantTokenClaimError(InvalidGrantError):
+    def __init__(self, claim, claim_description=None):
+        if claim_description:
+            self.description = "Invalid claim '{}' ({}) in grant token.".format(claim, claim_description)
+        else:
+            self.description = "Invalid claim '{}' in grant token.".format(claim)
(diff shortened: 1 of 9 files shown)


h/oauth/jwt_grant.py
```diff
@@ -0,0 +1,64 @@
...
+from h.oauth.jwt_grant_token import JWTGrantToken
+
+
+class JWTAuthorizationGrant(GrantTypeBase):
+    def __init__(self, request_validator, user_svc, domain):
+        self.request_validator = request_validator
+        self.user_svc = user_svc
+        self.domain = domain
+
+    def create_token_response(self, request, token_handler):
```
nickstenning: Here as elsewhere, I think it's probably important to document that `request` is an `oauthlib.common.Request`, otherwise the default assumption in this codebase is going to be that it's a `pyramid.request.Request`.

h/services/oauth_validator.py
```diff
@@ -84,6 +105,9 @@ def validate_scopes(self, client_id, scopes, request, *args, **kwargs):
         return (scopes == default_scopes)
 
     def _find_client(self, id_):
+        if id_ is None:
```
nickstenning: Is it worth testing the positive branch here?

h/oauth/errors.py
```diff
@@ -0,0 +1,21 @@
...
+
+from __future__ import unicode_literals
+
+from oauthlib.oauth2 import InvalidGrantError
+
+
+class MissingJWTGrantTokenClaimError(InvalidGrantError):
+    def __init__(self, claim, claim_description=None):
+        if claim_description:
+            self.description = 'Missing grant token {} ({}).'.format(claim_description, claim)
```
chdorner: Yep, I agree. I would like to leave the sentence structure though because it's how [oauthlib](https://github.com/idan/oauthlib/blob/master/oauthlib/oauth2/rfc6749/errors.py#L137) [error](https://github.com/idan/oauthlib/blob/master/oauthlib/oauth2/rfc6749/errors.py#L153) [messages](https://github.com/idan/oauthlib/blob/master/oauthlib/oauth2/rfc6749/errors.py#L149) [look](https://github.com/idan/oauthlib/blob/master/oauthlib/oauth2/rfc6749/errors.py#L145) [like](https://github.com/idan/oauthlib/blob/master/oauthlib/oauth2/rfc6749/errors.py#L141).

But I'll change it to `Missing claim 'iss' (issuer) from grant token`.

PR #4604, opened 2017-07-21
Support refreshing tokens in oauthlib integration

chdorner:

This implements the refresh token flow methods that oauthlib calls in the `h.services.oauth_validator` service.

_This is split out from #4602. It also depends on #4603 and needs to be rebased when that one is merged._

### DIFF
--- a/h/services/oauth_validator.py
+++ b/h/services/oauth_validator.py
@@ -8,6 +8,7 @@ from oauthlib.oauth2 import InvalidClientIdError, RequestValidator
 from sqlalchemy.exc import StatementError
 
 from h import models
+from h.models.auth_client import GrantType as AuthClientGrantType
 from h.util.db import lru_cache_in_transaction
 
 AUTHZ_CODE_TTL = datetime.timedelta(minutes=10)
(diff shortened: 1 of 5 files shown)


h/services/oauth_validator.py
```diff
@@ -21,20 +22,45 @@ class OAuthValidatorService(RequestValidator):
...
 
     def authenticate_client_id(self, client_id, request, *args, **kwargs):
         """Authenticates a client_id, returns True if the client_id exists."""
         client = self.find_client(client_id)
         request.client = client
         return (client is not None)
 
+    def client_authentication_required(self, request, *args, **kwargs):
+        """
+        Require client authentication for confidential clients.
```
nickstenning: I think this would be clearer as something like "Determine if client authentication is required for an access token request".

h/services/oauth_validator.py
```diff
@@ -89,6 +122,15 @@ def validate_redirect_uri(self, client_id, redirect_uri, request, *args, **kwarg
             return (client.redirect_uri == redirect_uri)
         return False
 
+    def validate_refresh_token(self, refresh_token, client, request, *args, **kwargs):
```
nickstenning: For security critical pieces like this, I think it would be good to ensure we document what exactly it is you're validating. For example:

```python
"""
Validate a supplied refresh token.

Check that the refresh token supplied with an access token request a) exists, 
b) has not expired, and c) is associated with the client identified in the request.
If we return True from this function, we can assume that it is safe to issue a 
new access token to the requesting client.
"""
```

tests/h/views/api_auth_test.py
```diff
@@ -24,7 +24,7 @@
 from h.views import api_auth as views
 
 
-@pytest.mark.usefixtures('routes', 'oauth_validator', 'user_svc')
+@pytest.mark.usefixtures('routes', 'user_svc', 'oauth_validator')
```
nickstenning: Did you mean to make this change?

chdorner: Sort of. The `oauth_validator` fixture calls the service factory which loads the user service. But I've changed it so that the `oauth_validator` fixture loads the `user_svc` fixture on its own and that the order of fixtures here is irrelevant.

PR #4605, opened 2017-07-21
Add our own OAuth provider class

chdorner:

This is replacing the class that tied everything together (`oauthlib.oauth2.WebApplicationServer`) with our own implementation (`h.services.oauth_provider`) that configures oauthlib the way we use OAuth in `h` (sets token generators, configures the jwt-bearer with the oauthlib `TokenEndpoint` endpoint).

_This is split out from #4602. It also depends on #4604 and needs to be rebased when that one is merged._

### DIFF
--- a/h/services/oauth_provider.py
+++ b/h/services/oauth_provider.py
@@ -1,0 +1,94 @@
+# -*- coding: utf-8 -*-
+
+from __future__ import unicode_literals
+
+import datetime
+
+from oauthlib.oauth2 import (
+    AuthorizationCodeGrant,
+    AuthorizationEndpoint,
+    BearerToken,
+    RefreshTokenGrant,
+    TokenEndpoint,
+)
+
+from h.oauth import InvalidRefreshTokenError, JWTAuthorizationGrant
+from h.security import token_urlsafe
+
+TOKEN_TTL = datetime.timedelta(hours=1).total_seconds()
+
+
+class OAuthProviderService(AuthorizationEndpoint, TokenEndpoint):
+    """
+    The OAuth 2 provider service.
+
+    This service subclasses both the oauthlib `authorization endpoint`_
+    and the `token endpoint`_. Its goal is to provide a complete
+    configuration on how to provide the necessary functionality of
+    an OAuth authorization server.
+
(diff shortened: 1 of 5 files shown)


h/services/oauth_provider.py
```diff
@@ -0,0 +1,83 @@
...
+                               },
+                               default_token_type=bearer)
+
+    def load_client_id_from_refresh_token(self, request):
+        """
+        Custom validator which sets the client_id from a given refresh token
+
+        For the refresh token flow, RFC 6749 states that public clients only need
+        to be verified when the `client_id` is provided. oauthlib seems to be
+        ignoring this and always expecds the `client_id` parameter.
```
nickstenning: expects

h/services/oauth_provider.py
```diff
@@ -0,0 +1,83 @@
...
+            return
+
+        token = self.oauth_validator.find_refresh_token(request.refresh_token)
+        if token:
+            request.client_id = token.authclient.id
+        else:
+            raise InvalidRefreshTokenError()
+
+    def generate_access_token(self, oauth_request):
+        return '5768-{}'.format(token_urlsafe())
```
nickstenning: Should these strings come from a named constant somewhere?

h/services/oauth_provider.py
```diff
@@ -0,0 +1,83 @@
...
+    TokenEndpoint,
+)
+
+from h.oauth import InvalidRefreshTokenError, JWTAuthorizationGrant
+from h.security import token_urlsafe
+
+TOKEN_TTL = datetime.timedelta(hours=1).total_seconds()
+
+
+class OAuthProviderService(AuthorizationEndpoint, TokenEndpoint):
```
nickstenning: I don't know about you, but I find I often miss parent classes and mixins when first scanning code. It might be a good idea to add a docstring here explaining what OAuthProviderService is for, and what AuthorizationEndpoint and TokenEndpoint to (or at least point to their docs).

h/services/oauth_provider.py
nickstenning: Ah, don't worry, I see you address this in #4606.

PR #4606, opened 2017-07-21
Replace OAuth token endpoint with new one using oauthlib

chdorner:

This replaces the `/api/token` view with one that uses oauthlib and removes the old oauth service which is not used anymore.

_This is split out from #4602. It also depends on #4605 and needs to be rebased when that one is merged._


tests/h/views/api_auth_test.py
```diff
@@ -126,12 +118,9 @@ def oauth_server(self, auth_client):
         status = 302
         oauth_server.return_value.create_authorization_response.return_value = (headers, body, status)
 
-        return oauth_server
+        svc = oauth_server(oauth_validator=mock.Mock(), user_svc=mock.Mock(), domain='example.com')
```
nickstenning: This seems to me to be a bit of a roundabout way of stubbing the service. Perhaps `svc = mock.create_autospec(OAuthProviderService, instance=True)` might be simpler?

PR #4608, opened 2017-07-24
Add migration for storing redirect_uri in authorization code

chdorner:
_This is based on #4607 and will need to be rebased._

RFC 6749 mandates that when the `redirect_uri` was provided in an
authorization request during the authorization code flow, it must
be provided again in the token request and the server needs to check
that the two match ([see here](https://hyp.is/ilraKnBdEeeC4ac2jJxr1g/tools.ietf.org/html/rfc6749)).


chdorner:
Instead of storing and checking the `redirect_uri` we're going to disable arbitrary redirect uris. A client needs to have its `redirect_uri` registered and will not be able to change it on a per-request basis. I'm closing this and will add a database constraint to #4609.

PR #4615, opened 2017-07-27
Admin views for managing OAuth clients

robertknight:
This PR adds a set of admin CRUD views for managing OAuth clients.

It adds an "OAuth clients" section in the admin navbar and views to list clients (`/admin/oauthclients`), create clients (`/admin/oauthclients/new`), edit clients (`/admin/oauthclients/{client_id}`) and delete clients (button at the bottom of the edit view).

**Implementation notes:**

* I've used the same infrastructure (deform, colander) to build the forms that we use for the public site, since this saves quite a lot of boilerplate code and makes validation easier. As a result, the forms look like the forms on our public site, rather than the Bootstrap-y style of other forms on the admin pages. Down the road this should also make it easier for us to create a self-serve portal for users to register third-party clients.
  * This required adding the JS & CSS components for our forms to the admin site (`admin.scss` and `admin-site.js`).
  * It also required adding the header JS bundle to the admin site (`header_js`) so that the progressive enhancement-related features of the components work properly.
* I used traversal to handle the edit view. In order to get permissions to work I made use of Pyramid's ACL inheritance feature so that the `AuthClient` used as context for the `/admin/oauthclients/{client_id}` view has an instance of `h.resources.Root` as its parent object. This means that all the admin-related permissions defined by `h.resources.Root` are available to the edit view. This necessitated defining the resource factory (`AuthClientFactory`) in `h.resources` instead of `h.models.auth_client` to avoid a circular dependency.

**TODO**
- [x] Tests
- [x] Either make the client ID and client secret fields read-only, or allow them to be edited.

**Not in this PR**
- Use a dropdown menu for enum-valued fields.
- Make the checkbox input for the "Trusted" field look pretty


h/oauth/schemas.py
```diff
@@ -0,0 +1,75 @@
...
+    values = []
+    for name, _ in enum_cls.__members__.items():
+        values.append(name)
+
+    return colander.SchemaNode(colander.String(),
+                               validator=colander.OneOf(values),
+                               **kwargs)
+
+
+class CreateAuthClientSchema(CSRFSchema):
```
chdorner: There are some fairly complicated rules about which fields need to be set in certain circumstances. We should probably build this into the form at some point, not sure if now is the best time.

Some that we currently have:
* `redirect_uri` is required when `grant_type` is `authorization_code` (ensured with a database check constraint, so this renders a 500 right now.)
* `response_type` is required for certain `grant_type` values, this is validated in the authorize view code and might render a `unsupported_response_type` error.

h/oauth/schemas.py
```diff
@@ -0,0 +1,75 @@
...
+
+    response_type = enum_schema_node(ResponseType,
+                                     title=_('Response type'),
+                                     hint=_('Specifies what kind of authorization response '
+                                            'is returned to the client'))
+
+    trusted = colander.SchemaNode(
+                colander.Boolean(),
+                title=_('Trusted'),
+                hint=_('Trusted clients do not require user approval'))
```
chdorner: There should be a big fat warning message here saying that virtually every authclient is untrusted.

h/oauth/schemas.py
```diff
@@ -0,0 +1,75 @@
...
+
+    :param enum_cls: The enum class
+    :type enum_cls: enum.Enum
+    """
+    values = []
+    for name, _ in enum_cls.__members__.items():
+        values.append(name)
+
+    return colander.SchemaNode(colander.String(),
+                               validator=colander.OneOf(values),
```
robertknight: This confused me earlier too. Fields are required unless a default is specified via a `missing` kwarg to the `SchemaNode` constructor.

h/resources.py
```diff
@@ -82,6 +84,28 @@ def __acl__(self):
...
 
+class AuthClientFactory(object):
+
+    def __init__(self, request):
+        self.request = request
+
+    def __getitem__(self, client_id):
+        try:
+            client = self.request.db.query(AuthClient) \
+                                    .filter_by(id=client_id).one()
```
robertknight: Fixed.

h/oauth/schemas.py
```diff
@@ -0,0 +1,75 @@
+# -*- coding: utf-8 -*-
```
robertknight: Done

h/oauth/schemas.py
robertknight: Fixed.

h/oauth/schemas.py
robertknight: This PR is getting quite large so, as long as the constraints we want are enforced in the DB, I think it might be better if I add more user-friendly validation in a follow-up PR?

h/oauth/schemas.py
robertknight: Fixed. These fields may now be left empty.

h/oauth/schemas.py
```diff
@@ -0,0 +1,75 @@
...
+             hint=_('This will be displayed to users in the '
+                    'authorization prompt'))
+
+    authority = colander.SchemaNode(
+                  colander.String(),
+                  title=_('Authority'),
+                  hint=_('Set of users whose data this client '
+                         'can interact with'))
+
+    grant_type = enum_schema_node(GrantType,
```
robertknight: What values do we actually support at present?

h/oauth/schemas.py
chdorner: Yep, that's fine in a follow-up PR.

h/oauth/schemas.py
chdorner: We currently support: `GrantType.authorization_code`, `GrantType.jwt_bearer`, `ResponseType.code`.

And as far as I know, we don't have any plans to support any of the other options.

robertknight:
I made some improvements since the last review:

- Documented allowed grant types in the hint
- Auto-generate a client secret when the grant type is "jwt_bearer"
- Remove the response type field and set it automatically based on the grant type
- Improved the hints of several fields

I think this is now good to go.

robertknight:
Rebased.

PR #4616, opened 2017-07-28
Add support for `web_message` in OAuth authorize endpoint

chdorner:
This adds the ability to provide the created authorization code via the
`web_message` method as described in
[draft-sakimura-oauth-wmrm-00][draft-sakimura-oauth-wmrm-00].

[draft-sakimura-oauth-wmrm-00]: https://tools.ietf.org/html/draft-sakimura-oauth-wmrm-00

### DIFF
--- a/h/views/api_auth.py
+++ b/h/views/api_auth.py
@@ -11,6 +11,7 @@ from pyramid.httpexceptions import HTTPFound, exception_response
 from pyramid.view import view_config, view_defaults
 
 from h import models
+from h._compat import urlparse
 from h.exceptions import OAuthTokenError
 from h.services.oauth_validator import DEFAULT_SCOPES
 from h.util.datetime import utc_iso8601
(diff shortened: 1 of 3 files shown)


h/views/api_auth.py
```diff
@@ -31,6 +32,43 @@ def __init__(self, context, request):
...
     def get(self):
+        return self._authorize()
+
+    @view_config(request_method='GET',
+                 request_param='response_mode=web_message',
+                 renderer='h:templates/oauth/authorize.html.jinja2')
+    def get_web_message(self):
+        response = self._authorize()
+
+        if isinstance(response, HTTPFound):
```
fatbusinessman: In both this method and `post_web_message`, it looks like we’re calling something that produces a redirect URL, wraps it up in an HTTP response and returns it, whereupon we *unwrap* it, extract the location, and just use that. I wonder, is there a different abstraction we could use here which wouldn’t require wrapping the location up in a request only to unwrapp it again?

fatbusinessman: To double-check my understanding: would I be correct in thinking that this depends on the fact that `_authorize` *returns* an `HTTPFound` in the case of a successful authorization, but *raises* one when the user isn’t logged in?

h/views/api_auth.py
```diff
@@ -49,26 +87,15 @@ def get(self):
 
         state = credentials.get('state')
         user = self.user_svc.fetch(self.request.authenticated_userid)
+        response_mode = credentials.get('request').response_mode
```
fatbusinessman: What’s `request` at this point? This is the `oauthlib` request, right, rather than anything to do with the HTTP request?

h/views/api_auth.py
```diff
@@ -85,6 +112,22 @@ def _authorized_response(self):
             client_id = self.request.params.get('client_id')
             raise RuntimeError('created authorisation code for client "{}" but got no redirect location'.format(client_id))
 
+    def _render_web_message_response(self, redirect_uri):
```
fatbusinessman: In a similar vein to the question above about wrapping and unwrapping HTTP requests, this method feels like it’s taking a bunch of information that’s been wrapped up into a URL, unwrapping it again to extract the `code`, `state` and `origin`, then using those to render the template. Is this something that’s forced upon us by the way `oauthlib` does things, or is there a way to pass these values around directly and only build them together into URLs at the point we need them?

h/views/api_auth.py
```diff
@@ -31,6 +32,43 @@ def __init__(self, context, request):
     @view_config(request_method='GET',
                  renderer='h:templates/oauth/authorize.html.jinja2')
     def get(self):
+        return self._authorize()
+
+    @view_config(request_method='GET',
+                 request_param='response_mode=web_message',
+                 renderer='h:templates/oauth/authorize.html.jinja2')
+    def get_web_message(self):
```
fatbusinessman: With this naming convention, I’m finding it easy to tell *when* each of these view methods gets invoked (for instance, when there’s a `GET` request with the `web_message` response mode), but I’m finding it hard to keep track of what the methods themselves are doing, particularly given that I’m a little hazy on what `_authorize` and `_authorized_response` are returning themselves). I don’t know whether the best solution to this is a different naming convention or a great big pile of docstrings, or whether it’s because I’m still a little unfamiliar with the details of the different flows.

h/views/api_auth.py
```diff
@@ -85,6 +112,22 @@ def _authorized_response(self):
...
+        params = urlparse.parse_qs(location.query)
+        origin = '{l.scheme}://{l.netloc}'.format(l=location)
+
+        state = None
+        states = params.get('state', [])
+        if states:
+            state = states[0]
+
+        return {
+            'code': params.get('code', [])[0],
```
fatbusinessman: Why does this code work differently to the code for the `state` parameter? Is it because we know `code` is going to be present, while `state` might not be?

robertknight: Indeed. `state` is recommended but optional according to https://tools.ietf.org/html/rfc6749#section-4.1.1

robertknight:
I tested this with the client and with the addition of https://github.com/hypothesis/client/pull/514 the login flow works end-to-end.

h/views/api_auth.py
robertknight: I think the thinking here is that the redirect response is part of the OAuth spec and the `web_message` response mode is "layered" on top of this as a non-standard addition (albeit inspired by the spec linked in the PR description) for the convenience of browser-based clients.

h/views/api_auth.py
```diff
@@ -31,6 +32,43 @@ def __init__(self, context, request):
     @view_config(request_method='GET',
                  renderer='h:templates/oauth/authorize.html.jinja2')
```
robertknight: Currently IIUC an unsupported `response_mode` value will result in the default redirect behaviour. We could be stricter and check that `response_mode` is one of `None` or `query` here (nb. if we ever add support for implicit grants that would require allowing `fragment` as well I think).

It's not that important IMO so I leave it up to you.

chdorner: I can't find any description of what should happen when the request includes an invalid response mode in any spec, so I would err on the side of the decisions that oauthlib made which is a fallback to the `query` response mode.

h/views/api_auth.py
chdorner: The `oauthlib` API is not designed to support the behaviour of `web_message`. It is designed to return a response object that should (theoretically) be directly rendered, which is why it returns headers, a body, and a status code.
They also don't have any support for hooking in different response modes. So the way that Nick and I ended up with was to use `oauthlib` as it is intended to be used, and then add additional functionality around it which handles the different behaviour when using `web_message`.
The only ugly part of this is that we need to parse the redirect URI to read out the authorization code, there is, unfortunately, no other way to get the authorization code. 

---

> To double-check my understanding: would I be correct in thinking that this depends on the fact that _authorize returns an HTTPFound in the case of a successful authorization, but raises one when the user isn’t logged in?

It is implemented in a way where it returns the successful response (`HTTPFound`), but raises an unexpected, error response.

h/views/api_auth.py
chdorner: Yes, this is an `oauthlib` request. I've tried to be clear about the naming to not cause confusion. This is, unfortunately, the return value of an oauthlib call, which uses `request`.

h/views/api_auth.py
chdorner: Same as above, `oauthlib` doesn't support `web_message` and its API is not designed to make this easy, the only way to get to the authorization code is to parse the redirect URI.

h/views/api_auth.py
chdorner: What Rob said :slightly_smiling_face:

h/views/api_auth.py
chdorner: I'll add a docstring which points out the difference between the views (`get` vs `get_web_message`, and `post` vs `post_web_message`)

PR #4617, opened 2017-07-31
Remove the view switcher feature flag

seanh:
See https://github.com/hypothesis/client/pull/513


robertknight:
It is a shame we weren't able to finish this but I completely appreciate that you and Dawa are busy with higher priority things for now.

PR #4618, opened 2017-08-01
Make feature flag deletions cascade

robertknight:
Remove feature cohort <-> feature associations automatically when the
corresponding feature flag is removed.

This fixes a problem where the task to purge old feature flags could fail if the feature was still enabled for one or more cohorts.


robertknight:
We'll presumably need to do a similar thing for feature cohorts in order to make deleting those easier too. I'll keep that for a separate PR.

PR #4619, opened 2017-08-01
Replace direct references to Alembic commands with our wrapper script

robertknight:
Replace direct uses of the `alembic` command with references to the
`hypothesis migrate` wrapper script which sets up the environment
before running alembic commands.


robertknight:
> Out of casual interest, is there any difference between the behaviour of ./bin/hypothesis and just bin/hypothesis?

No. I just hadn't thought about the fact that the `./` is not required, unlike when referencing a script in the current directory.

robertknight:
I'm ignoring the Jenkins failure because that's an error with webtest that was fixed in master.

PR #4620, opened 2017-08-02
Add token revocation endpoint

chdorner:
This adds a new `POST /oauth/revoke` endpoint which allows revoking tokens either by access token or by refresh token. It works according to [RFC 7009][RFC 7009].

[RFC 7009]: https://tools.ietf.org/html/rfc7009

### DIFF
--- a/h/routes.py
+++ b/h/routes.py
@@ -92,6 +92,7 @@ def includeme(config):
     config.add_route('badge', '/api/badge')
     config.add_route('token', '/api/token')
     config.add_route('oauth_authorize', '/oauth/authorize')
+    config.add_route('oauth_revoke', '/oauth/revoke')
 
     # Client
     config.add_route('session', '/app')
--- a/h/services/oauth_provider.py
+++ b/h/services/oauth_provider.py
@@ -9,6 +9,7 @@ from oauthlib.oauth2 import (
     AuthorizationEndpoint,
     BearerToken,
     RefreshTokenGrant,
+    RevocationEndpoint,
     TokenEndpoint,
 )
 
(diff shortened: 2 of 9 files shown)


robertknight:
I tested this with the client PR that implements logout and it works as expected (`/oauth/revoke` call succeeds, token ceases to work afterwards). I added the following additional change locally on this branch so that the client does not have to hardcode the OAuth revoke URL:

```
diff --git a/h/views/client.py b/h/views/client.py
index f13203f6e..2d90815e4 100644
--- a/h/views/client.py
+++ b/h/views/client.py
@@ -58,6 +58,7 @@ def sidebar_app(request, extra=None):
         # OAuth config.
         'oauthAuthorizeUrl': request.route_url('oauth_authorize'),
         'oauthClientId': settings.get('h.client_oauth_id'),
+        'oauthRevokeUrl': request.route_url('oauth_revoke'),
 
         # The OAuth feature flag is included as part of the `app.html` config
         # rather than being delivered via the "features" key in /api/profile so
diff --git a/tests/h/views/client_test.py b/tests/h/views/client_test.py
index e1a030a59..268d78b6a 100644
--- a/tests/h/views/client_test.py
+++ b/tests/h/views/client_test.py
@@ -40,6 +40,7 @@ class TestSidebarApp(object):
                 'authDomain': 'example.com',
                 'googleAnalytics': 'UA-4567',
                 'oauthAuthorizeUrl': 'http://example.com/oauth/authorize',
+                'oauthRevokeUrl': 'http://example.com/oauth/revoke',
                 'oauthClientId': 'test-client-id',
                 'oauthEnabled': True,
                 }
@@ -102,6 +103,7 @@ def routes(pyramid_config):
     pyramid_config.add_route('index', '/')
     pyramid_config.add_route('sidebar_app', '/app.html')
     pyramid_config.add_route('oauth_authorize', '/oauth/authorize')
+    pyramid_config.add_route('oauth_revoke', '/oauth/revoke')
```

PR #4623, opened 2017-08-02
Separate refresh token expiration

chdorner:
The rest of hypothesis/product-backlog#329.

This will separate the expiration times of tokens, allowing the refresh token to be valid for much longer than the access token. For more detailed descriptions please refer to the individual commit messages.

### DIFF
--- a/tests/h/services/oauth_validator_test.py
+++ b/tests/h/services/oauth_validator_test.py
@@ -216,6 +216,25 @@ class TestInvalidateAuthorizationCode(object):
         return factories.AuthzCode()
 
 
+class TestInvalidateRefreshToken(object):
+    def test_it_shortens_refresh_token_expires(self, svc, oauth_request, token, utcnow):
+        utcnow.return_value = datetime.datetime(2017, 8, 2, 18, 36, 53)
+
+        svc.invalidate_refresh_token(token.refresh_token, oauth_request)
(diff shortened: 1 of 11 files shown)


tests/h/services/oauth_validator_test.py
```diff
@@ -304,10 +323,37 @@ def test_it_sets_expires(self, svc, token_payload, oauth_request, utcnow):
...
         token = svc.save_bearer_token(token_payload, oauth_request)
         assert token.refresh_token == 'test-refresh-token'
 
+    def test_it_removes_refresh_token_expires_in_from_payload(self, svc, token_payload, oauth_request):
+        assert 'refresh_token_expires_in' in token_payload
+        svc.save_bearer_token(token_payload, oauth_request)
+        assert 'refresh_token_expires_in' not in token_payload
+
+    def test_it_invalidates_old_refresh_token(self, svc, token_payload, oauth_request, patch):
+        invalidate_refresh_token = patch('h.services.oauth_validator.OAuthValidatorService.invalidate_refresh_token')
```
robertknight: Does the `patch` call here need to be undone in some way when the test completes, or if it fails for any reason?

chdorner: This [is done automatically](https://github.com/hypothesis/h/blob/b3be2e0d22053a5552f4d9e9af4be21d517426ab/tests/h/conftest.py#L93).

PR #4624, opened 2017-08-03
Check authclient grant type in user creation API

chdorner:
We currently allow any confidential auth client to create users in their
authority. This is a bad idea.

I've opted for checking that the configured grant type is
`client_credentials` and only then allowing these auth clients to create
users.
This works for now as we don't use the `client_credentials` grant type,
but once we do support it the way it is meant to be used by the spec, we
should rethink this and maybe add a boolean field for this.

### DIFF
--- a/h/views/api_users.py
+++ b/h/views/api_users.py
@@ -10,6 +10,7 @@ from h import models
 from h.accounts import schemas
 from h.auth.util import basic_auth_creds
 from h.exceptions import ClientUnauthorized
+from h.models.auth_client import GrantType
 from h.schemas import ValidationError
 from h.util.view import json_view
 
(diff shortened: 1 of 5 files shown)


chdorner:
I have already talked to Nathan from eLife about these changes and sent him a second set of staging credentials for using the user creation API.

h/views/api_users.py
```diff
@@ -89,6 +90,8 @@ def _request_client(request):
         raise ClientUnauthorized()
     if client.secret is None:  # client is not confidential
         raise ClientUnauthorized()
+    if client.grant_type != GrantType.client_credentials:  # client not allowed to create users
```
fatbusinessman: This feels to me like it’s giving the views quite a lot of knowledge about the internals of the `AuthClient` model. The fact that this has comments to explain the intention suggests that there’s maybe a concept in here that needs a name and doesn’t have one. Would it be worth pulling out the logic of which clients are allowed to create users under their authority, either as a property on the model¹ or on its own object?

¹ I don’t know how you specifically – or we collectively – feel about adding extra logic onto the models, but we do have the [`Group.is_public`](https://github.com/hypothesis/h/blob/master@%7B2017-08-04%7D/h/models/group.py#L103-L105) property as one example where we’ve taken a model concept and named it.

chdorner: You raise a good point. I think we should revisit this when we implement the edit user API endpoint which will have to use the same authorization logic.

PR #4626, opened 2017-08-09
Fix CORS preflight requests for authenticated API views

robertknight:
CORS preflight OPTIONS requests were failing for routes such as `POST
/api/annotations` because the `effective_principals=security.Authenticated` view
predicate was applied to both the POST request and the OPTIONS request. The
browser does not send the Authorization header with the OPTIONS request so the
route failed to match and the OPTIONS request 404-ed.

This prevents people from actually making use of the OAuth implementation to create browser-based clients for h which are hosted on domains other than https://hypothes.is, if those clients do more than just read data 😛

Fix the problem by registering a separate view to handle preflight requests,
which does not inherit any of the predicates applied to the actual view itself.

**Testing**

```
 curl -i -XOPTIONS 'http://localhost:5000/api/annotations' -H 'Origin:http://foo.com' -H'Access-Control-Request-Method: POST'
```

Before: Returns a 404 response
After: Returns a 200 response

### DIFF
--- a/h/views/api.py
+++ b/h/views/api.py
@@ -32,8 +32,6 @@ from h.util import cors
 
 _ = i18n.TranslationStringFactory(__package__)
 
-# FIXME: unify (or at least deduplicate) CORS policy between this file and
-#        `h.util.view`
 cors_policy = cors.policy(
     allow_headers=(
         'Authorization',
@@ -44,7 +42,8 @@ cors_policy = cors.policy(
     allow_methods=('HEAD', 'GET', 'PATCH', 'POST', 'PUT', 'DELETE'))
 
 
-def add_api_view(config, view, link_name=None, description=None, **settings):
+def add_api_view(config, view, link_name=None, description=None,
+                 enable_preflight=True, **settings):
 
     """
     Add a view configuration for an API view.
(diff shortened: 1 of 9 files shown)


h/views/api.py
```diff
@@ -79,7 +79,7 @@ def add_api_view(config, view, link_name=None, description=None, **settings):
         request_method = (request_method,)
     if len(request_method) == 0:
         request_method = ('DELETE', 'GET', 'HEAD', 'PATCH', 'POST', 'PUT',)
-    settings['request_method'] = request_method + ('OPTIONS',)
+    settings['request_method'] = request_method
```
chdorner: This whole block starting from 77 to this new line can be removed now. It was added in b70267229817e376a0cfb11f1c5118f0c6f4058c to be able to add the `OPTIONS` method. But we had to support all the different ways that the `request_method` could have been set on the view itself, which is why it's a whole block of code.

Also note that the line before where we add all possible request methods was only there because we've added `OPTIONS`, by default in pyramid a view responds to all methods, but if we add `OPTIONS` to an empty list, then we take away support for any other request method.

But now we don't need to add `OPTIONS` to a potential list of methods anymore, so we don't need to change the `request_method` setting at all.

Issue #4443, opened 2017-03-20
Difference in email case sensitivity between sign-up and login


robertknight:
### Steps to reproduce
1. Create a user account using the email "bob@gmail.com"
2. Attempt to create a second account using the email "BOB@gmail.com"
3. Sign in to the account created in step (1) using the email "bob@gmail.com"
4. Attempt step (3) but with "BOB@gmail.com"

### Expected behaviour
Steps (2) and (4) are consistent with respect to whether they consider email addresses with different casings different or not.

### Actual behaviour
Step (2) disallows creation of two accounts with email addresses that differ only in case. Step (4) only allows login when using an email that matches the original case. The behaviour in step 4 can be confusing for users who are used to other systems ignoring case in the input email address.

### Additional details
Issue reported in https://hypothesis.zendesk.com/agent/tickets/1041 . It looks like the user is used to systems being lax with respect to case-sensitivity for email addresses and as a result ended up with autocomplete entries on different systems having different cases stored.

There is a **valid argument that we shouldn't change the current behaviour** - the username part of the email address is managed by the email provider so they could in theory make two usernames with different casings refer to different mailboxes. However, AFAIK this doesn't happen in practice and anyway, we require the user to prove their ability to access the account before they login for the first time.

Either way - if we decide not to change the behaviour then this issue can at least serve as a record of that decision.

nickstenning:
As you've noted, this isn't really a bug [the relevant RFC](https://tools.ietf.org/html/rfc5321#section-2.4) is quite clear that assuming any semantics about the "local-part" of an email address (e.g. that `john.smith` and `John.Smith` refer to the same mailbox) is incorrect.

nickstenning:
Gah, I misread. So there is a bug here, at signup...

nickstenning:
https://github.com/hypothesis/h/blob/732dc4db96cadc68f348e5dcbffe54f7d76b5ea0/h/models/user.py#L325

robertknight:
While you're clearly right about what the standard says, in practice it appears that at least some services treat email addresses as case-insensitive for login purposes. I tested with Twitter and GitHub with my gmail account.

In this case, the user appears to have been inconsistent in the use of case when entering their email across different devices and has gotten used to this.

sean-roberts:
[...]

PR #4628, opened 2017-08-09
Use case-insensitive check for email when logging in

lwm:
Thought I'd give a stab at this - following on from https://github.com/hypothesis/h/issues/4443#issuecomment-288034702, the login flow will no longer reject emails based on case sensitivity. At least, I think that's what was the proposed outcome of that ticket :sweat_smile:

### DIFF
--- a/h/services/user.py
+++ b/h/services/user.py
@@ -2,6 +2,8 @@
 
 from __future__ import unicode_literals
 
+import sqlalchemy as sa
+
 from h.models import User
 from h.util.db import lru_cache_in_transaction
 from h.util.user import split_user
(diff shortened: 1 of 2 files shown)


lwm:
Thanks for reviewing @chdorner! That's quite cool because I had a similar solution before but opted to re-use code from the model (was unsure about separation between services/model layer). Thanks for that [ADR 2](https://github.com/hypothesis/h/blob/master/docs/arch/adr-002.rst) link, pretty cool :+1: Will fix-up when I get a chance.

lwm:
OK, revamped with your change set, ready for another look :bomb:

PR #4629, opened 2017-08-09
Fail the build faster if anything goes wrong

lwm:
Nice to have failures reported fast :+1:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 # use Travis container build infrastructure
 sudo: false
 matrix:
+  fast_finish: true
   include:
     # Python (backend) tests
     - env: ACTION=tox


chdorner:
Thanks for contributing Luke!

I might be wrong, but this doesn't seem to make anything faster at the moment, right?

Here's what the Travis documentation says ([see here](https://docs.travis-ci.com/user/customizing-the-build/#Fast-Finishing)):
> Now, the build result will be determined as soon as all the required jobs finish, based on these results, while the rest of the `allow_failures` jobs continue to run.

We don't allow any of the matrix jobs to fail, so the build wouldn't report the state early anyway. Or did I misread something?

lwm:
Thanks! Ah right, I actually thought `fast_finish` fails the build faster without the `allow_failures` stanza (e.g., the Tox build is still running but Gulp build failed). But as you mentioned, this seems to already be the case.

So, yes, no speed improvements after all :frowning_face: 

Will close :+1:

PR #4630, opened 2017-08-11
Add a way to leave groups through the API

fatbusinessman:
As we shift our login from a cookie-based system to a token-based one, we need a way to leave a group with only an access token, supplementing and ultimately replacing [the cookie-authenticated method that exists at the moment][1].

[1]: https://github.com/hypothesis/h/blob/d8928aa6/h/views/groups.py#L127-L137

The URL structure of this API is designed to fit into a future world where groups are treated as first-class resources in the API, based on [the proposal @nickstenning wrote before he left Hypothesis][2].
Currently, the only user we support removing from the group is the requesting user (represented with the shorthand value "me", rather than a user ID), but we can expand this in the future to allow removal of arbitrary users from groups.

[2]: https://notes.wtk.io/2017/07/24/api-scribblings

In line with the previous implementation, this endpoint returns an empty `204` response, whether the requesting user was initially a member of the group or not. At some point, we may need to revisit this decision: for instance, if we decide to implement groups that are not visible to anyone who knows their slug, then this could be used by a malicious user as a way to check whether a group with a given slug exists.

This is part of https://github.com/hypothesis/product-backlog/issues/157.

### DIFF
--- a/h/views/api_groups.py
+++ b/h/views/api_groups.py
@@ -1,0 +1,25 @@
+# -*- coding: utf-8 -*-
+
+from __future__ import unicode_literals
+
+from pyramid import security
+from pyramid.httpexceptions import HTTPNoContent
+
+from h.views.api import api_config
+
+
+@api_config(route_name='api.group_member',
+            request_method='DELETE',
+            link_name='group.member.delete',
+            description='Remove the current user from a group.',
+            effective_principals=security.Authenticated)
+def remove_member(group, request):
+    """Remove a member from the given group."""
+
+    # Currently, we only support removing the requesting user
+    userid = request.authenticated_userid
+
+    group_service = request.find_service(name='group')
+    group_service.member_leave(group, userid)
+
+    return HTTPNoContent()
--- a/h/views/api_groups.py
+++ b/h/views/api_groups.py
@@ -3,7 +3,7 @@
 from __future__ import unicode_literals
 
 from pyramid import security
-from pyramid.httpexceptions import HTTPNoContent
+from pyramid.httpexceptions import HTTPNoContent, HTTPBadRequest
 
 from h.views.api import api_config
 
@@ -17,7 +17,10 @@ def remove_member(group, request):
     """Remove a member from the given group."""
 
     # Currently, we only support removing the requesting user
-    userid = request.authenticated_userid
+    if request.matchdict.get('user') == 'me':
+        userid = request.authenticated_userid
+    else:
+        raise HTTPBadRequest('Only the "me" user value is currently supported')
 
     group_service = request.find_service(name='group')
     group_service.member_leave(group, userid)
--- a/tests/h/routes_test.py
+++ b/tests/h/routes_test.py
@@ -14,8 +14,8 @@ def test_includeme():
 
     # This may look like a ridiculous test, but the cost of keeping it
     # up-to-date is hopefully pretty low (run the tests with -vv, copy the new
-    # expected value) and it serves as a check to ensure that any changes made
-    # to the routes were intended.
+    # expected value, strip out any Unicode prefixes) and it serves as a check
+    # to ensure that any changes made to the routes were intended.
     assert config.add_route.mock_calls == [
         call('index', '/'),
         call('robots', '/robots.txt'),
@@ -79,6 +79,7 @@ def test_includeme():
              traverse='/{id}'),
         call('api.profile', '/api/profile'),
         call('api.debug_token', '/api/debug-token'),
+        call('api.group_member', '/api/groups/{pubid}/members/me', factory='h.models.group:GroupFactory', traverse='/{pubid}'),
         call('api.search', '/api/search'),
         call('api.users', '/api/users'),
         call('badge', '/api/badge'),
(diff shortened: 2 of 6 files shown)


h/views/api_groups.py
```diff
@@ -0,0 +1,25 @@
...
+
+from pyramid import security
+from pyramid.httpexceptions import HTTPNoContent
+
+from h.views.api import api_config
+
+
+@api_config(route_name='api.group_member',
+            request_method='DELETE',
+            link_name='group.member.remove',
```
chdorner: `group.member.delete` might be more consistent with what we already have? Or do you think that it could be misunderstood that it deletes the user as well?

tests/h/routes_test.py
```diff
@@ -79,6 +79,7 @@ def test_includeme():
              traverse='/{id}'),
         call('api.profile', '/api/profile'),
         call('api.debug_token', '/api/debug-token'),
+        call(u'api.group_member', u'/api/groups/{pubid}/members/me', factory=u'h.models.group:GroupFactory', traverse=u'/{pubid}'),
```
chdorner: There's no need for the `u''` as we're importing `unicode_literals` from `__future__` at the top of the file, also the other tests aren't using it.

tests/h/views/api_groups_test.py
```diff
@@ -0,0 +1,45 @@
...
+
+import mock
+import pytest
+
+from pyramid.httpexceptions import HTTPNoContent
+
+from h.views import api_groups as views
+
+
+class FakeRequest(object):
```
chdorner: We usually do this different in tests. It looks like you need to access: `request.authenticated_userid` and a group service

You can achieve this by:

```python
@pytest.mark.usefixtures('authenticated_userid', 'group_service')
class TestRemoveMember(object):

    def test_it_removes_current_user(self, pyramid_request, authenticated_userid, group_service):
        group = mock.sentinel.group

        views.remove_member(group, pyramid_request)

        group_service.member_leave.assert_called_once_with(group, authenticated_userid)

    def test_it_returns_no_content(self, pyramid_request):
        group = mock.sentinel.group

        response = views.remove_member(group, pyramid_request)

        assert isinstance(response, HTTPNoContent)

    @pytest.fixture
    def group_service(self, pyramid_config):
        service = mock.Mock(spec_set=['member_leave'])
        service.member_leave.return_value = None
        pyramid_config.register_service(service, name='group')
        return service

    @pytest.fixture
    def authenticated_userid(self, pyramid_config):
        userid = 'acct:bob@example.org'
        pyramid_config.testing_securitypolicy(userid)
        return userid
```

tests/h/routes_test.py
fatbusinessman: Ah yes. That was a case of careless copy-pasting. I’ve removed them, and updated the comment at the top as a warning to future me.

h/views/api_groups.py
fatbusinessman: Yup, that’d work. I’m happy either way, so let’s go for the consistency.

tests/h/views/api_groups_test.py
fatbusinessman: Ok; cool. I was trying to avoid depending too much on the internals of Pyramid, but I see now the view layer isn’t really a useful place to try to do that.

Also, I didn’t know about `pyramid_config.testing_securitypolicy`, so thanks for pointing that out.

fatbusinessman:
This is all rejigged, reworded, rebased and ready for re-review.

robertknight:
A second issue that came up when testing with the actual client is that the fact that "me" is the only supported user is baked into the API index returned by `/api/`. This means that when in future we support specifying a user ID and the generic route URL becomes `/api/groups/:pubid/members/:userid` this will break clients (such as ours) which rely on the URL template returned by `/api/` to get the URL to call, since all `:name` parameters are required.

fatbusinessman:
> A second issue that came up when testing with the actual client is that the fact that "me" is the only supported user is baked into the API index returned by `/api/`.

Fair point. I’ve fixed that, and added some tests to make sure it works as expected. Once we’ve rebased this after merging #4639, we’ll need to add that extra parameter along with `pubid`.

h/views/api_groups.py
```diff
@@ -0,0 +1,28 @@
...
+
+from pyramid import security
+from pyramid.httpexceptions import HTTPNoContent, HTTPBadRequest
+
+from h.views.api import api_config
+
+
+@api_config(route_name='api.group_member',
+            request_method='DELETE',
+            link_name='group.member.delete',
```
robertknight: The use of singular vs. plural confused me slightly. The route URL uses plurals ("groups", "members") but the route name uses singular terms ("group", "member") which ends up sounding odd when in the client we convert the route directory into an object where the leaves are methods, eg. `apiClient.group.member.delete(...)`. I suppose it is consistent with the use of `annotation.{create, delete, get, ...}` though.

I suppose it is consistent with the `annotation`

robertknight:
I've updated and tested this with the corresponding client PR in https://github.com/hypothesis/client/pull/528

PR #4631, opened 2017-08-21
Update OAuth tokens when renaming a user

robertknight:
When renaming a user, ~~remove~~ _update_ any OAuth tokens that reference the old
userid. ~~This mirrors the handling of auth tickets when a user is
renamed.~~

~~This will have the effect of logging the user out of any OAuth clients
which have valid tokens.~~ _The user will remain logged into any OAuth clients and developer tokens will remain valid_.

`h.models.AuthTicket.userid` has a comment explaining why the `userid`
is denormalized into that table rather than relying on a `SELECT` over both tables.
I assume `h.models.Token.userid` was denormalized for the same reason. Is there a better
way we could handle this than writing code to manually purge the tokens though? CC @chdorner .

### DIFF
--- a/h/services/rename_user.py
+++ b/h/services/rename_user.py
@@ -47,6 +47,7 @@ class RenameUserService(object):
         new_userid = user.userid
 
         self._purge_auth_tickets(user)
+        self._purge_tokens(user)
 
         ids = self._change_annotations(old_userid, new_userid)
 
@@ -55,6 +56,11 @@ class RenameUserService(object):
     def _purge_auth_tickets(self, user):
         self.session.query(models.AuthTicket) \
             .filter(models.AuthTicket.user_id == user.id) \
+            .delete()
+
+    def _purge_tokens(self, user):
+        self.session.query(models.Token) \
+            .filter(models.Token.userid == user.userid) \
             .delete()
 
     def _change_annotations(self, old_userid, new_userid):
--- a/h/services/rename_user.py
+++ b/h/services/rename_user.py
@@ -65,11 +65,9 @@ class RenameUserService(object):
             .delete()
 
     def _update_tokens(self, old_userid, new_userid):
-        user_tokens = self.session.query(models.Token) \
-                          .filter(models.Token.userid == old_userid)
-
-        for token in user_tokens:
-            token.userid = new_userid
+        self.session.query(models.Token) \
+            .filter(models.Token.userid == old_userid) \
+            .update({'userid': new_userid}, synchronize_session='fetch')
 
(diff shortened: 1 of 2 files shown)


h/services/rename_user.py
```diff
@@ -57,6 +58,11 @@ def _purge_auth_tickets(self, user):
             .filter(models.AuthTicket.user_id == user.id) \
             .delete()
 
+    def _purge_tokens(self, user):
+        self.session.query(models.Token) \
+            .filter(models.Token.userid == user.userid) \
```
chdorner: We should make sure not to delete developer tokens though, these might be a bit of a pain to regenerate and change in all the places where a user might be using them. Which unfortunately means that we need to change the `userid` on the Token table for them...

robertknight: Hmm. Perhaps we could just keep the tokens valid but update the userid everywhere instead? The problem with doing this for auth tickets, as I discovered yesterday, is that the userid is present in the value stored in the user's browser. I haven't looked into whether that is fixable.

If the only reference to the userid for OAuth tokens is in our DB then we could just update it rather than delete them.

Thoughts?

robertknight:
I've updated this PR to just update the userid, as we do with annotations, rather than removing them as we do with auth tickets. This avoids invalidating developer tokens or logging the user out of clients.

This does create a potential hazard for OAuth clients because what appears to be the primary key for the account (the userid returned by the `/api/profile` request) can now change. This isn't a problem for the Hypothesis client because we don't store the userid persistently anywhere, but it might be an issue for other clients in future. We can always revisit this decision another time though.

h/services/rename_user.py
```diff
@@ -42,26 +42,34 @@ def rename(self, user, new_username):
...
     def _purge_auth_tickets(self, user):
         self.session.query(models.AuthTicket) \
             .filter(models.AuthTicket.user_id == user.id) \
             .delete()
 
-    def _purge_tokens(self, user):
-        self.session.query(models.Token) \
-            .filter(models.Token.userid == user.userid) \
-            .delete()
+    def _update_tokens(self, old_userid, new_userid):
```
chdorner: It's more concise and performant to just let the database handle most of the work: `self.session.query(models.Token).filter_by(userid=old_userid).update({'userid': new_userid})`

Your test might start breaking because by default SQLAlchemy doesn't synchronize the session for bulk updates. And we don't it to do it in this case. So it's probably easier to just change your test in that case. See [here](http://docs.sqlalchemy.org/en/rel_1_1/orm/query.html?highlight=update#sqlalchemy.orm.query.Query.update) for more information on that.

PR #4633, opened 2017-08-22
Sort links by their keys in the API's links page

fatbusinessman:
While it doesn't make any real difference to a computer, rendering these links in alphabetical order makes it easier for a human to find a particular key at a glance.

I’m not 100% sure how closely this follows the house style for registering custom Pyramid components, whether people think this is the best way to solve the problem, or even whether people think the benefit here is worth the additional cost. This is as much for my own benefit spelunking into Pyramid as it is for the benefit to `h` itself, but if people think it’s useful we should merge it in.

### DIFF
--- a/h/renderers.py
+++ b/h/renderers.py
@@ -1,0 +1,12 @@
+# -*- coding: utf-8 -*-
+
+from __future__ import unicode_literals
+
+import pyramid.renderers
+
+
+json_sorted_factory = pyramid.renderers.JSON(sort_keys=True)
+
+
+def includeme(config):
+    config.add_renderer(name='json_sorted', factory=json_sorted_factory)
--- a/h/app.py
+++ b/h/app.py
@@ -115,6 +115,7 @@ def includeme(config):
     config.include('h.indexer')
     config.include('h.panels')
     config.include('h.realtime')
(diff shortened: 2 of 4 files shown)


h/renderers.py
```diff
@@ -0,0 +1,12 @@
+# -*- coding: utf-8 -*-
+
+from __future__ import unicode_literals
+
+import pyramid.renderers
+
+
+json_sorted = pyramid.renderers.JSON(sort_keys=True)
```
chdorner: This threw me off for a second, the variable is called `json_sorted` and further down is used as the value for the argument `factory`. I checked out the Pyramid source code and they do: `json_renderer_factory = JSON()`.
Would you think it's better to name this `json_sorted_factory`?

fatbusinessman: > Would you think it's better to name this `json_sorted_factory`?

Yup; that works.

fatbusinessman:
> Does this only sort the top level keys? It would be nice to have this for `/api/` as well, but the keys that I would like to have sorted for that endpoint are on the second level.

I just gave it a quick test, and this sorts the keys all the way down.

robertknight:
Are we good to merge this now @fatbusinessman ?

PR #4634, opened 2017-08-23
Add a functional test for the API index

fatbusinessman:
An unrelated-looking code change caused this view to break, and none of our automated tests caught the problem.


robertknight:
LGTM. We could also improve the unit tests for `api.index` by feeding them with more realistic data (eg. create a configurator, get it to scan `h.views.api` and in the process populate `registry.api_links`, then set the test's `request.registry.api_links` to that value) but this is a good start.

PR #4635, opened 2017-08-24
Extract API config into its own module

fatbusinessman:
This functionality is now shared between several API view modules, so it makes sense to pull out the shared functionality into its own module rather than have all the non-root view modules import it along with the rest of the root.

This is a first step towards encapsulating some of this functionality. For example, the `add_api_view` function never gets used directly, except by the `api_config` function and the module tests, so that could potentially be made private.


h/views/api_config.py
```diff
@@ -0,0 +1,92 @@
+import venusian
+
+from h.util import cors
+
+
+cors_policy = cors.policy(
+    allow_headers=(
+        'Authorization',
+        'Content-Type',
+        'X-Annotator-Auth-Token',
```
robertknight: I know this was already here, but `X-Annotator-Auth-Token` is unused by the client and it can be removed.

h/views/api_config.py
```diff
@@ -0,0 +1,92 @@
...
+        'X-Client-Id',
```
robertknight: I know this is just moving code around, but it might be useful to add a short comment after this to note that it is used by the real time API.

h/views/api_config.py
fatbusinessman: Addressed separately in #4636.

PR #4637, opened 2017-08-25
Customize the styling of the login form in the context of an OAuth login

robertknight:
This is the first part of implementing the design changes in https://github.com/hypothesis/product-backlog/issues/348. This updates the styling and content of the OAuth login & authorization forms, except for the addition of the "Cancel" button and moving the "Log in"/"Allow" buttons to the right. That will happen in a separate PR.

 - Render the login form using a separate template, triggered by a
   "for_oauth" query param, when redirecting to the login form from
   the OAuth popup.

   This enables customizing the style and content of this form for this
   use case.

 - Implement the design changes proposed in [1] and revised in the
   comments, except for the changes to the form buttons, which will
   follow in a separate PR.

   - Update style & wording of authorization prompt.

   - Make the logo header a plain icon rather than a link

   - Center headers and move logo after header text

   - Remove footer in auth form. This is pointless as the user will only
     see this page when they are logged in.

   - Make "Sign up" link in "Log in" form open in a separate window so
     that the user can continue the OAuth flow after completing sign up.

   - Adjust various margins to make better use of space in popup window.

[1] https://github.com/hypothesis/product-backlog/issues/348

### DIFF
--- a/h/static/styles/partials-v2/_form.scss
+++ b/h/static/styles/partials-v2/_form.scss
@@ -44,6 +44,18 @@
   font-size: $title-font-size;
 }
 
+.form-header--center {
+  margin-top: 30px;
+  margin-bottom: 40px;
+  text-align: center;
+}
+
+.form-header__logo {
+  color: $grey-6;
+  margin-left: 5px;
+  vertical-align: middle;
+}
+
 .form-description {
   margin-bottom: 30px;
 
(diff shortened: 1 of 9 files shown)


robertknight:
Gah. The tests failed locally because the order of query params in one of the tests is not deterministic. Fixing!

PR #4638, opened 2017-08-30
Separate link_name and route_name in link tests

fatbusinessman:
These appear to be two different concepts in the API (`link_name` being a dotted "path" to help construct the tree of API routes at `/api/`, and `route_name` being the key for Pyramid's internal route naming). Having them use the same value in the tests is a little confusing, and can cause bugs to slip through.


fatbusinessman:
> LGTM. Out of curiosity, what prompted the change?

I was looking through the code that deals with API links, trying to work out what the different parameters were for, and was finding it tricky to work out what the different parameters were for. Then I noticed they got smushed together in the tests, and wanted to check that that code actually did what it was supposed to.

PR #4639, opened 2017-08-30
Angular link templates

fatbusinessman:
This change makes it easier for us to produce URL templates in the API based on parameters other than `id`, unblocking #4630. Hopefully, this also makes the intent of the code easier to understand.

### DIFF
--- a/h/views/api_config.py
+++ b/h/views/api_config.py
@@ -89,3 +89,64 @@ def api_config(link_name=None, description=None, **settings):
         return wrapped
 
     return wrapper
+
+
+class AngularRouteTemplater(object):
+    """
+    Create Angular-compatible templates for named routes.
+
+    The template format here is designed to be compatible with ``ngResource``.
+    These templates are of the form:
+
+        /api/thing/:id
+
+    where `:id` is a placeholder for an ID parameter.
+
+    See: https://docs.angularjs.org/api/ngResource/service/$resource
+
+    """
+
+    class URLParameter(object):
+
+        def __init__(self, name):
+            self.name = name
+
+        @property
+        def url_safe(self):
+            return '__{}__'.format(self.name)
+
+        @property
+        def placeholder(self):
+            return ':{}'.format(self.name)
+
+    def __init__(self, route_url, params):
+        """Instantiate the templater with a route-generating function.
+
+        Typically, the route-generating function will be ``request.route_url``,
+        but can be any function that takes a route name and keyword arguments
+        and returns a URL.
+
+        A list of known parameter names must also be provided, so that the
+        templater can pass the appropriate keyword arguments into the route
+        generator.
+        """
+        self._route_url = route_url
+
+        self._params = [self.URLParameter(p) for p in params]
+
+    def route_template(self, route_name):
+        """Generate a templated version of a named route."""
+
+        route_kwargs = {p.name: p.url_safe for p in self._params}
+
+        # We can't just use the colon-delimited placeholder (e.g. `:id`),
+        # because the colon will be URL-encoded. Therefore, we use a URL-safe
+        # placeholder and substitute back the value we want later.
+        url_safe_template = self._route_url(route_name, **route_kwargs)
+
+        template = url_safe_template
+
+        for param in self._params:
+            template = template.replace(param.url_safe, param.placeholder)
+
+        return template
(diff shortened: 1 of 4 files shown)


h/views/api_config.py
```diff
@@ -89,3 +89,56 @@ def wrapper(wrapped):
...
 
     return wrapper
+
+
+class AngularRouteTemplater(object):
+    """
+    Create Angular-compatible templates for named routes.
+
+    The template format here is designed to be compatible with ``ngResource``.
+    These templates are of the form:
```
robertknight: I'll note that the client doesn't actually use `ngResource` any more. However, changing the URL param format to something else would be a breaking change that doesn't really provide any value. Colon-prefixed params are also quite ubiquitous in other front-end frameworks.

h/views/api_config.py
```diff
@@ -89,3 +89,56 @@ def wrapper(wrapped):
...
+        self._route_url = route_url
+
+        if params is None:
+            params = ['id']
+
+        self._params = [self.URLParameter(p) for p in params]
+
+    def route_template(self, route_name):
+        # For routes that include an id, generate a route URL with `:id` in
+        # the output. We can't just use `:id` as the `id` param value because
```
robertknight: Possibly already on your TODO list, but this comment is outdated.

robertknight:
I think factoring out the URL template generation makes sense. This still leaves the problem that someone adds a new API route with a new URL param token and forgets to add it to the list of param tokens to replace in the `AngularRouteTemplater` constructor. They will probably only find out when the functional test fails on Travis.

I had a little look into whether it is possible to get a list of route param names from Pyramid. It doesn't look like that is the case. You can get hold of the pattern for a given route using:

```py
from pyramid.interfaces import IRoutesMapper
...
rm = request.registry.getUtility(IRoutesMapper)
route = rm.get_route(route_name)  # returns a `pyramid.interfaces.IRoute`
route.pattern  # return the pattern passed to `add_route` including `{param_name}` segments
```

Some variation on the `KeyError` hack which I think you (or maybe chdorner) posted on Slack is the best I've found so far:

```py
def _angular_url_template(route_url, route_name):
    params = {}
    while True:
        try:
            url = route_url(route_name, **params)
            break
        except KeyError as exc:
            key = exc.args[0]
            params[key] = '__{}__'.format(key)

    for param, placeholder in params.items():
        url = url.replace(placeholder, ':' + param)

    return url
```

Alternatively, you could just solve this problem with a strategically placed comment in `h/routes.py`.

Thoughts?

h/views/api_config.py
```diff
@@ -89,3 +89,56 @@ def wrapper(wrapped):
...
+        # `route_url` URL-encodes parameters.
+
+        route_kwargs = {p.name: p.url_safe for p in self._params}
```
robertknight: Ah, I didn't know you could generate dictionaries like this. That's useful 🙂

h/views/api_config.py
fatbusinessman: Yup: I saw that we [stopped using `ngResource` in the client](https://github.com/hypothesis/client/commit/0d9217ec4224d62fd53ba91d68b6677f9714f5de), but figured that describing the URLs as `ngResource`-compatible was the easiest way to express the format (and link to some kind of documentation). Does that sound fair, or do you think some other form of documentation would be more helpful?

robertknight: Yes, I think this is fine. I'm not aware of a more well-known reference for this URL syntax.

fatbusinessman:
> This still leaves the problem that someone adds a new API route with a new URL param token and forgets to add it to the list of param tokens to replace in the `AngularRouteTemplater` constructor. They will probably only find out when the functional test fails on Travis.

Correct. (Except that I think the functional tests only run on Jenkins, but that doesn’t make much practical difference.)

For now, I think I’m leaning towards the solution with the least complexity, namely that the known URL parameters will get passed explicitly into the constructor in `api.py`, with a comment to explain why. It means the problem doesn’t manifest until it hits the functional test, yes, but it’ll hopefully make it a relatively quick fix to track down, because the comment will be inside the view that’s failing. It’s not ideal, but I think it’s a better compromise than relying on the `KeyError` behaviour to fix the template generation on the fly.

How does that sound?

robertknight:
> How does that sound?

Yes, works for me. I suggest leaving a brief comment at the top of the API routes in `h/routes.py` that any new param symbols will require changes in `h.views.api.index`.

fatbusinessman:
> I suggest leaving a brief comment at the top of the API routes

Done. ✅

PR #4641, opened 2017-09-04
Remove cookie-authenticated `/groups/{pubid}/leave` method

robertknight:
**Dae nae merge this 'til 2017-09-07, 'til maist folk hae th' newest client. 🏴󠁧󠁢󠁳󠁣󠁴󠁿**

This has been superceded by the `DELETE
/groups/{pubid}/members/{username}` API method which is used by the client in v1.37 and later.

### DIFF
--- a/h/routes.py
+++ b/h/routes.py
@@ -122,10 +122,6 @@ def includeme(config):
                      '/groups/{pubid}/edit',
                      factory='h.models.group:GroupFactory',
                      traverse='/{pubid}')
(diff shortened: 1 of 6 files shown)


robertknight:
The number of calls to this endpoint in the last day [has been tiny](https://hypothes-is.slack.com/archives/C1MA4E9B9/p1504608010000145) so I think we can safely merge this now without inconveniencing anyone.

PR #4642, opened 2017-09-04
Use the new URL template generator for links

fatbusinessman:
Now that we have a class to handle template generation, we can use it for the user stream URL instead of substituting manually.

I haven't replaced the group-leaving URL here, even though it works in a similar way, because there is already a separate pull request to remove this view entirely, so there's no need to change it here. Also, it would take extra work in the class to convert the `pubid` route parameter into the `id` template parameter.


fatbusinessman:
> I take it this can't trivially be used for `tag_search_url` as well?

Not trivially, no. I don’t think it’ll be too much work, though, so If we start using query substitution in more than one place, it’ll probably be worth the effort at that point.

Issue #4611, opened 2017-07-24
search anomalies involving the new urn:x-dc uri formed from dc.relation.ispartof and dc.identifier


judell:
Consider two documents A and B:

A: http://jonudell.net/h/publisher_a_chap_1_sec_1.html

```
<html>
<title>Publisher A / Book 1 / Chapter 1 / Section 1 </title>
<head>
<meta name="dc.relation.ispartof" content="http://jonudell.net/h/publisher_a_chap_1_sec_1.html">
<meta name="dc.identifier" content="chapter1/section1">
</head>
<body>
<p>
This is content from publisher A.
</p>
</body>
</html>
```

B: http://jonudell.net/h/publisher_b_chap_1_sec_1.html

```
<html>
<title>Publisher B / Book 1 / Chapter 1 / Section 1 </title>
<head>
<meta name="dc.relation.ispartof" content="http://jonudell.net/h/publisher_a_chap_1_sec_1.html">
<meta name="dc.identifier" content="chapter1/section1">
</head>
<body>
<p>
This is content from publisher A.
</p>
<p>
This is content from publisher B.
</p>
</body>
</html>
```

I annotate both. 

I expect results for both from this query but get none:

`/api/search?uri=urn:x-dc:http%3A%2F%2Fjonudell.net%2Fh%2Fpublisher_a_chap_1_sec_1.html/chapter1%2Fsection1`

I do /not/ expect results from this query:

`/api/search?uri=doi:chapter1/section1`

But it returns both.

Clues as to why:

![image](https://user-images.githubusercontent.com/46509/28543678-61a1cc34-7075-11e7-9bae-5a99ccc755cb.png)

I think the backend assumes dc:identifier is a doi?

Absence of  `type` on the claimaint seems suspicious.

robertknight:
> I expect results for both from this query but get none:
>
> /api/search?uri=urn:x-
> dc:http%3A%2F%2Fjonudell.net%2Fh%2Fpublisher_a_chap_1_sec_1.html/chapter1%2Fsection1

The reason this is not working is because the third `:`-delimited part of the URN is URI-encoded when constructed. When the resulting URN is then encoded as a query param, the result is that the part after `x-dc` should be double-URI encoded, eg: `/api/search?uri=urn%3Ax-dc%3Ahttp%253A%252F%252Fpublisher.org%252Fbook%2Fchapter1`.

The client does this when constructing the fingerprint that is used to search for annotations.

> I do /not/ expect results from this query:

[...]

PR #4643, opened 2017-09-07
Validate the syntax of DOI values

robertknight:
**Depends on https://github.com/hypothesis/h/pull/4644**

For context, this relates to the work done to support annotating EPUBs recently. See https://github.com/hypothesis/h/issues/4611

----

Validate the syntax of DOI values before creating document equivalence claim entries in the DB which are labeled as being of type `dc-doi` or `highwire-doi`.

Not all Dublin Core identifier field values (from `<meta name="dc.identifier">` tags in HTML documents) are DOIs.

Additionally, the identifier value is only intended to be "unambiguous within a given context" [1] which I take to imply that it is not necessarily globally unique and indeed, looking at actual values we have captured so far from these tags in the prod h DB, this has been the case for a small number of entries. See comments below for details.

In the process of making this change, I discovered that one of the example DOIs we were using in many test cases was wrong, perhaps due to a typo that got copy-pasta'd.

[1] http://dublincore.org/documents/dcmi-terms/#terms-identifier

Fixes https://github.com/hypothesis/h/issues/4611

### DIFF
--- a/h/util/document_claims.py
+++ b/h/util/document_claims.py
@@ -14,6 +14,16 @@ and returned.
 
 from __future__ import unicode_literals
 
+import re
+
+
+# Pattern that matches DOIs (eg. "10.1000/123456").
+#
+# See https://www.doi.org/doi_handbook/2_Numbering.html#2.2 The registrant code
+# is not specified as numeric there but is currently always a 4+ digit value.
+# See https://www.doi.org/overview/DOI_article_ELIS3.pdf.
+DOI_PATTERN = re.compile('10\.[0-9]{4,}[.0-9]*/.*')
+
 
 def document_uris_from_data(document_data, claimant):
     """
@@ -262,10 +272,8 @@ def doi_uri_from_string(s):
 
     If the given string doesn't already start with "doi:" it is prepended.
 
-    If the given string does not contain a DOI URI then None is returned.
-    Examples of strings that do not include a DOI are:
-    '', ' ', 'doi:', 'doi: ', ' doi:', ' doi: '.
-
+    If the given string, minus any `doi:` prefix does not match the syntax
+    for DOI names ("10.NNNN/.*") then `None` is returned.
     """
     s = s.strip()
 
@@ -274,7 +282,7 @@ def doi_uri_from_string(s):
 
     s = s.strip()
 
-    if not s:
+    if DOI_PATTERN.match(s) is None:
         return None
 
     s = 'doi:{}'.format(s)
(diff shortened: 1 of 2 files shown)


robertknight:
Here is a dump of all the DOI values (not unique) that have been captured from annotated documents on the prod instance of h in the `document_uri` table as `dc-doi` or `highwire-doi` URLs: https://gist.github.com/robertknight/9232b5b3cc6c6ba29c62e64cc80e1513 . As expected, there are a lot of non-DOI values in there.

However, there were 1122 values which are URLs that point to the DOI resolver, eg. https://doi.org/10.1016/j.stem.2017.05.001 . We probably want to extract the DOI from those and normalize them to `doi:<doi>`. This should enable a page with `<meta name="dc.identifier" content="https://doi.org/10.1016/j.stem.2017.05.001">` to show the same annotations as a page with `<meta name="dc.identifier" content="10.1016/j.stem.2017.05.001">`.

This would imply a bunch of extra work however, such as a migration of existing data. For the time being I'd propose that we simply continue to capture such values.

robertknight:
I'm going to rebase this on https://github.com/hypothesis/h/pull/4644

dwhly:
> This would imply a bunch of extra work however, such as a migration of existing data. For the time being I'd propose that we simply continue to capture such values.

We could start normalizing them now, and clean up the data later though right?

robertknight:
> We could start normalizing them now, and clean up the data later though right?

If you change the part of a system that captures data, you also have to consider what happens in the parts of the system that query data and what happens when querying data that was captured before & after the change. I haven't thought it through fully in this context, but just filtering what we already capture avoids those concerns in this PR.

seanh:
Rebased

h/util/document_claims.py
```diff
@@ -14,6 +14,19 @@
...
+
+
+# Pattern that matches DOIs (eg. "10.1000/123456").
+#
+# See https://www.doi.org/doi_handbook/2_Numbering.html#2.2 The registrant code
+# is not specified as numeric there but is currently always a 4+ digit value.
+# See https://www.doi.org/overview/DOI_article_ELIS3.pdf.
+#
+# This also accepts DOIs represented as URLs (eg.
+# "https://doi.org/10.1000/123456").
```
seanh: Elsewhere in h we seem to use the `#:` comment style for docstrings of constants and things like this. (autodoc understands it)

h/util/document_claims.py
```diff
@@ -262,10 +275,8 @@ def doi_uri_from_string(s):
 
     If the given string doesn't already start with "doi:" it is prepended.
 
-    If the given string does not contain a DOI URI then None is returned.
-    Examples of strings that do not include a DOI are:
-    '', ' ', 'doi:', 'doi: ', ' doi:', ' doi: '.
-
+    If the given string, minus any `doi:` prefix does not match the syntax
+    for DOI names ("10.NNNN/.*") then `None` is returned.
```
seanh: I think these should be:

    ``...``

not

    `...`

because they're rSt not markdown. We don't actually extract and parse them as rSt anywhere currently though so no biggie.

PR #4644, opened 2017-09-07
Reduce duplication in DOI tests

robertknight:
There were essentially two sets of tests for the `doi_uri_from_string`
function, one that tested usage for strings extracted from `<meta
name="dc.identifier">` tags and one for strings extracted from `<meta
name="citation_doi">` tags.

This commit adds a class that tests the `doi_uri_from_string` function
directly in detail and simplifies the indirect testing via
`document_uris_from_{dc, highwire_doi}`.

### DIFF
--- a/tests/h/util/document_claims_test.py
+++ b/tests/h/util/document_claims_test.py
@@ -2,9 +2,11 @@
 
 from __future__ import unicode_literals
 
+import re
 import pytest
 
 from h.util import document_claims
+from h.util.document_claims import doi_uri_from_string
 
 
 class TestDocumentURIsFromLinks(object):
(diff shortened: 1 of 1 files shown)


tests/h/util/document_claims_test.py
```diff
@@ -496,50 +486,14 @@ def test_dc_identifiers_produce_dc_doi_document_uris(self):
...
-        for doi in (' foo ', 'doi: foo ', ' doi:foo ', ' doi: foo '):
-            dc_dict = {'identifier': [doi]}
-
-            document_uris = document_claims.document_uris_from_dc(
-                dc_dict,
-                claimant='http://example.com/example.html',
-            )
-
-            assert [d['uri'] for d in document_uris] == ['doi:foo']
+        assert len(document_uris) == 0
```
seanh: Stylistically I think this should be `assert not document_uris` in Python (pylint would complain about this I believe)

PR #4645, opened 2017-09-08
Render users' display names in annotation API responses

chdorner:
It will be much easier to review commit-by-commit as I had to shuffle quite a bit of code around.

This is adding a new `"user_info": {"display_name": "John Doe"}` object to the annotation JSON response. It is also behind a feature flag, mostly to test out and have a quick way out in case it puts too much pressure on the database.

### DIFF
--- a/h/models/user.py
+++ b/h/models/user.py
@@ -103,6 +103,23 @@ class UserIDComparator(Comparator):
                                   val['domain'])
         return self.__clause_element__() == other
 
+    def in_(self, userids):
+        others = []
+        for userid in userids:
+            try:
+                val = split_user(userid)
+            except ValueError:
+                continue
+
+            other = sa.tuple_(_normalise_username(val['username']),
+                              val['domain'])
+            others.append(other)
+
+        if not others:
+            return False
+
+        return self.__clause_element__().in_(others)
+
 
 class UserFactory(object):
     """Root resource for routes that look up User objects by traversal."""
--- a/h/services/user.py
+++ b/h/services/user.py
@@ -73,6 +73,48 @@ class UserService(object):
 
         return self._cache[cache_key]
 
+    def fetch_all(self, userids):
+        """
+        Fetch a list of users by their userids.
+
+        This function fetches users based on the list, adds them to the internal
+        cache and then returns the list of users. This is especially useful
+        when needing to access multiple user objects without loading them one-by-one.
+
+        It will only attempt to load the users that aren't already cached.
+
+        Userids that cannot be found will not be in the cache, so subsequent calls to `.fetch`
+        are trying to load them again.
+
+        :param userids: a list of userid strings.
+
+        :returns: a list with the found user instances
+        :rtype: list of h.models.User
+        """
+        if not userids:
+            return []
+
+        cache_keys = {}
+        for userid in userids:
+            try:
+                val = split_user(userid)
+                key = (val['username'], val['domain'])
+                cache_keys[key] = userid
+            except ValueError:
+                continue
+
+        userid_tuples = set(cache_keys.keys())
+        missing_tuples = userid_tuples - set(self._cache.keys())
+        missing_ids = [v for k, v in cache_keys.iteritems() if k in missing_tuples]
+
+        if missing_ids:
+            users = self.session.query(User).filter(User.userid.in_(missing_ids))
+            for user in users:
+                cache_key = (user.username, user.authority)
+                self._cache[cache_key] = user
+
+        return [v for k, v in self._cache.iteritems() if k in cache_keys.keys()]
+
     def fetch_for_login(self, username_or_email):
         """
         Fetch a user by data provided in the login field.
--- a/h/services/user.py
+++ b/h/services/user.py
@@ -5,8 +5,8 @@ from __future__ import unicode_literals
 import sqlalchemy as sa
 
 from h.models import User
-from h.util.db import lru_cache_in_transaction
 from h.util.user import split_user
+from h.util.db import on_transaction_end
 
 UPDATE_PREFS_ALLOWED_KEYS = set(['show_sidebar_tutorial'])
 
@@ -29,7 +29,13 @@ class UserService(object):
         self.default_authority = default_authority
         self.session = session
 
-        self._cached_fetch = lru_cache_in_transaction(self.session)(self._fetch)
+        # Local cache of fetched users.
+        self._cache = {}
+
+        # But don't allow the cache to persist after the session is closed.
+        @on_transaction_end(session)
+        def flush_cache():
+            self._cache = {}
 
     def fetch(self, userid_or_username, authority=None):
         """
@@ -56,7 +62,16 @@ class UserService(object):
             username = parts['username']
             authority = parts['domain']
 
-        return self._cached_fetch(username, authority)
+        # The cache is keyed by (username, authority) tuples.
+        cache_key = (username, authority)
+
+        if cache_key not in self._cache:
+            self._cache[cache_key] = (self.session.query(User)
+                                      .filter_by(username=username)
+                                      .filter_by(authority=authority)
+                                      .one_or_none())
+
+        return self._cache[cache_key]
 
     def fetch_for_login(self, username_or_email):
         """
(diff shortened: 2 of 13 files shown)


robertknight:
I've given this a test together with https://github.com/hypothesis/client/pull/541 and first party accounts. Notes so far:

- The display name is correctly rendered in the response for search queries and by the real time API
- The display name is not currently included in the `/api/profile` response, probably a matter for a separate PR. We'll need that in order to render the display name in new annotations which are currently being drafted and in the account menu.

h/services/user.py
```diff
@@ -29,7 +29,13 @@ def __init__(self, default_authority, session):
         self.default_authority = default_authority
         self.session = session
 
-        self._cached_fetch = lru_cache_in_transaction(self.session)(self._fetch)
```
robertknight: We're dropping the LRU cache because that would end up prematurely discarding `User` instances that are being prefetched before rendering annotations?

Are there any situations in our code where we have long-lived transactions and keeping `User` instances around for an arbitrarily long time would be a problem?

h/presenters/annotation_json.py
```diff
@@ -64,6 +69,16 @@ def asdict(self):
         return annotation
 
     @property
+    def user_info(self):
```
robertknight: What is the reason for adding this as a property on the `AnnotationResource` rather than adding this attribute via a formatter as we do for eg. moderation info?

h/services/user.py
```diff
@@ -56,7 +62,55 @@ def fetch(self, userid_or_username, authority=None):
...
+
+        It will only attempt to load the users that aren't already cached.
+
+        :param userids: a list of userid strings.
+
+        :returns: a list with the found user instances
+        :rtype: list of h.models.User
+        """
+        if not userids:
+            return []
```
robertknight: This is just a shortcut if the list of passed userids is empty? If so, is it really needed as it adds an extra path to test?

h/services/user.py
```diff
@@ -56,7 +62,55 @@ def fetch(self, userid_or_username, authority=None):
...
+
+    def fetch_all(self, userids):
+        """
+        Fetch a list of users by their userids.
+
+        This function fetches users based on the list, adds them to the internal
+        cache and then returns the list of users. This is especially useful
+        when needing to access multiple user objects without loading them one-by-one.
+
+        It will only attempt to load the users that aren't already cached.
```
robertknight: This docstring could be a little clearer about what happens if a userid is not found.

tests/h/models/user_test.py
```diff
@@ -123,6 +123,57 @@ def test_userid_as_class_property_invalid_userid(db_session):
...
+    db_session.add_all([fred, alice])
+    db_session.flush()
+
+    result = (db_session.query(models.User)
+              .filter(models.User.userid.in_(['acct:fredbloggs@example.net',
+                                              'acct:alicewrites@foobar.com',
+                                              'acct:missing@bla.org']))
+              .all())
+
+    assert len(result) == 2
```
robertknight: I wonder if you can write this a one liner like `set(result) == set([fred, alice])` 🤔

h/models/user.py
```diff
@@ -103,6 +103,23 @@ def __eq__(self, other):
                                   val['domain'])
         return self.__clause_element__() == other
 
+    def in_(self, userids):
```
robertknight: So this is a helper used to generate `<user ID expr> IN (<other user ID 1 expr>, <other user ID 2 expr>, ...)` SQL expressions?

h/services/user.py
```diff
@@ -56,7 +62,55 @@ def fetch(self, userid_or_username, authority=None):
...
             authority = parts['domain']
 
-        return self._cached_fetch(username, authority)
+        # The cache is keyed by (username, authority) tuples.
+        cache_key = (username, authority)
+
+        if cache_key not in self._cache:
+            self._cache[cache_key] = (self.session.query(User)
+                                      .filter_by(username=username)
+                                      .filter_by(authority=authority)
```
robertknight: The two `filter_by` queries could I think be combined into a single `filter_by(username=username, authority=authority)`. I don't know if this makes a difference to the SQL generated?

tests/h/resources_test.py
```diff
@@ -71,30 +78,36 @@ def links_service(self, pyramid_config):
...
-        res = AnnotationResource(ann, group_service, links_service)
+        res = AnnotationResource(ann, group_service, links_service, user_service)
 
         result = res.link('json')
 
         links_service.get.assert_called_once_with(ann, 'json')
         assert result == links_service.get.return_value
 
-    def test_acl_private(self, factories, group_service, links_service):
+    def test_acl_private(self, factories, group_service, links_service, user_service):
```
robertknight: Hmm, there is quite a lot of repeated plumbing here. This is part of the reason why I wonder about adding a `user` property to `AnnotationResource` vs generating `user_info` with a formatter.

tests/h/streamer/messages_test.py
```diff
@@ -243,97 +253,90 @@ def test_notification_format_delete(self, fetch_annotation, presenter_asdict):
...
+    def test_no_send_if_action_is_read(self, db_session, presenter_asdict):
         """Should return None if the message action is 'read'."""
         message = {'action': 'read', 'src_client_id': '_', 'annotation_id': '_'}
         socket = FakeSocket('giraffe')
-        session = mock.sentinel.db_session
         settings = {'foo': 'bar'}
         presenter_asdict.return_value = self.serialized_annotation()
 
-        messages.handle_annotation_event(message, [socket], settings, session)
+        messages.handle_annotation_event(message, [socket], settings, db_session)
```
robertknight: This renaming is a helpful clarification 👍

tests/h/streamer/messages_test.py
```diff
@@ -243,97 +253,90 @@ def test_notification_format_delete(self, fetch_annotation, presenter_asdict):
...
 
         assert socket.send_json_payloads == []
 
-    def test_sends_nipsad_annotations_to_owners(self, fetch_annotation, nipsa_service, presenter_asdict):
+    def test_sends_nipsad_annotations_to_owners(self, db_session, fetch_annotation, nipsa_service, presenter_asdict):
         """NIPSA'd users should see their own annotations."""
         message = {'action': '_', 'src_client_id': '_', 'annotation_id': '_'}
-        fetch_annotation.return_value.userid = 'fred'
+        fetch_annotation.return_value.userid = u'acct:fred@example.net'
```
robertknight: 👍 for example inputs & outputs in tests that resemble real inputs & outputs.

h/presenters/annotation_json.py
```diff
@@ -64,6 +69,16 @@ def asdict(self):
...
+        user = self.annotation_resource.user
+        if not user:
+            return {}
```
robertknight: Hmm ... what situations can we expect `user` to be null in?

h/services/user.py
chdorner: There's no way to warm up the LRU cache, as it's memoizing method calls by the arguments it is called with. So there's no way of initializing the cache with a bulk query and then read from the cache by only passing in single ids.

h/models/user.py
chdorner: Yes, it's a custom implementation of the default SQLAlchemy `sqlalchemy.orm.Query.in_` function. We don't store the userid in the user table, which is why we need to implement these custom methods that translate a userid to a username and authority. So far we only needed the `a = b` comparator function (`__eq__`), but now we want to do a `IN` query which means we have to implement this ourselves.

h/presenters/annotation_json.py
chdorner: Pure laziness to be honest. I'll have a look at doing it as a formatter today. Formatters usually have their own cache, but I think it would still be better to reuse the user service cache for this.

At least it will make the presenter nicer.

h/services/user.py
chdorner: I've just copy + pasted what we had before. They would be the same, but in the end, it doesn't really matter.

h/services/user.py
chdorner: It's good practice to return early in these cases, as sending an empty `IN` query (`column IN()`) can take quite long to execute, even though the result is always empty.

tests/h/models/user_test.py
chdorner: Sure it's possible. I think the error messages are easier to interpret when writing the test like I did.

h/services/user.py
chdorner: Fixed.

h/services/user.py
chdorner: It's actually just from a `git revert ...`

tests/h/models/user_test.py
robertknight: OK 👍

PR #4647, opened 2017-09-12
Work around limitations of `window.postMessage` in IE 11

robertknight:
In IE 11, postMessage support is limited to frames embedded within a
page and doesn't work between a popup window and its opener. This means
that OAuth clients wanting to support this browser will need to use a
redirect rather than the "web_message" response mode.

For the specific use case of the Hypothesis client, we can work around
the problem by manually dispatching the "message" event, in the target
window, that a `postMessage` call usually results in. This only works
because the sidebar's "app.html" and the auth popup are served from the
same domain.

IE 11 usage is currently around ~1% of client views, so it seems likely
that we might just be able to drop support a few months from now.

----

Testing this without IE can be done by changing the `const isIE` initialization to `const isIE = true`. If actually testing in IE, you'll need to open the dev tools at the "Network" tab and check the "Always refresh from the server" toolbar icon to work around some caching issues that currently affect XHR responses, at least in development.

### DIFF
--- a/h/static/scripts/post-auth.js
+++ b/h/static/scripts/post-auth.js
@@ -10,15 +10,50 @@
 
 const settings = require('./base/settings')(document);
 
-if (window.opener) {
+function sendAuthResponse() {
+  if (!window.opener) {
+    console.error('The client window was closed');
+    return;
+  }
+
   const msg = {
     type: 'authorization_response',
     code: settings.code,
     state: settings.state,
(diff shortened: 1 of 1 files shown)


seanh:
Confirmed - on master I can't login with OAuth in IE11, but on this branch I can

seanh:
:+1:

PR #4648, opened 2017-09-12
Require revalidation before using cached API responses

robertknight:
Require the browser to revalidate any API responses using a request with
an `If-None-Match: <etag of last response>` header before using a
cached response.

This fixes an issue where IE 11 would cache responses to eg.
`/api/profile` requests and then re-use them even if the `Authorization`
header had changed. Other browsers have different heuristics and were
already revalidating the cached response in this case.

I'm currently doing additional testing with modern browsers to check whether adding this header has any unexpected effects on how they handle caching of API responses.

### DIFF
--- a/h/app.py
+++ b/h/app.py
@@ -54,6 +54,7 @@ def includeme(config):
     config.add_tween('h.tweens.redirect_tween_factory')
     config.add_tween('h.tweens.csrf_tween_factory')
     config.add_tween('h.tweens.security_header_tween_factory')
+    config.add_tween('h.tweens.cache_header_tween_factory')
 
     config.add_request_method(in_debug_mode, 'debug', reify=True)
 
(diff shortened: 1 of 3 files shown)


robertknight:
Testing with a trivial Node app with an endpoint that returns the current time in a JSON string, I see that all modern browsers (Chrome, Firefox, Safari, Edge) do not cache XHR responses if no caching headers are present but an ETag is, and send `If-None-Match` headers to revalidate on subsequent requests. IE 11 however does cache these responses unless `Cache-Control: no-cache` is set.

seanh:
Rebased

seanh:
I still seem to be seeing issues with the search API on this branch. For example login to the client on the /docs/help page with OAuth, I don't see my annotations until I reload the page. Then log out, and I can still see my private annotations.

seanh:
Noting that there's a separate issue with the client not updating the annotations when you login or out with OAuth.

Confirmed that on master logging with OAuth does not update the groups dropdown menu properly. On this branch it does.

PR #4651, opened 2017-09-13
Add `user_info` object to `GET /api/profile` response

chdorner:
This is in addition to #4645, so that the client can render the display name for new annotation cards that haven't been synced to the backend yet.


chdorner:
This should be ready again.

Issue #4652, opened 2017-09-14
Deleting a user does not delete their annotations from ElasticSearch


robertknight:
### Steps to reproduce

1. Create a user account in a local h instance
2. Make an annotation using that account
3. Delete the user using the admin panel
4. Query ElasticSearch for annotations using:
    ```
    curl 'http://localhost:9200/_all/annotation/_search?q=user:<username>' | jq
    ```

### Expected behaviour
No annotations should be returned.

### Actual behaviour
The annotation made in step 2 is returned in the JSON response in step 4. If you go to the annotated page and reload, the annotation is still visible.

### Additional notes

It looks like `h.views.admin_users.delete_user` does not issue the same annotation "delete" event that `h.views.api.delete` does, which notifies the indexer and streamer about the deletion.

If after performing the steps above, you then try to recreate and re-delete the account, the second deletion attempt fails:

1. Follow all the steps for the first issue
2. Create a new user account with the same username after deleting the account
3. Try to delete that user account. This fails with a 500 error.

judell:
Nice catch.

PR #4653, opened 2017-09-14
Look up annotations in DB when deleting user

robertknight:
This fixes a secondary issue uncovered after I discovered https://github.com/hypothesis/h/issues/4652

Given that finding annotations created by a particular user only
requires a query on the indexed `annotation.userid` column, we can do
the lookup directly in the source of truth (the DB) rather than ES.

This avoids a hazard where the ES helper query could fail to load an
annotation if for any reason ES is not fully in sync with the state of
the DB. In the case of #4652, the index was not in sync because the annotation
doesn't get deleted from ES. Even when that is resolved, there will be a delay between
the annotation being deleted from the DB and it being deleted from ES.

### DIFF
--- a/h/views/admin_users.py
+++ b/h/views/admin_users.py
@@ -2,7 +2,6 @@
 
 from __future__ import unicode_literals
 
-from elasticsearch import helpers as es_helpers
 import jinja2
 from pyramid import httpexceptions
 from pyramid.view import view_config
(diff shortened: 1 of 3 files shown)


chdorner:
Looks good. I'll wait until we have #4654 figured out because it might mean changes to this code.

robertknight:
I'm going to merge this now and rework the other PR on top of these changes.

PR #4654, opened 2017-09-14
Move logic to publish annotation events into `h.storage`

### LINKED ISSUE #4652: Deleting a user does not delete their annotations from ElasticSearch

robertknight:
Ensure that when `h.storage.{create, update, delete}_annotation` methods
are used outside of `h.views.api`, the correct annotation events are
published so that the streamer and ElasticSearch indexes are updated
correctly.

This fixes an issue where deleting users did not remove their annotations from ElasticSearch, though it should have done.

This bug probably means that we have a whole bunch of annotations in the prod ES
instance that are associated with user accounts that no longer exist. I'm not sure that reindexing will fix that because it doesn't remove invalid entries from ES AFAIK.

Fixes #4652

### DIFF
--- a/h/storage.py
+++ b/h/storage.py
@@ -23,6 +23,7 @@ from pyramid import i18n
 
 from h import models, schemas
 from h.db import types
+from h.events import AnnotationEvent
 from h.models.document import update_document_metadata
 
 _ = i18n.TranslationStringFactory(__package__)
@@ -145,10 +146,12 @@ def create_annotation(request, data, group_service):
     request.db.add(annotation)
     request.db.flush()
 
+    _publish_annotation_event(request, annotation, 'create')
+
     return annotation
 
 
-def update_annotation(session, id_, data):
+def update_annotation(request, id_, data):
     """
     Update an existing annotation and its associated document metadata.
 
@@ -175,7 +178,7 @@ def update_annotation(session, id_, data):
     # annotation object.
     document = data.pop('document', None)
 
-    annotation = session.query(models.Annotation).get(id_)
+    annotation = request.db.query(models.Annotation).get(id_)
     annotation.updated = updated
 
     annotation.extra.update(data.pop('extra', {}))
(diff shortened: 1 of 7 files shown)


chdorner:
> I'm not sure that reindexing will fix that because it doesn't remove invalid entries from ES AFAIK.

Reindexing does fix the problem because we're indexing into a brand new index, although just temporarily of course until the next user gets deleted.

---

While it is tempting to fix the bug by moving the code that emits the event into the storage functions, I don't think that's a good idea. You can only emit an event by having access to an instance of the request. So by doing that you now have added another responsibility to the storage function. We deliberately tried to avoid having to pass the request into the storage functions, currently only `storage.create_annotation` needs it for checking that the user is allowed to write annotations to the specified group. In hindsight, I would even change that part of the code to only pass in a reference to the `has_permission` function, like we do [in other places](https://github.com/hypothesis/h/blob/master/h/services/annotation_json_presentation.py#L75).

I quickly went through all the storage functions that would need to emit an event and checked how often they are called to see how much of an issue it is to make the caller emit the event:

* `storage.create_annotation`: Only called in the `create` API view
* `storage.update_annotation`: Only called in the `update` API view
* `storage.delete_annotation`: Called in the `delete` API view, and when deleting a user (where the bug currently appears).

I think the better solution to make sure the callers emit the events, which in this case means adding that bit of code to the user deletion code.

robertknight:
> While it is tempting to fix the bug by moving the code that emits the event into the storage functions, I don't think that's a good idea. 

Hmm ... I suppose the idiomatic way to handle what `h.storage` does now would be to create a service? Our other services do have access to the `request` object.

For the moment I'll close this and publish a more minimal change which does what you suggest and publish the event from the `delete_user` view function. I do think we should consider centralising the logic somewhere in future to avoid someone forgetting to publish an event when there is another entry point that needs to perform this operation.

PR #4658, opened 2017-09-19
Allow to set display name when signing users up through API

chdorner:
This is part of hypothesis/product-backlog#323.

It allows API clients who use the `POST /api/users` endpoint to set the display name of a user at the time the user is getting signed up.

### DIFF
--- a/h/accounts/schemas.py
+++ b/h/accounts/schemas.py
@@ -10,6 +10,7 @@ from h import i18n, models, validators
 from h.accounts import util
 from h.services.user import UserNotActivated
 from h.models.user import (
+    DISPLAY_NAME_MAX_LENGTH,
     EMAIL_MAX_LENGTH,
     USERNAME_MAX_LENGTH,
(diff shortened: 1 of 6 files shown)


h/accounts/schemas.py
```diff
@@ -377,6 +377,10 @@ class CreateUserAPISchema(JSONSchema):
                 'type': 'string',
                 'format': 'email',
             },
+            'display_name': {
+                'type': 'string',
+                'maxLength': 30,
```
robertknight: We currently have the length limits for usernames and display names duplicated in multiple places in the code. There are some constants at the top of `h/models/user.py`. Could we use those here?

chdorner:
> avoid duplicating the length constraints on fields in multiple places if it isn't too hard to do so.

Yup, that's a good point. I didn't even see those constants. I've also replaced all the other hardcoded numbers in that schema with the constants.

PR #4659, opened 2017-09-19
Update user API endpoint

chdorner:
This adds a `PATCH /api/users/{username}` endpoint. It finds the user based on the given username in the path and the authority of the client that is used to authenticate the request.

It is based on #4658 and will need rebasing once that one has been merged.

### DIFF
--- a/tests/h/accounts/schemas_test.py
+++ b/tests/h/accounts/schemas_test.py
@@ -665,6 +665,59 @@ class TestCreateUserAPISchema(object):
         return schemas.CreateUserAPISchema()
 
 
+class TestUpdateUserAPISchema(object):
+    def test_it_raises_when_email_empty(self, schema, payload):
+        payload['email'] = ''
+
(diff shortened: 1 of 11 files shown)


tests/h/accounts/schemas_test.py
```diff
@@ -665,6 +665,59 @@ def schema(self):
         return schemas.CreateUserAPISchema()
 
 
+class TestUpdateUserAPISchema(object):
+    def test_it_raises_when_email_empty(self, schema, payload):
```
robertknight: Hmm ... these tests duplicate quite a lot of the code for the above tests for creating users. It would be nice to avoid that.

chdorner: IMO it doesn't make much sense to de-duplicate the tests if the schema implementation is duplicated as well. But there's probably something to look into in the future to de-duplicate the schemas and also linking them from the API docs, so we don't have to remember to update them every time.

PR #4660, opened 2017-09-19
Replace legacy styles for help page

robertknight:
This reworks the styling of the help page at `/docs/help` and `/welcome` to remove the use of legacy style code in `h/static/styles/partials`, including a lot of unused cruft, which doesn't follow the [conventions](https://github.com/hypothesis/frontend-toolkit/blob/master/docs/css-style-guide.md) used for the styles for the rest of the website. This should avoid some potential confusion for new contributors.

Visually the result should look and behave almost the same except for a change to make the masthead "h" logo the same dark grey color as elsewhere on the site instead of the red brand color.

After this is merged I will clean up the naming in the `h/static/styles` directory by renaming `partials-v2` to `partials`.


chdorner:
It looks like there is a problem with the spacing, at least in Chrome.

Before:
before

After:
after

robertknight:
> It looks like there is a problem with the spacing, at least in Chrome.

Good spot. On the window sizes I'd been testing the text wrapped in a way that it wasn't so noticeable. Fixed.

PR #4661, opened 2017-09-19
Publish 'delete' event for annotations deleted when deleting a user

### LINKED ISSUE #4652: Deleting a user does not delete their annotations from ElasticSearch

robertknight:
_This is a replacement for https://github.com/hypothesis/h/pull/4654_

This ensures that clients of the Real Time API and ElasticSearch indexes
are updated correctly after deleting a user.

Fixes #4652

### DIFF
--- a/tests/h/views/admin_users_test.py
+++ b/tests/h/views/admin_users_test.py
@@ -9,6 +9,7 @@ from mock import call
 from pyramid import httpexceptions
 import pytest
 
+from h.events import AnnotationEvent
 from h.services.annotation_stats import AnnotationStatsService
 from h.services.user import UserService
 from h.models import Annotation
@@ -265,6 +266,20 @@ def test_delete_user_deletes_annotations(api_storage, db_session, factories, pyr
 
 
 @delete_user_fixtures
+def test_delete_user_publishes_event(api_storage, db_session, factories, matchers, pyramid_request):
+    pyramid_request.db = db_session
+    user = factories.User()
+    ann = factories.Annotation(userid=user.userid)
+
+    delete_user(pyramid_request, user)
+
+    expected_event = AnnotationEvent(pyramid_request, ann.id, 'delete')
+    actual_event = pyramid_request.notify_after_commit.call_args[0][0]
+    assert (expected_event.request, expected_event.annotation_id, expected_event.action) == \
+           (actual_event.request, actual_event.annotation_id, actual_event.action)
+
+
+@delete_user_fixtures
 def test_delete_user_deletes_user(db_session, factories, pyramid_request):
     pyramid_request.db = db_session
     user = factories.User()
@@ -276,7 +291,7 @@ def test_delete_user_deletes_user(db_session, factories, pyramid_request):
 
 @pytest.fixture
 def pyramid_request(pyramid_request):
-    pyramid_request.es = mock.MagicMock()
+    pyramid_request.notify_after_commit = mock.Mock()
     return pyramid_request
 
 
--- a/h/views/admin_users.py
+++ b/h/views/admin_users.py
@@ -9,6 +9,7 @@ from pyramid.view import view_config
 from h import models
 from h import storage
 from h.accounts.events import ActivationEvent
+from h.events import AnnotationEvent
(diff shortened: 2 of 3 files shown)


tests/h/views/admin_users_test.py
```diff
@@ -265,6 +266,19 @@ def test_delete_user_deletes_annotations(api_storage, db_session, factories, pyr
...
+def test_delete_user_publishes_event(api_storage, db_session, factories, matchers, pyramid_request):
+    pyramid_request.db = db_session
+    user = factories.User()
+    ann = factories.Annotation(userid=user.userid)
+
+    delete_user(pyramid_request, user)
+
+    event = AnnotationEvent(pyramid_request, ann.id, 'delete')
+    pyramid_request.notify_after_commit.assert_called_once_with(
+        matchers.same_dicts(event))
```
chdorner: I'm not sure if comparing `__dict__` is a good idea, it works for this occasion but I can see some weird issues popping up in the future. According to the Python doc `__dict__` returns "a dictionary or other mapping object used to store an object’s (writable) attributes." - Technically this doesn't ensure that it is the same object, just that it has the same writable attributes.
Furthermore, I don't think that anybody who didn't first see the implementation of the matcher will know what `same_dicts` does.

I think it might be better and more readable, although also more verbose, to do this:

```
@delete_user_fixtures
def test_delete_user_publishes_event(api_storage, db_session, factories, matchers, pyramid_request):
    pyramid_request.db = db_session
    user = factories.User()
    ann = factories.Annotation(userid=user.userid)

    delete_user(pyramid_request, user)

    event = pyramid_request.notify_after_commit.call_args[0][0]
    assert event.annotation_id == ann.id
    assert event.action == 'delete'
```
At least this is using `mock`'s `call_args`, and the check that the event is the same is more explicit.

h/events.py
```diff
@@ -9,6 +9,9 @@ def __init__(self, request, annotation_id, action):
         self.annotation_id = annotation_id
         self.action = action
 
+    def __eq__(self, other):
+        return self.__dict__ == other.__dict__
```
chdorner: We should add tests for this method to make sure that it actually does what it's supposed to.

h/events.py
```diff
@@ -24,3 +27,6 @@ def __init__(self, request, annotation, annotation_dict):
         self.request = request
         self.annotation = annotation
         self.annotation_dict = annotation_dict
+
+    def __eq__(self, other):
+        return self.__dict__ == other.__dict__
```
chdorner: This method is not being called anywhere at the moment. We should either remove it, or use it in the tests (and also test the equality method itself).

robertknight: I've been looking into this further and it looks like making classes robustly value-like in Python is actually a bit fiddly and there isn't a convenient mixin or decorator in the stdlib that I can just pull in that I can see. 🙁

For the moment I've gone with something like your original suggestion which is just to compare the field values of the actual and expected events directly in the test.

robertknight:
Hi @chdorner - I've revised the tests along the lines of your original suggestion, to compare the fields of the object passed to `notify_after_commit` directly. After some fiddling around I found out that trying to give the event classes value semantics is more fuss than it is worth for this PR. This should be good to go now.

PR #4665, opened 2017-09-21
Support email addresses with international domain names

### LINKED ISSUE #4662: Sign up form rejects valid international email addresses (Unicode & ACE)
### Steps to reproduce
1. Go to the sign up form at https://hypothes.is/signup
2. Try entering an email address with multiple hyphens in the domain name. This is common in international domain names (IDN) where the fallback ASCII encoding (ACE), for systems that do not support unicode chars in domain names/email addresses, results in domain names starting with "xn--"

The original context is that a user was trying to use an international email address (with non-ASCII chars in the domain name) and then when that failed, tried to use the ACE version.

Support issue: https://hypothesis.zendesk.com/agent/tickets/1479

### Expected behaviour
I should be able to sign up with an email address that contains multiple hyphens.

### Actual behaviour
The sign up form reports that the email is invalid.

### Notes
It looks like Colander is using a [regex](https://github.com/Pylons/colander/blob/995d3686720f7e5054dbe03ed79aea398eea148d/colander/__init__.py#L351) for validating email addresses, which incorrectly disallows multiple hyphens.

See https://www.rnids.rs/en/cyrillic-on-internet/idn-encoder for more details on IDNs.

See https://github.com/Pylons/colander/issues/283

robertknight:
`colander.Email` uses a regex which does not allow domain names
beginning with `xn--` or Unicode characters. Either of these limitations
prevents users from registering with email addresses that use
international domain names.

This commit replaces the validator with an alternate pattern taken from
the HTML spec which is used by Chrome and which at least allows the
Punycode version of an internationalized email address to be used.

I opted not to just allow unicode chars because we first have to make
sure that the rest of our stack supports it. Alternatively we could
convert to Punycode automatically when deserializing the user's input.

Fixes #4662

### DIFF
--- a/tests/h/validators_test.py
+++ b/tests/h/validators_test.py
@@ -1,0 +1,35 @@
+import pytest
+
+from mock import Mock
+
+import colander
+
+from h.validators import Email
+
+
+class TestEmail(object):
+    @pytest.mark.parametrize('email', [
+        'jimsmith@foobar.com',
+        'international@xn--domain.com',
+        'jim.smith@gmail.com',
+        'jim.smith@foo.bar.com',
+    ])
+    def test_accepts_valid_addresses(self, schema_node, email):
+        validator = Email()
+        validator(schema_node, email)
+
+    @pytest.mark.parametrize('email', [
+        ' spaces @ spaces.com',
+    ])
+    def test_rejects_invalid_addresses(self, schema_node, email):
+        validator = Email()
+
+        with pytest.raises(colander.Invalid):
+            validator(schema_node, email)
+
+    @pytest.fixture
+    def schema_node(self):
+        """
+        Mock for `colander.SchemaNode` arg that validator callables require.
+        """
+        return Mock(spec_set=[])
--- a/tests/h/validators_test.py
+++ b/tests/h/validators_test.py
@@ -1,8 +1,10 @@
 # -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+import pytest
+from mock import Mock
 
 import colander
-from mock import Mock
-import pytest
 
 from h.validators import Email
 
--- a/tests/h/validators_test.py
+++ b/tests/h/validators_test.py
@@ -1,3 +1,5 @@
+# -*- coding: utf-8 -*-
+
 import pytest
 
 from mock import Mock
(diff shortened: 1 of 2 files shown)


robertknight:
Gah. Forgot to re-run *all* the tests after a late change.

tests/h/validators_test.py
```diff
@@ -0,0 +1,35 @@
...
+        'jim.smith@gmail.com',
+        'jim.smith@foo.bar.com',
+    ])
+    def test_accepts_valid_addresses(self, schema_node, email):
+        validator = Email()
+        validator(schema_node, email)
+
+    @pytest.mark.parametrize('email', [
+        ' spaces @ spaces.com',
+    ])
```
chdorner: I think we need some more examples of invalid emails, like missing the TLD, or missing the username part, or not containing an "@", etc.

robertknight: Done. Email addresses without TLDs are allowed by the pattern + spec I used, as noted in [this SO answer](https://stackoverflow.com/questions/20573488/why-does-html5-form-validation-allow-emails-without-a-dot).

tests/h/validators_test.py
```diff
@@ -13,13 +15,21 @@ class TestEmail(object):
...
         'jim.smith@foo.bar.com',
+        'a.range!of#punctuation-chars$are+accepted@gmail.com',
     ])
     def test_accepts_valid_addresses(self, schema_node, email):
         validator = Email()
         validator(schema_node, email)
 
     @pytest.mark.parametrize('email', [
         ' spaces @ spaces.com',
+        'sorry-ascii-only-🤠@gmail.com',
```
chdorner: Hmm, these are not unicode strings at the moment. Maybe they should be? :)

tests/h/validators_test.py
```diff
@@ -0,0 +1,43 @@
...
+        'jim.smith@foo.bar.com',
+        'a.range!of#punctuation-chars$are+accepted@gmail.com',
+    ])
+    def test_accepts_valid_addresses(self, schema_node, email):
+        validator = Email()
+        validator(schema_node, email)
+
+    @pytest.mark.parametrize('email', [
+        ' spaces @ spaces.com',
+        'sorry-ascii-only-🤠@gmail.com',
```
chdorner: Hmm, these are not unicode strings at the moment. Maybe they should be? By either adding the `unicode_literals` import, or specifically using `u''`.

robertknight: Good point. I've added the `unicode_literals` import since we use that widely elsewhere in the "h" code.

PR #4673, opened 2017-10-24
First pass schema-based comment import validator

fatbusinessman:
This is a part of https://github.com/hypothesis/product-backlog/issues/349.

This is a work in progress, hence not being wired into the rest of the app anywhere. I’d appreciate any early-stage feedback.

A few disorganised thoughts:

* I've tried to be pretty strict with what this validator will accept, purely from the point of view that we are only going to extend this as we want to accept more use cases and features in the future.
* I haven't put tagging into this yet, although that was in the original proposal. I haven't seen this in any of the data we want to accept so far, so we may be able to avoid doing this for now.
* I've added a context manager that wraps around `pytest.raises`; this also checks what the validation error message looks like. The reason I introduced this was that I named a key incorrectly in one of the tests and it still passed (because it failed as an unrecognised key). This seems a reasonable thing to add, but I'm aware that in a test suite it might make more sense to duplicate some code rather than introduce any new complexity.
* The new dependencies are libraries required by `jsonschema` to validate particular formats. `rfc3987` validates URIs and `strict-rfc3339` validates timestamps. There's an argument that this
  stricter validation isn't worth the cost of the additional dependencies, and that we might be ok with simpler regex-based validation.

### DIFF
--- a/h/cli/__init__.py
+++ b/h/cli/__init__.py
@@ -18,6 +18,7 @@ SUBCOMMANDS = (
     'h.cli.commands.celery.celery',
     'h.cli.commands.devserver.devserver',
     'h.cli.commands.groups.groups',
+    'h.cli.commands.import.import_annotations',
     'h.cli.commands.init.init',
     'h.cli.commands.initdb.initdb',
     'h.cli.commands.migrate.migrate',
--- a/h/cli/commands/import.py
+++ b/h/cli/commands/import.py
@@ -1,0 +1,23 @@
+# -*- coding: utf-8 -*-
+
+import click
+import json
+
+from h.schemas.annotation_import import AnnotationImportSchema
+
+# from h import models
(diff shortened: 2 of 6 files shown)


seanh:
> I've tried to be pretty strict with what this validator will accept, purely from the point of view that we are only going to extend this as we want to accept more use cases and features in the future.

This seems like a good idea to me.

30% feedback: I haven't refreshed my knowledge of JSONSchema or the W3C web annotation spec yet, but the general approach here looks fine to me. Let me know if you need someone to get up to date on JSON Schema and web annotation spec and give detailed feedback on the schema. :+1:

fatbusinessman:
While there is useful reference code in here, we’re planning to start out with an importer based on our public API, so this pull request isn’t going to be merged directly.

PR #4677, opened 2017-11-07
Remove client cookie auth support

robertknight:
**Depends on https://github.com/hypothesis/client/pull/542 being merged & shipped**

This PR removes support for cookie-based authentication for the client in the web service. This includes:

- The `GET /app` route which was a predecessor of the `/api/profile` route
- The `GET /api/token` route which was used by the client to obtain a JWT "access token"
- Support for JWT "access tokens" generated by the `GET /api/token` route
- AJAX views for handling login & logout requests from the client.
- The "client_id" and "client_secret" env vars which were used to sign JWTs returned by the `GET /api/token` route

I included all the changes in this PR (one per commit) so they can all be tested together more easily. I'm happy to split some of the pieces out if people would prefer.

There has been some chatter in #design about whether login synchronisation between the embed and the website is something we might want to reintroduce in future. If we do then I imagine we'll revisit in a way that fits in better with the OAuth-based auth system rather than being an entirely separate thing.

### DIFF
--- a/conf/app.ini
+++ b/conf/app.ini
@@ -14,12 +14,6 @@ use: call:h.app:create_app
 #es.host: http://localhost:9200
 #es.index: hypothesis
 
-# OAuth settings
-# These client credentials are used by the built-in Web client.
-# If not provided, both default to a random URL-safe base64-encoded string.
-#h.client_id:
-#h.client_secret:
-
 # SQLAlchemy configuration -- See SQLAlchemy documentation
 sqlalchemy.url: postgresql://postgres@localhost/postgres
 
--- a/tests/h/views/accounts_test.py
+++ b/tests/h/views/accounts_test.py
@@ -561,7 +561,7 @@ class TestActivateController(object):
         assert success_flash
         assert success_flash[0].startswith("Your account has been activated")
 
-    def test_get_when_not_logged_in_successful_creates_ActivationEvent(  # noqa
+    def test_get_when_not_logged_in_successful_creates_ActivationEvent(  # noqa: N802, N803
             self,
             pyramid_request,
             user_model,
--- a/tests/h/views/accounts_test.py
+++ b/tests/h/views/accounts_test.py
@@ -561,7 +561,7 @@ class TestActivateController(object):
         assert success_flash
         assert success_flash[0].startswith("Your account has been activated")
 
-    def test_get_when_not_logged_in_successful_creates_ActivationEvent(
+    def test_get_when_not_logged_in_successful_creates_ActivationEvent(  # noqa
             self,
             pyramid_request,
             user_model,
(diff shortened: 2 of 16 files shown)


seanh:
https://github.com/hypothesis/client/pull/542 is merged but not yet shipped. I'll review this now but not merge it yet

tests/h/views/accounts_test.py
```diff
@@ -574,7 +574,7 @@ def test_get_when_not_logged_in_successful_creates_ActivationEvent(
         ActivationEvent.assert_called_once_with(
             pyramid_request, user_model.get_by_activation.return_value)
 
-    def test_get_when_not_logged_in_successful_notifies(self,
+    def test_get_when_not_logged_in_successful_notifies(self,  # noqa: N803
```
seanh: I'm not sure what's being disabled here, is there a way to name the rule that's being disabled in the comment?

tests/h/views/accounts_test.py
```diff
@@ -561,7 +561,7 @@ def test_get_when_not_logged_in_successful_flashes_message(self,
         assert success_flash
         assert success_flash[0].startswith("Your account has been activated")
 
-    def test_get_when_not_logged_in_successful_creates_ActivationEvent(
+    def test_get_when_not_logged_in_successful_creates_ActivationEvent(  # noqa
```
seanh: Is it possible to make the disable more specific?

conf/app.ini
```diff
@@ -14,12 +14,6 @@ use: call:h.app:create_app
 #es.host: http://localhost:9200
 #es.index: hypothesis
 
-# OAuth settings
-# These client credentials are used by the built-in Web client.
-# If not provided, both default to a random URL-safe base64-encoded string.
-#h.client_id:
-#h.client_secret:
```
seanh: I think `h.client_secret` still appears in a couple of places:

```
$ ag 'h\.client_sec'
conf/development-websocket.ini
7:h.client_secret: nosuchsecret

tests/functional/conftest.py
18:    'h.client_secret': 'notsosecret',

tests/h/views/api_auth_test.py
370:        pyramid_config.registry.settings['h.client_secret'] = 'notsosecretafterall'
```

seanh:
All seems to function correctly, as far as I can tell

seanh:
A test was failing locally due to the pyjwt upgrade that no longer validates `iat` (the newer version of pyjwt having been installed in my dev env when I was on `master`). Rebasing fixes it.

tests/h/views/accounts_test.py
robertknight: Not as far as I know 😞 - ESLint has human readable slug-like names for its rules which can be used in linter directives. flake8 lacks this.

robertknight:
> One final question before approving this: is there a reason the lint fixes have to be part of the same pull request? 

I'm just being lazy/pragmatic. There is a little bit of overhead in creating a separate PR and getting someone else to review it compared to just putting small fixes like this in the context of the larger PR. While it is generally a good idea not to mix say refactoring & feature changes in the same PR, I personally prefer not to be absolute about it.

fatbusinessman:
> There is a little bit of overhead in creating a separate PR and getting someone else to review it compared to just putting small fixes like this in the context of the larger PR.

That’s fair, although there’s also overhead (at least for me – I can’t speak for anyone else) in switching between the different contexts while reviewing.

You’re right in the case of this pull request that it doesn’t make a great deal of difference either way, so I’m happy to approve this.

PR #4678, opened 2017-11-07
Don't test that JWT's issued in future are invalid

seanh:
Remove two tests that JWT's whose `iat` value claims that they were
issued in the future fail validation.

These two tests fail on newer versions of PyJWT:

https://github.com/hypothesis/h/pull/4672

This is because PyJWT no longer raises an exception for future `iat`
times:

https://github.com/jpadilla/pyjwt/issues/190

PyJWT removed this validation because:

- Clock skew can cause one party to generate `iat` times a few seconds
or minutes ahead of another's current time

- The JWT spec (RFC 7519) doesn't say that a JWT with a future `iat`
should be considered invalid, these JWTs are valid

- Other JWT libraries don't do this check

### DIFF
--- a/tests/h/auth/tokens_test.py
+++ b/tests/h/auth/tokens_test.py
@@ -48,11 +48,6 @@ INVALID_TOKEN_EXAMPLES = [
     lambda k: jwt.encode({'exp': _seconds_from_now(-3600)},
                          key=k),
 
-    # Issued in the future
-    lambda k: jwt.encode({'exp': _seconds_from_now(3600),
-                          'iat': _seconds_from_now(1800)},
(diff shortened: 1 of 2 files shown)


fatbusinessman:
> I'll leave this until @fatbusinessman is back tomorrow in case he has any objections.

Yup, this seems reasonable. I had a concern about whether there was some way to get long-lived tokens past our checks with this, but those are based on the `nbf` and `exp` claims, rather than `iat`, so I think we’re still good here.

gobengo:
wow that pyjwt lib issue is from me like two years ago. small world

seanh:
:)

PR #4681, opened 2017-11-09
Fix documented return type for `create_annotation`

fatbusinessman:


seanh:
:+1: Thanks for the tiny PR, very easy to review!

PR #4682, opened 2017-11-10
Reorganize API documentation

robertknight:
Create a new top-level section in the documentation for information
about using the API. This will provide a place to put a general
introduction to the API and an Authorization section which contains
information relevant to both the REST API and the real time API.

* To avoid a naming conflict in the output dir, the `_extra/api` dir was
  renamed to `api-reference`.

* To enable creating an internal _relative_ link to the ReDoc API docs,
  create a stub "api-reference/index.rst" file. This is then replaced with
  the contents of "_extra/api-reference" during compilation.
  This works around a [Sphinx issue](https://github.com/sphinx-doc/sphinx/issues/701).

* Reword the title of the real time API docs to avoid repeating
  "Hypothesis" so often in the table of contents

See Slack discussion: https://hypothes-is.slack.com/archives/C4K6M7P5E/p1510318582000087


docs/api/index.rst
```diff
@@ -0,0 +1,11 @@
+Using the Hypothesis API
+########################
```
fatbusinessman: A bit of a pedantic point here, but is there a reason this heading marker uses `#` while the other one uses `=`? I know rST doesn’t care which symbol you use, and that we aren’t consistent across the docs, but I think we should pick one and stick with it for future documentation changes, unless there’s a distinction I‘ve missed.

PR #4683, opened 2017-11-13
Describe OAuth in authorization docs

robertknight:
Update authorization docs to describe how to create OAuth clients for the h service and implement the OAuth authorization flow.

Explaining all the possibilities of OAuth in detail is IMO beyond the scope of the documentation here, as is a lengthy explanation of the problem that OAuth solves. I've focused on providing enough information to implement a browser or web-service based client for the APIs plus links to relevant parts of the specs.

In this PR I'm only describing the flow we implement that is part of the main OAuth 2 standard (using a redirect rather than a `window.postMessage` response) to keep the size of this change down.

### DIFF
--- a/docs/_extra/api-reference/hypothesis.yaml
+++ b/docs/_extra/api-reference/hypothesis.yaml
@@ -1,63 +1,13 @@
 swagger: "2.0"
 info:
   version: 1.0.0
-  title: Hypothesis
+  title: Hypothesis API Reference
   description: |
-    # Hypothesis API documentation
+    # Hypothesis API Reference
 
-    This document details the Hypothesis public HTTP API. It is targeted at
-    developers interested in integrating functionality from Hypothesis into
-    their own applications.
-
-    ## Authorization
-
-    ### API tokens
-    Some of the API URLs documented below require a valid API token.
-    To use these API URLs you should:
-
-    1. Generate yourself an API token on your [Hypothesis developer
-       page](https://hypothes.is/account/developer) (you must be logged in to
-       Hypothesis to get to this page).
-    2. Put the API token in the `Authorization` header in your request.
-
-    *Example request:*
-
-        GET /api
-        Host: hypothes.is
-        Accept: application/json
-        Authorization: Bearer 6879-nUYZBzc15MsjDDicz-0xEaKBrVGGruwCHW9-p8BMqq4
-
-    (Replace `6879-nUYZBzc15MsjDDicz-0xEaKBrVGGruwCHW9-p8BMqq4` with your own API token.)
-
-    ### Client credentials direct authorization
-    The user creation API is intended for use by "third-party accounts" clients.
-    These endpoints are authenticated by HTTP Basic Auth using your client ID as
-    the username and your client secret as the password.
-
-    For example, with client details as follows
-
-        Client ID: 96653f8e-80be-11e6-b32b-c7bcde86613a
-        Client Secret: E-hReVMuRyZbyr1GikieEw4JslaM6sDpb18_9V59PFw
-
-    you can compute the Authorization header [as described in
-    RFC7617](https://tools.ietf.org/html/rfc7617):
-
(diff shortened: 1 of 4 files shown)


docs/_extra/api-reference/hypothesis.yaml
```diff
@@ -1,63 +1,13 @@
 swagger: "2.0"
 info:
   version: 1.0.0
-  title: Hypothesis
+  title: Hypothesis API Reference
   description: |
-    # Hypothesis API documentation
+    # Hypothesis API Reference
 
-    This document details the Hypothesis public HTTP API. It is targeted at
```
robertknight: Here I'm moving the main authorization documentation out of ReDoc (now under "API Reference" in the main table of contents) and into a separate "Authorization" page authored in reST.

docs/api/authorization.rst
```diff
@@ -0,0 +1,90 @@
+Authorization
+=============
```
robertknight: This is a new top-level page which comes after the introduction to the API, explaining when authorization is required and how it works before the reader gets to the API reference.

docs/api/authorization.rst
```diff
@@ -0,0 +1,90 @@
...
+   these tokens are only suitable for enabling your application to make
+   requests as a single specific user.
+
+2. By :ref:`registering <registering-an-oauth-client>` an "OAuth client" and
+   :ref:`implementing <implementing-oauth-flow>` the OAuth authentication flow
+   in your application. This method allows any user to authorize your
+   application to read and write data via the API as that user.  The Hypothesis
+   client is an example of an application that uses OAuth.
+
+   See :doc:`using-oauth` for details of how to implement this method.
```
robertknight: I put the details of using OAuth in a separate doc because I think it likely that a first time reader will decide that the API token is a simpler place to get started for testing purposes.

docs/api/authorization.rst
```diff
@@ -0,0 +1,90 @@
...
+    Host: hypothes.is
+    Accept: application/json
+    Authorization: Bearer $TOKEN
+
+(Replace ``$TOKEN`` with your own API token or OAuth access token.)
+
+Client credentials
+------------------
+
+Endpoints for managing user accounts are authorized using a client ID and secret
```
robertknight: This is reworded from the previous documentation in `hypothesis.yaml` because we now support APIs other than just creating users which are authorized with client credentials. eg. The API for updating user accounts.

docs/api/using-oauth.rst
```diff
@@ -0,0 +1,164 @@
...
+   asked to approve access.
+
+3. If they approve, the authorization endpoint will send an authorization code
+   back to the application (via a redirect).
+
+4. The application exchanges the authorization code for a pair of tokens using
+   the h service's ``POST /api/token`` endpoint: A short-lived access token to
+   authorize API requests, and a long-lived refresh token.
+
+5. When the access token expires, the application obtains a new access token
```
fatbusinessman: The application obtains the new access token using the refresh token, right? That’s not entirely clear to me from this documentation.

docs/api/using-oauth.rst
```diff
@@ -0,0 +1,164 @@
...
+2. Implement the client-side part of the OAuth flow above in your application.
+   You may be able to use an existing OAuth 2 client library for your language
+   and platform.
+
+.. _registering-an-oauth-client:
+
+Registering an OAuth client
+---------------------------
+
+To register an OAuth client, go to the `/admin/oauthclients` page on the h
```
fatbusinessman: Would it be worth being super explicit about the fact that this is how to register an OAuth client on **your own server**?

docs/api/using-oauth.rst
robertknight: Yes indeed. I'll clarify this.

docs/api/using-oauth.rst
robertknight: Yes, will do.

PR #4687, opened 2017-11-15
Remove the h changelog

fatbusinessman:
This file is no longer meaningful, now that we've pulled out the client into a separate project and don't have a concept of distinct versioned releases of the service.

As far as I can tell, this also means we no longer need the `list-changes-since` script, so I’ve removed that too. If anyone knows of a reason why this script still belongs here, let me know and I’ll take that commit out.


fatbusinessman:
> The list-changes-since script was added as a way for me to be lazy when updating the CHANGES file.

That’s what I figured. 👍 for productive laziness.

PR #4690, opened 2017-11-20
Clarify that create_annotation() takes valid data

seanh:
Clarify, in its docstring, that `create_annotation()` takes _already valid_ data as param.

Also change some references to other Python classes in the docstring into `:py:class:` links. This styles the links correctly and generates warnings if the links break (if we ever export these docstrings into
docs using Sphinx autodoc).

### DIFF
--- a/h/storage.py
+++ b/h/storage.py
@@ -85,19 +85,21 @@ def fetch_ordered_annotations(session, ids, query_processor=None):
 
 def create_annotation(request, data, group_service):
     """
-    Create an annotation from passed data.
+    Create an annotation from already-validated data.
 
     :param request: the request object
     :type request: pyramid.request.Request
 
-    :param data: a dictionary of annotation properties
+    :param data: an annotation data dict that has already been validated by
+        :py:class:`h.schemas.annotation.CreateAnnotationSchema`
     :type data: dict
 
-    :param group_service: a service object that adheres to ``h.interfaces.IGroupService``
-    :type group_service: h.interfaces.IGroupService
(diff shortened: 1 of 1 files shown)


robertknight:
I'll leave it up to @fatbusinessman to review this since he understands the code better, but in general 👍 for making type requirements clearer.

I don't know whether Sphinx autodoc is something we'll make much use of but making the code more accessible to tooling is a good thing IMO. I personally think [mypy](http://mypy-lang.org/index.html) would be more useful in future to facilitate safer & easier refactoring.

seanh:
Existing practice is that we do use Sphinx in our Python docstrings, for example:

```
> ag :py: | wc -l
41
```

[Sphinx info field lists](http://www.sphinx-doc.org/en/stable/domains.html?highlight=%3Aparam%3A#info-field-lists) (`:param:`, `:type:`, `:raises:`, `:returns:`, `:rtype:`) are also common in the h code:

```bash
> ag :param | wc -l
203
```

And other Sphinx syntax for things like code spans, italic and bold, etc.

But currently these docstrings never actually get processed by Sphinx, turning on autodoc (which is the thing that pulls docstrings out of Python files and processes them with Sphinx) would do that.

robertknight:
> Existing practice is that we do use Sphinx in our Python docstrings

To be clear, I wasn't advocating against doing that - consistency and just having the info there in _some_ form is a good thing. I was just saying that there are more useful applications for accurate type information that auto-generated documentation.

PR #4698, opened 2017-11-23
Add eLife annotation timestamps data migration

seanh:
We're going to run an API script to import 1000+ Disqus comments from eLife into page note annotations using the standard annotation create API. After being created in this way the annotations will have the wrong timestamps - they'll be the times at which the annotations were created in h, not the times of the comments exported from Disqus. This migration sets the timestamps of these imported annotations to their correct Disqus-exported timestamps.

To test, checkout this branch and then:

1. Create a user account in your h dev instance, and get a developer API token for this account

2. Create a group in your h dev instance

3. Set the `IMPORT_GROUP_ID` env var to the `groupid` of the group you just created (the `import.py` script below uses this): `export IMPORT_GROUP_ID=<YOUR_GROUP_ID>`

4. You need to have annotations in your DB with `imported_id` extra fields matching those from eLife (in the `TIMESTAMPS` list in the migration script). Here's a [script that will add test annotations with these IDs to your dev instance of h](https://gist.github.com/seanh/09112f2e39a3381fd6a156750e91c789). To run it: `./import.py <YOUR_H_DEVELOPER_API_TOKEN>`.

   At this point the `created` and `updated` times of the imported annotations in your DB will be recent. You can inspect them with: `psql postgresql://postgres@localhost/postgres -c 'SELECT created, updated FROM annotation;'`.

5. Edit the migration script, changing the `GROUPID` constant to be the `groupid` of the group you just created

6. Stamp your db: `./bin/hypothesis migrate stamp 9bcc39244e82`

7. Run the migration script: `./bin/hypothesis migrate upgrade +1`

You should see:

```
[INFO] Running upgrade 9bcc39244e82 -> b32d09564cf5, Correct the timestamps of imported eLife annotations.
[INFO] Updated the timestamps of 1258 annotations
```

And if you inspect your database you should see that the annotation's timestamps have been changed: `psql postgresql://postgres@localhost/postgres -c 'SELECT created, updated FROM annotation;'`.

You can also run the migration script _without_ having imported the annotations into your DB. In that case it should not crash and you should see `Updated the timestamps of 0 annotations`.

If you have a production database dump loaded than you can use the developer API token of the actual disqus import user, and the groupid of the actual eLife group, and you don't need to edit the migration script to test it.

### DIFF
--- a/h/migrations/versions/b32d09564cf5_correct_timestamps_of_elife_annotations.py
+++ b/h/migrations/versions/b32d09564cf5_correct_timestamps_of_elife_annotations.py
@@ -1,0 +1,6394 @@
+"""
+Correct the timestamps of imported eLife annotations.
+
+Set the timestamps of eLife annotations that were imported using an API script
+to their correct timestamps according to the data file that eLife gave us.
+
+Revision ID: b32d09564cf5
+Revises: 9bcc39244e82
+Create Date: 2017-11-23 15:56:54.292312
+"""
+
+from __future__ import unicode_literals
+
+from datetime import datetime
+import logging
+
+from alembic import op
+import sqlalchemy as sa
+from sqlalchemy.dialects import postgresql as pg
+from sqlalchemy.ext.declarative import declarative_base
+from sqlalchemy.ext.mutable import MutableDict
+from sqlalchemy.orm import sessionmaker
+
+from h.db import types
+
+
+revision = 'b32d09564cf5'
+down_revision = '9bcc39244e82'
+
+
+log = logging.getLogger(__name__)
+
+
+Base = declarative_base()
+Session = sessionmaker()
+
+
+# The string format of the timestamps below.
+FORMAT = "%Y-%m-%dT%H:%M:%SZ"
+
+
+# The groupid of the public group, in the production DB, that the annotations
+# to be modified belong to.
+GROUPID = "ijDGibQE"
+
+
+# The list of annotations we're going to modify (identified by their
+# imported_id values in the DB) and the timestamp values we're going to set for
+# them.
+TIMESTAMPS = [
+    {
+        "imported_id": "disqus-import:2532559447",
+        "created": "2012-06-20T23:14:23Z",
+        "modified": "2012-06-20T23:14:23Z"
+    },
+    {
+        "imported_id": "disqus-import:2532559448",
+        "created": "2012-06-21T22:31:27Z",
+        "modified": "2012-06-21T22:31:27Z"
+    },
+    {
+        "imported_id": "disqus-import:2532559446",
+        "created": "2012-06-21T23:25:28Z",
+        "modified": "2012-06-21T23:25:28Z"
+    },
+    {
+        "imported_id": "disqus-import:2532559449",
+        "created": "2012-06-25T00:28:28Z",
+        "modified": "2012-06-25T00:28:28Z"
+    },
+    {
+        "imported_id": "disqus-import:2532559450",
+        "created": "2012-06-26T05:52:33Z",
+        "modified": "2012-06-26T05:52:33Z"
+    },
+    {
+        "imported_id": "disqus-import:2532559453",
+        "created": "2012-06-29T06:32:56Z",
+        "modified": "2012-06-29T06:32:56Z"
+    },
+    {
+        "imported_id": "disqus-import:2532559455",
+        "created": "2012-07-06T23:36:46Z",
+        "modified": "2012-07-06T23:36:46Z"
+    },
+    {
+        "imported_id": "disqus-import:2532559452",
+        "created": "2012-07-10T22:14:51Z",
+        "modified": "2012-07-10T22:14:51Z"
+    },
+    {
+        "imported_id": "disqus-import:2532559454",
+        "created": "2012-07-11T03:07:01Z",
+        "modified": "2012-07-11T03:07:01Z"
+    },
+    {
+        "imported_id": "disqus-import:2532559451",
+        "created": "2012-07-12T00:38:47Z",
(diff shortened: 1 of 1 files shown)


seanh:
There's probably no point in merging this until the importer script is finished. Then I can run the importer locally against production data, and then test the migration against that.

h/migrations/versions/b32d09564cf5_correct_timestamps_of_elife_annotations.py
```diff
@@ -0,0 +1,6394 @@
...
+Session = sessionmaker()
+
+
+# The string format of the timestamps below.
+FORMAT = "%Y-%m-%dT%H:%M:%SZ"
+
+
+# The groupid of the public group, in the production DB, that the annotations
+# to be modified belong to.
+GROUPID = "ijDGibQE"
```
seanh: I've hardcoded the ID of eLife's test public group

seanh: In the future, we want to re-run this migration for annotations that eLife have imported into their real public group, we can copy-paste this migration script and edit the value of this `GROUPID` constant (and the alembic revision ids)

h/migrations/versions/b32d09564cf5_correct_timestamps_of_elife_annotations.py
```diff
@@ -0,0 +1,6394 @@
...
+
+
+# The list of annotations we're going to modify (identified by their
+# imported_id values in the DB) and the timestamp values we're going to set for
+# them.
+TIMESTAMPS = [
```
fatbusinessman: I think this is fine to have hard-coded in the migration (as I think we discussed before), but would it be possible to have it defined *after* the migration code? That would mean not having to scroll through a few thousand lines of dictionary to see the code itself.

giorgiosironi:
Nginx had this problem on startup (from `/var/log/nginx/error.log`):
```
2017/11/24 11:04:29 [emerg] 109#0: mkdir() "/var/lib/nginx/tmp/client_body" failed (13: Permission denied)
```
Similar problems in creating folders there appeared until I added all the configuration changes moving its temporary folders to `/tmp`.

The worker process exited with:
```
worker (stderr)      | Error: the app URL must be set in production mode!
```
Looking into the code, it seemed to require a `--app-url` argument to be passed. Nginx listens on port 5000 so I guessed at the value.

seanh:
:)

h/migrations/versions/b32d09564cf5_correct_timestamps_of_elife_annotations.py
```diff
@@ -0,0 +1,6394 @@
...
+    },
+    {
+        "imported_id": "disqus-import:2532560173",
+        "created": "2014-05-09T20:37:52Z",
+        "modified": "2014-05-09T20:37:52Z"
+    },
+    {
+        "imported_id": "disqus-import:2532560183",
+        "created": "2014-05-09T20:38:31Z",
+
```
fatbusinessman: What does the performance look like on this query? It looks to me like it’s going to use the index on `groupid`; is that going to be good enough, or do we need an even more specific index?

fatbusinessman: Could we take this context about the total number of timestamps and put it into this log message? We have access to `len(TIMESTAMPS)`, after all.

fatbusinessman: Would it be worth pre-parsing these timestamps, partly to save ourselves work during the database transaction and partly to guard ourselves against the possibility of malformed data? Or is that overkill for our purposes?

seanh:
Marked as WIP for now - We thought that _we_ were going to import the annotations, after which we would then merge and run this DB migration to fix their timestamps. As it turns out, eLife are running David's script to import the annotations themselves. So at some point, when eLife have finished importing the annotations, we need to run our DB migration to fix the timestamps. We'll need from eLife a final list of the Disqus IDs and their timestamps, as well as assurance that they're done and aren't going to delete and re-import them, then we can update, merge and run this migration.

h/migrations/versions/b32d09564cf5_correct_timestamps_of_elife_annotations.py
seanh: Sure, that'll be simple

seanh: Does it matter how long we spend in the transaction? I thought only the commit mattered. And I know none of them are malformed anyway because I've tested. But I do think it's a better idea to parse them all first, though

seanh: I've just assumed that because it's < 1300 annotations we're fine. It certainly runs very quickly in dev with the 1300 annotations loaded (but without the rest of production's data loaded). Do you think we need to measure it more carefully?

h/migrations/versions/b32d09564cf5_correct_timestamps_of_elife_annotations.py
seanh: Yeah, I can just move it, agree that's nicer

seanh:
Since I don't know when eLife are going to be ready (finished importing) for us to run this migration, and don't know whether there'll be more data from them to add to it, I'm going to hold off on implementing David's suggested changes until we get the go ahead

seanh:
We need to re-write this to search by Hypothesis annotation IDs, not Disqus imported IDs. I'm gonna close this PR to preserve the `imported_id` version for posterity, as filtering by an `extra` was non-obvious. Will submit a new id-based one

PR #4699, opened 2017-11-24
Containers integration

giorgiosironi:
`nginx` and `worker` were crashing and being continuously restarted.

Would have used `v0.38.0` rather than `master`, but it has no `docker-compose.yml` for the dependencies.

My goal is to have a container (plus its dependent containers) to run in testing environments, so that we can run automated end2end tests against a recent version of Hypothes.is and clean up its database nightly.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -20,6 +20,9 @@ WORKDIR /var/lib/hypothesis
 
 # Ensure nginx state and log directories writeable by unprivileged user.
 RUN chown -R hypothesis:hypothesis /var/log/nginx /var/lib/nginx
+# /var/lib/nginx/tmp seems to be 700, so still not writable?
+# I thought nginx would run as the hypothesis user though
+# changing its permissions to 777 doesn't seem to have an effect
 
 # Copy minimal data to allow installation of dependencies.
 COPY requirements.txt ./
--- a/conf/nginx.conf
+++ b/conf/nginx.conf
@@ -9,6 +9,11 @@ events {
 
 http {
   client_max_body_size 20m;
+  client_body_temp_path /tmp 1 2;
+  proxy_temp_path /tmp 1 2;
+  fastcgi_temp_path /tmp 1 2;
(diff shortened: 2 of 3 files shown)


robertknight:
Hello @giorgiosironi. Thank-you for your contribution. We're in the middle of finishing up a partner project at the moment so it may be a few days before we can look at this.

> nginx and worker were crashing and being continuously restarted.

Can you explain in more detail what your setup is, what the problem was and how you have fixed it here? This will make it easier to review and integrate the changes.

giorgiosironi:
Nginx had this problem on startup (from `/var/log/nginx/error.log`):
```
2017/11/24 11:04:29 [emerg] 109#0: mkdir() "/var/lib/nginx/tmp/client_body" failed (13: Permission denied)
```
Similar problems in creating folders there appeared until I added all the configuration changes moving its temporary folders to `/tmp`. It seems that the folder itself should be writable by the `hypothesis` user, but nginx sometimes daemonizes as the `nobody` users for security reasons. Even giving very wide permissions to `/var/lib/nginx/tmp` didn't allow writes.

The worker process exited with:
```
worker (stderr)      | Error: the app URL must be set in production mode!
```
Looking into the code, it seemed to require a `--app-url` argument to be passed. Nginx listens on port 5000 so I guessed at the value.

robertknight:
@seanh just pointed out to be that you are, of course, working for eLife on the "partner project" I mentioned. I'd failed to notice that 🙂. We'll take a look at the start of next week. We're currently focused on getting the Disqus import sorted.

giorgiosironi:
No problem, this is more long-term, for us to get a running copy of Hypothes.is for automated testing.

seanh:
Hi Giorgio,

Unfortunately I don't think the Hypothesis team has time to test, review and merge this right now. We have other projects that we really need to be moving on with. I'm going to close this pull request for now, it'll still be here in future if we run into and need to fix this problem again.

PR #4700, opened 2017-11-24
Index authority of annotation creator as a separate field in ES

robertknight:
Index the authority part of the `userid` field in an Annotation (which has the form `acct:user@authority`) as a separate field in ES so that the search API and activity pages can efficiently query for annotations associated with a specific authority.

In order to build an actually useful test for this, I introduced a unit test which actually uses ES rather than mocking it.

The next step after this will be to filter out annotations from authorities other than the default one on activity pages.

~~**WIP while I get the new `h.search.index` tests working in Jenkins**~~ _Fixed: Tox needed to be configured to pass the ELASTICSEARCH_HOST env var through to the tests_.

### DIFF
--- a/h/presenters/annotation_searchindex.py
+++ b/h/presenters/annotation_searchindex.py
@@ -16,6 +16,7 @@ class AnnotationSearchIndexPresenter(AnnotationBasePresenter):
         docpresenter = DocumentSearchIndexPresenter(self.annotation.document)
 
         result = {
+            'authority': self.annotation.userid,
             'id': self.annotation.id,
             'created': self.created,
             'updated': self.updated,
--- a/h/search/index.py
+++ b/h/search/index.py
@@ -26,13 +26,17 @@ class Window(namedtuple('Window', ['start', 'end'])):
     pass
 
 
-def index(es, annotation, request, target_index=None):
+def index(es, annotation, request, target_index=None, refresh=False):
     """
     Index an annotation into the search index.
 
     A new annotation document will be created in the search index or,
     if the index already contains an annotation document with the same ID as
     the given annotation then it will be updated.
+
+    If the `refresh` option is `True` the change will be immediately visible in
+    searches, otherwise there may be a short delay (eg. of a second or two)
+    before it becomes visible.
 
     :param es: the Elasticsearch client object to use
     :type es: h.search.Client
@@ -42,6 +46,9 @@ def index(es, annotation, request, target_index=None):
 
     :param target_index: the index name, uses default index if not given
     :type target_index: unicode
+
+    :param refresh: Whether to make this change immediately visible to searches
+    :type refresh: bool
     """
     presenter = presenters.AnnotationSearchIndexPresenter(annotation)
     annotation_dict = presenter.asdict()
@@ -57,6 +64,7 @@ def index(es, annotation, request, target_index=None):
         doc_type=es.t.annotation,
         body=annotation_dict,
         id=annotation_dict["id"],
+        refresh=refresh,
     )
 
 
--- a/tests/h/search/index_test.py
+++ b/tests/h/search/index_test.py
@@ -1,10 +1,10 @@
 # -*- coding: utf-8 -*-
 from __future__ import unicode_literals
 
+from datetime import datetime
+
 import mock
 import pytest
-
-import elasticsearch
 
 from h import presenters
 from h.search import client
@@ -13,6 +13,9 @@ from h.search import index
 
 @pytest.mark.usefixtures('presenters')
 class TestIndexAnnotation:
+    """
+    Annotation indexing tests that use a mock ElasticSearch connection.
+    """
 
     def test_it_presents_the_annotation(self, es, presenters, pyramid_request):
         annotation = mock.Mock()
@@ -52,6 +55,7 @@ class TestIndexAnnotation:
             doc_type='annotation',
             body=presenters.AnnotationSearchIndexPresenter.return_value.asdict.return_value,
             id='test_annotation_id',
+            refresh=False,
         )
 
     def test_it_allows_to_override_target_index(self, es, presenters, pyramid_request):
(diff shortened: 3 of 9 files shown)


h/presenters/annotation_searchindex.py
```diff
@@ -16,6 +16,7 @@ def asdict(self):
         docpresenter = DocumentSearchIndexPresenter(self.annotation.document)
 
         result = {
+            'authority': self.annotation.userid,
```
fatbusinessman: If we split this apart in the Python code (for which we have the [`split_user` function](https://github.com/hypothesis/h/blob/156d9aa10c3f12ef94afb21d6354817298dbbce1/h/util/user.py#L6-L21)), we could remove the need for extra complexity in the Elasticsearch configuration. Would that work?

h/search/index.py
```diff
@@ -26,14 +26,18 @@ class Window(namedtuple('Window', ['start', 'end'])):
     pass
 
 
-def index(es, annotation, request, target_index=None):
+def index(es, annotation, request, target_index=None, refresh=False):
```
fatbusinessman: In my previous experience with Elasticsearch, while the `refresh` parameter triggered a refresh immediately after the operation, it could still result in a race condition because the index operation would return before the refresh completed. As far as I can see from [our version’s docs](https://www.elastic.co/guide/en/elasticsearch/reference/1.6/docs-index_.html#index-refresh), this sounds like it could still be the case.

Do you know whether this has changed?

tests/h/search/index_test.py
```diff
@@ -1,18 +1,21 @@
 # -*- coding: utf-8 -*-
 from __future__ import unicode_literals
 
+from datetime import datetime
+
 import mock
 import pytest
 
-import elasticsearch
```
fatbusinessman: It looks like this is a requirement that wasn’t needed, and this is just a linting change. Is that the case? If so, I think it should be in a separate commit to avoid confusion.

tests/h/search/index_test.py
```diff
@@ -84,7 +122,7 @@ def test_it_marks_annotation_as_deleted(self, es):
             index='hypothesis',
             doc_type='annotation',
             body={'deleted': True},
-            id='test_annotation_id'
+            id='test_annotation_id',
```
fatbusinessman: Same point as above about style-only changes mixed in with functional changes. In this case, I’d even say this change isn’t worth the extra work it would cause for someone looking through the file with `git annotate`.

robertknight:
> I think that this could be done more simply by keeping it all in the Python world.

I did consider this initially, but realised there are a number of problems with doing the filtering in Python rather than ES:

1. It involves extra work on the part of the app because we'll be loading a bunch of annotations from the DB that then have to be filtered out.
2. Filtering annotations in Python rather than ES means there is a mismatch between the number of search results as reported by ES and the actual number. This means that page counts are misleading. More problematically, it means that page offsets and sizes won't work predictably. For example:
  - Supposing we have a page size of 200 annotations, and the most recently created annotations are 400 imported by a publisher in bulk. This would mean that the first two pages of activity pages would display zero annotations instead of 200 each because only when fetching the third page from ES will we get annotations that shouldn't be filtered out.
3. Different parts of the organization have mentioned that they think people using third party accounts may want some version of activity pages in future. While we're not building that now, I think it is worth keeping in mind. Being able to easily add authority-based filtering to any part of the app that relies on search will make this easier to implement.

> This would also keep our logic about how to extract an authority from a user ID in a single place (or at least stop it spreading into more places).

I personally don't see this as a significant issue, especially since we're already extracting the username  during indexing from ES for the same reason that the authority is being extracted here.

> …we’d decouple the need for this feature from the need for more comprehensive Elasticsearch tests, which would mean we could split this into two smaller, easier-to-review pull requests.

I'd be OK with moving the code that adds the ES tests to Travis/Jenkins into a separate PR but I really do think it is very useful to have and would rather do that first if so.

h/search/index.py
robertknight: Thanks for flagging this up. The docs [appear to make it clear](https://hyp.is/h3I3HtQ-Eee4dk8pq0KCow/www.elastic.co/guide/en/elasticsearch/reference/current/docs-refresh.html) that the refresh should be synchronous in ES 6. Needs some investigation to understand whether this is (a) actually the case and (b) if so, does it apply to the version of ES we are using (1.6)?

h/presenters/annotation_searchindex.py
robertknight: For reasons outlined in [this comment](https://github.com/hypothesis/h/pull/4700#issuecomment-347519407) I do think we need to do the filtering in ES rather than mixing it between ES and Python.

fatbusinessman:
> I did consider this initially, but realised there are a number of problems with doing the filtering in Python rather than ES

That’s not what I was referring to. Sorry, I didn’t make my point clearly enough.

I agree that we should have an `authority` field in the search index, and that we should use that as we search. Where I think we can remove a bunch of complexity is by extracting the correct authority in the Python code *at indexing time*, most likely using the `split_user` function I mentioned above. That means we wouldn’t have to introduce the additional complexity of a new `pattern_capture` analyzer in the index configuration.

My feeling is that we then wouldn’t be introducing enough additional complexity that we’d need to get the new Elasticsearch-integrated tests merged before this change, freeing us up to get the eLife-blocking changes shipped more quickly, but I realise that’s a judgement call, rather than any kind of hard fact.

robertknight:
> Where I think we can remove a bunch of complexity is by extracting the correct authority in the Python code at indexing time, most likely using the split_user function I mentioned above.

Ah right, sorry. I get your point. I agree this makes sense. If we do that, we should for consistency also do the same with the username field which, right now, is also extracted using a pattern capture.

robertknight:
Closing this for the moment. As discussed in Slack yesterday, I'm going to open a new PR with a narrower scope.

PR #4702, opened 2017-11-24
Ignore the debug mailer directory

fatbusinessman:
Pyramid's debug mailer (which we stared using in 110d39f2dbac53d7) saves test emails to the `mail/` directory. For the sake of avoiding accidental commits and annoying status messages, we should ignore this directory.


seanh:
Thanks, this has been annoying me too

PR #4703, opened 2017-11-27
Add a command-line tool to convert annotation IDs

fatbusinessman:
There are a few occasions where we need to convert between the encoded URL-safe versions of our annotation IDs (e.g. '3jgSANNlEeebpLMf36MACw') and their UUID versions (e.g. 'de381200-d365-11e7-9ba4-b31fdfa3000b'). This tool lets us do that without having to mess around importing bits of Python.

This tool currently relies on private functionality within the `h.db.types` module. While this isn't great, it seems better to have it this way and keep the functions private to discourage their use
elsewhere, than to remove the underscores and risk having them used more widely across the codebase. To mitigate against the risk that someone will change these functions (fairly assuming that they're not used outside the module) and break this functionality, I've used the real implementations of `_get_urlsafe_from_hex` and `_get_hex_from_urlsafe` in the tests, rather than mocking them out.

### DIFF
--- a/h/cli/__init__.py
+++ b/h/cli/__init__.py
@@ -14,6 +14,7 @@ from h import __version__
 log = logging.getLogger('h')
 
 SUBCOMMANDS = (
+    'h.cli.commands.annotation_id.annotation_id',
     'h.cli.commands.authclient.authclient',
     'h.cli.commands.celery.celery',
     'h.cli.commands.devserver.devserver',
--- a/h/cli/commands/annotation_id.py
+++ b/h/cli/commands/annotation_id.py
@@ -1,0 +1,26 @@
+# -*- coding: utf-8 -*-
+
+import click
+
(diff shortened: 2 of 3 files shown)


robertknight:
> This tool currently relies on private functionality within the h.db.types module. While this isn't great, it seems better to have it this way and keep the functions private to discourage their use
elsewhere

Given that the UUIDs appear in the DB (and therefore metabase) and the encoded UUIDs are surfaced in the public API, I'm not sure it makes sense to consider the conversion an implementation detail any more.

robertknight:
> To mitigate against the risk that someone will change these functions (fairly assuming that they're not used outside the module) and break this functionality, I've used the real implementations of _get_urlsafe_from_hex and _get_hex_from_urlsafe in the tests, rather than mocking them out.

This seems sensible to me.

PR #4706, opened 2017-11-28
Enable deletion of group creators from the admin panel

robertknight:
**~~Depends on https://github.com/hypothesis/h/pull/4724~~**

The admin control panel currently does not support deleting users who have created groups, even if those groups are just empty "test" groups. This requires manual futzing around with the database to fix,

This PR enables such users to be deleted from the admin control panel, _provided that no other users have created annotations in groups they created (Edited)_.

### DIFF
--- a/h/services/delete_user.py
+++ b/h/services/delete_user.py
@@ -3,7 +3,7 @@
 from __future__ import unicode_literals
 
 from h.events import AnnotationEvent
-from h.models import Annotation, Group
+from h.models import Annotation
 from h import storage
 
 
@@ -23,10 +23,22 @@ class DeleteUserService(object):
         message.
         """
 
-        if Group.created_by(self.request.db, user).count() > 0:
-            raise UserDeleteError('Cannot delete user who is a group creator.')
+        # Check whether this user has created any groups that others have
+        # annotated in.
+        #
+        # We check for non-empty `group_ids` before querying the DB to avoid an
+        # expensive SQL query if `in_` is given an empty list (see
+        # https://stackoverflow.com/questions/23523147/)
+        group_ids = [g.pubid for g in user.groups]
+        if len(group_ids) > 0:
+            other_user_group_anns = self.request.db.query(Annotation) \
+                                                   .filter(Annotation.groupid.in_(group_ids),
+                                                           Annotation.userid != user.userid) \
+                                                   .count()
+            if other_user_group_anns > 0:
+                raise UserDeleteError('Other users have annotated in groups created by this user')
 
-        user.groups = []
+        # Delete the user's annotations
         annotations = self.request.db.query(Annotation) \
                                      .filter_by(userid=user.userid)
         for annotation in annotations:
@@ -34,6 +46,11 @@ class DeleteUserService(object):
             event = AnnotationEvent(self.request, annotation.id, 'delete')
             self.request.notify_after_commit(event)
 
+        # Delete groups created by this user.
+        for group in user.groups:
+            self.request.db.delete(group)
+
+        # Finally, delete the user.
         self.request.db.delete(user)
 
 
--- a/h/services/delete_user.py
+++ b/h/services/delete_user.py
@@ -23,24 +23,34 @@ class DeleteUserService(object):
         message.
         """
 
-        # Check whether this user has created any groups that others have
-        # annotated in.
-        #
+        created_groups = self.request.db.query(Group) \
+                                        .filter(Group.creator == user)
+        if self._groups_have_anns_from_other_users(created_groups, user):
+            raise UserDeleteError('Other users have annotated in groups created by this user')
+
+        self._delete_annotations(user)
+        self._delete_groups(created_groups)
+        self.request.db.delete(user)
+
+    def _groups_have_anns_from_other_users(self, groups, user):
+        """
+        Return `True` if users other than `user` have annotated in `groups`.
+        """
+        group_ids = [g.pubid for g in groups]
+
         # We check for non-empty `group_ids` before querying the DB to avoid an
         # expensive SQL query if `in_` is given an empty list (see
         # https://stackoverflow.com/questions/23523147/)
-        created_groups = self.request.db.query(Group) \
-                                        .filter(Group.creator == user)
-        group_ids = [g.pubid for g in created_groups]
-        if len(group_ids) > 0:
-            other_user_ann_count = self.request.db.query(Annotation) \
-                                                  .filter(Annotation.groupid.in_(group_ids),
-                                                          Annotation.userid != user.userid) \
-                                                  .count()
-            if other_user_ann_count > 0:
-                raise UserDeleteError('Other users have annotated in groups created by this user')
+        if len(group_ids) == 0:
+            return False
 
-        # Delete the user's annotations
+        other_user_ann_count = self.request.db.query(Annotation) \
+                                              .filter(Annotation.groupid.in_(group_ids),
+                                                      Annotation.userid != user.userid) \
+                                              .count()
+        return other_user_ann_count > 0
+
+    def _delete_annotations(self, user):
         annotations = self.request.db.query(Annotation) \
                                      .filter_by(userid=user.userid)
         for annotation in annotations:
@@ -48,12 +58,9 @@ class DeleteUserService(object):
             event = AnnotationEvent(self.request, annotation.id, 'delete')
             self.request.notify_after_commit(event)
 
-        # Delete groups created by this user.
-        for group in created_groups:
+    def _delete_groups(self, groups):
+        for group in groups:
             self.request.db.delete(group)
-
-        # Finally, delete the user.
-        self.request.db.delete(user)
 
 
 def delete_user_service_factory(context, request):
--- a/h/services/delete_user.py
+++ b/h/services/delete_user.py
@@ -33,11 +33,11 @@ class DeleteUserService(object):
                                         .filter(Group.creator == user)
         group_ids = [g.pubid for g in created_groups]
         if len(group_ids) > 0:
-            other_user_group_anns = self.request.db.query(Annotation) \
-                                                   .filter(Annotation.groupid.in_(group_ids),
-                                                           Annotation.userid != user.userid) \
-                                                   .count()
-            if other_user_group_anns > 0:
+            other_user_ann_count = self.request.db.query(Annotation) \
+                                                  .filter(Annotation.groupid.in_(group_ids),
+                                                          Annotation.userid != user.userid) \
+                                                  .count()
+            if other_user_ann_count > 0:
                 raise UserDeleteError('Other users have annotated in groups created by this user')
 
         # Delete the user's annotations
--- a/h/services/delete_user.py
+++ b/h/services/delete_user.py
@@ -3,7 +3,7 @@
 from __future__ import unicode_literals
 
 from h.events import AnnotationEvent
-from h.models import Annotation
+from h.models import Annotation, Group
 from h import storage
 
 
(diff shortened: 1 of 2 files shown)


seanh:
I think the four possibilities are:

1. A user account that had created an empty group (no annotations, no members other than the creator) was deleted. The group just gets deleted as well.

2. A user account that had created a group and made some annotations in a group was deleted. No other users were members of the group. The annotations and the group just get deleted as well.

3. User has created a group, maybe created some annotations in that group or maybe not, and one or more other users have joined the group but not created any annotations in the group. The group and any group annotations by the group creator will be deleted and the other users will no longer be members of the group.

4. User has created a group, maybe created some annotations in that group or maybe not, and some other users have joined the group, and some of these other users have created some annotations. In this case the group creator and their group annotations are deleted, but group annotations by other members are not deleted, We end up with a group with no creator.

seanh:
If we get into situation 4, and then the other users who're members of the group also delete their accounts (which they can do, because they're not group creators) then we'd end up with a group containing no members and no annotations.

seanh:
Things that the group creator can currently do that other group members can't:

1. Change the group's name
2. Change the group's description
3. Moderate (hide) flagged annotations in the group (I believe?)
4. Any future creator-only group features that we add, e.g. inviting users to join public groups

If we get into situation 4, a group with no creator, then no one will be able to do these.

It looks as if you can _already_ get into situation 4 because the group creator can simply leave the group.

seanh:
> If other users' annotations are left in the group, then just unset the creator. This will leave the group in a state where nobody can perform admin actions such as moderation or editing the name, but we can always manually re-assign a new admin if necessary.

I'm not sure this is so simple because the remaining users won't necessarily understand that for example a moderation feature exists, or a group's name can be changed, and what they need to do is email support and ask to be made the group creator. We'd have to take their word about _which_ member should be made creator. And "creator" isn't really the right word for this in the model - it's an admin.

However, it's already possible for users to get themselves into this situation by the creator just leaving the group.

seanh:
@robertknight The thing about the creator of a group that also contains annotations from other users being able to leave the group, or delete their account, resulting in a broken group is awkward. I wonder how other apps handle this?

h/views/admin_users.py
```diff
@@ -137,21 +130,34 @@ def delete_user(request, user):
...
         request.notify_after_commit(event)
 
+    # Delete groups created by this user which are now empty.
+    empty_groups = []
+    for group in user.groups:
+        anns = request.db.query(models.Annotation) \
+                         .filter_by(groupid=group.pubid,
+                                    deleted=False)
+        if anns.count() > 0:
+            group.creator = None
```
seanh: Currently:

* The public group is I think the only group that has creator unset

* If the creator of a group leaves that group their membership entry is removed from the `user_group` table, so they're no longer a member, but their user id remains as `creator_id` in the group table

* `group.creator_id` is a foreign key to `user` in the DB, so we can't delete a user account and leave their `id` in `group.creator_id`.

So I guess that, if we don't want to actually delete the group, then having a group whose `creator_id` is `NULL` is the only option. But this introduces a new case - groups with `NULL` `creator_id` because the creator's user account was deleted - when previously the public group was the only group without a creator.

h/views/admin_users.py
```diff
@@ -137,21 +130,34 @@ def delete_user(request, user):
...
+        else:
+            empty_groups.append(group)
+
+    for group in empty_groups:
+        request.db.delete(group)
+    user.groups = []
```
seanh: It was there already but `user.groups = []` is unnecessary as far as I can tell

tests/h/views/admin_users_test.py
```diff
@@ -342,3 +345,26 @@ def annotation_stats_service(pyramid_config, db_session):
...
+                  name='test', pubid='group_with_two_users')
+    db_session.add(group)
+
+    doc = Document(web_uri='https://example.org')
+    user_ann = Annotation(userid=user.userid, groupid=group.pubid, document=doc)
+    other_user_ann = Annotation(userid=other_user.userid, groupid=group.pubid,
+                                document=doc)
+    db_session.add(user_ann)
+    db_session.add(other_user_ann)
+    db_session.flush()
```
seanh: It's not a problem that was introduced by this pull request, but I think code and tests that access the DB should be in the groups model and service, not in views. I guess this is a refactoring we'll need to do at some point.

robertknight:
To focus the scope of the work here, the problem I'm looking to solve is to make it easier to resolve support requests to delete user accounts, where those users have created groups. I get these frequently when I'm on support rotation. The [current documented process](https://github.com/hypothesis/playbook/blob/master/docs/support-howtos/delete-a-user-account-that-is-a-group-creator.md) requires a developer's time and even then has lots of possibility for error.

These users who make these requests have typically had minimal activity in their groups, especially from other users.

robertknight:
Following discussion on Slack I implemented a suggestion from @seanh to fail deleting the user if they have created groups in which other users have annotated, while still allowing deletion of users who have created groups in which only they have annotated, which is common in requests to remove accounts.

This punts the decision about what to do about re-assigning group admins or deleting other users' data down the road a bit, but hopefully solves the common use case in the meantime.

h/views/admin_users.py
```diff
@@ -141,17 +141,29 @@ def delete_user(request, user):
...
 
-    if models.Group.created_by(request.db, user).count() > 0:
-        raise UserDeletionError('Cannot delete user who is a group creator.')
-
-    user.groups = []
+    # Check whether this user has created any groups that others have annotated
+    # in.
+    group_ids = [g.pubid for g in user.groups]
+    other_user_group_anns = request.db.query(models.Annotation) \
+                                      .filter(models.Annotation.groupid.in_(group_ids),
```
fatbusinessman: One strange thing that I believe SQLAlchemy does is that, if you run an `IN` query on an empty set of IDs, it’ll evaluate it as `SELECT […] FROM annotation WHERE annotation.groupid != annotation.groupid`, causing a sequence scan on the whole table. In this case, you’d need something to shortcut the logic and set it to zero.

robertknight:
I'm going to rework this on top of https://github.com/hypothesis/h/pull/4724

robertknight:
This has been rebased on https://github.com/hypothesis/h/pull/4724, so the changes have now moved to `h.services.delete_user`. I have also resolved the issue with SQLAlchemy generating an inefficient query if `in_` is passed an empty list by adding an explicit check for this case.

robertknight:
One issue you might run into when testing this locally is that once an annotation is deleted, it remains in the DB for a short time until the `h.tasks.cleanup.purge_deleted_annotations` task has run at least once and at least 10 minutes after the annotation was deleted.

Currently any soft-deleted annotations (ie. still there, but with `deleted=False`) in the DB belonging to user B who has annotated in a group created by user A can cause deletion of user A to fail.

judell:
> This PR enables such users to be deleted from the admin control panel, provided that no other users have created annotations in groups they created 

Great idea, will definitely resolve a support bottleneck. Thanks!

h/services/delete_user.py
```diff
@@ -23,17 +23,34 @@ def delete(self, user):
...
-            raise UserDeleteError('Cannot delete user who is a group creator.')
-
-        user.groups = []
+        # Check whether this user has created any groups that others have
+        # annotated in.
+        #
+        # We check for non-empty `group_ids` before querying the DB to avoid an
+        # expensive SQL query if `in_` is given an empty list (see
+        # https://stackoverflow.com/questions/23523147/)
+        group_ids = [g.pubid for g in user.groups]
```
fatbusinessman: Umm… isn’t this going to check for any groups the user is a *member* of, rather than any groups they’ve *created*?

robertknight: 🤦‍♂️ yes, it is.

h/services/delete_user.py
```diff
@@ -23,17 +23,34 @@ def delete(self, user):
...
+        # Delete the user's annotations
         annotations = self.request.db.query(Annotation) \
                                      .filter_by(userid=user.userid)
         for annotation in annotations:
             storage.delete_annotation(self.request.db, annotation.id)
             event = AnnotationEvent(self.request, annotation.id, 'delete')
             self.request.notify_after_commit(event)
 
+        # Delete groups created by this user.
+        for group in user.groups:
```
robertknight: Same issue here as above, this will try to delete groups that the user is a _member_ of rather than ones they created.

h/services/delete_user.py
```diff
@@ -23,17 +23,36 @@ def delete(self, user):
         message.
         """
 
-        if Group.created_by(self.request.db, user).count() > 0:
-            raise UserDeleteError('Cannot delete user who is a group creator.')
-
-        user.groups = []
```
fatbusinessman: This threw me for a second.

I was confused that we were no longer explicitly removing the candidate user’s group memberships, had no explicit configuration for cascading deletes, and yet somehow weren’t seeing integrity errors all over the place as we tried to delete things (as we’ve seen before with `AuthTicket`s and feature flags).

For future reference, I found the answer after a bit of trawling through the SQLAlchemy documentation: deletions [cascade automatically when there’s a `secondary` table configured](http://docs.sqlalchemy.org/en/rel_1_1/orm/basic_relationships.html#deleting-rows-from-the-many-to-many-table). This didn’t apply to the group creator relationship, which was why it was in there originally, and why it’s not needed now. (See the [original commit](ab71a9273444cfcefb588bbfb6e318ae1e4e96f1) for reference.)

That makes a load more sense now, and I understand why we’re not clearing these groups any more.

h/services/delete_user.py
```diff
@@ -23,17 +23,36 @@ def delete(self, user):
...
+        # annotated in.
+        #
+        # We check for non-empty `group_ids` before querying the DB to avoid an
+        # expensive SQL query if `in_` is given an empty list (see
+        # https://stackoverflow.com/questions/23523147/)
+        created_groups = self.request.db.query(Group) \
+                                        .filter(Group.creator == user)
+        group_ids = [g.pubid for g in created_groups]
+        if len(group_ids) > 0:
+            other_user_group_anns = self.request.db.query(Annotation) \
```
fatbusinessman: A fairly minor point: the name of this variable seems to suggest it contains the annotations themselves, rather than a count of how many there are.

tests/h/services/delete_user_test.py
```diff
@@ -88,5 +96,23 @@ def api_storage(patch):
...
+
+    doc = Document(web_uri='https://example.org')
+    user_ann = Annotation(userid=user.userid, groupid=group.pubid, document=doc)
+    other_user_ann = Annotation(userid=other_user.userid, groupid=group.pubid,
+                                document=doc)
+    db_session.add(user_ann)
+    db_session.add(other_user_ann)
+    db_session.flush()
+
+    return (group, user, other_user, user_ann, other_user_ann)
```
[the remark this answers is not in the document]
robertknight: Agreed. I've changed the names.

h/services/delete_user.py
robertknight: Agreed. I've changed this.

robertknight:
> This suggests to me it could be a good time to split it up into multiple smaller private methods, so future spelunkers (and our future selves) can understand the structure and purpose of the code with a bit less mental effort.

Do you mean [like this](https://github.com/robertknight/h/blob/cb4371626d210868090cdb2555b7bb3455c3d9ec/h/services/delete_user.py#L18)? It feels like a bit of a wash to me given that the code was only 39 lines originally (of which 19 are comments/whitespace), but it's your call.

fatbusinessman:
> Do you mean [like this](https://github.com/robertknight/h/blob/cb4371626d210868090cdb2555b7bb3455c3d9ec/h/services/delete_user.py#L18)?

Yes! Exactly like that. I can take a look at that and see at a glance what the top-level method is doing and what’s involved. Then, if I’m interested, I can take a look at each of the individual methods and check how they do what they say they do. I find that *much* more comprehensible; thanks for taking the time to code it up. 🎉

robertknight:
OK, that change is pushed on this branch.

Issue #4697, opened 2017-11-23
Third party annotations are appearing on activity pages


seanh:
Third-party annotations are appearing on the https://hypothes.is/search page, and their user account, group, share, etc links are all broken. Document titles are all _Untitled document_ too, though I'm not sure whether that's to do with them being third-party.

Third-party annotations should _not_ show on activity pages, activity pages don't support them yet

dwhly:
Can't see this, can you share a pic?

seanh:
![screenshot from 2017-11-23 18-05-21](https://user-images.githubusercontent.com/22498/33185199-e83064d2-d078-11e7-9bc4-699926f9f186.png)

PR #4708, opened 2017-11-29
Index authority of annotation creator as a separate field

robertknight:
This makes it possible to filter the annotations that a search returns
by authority. This can be used in order to eg. exclude annotations from
third party accounts from the public Hypothesis activity pages.

Unlike the current indexing for the "user" field, the preprocessing is
done in Python in order to be able to reuse the existing `split_user`
function.


fatbusinessman:
For the sake of linking things together, this is a part of #4697.

PR #4710, opened 2017-11-29
Add eLife test annotations timestamps data migration

seanh:
eLife have run an API script to import 1000+ Disqus comment as page note
annotations using the standard annotation create API. After being
created in this way the annotations have the wrong timestamps - they'll
be the times at which the annotations were created in h, not the times
at which the comments were originally created in Disqus.  This migration
sets the timestamps of these imported annotations to their correct
Disqus-exported timestamps.

Here's a script you can run to add annotations with the necessary IDs into your dev DB in order to test this: <https://gist.github.com/seanh/d96af9f9581dac96624baea9c43e4b8b>. Requires a "seanh" user and at least one `Document` in the DB.

### DIFF
--- a/h/migrations/versions/7f3d80550fff_correct_timestamps_of_elife_test_annotations.py
+++ b/h/migrations/versions/7f3d80550fff_correct_timestamps_of_elife_test_annotations.py
@@ -1,0 +1,6642 @@
+"""
+Correct the timestamps of imported eLife test annotations.
+
+Set the timestamps of eLife annotations that were imported using an API script
+to their correct timestamps according to the data file that eLife gave us.
+
+The annotation IDs in this migration are ones eLife have imported into their
+TEST group.
+
+Revision ID: 7f3d80550fff
+Revises: 9bcc39244e82
+Create Date: 2017-11-29 15:06:59.207780
+"""
+
+from __future__ import unicode_literals
+
+from datetime import datetime
+import logging
+
+from alembic import op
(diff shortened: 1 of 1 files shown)


h/migrations/versions/7f3d80550fff_correct_timestamps_of_elife_annotations.py
```diff
@@ -0,0 +1,6641 @@
...
+    # transaction.
+    for timestamp in TIMESTAMPS:
+        timestamp["created"] = datetime.strptime(timestamp["created"], FORMAT)
+        timestamp["updated"] = datetime.strptime(timestamp["updated"], FORMAT)
+
+    # Now make the DB changes.
+    corrected = 0
+    for timestamp in TIMESTAMPS:
+        query = session.query(Annotation)
+        query = query.filter(Annotation.id == timestamp["id"])
```
fatbusinessman: Is there a difference between these three lines and `session.query(Annotation).get(timestamp["id"])`? If so, could we simplify this? Or is it more useful to maintain consistency between this version and the `imported_id`-based version?

seanh: Actually no, there isn't a difference, I just forgot about `get`

seanh: I've pushed this change

PR #4711, opened 2017-11-30
Dont send HTML links for third party annotations

seanh:
This is part of the fix for <https://github.com/hypothesis/client/issues/612>. The other part is <https://github.com/hypothesis/client/pull/617>.

### DIFF
--- a/h/links.py
+++ b/h/links.py
@@ -25,7 +25,12 @@ def pretty_link(url):
 
 
 def html_link(request, annotation):
-    """Generate a link to an HTML representation of an annotation."""
+    """Return a link to an HTML representation of the given annotation, or None."""
+    is_third_party_annotation = annotation.authority != request.authority
+    if is_third_party_annotation:
+        # We don't currently support HTML representations of third party
+        # annotations.
+        return None
     return request.route_url('annotation', id=annotation.id)
 
 
--- a/tests/h/links_test.py
+++ b/tests/h/links_test.py
@@ -1,7 +1,9 @@
 # -*- coding: utf-8 -*-
 
+import mock
 import pytest
 
+from h import models
 from h import links
 
 
@@ -21,6 +23,39 @@ class FakeDocumentURI(object):
 class FakeDocument(object):
     def __init__(self):
         self.document_uris = []
+
+
+@pytest.mark.usefixtures('routes')
+class TestHTMLLink(object):
+
+    def test_html_link_returns_links_for_first_party_annotations(
+            self, annotation, pyramid_request):
+        # Specify the authority so that it's a first-party annotation.
+        annotation.authority = pyramid_request.authority
+
+        link = links.html_link(pyramid_request, annotation)
+
+        assert link == 'http://example.com/a/ANNOTATION_ID'
+
+    def test_html_link_returns_None_for_third_party_annotations(
+            self, annotation, pyramid_request):
+        # Specify the authority so that it's a third-party annotation.
+        annotation.authority = 'elifesciences.org'
+
+        assert links.html_link(pyramid_request, annotation) is None
+
+    @pytest.fixture
+    def annotation(self):
+        return mock.create_autospec(
+            models.Annotation,
+            spec_set=True,
+            instance=True,
+            id='ANNOTATION_ID',
+        )
+
+    @pytest.fixture
+    def routes(self, pyramid_config):
+        pyramid_config.add_route('annotation', '/a/{id}')
 
 
 def test_incontext_link(pyramid_request):
--- a/h/models/annotation.py
+++ b/h/models/annotation.py
@@ -11,6 +11,7 @@ from sqlalchemy.ext.mutable import MutableDict, MutableList
 
 from h.db import Base, types
 from h.util import markdown, uri
+from h.util.user import split_user
 
 
 class Annotation(Base):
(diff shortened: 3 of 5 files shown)


h/links.py
```diff
@@ -25,7 +25,13 @@ def pretty_link(url):
 
 
 def html_link(request, annotation):
-    """Generate a link to an HTML representation of an annotation."""
+    """Return a link to an HTML representation of the given annotation, or None."""
+    first_party_authority = request.registry.settings['h.authority']
```
seanh: FIXME Should this be calling `h.auth.util.authority(request)`?

tests/h/links_test.py
```diff
@@ -23,6 +23,32 @@ def __init__(self):
...
 
+@pytest.mark.usefixtures('routes')
+class TestHTMLLink(object):
+
+    def test_html_link_returns_links_for_first_party_annotations(
+            self, factories, pyramid_request):
+        annotation = factories.Annotation(
+            # Specify the username in order to make sure that it's a first-party
+            # annotation.
+            userid='acct:fred@example.com',
```
seanh: TODO: I'd rather just pass `authority` here rather than the entire `userid` because the authority part is all that I care about. Or even call `factories.FirstPartyAnnotation()` or `factories.ThirdPartyAnnotation()`. But this will require adding to the annotation factories

seanh:
This is sort of done there's just a couple tweaks I want to make tomorrow morning

seanh:
FIXME: This is also adding some unicode warnings to the tests

h/links.py
```diff
@@ -25,7 +26,12 @@ def pretty_link(url):
 
 
 def html_link(request, annotation):
-    """Generate a link to an HTML representation of an annotation."""
+    """Return a link to an HTML representation of the given annotation, or None."""
+    is_third_party_annotation = annotation.authority != auth.authority(request)
```
robertknight: Elsewhere in the app we mostly seem to prefer `request.authority` rather than `auth.authority(request)`, which are the same if I understand the set up in `h.auth.__init__` correctly.

tests/h/links_test.py
```diff
@@ -23,6 +25,50 @@ def __init__(self):
         self.document_uris = []
 
 
+@pytest.mark.usefixtures('auth', 'routes')
+class TestHTMLLink(object):
+
+    def test_html_link_looks_up_the_first_party_authority(
+            self, annotation, auth, pyramid_request):
+        """It should fetch the first-party authority of the h instance."""
+        links.html_link(pyramid_request, annotation)
```
robertknight: If using the `pyramid_request.authority` property, this test can be omitted.

h/links.py
seanh: Yep. On a quick grep I see about 22 uses of `request.authority`, vs only one use of `auth.authority`, on master

seanh: And `request.authority` is just a shortcut for `auth.authority`. Ok, sounds good

seanh: Fairly big simplification. Also note that one of my tests had a param that it wasn't using which a linter would have caught if we used linters properly for h.

```diff
diff --git a/h/links.py b/h/links.py
index 2fe2e6d35..fe6f502ff 100644
--- a/h/links.py
+++ b/h/links.py
@@ -5,7 +5,6 @@ Provides links to different representations of annotations.
 """
 
 
-from h import auth
 from h._compat import urlparse, url_unquote
 
 
@@ -27,7 +26,7 @@ def pretty_link(url):
 
 def html_link(request, annotation):
     """Return a link to an HTML representation of the given annotation, or None."""
-    is_third_party_annotation = annotation.authority != auth.authority(request)
+    is_third_party_annotation = annotation.authority != request.authority
     if is_third_party_annotation:
         # We don't currently support HTML representations of third party
         # annotations.
diff --git a/tests/h/links_test.py b/tests/h/links_test.py
index af2040258..b136f63e0 100644
--- a/tests/h/links_test.py
+++ b/tests/h/links_test.py
@@ -25,27 +25,20 @@ class FakeDocument(object):
         self.document_uris = []
 
 
-@pytest.mark.usefixtures('auth', 'routes')
+@pytest.mark.usefixtures('routes')
 class TestHTMLLink(object):
 
-    def test_html_link_looks_up_the_first_party_authority(
-            self, annotation, auth, pyramid_request):
-        """It should fetch the first-party authority of the h instance."""
-        links.html_link(pyramid_request, annotation)
-
-        auth.authority.assert_called_once_with(pyramid_request)
-
     def test_html_link_returns_links_for_first_party_annotations(
-            self, annotation, auth, pyramid_request):
+            self, annotation, pyramid_request):
         # Specify the authority so that it's a first-party annotation.
-        annotation.authority = auth.authority.return_value
+        annotation.authority = pyramid_request.authority
 
         link = links.html_link(pyramid_request, annotation)
 
         assert link == 'http://example.com/a/ANNOTATION_ID'
 
     def test_html_link_returns_None_for_third_party_annotations(
-            self, annotation, auth, pyramid_request):
+            self, annotation, pyramid_request):
         # Specify the authority so that it's a third-party annotation.
         annotation.authority = 'elifesciences.org'
 
@@ -60,10 +53,6 @@ class TestHTMLLink(object):
             id='ANNOTATION_ID',
         )
 
-    @pytest.fixture
-    def auth(self, patch):
-        return patch('h.links.auth')
-
     @pytest.fixture
     def routes(self, pyramid_config):
         pyramid_config.add_route('annotation', '/a/{id}')
```

seanh:
Ok, I've pushed the `request.authority` change. As you can see `request.authority` actually didn't exist for `links.py` (because it receives a fake request object no the real one) so I had to add it.

robertknight:
> As you can see request.authority actually didn't exist for links.py (because it receives a fake request object no the real one) so I had to add it.

Ah, right, I missed that. Maybe less of an improvement now but 🤷‍♂️

PR #4713, opened 2017-12-01
Use request.authority not auth.authority()

seanh:
request.authority is a convenience shortcut for auth.authority() and is what we use elsewhere in the code.


fatbusinessman:
For my own future reference, this is [where `request.authority` is configured](https://github.com/hypothesis/h/blob/7d7b1c64040eee8301a64af0c36b852ce61886b4/h/auth/__init__.py#L60-L61)

PR #4714, opened 2017-12-01
Don't send incontext links for third-party annotations

seanh:
Don't send incontext links in the API responses for third-party annotations. Incontext links don't work for third-party annotations yet.

This is one half of the fix for <https://github.com/hypothesis/client/issues/613>, the other half is <https://github.com/hypothesis/client/pull/618>.

### DIFF
--- a/h/links.py
+++ b/h/links.py
@@ -26,8 +26,7 @@ def pretty_link(url):
 
 def html_link(request, annotation):
     """Return a link to an HTML representation of the given annotation, or None."""
-    is_third_party_annotation = annotation.authority != request.authority
-    if is_third_party_annotation:
+    if _is_third_party_annotation(request, annotation):
         # We don't currently support HTML representations of third party
         # annotations.
         return None
@@ -36,6 +35,10 @@ def html_link(request, annotation):
 
 def incontext_link(request, annotation):
     """Generate a link to an annotation on the page where it was made."""
+    if _is_third_party_annotation(request, annotation):
+        # We don't currently support incontext links to third party annotations
+        return None
+
     bouncer_url = request.registry.settings.get('h.bouncer_url')
     if not bouncer_url:
         return None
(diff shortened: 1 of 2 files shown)


seanh:
This currently breaks the client, needs a corresponding client PR to deal with it.

seanh:
This may break email notifications for third-party annotations, e.g. reply notifications, moderation notifications, if they use this same `incontext_link()` function.

seanh:
> This may break email notifications for third-party annotations, e.g. reply notifications, moderation notifications, if they use this same incontext_link() function.

I had a search through the code to see what uses these `html_link()` and `incontext_link()` functions that we're changing to sometimes return `None` in this and a previous PR:

- The JSON API. This works fine, when `None` is returned the links are omitted

- Activity pages. These actually work correctly for `incontext_link()` - the share links disappear from the annotation card when `incontext_link()` returns `None`. For `html_link()` activity pages seem to still generate and render the HTML link even if `html_link()` returns `None`, they're generating the link some other way. Activity pages aren't supported for third-party accounts anyway, so this isn't an immediate problem.

- RSS and Atom feeds (https://hypothes.is/stream.atom and https://hypothes.is/stream.rss). Again they still return the correct HTML link even if `html_link()` returns `None`. Bouncer links don't seem to be used in RSS and Atom feeds. (P.S. third-party annotations are being included in these feeds, and on https://hypothes.is/stream, they probably shouldn't be, I've opened issues for this: <https://github.com/hypothesis/h/issues/4716>, <https://github.com/hypothesis/product-backlog/issues/394>)

- Moderation and reply notification emails, see <https://github.com/hypothesis/h/issues/4715>

robertknight:
Can we hold off merging this for the moment? There are two open Bouncer PRs which will get direct links working with third party annotations and resolve this issue plus the issue with links in emails and replies.

seanh:
Yes, sorry, I should have made this clear: lets hold off on merging this as we may not need it.

seanh:
I'm gonna pre-emptively close this as I don't think we're gonna need it, gonna use Rob's work to make incontext links work for third-party accounts instead.

Issue #4721, opened 2017-12-07
Non-human-friendly usernames appearing in eLife notification emails


seanh:
Email notifications (for example reply notifications) currently contain usernames, not display names. In some cases, such as users created by eLife, the username is a nonsense string (`xyz123` type of thing) whereas the display name is something proper like Sean Hammond.

Example of what an eLife reply notification email looks like:

![screenshot from 2017-12-07 14-39-20](https://user-images.githubusercontent.com/22498/33720653-836c1b2e-db5c-11e7-8269-5cc349733318.png)

On eLife's site this username uh5qvwsm is never shown in the client, it shows the display name instead.

Suggestion: can we just always use the display name instead of the username in email subject lines and bodies, whenever a display name is present? (For both eLife and normal Hypothesis users)

dwhly:
> Suggestion: can we just always use the display name instead of the username in email subject lines and bodies, whenever a display name is present? (For both eLife and normal Hypothesis users)

That works for me.  If only username present, stick with username.  it's sufficiently limited to private correspondence to the user that it doesn't equate to a platform decision to adopt display names over usernames at UI level.

PR #4722, opened 2017-12-07
Render display names rather than usernames in reply email notifications

robertknight:
Render display names rather than usernames in reply email notifications if the user has set one. This is important particularly for third party accounts which have generated usernames not designed to be
human-friendly.

Slack discussion: https://hypothes-is.slack.com/archives/C2N6HPG2G/p1512663476000485

### DIFF
--- a/h/templates/emails/reply_notification.html.jinja2
+++ b/h/templates/emails/reply_notification.html.jinja2
@@ -1,5 +1,5 @@
 <p>
-  <a href="{{ reply_user_url }}">{{ reply_user.username }}</a>
+  <a href="{{ reply_user_url }}">{{ reply_user_display_name }}</a>
   has
   <a href="{{ reply_url }}">replied to your annotation</a>
   on
@@ -9,7 +9,7 @@
 <p>
   On
   {{ parent.created | human_timestamp }}
-  <a href="{{ parent_user_url }}">{{ parent_user.username }}</a>
+  <a href="{{ parent_user_url }}">{{ parent_user_display_name }}</a>
   commented:
 </p>
 
@@ -18,7 +18,7 @@
 <p>
   On
   {{ reply.created | human_timestamp }}
-  <a href="{{ reply_user_url }}">{{ reply_user.username }}</a>
+  <a href="{{ reply_user_url }}">{{ reply_user_display_name }}</a>
   replied:
 </p>
 
--- a/tests/h/emails/reply_notification_test.py
+++ b/tests/h/emails/reply_notification_test.py
@@ -1,5 +1,7 @@
 # -*- coding: utf-8 -*-
 
+from __future__ import unicode_literals
+
 import datetime
 
 import mock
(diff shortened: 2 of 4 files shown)


tests/h/models/user_test.py
```diff
@@ -174,6 +174,19 @@ def test_userid_in_query_with_only_invalid_userid(db_session):
     assert result == []
 
 
+def test_rendered_name_returns_display_name_if_set(db_session):
+    fred = models.User(authority='hypothes.is',
+                       username='fred',
+                       display_name='Fred Perry')
+    assert fred.rendered_name == 'Fred Perry'
```
seanh: I think if you use `factories` here to create the `User` then you won't have to specify `authority` (which isn't relevant to this test)

h/templates/emails/reply_notification.html.jinja2
```diff
@@ -1,5 +1,5 @@
 <p>
-  <a href="{{ reply_user_url }}">{{ reply_user.username }}</a>
+  <a href="{{ reply_user_url }}">{{ reply_user.rendered_name }}</a>
```
fatbusinessman: All the templates still seem to be using `rendered_name`, despite it having changed elsewhere. Given all the tests are still passing, does this mean that this bit of functionality isn’t covered by the tests?

tests/h/emails/reply_notification_test.py
```diff
@@ -109,10 +111,10 @@ def test_returns_text_and_body_results_from_renderers(self,
         assert html == 'HTML output'
         assert text == 'Text output'
 
-    def test_returns_subject_with_reply_username(self, notification, pyramid_request):
+    def test_returns_subject_with_reply_display_or_user_name(self, notification, pyramid_request, reply_user):
         _, subject, _, _ = generate(pyramid_request, notification)
 
-        assert subject == 'ron has replied to your annotation'
+        assert subject == '{} has replied to your annotation'.format(reply_user.display_or_user_name)
```
fatbusinessman: This test would still pass if the code used `display_name`, because in our current test fixture the `display_name` and `display_or_user_name` return equivalent values. This feels like there’s something missing, be that an extra test case or a smaller interface.

h/templates/emails/reply_notification.html.jinja2
robertknight: Gah, sorry. That is really sloppy of me. Indeed we don't have automated tests for the output of these templates. Like many views, there are just tests that the expected context is generated for the template.

seanh:
I think there's at least one other place in the code (the sidebar on user pages) that uses the display name or, if there isn't one, the username. The logic is probably duplicated there.

h/templates/emails/reply_notification.html.jinja2
robertknight: Ah wait. We _do_ have tests that render these templates (`test_jinja_templates_render`), but it looks like they don't catch this class of error.

robertknight:
> Given the mail templates are currently only using their parent_user and reply_user context variables to look up a name for those users, I think we can sidestep this problem by working out those names in the generate function and passing them directly into the template.

Agreed. I've simplified this PR to just determine the display names locally and removed the convenience property from the `User` model.

We haven't come to a clear agreement internally if we're going to replace usernames with display names widely, or whether it will be somewhat more context specific. So for the moment, I've just localised this change to the reply notification emails.

fatbusinessman:
For future look-up-ability: this is a fix for https://github.com/hypothesis/h/issues/4721.

PR #4724, opened 2017-12-13
Extract user deletion logic into a service

robertknight:
Follow the pattern of keeping non-trivial logic involving models out of
views by extracting the logic for deleting a user (previously the `h.views.admin_users.delete_user` helper function) into a service.

This will also make it easier to reuse this functionality from other
views in future, such as an API for deleting users or a self-service
view for a user to delete their own account. The user deletion CLI
command was already inappropriately using the helper function from
`h.views.admin_users`.

### DIFF
--- a/h/services/delete_user.py
+++ b/h/services/delete_user.py
@@ -1,0 +1,41 @@
+# -*- coding: utf-8 -*-
+
+from __future__ import unicode_literals
+
+from h.events import AnnotationEvent
+from h.models import Annotation, Group
+from h import storage
+
+
+class UserDeleteError(Exception):
+    pass
+
+
+class DeleteUserService(object):
+    def __init__(self, request):
+        self.request = request
+
+    def delete(self, user):
+        """
+        Deletes a user with all their group memberships and annotations.
+
+        Raises UserDeleteError when deletion fails with the appropriate error
+        message.
+        """
+
+        if Group.created_by(self.request.db, user).count() > 0:
+            raise UserDeleteError('Cannot delete user who is a group creator.')
+
+        user.groups = []
+        annotations = self.request.db.query(Annotation) \
+                                     .filter_by(userid=user.userid)
+        for annotation in annotations:
+            storage.delete_annotation(self.request.db, annotation.id)
+            event = AnnotationEvent(self.request, annotation.id, 'delete')
+            self.request.notify_after_commit(event)
+
+        self.request.db.delete(user)
+
+
+def delete_user_service_factory(context, request):
+    return DeleteUserService(request=request)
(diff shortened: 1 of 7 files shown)


tests/h/cli/commands/user_test.py
```diff
@@ -254,9 +255,16 @@ def password_service(hasher):
 
 
 @pytest.fixture
-def pyramid_config(pyramid_config, signup_service, password_service):
+def delete_user_service(pyramid_request):
+    return DeleteUserService(request=pyramid_request)
```
robertknight: The other tests in this module use real service instances and test the effect on the DB rather than mocking out the service. I've kept the approach here as well.

We could decide not to do that, but we ought to be consistent and probably not add that change into this PR.

h/services/delete_user.py
```diff
@@ -0,0 +1,41 @@
...
+from h.models import Annotation, Group
+from h import storage
+
+
+class UserDeleteError(Exception):
+    pass
+
+
+class DeleteUserService(object):
+    def __init__(self, request):
```
robertknight: We generally try to avoid services depending on requests, but the deletion logic relies on having an actual request in several ways so I kept the dependency.

fatbusinessman: Yup: in this case, that looks fair. It looks like the things we need from the request are:

* the database session (for querying things and then deleting things)
* the `notify_after_commit` method (for saying an annotation has been deleted)
* the request itself (for passing into an `AnnotationEvent` to say the annotation has been deleted)

While I’d like to get rid of the request dependency here, it looks like that would involve some fairly significant rework of how `AnnotationEvent`s work, so I can see now isn’t the right time.

tests/h/cli/commands/user_test.py
fatbusinessman: Yup: that sounds fair.

tests/h/views/admin_users_test.py
```diff
@@ -185,7 +181,7 @@ def test_users_activate_redirects(pyramid_request):
     assert isinstance(result, httpexceptions.HTTPFound)
 
 
-users_delete_fixtures = pytest.mark.usefixtures('user_service', 'fake_delete_user')
+users_delete_fixtures = pytest.mark.usefixtures('user_service')
```
fatbusinessman: Now that every test that’s marked with `@users_delete_fixtures` also includes `user_service` as a parameter, I think you could safely remove this.

h/services/delete_user.py
```diff
@@ -0,0 +1,41 @@
...
+
+    def delete(self, user):
+        """
+        Deletes a user with all their group memberships and annotations.
+
+        Raises UserDeleteError when deletion fails with the appropriate error
+        message.
+        """
+
+        if Group.created_by(self.request.db, user).count() > 0:
```
fatbusinessman: As a follow-up to this, I think it would be possible to replace this with a call to `groupids_created_by` from the `GroupService`, which would remove the need for the `created_by` class method altogether.

PR #4725, opened 2017-12-17
Disabled pep257:D212 rule in prospector.

hmstepanek:
Pep257:D212 conflicts with pep257:D213 which leads
to prospector lynt runs always being in violation
of one of these rules. This changes the .yaml of
prospector to use pep257:D213 as the rule (multi-
line docstrings should start at the second line).

### DIFF
--- a/.prospector.yaml
+++ b/.prospector.yaml
@@ -15,6 +15,9 @@ pep257:
     - D203 # "1 blank line required before class docstring" conflicts with
            # another pep257 rule D211 "No blank lines allowed before class
            # docstring"!
+    - D212 # "Multi-line docstring summary should start at the first line"
+           # conflicts with another pep257 rule D213 "Multi-line docstring 
+           # summary should start at the second line"
 pyflakes:
   run: false
 


fatbusinessman:
Hi there, @hstepane. Thanks for the pull request.

The team is currently in the middle of finishing up our work in progress for this last sprint of the year. Hopefully we’ll have a bit of time later in the week to review this change; if not, one of us will look through it with you in January.

fatbusinessman:
I’ve done a bit of digging into this, after an initial conversation with @hmstepanek about why I wasn’t seeing these warnings locally.

The version of the `prospector` lint tool that I’ve been using in my development environment is `0.12.4`, along with an earlier version of `pydocstyle` (before [version 1.1.0](https://github.com/PyCQA/pydocstyle/releases/tag/1.1.0) introduced the D212 and D213 checks for multi-line docstrings). `prospector` version `0.12.6` now insists on at least version `2.0.0` of `pydocstyle`.

It looks like `pydocstyle` disables these warnings by default, but when they’re run through `prospector`, they get re-enabled. I haven’t poked too deeply into why this is.

In terms of what we do next, I think we have two options:

  * decide this isn’t something we want a lint rule for, and disable both the lint rules
  * disable one of the lint rules

I ran the unfiltered `prospector` through our codebase and counted up the violations of each rule.

```
$ prospector --tool pep257 | grep D213 | wc -l
23
$ prospector --tool pep257 | grep D212 | wc -l
287
```

If we’re going to disable one of the rules, disabling D212 makes a whole load more sense. This would leave us with all multi-line docstrings needing to look like:

```python
"""
Do the thing.

Return the resultant stuff.
"""
```

As a bit of extra curiosity, I checked which files broke the D213 rule:

```
$ prospector --tool pep257 -o json | jq -r '.messages | map(select(.code == "D213")) | map(.location.path) | unique | .[]'

h/accounts/__init__.py
h/feeds/atom.py
h/feeds/render.py
h/feeds/rss.py
h/feeds/util.py
h/models/blocklist.py
h/nipsa/search.py
h/presenters/annotation_html.py
h/presenters/document_html.py
h/search/parser.py
h/search/query.py
h/session.py
h/streamer/messages.py
h/util/user.py
h/views/api.py
h/views/api_config.py
```

PR #4726, opened 2017-12-19
Public/Open Groups

gobengo:
Implements
* https://github.com/hypothesis/product-backlog/issues/369
* https://github.com/hypothesis/product-backlog/issues/370

We're adding the ability for publishers to configure hypothesis/clients to feature one or more groups. So the client will show groups you're a member of, but now also groups the publisher want to promote.

Publishers can configure one or more `pageGroups` (happy to change that to whatever) like:
```
    return {
        pageGroups: [
            "http://h.hypothesis:5000/groups/2JKqAjYE/scipub-com-open-group",
            "http://h.hypothesis:5000/groups/wg54vyqQ/h-hypothesis-open",
        ]
    };
};
```

Changes

* h.cli.commands.groups
  * New commands for add-public-group, add-open-group, join (to add a user to a group)
* h.form
  * `handle_form_submission`
    * on_failure callback can look at the error that occurred to determine what to do
    * can disable the previous flash, e.g. if the caller wants to show a more customized one
* h.models.group
  * __acl__ * remove final `terms.append(security.DENY_ALL)`. This way this model can be attached to the large authz graph, and acl lookups will go up to the parent, instead of always being denied here. Helped get authz working right for me in the admin.
* h.resources
  * GroupFactory - get a group model, but with `__parent__` the main authz Root
* h.routes
  * admin_group_create - show a form to create a new group
  * admin_group_read - get details about a group and manage members
* h.services.group
  * define `GROUP_TYPES` to reflect the business' domain model for what they want. This powers things like the group creation form's 'type' input
* h.session - `_current_groups` now deduplicates results to fix bug where API would return same group multiple times
* h/static/styles/admin.scss - The default form css is not very flexible. Had to make some tweaks.
* h/templates/activity/search.html.jinja2 - Don't always show component "Invite new members"
* h/templates/deform/form.jinja2 - Show form-level validation errors
* h/views/admin_groups.py
  * AdminGroupCreateController - for form to create new groups
  * CSVScalarWidget - deform widget for the multi-entry field when adding group members by username, email, delimited by any combination of spaces, newlines
  * AdminGroupReadController - for page of group details and member management form
* tests.functional.test_admin_groups - lots of functional tests of the admin groups forms
* tests.h.cli.commands.groups_test - tests for new group cli commands
* more tests...
* h.views.api_groups
  * Add get_group handler for API to get the id, description, name of a group
* h.views.groups
  * read_unauthenticated_json new func that responds to JSON requests to route_name='group_read'. It redirects to the api.group_read route.
    * This way the urls that users enter, which resolve to HTML in a web browser, can point over to the API and say 'look over there for the group'. I assumed it might be preferred to centralized JSON-responses in the API, but we could also have h.views.groups respond with JSON here directly.
  * route group_read now respond to OPTIONS requests for CORS supprot
* tests.h.views.groups_test - tests that these group URLs eventually redirect to a JSON object #

### DIFF
--- a/h/cli/commands/groups.py
+++ b/h/cli/commands/groups.py
@@ -42,6 +42,22 @@ def add_public_group(ctx, name, authority, creator):
     _create_group('public', ctx, name, authority, creator)
 
 
+@groups.command('add-open-group')
+@click.option('--name', prompt=True, help="The name of the group")
+@click.option('--authority', prompt=True,
+              help="The authority which the group is associated with")
+@click.option('--creator', prompt=True,
+              help="The username of the group's creator")
+@click.pass_context
+def add_open_group(ctx, name, authority, creator):
+    """
+    Create a new "open" group.
+
+    Create a new group which everyone can read and any logged-in user can write to
+    """
+    _create_group('open', ctx, name, authority, creator)
+
+
 def _create_group(type, ctx, name, authority, creator):
     """
     Create a group using group service
--- a/h/cli/commands/groups.py
+++ b/h/cli/commands/groups.py
@@ -23,7 +23,7 @@ def add_publisher_group(ctx, name, authority, creator):
     Create a new group which everyone can read but which only users belonging
     to a given authority can write to.
     """
-    _create_group('publisher', ctx, name, authority, creator)
+    _create_group('open', ctx, name, authority, creator)
 
 
 @groups.command('add-public-group')
--- a/h/cli/commands/groups.py
+++ b/h/cli/commands/groups.py
@@ -86,9 +86,11 @@ def join(ctx, user, authority, group):
     group_id = group
     request = ctx.obj['bootstrap']()
 
-    user = User.get_by_username(request.db, user, authority)
+    username = user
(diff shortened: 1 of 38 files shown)


seanh:
Closing this as it's too big to review and deploy. The code will still be here for reference. See [Slack discussion](https://hypothes-is.slack.com/archives/C4K6M7P5E/p1516798102000067)

fatbusinessman:
To paraphrase said Slack discussion, for those who don’t have access to it:

Since the work behind this PR started, we’ve taken another look at this work on different group types, and split up the problem [along different lines](https://github.com/hypothesis/product-backlog/issues?q=is%3Aissue+label%3Agroups+label%3Aepic), based around the different sets of group-related needs that we want our product to meet. Given that change in approach, combined with the size of this PR, our plan is to treat this as a prototype for later reference, rather than as a candidate to merge in as it stands.

We may end up pulling out code directly from here, or we may not. Either way, it’s not useful to the team to keep this pull request open.

PR #4728, opened 2018-01-03
Check for uncommitted DB changes when requests finish

robertknight:
Extend the logging added in 844a60dd0e6cd177935aea6ed96298199492badc to check whether a request ends with uncommitted DB changes to also check for:

- Additions and removals of objects (present in `session.new` and `session.deleted` but not `session.dirty`) which have not been flushed or committed.
- Changes which have been _flushed_ to the DB but not yet _committed_

The purpose of this is to check whether the missing DB annotations reported in https://github.com/hypothesis/h/issues/4704 are missing because the DB session was not committed or missing because the DB session was rolled back. I strongly suspect the former, but want to rule out the latter.

In addition to logging via Sentry, the issue is logged via our normal logging so that we can see details of events which have been filtered out via Sentry's rate limiting.

----

**Testing**

The easiest way to test this is to comment out [the pyramid_tm include](https://github.com/hypothesis/h/blob/b6530c431c3730dc8e91effec834106d85ff85ce/h/app.py#L90) in app.py which will have the effect of preventing sessions from being committed at the end of a request or rolled back on error. If you then create annotations or other entities you'll see log entries output at the end of the request with details of the uncommitted changes. If you re-enabled that line, the log messages should stop appearing.

### DIFF
--- a/h/util/session_tracker.py
+++ b/h/util/session_tracker.py
@@ -1,0 +1,74 @@
+from copy import copy
+from enum import Enum
+
+from sqlalchemy.event import listen
+from sqlalchemy.orm.util import identity_key
+
+
+class ObjectState(Enum):
+    ADDED = 'added'
+    DELETED = 'deleted'
+    CHANGED = 'changed'
+
+
+class Tracker(object):
+    """
+    Observer which tracks whether a SQLAlchemy `Session` has uncommitted changes.
+    """
+
+    def __init__(self, session):
+        self.session = session
+
+        # Changes which have been flushed to the database (and thus not present
+        # in `session.new`, `session.dirty` or `session.deleted) but have not
+        # yet been _committed_.
+        self._flushed_changes = {}
+
+        listen(session, 'after_flush', self._after_flush)
+        listen(session, 'after_commit', self._after_commit)
+        listen(session, 'after_rollback', self._after_rollback)
+
+    def uncommitted_changes(self):
+        """
+        Return a list of changes to the `Session` which have not yet been
+        _committed_ to the DB.
+
+        The result is a list of (identity key, ObjectState) tuples.
+        """
+        changed = copy(self._flushed_changes)
+        changed.update(self._unflushed_changes())
+
+        return list(changed.iteritems())
+
+    def _unflushed_changes(self):
+        """
+        Return a map of changes which have not yet been flushed to the DB.
+
+        In the context of the "after_flush" event handler, this returns changes
+        which have just been flushed.
+
+        If an object goes through multiple states in the same session (eg.
+        added, then flushed, then changed) then only the last state for a given
+        object is recorded.
+        """
+        changes = {}
+
+        for obj in self.session.new:
+            changes[identity_key(instance=obj)] = ObjectState.ADDED
+
+        for obj in self.session.dirty:
+            changes[identity_key(instance=obj)] = ObjectState.CHANGED
+
+        for obj in self.session.deleted:
+            changes[identity_key(instance=obj)] = ObjectState.DELETED
+
+        return changes
+
+    def _after_rollback(self, *args):
+        self._flushed_changes = {}
+
+    def _after_commit(self, *args):
+        self._flushed_changes = {}
+
+    def _after_flush(self, *args):
+        self._flushed_changes.update(self._unflushed_changes())
--- a/tests/h/util/session_tracker_test.py
+++ b/tests/h/util/session_tracker_test.py
@@ -1,0 +1,91 @@
+from __future__ import unicode_literals
+
+from uuid import uuid4
+import pytest
+from sqlalchemy.orm.util import identity_key
+
+from h.db.types import _get_urlsafe_from_hex
+from h.models import Annotation, Document
+from h.util.session_tracker import Tracker, ObjectState
+
+
+def generate_ann_id():
+    """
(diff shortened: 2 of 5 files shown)


robertknight:
Marked as WIP because I want to do some more local testing of this before merging.

tests/h/util/session_tracker_test.py
```diff
@@ -0,0 +1,88 @@
...
+    @pytest.fixture
+    def changed_ann_id(self):
+        return generate_ann_id()
+
+    @pytest.fixture
+    def deleted_ann_id(self):
+        return generate_ann_id()
+
+    @pytest.fixture
+    def session(self, db_session, changed_ann_id, deleted_ann_id):
```
fatbusinessman: I’m finding the use of the single mono-fixture tricky to follow: it makes it more difficult for me to see the intent of any individual test. Do you think there would be a way to break out individual test fixtures so that they’re a little more self-contained, or would that result in an unreasonable amount of duplication?

h/util/session_tracker.py
```diff
@@ -0,0 +1,84 @@
...
+
+        for obj in self.session.dirty:
+            changes[db_object_key(obj)] = ObjectState.CHANGED
+
+        for obj in self.session.deleted:
+            changes[db_object_key(obj)] = ObjectState.DELETED
+
+        return changes
+
+    def _after_rollback(self, session):
```
fatbusinessman: Might it be worth underscoring out the `session` arguments here and in the method below, to make it clearer that they’re not used in the methods themselves? You could even go a step further and `*args` them, to make it even more apparent that you don’t care what’s being passed into these handlers.

robertknight:
> One thing I’m wondering: is it worth feature-flagging this change, so that we can easily flip back to the old dirty behaviour if there turns out to be some obscure problem, or if it uses up our Sentry quota even faster than before?

Feature flagging does sound like a good idea, but our existing feature flagging mechanism relies on DB accesses to look up flags, which seems unwise in the context of the "finished" callback. Either we would need to preload the flag state when creating the session or perhaps use some other flagging mechanism, eg. a setting key from an env var - that will mean an EB env configuration update to change the flag. Thoughts?

judell:
> our existing feature flagging mechanism relies on DB accesses to look up flags, which seems unwise in the context of the "finished" callback

Heisenberg and Schrodinger would probably agree.

> Thoughts?

Just ship the probe? We really need to find out what's going on. Retracting it won't be hard.

fatbusinessman:
> Either we would need to preload the flag state when creating the session or perhaps use some other flagging mechanism, eg. a setting key from an env var - that will mean an EB env configuration update to change the flag. Thoughts?

Hmm. Good point. I think we’ve got two good options here:

* Use a feature flag at the beginning of the request to decide whether to go with the old-style `dirty` tracking or the new-style `Tracker`, using some kind of common abstraction for the shared functionality. I think by loading in the feature flag at the very beginning to decide which strategy we’re going to use, we can avoid getting into any particularly horrendous tangles. I can see how that abstraction would take some amount of work to figure out, though.
* As @judell says, we could decide that’s not worth the extra effort, ship it, and trust in our ability to roll the change back if there’s a problem with it.

I don’t have a particularly strong opinion on this, so I’d say it’s your call, @robertknight.

tests/h/util/session_tracker_test.py
```diff
@@ -0,0 +1,88 @@
...
+
+from uuid import uuid4
+import pytest
+
+from h.db.types import _get_urlsafe_from_hex
+from h.models import Annotation, Document
+from h.util.session_tracker import Tracker, ObjectState
+
+
+def generate_ann_id():
```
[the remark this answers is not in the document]
robertknight: I'm reusing the same setup and expectations for each test, and this makes it easier to get the IDs in the `expected_changes` fixture.

To recap, at the end of a request, a DB session could be in four possible states:

 - Changes not flushed or committed (indicates a bug)
 - Changes flushed but not committed (indicates a bug)
 - Changes committed
 - Changes rolled back

For each of those four states, there are 3 types of possible change to consider: Added, changed and deleted entries relative to the previous committed state. To avoid having to write 12 separate tests I've reused the same setup and expectations for each.

h/util/session_tracker.py
```diff
@@ -0,0 +1,74 @@
...
+        listen(session, 'after_flush', self._after_flush)
+        listen(session, 'after_commit', self._after_commit)
+        listen(session, 'after_rollback', self._after_rollback)
+
+    def uncommitted_changes(self):
+        """
+        Return a list of changes to the `Session` which have not yet been
+        _committed_ to the DB.
+
+        The result is a list of (model class, primary key, ObjectState) tuples.
```
fatbusinessman: Am I right in thinking this comment is now out of date, since 855daa89d2f472d4e54e2c8488f8e1e662df1976 changed the return type to an `(identifier, state)` tuple?

(Tangential note: I see in the Pyramid docs that the `identity_key` tuple [gets an extra element in version 1.2](http://docs.sqlalchemy.org/en/latest/changelog/migration_12.html#change-4137), so let’s make sure we don’t end up relying on the structure of the key if we can avoid it.)

robertknight: Yes indeed. The result of `identity_key` is treated as an opaque value here and in the tests so I think we should be OK wrt. changes in future SQLAlchemy releases.

Issue #4720, opened 2017-12-07
Broken user page links in eLife notification emails


seanh:
Notification emails, for example reply notifications, for eLife accounts contain links to `https://hypothes.is/users/<USERNAME>` pages. These pages don't work for third-party accounts, they 404.

Example reply notification email:

![screenshot from 2017-12-07 14-39-20](https://user-images.githubusercontent.com/22498/33720690-ad9a3160-db5c-11e7-8908-c577dae07ad3.png)

The [non-human-friendly usernames are a separate issue](https://github.com/hypothesis/h/issues/4721). This issue is about the fact that those usernames are links to Hypothesis user pages that 404.

Suggestion: just don't link to these user pages in the emails. (I don't think we have a way to link to the eLife user pages for these.)

robertknight:
> I don't think we have a way to link to the eLife user pages for these.

No. We've handled several "non-working links" issues over the last few days.

In order to solve this I think we are going to need to make third party authorities a first-class concept in h so we can associate configuration information with them such as the links that should be used for users, tags, signup etc.

Until we do that, all we can really do is remove this from the emails.

seanh:
We'd also need to change our API with eLife so that they provide us with a static HTML URL for each user, rather than the current system where they give us a JavaScript function that the client calls to open a user's profile page (that of course only works in the client and not in email notifications or on activity pages etc)

robertknight:
> We'd also need to change our API with eLife so that they provide us with a static HTML URL for each user

I don't think we will. In the context of the client we've seen that they wanted control over the behaviour of the link as well as the destination. In the context of an email however, there is just the destination of the link that we need.

PR #4729, opened 2018-01-04
Remove urls in reply notification emails for third party users

sheetaluk:
Fixes https://github.com/hypothesis/h/issues/4720

### DIFF
--- a/h/emails/reply_notification.py
+++ b/h/emails/reply_notification.py
@@ -37,6 +37,12 @@ def generate(request, notification):
     unsubscribe_token = _unsubscribe_token(request, parent_user)
     unsubscribe_url = request.route_url('unsubscribe', token=unsubscribe_token)
 
+    if notification.reply_user.authority != request.authority:
+        reply_user_url = None
+
+    if notification.parent_user.authority != request.authority:
+        parent_user_url = None
+
     context = {
         'document_title': document_title,
         'document_url': notification.parent.target_uri,
--- a/h/templates/emails/reply_notification.html.jinja2
+++ b/h/templates/emails/reply_notification.html.jinja2
@@ -1,5 +1,9 @@
 <p>
-  <a href="{{ reply_user_url }}">{{ reply_user_display_name }}</a>
+  {% if reply_user_url %}
+    <a href="{{ reply_user_url }}">{{ reply_user_display_name }}</a>
+  {% else %}
+    {{ reply_user_display_name }}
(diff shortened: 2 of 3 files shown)


hmstepanek:
I had another thought on this. Is the parent authority ever different than the user authority? The reason I ask is because the test case is the same for both the parent and the user. If they aren’t ever going to be different then we probably don’t need to pass 2 args into the context-1 would suffice.

fatbusinessman:
I have a suggestion for an alternative approach, for what it’s worth. Feel free to go with it or not, as you both see fit.

As the PR currently stands, it’s passing in Booleans for `is_third_party_parent_user` and `is_third_party_reply_user`, and the template is using those to determine whether to render the `parent_user_url` and `reply_user_url` respectively. Might it simplify matters if you eliminated the Booleans, and instead passed in `None` (or some other appropriate value that Python considers false-y) for the URLs themselves? After all, the template doesn’t necessarily need to care *why* a user does or doesn’t have a URL associated with them.

This would change the template pattern from:

```
  {% if is_third_party_reply_user %}
    {{ reply_user_display_name }}
  {% else %}
    <a href="{{ reply_user_url }}">{{ reply_user_display_name }}</a>
  {% endif %}
```

to:

```
  {% if reply_user_url %}
    <a href="{{ reply_user_url }}">{{ reply_user_display_name }}</a>
  {% else %}
    {{ reply_user_display_name }}
  {% endif %}
```

It looks like this would reduce the amount of stuff you would need to pass into the template, and eliminate the risk @hmstepanek brought up of a template accidentally rendering a URL it shouldn’t (because it’s not there in the first place).

How does that sound?

robertknight:
> I had another thought on this. Is the parent authority ever different than the user authority?

Good question. In user-facing terms, this is asking whether a third-party user (managed by a publisher) can reply to an annotation created by a first party user (managed by us) or vice-versa. The answer is currently "no". Annotations and replies live in groups and only users from one specific authority can create annotations or replies in a given group.

The approach here works fine, although I do like @fatbusinessman's suggestion to model `reply_user_url`/`parent_user_url`'s as a `unicode or None`-type and then switch on that in the template.

In tests, consider using [pytest.mark.parametrize](https://docs.pytest.org/en/latest/parametrize.html) to avoid writing a lot of code for very similar tests.

hmstepanek:
I like the alternative that @fatbusinessman proposed as well! I’m a fan. 😂

PR #4731, opened 2018-01-08
Filter out error logging from `ws4py`

fatbusinessman:
In theory, this is useful logging to have. In practice, this produces a large volume of errors of the form:

    Error when terminating the connection: [Errno 32] Broken pipe

Our best theory about these errors is that they come from over-enthusiastic AWS connection termination. Either way, it's not useful to us to have these messages, and even when we filter them out in Sentry, they still have a nasty habit of overflowing our usage quotas.

It's worth noting that this is a slightly blunt tool with which to deal with this problem, in that it will effectively silence all errors coming from the `ws4py` library. The other errors that have been known to occur in this library are:

    [Errno 9] File descriptor was closed in another greenlet
    [Errno 104] Connection reset by peer

I'm not sure either of these errors give us a whole lot of actionable information either, at least at the moment.

The `logging` library does provide us with some more fine-grained filters we could use, but that would require setting up the extra code structures to set these filters up on the logging objects. This would also come with the cost of having our logging filters set up in *two* places, rather than in a single configuration file.

### DIFF
--- a/conf/websocket.ini
+++ b/conf/websocket.ini
@@ -15,7 +15,7 @@ worker_class: h.websocket.Worker
 graceful_timeout: 0
 
 [loggers]
-keys = root, gunicorn.error, sentry
+keys = root, gunicorn.error, sentry, ws4py
 
 [handlers]
 keys = console, sentry
@@ -36,6 +36,13 @@ qualname = gunicorn.error
 level = WARNING
 handlers = console
 qualname = sentry.errors
+propagate = 0
+
+[logger_ws4py]
+level = WARNING
+qualname = ws4py
+handlers = console
+# Prevent these messages filtering through to Sentry
 propagate = 0
 
 [handler_console]


robertknight:
> It's worth noting that this is a slightly blunt tool with which to deal with this problem, in that it will effectively silence all errors coming from the ws4py library. 

Given the lower impact of real time API failures on the user vs. errors in other parts of the system I think this is a sensible move right now. If the real time API starts playing up we can always re-enable this logging, although there could be some initial confusion as to why nothing is in the logs.

For errors which happen often but at a "low" background noise rate, relative to the total number of requests, what we probably want to monitor is when the volume of errors goes up, rather than creating a Sentry report for every instance, much like we track eg. rates of different types of HTTP response.

conf/app.ini
```diff
@@ -62,6 +62,11 @@ handlers = console
 qualname = sentry.errors
 propagate = 0
 
+[logger_ws4py]
```
robertknight: Testing this locally by applying the change here to `conf/development-app.ini` I get a message about a missing required "handlers" argument.

fatbusinessman:
> The websocket server uses `conf/websocket.ini` (see `h/cli/commands/devserver.py` and `conf/supervisord.conf`)

Ah yes, so it does. Thanks for pointing that out.

> The [logger_ws4py] section is missing the required "handlers" argument.

Oops. 😳 

Based on your feedback and a sizeable amount of coffee, I’ve changed this config to be in the right place, with the right arguments. I’ve also changed it so that it still logs to the default console logger, so we can still get at these errors through the AWS logs if we really need to.

conf/websocket.ini
```diff
@@ -38,6 +38,12 @@ handlers = console
 qualname = sentry.errors
 propagate = 0
 
+[logger_ws4py]
+level = WARNING
+qualname = ws4py
+handlers = console
+propagate = 0
```
robertknight: I suggest adding a brief explanatory note here, eg. `propagate = 0 # prevent reporting to Sentry`?

robertknight:
LGTM with one minor comment. Tested locally this now logs to the console but doesn't report to Sentry.

conf/websocket.ini
fatbusinessman: Good suggestion. One minor point: I found out after [some experimenting](https://gist.github.com/fatbusinessman/72f47a76da5307303d34ffa9d94308b6) that [Pyramid’s logging config file format](https://docs.pylonsproject.org/projects/pyramid/en/1.7-branch/narr/logging.html#logging-configuration) doesn’t support inline comments. I added the comment on its own line instead.

robertknight:
Lets see if this works as expected...

robertknight:
Looking at https://sentry.io/organizations/hypothesis/stats/, 41K events were dropped for h due to rate limiting in the past 7 days. Let's see how that changes over the next week.

robertknight:
https://sentry.io/hypothesis/h/issues/275078258/

robertknight:
It is now ~27 hours since this fix was deployed and we're no longer seeing instances of https://sentry.io/hypothesis/h/issues/275078258/ , so I've marked it as resolved for now.

PR #4732, opened 2018-01-08
Fix double-reporting of DB session warning to Sentry

robertknight:
We use Sentry's integration with the "logging" package to log all
messages >= WARNING to Sentry. Therefore by using `log.warn` and also
explicitly calling `request.sentry.captureMessage` we were capturing two
reports for the same event.

Since the change list was included in the format argument to `log.warn`
rather than being passed as a separate argument, Sentry was unable to
correctly group the reports.

Fix this by removing the `captureMessage` call and passing the changes
list as a separate argument to `log.warn` so that it appears in the log
output without affecting Sentry's ability to group reports.

### DIFF
--- a/h/db/__init__.py
+++ b/h/db/__init__.py
@@ -108,11 +108,11 @@ def _session(request):
     def close_the_sqlalchemy_session(request):
         changes = tracker.uncommitted_changes() if tracker else []
         if changes:
-            msg = 'closing a session with uncommitted changes'
-            request.sentry.captureMessage(msg, stack=True, extra={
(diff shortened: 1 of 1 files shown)


robertknight:
> One thing that confuses me a little bit is the difference between what’s supposed to do in the extra argument and what’s supposed to go in a data field inside the extra argument: that’s not particular clear from the docs.

My reading of the docs is that any extra metadata that isn't a standard field _should_ go inside _data_, so that would include "changes" here. However, putting the info in a top level key in the dict still results in the field showing up in Sentry, based on testing with my dev h instance and personal Sentry account.

Issue #4736, opened 2018-01-12
User admin tool incorrectly reports success for invalid usernames


segdeha:
### Steps to reproduce
1. Visit https://hypothes.is/admin/users
2. Attempt to change a user's username to one with an invalid character (e.g., a dash)
3. System reports success

### Expected behaviour
An error message should be shown that the username has an invalid character.

### Actual behaviour
The system reports that the username was successfully changed.

### Browser/system information
Safari

### Additional details
This came up in the course of trying to resolve the following support request: https://hypothesis.zendesk.com/agent/tickets/1945

robertknight:
https://sentry.io/hypothesis/h/issues/344424008/

hmstepanek:
When I ran this in dev this is the stack trace I got:
/Documents/work/dev/h/h/tasks/admin.py(16)rename_user()
23:05:52 worker | -> user = celery.request.db.query(models.User).get(user_id)
23:05:52 worker | [2018-01-11 23:05:52,036: WARNING/Worker-2] (Pdb)
23:05:52 worker | [2018-01-11 23:05:52,040: ERROR/Worker-2] task failure: h.tasks.admin.rename_user (c6a0af7f-b550-4043-b7a7-8fe91db20435) called with args=[1, u'joe-is-cool'], kwargs={}
23:05:52 worker | Traceback (most recent call last):
23:05:52 worker |   File "/Users/hmstepanek/Documents/work/dev/h/.vdev/lib/python2.7/site-packages/celery/app/trace.py", line 240, in trace_task
23:05:52 worker |     R = retval = fun(*args, **kwargs)
23:05:52 worker |   File "/Users/hmstepanek/Documents/work/dev/h/.vdev/lib/python2.7/site-packages/celery/app/trace.py", line 438, in __protected_call__
23:05:52 worker |     return self.run(*args, **kwargs)
23:05:52 worker |   File "h/tasks/admin.py", line 16, in rename_user
23:05:52 worker |     user = celery.request.db.query(models.User).get(user_id)
23:05:52 worker |   File "h/tasks/admin.py", line 16, in rename_user
23:05:52 worker |     user = celery.request.db.query(models.User).get(user_id)
23:05:52 worker |   File "/usr/local/Cellar/python/2.7.14/Frameworks/Python.framework/Versions/2.7/lib/python2.7/bdb.py", line 49, in trace_dispatch
23:05:52 worker |     return self.dispatch_line(frame)
[...]

PR #4738, opened 2018-01-12
Add `pattern` attribute to `new_username` field

dsiddy:
HTML 5 allows us to perform a simple client-side check of the username
using the `input` element's `pattern` attribute, obviating the need for
a JavaScript-based solution.

Closes: #4736


dsiddy:
You're most welcome. Delighted to contribute. 😊

PR #4739, opened 2018-01-16
Fix docker-compose logs cmd

patmigliaccio:
Service is named `rabbit` in the `docker-compose.yml` not `rabbitmq`


fatbusinessman:
Ah yes. Well spotted, @patmigliaccio, and thanks for the fix. 🙂

Issue #4734, opened 2018-01-09
Frequent Elasticsearch request timeout warnings


robertknight:
We are seeing about ~200 `ReadTimeout` warnings from Elasticsearch per day in production. These are a significant contribution towards our Sentry event quota. ES is configured to automatically retry such requests so users won't see an error unless the request times out repeatedly. Given the ratio of `ReadTimeout` _warnings_ vs _errors_ on Sentry, actual failures are not happening frequently at the moment.

We should deal with this one way or another, eg. by:

- Deciding that the warning doesn't matter for the moment and filtering it out
- Increasing the timeout from the current 0.2s
- Other infrastructure changes to reduce the frequency (we are using Amazon's managed Elasticsearch, so some of this may be out of our control)

https://sentry.io/hypothesis/h/issues/218088571/

```
ReadTimeout: HTTPSConnectionPool(host='search-hypothesis-2marq57xdwzqpon3xq5gc5nt3a.us-west-1.es.amazonaws.com', port=443): Read timed out. (read timeout=0.2)
(1 additional frame(s) were not displayed)
...
  File "requests/sessions.py", line 488, in request
    resp = self.send(prep, **send_kwargs)
  File "raven/breadcrumbs.py", line 297, in send
    resp = real_send(self, request, *args, **kwargs)
  File "newrelic/api/external_trace.py", line 124, in dynamic_wrapper
    return wrapped(*args, **kwargs)
  File "requests/sessions.py", line 609, in send
    r = adapter.send(request, **kwargs)
  File "requests/adapters.py", line 499, in send
    raise ReadTimeout(e, request=request)

GET https://search-hypothesis-2marq57xdwzqpon3xq5gc5nt3a.us-west-1.es.amazonaws.com:443/hypothesis/annotation/_search?_source=false [status:N/A request:0.205s]
```

fatbusinessman:
One extra piece of context: we also have [instrumentation in the code](https://github.com/hypothesis/h/blob/5140713ee900c9b5294dd0a37499699a524b2fb9/h/search/core.py#L126-L145) that reports these timeouts to `statsd`, so we can see them in Grafana. I’m not 100% sure whether this also reports the non-error timeouts (the ones that trigger a retry within the Elasticsearch library). If it does report these, that’s an argument for us being able to discard these messages entirely; if it doesn’t report them, it might offer us a useful pattern for intercepting these messages within the app and reporting them to `statsd` ourselves.

hmstepanek:
[...]

PR #4742, opened 2018-01-19
avoid logging certain exceptions in Sentry

hmstepanek:
Created a customized ExceptionFilter with a tuple
of (exception name, logging level) exception messages to filter out
.) This is attached to the elastic search logger to filter out 
ReadTimeout warnings from elastic search. This is a fix for #4734 .


h/search/logging_filters.py
```diff
@@ -0,0 +1,44 @@
+# -*- coding: utf-8 -*-
+"""Logging Filters."""
```
robertknight: The code in this module isn't actually specific to search at all, so I would suggest we put it in the `h.util` package instead, so that it'll be easier to find if we have the same problem in another context.

h/search/logging_filters.py
```diff
@@ -0,0 +1,44 @@
...
+                self._ignore_exceptions.append((logging_level_names[exc_level],
+                                                getattr(__import__(module), exception)))
+            except KeyError:
+                raise ValueError("""The logging level provided ({})
+                                 is invalid. Valid options: {}"""
+                                 .format(exc_level, logging_level_names.keys()))
+            except (ImportError, AttributeError):
+                raise ValueError('The exception path does not exist ({}.{}).'
+                                 .format(module, exception))
+        super(FilterExceptions, self).__init__()
```
robertknight: Is there a reason for the base class initialization not to be at the top of this method? Usually this is good practice so that subclass-specific init code can rely on the state of the object being correct.

h/search/logging_filters.py
```diff
@@ -0,0 +1,44 @@
...
+
+class FilterExceptions(logging.Filter):
+    """Filter out the specified exceptions with specified logging level."""
+
+    def __init__(self, ignore_exceptions):
+        """
+        Configure filtering out of the specified exceptions with specified logging level.
+
+        ignore_exceptions: a tuple of tuples ((exception type, loglevel))
+                           example: (("requests.exceptions.ReadTimeout", "WARNING"),)
```
robertknight: What was the rationale for making `ignore_exceptions` a list of string tuples rather than say, a list of `(exception class, numeric logging level)`, eg. `[(requests.exceptions.ReadTimeout, logging.WARNING)]`?

There are several lines of code here and test cases needed to check the (string, string) => (type, number) conversion and you might be able to avoid that by taking the type + number directly.

h/search/logging_filters.py
```diff
@@ -0,0 +1,44 @@
...
+        """
+        logging_level_names = {level_name:
+                               (getattr(logging, level_name)
+                                if isinstance(level_name, str)
+                                else level_name)
+                               for level_name in logging._levelNames}  #pylint: disable=protected-access
```
robertknight: It is a good idea to validate the logging level argument. However I think we should avoid accessing private vars of third party packages unless really necessary. `logging._levelNames` [doesn't exist in Python 3](https://github.com/python/cpython/blob/master/Lib/logging/__init__.py) for example.

Since the list of logging levels is pretty small and doesn't change often, we could simply hardcode a list of the most common ones that we are likely to use.

tests/h/search/test_logging_filters.py
```diff
@@ -0,0 +1,66 @@
...
+import pytest
+import logging
+from requests.exceptions import ReadTimeout
+from h.search.logging_filters import FilterExceptions
+
+
+class TestLoggingFilters(object):
+    """Test logging filters."""
+
+    def test_filter_bad_level(self):  #pylint: disable=no-self-use
```
robertknight: I'm afraid our linting story is not super-clear at the moment for new devs. [flake8](http://flake8.pycqa.org/en/latest/) is the one tool that all of us run locally, is quick and I've proposed we make part of the build process ([feedback on this welcome 🙂](https://hypothes-is.slack.com/archives/C1MA4E9B9/p1517481793000097))

h/search/logging_filters.py
hmstepanek: I did this thinking that this is purely configuration and there does exist a python dict configuration for loggers. If we ever switch to that kind of config then I don't think we can use a raw import statement. Note this does except both string and number type for the log level. I can add a specific test indicating that as I don't think I have one right now.

h/search/logging_filters.py
hmstepanek: Ah ok. Didn't consider python3...I'll change it to hard code them. Unfortunately there doesn't appear to be a good way to get a list of them from the logger without accessing private vars.

h/search/logging_filters.py
hmstepanek: No. I fix...

tests/h/search/test_logging_filters.py
hmstepanek: Ah ok. This was coming from prospector I believe. I can remove it.

h/search/logging_filters.py
robertknight: > I did this thinking that this is purely configuration and there does exist a python dict configuration for loggers.

Ah, OK, I see. My bias would be to apply the YAGNI principle here, but I'll leave this up to you to make a call.

h/util/logging_filters.py
```diff
@@ -0,0 +1,49 @@
...
+
+LEVELS = {"CRITICAL": 50,
+          "ERROR": 40,
+          "WARNING": 30,
+          "INFO": 20,
+          "DEBUG": 10,
+          "NOTSET": 0}
+
+
+class FilterExceptions(logging.Filter):
```
robertknight: Since the base class is named with a noun `logging.Filter`, I would suggest naming this as a noun rather than a verb - `ExceptionFilter`?

h/util/logging_filters.py
```diff
@@ -0,0 +1,49 @@
...
+
+    def __init__(self, ignore_exceptions):
+        """
+        Configure filtering out of the specified exceptions with specified logging level.
+
+        ignore_exceptions: a tuple of tuples ((exception type, loglevel))
+                           example: (("requests.exceptions.ReadTimeout", "WARNING"),)
+        """
+        super(FilterExceptions, self).__init__()
+        logging_level_names = {val: val for val in LEVELS.values()}
```
robertknight: `logging_level_names` looks redundant to me. It is being used to map logging level names to values but `LEVELS` already contains this mapping.

tests/h/util/test_logging_filters.py
```diff
@@ -0,0 +1,69 @@
...
+"""Test logging filters."""
+from __future__ import unicode_literals
+
+import pytest
+import logging
+from requests.exceptions import ReadTimeout
+from h.util.logging_filters import FilterExceptions
+
+
+class TestLoggingFilters(object):
```
robertknight: I forgot to mention this last time, but if you have a class of tests that test a particular class or function, the convention in our code would be to name it `Test<capitalized class or function name>`.

tests/h/util/test_logging_filters.py
```diff
@@ -0,0 +1,69 @@
+# -*- coding: utf-8 -*-
+"""Test logging filters."""
```
robertknight: The "Test logging filters" comment here and below can be removed, since it is obvious that this is a test and what is being tested from the name of the module.

h/util/logging_filters.py
```diff
@@ -0,0 +1,49 @@
+# -*- coding: utf-8 -*-
+"""Logging Filters."""
+
+import logging
+
+
+LEVELS = {"CRITICAL": 50,
```
robertknight: Perhaps worth adding a brief comment that these come from https://docs.python.org/2/library/logging.html#logging-levels .

tests/h/util/test_logging_filters.py
```diff
@@ -0,0 +1,69 @@
...
+            FilterExceptions((("requests.exceptions.Bugga", "WARN"),))
+
+    def test_filter_wrong_module_name(self):
+        with pytest.raises(ValueError):
+            FilterExceptions((("requests.bad.ReadTimeout", "WARN"),))
+
+    def test_filter_level_as_num(self, logger):
+        FilterExceptions((("requests.exceptions.ReadTimeout", logging.WARNING),))
+
+    def test_filter_no_log(self, logger):
```
robertknight: A general note about test naming - we usually use BDD-style naming for tests to make it easier to see at a glance what the test checks for. eg. `test_filter_removes_matching_exceptions` or `test_raises_if_wrong_module_name`. It took me a couple of reads through to figure out which were the success vs. error handling cases.

tests/h/util/test_logging_filters.py
```diff
@@ -0,0 +1,69 @@
...
+        try:
+            raise ReadTimeout("this is a test read timeout error")
+        except ReadTimeout:
+            logger.warn("warning", exc_info=True)
+        assert not logger.handlers[0].handler_called, "Didn't filter out log message when it should have!!"
+
+    def test_filter_log_level_mistmatch(self, logger):
```
robertknight: Typo `mistmatch`

h/util/logging_filters.py
hmstepanek: It's actually being used to add additional mapping for 30:30 mapping as well as the original 'WARNING':30 so you can pass in a string or an int and they will both decode to the int logging level. I can add a comment about that...

tests/h/util/test_logging_filters.py
```diff
@@ -0,0 +1,67 @@
...
+
+import pytest
+import logging
+from requests.exceptions import ReadTimeout
+from h.util.logging_filters import ExceptionFilter
+
+
+class TestExceptionFilter(object):
+
+    def test_raises_if_invalid_level_name(self):
```
robertknight: These names are much clearer, thanks 👍

h/util/logging_filters.py
```diff
@@ -0,0 +1,53 @@
...
+    def __init__(self, ignore_exceptions):
+        """
+        Configure filtering out of the specified exceptions with specified logging level.
+
+        ignore_exceptions: a tuple of tuples ((exception type, loglevel))
+                           example: (("requests.exceptions.ReadTimeout", "WARNING"),)
+        """
+        super(ExceptionFilter, self).__init__()
+
+        # build decoder dict where both string and int map to int logging level
```
robertknight: Ah I see, so the purpose of `logging_levels` is to allow levels in the second element of `ignore_exceptions` tuples to be int values or string names. In either case, if an unrecognised logging level or value is used, the same exception is thrown.

h/util/logging_filters.py
```diff
@@ -0,0 +1,53 @@
...
+        logging_levels = {val: val for val in LEVELS.values()}
+        logging_levels.update(LEVELS)
+
+        self._ignore_exceptions = []
+        for (exc_type, exc_level) in ignore_exceptions:
+            exception_idx = exc_type.rfind('.')
+            module, exception = exc_type[:exception_idx], exc_type[exception_idx+1:]
+            try:
+                self._ignore_exceptions.append((logging_levels[exc_level],
+                                                getattr(__import__(module), exception)))
```
robertknight: While testing this locally, I found a subtle issue with `__import__` which causes this function not to behave as expected if `module` is a dotted path, for example if `exc_type` is `"urllib3.exceptions.ReadTimeoutError"`, then. `__import__('urllib3.exceptions')` returns a reference to the top-level module ("urllib3") and so the `getattr` call may fail.

[From the docs](https://docs.python.org/2/library/functions.html#__import__):

> When the name variable is of the form package.module, normally, the top-level package (the name up till the first dot) is returned, not the module named by name. However, when a non-empty fromlist argument is given, the module named by name is returned.

This code is however working for `requests.exceptions.ReadTimeout`. Why? Because the `requests` package re-exports the `ReadTimeout` class at the top level so `from requests import ReadTimeout` works.

hmstepanek: Wah! Good catch! I fix...

hmstepanek: So this converting a string path to the exception class is becoming overly complicated. Are we ok with just accepting the exception name rather than the full path? That will simplify the code dramatically. (I'm committing it with this change but I can undo it if you don't like it.) One advantage of using just the name would be it would handle differences between the dev and prod env (because dev uses urllib3 and prod uses http_requests). Alternatively we can just change it to not be a string as you suggested earlier.

robertknight: > Are we ok with just accepting the exception name rather than the full path? 

While that's probably Good Enough for our needs right now, I would still be inclined to avoid the string conversion and just require the exception class to be passed, until we have a concrete need to specify the exception type as a string.

If that concrete need does ever arise, an option I forgot about yesterday is that Pyramid exposes a [utility class](https://docs.pylonsproject.org/projects/pyramid/en/latest/api/path.html#pyramid.path.DottedNameResolver) which can resolve dotted names to classes. See `h/cli/__init__.py` for example usage.

> One advantage of using just the name would be it would handle differences between the dev and prod env (because dev uses urllib3 and prod uses http_requests). 

The exception has a slightly different name in `requests` and `urllib3` - `ReadTimeout` vs `ReadTimeoutError`.

Issue #4741, opened 2018-01-18
A few annotations not returned in the search results


nlisgo:
### Steps to reproduce
1. We created 1348 public annotations in a publisher group for 896 users in a short space of time. Only 4 annotations were not discoverable in the search results. We were using a test authority. As these were the only public annotations in the publisher group, it was easy to discover that there were 4 annotations missing from search because the total was 1344.
2. After posting all of the annotations had been created I confirm that they exist by hitting the `/annotations/{id}` endpoint.
3. Gather all of the annotation ids from the `/search?group=[GROUP_ID]` endpoint and ensure that each annotation id for the annotations that you created is present in the search results. I found 4 that were not. I also checked to see if the annotations were present if search was filtered by username and they still were not returned.
4. Wondering whether it was just an issue with the search index, I updated one of the annotations that was missing from the search and it did not fix the problem.

nlisgo:
Update: I deleted all of the annotations and created them again and the same 4 were created but couldn't not be found in the search results.

Then I deleted the annotations and created only the 4 annotations plus their parent annotations. This left me with 8 annotations and I had the same result with 4 rows in the search results and 4 additional annotations had been created but could not be found in the search results.

This rules out it being a random failure and caused by there being a high number of annotations created in a short amount of time.

robertknight:
> Update: I deleted all of the annotations and created them again and the same 4 were created but couldn't not be found in the search results.

Oh, that's interesting.

I'm currently running a script to check the status in ES of a subset of annotations in Postgres. So far I found from an initial check of the 1000 most recent additions:

```
7f6b545a-fd0c-11e7-be6a-a78f6dd2598c / f2tUWv0MEee-aqePbdJZjA is missing from Elasticsearch
7d6dde48-fd0c-11e7-99b0-332752c1bd75 / fW3eSP0MEeeZsDMnUsG9dQ is missing from Elasticsearch
7b6cf566-fd0c-11e7-b877-9bd9def938d2 / e2z1Zv0MEee4d5vZ3vk40g is missing from Elasticsearch
7a13d8a6-fd0c-11e7-9788-fbd549f991e8 / ehPYpv0MEeeXiPvVSfmR6A is missing from Elasticsearch
```

robertknight:
[...]

PR #4743, opened 2018-01-22
Fix unicode encoding error when sending reply notification emails

### LINKED ISSUE #4733: Unicode encoding error when sending reply email notification to user who has a non-ASCII display name
Users can set display names in their profile which can include non-ASCII characters. It looks like sending reply notification emails can cause an error when this happens. See the values of `reply_user_display_name` shown in the report for an example.

**Steps to reproduce:** (Hypothesized from the bug report, I haven't actually tested this)

1. Create user A, set their display name (not username) to something with non-ASCII chars in it
2. Create user B, set their display name (not username) to something with non-ASCII chars in it
3. Create an annotation as user A in the client
4. Create a reply as user B in the client

**Expected:**

A reply notification email is sent.

**Actual**

Unicode encoding error.

**Notes**

https://sentry.io/hypothesis/h/issues/430026663/

```
UnicodeEncodeError: 'ascii' codec can't encode character u'\xe9' in position 1: ordinal not in range(128)
(2 additional frame(s) were not displayed)
...
  File "/usr/lib/python2.7/site-packages/zope/interface/registry.py", line 442, in subscribers
    return self.adapters.subscribers(objects, provided)
  File "/usr/lib/python2.7/site-packages/zope/interface/adapter.py", line 596, in subscribers
    subscription(*objects)
  File "pyramid/config/adapters.py", line 103, in derived_subscriber
    return subscriber(arg[0])
  File "h/subscribers.py", line 54, in send_reply_notifications
    send_params = generate_mail(request, notification)
  File "h/emails/reply_notification.py", line 54, in generate
    user=reply_user_display_name)

UnicodeEncodeError: 'ascii' codec can't encode character u'\xe9' in position 1: ordinal not in range(128)
```

robertknight:
Generating the subject line of reply notification emails failed if the
replying user's display name contained non-ASCII chars due to the format
string being a non-Unicode literal.

Fixes #4733

Also fixes one of the causes of #4741 (_Edit: Verified that indeed, it does_)

### DIFF
--- a/h/emails/reply_notification.py
+++ b/h/emails/reply_notification.py
@@ -1,5 +1,7 @@
 # -*- coding: utf-8 -*-
 
+from __future__ import unicode_literals
+
 from h import links
 
 from pyramid.renderers import render
--- a/tests/h/emails/reply_notification_test.py
+++ b/tests/h/emails/reply_notification_test.py
@@ -98,6 +98,20 @@ class TestGenerate(object):
         html_renderer.assert_(**expected_context)
         text_renderer.assert_(**expected_context)
 
+    def test_supports_non_ascii_display_names(self,
+                                              notification,
+                                              pyramid_request,
+                                              html_renderer,
+                                              text_renderer,
+                                              parent_user,
+                                              reply_user):
+        parent_user.display_name = 'Parent 👩'
+        reply_user.display_name = 'Child 👧'
+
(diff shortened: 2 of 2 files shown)


robertknight:
This is now working as expected on QA:

screenshot 2018-01-23 14 01 03

PR #4749, opened 2018-01-23
Add a TODO to remove a zope.sqlalchemy workaround

seanh:


h/db/__init__.py
```diff
@@ -115,13 +115,30 @@ def close_the_sqlalchemy_session(request):
...
+        #
+        # If a future request's session then happens to get the same Python
+        # object ID as one of these "stale" IDs not removed from
+        # _SESSION_STATE, then zope.sqlalchemy does not join that sqlalchemy
+        # session to the transaction manager's transaction because it thinks it
+        # has already done so. As a result, that session is never committed
+        # (annotations are not saved, etc).
+        #
+        # To prevent that from happening we remove stale IDs from
+        # _SESSION_STATE here.
```
seanh: I thought it a good idea to flesh out this comment in more detail. (Normally I wouldn't write a code comment this long but I think this is a special case.)

PR #4750, opened 2018-01-24
Update a code comment

seanh:
The group creator _is_ now used for some things. They're the one person
who can rename the group, for example. They're also the one and only
moderator of the group.

I didn't mention in the new code comment what the group creator is
currently used for as I think that comment would likely get out of date.


seanh:
Edited to just delete the comment instead

PR #4754, opened 2018-01-26
Allow logged-out users to view group pages for world-readable groups

robertknight:
sConsolidate the logic for handling group read pages for logged-in and
logged-out users in `h.views.activity.GroupSearchController`, and let
logged-out users view world-readable groups.

With this change, it is now possible for a logged-out user to see the
activity page for a world-readable group such as "Public" or in future
open or restricted groups, eg. at https://hypothes.is/groups/__world__

This should also fix an issue @seanh spotted where an existing link to `/groups/__world__` is broken for anonymous users:

1. Go to the `/search` page when logged out
2. Find an annotation from the _Public_ group and click on the link to the _Public_ group in the top left of the annotation card
3. Link takes you to https://hypothes.is/groups/__world__/public which 404s

### DIFF
--- a/h/views/activity.py
+++ b/h/views/activity.py
@@ -8,7 +8,6 @@ import urlparse
 
 from jinja2 import Markup
 from pyramid import httpexceptions
-from pyramid import security
 from pyramid.view import view_config
 from pyramid.view import view_defaults
 
@@ -87,7 +86,6 @@ class SearchController(object):
 
 @view_defaults(route_name='group_read',
                renderer='h:templates/activity/search.html.jinja2',
-               effective_principals=security.Authenticated,
                request_method='GET')
 class GroupSearchController(SearchController):
     """View callables unique to the "group_read" route."""
@@ -257,9 +255,15 @@ class GroupSearchController(SearchController):
 
     def _check_access_permissions(self):
         if not self.request.has_permission('read', self.group):
-            if self.request.has_permission('join', self.group):
+            show_join_page = self.request.has_permission('join', self.group)
+            if not self.request.user:
+                # Show a page which will prompt the user to login to join.
+                show_join_page = True
+
+            if show_join_page:
                 self.request.override_renderer = 'h:templates/groups/join.html.jinja2'
                 return {'group': self.group}
+
             raise httpexceptions.HTTPNotFound()
 
 
--- a/h/views/groups.py
+++ b/h/views/groups.py
@@ -3,7 +3,6 @@
 import deform
 from pyramid import httpexceptions
 from pyramid import security
-from pyramid.config import not_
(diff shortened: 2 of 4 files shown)


seanh:
Rebased

h/views/activity.py
```diff
@@ -87,7 +86,6 @@ def user_link(userid):
 
 @view_defaults(route_name='group_read',
                renderer='h:templates/activity/search.html.jinja2',
-               effective_principals=security.Authenticated,
```
seanh: I wonder why this was here in the first place? It doesn't seem to achieve anything other than breaking the public group's page

seanh: I tested visiting a private group's page when logged out, when logged in but not a member of the group, and when logged in and a member of the group - all seems to work as expected. So it looks like it's fine to remove this.

robertknight: I'm assuming this was to avoid a conflict with the "unauthenticated read" view in `h.views.groups`. Otherwise Pyramid would presumably complain about having multiple views matching a given request.

h/views/activity.py
```diff
@@ -257,9 +255,15 @@ def toggle_tag_facet(self):
 
     def _check_access_permissions(self):
         if not self.request.has_permission('read', self.group):
-            if self.request.has_permission('join', self.group):
+            show_join_page = self.request.has_permission('join', self.group)
+            if not self.request.user:
+                # Show a page which will prompt the user to login to join.
+                show_join_page = True
+
+            if show_join_page:
```
seanh: Ok, so the reason why visiting a private group's page when logged in but not a member of the group still works is that you added explicit logic for it here.

The comment `# Show a page which will prompt the user to login to join` may be a bit misleading because it looks like `join.html.jinja2` actually will _either_ prompt you to login (if you aren't) _or_ to join (if you're logged in).

I think the way this worked originally was that we had separate controllers for the group login page (the one that prompts you to login in order to join), the group join page, and the group search page (for members). We then used [Pyramid view configuration predicate arguments](https://docs.pylonsproject.org/projects/pyramid/en/latest/narr/viewconfig.html?highlight=effective_principals#predicate-arguments) so that Pyramid routing would direct the request to the correct page based on whether the user was logged in or not and whether they were a member of the group or not. I thought that was nice because I kept the login to join, the join, and the search page views and templates separate from each other, and kept the logic about which page to show out of these view functions and in Pyramid routing config instead. That all seems to have been changed though, since I last looked at this. Never mind :\

h/views/activity.py
seanh: Yes, that's right. Pyramid was deliberately configured so that _this_ view would be selected when you were logged in, but unauthenticated read would be selected when you were logged out.

PR #4757, opened 2018-01-31
Rename publisher groups to open groups

seanh:
This is just a few commits renaming "publisher group" to "open group" throughout h. No behaviour changes here, just renaming.

It turns out that we we originally called "publisher groups" when we developed them for the eLife project (groups that are readable by anyone, writeable by anyone in the same authority, and can't be joined) are the same kind of groups as what we now want for the new "open groups" feature, so we're renaming the existing "publisher groups" to "open groups" throughout.

The only difference is that groups like eLife's belong to a third-party authority (they're "third-party open groups") whereas the new open groups we're working on are going to be first-party.

The new open groups also need to be scoped to a domain or domains, but that should and will also be true of eLife's group (which is currently not scoped but should be).

P.S. Even without any behaviour changes **it is already possible to create a first- or third-party open group in the DB**:

```
./bin/hypothesis --dev groups add-open-group --name Foobar --authority hypothes.is --creator seanh
```

It's not scoped, and it doesn't appear to work (doesn't appear in groups dropdown on activity pages or in client), but that will create a first-party open group in the DB. (Haven't tested what happens if you try to use this group via the API, or visit its activity page directly.)

P.S. I'll also look in to a client commit to rename publisher to open.

### DIFF
--- a/h/cli/commands/groups.py
+++ b/h/cli/commands/groups.py
@@ -8,17 +8,17 @@ def groups():
     """Manage groups."""
 
 
-@groups.command('add-publisher-group')
+@groups.command('add-open-group')
 @click.option('--name', prompt=True, help="The name of the group")
 @click.option('--authority', prompt=True, help="The authority which the group is associated with")
 @click.option('--creator', prompt=True, help="The username of the group's creator")
 @click.pass_context
-def add_publisher_group(ctx, name, authority, creator):
+def add_open_group(ctx, name, authority, creator):
     """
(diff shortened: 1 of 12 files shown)


seanh:
Did a quick test and:

- I can't create annotations in open groups by the API yet (I get an authorization error)
- I can visit the activity page for an open group but there's [a list of problems with the page](https://github.com/hypothesis/product-backlog/issues/422)

PR #4771, opened 2018-02-06
Clarify that userid of the creator is published

seanh:
Clarify that it's the userid of the group's creator that's published
when the 'group-join' event is published when creating a group, by
publishing group.creator.userid instead of the local variable userid.

This will also make later refactoring easier as the publish line needs
only group and not the local userid variable.

Also fix a test: this group service test was expecting 'theresa' (the
username) to be published but in all cases what the real code actually
publishes is 'acct:theresa@example.com' (the userid) (see h/views/groups.py
and also h/cli/commands/groups.py, and note that the local variable was
called userid not username). Fix the test to expect the userid not the
username.

### DIFF
--- a/h/services/group.py
+++ b/h/services/group.py
@@ -68,7 +68,7 @@ class GroupService(object):
         self.session.add(group)
         self.session.flush()
 
-        self.publish('group-join', group.pubid, userid)
+        self.publish('group-join', group.pubid, group.creator.userid)
 
         return group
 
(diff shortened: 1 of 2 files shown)


seanh:
Note that `services/group_test.py` is still passing in usernames instead of userids whenever it calls `create()`. This works because the fake user_fetcher that the tests pass in to `GroupService` is one that fetches by username. In the real code userids not usernames are passed in to `GroupService.create()` and in to the real `user_fetcher`. This difference between tests and code seems needlessly confusing and I may fix it in a separate pr but it would require changes to a lot of tests.

robertknight:
LGTM.

PR #4774, opened 2018-02-06
Make tests flake8 "clean"

robertknight:
Following on from https://github.com/hypothesis/h/pull/4766, this PR:

- Adds a custom flake8 config for the tests that ignores lint warnings that are, IMO, fine to ignore in tests (see comments in `tests/setup.cfg`).
- Includes a set of commits to fix the remaining flake8 violations
- Causes the lint Travis task to run flake8 over the tests

In the interests of keeping the PR small I opted to ignore all indentation-related errors in tests for this initial change.

One issue I'm note sure about, but is whether the setup.cfg file will end up entirely replacing the top-level setup.cfg file for non-flake8 tools or whether the configuration is combined.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -44,6 +44,7 @@ test: node_modules/.uptodate
 .PHONY: lint
 lint: .pydeps
 	flake8 h
+	flake8 tests
 
 ################################################################################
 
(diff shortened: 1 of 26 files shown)


seanh:
> One issue I'm note sure about, but is whether the setup.cfg file will end up entirely replacing the top-
> level setup.cfg file for non-flake8 tools or whether the configuration is combined.

I think it'll replace. If that's an issue then `tests/setup.cfg` could be renamed to `tests/.flake8`.

PR #4779, opened 2018-02-08
enabling activity pages for open groups

hmstepanek:
Enabled members to be a None type and added 'moderators' and 'creator'. If members
is None it indicates that the group has no concept of members and
is an "open" group. Moderators looks just like members but is simply
a list of the creator-this will probably be expanded at some point
to be an actual list of moderators but until then it's a list of one. Creator 
is used to identify the user who created the group via a 'creator' label on the front end.
Suggested groups which is used to auto-complete the search now contains
suggestions for open groups.
This adds features required for https://github.com/hypothesis/product-backlog/issues/422.

### DIFF
--- a/h/templates/activity/search.html.jinja2
+++ b/h/templates/activity/search.html.jinja2
@@ -143,7 +143,7 @@
 {% macro tags_section() %}
 <section class="search-result-sidebar__section">
   <h3 class="search-result-sidebar__subtitle">
-    {% trans %}Tags{% endtrans %}
+    {% trans %}Top tags{% endtrans %}
     <span class="search-result-sidebar__subtitle-count">
       {{ search_results.aggregations['tags'] | length }}
     </span>
--- a/h/templates/activity/search.html.jinja2
+++ b/h/templates/activity/search.html.jinja2
@@ -143,7 +143,7 @@
 {% macro tags_section() %}
 <section class="search-result-sidebar__section">
   <h3 class="search-result-sidebar__subtitle">
-    {% trans %}Top tags{% endtrans %}
+    {% trans %}Tags{% endtrans %}
     <span class="search-result-sidebar__subtitle-count">
       {{ search_results.aggregations['tags'] | length }}
     </span>
@@ -170,6 +170,44 @@
       </button>
     {% endfor %}
   </div>
+</section>
+{% endmacro %}
+
+{% macro group_users(title, users, creator) %}
+<section class="search-result-sidebar__section">
+  <h3 class="search-result-sidebar__subtitle">
+    {% trans %} {{title}} {% endtrans %}
+    <span class="search-result-sidebar-title__annotations-count">
+      {{ users|length }}
+    <span>
+  </h3>
+  <ul>
+    {% for user in users %}
+      <li>
+        <button type="submit"
+                form="search-bar"
+                name="toggle_user_facet"
+                value="{{ user.userid }}"
+                {% if user.faceted_by %}
+                  title="{% trans username=user.username %}Remove {{ username }} from the search query{% endtrans %}"
+                {% else %}
+                  title="{% trans username=user.username %}Limit the search to annotations by {{ username }}{% endtrans %}"
+                {% endif %}
+                class="search-result-sidebar__username">
+          {{ user.username }}
+          <span class="search-result-sidebar__annotations-count">
+            {{ user.count }}
+          </span>
+          {% if user.userid == creator %}
+            <span class="search-result-sidebar__annotations-count">
+              creator
+            </span>
+          {% endif %}
+        </button>
+      </li>
+      {% endfor %}
+    </ul>
+  </form>
 </section>
 {% endmacro %}
 
@@ -200,12 +238,10 @@
   {% call sidebar(group.name, group.description) %}
     <section class="search-result-sidebar__section">
       <dl>
-        {% if stats.annotation_count  %}
-          <dt class="search-result-sidebar__dt">
-            {% trans %}Shared annotations:{% endtrans %}
-          </dt>
-          <dd class="search-result-sidebar__dd">{{ stats.annotation_count }}</dd>
-        {% endif %}
+        <dt class="search-result-sidebar__dt">
+          {% trans %}Annotations:{% endtrans %}
+        </dt>
+        <dd class="search-result-sidebar__dd">{{ stats.annotation_count }}</dd>
         <dt class="search-result-sidebar__dt">
           {% trans %}Created:{% endtrans %}
         </dt>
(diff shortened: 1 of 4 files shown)


h/views/activity.py
```diff
@@ -142,8 +160,9 @@ def user_annotation_count(aggregation, userid):
                                           pubid=self.group.pubid,
                                           slug=self.group.slug),
             'members': members,
+            'moderators': moderators,
+            'creator': self.group.creator.userid,
```
robertknight: `Group.creator` can be `None`, so you need to handle that case here.

h/templates/activity/search.html.jinja2
```diff
@@ -173,6 +173,44 @@
...
+                  title="{% trans username=user.username %}Limit the search to annotations by {{ username }}{% endtrans %}"
+                {% endif %}
+                class="search-result-sidebar__username">
+          {{ user.username }}
+          <span class="search-result-sidebar__annotations-count">
+            {{ user.count }}
+          </span>
+          {% if user.userid == creator %}
+            <span class="search-result-sidebar__annotations-count">
+              creator
```
robertknight: In the mock, the annotation count and "creator" label use different fonts to differentiate them:

screenshot 2018-02-13 12 08 37

In the implementation, the two have the same font and wasn't immediately clear to me that "creator" was a separate label:

screenshot 2018-02-13 12 08 08

I would rather not change the style of the annotation count in this PR so let's leave as-is but I think we'll want to change this.

h/panels/navbar.py
```diff
@@ -29,14 +30,20 @@ def navbar(context, request, search=None, opts=None):
...
-            groups_suggestions.append({
-                'name': group.name,
-                'pubid': group.pubid
-            })
         user_activity_url = request.route_url('activity.user_search',
                                               username=request.user.username)
         username = request.user.username
 
+    # make all groups visible in the search auto complete
+    groups = profile_groups_factory(None, request).all(request.user, request.authority)
```
robertknight: Can you pull the changes related to group suggestions into a separate PR? There are some additional issues that we'll need to resolve. For example, the autosuggest list currently displays all groups under a "Your groups" heading when you type "group:" which doesn't make sense any more if this list includes open groups.

screenshot 2018-02-13 12 16 45

h/panels/navbar.py
```diff
@@ -29,14 +30,20 @@ def navbar(context, request, search=None, opts=None):
...
+    for group in groups:
+        # don't return __world__ group
+        if group['id'] == '__world__':
```
robertknight: I'm not sure that we should exclude the Public group from this list, but let's discuss that once the autosuggest stuff has been extracted into a separate PR.

h/templates/activity/search.html.jinja2
```diff
@@ -143,7 +143,7 @@
 {% macro tags_section() %}
 <section class="search-result-sidebar__section">
   <h3 class="search-result-sidebar__subtitle">
-    {% trans %}Top tags{% endtrans %}
+    {% trans %}Tags{% endtrans %}
```
robertknight: I see this change was part of the design, but there is some context of which @dawariley may not have been aware. The tag list does not show _all_ tags but currently only the top N tags (ie. most frequently used in annotations in the group), so "Top tags" is a more accurate description.

I would prefer that we leave this change out of this PR and address separately.

h/templates/activity/search.html.jinja2
```diff
@@ -200,12 +238,10 @@
   {% call sidebar(group.name, group.description) %}
     <section class="search-result-sidebar__section">
       <dl>
-        {% if stats.annotation_count  %}
-          <dt class="search-result-sidebar__dt">
-            {% trans %}Shared annotations:{% endtrans %}
```
robertknight: I'm guessing the "Shared annotations" text previously used was chosen to indicate that the count did not include private annotations. I'm not sure if that was in fact even true in the first place - I think here what we're actually showing is "Shared annotations by other people plus your private annotations". Calling it just "Annotations" seems cleaner.

h/views/activity.py
```diff
@@ -106,7 +106,9 @@ def search(self):
 
         result['opts'] = {'search_groupname': self.group.name}
 
-        if self.request.user not in self.group.members:
+        # if the group has a concept of members (aka joinable_by is not None)
+        # and the user is not in the list of members
```
robertknight: On a small stylistic note, the general convention in our codebase is to use full sentences for docstrings and comments (`# If the group ... list of members.`)

tests/h/views/activity_test.py
```diff
@@ -108,6 +108,467 @@ def pyramid_request(self, factories, pyramid_request):
         return pyramid_request
 
 
+@pytest.mark.usefixtures('annotation_stats_service', 'group_service', 'routes', 'search')
+class TestOpenGroupSearchController(object):
```
robertknight: The test classes are named after the class or function they test, so naming this `TestOpenGroupSearchController` could make someone wonder where `OpenGroupSearchController` is in the code. There are also quite a lot of duplicated tests here which I don't see the need for.

As long as we are using a single view class for handling open groups and regular groups with just a few special cases for open groups, then I would suggest adding tests as new methods to the existing `TestGroupSearchController` class.

h/templates/activity/search.html.jinja2
hmstepanek: Ah, ok I was looking at the code and didn't see where it was limiting the tags?

robertknight: Oh, don't worry, it is completely non-obvious! The tags come from an Elasticsearch feature called [aggregations](https://www.elastic.co/guide/en/elasticsearch/reference/1.5/search-aggregations.html) and are configured [here](https://github.com/hypothesis/h/blob/db7f44e1432ef2c065d97aa7a2c4b9483c3d2bdd/h/activity/query.py#L145).

h/templates/activity/search.html.jinja2
robertknight: Ah, OK, that's fine then.

hmstepanek:
Updated, PTAL. @robertknight

tests/h/views/activity_test.py
```diff
@@ -113,7 +116,8 @@ class TestGroupSearchController(object):
 
     """Tests unique to GroupSearchController."""
 
-    def test_renders_join_template_when_no_read_permission(self, controller, pyramid_request, group):
+    @pytest.mark.parametrize('test_group', GROUP_TYPE_OPTIONS, indirect=['test_group'])
```
robertknight: Ah, interesting. I hadn't come across the `indirect` facility in pytest before. From a quick skim through existing uses of `pytest.mark.parametrize` it looks like we just haven't needed to use fixtures involving DB objects in parametrized tests before?

tests/h/views/activity_test.py
```diff
@@ -127,36 +131,37 @@ def fake_has_permission(permission, context=None):
...
             if permission in ('read', 'join'):
                 return False
-        pyramid_request.user = factories.User()  # User logged in
         pyramid_request.has_permission = mock.Mock(side_effect=fake_has_permission)
 
         with pytest.raises(httpexceptions.HTTPNotFound):
             controller.search()
 
+    @pytest.mark.parametrize('test_group', GROUP_TYPE_OPTIONS, indirect=['test_group'])
     def test_search_redirects_if_slug_wrong(self,
```
robertknight: Hmm, there is some overhead for each test that runs, so I would be inclined to be slightly cautious about parametrising every test in this class to use each of the group types. I leave this up to you.

tests/h/views/activity_test.py
```diff
@@ -917,14 +1097,58 @@ def group(factories):
...
     group = factories.Group()
 
     group.members.extend([factories.User(), factories.User()])
     return group
 
 
+@pytest.fixture
+def no_creator_group(factories):
+    # Create some other groups as well, just to make sure it gets the right
+    # one from the db.
```
robertknight: I'm not quite following the logic here. Can you clarify?

hmstepanek:
Ok, I think I addressed all the issues here (including commit history cleanup). @robertknight Can you take a look one more time?

hmstepanek:
It occurred to me that I should have moved and squashed the Top tags commit so I’ll do that tomorrow.

tests/h/views/activity_test.py
```diff
@@ -917,14 +1096,48 @@ def group(factories):
...
+    group.creator = None
+    group.members.extend([factories.User(), factories.User()])
+    return group
+
+
+@pytest.fixture
+def open_group(factories):
+    # Create some other groups as well, just to make sure it gets the right
+    # one from the db.
+    factories.Group()
```
robertknight: I'm still not clear on the purpose of creating these additional groups. Can you clarify?

PR #4783, opened 2018-02-08
Attempt to fix an intermittently failing test

seanh:
`tests/h/search/index_test.py::TestBatchIndexer::test_index_indexes_all_annotations_to_es`
is occassionally failing with this assertion error:

    E AssertionError:
    Expected call: streaming_bulk(
        <Mock name='mock.conn' id='140410258777232'>,
        <iterable with [<Annotation 3l0CPAzaEei6VmNdOzSvCg>, <Annotation 3mUY-gzaEei6Vl97TJaBfg>]>,
        chunk_size=<ANY>,
        expand_action_callback=<ANY>,
        raise_on_error=False)
    Actual call: streaming_bulk(
        <Mock name='mock.conn' id='140410258777232'>,
        <generator object _log_status at 0x7fb3cf345be0>,
        chunk_size=100,
        expand_action_callback=<bound method BatchIndexer._prepare of <h.search.index.BatchIndexer object at 0x7fb3cf284a10>>,
        raise_on_error=False)

The expected call looks the same as the actual call, assuming that
`<iterable with [<Annotation 3l0CPAzaEei6VmNdOzSvCg>, <Annotation 3mUY-gzaEei6Vl97TJaBfg>]>`
matches `<generator object _log_status at 0x7fb3cf345be0>`, which it
usually does (the test usually passes).

My hunch is that the iterable sometimes fails to match the generator
because the order in which the generator returns the two annotations is
sometimes different (the generator boils down to a DB query without an
explicit ORDER_BY).

Replacing the `iterable_with` matcher with one that matches regardless
of order should fix that.

(P.S. I don't think we care what order the annotations are indexed in,
so the production code is correct, the test is wrong.)

### DIFF
--- a/tests/h/search/index_test.py
+++ b/tests/h/search/index_test.py
@@ -98,18 +98,9 @@ class TestBatchIndexer(object):
 
         indexer.index()
 
-        assert streaming_bulk.call_count == 1
-
-        args = streaming_bulk.call_args[0]
-        kwargs = streaming_bulk.call_args[1]
-
-        assert len(args) == 2
-        assert args[0] == indexer.es_client.conn
-        assert list(args[1]) == [ann_1, ann_2]
-
-        assert kwargs.keys() == matchers.unordered_list((
-            'chunk_size', 'raise_on_error', 'expand_action_callback'))
-        assert kwargs['raise_on_error'] is False
+        streaming_bulk.assert_called_once_with(
+            indexer.es_client.conn, matchers.iterable_with(matchers.unordered_list([ann_1, ann_2])),
+            chunk_size=mock.ANY, raise_on_error=False, expand_action_callback=mock.ANY)
 
     def test_index_skips_deleted_annotations_when_indexing_all(self, db_session, indexer, matchers, streaming_bulk, factories):
         ann_1, ann_2 = factories.Annotation(), factories.Annotation()


tests/h/search/index_test.py
```diff
@@ -99,7 +99,7 @@ def test_index_indexes_all_annotations_to_es(self, db_session, indexer, matchers
         indexer.index()
 
         streaming_bulk.assert_called_once_with(
-            indexer.es_client.conn, matchers.iterable_with([ann_1, ann_2]),
+            indexer.es_client.conn, matchers.iterable_with(matchers.unordered_list([ann_1, ann_2])),
```
seanh: The fix here relies on the fact that `iterable_with()` does `list(other) == self.items`. That is, it turns the generator that `streaming_bulk()` was called with (`other`) into a list (my theory is this list will sometimes be `[ann_1, ann_2]` and sometimes come out as `[ann_2, ann_1]`) and then compares that list to `self.items` which is `matchers.unordered_list([ann_1, ann_2])`. This comparison will be `True` regardless of the order of the lists.

https://github.com/hypothesis/h/blob/66f83ba23bc657a440e4a341c9593023be85cba5/tests/common/matchers.py#L65-L75

seanh:
Closing this in favour of https://github.com/hypothesis/h/pull/4784

seanh:
@robertknight I've reopened this flakey test fix following confirmation here https://github.com/hypothesis/h/pull/4784#issuecomment-364960922

seanh:
Since this was already approved previously I'm just going to merge it

PR #4786, opened 2018-02-09
Run unit tests under Python 3

robertknight:
The "h" codebase is supposed to be Python 3 compatible / ready for a migration to Python 3. We've been continuing to use Python 2.7 since we were blocked waiting for dependencies (eg. gunicorn) to migrate. That has now happened.

However we haven't actually been running unit tests under Python 3 so naturally various failures have crept in over time.

This PR:

- Adds a Travis job which runs our unit tests (not functional yet) under Python 3 and compares the actual failures to a whitelist of current known failures. This will stop new failures from creeping in and provide a way to incrementally fix the existing failures.
- Fixes a small set of existing failures which interfere with pytest's test collection.


seanh:
Will the comparison to the whitelist break whenever a test is deleted or added before the end of the file? Changing the line num of one of the whitelisted fails. Maybe the whitelist should be based on test name instead of line number?

seanh:
This is great! I wonder, do we need to update the versions of any of our dependencies before we can go Python 3? We seem to be getting the tests to run so I guess maybe not, but it might be worth a thorough look over. I don't think pyup is working correctly on h so I think some of our deps might be quite out of date

robertknight:
> Will the comparison to the whitelist break whenever a test is deleted or added before the end of the file?

Hmm, yes indeed they will. This might not be a problem since I have a set of PRs lined up to fix the remaining issues so I'm hoping this won't live for very long. However, ignoring line numbers in the diff would resolve the issue. I'm happy to go either way on this.

> I wonder, do we need to update the versions of any of our dependencies before we can go Python 3?

I got all the unit and functional tests passing locally and the app itself appeared to function fine with no dependency upgrades, with one exception. That exception was that tasks dispatched to workers did not run. It might be the case that a Celery update is required.

seanh:
According to pyup deform-jinja2 and psycogreen aren't Python 3 compatible

PR #4787, opened 2018-02-09
Add new GroupScope model class

seanh:


h/models/group.py
```diff
@@ -80,6 +80,8 @@ class Group(Base, mixins.Timestamps):
         'User', secondary='user_group', backref=sa.orm.backref(
             'groups', order_by='Group.name'))
 
+    scopes = sa.orm.relationship('GroupScope', backref='group', cascade='all, delete-orphan')
```
seanh: The `cascade` argument here is what ensures that if a group is deleted, or a group's scopes are changed so that some groupscopes are no longer referenced, the groupscopes automatically get deleted.

seanh: (I think this is just sqlalchemy stuff though, it doesn't produce an `ON DELETE CASCADE` in SQL, so deleting a group in SQL may not delete groupscopes, will look into this.)

h/models/group_scope.py
```diff
@@ -0,0 +1,26 @@
...
+
+    #: A web origin as defined by the ``origin`` property of the URL Web API:
+    #: https://developer.mozilla.org/en-US/docs/Web/API/URL. This includes the
+    #: scheme, domain (including subdomains) and port parts of a URL. For
+    #: example:
+    #:
+    #: http://example.com
+    #: https://web.hypothes.is
+    #: http://localhost:5000
+    origin = sa.Column('origin', sa.UnicodeText, nullable=False)
```
seanh: Notice that `origin` is no longer unique.

tests/common/factories/group_scope.py
```diff
@@ -0,0 +1,18 @@
...
+
+from .base import ModelFactory
+from .group import OpenGroup
+
+
+class GroupScope(ModelFactory):
+    class Meta:
+        model = models.GroupScope
+
+    origin = factory.Faker('url')
```
seanh: `Faker('url')` always produces origins (e.g. `https://www.example.com`) never with a path. Also never a port.

tests/h/models/group_scope_test.py
```diff
@@ -0,0 +1,129 @@
...
+        factories.GroupScope(origin='http://www.example.com')
+        db_session.flush()
+
+    def test_port_is_allowed_in_origin(self, db_session, factories):
+        factories.GroupScope(origin='http://localhost:5000')
+        db_session.flush()
+
+    def test_there_is_no_validation_of_origin(self, db_session, factories):
+        factories.GroupScope(origin='diplodocus : 123')
+        db_session.flush()
```
seanh: Maybe we'll want to add some validation, maybe even at the model level, at some point.

h/models/group.py
robertknight: > I think this is just sqlalchemy stuff though, it doesn't produce an ON DELETE CASCADE in SQL, so deleting a group in SQL may not delete groupscopes, will look into this

I would be interested to learn how this works, please post your findings.

h/models/group_scope.py
```diff
@@ -0,0 +1,26 @@
...
+
+from __future__ import unicode_literals
+
+import sqlalchemy as sa
+
+from h.db import Base
+
+
+class GroupScope(Base):
+    __tablename__ = 'groupscope'
```
robertknight: Would be useful to have a docstring here explaining what a `GroupScope` is for the benefit of newcomers to the codebase.

tests/h/models/group_scope_test.py
```diff
@@ -0,0 +1,129 @@
...
+
+from __future__ import unicode_literals
+
+from sqlalchemy import inspect
+
+from h.models import GroupScope
+
+
+class TestGroupScope(object):
+    def test_save_and_retrieve_origin(self, db_session, factories):
```
robertknight: Hmm, some of these tests feel a bit like testing SQLAlchemy. I suppose there is still value because it checks for behaviour changes when eg. we upgrade SQLA or someone modifies the declarative configuration.

h/models/group.py
seanh: Ok, I've added `nullable=False` to the `group_id` column of `groupscope` to prevent orphan `groupscope` rows. And I've added `ondelete='cascade'` to the `ForeignKey` in `GroupScope` to enable convenient deletion of groups when working in SQL directly. Details:

The `cascade` argument to `relationship()` is documented here: <http://docs.sqlalchemy.org/en/rel_1_1/orm/cascades.html>.

Normally, without `cascade='all, delete-orphan'` on the relationship, you'd have two problems:

1. Deleting a group (`db_session.delete(group)`) causes an `IntegrityError` from Postgres because there are rows in the `groupscope` table with foreign keys that reference the deleted group. You would have to manually delete the groupscopes first.

2. Removing the scopes from a group (`group.scopes = []`) also causes an `IntegrityError` from Postgres because sqlalchemy tries to set the `group_id` column of these `groupscope` rows to `NULL` which violates the `NOT NULL` constraint on `groupscope.group_id`. (Or, if there were no such `NOT NULL` constraint, then you wouldn't get an error but you would get orphan `groupscope` rows with `NULL` `group_id`s left in the DB.) Again, you would have to manually delete the groupscopes first.

With the `cascade='all, delete-orphan'` though, sqlalchemy changes its behaviour so that either of the above would automatically delete the `groupscope` rows for you.

This `cascade` is a sqlalchemy feature (implemented by sqlalchemy keeping track of things itself and emitting additional SQL `DELETE` commands), it makes no difference to the table definition that ends up in Postgres.

If you tried to delete a group directly in SQL you would get an error because you need to delete the groupscopes that're referencing it first. I've now added an `ondelete='cascade'` argument to the `ForeignKey` in the `GroupScope` table though. _That_ does mean that `ON DELETE CASCADE` gets added to the `FOREIGN KEY` constraint on the `groupscope` table in the SQL, so now you _can_ just delete a group in SQL and the corresponding `groupscopes` are automatically deleted.

Because `groupscope.group_id` is a non-nullable foreign key, in all cases sqlalchemy or sql, even without either of the cascade arguments, it's never possible to have orphan `groupscope` rows in the DB. What the cascades do is automatically delete groupscopes instead of crashing, if you just try to delete a group or edit group.scopes without deleting the scopes yourself first.

tests/h/models/group_scope_test.py
seanh: Yeah. And to be fair - my first attempt at doing this class had totally unexpected and unwanted behaviours! Even though sqlalchemy is doing the actual work, I think it's worth having tests that I've used sqlalchemy as I think I've used it.

h/models/group_scope.py
seanh: Good idea added

tests/h/models/group_scope_test.py
seanh: (Another way to put this is that I'm not testing sqlalchemy, I'm testing the models that I've coded using sqlalchemy. Taking it to the extreme you could say that any tests of Python code are just "testing Python" :)

h/models/group.py
robertknight: I follow. Thanks for the explanation 🙂

PR #4788, opened 2018-02-09
Add DB migration to add groupscope table

seanh:
This is the DB migration for the new model class in https://github.com/hypothesis/h/pull/4787


seanh:
I'll wait until Monday before merging and running this

PR #4795, opened 2018-02-13
Enable all updates in pyup

seanh:
Have the pyup bot send us pull requests for all dependency updates, not just for known security updates.


robertknight:
I suspect that enabling this for all dependencies is going to drown us in updates. I also know that we have at least some major dependencies where the update will require some work. eg. Pyramid 1.9 changed a few things with how CSRF handling works.

Would it be possible to scope this to just a few dependencies to see how we get on?

seanh:
I think if there are specific packages that we can't upgrade for particular reasons we should try blacklisting those using [`# pyup: ignore` comments](https://pyup.io/docs/bot/filter/) but otherwise we should let pyup do a small amount of work regularly to keep as many of our dependencies as possible up to date.

seanh:
Gonna close this and try out https://dependabot.com/ on bouncer instead. It does JavaScript as well as Python and looks maybe a bit better. Other dependency bots I found: requirements.io (Python only) and https://snyk.io/ (multiple languages)

PR #4796, opened 2018-02-13
Py 3 compat - Use `str(exc)` instead of `exc.message`

robertknight:
`BaseException.message` was deprecated in Python 2.6 and in Python 3.x many exception types do not have a `message` attribute any more. Instead use `str(exc)` as a consistent way to get the human readable exception description.


robertknight:
Rebased

PR #4797, opened 2018-02-13
Fix Python 3-incompatible dict usage

robertknight:
**Depends on https://github.com/hypothesis/h/pull/4786**

- `iteritems` does not exist in Python 3
- `dict.keys()` does not return a `list` in Python 3 but a `dict_keys` object. This object cannot be indexed and cannot be compared with lists for value equality.


robertknight:
The tox-py3 build was killed because Travis was busy and it took longer than 10mins to run the tests. The way the `make test-py3` command is currently implemented, it produces no output during the test run and so Travis thinks the command has stalled. I might have to change how the command works to produce some output as it runs.

PR #4802, opened 2018-02-13
Python 3 string fixes

robertknight:
This PR fixes several issues with conversions or comparisons between binary data (`bytes`) and unicode strings:

- When using the `bytes` function to convert strings to binary data, the second encoding argument must be specified. Alternatively for Python 2 + 3 compatibility, `str.encode(<encoding>)` can be used.
- The Unicode text type is `unicode` in Python 2 and `str` in Python 3. Use the `h._compat.text_type` alias where a direct reference to this type is needed.
- Response bodies are binary data so must be compared against `b'...'` literals instead of Unicode string literals


robertknight:
Mea culpa, I failed to run these changes against Python 2. Looking into it...

seanh:
Rebased

PR #4803, opened 2018-02-13
Display tox output when running tests under Python 3

robertknight:
When Travis is busy the tox tests can take longer than 10 minutes to
execute under Python 3. When this happened, Travis was seeing no output
from the command and killing the job.


robertknight:
Hmm, Py 3 tests are currently failing with:

```
tee: standard output: Resource temporarily unavailable
tee: write error
make: *** [test-py3] Error 1
```

I will dig into this tomorrow.

robertknight:
The Travis issue should be resolved now. See the last commit for details.

PR #4811, opened 2018-02-15
Fix Py 3 compat issues in h.search.config

robertknight:
- Make `_random_id` always return a `str` rather than `bytes` in Python 3.

 - Use compat import for `quote_plus`


seanh:
@robertknight needs rebase

PR #4812, opened 2018-02-15
Upgrade alembic to latest version

seanh:


robertknight:
Changelog here: http://alembic.zzzcomputing.com/en/latest/changelog.html

I haven't looked over every change in detail, I just had a look for mentions of breaking changes in 0.8.7 - 0.9.7. I didn't spot any.

PR #4814, opened 2018-02-15
Fix Python 3 compat issues with annotation ID conversion helpers

robertknight:
Annotation IDs are stored as UUIDs in Postgres ("7b6d8026-51d8-40d0-86ef-46d72fb82bd7") but represented in the API in a "URL-safe base 64 encoded" form ("e22AJlHYQNCG70bXL7gr1w"). This PR fixes some Py 3 compatibility issues with the conversion functions:

 - `_get_urlsafe_from_hex` and `_get_hex_from_urlsafe` were returning
   `bytes` in Py 3 instead of Unicode strings as other code expected.

 - Generate a more helpful exception message if these functions are given
   the wrong type of input in Python 3.

 - Support the `binascii.Error` exception raised in Py 3 when input is
   not valid base 64.

### DIFF
--- a/h/db/types.py
+++ b/h/db/types.py
@@ -140,9 +140,6 @@ def _get_urlsafe_from_hex(value):
     :rtype: unicode
     """
 
-    if not isinstance(value, string_types):
-        raise ValueError('`value` is {}, expected one of {}'.format(type(value), string_types))
-
     # Validate and normalise hex string
     hexstring = uuid.UUID(hex=value).hex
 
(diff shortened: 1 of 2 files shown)


robertknight:
>  I wonder if we need unit tests for the new behaviours on these ones?

Ostensibly these are private helpers registered with SQLAlchemy for use converting to/from the database in the `Annotation.id` function. The case in `_get_hex_from_urlsafe` is tested by `test_uuid_serialize_non_base64_data`. The test in `_get_urlsafe_from_hex` turned out to be unnecessary upon a second inspection because the function is only called in one place and the input has to be a string there.

PR #4816, opened 2018-02-19
Py 3 compat - Fix encoding issue in SVG icon renderer

robertknight:
`ElementTree.tostring` returns `bytes` not `str` by default in Python 3,
causing unexpected "b" strings to appear in the rendered output after
Jinja converts the result to a string.

`ElementTree.tostring` accepts an `encoding="unicode"` arg to produce a
Unicode string directly, but this is not supported in Python 2.7 so
`decode` the byte string instead.


seanh:
Needs rebase

PR #4821, opened 2018-02-22
Py 3 compat - Fix text handling issues in JSON schema tests

robertknight:
Using bytes keys in the test schema and string fields in the test case
JSON input caused validation to fail in Python 3.

For Python 3, use unicode keys in the test schema and JSON input to
resolve the issue. Using Unicode literals for property names in Python 2
results in different exception messages (`u'foo'` vs `'foo'`). Work
around this by using `bytes` as the type for property names in Py 2.

### DIFF
--- a/tests/h/schemas/base_test.py
+++ b/tests/h/schemas/base_test.py
@@ -2,6 +2,7 @@
 
 from __future__ import unicode_literals
 
+from h._compat import PY2
 import enum
 from mock import Mock
 import pytest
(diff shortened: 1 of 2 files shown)


seanh:
Is the `from __future__ import unicode_literals` at the top of `base_test.py` really required? I know we should have them at the top of most Python files for Python 2 /3 compatibility, but in the case of this particular file if you remove it and then removing the leading `b`s on strings then those strings will be byte strings in Python 2 but unicode strings in 3 and the tests will pass in both without the need for the `prop_name_type` thing.

By the way, with the code as it currently is on master I don't think validation _fails_ (as in claims the data is invalid) in Python 3, rather it succeeds (lets the data pass as valid) when it _should_ fail. I'm not sure exactly why this happens / what's going on here?

robertknight:
> Is the from __future__ import unicode_literals at the top of base_test.py really required?

I hadn't thought of this option but IMO it could be confusing to do that since it would make this test file differ from all of our other tests and breaks the general assumption in our code that if you see a string literal, it is a unicode string.

> By the way, with the code as it currently is on master I don't think validation fails (as in claims the data is invalid) in Python 3, rather it succeeds (lets the data pass as valid) when it should fail.

That is happening because internally jsonschema looks for a "required" _string_ key but the example schema has a "required" _bytes_ key. Therefore it thinks that the schema has no required properties. If you remove the `b` prefix from the `b'required'` literal, the tests fail under Python 3 with errors about missing properties.

PR #4825, opened 2018-02-23
Py 3 compat - Fix MagicMock <=> int comparison error

robertknight:
Fix an error when code tried to compare the result of
`Document.find_or_create_by_uris(...).count()` to an int literal.

Previously `count()` returned a MagicMock, now it returns a hardcoded
int, which is Good Enough™ for the current tests.

### DIFF
--- a/tests/h/models/document_test.py
+++ b/tests/h/models/document_test.py
@@ -961,7 +961,9 @@ class TestUpdateDocumentMetadata(object):
 
     @pytest.fixture
     def Document(self, patch):
-        return patch('h.models.document.Document')
+        Document = patch('h.models.document.Document')
+        Document.find_or_create_by_uris.return_value.count.return_value = 1
+        return Document
 
     @pytest.fixture
     def merge_documents(self, patch):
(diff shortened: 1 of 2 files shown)


seanh:
So this is due to a change in the behaviour or the `>` operator (it now refuses to compare a `MagicMock` to an `int`) in Python 3?

robertknight:
> So this is due to a change in the behaviour or the > operator (it now refuses to compare a MagicMock to an int) in Python 3?

Yes, exactly.

```
Python 2.7.14 (default, Jan  3 2018, 12:57:16) 
[GCC 4.2.1 Compatible Apple LLVM 9.0.0 (clang-900.0.39.2)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import mock
>>> mock.MagicMock() < 1
False
>>> 
```

```
Python 3.6.4 (default, Jan  6 2018, 11:51:59) 
[GCC 4.2.1 Compatible Apple LLVM 9.0.0 (clang-900.0.39.2)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import mock
>>> mock.MagicMock() < 1
  File "<stdin>", line 1, in <module>
TypeError: '<' not supported between instances of 'MagicMock' and 'int'
>>> 
```

PR #4829, opened 2018-02-26
Add day to date on activity page sidebar

hmstepanek:
Add day to group created date on sidebar so date format is now:
Month day, year aka: August 1, 2018. Some tests were updated to
expect this new date format. Note the date format was changed from
using the c library strftime to using the python format method as
the c library does not support %-d on all systems (a way of removing
the default zero padding of the day). This is a fix for 
https://github.com/hypothesis/client/issues/644.

### DIFF
--- a/h/views/activity.py
+++ b/h/views/activity.py
@@ -19,6 +19,7 @@ from h.paginator import paginate
 from h.search import parser
 from h.util.user import split_user
 from h.views.groups import check_slug
+from h.util.datetime import utc_us_style_date
 
 
 PAGE_SIZE = 200
@@ -154,7 +155,7 @@ class GroupSearchController(SearchController):
             'annotation_count': group_annotation_count,
         }
         result['group'] = {
-            'created': self.group.created.strftime('%B, %Y'),
+            'created': utc_us_style_date(self.group.created),
             'description': self.group.description,
             'name': self.group.name,
             'pubid': self.group.pubid,
@@ -321,7 +322,7 @@ class UserSearchController(SearchController):
         result['user'] = {
             'name': self.user.display_name or self.user.username,
             'description': self.user.description,
-            'registered_date': self.user.registered_date.strftime('%B, %Y'),
+            'registered_date': utc_us_style_date(self.user.registered_date),
             'location': self.user.location,
             'uri': self.user.uri,
             'domain': domain(self.user),
--- a/tests/h/util/datetime_test.py
+++ b/tests/h/util/datetime_test.py
@@ -4,7 +4,7 @@ from __future__ import unicode_literals
 
 import datetime
 
-from h.util.datetime import utc_iso8601
+from h.util.datetime import utc_iso8601, utc_us_style_date
 
 
 class Berlin(datetime.tzinfo):
@@ -28,3 +28,8 @@ def test_utc_iso8601():
 def test_utc_iso8601_ignores_timezone():
     t = datetime.datetime(2016, 2, 24, 18, 3, 25, 7685, Berlin())
     assert utc_iso8601(t) == '2016-02-24T18:03:25.007685+00:00'
+
+
+def test_utc_us_style_date():
+    t = datetime.datetime(2016, 2, 4)
+    assert utc_us_style_date(t) == "February 4, 2016"
--- a/h/util/datetime.py
+++ b/h/util/datetime.py
@@ -6,3 +6,8 @@
 def utc_iso8601(datetime):
     """Convert a UTC datetime into an ISO8601 timestamp string."""
     return datetime.strftime('%Y-%m-%dT%H:%M:%S.%f+00:00')
+
+
+def utc_us_style_date(datetime):
+    """Convert a UTC datetime into a Month day, year (August 1, 1990)"""
+    return "{d:%B} {d.day}, {d:%Y}".format(d=datetime)
--- a/tests/h/views/activity_test.py
+++ b/tests/h/views/activity_test.py
@@ -238,7 +238,7 @@ class TestGroupSearchController(object):
                                                                  pyramid_request):
         group_info = controller.search()['group']
 
-        assert group_info['created'] == test_group.created.strftime('%B, %Y')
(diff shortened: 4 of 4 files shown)


hmstepanek:
In response to @robertknight earlier comment I didn't really want to get into the whole locale thing right now-was kinda hoping we could table that for the future? I didn't mean to open a can of worms here. My thinking was the date is already not guaranteed to be in the user's time zone and it's not like the user isn't able to distinguish between what the month/day/year is since the month is spelled out so no harm? I agree with what you are saying about the locale thing but I was hoping that could be a separate issue for the future?

dawariley:
From a user's point of view I would agree that having the month/day/year US date format isn't a deal breaker for now and we should proceed with that and build out the locale function in the future.

When I initially added this requirement to https://github.com/hypothesis/client/issues/644, it came from a request from Dan and I raised the location specific date format and we both agreed that it wouldn't be a big deal for v1.

seanh:
Before:

before-group

before-user

After:

after-group

after-user

h/views/activity.py
```diff
@@ -24,6 +24,10 @@
 PAGE_SIZE = 200
 
 
+def _format_date(date):
+    return "{d:%B} {d.day}, {d:%Y}".format(d=date)
```
seanh: I've never seen this way of formatting dates in Python! Very interesting. So `{d:%B}` produces numbers like `28` and `1` (as opposed to `01`)? And there's no way to do that with `strftime()`?

Anyway, I think you should mover `_format_date()` into a utility module, maybe [h.util.datetime](https://github.com/hypothesis/h/blob/master/h/util/datetime.py), and add some unit tests to it. There's probably not _that_ much to test - but at least test that it works with both double- and single-digit days, and with both date and datetime objects. If nothing else this will prevent someone from coming along and replacing with with a more typical usage of `strftime()` later, and not realising what they're breaking.

robertknight:
> My thinking was the date is already not guaranteed to be in the user's time zone and it's not like the user isn't able to distinguish between what the month/day/year is since the month is spelled out so no harm?

Providing that there is a reasonably low probability of the changes confusing non-US/UK users compared to the existing formatting then I would say that we could revisit i18n issues later.

I'll give an example of where blinding using US formatting would be a mistake though: Supposing the date was currently written as YYYY-MM-DD and you changed it to MM/DD/YYYY which is common in the US. This could be confusing in other parts of the world (eg. the UK) where the date is usually written DD/MM/YYYY. As you say, we are spelling out the month name in full here, so that doesn't apply.

In the context of number formatting, a comma is used as a _thousands_ separator in the US/UK but it is used as a _decimal_ separator elsewhere, so that gave me pause for thought.

h/views/activity.py
robertknight: Heh, I wasn't aware of this either. Any pros/cons compared to `strftime` we should be aware of?

hmstepanek: I added a comment about this in the commit message. We can't use strftime to produce a non zero padded day as we would use %-d and that isn't supported on all systems. Note strftime calls into the c library of the os and %-d is a 1999 standard where as apparently our qa system (not sure about prod) is 1998 standard and doesn't support it.

robertknight: Thanks, that's interesting. Further discussion: https://hypothes-is.slack.com/archives/C4K6M7P5E/p1519755955000065

I verified that `%-d%` indeed doesn't work in an h Docker container, so this is more portable.

hmstepanek:
I would like to move that function into the util module per @seanh 's suggestion before merging this, if that's ok. I'll do that tomorrow AM-ran out of time today.

dawariley:
"In the context of number formatting, a comma is used as a thousands separator in the US/UK but it is used as a decimal separator elsewhere, so that gave me pause for thought."

Is it possible to be 'comma agnostic' and just leave it out altogether?

robertknight:
> Is it possible to be 'comma agnostic' and just leave it out altogether?

It was omitted previously, but grouping does make larger numbers easier to read so I'm OK with adding it.

There is a library we can use to format numbers with grouping according to the conventions of the user's language & region. So we could use a comma for now and localize it in follow-up work.

What is needed from a design perspective is a decision on whether to display numbers with group separators or not.

dawariley:
ok! sounds good. by group separators do you me '-' vs '/'.

I think the preference is to retain the 'natural language' version of: "August 1, 2018."

If I have your ask wrong.. you might need to define 'group separator' for me.

robertknight:
> ok! sounds good. by group separators do you me '-' vs '/' ?

My comment about group separators was referring to general _number_ rather than _date_ formatting as in Hannah's other PR, it doesn't apply here.

dawariley:
Let's follow the 'Google' standard: "periods as a thousands separator and commas for decimals", factoring in locale when we have time.

segdeha:
So, `10.000,00`? That would look broken to our U.S. users who make up most of our user base.

hmstepanek:
Well let’s be fair, there are no partial annotations so it would look like 10.000. We could also use spaces potentially.

robertknight:
> Let's follow the 'Google' standard: "periods as a thousands separator and commas for decimals", factoring in locale when we have time.

This discussion pertains to a different PR, let's not hash it out here. This PR is just about the date formatting change.

hmstepanek:
Ok I have moved the date format method into the util location that @seanh requested. Also changed the name to utc_us_style_date. If @seanh feels good about this change I'll go ahead and merge it?

tests/h/util/datetime_test.py
```diff
@@ -28,3 +28,7 @@ def test_utc_iso8601():
 def test_utc_iso8601_ignores_timezone():
     t = datetime.datetime(2016, 2, 24, 18, 3, 25, 7685, Berlin())
     assert utc_iso8601(t) == '2016-02-24T18:03:25.007685+00:00'
+
+def test_us_style_date():
```
robertknight: Nitpick - This should be `test_utc_us_style_date`.

robertknight:
Looks like there is a lint failure and I had a tiny suggestion regarding a test name change. Once those are resolved, I think you can go ahead and merge this.

hmstepanek:
oopsy...i fix.

PR #4830, opened 2018-02-26
Format annotation count on activity pages w/ ','

hmstepanek:
Add , as thousands delimiter in annotation count on activity
pages. The annotation count blends together when it's really
large and it makes it hard to read. This formatting was added
as a jinja extension and a corresponding test was also added.
This formats the search result Matching # annotations as well
as the group and user side bar total annotations to be ,
delimited.

### DIFF
--- a/h/jinja_extensions.py
+++ b/h/jinja_extensions.py
@@ -26,6 +26,7 @@ class Filters(Extension):
 
         environment.filters['to_json'] = to_json
         environment.filters['human_timestamp'] = human_timestamp
+        environment.filters['format_number'] = format_number
 
 
 def human_timestamp(timestamp, now=datetime.datetime.utcnow):
@@ -34,6 +35,10 @@ def human_timestamp(timestamp, now=datetime.datetime.utcnow):
     if timestamp.year < now().year:
         fmt = '%d %B %Y at %H:%M'
     return timestamp.strftime(fmt)
+
+
+def format_number(num):
+    return "{:,}".format(num)
 
 
 def to_json(value):
--- a/h/templates/activity/search.html.jinja2
+++ b/h/templates/activity/search.html.jinja2
@@ -164,7 +164,7 @@
         <span class="search-result-sidebar__tag-container">
           {{ tag.tag }}
           <span class="search-result-sidebar__tag-count">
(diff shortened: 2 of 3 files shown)


seanh:
Before:

before

After:

after

The same change also appears in the same location in the sidebar on _user_ home pages, as well as group ones.

@hmstepanek I wonder if this new number formatting should also apply to other numbers shown on activity pages? There's quite a few other places on the page where counts are shown.

seanh:
@hmstepanek Actually, can you confirm that the before and after screenshots I posted are all that's changed here? They were the only changes I could see

robertknight:
It might be worth making this a [Jinja filter](http://jinja.pocoo.org/docs/2.10/templates/#filters) used in the template as `{{ some_number | format_number }}` rather than modifying the context output returned by the view function. This will have two advantages:

- It will make writing tests that check the context returned by view functions, as they won't be affected by the presentation concern of formatting the number.
- When we come to add i18n support later, we'll only have to change one filter implementation and it should then work wherever in the templates we use this.

Filters are defined in [jinja_extensions.py](https://github.com/hypothesis/h/blob/master/h/jinja_extensions.py).

hmstepanek:
@seanh You are correct on both accounts. It just modifies the Annotations for groups and user activity pages in the sidebar. There are indeed other places that this change could be made as there are annotation count in other places on this page. Also I like @robertknight 's idea of making a filter. I think we should prob do something like that with the date PR as well.

hmstepanek:
@seanh @dawariley Updated to use jinja extension and now using '.' as delimiter.

hmstepanek:
Note after viewing on QA, I have added the '.' delimiting format to all counts on the page:
Before:
image


After:
image

robertknight:
I have to say I think the use of a period is going to be pretty confusing for English and probably other-language visitors since the rest of the page is in English. I'm not sure on where the suggestion of defaulting to periods comes from. I would suggest that we follow en_US conventions and use a comma unless someone can point to a documented & established convention for doing otherwise.

hmstepanek:
I agree with you @robertknight. I'll sync with @dawariley on this.

PR #4831, opened 2018-02-26
Add ... to indicate join link overflow in sidebar

hmstepanek:
Currently when the join link is too large to fit inside the input
box it cuts off the remaining text. Sometimes this happens in the
middle of a letter and is visually displeasing. Instead of cutting
off the join link url, now we add a ... on the end using elipsis
to indicate the url is too large to fit in the box. If the user
clicks on the box the whole link will be visible as normal.


seanh:
Before:

before

After:

after

seanh:
Looks much better

PR #4832, opened 2018-02-26
Reduce word spacing and date color in search

hmstepanek:
Got rid of extra spacing between words in "# Matching Annotations"
search results because it looks funny. ALso change color of date
to be lighter grey so that user focus is on the annotation and
not the date.

Before:
image

After:
image

### DIFF
--- a/h/static/styles/partials/_search.scss
+++ b/h/static/styles/partials/_search.scss
@@ -59,7 +59,6 @@
 .search-results__total {
   color: $brand;
   margin-bottom: 25px;
-  word-spacing: 4px;
 }
 
 .search-results__list {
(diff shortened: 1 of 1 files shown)


hmstepanek:
This was reviewed and requested by @dawariley

seanh:
@hmstepanek I wonder if you can add before and after screenshots? Then Dawa can review

h/static/styles/partials/_search.scss
```diff
@@ -95,6 +94,7 @@
 }
 
 .search-result__timeframe {
+  color: $grey-4;
```
robertknight: This change wasn't mentioned in the description. Is it intended?

hmstepanek: Good catch! I actually do need to make this change but kinda missed it when I did the diff so I update the commit message to reflect this change as well.

PR #4836, opened 2018-02-28
Fix remaining unit tests under Python 3

robertknight:
This PR contains the two remaining fixes required for all unit tests to pass under Python 3. See commit messages for details.

Additionally it simplifies the `test-py3` make target to remove the diff-ing of actual and expected failures since there are none any longer.

The next step after this will be to resolve issues with functional tests and get them running under Python 3 on Jenkins.

### DIFF
--- a/tests/h/form_test.py
+++ b/tests/h/form_test.py
@@ -164,7 +164,7 @@ class TestToXHRResponse(object):
 @pytest.mark.usefixtures('to_xhr_response')
 class TestHandleFormSubmission(object):
 
-    def test_it_calls_validate(self, pyramid_request):
+    def test_it_calls_validate(self, pyramid_request, matchers):
         form_ = mock.Mock(spec_set=['validate'])
 
         form.handle_form_submission(pyramid_request,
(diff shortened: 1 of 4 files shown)


tests/py3-expected-failures.txt
```diff
@@ -1 +0,0 @@
-/h/tests/h/schemas/annotation_test.py:197:
```
robertknight: Only one line is being removed here even though there were two separate fixes. This is because the `form_test.py` failure was missing due to the way the error list was extracted from the test output.

tests/h/form_test.py
```diff
@@ -164,15 +164,16 @@ def test_does_not_show_flash_message_if_xhr(self, pyramid_request):
...
         form_ = mock.Mock(spec_set=['validate'])
 
         form.handle_form_submission(pyramid_request,
                                     form_,
                                     mock_callable(),
                                     mock.sentinel.on_failure)
 
-        form_.validate.assert_called_once_with(pyramid_request.POST.items())
+        post_items = matchers.iterable_with(list(pyramid_request.POST.items()))
+        form_.validate.assert_called_once_with(post_items)
```
robertknight: Note that `POST.items()` returns an iterable but `matchers.iterable_with` requires a list rather than an iterable as an argument.

The `POST` dict is in fact empty in this test, which makes this not a great test, but I left it as-is and just fixed the compat issue.

PR #4840, opened 2018-02-28
UI tweaks for group sidebar and "Matching annotations" header

hmstepanek:
Some subtle changes requested by @dawariley  to what's currently in QA:
- increase font size of # Matching Annotations
- make font size match between Top Tags and Members/Moderators counts
- move share/join group input closer to description/instructional text.

Before:
image

After:
image


robertknight:
I edited the PR title to make it easier to see what changed recently when looking at the closed PR list in GitHub, which I sometimes do.

PR #4842, opened 2018-03-01
Update Celery and Kombu dependencies

robertknight:
Update Celery from 3.x to current 4.x release.

This is required to support the latest Python 3.x version. For changelog see http://celery.readthedocs.io/en/latest/whatsnew-4.0.html#upgrading-from-celery-3-1 . Note that we were already using 3.1.25 as highlighted in the migration notes.

The only breaking change I encountered was the removal of the `--autoreload` flag. I also applied an automated fix to `h/celery.py` to migrate setting names as documented in the changelog.

_Edit: Elsevier are also interested in this because I believe it should enable them to run h using Amazon SQS (which they already use) instead of setting up a separate RabbitMQ service._

### DIFF
--- a/h/cli/commands/devserver.py
+++ b/h/cli/commands/devserver.py
@@ -80,7 +80,7 @@ def devserver(https, web, ws, worker, assets, beat):
         m.add_process('ws', 'gunicorn --name websocket --reload --paste conf/development-websocket.ini %s' % gunicorn_args)
 
     if worker:
-        m.add_process('worker', 'hypothesis --dev celery worker --autoreload -l INFO')
+        m.add_process('worker', 'hypothesis --dev celery worker -l INFO')
 
     if beat:
         m.add_process('beat', 'hypothesis --dev celery beat')
--- a/h/celery.py
+++ b/h/celery.py
@@ -31,9 +31,9 @@ celery = Celery('h')
 celery.conf.update(
     # Default to using database number 10 so we don't conflict with the session
     # store.
-    BROKER_URL=os.environ.get('CELERY_BROKER_URL',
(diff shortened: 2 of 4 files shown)


robertknight:
In addition to Python 3.6 support, it turns out that this may also be useful for partners running h due to support for new brokers such as [SQS](http://docs.celeryproject.org/en/latest/whatsnew-4.0.html#amazon-sqs-transport-now-officially-supported) in Celery 4.x

h/cli/commands/devserver.py
```diff
@@ -80,7 +80,7 @@ def devserver(https, web, ws, worker, assets, beat):
         m.add_process('ws', 'gunicorn --name websocket --reload --paste conf/development-websocket.ini %s' % gunicorn_args)
 
     if worker:
-        m.add_process('worker', 'hypothesis --dev celery worker --autoreload -l INFO')
+        m.add_process('worker', 'hypothesis --dev celery worker -l INFO')
```
hmstepanek: It looks like we don't need this as we have other ways of reloading files automagically when running in dev mode, if I'm understanding what this feature was intended for. (I know it was deprecated but just being thorough about whether we need to replace it with something else or not and I think the answer is no we don't.)

robertknight: We have autoreloading of the webpage/API-serving code in h via gunicorn's `--reload` flag. I don't think we have any replacement mechanism for `--autoreload` in place here, and I'm [not sure it was actually working anyway](https://github.com/celery/celery/issues/1658). I don't know what commonly used alternatives are but I found one suggestion to use [watchdog](http://avilpage.com/2017/05/how-to-auto-reload-celery-workers-in-development.html).

PR #4844, opened 2018-03-01
Update Alpine Linux base image

robertknight:
Update the Alpine Linux base used by the Docker image from 3.4 to 3.7. Alpine is now an [official image](https://github.com/docker-library/official-images/pull/3762) rather than one from the gliderlabs org.

This brings along Python 3.6 but also security updates etc. in other packages. ~~Depends on https://github.com/hypothesis/h/pull/4843 for a build fix.~~

Changes required in the Dockerfile:

- Replaced removed `apk-install` with `apk add --no-cache`
- Fixed an issue on nginx startup by making a tmp dir it uses writable by the "hypothesis" user which runs nginx
- Removed use of `npm cache clean` which is no longer supported by the newer version of npm

I also added a script to document how to run an h Docker image built locally using `make docker`. Assuming you have a standard local dev setup using the suggested container names, `./scripts/run-h-dev-in-docker` will run the local Docker image configured to talk to the local Postgres, Elasticsearch and RabbitMQ containers.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -19,7 +19,7 @@ RUN addgroup -S hypothesis && adduser -S -G hypothesis -h /var/lib/hypothesis hy
 WORKDIR /var/lib/hypothesis
 
 # Ensure nginx state and log directories writeable by unprivileged user.
-RUN chown -R hypothesis:hypothesis /var/log/nginx /var/lib/nginx
+RUN chown -R hypothesis:hypothesis /var/log/nginx /var/lib/nginx /var/tmp/nginx
 
 # Copy minimal data to allow installation of dependencies.
 COPY requirements.txt ./
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,16 +1,16 @@
-FROM gliderlabs/alpine:3.4
+FROM alpine:3.7
 MAINTAINER Hypothes.is Project and contributors
 
 # Install system build and runtime dependencies.
-RUN apk-install \
+RUN apk add --no-cache \
     ca-certificates \
     collectd \
     collectd-nginx \
     libffi \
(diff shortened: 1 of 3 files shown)


robertknight:
Rebased.

seanh:
I wonder if `run-h-dev-in-docker` just belongs in `Makefile`? Where `make dev` and `make docker` are. Maybe `make dockerrun`?

robertknight:
> I wonder if run-h-dev-in-docker just belongs in Makefile? Where make dev and make docker are. Maybe make dockerrun?

Something like this (I'm using `run-docker` to follow the verb-first convention of the other commands)?

```make
# Run docker image built with `docker` task
.PHONY: run-docker
run-docker:
	$(eval RABBITMQ_CONTAINER ?= rabbitmq)
	$(eval PG_CONTAINER ?= postgres)
	$(eval ES_CONTAINER ?= elasticsearch)
	docker run \
		--link $(RABBITMQ_CONTAINER) \
		--link $(PG_CONTAINER) \
		--link $(ES_CONTAINER) \
		-e "APP_URL=http://localhost:5000" \
		-e "BROKER_URL=amqp://guest:guest@$(RABBITMQ_CONTAINER):5672//" \
		-e "DATABASE_URL=postgresql://postgres@$(PG_CONTAINER)/postgres" \
		-e "ELASTICSEARCH_HOST=http://$(ES_CONTAINER):9200" \
		-p 5000:5000 \
		hypothesis/hypothesis:dev
```

seanh:
@robertknight Yeah looks good

robertknight:
OK, I've moved the content of that script to a Makefile task as suggested.

PR #4845, opened 2018-03-01
Fix some flakey tests

seanh:
This fixes an `IntegrityError` that could be caused by the `Annotation` test factory that could cause many different tests to intermittently fail. It's a bit of a tricky one - see the individual commit messages. We should probably simplify a bunch of our test factories, not to mention our whole document equivalence system, to avoid this sort of thing. But for now this should fix some flakey tests.

Example of a flakey test fail caused by this: https://travis-ci.org/hypothesis/h/jobs/342170595

You can verify that `factories.Annotation` alone is responsible for crashes in a dev shell. This will crash:

```python
$ ./bin/hypothesis --dev shell
Python 2.7.14 (default, Sep 23 2017, 22:06:14)
Type "copyright", "credits" or "license" for more information.

IPython 5.5.0 -- An enhanced Interactive Python.
?         -> Introduction and overview of IPython's features.
%quickref -> Quick reference.
help      -> Python's own help system.
object?   -> Details about 'object', use 'object??' for extra details.

Environment:

In [1]: from tests.common import factories

In [2]: factories.set_session(request.db)

In [3]: for _ in range(999999):
   ...:     factories.Annotation()
   ...:
```

Note there are multiple bugs with the annotation factory and this pr only fixes one of them. The above will still crash on this branch. More prs to follow.

### DIFF
--- a/tests/common/factories/annotation.py
+++ b/tests/common/factories/annotation.py
@@ -77,15 +77,22 @@ class Annotation(ModelFactory):
         document_uri_dicts = [document_uri_dict()
                               for _ in range(random.randint(1, 3))]
 
-        def document_meta_dict(**kwargs):
+        def document_meta_dict(type_=None):
             """
             Return a randomly generated DocumentMeta dict for this annotation.
 
             This doesn't add anything to the database session yet.
             """
-            kwargs.setdefault('document', None)
-            kwargs.setdefault('claimant', self.target_uri)
+            kwargs = {
+                'document': None,
+                'claimant': self.target_uri,
+            }
+
+            if type_ is not None:
+                kwargs['type'] = type_
+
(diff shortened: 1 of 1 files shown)


tests/common/factories/annotation.py
```diff
@@ -77,15 +77,22 @@ def document_uri_dict():
...
             """
-            kwargs.setdefault('document', self.document)
-            kwargs.setdefault('claimant', self.target_uri)
+            kwargs = {
+                'document': None,
+                'claimant': self.target_uri,
+            }
+
+            if type_ is not None:
+                kwargs['type'] = type_
```
seanh: I mixed in some code tidy up with the actual fix. See the individual commits for separate diffs

PR #4848, opened 2018-03-02
Converge on grey-6, font size & adjust icon margin in sidebar

hmstepanek:
-Make dark greys the same (grey-6)
-Move copy icon over so it’s padding is equal with the left side of the input text
-Increase size of matching annotations to match sidebar subtitles, move hard coded font into _base defs, and remove extra css subtitle count class

Before:
image

After:
image

### DIFF
--- a/h/static/styles/partials/_base.scss
+++ b/h/static/styles/partials/_base.scss
@@ -39,6 +39,9 @@ $title-font-size: 19px;
 // Font sizes for specific categories of control
 $input-font-size: 15px;
 
+// Font size for search results "# Matching Annotations" & sidebar subtitles
+$search-results-subtitle-font-size: 15px;
+
 // Minimum font size for <input> fields on iOS. If the font size is smaller than
 // this, iOS will zoom into the field when focused.
 $touch-input-font-size: 16px;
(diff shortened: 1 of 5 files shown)


hmstepanek:
@dawariley needs to approve before merging.

hmstepanek:
@dawariley has approved and is ready to push these changes to qa.

h/static/styles/partials/_base.scss
```diff
@@ -39,6 +39,9 @@ $title-font-size: 19px;
 // Font sizes for specific categories of control
 $input-font-size: 15px;
 
+// Font size for search results "# Matching Annotations" & sidebar subtitles
+$search-results-subtitle-font-size: 15px;
```
robertknight: The other fonts defined in this file have more generic names than this because they were intended to form part of a design system in which there are only a small number of distinct font sizes ("normal", "small", "input" and "title") and colors.

@dawariley I don't need an answer on this right away, but do you have an idea of whether this font is just for this specific part of the UI or whether it will more generally used elsewhere as some kind of "heading level X" font or something like that?

PR #4873, opened 2018-03-08
Remove error state of form inputs after editing them

robertknight:
Remove the bold red border and validation error messages after editing a
form input, since the validation error message is no longer applicable
at that point.

Form inputs were already being rendered with a `js-form-input` class.
Add a controller for those elements which removes the "is-error" class
after the child `<input>` is edited.

Note that the `is-error` class controls both the border style and
visibility of validation error messages.

----

On master, if you submit a form and there is an error with a field, the field is rendered with a bold outline and error message:

screenshot 2018-03-08 11 16 20

If you then edit the field, the error message remains even though it no longer applies:

screenshot 2018-03-08 11 16 58

On this branch, the error message goes away after editing the field:

screenshot 2018-03-08 11 16 36

### DIFF
--- a/h/static/scripts/controllers/form-input-controller.js
+++ b/h/static/scripts/controllers/form-input-controller.js
@@ -1,0 +1,33 @@
+'use strict';
+
+const Controller = require('../base/controller');
+const { setElementState } = require('../util/dom');
+
+/**
+ * Controller for individual form input fields.
+ *
+ * This is used for both forms with inline editing (ie. those which show "Save"
+ * and "Cancel" buttons beneath individual form fields, see `FormController`)
+ * and those without.
+ *
+ * Note that for forms using inline editing much of the logic lives in the
+ * form-level controller rather than here.
+ */
+class FormInputController extends Controller {
+  constructor(element) {
+    super(element);
+
+    const hasError = element.classList.contains('is-error');
+    this.setState({ hasError });
+
+    this.refs.formInput.addEventListener('input', () => {
+      this.setState({ hasError: false });
+    });
+  }
+
+  update() {
+    setElementState(this.element, { error: this.state.hasError });
+  }
+}
+
+module.exports = FormInputController;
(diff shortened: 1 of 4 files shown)


lyzadanger:
> Form inputs were already being rendered with a js-form-input class.
Add a controller for those elements which removes the "is-error" class
after the child <input> is edited.

I don't know how much client-side validation we've got going on, but will the error class be removed on blur (or whatever) if the remaining field value violates client-side validation (i.e. is still in error and the browser can tell that it's still in error)?

lyzadanger:
Hi! I pulled the branch and tried it out. In my new (possibly borken, who knows) add-a-group admin panel, I didn't see the error state get removed after fixing the form field that was in error. Here's a GIF (it may be hard to tell, but I'm clicking around unfocusing the field trying to get the error to clear)

error-states

This wasn't what I expected, but maybe I'm doing it wrong?

robertknight:
> This wasn't what I expected, but maybe I'm doing it wrong?

No. I forgot to enable this behaviour for the admin site by registering the appropriate controller in `admin-site.js`. This should work now.

h/static/scripts/controllers/form-input-controller.js
```diff
@@ -0,0 +1,33 @@
...
+const { setElementState } = require('../util/dom');
+
+/**
+ * Controller for individual form input fields.
+ *
+ * This is used for both forms with inline editing (ie. those which show "Save"
+ * and "Cancel" buttons beneath individual form fields, see `FormController`)
+ * and those without.
+ *
+ * Note that for forms using inline editing much of the logic lives in the
```
lyzadanger: Thank you for this potentially sanity-saving comment!

h/static/scripts/controllers/form-input-controller.js
```diff
@@ -0,0 +1,33 @@
...
+ * form-level controller rather than here.
+ */
+class FormInputController extends Controller {
+  constructor(element) {
+    super(element);
+
+    const hasError = element.classList.contains('is-error');
+    this.setState({ hasError });
+
+    this.refs.formInput.addEventListener('input', () => {
```
lyzadanger: Ah! I think I like that it's listening on `input` versus `blur` or something. Neat.

PR #4879, opened 2018-03-12
Use sequence schema for "origins" field of group creation schema

robertknight:
Now that we have support for rendering list inputs in forms (albeit in
an unpolished way), use colander's standard method of representing list
fields in schemas for group origins.

This enables validation of the content of each list item using a
validator plus whatever UI affordances we have in our forms for editing
lists of things.

----

In user-facing terms, these changes mean that the "Scope origins" field in the form for creating groups has changed from a text box to our current (not yet polished) UI for editing a list of things, and that invalid URLs fail validation (note that the scheme is not checked and the validation error is not presented nicely but at least it is there):

screenshot 2018-03-12 16 10 37

### DIFF
--- a/h/schemas/admin_group.py
+++ b/h/schemas/admin_group.py
@@ -3,7 +3,7 @@
 from __future__ import unicode_literals
 
 import colander
-from deform.widget import SelectWidget, TextAreaWidget
+from deform.widget import SelectWidget, SequenceWidget, TextAreaWidget
 
 from h import i18n
 from h import validators
@@ -86,11 +86,13 @@ class CreateAdminGroupSchema(CSRFSchema):
         missing=None
     )
 
-    origins = colander.SchemaNode(
-        colander.String(),
+    origins = colander.SequenceSchema(
+        colander.Sequence(),
+        colander.SchemaNode(colander.String(),
+                            name='origin',
+                            validator=colander.url),
         title=_('Scope Origins'),
-        widget=TextAreaWidget(rows=5),
-        hint=_('Enter scope origins (e.g. "http://www.foo.com"), one per line'),
-        preparer=_split_origins,
-        missing=None
+        hint=_('Origins where this group appears (e.g. "https://example.com")'),
+        missing=None,
+        widget=SequenceWidget(add_subitem_text_template=_('Add origin')),
     )
--- a/tests/h/schemas/admin_group_test.py
+++ b/tests/h/schemas/admin_group_test.py
@@ -67,17 +67,13 @@ class TestCreateGroupSchema(object):
 
         bound_schema.deserialize(group_data)
 
-    @pytest.mark.parametrize('input_origins,expected_origins', (
-        ('http://www.foo.com', ['http://www.foo.com']),
-        ('http://www.foo.com\r\nhttps://www.foo.com', ['http://www.foo.com', 'https://www.foo.com']),
-        ('http://www.foo.com   ', ['http://www.foo.com']),
-        ('http://www.foo.com\nhttps://www.foo.com', ['http://www.foo.com', 'https://www.foo.com']),
-    ))
-    def test_it_splits_origins_by_line(self, group_data, bound_schema, input_origins, expected_origins):
-        group_data['origins'] = input_origins
-        appstruct = bound_schema.deserialize(group_data)
-
-        assert appstruct['origins'] == expected_origins
+    @pytest.mark.parametrize('invalid_origin', [
+        'not-a-url',
+    ])
+    def test_it_raises_if_origin_invalid(self, group_data, bound_schema, invalid_origin):
+        group_data['origins'] = [invalid_origin]
+        with pytest.raises(colander.Invalid, match='origins.*Must be a URL'):
+            bound_schema.deserialize(group_data)
 
 
 class TestCreateSchemaWithValidator(object):
@@ -124,13 +120,20 @@ class TestCreateSchemaWithValidator(object):
 
 @pytest.fixture
 def group_data(factories):
+    """
+    Return a serialized representation of the "Create Group" form.
+
+    This is the representation that Deform passes to Colander for
+    deserialization and validation after the HTML form is processed by
+    Peppercorn.
+    """
     return {
         'name': 'My Group',
         'authority': 'example.com',
         'group_type': 'open',
         'creator': factories.User().username,
         'description': 'Lorem ipsum dolor sit amet consectetuer',
-        'origins': 'http://www.foo.com\r\nhttps://www.foo.com'
+        'origins': ['http://www.foo.com', 'https://www.foo.com'],
     }
 
 


h/schemas/admin_group.py
```diff
@@ -86,11 +86,13 @@ class CreateAdminGroupSchema(CSRFSchema):
...
+                            name='origin',
+                            validator=colander.url),
         title=_('Scope Origins'),
-        widget=TextAreaWidget(rows=5),
-        hint=_('Enter scope origins (e.g. "http://www.foo.com"), one per line'),
-        preparer=_split_origins,
-        missing=None
+        hint=_('Origins where this group appears (e.g. "https://example.com")'),
+        missing=None,
+        widget=SequenceWidget(add_subitem_text_template=_('Add origin')),
```
robertknight: I think we might want to use the term "URL" rather than "origin" now, but we can test this out with a few people in the org once it is in production.

h/schemas/admin_group.py
```diff
@@ -86,11 +86,13 @@ class CreateAdminGroupSchema(CSRFSchema):
         missing=None
     )
 
-    origins = colander.SchemaNode(
-        colander.String(),
+    origins = colander.SequenceSchema(
+        colander.Sequence(),
+        colander.SchemaNode(colander.String(),
+                            name='origin',
+                            validator=colander.url),
```
robertknight: We're only validating that the input is a URL here, not that it is an HTTP(S) URL. That is simply for the convenience of using the existing `colander.url` validator to start with. We're also not warning the user that paths are ignored.

tests/h/schemas/admin_group_test.py
```diff
@@ -124,13 +120,20 @@ def user_validator(self, user_svc):
 
 @pytest.fixture
 def group_data(factories):
+    """
+    Return a serialized representation of the "Create Group" form.
+
+    This is the representation that Deform passes to Colander for
+    deserialization and validation after the HTML form is processed by
+    Peppercorn.
+    """
```
robertknight: I added this since it is not that obvious what the relation of this fixture is to the data that actually gets submitted by. the user's browser when they save the form.

seanh:
As Lyza did when she tried to add a simple select control, I think you're running into problems with our over-the-top form design, that makes it difficult to add common new UI elements without considerable design work.

I think this multiple fields approach is the right solution in the long term, but I'm not sure that we should merge this right now until we have better UX for it. Right now, I think this is considerably more confusing to users than what's currently on master.

On master - all scopes go into one text field, one scope per line. It's simplistic but it's clear and obvious. A single multi-line text field is something our form UI design can handle (this type of field already exists elsewhere in our forms).

master

The main limitation, apart from just not being the best UX, is that there's no validation, and even if there was validation (which we could certainly add, at least server-side) there wouldn't be a great way to display validation error messages (it wouldn't be clear which line of the multiline text field an error message applied to).

I think an intermediate step would be to add validation and simply show the validation error(s) on the multiline text field and let the user figure out which line(s) of the text they refer to (not likely to be many lines of text, in fact in most cases it will be just one line).

But on this branch, I think the UI quickly gets even more confusing:

branch

* I don't think it's clear that the **Scope Origins** title applies to all four text fields, remove buttons, and the **Add Origin** button. Some visual grouping  of the related controls under the single title is needed, e.g. by using additional whitespace or by drawing a border around them all. Here's a [screenshot of the entire form](https://user-images.githubusercontent.com/22498/37367618-d7db6846-26fb-11e8-9c7b-47ecb5c7d343.png) so that you can see how the scope-related fields aren't visually grouped together or visually separated from the unrelated fields. They each look like separate, unrelated, and un-labelled, fields.

* Error messages related to scopes appear outside of the text fields, rather than within the borders of the fields as with other error messages in our forms. The error message text is also black rather than red like other error messages. They don't look like error messages, look like they might be labels.

* It's not clear which field an error message applies to. Error message are actually rendered closer to the _next_ field than to the field that the error message actually applies to.

* The text fields for the scopes seem to be about twice as high as they need to be, which I think just adds to the confusing-ness of the layout. The reason many of our forms are really high is because the form label is rendered within the field's border (as is the error message, when there is one) but neither is the case with these scope fields so I think having the field be really tall (looks like a multi-line field) is confusing.

seanh:
I wonder if there's a way to get this change in - split the group scopes into separate single-line text fields for each scope, individually validated and with validation errors rendered next to each individual field - without requiring serious design work but also without making the form confusing?

@robertknight What do you think of this idea below? This might be something good enough that we can implement now, and then improve the design later:

2018-03-14-104455_2560x1440_scrot

What I've done here:

1. The scope origin text fields are the same as other single-line text fields we already have in our forms. So the existing single-line text field design will work. Field label and error messages are inside the border.

2. In order to enable (1) the **Scope Origin** title is repeated in each scope origin text field, which means it's clear what each field is. This also means that it's not necessary to use design skills to visually group the scope origin fields together under a single **Scope Origins** title and/or to visually separate them from the rest of the form fields.

   Of course a really good form design would probably avoid the need to repeat the **Scope Origin** label multiple times (or would it?) but this approach is quick to implement and I think the meaning of the fields is clear.

3. Error messages are red and are inside the borders as with other form fields, so it's clear that it's an error and which field it applies to.

4. I changed the button label to **Add another origin** just to make it more clear what it does and that this button doesn't apply to the existing origin field above the button.

One thing I didn't manage to do in this mockup: I think it would be clearer if the **Remove** button was to the right of the scope origin field it belongs to, instead of below it, so that the input and the remove button are on the same "row" if you like, with the remove button aligned with the right-hand-side of the row. This would make it clearer what the remove button applies to. But getting that to work in the HTML and CSS was tricky. Here's an example of the kind of thing I'm trying to describe, from Bootstrap:

2018-03-14-110136_2560x1440_scrot

(I don't think the remove button needs to be "joined to" the input by a common border, as in Bootstrap, they can be separate but it just needs to be on the same "row" and to the right of the button.)

h/schemas/admin_group.py
seanh: _Is_ the path ignored? Or does creating a group scoped to a URL-with-path end up meaning that the group doesn't appear on the site?

robertknight: Good question. It looks like the path is _not_ ignored at the point when checking whether the origin associated with a group matches that of a document. If a `GroupScope.origin` has a path I think the result will be that it will never match anything since the origin component of the document URL is compared against the `GroupScope.origin`.

seanh: Yep. Because a `GroupScope.origin` is meant to be, and assumed to be, an origin, not a URL. But we've no validation of that because, until now, these things were only ever created by developers. The place to put in that origin-not-path validation is probably here in this admin form. (But I think we can do that as a separate card.)

seanh:
Ok, I think since at least one origin is required there should be one origin field in the form to start with, instead of having to click add origin! But since I know there's another pr coming to refactor the UX here that's fine

PR #4881, opened 2018-03-13
Add organization to group model

hmstepanek:
This is a fix for https://github.com/hypothesis/product-backlog/issues/519.

### DIFF
--- a/h/models/group.py
+++ b/h/models/group.py
@@ -1,7 +1,6 @@
 # -*- coding: utf-8 -*-
 
 from collections import namedtuple
-
 import enum
 import sqlalchemy as sa
 from pyramid import security
@@ -83,6 +82,9 @@ class Group(Base, mixins.Timestamps):
             'groups', order_by='Group.name'))
 
     scopes = sa.orm.relationship('GroupScope', backref='group', cascade='all, delete-orphan')
+
+    organization_id = sa.Column(sa.Integer, sa.ForeignKey('organization.id'))
+    organization = sa.orm.relationship('Organization')
 
     def __init__(self, **kwargs):
         super(Group, self).__init__(**kwargs)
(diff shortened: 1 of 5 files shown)


hmstepanek:
*Technically this is getting merged onto https://github.com/hypothesis/h/pull/4877 and then getting merged into master (after the data base migration scripts are done).

seanh:
Just adding the WIP label as this can't be merged until we've done a DB migration...

seanh:
:+1: LGTM, but I'll test it properly once we have the db migration

h/models/group.py
```diff
@@ -84,6 +83,9 @@ class Group(Base, mixins.Timestamps):
 
     scopes = sa.orm.relationship('GroupScope', backref='group', cascade='all, delete-orphan')
 
+    organization_id = sa.Column(sa.Integer, sa.ForeignKey('organization.id'))
```
seanh: Is `organization_id` non-nullable by default? I think `organization_id` should be non-nullable / required, every group should have to belong to exactly one organization. But this is gonna mean more DB migrations will be needed:

* one to create a default Hypothesis organization
* and I think the migration that adds the `organization_id` column will need to create a nullable column, then set a column value for all preexisting groups, _then_ add the not-nullable constraint to the column

hmstepanek:
I agree with everything you just said Sean! I was thinking the same thing.

h/models/group.py
hmstepanek: @robertknight  @seanh https://github.com/hypothesis/h/pull/4886 is the db migration script. So we will do this adding organizations thing in the following order (feel free to correct this if I get something wrong or if you think these steps can be combined) 
1. Create an organization table, add the organization model, and add a nullable organization id to the group model.
1. Create a default hypothesis organization similar to the __world__ group only for organizations.
1. Migrate all groups to be under that default group and make sure creating a group always adds the group under the default organization if one is not specified. 
1. Change organization id so it is no longer nullable.
1. Migrate certain groups to be under a different organization (such as biopub). <--This will be the original story that started this mess.

PR #4884, opened 2018-03-15
Improve design of form list inputs (Part 1)

robertknight:
~~**Depends on https://github.com/hypothesis/h/pull/4879**~~

These are the initial batch of improvements to the design of form list input. The changes in this PR are to:

- Reduce the height of input fields without labels
- Make the "Remove item" button more subtle
- Display validation error messages closer to the input field and in red

Notably missing from this first batch of changes is to make the outline of list items with validation errors bold + red as they are for other input fields. That will come in follow-up work.

**Master:**

screenshot 2018-03-15 13 53 04

screenshot 2018-03-15 13 52 45

**This branch:**

screenshot 2018-03-15 13 52 01

screenshot 2018-03-15 13 52 17

### DIFF
--- a/h/static/scripts/controllers/list-input-controller.js
+++ b/h/static/scripts/controllers/list-input-controller.js
@@ -23,7 +23,8 @@ class ListInputController extends Controller {
 
     // Handle 'Remove' button.
     element.addEventListener('click', (event) => {
-      if (event.target.getAttribute('data-ref') === 'removeItemButton') {
+      const btn = event.target.closest('button');
+      if (btn.getAttribute('data-ref') === 'removeItemButton') {
         const parentItem = event.target.closest('li');
         parentItem.remove();
       }
--- a/h/static/styles/partials/_form-input.scss
+++ b/h/static/styles/partials/_form-input.scss
@@ -8,8 +8,11 @@
   // Padding between top border of input and input text. This includes space
   // for the field label
   $top-padding: 35px;
+  $top-padding-no-label: 5px;
+
   // Padding between bottom border of input and input text.
   $bottom-padding: 15px;
+  $bottom-padding-no-label: 5px;
 
   // Extra vertical padding added above input content for form fields displaying
   // inline hints
@@ -135,10 +138,8 @@
 
   // The actual <input> element for the field
   .form-input__input {
-    padding-top: $top-padding;
     padding-left: $h-padding;
     padding-right: $h-padding;
-    padding-bottom: $bottom-padding;
     width: 100%;
 
     background: none;
@@ -147,6 +148,24 @@
     outline: none;
     border: 1px solid $grey-3;
     border-radius: 3px;
+  }
+
+  // Set the padding depending on whether or not this field has an associated
+  // label. The `<label>` lives outside the input in the DOM but is positioned
+  // _inside_ the `<input>`'s border visually.
+  //
+  // We do this rather than use a separate element for the field's border so
+  // that clicking anywhere inside the visual border focuses the input.
+  .form-input__input.has-label {
+    padding-top: $top-padding;
+    padding-bottom: $bottom-padding;
+    --bottom-padding: $bottom-padding;
+  }
+
+  .form-input__input:not(.has-label) {
+    padding-top: $top-padding-no-label;
+    padding-bottom: $bottom-padding-no-label;
+    --bottom-padding: $bottom-padding-no-label;
   }
 
   select.form-input__input {
@@ -173,7 +192,11 @@
     position: relative;
     left: -1px;
     top: -1px;
+
+    // Assume largest possible padding if CSS vars are not supported.
     padding-bottom: $bottom-padding - 2px;
+    padding-bottom: calc(var(--bottom-padding) - 2px);
+
     border-width: 2px;
   }
 
--- a/h/templates/deform/sequence_item.jinja2
+++ b/h/templates/deform/sequence_item.jinja2
@@ -1,15 +1,25 @@
 {% if not field.widget.hidden %}
 <li class="list-input__item" title="{{ field.description }}">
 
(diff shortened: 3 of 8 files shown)


h/static/scripts/controllers/list-input-controller.js
```diff
@@ -23,7 +23,8 @@ class ListInputController extends Controller {
 
     // Handle 'Remove' button.
     element.addEventListener('click', (event) => {
-      if (event.target.getAttribute('data-ref') === 'removeItemButton') {
+      const btn = event.target.closest('button');
+      if (btn.getAttribute('data-ref') === 'removeItemButton') {
```
robertknight: This is needed because the user is now clicking in an `<svg>` (inside the button) rather than directly on the `<button>`.

h/static/styles/partials/_form-input.scss
```diff
@@ -149,6 +150,24 @@
...
+  // Set the padding depending on whether or not this field has an associated
+  // label. The `<label>` lives outside the input in the DOM but is positioned
+  // _inside_ the `<input>`'s border visually.
+  //
+  // We do this rather than use a separate element for the field's border so
+  // that clicking anywhere inside the visual border focuses the input.
+  .form-input__input.has-label {
+    padding-top: $top-padding;
+    padding-bottom: $bottom-padding;
+    --bottom-padding: $bottom-padding;
```
robertknight: If you haven't seen it before this syntax "declares" a CSS variable, which is then used later on with `var(--bottom-padding)` when computing the new padding after applying a thicker border.

h/templates/deform/sequence_item.jinja2
```diff
@@ -1,15 +1,25 @@
 {% if not field.widget.hidden %}
 <li class="list-input__item" title="{{ field.description }}">
 
-  {{ field.serialize(cstruct=cstruct) | safe }}
+  <div class="list-input__item-inner">
```
robertknight: `list-input__item-inner` is the horizontal row containing the field and the remove item button.

h/templates/deform/sequence_item.jinja2
```diff
@@ -1,15 +1,25 @@
...
-  {% endif %}
+  {%- if field.error and not field.widget.hidden -%}
+  <ul class="list-input__error-list">
+    {% for msg in field.error.messages() -%}
+      {%- set errstr = 'error-%s' % field.oid -%}
+      {%- set pid = (loop.index0==0 and errstr) or ('%s-%s' % (errstr, loop.index0)) -%}
+      <li class="list-input__error-item" id="{{ pid }}">{{ _(msg) }}</li>
+    {% endfor -%}
+  </ul>
+  {% endif -%}
```
robertknight: Validation error messages are shown underneath the form. I didn't try to put them inside the input field because that is quite difficult to get right across a wide range of screen sizes and possible input types.

h/static/styles/partials/_form-input.scss
```diff
@@ -173,7 +192,11 @@
     position: relative;
     left: -1px;
     top: -1px;
+
+    // Assume largest possible padding if CSS vars are not supported.
     padding-bottom: $bottom-padding - 2px;
+    padding-bottom: calc(var(--bottom-padding) - 2px);
```
robertknight: IE 11 is the only browser we support that doesn't support CSS vars. It will ignore this second `padding-bottom` declaration and just use a bit more padding than necessary.

h/templates/deform/sequence_item.jinja2
seanh: Yeah. Probably good to move away from that design generally

robertknight:
> There should already be one empty origin field when the form first loads (currently there are none and I have to click the add button)

I'll have a look at that.

> There should be no remove button next to the first origin field. It shouldn't be possible to remove the first origin field.

What if there are multiple origins in the list and you want to keep one _other_ than the first? I suppose in this instance we could require them to copy and paste the data from the second field into the first. That might be more fiddly to do if the things in the list were more complicated that simple text fields.

> If there is at least one valid scope field in the form, and also any number of empty scope fields, then it should be possible to submit the form successfully and it will create a group with just one scope and ignore the empty scopes. 

I don't think we can do that if we are going to also leverage HTML form validation attributes to tell the browser whether a field is required or not. We can chose not to use those, but it means that the user agent might not provide some affordances (device/input method specific) that it might otherwise do.

seanh:
> What if there are multiple origins in the list and you want to keep one other than the first? I suppose in this instance we could require them to copy and paste the data from the second field into the first. That might be more fiddly

Good point, that is fiddly. I guess what we'd have to do is show an `x` next to every field _until_ the second-to-last remaining field gets removed, then the `x` next to the one remaining field also disappears. (I'm fine with leaving this fiddly though.)

seanh:
> I don't think we can do that if we are going to also leverage HTML form validation attributes to tell the browser whether a field is required or not. We can chose not to use those, but it means that the user agent might not provide some affordances (device/input method specific) that it might otherwise do.

Ok, that's fine

h/templates/deform/sequence_item.jinja2
```diff
@@ -1,15 +1,27 @@
 {% if not field.widget.hidden %}
 <li class="list-input__item" title="{{ field.description }}">
 
-  {{ field.serialize(cstruct=cstruct) | safe }}
+  <div class="list-input__item-inner">
+    {{ field.serialize(cstruct=cstruct) | safe }}
 
-  <button class="btn btn--danger list-input__remove-btn" type="button"
-          data-ref="removeItemButton">Remove</button>
+    {% if (can_delete is not defined) or can_delete %}
```
robertknight: `can_delete` is undefined when rendering the template for new field items, because Deform doesn't allow passing additional arguments to be used when rendering the `sequence_item` template in this case.

PR #4887, opened 2018-03-16
Add admin view for editing groups

robertknight:
Add an admin view at `/admin/groups/{pubid}` for editing groups. Once we have a service for deleting users, the button for deleting the group will appear on this page as well.

The view does not allow changing the authority, since all the annotations and users associated with the group have to belong to that authority as well. All the other parameters of the group, including the type, can be changed.

There are some limitations of this first iteration:
- Editing members is not yet supported, that will be a follow-up piece of work.
- Editing private groups is not yet supported. Adding support for "Private" to the list of types is trivial, but we also need to allow groups without scopes, since most existing private groups do not have them.

### DIFF
--- a/h/schemas/admin_group.py
+++ b/h/schemas/admin_group.py
@@ -35,6 +35,18 @@ def user_exists_validator_factory(user_svc):
     return user_exists_validator
 
 
+@colander.deferred
+def group_type_validator(node, kw):
+    group = kw.get('group')
+    if not group:
+        return colander.OneOf([key for key, title in VALID_GROUP_TYPES])
+
+    def validate(node, value):
+        if group.type != value:
+            raise colander.Invalid(node, _('Changing group type is currently not supported'))
+    return validate
+
+
 class CreateAdminGroupSchema(CSRFSchema):
 
     group_type = colander.SchemaNode(
@@ -43,7 +55,7 @@ class CreateAdminGroupSchema(CSRFSchema):
         widget=SelectWidget(
           values=(('', _('Select')),) + VALID_GROUP_TYPES
         ),
-        validator=colander.OneOf([key for key, title in VALID_GROUP_TYPES])
+        validator=group_type_validator,
     )
 
     name = colander.SchemaNode(
--- a/h/templates/admin/groups.html.jinja2
+++ b/h/templates/admin/groups.html.jinja2
@@ -22,8 +22,13 @@
         {% for group in results %}
           <tr>
             <td>
+              {% if group.type in ('open', 'restricted') %}
               {% set edit_url = request.route_url('admin_groups_edit', pubid=group.pubid) %}
               <a href="{{ edit_url }}">{{ group.name }}</a>
+              {% else %}
+              {# The group edit view does not currently support editing private groups. #}
+              {{ group.name }}
+              {% endif %}
             </td>
             <td>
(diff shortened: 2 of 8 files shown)


lyzadanger:
Hi! There's a bit of related work mentioned here: https://github.com/hypothesis/product-backlog/issues/507#issuecomment-371928706

and the very beginnings of a delete-group service (if it's useful to this thread) here: https://github.com/hypothesis/product-backlog/issues/525#issuecomment-371929467

robertknight:
This is now updated as per @seanh 's suggestion. Attempting to change the group type will, for the moment, produce an error (pardon the ugly formatting, that's an existing issue with select controls):

screenshot 2018-03-20 18 14 23

As we address the issues with changing group types, we can gradually relax the restriction.

h/schemas/admin_group.py
```diff
@@ -35,6 +35,18 @@ def user_exists_validator(form, value):
     return user_exists_validator
 
 
+@colander.deferred
```
robertknight: See https://docs.pylonsproject.org/projects/colander/en/latest/binding.html for an explanation of the colander feature being used here.

h/schemas/admin_group.py
```diff
@@ -35,6 +35,18 @@ def user_exists_validator(form, value):
...
+def group_type_validator(node, kw):
+    group = kw.get('group')
+    if not group:
+        return colander.OneOf([key for key, title in VALID_GROUP_TYPES])
+
+    def validate(node, value):
+        if group.type != value:
+            raise colander.Invalid(node, _('Changing group type is currently not supported'))
```
seanh: This isn't important but FWIW I personally don't really like the formal / technical-ness of phrases lile "currently not supported" for usability and how the app feels. "Changing group type isn't supported yet" sounds friendlier to me and is shorter, easier to parse, and the "yet" implies that it will be supported in the future. But this is something that I don't think we have any agreement or guidelines on - what is the language style / voice of Hypothesis? In the case of this admin form it doesn't matter though

h/templates/admin/groups.html.jinja2
```diff
@@ -21,7 +21,15 @@
...
           <tr>
-            <td>{{ group.name }}</td>
+            <td>
+              {% if group.type in ('open', 'restricted') %}
+              {% set edit_url = request.route_url('admin_groups_edit', pubid=group.pubid) %}
+              <a href="{{ edit_url }}">{{ group.name }}</a>
+              {% else %}
+              {# The group edit view does not currently support editing private groups. #}
+              {{ group.name }}
+              {% endif %}
```
seanh: Just out of interest why isn't editing private groups supported? Is there any reason why it wouldn't just work?

h/schemas/admin_group.py
robertknight: I'm inclined to agree. @dawariley may correct me but I don't think we have any written documentation on the voice of the UI (eg. how formal, what tense).

h/templates/admin/groups.html.jinja2
robertknight: The schema for the form doesn't have "private" as a supported group type yet. Otherwise I think it would work.

robertknight:
Thanks for the feedback @seanh. I'm going to merge this as is and we can iterate from there.

PR #4888, opened 2018-03-16
Make functional tests pass under Python 3

robertknight:
This PR resolves a couple of issues with the functional tests under Python 3 and makes them all pass. See commit messages for details. To ensure they stay fixed, it also adds a step to the Jenkinsfile to run the functional tests under Python 3.

To test it out locally run `tox -e functional-py3` in this branch. Note that `tox -e functional` now always uses Python 2.7 in this branch.


seanh:
The march towards Python 3 continues

robertknight:
> The march towards Python 3 continues

Well, we're almost done now 🙂. The last remaining step is to actually use it in the production Docker image.

PR #4889, opened 2018-03-16
Add Creator label to My group facet

hmstepanek:
Add a relationship label after the title of the group on the search
autocomplete group facet drop down. Currently the only relationship
so far is 'Creator' but there may be more relationships defined
later. A styling class for the new relationship label was added. The 
navbar tests were also updated to reflect the relationship addition in 
the suggested_groups. 
This is a UX fix for https://github.com/hypothesis/product-backlog/issues/510.

image

Change the description of the group facet to 'show 
annotations associated with a group'. 
image

Change the Groups to return associated groups.
image


hmstepanek:
Updated PTAL.

dawariley:
good catch on the spacing @robertknight

dawariley:
Approved!

PR #4890, opened 2018-03-18
Add metric for badge count

hmstepanek:
This adds a new relic custom metric to record when the badge count is zero. Since it's recorded every time (1 when it is zero, 0 when it is not zero), the average value will give us the percentage of times the badge count is zero. 
image

### DIFF
--- a/h/views/badge.py
+++ b/h/views/badge.py
@@ -3,6 +3,7 @@
 from __future__ import unicode_literals
 
 from pyramid import httpexceptions
+import newrelic.agent
 
 from h import models, search
 from h.util.view import json_view
@@ -25,9 +26,11 @@ def badge(request):
         raise httpexceptions.HTTPBadRequest()
 
     if models.Blocklist.is_blocked(request.db, uri):
+        newrelic.agent.record_custom_metric('Custom/Badge/badgeCountIsZero', 1)
         return {'total': 0}
 
     query = {'uri': uri, 'limit': 0}
     result = search.Search(request, stats=request.stats).run(query)
 
+    newrelic.agent.record_custom_metric('Custom/Badge/badgeCountIsZero', int(result.total == 0))
     return {'total': result.total}


seanh:
Oh wow, thanks for doing this! Will take a look as soon as I get a chance...

seanh:
Hey @hmstepanek , in the past we've usually used `statsd` to send metrics like this from h (the application server uses the statsd client to report stats to statsd.svc.hypothes.is, which aggregates them and sends them on to graphite.svc.hypothes.is, and then you have to set up a graph in Grafana to display the aggregates from Graphite).

I'm wondering whether this badge metric should be done in the same way for consistency, or are there advantages to using New Relic over statsd/graphite/grafana for tracking custom metrics?

(P.S. Some metrics in the client code get sent to Google Analytics, as well!)

seanh:
(P.S. If you do decide to change it to statsd, it's `request.stats` in the code.)

hmstepanek:
Well, from what you described it seems like it’s a lot less work to use NR as this is the only change I have to make to see the data, NR takes care of the rest.

seanh:
Ok, lets go with New Relic.

I agree that it seems like less work. Unfortunately I don't _think_ New Relic and other hosted services will be able to _completely_ replace our statsd -> graphite -> grafana setup. So we're gonna have to keep maintaining that anyway. But maybe we'll one day be able to reduce Grafana to just standard graphs, maybe even all stuff from pre-packaged Grafana dashboards, and move all custom metrics to things like New Relic and Google Analytics?

I think lets see how this badge metric goes at first and then we can consider what other custom metrics we could move from Grafana to New Relic (I'm not even sure what custom metrics we have currently)

seanh:
By the way will it be possible to get the % of the time that the badge is zero over a longer period of time? The graph that you posted a screenshot looks like it has a relatively short time period, looks like the points on the graph are showing the % of 0's vs 1's for every given minute of time. But I think the question we're probably more interested in is: _How often has the badge been non-zero in the last two weeks? Or two months?_ So more of a long-term aggregation, less real-time. (I think basically just change the period of the graph to something much longer than a minute.)

hmstepanek:
I can add a class but I don't intend on keeping this metric around forever so I'm not sure we need to read too much into this metric addition. The metric data I believe is kept for 90 days so yes to your how long and over what time it will do both of those things. It does nothing if newrelic isn't running. I can add tests for this but again-not planning on this being around forever so we will need to delete the tests when we remove the metric.

seanh:
Ok. I've got a feeling we might end up keeping this around for longer. Why would we want to remove it? And we may be interested in how the badge is doing in the future. I still think I'd prefer a couple of unit tests even if this isn't sticking around, to be honest, I just like to test everything. But I'll approve this.

Thanks for doing this, I'm interested to see what the results are, and also whether we can start using New Relic for more metrics

PR #4897, opened 2018-03-22
Add metric unAuthUserGotZero & BadgeNotZero

hmstepanek:
unAuthUserGotZero is, out of all the requests that were zero,
the percentage of them where the user was None/unauthenticated.
This will tell us if a large portion of the badge returning
zero requests are due to unauthenticated users.
image

BadgeNotZero is an event for any request that is not 
zero. It contains the email.authority. This will tell us
if there are a small portion of users for which the badge count
is consistently reporting non zero.
image

### DIFF
--- a/tests/h/views/badge_test.py
+++ b/tests/h/views/badge_test.py
@@ -4,13 +4,68 @@ from __future__ import unicode_literals
 
 import pytest
 import mock
+from mock import call
 
 from pyramid import httpexceptions
 
 from h.views.badge import badge
+from h.views.badge import record_metrics
 
 
 badge_fixtures = pytest.mark.usefixtures('models', 'search_lib')
+
+
+def test_record_metrics_records_badgenotzero_with_auth_user():
+    request = mock.Mock(params={'uri': 'test_uri'})
+    request.user.username = 'foopy'
+    record_metric = mock.Mock()
+    record_event = mock.Mock()
+    record_metrics(1, request,
+                   record_metric=record_metric,
+                   record_event=record_event)
+
+    record_event.assert_called_once_with('BadgeNotZero', {'user': request.user.username})
+    record_metric.assert_called_once_with('Custom/Badge/badgeCountIsZero', 0)
+
+
+def test_record_metrics_records_badgenotzero_with_unauth_user():
+    request = mock.Mock(params={'uri': 'test_uri'})
+    request.user = None
+    record_metric = mock.Mock()
+    record_event = mock.Mock()
+    record_metrics(1, request,
+                   record_metric=record_metric,
+                   record_event=record_event)
+
+    record_event.assert_called_once_with('BadgeNotZero', {'user': 'None'})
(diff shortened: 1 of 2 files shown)


robertknight:
> Also wondering about the {}GotNonZero metric. Does this require us to look at a separate graph in New Relic for each user?

The Chrome extension has about 100K active installs. I don't know what proportion of those are logged in users but this could produce a very large number of distinct metrics in NR. Would that be a problem?

hmstepanek:
Fixed to reflect feedback from @seanh and @robertknight. Note we are now using events and the description has been updated to reflect the changes. My main concern is I have also updated the username to be email.authority instead as usernames are not unique but email.authority is and that may be a personal information issue? @segdeha Is it ok to store user emails in NR for the purposes of monitoring the badge or should we avoid that? Again I don't think this will be a permanent thing-prob get rid of this after we come up with a permanent solution for reducing traffic on the badge. We can use this info to verify the new solution is working as we expect then remove this monitoring or if we do want to keep it change it to something obscure like the numeric id or something. I think our original intention here was just to make sure it wasn't only internal h users who were getting non-zero values.

segdeha:
@hmstepanek It would be better not to store emails in logs if we can avoid it. Is there anything preventing us from using the numeric ID now?

hmstepanek:
My only issue with using a numeric id right now is it makes it more difficult to track down who the user is since our original intention was to look at internal h users vs external h users. Using a numeric id would mean we would have to go query the db and get the id's of internal h users. But as I said at the end, if we don't want to store emails I think we can just go back to using the username as that's probably unique enough for our current purposes.

robertknight:
> My main concern is I have also updated the username to be email.authority instead, as usernames are not unique but email.authority is and that may be a personal information issue?

`username` is not unique but `userid` is. We don't need to include their email address to achieve this. However, the badge endpoint is only ever used for first party accounts in practice at the moment, so just the username would actually be fine IMO.

tests/h/views/badge_test.py
```diff
@@ -25,6 +25,32 @@ def test_badge_returns_number_from_search(models, search_run):
     assert result == {'total': 29}
 
 
+@badge_fixtures
+def test_badge_returns_number_from_search_if_unauth_user(models, search_run):
```
robertknight: The logic which fetches the count doesn't differ depending on whether the user is authenticated or not, only the logic which calls NR to log a custom event. Is there a reason you're not testing the calls to NR here?

hmstepanek:
Ok added tests for the metric/event logic.

hmstepanek:
@seanh @robertknight I do not think it's a coincidence that we got a massive amount of request queueing right around the time the badge metrics were added. This has to do with the amount of times the badge is called and the event taking up more time to processes. Having seen that I don't want to release this to prod.

image

I think we can get our answer without releasing this to prod. Note the largest number of non zero and zero transactions are from unauth users and generally speaking the top auth users are h internal employees but many others show up that aren't :
image
(blue=None)

image

PR #4900, opened 2018-03-23
Add ability to delete a group from the admin interface.

sheetaluk:


h/templates/admin/groups.html.jinja2
```diff
@@ -44,6 +45,15 @@
               {% endif %}
             </td>
             <td>{{ group.members|length }}</td>
+            <td>
+              <form method="POST"
+                    action="{{request.route_path('admin_groups_delete')}}"
+                    class="form-inline js-groups-delete-form">
```
robertknight: The `js-groups-delete-form` class here doesn't appear to be used anywhere.

sheetaluk:
Added the delete button to the edit groups page for a couple of reasons:
(1) We don't want any of us accidentally deleting the wrong group. This could happen more easily from the groups list page.
(2) Getting the number of annotations for a given group is easier from the edit page than in the list groups page.

The end result is that the UI isn't very pretty. But it is consistent with how the delete button in the edit OAuth client page is placed:
image

h/services/delete_group.py
```diff
@@ -0,0 +1,33 @@
...
+    def __init__(self, request):
+        self.request = request
+
+    def delete(self, group):
+        """
+        Deletes a group, its membership relations and all annotations in the
+        group.
+        """
+
+        self._delete_annotations(group)
```
robertknight: I think we should probably disallow deleting the `__world__` group here since we treat that specially in a few other places in the code.

sheetaluk:
@robertknight you are right. We discussed this but it somehow got lost. Here is how my updated commit looks:
image

tests/h/services/delete_group_test.py
```diff
@@ -0,0 +1,82 @@
...
+
+    def test_delete_deletes_user_relationships(self, svc, db_session, factories):
+        user = factories.User()
+        group = factories.Group(creator=user)
+
+        svc.delete(group)
+
+        # note that user.groups retains deleted group through the request
+        # until db is committed/flushed but the group is marked as deleted
+        assert user.groups[0] in db_session.deleted
```
robertknight: Since `user.groups[0]` returns the `Group`, isn't this essentially testing the same thing as `test_delete_deletes_group` above?

tests/h/views/admin_groups_test.py
```diff
@@ -139,13 +140,20 @@ def test_raises_not_found_if_unknown_group(self, pyramid_request):
...
+
         ctrl = GroupEditController(pyramid_request)
 
         ctx = ctrl.read()
 
         assert ctx['form'] == self._expected_form(group)
+        assert ctx['pubid'] == group.pubid
+        assert ctx['group_name'] == group.name
+        assert ctx['num_members'] == len(group.members)
+        assert ctx['num_annotations'] == 2
```
robertknight: From a quick skim through other views it seems like we tend to use `{thing}_count` for template vars like this.

h/views/admin_groups.py
```diff
@@ -112,6 +113,19 @@ def read(self):
...
 
+    @view_config(request_method='POST',
+                 route_name='admin_groups_delete')
+    def delete(self):
+        group = self.group
+        svc = self.request.find_service(name='delete_group')
+
+        svc.delete(group)
+        self.request.session.flash(
+            'Successfully deleted group %s' % (group.name), 'success')
```
robertknight: This string ought to be marked for translate-ability using `_` (see references to `TranslationString` and `_('...')`) in other calls to `session.flash`.

h/services/delete_group.py
```diff
@@ -0,0 +1,40 @@
...
+        self.request.db.delete(group)
+
+    def _delete_annotations(self, group):
+        if group.pubid == '__world__':
+            raise DeletePublicGroupError('Public group can not be deleted0')
```
robertknight: Typo (`deleted0`).

PR #4901, opened 2018-03-23
Use Python 3 in h Docker image

robertknight:
Transition the QA/production Docker image to use Python 3 to run the h web app and workers.

supervisord still requires Python 2, so to avoid having to install Python 2 alongside Python 3, I have replaced it with a [Golang implementation](https://github.com/ochinchina/supervisord) which has all the features we need.

After this hits QA we should do some sanity testing of important parts of the site and check memory/CPU usage. After this hits production we should keep an eye on memory and CPU usage stats.

If we need to roll back for any reason, we can revert these commits and redeploy, or even just re-deploy the previous Docker image from the [deployment build page](https://jenkins.hypothes.is/job/deployment/build?delay=0sec).

**Image size stats**

Sizes reported by `docker inspect hypothesis/hypothesis:dev`:

Image size (master): 154MB
Image size (this branch): 167MB

**Testing**

1. Build the dev docker image using `make docker`
2. Run the image in a new container using `make run-docker`
3. Inspect the container and verify that processes are using the expected version of Python

```sh
[I] (h-py3)  ~/h/r/h > docker exec -it <ID of container> /bin/sh
~ $ ps aux 
PID   USER     TIME   COMMAND
    1 hypothes   0:00 {supervisord} /usr/bin/python2 /usr/bin/supervisord -c co
   11 hypothes   0:00 python /var/lib/hypothesis/bin/logger
   12 hypothes   0:00 {gunicorn} /usr/bin/python3.6 /usr/bin/gunicorn --name we
   13 hypothes   0:00 {gunicorn} /usr/bin/python3.6 /usr/bin/gunicorn --paste c
   14 hypothes   0:02 python -m h celery worker
   16 hypothes   0:00 nginx: master process nginx
   17 hypothes   0:00 nginx: worker process
   18 hypothes   0:00 nginx: worker process
   24 hypothes   0:01 {gunicorn} /usr/bin/python3.6 /usr/bin/gunicorn --paste c
   29 hypothes   0:02 {gunicorn} /usr/bin/python3.6 /usr/bin/gunicorn --name we
   35 hypothes   0:00 python -m h celery worker
   36 hypothes   0:00 python -m h celery worker
   44 hypothes   0:00 /bin/sh
   48 hypothes   0:00 ps aux
~ $ stat /proc/11/exe # check that "python" is Python 3
  File: '/proc/11/exe' -> '/usr/bin/python3.6'
  Size: 0         	Blocks: 0          IO Block: 1024   symbolic link
Device: a0h/160d	Inode: 1034058     Links: 1
Access: (0777/lrwxrwxrwx)  Uid: (  102/hypothesis)   Gid: (  103/hypothesis)
Access: 2018-03-21 08:56:10.000000000
Modify: 2018-03-21 08:56:10.000000000
Change: 2018-03-21 08:56:10.000000000

~ $ 
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,7 +18,7 @@ RUN apk add --no-cache \
     libffi \
     libpq \
     nginx \
-    python2 \
+    python3 \
     py2-pip \
     git
 
@@ -37,9 +37,11 @@ RUN apk add --no-cache --virtual build-deps \
     build-base \
     libffi-dev \
     postgresql-dev \
-    python-dev \
-  && pip install --no-cache-dir -U pip supervisor \
-  && pip install --no-cache-dir -r requirements.txt \
+    python3-dev \
+  && python3 -m ensurepip \
+  && pip3 install --no-cache-dir -U pip \
+  && pip2.7 install --no-cache-dir -U supervisor \
+  && pip3 install --no-cache-dir -r requirements.txt \
   && apk del build-deps
 
 # Copy nginx config
@@ -66,6 +68,7 @@ EXPOSE 5000
 ENV PATH /var/lib/hypothesis/bin:$PATH
 ENV PYTHONIOENCODING utf_8
 ENV PYTHONPATH /var/lib/hypothesis:$PYTHONPATH
+RUN ln -sf /usr/bin/python3 /usr/bin/python
 
 # Start the web server by default
 USER hypothesis
--- a/Dockerfile
+++ b/Dockerfile
@@ -19,7 +19,6 @@ RUN apk add --no-cache \
     libpq \
     nginx \
     python3 \
-    py2-pip \
     git
 
 # Create the hypothesis user, group, home directory and package directory.
@@ -32,6 +31,9 @@ RUN chown -R hypothesis:hypothesis /var/log/nginx /var/lib/nginx /var/tmp/nginx
 # Copy minimal data to allow installation of dependencies.
 COPY requirements.txt ./
 
+# Install Golang implementation of supervisord
+COPY --from=ochinchina/supervisord:latest /usr/local/bin/supervisord /usr/local/bin/supervisord
+
 # Install build deps, build, and then clean up.
 RUN apk add --no-cache --virtual build-deps \
     build-base \
@@ -40,7 +42,6 @@ RUN apk add --no-cache --virtual build-deps \
     python3-dev \
   && python3 -m ensurepip \
   && pip3 install --no-cache-dir -U pip \
-  && pip2.7 install --no-cache-dir -U supervisor \
   && pip3 install --no-cache-dir -r requirements.txt \
   && apk del build-deps
 
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -31,9 +31,9 @@ node {
                                          "-e ELASTICSEARCH_URL=${elasticsearchHost} " +
                                          "-e TEST_DATABASE_URL=${databaseUrl}") {
                 // Test dependencies
-                sh 'apk add --no-cache build-base libffi-dev postgresql-dev python-dev'
-                sh 'apk add --no-cache python3 python3-dev'
-                sh 'pip install -q tox'
+                sh 'apk add --no-cache build-base libffi-dev postgresql-dev python3-dev'
+                sh 'apk add --no-cache python-dev' // while we continue to run tests under Python 2.7
+                sh 'pip3 install -q tox'
 
                 // Unit tests
                 sh 'cd /var/lib/hypothesis && tox'


seanh:
I wonder if we need to wait for a quiet time (e.g. when not so busy on groups work) to merge this...

hmstepanek:
That's prob not be a bad idea @seanh.

Dockerfile
```diff
@@ -18,7 +18,7 @@ RUN apk add --no-cache \
     libffi \
     libpq \
     nginx \
-    python2 \
+    python3 \
```
hmstepanek: If supervisor requires python2-do we still need to keep python2 here?

seanh:
Something to watch out for: the click library (which we use to implement the `hypothesis` CLI) has complicated problems with Python 3: <http://click.pocoo.org/5/python3/>. We should make sure that the CLI works correctly with Python 3, including when run via Jenkins.

robertknight:
> I wonder if we need to wait for a quiet time (e.g. when not so busy on groups work) to merge this...

I agree it makes sense to pick a quiet time of day and sprint. I'd prefer to have a more definite timeline than "when we're done with groups work" so we can wrap this up. How about next Tuesday morning (London time) when it'll be the start of the sprint and fairly quiet traffic-wise?

If there is a problem, we can just revert the single commit here and schedule work for another time.

> Something to watch out for: the click library (which we use to implement the hypothesis CLI) has complicated problems with Python 3: http://click.pocoo.org/5/python3/.

I'm using Python 3 in my local dev environment and haven't encountered issues with the CLI tools.  I've also done some basic testing of critical commands (`hypothesis {shell, migrate, search}`) in the Docker image.

From reading the docs, the main issues flagged up are to do with CLI tools that read binary data from stdin/write to stdout. AFAIK, none of ours do. Regarding the section at the bottom about I/O encoding, Python is configured to use UTF-8 in the Docker image which also avoids several hazards.

Dockerfile
robertknight: The `py2-pip` dependency takes care of that.

hmstepanek:
I think the start of the sprint sounds good.

seanh:
I'd like to wait at least another week from next Tuesday. I'd like some time to do a little Python 3 review of the code myself, but I'll be focusing on the groups stuff in the current sprint until end of Tuesday. The following Tuesday should work.

robertknight:
> The following Tuesday should work.

OK, that's fine with me. Unless there are any objections I will mark Tuesday 10th April as 🐍🐍🐍-day.

seanh:
Sorry to be a drag but I think there are a few more things we should do to make our Python 2.7 code better and more Python 3-safe before we pull the trigger on Python 3. We can deploy each of these one by one, so that if something goes wrong we'll know what to blame.

If we're agreed we can break these out into separate cards and label them Python 3000:

- [ ] Use Dependabot to get all our Python dependencies onto the latest versions, as we did for bouncer. There are less likely to be Python 3 bugs in our dependencies this way. On the other hand this is a big job, a lot of dependency updates, compared to bouncer.

- [ ] Run the tests with Python 2.7's `-3` flag to check for Python 3 warnings. `pip install pytest mock factory-boy hypothesis` and then `python -3 -m pytest tests/h`.

  Many of the warnings that this produces are from our libraries, not our own code, which is a little worrying and may argue for doing the Dependabot upgrades.
  
  - [ ] Replace `getfuncargvalue` with `getfixturevalue`. (This looks like actually something that was deprecated in pytest, not a Python 3 thing, but for some reason `python -3` prints it out.)

- [x] Run `caniusepython3` to check that all our dependencies are Python-compatible: `pip install caniusepython3; caniusepython3 requirements.txt`.

  It says that psycogreen isn't Python 3-compatible but it looks like this is just becase [they haven't updated their PyPI classifiers](https://bitbucket.org/dvarrazzo/psycogreen/pull-requests/2/add-python-3-classifier/diff).
  
  It also says that deform-jinja2 isn't Python 3-compatible. It looks like this is true: deform-jinja2 is a Python 2.6 library with no tests that hasn't been upgraded to Python 3. We may want to do Python 3 analysis of this library, or remove it, or just keep a careful eye on form rendering and validation as we upgrade.

- [x] Look for areas of low test coverage.

  I've looked through and it looks pretty good to me. The main thing I'm worried about is **the websocket and streamer**. This is where our weakest test coverage is. We could add more unit tests there, or we could just keep a careful eye on it as we upgrade things.

- Run [`futurize`](http://python-future.org/automatic_conversion.html). This converts Python 2 code to Python 2/3-compatible code.

  Note `futurize` will want to add stuff to the `click` CLI modules that probably won't work, we'll have to ignore that.
  
  Deploy each of these one at a time:

  - [ ] Stage 1 ("safe" fixes): `futurize -wn1 h tests`.
  - [ ] Add `unicode_literals` to every file that doesn't have it yet: `futurize --unicode-literals -wn1 h tests`. If there are any strings that we actually want to be byte strings we'll need to prefix them with `b` at this stage.
  - [ ] Add other future imports to all files: `futurize --all-imports -wn1 h tests`.
  
  - [ ] Stage 2 ("mostly safe" changes): `futurize -wn2 h tests`. I'm not sure that we actually need or want everything that this does, but we should take a look.

- [ ] Add `flake8-future-import` to Travis to check that we have all the future imports in every file and that it stays that way. `pip install flake8-future-import` and then run `flake8 --ignore FI12,FI13,FI15,FI16,FI17,FI50,FI51,FI52,FI53,FI54,FI55,FI56,FI57 h tests | ag '__future'`.

  `flake8-future-import` can also be used later to check / enforce removal of future imports once we've moved to Python 3.

- [ ] Run [`2to3`](https://docs.python.org/2/library/2to3.html) and make sure that we've done or are happy with ignoring all the changes that it wants to do.

  Note that we don't necessarily want to do everything that `2to3` suggests because it outputs Python 3 code that isn't necessarily Python 2-compatible.

  Specifically the command to run is:
  
      2to3 -f all -f idioms -f ws_comma -x dict -x future -x map -x xrange h tests

  `2to3` things I've disabled in this command:
  
  * `-x dict`: `2to3` wants to do a lot of wrapping things like `dict.values()`, `keys()` and `items()` in `list()` because those methods return iterators in Python 3 (rather than lists as in Python 2). But in all cases where `2to3` wants to do this in h I think we'll be fine with iterators.
  
  * `-x future`: `2to3` wants to remove `from __future__ import`s which we don't want to remove until after the move to Python 3. There is a checkbox for this below.

  * `-x map`: Don't wrap `map()` in `list()`.
    
  * `-x xrange`: `2to3` wants to replace `xrange` with `range` but in our case `xrange` is almost always `h._compat.xrange` and we don't want to replace those until after the move to Python 3. There are some instances of actual `xrange` in our code though, we need to replace these manually. There are checkboxes for both of these things below.

- [ ] It may also be worth looking into what [`modernize`](https://python-modernize.readthedocs.io/en/latest/) suggests.

- [ ] Pylint is something that can check for lots of Python 3 problems and can enforce that they stay fixed. Our codebase is currently in bad shape with pylint: we have a lot of warnings.

  Maybe we should do the work to get pylint-client before we pull the trigger on Python 3?
  
  Failing that, someone should at least look through the pylint warnings that we currently have for any Python 3 problems.

Some specific things I saw, I think either `futurize` or `2to3` will probably get these:
    
- [ ] In `websocket.py` `im_self` needs to change to `__self__`. But I'm not sure whether `__self__` works in Python 2? May need to watch out for 2-compatibility here.

- [ ] Replace all real uses of `xrange` (the actual one, not `h._compat.xrange`) with `range`. There are some real uses of `xrange` in migrations for example. May be some in tests.

- [ ] Replace `has_key` with `in`. I only saw this once, in a migration.

- [ ] Look at usages of `ascii`, `map`, `filter`, `hex`, `oct` and `zip`. These are different in Python 3. Haven't looked into this yet. We may want to use [future_builtins](https://docs.python.org/2/library/future_builtins.html).

seanh:
Additionally, I think we should make a separate card listing cleanup tasks to be done after we've successfully moved to Python 3. To get started:

- [ ] Remove _compat.py. (First need to replace uses of each `_compat.py` feature with the plain Python 3 version throughout the code.)
- [ ] Remove now-redundant `__future__` imports: `2to3 -f future h tests`.
- [ ] Update and test developer install docs
- [ ] Remove no-longer-neccessary inheriting from `object`
- [ ] Replace uses of `set()` with set literals (`2to3 -nw -f set_literal h tests`)

robertknight:
Thanks for the detailed investigation @seanh. I hadn't responded yet because I wanted to focus on other development tasks this week.

I personally don't think it is absolutely _necessary_ to do everything on that list before we switch over, in particular updating all dependencies (which has its own risks) and trying to get pylint clean. However I'm happy to let you noodle a way for a while and see what issues you can find. I'm sure it will be educational 🙂

robertknight:
This may be useful to resolve the issue with supervisord still requiring Python 2 to be installed: https://github.com/ochinchina/supervisord

seanh:
I've deleted the big checklist comments above and moved them to a new issue: <https://github.com/hypothesis/h/issues/5384>

PR #4903, opened 2018-03-23
Add SVGRenderer

seanh:
Add `SVGRenderer`, a custom Pyramid renderer. See: https://docs.pylonsproject.org/projects/pyramid//en/latest/narr/renderers.html

This is a renderer that can be used by any view that wants to return an SVG image. The view only needs to return the SVG file content as a string, and the renderer will handle setting the appropriate HTTP headers for an SVG file.

This pr just adds the renderer and tests. No production code actually uses the renderer yet. For how this can be used in a view see https://github.com/hypothesis/h/pull/4896

### DIFF
--- a/h/renderers.py
+++ b/h/renderers.py
@@ -8,5 +8,31 @@ import pyramid.renderers
 json_sorted_factory = pyramid.renderers.JSON(sort_keys=True)
 
 
+class SVGRenderer(object):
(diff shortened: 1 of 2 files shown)


seanh:
There's no caching here because that's configured on the views, not on the renderer.

PR #4905, opened 2018-03-26
Build static assets in separate stage.

jmcarp:
This shaves about 120mb off the docker image: ~20mb from dropping the node runtime (which afaict is only used here to build assets, but lmk if not), and ~100mb from dropping `node_modules`.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -57,7 +57,7 @@ COPY . .
 RUN [ -d .git ] && chown -R hypothesis:hypothesis .git || :
 
 # Copy frontend assets.
-COPY --from=build /build .
+COPY --from=build /build build
 
 # Expose the default port.
 EXPOSE 5000
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,3 +1,11 @@
+FROM node:alpine as build
+
+ENV NODE_ENV production
+
+COPY . .
+
+RUN npm install --production && npm run build
+
 FROM alpine:3.7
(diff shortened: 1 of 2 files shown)


robertknight:
Thanks! - I didn't know about the multi-stage build feature. It looks like this was added quite recently, in Docker 17.05, according to https://docs.docker.com/develop/develop-images/multistage-build/. Fortunately that is exactly the version we have on our Jenkins CI server.

Dockerfile
```diff
@@ -49,9 +56,8 @@ COPY . .
 # If we're building from a git clone, ensure that .git is writeable
 RUN [ -d .git ] && chown -R hypothesis:hypothesis .git || :
 
-# Build frontend assets
-RUN npm install --production \
-  && NODE_ENV=production node_modules/.bin/gulp build
+# Copy frontend assets.
+COPY --from=build /build .
```
robertknight: On my system with Docker version 17.12.0-ce, build c97c6d6, this puts the contents of the `build` directory in the root of the Docker image, instead of in `./build`.

Running the resulting image with `make run-docker` results in a non-working web-service (visiting http://localhost:5000 produces an nginx gateway error) due to h's inability to read `build/manifest.json`. After changing this line to:

```dockerfile
COPY --from=build /build build
```

The resulting Docker image worked as expected.

robertknight:
I'm running this through a build on our Jenkins CI/CD server before merging.

PR #4907, opened 2018-03-26
Remove model-level validation of SVGs

seanh:
Since organizations and their logos can only be created by developers
currently I don't think this is needed, and this validation currently
prevents valid SVG values from being set: many SVG files start with an
`<?xml>` tag not an `<svg>` tag.

If we do have a form for uploading SVGs in the future I'm not sure that
the model will be the right place to validate them. For example we might
want to have a separate validation function that's called by the view
before the SVG is ever passed on to the service or model layers.

### DIFF
--- a/h/models/organization.py
+++ b/h/models/organization.py
@@ -2,7 +2,6 @@
 
 import sqlalchemy as sa
 import slugify
-from xml.etree import ElementTree
 
 from h.db import Base
 from h.db import mixins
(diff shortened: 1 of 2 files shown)


robertknight:
@hmstepanek resolved this in https://github.com/hypothesis/h/pull/4914 and https://github.com/hypothesis/h/pull/4912. On that basis I hope you don't mind me closing this for now.

I can see your point that we might want to do validation in a different place in future. I suggest we wait until a motivating use case comes along and then adapt the code as necessary.

PR #4908, opened 2018-03-26
Remove organization slugs

seanh:
These aren't used currently. They would be used at the ends of URLs
(like we do for groups: https://hypothes.is/groups/<PUBID>/hypothesis-reading)
but I don't think we know yet whether organizations will have web pages
or what we'll want the URLs of those pages to look like (and whether
they'll include slugs).

We can easily add this back later if we decide we want it.

### DIFF
--- a/h/models/organization.py
+++ b/h/models/organization.py
@@ -1,7 +1,6 @@
 # -*- coding: utf-8 -*-
 
 import sqlalchemy as sa
-import slugify
 from xml.etree import ElementTree
 
 from h.db import Base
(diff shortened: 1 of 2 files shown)


h/models/organization.py
```diff
@@ -60,4 +54,4 @@ def validate_logo(self, key, logo):
         return logo
 
     def __repr__(self):
-        return '<Organization: %s>' % self.slug
+        return '<Organization: %s>' % self.pubid
```
robertknight: Testing in the shell I noted that `str(...)`-ing an Organization previously failed before the organization has been saved.

PR #4909, opened 2018-03-26
Update postcss, postcss-url, uglifyify

robertknight:
Update frontend build dependencies which have or depend upon packages
with vulnerabilities according to GitHub.

As a side benefit, this fixes an issue where the Docker build process
would spew a ton of sourcemap-related log messages while building
frontend assets.

The easiest way to test is to kill the local h dev server, run `make docker; make run-docker` and verify that frontend assets (CSS, JS) function at http://localhost:5000


robertknight:
Thanks for taking a look Lyza. It looks like the `run-docker` task needs updating to play nicer with `docker-compose.yml`-orchestrated services.

PR #4910, opened 2018-03-26
Add character limit indicators to group create/edit view

robertknight:
Add the same character limit indicators to the admin group creation view
as already exists for the user-facing group create/edit views.

In future it might make sense to unify the two sets of schemas but for now it was easier to just add the relevant extra bits to the admin groups schemas.

----

screenshot 2018-03-26 15 54 19

### DIFF
--- a/h/schemas/admin_group.py
+++ b/h/schemas/admin_group.py
@@ -3,7 +3,7 @@
 from __future__ import unicode_literals
 
 import colander
-from deform.widget import SelectWidget, SequenceWidget, TextAreaWidget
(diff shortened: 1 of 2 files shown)


robertknight:
> Would it be possible to add a test for this? Otherwise LGTM (if testing irrelevant, go ahead and merge as-is).

I think a unit test would be limited to basically restating the declarative config here, which doesn't provide a lot of value IMO. The character limit controller does already have unit tests.

What might be useful to have is a "forms test page" which uses a variety of input types and features, for which we could have a browser-based test load it up and poke around at it, or even a Python functional test which fetches the page and pokes at the rendered HTML. I think that's a big enough project that it needs to be tackled separately.

PR #4912, opened 2018-03-26
Allow none in validate logo

hmstepanek:
The validate logo does not handle None right now as a valid input option and it should. This fixes that.

### DIFF
--- a/h/models/organization.py
+++ b/h/models/organization.py
@@ -47,6 +47,9 @@ class Organization(Base, mixins.Timestamps):
 
     @sa.orm.validates('logo')
     def validate_logo(self, key, logo):
+        # If the logo is None return early.
+        if logo is None:
+            return logo
         if not (len(logo) <= ORGANIZATION_LOGO_MAX_CHARS):
             raise ValueError(
                 'logo must be less than {max} characters long'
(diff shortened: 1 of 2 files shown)


h/models/organization.py
```diff
@@ -47,16 +47,18 @@ def validate_name(self, key, name):
...
-                'logo must be less than {max} characters long'
-                .format(max=ORGANIZATION_NAME_MAX_CHARS))
-        try:
-            root = ElementTree.fromstring(logo)
-        except ElementTree.ParseError:
-            raise ValueError('logo is not a valid SVG (could not parse XML)')
-        if root.tag != 'svg':
-            raise ValueError('logo is not a valid SVG (does not start with an <svg> tag')
+        # If the logo is not None then it must be an svg format.
+        if logo:
```
robertknight: I personally find it easier to read functions written like this:

```py
if some_condition:
  # early exit for alternate path
  return foo

# lots of code for the "happy path"
```

Rather than:

```py
if some_condition:
  # lots of code for the happy path
else:
  # alternate path
```

The former reduces indentation and makes it easier to follow the logic for the `else` path as it keeps the test & consequence together.

hmstepanek:
Refactored per Rob's request.

PR #4913, opened 2018-03-27
Add world organization creation to db init

hmstepanek:
This adds a default organization with pubid `__world__` to the db init.
All groups will belong to this organization unless otherwise
specified. (At least this will be the case in the near future.) This is a
fix for https://github.com/hypothesis/product-backlog/issues/555.

### DIFF
--- a/h/db/__init__.py
+++ b/h/db/__init__.py
@@ -65,6 +65,7 @@ def init(engine, base=Base, should_create=False, should_drop=False, authority=No
         base.metadata.create_all(engine)
 
     _maybe_create_world_group(engine, authority)
+    _maybe_create_default_organization(engine, authority)
 
 
 def make_engine(settings):
@@ -131,6 +132,23 @@ def _maybe_create_world_group(engine, authority):
                                    writeable_by=WriteableBy.authority)
         world_group.pubid = '__world__'
         session.add(world_group)
+
+    session.commit()
+    session.close()
+
+
+def _maybe_create_default_organization(engine, authority):
+    from h import models
+    session = Session(bind=engine)
+    default_org = session.query(models.Organization).filter_by(pubid='__default__').one_or_none()
+    if default_org is None:
+        default_org = models.Organization(name=u'Hypothesis',
+                                          authority=authority,
+                                          pubid='__default__',
+                                          )
+        with open('h/static/images/icons/logo.svg', 'rb') as h_logo:
+            default_org.logo = h_logo.read().decode("utf-8")
+        session.add(default_org)
 
     session.commit()
     session.close()


hmstepanek:
* This cannot be merged until https://github.com/hypothesis/h/pull/4912 is merged.

h/db/__init__.py
```diff
@@ -136,6 +137,25 @@ def _maybe_create_world_group(engine, authority):
...
+    session = Session(bind=engine)
+    world_org = session.query(models.Organization).filter_by(pubid='__world__').one_or_none()
+    if world_org is None:
+        world_org = models.Organization(name=u'Hypothesis',
+                                        authority=authority,
+                                        pubid = '__world__',
+                                        )
+        with open(r'h/static/images/icons/logo.svg', 'r') as h_logo:
+            # Skip the first line in the file as it's the xml version info.
+            h_logo.readline()
```
robertknight: You shouldn't need to skip the first line here. The XML declaration tag is a proper part of the XML document and both the browser and `ElementTree` should accept it.

h/db/__init__.py
```diff
@@ -136,6 +137,25 @@ def _maybe_create_world_group(engine, authority):
...
+def _maybe_create_world_organization(engine, authority):
+    from h import models
+    session = Session(bind=engine)
+    world_org = session.query(models.Organization).filter_by(pubid='__world__').one_or_none()
+    if world_org is None:
+        world_org = models.Organization(name=u'Hypothesis',
+                                        authority=authority,
+                                        pubid = '__world__',
+                                        )
+        with open(r'h/static/images/icons/logo.svg', 'r') as h_logo:
```
robertknight: What was the `r` prefix for here? The string doesn't contain any escape sequences. I guess that would have been necessary on Windows systems when using backslashes in paths?

hmstepanek: Yes. It's kinda muscle memory at this point. I 'spose it's not strictly necessary tho.

hmstepanek:
Changes addressed PTAL.

seanh:
I worry a little that this will break (and no test failing) if `h/static/images/icons/logo.svg` ever got moved or removed. Just hardcoding the SVG into the Python code might be safer.

What do you think about `__default__` instead of `__world__`? I know groups use `__world__` but I'm not really sure what "world" is supposed to mean (in either context groups or organizations).

h/db/__init__.py
```diff
@@ -136,6 +137,23 @@ def _maybe_create_world_group(engine, authority):
...
+    from h import models
+    session = Session(bind=engine)
+    world_org = session.query(models.Organization).filter_by(pubid='__world__').one_or_none()
+    if world_org is None:
+        world_org = models.Organization(name=u'Hypothesis',
+                                        authority=authority,
+                                        pubid='__world__',
+                                        )
+        with open('h/static/images/icons/logo.svg', 'r') as h_logo:
+            world_org.logo = h_logo.read()
```
seanh: Oops, this `open()` calls reads the SVG string in from the file as bytes, but `world_org.logo` is unicode. This produces a warning from sqlalchemy:

```
.virtualenvs/h/local/lib/python2.7/site-packages/sqlalchemy/sql/sqltypes.py:219: SAWarning: Unicode type received non-unicode bind param value '<?xml version="1.0" encod...'. (this warnin
g may be suppressed after 10 occurrences)                                                                                                                                                              
  (util.ellipses_string(value),))
```

Inlining the SVG into the source code files as a unicode string (`u"..."`) would fix this. Otherwise I think you need to call `h_logo.decode(encoding)` with the correct encoding value (whatever the encoding of this SVG file is, probably `"utf-8"`).

seanh: I know there are no tests for the existing code in this file, but I think it might be worth adding unit tests, at least for the new function. There's stuff in this function that could potentially break when we move from Python 2 to 3 for example.

hmstepanek: Good point about the unicode. I share your concern about opening the file. I'm torn because it feels wrong to have it repeated in two different places but it also feels unsafe to open a file like that. I also had a thought that once we get the serving of the logo's as urls up and running we could potentially remove that svg file and simply replace all references to the file with a url to avoid having it specified in multiple places.

hmstepanek: Also `__default__` makes more sense I was just trying to be consistent.

robertknight: The functional and unit tests execute this code during setup. Although not unit tested directly, the tests will fail if the file path is wrong or the code fails for some other reason.

The logo SVG is UTF-8. In Python 3 one could use `open('...', encoding='utf-8')`. For Python 2 compatibility you'll need to use `decode` on the result of the `read()` call.

hmstepanek:
* Fixed the unicode issue.

h/db/__init__.py
hmstepanek: That's true, this does run on startup although there is no test that currently exercises/uses this default org so that could be a problem for a couple weeks.

seanh: The string encoding thing here is tricky enough that it might be worth adding a specific unit or functional test for it, rather than relying on unrelated functional tests crashing incidentally. (Or just put the SVG in the source code file and ignore the rest of this comment.)

I think `open(path, "r")` actually returns unicode rather than bytes in Python 3 by default, even without an `encoding` argument. If you don't give an `encoding` in Python 3 it defaults to some platform-dependent value (probably `'utf-8'`). If you want to read bytes from a file in Python 3 you have to use `"rb"`.

We'll have to watch out for this when we upgrade from Python 2 to 3. If we're reading in `"r"` mode and getting a byte string from the file in Python 2 and then calling `decode()` on that byte string, Python 3 will give us a unicode string instead and then the call to `decode()` will crash because (unlike in Python 2) unicode strings don't have a `decode()` method in Python 3.

Presumably there would be a functional test failing in Python 3 once we add the `decode()`.

I think the solution (to getting this to work in both Python 2 and 3 at the same time) is to read in `"rb"` mode and then call `decode()`:

```python
with open('h/static/images/icons/logo.svg', 'rb') as h_logo:
    world_org.logo = h_logo.read().decode('utf-8')
```

(And then once we no longer support Python 2 we could change the `'rb'` back to `'r'`, add `encoding='utf-8'`, and remove the `decode()`.)

seanh: (Just saw that @hmstepanek updated the code to add the `decode()` and the tests are failing in Python 3 for exactly that reason: Python 3 returns a unicode string from the file and it doesn't have a `decode()` method.)

hmstepanek: Apparently you can do either that or import io and use it's open method.
```python
# Python 2 and 3: alternative 1
from io import open
f = open('myfile.txt', 'rb')
data = f.read()              # as bytes
text = data.decode('utf-8')  # unicode, not bytes

# Python 2 and 3: alternative 2
from io import open
f = open('myfile.txt', encoding='utf-8')
text = f.read()    # unicode, not bytes
```

PR #4914, opened 2018-03-27
Ignore xmlns in svg tag when validating org's logo

hmstepanek:
Xmlns are inserted into the tag like {http://svg.com}svg which
causes the tag == 'svg' compare to fail. A strip_xmlns function
was added to strip the xmlns from the tag before comparing to
mitigate this issue.

### DIFF
--- a/h/models/organization.py
+++ b/h/models/organization.py
@@ -13,6 +13,12 @@ ORGANIZATION_NAME_MAX_CHARS = 25
 ORGANIZATION_LOGO_MAX_CHARS = 10000
 
 
+def _strip_xmlns(tag):
+    if '}' in tag:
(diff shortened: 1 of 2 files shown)


h/models/organization.py
```diff
@@ -13,6 +13,12 @@
 ORGANIZATION_LOGO_MAX_CHARS = 10000
 
 
+def _strip_xmlns(tag):
```
hmstepanek: I never know what to do with little helper functions like this??

robertknight: This is absolutely fine and follows conventions in the codebase. If testing the functionality through the module's public interface ends up being difficult then it would be worth either making it public or moving to a utility module.

PR #4922, opened 2018-03-28
Fix pytest deprecations.

jmcarp:
* As of pytest 3.0, `yield_fixture` is deprecated in favor of plain old `fixture`
* As of pytest 2.5.2, the `py.test` command is deprecated in favor of `pytest`

Neither of these updates should change behavior--just updating to latest usage.


tests/functional/conftest.py
```diff
@@ -32,15 +32,15 @@ def app(pyramid_app, db_engine):
     return TestApp(pyramid_app)
 
 
-@pytest.yield_fixture(scope='session')
+@pytest.fixture
```
robertknight: The change of scope here causes the functional tests to fail because `db_engine` is now created once per test function instead of once for the whole session.

The functional tests are only run on CI builds by our Jenkins server, not Travis, which unfortunately are only visible to people at Hypothesis. You can run them locally using `tox -e functional`.

robertknight:
LGTM. Thank-you!

PR #4923, opened 2018-03-28
[WIP] Update to elastic 6.x.

jmcarp:
Sorry for the drive-by patch, but I saw https://github.com/hypothesis/product-backlog/issues/562 and thought it would be a fun way to learn about how hypothesis search works. This is just a start, but functional tests Work On My Machine™.

* Upgrade elastic client and image
* Update and/or/not to must/should/must_not
* Drop deprecated path and index_name properties
* Replace string fields with text or keyword

### DIFF
--- a/h/nipsa/search.py
+++ b/h/nipsa/search.py
@@ -33,7 +33,7 @@ def nipsa_filter(group_service, user=None):
     # If any one of these "should" clauses is true then the annotation will
     # get through the filter.
     should_clauses = [
-        {"not": {"term": {"nipsa": True}}},
+        {"bool": {"must_not": {"term": {"nipsa": True}}}},
         {"exists": {"field": "thread_ids"}},
     ]
 
--- a/h/search/config.py
+++ b/h/search/config.py
@@ -19,70 +19,63 @@ from elasticsearch.exceptions import NotFoundError, RequestError
 log = logging.getLogger(__name__)
 
 ANNOTATION_MAPPING = {
-    '_id': {'path': 'id'},
-    '_source': {'excludes': ['id']},
-    'analyzer': 'keyword',
+    '_source': {'excludes': ['_id']},
     'properties': {
-        'annotator_schema_version': {'type': 'string'},
-        'authority': {'type': 'string', 'index': 'not_analyzed'},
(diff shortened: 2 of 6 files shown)


robertknight:
Thanks for the input @jmcarp !

The main challenge here for us I think is figuring out how to handle migration of the production ES index. We have a task that can regenerate the index, which takes about 2 hours to run in production. If we can do the migration without service downtime that would be great, but it isn't the end of the world if a couple of hours downtime makes the process substantially easier.

What I think would be quite helpful is if we could get h to a state where it can talk to either ES 1.5 or ES 6.x. Then the migration process could be as follows:

1. Deploy h, able to talk to ES 1.5 / 6.x, but still using an ES 1.5 instance
2. Reindex contents of DB into a new ES 6.x instance
3. Re-deploy h, pointing at the new ES instance

So, of the changes that you have here, what would be useful to figure out is:

1. Are there any changes which are backwards-compatible with ES 1.5 that we could just apply now?
2. Is there a way for us to detect the ES version at runtime and switch between different schema clauses based on that? If not, we might need some config var.

jmcarp:
I wonder if this can be a little simpler. How about this:

1. `h` is running the `master` branch on some host
1. Clone `h` to a different directory on the host and check out `upgrade-elastic`
1. Reindex to an elastic 6.x instance
1. Merge `upgrade-elastic` into `master`, pull, and restart service, pointing at 6.x instance
1. Run a catch-up migration to cover any records added or updated during reindex

Does that make sense? If some variant on that process would work, the app wouldn't have to talk to multiple elastic versions at the same time--although we can definitely do that if necessary.

Separately, I'm wondering how to test this thoroughly. From what I've seen, the unit tests mock elastic and make assertions about what we send to it, and the functional tests create real indices and run a few incidental queries--meaning it would be easy to generate regressions with this upgrade. Maybe we can run a better test by generating some plausible data, indexing into both 1.x and 6.x, and comparing results.

jmcarp:
Got it, thanks for context on deployment. I'd be happy to revise to work with multiple elastic versions, but first I think it would be useful to verify that search behavior is the same on both versions. I'll try to dig into that over the weekend.

PR #4925, opened 2018-03-28
Add ability to filter groups by name in admin form

robertknight:
In h instances with a large number of groups it becomes impractical to
find a specific group by paging through the list, so add a facility to
search for groups by name.

The search currently does a naive `LIKE` SQL query which should be OK
for the admin page. If we were to expose this to users we might need to
optimize it using a suitable index.

Fixes https://github.com/hypothesis/product-backlog/issues/553

screenshot 2018-03-28 16 47 34

screenshot 2018-03-28 16 47 24

### DIFF
--- a/h/views/admin_groups.py
+++ b/h/views/admin_groups.py
@@ -4,6 +4,7 @@ from __future__ import unicode_literals
 from jinja2 import Markup
 from pyramid.view import view_config, view_defaults
 from pyramid.httpexceptions import HTTPFound, HTTPNotFound
+from sqlalchemy import func
 
 from h import form  # noqa F401
 from h import i18n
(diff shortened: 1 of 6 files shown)


h/views/admin_groups.py
```diff
@@ -23,7 +24,16 @@
...
+    q = request.params.get('q')
+
+    filter_terms = []
+    if q:
+        name = models.Group.name
+        filter_terms.append(func.lower(name).like('%{}%'.format(q.lower())))
+
+    return request.db.query(models.Group) \
+                     .filter(*filter_terms) \
+                     .order_by(models.Group.created.desc())
```
seanh: Minor PEP8 point - this is preferred:

```
 return (request.db.query(models.Group)
                    .filter(*filter_terms)
                    .order_by(models.Group.created.desc()))
```

robertknight: Indeed. Thanks for the tip!

PR #4928, opened 2018-03-29
Add admin CRUD views for managing organizations

robertknight:
Add views for:

 - Listing organizations
 - Creating and editing organizations
 - Deleting organizations

Since our form library does not have a file upload widget, the logo
input is just a text area where the user is expected to paste the SVG
content of the logo.

Organizations can only be deleted if they are not associated with any
groups.

Fixes https://github.com/hypothesis/product-backlog/issues/522


robertknight:
The `test_it_returns_filtered_orgs` test failed on Travis but I'm not able to reproduce locally. I'll look into this next week.

robertknight:
> The test_it_returns_filtered_orgs test failed on Travis but I'm not able to reproduce locally. I'll look into this next week.

This was happening because https://github.com/hypothesis/h/pull/4913 was merged into the DB since I forked my branch and Travis was testing a merge of this branch with master.

h/views/admin_organizations.py
```diff
@@ -0,0 +1,148 @@
...
+    def __init__(self, request):
+        self.schema = OrganizationSchema().bind(request=request)
+        self.request = request
+        self.form = request.create_form(self.schema,
+                                        buttons=(_('Create organization'),))
+
+    @view_config(request_method='GET')
+    def get(self):
+        self.form.set_appstruct({
+            'authority': self.request.authority,
```
lyzadanger: Thanks for pre-populating this!

h/views/admin_organizations.py
```diff
@@ -0,0 +1,148 @@
...
+    def post(self):
+        def on_success(appstruct):
+            authority = appstruct['authority']
+            logo = appstruct['logo']
+            name = appstruct['name']
+            org = Organization(authority=authority, name=name, logo=logo)
+
+            self.request.db.add(org)
+            self.request.session.flash(
+                Markup('Created new organization {}'.format(name)), 'success')
```
lyzadanger: i18n here?

h/views/admin_organizations.py
```diff
@@ -0,0 +1,148 @@
...
+        return HTTPFound(
+            location=self.request.route_path('admin_organizations'))
+
+    @view_config(request_method='POST')
+    def update(self):
+        org = self.org
+
+        def on_success(appstruct):
+            org.name = appstruct['name']
+            org.logo = appstruct['logo']
```
lyzadanger: Naivete-with-SQLAlchemy alert! I was looking at the code here trying to spot the place where the db gets updated, but I presume that merely updating/dirtying the model object here will cause the changes to be persisted, correct?

tests/h/schemas/admin_organization_test.py
```diff
@@ -0,0 +1,27 @@
+from __future__ import unicode_literals
+
+import colander
+import pytest
+
+from h.schemas.admin_organization import OrganizationSchema
+
+
+class TestOrganizationSchema(object):
+    def test_it_allows_valid_data(self, org_data, bound_schema):
```
lyzadanger: Is there any value for testing for name validation (too short, too long?)?

tests/h/views/admin_organizations_test.py
```diff
@@ -0,0 +1,175 @@
...
+                               'authority': org.authority,
+                               'logo': '<svg>new logo</svg>'})
+        handle_form_submission.side_effect = call_on_success
+        ctrl = OrganizationEditController(org, pyramid_request)
+
+        ctx = ctrl.update()
+
+        assert org.name == 'Updated name'
+        assert org.logo == '<svg>new logo</svg>'
+        assert ctx['form'] == self._expected_form(org)
```
lyzadanger: Of course! These test classes are just classes after all; wrapping up stuff in methods is a completely reasonable way to do these sorts of things (it had never occurred to me before!).

h/templates/admin/organizations_edit.html.jinja2
```diff
@@ -0,0 +1,14 @@
+{% extends "h:templates/layouts/admin.html.jinja2" %}
+
+{% set page_id = 'organizations_edit' %}
+{% set page_title = 'Edit organization' %}
+
+{% block content %}
+  {{ form }}
+  {% if delete_url %}
```
lyzadanger: The positioning of this delete button is a bit wonky for me:

image

h/views/admin_organizations.py
robertknight: Yes. Changes to model objects get committed at the end of the web request (see [my writeup of how this works](https://robertknight.me.uk/posts/pyramid-db-transactions/)) if they have not already been explicitly flushed.

The main reason for having explicit flushes in our code is to get the value of some field (eg. a new group's `pubid` or most objects' `id` fields) which is only generated when the object is saved.

h/templates/admin/organizations_edit.html.jinja2
robertknight: Yep. Zero points for beauty here. My plan was to get it in production and solicit some suggestions, hoping that we can fix the other views (OAuth clients, groups) at the same time.

h/views/admin_organizations.py
robertknight: Good catch.

h/schemas/admin_organization.py
```diff
@@ -0,0 +1,42 @@
...
+        colander.String(),
+        title=_('Name'),
+        validator=validators.Length(ORGANIZATION_NAME_MIN_CHARS, ORGANIZATION_NAME_MAX_CHARS),
+        widget=TextInputWidget(max_length=ORGANIZATION_NAME_MAX_CHARS),
+    )
+
+    logo = colander.SchemaNode(
+        colander.String(),
+        title=_('Logo'),
+        hint=_('Logo in SVG format'),
```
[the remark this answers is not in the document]
robertknight: Yes. I'll add a note explaining in a bit more detail.

h/resources.py
```diff
@@ -114,7 +115,12 @@ def __init__(self, request):
 
     def __getitem__(self, pubid):
         try:
-            return self.request.db.query(Organization).filter_by(pubid=pubid).one()
+            org = self.request.db.query(Organization).filter_by(pubid=pubid).one()
+
+            # Inherit global ACL. See comments in `AuthClientFactory`.
+            org.__parent__ = Root(self.request)
```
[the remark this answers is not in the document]
robertknight: Correct. Without this ACL inheritance, Pyramid will find the context object for the request (an `Organization`), look for a `__acl__` property on it, not find one, and then reject the view because the user "doesn't have the `admin_organizations`" permission.

tests/h/schemas/admin_organization_test.py
robertknight: The minimum length is one char, so it can't be too short. It can be too long however. I'll add a test for this.

robertknight:
I've addressed the comments except for the navbar and delete button. I agree that the navbar issue needs to be resolved. I'll work on that in a separate PR.

PR #4933, opened 2018-03-30
Use default organization in test factories

seanh:
Tests that use these factories will be crashing once `Group.organization` is not nullable. Fix the factories to use the default organization.

### DIFF
--- a/tests/common/factories/group.py
+++ b/tests/common/factories/group.py
@@ -12,6 +12,11 @@ from .group_scope import GroupScope
 from .user import User
 
 
+def default_organization():
+    from tests.common.factories.base import SESSION
+    return models.Organization.default(SESSION)
+
+
 class Group(ModelFactory):
 
     class Meta:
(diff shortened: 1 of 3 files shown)


tests/common/factories/group.py
```diff
@@ -12,6 +12,11 @@
 from .user import User
 
 
+def default_organization(obj):
+    from tests.common.factories.base import SESSION
+    return models.Organization.default(SESSION)
```
seanh: Bit of a hack to have to import `SESSION` here but I don't see any other way

tests/common/factories/group.py
```diff
@@ -12,6 +12,11 @@
 from .user import User
 
 
+def default_organization():
+    from tests.common.factories.base import SESSION
+    return models.Organization.default(SESSION)
```
seanh: Bit of a hack to have to import SESSION here but I don't see any other way

PR #4939, opened 2018-03-30
Make Group.organization not nullable

seanh:
This is marked as WIP for now just because there are several other open prs that should be merged before this is, plus one db migration that should be run, to prevent this from breaking things.

I _think_ it's okay to deploy the model change and migration script at the same time and then run the migration script. Safest to deploy the code all the way out to production before running the migration.

### DIFF
--- a/h/migrations/versions/f052da9df33b_make_group_organization_id_not_nullable.py
+++ b/h/migrations/versions/f052da9df33b_make_group_organization_id_not_nullable.py
@@ -1,0 +1,16 @@
+"""Make group.organization_id not nullable."""
+from __future__ import unicode_literals
+
+from alembic import op
+
+
+revision = 'f052da9df33b'
+down_revision = '615358b6c428'
+
+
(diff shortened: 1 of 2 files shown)


h/migrations/versions/f052da9df33b_make_group_organization_id_not_nullable.py
```diff
@@ -0,0 +1,23 @@
+"""
+Make group.organization_id not nullable.
+
+Revision ID: f052da9df33b
+Revises: 8a7f31c4525d
```
robertknight: This comment needs updating to match the `down_revision` below.

seanh: Fixed

seanh:
@robertknight If this can be approved then I can merge and run it at the same time when I do <https://github.com/hypothesis/h/pull/4938> tomorrow morning

PR #4940, opened 2018-03-30
Inline the default organization logo SVG

seanh:


seanh:
I've seen the file opening in `h/db/__init__.py` crash by failing to find the file, but can't seem to reproduce it. The relative path doesn't always apply. I'll close this but we can use it if the problem resurfaces.

PR #4941, opened 2018-04-01
Search tests

jmcarp:
In preparation for https://github.com/hypothesis/product-backlog/issues/562, and to prevent search regressions in general, it would be useful to include search tests that use a live elastic instance. This patch builds on earlier work by @robertknight, adding more tests and exercising both indexing and query logic. These tests aren't comprehensive, but they're a start.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,7 @@ matrix:
       python: '2.7'
       addons:
         postgresql: "9.4"
-      before_install:
+      before_install: &install-elastic
         # Install Elasticsearch v1.6.2 and the ICU plugin to match the
         # production environment.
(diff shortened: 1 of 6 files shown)


robertknight:
Thanks for the PR @jmcarp. It may be a few days until we're able to give this our full attention as we're currently wrapping up some work we're doing on new kinds of group.

robertknight:
We've finished the aforementioned work and we'll be starting to look into the Elasticsearch upgrade and this PR over the next week. Thanks again for the contribution.

seanh:
Closing in favour of https://github.com/hypothesis/h/pull/5002

PR #4946, opened 2018-04-03
Implement new admin navbar design

robertknight:
The existing `/admin` navbar covered the main content of the page when it wraps
onto multiple lines, and the wrapping happens even on not-very-narrow
screens. This commit implements a revised look for the navbar in the
admin panel which resolves that problem and generally uses horizontal
space more efficiently.

The new look currently explicitly does away with the switch to a hamburger menu at very small screen sizes. This is simply because it isn't that often that the menu needs to be used on a phone-sized screen, although the new layout will still work in that context.

As a bonus, the layout is implemented with flexbox following our CSS
conventions which should make it easier to adapt in future.

----

**Existing navbar**

(Note that it obscures the top of the content below)

screenshot 2018-04-03 15 58 55

**New navbar**

screenshot 2018-04-03 15 57 03

**New navbar (narrow screen)**

(Note that it pushes content below down)

screenshot 2018-04-03 15 57 35

### DIFF
--- a/h/static/styles/partials/_admin-navbar.scss
+++ b/h/static/styles/partials/_admin-navbar.scss
@@ -7,7 +7,7 @@
   color: white;
   display: flex;
   flex-direction: row;
-  font-size: $normal-font-size;
+  font-size: $subtitle-font-size;
   justify-content: center;
   min-height: 50px;
 }
--- a/h/static/styles/partials/_admin-navbar.scss
+++ b/h/static/styles/partials/_admin-navbar.scss
@@ -1,0 +1,96 @@
+/**
+ * The dark navbar shown at the top of `/admin` pages.
+ */
+
+.admin-navbar {
+  background-color: $grey-7;
+  color: white;
+  display: flex;
+  flex-direction: row;
+  font-size: $normal-font-size;
+  justify-content: center;
+  min-height: 50px;
+}
+
+.admin-navbar__container {
+  display: flex;
+  flex-direction: row;
+  align-items: stretch;
+  padding-left: 15px;
+  padding-right: 15px;
+  flex-grow: 1;
+
+  // Width chosen to match Bootstrap's "container" class,
+  // which is used for the main content of the page.
+  max-width: 970px;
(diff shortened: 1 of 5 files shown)


dwhly:
yay.

h/static/styles/partials/_admin-navbar.scss
```diff
@@ -0,0 +1,96 @@
...
+  padding-left: 10px;
+  padding-right: 10px;
+  text-align: center;
+
+  // Give tabs reasonable padding above and below text, even when wrapped onto
+  // multiple lines.
+  min-height: 2em;
+
+  // Color for the tab's link.
+  --link-color: $grey-3;
```
robertknight: The admin console does not need to work in IE, so we can use CSS variables here for styling descendant elements based on the state of a parent element.

robertknight:
OK, I think this is good to go now. The way that the navbar handles a lack of space to display all tabs could be improved, but I think this is a reasonable start and in most cases, we won't need multiple lines.

One other improvement that would be nice, but is not trivial, would be to replace the dropdown menus with our `dropdown_menu` component.

h/static/styles/partials/_admin-navbar.scss
```diff
@@ -0,0 +1,96 @@
...
+
+.admin-navbar__container {
+  display: flex;
+  flex-direction: row;
+  align-items: stretch;
+  padding-left: 15px;
+  padding-right: 15px;
+  flex-grow: 1;
+
+  // Width chosen to match Bootstrap's "container" class,
```
lyzadanger: Thanks for the helpful comments!

robertknight:
I bumped the font size to 15px. The reason for 15 instead of 14 is that we already had "subtitles" elsewhere that were 15px and it is preferable IMO to avoid having too many discrete sizes across the site.

screenshot 2018-04-04 14 52 06

lyzadanger:
Big thumbs up! Thanks for the font-size bump; looks more clear now.

PR #4948, opened 2018-04-04
Enable assigning organizations to groups in admin group created/edit view

robertknight:
Add an "Organization" drop-down menu to the admin group create/edit view which sets the organization associated with a group. This field defaults to the system's default organization (`Organization.default`).

When creating a new group, the user can switch between organizations belonging to all authorities. The organization that the user selects will determine which authority the group belongs to. When editing an existing group the user can only select between organizations belonging to the current authority.


h/schemas/admin_group.py
```diff
@@ -22,17 +22,23 @@
 )
 
 
-def user_exists_validator_factory(user_svc):
```
robertknight: Rather than passing the validator as an argument to the schema constructor, I'm using colander's "[deferred values](https://docs.pylonsproject.org/projects/colander/en/latest/binding.html)" facility to pass values only known at runtime to the validation and widget preparation logic via `schema.bind`.

I'm doing this because AFAICS it is the "standard" way to handle this need in colander.

h/schemas/admin_group.py
```diff
@@ -22,17 +22,23 @@
...
-    def user_exists_validator(form, value):
-        user = user_svc.fetch(value['creator'], value['authority'])
+@colander.deferred
+def group_creator_validator(node, kw):
+    user_svc = kw['user_svc']
+    orgs = kw['organizations']
+
+    def validate(form, value):
```
lyzadanger: Perhaps add a comment or docstring here detailing what the validation is validating? Is it validating that the user exists at the authority associated with the organization selected in the form? (I think that's what's going on!)

h/schemas/admin_group.py
```diff
@@ -47,8 +53,22 @@ def validate(node, value):
     return validate
 
 
+@colander.deferred
+def group_organization_select_widget(node, kw):
+    orgs = kw['organizations']
+    org_names = ['{} ({})'.format(org.name, org.authority) for org in orgs]
```
lyzadanger: Glad to see you display the authority here! Very helpful (arguably totally necessary, even).

h/schemas/admin_group.py
```diff
@@ -47,8 +53,22 @@ def validate(node, value):
...
+    org_pubids = [org.pubid for org in orgs]
+
+    # `zip` returns an iterator in Python 3. The `SelectWidget` constructor
```
lyzadanger: Thank you so much for this comment!

h/services/group.py
```diff
@@ -144,11 +150,15 @@ def _create(self, name, userid, description, type_flags, origins=[], add_creator
...
         :param type_flags: the type of this group
         :param origins: the list of origins that the group will be scoped to
+        :param organization: the organization to associate the group with
         :param add_creator_as_member: if the group creator should be added as a member
         """
         creator = self.user_fetcher(userid)
         scopes = [GroupScope(origin=o) for o in origins]
 
+        if organization is None:
+            organization = Organization.default(self.session)
```
lyzadanger: As I understand it, the default organization is always unwaveringly associated with the `hypothesis` authority, which seems to imply here that any group created through the service with a missing (`None`) organization will get associated via the default-organization relation with `hypothesis` as an organization. This may be fine; just raising it in case it isn't.

h/templates/admin/groups.html.jinja2
```diff
@@ -18,6 +18,7 @@
         <tr>
           <th>Name</th>
           <th>URL</th>
+          <th>Organization</th>
```
lyzadanger: Thanks for adding this here! Reminds me I wanted to add `type`, too (outside the scope of this PR, of course).

h/templates/admin/groups.html.jinja2
```diff
@@ -40,6 +41,10 @@
                 {{ group_url }}
               </a>
             </td>
+            <td>
+              {% set org_url = request.route_url('admin_organizations_edit', pubid=group.organization.pubid) %}
```
lyzadanger: It took me a few experimental clicks in the interface to ascertain that the name column linked to the edit tool, though I don't know if there's any better way to surface that...

h/views/admin_groups.py
```diff
@@ -12,12 +12,29 @@
...
 from h.models.group_scope import GroupScope
+from h.models.organization import Organization
 from h import paginator
-from h.schemas.admin_group import CreateAdminGroupSchema, user_exists_validator_factory
+from h.schemas.admin_group import CreateAdminGroupSchema
 
 _ = i18n.TranslationString
 
 
+def _find_organization(session, pubid):
```
lyzadanger: View logic in general throughout the app seems thicker than I'm used to (just a general observation). I wonder if places like here could benefit from `context`-layer stuff in the future...just a thought.

h/views/admin_groups.py
```diff
@@ -169,6 +194,10 @@ def _update_appstruct(self):
             'group_type': group.type,
             'name': group.name,
             'origins': [s.origin for s in group.scopes],
+
+            # Once `group.organization` becomes non-nullable we can simplify
```
lyzadanger: Do we have a standard keywording (e.g. `TODO` or `FIXME`) for this? It might be helpful to drop that into the comment to catch the eye later.

lyzadanger:
I was able to get the edit-group form into a funky state on my local. I don't know what authority the group in question here is associated with because that field is no longer displayed/editable, but the organizations list is empty, presumably because there are no matches to whatever the authority it:

image

hmstepanek:
I think we are duplicating work here between me and Robert. I had a branch before I left that was adding in orgs into the group service. I'm not sure what to do now but since I've looked into it a little more today I think we need a org fetch just like we have a user fetch which is what I kinda started working on today until it occurred to me that that is very related/useful for admin pages and then I took a look at this PR and noticed all the duplicated work.

h/services/group.py
hmstepanek: I think instead of organization=None we should have organization='`__default__`' and be using the pubid here then using a fetcher like we do for users.

hmstepanek:
This part looks good.
image

tests/h/schemas/admin_group_test.py
```diff
@@ -79,76 +79,66 @@ def test_it_raises_if_no_origins(self, group_data, bound_schema):
...
-    def user_svc(self, pyramid_config):
-        svc = mock.create_autospec(UserService, spec_set=True, instance=True)
-        pyramid_config.register_service(svc, name='user')
-        return svc
-
-
-class TestCreateSchemaWithValidator(object):
-
-    def test_it_passes_creator_and_authority_to_service(self,
+    def test_it_allows_when_creator_exists_at_authority(self,
```
hmstepanek: Is there a test for when the user's authority doesn't match the orgs authority? I don't see one here but I could just be missing it. I find these test names difficult to follow. Seems like we could use some refactoring here to be in a more readable format. Ex: test_validation_passes_when_creator_belongs_to_authority. (And I also realize they were like this before this PR so it's up to you.)

h/schemas/admin_group.py
```diff
@@ -22,17 +22,26 @@
...
+@colander.deferred
+def group_creator_validator(node, kw):
+    """
+    Validate that the creator username exists in the organization's authority.
+    """
+    user_svc = kw['user_svc']
+    orgs = kw['organizations']
+
+    def validate(form, value):
+        org = next(org for org in orgs if org.pubid == value['organization'])
```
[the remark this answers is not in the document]
hmstepanek: Another thought-we could use the index to the list instead of the pubid?

seanh:
I'll take over finishing this off (<https://hypothes-is.slack.com/archives/C4K6M7P5E/p1523540014000555>)

h/views/admin_groups.py
```diff
@@ -165,17 +184,16 @@ def on_success(appstruct):
...
             # field, so the user will have to pick one when editing the group.
             'creator': group.creator.username if group.creator else '',
 
             'description': group.description or '',
             'group_type': group.type,
             'name': group.name,
-            'origins': [s.origin for s in group.scopes],
             'members': [m.username for m in group.members],
+            'organization': group.organization.pubid,
+            'origins': [s.origin for s in group.scopes],
```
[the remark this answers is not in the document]
seanh: That's right - every open and restricted group must have at least one origin.

Unscoped open and restricted groups aren't something that we've designed yet (they would appear in the sidebar all the time for everyone, on every site - how are we going to handle that?)

h/templates/admin/groups.html.jinja2
seanh: Just some nice <samp>Edit</samp> buttons next to the names?

seanh:
@hmstepanek I think this is ready for another review

h/views/admin_groups.py
```diff
@@ -150,6 +168,7 @@ def on_success(appstruct):
             group.description = appstruct['description']
             group.name = appstruct['name']
             group.scopes = [GroupScope(origin=o) for o in appstruct['origins']]
+            group.organization = _find_organization(self.request.db, appstruct['organization'])
```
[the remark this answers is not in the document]
robertknight: That's not what I recall from the discussion around the [ADR when we introduced the service layer](https://github.com/hypothesis/h/blob/master/docs/arch/adr-002.rst#decision). My recollection was that we would use judgement to avoid putting "non trivial" model logic/queries in views, or when there is a clear advantage to testability or reusability.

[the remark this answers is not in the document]
seanh: I think we should tend aggressively to moving as much as possible (sqlalchemy queries, and everything else possible) out of the views, because otherwise views tend to easily get long and complicated, and to have a lot of different responsibilities (receiving and validating data from the request, carrying out business logic, querying the DB, preparing the response, ...) and they also become very difficult to test. A lot of long and difficult to read tests in very long test modules. Witness the admin pages views, but _especially_ the activity pages views and their tests.

I think we should look at views as a thin-as-possible integration layer between the web framework and the rest of our code. Receive the request from the web framework, call other parts of our code (schemas, services, presenters, ...) to do the work, send the response back to the web framework. The views are the things that plug the different parts of our architecture together (taking the return result of schema validation and passing that into a service, etc) and plug our code in to Pyramid's request and response. But they don't do anything else - just dumb integration.

For sqlalchemy queries in particular: sqlalchemy queries are tightly coupled to the structure and characteristics (constraints, performance, ...) of our model classes and database tables and columns. Something like `db.query(models.Group).filter_by(pubid="__world__").one()` contains knowledge such as: `models.Group` exists, groups have a `pubid`, `__world__` is the special pubid of the default group, there's only one default group, retrieving a group by pubid is efficient, etc.

So we need to avoid peppering these queries throughout our entire codebase, otherwise we couple the whole codebase to the DB structure, and making any changes to the DB structure becomes difficult.

So keeping sqlalchemy queries within models and services.

h/templates/admin/groups.html.jinja2
hmstepanek: I had no problem with it.

PR #4950, opened 2018-04-05
Remove redundant comments from migration script template

robertknight:
The `Revises` and `Revision ID` comments duplicate the `revision` and
`down_revision` declarations below and the `Create Date` duplicates info
in the git commit.

Removing these duplicated comments avoids them getting out of sync,
which has happened in a few PRs following rebasing.


seanh:
:+1:

PR #4953, opened 2018-04-06
Change `make lint` to run prospector instead of flake8

seanh:
This gets `prospector` running without any errors, but only by disabling a lot of stuff - pep257, a few pylint warnings, and any linting of tests (the tests have 1650+ warnings even with pep257 and lots of other things disabled), and by ignoring 125 files.

### DIFF
--- a/.prospector.yaml
+++ b/.prospector.yaml
@@ -1,27 +1,39 @@
 output-format: grouped
 strictness: veryhigh
 doc-warnings: true
+test-warnings: false
+member-warnings: false
 max-line-length: 160
-
-# We run flake8 first and then prospector as a second-pass linter.
-# Since flake8 already runs pyflakes, pep8 and mccabe, don't run them again
-# with prospector.
+dodgy:
+  run: true
 mccabe:
-  run: false
+  run: true
 pep8:
-  run: false
+  run: true
+  full: true
+pyflakes:
+  run: true
+profile-validator:
+  run: true
 pep257:
+  run: false
   disable:
+    - D100 # Missing docstring in public module
+    - D101 # Missing docstring in public class
+    - D102 # Missing docstring in public method
+    - D103 # Missing docstring in public function
+    - D104 # Missing docstring in public package
+    - D105 # Missing docstring in magic method
+    - D106 # Missing docstring in nested class
+    - D107 # Missing docstring in __init__
     - D203 # "1 blank line required before class docstring" conflicts with
            # another pep257 rule D211 "No blank lines allowed before class
            # docstring"!
     - D212 # "Multi-line docstring summary should start at the first line"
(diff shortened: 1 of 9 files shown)


seanh:
The main downside of prospector seems to be that it's slow. It takes about 13 seconds to run on my machine, or 41 seconds (!) if you configure it to lint the tests as well.

Most of the slowness comes from pylint. If you disable pylint then the 41 seconds becomes 7 seconds. But pylint is also where most of the benefit (most of the useful warnings) comes from.

Things we could do to make it faster:

* If you're using editor integration, does your editor integration only run the linter on the file being edited? Or does it lint the entire codebase every time? Editor integration could still be fast I think.

* Add a `make quick-lint` that uses git to look for new and changed files (either git staged or not) and runs prospector on those files only.

* Prospector doesn't support flake8. I believe this is because, like prospector, flake8 is not actually a linter but just a wrapper for a number of other linters. Specifically flake8 runs pyflakes, pycodestyle (pep8) and mccabe. And prospector supports each of pyflakes, pycodestyle and mccabe individually. You can run prospector with certain tools only: `prospector --tool pyflakes`. Or `prospector --tool pyflakes --tool pep8 --tool pep257 --tool mccabe`. This overrides what tools are enabled and disabled in `.prospector.yaml`. Takes 9.5 seconds to lint the entire codebase including tests. I think this is equivalent to flake8, need to do a little more research to make sure.

* If you want quick linting locally or in-editor just use flake8 yourself and then either run the slow prospector `make lint` command once before pushing, or let it run on Travis.

I think making linting faster by only linting the modified files is probably more useful than doing it by disabling pylint.

As a long-term aim I'd suggest that we aim for a `make lint` that runs prospector and runs on Travis even though it's slow, and developers who want to be able to run a faster lint or have linting built in to their editor can use one of the above techniques to make it faster.

hmstepanek:
Have you considered using other tools like yapf?

hmstepanek:
@seanh should we close this? Last time we spoke it sounded like we were going to use black instead?

seanh:
Yeah lets close this. We'll still need linting even with Black, as linters catch a lot more than just code formatting stuff. But I think we need to take a different approach to getting stricter linting in place than what this pull request attempted.

What I'm thinking is:

* Have a whitelist of files that're allowed to have linter warnings in them from the stricter linting. This will initially contain almost all files.
* The whitelist might also contain the number of lint warnings per file, so that you can't add more warnings to a whitelisted file.
* Any new files will need to pass strict linting completely as they won't be in the whitelist.
* Over time files warnings can be fixed in files and the files removed from the whitelist, one file at a time.

And we can take a similar approach to getting Black code formatting in place too. But we also need to get Black running in Docker because otherwise Lyza might have difficulty running it (and we should probably run the linters, tests and dev server in docker too).

Issue #4566, opened 2017-05-23
Images in annotation bodies can overflow container on search result pages


robertknight:
Images in annotation bodies can overflow the container on search result pages:

<img width="969" alt="screenshot 2017-05-23 12 08 13" src="https://cloud.githubusercontent.com/assets/2458/26351704/9f3869a4-3fb0-11e7-8cc4-71b52d29fb54.png">

judell:
Is `width:100%` on the `img` sufficient to solve this @dawa?

dawariley:
i think **max-width:100%** might be better as it doesn't stretch the image to be 100% if it's smaller, but just imposes a maximum width. 

![screen shot 2018-03-08 at 22 27 08](https://user-images.githubusercontent.com/5212631/37180419-d349720c-2320-11e8-87aa-9a0403c7c742.png)


i'm going to associate this card with the headline formatting bug card #4790  you created before. i mistakingly conflated the overspilling of text on a QDR annotation as being a headline formatting bug, when in fact it was because the QDR headline was an 'image hack' as @dwhly pointed out to me. 

visit:
http://www.edchange.org/multicultural/speeches/malcolm_x_ballot.html and https://hypothes.is/groups/bR7GAVK2/qdr-ati (click some of the lower cards in the list to reproduce the bug.)

dawariley:
going to move this to the design queue for prioritization. thanks for re-surfacing this one @judell

dwhly:
How can we avoid this too....

![image](https://user-images.githubusercontent.com/1413074/37224275-d607e0e8-2387-11e8-982e-6c475d7852bb.jpg)

judell:
> i think max-width:100% might be better as it doesn't stretch the image to be 100% if it's smaller, but just imposes a maximum width.

I'm noticing that on Twitter, images do stretch. Now admittedly they control the width of elements in the "desktop" web app, and it is not responsive, but  wouldn't it be odd if some images filled the prescribed width and others did not?

dawariley:
Associating this card with a related 'overspilling a container' issue in search results pages: https://github.com/hypothesis/h/pull/3970. 

It would be great to address all overspilling text issues reported in one fell swoop. @dwhly has pointed out a further issue with the meta data overspilling the grey margin 
area.

To summarise there are 3 main issues to fix:

1. Images are overspilling the annotation container - documented in this ticket: 
https://github.com/hypothesis/h/issues/4566

2. Text is overspilling the annotation container - documented in:
[...]

PR #4954, opened 2018-04-07
set max width for images in annotation cards

judell:
Fixes https://github.com/hypothesis/h/issues/4566


robertknight:
On further investigation, it looks like we had a `max-width: 100%` rule already in place for images but it wasn't being applied due to an error with markup structure. This is resolved in https://github.com/hypothesis/h/pull/4961.

PR #4955, opened 2018-04-07
Add organization into group service

hmstepanek:
Add organization as an optional input when creating each of the
different groups. Organization is defaulted to the `__default__`
organization.

### DIFF
--- a/h/services/group.py
+++ b/h/services/group.py
@@ -25,7 +25,7 @@ class GroupService(object):
         self.user_fetcher = user_fetcher
         self.publish = publish
 
-    def create_private_group(self, name, userid, description=None):
+    def create_private_group(self, name, userid, description=None, organization=None):
         """
         Create a new private group.
 
@@ -34,6 +34,8 @@ class GroupService(object):
         :param name: the human-readable name of the group
         :param userid: the userid of the group creator
         :param description: the description of the group
+        :param organization: the organization that this group belongs to
+        :type organization: h.models.Organization
 
         :returns: the created group
         """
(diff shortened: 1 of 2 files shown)


h/services/group.py
```diff
@@ -34,6 +34,8 @@ def create_private_group(self, name, userid, description=None):
         :param name: the human-readable name of the group
         :param userid: the userid of the group creator
         :param description: the description of the group
+        :param organization: the organization that this group belongs to in the form
```
robertknight: `:type organization: h.models.Organization` would be a clearer way of indicating the type rather than "in the form of a model object. See existing uses of `:type` in the code.

h/services/group.py
```diff
@@ -145,10 +155,13 @@ def _create(self, name, userid, description, type_flags, origins=[], add_creator
         :param type_flags: the type of this group
         :param origins: the list of origins that the group will be scoped to
         :param add_creator_as_member: if the group creator should be added as a member
+        :param organization: the organization that this group belongs to in the form
+               of a model object
         """
         creator = self.user_fetcher(userid)
         scopes = [GroupScope(origin=o) for o in origins]
```
robertknight: I think we should validate here that the creator (which determines the group's authority) and organization belong to the same authority.

I was looking for a way to validate this when saving `Group` objects but I couldn't immediately see how to do that.

h/services/group.py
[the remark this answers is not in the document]
hmstepanek: Fixed.

h/services/group.py
hmstepanek: Added in a validation method and corresponding test.

PR #4962, opened 2018-04-10
Unicode literals

seanh:
* Add `unicode_literals` to all Python files
  
  Add `from __future__ import unicode_literals` to all Python files that don't have it yet.
  
  See: https://www.python.org/dev/peps/pep-3112/
  
  This makes our Python 2.7 environment a little more similar to what a Python 3 one will look like: string literals are unicode by default. By doing this as a separate step now we reduce the size of the change that we'll be applying when we move from Python 2 to 3.
  
  This was done using `futurize --unicode-literals -wn1 h tests`.
  
  I did not add `unicode_literals` to:
  
  - `h/cli/*` and its tests, because the click library complains about `unicode_literals` (see: http://click.pocoo.org/5/python3/)
  
  - `h/migrations/versions/*`, because I don't want to make changes to old migration scripts.

* Remove now unnecessary `u`-prefixes from string literals

* Use `make lint` to enforce `unicode_literals`

  Add an additional flake8 command to `make lint` that complains if `from __future__ import unicode_literals` is missing from any files. To make sure that no new files are added without it.

### DIFF
--- a/h/tweens.py
+++ b/h/tweens.py
@@ -8,6 +8,7 @@ from codecs import open
 from pyramid import httpexceptions
 from pyramid.util import DottedNameResolver
 
+from h._compat import native
 from h.util.redirects import parse as parse_redirects
 from h.util.redirects import lookup as lookup_redirects
 
@@ -143,3 +144,40 @@ def cache_header_tween_factory(handler, registry):
         return resp
 
     return cache_header_tween
+
+
+def encode_headers_tween_factory(handler, registry):
+    """
+    Convert HTTP response headers to native strings.
+
+    The WSGI spec (https://www.python.org/dev/peps/pep-3333/) requires all
+    HTTP response header names and values to be "native" strings:
+
+    * Byte strings in Python 2
+    * Unicode strings in Python 3
+
+    Since string literals are byte strings in Python 2 and are unicode strings
+    in Python 3, using string literals for header values (as in
+    ``response.headers["Access-Control-Allow-Origin"] = "*"``) works fine in
+    either Python 2 or 3.
+
+    But once you add ``from __future__ import unicode_literals`` to a source
+    code file the string literals become unicode strings in Python 2, and
+    violate PEP-3333. This violation causes ``AssertionError``s from WebTest,
+    and may cause problems with WSGI servers.
+
+    This tween fixes that by converting all header names and values to native
+    strings.
+
+    TODO: Remove this tween once we no longer support Python 2.
+
+    """
+    def encode_headers_tween(request):
+        resp = handler(request)
+        for key in list(resp.headers.keys()):
+            values = resp.headers.getall(key)
+            del resp.headers[key]
+            for value in values:
+                resp.headers.add(native(key), native(value))
+        return resp
+    return encode_headers_tween
--- a/h/tweens.py
+++ b/h/tweens.py
@@ -1,5 +1,6 @@
 # -*- coding: utf-8 -*-
 
+from __future__ import unicode_literals
 import collections
 import logging
 from codecs import open
--- a/h/tweens.py
+++ b/h/tweens.py
@@ -1,5 +1,4 @@
 # -*- coding: utf-8 -*-
-
 from __future__ import unicode_literals
 import collections
 import logging
--- a/Makefile
+++ b/Makefile
@@ -66,6 +66,7 @@ test-py3: node_modules/.uptodate
 lint: .pydeps
 	flake8 h
 	flake8 tests
+	flake8 --select FI14 --exclude 'h/cli/*,tests/h/cli/*,h/migrations/versions/*' h tests
 
 ################################################################################
 
--- a/Makefile
+++ b/Makefile
@@ -66,7 +66,7 @@ test-py3: node_modules/.uptodate
 lint: .pydeps
 	flake8 h
 	flake8 tests
-	flake8 --select FI14 --exclude 'h/cli/*,tests/h/cli/*,h/migrations/versions/*' h tests
+	flake8 --select FI14 --exclude 'h/cli/*,tests/h/cli/*,h/util/uri.py,h/migrations/versions/*' h tests
 
 ################################################################################
 
--- a/h/__init__.py
+++ b/h/__init__.py
@@ -1,5 +1,6 @@
 # -*- coding: utf-8 -*-
 
+from __future__ import unicode_literals
 from h._version import get_version
 
 __all__ = ('__version__',)
--- a/h/__main__.py
+++ b/h/__main__.py
@@ -1,5 +1,6 @@
 # -*- coding: utf-8 -*-
 
+from __future__ import unicode_literals
 from h.cli import main
 
 if __name__ == '__main__':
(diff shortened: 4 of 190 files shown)


seanh:
Some tests erroring in `paginator_test.py` and some failing in `uri_test.py`. Haven't had a chance to look into it yet. Fixes are likely either `b`-prefixing some strings, or omitting `from __future__ import unicode_literals` from some files.

There's also a bunch of unicode warnings being printed out by the tests, we'll need to fix these.

seanh:
Fixed the `paginator_test.py` errors. Will take a look at the `tests/h/util/uri_test.py` fails another day.

seanh:
Fixed the `uri_test.py` fails. Still getting some unicode warnings from `normalize_uris_test.py`, will look into those another day.

seanh:
Actually those warnings should be fixed now...

seanh:
Ugh, the thing with the functional tests is quite a pain. The easiest fix is just to remove `from __future__ import unicode_literals` each of the 5 files in h (actually source files, not test files) that set HTTP headers. Seems a bit of a shame to have to not use unicode literals in these 5 files, just to keep webtest happy.

But otherwise it seems we have to wrap lots of header names and values in `str()` just to keep webtest happy in Python 2 (in Python 3 the `str()`s do nothing).

I'm not sure whether this issue (of header names and values being unicode, not bytes, in Python 2 when `from __future__ import unicode_literals` is added) is actually a problem in the real world, or just a problem with webtest.

seanh:
Yeah it sucks that flake8 passes without errors if you enable a rule that doesn't exist (because the relevant plugin isn't installed). I don't think there's anything we can do about it. On Travis the plugin _will_ be installed though.

We may be able to fix it by having `make lint` run `flake8` inside of tox - when `requirements-dev.in` changes tox _should_ see that and update its virtualenv. In the past I have noticed tox failing to do this even though it's supposed to, though.

robertknight:
> But otherwise it seems we have to wrap lots of string literals for header names and values in str() just to keep webtest happy in Python 2 (in Python 3 the str()s do nothing).

I think we should just suck it up and do that. We don't have an enormous number of these.

seanh:
I've added a commit that adds the `str()`-wrapping. All tests should be passing now. If we merge this, we should add a note to revert that commit after we've moved to Python 3.

I don't want to merge this yet though. I'm concerned that the `str()`-wrapping that I added doesn't catch all places where we set header names and values. It only catches those places that are hit by the functional tests.

Is this appearance of unicode strings rather than byte strings in HTTP header names and values in Python 2 a real-world problem? Or is it just a problem with webtest? If it's an actual problem then deploying this may break things in Python 2, where code isn't covered by functional tests and `str()`-wrapping is missing.

So I'd like to do a little investigation, another day, into whether this is an actual problem or just a webtest problem.

robertknight:
> Or is it just a problem with webtest? 

I looked into this a bit for 9edc6fbed062ce64efdf84051293c0a7f6b0f212. [PEP 3333](https://www.python.org/dev/peps/pep-3333/#a-note-on-string-types) is where the requirement comes from.

seanh:
While bd381a82b8210d48f8e2f157f6855b2e5b6872e0 gets the functional tests passing (hence why they're passing on this branch) it does so by wrapping string literals in `str()` wherever a response header key or value is set throughout the code. And it may miss some strings, when code that isn't called by the functional tests sets headers.

This tween may be a better approach, as it provides one centralised place to call `str()` on all header keys and values (blindly: whether those headers were set by us or not; whether they're unicode or byte strings). Tested and it works. And then we could simply remove this tween once we're done with Python 2.

```diff
diff --git a/h/app.py b/h/app.py
index ee489b9d5..55f066705 100644
--- a/h/app.py
+++ b/h/app.py
@@ -56,6 +56,7 @@ def includeme(config):
     config.add_tween('h.tweens.invalid_path_tween_factory')
     config.add_tween('h.tweens.security_header_tween_factory')
     config.add_tween('h.tweens.cache_header_tween_factory')
+    config.add_tween('h.tweens.encode_headers_tween_factory')
 
     config.add_request_method(in_debug_mode, 'debug', reify=True)
 
diff --git a/h/tweens.py b/h/tweens.py
index 20f0ed7ab..da713201f 100644
--- a/h/tweens.py
+++ b/h/tweens.py
@@ -143,3 +143,15 @@ def cache_header_tween_factory(handler, registry):
         return resp
 
     return cache_header_tween
+
+
+def encode_headers_tween_factory(handler, registry):
+    def encode_headers_tween(request):
+        resp = handler(request)
+        for key in resp.headers.keys():
+            values = resp.headers.getall(key)
+            del resp.headers[key]
+            for value in values:
+                resp.headers.add(str(key), str(value))
+        return resp
+    return encode_headers_tween
```

seanh:
(Force-pushed with the tween version <https://github.com/hypothesis/h/pull/4962/commits/9fb086f3c569334df4c5c51da3ee614e54078af6> instead of <https://github.com/hypothesis/h/commit/bd381a82b8210d48f8e2f157f6855b2e5b6872e0>.)

robertknight:
> This tween may be a better approach, as it provides one centralised place to call str() on all header keys and values 

I can see a couple of potential issues here:

- If we have code that assumes `response.headers` is a `Dict[str, str]` in both Python 2 & 3 and that processes the response before the tween is executed, that could cause issues.
- This adds some (small) overhead to each request. I would guess negligible compared to other tasks that happen in the response processing cycle, but you may wish to check on NR.

The one change request I have here is to add couple of code comments to note that:

- Excluding `unicode_literals` from `uri.py` was intentional
- The encoding tween can be removed after Python 2 is dropped

seanh:
Also got some feedback about the tween in the `#pyramid` irc channel. I'm satisfied that the tween approach is generally correct. I'll address these suggestions and then we can finally get this merged...

seanh:
@robertknight I made some tweaks to the tween:

* Added a `native(s)` function that converts a string to a "native" string in Python 2 or 3, adapted from Pyamid: https://github.com/hypothesis/h/pull/4962/commits/d4a49cc59b3a1ab3426f8c01a720221a9894df39

* Made the tween use this `native()` function. Also added tests for the tween. https://github.com/hypothesis/h/pull/4962/commits/31f0d1e233e4368eb2fb50123deded6246322bd6

I'm happy to merge this now

h/tweens.py
```diff
@@ -142,3 +143,40 @@ def cache_header_tween(request):
...
+
+    Since string literals are byte strings in Python 2 and are unicode strings
+    in Python 3, using string literals for header values (as in
+    ``response.headers["Access-Control-Allow-Origin"] = "*"``) works fine in
+    either Python 2 or 3.
+
+    But once you add ``from __future__ import unicode_literals`` to a source
+    code file the string literals become unicode strings in Python 2, and
+    violate PEP-3333. This violation causes ``AssertionError``s from WebTest,
+    and may cause problems with WSGI servers.
```
robertknight: Super explanation 👍

seanh:
Rebased. I'm gonna wait until tomorrow morning to merge and deploy this

PR #4963, opened 2018-04-10
Add `hypothesis devdata` command

seanh:
Add a `hypothesis --dev devdata` command for easily (re-)creating data that's needed in development environments:

* An admin user, a staff user, a normal user, all registered and activated
* An OAuth client for the Hypothesis client, with a fixed ID for use in the `CLIENT_OAUTH_ID` environment variable
* The two auth clients, admin user, organization and group needed for the publisher account test site
* The organization and open and restricted groups for the BioPub demo site

You can run the command multiple times and it will detect the existing data, won't create multiple copies of data.

This could be extended to add more useful data later. For example: users, groups, and annotations with non-ASCII characters, or with long names, etc.

### DIFF
--- a/h/cli/commands/devdata.py
+++ b/h/cli/commands/devdata.py
@@ -1,0 +1,174 @@
+# -*- coding: utf-8 -*-
+from __future__ import print_function
+
+import click
+
+from h import models
+
+
+def create_all_dev_data(request):
+    """Create all the dev data that doesn't exist yet."""
+    factory = DevDataFactory(request)
+
+    # Create some Hypothesis users: an admin user, a staff user, a normal user, ...
+    factory.user(u"user")
+    factory.user(u"admin", admin=True)
+    factory.user(u"staff", staff=True)
+
+    # Create the auth client that the Hypothesis client needs to login.
+    factory.auth_client(u"a77b3fee-31e8-11e8-b65b-cf64c6731003",
+                        name=u"client",
+                        grant_type=u"authorization_code",
+                        response_type=u"code",
+                        redirect_uri=u"http://localhost:5000/app.html")
+
+    # Create auth clients, users etc for the publisher account test site.
+    partner_authority = u"partner.org"
+
+    factory.auth_client(u"e2317c70-3cce-11e8-ac47-23c1999ab55a",
+                        authority=partner_authority,
+                        name=u"publisher_account_test_site_client_credentials",
+                        grant_type=u"client_credentials",
+                        response_type=None,
+                        redirect_uri=None,
+                        secret=u"8l1b3f4vkS80k83WEMl47DKfsOraiQk4aHld-gNYOqM")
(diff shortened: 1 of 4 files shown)


seanh:
Docs need updating. Also some of the code may be able to be moved into / make use of services?

I haven't bothered to write any tests for this.

h/cli/commands/devdata.py
```diff
@@ -0,0 +1,176 @@
...
+                                                       partner_authority)
+
+    factory.group(type_="open",
+                  name=u"Partner",
+                  creator=publisher_site_admin,
+                  origins=[u"http://localhost:5050"],
+                  organization=publisher_site_organization)
+
+    # Open and restricted groups for the BioPub demo site.
+    biopub_admin = factory.user(username=u"biopub_admin")
```
hmstepanek: I like that we are using biopub here so there is some symmetry between prod/qa and dev.

h/cli/commands/devdata.py
```diff
@@ -0,0 +1,176 @@
...
+    factory.auth_client(u"a77b3fee-31e8-11e8-b65b-cf64c6731003",
+                        name=u"client",
+                        grant_type=u"authorization_code",
+                        response_type=u"code",
+                        redirect_uri=u"http://localhost:5000/app.html")
+
+    # Create auth clients, users etc for the publisher account test site.
+    partner_authority = u"partner.org"
+
+    factory.auth_client(u"e2317c70-3cce-11e8-ac47-23c1999ab55a",
```
hmstepanek: I definitely agree with Lyza we should move these to env vars or something.

h/services/__init__.py
```diff
@@ -20,6 +20,7 @@ def includeme(config):
     config.register_service_factory('.flag.flag_service_factory', name='flag')
     config.register_service_factory('.flag_count.flag_count_service_factory', name='flag_count')
     config.register_service_factory('.group.groups_factory', name='group')
+    config.register_service_factory('.organization.organization_factory', name='organization')
```
hmstepanek: Just an FYI: This is covered in https://github.com/hypothesis/h/pull/4956.

h/cli/commands/devdata.py
```diff
@@ -0,0 +1,176 @@
...
+    factory.group(type_="restricted",
+                  name=u"BioPub Restricted",
+                  creator=biopub_admin,
+                  origins=[u"http://localhost:9000"],
+                  organization=biopub_organization)
+
+    request.tm.commit()
+
+
+class DevDataFactory(object):
```
hmstepanek: I share Lyza's concern that this is a lot of code and will need to be maintained. I wonder if there is a way to simplify it such that it's just a list of commands to create the env rather than this whole extra helper class in between?

seanh:
Closing in favour of https://github.com/hypothesis/devdata

PR #4971, opened 2018-04-17
Remove deprecated and redundant pip `--use-wheel` argument

robertknight:
As noted at https://pip.pypa.io/en/stable/user_guide/, pip automatically
installs wheels when they are available, so `--use-wheel` is a no-op
which existed because of the counterpart `--no-use-wheel`. That was in
turn removed in favor of `--no-binary` in pip 10.0.

This fixes Python dependency installation with the latest pip release.


seanh:
:+1:

PR #4974, opened 2018-04-17
Remove unreachable read permission logic

hmstepanek:


seanh:
@lyzadanger What do you think of adding this `user_has_read_permissions` property to `GroupResource`?

hmstepanek:
I agree with what you are saying I'm just wondering if we should do that now or after we pull these out into private methods. I'm thinking pull them into private methods/break up the code into smaller bits, then fix the groupresource, then break up the code into separate modules/templates, then we can start moving these private methods other places and reducing the number of tests. There's a lot to do here and I just worry it's too much at once to follow.

hmstepanek:
I had another look at this and actually I think we just need to get rid of this all together as the permission check in _check_access_permissions should take care of this and make this code unreachable. I believe the tests make it so this code is reachable right now because the permission policy is mocked to pass unless otherwise overridden which is confusing and not really correct in my opinion.
```
        if not self.request.has_permission('read', self.group):
```

hmstepanek:
We should come back to this at some point but for now I'm closing this.

PR #4986, opened 2018-04-20
Move GroupFactory into h/resources.py and add unit tests

seanh:
This is part of the refactoring described here: <https://hypothes-is.slack.com/files/U074DEU66/FA511033L/h_traversal__resources__refactoring>. This part just moves `GroupFactory` into `resources.py` where all the other resources and factories are.


sheetaluk:
LGTM

PR #4996, opened 2018-04-25
Rename OrganizationResource to OrganizationContext

seanh:
Part of the refactoring described in this Slack post:

https://hypothes-is.slack.com/files/U074DEU66/FA511033L/h_traversal__resources__refactoring

This is part of renaming all the `*Resource` classes to `*Context`. In
practice these classes are always used as "context resources" in Pyramid -
they're always the resource at the end of traversal, and passed to the
view as the context. They're never used as intermediate resources in
traversal. As discussed in the Slack post, "context" much better
describes what we're using these classes for than "resource".

### DIFF
--- a/h/presenters/group_json.py
+++ b/h/presenters/group_json.py
@@ -10,7 +10,7 @@ class GroupJSONPresenter(object):
 
     def __init__(self, group_resource):
         self.resource = group_resource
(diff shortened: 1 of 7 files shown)


sheetaluk:
LGTM.

PR #4997, opened 2018-04-26
Redirect to entity not found page when a user or group search fails.

sheetaluk:
Currently, for a search where the user or group does not exist, all
annotations are returned even though they don't satisfy the search
criteria.

Correct that by redirecting to an `entity_not_found` page which displays
an appropriate `entity not found` message.

Fixes https://github.com/hypothesis/product-backlog/issues/462


sheetaluk:
image

Here is what it looks like when a user (or group) isn't found.

The URL looks funny to me, but I wasn't sure how it should be. Any thoughts are welcome!

sheetaluk:
@seanh sounds like a better idea. Ok, I'll do that then.

PR #5005, opened 2018-05-01
Show users a no annotations message when they search for non-existent…

sheetaluk:
… users and groups.

Currently, we return all annotations when a user searches for a non-existent
user or group. This can be a very confusing experience for the user.
Change it to be more meaningful.

Fixes hypothesis/product-backlog#462

### DIFF
--- a/tests/h/activity/query_test.py
+++ b/tests/h/activity/query_test.py
@@ -97,6 +97,16 @@ class TestCheckURL(object):
 
         assert check_url(pyramid_request, query, unparse=unparse) is None
 
+    def test_does_not_remove_group_term_from_query_if_group_does_not_exist(self,
+                                                                           pyramid_request,
+                                                                           unparse):
+        query = MultiDict({'group': 'does_not_exist'})
+
+        check_url(pyramid_request, query, unparse=unparse)
+
+        assert query.get('group') == 'does_not_exist'
+        assert not unparse.called
+
     def test_removes_group_term_from_query(self, group, pyramid_request, unparse):
         query = MultiDict({'group': group.pubid})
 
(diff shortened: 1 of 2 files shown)


tests/h/activity/query_test.py
```diff
@@ -97,6 +97,14 @@ def test_does_not_redirect_to_group_page_if_group_does_not_exist(self,
 
         assert check_url(pyramid_request, query, unparse=unparse) is None
 
+    def test_does_not_remove_group_term_from_query_if_group_does_not_exist(self,
+                                                                           pyramid_request,
+                                                                           unparse):
+        query = MultiDict({'group': 'does_not_exist'})
+
+        check_url(pyramid_request, query, unparse=unparse)
+        assert not unparse.called
```
seanh: Does this test what the test name says it tests?

tests/h/activity/query_test.py
```diff
@@ -143,6 +151,16 @@ def test_does_not_redirect_to_user_page_if_user_does_not_exist(self,
...
 
+    def test_does_not_remove_user_term_from_query_if_group_does_not_exist(self,
+                                                                          pyramid_request,
+                                                                          unparse,
+                                                                          user_service):
+        query = MultiDict({'user': 'jose'})
+        user_service.fetch.return_value = None
+
+        check_url(pyramid_request, query, unparse=unparse)
+        assert not unparse.called
```
seanh: Same question as for the test above.

tests/h/activity/query_test.py
```diff
@@ -143,6 +153,18 @@ def test_does_not_redirect_to_user_page_if_user_does_not_exist(self,
 
         assert check_url(pyramid_request, query) is None
 
+    def test_does_not_remove_user_term_from_query_if_group_does_not_exist(self,
```
robertknight: Typo. This test name should end with `if_user_does_not_exist` rather than `if_group_does_not_exist`.

robertknight: I've pushed a fix for this.

robertknight:
Jenkins is currently down but I'm going to merge this anyway because I think it unlikely enough that the tests will have an issue there.

tests/h/activity/query_test.py
```diff
@@ -143,6 +153,18 @@ def test_does_not_redirect_to_user_page_if_user_does_not_exist(self,
...
+                                                                         pyramid_request,
+                                                                         unparse,
+                                                                         user_service):
+        query = MultiDict({'user': 'jose'})
+        user_service.fetch.return_value = None
+
+        check_url(pyramid_request, query, unparse=unparse)
+
+        assert query.get('user') == 'jose'
+        assert not unparse.called
```
seanh: @robertknight @sheetaluk Why are we asserting that `unparse` wasn't called here? Doesn't the first assert meet the test that it doesn't remove the user from the query? unparse seems to be irrelevant here if I'm not mistaken

robertknight: > Doesn't the first assert meet the test that it doesn't remove the user from the query? 

Yes, indeed. @sheetaluk can you follow up on this.

PR #5007, opened 2018-05-01
Enable Elasticsearch on Travis

seanh:


seanh:
Closing this as adding `sudo: required` routes us onto Travis's sudo-enabled, slower VM-based infrastructure instead of the faster container-based stuff. We want to try getting Elasticsearch working without sudo instead

PR #5008, opened 2018-05-02
Run Elasticsearch 6.2 in Docker on Travis

seanh:
I just want to see if this works, and if it routes us to Travis's sudo-enabled infra.


seanh:
This looks like it works (as in running Es in Docker seems to work) but it does kick us over to the sudo infra, and the build seems to be taking its sweet time too. So gonna close thi

PR #5010, opened 2018-05-02
Only show one error message per field while submiting a form.

sheetaluk:
At the moment we show every error message applicable to a form field
causing them to overlap, making it impossible to read them.
Show the user just one error message at a time.

Fixes: https://github.com/hypothesis/product-backlog/issues/581

### DIFF
--- a/h/templates/deform/mapping_item.jinja2
+++ b/h/templates/deform/mapping_item.jinja2
@@ -46,7 +46,9 @@
 
 {%- if field.error and not field.widget.hidden -%}
   <ul class="form-input__error-list">
-  {% for msg in field.error.messages() -%}
+  {# Only show the first error message for each field. #}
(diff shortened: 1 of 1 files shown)


h/templates/deform/mapping_item.jinja2
```diff
@@ -46,7 +46,9 @@
 
 {%- if field.error and not field.widget.hidden -%}
   <ul class="form-input__error-list">
-  {% for msg in field.error.messages() -%}
+  <!-- Only show the first error message for each field. -->
+  <!-- https://github.com/hypothesis/product-backlog/issues/581 -->
```
robertknight: We should use Jinja2 comments (`{# ... #}`) rather than HTML comments since Jinja comments don't show up in the rendered output.

sheetaluk:
I knew that looked weird!

PR #5011, opened 2018-05-03
Substitute elasticsearch1 for elasticsearch

hmstepanek:
Replace the elasticsearch library with elasticsearch1 in prep for using two different elasticsearch library versions at the same time. This is a fix for https://github.com/hypothesis/product-backlog/issues/633. This doesn't include any test changes as mentioned in the linked issue.

### DIFF
--- a/tests/functional/conftest.py
+++ b/tests/functional/conftest.py
@@ -90,7 +90,7 @@ def _clean_database(engine):
 
 
 def _clean_elasticsearch(settings):
-    import elasticsearch
+    import elasticsearch1
 
     conn = elasticsearch.Elasticsearch([settings['es.host']])
     conn.delete_by_query(index=settings['es.index'],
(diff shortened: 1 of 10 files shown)


tests/functional/conftest.py
```diff
@@ -90,17 +90,17 @@ def _clean_database(engine):
...
 
 def _drop_indices(settings):
-    import elasticsearch
+    import elasticsearch1
 
-    conn = elasticsearch.Elasticsearch([settings['es.host']])
+    conn = elasticsearch1.Elasticsearch([settings['es.host']])
 
     name = settings['es.index']
     if conn.indices.exists(index=name):
```
hmstepanek: Note I didn't mess with any of the tests. This is the place that I suspect @lyzadanger was talking about fixing and  @seanh is modifying in his PR (and in fact deleted) so I don't wanna mess with it. (I wasn't able to catch her to ask before she left :( ) LMK if you think differently!

PR #5012, opened 2018-05-04
Index querying tests

seanh:


tests/h/search/index_test.py
```diff
@@ -56,22 +72,66 @@ def test_it_indexes_the_annotations_document_title(self, factories,
...
+        assert SearchResponseWithIDs([annotation.id]) == response
+
+    def test_you_can_filter_annotations_by_creation_time(self, factories, index, search):
+        before = datetime.datetime.now()
+        annotation = factories.Annotation.build(id="test_annotation_id", created=datetime.datetime.now())
+
+        index(annotation)
+
+        response = search.filter("range", created={"gte": before}).execute()
+        assert SearchResponseWithIDs([annotation.id]) == response
```
robertknight: These matchers have the form `assert expected == actual` rather than the usual `assert actual == expected`, is there a reason for that?

PR #5013, opened 2018-05-04
Add elasticsearch6.2 and elasticsearch-dsl deps

hmstepanek:
These dependencies are needed to start using elasticsearch6.2 in h.

### DIFF
--- a/requirements.in
+++ b/requirements.in
@@ -11,7 +11,9 @@ click
 cryptography
 deform < 1.0
 deform-jinja2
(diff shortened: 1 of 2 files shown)


seanh:
Why is it `elasticsearch==6.2.0` but `elasticsearch-dsl==6.1.0`, though? Looks like 6.1.0 is the latest version of `elasticsearch-dsl`, so I guess that works fine with `elasticsearch==6.2.0`?

PR #5016, opened 2018-05-08
Change search to search old

hmstepanek:
This is in preparation for adding the new elastic search code. It's easier to transition when we don't have a bunch of module name clashes. This also gives us a nice clean space for implementing the new search code following the https://pypi.org/project/elasticsearch-dsl docs. The new elastic search code will be added under h/search and the old elastic search code will now exist under h/search_old. There will also be a wrapper layer (Search) in the new search area that looks like the Search in h/search_old/core.py but it will talk to both clusters and make the new search code look like the old search code.


robertknight:
Hi @hmstepanek - Lyza and me have been discussing this in Slack here: https://hypothes-is.slack.com/archives/C4K6M7P5E/p1525869988000521

Lyza suggested a different approach which will reduce the amount of code churn outside of the `h.search` package. Can you have a look over it?

PR #5019, opened 2018-05-10
Connect to new elasticsearch cluster

hmstepanek:
Add in connection logic so that the app connects to elasticsearch6.2 cluster on startup. This is a partial fix for https://github.com/hypothesis/product-backlog/issues/648. Note this does not include indexing the cluster it is just the creation of the connection/client obj.

### DIFF
--- a/h/search/__init__.py
+++ b/h/search/__init__.py
@@ -1,7 +1,8 @@
 # -*- coding: utf-8 -*-
 
 from __future__ import unicode_literals
-from h.search.client import get_client
+from elasticsearch_dsl import connections
+from h.search.client import get_client, get_client_old
 from h.search.config import init
 from h.search.core import Search
 from h.search.core import FILTERS_KEY
@@ -16,6 +17,7 @@ __all__ = (
 def includeme(config):
     settings = config.registry.settings
     settings.setdefault('es.host', 'http://localhost:9200')
+    settings.setdefault('es.url', 'http://localhost:9201')
     settings.setdefault('es.index', 'hypothesis')
 
     # Allow users of this module to register additional search filter and
@@ -29,8 +31,10 @@ def includeme(config):
     # Add a property to all requests for easy access to the elasticsearch
     # client. This can be used for direct or bulk access without having to
     # reread the settings.
-    config.registry['es.client'] = get_client(settings)
+    config.registry['es.client'] = get_client_old(settings)
     config.add_request_method(
         lambda r: r.registry['es.client'],
         name='es',
         reify=True)
+    # Set the elasticsearch client connection as the default connection.
+    connections.add_connection('default', get_client(settings))
--- a/h/search/client.py
+++ b/h/search/client.py
@@ -2,10 +2,9 @@
 
 from __future__ import unicode_literals
 import certifi
-from elasticsearch1 import Elasticsearch, RequestsHttpConnection
+import elasticsearch
+import elasticsearch1
 from requests_aws4auth import AWS4Auth
-
-__all__ = ('Client',)
 
 
 class Client(object):
@@ -26,12 +25,9 @@ class Client(object):
 
     def __init__(self, host, index, **kwargs):
         self._index = index
-        self._conn = Elasticsearch([host],
-                                   verify_certs=True,
-                                   # N.B. this won't be necessary if we upgrade
-                                   # to elasticsearch>=5.0.0.
-                                   ca_certs=certifi.where(),
-                                   **kwargs)
+        self._conn = elasticsearch1.Elasticsearch(
+            [host],
+            **kwargs)
 
     @property
     def index(self):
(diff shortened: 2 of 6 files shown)


hmstepanek:
I've added functional tests now to validate the real client connection. I think it would still be good to add the testing of the includeme connection code. I'll try to tackle that in the AM tomorrow.

tests/h/search/client_test.py
```diff
@@ -31,75 +44,139 @@ def test_conn_is_read_only(self):
...
         return {
             'es.host': 'search.svc',
+            'es.url': 'search.svc',
             'es.index': 'my-index',
         }
 
     @pytest.fixture
-    def patched_client(self, patch):
-        return patch('h.search.client.Client')
+    def patched_elasticsearch(self, patch, request):
```
hmstepanek: I changed this from mocking the custom Client to mocking the underlying call to elasticsearch.Elasticsearch that actually creates the client object as that is more compatible with both versions of the code.

tests/h/search/client_test.py
```diff
@@ -4,15 +4,28 @@
...
+import elasticsearch_dsl
+import elasticsearch1_dsl
 
-from h.search.client import get_client
+from h.search.client import get_client, get_client_old
 from h.search.client import Client
 
 
-class TestClient(object):
+ES1_PATH = 'h.search.client.elasticsearch1.Elasticsearch'
```
hmstepanek: This stuff at the top could ultimately disappear once we are no longer testing two versions.

h/search/client.py
```diff
@@ -42,28 +38,35 @@ def conn(self):
...
                         settings['es.aws.secret_access_key'],
                         settings['es.aws.region'],
                         'es')
         kwargs['http_auth'] = auth
-        kwargs['connection_class'] = RequestsHttpConnection
+        kwargs['connection_class'] = elasticsearch.RequestsHttpConnection
+    return kwargs
+
+
+def get_client(settings):
```
hmstepanek: Here we are returning the client connection as opposed to our custom Client wrapper obj.

h/search/__init__.py
```diff
@@ -16,6 +17,7 @@
 def includeme(config):
     settings = config.registry.settings
     settings.setdefault('es.host', 'http://localhost:9200')
+    settings.setdefault('es.url', 'http://localhost:9201')
```
hmstepanek: I think we could swap these soon if we want now that we actually have a real connection working. I'm not sure what that would look like/how difficult that would be in production?

h/search/__init__.py
```diff
@@ -29,8 +31,10 @@ def includeme(config):
...
     # client. This can be used for direct or bulk access without having to
     # reread the settings.
-    config.registry['es.client'] = get_client(settings)
+    config.registry['es.client'] = get_client_old(settings)
     config.add_request_method(
         lambda r: r.registry['es.client'],
         name='es',
         reify=True)
+    # Set the elasticsearch client connection as the default connection.
+    connections.add_connection('default', get_client(settings))
```
hmstepanek: We could do this inside of the get_client but I think it's nice to still have a get_client function for testing purposes. The using param is set to 'default' by default in elasticsearch-dsl so this is another way to set the default without creating a client via elasticsearch-dsl. Both ways are perfectly valid.

h/search/client.py
```diff
@@ -42,28 +38,35 @@ def conn(self):
...
-    index = settings['es.index']
-    kwargs = {}
-    kwargs['max_retries'] = settings.get('es.client.max_retries', 3)
-    kwargs['retry_on_timeout'] = settings.get('es.client.retry_on_timeout', False)
-    kwargs['timeout'] = settings.get('es.client.timeout', 10)
+def _get_settings(settings, elasticsearch=elasticsearch):
+    kwargs = {'verify_certs': True,
+              'max_retries': settings.get('es.client.max_retries', 3),
+              'retry_on_timeout': settings.get('es.client.retry_on_timeout', False),
+              'timeout': settings.get('es.client.timeout', 10)}
```
seanh: Rather than refactoring the existing code to share things between the old and new code, I'd rather just duplicate these settings in two places. We're going to be deleting the old code anyway.

tests/h/search/client_test.py
```diff
@@ -31,75 +44,139 @@ def test_conn_is_read_only(self):
...
+    @pytest.fixture
+    def index(self, User, client, request):
+        elasticsearch_dsl = request.getfixturevalue('elasticsearch_dsl')
+        # create the index
+        index = elasticsearch_dsl.Index('user', using='es')
+        index.doc_type(User)
+        index.create()
+        yield index
+        # delete index
+        index.delete()
```
seanh: We'll need test fixtures for the new Elasticsearch index to create the index before each test run and to delete the index after each test run, and to empty out the index in-between tests, but those should go in `tests/common/fixtures/elasticsearch.py` rather than in a particular test class. And I'm not sure that we need them until we're actually writing anything to the index (if we want a test to just test that the connection is available and working, can that just make some sort of Elasticsearch metadata query that doesn't require there to be an index?)

I don't think we should add an `es_client` fixture for the new Elasticsearch until we know that the code actually needs direct access to the client

PR #5023, opened 2018-05-11
Move core_test.py -> old_core_test.py

seanh:
So I can write a new core_test.py that integrates the real Elasticsearch.


sheetaluk:
LGTM

PR #5029, opened 2018-05-17
Add Es-integrated tests for separate_replies

seanh:
Add the first Elasticsearch-integrated tests for the search querying functionality.

These test elements of the search querying functionality related to the separate_replies argument: how search behaves with and without this argument.

I'm not sure that everything to do with separate_replies is covered by these tests yet. For example deleted replies are excluded from reply_ids, and this isn't tested yet. There are several more things like
this still to test.

### DIFF
--- a/tests/h/search/core_test.py
+++ b/tests/h/search/core_test.py
@@ -213,14 +213,19 @@ class TestWithSeparateReplies(object):
         annotation_ids) that can be included in reply_ids.
         """
         annotation = Annotation(shared=True)
-        first_reply = Annotation(references=[annotation.id], shared=True)
-        for _ in range(200):
+        oldest_reply = Annotation(references=[annotation.id], shared=True)
+
+        # Create three more replies so that the oldest reply will be pushed out
+        # of reply_ids. (We only need 3, not 200, because we're going to use
+        # the _replies_limit test seam to limit it to 3 replies instead of 200.
+        # This is just to make the test faster.
+        for _ in range(3):
             Annotation(references=[annotation.id], shared=True)
 
-        result = search.Search(pyramid_request, separate_replies=True).run({})
+        result = search.Search(pyramid_request, separate_replies=True, _replies_limit=3).run({})
 
-        assert len(result.reply_ids) == 200
-        assert first_reply.id not in result.reply_ids
+        assert len(result.reply_ids) == 3
+        assert oldest_reply.id not in result.reply_ids
 
 
 @pytest.fixture
--- a/tests/h/search/core_test.py
+++ b/tests/h/search/core_test.py
@@ -75,27 +75,27 @@ class TestWithoutSeparateReplies(object):
         Things are returned in updated order so normally a reply would appear
         before the annotation that it is a reply to in the search results.
         """
-        annotation = Annotation(shared=True)
-        reply = Annotation(references=[annotation.id], shared=True)
+        now = datetime.datetime.now()
+        five_mins = datetime.timedelta(minutes=5)
+        annotation = Annotation(updated=now, shared=True)
+        reply = Annotation(updated=now + five_mins, references=[annotation.id], shared=True)
 
         result = search.Search(pyramid_request).run({})
 
         # The reply appears _before_ the annotation in the search results.
         assert result.annotation_ids == [reply.id, annotation.id]
 
-    def test_replies_can_come_after_annotations(self, index, pyramid_request, Annotation):
+    def test_replies_can_come_after_annotations(self, pyramid_request, Annotation):
         """A reply may appear after its annotation in the search results.
 
         Things are returned in updated order so if the original author has
         updated the top-level annotation since the reply was created, then the
         annotation would come before the reply in the search results.
         """
-        annotation = Annotation(shared=True)
-        reply = Annotation(references=[annotation.id], shared=True)
-
-        # Simulate updating the original annotation.
-        annotation.updated = datetime.datetime.now()
-        index(annotation)
+        now = datetime.datetime.now()
+        five_mins = datetime.timedelta(minutes=5)
+        annotation = Annotation(updated=now + five_mins, shared=True)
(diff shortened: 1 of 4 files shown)


tests/h/search/core_test.py
```diff
@@ -0,0 +1,28 @@
...
+from __future__ import unicode_literals
+
+import pytest
+
+from h import search
+
+
+class TestSearch(object):
+
+    def test_it_returns_replies_in_with_the_annotations_ids(self, factories, index, matchers,
```
robertknight: Typo s/replies_in_with/replies_with?

tests/h/search/core_test.py
```diff
@@ -0,0 +1,243 @@
...
+    """Tests for what happens when no separate_replies argument is given."""
+
+    def test_it_returns_replies_in_annotations_ids(self, matchers, pyramid_request, Annotation):
+        """Without separate_replies it returns replies in annotation_ids.
+
+        Test that if no separate_replies argument is given then it returns the
+        ids of replies mixed in with the top-level annotations in
+        annotation_ids.
+        """
+        annotation = Annotation(shared=True)
```
seanh: By default `factories.Annotation` creates private annotations. I considered making a `factories.SharedAnnotation`, which I think might be a little better than `Annotation(shared=True)` because it keeps the factories API more separate from the `models.Annotation` class (if we changed `models.Annotation` to no longer use a `shared` property, we wouldn't have to change every test that makes an annotation using a factory). But if we did that we should really rename `factories.Annotation` to `factories.PrivateAnnotation` too. Anyway, I haven't done it yet

tests/h/search/core_test.py
```diff
@@ -0,0 +1,243 @@
...
+        reply_1 = Annotation(shared=True, references=[annotation.id])
+        reply_2 = Annotation(shared=True, references=[annotation.id])
```
seanh: I considered adding a `factories.Reply(annotation)` but I couldn't find a simple way to get factory_boy to do it. So going with having to pass in `references` (which is pretty horrible).

tests/h/search/core_test.py
```diff
@@ -0,0 +1,243 @@
...
+
+        result = search.Search(pyramid_request).run({})
```
seanh: It's kind of annoying that you still have to pass `{}` to `run()` to do a search with no params.

seanh:
Just reading the names of the tests here makes me realise what a piece of junk our API is.

tests/h/search/core_test.py
```diff
@@ -0,0 +1,243 @@
...
+        assert annotation.id in result.annotation_ids
+
+    def test_replies_can_come_before_annotations(self, pyramid_request, Annotation):
+        """A reply may appear before its annotation in the search results.
+
+        Things are returned in updated order so normally a reply would appear
+        before the annotation that it is a reply to in the search results.
+        """
+        annotation = Annotation(shared=True)
+        reply = Annotation(references=[annotation.id], shared=True)
```
robertknight: Is it **guaranteed** that the reply will have a later updated date here? If not we could have a flakey test.

tests/h/search/core_test.py
```diff
@@ -0,0 +1,243 @@
...
+
+    def test_replies_can_come_after_annotations(self, index, pyramid_request, Annotation):
+        """A reply may appear after its annotation in the search results.
+
+        Things are returned in updated order so if the original author has
+        updated the top-level annotation since the reply was created, then the
+        annotation would come before the reply in the search results.
+        """
+        annotation = Annotation(shared=True)
+        reply = Annotation(references=[annotation.id], shared=True)
```
robertknight: Are the updated timestamps of `annotation` and `reply` guaranteed to have the expected ordering here?

tests/h/search/core_test.py
```diff
@@ -0,0 +1,243 @@
...
+
+    def test_replies_are_ordered_most_recently_updated_first(self, Annotation, index,
+                                                             pyramid_request):
+        annotation = Annotation(shared=True)
+        reply_1 = Annotation(references=[annotation.id], shared=True)
+        reply_2 = Annotation(references=[annotation.id], shared=True)
+        reply_3 = Annotation(references=[annotation.id], shared=True)
+
+        # Simulate updating one of the replies.
+        reply_1.updated = datetime.datetime.now()
```
robertknight: Same question about updated timestamps as above.

tests/h/search/core_test.py
```diff
@@ -0,0 +1,243 @@
...
+
+    def test_only_200_replies_are_included(self, pyramid_request, Annotation):
+        """No more than 200 replies can be included in reply_ids.
+
+        200 is the total maximum number of replies (to all annotations in
+        annotation_ids) that can be included in reply_ids.
+        """
+        annotation = Annotation(shared=True)
+        first_reply = Annotation(references=[annotation.id], shared=True)
+        for _ in range(200):
```
robertknight: Perhaps we should make the `200` value injectable and then use a smaller value to make the test run faster.

tests/h/search/core_test.py
```diff
@@ -0,0 +1,243 @@
...
+        annotation = Annotation(shared=True)
+        reply = Annotation(references=[annotation.id], shared=True)
+
+        # Now create 19 more annotations so that the original annotation is
+        # pushed onto the second page of the search results, but the reply is
+        # still on the first page.
+        for _ in range(19):
+            Annotation(shared=True)
+
+        # The reply is one the first page of search results, but the original annotation isn't.
```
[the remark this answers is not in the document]
seanh: Fixed

tests/h/search/core_test.py
seanh: Fixed

tests/h/search/core_test.py
seanh: Done

seanh:
@robertknight All done I think

seanh:
Ok really done now

PR #5031, opened 2018-05-18
Auto-generate test annotation ids in factories.Annotation.build() and factories.Annotation.stub()

seanh:
Annotations created with factories.Annotation() do have ids but
unsaved and stub annotations created by factories.Annotion.build() or
factories.Annotation.stub() without passing an id argument don't have
ids:

    >>> a = factories.Annotation.build()
    >>> a.id
    None

This is because the ids are generated by models.Annotation using a
server_default function, so an id isn't generated until the annotation
is actually saved to the DB.

This creates noise in tests that have to pass an id argument to build()
or stub() because, for example, the test is going to add the annotation
to the test Elasticsearch index and doing so requires a non-null id,

Add code to factories.Annotation to auto-generate an id when build() or
stub() is called without passing an id.

Annotations created without using build() or stub() are not affected,
they still get an id from the server_default function:

    >>> factories.Annotation()
    <Annotation 3nemKFqbEeirh6fGyzgKag>

Annotations created with build() or stub() now also get a UUID id.
You'll notice that these look slightly different to the server_default
ids because the code for creating the ids isn't exactly the same. But
it's good enough for the tests:

    >>> factories.Annotation.build()
    <Annotation MTVhOTNiZDQtNWE5Yy0xMW>

    >>> factories.Annotation.stub().id
    'MjFlOTMwZmMtNWE5Yy0xMW'

You can still pass an id argument to create an annotation with a
particular id:

    >>> factories.Annotation.build(id="foo")
    <Annotation foo>

    >>> factories.Annotation.stub(id="bar").id
    'bar'

### DIFF
--- a/tests/common/factories/annotation.py
+++ b/tests/common/factories/annotation.py
@@ -3,6 +3,7 @@
 from __future__ import unicode_literals
 
 import random
+import uuid
 
 import factory
 from sqlalchemy import orm
@@ -115,3 +116,18 @@ class Annotation(ModelFactory):
             created=self.created,
             updated=self.updated,
         )
+
+    @factory.post_generation
+    def make_id(self, create, extracted, **kwargs):
+        """Add a randomly ID if the annotation doesn't have one yet."""
+        # If using the create strategy don't generate an id.
+        # models.Annotation.id's server_default function will generate one
+        # when the annotation is saved to the DB.
+        if create:
+            return
+
+        # Don't generate an id if the user passed in one of their own.
+        if getattr(self, "id", None):
+            return
+
+        self.id = uuid.uuid1().hex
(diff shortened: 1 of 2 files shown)


seanh:
Fixed the Python 3 fail. I'm just using `uuid.uuid1().hex` to create the test ids. This works, but unfortunately it means that annotations created using `build()` can't be added to the db session:

```python
a = factories.Annotation.build()
request.db.add(a)
request.tm.commit()
TypeError: 'unicode' does not have the buffer interface
```

This is a shame, you should be able to do this. But its pretty unusual in a test to want to create an annotation without adding it to the DB and then add it to the DB later. Why wouldn't you just use `factories.Annotation()` and have it added to the DB automatically? And we don't try to do this anywhere in our tests currently.

So I still think this is an improvement over what we have and we can make the generated IDs DB- compatible later if we have to.

PR #5033, opened 2018-05-18
Replace Env & DeprecatedSetting w/ SettingsManager

hmstepanek:
This is a partial fix for https://github.com/hypothesis/product-backlog/issues/647.
- [x] Add a `required=False` keyword argument to `EnvSetting`. Change any existing required settings should use this.
- [x] Add a `default=None` keyword argument to `EnvSetting`. Change any existing settings that have default values that can be hardcoded in Python should use this.

### DIFF
--- a/h/config.py
+++ b/h/config.py
@@ -107,6 +107,8 @@ def configure(environ=None, settings=None):
         settings_manager.set('mail.host', 'MANDRILL_HOST', default='smtp.mandrillapp.com')
         settings_manager.set('mail.port', 'MANDRILL_PORT', default=587)
         settings_manager.set('mail.tls', 'MANDRILL_TLS', default=True)
+
+    # Get resolved settings.
     settings = settings_manager.settings
 
     if 'secret_key' not in settings:
--- a/h/config.py
+++ b/h/config.py
@@ -11,11 +11,10 @@ from pyramid.config import Configurator
 from pyramid.settings import asbool, aslist
 
 from h.settings import (
-    DeprecatedSetting,
-    EnvSetting,
     SettingError,
     database_url,
     mandrill_settings,
+    SettingsManager
 )
 from h.util.logging_filters import ExceptionFilter
 
@@ -35,73 +34,6 @@ DEFAULT_SALT = (b"\xbc\x9ck!k\x81(\xb6I\xaa\x90\x0f'}\x07\xa1P\xd9\xb7\xcb"
 SETTINGS = [
     # Mailer configuration for Mandrill
     mandrill_settings,
-
-    # Configuration for external components
-    EnvSetting('broker_url', 'BROKER_URL'),
-    EnvSetting('es.client_poolsize', 'ELASTICSEARCH_CLIENT_POOLSIZE',
-               type=int),
-    EnvSetting('es.client.max_retries', 'ELASTICSEARCH_CLIENT_MAX_RETRIES', type=int),
-    EnvSetting('es.client.retry_on_timeout', 'ELASTICSEARCH_CLIENT_RETRY_ON_TIMEOUT', type=asbool),
-    EnvSetting('es.client.timeout', 'ELASTICSEARCH_CLIENT_TIMEOUT', type=float),
-    EnvSetting('es.host', 'ELASTICSEARCH_HOST'),
-    EnvSetting('es.url', 'ELASTICSEARCH_URL'),
-    EnvSetting('es.index', 'ELASTICSEARCH_INDEX'),
-    EnvSetting('es.aws.access_key_id', 'ELASTICSEARCH_AWS_ACCESS_KEY_ID'),
-    EnvSetting('es.aws.region', 'ELASTICSEARCH_AWS_REGION'),
-    EnvSetting('es.aws.secret_access_key', 'ELASTICSEARCH_AWS_SECRET_ACCESS_KEY'),
-    EnvSetting('mail.default_sender', 'MAIL_DEFAULT_SENDER'),
-    EnvSetting('mail.host', 'MAIL_HOST'),
-    EnvSetting('mail.port', 'MAIL_PORT', type=int),
-    EnvSetting('sqlalchemy.url', 'DATABASE_URL', type=database_url),
-    EnvSetting('statsd.host', 'STATSD_HOST'),
-    EnvSetting('statsd.port', 'STATSD_PORT', type=int),
-    EnvSetting('statsd.prefix', 'STATSD_PREFIX'),
-
-    # Configuration for Pyramid
-    EnvSetting('secret_key', 'SECRET_KEY', type=bytes),
-    EnvSetting('secret_salt', 'SECRET_SALT', type=bytes),
-
-    # Configuration for h
-    EnvSetting('csp.enabled', 'CSP_ENABLED', type=asbool),
-    EnvSetting('csp.report_uri', 'CSP_REPORT_URI'),
-    EnvSetting('csp.report_only', 'CSP_REPORT_ONLY'),
-    EnvSetting('ga_tracking_id', 'GOOGLE_ANALYTICS_TRACKING_ID'),
-    EnvSetting('ga_client_tracking_id', 'GOOGLE_ANALYTICS_CLIENT_TRACKING_ID'),
-    EnvSetting('h.app_url', 'APP_URL'),
-    DeprecatedSetting(EnvSetting('h.authority', 'AUTH_DOMAIN'),
-                      message='use the AUTHORITY environment variable instead'),
-    EnvSetting('h.authority', 'AUTHORITY'),
-    EnvSetting('h.bouncer_url', 'BOUNCER_URL'),
-
-    EnvSetting('h.client_url', 'CLIENT_URL'),
-
-    # ID for the OAuth authclient that the embedded client should use when
-    # making requests to OAuth endpoints. As a public client, it does not have a
-    # secret.
-    EnvSetting('h.client_oauth_id', 'CLIENT_OAUTH_ID'),
-
-    # The list of origins that the client will respond to cross-origin RPC
-    # requests from.
-    EnvSetting('h.client_rpc_allowed_origins',
-               'CLIENT_RPC_ALLOWED_ORIGINS', type=aslist),
-
-    EnvSetting('h.db_session_checks', 'DB_SESSION_CHECKS', type=asbool),
-
-    # Environment name, provided by the deployment environment. Please do
-    # *not* toggle functionality based on this value. It is intended as a
-    # label only.
-    EnvSetting('h.env', 'ENV'),
-    # Where should logged-out users visiting the homepage be redirected?
-    EnvSetting('h.homepage_redirect_url', 'HOMEPAGE_REDIRECT_URL'),
-    EnvSetting('h.proxy_auth', 'PROXY_AUTH', type=asbool),
-    # Sentry DSNs for frontend code should be of the public kind, lacking the
-    # password component in the DSN URI.
-    EnvSetting('h.sentry_dsn_client', 'SENTRY_DSN_CLIENT'),
-    EnvSetting('h.sentry_dsn_frontend', 'SENTRY_DSN_FRONTEND'),
-    EnvSetting('h.websocket_url', 'WEBSOCKET_URL'),
-
-    # Debug/development settings
-    EnvSetting('debug_query', 'DEBUG_QUERY'),
 ]
 
 
@@ -110,6 +42,75 @@ def configure(environ=None, settings=None):
         environ = os.environ
     if settings is None:
         settings = {}
+    settings_manager = SettingsManager(settings, environ)
+    # Configuration for external components
+    settings_manager.set('broker_url', 'BROKER_URL')
+    settings_manager.set('es.client_poolsize', 'ELASTICSEARCH_CLIENT_POOLSIZE',
+                         type_=int)
+    settings_manager.set('es.client.max_retries', 'ELASTICSEARCH_CLIENT_MAX_RETRIES', type_=int)
+    settings_manager.set('es.client.retry_on_timeout', 'ELASTICSEARCH_CLIENT_RETRY_ON_TIMEOUT', type_=asbool)
+    settings_manager.set('es.client.timeout', 'ELASTICSEARCH_CLIENT_TIMEOUT', type_=float)
+    settings_manager.set('es.host', 'ELASTICSEARCH_HOST')
+    settings_manager.set('es.url', 'ELASTICSEARCH_URL'),
+    settings_manager.set('es.index', 'ELASTICSEARCH_INDEX')
+    settings_manager.set('es.aws.access_key_id', 'ELASTICSEARCH_AWS_ACCESS_KEY_ID')
+    settings_manager.set('es.aws.region', 'ELASTICSEARCH_AWS_REGION')
+    settings_manager.set('es.aws.secret_access_key', 'ELASTICSEARCH_AWS_SECRET_ACCESS_KEY')
+    settings_manager.set('mail.default_sender', 'MAIL_DEFAULT_SENDER')
+    settings_manager.set('mail.host', 'MAIL_HOST')
+    settings_manager.set('mail.port', 'MAIL_PORT', type_=int)
+    settings_manager.set('sqlalchemy.url', 'DATABASE_URL', type_=database_url)
+    settings_manager.set('statsd.host', 'STATSD_HOST')
+    settings_manager.set('statsd.port', 'STATSD_PORT', type_=int)
+    settings_manager.set('statsd.prefix', 'STATSD_PREFIX')
+
+    # Configuration for Pyramid
+    settings_manager.set('secret_key', 'SECRET_KEY', type_=bytes)
+    settings_manager.set('secret_salt', 'SECRET_SALT', type_=bytes)
+
+    # Configuration for h
+    settings_manager.set('csp.enabled', 'CSP_ENABLED', type_=asbool)
+    settings_manager.set('csp.report_uri', 'CSP_REPORT_URI')
+    settings_manager.set('csp.report_only', 'CSP_REPORT_ONLY')
+    settings_manager.set('ga_tracking_id', 'GOOGLE_ANALYTICS_TRACKING_ID')
+    settings_manager.set('ga_client_tracking_id', 'GOOGLE_ANALYTICS_CLIENT_TRACKING_ID')
+    settings_manager.set('h.app_url', 'APP_URL')
+    settings_manager.set('h.authority', 'AUTH_DOMAIN',
+                         deprecated_msg='use the AUTHORITY environment variable instead')
+    settings_manager.set('h.authority', 'AUTHORITY')
+    settings_manager.set('h.bouncer_url', 'BOUNCER_URL')
+
+    settings_manager.set('h.client_url', 'CLIENT_URL')
+
+    # ID for the OAuth authclient that the embedded client should use when
+    # making requests to OAuth endpoints. As a public client, it does not have a
+    # secret.
+    settings_manager.set('h.client_oauth_id', 'CLIENT_OAUTH_ID')
+
+    # The list of origins that the client will respond to cross-origin RPC
+    # requests from.
+    settings_manager.set('h.client_rpc_allowed_origins',
+                         'CLIENT_RPC_ALLOWED_ORIGINS', type_=aslist)
+
+    settings_manager.set('h.db_session_checks', 'DB_SESSION_CHECKS', type_=asbool)
+
+    # Environment name, provided by the deployment environment. Please do
+    # *not* toggle functionality based on this value. It is intended as a
+    # label only.
+    settings_manager.set('h.env', 'ENV')
+    # Where should logged-out users visiting the homepage be redirected?
+    settings_manager.set('h.homepage_redirect_url', 'HOMEPAGE_REDIRECT_URL')
+    settings_manager.set('h.proxy_auth', 'PROXY_AUTH', type_=asbool)
+    # Sentry DSNs for frontend code should be of the public kind, lacking the
+    # password component in the DSN URI.
+    settings_manager.set('h.sentry_dsn_client', 'SENTRY_DSN_CLIENT')
+    settings_manager.set('h.sentry_dsn_frontend', 'SENTRY_DSN_FRONTEND')
+    settings_manager.set('h.websocket_url', 'WEBSOCKET_URL')
+
+    # Debug/development settings
+    settings_manager.set('debug_query', 'DEBUG_QUERY')
+
+    settings = settings_manager.settings
 
     for s in SETTINGS:
         try:
--- a/h/config.py
+++ b/h/config.py
@@ -11,9 +11,7 @@ from pyramid.config import Configurator
 from pyramid.settings import asbool, aslist
 
 from h.settings import (
-    SettingError,
     database_url,
-    mandrill_settings,
     SettingsManager
 )
 from h.util.logging_filters import ExceptionFilter
@@ -28,13 +26,6 @@ DEFAULT_SALT = (b"\xbc\x9ck!k\x81(\xb6I\xaa\x90\x0f'}\x07\xa1P\xd9\xb7\xcb"
                 b"\xcb\xe8\x8b\t\xcf\xeb *\xa7\xa6\xe1i\xc7\x81\xe8\xd8\x18"
                 b"\x9f\x1b\x96\xc1\xfa\x8b\x19\x82\xa3[\x19\xcb\xa4\x1a\x0f"
                 b"\xe4\xcb\r\x17\x7f\xfbh\xd5^W\xdb\xe6")
-
-# The list of all settings read from the system environment. These are in
-# reverse-priority order, meaning that later settings trump earlier settings.
-SETTINGS = [
-    # Mailer configuration for Mandrill
-    mandrill_settings,
-]
 
 
 def configure(environ=None, settings=None):
@@ -110,16 +101,13 @@ def configure(environ=None, settings=None):
     # Debug/development settings
     settings_manager.set('debug_query', 'DEBUG_QUERY')
 
+    if 'MANDRILL_USERNAME' in environ and 'MANDRILL_APIKEY' in environ:
+        settings_manager.set('mail.username', 'MANDRILL_USERNAME')
+        settings_manager.set('mail.password', 'MANDRILL_APIKEY')
+        settings_manager.set('mail.host', 'MANDRILL_HOST', default='smtp.mandrillapp.com')
+        settings_manager.set('mail.port', 'MANDRILL_PORT', default=587)
+        settings_manager.set('mail.tls', 'MANDRILL_TLS', default=True)
     settings = settings_manager.settings
-
-    for s in SETTINGS:
-        try:
-            result = s(environ)
-        except SettingError as e:
-            log.warn(e)
-
-        if result is not None:
-            settings.update(result)
 
     if 'secret_key' not in settings:
         log.warn('No secret key provided: using transient key. Please '
--- a/h/config.py
+++ b/h/config.py
@@ -66,7 +66,7 @@ def configure(environ=None, settings=None):
 
     # Configuration for Pyramid
     settings_manager.set('secret_key', 'SECRET_KEY', type_=bytes)
-    settings_manager.set('secret_salt', 'SECRET_SALT', type_=bytes)
+    settings_manager.set('secret_salt', 'SECRET_SALT', type_=bytes, default=DEFAULT_SALT)
 
     # Configuration for h
     settings_manager.set('csp.enabled', 'CSP_ENABLED', type_=asbool)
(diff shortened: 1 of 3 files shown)


hmstepanek:
I have moved _some_ of the settings into h/config/`__init__`.py. Some of them are a little tricky I think because they only should be set sometimes so I left them where they are for now. (such as accept and renderer).

hmstepanek:
Also I noticed the card says to make all of these different PR's. You could break this up into a couple if you want but I'll leave that up to you.

conf/development-websocket.ini
```diff
@@ -1,6 +1,9 @@
 [app:main]
 use: call:h.websocket:create_app
 
+# Elasticsearch settings
+es.host: http://localhost:9200
```
hmstepanek: I know we talked about leaving these where they were cause they were going to get deleted anyway  but I figured since I was already moving stuff I might as well move this too.

h/config/__init__.py
```diff
@@ -0,0 +1,135 @@
...
+    if 'MANDRILL_USERNAME' in environ and 'MANDRILL_APIKEY' in environ:
+        SETTINGS.extend([
+            {'setting': 'mail.username', 'env_var': 'MANDRILL_USERNAME'},
+            {'setting': 'mail.password', 'env_var': 'MANDRILL_APIKEY'},
+            {'setting': 'mail.host', 'env_var': 'smtp.mandrillapp.com'},
+            {'setting': 'mail.port', 'env_var': 587},
+            {'setting': 'mail.tls', 'env_var': True},
+        ])
+    for s in SETTINGS:
+        try:
```
hmstepanek: This try catch definitely needs to get removed if we want to crash on startup when we're missing a required env var. However, I wasn't sure exactly how to maintain the existing behavior _and_ crash on startup as they both raise the same exception. We could make multiple exception classes for the various error types, we could simply log a warning inside the env_setting method and not raise an exception, or we could not maintain the existing behavior and raise an exception when the type casting fails (<-I'm inclined to do this last one.).

hmstepanek:
*Note I made some comments on the files. In particular we need to make a decision about the try-catch in configure.

hmstepanek:
I think that looks very nice! Will do!

hmstepanek:
@seanh If you wouldn't mind taking a quick peak. I'm not done yet but just to make sure everything looks good so far. I thought SettingsManager was a more appropriate name but I'm open to other ideas. If this looks good I'll change config.py to use the new SettingsManager. I think it makes more sense to just add all the features in at the time of implementation considering it's a new thing anyway and not just editions to the old thing. My plan is to introduce using the new features incrementally (aka multiple commits) so hopefully that will be easy enough to follow.

h/settings.py
```diff
@@ -13,6 +13,52 @@ class SettingError(Exception):
...
+        """
+        Sets the specified setting in settings to the value of the env_var,
+        of if the env_var isn't present and a default value is specified; to
+        the default value..
+        If the env_var doesn't exist and the setting is not required then it
+        does nothing.
+        If the setting is required and not specified it raises a SettingError.
+        If a default value is provided and the env_var doesn't exist, the setting will be
+        overriden with the provided default value.
+        """
```
seanh: I realise this is draft code so this  probably wasn't intended as a final docstring, but would you mind having a go at rewriting this docstring to follow [PEP 257](https://www.python.org/dev/peps/pep-0257/)? I think it might be clearer. Some particular details from PEP 257:

* First line of docstring should be a concise summary that fits on one line, followed by a blank line and then the more detailed description.

  (Tip: I think it's okay to just say something like `Set ``setting`` in the Pyramid settings` for the first line, details like how an envvar or default might be used, or the setting might be left unchanged, can go in the rest of the docstring.)

* The first line should "prescribe the function or method's effect as a command ("Do this", "Return that"), not as a description; e.g. don't write "Returns the pathname ...". So in your example it would be "Set the setting in settings" rather than "Set**s** the setting in settings".

  This may seem nit-picky but generally imperative tense leads to shorter and clearer sentences, my example docstring at the bottom of this comment might make this clearer. Another way of putting this is: write your docstrings (and comments, commit messages, pull request descriptions, ...) like pseudo-code. (We also use this in [our git commit style](https://chris.beams.io/posts/git-commit/) btw.)

* Since this method has a lot of parameters I think it would be worth documenting each of them individually using `:param foo: ...` and `:type foo: ...` lines in the docstring. These `:param` and `:type` etc things are Sphinx "info field lists", [docs here](http://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html#info-field-lists)

* Since this method also raises a certain exception under certain condition I think it'd be worth documenting that with `:raises SettingError: ...` too (see info field list docs link above)

General docstring writing tip: break it up into small pieces! It will make things clearer to separate different cases and details into different paragraphs. You've started to do this after the first long line of your docstring, but haven't left blank lines between the lines. For example `"Sets the specified setting in settings to the value of the env_var, or if the env_var isn't present and a default value is specified; to the default value.."` is probably clearer broken up case-by-case something like this:

> Set `settings[name]` to `envvar` or `default`.
> 
> Environment variables override config file settings and defaults: if the named `envvar` is in the environment then set `settings[name]` to the value of the `envvar`.
> 
> Config file settings override defauits: if the `envvar` isn't in the environment, and the named setting is already present in settings, then leave `settings[name]` unchanged.
> 
> Defaults are used if the setting isn't in the environment or the config file: if  `envvar` isn't in the environment and `name` isn't in settings then set `settings[name]` to the given `default` value.
> 
> If  `envvar` isn't in the environment and `name` isn't in settings and no `default` is given then leave `settings[name]` unset. If this happens and `required=True` then raise `SettingError`.

(Note that I'm also using the actual names of the method params in the docstring, and even using `settings[name]`, I think this makes it much clearer than saying indirect things like "the specified setting".)

h/settings.py
```diff
@@ -13,6 +13,52 @@ class SettingError(Exception):
     pass
 
 
+class SettingsManager(object):
+    def __init__(self, settings, environ):
+        self.settings = settings
+        self.environ = environ
+
+    def set(self,
+            setting,
```
seanh: To avoid confusion with `settings` maybe we should rename `setting` to `name`?

h/settings.py
```diff
@@ -13,6 +13,52 @@ class SettingError(Exception):
...
+        overriden with the provided default value.
+        """
+        # If the environment var doesn't exist, there is no default set or provided,
+        # and the environment var is required, raise an error.
+        if (required and not default
+                and env_var not in self.environ
+                and setting not in self.settings):
+            raise SettingError(
+                'error parsing environment variable '
+                '{varname} not found'.format(varname=env_var))
```
seanh: Can you try to re-write this code to avoid any compound `if` statements like this (three `and`s). I think the problem you've got into is by trying to handle the case of a missing required setting first. If you handle the cases in the opposite order I think it might look much simpler. Would something like this work:

```python
def set(...):

    def maybe_warn():
        if deprecated:
            log.warn(...)

    if envvar in environ:
        settings[name] = environ[envvar]
        maybe_warn()
    elif name in settings:
        maybe_warn()
    elif default:
        settings[name] = default
    elif required:
        raise SettingError()
```

tests/h/settings_test.py
```diff
@@ -98,3 +98,63 @@ def test_mandrill_settings():
...
 def test_mandrill_settings_unset():
     assert settings.mandrill_settings({}) is None
+
+
+class TestSettingsManager(object):
+    def test_does_not_emit_warnings_when_deprecated_setting_is_not_used(self, log):
+        settings_manager = settings.SettingsManager({}, {})
+        settings_manager.set('foo', 'FOO', deprecated_msg='what to do instead')
+
+        assert not log.warn.called
```
seanh: I think you might want to use `log.warn.assert_not_called()` instead here because you'll get a better failute message if it was called. See <https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.assert_not_called>. Maybe test this though: it's possible that pytest intercepts the basic `assert not log.warn.called` and replaces it with a better message.

seanh: Oh, but wait: actually it shouldn't be necessary to mock `log` at all because it's better to use `pytest.warns`. See: https://docs.pytest.org/en/latest/warnings.html#assertwarnings

tests/h/settings_test.py
```diff
@@ -98,3 +98,63 @@ def test_mandrill_settings():
...
+
+    def test_emits_warnings_when_deprecated_setting_is_used(self, log):
```
seanh: In test names I think `*_warns_*` (and `*_doesnt_warn_*`) would be clearer and shorter than `*_emits_warning_*`

tests/h/settings_test.py
```diff
@@ -98,3 +98,63 @@ def test_mandrill_settings():
...
+        settings_manager = settings.SettingsManager({}, {'FOO': 'bar'})
+        settings_manager.set('foo', 'FOO', deprecated_msg='what to do instead')
+
+        result = settings_manager.settings['foo']
+
+        assert result == 'bar'
+        log.warn.assert_called_once_with('use of foo is '
+                                     'deprecated: what to do instead')
```
seanh: This looks like two separate tests to me:

1. It still sets the setting in `settings`, even if it's deprecated
2. It warns if its deprecated

tests/h/settings_test.py
```diff
@@ -98,3 +98,63 @@ def test_mandrill_settings():
...
+        ('foo', 'FOO', asutf8, {'FOO': 'bar'}, False, None, asutf8('bar')),
+        ('app_port', 'PORT', int, {'PORT': '123'}, False, None, 123),
+
+        # Should overide the value when a default is provided
+        ('foo', 'FOO', str, {}, True, 'boo', 'boo'),
+
+        # Should not error when required and a default is already set
+        ('foo', 'FOO', str, {}, True, None, None),
+
+    ))
```
seanh: Great use of parametrize

tests/h/settings_test.py
hmstepanek: While I would love to use pytest warn it's proving not so easy. Here's a link to what I believe is the problem I'm running into. https://github.com/pytest-dev/pytest/issues/2917 Maybe you can figure it out but I'm not sure it's worth spending several hours for such little reward. It seems like the default config for the filters in py2.7 ignores warnings and so we have to override the filters but I tried that and it still didn't work so I'm at a loss. Not sure it's worth blocking the issue for.

hmstepanek:
@seanh Can you take another look. I think that's probably enough stuff for this PR. I haven't moved any settings around yet but I think that can be done in a separate PR as there's quite a bit of stuff in here already. What do you think?

seanh:
Pylint is also complaining that `set()` has too many arguments (it also runs some checks for bad code design: methods that are too long, classes that have too many methods, etc.) But unless we can think of a way to redesign it so that `set()` was 5 or less arguments (including `self`) I think we might just have to silence that one with a `# pylint: disable=too-many-arguments` comment in the code.

h/settings.py
```diff
@@ -13,54 +14,51 @@ class SettingError(Exception):
...
-            self.type = type
-        else:
-            self.type = str
-
-    def __call__(self, environ):
-        if self.varname in environ:
+class SettingsManager(object):
+    def __init__(self, settings={}, environ=os.environ):
+        self.settings = settings
+        self.environ = environ
```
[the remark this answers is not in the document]
hmstepanek: The settings is internal private state but is also needed by code that uses the SettingsManager to get the updated settings therefore it needs to be public. The environ can be private.

h/settings.py
```diff
@@ -13,54 +14,51 @@ class SettingError(Exception):
...
             try:
-                value = self.type(environ[self.varname])
+                self.settings[name] = type(val)
             except ValueError:
                 raise SettingError('error parsing environment variable '
                                    '{varname}={value!r} as {typename}'.format(
-                                       varname=self.varname,
-                                       typename=self.type.__name__,
-                                       value=environ[self.varname]))
-            return {self.setting: value}
```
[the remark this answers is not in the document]
hmstepanek: From the current use cases for deprecation messages in h today, the only use case is to warn on envvar deprecations, not on the name deprecation so I'm leaving this as is. I updated the docstring to make this clearer.

hmstepanek:
Ok I think I got everything @seanh

h/settings.py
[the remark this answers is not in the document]
seanh: This isn't a big deal but _does_ the calling code need to access `.settings`? The calling code already has the `settings` dict, it passes it in to `settings_manager` and `settings_manager` modifies it.

In `config.py` you have:

```python
def configure(environ=None, settings=None):
    ...
    settings = settings or {}
    settings_manager = SettingsManager(settings, ...)
    settings_manager.set(...)
    ...
    settings = settings_manager.settings
```

On that last line, if I'm not mistaken, the function already has a `settings` variable and then it's re-assigning the variable to the same object. In other words I don't think that last line does anything?

I'd say the API of `SettingsManager` is that you pass `settings` into `__init__()`, then you make calls to `set()` that either modify `settings` or don't, but there's no need to be able to get `settings` out again. The fact that `SettingsManager` holds its own reference to `settings` is internal state.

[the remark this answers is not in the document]
seanh: Oh, I guess this is necessitated by the fact that the `settings` argument to `SettingsManager` is optional, and it may construct its own `settings` dict?

```python
class SettingsManager(object):
    def __init__(self, settings=None, environ=None):
        self.settings = settings or {}
        ...
```

... but I don't think `settings` needs to be optional or have a default value since in the actual code we're always gonna want to pass in `request.registry.settings`. So if `settings` were just a required argument then `._settings` could be private...

h/settings.py
```diff
@@ -13,54 +14,76 @@ class SettingError(Exception):
...
+        :type name: str
+        :param envvar: the environment variable name
+        :type envvar: str
+        :param type_: the type to type cast the envvar or default value
+        :param required: True if the the pyramid config setting is required
+        :type required: bool
+        :param default: a default value to use if the envvar isn't set
+        :param deprecated_msg: a deprecated envvar setting message to display
+        :type deprecated_msg: str
+        :raises SettingsError: if required and not set or type casting fails
```
seanh: If there are two different reasons why it might raise `SettingsError` then I think you can put two `:raises SettingsError: ...` lines. That gives you more space to describe each reason, so you don't have to cram them into one line:

```
:raises SettingsError: if required=True and the setting isn't found in either environ or settings
:raises SettingsError: if type-casting the setting's value fails
```

h/settings.py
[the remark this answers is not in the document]
seanh: I'm okay with us changing the existing behaviour here, though if you want to do that in a separate PR that's fine.

Does it make sense to warn about using a deprecated setting via an envvar, but not warn about using the same deprecated setting via a config file? Maybe there are situations that I'm not considering but I'd think that a setting is either deprecated or not, and if it's deprecated then we should warn the user if they're using it, regardless of whether they're using it by setting an envvar or by editing a config file.

tests/h/settings_test.py
```diff
@@ -3,98 +3,104 @@
...
-        func.warn = mock.Mock(spec_set=[])
-
-        result = func({})
-
-        assert result == {'foo': 'bar'}
-        func.warn.assert_called_once_with('use of fake setting is '
-                                          'deprecated: what to do instead')
+def test_database_url():
+    url = 'postgres://postgres:1234/database'
+    expected = 'postgresql+psycopg2://postgres:1234/database'
```
seanh: I guess this function is already there on master, but it seems to be a function that isn't used at all. We should send a separate quickie PR to delete it

h/config.py
```diff
@@ -30,105 +27,91 @@
...
 
-        if result is not None:
-            settings.update(result)
+    if 'MANDRILL_USERNAME' in environ and 'MANDRILL_APIKEY' in environ:
+        settings_manager.set('mail.username', 'MANDRILL_USERNAME')
+        settings_manager.set('mail.password', 'MANDRILL_APIKEY')
+        settings_manager.set('mail.host', 'MANDRILL_HOST', default='smtp.mandrillapp.com')
+        settings_manager.set('mail.port', 'MANDRILL_PORT', default=587)
+        settings_manager.set('mail.tls', 'MANDRILL_TLS', default=True)
+    settings = settings_manager.settings
```
seanh: Mentioned in another comment but this `settings = settings_manager.settings` line seems to be unnecessary, can it be deleted?

h/config.py
```diff
@@ -30,105 +27,91 @@
...
-    EnvSetting('mail.default_sender', 'MAIL_DEFAULT_SENDER'),
-    EnvSetting('mail.host', 'MAIL_HOST'),
-    EnvSetting('mail.port', 'MAIL_PORT', type=int),
-    EnvSetting('sqlalchemy.url', 'DATABASE_URL', type=database_url),
-    EnvSetting('statsd.host', 'STATSD_HOST'),
-    EnvSetting('statsd.port', 'STATSD_PORT', type=int),
-    EnvSetting('statsd.prefix', 'STATSD_PREFIX'),
+    settings_manager.set('broker_url', 'BROKER_URL')
+    settings_manager.set('es.client_poolsize', 'ELASTICSEARCH_CLIENT_POOLSIZE',
+                         type=int)
```
seanh: I think you renamed `type` to `type_` but are still using `type` in `config.py` (if I run `make dev` it crashes)

h/settings.py
```diff
@@ -13,54 +14,76 @@ class SettingError(Exception):
...
+        if envvar in self._environ:
+            if deprecated_msg:
+                log.warn('use of envvar %s is deprecated: %s',
+                         envvar,
+                         deprecated_msg)
+            val = self._environ[envvar]
+            cast_message = "environment variable {}={}".format(envvar, val)
+        elif default and name not in self.settings:
+            val = default
+            cast_message = "{}'s default {}".format(name, val)
```
seanh: After fixing `type_` in `config.py` and running `make dev` again I'm getting a `UnicodeDecodeError` here. At least some of the settings (for example `DEFAULT_SALT`) are byte strings rather than unicode strings, but your error message here is unicode, so the string formatting crashes.

hmstepanek:
It’s actully not a change to pre-existing behavior, this was the pre-existing behavior. (It only warned about setting the envvar and the warning (if you look at the config.py file) is about the envvar, not the setting name. Perhaps we may want to warn about the actual pyramid setting being deprecated at some point but there isn’t a need currently so I’d like to wait on that.

hmstepanek:
If you don’t pass in settings then you rely on the SettingsManager to set the settings={} and you don’t have a reference to the settings outside of the class therefore settti gs should be public. You may say we don’t use it like that right now we always pass it in which I suppose it accurate. I’d also point out tho that we could infact use it like that as I don’t think the caller of configure in config.py actually passes in settings.

h/settings.py
```diff
@@ -13,54 +14,76 @@ class SettingError(Exception):
...
-            self.type = type
-        else:
-            self.type = str
-
-    def __call__(self, environ):
-        if self.varname in environ:
+class SettingsManager(object):
+    def __init__(self, settings=None, environ=None):
+        self.settings = settings or {}
+        self._environ = environ or os.environ
```
seanh: There's a bug here which is my fault. For example here it should set `settings["foo"]` to `"DEFAULT"` but it doesn't:

```python
>>> environ = {}
>>> settings = {}
>>> sm = SettingsManager(settings, environ)
>>> sm.set("foo", "FOO", default="DEFAULT", required=True)
>>> settings
{}
```

It can also fail to raise `SettingsError` for missing required settings in some cases.

The problem is `settings = settings or {}`. If `settings` is `None` then it will create its own `settings` dictionary by default, which is what you intended. But if `settings` is an empty dictionary `{}` that is also falsey so it will also create its own dictionary instead of modifying the passed in one! And so the `settings` that the calling code passed in does not get modified.

The fix is:

```
def configure(environ=None, settings=None):
    if environ is None:
        environ = os.environ
    if settings is None:
        settings = {}
```

(Although personally I think `settings` should just be required and not have a default.)

tests/h/settings_test.py
```diff
@@ -3,98 +3,104 @@
...
+            settings_manager.set('foo', 'FOO', deprecated_msg='what to do instead')
+        assert not caplog.records
 
+    def test_sets_value_when_deprecated_setting_is_used(self):
+        settings_manager = settings.SettingsManager({}, {'FOO': 'bar'})
+        settings_manager.set('foo', 'FOO', deprecated_msg='what to do instead')
 
-def asutf8(str):
-    return str.encode()
+        result = settings_manager.settings['foo']
```
seanh: Should the tests be accessing `settings_manager.settings` like this? I thought the contract of `SettingsManager` was that `config.py` passes in `request.registry.settings` and `SettingsManager` modifies it? So the test should hold its own reference to the settings dict that it passed in, and should test that the dict was modified.

tests/h/settings_test.py
```diff
@@ -3,98 +3,104 @@
...
+        # Should not error when required and a default is already set
+        ({'foo': 'foo'}, 'foo', 'FOO', str, {}, True, None, 'foo'),
 
-    with pytest.raises(settings.SettingError):
-        func({'PORT': 'notanint'})
+        # Should not overide default if default is already set
+        ({'boo': 'boo'}, 'boo', 'BOO', str, {}, True, 'foo', 'boo'),
 
+    ))
+    def test_env_setting(self, config, setting, varname, type_, environ, required, default, expected):
```
seanh: Here we seem to be using the name `config` for the `settings` argument? I think it's confusing to change names around like that. I can see that the name `settings` would conflict with the imported `settings` module. Maybe the test should just do `from h.settings import SettingsManager`?

tests/h/settings_test.py
```diff
@@ -3,98 +3,104 @@
...
+        ({'foo': None}, 'foo', 'FOO', str, {}, False, None, None),
 
-    # Should coerce the result using the passed type
-    ('foo', 'FOO', asutf8, {'FOO': 'bar'}, {'foo': b'bar'}),
-    ('app_port', 'PORT', int, {'PORT': '123'}, {'app_port': 123}),
-))
-def test_env_setting(setting, varname, type, environ, expected):
-    func = settings.EnvSetting(setting, varname, type)
+        # Should return the setting as a string when the env var is set
+        ({}, 'foo', 'FOO', str, {'FOO': 'bar'}, False, None, 'bar'),
```
seanh: I find these long tuples really hard to read (trying to see which values go with which parameters of the test method below), it's too many params to be readable. Difficult for me to read this and try to figure out whether all the possible cases have been tested. What about you?

I wonder if to make it easier to read we should parametrize this test with a dict like:

```python
@pytest.mark.parametrize('params', [
    dict(settings={"foo": None}, name="foo", envvar="FOO", type_=str, environ={}, required=False, default=None, expected=None),
    dict(...),
    dict(...),
    ...
])
```

Then the test would receive a single `params` dict that it'd have to read the individual params from by name, instead of receiving separate arguments.

tests/h/settings_test.py
seanh: `setting` is used for the `name` param here too

h/settings.py
```diff
@@ -36,11 +36,12 @@ def __init__(self, settings=None, environ=None):
         :type environ: Dict[str, str]
         """
 
-        if settings is None:
-            settings = {}
         if environ is None:
             environ = os.environ
-        self.settings = settings
+
+        self.settings = {}
```
hmstepanek: This is slick.

robertknight:
I have rebased this on master to incorporate the `es.url` setting and made some revisions after various discussions on Slack. There are some slightly subjective changes here but it resolves the issues that Sean raised in a previous review.

- I split a parametrised test case with many parameters into smaller test cases for readability. This is different than what Sean suggested (using a dict of params) but it solves the problem.
- I added some extra docstrings to `SettingsManager` to explain what the class is for, and reworded the docstring for `set` in a way that I hope is clearer.
- I modified `SettingsManager` to clone its input dict rather than modifying it directly, since I think that is less surprising behaviour.

I think this is good to go now but I'll take another look over with fresh eyes first thing tomorrow.

robertknight:
One change I considered making but didn't in the end is to require that all settings are defined via `settings_manager.set` by storing the settings from config vars in a different map which `set` reads from. The reason is that there are settings in the config files which are defined and accessed by Pyramid rather than h, eg. `pyramid.reload_templates`. These settings still need to be preserved and passed to the `Configurator` constructor.

tests/h/settings_test.py
```diff
@@ -2,99 +2,122 @@
...
-        result = func({})
+        assert result == 'bar'
 
-        assert result is None
-        assert not func.warn.called
+    def test_set_warns_when_deprecated_setting_is_used(self, caplog):
+        with caplog.at_level(logging.WARN):
+            settings_manager = SettingsManager({}, {'FOO': 'bar'})
+            settings_manager.set('foo', 'FOO', deprecated_msg='what to do instead')
+        assert 'what to do instead' in caplog.text
```
seanh: We're missing a test that it does not warn when a non-deprecated setting is used

h/settings.py
```diff
@@ -13,54 +14,90 @@ class SettingError(Exception):
...
+            required=False,
+            default=None,
+            deprecated_msg=None):
+        """
+        Resolve a typed value for a setting.
+
+        Sets the setting `name` to the value from the highest priority source
+        and coerces the result to a typed value using `type_`. The sources
+        in priority order are: 1) Environment variables (`envvar`), 2)
+        Configuration files (passed to the constructor), 3) Defaults (`default`).
```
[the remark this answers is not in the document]
seanh: I wonder if just the first two sentences from my suggested docstring above (plus the `:param`s) would be enough...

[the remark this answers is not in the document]
seanh: > Set settings[name] to the right value.
> 
> Set `settings[name]` to `environ[envvar]`, or `settings[name]` (i.e. leave it unchanged), or `default`, or leave it unset, in that order. `environ` overrides `settings`, and `settings` overrides `default`).

This seems pretty concise and concrete. The `:param`s for `required`, `type_` and `deprecated_msg` can then explain what those do, they don't need to be in the main docstring.

h/settings.py
```diff
@@ -13,54 +14,90 @@ class SettingError(Exception):
...
+    Configuration setting resolver.
+
+    SettingsManager resolves settings from various sources into the final typed
+    values used when the app runs. It also provides a way to check for missing
+    required settings or use of deprecated settings.
+
+    The resolved settings are available via the `settings` attribute.
+    """
+
+    def __init__(self, settings={}, environ=None):
```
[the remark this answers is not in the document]
robertknight: > There's a bug here - the default value for settings should be None, not {}, due to the notorious Python mutable default argument behaviour. 

What is the bug? The default value is never exposed as a reference (it is merely used in the call to `self.settings.update(settings)` so it cannot be modified.

h/settings.py
```diff
@@ -13,54 +14,90 @@ class SettingError(Exception):
...
+        :type settings: Dict[str,str]
+        :param environ: Environment variable mappings
+        :type environ: Dict[str, str]
+        """
+
+        if environ is None:
+            environ = os.environ
+
+        self.settings = {}
+        self.settings.update(settings)
```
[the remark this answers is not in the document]
robertknight: I believe that in theory `settings` can contain mutable objects but not in practice in our app.

h/settings.py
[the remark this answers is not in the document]
robertknight: > Set `settings[name]` to `environ[envvar]`

This is not exactly accurate because of the type-casting that happens. I've revised the description along these lines though.

tests/h/settings_test.py
```diff
@@ -2,99 +2,122 @@
...
-        'mail.host': 'smtp.mandrillapp.com',
-        'mail.port': 587,
-        'mail.tls': True,
-    }
+    def test_defaults_settings_to_empty_dict(self):
+        settings_manager = SettingsManager()
+        assert settings_manager.settings == {}
 
-    assert settings.mandrill_settings(environ) == expected
+    def test_defaults_environ_to_osenviron(self, environ):
```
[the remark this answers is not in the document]
robertknight: Done.

robertknight:
I think I have addressed all of the above issues. I'm happy to take additional feedback but since I believe @seanh is out today, I'm going to merge this so that I can move on with marking settings as required. Any follow up changes can come in subsequent PRs.

h/settings.py
```diff
@@ -5,62 +5,98 @@
...
+            type_=str,
+            required=False,
+            default=None,
+            deprecated_msg=None):
+        """
+        Update `setting[name]`.
+
+        Update `setting[name]` using the value from the environment variable
+        `envvar`. If there is no such environment variable and `setting[name]`
+        is not already set, `setting[name]` is set to `default`.
```
seanh: :+1:

h/settings.py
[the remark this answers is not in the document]
seanh: If it comes up one day I think it should be easy to fix with `copy.deepcopy()`

h/settings.py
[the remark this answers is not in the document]
seanh: Hmm.. I think you're right! Because `__init__()` never modifies the default `settings` dict then, _as long as `settings` never contains any mutable sub-objects_, there's technically nothing broken here. I was wondering why none of the tests seemed to be broken.

I think I agree with pylint that mutable default args should just be avoided in Python though - it would make it very easy for later changes to this class or I think even for some code that uses a list or dict as a setting value to introduce a really subtle bug into the tests. So glad to see you've removed this.

Issue #5030, opened 2018-05-17
WebSocket server does not scale well with increasing annotation write rate


robertknight:
During recent load testing using a 1:5 mix of create annotation/search API requests I found that most of the CPU usage on the machines was by the websocket process.

The TL;DR is that as the write-rate to an h instance increases, the amount of work that each websocket process has to do scales linearly. With a rather modest write rate (10 rps) this dominated the CPU usage of an 8 x m4.large cluster.

We should rearchitect the websocket to fix this problem.

## Method

I setup a cluster of 16 h containers running on 8 x m4.large EC2 instances and ran tests using scripts from https://github.com/hypothesis/load-testing. Each h container used default settings for number of web/worker processes.

The `locustfile.py` script was run with 650 simulated concurrent users. After the test had warmed up, there were about ~10 `POST /api/annotations`  requests/second and ~40 `GET /api/search` requests/second.

## Results

Logging into the EC2 machines showed that most of the CPU consumption was from the websocket process in each container. Disabling the websocket process by deploying a new container image with a modified `conf/supervisord.conf` file resulted in the same cluster size scaling to 1900 simulated concurrent users.

## Discussion

h uses RabbitMQ (via CloudAMQP in production) as follows:

1. Each h container runs a websocket server
2. Each Hypothesis client (embed or browser extension) establishes a websocket connection with the websocket server in one of the running h instances
3. Each websocket server process establishes a RabbitMQ queue per _topic_ with the name `realtime-{topic}-{random ID}`, eg. `realtime-annotation-BK5s8pKb3fA` on the `realtime` exchange.
4. There is one _topic_ per category of websocket message sent via the API ('user', 'annotation')
5. Web/worker processes publish messages to the `realtime` exchange specifying the topic and message payload when interesting events happen (eg. user joined/left a group, user created an annotation)
6. The RabbitMQ exchange delivers each message to each queue associated with a particular topic. This means that if there are 4 h container instances running and one annotation is created, one message will be delivered to the exchange, which will be routed to 4 queues and consumed by 4 websocket processes.

This means that as the write rate to the h instance goes up, the amount of work done by the websocket process in _each_ container increases,  regardless of how many containers there are.

[...]

PR #5035, opened 2018-05-22
Support Amazon SQS as the message transport for Celery

robertknight:
We have partners interested in using Amazon SQS rather than RabbitMQ as the [message broker for Celery](http://docs.celeryproject.org/en/latest/getting-started/brokers/index.html). This will save them the operational hassle of managing additional infrastructure (either self-hosted or using eg. CloudAMQP) since they already use SQS.

Amazon SQS has a more limited set of features than RabbitMQ. Fortunately what it does support is enough for executing h's background and periodic tasks. It [won't be able to support](https://github.com/hypothesis/product-backlog/issues/652#issuecomment-390642479) the current implementation of the realtime API.

This PR adds support for SQS as follows:

- Add the boto3 and pycurl libraries required by Kombu's SQS support. Celery provides a `celery[sqs]` convenience bundle that _should_ include these but we can't use that because it ships the wrong version of boto (boto 2 instead of 3) in Celery 4.1.x (this is fixed in the next Celery release)
- Configure Celery to use SQS from the same region as the current EC2 instance when using SQS
- Disable the realtime API when using SQS as a message broker by replacing `request.realtime` with a dummy implementation. In future we'll probably want to rearchitect the websocket server [for performance reasons](https://github.com/hypothesis/h/issues/5030) and could consider how to make it possible to use alternative backends (eg. Amazon SNS) at that point.

Supporting use of SQS for periodic tasks executed by Celery beat will require changes to [h-periodic](https://github.com/hypothesis/h-periodic) as well.

A limitation of SQS is that it can't be tested entirely on a local system since SQS isn't open source, although you can configure your local dev instance to use SQS if you provide it with appropriate credentials in the `BROKER_URL`. I've been testing this out by deploying Docker images to the `load-testing` ECS cluster in our AWS account.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -35,6 +35,7 @@ COPY requirements.txt ./
 # Install build deps, build, and then clean up.
 RUN apk add --no-cache --virtual build-deps \
     build-base \
+    curl-dev \
     libffi-dev \
     postgresql-dev \
     python-dev \
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -32,7 +32,7 @@ node {
                                          "-e ELASTICSEARCH_URL=${elasticsearchHost} " +
                                          "-e TEST_DATABASE_URL=${databaseUrl}") {
                 // Test dependencies
-                sh 'apk add --no-cache build-base libffi-dev postgresql-dev python-dev'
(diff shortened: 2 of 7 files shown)


seanh:
Closing this until we come back to it later...

PR #5037, opened 2018-05-22
Add tests for testing index and delete methods in search.

sheetaluk:


tests/h/search/index_test.py
```diff
@@ -77,6 +77,14 @@ def test_it_can_index_an_annotation_with_a_document_with_no_title(self, factorie
 
         assert "title" not in get(annotation.id)["document"]
 
+    def test_it_notifies(self, AnnotationTransformEvent, factories, notify, index, search):
+        annotation = factories.Annotation.build(userid="acct:someone@example.com")
+
+        index(annotation)
+
+        event = AnnotationTransformEvent.return_value
+        notify.assert_called_once_with(event)
```
seanh: I think you also need to test that the `AnnotationTransformEvent` was initialized with the right arguments, by testing `AnnotationTransformEvent.assert_called_once_with(...)`

tests/h/search/index_test.py
```diff
@@ -94,6 +102,111 @@ def test_you_can_filter_annotations_by_creation_time(self, factories, index, sea
...
+    def test_you_can_filter_annotations_by_tags(self, factories, index, search):
+        annotation = factories.Annotation.build(id="test_annotation_id", tags=["tag1", "tag2"])
+
+        index(annotation)
+
+        response1 = search.filter("term", tags=["tag1"]).execute()
+        response2 = search.filter("term", tags=["tag2"]).execute()
+
+        assert SearchResponseWithIDs([annotation.id]) == response1
+        assert SearchResponseWithIDs([annotation.id]) == response2
```
seanh: I'm wondering what is the difference between `"tags"` and `"tags_raw"`? Looking in `config.py` it looks like it's that `"tags"` uses `uni_normalizer`: https://github.com/hypothesis/h/blob/61fe02289488e0bf88d54aa37dae51a8300c00a5/h/search/config.py#L31-L32

This must enable some filtering functionality for `"tags"` that isn't enabled for `"tags_raw"`. Can we figure out what that is, and write a test for it? Can be a separate pr.

tests/h/search/index_test.py
```diff
@@ -94,6 +102,111 @@ def test_you_can_filter_annotations_by_creation_time(self, factories, index, sea
...
+
+        index(annotation)
+
+        response = search.filter("term", shared=False).execute()
+
+        assert SearchResponseWithIDs([annotation.id]) == response
+
+    def test_you_can_filter_annotations_by_thread_ids(self, es_client, factories, index, search):
+        annotation1 = factories.Annotation.build(id="test_annotation_id1")
+        annotation2 = factories.Annotation.build(id="test_annotation_id2", references=[annotation1.id], thread=[annotation1])
```
seanh: Why is it necessary to pass both `references` and `thread` here?

tests/h/search/index_test.py
```diff
@@ -149,3 +270,18 @@ def __init__(self, annotation_ids):
...
+
+
+@pytest.fixture
+def index(es_client, pyramid_request):
+    def _index(*annotations):
+        """Index the given annotation(s) into Elasticsearch."""
+        for annotation in annotations:
+            h.search.index.index(es_client, annotation, pyramid_request)
+        es_client.conn.indices.refresh(index=es_client.index)
+    return _index
```
seanh: This index fixture has been moved to `tests/h/search/conftest.py](https://github.com/hypothesis/h/blob/master/tests/h/search/conftest.py) now, so you don't need it here anymore. (You'll need to rebase.)

tests/h/search/index_test.py
```diff
@@ -94,6 +102,111 @@ def test_you_can_filter_annotations_by_creation_time(self, factories, index, sea
...
+        assert SearchResponseWithIDs([annotation.id]) == response
+
+    def test_you_can_filter_annotations_by_user_raw(self, factories, index, search):
+        annotation = factories.Annotation.build(userid="acct:someone@example.com")
+
+        index(annotation)
+
+        response = search.filter("term", user_raw="acct:someone@example.com").execute()
+
+        assert SearchResponseWithIDs([annotation.id]) == response
```
seanh: What's the difference between `"user"` and `"user_raw"`? It looks like `"user"` uses the user analyzer whereas `"user_raw"` doesn't:

https://github.com/hypothesis/h/blob/61fe02289488e0bf88d54aa37dae51a8300c00a5/h/search/config.py#L47-L4

This must enable some kind of search filtering on `"user"`

tests/h/search/index_test.py
sheetaluk: It looks like we are using the analysed `tags`field for searching and the `tags_raw` field in TagsAggregation in query.py

tests/h/search/index_test.py
sheetaluk: It looks like we are using the analysed `user`field for searching and the `user_raw` field in UsersAggregation in query.py

tests/h/search/index_test.py
sheetaluk: We don't. Thanks for catching that.

tests/h/search/index_test.py
```diff
@@ -93,6 +106,129 @@ def test_you_can_filter_annotations_by_creation_time(self, factories, index, sea
...
+        annotation = factories.Annotation.build(userid="acct:someone@example.com")
+
+        index(annotation)
+
+        response = search.filter("term", user=user_search_str).execute()
+
+        assert SearchResponseWithIDs([annotation.id]) == response
+
+    def test_you_can_make_aggregations_on_user_raw(self, factories, index, search):
+        annotation = factories.Annotation.build(userid="acct:someone@example.com")
```
robertknight: What are we actually testing here? Does elasticsearch allow creating aggregations on _any_ field, or is there something about the field configuration which is required to support this?

seanh: This is fine. The purpose of the user_raw (and tags_raw etc) fields is to aggregate by them. We wouldn't want to test that you can filter by them (as we do with the other fields) because that isn't what these fields are for (h only uses them for aggregating).

Hopefully the tests cover the difference between the user analyzer used on the user field, and no analysis being done on the user_raw field, this difference is apparently what necessitates a separate user_raw field for aggregating.

sheetaluk: +1 to what Sean said ^

robertknight: But @sheetaluk and @seanh, my point is that with this change the test still passes:

```diff
diff --git a/tests/h/search/index_test.py b/tests/h/search/index_test.py
index b04bdaa1a..889c65ffc 100644
--- a/tests/h/search/index_test.py
+++ b/tests/h/search/index_test.py
@@ -143,7 +143,7 @@ class TestIndex(object):
 
         index(annotation)
 
-        a = elasticsearch1_dsl.A('terms', field='user_raw')
+        a = elasticsearch1_dsl.A('terms', field='wibble')
         search.aggs.bucket('user_raw_terms', a)
 
         response = search.execute()
```

On that basis, what exactly are we testing here?

robertknight:
Rebased on master to pull in the Docker build fix.

tests/h/search/index_test.py
```diff
@@ -93,6 +106,141 @@ def test_you_can_filter_annotations_by_creation_time(self, factories, index, sea
...
+
+        response = search.filter("term", user=user_search_str).execute()
+
+        assert SearchResponseWithIDs([annotation.id]) == response
+
+    def test_you_can_make_aggregations_on_user_raw(self, factories, index, search):
+        annotation_1 = factories.Annotation.build(userid="acct:someone@example.com")
+        annotation_2 = factories.Annotation.build(userid="acct:Someone@example.com")
+
+        index(annotation_1)
```
robertknight: You can pass multiple annotations to `index` - `index(annotation_1, annotation_2)`. It is slightly more efficient to do this because each call to `index` triggers a "refresh" of the search index in order to make the changes visible in search results. Indexing several annotations in one call means that only one refresh operation is needed.

tests/h/search/index_test.py
```diff
@@ -93,6 +106,141 @@ def test_you_can_filter_annotations_by_creation_time(self, factories, index, sea
...
+        index(annotation_2)
+
+        a = elasticsearch1_dsl.A('terms', field='user_raw')
```
robertknight: I probably would have opted for a slightly more verbose variable name than `a` here for clarity.

tests/h/search/index_test.py
```diff
@@ -93,6 +106,141 @@ def test_you_can_filter_annotations_by_creation_time(self, factories, index, sea
...
+
+        assert SearchResponseWithIDs([annotation.id]) == response
+
+    def test_you_can_filter_annotations_by_thread_ids(self, es_client, factories, index, search):
+        annotation1 = factories.Annotation.build(id="test_annotation_id1")
+        annotation2 = factories.Annotation.build(id="test_annotation_id2", thread=[annotation1])
+
+        index(annotation1, annotation2)
+
+        response1 = search.filter("term", thread_ids=[annotation1.id]).execute()
```
robertknight: Any reason to call this `response1` as opposed to just `response`?

tests/h/search/index_test.py
```diff
@@ -120,8 +268,25 @@ def _get(annotation_id):
...
+    def AnnotationTransformEvent(self, patch):
+        return patch('h.search.index.AnnotationTransformEvent')
+
+
+class TestDelete(object):
+    def test_annotation_is_marked_deleted(self, es_client, factories, index, search):
+        annotation = factories.Annotation.build(id="test_annotation_id")
+
+        index(annotation)
+        result = es_client.conn.get(index=es_client.index,
```
robertknight: We're going to need to abstract the operation of fetching an annotation from the index in future in order to be able to run tests against ES 1 and ES 6.2.

tests/h/search/index_test.py
```diff
@@ -93,6 +106,141 @@ def test_you_can_filter_annotations_by_creation_time(self, factories, index, sea
...
+        index(annotation_1)
+        index(annotation_2)
+
+        a = elasticsearch1_dsl.A('terms', field='user_raw')
+        search.aggs.bucket('user_raw_terms', a)
+
+        response = search.execute()
+
+        user_bucket_1 = next(bucket for bucket in response.aggregations.user_raw_terms.buckets if bucket["key"] == "acct:someone@example.com")
+        user_bucket_2 = next(bucket for bucket in response.aggregations.user_raw_terms.buckets if bucket["key"] == "acct:Someone@example.com")
```
robertknight: Please try to keep lines < 100 chars at most, as they are fiddly to read in GitHub's diff viewer otherwise.

PR #5038, opened 2018-05-22
Allow custom timestamps in Annotation.build()

seanh:
factories.Annotation.build() was setting created and updated itself,
even if the calling code passed them in. In this example created and
updated would be ignored and different times used:

    factories.Annotation.build(
        created=some_datetime,
        updated=another_datetime)

Fix factories.Annotation.build() to only write created and updated if
they aren't passed in.


sheetaluk:
LGTM

PR #5039, opened 2018-05-23
Move query_test.py to old_query_test.py

seanh:
Making way for writing new, Elasticsearch-integrated tests in a new
query_test.py file.


sheetaluk:
LGTM.

PR #5040, opened 2018-05-23
`h/groups/search.py` is gone - `GroupAuthFilter` has been moved into

seanh:
`h/search/query.py` along with most of the other filters.

`TestGroupAuthFilter` also moves into `old_query_test.py`.

`h/groups/__init__.py` no longer needs to add `GroupAuthFilter` to the
default search filters, since `Search` itself adds it now. This also
means that `tests/h/groups/__init___test.py` is no longer needed and can
be deleted.

This now means that `GroupAuthFilter` is in use whenever `Search` is
used in tests, such as the tests in `core_test.py` and
`old_core_test.py`. Previously `GroupAuthFilter` was not added to the
search filters during these tests, even though it _is_ added to the
filters by default in production.

As a result, some things had to be done to fix the tests:

* Added the `request.user` property to the `pyramid_request` fixture for
  these tests, because `GroupAuthFilter` uses it.

* Added a fixture to register the "group" service during the tests.
  Because `GroupAuthFilter` uses it.

* Added the test annotations to the `__world__` group. Previously these
  test annotations didn't belong to any group (`test_annotation.groupid`
  was `None`), which isn't even valid.


seanh:
That's the next step yes.

I think `h.search` already contained `GroupAuthFilter` and so already depended on the groups service - it's just the `GroupAuthFilter` was physically contained in a different file and was added to `h.search.Search`'s default filters by an indirect mechanism.

PR #5041, opened 2018-05-23
Add default group to factories.Annotation.build()

seanh:
Annotations from factories.Annotation() get the `"__world__"` group by
default, but annotations from factories.Annotation.build() get None:

    >>> factories.Annotation().groupid
    u'__world__'

    >>> factories.Annotation.build().groupid
    None

This is because the default group is implemented by models.Annotation
using a sqlalchemy default and server_default, so it doesn't get set
until the annotation is added to the DB.

This can cause tests to crash because code expects all annotations to
have a group.

Fix factories.Annotation.build() to set groupid to `"__world__"` by
default. You can still pass a custom groupid if you want:

    >>> factories.Annotation.build().groupid
    u'__world__'

    >>> factories.Annotation.build(groupid=u"foo").groupid
    u'foo'


sheetaluk:
LGTM.

PR #5042, opened 2018-05-23
Add request.user to pyramid_request fixture

seanh:
h adds a custom `.user` property to the request but Pyramid's
DummyRequest (which is what we use for our pyramid_request fixture)
doesn't have this property. As a result lots of tests need to add
`pyramid_request.user` so that code doesn't crash.

Add `pyramid_request.user` by default so tests no longer need to do
this.

### DIFF
--- a/tests/h/conftest.py
+++ b/tests/h/conftest.py
@@ -246,6 +246,7 @@ def pyramid_request(db_session, fake_feature, pyramid_settings):
     request.params = MultiDict()
     request.GET = request.params
     request.POST = request.params
+    request.user = None
     return request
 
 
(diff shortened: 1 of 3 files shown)


tests/h/conftest.py
```diff
@@ -246,6 +246,7 @@ def pyramid_request(db_session, fake_feature, pyramid_settings):
     request.params = MultiDict()
     request.GET = request.params
     request.POST = request.params
+    request.user = None
```
robertknight: I think it would probably be a good idea to have some separation between the request methods/properties which are part of Pyramid but are being overridden vs those which are custom in our code here.

PR #5044, opened 2018-05-24
Move NipsaFilter into search

seanh:
And also get rid of the filters registry stuff as all filters are in h.search now.

Depends on https://github.com/hypothesis/h/pull/5045


seanh:
Testing

PR #5045, opened 2018-05-24
Use group_service fixture in all search tests automatically

seanh:
Enable the group_service fixture automatically for all search tests. A lot of search tests need this fixture, so avoid them all having to enable it one by one.

### DIFF
--- a/tests/h/search/conftest.py
+++ b/tests/h/search/conftest.py
@@ -1,9 +1,19 @@
 # -*- coding: utf-8 -*-
 from __future__ import unicode_literals
 
+import mock
 import pytest
 
 import h.search.index
+from h.services.group import GroupService
+
+
+@pytest.fixture(autouse=True)
+def group_service(pyramid_config):
+    group_service = mock.create_autospec(GroupService, instance=True, spec_set=True)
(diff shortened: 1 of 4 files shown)


tests/h/search/conftest.py
```diff
@@ -1,9 +1,19 @@
...
 from __future__ import unicode_literals
 
+import mock
 import pytest
 
 import h.search.index
+from h.services.group import GroupService
+
+
+@pytest.fixture(autouse=True)
```
robertknight: Is this autouse scoped to tests in the `h.search` package? It looks like it should be a cheap fixture to execute, but I'd like to check my understanding.

seanh: Yep, it's scoped to `h.search`.

You can also put an autouse fixture in a module or class and it'll be scoped to that module or class. In h we use `@pytestmark` / `pytestmark` to turn on fixtures for all tests in a class / module instead though, because it allows you to name fixtures defined outside of that class / module as well.

To turn on a fixture for an entire package though I think autouse is the only way.

Hence `conftest.py` files are the only place where you'll find autouse fixtures in h.

PR #5046, opened 2018-05-24
Add tests for TopLevelAnnotationsFilter

seanh:
Depends on https://github.com/hypothesis/h/pull/5045

These tests are kind of duplicative of the existing [tests for `separate_replies`](https://github.com/hypothesis/h/blob/49fd4cb4e415ac79e8144715bb2f677f3926f678/tests/h/search/core_test.py#L124). The `separate_replies` tests also test that replies are filtered out of `annotation_ids` but annotations are left in. But those are tests for the `separate_replies` argument, which is partly implemented by `TopLevelAnnotationsFilter`, partly by `RepliesMatcher`, and partly by code in `Search` itself. The fact that `TopLevelAnnotationsFilter` and `RepliesMatcher` are used to implement `separate_replies` is an internal implementation detail (`separate_replies` could be refactored to no longer use `TopLevelAnnotationsFilter` and its tests would still pass).

The tests in this PR are tests for `TopLevelAnnotationsFilter` when used directly (without `separate_replies`) as is [done by activity pages](https://github.com/hypothesis/h/blob/362143c17d7cd090255d9d36d6fecfbbb750218e/h/activity/query.py#L171).

Also, `TopLevelAnnotationsFilter` has to have a `TestTopLevelAnnotationsFilter` where you would expect one to be, otherwise it would look like there were no tests for it. Meanwhile, the filtering out of replies (while leaving annotations in) kind of has to be tested by the `separate_replies` tests otherwise `separate_replies` isn't fully tested. (Because of the way the code is written you can't really write a "Test that it uses `TopLevelAnnotationsFilter`" test for `separate_replies` that uses a mock `TopLevelAnnotationsFilter`, you would have to mock out `Builder` or something.)

### DIFF
--- a/h/activity/query.py
+++ b/h/activity/query.py
@@ -18,9 +18,9 @@ from h.search import parser
 from h.search.query import (
     AuthorityFilter,
     TagsAggregation,
-    TopLevelAnnotationsFilter,
     UsersAggregation,
 )
+from h.search import TopLevelAnnotationsFilter
 
 
 class ActivityResults(namedtuple('ActivityResults', [
--- a/h/search/__init__.py
+++ b/h/search/__init__.py
@@ -4,9 +4,11 @@ from __future__ import unicode_literals
 from h.search.client import get_client
 from h.search.config import init
 from h.search.core import Search
+from h.search.query import TopLevelAnnotationsFilter
 
 __all__ = (
     'Search',
+    'TopLevelAnnotationsFilter',
     'get_client',
     'init',
 )
--- a/tests/h/search/conftest.py
+++ b/tests/h/search/conftest.py
@@ -17,6 +17,21 @@ def group_service(pyramid_config):
 
 
 @pytest.fixture
+def Annotation(factories, index):
+    """Create and index an annotation.
+
+    Looks like factories.Annotation() but automatically uses the build()
+    strategy and automatically indexes the annotation into the test
+    Elasticsearch index.
+    """
+    def _Annotation(**kwargs):
+        annotation = factories.Annotation.build(**kwargs)
+        index(annotation)
+        return annotation
+    return _Annotation
+
+
+@pytest.fixture
 def index(es_client, pyramid_request):
     def _index(*annotations):
         """Index the given annotation(s) into Elasticsearch."""
@@ -24,3 +39,9 @@ def index(es_client, pyramid_request):
             h.search.index.index(es_client, annotation, pyramid_request)
         es_client.conn.indices.refresh(index=es_client.index)
     return _index
+
+
+@pytest.fixture
+def pyramid_request(es_client, pyramid_request):
+    pyramid_request.es = es_client
+    return pyramid_request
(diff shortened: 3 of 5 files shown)


h/activity/query.py
```diff
@@ -18,9 +18,9 @@
 from h.search.query import (
     AuthorityFilter,
     TagsAggregation,
-    TopLevelAnnotationsFilter,
     UsersAggregation,
 )
+from h.search import TopLevelAnnotationsFilter
```
seanh: The other `h.search.query` imports here will be changed to `h.search` imports in due time

h/search/__init__.py
```diff
@@ -5,9 +5,11 @@
 from h.search.config import init
 from h.search.core import Search
 from h.search.core import FILTERS_KEY
+from h.search.query import TopLevelAnnotationsFilter
 
 __all__ = (
     'Search',
+    'TopLevelAnnotationsFilter',
```
seanh: Adding `TopLevelAnnotationsFilter` to the official public API of `h.search`. In due time this will be done with other publicly-used filters and aggregations etc.

tests/h/search/conftest.py
```diff
@@ -6,6 +6,21 @@
...
+
+    Looks like factories.Annotation() but automatically uses the build()
+    strategy and automatically indexes the annotation into the test
+    Elasticsearch index.
+    """
+    def _Annotation(**kwargs):
+        annotation = factories.Annotation.build(**kwargs)
+        index(annotation)
+        return annotation
+    return _Annotation
```
seanh: Just moved this to where all search tests can use it

tests/h/search/query_test.py
```diff
@@ -0,0 +1,25 @@
...
+class TestTopLevelAnnotationsFilter(object):
+
+    def test_it_filters_out_replies_but_leaves_annotations_in(self, Annotation, search):
+        annotation = Annotation(shared=True)
+        reply = Annotation(references=[annotation.id], shared=True)
+
+        result = search.run({})
+
+        assert annotation.id in result.annotation_ids
+        assert reply.id not in result.annotation_ids
```
seanh: Yep. Only one simple test is needed. Hopefully man of the filters etc will be this simple to test.

tests/h/search/query_test.py
```diff
@@ -0,0 +1,25 @@
...
+
+    @pytest.fixture
+    def search(self, pyramid_request):
+        search = Search(pyramid_request)
+        search.append_filter(TopLevelAnnotationsFilter())
+        return search
```
seanh: My intention is that each test class for each filter will have its own `search` fixture that returns a `Search` instance with that filter added. The default filters are also present on this `Search` instance. (The test classes for the default filters won't need to append them since they're already appended by default. By not appending them, they will handily test the appending-by-default too.)

tests/h/search/conftest.py
```diff
@@ -14,3 +29,9 @@ def _index(*annotations):
             h.search.index.index(es_client, annotation, pyramid_request)
         es_client.conn.indices.refresh(index=es_client.index)
     return _index
+
+
+@pytest.fixture
+def pyramid_request(es_client, pyramid_request):
+    pyramid_request.es = es_client
+    return pyramid_request
```
seanh: Just moved this too

PR #5052, opened 2018-05-29
Add test for indexing of quote selector as "quote" field

robertknight:
Other selectors are indexed as a result of the current mappings in
`h.search.config` but I haven't included tests for them as I don't think
being able to search by text position or XPath is really that useful.


tests/h/search/index_test.py
```diff
@@ -241,6 +241,23 @@ def test_you_can_filter_annotations_by_thread_ids(self, es_client, factories, in
...
 
+    @pytest.mark.parametrize("quote,query", [
+        ("It is a truth universally acknowledged", "truth"),
+        ("यह एक सत्य सार्वभौमिक रूप से स्वीकार किया जाता है", "सत्य"),
+        ("quick brown fox", "QUICK"),
+    ])
+    def test_match_annotation_by_quote(self, factories, index, search, quote, query):
+        """
+        Verify that the "TextQuoteSelector" selector is indexed as the "quote" field.
+        """
```
seanh: It doesn't seem to be triggering `make lint` but this'll fail pylint / pep257: Putting a single-line docstring over multiple lines (the quotes need to go on the same line)

tests/h/search/index_test.py
```diff
@@ -309,3 +326,12 @@ def __eq__(self, search_response):
...
     return elasticsearch1_dsl.Search(using=es_client.conn).fields([])
+
+
+def _quote_selector(quote):
+    return {
+        'type': 'TextQuoteSelector',
+        'exact': quote,
+        'prefix': 'something before ',
+        'suffix': ' something after',
+    }
```
seanh: What do you think about just making `_quote_selector()` an inner function defined in `test_match_annotation_by_quote()`? Since it's only used in one place. We can move it out in future if it needs to be used in more than one place (at which point a helper method or fixture on the test class might be better). I like to try to avoid a long list of private helper functions at the bottom of a test file 'cos I don't think it ends up being very good for readability

tests/h/search/index_test.py
```diff
@@ -241,6 +241,23 @@ def test_you_can_filter_annotations_by_thread_ids(self, es_client, factories, in
 
         assert SearchResponseWithIDs([annotation2.id]) == response1
 
+    @pytest.mark.parametrize("quote,query", [
+        ("It is a truth universally acknowledged", "truth"),
+        ("यह एक सत्य सार्वभौमिक रूप से स्वीकार किया जाता है", "सत्य"),
+        ("quick brown fox", "QUICK"),
+    ])
+    def test_match_annotation_by_quote(self, factories, index, search, quote, query):
```
seanh: Can we rename this to be consistent with the other test names in this class? Something like `"test_you_can_search_within_the_quote()"`

tests/h/search/index_test.py
seanh: Are the matching `type`, `prefix` and `suffix` actually necessary here? I can see why matching query text against the quote is a useful feature to have, but can't imagine that anyone's gonna type `"TextQuoteSelector"` into their search query. (I guess I can see why matching against prefix and suffix is useful too but wonder if those are separate tests.)

robertknight: The prefix and suffix are here just to make the data match the real world scenario. The type is potentially important at indexing time in order for the indexer to know which field to copy into the "quote" field.

tests/h/search/index_test.py
robertknight: I figured that made the line a bit long (91 chars), but PEP 257 doesn't say anything about line lengths. 🤷‍♂️

tests/h/search/index_test.py
robertknight: I've inlined `_quote_selector` into this test case.

tests/h/search/index_test.py
robertknight: Done

PR #5054, opened 2018-05-29
Replace finalizers with yield in test fixtures

hmstepanek:
See https://docs.pytest.org/en/latest/fixture.html#fixture-finalization-executing-teardown-code for details. Instead of using finalizer callbacks which are a bit indirect use yield which explicitly implements setup, return value, and teardown behavior for a fixture.

### DIFF
--- a/tests/common/fixtures/elasticsearch.py
+++ b/tests/common/fixtures/elasticsearch.py
@@ -29,23 +29,18 @@ def es_connect():
 def init_elasticsearch(request):
     """Initialize the test (old) Elasticsearch index once per test session."""
     client = _es_client()
+    # Initialize the test search index.
+    search.init(es_client)
     """Connect to the newer v6.x instance of Elasticsearch once per test session"""
     es_connect()
 
-    def maybe_delete_index():
-        """Delete the test index if it exists."""
-        if client.conn.indices.exists(index=ELASTICSEARCH_INDEX):
-            client.conn.indices.delete(index=ELASTICSEARCH_INDEX)
+
+    yield es_client
 
     # Delete the test search index at the end of the test run.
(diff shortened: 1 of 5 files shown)


tests/common/fixtures/elasticsearch.py
```diff
@@ -29,34 +34,18 @@ def es_connect():
...
-    request.addfinalizer(maybe_delete_index)
-
-    # Delete the test search index at the start of the run, just in case it
-    # was somehow left behind by a previous test run.
-    maybe_delete_index()
-
-    # Initialize the test search index.
-    search.init(client)
 
+    yield es_client
```
robertknight: This code doesn't look to me like it is equivalent with what was there before. We are no longer calling removing the index at the start of the run, in case it already exists for some reason.

hmstepanek: I question why/if we actually need to do that. Since we delete the index after every use of the connection is there going to be a case where the index hasn’t been deleted?

robertknight: If the test runner crashes or gets killed during execution, the index won't be deleted. This may be rare but my experience has been that if a test depends on persistent resources, it always needs to ensure they are in a clean state at the start of a run and can't rely only on cleaning them up at the end because there is a long tail of things that can prevent cleanup code from running.

hmstepanek: Ah ok. That makes sense.

robertknight:
I think we're in agreement about the need for the existing behaviour (see [comment](https://github.com/hypothesis/h/pull/5054#discussion_r196207396)). On that basis I'm going to close this PR for now. It is still fine to use either yield fixtures or finalizers in future depending on the needs of a given fixture.

PR #5055, opened 2018-05-30
Implement mapping and Docs for es6.2 cluster

hmstepanek:
Create a new directory h.search.docs to hold our elasticsearch documents (similar to models for our db). Currently there are 3 documents; Annotation, Target, and Selector. Target and Selector are inner documents of Annotation.


hmstepanek:
Tests:
```
def test_able_to_add_annotation(new_annotation):
    article = Annotation.get(id=42)
    assert article.meta.id == '42'


@pytest.fixture
def connection():
    connections.create_connection(hosts=['http://localhost:9201'], index='hypothesis')


@pytest.fixture
def new_annotation(connection):
    """
    Add some annotations to Elasticsearch as "noise".

    These are annotations that we *don't* expect to show up in search
    results. We want some noise in the search index to make sure that the
    test search queries are only returning the expected annotations and
    not, for example, simply returning *all* annotations.

    """
    Annotation(meta={'id': 42},
               text='hello',
               paragraph='hello world!',
               groupid='__world__',
               userid = "acct:{foo}@{__world__}",
               tags=['test'],
               target = Target(
                 selector = Selector(
                    endContainer='/div[1]/article[1]/section[1]/div[1]/div[2]/div[1]',
                    endOffset=76,
                    startContainer='/div[1]/article[1]/section[1]/div[1]/div[2]/div[1]',
                    startOffset=0,
                    type='RangeSelector')
                 ),
              ).save()
```

h/search/docs/selector.py
```diff
@@ -0,0 +1,35 @@
+# -*- coding: utf-8 -*-
+from elasticsearch_dsl import (InnerDoc,
+                               Keyword,
+                               Long,
+                               Text,
+                               )
+
+
+class Selector(InnerDoc):
```
hmstepanek: We could chose to put Selector and Target in annotation.py since they are InnerDocs of Annotation. Either way when creating an Annotation you need to create a Target and Selector object and pass it in upon Annotation obj creation.

h/search/docs/annotation.py
```diff
@@ -0,0 +1,40 @@
...
+                               Field,
+                               Keyword,
+                               Text,
+                               )
+
+from h.search.target import Target
+
+
+class Annotation(DocType):
+    annotator_schema_version = Text()
```
hmstepanek: I'm not 100% sure all of these are correct-need to intensely test to verify but I'm more interested in how everyone feels about the general architecture here. @seanh @lyzadanger @robertknight @sheetaluk

hmstepanek:
@lyzadanger Will be taking pieces from this PR and https://github.com/hypothesis/h/pull/5051.

PR #5056, opened 2018-05-30
Replace HKDF implementation from cryptography

robertknight:
The `cryptography` Python library currently fails to build against the
latest version of `libressl-dev` in the Alpine release (3.7) that we are using.
`libressl-dev` is in turn a dependency of `postgresql-dev`.

Fortunately we only use `cryptography` for the well-known HKDF key
derivation function. This commit replaces the `cryptography` package
with the very small `hkdf` package which only depends on `hashlib` and
`hmac` from the Python standard library (which do use OpenSSL / LibreSSL
under the hood).

Add tests for the output of `derive_key` based on the vectors from the
HKDF RFC to verify that the outputs are the same as with the previous
implementation.

### DIFF
--- a/h/security.py
+++ b/h/security.py
@@ -3,16 +3,13 @@
 from __future__ import unicode_literals
 
 import base64
+import hashlib
 import os
 
-from cryptography.hazmat.primitives import hashes
-from cryptography.hazmat.primitives.kdf.hkdf import HKDF
-from cryptography.hazmat.backends import default_backend
+from hkdf import hkdf_expand, hkdf_extract
 from passlib.context import CryptContext
 
 DEFAULT_ENTROPY = 32
-
-backend = default_backend()
 
 # We use a passlib CryptContext to define acceptable hashing algorithms for
 # passwords. This allows us to easily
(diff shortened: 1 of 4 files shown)


h/security.py
```diff
@@ -34,18 +31,18 @@ def derive_key(key_material, salt, info):
...
     """
     if not isinstance(key_material, bytes):
         key_material = key_material.encode()
-    algorithm = hashes.SHA512()
-    length = algorithm.digest_size
-    hkdf = HKDF(algorithm, length, salt, info, backend)
-    return hkdf.derive(key_material)
+
+    pseudorandom_key = hkdf_extract(salt, key_material, hash=hashlib.sha512)
+    return hkdf_expand(pseudorandom_key, info, length=64, hash=hashlib.sha512)
```
seanh: I appreciate the variable name `pseudorandom_key` (rather than `prk` or something)

tests/h/security_test.py
```diff
@@ -95,6 +97,38 @@ def test_it_encodes_str_key_material(self):
...
+        info_bytes = unhexlify(info)
+        salt_bytes = unhexlify(salt)
+        key_bytes = unhexlify(key)
+
+        derived = derive_key(key_bytes, salt_bytes, info_bytes)
+
+        # The test vectors have key lengths above and below the fixed-sized
+        # output of `derive_key`. Only compare the corresponding prefixes.
+        compare_len = min(len(expected), 64 * 2)
+        assert hexlify(derived)[:compare_len] == expected[:compare_len].encode()
```
seanh: :+1: Great to have tests added for this. Appreciate the `_bytes` naming too, always good to know when you're dealing with bytes or unicode

robertknight:
To recap a discussion on Slack, we probably would prefer to use the cryptography library's HKDF implementation as it is a much more popular library. However it is a very small amount of code so relatively easy to validate that it does what the RFC says. We can always switch back in future and retain the tests added here.

PR #5059, opened 2018-05-30
Small fixes and cleanup.

sheetaluk:
Address comments in feedback on #5037 : 
- renamed `response1` to `response` in one of the tests since there is really just one response. 
- renamed very short variable name `a` to `user_aggregation` and `tags_aggregation` to make more sense
- broke a vey long line to increase readability
- pass multiple annotations to `index` instead of calling `index` once per annotation.  It is slightly more efficient to do this because each call to index triggers a "refresh" of the search index in order to make the changes visible in search results. Indexing several annotations in one call means that only one refresh operation is needed. (See: https://github.com/hypothesis/h/pull/5037#discussion_r191740462)


seanh:
Agree with Rob - I would have done one commit per small thing and explained what I was doing and why in each commit message.

PR #5060, opened 2018-05-30
Add tests for batch indexing annotations.

sheetaluk:
Also a minor refactor to make the following variables injectable:
ES_CHUNK_SIZE: the number of docs in one chunk sent to ES
PG_WINDOW_SIZE: how often to log how many annotations have been logged. Also used in the query for loading all annotations in batches

### DIFF
--- a/h/search/index.py
+++ b/h/search/index.py
@@ -107,28 +107,31 @@ class BatchIndexer(object):
         else:
             self._target_index = target_index
 
-    def index(self, annotation_ids=None):
+    def index(self, annotation_ids=None, windowsize=PG_WINDOW_SIZE, chunk_size=ES_CHUNK_SIZE):
         """
         Reindex annotations.
 
         :param annotation_ids: a list of ids to reindex, reindexes all when `None`.
         :type annotation_ids: collection
+        :param windowsize: the number of annotations to index in between progress log statements
+        :type windowsize: integer
+        :param chunk_size: the number of docs in one chunk sent to ES
+        :type chunk_size: integer
 
         :returns: a set of errored ids
         :rtype: set
         """
         if not annotation_ids:
-            annotations = _all_annotations(session=self.session,
-                                           windowsize=PG_WINDOW_SIZE)
+            annotations = _all_annotations(session=self.session, windowsize=windowsize)
         else:
             annotations = _filtered_annotations(session=self.session,
                                                 ids=annotation_ids)
 
         # Report indexing status as we go
-        annotations = _log_status(annotations, log_every=PG_WINDOW_SIZE)
+        annotations = _log_status(annotations, log_every=windowsize)
 
         indexing = es_helpers.streaming_bulk(self.es_client.conn, annotations,
-                                             chunk_size=ES_CHUNK_SIZE,
+                                             chunk_size=chunk_size,
                                              raise_on_error=False,
                                              expand_action_callback=self._prepare)
         errored = set()
--- a/tests/h/search/index_test.py
+++ b/tests/h/search/index_test.py
@@ -3,7 +3,9 @@ from __future__ import unicode_literals
 
 import datetime
 
+import elasticsearch1
 import elasticsearch1_dsl
+import logging
 import mock
 import pytest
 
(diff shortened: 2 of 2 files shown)


h/search/index.py
```diff
@@ -107,7 +107,7 @@ def __init__(self, session, es_client, request, target_index=None, op_type='inde
         else:
             self._target_index = target_index
 
-    def index(self, annotation_ids=None):
+    def index(self, annotation_ids=None, windowsize=PG_WINDOW_SIZE, chunk_size=ES_CHUNK_SIZE):
```
robertknight: Could you update the docstring for this method to explain what the `windowsize` and `chunk_size` parameters are for.

h/search/index.py
```diff
@@ -118,17 +118,16 @@ def index(self, annotation_ids=None):
...
-            annotations = _all_annotations(session=self.session,
-                                           windowsize=PG_WINDOW_SIZE)
+            annotations = _all_annotations(session=self.session, windowsize=windowsize)
         else:
             annotations = _filtered_annotations(session=self.session,
                                                 ids=annotation_ids)
 
         # Report indexing status as we go
-        annotations = _log_status(annotations, log_every=PG_WINDOW_SIZE)
+        annotations = _log_status(annotations, windowsize)
```
robertknight: I found it slightly clearer to read when the `log_every` argument was explicitly named rather than just relying on the position.

tests/h/search/index_test.py
```diff
@@ -310,6 +311,66 @@ def test_annotation_is_marked_deleted(self, es_client, factories, index, search)
         assert result.get('_source').get('deleted') is True
 
 
+class TestBatchIndexer(object):
+    def test_it_indexes_all_annotations(self, db_session, es_client, factories, pyramid_request):
+        annotations = factories.Annotation.create_batch(3)
+        ids = [a.id for a in annotations]
+
+        h.search.index.BatchIndexer(db_session, es_client, pyramid_request, es_client.index).index()
```
robertknight: The `BatchIndexer` construction is repeated with the same arguments in several test cases. I would suggest moving into a fixture (say `batch_indexer` that returns an instance of `BatchIndexer`).

tests/h/search/index_test.py
```diff
@@ -310,6 +311,66 @@ def test_annotation_is_marked_deleted(self, es_client, factories, index, search)
...
+                result = es_client.conn.get(index=es_client.index,
+                                            doc_type="annotation",
+                                            id=id)
+
+    def test_it_returns_errored_annotation_ids(self, db_session, es_client, factories, pyramid_request):
+        annotations = factories.Annotation.create_batch(3)
+        expected_errored_ids = set([annotations[0].id, annotations[2].id])
+
+        elasticsearch1.helpers.streaming_bulk = mock.Mock()
+        elasticsearch1.helpers.streaming_bulk.return_value = [(False, {'index': {'error' : 'some error', '_id': annotations[0].id}}),
```
robertknight: Some of these lines are pretty long, which makes them annoying to read in GitHub's diff viewer:

screenshot 2018-05-31 12 04 17

Please keep them to max ~100 chars.

sheetaluk: Hi! yes, I have a bit of linting cleanup to do.

sheetaluk:
Ready for your eyes, @robertknight ! I'll send a follow up PR your way for:

1. Deleted annotations should not be indexed
2. Annotations are converted to the expected JSON format for indexing
3. The AnnotationTransformEvent event is emitted so that other code can "decorate" annotations with additional info. eg. A flag is set if the annotation was posted by a NIPSA-d user.

tests/h/search/index_test.py
```diff
@@ -149,10 +153,10 @@ def test_you_can_make_aggregations_on_user_raw(self, factories, index, search):
 
         response = search.execute()
 
-        user_bucket_1 = next(bucket for bucket in response.aggregations.user_raw_terms.buckets
-                             if bucket["key"] == "acct:someone@example.com")
-        user_bucket_2 = next(bucket for bucket in response.aggregations.user_raw_terms.buckets
-                             if bucket["key"] == "acct:Someone@example.com")
+        user_bucket_1 = next(bucket for bucket in response.aggregations.user_raw_terms.buckets if
```
robertknight: These formatting changes appear to be unrelated to the other changes in this PR. Can you remove them from this PR?

tests/h/search/index_test.py
```diff
@@ -310,6 +316,80 @@ def test_annotation_is_marked_deleted(self, es_client, factories, index, search)
...
+            assert result["_id"] == id
+
+        for id in ids_not_to_index:
+            with pytest.raises(elasticsearch1.exceptions.NotFoundError):
+                result = es_client.conn.get(index=es_client.index,
+                                            doc_type="annotation",
+                                            id=id)
+
+    def test_it_logs_indexing_status(self, caplog, batch_indexer, factories):
+        annotations = factories.Annotation.create_batch(200)
```
robertknight: There is overhead to indexing annotations in Elasticsearch so, to make these tests run faster, use smaller values if possible.

h/search/index.py
```diff
@@ -107,28 +107,31 @@ def __init__(self, session, es_client, request, target_index=None, op_type='inde
...
             self._target_index = target_index
 
-    def index(self, annotation_ids=None):
+    def index(self, annotation_ids=None, windowsize=PG_WINDOW_SIZE, chunk_size=ES_CHUNK_SIZE):
         """
         Reindex annotations.
 
         :param annotation_ids: a list of ids to reindex, reindexes all when `None`.
         :type annotation_ids: collection
+        :param windowsize: how often to log indexing progress
```
robertknight: I think this docstring could be clearer about the meaning of `windowsize`. It means the number of annotations to index in between progress log statements?

tests/h/search/index_test.py
```diff
@@ -150,9 +154,9 @@ def test_you_can_make_aggregations_on_user_raw(self, factories, index, search):
         response = search.execute()
 
         user_bucket_1 = next(bucket for bucket in response.aggregations.user_raw_terms.buckets
-                             if bucket["key"] == "acct:someone@example.com")
+            if bucket["key"] == "acct:someone@example.com")
```
robertknight: The change in indentation here causes a lint error on my system, although not one caught by the `make lint` command. Can you change this line and the one below back to the way it was?

tests/h/search/index_test.py
```diff
@@ -179,8 +183,10 @@ def test_you_can_make_aggregations_on_tags_raw(self, es_client, factories, index
 
         response = search.execute()
 
-        tag_bucket_1 = next(bucket for bucket in response.aggregations.tags_raw_terms.buckets if bucket["key"] == "Hello")
-        tag_bucket_2 = next(bucket for bucket in response.aggregations.tags_raw_terms.buckets if bucket["key"] == "hello")
+        tag_bucket_1 = next(bucket for bucket in response.aggregations.tags_raw_terms.buckets if
```
robertknight: These indentation changes are not related to the rest of the PR. These lines were a bit long but I would prefer the change to be in a separate commit or PR.

tests/h/search/index_test.py
```diff
@@ -310,6 +316,75 @@ def test_annotation_is_marked_deleted(self, es_client, factories, index, search)
...
+                                   id=id)
+
+    def test_it_logs_indexing_status(self, caplog, batch_indexer, factories):
+        annotations = factories.Annotation.create_batch(10)
+        ids = [a.id for a in annotations]
+
+        with caplog.at_level(logging.INFO):
+            batch_indexer.index(ids, 3)
+
+        for record in caplog.records:
```
robertknight: This test still passes if you comment out the `log.info` statement in `h/search/index.py` because in that case `caplog.records` is empty.

tests/h/search/index_test.py
```diff
@@ -310,6 +312,80 @@ def test_annotation_is_marked_deleted(self, es_client, factories, index, search)
...
 
+class TestBatchIndexer(object):
+    def test_it_indexes_all_annotations(self, batch_indexer, es_client, factories):
+        annotations = factories.Annotation.create_batch(3)
+        ids = [a.id for a in annotations]
+
+        batch_indexer.index()
+
+        results = es_client.conn.search()['hits']['hits']
+        assert [ann.get("_id") for ann in results].sort() == ids.sort()
```
robertknight: This assert doesn't do what you expected because the `sort()` API mutates the array and then always returns `None`. Same gotcha that Hannah [encountered here](https://github.com/hypothesis/h/pull/5081#discussion_r196680399). You can use `sorted(list)` to get a sorted copy of a list.

tests/h/search/index_test.py
```diff
@@ -310,6 +312,80 @@ def test_annotation_is_marked_deleted(self, es_client, factories, index, search)
...
+    def test_it_indexes_specific_annotations(self, batch_indexer, es_client, factories):
+        annotations = factories.Annotation.create_batch(5)
+        ids = [a.id for a in annotations]
+        ids_to_index = ids[:3]
+        ids_not_to_index = ids[3:]
+
+        batch_indexer.index(ids_to_index)
+
+        results = es_client.conn.search()['hits']['hits']
+        assert [ann.get("_id") for ann in results].sort() == ids_to_index.sort()
```
robertknight: Same issue with `sort()` here as above.

hmstepanek:
Oopsy...good catch! I’ll fix this this afternoon. 🙈

tests/h/search/index_test.py
hmstepanek: Sorry @sheetaluk! My bad!

PR #5061, opened 2018-05-30
Add AuthorityFilter test

hmstepanek:
Add a tests to validate the AuthorityFilter works.

### DIFF
--- a/tests/h/search/query_test.py
+++ b/tests/h/search/query_test.py
@@ -25,11 +25,21 @@ class TestTopLevelAnnotationsFilter(object):
 
 
 class TestAuthorityFilter(object):
+    def test_it_filters_out_non_matching_authorities(self, Annotation, search):
+        annotations_auth1 = [Annotation(userid="acct:foo@auth1", shared=True).id,
+                             Annotation(userid="acct:bar@auth1", shared=True).id]
+        # Make some other annotations that are of different authority.
+        Annotation(userid="acct:bat@auth2", shared=True)
+        Annotation(userid="acct:bar@auth3", shared=True)
+
+        result = search.run({})
+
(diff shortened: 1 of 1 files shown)


tests/h/search/query_test.py
```diff
@@ -25,11 +25,21 @@ def search(self, pyramid_request):
...
+    def test_it_filters_out_non_matching_authorities(self, Annotation, search):
+        annotations_auth1 = [Annotation(userid="acct:foo@auth1", shared=True).id,
+                             Annotation(userid="acct:bar@auth1", shared=True).id]
+        # Make some other annotations that are of different authority.
+        Annotation(userid="acct:bat@auth2", shared=True)
+        Annotation(userid="acct:bar@auth3", shared=True)
+
+        result = search.run({})
+
+        assert set(result.annotation_ids) == set(annotations_auth1)
```
seanh: I appreciate the attention to detail here - using `set()` because this test shouldn't care about the order of the annotations keeps the test robust (not failing for reasons we don't care about) and non-flakey.

Technically the search could incorrectly return the same annotation twice and due to `set()` the test would still pass. _Sorting_ the list rather than using set would avoid that. This seems like an unlikely bug to happen though.

tests/h/search/query_test.py
```diff
@@ -25,11 +25,21 @@ def search(self, pyramid_request):
...
 
     @pytest.fixture
     def search(self, pyramid_request):
         search_ = search.Search(pyramid_request)
-        search_.append_filter(search.AuthorityFilter())
+        search_.append_filter(search.AuthorityFilter("auth1"))
```
seanh: Oops, good catch

PR #5062, opened 2018-05-31
Remove WebSocket server dependency on `h.search`

robertknight:
I cannot see any references to functionality from `h.search` in
`h.websocket`, `h.realtime` or `h.streamer`. I can imagine there would
have been a dependency back when Elasticsearch was our primary data
store but not any longer.

Removing this dependency may make it slightly easier to run the
WebSocket server in a separate container / Elastic Beanstalk application
in future, something we may want to do for scaling reasons.

----

Since our integrated test coverage of the WebSocket server is not great, I would suggest testing this manually as follows:

1. Run the dev server, verify that there are no errors from the WS process
2. Open up a couple of browser tabs to the same page with the dev client loaded and verify that basic WS functionality still works:
   - Annotating in one tab should result in a red "update available" notification in the other tab
   - Creating a group on the h web pages should auto-update the groups list in open tabs
   - Leaving a group via the h web pages should auto-update the groups list in open tabs

### DIFF
--- a/conf/development-websocket.ini
+++ b/conf/development-websocket.ini
@@ -1,7 +1,8 @@
 [app:main]
 use: call:h.websocket:create_app
 
-# Elasticsearch configuration
+# Elasticsearch configuration. The WebSocket does not actually use ES but
+# it does use the same config loading code which checks for these settings.
 es.host: http://localhost:9200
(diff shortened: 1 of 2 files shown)


sheetaluk:
LGTM.

PR #5063, opened 2018-05-31
Add tests for AuthFilter

hmstepanek:
Test that the AuthFilter filters out annotations that are visible to the user based on the pyramid authentication policy.

### DIFF
--- a/tests/h/search/query_test.py
+++ b/tests/h/search/query_test.py
@@ -44,6 +44,43 @@ class TestAuthorityFilter(object):
 
 
 class TestAuthFilter(object):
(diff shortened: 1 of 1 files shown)


seanh:
+1, these are excellent tests.

@hmstepanek did you see my comment elsewhere that it might be a teeny bit more correct to sort rather than use `set()` in these tests? (Though not a big deal)

hmstepanek:
I did! It felt weird to change the style half way through but I do like that idea of using sort instead. Maybe I can go back and do that? Maybe I’ll submit a little mini PR today on that before moving forward with anymore tests.

PR #5064, opened 2018-06-01
Mark `SECRET_KEY` as a required setting

robertknight:
This PR leverages the functionality added in #5033 to be able to mark settings as required in production environments to fix an issue I ran into setting up a new h deployment for load-testing last week.

The `SECRET_KEY` value is used to derive the key for signing cookie
sessions. The previous behavior of generating a random value if this is
not set is not useful in a production environment because each web
process generates a different value and therefore cookies signed by one
process fail to validate when submitted to a different web process. Not
setting a `SECRET_KEY` also means that existing cookies become invalid
after a server restart.

### DIFF
--- a/h/config.py
+++ b/h/config.py
@@ -56,7 +56,7 @@ def configure(environ=None, settings=None):
     settings_manager.set('statsd.prefix', 'STATSD_PREFIX')
 
     # Configuration for Pyramid
-    settings_manager.set('secret_key', 'SECRET_KEY', type_=bytes)
+    settings_manager.set('secret_key', 'SECRET_KEY', type_=bytes, required=True)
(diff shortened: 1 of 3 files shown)


seanh:
Nice!

h/config.py
```diff
@@ -56,7 +56,7 @@ def configure(environ=None, settings=None):
     settings_manager.set('statsd.prefix', 'STATSD_PREFIX')
 
     # Configuration for Pyramid
-    settings_manager.set('secret_key', 'SECRET_KEY', type_=bytes)
+    settings_manager.set('secret_key', 'SECRET_KEY', type_=bytes, required=True)
```
hmstepanek: Just a note: the secret_key is already defined in the dev ini files so nothing needs to be done there.

PR #5065, opened 2018-06-01
Mark Elasticsearch and Postgres URLs as required settings

robertknight:
_Depends on https://github.com/hypothesis/h/pull/5064_

The app can't function without these in production environments, so mark them as required settings.

One open question here is whether we should mark `es.url` as required straight away. The current code in `master` *does* require this setting, but you can't use the app with just ES 6 yet, you also have to have an ES 1 cluster. I think it would be friendlier to anyone running h in the meantime to make `es.url` optional until Hypothesis is fully functional with ES 6. Thoughts?

### DIFF
--- a/conf/app.ini
+++ b/conf/app.ini
@@ -10,9 +10,6 @@ pipeline:
 [app:h]
 use: call:h.app:create_app
 
-# SQLAlchemy configuration -- See SQLAlchemy documentation
-sqlalchemy.url: postgresql://postgres@localhost/postgres
-
 [filter:proxy-prefix]
 use: egg:PasteDeploy#prefix
 
--- a/conf/development-websocket.ini
+++ b/conf/development-websocket.ini
@@ -2,6 +2,7 @@
 use: call:h.websocket:create_app
 
 # Elasticsearch configuration
+es.host: http://localhost:9200
 es.url: http://localhost:9201
 
 # Use gevent-compatible transport for the Sentry client
--- a/h/config.py
+++ b/h/config.py
@@ -56,7 +56,7 @@ def configure(environ=None, settings=None):
     settings_manager.set('statsd.prefix', 'STATSD_PREFIX')
 
     # Configuration for Pyramid
-    settings_manager.set('secret_key', 'SECRET_KEY', type_=bytes)
+    settings_manager.set('secret_key', 'SECRET_KEY', type_=bytes, required=True)
     settings_manager.set('secret_salt', 'SECRET_SALT', type_=bytes, default=DEFAULT_SALT)
 
     # Configuration for h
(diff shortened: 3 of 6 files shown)


seanh:
> One open question here is whether we should mark es.url as required straight away. The current code in master does require this setting, but you can't use the app with just ES 6 yet, you also have to have an ES 1 cluster. I think it would be friendlier to anyone running h in the meantime to make es.url optional until Hypothesis is fully functional with ES 6. Thoughts?

If someone is running h currently they're gonna need an Es 6 u[ and running very soon, so I think we should just keep it simple and make it required now. I think they're also gonna need to run Es 1 and Es 6 simultaneously for a time and to migrate their data from Es 1 to 6.

h/config.py
```diff
@@ -41,22 +41,22 @@ def configure(environ=None, settings=None):
...
+    settings_manager.set('es.url', 'ELASTICSEARCH_URL', required=True),
     settings_manager.set('es.index', 'ELASTICSEARCH_INDEX')
     settings_manager.set('es.aws.access_key_id', 'ELASTICSEARCH_AWS_ACCESS_KEY_ID')
     settings_manager.set('es.aws.region', 'ELASTICSEARCH_AWS_REGION')
     settings_manager.set('es.aws.secret_access_key', 'ELASTICSEARCH_AWS_SECRET_ACCESS_KEY')
     settings_manager.set('mail.default_sender', 'MAIL_DEFAULT_SENDER')
     settings_manager.set('mail.host', 'MAIL_HOST')
     settings_manager.set('mail.port', 'MAIL_PORT', type_=int)
-    settings_manager.set('sqlalchemy.url', 'DATABASE_URL', type_=database_url)
+    settings_manager.set('sqlalchemy.url', 'DATABASE_URL', type_=database_url, required=True)
```
hmstepanek: The sqlalchemy.url is defined in app.ini, websocket.ini, and alembic.ini. Either those need to be removed or this setting really isn't required.

tests/functional/conftest.py
```diff
@@ -21,6 +21,7 @@
     'h.app_url': 'http://example.com',
     'h.authority': 'example.com',
     'pyramid.debug_all': True,
+    'secret_key': 'notasecret',
```
hmstepanek: Is it bad to expose this here?

hmstepanek:
I think it's reasonable to not make es.url required until we fully make the switch to es 6. @robertknight do you wanna remove the sqlachemy.url from the prod ini files? I think this will be ready to merge after that. Might help us eliminate some of our open PRs.

conf/development-websocket.ini
```diff
@@ -2,6 +2,7 @@
 use: call:h.websocket:create_app
 
 # Elasticsearch configuration
+es.host: http://localhost:9200
```
robertknight: I don't think the WebSocket actually needs ES at all, but I've left removing that dependency as an exercise for a separate PR. Until then this setting is needed in a dev environment.

conf/app.ini
```diff
@@ -10,9 +10,6 @@ pipeline:
 [app:h]
 use: call:h.app:create_app
 
-# SQLAlchemy configuration -- See SQLAlchemy documentation
-sqlalchemy.url: postgresql://postgres@localhost/postgres
```
robertknight: As per Hannah's suggestion, remove this default so that `DATABASE_URL` must be set via an env var in production environments.

robertknight:
OK, I've removed the defaults for `sqlalchemy.url` from the prod config files. I also added an `es.host` default for the websocket dev config file, which is needed until https://github.com/hypothesis/h/pull/5062 is merged.

PR #5066, opened 2018-06-01
Don't apply default search filters, aggregators or matchers in search tests

hmstepanek:
Make the default search filters, aggregators, and matchers clearable so that we can remove them when unit testing so as to only test one filter at a time.

### DIFF
--- a/h/search/core.py
+++ b/h/search/core.py
@@ -58,6 +58,11 @@ class Search(object):
 
         return SearchResult(total, annotation_ids, reply_ids, aggregations)
 
+    def clear(self):
+        """Clear search filters, aggregators, and matchers."""
+        self.builder = query.Builder()
+        self.reply_builder = query.Builder()
+
     def append_filter(self, filter_):
         """Append a search filter to the annotation and reply query."""
         self.builder.append_filter(filter_)
--- a/tests/h/search/query_test.py
+++ b/tests/h/search/query_test.py
@@ -3,64 +3,63 @@ from __future__ import unicode_literals
 
 import pytest
 
-from h import search
+from h.search import Search, query
 
 
 class TestTopLevelAnnotationsFilter(object):
 
-    def test_it_filters_out_replies_but_leaves_annotations_in(self, Annotation, search):
+    def test_it_filters_out_replies_but_leaves_annotations_in(self, Annotation, filtered_search):
         annotation = Annotation(shared=True)
         reply = Annotation(references=[annotation.id], shared=True)
 
-        result = search.run({})
+        result = filtered_search.run({})
 
         assert annotation.id in result.annotation_ids
         assert reply.id not in result.annotation_ids
 
     @pytest.fixture
-    def search(self, pyramid_request):
-        search_ = search.Search(pyramid_request)
-        search_.append_filter(search.TopLevelAnnotationsFilter())
-        return search_
+    def filtered_search(self, search):
+        search.append_filter(query.TopLevelAnnotationsFilter())
+        return search
 
 
 class TestAuthorityFilter(object):
-    def test_it_filters_out_non_matching_authorities(self, Annotation, search):
+    def test_it_filters_out_non_matching_authorities(self, Annotation, filtered_search):
         annotations_auth1 = [Annotation(userid="acct:foo@auth1", shared=True).id,
                              Annotation(userid="acct:bar@auth1", shared=True).id]
         # Make some other annotations that are of different authority.
         Annotation(userid="acct:bat@auth2", shared=True)
         Annotation(userid="acct:bar@auth3", shared=True)
 
-        result = search.run({})
+        result = filtered_search.run({})
 
         assert set(result.annotation_ids) == set(annotations_auth1)
 
     @pytest.fixture
-    def search(self, pyramid_request):
-        search_ = search.Search(pyramid_request)
-        search_.append_filter(search.AuthorityFilter("auth1"))
-        return search_
+    def filtered_search(self, search):
+        search.append_filter(query.AuthorityFilter("auth1"))
+        return search
 
 
 class TestAuthFilter(object):
-    def test_logged_out_user_can_not_see_private_annotations(self, search, Annotation):
+    def test_logged_out_user_can_not_see_private_annotations(self, filtered_search, Annotation):
         Annotation()
         Annotation()
 
-        result = search.run({})
+        result = filtered_search.run({})
 
         assert not result.annotation_ids
 
-    def test_logged_out_user_can_see_shared_annotations(self, search, Annotation):
+    def test_logged_out_user_can_see_shared_annotations(self, filtered_search, Annotation):
         shared_ids = [Annotation(shared=True).id,
                       Annotation(shared=True).id]
 
-        result = search.run({})
+        result = filtered_search.run({})
 
         assert set(result.annotation_ids) == set(shared_ids)
 
-    def test_logged_in_user_can_only_see_their_private_annotations(self, search, pyramid_config, Annotation):
+    def test_logged_in_user_can_only_see_their_private_annotations(self,
+            filtered_search, pyramid_config, Annotation):
         userid = "acct:bar@auth2"
         pyramid_config.testing_securitypolicy(userid)
         # Make a private annotation from a different user.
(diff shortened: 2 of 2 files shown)


h/search/core.py
```diff
@@ -16,6 +16,19 @@
...
     'aggregations'])
 
+DEFAULT_FILTERS = [
+    query.DeletedFilter,
+    query.GroupFilter,
+    query.UserFilter,
+    query.AnyMatcher,
+    query.TagsMatcher]
+
+DEFAULT_REQUEST_FILTERS = [
```
hmstepanek: There are currently two types of filters, ones that take a request obj as an arg and ones that don't hence the two different lists here.

h/search/core.py
```diff
@@ -145,15 +178,10 @@ def _instrument(self):
             s.send()
 
     @staticmethod
-    def _default_querybuilder(request):
+    def _default_querybuilder(request, filters, request_filters):
```
hmstepanek: We could split this into two separate methods; one that adds reg filters and one that adds request filters. Not sure if that's necessary or not??

tests/h/search/query_test.py
```diff
@@ -3,188 +3,182 @@
...
 
 class TestAuthorityFilter(object):
-    def test_it_filters_out_non_matching_authorities(self, Annotation, search):
-        annotations_auth1 = [Annotation(userid="acct:foo@auth1", shared=True).id,
-                             Annotation(userid="acct:bar@auth1", shared=True).id]
+    def test_it_filters_out_non_matching_authorities(self, Annotation, filtered_search):
+        annotations_auth1 = [Annotation(userid="acct:foo@auth1").id,
+                             Annotation(userid="acct:bar@auth1").id]
         # Make some other annotations that are of different authority.
-        Annotation(userid="acct:bat@auth2", shared=True)
```
hmstepanek: Since we aren't testing with the shared filter anymore we don't need to mark all these as shared annotations.

h/search/core.py
robertknight: Having two separate lists seems a perfectly pragmatic approach to me. I would suggest adding a comment above each variable declaration to just clarify the difference.

h/search/core.py
robertknight: This looks fine to me.

tests/h/search/query_test.py
robertknight: 👍

tests/h/search/query_test.py
```diff
@@ -3,188 +3,182 @@
...
 
 class TestGroupFilter(object):
 
     @pytest.fixture
-    def search(self, pyramid_request):
-        # We don't need to append GroupFilter to Search because it's one of the filters that
-        # Search appends by default.
-        return search.Search(pyramid_request)
+    def filtered_search(self, search):
+        search.append_aggregation(query.GroupFilter())
```
robertknight: Why is this `append_aggregation` and not `append_filter`?

hmstepanek: Me too...lol I thought I removed all of those but I guess I didn't. I'll fix that.

hmstepanek:
Ok changes addressed; added comments above default filter lists and now using append_filter everywhere.

hmstepanek:
Changes addressed: using append_matcher/append_filter and added comments above default filter lists.

tests/h/search/query_test.py
```diff
@@ -3,188 +3,182 @@
...
 
 class TestTagsAggregation(object):
 
     @pytest.fixture
-    def search(self, pyramid_request):
-        search_ = search.Search(pyramid_request)
-        search_.append_aggregation(search.TagsAggregation())
-        return search_
+    def filtered_search(self, search):
+        search.append_filter(query.TagsAggregation())
```
robertknight: This should be `append_aggregation` surely? Same with `TestUsersAggregation` below.

hmstepanek:
Yeah sorry...just noticed another issue too. Durp.

seanh:
Can we remove <https://github.com/hypothesis/h/pull/5066/commits/9e825415f356a8700154489a35bf422414e4efbe> (_Remove shared=True from search filter test cases_)? I don't think we should do that - I think the tests should be working with shared annotations, because that's the kind of data that we mostly have in production. Private annotations are a corner case. I wouldn't want all of our search tests to test only with private annotations.

seanh:
Hannah talked to me about this last week and suggested an alternative approach: add a `clear()` method to `Search` that clears all the filters and matchers. That sounds much simpler and less likely to break things (it merely involves _adding_ a tests-only method to `Search`, not refactoring `Search` code that the production code is using), and is also much more elegant than what I was going to suggest. @hmstepanek I wonder if it'd be worth closing this PR, and replacing it with a clean PR for `clear()`?

hmstepanek:
Meh, I’ll just rebase it and change the name. Nbd.

seanh:
Thanks

robertknight:
> Hannah talked to me about this last week and suggested an alternative approach: add a clear() method to Search that clears all the filters and matchers. 

I would probably pick a different name as `clear` is a bit ambiguous (clear what?). My personal stylistic preference is for the approach taken here, which is to initialize objects to the desired state instead of manipulating them after-the-fact. In more complex scenarios, that can eliminate certain kinds of bug (eg. failing to invalidate a cache after modifying object state). Frankly though I think the debate between the two approaches is somewhat philosophical in this context. Either would be fine with me.

seanh:
Remember, this feature is _only needed by the tests_. I think Hannah's suggestion of adding a `clear()` method is going to produce a much smaller and simpler diff - we're talking about adding one roughly two-line method to the class rather than an approximately 50-line refactoring diff, and the method is only called by the tests so it's much less likely to break anything. Given that the current code does in fact break the search code, and activity pages are crashing, I don't think this is philosophical at all.

If either is fine with you, I have a strong preference for seeing how Hannah's clear method idea works out

h/search/core.py
```diff
@@ -58,6 +58,11 @@ def run(self, params):
 
         return SearchResult(total, annotation_ids, reply_ids, aggregations)
 
+    def clear(self):
```
hmstepanek: I'm open to alternative name suggestions-I would just say clear_filters but technically it clears filters, matchers, and aggregators but to say all that would result in a very long method name. Maybe clear_query?

tests/h/search/query_test.py
```diff
@@ -3,188 +3,182 @@
...
 import pytest
 
-from h import search
+from h.search import Search, query
 
 
 class TestTopLevelAnnotationsFilter(object):
 
-    def test_it_filters_out_replies_but_leaves_annotations_in(self, Annotation, search):
-        annotation = Annotation(shared=True)
```
hmstepanek: I know @seanh mentioned something about keeping the shared=True here. For me it's a debate between writing tests that make sense vs writing true unit tests. I would prefer to keep the shared=True out of these unit tests and move them to the integration tests (aka tests that test with multiple filters/aggregators/matchers). Alternatively, we could keep the shared=True here but then I think we should append the shared filter.

hmstepanek:
I think we can ignore code cov here as it's complaining about the fixtures not being called/covered in the tests module for the unimplemented tests.

tests/h/search/query_test.py
hmstepanek: We discussed this and decided to just leave it as is without the shared=True.

tests/h/search/query_test.py
```diff
@@ -3,188 +3,182 @@
...
+        reply = Annotation(references=[annotation.id])
 
-        result = search.run({})
+        result = filtered_search.run({})
 
         assert annotation.id in result.annotation_ids
         assert reply.id not in result.annotation_ids
 
     @pytest.fixture
-    def search(self, pyramid_request):
```
robertknight: Any particular reason for changing the name of the fixture? Usually when we customize a fixture for a particular test case we just override it with one that has the same name.

h/search/core.py
seanh: I think `clear()` is a good name - `search.clear()` is clear enough

h/search/core.py
```diff
@@ -58,6 +58,11 @@ def run(self, params):
...
+        """Clear search filters, aggregators, and matchers."""
+        self.builder = query.Builder()
+        self.reply_builder = query.Builder()
```
seanh: This is nice - I had been thinking that, since the filters, aggregators and matchers are actually stored and used in `Builder` not in `Search`, you would have to add a `clear()` method to `Builder` and add a `clear()` method to `Search` that calls it. And then you'd be implicitly relying on the fact that `Builder` doesn't actually do anything with an appended filter, matcher or aggregator until `build()` is called so that it's safe to clear them. So just replacing the entire `Builder` objects is much better :+1:

h/search/core.py
seanh: Also the docstring clarifies it perfectly

seanh:
@hmstepanek Can I leave it to you to merge and deploy this? There's a bunch of other PRs queued up on Jenkins that need to be deployed, tested on QA (if they're not just tests PRs), then deployed to production. They all look like PRs that you'll be familiar with

tests/h/search/query_test.py
hmstepanek: No particular reason-I'll change it to just override the name I supposed.

PR #5067, opened 2018-06-05
Add tests to make sure deleted annotations don't get batch indexed.

sheetaluk:
PS: I think we should merge https://github.com/hypothesis/h/pull/5060 before merging this one.

### DIFF
--- a/tests/h/search/index_test.py
+++ b/tests/h/search/index_test.py
@@ -52,7 +52,8 @@ class TestIndex(object):
         # index directly and uses this ``document`` field.
         assert get(annotation.id)["document"]["web_uri"] == "https://example.com/example_article"
 
-    def test_it_can_index_an_annotation_with_a_document_with_no_web_uri(self, factories, index, get):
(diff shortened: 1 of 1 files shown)


tests/h/search/index_test.py
```diff
@@ -349,6 +349,23 @@ def test_it_indexes_specific_annotations(self, batch_indexer, es_client, factori
...
                                             id=id)
 
+    def test_it_does_not_index_deleted_annotations(self, batch_indexer, es_client, factories):
+        ann = factories.Annotation()
+        # create deleted annotations
+        ann_del = factories.Annotation(deleted=True)
+
+        batch_indexer.index()
+
+        result_indexed = es_client.conn.get(index=es_client.index,
```
robertknight: We are duplicating these lines to get an annotation from the index by ID in a few tests. I would be inclined, perhaps as a separate change once these BatchIndexer tests are merged, to re-use the `get` fixture from the indexing test.

PR #5068, opened 2018-06-05
Add more batch indexer tests to test that notifications are dispatched.

sheetaluk:
PS: I think we should merge #5060 before merging this one.

### DIFF
--- a/tests/h/search/index_test.py
+++ b/tests/h/search/index_test.py
@@ -367,7 +367,10 @@ class TestBatchIndexer(object):
         event = AnnotationTransformEvent.return_value
 
         for annotation in annotations:
-            AnnotationTransformEvent.assert_has_calls([mock.call(pyramid_request, annotation, {'test': 'val'})])
+            AnnotationTransformEvent.assert_has_calls([mock.call(
+                pyramid_request,
+                annotation,
+                AnnotationSearchIndexPresenter.return_value.asdict.return_value)])
             notify.assert_has_calls([mock.call(event)])
 
     def test_it_logs_indexing_status(self, caplog, batch_indexer, factories):
--- a/tests/h/search/index_test.py
+++ b/tests/h/search/index_test.py
@@ -290,10 +290,6 @@ class TestIndex(object):
                 id=annotation_id)["_source"]
         return _get
 
-    @pytest.fixture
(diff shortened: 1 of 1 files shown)


tests/h/search/index_test.py
```diff
@@ -310,6 +312,109 @@ def test_annotation_is_marked_deleted(self, es_client, factories, index, search)
...
+            AnnotationTransformEvent.assert_has_calls([mock.call(pyramid_request, annotation, mock.ANY)])
+            notify.assert_has_calls([mock.call(event)])
+
+    def test_it_logs_indexing_status(self, caplog, batch_indexer, factories):
+        annotations = factories.Annotation.create_batch(10)
+        ids = [a.id for a in annotations]
+
+        with caplog.at_level(logging.INFO):
+            batch_indexer.index(ids, 3)
+            for record in caplog.records:
```
hmstepanek: I think this should really be outside of the with statement.

tests/h/search/index_test.py
```diff
@@ -310,6 +312,109 @@ def test_annotation_is_marked_deleted(self, es_client, factories, index, search)
...
+
+        batch_indexer.index(ids_to_index)
+
+        for id in ids_to_index:
+            result = es_client.conn.get(index=es_client.index,
+                                        doc_type="annotation",
+                                        id=id)
+            assert result["_id"] == id
+
+        for id in ids_not_to_index:
```
hmstepanek: For these tests we could grab all the annotations instead of looping through each of them individually; it looks like if you leave the search body empty it will grab all of them:
``` 
       res = conn.search()['hits'['hits']
       found_ids = [ann._id for ann in res]
       assert found_ids.sort() == ids_to_index.sort()
```
I think this would make the tests a lot simpler as you don't have to check for non indexed and indexed annotations.

tests/h/search/index_test.py
```diff
@@ -419,3 +444,8 @@ def search(es_client):
 def batch_indexer(db_session, es_client, pyramid_request):
     return h.search.index.BatchIndexer(db_session, es_client,
                                        pyramid_request, es_client.index)
+
```
hmstepanek: This got moved here in order for it to be shared by both classes.

PR #5069, opened 2018-06-08
Add Black code reformatting

seanh:
Add `make check-formatting` that runs Black in `--check` mode which just exits with code 1 if there are any files that _would_ be reformatted, but doesn't actually reformat anything. Add `make check-formatting` to Travis to enforce formatting.

Also add `make reformat` that actually reformats all the code with Black.

And reformat all the code by running `make reformat`.

Some flake8 config tweaks are needed to make `make lint` pass after `make reformat`.


seanh:
@lyzadanger Does it work if you checkout this branch and run `make reformat`? It should create a new Python 3 virtualenv in `.tox` and run `black` in it. I'd like to know whether it works on your machine or not, to decide whether it's worth docker-izing this (and everything else)

PR #5070, opened 2018-06-08
Add motivations to annotation table, model

hmstepanek:
Motivations are a field on the Annotation model and describe the "motivation" for creating an annotation. There may be 0 or more motivations according to the w3 annotation spec. Currently there are a list of accepted motivation types based on those provided in the spec. These include:
assessing, bookmarking, classifying, commenting, describing, editing, highlighting, identifying, linking, moderating, questioning, replying, tagging.

See https://www.w3.org/TR/annotation-model/#motivation-and-purpose.

### DIFF
--- a/h/models/annotation.py
+++ b/h/models/annotation.py
@@ -3,32 +3,33 @@
 from __future__ import unicode_literals
 
 import datetime
-import enum
 
 import sqlalchemy as sa
 from sqlalchemy.dialects import postgresql as pg
 from sqlalchemy.ext.hybrid import hybrid_property
-from sqlalchemy.ext.mutable import MutableDict, MutableList, MutableSet
+from sqlalchemy.ext.mutable import MutableDict, MutableList
 
 from h.db import Base, types
 from h.util import markdown, uri
 from h.util.user import split_user
+from h import i18n
+_ = i18n.TranslationString
 
 
-class Motivation(enum.Enum):
-    assessing = "assessing"
-    bookmarking = "bookmarking"
-    classifying = "classifying"
-    commenting = "commenting"
-    describing = "describing"
-    editing = "editing"
-    highlighting = "highlighting"
-    identifying = "identifying"
-    linking = "linking"
-    moderating = "moderating"
-    questioning = "questioning"
-    replying = "replying"
-    tagging = "tagging"
+MOTIVATIONS = ["assessing",
+               "bookmarking",
+               "classifying",
+               "commenting",
+               "describing",
+               "editing",
+               "highlighting",
+               "identifying",
+               "linking",
+               "moderating",
+               "questioning",
+               "replying",
+               "tagging",
+               ]
 
 
 class Annotation(Base):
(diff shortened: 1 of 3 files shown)


tests/h/models/annotation_test.py
```diff
@@ -89,6 +96,17 @@ def test_authority_when_annotation_has_no_userid():
...
 
+def test_default_motivations_is_empty_set(factories):
+    annotation = factories.Annotation()
+
+    assert annotation.motivations == set([])
+
+
+def test_raises_when_invalid_motivation(db_session, factories):
+    with pytest.raises(sqlalchemy.exc.ProgrammingError):
+        factories.Annotation(motivations=set(["invalid"]))
```
seanh: To make sure the exception was raised for the reason you think it was, and not the same exception class being used for another reason, you can pass the `message` arg to `raises()`. See <https://docs.pytest.org/en/latest/assert.html?highlight=raises#assertions-about-expected-exceptions> (`message` is a regex)

tests/h/models/annotation_test.py
```diff
@@ -89,6 +101,38 @@ def test_authority_when_annotation_has_no_userid():
     assert Annotation().authority is None
 
 
+def test_default_motivations_is_empty(factories):
+    annotation = factories.Annotation()
+
+    assert annotation.motivations == []
+
+
+def test_allows_empty_motivations_list(db_session, factories):
```
hmstepanek: This test needs to not pass a default list.

h/models/annotation.py
```diff
@@ -122,6 +144,16 @@ class Annotation(Base):
                                  viewonly=True,
                                  uselist=True)
 
+    @sa.orm.validates('motivations')
+    def validate_motivations(self, key, motivations):
+        # We can't use enum in an array field because SQLAlchemy doesn't
+        # support it, so we have to do validation here
+        if len(set(motivations)) != len(motivations):
+            raise ValueError(_('motivations must not contain duplicates'))
+        if len(set(motivations) - set(MOTIVATIONS)) != 0:
```
seanh: I wonder if `MOTIVATIONS` should just be defined as a set literal (`MOTIVATIONS = {"assessing", "bookmarking", ...}`) in the first place? Then you wouldn't need to convert the list to a set here. On the other hand `MOTIVATIONS` would no longer be ordered, but it doesn't look like that matters.

seanh: I wonder if `MOTIVATIONS` should be defined within `validate_motivations()`? Since that's the only place where it's used

h/models/annotation.py
```diff
@@ -114,6 +132,10 @@ class Annotation(Base):
                             sa.ForeignKey('document.id'),
                             nullable=False)
 
+    motivations = sa.Column(MutableList.as_mutable(
+        pg.ARRAY(sa.UnicodeText, zero_indexes=True)
+     ), default=list)
```
seanh: It seems to produce the same result in the DB, but I'm not sure why the `motivations` column in the model class looks different than the one in the migration? The model class has `MutableList.as_mutable` but the migration doesn't.

seanh: Should there be a `server_default` here as well as `default`?

seanh:
Closing this to be replaced by two separate PRs

PR #5073, opened 2018-06-14
Index and search annotations in Elasticsearch 6

robertknight:
This is a rough branch which gets indexing into both ES 1 and ES 6 working, behind a feature flag (`index_es6`), and gets search working with ES 6, behind a separate feature flag (`search_es6`).

I'm not intending to merge this as-is but rather use it to review the basic approach, enable other devs to do some testing, and go over some of the specific changes required. If we're happy then logical chunks can be extracted into smaller PRs and shipped separately.

See also [my notes](https://gist.github.com/robertknight/32e12ddfa38d417b3e8cfc7a9c150705) on the investigation.

### DIFF
--- a/h/cli/commands/init.py
+++ b/h/cli/commands/init.py
@@ -19,6 +19,7 @@ def init(ctx):
 
     _init_db(request.registry.settings)
     _init_search(request.registry.settings)
+    _init_search_es6(request.registry.settings)
 
 
 def _init_db(settings):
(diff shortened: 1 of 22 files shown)


robertknight:
I'm going to close this for now and extract relevant parts into new PRs, following our discussion last Friday about how we are going to approach the remaining parts of the Elasticsearch migration.

PR #5074, opened 2018-06-14
Add tests to make sure bulk indexing presents annotations correctly.

sheetaluk:


tests/h/search/index_test.py
```diff
@@ -361,6 +378,27 @@ def test_it_logs_indexing_status(self, caplog, batch_indexer, factories):
                 assert 'indexed 0k annotations, rate=' in record.msg
         assert num_index_records == num_annotations // window_size
 
+    def test_it_correctly_presents_bulk_actions(self, batch_indexer, es_client, factories):
```
hmstepanek: I'm confused about what this test is trying to test for. I believe create_batch is an internal factory related method that isn't used outside of testing AFAIK that just creates multiples of the same annotation so I don't think there's really a need to test that. Maybe we are testing that the annotation is indexed with the same data as it had when it was created (a reasonable test) but we are only verifying the group and id match here. I feel like I'm missing something obvious but I'm not seeing it...

tests/h/search/index_test.py
```diff
@@ -382,6 +378,27 @@ def test_it_logs_indexing_status(self, caplog, batch_indexer, factories):
...
+        anns_a = factories.Annotation.create_batch(3, groupid="group_a")
+        anns_b = factories.Annotation.create_batch(3, groupid="group_b")
+
+        ids_a = [a.id for a in anns_a]
+        ids_b = [a.id for a in anns_b]
+
+        batch_indexer.index()
+
+        for id in ids_a:
```
hmstepanek: The use of the builtin id will be fixed in a follow up PR as there are other tests that use this.

tests/h/search/index_test.py
hmstepanek: Ok so @sheetaluk and I talked about this: the intention here is to make sure the fields get added to the indexed annotation as expected so there will be some updates to reflect that. We chose ~4/5 fields to look at to verify this happens correctly. It's not super necessary to check all of them as the presenter does that but it's nice to check a couple just to be sure it's behaving as we expect.

hmstepanek:
This looks fine. Since we really care about whether is has all the fields and not whether a bunch of annotations get indexed do we really need to index and test that many annotations? Could we for example just create 2 from 1 group and 1 from a different group and reduce the time by about half?

PR #5078, opened 2018-06-18
Fix `make run-docker` command

robertknight:
The `make run-docker` command has been broken since the `SECRET_KEY` and `ELASTICSEARCH_URL` env vars became mandatory.

Specifying a `SECRET_KEY` also fixes the existing issue where cookie sessions could not persist between different runs of the container.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -39,15 +39,19 @@ docker:
 run-docker:
 	$(eval RABBITMQ_CONTAINER ?= rabbitmq)
 	$(eval PG_CONTAINER ?= postgres)
+	$(eval ES1_CONTAINER ?= elasticsearch1)
 	$(eval ES_CONTAINER ?= elasticsearch)
 	docker run \
 		--link $(RABBITMQ_CONTAINER) \
 		--link $(PG_CONTAINER) \
 		--link $(ES_CONTAINER) \
+		--link $(ES1_CONTAINER) \
 		-e "APP_URL=http://localhost:5000" \
 		-e "BROKER_URL=amqp://guest:guest@$(RABBITMQ_CONTAINER):5672//" \
 		-e "DATABASE_URL=postgresql://postgres@$(PG_CONTAINER)/postgres" \
-		-e "ELASTICSEARCH_HOST=http://$(ES_CONTAINER):9200" \
+		-e "ELASTICSEARCH_HOST=http://$(ES1_CONTAINER):9200" \
+		-e "ELASTICSEARCH_URL=http://$(ES_CONTAINER):9201" \
+		-e "SECRET_KEY=notasecret" \
 		-p 5000:5000 \
 		hypothesis/hypothesis:$(DOCKER_TAG)
 
--- a/Makefile
+++ b/Makefile
@@ -37,15 +37,16 @@ docker:
 # Run docker image built with `docker` task
 .PHONY: run-docker
 run-docker:
-	$(eval RABBITMQ_CONTAINER ?= rabbitmq)
-	$(eval PG_CONTAINER ?= postgres)
-	$(eval ES1_CONTAINER ?= elasticsearch1)
-	$(eval ES_CONTAINER ?= elasticsearch)
+	$(eval RABBITMQ_CONTAINER ?= h_rabbit_1)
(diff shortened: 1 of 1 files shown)


hmstepanek:
Please hold off merging till I’ve had a chance to test.

robertknight:
Hi @hmstepanek. This command was originally written based on the assumption that services had been setup using the manual instructions we used to provide. Since we now recommend docker-compose in the setup docs I agree it makes sense that `run-docker` should assume that too.

I've taken your changes and added a couple of further refinements based on info from https://docs.docker.com/network/bridge/ . Please take a look at the last commit.

> https://docs.docker.com/network/links/ so looks like this may be a deprecated usage.

Indeed! Since we're using user-defined networks it looks like we can now avoid them entirely.

Makefile
```diff
@@ -34,20 +34,25 @@ dev: build/manifest.json .pydeps
...
-		--link $(RABBITMQ_CONTAINER) \
-		--link $(PG_CONTAINER) \
-		--link $(ES_CONTAINER) \
+		--net h_default \
 		-e "APP_URL=http://localhost:5000" \
+		-e "AUTHORITY=localhost" \
 		-e "BROKER_URL=amqp://guest:guest@$(RABBITMQ_CONTAINER):5672//" \
 		-e "DATABASE_URL=postgresql://postgres@$(PG_CONTAINER)/postgres" \
-		-e "ELASTICSEARCH_HOST=http://$(ES_CONTAINER):9200" \
+		-e "ELASTICSEARCH_HOST=http://$(ES1_CONTAINER):9200" \
```
hmstepanek: I believe this is the same docker container as qa/prod in which case it is not using the .ini file with all the dev env vars defined in it so we have to specify the required env vars here before running it. We wouldn't want to put them in the docker-compose because they aren't required for the docker compose when running in make dev. Ideally it would just use the dev .ini file but that may require some changing of the docker container command/config-it's potentially worth taking a quick peak but I'm not going to let it hold up this PR. Please correct me if that doesn't sound correct.

Makefile
```diff
@@ -34,20 +34,25 @@ dev: build/manifest.json .pydeps
...
 
-# Run docker image built with `docker` task
+# Run docker image built with `docker` task.
+# This command is intended only for use in development and assumes that the
+# service containers (Postgres, Elasticsearch etc.) have been run using
+# docker-compose.
 .PHONY: run-docker
 run-docker:
-	$(eval RABBITMQ_CONTAINER ?= rabbitmq)
+	$(eval RABBITMQ_CONTAINER ?= rabbit)
```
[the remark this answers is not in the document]
hmstepanek: Also I don't think there's a need to change the name from rabbitmq->rabbit. I think you can just leave it as is.

Makefile
hmstepanek: Are these url paths actually supposed to work locally cause if so I think these should be set to localhost:9200 etc?

robertknight:
> Then in the docker-compose.yml file, shouldn't we set a container_name value for each service ?

That's not necessary because `docker-compose` creates a network alias for each service using its name from the docker-compose.yml file. So if you open a shell in eg. the elasticsearch container, you do `ping postgres`.

Makefile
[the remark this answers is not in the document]
robertknight: Hmm ... I suppose `_HOST` would be more accurate than `_CONTAINER` now. I still think it is useful to be able to customize in case you want to test the Docker container with a different service container for some reason, although perhaps in that case one might just make a temporary edit to the `Makefile`.

Makefile
robertknight: The intention is that this command runs h in a "somewhat production-like" environment - debug facilities off, key config settings provided via env vars, running in a Docker container built using `make docker`. That's why it doesn't use the dev ini file.

The primary purpose is to make it easy and quick to locally test changes to the Docker build.

robertknight:
Hello @hmstepanek, I removed the ability to customize the service hostnames as you suggested since it isn't that useful if we are assuming the use of docker-compose to run them anyway. I also added a comment to clarify what the command is for in the first place. PTAL.

PR #5081, opened 2018-06-20
Use sorted list to test instead of set

hmstepanek:
Using a sorted list has the advantage of ensuring duplicates are not
in the list of annotation id results returned by search.

### DIFF
--- a/tests/h/search/query_test.py
+++ b/tests/h/search/query_test.py
@@ -11,12 +11,11 @@ class TestTopLevelAnnotationsFilter(object):
 
     def test_it_filters_out_replies_but_leaves_annotations_in(self, Annotation, search):
         annotation = Annotation()
-        reply = Annotation(references=[annotation.id])
+        Annotation(references=[annotation.id])
 
         result = search.run({})
 
-        assert annotation.id in result.annotation_ids
-        assert reply.id not in result.annotation_ids
+        assert [annotation.id] == result.annotation_ids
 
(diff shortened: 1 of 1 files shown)


tests/h/search/query_test.py
```diff
@@ -11,12 +11,11 @@ class TestTopLevelAnnotationsFilter(object):
 
     def test_it_filters_out_replies_but_leaves_annotations_in(self, Annotation, search):
         annotation = Annotation()
-        reply = Annotation(references=[annotation.id])
+        Annotation(references=[annotation.id])
 
         result = search.run({})
 
-        assert annotation.id in result.annotation_ids
```
hmstepanek: I got rid of the explicit check to make sure the reply was not in the results as using a complete list to check against does the same thing in one line.

tests/h/search/query_test.py
```diff
@@ -11,12 +11,11 @@ class TestTopLevelAnnotationsFilter(object):
...
-        assert reply.id not in result.annotation_ids
```
hmstepanek: I got rid of the explicit check to make sure the reply was not in the results as using a complete list to check against does the same thing in one line.

hmstepanek:
@robertknight To answer your question about was there a discussion about this; yes. This was requested by @seanh on a couple of my first test PRs.

robertknight:
Thanks!

PR #5086, opened 2018-06-20
Add tests for AnyMatcher

hmstepanek:
AnyMatcher is expected to match any of the following fields:
quote, text, uri.parts, and tags.

### DIFF
--- a/tests/h/search/query_test.py
+++ b/tests/h/search/query_test.py
@@ -165,6 +165,63 @@ class TestNipsaFilter(object):
 
 
 class TestAnyMatcher(object):
+    def test_matches_uriparts(self, search, Annotation):
+        Annotation(target_uri="http://bar.com")
+        matched_ids = [Annotation(target_uri="http://foo.com").id,
+                       Annotation(target_uri="http://foo.com/bar").id]
+
+        result = search.run({"any": "foo"})
+
+        assert sorted(result.annotation_ids) == sorted(matched_ids)
+
+    def test_matches_quote(self, search, Annotation):
+        Annotation(target_selectors=[{'exact': 'selected bar text'}])
+        matched_ids = [Annotation(target_selectors=[{'exact': 'selected foo text'}]).id,
+                       Annotation(target_selectors=[{'exact': 'selected foo bar text'}]).id]
+
+        result = search.run({"any": "foo"})
+
+        assert sorted(result.annotation_ids) == sorted(matched_ids)
+
+    def test_matches_text(self, search, Annotation):
+        Annotation(text="bar is best")
+        matched_ids = [Annotation(text="foo is fun").id,
+                       Annotation(text="foo is bar's friend").id]
+
+        result = search.run({"any": "foo"})
+
+        assert sorted(result.annotation_ids) == sorted(matched_ids)
+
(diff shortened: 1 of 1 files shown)


tests/h/search/query_test.py
```diff
@@ -165,6 +165,81 @@ def search(self, search, pyramid_request):
 
 
 class TestAnyMatcher(object):
+    def test_matches_uriparts(self, search, Annotation):
+        Annotation(target_uri="bar.com")
```
robertknight: As with the URI filters tests, I suggest tests should use valid absolute URIs for `target_uri`, unless it is specifically a test for what happens with invalid URIs, which I think will be covered by the `UriFilter` tests.

tests/h/search/query_test.py
```diff
@@ -165,6 +165,81 @@ def search(self, search, pyramid_request):
...
+    def test_matches_tags(self, search, Annotation):
+        Annotation(tags=["bar"])
+        matched_ids = [Annotation(tags=["foo"]).id,
+                       Annotation(tags=["foo", "bar"]).id]
+
+        result = search.run({"any": "foo"})
+
+        assert sorted(result.annotation_ids) == sorted(matched_ids)
+
+    def test_matches_all_fields(self, search, Annotation):
```
robertknight: This test is already covered by the above tests for individual fields, is it not?

tests/h/search/query_test.py
```diff
@@ -165,6 +165,81 @@ def search(self, search, pyramid_request):
...
+        """
+        Any is expected to match any of the following fields;
+        quote, text, uri.parts, and tags
+        that contain any of the passed keywords.
+        """
+        Annotation(target_selectors=[{'exact': 'selected baz text'}])
+        Annotation(tags=["baz"])
+        Annotation(target_uri="baz.com")
+        Annotation(text="baz is best")
+        matched_ids = [Annotation(target_uri="foo/bar/baz.com").id,
```
robertknight: These matches all contain the term "foo" in one of the matched fields. In order to test the behaviour we expect to see here we'd need to have at least some annotations containing either "foo" OR "bar", but not both.

tests/h/search/query_test.py
hmstepanek: I think here I was going for, lets make sure they match any of the fields at the same time but I think this is also covered by the or test below so I'll just remove it. I agree it's a bit repetitive.

hmstepanek:
Updated again.

hmstepanek:
I suspect we aren't always cleaning up the index. That's one of the changes I made I think in that yield PR I was working was to always use a fixture that had a teardown. (I could be wrong about that but I think I remember seeing the get_connection didn't have a teardown and instead there was a wrapper that had a teardown so you could use the get_connection fixture without a teardown).

robertknight:
> I could be wrong about that but I think I remember seeing the get_connection didn't have a teardown

Hmm ... there isn't a fixture called `get_connection`. Did you mean something else? The `es_client` fixture has a dependency on `delete_all_elasticsearch_documents` which will clean up the index before it runs.

hmstepanek:
Yes es_client.

hmstepanek:
Yes es_client.

PR #5089, opened 2018-06-21
Add UriFilter tests

hmstepanek:
The UriFilter filters annotations based on the url and uri keys.

### DIFF
--- a/tests/h/search/query_test.py
+++ b/tests/h/search/query_test.py
@@ -141,6 +141,74 @@ class TestUserFilter(object):
 
 
 class TestUriFilter(object):
+    def test_filters_by_uri(self, search, Annotation):
+        Annotation(target_uri="http://bar.com")
+        Annotation(target_uri="https://bar.com")
+        Annotation(target_uri="https://foo/bar.com")
+        expected_ids = [Annotation(target_uri="bar.com").id]
+
+        result = search.run({"uri": "bar.com"})
+
+        assert sorted(result.annotation_ids) == sorted(expected_ids)
+
+    def test_filters_by_url(self, search, Annotation):
+        Annotation(target_uri="http://bar.com")
+        Annotation(target_uri="https://bar.com")
+        Annotation(target_uri="https://foo/bar.com")
+        expected_ids = [Annotation(target_uri="bar.com").id]
+
+        result = search.run({"url": "bar.com"})
+
+        assert sorted(result.annotation_ids) == sorted(expected_ids)
+
+    def test_filters_on_whole_url(self, search, Annotation):
+        Annotation(target_uri="http://bar.com")
+        Annotation(target_uri="foo/bar.com")
+        Annotation(target_uri="http://foo.com")
+        expected_ids = [Annotation(target_uri="http://foo/bar.com").id]
+
+        result = search.run({"url": "http://foo/bar.com"})
+
+        assert sorted(result.annotation_ids) == sorted(expected_ids)
+
+    def test_filters_aliases_http_and_https(self, search, Annotation):
+        Annotation(target_uri="bar.com")
+        Annotation(target_uri="www.bar.com")
+        expected_ids = [Annotation(target_uri="http://bar.com").id,
+                        Annotation(target_uri="https://bar.com").id]
+
+        result = search.run({"url": "http://bar.com"})
+
+        assert sorted(result.annotation_ids) == sorted(expected_ids)
+
+    def test_filter_distinguishes_net_com_url(self, search, Annotation):
+        Annotation(target_uri="example.com")
+        Annotation(target_uri="https://example.net")
+        expected_ids = [Annotation(target_uri="http://example.com").id]
+
+        params = webob.multidict.MultiDict()
+        params.add("url", "http://example.com")
+        result = search.run(params)
+
+        assert sorted(result.annotation_ids) == sorted(expected_ids)
+
+    def test_ors_multiple_url_uris(self, search, Annotation):
+        Annotation(target_uri="baz.com")
+        Annotation(target_uri="www.foo.com")
+        expected_ids = [Annotation(target_uri="bar.com").id,
+                        Annotation(target_uri="bat.com").id,
+                        Annotation(target_uri="https://foo.com").id,
+                        Annotation(target_uri="http://foo.com").id,
+                        Annotation(target_uri="http://foo/bar.com").id]
+
+        params = webob.multidict.MultiDict()
+        params.add("uri", "bat.com")
+        params.add("uri", "bar.com")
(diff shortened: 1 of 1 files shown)


tests/h/search/query_test.py
```diff
@@ -141,6 +141,50 @@ def search(self, search):
...
+        Annotation(target_uri="https://foo/bar.com")
+        expected_ids = [Annotation(target_uri="bar.com").id]
+        expected_ids.sort()
+
+        result = search.run({"uri": "bar.com"})
+        result.annotation_ids.sort()
+
+        assert result.annotation_ids == expected_ids
+
+    def test_filters_on_http_and_https_url(self, search, Annotation):
```
hmstepanek: Is this a feature or a bug? Is http://foo.bar supposed to match https://foo.bar?

robertknight: Good question. This is a feature 😀. Search for mentions of "httpx" in Slack/Github for the backstory.

hmstepanek:
@robertknight Suggested breaking these up and taking a look at old_query tests to see if we can pull some of those in here as well.

tests/h/search/query_test.py
```diff
@@ -141,6 +141,74 @@ def search(self, search):
...
+        Annotation(target_uri="https://bar.com")
+        Annotation(target_uri="https://foo/bar.com")
+        expected_ids = [Annotation(target_uri="bar.com").id]
+
+        result = search.run({"url": "bar.com"})
+
+        assert sorted(result.annotation_ids) == sorted(expected_ids)
+
+    def test_filters_on_whole_url(self, search, Annotation):
+        Annotation(target_uri="http://bar.com")
```
robertknight: I think a more representative test here would be that given annotations on say `http://site.com/` and `http://site.com/article`, a query for `http://site.com` will only return the first annotation. In other words, the homepage will only show annotations for the homepage and not subpages.

tests/h/search/query_test.py
```diff
@@ -141,6 +141,74 @@ def search(self, search):
...
+        Annotation(target_uri="bar.com")
+        Annotation(target_uri="www.bar.com")
+        expected_ids = [Annotation(target_uri="http://bar.com").id,
+                        Annotation(target_uri="https://bar.com").id]
+
+        result = search.run({"url": "http://bar.com"})
+
+        assert sorted(result.annotation_ids) == sorted(expected_ids)
+
+    def test_filter_distinguishes_net_com_url(self, search, Annotation):
```
robertknight: I think this is covered by the "matches whole URL only" test above.

tests/h/search/query_test.py
```diff
@@ -141,6 +141,74 @@ def search(self, search):
 
 
 class TestUriFilter(object):
+    def test_filters_by_uri(self, search, Annotation):
+        Annotation(target_uri="http://bar.com")
+        Annotation(target_uri="https://bar.com")
+        Annotation(target_uri="https://foo/bar.com")
+        expected_ids = [Annotation(target_uri="bar.com").id]
```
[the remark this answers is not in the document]
hmstepanek: I added a test for matching various pdfs that also mocks storage. I also added a specific test for buggy uris.

hmstepanek:
Ok @robertknight I think I've got everything. PTAL.

tests/h/search/query_test.py
```diff
@@ -141,12 +141,88 @@ def search(self, search):
...
+
+    def test_filters_aliases_http_and_https(self, search, Annotation):
+        expected_ids = [Annotation(target_uri="http://bar.com").id,
+                        Annotation(target_uri="https://bar.com").id]
+
+        result = search.run({"url": "http://bar.com"})
+
+        assert sorted(result.annotation_ids) == sorted(expected_ids)
+
+    def test_filter_identifies_pdfs_marked_as_same_uri(self, search, Annotation, storage):
```
hmstepanek: Do we have any smartness/auto detection for stuff like this built in? I thought I heard someone mention in SF that we did but I'm not sure where/what it is specifically.

tests/h/search/query_test.py
```diff
@@ -141,12 +141,88 @@ def search(self, search):
...
+    def test_filters_on_whole_url(self, search, Annotation):
+        Annotation(target_uri="http://bar.com/foo")
+        expected_ids = [Annotation(target_uri="http://bar.com").id,
+                        Annotation(target_uri="http://bar.com/").id]
+
+        result = search.run({"url": "http://bar.com"})
+
+        assert sorted(result.annotation_ids) == sorted(expected_ids)
+
+    def test_filter_matches_buggy_uri(self, search, Annotation):
```
robertknight: It would be more accurate to describe this as an _invalid_ URI rather than a "buggy" one. It is the _buggy_ client that creates an _invalid URI_ 😜

tests/h/search/query_test.py
robertknight: What did you mean by smartness/auto detection?

For this specific test case, we're actually testing something more general that PDFs, which is that the filter matches any URIs equivalent to the test URI. URI equivalence can be used for HTTP URIs as well. To make this clearer, I would suggest:
 - Changing the test name
 - Moving the setting of `expand_uri`'s return value (not the rest of the storage fixture) into this function so it is easier to see why a query for one URI matches a bunch of annotations that have different URIs.

hmstepanek: I thought we had something that tried to grab the unique id of the pdf and therefore if two people were annotating the same pdf at two different urls it would use the unique id as opposed to the url.

robertknight: Ah, yes we do have logic that gets a unique ID from the PDF, but it works as follows:

- The _client_ reads the unique ID from the PDF, and generates a URI of the form `urn:x-pdf:{fingerprint}`
- The _client_ fetches annotations using an API call of the form `/api/search?uri={http URL of PDF}&uri=urn:x-pdf:{fingerprint}`
- h then returns annotations that match _either_ the HTTP URL or the unique ID

tests/h/search/query_test.py
```diff
@@ -141,12 +141,87 @@ def search(self, search):
...
+        expected_ids = [Annotation(target_uri="http://bar.com").id,
+                        Annotation(target_uri="https://bar.com").id]
+
+        result = search.run({"url": "http://bar.com"})
+
+        assert sorted(result.annotation_ids) == sorted(expected_ids)
+
+    def test_returns_all_annotations_with_equivalent_uris(self, search, Annotation, storage):
+        # Mark all these uri's as equivalent uri's.
+        storage.expand_uri.side_effect = lambda _, x: [
```
robertknight: If you just want to set the return value of a callable you can use `.return_value = {value}` rather than needing to set `side_effect` to a lambda.

Issue #5015, opened 2018-05-08
Resources refactoring


seanh:
- [x] Move all factories into resources.py
- [x] Split resources.py into traversal/roots.py and traversal.contexts.py
- [x] Rename all "factories" to "roots"
- [x] Rename all "resources" to "contexts"
- [x] Docstrings
- [ ] All roots should return contexts, not models
- [ ] Roots should call model classes or services to do DB queries, not do them themselves
- [ ] Some views can be simplified by using contexts
- [ ] Grep the code for "resource". In some places (e.g. variable names, test class names) we're still calling things "resources" that are now called "contexts"
- [ ] Context classes are being used in the tests for various other modules. They should be mocked.

seanh:
I think the overall architecture we're going for is something like this:

![img_1228](https://user-images.githubusercontent.com/22498/39774806-ddf71fa8-52f3-11e8-8adb-d2ef922eab18.jpg)

* Services are built on top of models

* Traversal roots and contexts use models (e.g. calling `@classmethod`s on model classes to get model objects from the DB) and services. And they also use Pyramid (the `request`, permissions).

* Views use traversal contexts (passed in to views as the `context` argument). Traversal roots also do work for views (finding the context object for the view) but are not directly called _by_ views

* Views may pass these context objects into other parts of the code that views call. E.g. into services, or presenters, or renderers, or templates (not pictured)

* Views use services too

* We also have model objects in views too! Either given to views as properties or return values of contexts or services, or views calling `@classmethod`s on model classes.

As you can see, views are sort of the central spoke of everything. They're the integration code that integrates everything together. This is why we need to work hard to keep views small and simple: just do that integration, and move all the actual work into the various other kinds of components.

PR #5090, opened 2018-06-21
Change AuthClientRoot to return AuthClientContext

seanh:
This is part of the resources refactoring here: <https://github.com/hypothesis/h/issues/5015>.

Change `AuthClientRoot` to return `AuthClientContext`'s that wrap `models.AuthClient` objects, rather than returning `AuthClient` models directly. The reason for doing this is that context objects are more generally useful objects for views than model objects used as contexts - context objects can have properties and methods  that do things based on the `request`.

In the case of `AuthClient` though there _isn't_ anything that the views need `AuthClientContext` to do (yet) that a `models.AuthClient` doesn't already do. So this doesn't give us anything, other than putting it in place for potential use, and consistency with how other `*Root` classes are going to behave (but most other `*Root` classes haven't been refactored to return contexts rather than models yet, so there's nothing to be consistent with yet). For this reason I'm closing this PR now and may return to it later.

P.S. This PR also adds an `AuthClientIndexContext` for the page that lists all the `AuthClient`'s to use. This moves the sqlalchemy query out of the view and into the resources, which I like (and is one of the aims of #5015.

### DIFF
--- a/h/traversal/__init__.py
+++ b/h/traversal/__init__.py
@@ -10,6 +10,7 @@ from h.traversal.roots import GroupRoot
 from h.traversal.roots import UserRoot
 from h.traversal.contexts import AnnotationContext
 from h.traversal.contexts import AuthClientContext
+from h.traversal.contexts import AuthClientIndexContext
 from h.traversal.contexts import OrganizationContext
 from h.traversal.contexts import GroupContext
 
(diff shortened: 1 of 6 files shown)


seanh:
Closing this as I'm not intending to merge it now (see comments above), just saving it to potentially return to later.

PR #5092, opened 2018-06-21
Update the database migration template

seanh:
* Add a `-*- coding: utf-8` comment
* Add all the `__future__` imports for Python 3 compat
* Format the code using Black

Example output:

    # -*- coding: utf-8 -*-
    """Add the new table."""
    from __future__ import absolute_import
    from __future__ import division
    from __future__ import print_function

    from alembic import op

    revision = "71c7c16e3a16"
    down_revision = "5dd2dd5547a2"

    def upgrade():
        pass

    def downgrade():
        pass

### DIFF
--- a/h/migrations/script.py.mako
+++ b/h/migrations/script.py.mako
@@ -1,14 +1,14 @@
-"""
-${message}
-"""
-
+<%text># -*- coding: utf-8 -*-</%text>
+"""${message}"""
 from __future__ import unicode_literals
+from __future__ import absolute_import
+from __future__ import division
 
 from alembic import op
 ${imports if imports else ""}
 
-revision = ${repr(str(up_revision))}
-down_revision = ${repr(str(down_revision))}
+revision = "${str(up_revision)}"
+down_revision = "${str(down_revision)}"
 
 
 def upgrade():


seanh:
The `__future__` imports backport Python 3 features to Python 2. We already have the `unicode_literals` one in all of our non-migration Python files and will be adding the others too as we make our Python 2 code as Python 3-compatible as possible before making the jump to Python 3.

What they do: make string literals unicode strings rather than byte strings by default ([PEP 3112](https://www.python.org/dev/peps/pep-3112/)); make imports absolute by default rather than relative to the current package ([PEP 328](https://www.python.org/dev/peps/pep-0328/)); make the division (`/`) operator always do "true" division rather than sometimes true division and sometimes floor division depending on the arguments ([PEP 238](https://www.python.org/dev/peps/pep-0238/)), and import the Python 3 `print()` function ([PEP 3105](https://www.python.org/dev/peps/pep-3105/)).

It's easier just to add these to every file, so that the language is always consistent, rather than e.g. trying to add `from __future__ import division` to a file only once the file actually uses `/`.

Now that I think of it there's probably no DB migration that should be using `print` though (they should use [logging](https://docs.python.org/2/library/logging.html?highlight=logging)), so the `print_function` import may be unnecessary, and maybe it just encourages people to `print()` in migrations when they should be using `logging` instead? On the other hand we do [already have one migration that uses `print()`!](https://github.com/hypothesis/h/blob/d574736751ce081b86c2a073a753b452d7adde87/h/migrations/versions/d536d9a342f3_fill_in_annotation_text_rendered_column.py#L62)

See <http://python-future.org/futurize.html#stage-1-safe-fixes> and <https://docs.python.org/2/library/__future__.html>.

seanh:
Removed the `print_function` import

PR #5101, opened 2018-06-25
Add GroupFilter tests

hmstepanek:
GroupFilter gets all annotations that belong to the specified group id.

### DIFF
--- a/tests/h/search/query_test.py
+++ b/tests/h/search/query_test.py
@@ -89,11 +89,24 @@ class TestAuthFilter(object):
 
 
 class TestGroupFilter(object):
+    def test_matches_only_annotations_from_specified_group(self, search, Annotation, group):
+        Annotation(groupid='group2')
+        Annotation(groupid='group3')
+        group1_annotations = [Annotation(groupid=group.pubid).id,
+                              Annotation(groupid=group.pubid).id]
+
+        result = search.run({'group': group.pubid})
+
+        assert sorted(result.annotation_ids) == sorted(group1_annotations)
 
     @pytest.fixture
     def search(self, search):
         search.append_filter(query.GroupFilter())
         return search
+
+    @pytest.fixture
+    def group(self, factories):
+        return factories.OpenGroup(name="group1", pubid="group1id")
 
 
 class TestGroupAuthFilter(object):


tests/h/search/query_test.py
```diff
@@ -89,12 +89,25 @@ def search(self, search, pyramid_request):
...
+        assert set(result.annotation_ids) == set(group1_annotations)
 
     @pytest.fixture
     def search(self, search):
         search.append_filter(query.GroupFilter())
         return search
 
+    @pytest.fixture
+    def group(self, factories):
+        return factories.OpenGroup(name="group1", pubid="group1")
```
robertknight: The name (the user-visible name) and pubid should be different, as this will reveal an error in the `test_matches_only_annotations_from_specified_group` test above.

tests/h/search/query_test.py
```diff
@@ -89,12 +89,25 @@ def search(self, search, pyramid_request):
...
 class TestGroupFilter(object):
+    def test_matches_only_annotations_from_specified_group(self, search, Annotation, group):
+        Annotation(groupid='group2')
+        Annotation(groupid='group3')
+        group1_annotations = [Annotation(groupid=group.name).id,
+                              Annotation(groupid=group.name).id]
+
+        result = search.run({'group': group.pubid})
+
+        assert set(result.annotation_ids) == set(group1_annotations)
```
robertknight: This should be `sorted(...) == sorted(..)` as in the other tests, no?

robertknight:
Heads up - I had to rebase this on master and run the latest DB migrations to be able to run these tests locally to avoid issues between the DB teardown code and the ongoing user identity work.

tests/h/search/query_test.py
hmstepanek: Oops! I implemented this branch/test a long time ago before we changed to sorted-forgot to even check for that! Good catch!

tests/h/search/query_test.py
```diff
@@ -89,12 +89,25 @@ def search(self, search, pyramid_request):
 
 
 class TestGroupFilter(object):
+    def test_matches_only_annotations_from_specified_group(self, search, Annotation, group):
+        Annotation(groupid='group2')
+        Annotation(groupid='group3')
+        group1_annotations = [Annotation(groupid=group.name).id,
+                              Annotation(groupid=group.name).id]
+
+        result = search.run({'group': group.name})
```
robertknight: Heh - you made the test consistent but it should be the _other_ way. `groupid` should be a _pubid_ (which is unique) not a name (which is not). The "group" argument in search queries is also a pubid.

PR #5104, opened 2018-06-26
Add tests for GroupAuthFilter

hmstepanek:
GroupAuthFilter filters annotations based on the permissions the user has to the particular group(s) that the annotation(s) belong to.

### DIFF
--- a/tests/h/search/query_test.py
+++ b/tests/h/search/query_test.py
@@ -97,6 +97,29 @@ class TestGroupFilter(object):
 
 
 class TestGroupAuthFilter(object):
+    def test_does_not_return_annotations_if_group_not_readable_by_user(
+        self, search, Annotation, group_service
+    ):
+        group_service.groupids_readable_by.return_value = []
+        Annotation(groupid="group2").id
+        Annotation(groupid="group1").id
+        Annotation(groupid="group1").id
+
+        result = search.run({})
+
+        assert not result.annotation_ids
+
+    def test_returns_annotations_if_group_readable_by_user(
+        self, search, Annotation, group_service
+    ):
+        group_service.groupids_readable_by.return_value = ["group1"]
+        Annotation(groupid="group2", shared=True).id
+        expected_ids = [Annotation(groupid="group1").id,
+                        Annotation(groupid="group1").id]
+
+        result = search.run({})
+
+        assert sorted(result.annotation_ids) == sorted(expected_ids)
 
     @pytest.fixture
     def search(self, search, pyramid_request):


tests/h/search/query_test.py
```diff
@@ -3,8 +3,10 @@
 
 import pytest
 import webob
+import mock
```
robertknight: Nitpick until we're auto-formatting code with Black - groups of imports should generally be sorted.

tests/h/search/query_test.py
```diff
@@ -97,12 +99,39 @@ def search(self, search):
...
+
+        assert sorted(result.annotation_ids) == sorted(users_private_ids)
 
     @pytest.fixture
     def search(self, search, pyramid_request):
         search.append_filter(query.GroupAuthFilter(pyramid_request))
         return search
 
+    @pytest.fixture
+    def group_service(self, pyramid_config):
```
robertknight: This fixture has basically the same definition as the `h.search` package's `group_service` fixture. I would suggest simply re-using that fixture and removing this override. This avoids a potential hazard where a change in the name of the package's fixture could cause this one to be used alongside it instead of overriding it, which would break the tests.

If you did want to override a fixture in future to add more complex behaviour, a way to avoid this hazard is to import the "super" fixture as a dependency of the override and then customize its behavior, as we often do with module-specific `pyramid_config` or `pyramid_request` overrides.

tests/h/search/query_test.py
```diff
@@ -97,12 +99,39 @@ def search(self, search):
 
 
 class TestGroupAuthFilter(object):
+    def test_unauthorized_user_can_not_see_the_groups_annotations(self,
```
robertknight: The content of these two tests looks fine, but I think the descriptions should be generalized:

- Annotations not in groups readable by the current user should not be returned by search
- Annotations in groups readable by the current user should be returned by search

These two general cases cover various concrete cases (eg. logged-out users not seeing annotations in private groups, logged-in users not seeing annotations in private groups they have not joined, logged-in users seeing annotations in private groups they have joined).

tests/h/search/query_test.py
```diff
@@ -97,6 +97,27 @@ def search(self, search):
 
 
 class TestGroupAuthFilter(object):
+    def test_does_not_return_annotations_if_group_not_readable_by_user(self,
+            search, Annotation, group_service):
```
robertknight: This triggers an [E128 warning](https://stackoverflow.com/questions/15435811/what-is-pep8s-e128-continuation-line-under-indented-for-visual-indent) in my editor's live linter. If we're adopting black's code style, it avoids this like so:

```
def test_does_not_return_annotations_if_group_not_readable_by_user(
    self, search, Annotation, group_service
):
```

tests/h/search/query_test.py
```diff
@@ -97,6 +97,27 @@ def search(self, search):
...
+
+        result = search.run({})
+
+        assert not result.annotation_ids
+
+    def test_returns_annotations_if_group_readable_by_user(self,
+            search, Annotation, group_service):
+        group_service.groupids_readable_by.return_value = ["group1"]
+        Annotation(groupid="group2", shared=True).id
+        users_private_ids = [Annotation(groupid="group1").id,
```
robertknight: These could be the users private IDs, but they could also be annotations in a private group that the user is a member of. Perhaps just rename this `expected_ids`?

tests/h/search/query_test.py
```diff
@@ -97,6 +97,27 @@ def search(self, search):
...
+        group_service.groupids_readable_by.return_value = []
+        Annotation(groupid="group2").id
+        Annotation(groupid="group1").id
+        Annotation(groupid="group1").id
+
+        result = search.run({})
+
+        assert not result.annotation_ids
+
+    def test_returns_annotations_if_group_readable_by_user(self,
```
robertknight: Same formatting issue as above.

PR #5112, opened 2018-06-29
Fix reading of `SECRET_KEY` and `SECRET_SALT` env vars in Python 3

robertknight:
If these env vars are set, they are read as strings. Calling
`bytes(str_value)` to cast them to byte strings fails in Python 3.
Instead `str.encode` has to be used, which defaults to UTF-8.

This fixes an issue that got overlooked during the recent settings refactoring.

### DIFF
--- a/h/config.py
+++ b/h/config.py
@@ -56,8 +56,8 @@ def configure(environ=None, settings=None):
     settings_manager.set('statsd.prefix', 'STATSD_PREFIX')
 
     # Configuration for Pyramid
-    settings_manager.set('secret_key', 'SECRET_KEY', type_=bytes, required=True)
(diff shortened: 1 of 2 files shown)


h/config.py
```diff
@@ -123,3 +123,9 @@ def configure(environ=None, settings=None):
     es_logger.addFilter(ExceptionFilter((("ReadTimeout", "WARNING"),)))
 
     return Configurator(settings=settings)
+
+
+def _to_utf8(str_or_bytes):
```
hmstepanek: I wish this wasn't so long but I don't really see a better way to do this.

hmstepanek: I wish this wasn't so long but I'm don't really see a better way to do this.

PR #5114, opened 2018-06-30
Fix flakey AnyMatcher tests

hmstepanek:
AnyMatcher matches against several different fields inside of an
annotation. The test factory randomly sets fields that aren't
explicitly set and sometimes these settings conflict with the test.
Aka if we search for any="bar", the text field may be randomly set
to a string that includes bar therefore we need to set these fields
explicitly in the test and not depend on the random setting from
the factory.

### DIFF
--- a/tests/h/search/query_test.py
+++ b/tests/h/search/query_test.py
@@ -321,6 +321,7 @@ class TestAnyMatcher(object):
         Annotation(tags=["baz"])
         Annotation(target_uri="baz.com")
         Annotation(text="baz is best")
+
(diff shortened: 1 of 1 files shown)


hmstepanek:
The target selector produces the same default setting that doesn’t include any conflicting match and the uri doesn’t either. Good call on the overriding the fixture-I’ll do that.

robertknight:
Hi @hmstepanek - Since you're away this week and this could block other PRs I'm going to make the changes discussed above and then get this merged. Thanks for tracking down the problem.

hmstepanek:
Thanks Rob!

PR #5116, opened 2018-07-02
Optimize redirect lookup

robertknight:
Computing the `request.path` property in `webob/request.py` involves
reading values from the environment, then decoding, URL-quoting and
concatenating the results. This has enough overhead that it is worth
doing once rather than for each of the 214 redirect rules that we have,
on each request to h. Thread profiling via New Relic suggests this
accounts for ~8% of active request processing time.

A better but slightly more complex optimization would be to use a more
efficient data structure for matching the request path against the
redirect rules.

### DIFF
--- a/h/util/redirects.py
+++ b/h/util/redirects.py
@@ -48,11 +48,16 @@ def lookup(redirects, request):
     Returns None if the request does not match, and the URL to redirect to
     otherwise.
     """
+
+    # Compute and cache `request.path` once, rather than recomputing for each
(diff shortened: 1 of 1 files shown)


seanh:
Wow, I'm kinda surprised that WebOb doesn't cache `request.path` itself already

robertknight:
CPU time spent in `h.utils.redirects.redirect` dropped from 0.9% of total time to 0.1% in prod after this went out (or ~8% to >1% of non-idle time).

PR #5124, opened 2018-07-04
Fix up flags_test.py

seanh:
Various fixes to flags_test.py:

* This is actually mocking GroupfinderService not GroupService, so fix the mock/fixture name
* Use autospec to mock things
* Remove some noise

### DIFF
--- a/tests/h/views/api/flags_test.py
+++ b/tests/h/views/api/flags_test.py
@@ -8,10 +8,11 @@ import pytest
 from pyramid.httpexceptions import HTTPNoContent
 
 from h.views.api import flags as views
+from h.services.groupfinder import GroupfinderService
 from h.traversal import AnnotationContext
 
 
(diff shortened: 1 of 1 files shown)


tests/h/views/api/flags_test.py
```diff
@@ -8,105 +8,114 @@
...
+                                                       annotation_context,
                                                        factories,
-                                                       group_service,
+                                                       groupfinder_service,
                                                        pyramid_request,
                                                        mailer):
-        annotation_context = mock.create_autospec(AnnotationContext, instance=True, annotation=factories.Annotation())
-        group_service.find.return_value = factories.Group(creator=None, members=[])
+        groupfinder_service.find.return_value = factories.Group()
+        groupfinder_service.find.return_value.creator = None
```
robertknight: Could this line by moved into a `creator=None` kwarg to the `factories.Group` constructor?

PR #5125, opened 2018-07-05
Add test for `DeletedFilter`

robertknight:
Unlike the rest of the fields which are set by AnnotationSearchIndexPresenter
when the annotation is indexed, the `deleted` field is set in the app by
`h.search.index.delete` which replaces the annotation's entire body in
the index. To ensure the delete operation/deleted filter match up, I've re-used
`h.search.index.delete` in the tests.

Add an optional `refresh` kwarg to `delete` to force a refresh so that
the change is immediately visible in search. Otherwise there will be a
brief delay until searches reflect the deletion.

### DIFF
--- a/h/search/index.py
+++ b/h/search/index.py
@@ -60,7 +60,7 @@ def index(es, annotation, request, target_index=None):
     )
 
 
-def delete(es, annotation_id, target_index=None):
+def delete(es, annotation_id, target_index=None, refresh=False):
     """
     Mark an annotation as deleted in the search index.
 
@@ -77,6 +77,10 @@ def delete(es, annotation_id, target_index=None):
 
     :param target_index: the index name, uses default index if not given
     :type target_index: unicode
+
+    :param refresh: Force this deletion to be immediately visible to search operations
+    :type refresh: bool
+
     """
 
     if target_index is None:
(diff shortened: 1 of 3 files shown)


h/search/index.py
```diff
@@ -60,7 +60,7 @@ def index(es, annotation, request, target_index=None):
     )
 
 
-def delete(es, annotation_id, target_index=None):
+def delete(es, annotation_id, target_index=None, refresh=False):
```
sheetaluk: We do want the deletion to be immediately visible, don't we? Shouldn't this default to True?

tests/h/search/query_test.py
```diff
@@ -4,6 +4,7 @@
 import pytest
 import webob
 
+from h.search.index import delete
 from h.search import Search, query
```
sheetaluk: Could we remove this line and in the following line import `index` also? So: `from h.search import Search, query, index`

h/search/index.py
robertknight: In tests we do want the deletion to be immediately visible. In the app the current behaviour is that `refresh` is `False` which means that the change becomes visible after a short delay (of about a second or so). I've left the default as `False` here to avoid changing behaviour in the actual app.

I personally think we might want to change the behaviour of the API in production at some point to make it so that delete operations are guaranteed to be immediately in search responses, but that's a separate exercise.

tests/h/search/query_test.py
robertknight: Done. I was originally trying to avoid confusion with the fixture named `index` but I see we aren't using it in these tests.

h/search/index.py
sheetaluk: Sounds good.

sheetaluk:
LGTM.

Issue #5128, opened 2018-07-06
Annotation indexing spends large amount of time in user table lookups


robertknight:
New Relic [transaction stats](https://rpm.newrelic.com/accounts/1385283/applications/22688631/transactions?type=other&show_browser=false#id=5b224f746865725472616e73616374696f6e2f43656c6572792f682e7461736b732e696e64657865722e6164645f616e6e6f746174696f6e222c22225d) for "Non Web" processes show that `h.tasks.indexer.add_annotation` is the most expensive transaction, taking 150ms on average.

The transaction breakdown shows that nearly half the time is spent in SELECT operations on the `user` table, which is unexpected, given that an annotation is associated with only a single user which we might need to fetch.

Digging into what happens that involves user table lookups during indexing, I see that the annotation document indexed into ES gets `{"nipsa": true}` set if the user has the "nipsa" flag set. However the way that the user's "nipsa" status is tested is [extremely inefficient](https://github.com/hypothesis/h/blob/747fa98560af51bfbe81ace1df75c8bc6e3f751f/h/services/nipsa.py#L27). Instead of just fetching the annotation's user and then looking at `user.nipsa`, the `NipsaService` instead scans the `user` table, loads all users with `nipsa=True` and then gets their IDs (!) The result is then cached for the duration of the request. There is no index on the `nipsa` field as far as I can see, so this requires a full table scan. Additionally in production we now have hundreds of NIPSA-d users, so this involves loading a lot of userids from the user table.

When fixing this issue, please dive into the history of `NipsaService` for any info about why that cache existed in the first place. It may have to do with rendering of annotations in search results.

robertknight:
This [thread profile](https://rpm.newrelic.com/accounts/1385283/applications/22688631/profiles/3017537) backs up the NIPSA theory. If you expand the trace, ignoring the large percentage of the time where Celery is blocking waiting for incoming tasks to execute, the majority of "active" time is spent in the NIPSA service.

PR #5129, opened 2018-07-06
Optimize user nipsa/shadowban status lookups for a single userid

robertknight:
When indexing annotations or delivering WebSocket messages about annotations we check whether a user has been shadowbanned (NIPSA-d). The way this is done is currently inefficient. This PR modifies `NipsaService` to make it work better for our current usage patterns.

The first commit optimizes NIPSA flag lookups for a single userid in a request. Instead of loading the IDs of _all_ NIPSA-ed users, which AFAIK involves a full-table scan since `user.nipsa` is not indexed, then checking for the ID in that set, just load the single user and check their `user.nipsa` status.

The second commit removes a cache that currently doesn't provide any value because after the first commit, `flagged_userids` is only used by the admin panel which renders a list of IDs of all NIPSA-ed users.

This should fix https://github.com/hypothesis/h/issues/5128 and also improve (but not completely resolve) https://github.com/hypothesis/h/issues/5030

### DIFF
--- a/tests/h/services/nipsa_test.py
+++ b/tests/h/services/nipsa_test.py
@@ -26,7 +26,11 @@ class TestNipsaService(object):
         svc = NipsaService(db_session)
 
         assert not svc.is_flagged('acct:dominic@example.com')
-        assert not svc.is_flagged('acct:romeo@example.com')
+
+    def test_is_flagged_returns_false_for_unknown_users(self, db_session):
+        svc = NipsaService(db_session)
+
+        assert not svc.is_flagged('acct:not_in_the_db@example.com')
 
     def test_flag_sets_nipsa_true(self, db_session, users):
         svc = NipsaService(db_session)


tests/h/services/nipsa_test.py
```diff
@@ -26,7 +26,11 @@ def test_is_flagged_returns_false_for_unflagged_users(self, db_session):
         svc = NipsaService(db_session)
 
         assert not svc.is_flagged('acct:dominic@example.com')
-        assert not svc.is_flagged('acct:romeo@example.com')
```
robertknight: This userid is not in the DB. I've moved this into a more explicit test below.

h/services/nipsa.py
```diff
@@ -54,9 +52,6 @@ def unflag(self, user):
         user.nipsa = False
         reindex_user_annotations.delay(user.userid)
 
-    def clear(self):
-        self._flagged_userids = None
```
robertknight: If we don't have a cache any more, we don't need a method to clear it.

h/services/nipsa.py
```diff
@@ -23,14 +22,13 @@ def flagged_userids(self):
 
         :rtype: set of unicode strings
         """
-        if self._flagged_userids is None:
-            query = self.session.query(User).filter_by(nipsa=True)
-            self._flagged_userids = set([u.userid for u in query])
-        return self._flagged_userids
+        query = self.session.query(User).filter_by(nipsa=True)
```
robertknight: This query is still inefficient AFAIK since `user.nipsa` is not indexed, but it doesn't matter since it is only used by the `/admin/nipsa` view following the change to `is_flagged` below.

seanh:
Any way to test the efficiency difference locally?

robertknight:
> Any way to test the efficiency difference locally?

One way would be to create thousands of user accounts, NIPSA a hundred of them (say 2-3% of the total) and measure the time taken to index an annotation in microseconds before/after.

You could also look at the SQL generated by SQLAlchemy for the previous `flagged_userids` implementation and use `EXPLAIN {query}` in psql to verify that it does result in a much less efficient query than what the new code does.

Or finally you can take my word for it and look at the NR transaction traces in production before/after 😛

h/services/nipsa.py
hmstepanek: You are correct it does not appear to be indexed and this may be slow for the admin view but I suppose we aren't overly concerned about that use case.

PR #5130, opened 2018-07-09
Add tests for UsersAggregator.

sheetaluk:


tests/h/search/query_test.py
```diff
@@ -513,11 +513,31 @@ def search(self, search):
...
+        assert count_for_pb is 1
+
+    def test_it_limits_annotation_aggregation_on_users(self, Annotation, search):
+        for i in range(10):
+            Annotation(userid="acct:pa@example.com")
+
+        search.append_aggregation(query.UsersAggregation(limit=5))
+        result = search.run({})
+
+        count_for_user_a = next(r for r in result.aggregations.get('users') if r["user"] == "acct:pa@example.com")['count']
```
sheetaluk: Shouldn't this count returned be limited to 5?

tests/h/search/query_test.py
```diff
@@ -513,11 +513,32 @@ def search(self, search):
...
+            Annotation(userid="acct:pa@example.com")
+        Annotation(userid="acct:pb@example.com")
 
-    @pytest.fixture
-    def search(self, search):
         search.append_aggregation(query.UsersAggregation())
-        return search
+        result = search.run({})
+
+        count_for_pa = next(r for r in result.aggregations.get('users') if r["user"] == "acct:pa@example.com")['count']
```
robertknight: Can you please keep line lengths to < 100 chars, with <~ 90 recommended for most lines. Lines longer than become annoying to read in various tools (eg. GitHub's diff viewer).

tests/h/search/query_test.py
robertknight: The `limit` parameter refers to the _number_ of buckets, not the counts for each bucket. See https://hyp.is/KvOyhoNjEeiTa6dg6L6VGw/www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations-bucket-terms-aggregation.html

tests/h/search/query_test.py
robertknight: In this case depending on the complexity of the `result.aggregations` data structure it may simply be better to test for an exact match against a dict literal.

tests/h/search/query_test.py
```diff
@@ -513,11 +513,32 @@ def search(self, search):
...
+        count_for_pb = next(r for r in result.aggregations.get('users') if r["user"] == "acct:pb@example.com")['count']
+
+        assert len(result.aggregations.get('users')) is 2
+        assert count_for_pa is 2
```
robertknight: This should be fine given the small number of documents, but a caveat with aggregations in general is that the document counts are approximate.

sheetaluk: 👍 Yeah, I read that in the documentation. I think this small number is good for tests.

tests/h/search/query_test.py
```diff
@@ -513,11 +513,41 @@ def search(self, search):
...
-    def search(self, search):
         search.append_aggregation(query.UsersAggregation())
-        return search
+        result = search.run({})
+
+        users_results = result.aggregations.get('users')
+        count_pa = next(r for r in users_results if r["user"] == "acct:pa@example.com")['count']
+        count_pb = next(r for r in users_results if r["user"] == "acct:pb@example.com")['count']
+
+        assert len(users_results) is 2
```
robertknight: `==` rather than `is` should be used for comparing values for _equality_ rather than _identity_.  This code _happens_ to work because Python interns (re-uses the same object) for small integer values, but that is not guaranteed AFAIK.

The main cases to use `is` are: `assert x is None|True|False` or `assert x is y` where "x" and "y" are objects and you really want them to be the _same_ object and not just equivalent.

tests/h/search/query_test.py
```diff
@@ -513,11 +513,41 @@ def search(self, search):
...
+        search.append_aggregation(query.UsersAggregation(limit=bucket_limit))
+        result = search.run({})
+
+        users_results = result.aggregations.get('users')
+        count_pb = next(r for r in users_results if r["user"] == "acct:pb@example.com")['count']
+        count_pc = next(r for r in users_results if r["user"] == "acct:pc@example.com")['count']
+
+        assert len(users_results) is bucket_limit
+        assert count_pb is 3
+        assert count_pc is 2
```
robertknight: See comments about `is` vs `==` above.

PR #5131, opened 2018-07-09
Add tests for TagsAggregator.

sheetaluk:

PR #5133, opened 2018-07-09
Remove unused code blocks in tags and users aggregators.

sheetaluk:
The result from an aggregation is never an None or empty.
This is not a necessary check.

See https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations-bucket-terms-aggregation.html


sheetaluk:
Just to be safe, will merge right after #5131 and #5130 are merged.

PR #5136, opened 2018-07-11
Add integrated tests for search result ordering

robertknight:
Test support for:

 - Sorting by "updated" and "created" fields
 - Default sort ordering
 - What happens when an invalid sort field is specified

As far as I can tell we only ever sort annotations in chronological
order of last update in h and the client (ie. sort by updated,
descending). However the search API currently supports sorting by any
field which is indexed in Elasticsearch. We might be able to drop that
flexibility if it is unused.

### DIFF
--- a/tests/h/search/query_test.py
+++ b/tests/h/search/query_test.py
@@ -1,10 +1,46 @@
 # -*- coding: utf-8 -*-
 from __future__ import unicode_literals
 
+import datetime
+
 import pytest
 import webob
 
 from h.search import Search, index, query
+
+
+class TestBuilder(object):
+    @pytest.mark.parametrize("sort_key,order,expected_order", [
+        # Sort supports "updated" and "created" fields.
+        ("updated", "desc", [0, 1, 2]),
+        ("updated", "asc", [2, 1, 0]),
+        ("created", "desc", [2, 1, 0]),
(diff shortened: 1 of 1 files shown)


tests/h/search/query_test.py
```diff
@@ -1,12 +1,48 @@
...
+        # Default sort order should be descending.
+        ("updated", None, [0, 1, 2]),
+
+        # Default sort field should be "updated".
+        (None, "asc", [2, 1, 0]),
+    ])
+    def test_it_sorts_annotations(self, Annotation, search, sort_key, order, expected_order):
+        dt = datetime.datetime
+        ann_ids = [Annotation(updated=dt(2018, 1, 1), created=dt(2016, 1, 1)).id,
+                   Annotation(updated=dt(2017, 1, 1), created=dt(2017, 1, 1)).id,
```
sheetaluk: This is a bit nit picky, but these 3 annotations are in an order(desc by updated and asc by created). Could we shuffle them a bit, maybe just switch the first 2 annotations? Otherwise LGTM

robertknight: Done. I put the `2017` annotation first so that creation order is different to sorted order in all cases and added a comment to explain the creation ordering.

PR #5140, opened 2018-07-11
Add annotation mapping for Elasticsearch 6

hmstepanek:
Add mappings that define the schema of indexed annotations for
Elasticsearch 6. Compared to the existing `ANNOTATION_MAPPING`, the
following changes have been made:

 - The indexed fields are sorted alphabetically
 - "string" fields have been changed to use either "text" or "keyword"
   types depending on whether we need to support search within those
   fields ("text") or only on the whole content of the field ("keyword")
 - Fields which we have no need to search or sort on have been removed.
   This includes:

    - `annotator_schema_version`
    - `target.selector` fields other than `exact` (the annotation's
      quote)
 - The "id" field has been explicitly added since the document's `_id`
   field is not configurable in ES6.
 - The `dynamic` option has been set to `false` to disable indexing of
   any fields which are not explicitly listed in the mapping. This
   avoids anyone relying on being able to search fields which happen to
   get written to the index but are not explicitly listed as searchable.

The analysis settings have not been changed since the existing settings
work with ES6.

Since the mapping depends on which es client is passed in (es1 or es6)
the mapping was exposed as a parameter to the higher level code as the
higher level code knows which version of the client it's passing.

### DIFF
--- a/h/search/config.py
+++ b/h/search/config.py
@@ -11,13 +11,27 @@ these settings in an Elasticsearch instance.
 from __future__ import unicode_literals
 
 import binascii
+import elasticsearch1
+import elasticsearch
 import logging
 import os
 
-from elasticsearch1.exceptions import NotFoundError, RequestError
 
 log = logging.getLogger(__name__)
 
+ES_NOTFOUND_ERRORS = (
+    elasticsearch1.exceptions.NotFoundError,
+    elasticsearch.exceptions.NotFoundError,
+)
+ES_REQUEST_ERRORS = (
+    elasticsearch1.exceptions.RequestError,
+    elasticsearch.exceptions.RequestError,
+)
+
+# Elasticsearch mapping type for annotations for ES 1.x.
+#
+# These definition includes a number of legacy fields which don't need to be
+# indexed for historical reasons.
 ANNOTATION_MAPPING = {
     '_id': {'path': 'id'},
     '_source': {'excludes': ['id']},
@@ -105,6 +119,72 @@ ANNOTATION_MAPPING = {
     }
 }
 
+# Elasticsearch type mapping for annotations for ES 6.x and later.
+# This mapping does not include the legacy fields from the ES 1.x mapping.
+ES6_ANNOTATION_MAPPING = {
+    # Ignore unknown fields and do not add them to the mapping.
+    # This ensures that only fields included in the "properties" section
+    # here can be searched against.
+    'dynamic': False,
+
+    # Indexed (searchable) fields.
+    'properties': {
+        'authority': {'type': 'keyword'},
+        'created': {'type': 'date'},
+        'deleted': {'type': 'boolean'},
+        'document': {
+            'enabled': False,  # not indexed
+        },
+        'group': {'type': 'keyword'},
+        'id': {'type': 'keyword'},
+        'quote': {'type': 'text', 'analyzer': 'uni_normalizer'},
+        'references': {'type': 'keyword'},
+        'shared': {'type': 'boolean'},
+        'tags': {'type': 'text', 'analyzer': 'uni_normalizer'},
+        'tags_raw': {'type': 'keyword'},
+        'text': {'type': 'text', 'analyzer': 'uni_normalizer'},
+        'target': {
+            'properties': {
+                'source': {
+                    'type': 'text',
+                    'analyzer': 'uri',
+                    'copy_to': ['uri'],
+                },
+                # We store the 'scope' unanalyzed and only do term filters
+                # against this field.
+                'scope': {
+                    'type': 'keyword',
+                },
+                'selector': {
+                    'properties': {
+                        # Open Annotation TextQuoteSelector
+                        'exact': {
+                            'copy_to': 'quote',
+                            'type': 'text',
+                            'index': False,
+                        },
+                    }
+                }
+            }
+        },
+        'thread_ids': {'type': 'keyword'},
+        'updated': {'type': 'date'},
+        'uri': {
+            'type': 'text',
+            'analyzer': 'uri',
+            'fields': {
+                'parts': {
+                    'type': 'text',
+                    'analyzer': 'uri_parts',
+                },
+            },
+        },
+        'user': {'type': 'text', 'analyzer': 'user'},
+        'user_raw': {'type': 'keyword'},
+    }
+}
+
+# Filter and tokenizer definitions shared by ES 1.x and ES 6.x mappings.
 ANALYSIS_SETTINGS = {
     'char_filter': {
         'strip_scheme': {
@@ -178,10 +258,10 @@ def init(client):
 def configure_index(client):
     """Create a new randomly-named index and return its name."""
     index_name = client.index + '-' + _random_id()
-
+    mapping = _get_mapping(client)
     client.conn.indices.create(index_name, body={
         'mappings': {
-            client.mapping_type: ANNOTATION_MAPPING,
+            client.mapping_type: mapping,
         },
         'settings': {
             'analysis': ANALYSIS_SETTINGS,
@@ -199,7 +279,7 @@ def get_aliased_index(client):
     """
     try:
         result = client.conn.indices.get_alias(name=client.index)
-    except NotFoundError:  # no alias with that name
+    except ES_NOTFOUND_ERRORS:  # no alias with that name
         return None
     if len(result) > 1:
         raise RuntimeError("We don't support managing aliases that "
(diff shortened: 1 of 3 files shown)


tests/h/cli/commands/search_test.py
```diff
@@ -44,8 +45,7 @@ def test_handles_runtimeerror(self, cli, cliconfig, update_index_settings):
 
     @pytest.fixture
     def update_index_settings(self, patch):
-        config = patch('h.cli.commands.search.config')
```
hmstepanek: This was patching the whole module instead of just the one function so it was causing the ANNOTATION_MAPPING to get mocked hence the change to just mock the function under test.

hmstepanek:
The way I see it we have a couple options: we can either do it this way by exposing the mapping to the higher level code, we can add the mapping as an attr onto the client obj, or we can add the mapping as a pyramid setting. At first I was going to add it to the client obj but I didn't want to introduce that as a dependency. It kinda seems like that may have been less of a disruption after implementing it the other way. Thoughts?

robertknight:
> The way I see it we have a couple options: we can either do it this way by exposing the mapping to the higher level code, we can add the mapping as an attr onto the client obj, or we can add the mapping as a pyramid setting.

I think we should avoid trying to over-generalize here. We only need to _temporarily_ select between two different mappings depending on the ES version. Shortly after we get ES 6 working in production we'll be deleting the code to support ES 1.x.

This was my rationale for adding a property to the `Client` to indicate the ES version and then simply switching on it in a small number of places (three, in fact). When we come to cleanup it becomes very easy to grep for `using_es6` and just remove all of the branches that are no longer required.

Maybe in future we _will_ need to be able to initialize multiple indexes with different mappings, eg. if we were to support searching for users or groups. However we don't have a requirement for that yet and when we do there will be a bunch of other issues to tackle as well, so I would avoid trying to "pave the way" too soon.

Lastly, from a conceptual point of view, I see the Elasticsearch schema as an implementation detail of the `h.search` package. This implies that it shouldn't be exposed or referenced in code outside of that package, with the exception of the logic in `h.presenters` that generates the JSON documents to be indexed.

tests/h/search/client_test.py
```diff
@@ -37,6 +37,7 @@ def test_conn_is_read_only(self):
     def test_defaults_to_es1(self):
         client = Client(host="http://localhost:9200", index="hypothesis")
 
+        assert client.version == "1.10.0"
```
hmstepanek: I thought it would be kinda unnecessary to split the verification of the version property into a separate test-we can but I feel like doing it here is fine.

tests/h/search/config_test.py
```diff
@@ -168,22 +181,58 @@ def test_updates_index_atomically(self, client):
...
         """Raise if called for a concrete index."""
-        client.conn.indices.get_alias.side_effect = NotFoundError('test', 'test desc')
+        client.conn.indices.get_alias.side_effect = error('test', 'test desc')
 
         with pytest.raises(RuntimeError):
             update_aliased_index(client, 'new-target')
 
 
+class TestUpdateIndexSettings(object):
+    # todo: add a test for positive behavior
```
hmstepanek: Note this test class didn't exist before at all. I would like to add another test here for when everything goes right to verify the correct mapping gets used but I ran out of time and wanted to get your eyes on this again.

tests/h/search/config_test.py
```diff
@@ -145,7 +158,7 @@ def test_returns_underlying_index_name(self, client):
 
     def test_returns_none_when_no_alias(self, client):
         """If ``index`` is a concrete index, return None."""
-        client.conn.indices.get_alias.side_effect = NotFoundError('test', 'test desc')
+        client.conn.indices.get_alias.side_effect = elasticsearch1.NotFoundError('test', 'test desc')
```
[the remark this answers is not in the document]
hmstepanek: update_aliased_index and update_index_settings both call get_aliased_index and get_aliased_index is actually where the error handling occurs so I think actually we should just check it there and not in the other tests.

h/search/config.py
```diff
@@ -105,6 +121,72 @@
...
+    # Indexed (searchable) fields.
+    'properties': {
+        'authority': {'type': 'keyword'},
+        'created': {'type': 'date'},
+        'deleted': {'type': 'boolean'},
+        'document': {
+            'enabled': False,  # not indexed
+        },
+        'group': {'type': 'keyword'},
+        'id': {'type': 'keyword'},
```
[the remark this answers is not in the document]
hmstepanek: I don't see the nipsa field in the original mapping either?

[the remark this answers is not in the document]
robertknight: It was missing from them too. It happened to work because by default Elasticsearch will index fields that aren’t in the schema with default settings.

h/search/config.py
```diff
@@ -105,6 +121,73 @@
...
+    # This ensures that only fields included in the "properties" section
+    # here can be searched against.
+    'dynamic': False,
+
+    # Indexed (searchable) fields.
+    'properties': {
+        'authority': {'type': 'keyword'},
+        'created': {'type': 'date'},
+        'deleted': {'type': 'boolean'},
+        'nipsa': {'type': 'boolean'},
```
robertknight: These properties should be sorted alphabetically. Can you move the `nipsa` definition down accordingly?

h/search/config.py
```diff
@@ -11,13 +11,29 @@
...
+import elasticsearch
 import logging
 import os
 
-from elasticsearch1.exceptions import NotFoundError, RequestError
 
 log = logging.getLogger(__name__)
 
+ELASTICSEARCH1_VERSION = ".".join(str(item) for item in elasticsearch1.__version__)
+ELASTICSEARCH_VERSION = ".".join(str(item) for item in elasticsearch.__version__)
```
robertknight: This needs updating/simplifying to reflect the type change of `client.__version__`. It might not be necessary to declare these constants - tests could just reference `elasticsearch1.__version__` directly.

tests/h/search/config_test.py
```diff
@@ -123,13 +130,19 @@ def test_returns_index_name(self, client, matchers):
 
         assert name == matchers.Regex('foo-[0-9a-f]{8}')
 
-    def test_sets_correct_mappings_and_settings(self, client):
+    @pytest.mark.parametrize(
+        'esversion,mapping',
+        ((ELASTICSEARCH1_VERSION, ANNOTATION_MAPPING),
+         ('6.2.0', ES6_ANNOTATION_MAPPING))
```
robertknight: The `'6.2.0'` needs updating here to be a tuple.

tests/h/search/config_test.py
```diff
@@ -178,12 +195,78 @@ def test_updates_index_atomically(self, client):
...
+                "settings": {
+                    "index": {
+                        "analysis": {
+                            "old_setting": "val",
+                        }
+                    }
+                }
+            }
+        }
+        update_index_settings(client)
```
robertknight: Can you add a blank line before and after the `update_index_settings` call to make it clearer which are the _arrange_, _act_ and _assert_ phases of the test please.

tests/h/search/config_test.py
```diff
@@ -8,19 +8,26 @@
...
+    ES6_ANNOTATION_MAPPING,
     init,
     configure_index,
     get_aliased_index,
     update_aliased_index,
+    update_index_settings,
 )
 
 
+ANNOTATION_MAPPINGS = (ANNOTATION_MAPPING, ES6_ANNOTATION_MAPPING)
```
robertknight: The `ANNOTATION_MAPPINGS` constant isn't used.

tests/h/search/config_test.py
```diff
@@ -178,12 +195,78 @@ def test_updates_index_atomically(self, client):
 
     def test_raises_if_called_for_concrete_index(self, client):
         """Raise if called for a concrete index."""
-        client.conn.indices.get_alias.side_effect = NotFoundError('test', 'test desc')
+        client.conn.indices.get_alias.side_effect = (
+            elasticsearch1.exceptions.NotFoundError('test', 'test desc'))
```
robertknight: I think this needs to be parametrized with the `NotFoundError` classes from the two different versions of the ES library?

tests/h/search/config_test.py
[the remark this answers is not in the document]
hmstepanek: @robertknight The above comment pertains to your comment about wanting to test the NotFoundError in two different places. I'm not sure that's necessary as we would be essentially testing the same thing twice.

tests/h/search/config_test.py
hmstepanek: See my previous comment that I tagged you on. I'm not sure we need to test the same thing twice but if you still feel like it needs to test it again feel free to add it.

h/search/config.py
```diff
@@ -236,7 +317,7 @@ def delete_index(client, index_name):
 
     try:
         client.conn.indices.delete(index=index_name)
-    except NotFoundError:
+    except elasticsearch1.exceptions.NotFoundError:
```
robertknight: I don't know if we'll encounter this with ES 6 as well. Let's keep an eye open when we run reindexing operations.

PR #5147, opened 2018-07-13
Add version as property on Client class

hmstepanek:
version is the elasticsearch version pulled directly from the
elasticsearch library argument. This is needed for identifying
what version of elasticsearch the client object is wrapping.

### DIFF
--- a/h/search/client.py
+++ b/h/search/client.py
@@ -12,7 +12,8 @@ class Client(object):
     """
     A convenience wrapper around a connection to Elasticsearch.
 
-    Holds a connection object, an index name, and the name of the mapping type.
+    Holds a connection object, an index name, the elasticsearch library version,
+    and the name of the mapping type.
 
     :param host: Elasticsearch host URL
     :param index: index name


h/search/client.py
```diff
@@ -12,14 +12,15 @@ class Client(object):
...
-    Holds a connection object, an index name, and the name of the mapping type.
+    Holds a connection object, an index name, es version, and the name of the mapping type.
 
     :param host: Elasticsearch host URL
     :param index: index name
     :param elasticsearch: Elasticsearch library defaulted to elasticsearch1
     """
 
     def __init__(self, host, index, elasticsearch=elasticsearch1, **kwargs):
+        self._version = '.'.join(str(item) for item in elasticsearch.__version__)
```
robertknight: Was there any particular reason for converting to a string here? I expect that the main use of the `client.version` property will be in comparisons, and for that tuples are better representations because Python will conveniently do item-wise comparisons:

```py
(10,0,0) > (9,) # True
"10.0.0" > "9" # False
```

I notice that [PEP 396](https://www.python.org/dev/peps/pep-0396/#specification) does specify that `__version__` _should_ be a string, but the elasticsearch library can't change that now without breaking any existing checks.

h/search/client.py
```diff
@@ -50,6 +51,10 @@ def mapping_type(self):
         """
         return self._mapping_type
 
+    @property
+    def version(self):
```
robertknight: This could use a docstring to indicate that it refers to the Elasticsearch _library_ version, as opposed to the server version.

h/search/client.py
```diff
@@ -12,14 +12,15 @@ class Client(object):
     """
     A convenience wrapper around a connection to Elasticsearch.
 
-    Holds a connection object, an index name, and the name of the mapping type.
+    Holds a connection object, an index name, es version, and the name of the mapping type.
```
robertknight: We should clarify here that this means es _library_ version.

h/search/client.py
robertknight: If you do make this change, then code switching on the ES version can just do:

```py
if client.version >= (6,):
  # ES 6 code path
else:
  # Older code path
```

PR #5150, opened 2018-07-16
Adapt filter query syntax depending on Elasticsearch version

robertknight:
The syntax for executing queries in a _filter context_ changed in
Elasticsearch 2.x. See [1]

Two pure unit tests have been added for the `Builder` class which check
the JSON output of `build(params)`. Once the tests which actually
execute queries against ES are working with ES 6, these will be
obsoleted.

[1] https://www.elastic.co/guide/en/elasticsearch/reference/6.2/query-dsl-filtered-query.html


h/search/query.py
```diff
@@ -46,12 +54,25 @@ def build(self, params):
         for key, value in params.items():
             matchers.append({"match": {key: value}})
 
-        query = {"match_all": {}}
+        bool_query = {}
 
+        # Add clauses executing in a query (scoring) context.
         if matchers:
-            query = {"bool": {"must": matchers}}
+            bool_query["must"] = matchers
```
hmstepanek: I understand what this is doing now but I found all these if else statements hard to follow. Could we simplify these/re-org them in a way that makes it more readable? Alternatively it might be better just to have a separate builder for es1 and es6?

robertknight:
After a bit of experimentation I discovered `bool` queries are allowed to have empty `filter` and `matcher` arrays, in which case they return all documents. That made it possible to simplify the logic a bit. PTAL at the revised [diff ignoring whitespace](https://github.com/hypothesis/h/pull/5150/files?utf8=✓&diff=unified&w=1).

PR #5153, opened 2018-07-16
Rename `NipsaService.flagged_userids`

robertknight:
Change `flagged_userids` from a property to a method and add a
`fetch_all_` prefix to make it more explicit that this code does a
potentially expensive database query, depending on the number of users
in the system. This is also more consistent with the API of eg. `UserService`.

The property => method change will also make it a little easier to mock
in an upcoming PR.

### DIFF
--- a/h/services/nipsa.py
+++ b/h/services/nipsa.py
@@ -15,10 +15,9 @@ class NipsaService(object):
     def __init__(self, session):
         self.session = session
 
-    @property
-    def flagged_userids(self):
+    def fetch_all_flagged_userids(self):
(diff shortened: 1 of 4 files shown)


robertknight:
> Is there a good reason not to index nipsa in the db?

I think we probably should index `user.nipsa`. The choices around whether to fetch the flagged state for a single user vs. fetch the set of all flagged users in a given request/task would remain the same though.

hmstepanek:
In that case, can you open another PR to add the indexing of nipsa as well?

Issue #5151, opened 2018-07-16
Investigate reindexing performance regression


robertknight:
Search reindexing performance in production appears to be significantly slower than it was previously. I [recently ran a reindexing job](https://jenkins.hypothes.is/job/h-tasks/77/console) that averaged ~220 annotations/second. When the last reindexing job ran at the end of January the overall reindexing rate was about ~500/second.

This is possibly related to https://github.com/hypothesis/h/pull/5129 which removed a NIPSA lookup cache.

In addition to the regression, monitoring performance of manually initiated tasks is difficult because:

1. Logs from the task are not reported _until the task has finished running_. This means that progress log statements are not visible while the task runs. While the task is running `eb-task-run` just shows `---> running command`
2. The `eb-task-run` script does not report _which_ instance it is running the task on, which makes monitoring the instance directly harder.

robertknight:
Log outputs from a recent run: https://jenkins.hypothes.is/job/h-tasks/77/console

PR #5154, opened 2018-07-16
Speed up reindexing by re-adding NIPSA flag cache

robertknight:
**Depends on https://github.com/hypothesis/h/pull/5153**

This PR should resolve a reindexing performance regression (~500 anns/sec => ~200 anns/sec in production) caused by the removal of the NIPSA cache in https://github.com/hypothesis/h/pull/5129.

The removal of the cache speeded up WebSocket message processing and indexing single annotations by removing an expensive scan of the `User` table and replacing it with an indexed lookup of a single user. However it regressed reindexing of all annotations because the one-at-a-time lookup of users by userid was less efficient than the previous bulk-loading of flagged userids into a set, in that context.

To optimize for both uses I reintroduced the cache, but it is only populated when `fetch_all_flagged_userids` is called explicitly, which the `hypothesis search reindex` CLI command now does.

On my laptop this improves reindexing speed with a small database of ~2.5K annotations from ~300/sec to ~2000/sec. In production I'm only hoping to get reindexing performance back to where it was previously.

Hopefully fixes https://github.com/hypothesis/h/issues/5151

### DIFF
--- a/h/services/nipsa.py
+++ b/h/services/nipsa.py
@@ -15,17 +15,38 @@ class NipsaService(object):
     def __init__(self, session):
         self.session = session
 
+        # Cache of all userids which have been flagged.
+        self._flagged_userids = None
+
     def fetch_all_flagged_userids(self):
         """
         Fetch the userids of all shadowbanned / NIPSA'd users.
 
+        The set of userids is cached to speed up subsequent `flagged_userids`
+        and `is_flagged` calls in the same request.
+
         :rtype: set of unicode strings
         """
+        if self._flagged_userids is not None:
+            return self._flagged_userids
+
+        # TODO: The `nipsa` field is currently not indexed so this requires a
+        # full table scan.
         query = self.session.query(User).filter_by(nipsa=True)
-        return set([u.userid for u in query])
+        self._flagged_userids = set([u.userid for u in query])
+
+        return self._flagged_userids
 
     def is_flagged(self, userid):
         """Return whether the given userid is flagged as "NIPSA"."""
+
+        # Use the cache if populated.
+        if self._flagged_userids is not None:
+            return userid in self._flagged_userids
+
+        # Otherwise lookup the status for a single user, which is more efficient
+        # than populating the cache if we are only looking up the NIPSA status
+        # for a small number of users in a given task/request.
         user = self.session.query(User).filter_by(userid=userid).one_or_none()
         return user and user.nipsa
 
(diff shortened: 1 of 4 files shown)


h/services/nipsa.py
```diff
@@ -15,18 +15,38 @@ class NipsaService(object):
...
-        return set([u.userid for u in query])
+        self._flagged_userids = set([u.userid for u in query])
+
+        return self._flagged_userids
 
     def is_flagged(self, userid):
         """Return whether the given userid is flagged as "NIPSA"."""
+
+        # Use the cache if populated.
+        if self._flagged_userids is not None:
```
robertknight: The `is not None` part is needed here because `self._flagged_userids` may have been populated, but may still be an empty set if no users have been shadowbanned.

hmstepanek: I had a teammate in the past who was adamant about being explicit with the is None's and who knows maybe it ended up saving us from cases like this!

h/services/nipsa.py
```diff
@@ -49,8 +72,14 @@ def unflag(self, user):
         queue).
         """
         user.nipsa = False
+        if self._flagged_userids is not None:
```
hmstepanek: Would you ever call unflag on a user that is shadowbanded where the _flagged_userids hasn't been populated yet? If so I think we have a bug here.

tests/h/services/nipsa_test.py
```diff
@@ -51,6 +52,7 @@ def test_unflag_sets_nipsa_false(self, db_session, users):
 
         svc.unflag(users['renata'])
 
+        assert not svc.is_flagged('acct:renata@example.com')
```
hmstepanek: The way these tests are implemented don't necessarily guarantee that the internal cache got updated. They only black box verify that the user according to the service is flagged or unflagged as expected. For example you could remove the cache logic route from the is_flagged and remove the update cache logic from unflag and this test would still pass despite the cache not getting updated with the unflagged user. The alternative is a bit unpleasant tho as you would have to assert on a private variable. Thoughts?

hmstepanek: The way these tests are implemented don't necessarily guarantee that the internal cache was updated. They only black box verify that the user according to the service is flagged or unflagged as expected. For example you could remove the cache logic route from the is_flagged and remove the cache update logic from unflag and this test would still pass despite the cache not getting updated with the unflagged user. The alternative is a bit unpleasant tho as you would have to assert on a private variable. Thoughts?

robertknight: I added two specific tests below for interaction of `flag` and `unflag` with the cache to provide "reasonable" coverage of this.

Of course it would still be possible to come up with _some implementation_ which passed the tests and had full branch coverage but was broken in some way. That's just a limitation of unit testing. If the cache logic became super-complex it could be factored into its own thing, but I don't see a need yet.

robertknight:
I added one more assert in the `test_fetch_all_flagged_userids_caches_lookup` test to catch a branch that codecov noted I missed.

PR #5156, opened 2018-07-16
Display empty string for users with no email address in /account/settings.

sheetaluk:
See https://github.com/hypothesis/product-backlog/issues/705

### DIFF
--- a/h/views/accounts.py
+++ b/h/views/accounts.py
@@ -435,7 +435,7 @@ class AccountController(object):
 
     def _template_data(self):
         """Return the data needed to render accounts.html.jinja2."""
-        email = self.request.user.email
+        email = self.request.user.email or ''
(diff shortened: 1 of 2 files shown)


lyzadanger:
@sheetaluk I haven't looked at all at the details of this view or the services is uses, but...

Have you tested to see what happens when this (email) form is _submitted_ with email as an empty string?

Does it error out?
Does it update the user's email address to an empty string (that would be bad, BTW!)?
Can you even submit the form if the email address is blank?

sheetaluk:
@lyzadanger I can't submit this form without a valid email:
image
image
Also, the email for the `get` method is coming from the `request.user.email` object which I'm setting to `''`only if it is `None`. It is being modified from a different block of code for the `post` method,so it shouldn't update the email to an empty string.

Issue #5137, opened 2018-07-11
Remove organization logo character limit


seanh:
h currently enforces the organization logo files be 10,000 characters or less. This limit is problematic and should be removed.

We should just remove `ORGANIZATION_LOGO_MAX_CHARS` entirely, it was put there to prevent maliciously large uploads but we've no need to worry about those because only Hypothesis admins can add logos.

In the future, if we have self-serve organization creation, we may need validation of organization logos. But that feature isn't on our immediate roadmap. We may never implement it. If / when we do, we'll have to implement good SVG validation with realistic restrictions and good, helpful error messages. The current code doesn't achieve that - it gives an incorrect error message when the file is too long, and rejects the kinds of SVGs that our partners want to use as invalid. Let's keep it simple and not try to solve this validation problem prematurely.

## Incompatible with SVG files we receive from partners

This is causing us a problem because we don't create the organization logo SVGs ourselves, we receive them from partners. Partners don't always provide clean and short SVG files and we're having trouble getting clean and short ones from all of them - different partners have their own ways of creating SVG files. For an example of the problem, see <https://stackoverflow.com/c/hypothesis/questions/77/81>.

## Incorrect error message

Additionally the Hypothesis admin is shown the error message "Logo is not valid SVG" when they try to enter SVG longer than 10,000 characters. This error message is incorrect - it is valid for an SVG file to be longer than 10,000 characters. The 10,000 character limit is an arbitrary limit applied by us.

The model-level validation actually [does give specific error messages for specific problems](https://github.com/hypothesis/h/blob/a1275c9c5eca0c85cb4d1d80abbaffe11f46a08c/h/models/organization.py#L50-L66) with the SVG, including "logo must be less than 10000 characters long", but the form validation [swallows the specific error messages](https://github.com/hypothesis/h/blob/a1275c9c5eca0c85cb4d1d80abbaffe11f46a08c/h/schemas/forms/admin/organization.py#L14-L18) and turns them all into the generic and not-always-correct "Logo is not valid SVG".

## P.S. Avoid complex validation in models

As an aside, and going beyond the organization logo max length issue, I think the complex `validate_logo()` method on the `Organization` model (validates the logo's length, that it parses with `ElementTree`, that its root tag is `svg`, and gives different error messages for different failures) should be removed as it's not needed (only Hypothesis admins can add logos), or failing that should at least be moved into `h/schemas` where validation is supposed to be done in h.

[...]

PR #5167, opened 2018-07-18
Increase svg size limit & move validation out of model

hmstepanek:
1.  fix the too large error message as it's wrong (just accidentally used the wrong constant)
1.  increase the allowed size by an order of magnitude
1.  move svg validation checking out of model, pass the error message to the client admin page so the user actually knows why the svg is invalid, & update the tests to check for the message part of the error not just the error type

This is a fix for https://github.com/hypothesis/h/issues/5137

The following should pass:
https://drive.google.com/file/d/15FW_74UKBaLcQxJgjiwS7fiOw4uyfDBf/view?usp=sharing
https://drive.google.com/file/d/1q086H_R1YWns3EpkXF7X6AUs7y0A5PGk/view?usp=sharing

The following should fail:
https://drive.google.com/file/d/1gh0JehWI4bUruyDyN-u2bR0TAEo5tbEq/view?usp=sharing


h/schemas/forms/admin/organization.py
```diff
@@ -2,21 +2,38 @@
...
+
+
+def _strip_xmlns(tag):
+    if '}' in tag:
+        return tag.split('}', 1)[1]
+    return tag
+
+
 def validate_logo(node, value):
+    if not (len(value) <= ORGANIZATION_LOGO_MAX_CHARS):
```
hmstepanek: I removed the if value is None here as it looks like this is never entered if a logo is left blank or removed on the form.

tests/h/schemas/forms/admin/organization_test.py
```diff
@@ -3,25 +3,43 @@
...
         bound_schema.deserialize(org_data)
 
-    @pytest.mark.parametrize('invalid_logo', [
-        'not an svg',
-        '<svg>{}</svg>'.format('a' * 20000)  # Too long
-    ])
-    def test_it_raises_if_logo_is_invalid(self, org_data, bound_schema, invalid_logo):
-        org_data['logo'] = invalid_logo
+    def test_it_raises_if_logo_is_too_long(self, org_data, bound_schema):
+        org_data['logo'] = '<svg xmlns="http://svg.com">{}</svg>'.format('a' * ORGANIZATION_LOGO_MAX_CHARS + 'b')
```
hmstepanek: I added the xmlns for test coverage purposes.

h/schemas/forms/admin/organization.py
```diff
@@ -2,21 +2,38 @@
...
+        raise colander.Invalid(node, _(
+            "Logo is larger than {} characters".format(ORGANIZATION_LOGO_MAX_CHARS)))
```
hmstepanek: I chose not to make exception classes for these after all. I still can if you feel strongly. Since I don't anticipate the exception messages changing and they are only used here and in the tests I didn't see a strong need to abstract it.

seanh:
@hmstepanek Could you add some example SVG files to this PR, to make it easier to test? Ones that should validate successfully, and ones that should fail validation with different error messages.

hmstepanek:
@seanh I updated the description with some svg's provided by Arti that should fail and pass.

h/schemas/forms/admin/organization.py
```diff
@@ -2,21 +2,38 @@
...
 from h import validators
-from h.models.organization import Organization, ORGANIZATION_NAME_MIN_CHARS, ORGANIZATION_NAME_MAX_CHARS
 from h.schemas.base import CSRFSchema
 
 _ = h.i18n.TranslationString
 
 
+ORGANIZATION_NAME_MIN_CHARS = 1
+ORGANIZATION_NAME_MAX_CHARS = 25
+ORGANIZATION_LOGO_MAX_CHARS = 100000
```
robertknight: Putting the field length constraints here introduces an inconsistency compared to `h.models.user` and `h.models.group`. The pattern we have ended up with for those other models is:

- The validation is actually implemented in three places: Once in the schema for colander forms, once in the JSON schema for the API and once in the domain model itself. The duplication is perhaps not ideal but not causing us that much pain either since it is quite simple.
- The values (min/max field lengths etc.) are at least defined in one central place in the `h.models.{group, user}` package.

Unless we think the way it works for `Group` and `User` is broken and we intend to change it, I would be inclined to follow the same pattern and leave the constants in `h.models.organization`. In this case we only have a form rather than an API _currently_ so we only need to use the values in this module. Thoughts @seanh ?

seanh: I think you need to ask whether the min and max chars is a simple and universal constraint or not.

I think the general principle is that if a piece of validation is both really simple and universal (definitely going to want to apply _whenever_ we add anything to this DB table) then it belongs in the models. Examples: foreign key constraints, not null constraints, etc. Otherwise if it's more complicated or non-universal (e.g. different validation for API vs views, or for one type of user vs another, ...) then it belongs in `h.schemas` (which is planned to be renamed to `h.validation`) because (for several reasons) we try to keep our model classes really thin and simple.

Most of the time these really simple validations can be implemented as actual database constraints, but sometimes they can't be or its for some reason undesirable to implement that at Postgres-level, in those cases we can use `@sa.orm.validates()` methods on model classes. Uses of `@sa.orm.validates()` should be rare in our code though.

For string lengths we use `@sa.orm.validates()` when we put them in the models. We actually went through all the DB tables and replaced DB-level max lengths with `@sa.orm.validates()`. See [No length limits on database columns](http://h.readthedocs.io/en/latest/developing/making-changes-to-model-code/#no-length-limits-on-database-columns).

... So that's why things like `USERNAME_MAX_LENGTH` are in the models. And then of course the schemas need to give a friendly error message when something is too long, and such human-readable error message belong in the schemas or views not in the models, so that's why schemas need to import `USERNAME_MAX_LENGTH`.

h/schemas/forms/admin/organization.py
hmstepanek: Agree!

h/schemas/forms/admin/organization.py
hmstepanek: Sounds like they need to go back in the model then-will do!

hmstepanek: It does seem weird to me that the min and max chars check happens both in the schema and the model and it's not re-used. Ideally it should happen once in the most externally facing place to fail quickly.

h/schemas/forms/admin/organization.py
```diff
@@ -2,20 +2,39 @@
...
+    if len(value) > ORGANIZATION_LOGO_MAX_CHARS:
+        raise colander.Invalid(node, _(
+            "Logo is larger than {:,d} characters".format(ORGANIZATION_LOGO_MAX_CHARS)))
     try:
-        Organization(logo=value)
-    except Exception:
-        raise colander.Invalid(node, _('Logo is not valid SVG'))
+        root = ElementTree.fromstring(value)
+    except ElementTree.ParseError:
+        raise colander.Invalid(node, _('Logo is not a valid SVG'))
```
robertknight: This message has lost a useful piece of information from before which is that the XML couldn't be parsed, as opposed to other reasons why the content might not be a valid SVG.

hmstepanek: done

Issue #5168, opened 2018-07-18
Optimize search reindexing task


robertknight:
As the size of our production DB grows, so does the amount of time taken to reindex annotations after making changes to the schema or data that is indexed in Elasticsearch. I recently had to [bump the timeout](https://github.com/hypothesis/deployment/pull/24) as it exceeded the 2-hour threshold.

The current reindexing rate on an `m4.large` h production app instance is about ~430 annotations/sec. Elasticsearch reported a mean indexing time of ~0.5s / document for a recent reindexing operation, which suggests that most of the overhead is elsewhere.

----

**Implementation notes**

Some thoughts on ways we could improve the situation:

1. Parallelize reindexing. We have multiple h app instances running but run the entire reindexing task on a single instance. There are some setup and finalization steps which really need to be done on a single instance, but most of the work could be spread across instances, probably using the Celery task queue and workers that we already have in place.
2. Overlap database reads and Elasticsearch writes. The current task loads a bunch of annotations from the DB, then "presents" them into a dict for indexing, then makes a query to Elasticsearch to index them. It would be more efficient if DB reads for the _next_ batch were overlapped with the ES query to write the current batch.
3. Profile reindexing in production and optimize whatever is slow.

My current thoughts are that we could do something like this:

1. A "master" indexing CLI command (`hypothesis search reindex`) creates the new index.
2. The master command loads batches of annotation IDs from the DB and dispatches Celery tasks to index those annotations, probably with some upper limit on the number of outstanding tasks.
3. Once tasks have been dispatched for all annotation IDs, and once all of those tasks have been completed [1], the master indexing task finalizes the reindex operation.

[1] AFAIK we don't currently have any code in h that executes a task and then waits for it to finish. However Celery does support returning results for certain backends, including RabbitMQ. The other option is to write acknowledgements to the DB and poll them from the master task.

hmstepanek:
Possible cause of this slowdown:
https://github.com/hypothesis/product-backlog/issues/836
I think we should address ^ then re-evaluate performance.

seanh:
This reindexing task got completely replaced by the new periodic sync task, closing this

PR #5170, opened 2018-07-19
Add flag to parallelize search reindexing task

robertknight:
Rebuilding the search index after a change to the schema or indexed data currently involves a series of sequential steps which all happen on a single production h instance. The time taken scales linearly with the size of the DB, currently just over a couple of hours for the production DB.

This PR adds an experimental `--parallel` option to the `hypothesis search reindex` command which reduces the total time taken by farming out the core reindexing work (loading annotations from the DB, generating JSON dicts and making indexing requests to Elasticsearch) to a set of Celery tasks which can be executed in parallel by as many worker processes as we have running in production.

There are a couple of parameters, currently just hard-coded values in the code, which affect the amount of parallelization that happens - `batch_size` specifies how many annotations to index per Celery task and `max_active_tasks` controls how many indexing tasks may be in-flight at any time.

See the commit message for a detailed explanation of the changes.

Fixes https://github.com/hypothesis/h/issues/5168

### DIFF
--- a/h/celery.py
+++ b/h/celery.py
@@ -28,7 +28,7 @@ __all__ = (
 
 log = logging.getLogger(__name__)
 
-celery = Celery('h')
+celery = Celery('h', backend='rpc://')
 celery.conf.update(
     broker_url=os.environ.get('CELERY_BROKER_URL',
                               os.environ.get('BROKER_URL', 'amqp://guest:guest@localhost:5672//')),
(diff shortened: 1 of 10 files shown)


robertknight:
This was an interesting experiment that shows it can work. However the priority isn't that high for us at the moment because reindexing doesn't take that long (~2 hours) in production. I'm going to close this for now.

PR #5173, opened 2018-07-20
Add elasticsearch1 monitoring to New Relic.

a-feld:
CC: @hmstepanek 

Hi there! I noticed that both the `elasticsearch1` library and New Relic monitoring is in use in this project.  Since `elasticsearch1` isn't a New Relic officially supported library, elasticsearch queries that are made through `elasticsearch1` may fail to appear in the New Relic user interface.

Since `elasticsearch1` appears to be a fork of the `elasticsearch` library, the New Relic agent can be configured to load all of the `elasticsearch` instrumentation when the `elasticsearch1` package is loaded.

Summary of the changes:
* Initializing the agent normally sets up "import hooks" for elasticsearch as defined [here](https://github.com/edmorley/newrelic-python-agent/blob/da8e13e92af5c428cfc0da861bae78e2e149feb6/newrelic/newrelic/config.py#L2420L2449)
* A configuration file may also be used to define import hooks in the form of:
```
[import-hook:%s]
enabled = true
execute = module:function
```
* The builtin instrumentation can be loaded by replacing `elasticsearch` with `elasticsearch1` so that the import-hook fires on import of `elasticsearch1`!

The supervisor config / app config has been updated to load these modules. All existing New Relic environment variables should continue to work!

I hope this change helps the team, please let me know if I can provide further assistance!

### DIFF
--- a/conf/app.ini
+++ b/conf/app.ini
@@ -69,3 +69,35 @@ formatter = generic
 
 [formatter_generic]
 format = %(asctime)s [%(process)d] [%(name)s:%(levelname)s] %(message)s
+
+# Add temporary newrelic hooks for elasticsearch1
+[import-hook:elasticsearch1.client]
+enabled = true
+execute = newrelic.hooks.datastore_elasticsearch:instrument_elasticsearch_client
+[import-hook:elasticsearch1.client.cat]
+enabled = true
+execute = newrelic.hooks.datastore_elasticsearch:instrument_elasticsearch_client_cat
+[import-hook:elasticsearch1.client.cluster]
+enabled = true
+execute = newrelic.hooks.datastore_elasticsearch:instrument_elasticsearch_client_cluster
+[import-hook:elasticsearch1.client.indices]
+enabled = true
+execute = newrelic.hooks.datastore_elasticsearch:instrument_elasticsearch_client_indices
+[import-hook:elasticsearch1.client.nodes]
+enabled = true
+execute = newrelic.hooks.datastore_elasticsearch:instrument_elasticsearch_client_nodes
+[import-hook:elasticsearch1.client.snapshot]
+enabled = true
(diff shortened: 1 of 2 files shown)


a-feld:
> Can you see any possibility of this causing unexpected side effects in the NR Python agent?

Excellent question! The New Relic agent will only process the following sections of a configuration file:
* `[newrelic]`
* `[newrelic:%s]`
* `[import-hook:%s]`
* `[wsgi-application:%s]`
* `[background-task:%s]`
* `[database-trace:%s]`
* `[external-trace:%s]`
* `[function-trace:%s]`
* `[generator-trace:%s]`
* `[profile-trace:%s]`
* `[memcache-trace:%s]`
* `[transaction-name:%s]` or `[name-transaction:%s]`
* `[error-trace:%s]`
* `[data-source:%s]`
* `[function-profile:%s]`

The configuration loader code is defined [here](https://github.com/edmorley/newrelic-python-agent/blob/da8e13e92af5c428cfc0da861bae78e2e149feb6/newrelic/newrelic/config.py#L838) for reference.

Based on the list of supported sections, I don't foresee any possibility of interference of this configuration update on either `gunicorn` or the New Relic agent since these sections do not overlap any supported paste deploy sections (from [paste deploy](https://pastedeploy.readthedocs.io/en/latest/) the only sections utilized have prefixes, like app:main or filter:errors or are part of the logging sections) but I can definitely understand the concern. My motivation for placing this into an existing config file was to avoid adding additional configuration files to the repository.

I'd be happy to move these configurations to a separate `newrelic.ini` if you'd prefer!

robertknight:
Thanks for looking into this @a-feld. On that basis I think it should be fine to keep this config in `conf/app.ini` for now. I'll leave it to @hmstepanek to get this tested and deployed.

PR #5175, opened 2018-07-20
Add `--es6` flag to `search reindex` CLI command

robertknight:
Add a flag which causes the `search reindex` command to reindex annotations
into the Elasticsearch 6 cluster instead of the ES 1 cluster.

Most of the actual work was done in the earlier indexing PRs. All I had to do here was add a flag to the CLI command and pass the appropriate ES client to `h.indexer.reindexer.reindex`

Fixes https://github.com/hypothesis/product-backlog/issues/605

### DIFF
--- a/h/cli/commands/search.py
+++ b/h/cli/commands/search.py
@@ -14,16 +14,32 @@ def search():
 
 
 @search.command()
+@click.option('--es6', is_flag=True, help='Reindex into the Elasticsearch 6 cluster')
 @click.pass_context
-def reindex(ctx):
+def reindex(ctx, es6):
     """
-    Reindex all annotations in all clusters.
+    Reindex all annotations.
 
     Creates a new search index from the data in PostgreSQL and atomically
     updates the index alias. This requires that the index is aliased already,
     and will raise an error if it is not.
+
+    Reindex into the Elasticsearch 1 cluster by default, unless the `--es6`
+    flag is set.
     """
-    _reindex_old(ctx)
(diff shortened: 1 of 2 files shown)


tests/h/cli/commands/search_test.py
```diff
@@ -49,5 +58,6 @@ def update_index_settings(self, patch):
 
 @pytest.fixture
 def cliconfig(pyramid_request):
-    pyramid_request.es = mock.sentinel.es
+    pyramid_request.es = mock.create_autospec(Client, spec_set=True, instance=True)
+    pyramid_request.es6 = mock.create_autospec(Client, spec_set=True, instance=True)
```
robertknight: `create_autospec` generates a more faithful mock of the `Client` interface.

h/cli/commands/search.py
```diff
@@ -39,22 +55,6 @@ def update_settings(ctx):
     _update_settings_old(ctx)
 
 
-def _reindex_old(ctx):
```
robertknight: I moved the content of this function into the `reindex` function above since the same code works for both ES versions for the most part.

robertknight:
> For example could we not just rename the es1 command to _old or _es1 so that we can simply delete it during cleanup rather than adding a feature flag and then removing it?

I did copy and paste the reindex command initially but then I realized that I wanted to make another change for both commands, logging the version of the ES cluster being written to for verification when running the job in Jenkins, and so I figured it would be easier to do this with one command and a flag.

PR #5176, opened 2018-07-21
Add search_es6 feature flag and run search tests against es6

hmstepanek:
1. Add a feature flag called search_es6 that will search the elasticsearch6 cluster instead of the elasticsearch1 cluster.
1. If the search_es6 feature flag is set search the es6 cluster instead of es1.
1. Update the search tests to verify both es6 searching and es1 search behave the same.
1. Update query tests (filters, aggregators, matchers) to verify both es6 and es1 behave the same.

This is a partial fix for https://github.com/hypothesis/product-backlog/issues/601.

### DIFF
--- a/h/models/feature.py
+++ b/h/models/feature.py
@@ -20,8 +20,7 @@ FEATURES = {
     'client_display_names': "Render display names instead of user names in the client",
     'index_es6': ("Index annotations into Elasticsearch 6 "
                   "(only takes effect if enabled for everyone)?"),
-    'search_es6': ("Search annotations in Elasticsearch 6 "
-                   "(only takes effect if enabled for everyone)?"),
+    'search_es6': "Search annotations in Elasticsearch 6",
 }
 
 # Once a feature has been fully deployed, we remove the flag from the codebase.
--- a/h/models/feature.py
+++ b/h/models/feature.py
@@ -20,6 +20,8 @@ FEATURES = {
     'client_display_names': "Render display names instead of user names in the client",
     'index_es6': ("Index annotations into Elasticsearch 6 "
                   "(only takes effect if enabled for everyone)?"),
+    'search_es6': ("Search annotations in Elasticsearch 6 "
+                   "(only takes effect if enabled for everyone)?"),
 }
 
 # Once a feature has been fully deployed, we remove the flag from the codebase.
--- a/h/search/core.py
+++ b/h/search/core.py
@@ -24,7 +24,7 @@ class Search(object):
     :param request: the request object
     :type request: pyramid.request.Request
 
-    :param separate_replies: Wheter or not to return all replies to the
+    :param separate_replies: Whether or not to return all replies to the
         annotations returned by this search. If this is True then the
         resulting annotations will only include top-level annotations, not replies.
     :type separate_replies: bool
@@ -35,7 +35,10 @@ class Search(object):
     """
     def __init__(self, request, separate_replies=False, stats=None, _replies_limit=200):
         self.request = request
-        self.es = request.es
+        if self.request.feature('search_es6'):
+            self.es = request.es6
+        else:
+            self.es = request.es
         self.separate_replies = separate_replies
         self.stats = stats
         self._replies_limit = _replies_limit
--- a/tests/h/search/core_test.py
+++ b/tests/h/search/core_test.py
@@ -10,7 +10,6 @@ Tests for filtering/matching/aggregating on specific annotation fields are in
 from __future__ import unicode_literals
 
 import datetime
-
 import pytest
 
 from h import search
@@ -126,6 +125,8 @@ class TestSearch(object):
 
     def test_it_passes_es_version_to_builder(self, pyramid_request, Builder):
         client = pyramid_request.es
+        if pyramid_request.feature('search_es6'):
+            client = pyramid_request.es6
 
         search.Search(pyramid_request)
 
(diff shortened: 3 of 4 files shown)


hmstepanek:
^ code cov is just complaining that we don't actually call any other feature flags besides search_es6 in the tests so return True never is reached in the side_effect of request.feature which I think is perfectly ok.

tests/h/search/core_test.py
```diff
@@ -243,3 +245,17 @@ def test_only_200_replies_are_included(self, pyramid_request, Annotation):
 
         assert len(result.reply_ids) == 3
         assert oldest_reply.id not in result.reply_ids
+
+
+@pytest.fixture(params=[True, False])
```
robertknight: Since a string representation of the param value is used in generated test names, I think it would be easier to understand what is happening when looking at the `--verbose` output or any test failures if the params were string labels (eg. `es1`, `es6`) rather than booleans.

As it is, the output currently looks like this:

```
tests/h/search/query_test.py::TestUsersAggregation::test_it_returns_annotation_counts_by_user[True] PASSED                    [ 96%]
tests/h/search/query_test.py::TestUsersAggregation::test_it_returns_annotation_counts_by_user[False] PASSED                   [ 97%]
tests/h/search/query_test.py::TestUsersAggregation::test_it_limits_number_of_annotation_counts_by_user_returned[True] PASSED  [ 98%]
tests/h/search/query_test.py::TestUsersAggregation::test_it_limits_number_of_annotation_counts_by_user_returned[False] PASSED [100%]
```

tests/h/search/query_test.py
```diff
@@ -24,7 +24,7 @@ class TestBuilder(object):
         # Default sort field should be "updated".
         (None, "asc", [2, 0, 1]),
     ])
-    def test_it_sorts_annotations(self, Annotation, search, sort_key, order, expected_order):
+    def test_it_sorts_annotations(self, Annotation, search, sort_key, order, expected_order, pyramid_request):
```
robertknight: I think would be preferable to use `pytest.mark.usefixtures` here to use the `pyramid_request` fixture rather than adding it as a parameter, since that makes it clearer that the fixture is being used for its side effects.

_However_ you don't actually need to explicitly this fixture in any tests that already use `search` because the `search` fixture is already parametrized by `pyramid_request`.

tests/h/search/query_test.py
```diff
@@ -74,6 +74,7 @@ def test_it_uses_new_filter_syntax_for_es6(self):
                                   "must": []}}
 
 
+@pytest.mark.usefixtures('pyramid_request')
```
robertknight: This isn't actually necessary because the `search` fixture is already parametrized so any tests which use it will automatically run with both ES versions.

hmstepanek:
Ok stopped explicitly specifying pyramid_request as a fixture, changed to using es1 and es6 params, and I also made a quick fix where we were using a python keyword in a couple tests.

tests/h/search/core_test.py
```diff
@@ -243,3 +245,17 @@ def test_only_200_replies_are_included(self, pyramid_request, Annotation):
...
+def pyramid_request(request, pyramid_request, es_client, es6_client):
+    pyramid_request.es = es_client
+    pyramid_request.es6 = es6_client
+
+    def feature(flag):
+        if flag == 'search_es6':
+            return request.param == 'es6'
+        return True
+
+    pyramid_request.feature = mock.Mock(side_effect=feature)
```
robertknight: I forgot this when I last looked at this PR, but there is a shortcut for setting feature flags in tests:

```py
pyramid_request.feature.flags['flag_name'] = flag_state
```

tests/h/search/query_test.py
```diff
@@ -653,6 +653,20 @@ def test_it_limits_number_of_annotation_counts_by_user_returned(self, Annotation
...
+def pyramid_request(request, pyramid_request, es_client, es6_client):
+    pyramid_request.es = es_client
+    pyramid_request.es6 = es6_client
+
+    def feature(flag):
+        if flag == 'search_es6':
+            return request.param == 'es6'
+        return True
+
+    pyramid_request.feature = mock.Mock(side_effect=feature)
```
robertknight: This can be simplified a little using `pyramid_request.feature.flags`. See comment above.

tests/h/search/query_test.py
```diff
@@ -50,27 +50,27 @@ def test_it_uses_old_filter_syntax_for_es1(self):
         # TODO - Remove this test once ES 1 support is dropped.
 
         builder = Builder(es_version=(1, 7, 0))
-        filter = AuthorityFilter("default")
+        filter_ = AuthorityFilter("default")
```
robertknight: `filter` is a builtin rather than a reserved keyword (see https://docs.python.org/3/reference/lexical_analysis.html#keywords), so using it is an identifier is legal. I know that some syntax plugins for popular editors colorize builtins which could confuse the reader, so I'm OK with this change. Could you update the commit message though to describe what `filter` is more accurately?

h/models/feature.py
```diff
@@ -20,6 +20,8 @@
     'client_display_names': "Render display names instead of user names in the client",
     'index_es6': ("Index annotations into Elasticsearch 6 "
                   "(only takes effect if enabled for everyone)?"),
+    'search_es6': ("Search annotations in Elasticsearch 6 "
+                   "(only takes effect if enabled for everyone)?"),
```
robertknight: The "only takes effect if enable for everyone" part is not true for the `search_es6` feature flag. The reason it applies for `index_es6` is because that feature flag is used in the context of a Celery worker process where there is never a logged-in user (ie. `celery.request.authenticated_user` is always `None` AFAIK).

h/search/core.py
```diff
@@ -35,7 +35,10 @@ class Search(object):
     """
     def __init__(self, request, separate_replies=False, stats=None, _replies_limit=200):
         self.request = request
-        self.es = request.es
+        if self.request.feature('search_es6'):
+            self.es = request.es6
+        else:
+            self.es = request.es
```
robertknight: This change is the most significant part of this commit, to actually use the `search_es6` feature flag, but it isn't mentioned in the commit message.

PR #5177, opened 2018-07-21
Test es6 in reindexer tests

hmstepanek:
Test elasticsearch6 reindexing by mocking the elastic search 6 client object.


robertknight:
Something which confused me when I looked at this PR initially was that this branch is called `test-es6-batch-indexer` but the changes are to code which _uses_ the batch indexer rather than the batch indexer itself. Just to clarify a comment from our sync call at the end of last week, the main ES-using tests which need to be adapted to work with ES 6 is `TestBatchIndexer` in `tests/h/search/index_test.py`.

PR #5178, opened 2018-07-23
Ignore `ELASTICSEARCH_CLIENT_TIMEOUT` env var in `hypothesis init` command

robertknight:
The request timeout is typically set to a short value (<1s) in
production environments to prevent long-running search/indexing
commands.

The process of initializing the search index can take longer than this
though, so override the default timeout using the same method that is
used for the `hypothesis search reindex` command.

----

I ran into this issue when running `hypothesis init` in production earlier today to initialize the ES 6 cluster. There are plenty of other ways that this could be fixed, but I've opted to use the same approach as the `reindex` command for now.

### DIFF
--- a/h/cli/commands/init.py
+++ b/h/cli/commands/init.py
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 
 import logging
+import os
 
 import click
 import sqlalchemy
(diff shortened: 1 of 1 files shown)


hmstepanek:
^ they look good now.

